PR #9571, opened 2026-07-12
fix(memory): use a tz-aware sentinel in the review-queue sort key to avoid a 500

ZachL111:
## Problem
`_review_conflict_sort_key` returned `(impact, cast(datetime, item.get('created_at') or datetime.min))`. `datetime.min` is naive, but stored `created_at` is tz-aware (`datetime.now(timezone.utc)`). On an impact tie, `list_review_conflicts`' `items.sort(key=_review_conflict_sort_key, reverse=True)` compares the two datetimes and raises `TypeError: can't compare offset-naive and offset-aware datetimes` -> HTTP 500 on `GET /v3/memories/review-queue`. Impact ties are common (defaults/fallbacks), so an item missing `created_at` triggers it.

## Fix
Use the tz-aware sentinel `datetime.min.replace(tzinfo=timezone.utc)`, exactly as `database/memory_ledger.py` already does for the same kind of sort.

## Test
`tests/unit/test_review_queue_sort_tz.py` sorts items with one tz-aware `created_at` and one missing (an impact tie) without raising, and confirms the ordering (impact desc, then created_at desc on a tie).

## Verification
- `pytest tests/unit/test_review_queue_sort_tz.py` -> 2 passed
- `black --check` clean; `check_module_stub_pollution` -> 0

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9571?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/review_queue.py
+++ b/backend/database/review_queue.py
@@ -138,7 +138,9 @@ def purge_stale_review_conflicts_for_memories(
 
 def _review_conflict_sort_key(item: Dict[str, Any]) -> tuple[float, datetime]:
     impact_value: float = cast(float, item.get('impact') or 0.0)
-    created_value: datetime = cast(datetime, item.get('created_at') or datetime.min)
+    # Use a tz-aware sentinel: stored created_at is tz-aware, and on an impact tie the sort compares
+    # the datetimes, so a naive datetime.min would raise "can't compare naive and aware" -> 500.
+    created_value: datetime = cast(datetime, item.get('created_at') or datetime.min.replace(tzinfo=timezone.utc))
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Backend bug fix — root cause stated, scoped, CI green. Merging.

PR #9572, opened 2026-07-12
fix(memory): no-op mark_consolidated on an absent short-term doc instead of 500ing resolve

ZachL111:
## Problem
`mark_consolidated` did `doc_ref.update(...)` with no existence check. Firestore `.update()` raises `NotFound` on a missing document (unlike `set`). Resolving a review conflict (`POST /v3/memories/review-queue/{id}/resolve`) calls `mark_consolidated` on the conflict's `source_short_term_id`, and canonical cohorts write `memory_items` rather than `short_term`, so that id can point at an absent `short_term` doc -> `NotFound` -> HTTP 500.

## Fix
Add an existence guard and no-op if the doc is gone, exactly as `database/memory_app_key_grants` already does for the same `update()`-on-a-missing-doc hazard.

## Test
`tests/unit/test_mark_consolidated_missing_doc.py`: with the doc absent, `mark_consolidated` returns without calling `.update()` (no raise); with it present, the update is applied.

## Verification
- `pytest tests/unit/test_mark_consolidated_missing_doc.py` -> 2 passed
- `black --check` clean; `check_module_stub_pollution` -> 0

- INV-MEM-1 (memory tiers) — this touches `database/short_term_memories.py`. The change is a defensive existence guard on `mark_consolidated`'s Firestore `.update()`; it does not alter the three-tier model or consolidation semantics.

### DIFF
--- a/backend/database/short_term_memories.py
+++ b/backend/database/short_term_memories.py
@@ -21,6 +21,11 @@ def _is_payload(value: object) -> TypeGuard[Payload]:
 
 def mark_consolidated(uid: str, short_term_id: str, commit_id: Optional[str]) -> None:
     doc_ref = db.collection(users_collection).document(uid).collection(short_term_collection).document(short_term_id)
+    # A conflict's source_short_term_id can point at an absent short-term doc (canonical cohorts write
+    # memory_items, not short_term). Firestore .update() raises NotFound on a missing doc (unlike set),
+    # which would surface as a 500 on resolve; no-op instead, mirroring memory_app_key_grants.
+    if not doc_ref.get().exists:
+        return
     now = datetime.now(timezone.utc)
     doc_ref.update(
         {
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for the focused fix. I reviewed the diff and the new unit coverage.

This looks like a reasonable defensive change: `mark_consolidated()` currently writes through Firestore `.update()`, so a missing legacy `short_term` document can raise instead of letting review-queue resolution complete. The added existence check makes the missing-doc case a no-op, while the test covers both absent and present document paths.

I’m not formally approving from automation because this touches memory/review-queue persistence behavior, so a human maintainer should confirm that no-op is the desired semantic when `source_short_term_id` points outside the legacy `short_term` collection. I don’t see a code-level blocker in this patch.

ZachL111:
Good question on the semantic. mark_consolidated lives in the legacy shadow short_term collection helpers, which the module docstring flags as data-drain only: canonical cohorts write memory_items, not short_term. So on a canonical review-queue resolve there is often no short_term shadow row for that id at all, and the old .update() then raised NotFound and failed the resolve.

The no-op is intentional: the resolve should complete whether or not a legacy shadow row exists, and there is nothing to mark consolidated when the row was never written to short_term. tombstone_source in the same module already tolerates the absent-row case by iterating an empty collection, so this keeps mark_consolidated consistent with it.

If you would rather make the missing-row case observable, I can add a debug log on the no-op instead of returning silently, but I would keep it non-fatal so review-queue resolution never 500s on an absent legacy doc.

kodjima33 (approved):
Backend bug fix: no-op mark_consolidated on an absent short-term doc instead of 500ing resolve; has test.

PR #9575, opened 2026-07-12
fix(goals): skip a malformed progress event instead of 500ing the goal detail page

ZachL111:
## What

`list_goal_progress_events` built a `GoalProgressEvent` per Firestore event doc with a bare list comprehension and no per-item guard:

```python
return [GoalProgressEvent.model_validate(_snapshot_dict(snapshot)) for snapshot in query.stream()]
```

`GoalProgressEvent` is `extra='forbid'` with required `event_id`, `goal_id`, `sequence` (>= 1), `kind`, `summary` (min_length=1), and `created_at`. One malformed or legacy event doc (a stray field, an empty summary, `sequence` < 1) raises `ValidationError`, which propagates out and 500s the whole goal detail page. `get_goal_detail` passes this list straight into the detail projection as `progress_events=`, and `GET /v1/goals/{goal_id}/detail` does not catch `ValidationError`, so it becomes an HTTP 500 for the entire page.

## Fix

Iterate and skip a malformed doc rather than fail the whole read:

```python
events: list[GoalProgressEvent] = []
for snapshot in query.stream():
    try:
        events.append(GoalProgressEvent.model_validate(_snapshot_dict(snapshot)))
    except ValidationError as e:
        logger.warning('Skipping malformed goal progress event %s: %s', getattr(snapshot, 'id', None), e)
return events
```

The catch is narrow (`ValidationError` only) so an unexpected error still propagates. This is the same failure class already guarded elsewhere: the workstreams list loops (`list_artifact_descriptors`, `list_continuation_checkpoints`) and the action-item readers in `routers/action_items.py` / `routers/developer.py` (`_safe_action_item_responses`).

## Tests

New `tests/unit/test_goal_progress_events_skip_malformed.py` drives the function directly with an injected fake client (database.goals is light):
- a malformed event in the middle of the stream is skipped, the valid ones are returned

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_goal_progress_events_skip_malformed.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check database/goals.py tests/unit/test_goal_progress_events_skip_malformed.py
  -> unchanged
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9575?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/goals.py
+++ b/backend/database/goals.py
@@ -2,12 +2,14 @@
 
 import hashlib
 import json
+import logging
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, cast
 from uuid import uuid4
 
 from google.cloud import firestore
 from google.cloud.firestore_v1 import FieldFilter
+from pydantic import ValidationError
 
 from database import _client
 from models.goal import (
@@ -21,6 +23,8 @@ from models.goal import (
     GoalType,
 )
 from models.task_intelligence import TaskWorkflowControl, TaskWorkflowMode
+
+logger = logging.getLogger(__name__)
 
 goals_collection = 'goals'
 goal_history_collection = 'goal_history'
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Thanks for the focused fix — this is a sensible resilience improvement. I reviewed the changed backend path and the narrow `ValidationError` catch looks appropriate: malformed/legacy progress-event documents are skipped so the goal detail response can still render, while unexpected exceptions still propagate instead of being silently hidden.

The added unit coverage also checks both important cases: a bad event in the stream is skipped, and a non-validation exception is not swallowed.

I’m leaving this as a positive signal rather than a formal approval because this touches backend goal/user-data handling and the maintainer gate requires human review for that surface. From the code review side, I don’t see a blocking issue in this diff.

kodjima33 (approved):
Backend bug fix: skip a malformed goal progress event instead of 500ing the goal detail page; has test.

PR #9576, opened 2026-07-12
fix(dev-keys): skip a malformed dev API key instead of 500ing GET /v1/dev/keys

ZachL111:
## What

`get_dev_keys_for_user` built a `DevApiKey` per stored key doc with no id fallback and no per-item guard:

```python
for doc in docs:
    key_dict = doc.to_dict()
    if "scopes" not in key_dict:
        key_dict["scopes"] = None
    keys.append(DevApiKey.model_validate(key_dict))
```

`DevApiKey` requires `id`, `name`, `key_prefix`, and `created_at` (no defaults). Firestore `doc.to_dict()` does not include the document id, and older key docs may lack an explicit `id` field, so one legacy or malformed key doc raises `ValidationError`. `GET /v1/dev/keys` (`routers/developer.py`, `listApiKeys`) returns that list directly, so a single bad row 500s the whole endpoint and the user cannot list or manage any of their keys.

## Fix

Fall back to the document id and skip a genuinely malformed key rather than failing the whole read:

```python
key_dict["id"] = key_dict.get("id") or doc.id
try:
    keys.append(DevApiKey.model_validate(key_dict))
except ValidationError as e:
    logger.warning("Skipping malformed dev API key %s: %s", doc.id, e)
```

This is the exact pattern already merged for the sibling module: `database/mcp_api_key.py` `get_mcp_keys_for_user` does the same id fallback plus a narrow `ValidationError` skip for `GET /v1/mcp/keys`. The catch is narrow so an unexpected error still propagates.

## Tests

New `tests/unit/test_dev_keys_legacy_doc.py` drives the function with an injected fake db proxy (database.dev_api_key is light):
- an explicit id is preserved, a doc missing the id field falls back to doc.id, and a malformed doc is skipped so the list survives

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_dev_keys_legacy_doc.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check database/dev_api_key.py tests/unit/test_dev_keys_legacy_doc.py
  -> unchanged
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/dev_api_key.py
+++ b/backend/database/dev_api_key.py
@@ -1,8 +1,10 @@
+import logging
 import uuid
 from datetime import datetime
 from typing import List, Optional, Tuple
 
 from google.cloud import firestore
+from pydantic import ValidationError
 
 import database.redis_db as redis_db
 from database._client import db
@@ -13,6 +15,8 @@ from database.memory_app_key_grants import (
 from models.dev_api_key import DevApiKey
 from utils.dev_api_keys import generate_dev_api_key, hash_dev_api_key
 from utils.scopes import READ_ONLY_SCOPES, Scopes
+
+logger = logging.getLogger(__name__)
 
 
 def create_dev_key(user_id: str, name: str, scopes: Optional[List[str]] = None) -> Tuple[str, DevApiKey]:
@@ -82,7 +86,13 @@ def get_dev_keys_for_user(user_id: str) -> List[DevApiKey]:
         # Ensure scopes field is present (None for backward compat)
         if "scopes" not in key_dict:
             key_dict["scopes"] = None
-        keys.append(DevApiKey.model_validate(key_dict))
+        # Older key docs may omit the id field; fall back to the document id, mirroring
+        # get_mcp_keys_for_user. Skip a malformed/legacy key rather than 500 the whole list.
+        key_dict["id"] = key_dict.get("id") or doc.id
+        try:
+            keys.append(DevApiKey.model_validate(key_dict))
+        except ValidationError as e:
+            logger.warning("Skipping malformed dev API key %s: %s", doc.id, e)
     return keys
 
 
--- a/backend/tests/unit/test_dev_keys_legacy_doc.py
+++ b/backend/tests/unit/test_dev_keys_legacy_doc.py
@@ -1,0 +1,81 @@
+"""get_dev_keys_for_user must fall back to doc.id and skip a malformed legacy key, not 500 the list.
+
+GET /v1/dev/keys built DevApiKey.model_validate(doc.to_dict()) per key with no id fallback and no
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Thanks for the focused fix here. I reviewed the diff and ran the new targeted unit test in an isolated local env:

- `python -m pytest tests/unit/test_dev_keys_legacy_doc.py -q` -> 2 passed
- `black --line-length 120 --skip-string-normalization --check database/dev_api_key.py tests/unit/test_dev_keys_legacy_doc.py` -> unchanged

The implementation looks directionally good: it mirrors the MCP-key reader's `doc.id` fallback, catches only `pydantic.ValidationError`, and keeps unexpected exceptions visible. The test covers preserving explicit ids, falling back for legacy docs, skipping malformed docs without 500ing the whole list, and not swallowing non-validation errors.

Because this touches developer API key listing/auth-adjacent data, I’m not formally approving it via automation; a human maintainer should still give the final security-sensitive sign-off. No blocking code issue found from my review.

kodjima33 (approved):
Backend bug fix: skip a malformed dev API key instead of 500ing GET /v1/dev/keys — approve; needs rebase (conflicts).

Git-on-my-level:
Thanks again for the focused bug fix. I took another look against current `main`, and this now appears to be superseded by the newer developer API key metadata-repair path there.

Current `main` has `get_dev_keys_for_user_with_repair_info()` projecting dev key metadata through `project_api_key_metadata(document_id=doc.id, ...)`, which already uses the Firestore document id as the public key id and repairs malformed display metadata before `DevApiKey.model_validate()`. That path also tracks bounded repair reasons instead of silently dropping visible/revocable keys.

Because this branch is now conflicting with `backend/database/dev_api_key.py` and the same user-facing failure mode appears covered on `main`, I’m marking this as `superseded` rather than requesting code changes. A human maintainer should make the final call, but I don’t see a remaining reason to merge this branch as-is.

ZachL111:
Closing as superseded by main's dev_api_key.py refactor. get_dev_keys_for_user was reworked into get_dev_keys_for_user_with_repair_info with a project_api_key_metadata projection/repair layer that normalizes legacy key documents, which is the same legacy/malformed-key concern this PR targeted. This branch patches the old structure and cannot rebase cleanly. If the new projection path can still raise on a malformed key, that is a fresh PR against the new structure.

PR #9577, opened 2026-07-12
fix(chat): use a tz-aware sentinel when sorting chat files with a missing created_at

ZachL111:
## What

`get_chat_files_desc` merges results across 30-id chunks when more than 30 file ids are requested, then sorts them in Python:

```python
results.sort(key=lambda x: x.get('created_at', datetime.min), reverse=True)
```

The default is a bare naive `datetime.min`. Stored `created_at` is tz-aware (`datetime.now(timezone.utc)`), so if any file doc is missing `created_at`, the sort compares the naive sentinel against a tz-aware value and raises `TypeError: can't compare offset-naive and offset-aware datetimes`, which 500s the request. This is a latent inconsistency rather than a hot path (the manual sort only runs on the >30-file branch), but it is a real crash, and `chat.py:456` was the lone sort site still using the naive sentinel.

## Fix

Use a tz-aware sentinel with the `or` fallback (so an absent or None `created_at` both sort last), matching the existing convention in the sibling sort sites:

```python
results.sort(key=lambda x: x.get('created_at') or datetime.min.replace(tzinfo=timezone.utc), reverse=True)
```

- `database/action_items.py` (`synced_items.sort(... or datetime.min.replace(tzinfo=timezone.utc))`)
- `database/task_recommendations.py`
- `database/memory_ledger.py`

## Tests

New `tests/unit/test_chat_files_sort_tz.py` drives the real >30-file chunked path with an injected fake db proxy:
- one file doc missing `created_at` no longer raises; it sorts last
- files with `created_at` present still order newest-first across chunks

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_chat_files_sort_tz.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check database/chat.py tests/unit/test_chat_files_sort_tz.py
  -> unchanged
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9577?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/chat.py
+++ b/backend/database/chat.py
@@ -452,8 +452,10 @@ def get_chat_files_desc(uid: str, files_id: Optional[List[str]] = None, limit: i
         chunk_ref = chunk_ref.order_by('created_at', direction=firestore.Query.DESCENDING)
         results.extend([_typed_doc(doc) for doc in chunk_ref.stream()])
 
-    # Sort all results by created_at and limit
-    results.sort(key=lambda x: x.get('created_at', datetime.min), reverse=True)
+    # Sort all results by created_at and limit. Use a tz-aware sentinel (mirroring the sort keys in
+    # action_items/task_recommendations/memory_ledger): stored created_at is tz-aware, so a naive
+    # datetime.min default would raise "can't compare offset-naive and offset-aware datetimes".
+    results.sort(key=lambda x: x.get('created_at') or datetime.min.replace(tzinfo=timezone.utc), reverse=True)
     return results[:limit]
 
 
--- a/backend/tests/unit/test_chat_files_sort_tz.py
+++ b/backend/tests/unit/test_chat_files_sort_tz.py
@@ -1,0 +1,59 @@
+"""database.chat.get_chat_files_desc must not TypeError sorting files with mixed tz-aware/missing created_at.
+
+In the >30-file chunked merge path, the per-file sort key used a bare naive datetime.min default. Stored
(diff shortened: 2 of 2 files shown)


Git-on-my-level:

I reviewed the changed `get_chat_files_desc` path and the new regression coverage. Using a timezone-aware fallback sentinel fixes the mixed naive/aware datetime crash in the >30-file chunked merge path, and the tests cover both the missing-`created_at` case and cross-chunk descending order. I also verified the sort behavior locally with stubbed service dependencies, and Black reports the changed files would be left unchanged.

I’m not giving a formal automation approval here because the maintainer gate for this run disallows approval on the current risk/check metadata, but I don’t see a code-level blocker in the patch itself. Human maintainer review should be enough to confirm the CI/check-rollup state before merge.

kodjima33 (approved):
Backend bug fix: tz-aware sentinel when sorting chat files with a missing created_at — approve; needs rebase (conflicts).

ZachL111:
Closing as superseded: main already sorts get_chat_files_desc with a tz-aware sentinel (the #9571 pattern), which fixes the naive-vs-aware TypeError this PR targeted. The only delta here is handling an explicit created_at=None (vs a missing key), which is not worth a separate PR.

PR #9578, opened 2026-07-12
fix(apps): skip a malformed usage-history doc instead of 500ing app usage/earnings

ZachL111:
## What

`get_app_usage_history` and `get_app_money_made` built a `UsageHistoryItem` per raw usage doc with a bare comprehension and no per-item guard:

```python
usage = [UsageHistoryItem(**x) for x in usage]
```

`UsageHistoryItem` requires `uid`, `timestamp`, and `type` (a `UsageHistoryType` enum). These results are Redis/process-cached and shared across callers, so one legacy or malformed usage document (an out-of-enum `type`, a missing `timestamp`) raises `ValidationError` and 500s the whole app usage-history and earnings enrichment for that app.

## Fix

Route both sites through a shared `_safe_usage_history_items` helper that skips a malformed record rather than failing the whole build, mirroring the `_safe_build_app` guard already in this module for the marketplace list builders:

```python
def _safe_usage_history_items(usage, app_id):
    items = []
    for x in usage:
        try:
            items.append(UsageHistoryItem(**x))
        except ValidationError as e:
            logger.warning("Skipping malformed usage history item for app %s: %s", app_id, [err['loc'][0] for err in e.errors()])
    return items
```

The catch is narrow (`ValidationError`) and logs only the app id plus offending field names, matching `_safe_build_app`.

## Tests

New `tests/unit/test_app_usage_history_skip.py` calls the pure helper directly (same seam as `test_apps_marketplace_poison_guard.py`, which imports `utils.apps` with no stubs):
- valid records build items
- a record missing timestamp/type and a record with an out-of-enum type are skipped, and the valid one survives
- empty input returns empty

## Verification

```
cd backend && ENCRYPTION_SECRET=... OPENAI_API_KEY=... python -m pytest \
  tests/unit/test_app_usage_history_skip.py tests/unit/test_apps_marketplace_poison_guard.py -q
  -> 7 passed
black --line-length 120 --skip-string-normalization --check utils/apps.py tests/unit/test_app_usage_history_skip.py
  -> unchanged
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9578?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_app_usage_history_skip.py
+++ b/backend/tests/unit/test_app_usage_history_skip.py
@@ -1,0 +1,48 @@
+"""get_app_usage_history / get_app_money_made must skip a malformed usage doc, not 500 the enrichment.
+
+Both functions built UsageHistoryItem(**x) per raw usage doc with no guard. UsageHistoryItem requires
+uid/timestamp/type (enum), and the results are Redis/process-cached and shared, so one legacy or
+malformed usage document (a bad type enum, a missing timestamp) raised ValidationError and 500'd the
+whole app usage/earnings enrichment. _safe_usage_history_items skips such a record and logs the app id
+plus offending field names, mirroring _safe_build_app in the same module. The helper is a pure
+function, so the test imports and calls it directly (no monkeypatch, no sys.modules).
+"""
+
+import os
+
+os.environ.setdefault('OPENAI_API_KEY', 'sk-test-not-real')
+os.environ.setdefault('ENCRYPTION_SECRET', 'omi_ZwB2ZNqB2HHpMK6wStk7sTpavJiPTFg7gXUHnc4tFABPU6pZ2c2DKgehtfgi4RZv')
+
+from datetime import datetime, timezone
+
+import utils.apps as apps_utils  # noqa: E402
+from models.app import UsageHistoryItem, UsageHistoryType  # noqa: E402
+
+
+def _valid_usage_dict():
+    return {
+        'uid': 'u1',
+        'timestamp': datetime(2026, 1, 1, tzinfo=timezone.utc),
+        'type': UsageHistoryType.chat_message_sent.value,
+    }
+
+
+def test_safe_usage_history_items_returns_items_for_valid_records():
+    items = apps_utils._safe_usage_history_items([_valid_usage_dict(), _valid_usage_dict()], 'app1')
+    assert len(items) == 2
+    assert all(isinstance(i, UsageHistoryItem) for i in items)
+
+
+def test_safe_usage_history_items_skips_malformed_records():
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Nice defensive fix — this cleanly extends the `_safe_build_app` resilience pattern already in this module to the usage-history builders, which had the same unguarded `UsageHistoryItem(**x)` comprehension over shared/cached data.

A few observations from static review:

1. **Correctness:** `_safe_usage_history_items` is well-scoped — it catches only `ValidationError` (not a broad `Exception`), logs the app id plus offending field names for observability, and returns properly typed `UsageHistoryItem` objects so the downstream filters in `get_app_money_made` still work. Both call sites are updated consistently.

2. **Resilience trade-off:** Silently skipping malformed records means usage/earnings could be undercounted for an app with corrupted data. The warning log mitigates this (visible, not silent), and 500ing the entire shared/cached enrichment for all users is clearly worse. This matches the established `_safe_build_app` philosophy. For future consideration: if undercounting becomes a concern, a metric/counter on skipped records would help track drift.

3. **Tests:** Good coverage — valid records, missing-field docs, bad-enum docs, and empty input. The test seam (direct import of the pure helper) mirrors `test_apps_marketplace_poison_guard.py`.

The code looks mergeable from a review standpoint. CI shows some cancelled runs on Hygiene/Formatting/Detect Changes alongside passing ones — the latest Backend unit suite, Hermetic Backend E2E, and Public Developer API contract all pass. A maintainer should confirm the final checks state before merge, but this is a small, focused, low-risk fix that reads well.

ZachL111:
Thanks for the thorough review. Point 2 is a fair call on undercount drift.

I deliberately kept this a straight mirror of `_safe_build_app` (log-only) so the two skip paths in this module stay consistent. If we want the drift trackable, the right shape per the fallback-telemetry rule in AGENTS.md would be the shared `record_fallback` helper in `utils/observability/fallback.py` (outcome `degraded`) rather than a one-off `*_skipped_total` counter, and it should land on both `_safe_build_app` and `_safe_usage_history_items` at once so marketplace skips and usage skips report identically.

I'll track that as a separate follow-up so this PR stays small and focused on the 500 fix, unless you'd rather I bundle the telemetry into both skip paths here.

kodjima33 (approved):
Backend bug fix: skip a malformed usage-history doc instead of 500ing app usage/earnings; has test.

PR #9582, opened 2026-07-12
fix: folder assignments reverted on upsert, tasks lost on reprocess/merge, recurring tasks re-delegating agents every 60s

skanderkaroui:
Three root-cause fixes from an end-to-end review of the folder-assignment, action-item-extraction, and recurring-task-delegation pipelines. Each closes a failure class at the shared function all callers route through, with regression tests.

## 1. Conversations stopped being auto-filed into folders (and were misfiled before that)

**Root cause A (regression, cb7cea49e):** `upsert_conversation` preserved user-owned fields whenever the key existed in the stored doc. The in-progress stub written at transcribe time dumps `folder_id: None`, so `'folder_id' in existing` was always true and every AI folder assignment made during processing was silently reverted to null. The "AI assigned conversation to folder" log still fired, so the failure was invisible.
**Durable guard:** preserve only non-null values — null means "never user-set". A real concurrent user move is still preserved (transaction retry re-reads it, covered by the existing retry test). Regression test: stub-null `folder_id` no longer clobbers the assignment; non-null user values still win.

**Root cause B (since a9f542a5e):** removing the 'Other' system folder left no folder with `is_default=True`, so `get_default_folder_id()` returns None and the low-confidence guard (`confidence < threshold AND default_folder_id`) was dead — every sub-0.7 guess, including the 0.5 schema default when the LLM omits confidence, was accepted verbatim and conversations were filed into whichever folder the model half-liked.
**Durable guard:** low confidence now always falls back; with no default folder the conversation stays unfiled rather than misfiled. New test file pins the fallback matrix (no-default, with-default, confident-accept, omitted-confidence).

## 2. Action items silently vanished on reprocess and merge

**Root cause:** `_fetch_dedup_candidates` fed the conversation's own still-open action items (and, after a merge, the source conversations') to the extraction prompt as duplicate candidates. The LLM suppressed re-extracting them, and `_save_action_items` then deleted the old items and saved only the surviving subset — deleting the user's tasks.
**Durable guard:** dedup candidates now exclude items whose `conversation_id` is the conversation being processed or any id in its `merge_metadata.source_conversation_ids`, in the one helper all three extraction paths (external-integration, reprocess, streaming) route through. Regression tests pin both the exclusion and the unchanged no-context path.

## 3. Recurring tasks re-delegated an agent investigation every 60 seconds, forever

**Root cause:** the scheduler's claimed dedup did not exist. `getDueRecurringTasks` returns every incomplete recurring task with `dueAt <= now`, nothing advances `dueAt`, and both guards read fields only the legacy tmux path ever wrote: `chatSessionId` has no writers on the kernel path (permanently nil) and `agentStartedAt` was never stamped by `investigateInBackground`. Every due recurring task fired a fresh agent investigation per 60s tick — unbounded duplicate agent runs and thread spam.
**Durable guard:** `investigateInBackground` stamps `agentStartedAt` (new narrow `ActionItemStorage.updateAgentStartedAt` that leaves the other agent fields alone) before sending, and the scheduler gates all recurrence kinds on it — one investigation per task per 4 hours (the pre-existing daily heuristic, now actually live). The dead `chatSessionId` branch is deleted. `investigateInBackground` also now guards `activeTaskId == task.id`: `openThread` can early-exit (`isOpening`) or fail without resetting state when `revealPanel` is false, which previously sent this task's prompt into a previous task's thread with the wrong context packet.

**Recurrence evidence (failure class):** items 1 and 3 are the same class — a guard reading state that no live writer maintains (`is_default` after the folder removal; `chatSessionId`/`agentStartedAt` after the tmux-path migration). Both fixes move the gate onto state the live path actually writes.

**Deliberately deferred:** advancing `dueAt` per recurrence rule (real recurrence semantics — needs its own UX spec); reinstating a default system folder (unfiled is the safe state; product call).

- INV-CHAT-1

## Verification

- Backend full suite via `bash test.sh`: 542 test files, exit 0.
- Backend hermetic E2E harness (`bash backend/testing/e2e/run.sh`): all passed, exit 0 — includes the conversation-processing seam tests that exercise `process_conversation`/`upsert_conversation` through real routes.
- New/updated unit tests, run isolated under the pinned 3.11 venv: `test_conversation_revision_contract.py` (11 passed), `test_conversation_folder_assignment.py` (4 passed), `test_process_conversation_usage_context.py` (17 passed).
- Desktop: `xcrun swift build -c debug --package-path Desktop` clean; `xcrun swift test --filter RecurringTaskSchedulerGateTests` 4/4 (gate is a pure function with injected `now`, no wall-clock waits).
- `./scripts/agent-logic-harness.sh` green (Swift focused lifecycle/state tests + agent runtime 411 tests + pi-mono-extension package tests).
- `./scripts/desktop-core-harness.sh --self-check` green (flow lint 52 flows, gauntlet self-check, backend desktop contracts 8/8). Tier-2 hermetic run could not start on this machine: the firestore emulator port (8085, hardcoded) is held by a concurrently running dev harness in another worktree (pr9505) — not a failure of this change; can be rerun once that stack is down.
- `make preflight` green after adding the invariant citation above.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

(Folder assignments were reverted because the conversation upsert path also wrote folder state, competing with the folder-assignment owner; the fix restores one authoritative owner for that transition.)


Git-on-my-level (dismissed):
Thanks for digging into these regressions — the fixes are well-scoped and the added regression coverage is helpful. I found one blocking edge case in the folder-preservation change, though.

The `upsert_conversation` change now treats every existing `folder_id: None` as “never user-set” and allows processing output to overwrite it. That fixes the initial in-progress stub case, but `folder_id` is also user-owned and the folder API permits explicitly moving a conversation to no folder (`PATCH /v1/conversations/{conversation_id}/folder` with `folder_id: null`). If a user clears/removes the folder while processing is in flight, the transaction will re-read `folder_id: None` and no longer preserve that concurrent user mutation, so the AI-assigned folder can be written back over the user's explicit unfile action.

Could you narrow the exception to the actual stub/initial-processing case (for example by checking an existing in-progress/stub status or another reliable marker), and add a regression test where an existing completed conversation with `folder_id: None` is preserved over an incoming AI `folder_id`? The non-null user-folder preservation test is good, but we also need to protect the explicit-null user-owned state.

The task de-dup and recurring-task gate changes look directionally good to me from static review. I also tried the targeted backend tests locally, but this checkout does not have the backend dependencies installed (`langchain_core` / `google` import failures); GitHub's backend and desktop checks for this head are green.

kodjima33 (approved):
Three root-cause fixes (folder assignment revert, task loss on reprocess, recurring-task re-delegation) with regression tests — approve; needs a rebase (conflicts) before it can land.

skanderkaroui:
Addressed the blocking review point in c371453f1.

**Root cause acknowledged:** treating every existing `folder_id: None` as never-user-set let processing output overwrite an explicit user unfile (`PATCH .../folder` with `folder_id: null`), including when the unfile raced an in-flight processing run.

**Fix — ownership marker, not a status heuristic:** the user-facing folder moves (single + bulk) now stamp `folder_user_set: true` on the conversation doc, and `upsert_conversation` preserves `folder_id` — explicit null included — whenever the marker is present. A stub's never-touched null stays overwritable, so the original regression fix is intact. A status-based guard was considered and rejected because it misses exactly the raced-unfile case (the doc is still the in-progress stub when the user clears the folder).

**Regression tests added** (`test_conversation_revision_contract.py`, 18 passed locally):
- completed conversation with explicit-null folder preserved over incoming AI `folder_id` (the case you asked for)
- unfile raced against in-flight processing (stub status + marker) preserved
- untouched stub null still overwritable by the AI assignment

API models are unaffected by the new Firestore field (pydantic `extra=ignore`).

Git-on-my-level (commented):
Thanks for the follow-up here. The new `folder_user_set` marker plus the added regression cases address the explicit-null folder preservation concern I raised earlier: future user folder moves/clears are marked as user-owned, processing upserts preserve both non-null folder choices and marker-backed `folder_id: null`, and the original stub-null AI-assignment path remains covered.

From this pass, the task de-dup change and the recurring-task investigation gate also look coherent and are covered by targeted tests. I did not find a new blocking issue in static review. I also verified the PR head is now `c371453f19fb174d42c79600a91ce15e0c07c1d2`; GitHub reports the backend and desktop checks passing. My local targeted pytest attempt still cannot collect in this sandbox because backend dependencies are not installed (`langchain_core` / `google`), but `py_compile` on the touched backend files and `git diff --check` both pass locally.

I’m not formally approving because this PR still touches backend persistence semantics and desktop agent scheduling behavior, and it carries `needs-maintainer-review`; a human maintainer should make the final merge call. But from this maintainer-review pass, this is a positive signal and the previous requested-change item appears resolved.

undivisible (approved):
## Approve

Three root causes, three shared writers — not three more observers. `folder_user_set` keeps explicit unfile from being gaslit by the stub, dedup stops eating the conversation’s own tasks on reprocess, and `agentStartedAt` ends the every-60s agent séance. Prior explicit-null concern is covered. Rebase the three dirty files and ship.

undivisible:
Approved on the follow-up head — `folder_user_set`, self/merge dedup exclusion, `agentStartedAt` 4h gate. Prior explicit-null concern is covered.

Still DIRTY in `ActionItemStorage`, `RecurringTaskScheduler`, and one process-conversation test. Rebase those three; once clean this unblocks real user data-loss fixes. (If GitHub still shows an old CHANGES_REQUESTED from another reviewer after rebase, ask them to re-approve — substance is done.)

Git-on-my-level (commented):
Thanks for the latest merge/update. I re-reviewed the current head `231df9addd4377886cbd938f57b66ff2202d5c05` against the prior automation feedback.

The original explicit-null folder blocker looks resolved on this head: `folder_user_set` marks user folder moves/clears, processing upserts preserve marker-backed `folder_id: null`, and the added regression cases cover the completed-conversation, in-flight-processing, and initial-stub-null paths. The action-item de-dup and recurring-task investigation gate still look coherent from static review, and the backend unit suite is green in CI.

I’m not approving because the current head is not merge-ready: GitHub now reports failing checks. The actionable CI item I found is `Public Developer API contract`, which says `docs/api-reference/app-client-openapi.json` is stale and should be regenerated with `backend/scripts/export_openapi.py --write ../docs/api-reference/app-client-openapi.json`. The desktop aggregate check is also red because the underlying Desktop Swift static/test jobs were cancelled, so that lane needs a clean rerun/pass before merge.

I did not find a new code-level blocker in this pass, and I’ll treat the earlier folder-preservation change request as stale/resolved. Human maintainer review is still required because this changes backend persistence semantics and desktop agent scheduling behavior, and the PR carries `needs-maintainer-review`.

---

Issue #8991, opened 2026-07-04
feat(windows): Local File Indexing for AI Context


thesohamdatta:
## Description
Implement the Local File Indexing feature on Windows to ensure strict parity with the macOS application's `FileIndexerService.swift`. This will recursively scan predefined developer/user directories and index their metadata into the local `indexed_files` SQLite table.

## Requirements
- Node.js `fs` recursive scan restricted to standard folders (`Documents`, `Desktop`, `Downloads`, `Developer`, `Projects`, `Code`, `src`, `repos`, `Sites`).
- Max recursion depth: 3.
- Exclude common heavy directories (`node_modules`, `.git`, `.cache`, `build`, etc.).
- Exclude files > 500 MB.
- Save to `better-sqlite3` in batches of 500.
- Update `pendingFileIndexingChat` flag on completion for the chat UI to trigger the initial analysis notification.
- Use strict TypeScript types (Matt Pocock standard) and follow triage review.

## References
- Spec: `docs/superpowers/specs/2026-07-04-local-file-indexing-design.md`
- Plan: `docs/superpowers/plans/2026-07-04-local-file-indexing.md`

tianmind-studio:
Current-main implementation audit (`73ce2d651`): the core Windows file-indexing feature is now present, mostly through the full Windows-port landing and subsequent hardening.

Implemented today:

- typed Node `fs` scan planner with the requested Documents/Desktop/Downloads + developer roots and Windows Start Menu app roots
- `MAX_DEPTH = 3`, 500 MB file cap, hidden-file filtering, and the expanded macOS-parity/build-cache skip set
- `better-sqlite3` persistence in `indexed_files`
- incremental mtime diffing with one atomic transaction, safe deletion of genuinely removed paths, and subtree protection for unreadable/absent roots (stronger than the original clear-and-reinsert plan)
- onboarding scan/status UI, manual re-scan, and a consent-gated startup refresh for already-populated indexes

Native Windows verification: the four file-index policy/planner suites pass **57/57**; the onboarding integration suite passes **3/3**.

Two acceptance details from this issue are not literal current-main contracts:

- there is no `pendingFileIndexingChat` flag; onboarding now updates app/graph context via `addAppNodes()` and fire-and-forget `runAppIndexing()` instead of the originally proposed initial-analysis notification flag
- writes are atomic/incremental but not chunked into explicit batches of 500

[...]

PR #9595, opened 2026-07-13
fix(windows): skip generated file index directories

tianmind-studio:
## Summary
- aligns the Windows local file index skip list with macOS `FileIndexScanPolicy.standard`
- skips generated, dependency, build, and cache directories such as `.venv`, `.cache`, `build`, `dist`, `.next`, `target`, `vendor`, and `Library`
- preserves case-insensitive Windows matching and adds regression coverage for exact ignored names and similarly named allowed folders

## Why
Part of #8991. Windows already has the local file indexing pipeline, but its scan policy only skipped four directories. Developer roots can contain large generated folders, so the first onboarding/indexing scan could waste time and fill `indexed_files` with dependency and build metadata that macOS intentionally ignores.

None. This only narrows local Windows metadata indexing exclusions; it does not change memory tiers, auth/session behavior, chat continuity, agent control, integrations, or UI styling.

## Validation
- `pnpm exec vitest run src/main/fileIndex/fileTypes.test.ts src/main/fileIndex/scanRoots.test.ts src/main/fileIndex/scanRules.test.ts` -> 37 passed
- `pnpm exec prettier --check src/main/fileIndex/scanRules.ts src/main/fileIndex/scanRules.test.ts`
- `pnpm run typecheck:node`
- `pnpm run typecheck:web`
- `pnpm test` -> 82 files passed, 1 skipped; 577 tests passed, 3 skipped
- `pnpm run build`
- `git diff --check`

I did not run a default live file-index scan against the real Windows profile because that would enumerate personal Documents/Desktop metadata. The covered path is the same scan policy used by the real indexer.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9595?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/fileIndex/scanRules.test.ts
+++ b/desktop/windows/src/main/fileIndex/scanRules.test.ts
@@ -1,23 +1,54 @@
 import { describe, it, expect } from 'vitest'
 import { shouldVisitDir, shouldIndexFile, SKIP_DIRS, MAX_DEPTH, MAX_FILE_SIZE } from './scanRules'
 
+const GENERATED_AND_DEPENDENCY_DIRS = [
+  '.Trash',
+  'node_modules',
+  '.git',
+  '__pycache__',
+  '.venv',
+  'venv',
+  '.cache',
+  '.npm',
+  '.yarn',
+  'Pods',
+  'DerivedData',
+  '.build',
+  'build',
+  'dist',
+  '.next',
+  '.nuxt',
+  'target',
+  'vendor',
+  'Library',
+  '.local',
+  '.cargo',
+  '.rustup'
+] as const
+
 describe('shouldVisitDir', () => {
-  it('skips noise directories', () => {
-    expect(shouldVisitDir('node_modules', 1)).toBe(false)
-    expect(shouldVisitDir('.git', 1)).toBe(false)
-    expect(shouldVisitDir('__pycache__', 1)).toBe(false)
-    expect(shouldVisitDir('.Trash', 1)).toBe(false)
-    expect(SKIP_DIRS.has('.Trash')).toBe(true)
+  it.each(GENERATED_AND_DEPENDENCY_DIRS)('skips ignored directory %s', (name) => {
+    expect(SKIP_DIRS.has(name)).toBe(true)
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @tianmind-studio — this is a clean, well-scoped parity fix. A few notes for the maintainer who merges:

**Verified**
- The new Windows `SKIP_DIRS` (22 entries) matches macOS `FileIndexScanPolicy.standard.skipFolders` exactly, including order, so the cross-platform alignment claim holds.
- Matching is exact-name (case-normalized), not prefix-based, so legitimately-named project folders like `build-tools` or `vendor-notes` are still indexed. Good that the test covers this.
- Case-insensitive comparison on Windows is the right adaptation (NTFS is case-insensitive by default), distinct from macOS exact-case matching. Reasonable divergence.
- Targeted tests are added and the relevant checks that actually ran (`Typecheck · Lint · Test`, `Windows build smoke`) are green.

**For the maintainer**
- CI shows several matrix checks as `skipped`/`cancelled` for this path (Formatting, Hygiene, some Detect Changes runs) — these are path-filtered jobs that did not execute on this `desktop/windows` change, not failures. Nothing appears red for this head SHA.
- This touches the local file-index scan policy, which influences what personal Documents/Desktop metadata gets indexed. Worth a quick product judgment call that the expanded skip list is the desired behavior on Windows (I think it clearly is, since it only narrows indexing and mirrors macOS).
- One nit, not blocking: it might be worth a short follow-up to add a shared cross-platform constant or a sync test that asserts the Windows set equals the macOS `FileIndexScanPolicy.standard` set, so the two do not silently drift again. The current code comment ("Keep this aligned…") relies on manual discipline.

No request changes from me. Leaving for a maintainer to merge.

kodjima33 (approved):
Desktop/Windows fix: align file-index skip list with macOS scan policy so generated/dependency dirs aren't indexed; has tests.

PR #9048, opened 2026-07-05
feat: support Hardware AAD feature and improve power consumption

TuEmb:
This PR is to:

### 1. Hardware AAD (new t5838_aad.c/.h, mic.c; removes aad.c/aad.h)

On silence, the T5838 is clocked into its built-in AAD sleep (mic off, ~µA) and wakes on the WAKE pin (P1.02). Owned inside mic.c since it shares the PDM peripheral.
CLK + THSEL parked high during sleep so they don't fight the TXS0104 level-shifter's pull-ups (~1 mA leak).

### 2. Offline / idle power (sd_card.c, mic.c, main.c)

SD-NAND powered off during offline AAD sleep, remounted on wake (writes buffered across the remount); CS parked low; full bus disconnect at shutdown.
Status LED off during offline sleep.
Deep-sleep (system-off): mic_off() now cuts PDM_EN so mic + shifter fully power down → ~25 µA (IMU + OTA flash stay functional).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9048?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #9600, opened 2026-07-13
Review swarm: 7 fixes from 2-day merged-PR review

Git-on-my-level:
## Summary

Review-swarm findings from a 10-agent code review of the last 2 days of merged PRs (62 PRs / 1072 files). Seven isolated bugs were found and fixed; no behavior change beyond the fixes. Each commit is independent and reviewable on its own.

### High-severity correctness
- **`fix(chat)`: tz-aware `created_at` sort sentinel** — `get_chat_files_desc` (>30 file_ids merge path) used naive `datetime.min`, which TypeError-compares against tz-aware Firestore dates and 500s. Last instance of the #9571 failure class; all four `datetime.min` sort sentinels in `backend/` are now tz-aware.
- **`fix(agent-control)`: derive consent for `request_permission` in PTT/voice mode** — both voice paths never built a `PermissionRequestAuthorization`, so the consent gate always silently denied `request_permission` while the system prompt told the model to call it. The feature was non-functional in voice mode. Fixed by deriving authorization from the spoken turn transcript (mirrors the text path; fail-closed).

### Medium
- **`fix(translation)`: NLLB native-language-code case resolution** — `_resolve_nllb_code` lowercased input before checking the mixed-case `NLLB_TO_BCP47` keys, making the native-code passthrough dead. README documents `eng_Latn` as valid input.
- **`fix(nllb)`: autoscale on active requests instead of CPU-only** — the NLLB GPU HPA scaled only on CPU, which stays low during GPU inference, so it never scaled past minReplicas. Added a concurrency-based External metric on the existing `nllb_active_requests` gauge (`requestsPerPod: 4` in prod), mirroring the parakeet pattern, and wired the adapter rule into the cluster-wide prometheus-adapter (dev+prod). CPU retained as a secondary floor.
- **`fix(firmware)`: propagate `t5838_aad_init` failure from `mic_start`** — the init powers the 1.8V mic rail; on failure it only logged and returned 0 (false-ready at the caller: mic "starts" but captures silence). `main()` already handles a nonzero `mic_start` return via `error_microphone()`, so propagate the failure. Closes the #9048 false-ready goal at the caller level.

### Low / defense-in-depth
- **`fix(tasks)`: re-add defensive newest-first sort for suggested candidates** — #9591 removed the client-side sort, leaving the top-5 slice dependent on backend return order with no test. Re-added the sort as defense-in-depth.
- **`fix(ci)`: drop duplicate isinstance-ratchet exempt entry** — two commits registered the same script in the manifest `exempt:` list.

### Desktop changelog
Fragments added for the two user-visible desktop fixes (PTT permission consent, suggested-tasks ordering).

- INV-CHAT-1

## Verification
- Each fix reviewed hunk-by-hunk by a dedicated reviewer; the chat.py fix was cross-confirmed by two reviewers (sync/memory found it, API owner fixed it).
- Pre-push manifest checks pass (check-manifest-contract, diff-hygiene, architecture-guardrails, product-invariants with this citation, desktop-changelog-entry, brand-ui INV-UI-1, backend-import-purity, backend-async-blockers).
- No new tests added in this PR; the underlying behavior is covered by the existing suite for each touched area. The tz-sentinel and consent fixes are the kind of regression a follow-up test should pin (noted below).

## Notes / follow-ups
- The NLLB HPA active-requests metric requires the documented prometheus-adapter helm upgrade (same manual step parakeet needs) before it takes effect in-cluster.
- Reported-but-not-fixed concerns from the review (out of scope here): NLLB model re-downloaded every pod restart (emptyDir), gateway 13-label metrics cardinality monitoring, a static guard to prevent recurrence of the naive-`datetime.min` sort class.

🤖 Generated with Claude Code

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9600?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
@cursor fix CI

PR #9601, opened 2026-07-13
fix(desktop): restore PTT permission requests and device-scoped recommendations

Git-on-my-level:
## Summary

- let an explicitly requested PTT permission reach the canonical executor without a redundant in-agent confirmation; macOS remains the grant authority
- bind What Matters Now device scope to authenticated headers while accepting the legacy bare-hash alias during the desktop rollout
- keep the disabled canonical-maintenance cron regression test within the repository's fast-unit timing budget

The PTT direct-tool path omitted the authorization required by the shared permission executor, so it failed before macOS could present its prompt. PTT now derives the same single-use authorization from the active voice utterance as primary chat: an explicit request proceeds directly to the native prompt, while an unrelated request cannot open System Settings.

Desktop supplied a bare device hash in a query parameter while the backend compares the authenticated platform-prefixed device identity. The client no longer sends the redundant query parameter; the backend accepts only the equivalent legacy bare hash from the same authenticated request during rollout, rejects other devices, and canonicalizes accepted legacy IDs before persisting either snapshot type.

The CI failure was test-only: the disabled-cron regression test built a fake Firestore control document and patched the maintenance callback even though the production gate returns before either can be read. The test now exercises that gate with no unreachable setup.

- INV-AUTH-1
- INV-CHAT-1

## Verification

- `BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-focused-goals-tests.txt BACKEND_PYTEST_WORKERS=1 bash backend/test.sh` (14 passed)
- `cd desktop/macos && xcrun swift test --package-path Desktop --filter APIClientCandidateTests` (11 passed)
- `cd desktop/macos && xcrun swift test --package-path Desktop --filter PermissionRequestAuthorizationTests` (7 passed)
- `cd desktop/macos && xcrun swift test --package-path Desktop --filter RealtimeHubSpawnAgentTests` (17 passed)
- `cd backend && ../backend/.venv/bin/python -m pytest tests/unit/test_workstream_router_contract.py -q` (16 passed)
- `cd desktop/macos && python3 scripts/check_desktop_test_quality.py`
- `OMI_APP_NAME=omi-main-head-ecb0daf ./run.sh --yolo`, then `OMI_APP_NAME=omi-main-head-ecb0daf OMI_AUTOMATION_PORT=47919 ./scripts/omi-ctl state`
- `OMI_PR_BODY_FILE=/tmp/omi-pr-body.md make preflight`
- `cd backend && BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-backend-unit-failures.txt bash test.sh` (9 passed)


Git-on-my-level:
@cursor resolve merge conflict

Git-on-my-level:

PR #9606, opened 2026-07-13
fix(recommendations): skip a malformed decision record instead of 500ing get_decisions

ZachL111:
## What

`get_decisions` decoded stored decision records with an unguarded comprehension inside a Firestore transaction:

```python
records = [DecisionRecord.model_validate(record) for record in raw_records]
```

`DecisionRecord` is `ConfigDict(extra='forbid')` with several required fields, so one legacy or schema-drifted audit record in the stored `decisions` list (a stray field from a newer or older writer, or a missing required field) raises `ValidationError`. That propagates out of `get_decisions` and 500s the recommendation read that calls it (`utils/task_intelligence/recommendations.py`).

## Fix

Move the decode into a pure `_decision_records` helper that skips a malformed record rather than failing the whole batch, and rewire `get_decisions` to call it:

```python
def _decision_records(raw_records, evaluation_id):
    records = []
    for record in raw_records:
        try:
            records.append(DecisionRecord.model_validate(record))
        except ValidationError as e:
            logger.warning('Skipping malformed decision record in evaluation %s: %s', evaluation_id, e)
    records.sort(key=lambda record: record.subject_id)
    return records
```

The catch is narrow (`ValidationError` only) so an unexpected error still propagates. Extracting the helper also makes the skip behavior unit-testable without the surrounding `@firestore.transactional` read machinery. This matches the skip-malformed guards already used across the backend list readers.

## Tests

New `tests/unit/test_decision_records_skip_malformed.py` drives the helper directly:
- a malformed record in the middle of the list is skipped, the rest are returned sorted by subject_id

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_decision_records_skip_malformed.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check database/task_recommendations.py tests/unit/test_decision_records_skip_malformed.py
  -> unchanged
python scripts/check_module_stub_pollution.py -> Checked 614 file(s); 0 violation(s)
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/task_recommendations.py
+++ b/backend/database/task_recommendations.py
@@ -2,11 +2,13 @@
 
 import hashlib
 import json
+import logging
 from datetime import datetime, timezone
 from typing import Any, Optional, cast
 
 from google.cloud import firestore
 from google.cloud.firestore_v1 import FieldFilter
+from pydantic import ValidationError
 
 from database._client import get_firestore_client
 from models.task_recommendation import (
@@ -23,6 +25,8 @@ from models.task_recommendation import (
     WhatMattersNowProjection,
 )
 from models.task_intelligence import TaskWorkflowControl, TaskWorkflowMode
+
+logger = logging.getLogger(__name__)
 
 FEEDBACK_COLLECTION = 'task_feedback'
 OUTCOMES_COLLECTION = 'task_outcomes'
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for the focused fix — this looks like the right failure mode to handle. I reviewed the diff and ran the targeted backend tests locally:

- `ENCRYPTION_SECRET=... .venv/bin/python -m pytest tests/unit/test_decision_records_skip_malformed.py -q` → 2 passed
- `ENCRYPTION_SECRET=... .venv/bin/python -m pytest tests/unit/test_task_recommendations.py tests/unit/test_decision_records_skip_malformed.py -q` → 41 passed
- `uvx black --check --line-length 120 --skip-string-normalization database/task_recommendations.py tests/unit/test_decision_records_skip_malformed.py` → unchanged

The narrow `ValidationError` catch is good: malformed stored decision audit records are skipped, while unexpected exceptions still propagate. The regression coverage also checks the sort behavior and the non-ValidationError path.

I’m not formally approving from automation because this touches the backend recommendation/debug-decision read path and user-data-adjacent persistence, so it should still get maintainer sign-off before merge. But from this review, this is a clean, well-scoped bug fix and I don’t see requested changes.

kodjima33 (approved):
Backend bug fix: skip a malformed decision record instead of 500ing get_decisions; has test.

PR #9608, opened 2026-07-13
fix(chat-file): skip a malformed file doc instead of 500ing the chat-file flow

ZachL111:
## What

Three sites in `utils/other/chat_file.py` turned raw file docs into `FileChat` objects with an unguarded comprehension:

```python
files = [FileChat(**f) for f in files_data]   # + two more (line ~283, ~386)
```

`FileChat` requires `id`, `name`, `mime_type`, `openai_file_id`, and `created_at`. The file docs come from `chat_db.get_chat_files` / `get_chat_files_desc`, so one legacy or partial file document (missing `openai_file_id`, `mime_type`, or `created_at`) raises `ValidationError` and 500s the flows that build these objects: the async file-read path, `_fill_question` (answering with file context), and `cleanup` (session teardown).

## Fix

Route all three sites through a shared `_safe_file_chats` helper that skips a malformed record rather than failing the whole batch, mirroring `utils/apps.py` `_safe_build_app`:

```python
def _safe_file_chats(files_data):
    files = []
    for f in files_data:
        try:
            files.append(FileChat(**f))
        except ValidationError as e:
            logger.warning("Skipping malformed chat file %s: %s", f.get('id') if isinstance(f, dict) else None, [err['loc'][0] for err in e.errors()])
    return files
```

The catch is narrow (`ValidationError` only) so an unexpected error still propagates, and it logs the file id plus offending field names only.

## Tests

New `tests/unit/test_chat_file_skip_malformed.py` calls the pure helper directly (imports `utils.other.chat_file`, no stubs):
- valid records build FileChat objects
- a record missing required fields is skipped and the valid ones survive
- empty input returns empty

## Verification

```
cd backend && ENCRYPTION_SECRET=... OPENAI_API_KEY=... python -m pytest tests/unit/test_chat_file_skip_malformed.py -q
  -> 3 passed
black --line-length 120 --skip-string-normalization --check utils/other/chat_file.py tests/unit/test_chat_file_skip_malformed.py
  -> unchanged
python scripts/check_module_stub_pollution.py -> Checked 614 file(s); 0 violation(s)
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_chat_file_skip_malformed.py
+++ b/backend/tests/unit/test_chat_file_skip_malformed.py
@@ -1,0 +1,46 @@
+"""utils.other.chat_file._safe_file_chats skips a malformed file doc instead of 500ing the chat-file flow.
+
+get_chat_files / get_chat_files_desc results were turned into FileChat objects with unguarded
+[FileChat(**f) for f in ...] comprehensions at three sites. FileChat requires id/name/mime_type/
+openai_file_id/created_at, so one legacy or partial file document raised ValidationError and 500'd
+the whole attach / answer / cleanup flow. All three sites now route through _safe_file_chats, which
+skips a malformed record (mirroring utils.apps._safe_build_app). The helper is pure, so the test
+imports and calls it directly.
+"""
+
+import os
+
+os.environ.setdefault('OPENAI_API_KEY', 'sk-test-not-real')
+os.environ.setdefault('ENCRYPTION_SECRET', 'omi_ZwB2ZNqB2HHpMK6wStk7sTpavJiPTFg7gXUHnc4tFABPU6pZ2c2DKgehtfgi4RZv')
+
+from datetime import datetime, timezone
+
+import utils.other.chat_file as cf  # noqa: E402
+from models.chat import FileChat  # noqa: E402
+
+
+def _valid_file_dict():
+    return {
+        'id': 'f1',
+        'name': 'a.png',
+        'mime_type': 'image/png',
+        'openai_file_id': 'oai-1',
+        'created_at': datetime(2026, 1, 1, tzinfo=timezone.utc),
+    }
+
+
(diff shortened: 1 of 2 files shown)


Git-on-my-level (approved):
Good robustness fix. Routing all three `FileChat(**f)` comprehension sites through `_safe_file_chats` mirrors the established `utils.apps._safe_build_app` pattern, and the narrow `ValidationError` catch correctly lets unexpected errors propagate.

Verified:
- `FileChat` requires `id/name/mime_type/openai_file_id/created_at` with no defaults, so a partial doc would indeed 500 all three flows.
- The helper is structurally identical to `_safe_build_app` (same exception, same warning shape with file id + offending field names).
- `f.get('id')` in the except is safe: only dicts reach the ValidationError path; non-dict elements raise TypeError and propagate.
- The `cleanup` path is strictly improved: previously a single malformed file aborted the entire loop, leaving subsequent OpenAI files orphaned; now all valid files get cleaned up and only malformed ones are skipped (with a logged warning).

Tests cover valid records, malformed-skip, and empty input against the pure helper. Backend unit suite, E2E, and API contract checks all green.

Note: the `ENCRYPTION_SECRET` in the test file matches the repo-wide fixture used across 30+ existing tests and conftest.py, so that's fine.

Thanks for the clean fix, @ZachL111!

kodjima33 (approved):
Backend bug fix: skip malformed file docs instead of 500ing the chat-file flow; has test.

PR #9609, opened 2026-07-13
fix(windows): select primary display capture source

### LINKED ISSUE #9607: fix(windows): select the primary display by Electron display ID
## Problem

`desktop/windows/src/main/rewind/sourceId.ts` currently treats `desktopCapturer.getSources({ types: ['screen'] })[0]` as the primary display.

That assumption is not tied to Electron's Screen API. On a multi-display Windows setup, a reordered source list can make both Rewind capture and the `screen:readNow` fallback read a secondary monitor while the code and UI describe the source as primary.

Electron exposes the supported mapping for this: `DesktopCapturerSource.display_id` corresponds to `screen.getPrimaryDisplay().id`. The source `id` itself uses a sequential screen number and should not be inferred from display ordering.

Official API reference: https://www.electronjs.org/docs/latest/api/structures/desktop-capturer-source

## Proposed fix

- Match `source.display_id` against `String(screen.getPrimaryDisplay().id)`.
- Fall back to the first source only when Electron does not expose a matching display ID.
- Preserve the existing single-flight cache and display-change invalidation.
- Add regression tests for reordered sources, unavailable IDs, empty source lists, caching, and invalidation.

## Scope

Windows desktop only. This changes source selection, not capture cadence, image retention, OCR, privacy filters, or upload behavior.

## Local observation

A real Electron 39 probe on a single-display Windows machine returned one source whose `display_id` matched the primary display ID. The multi-display ordering case will be covered deterministically with mocked Electron sources.

tianmind-studio:
## Summary

- selects the Windows Rewind capture source by matching Electron's `DesktopCapturerSource.display_id` to `screen.getPrimaryDisplay().id`
- keeps the existing first-source fallback when Electron does not expose a usable display ID
- preserves session caching and display-change invalidation, with regression coverage for reordered multi-display sources

Fixes #9607.

## Root cause

`fetchPrimarySourceId()` returned `sources[0]` and treated the list position as primary-display identity. Electron exposes a separate `display_id` field for matching a desktop source to the Screen API, so a multi-display source list that is not primary-first could make both Rewind capture and the `screen:readNow` fallback read a secondary monitor.

## Durable guard

The selection boundary now uses Electron's documented display identity and falls back only when no match is available. The new test puts the primary source second and fails against the old implementation; additional cases cover unavailable IDs, no sources, caching, and display-metrics invalidation.

## Real-path exercise

Ran an Electron 39 metadata-only probe on Windows. The current single-display machine reported one source whose `display_id` matched `screen.getPrimaryDisplay().id`, and the matching source ID was valid for `chromeMediaSourceId`. The probe did not capture or persist screen pixels. Multi-display ordering is exercised deterministically in the regression test.

None. This only corrects Windows desktop source selection. It does not change capture cadence, retention, OCR, privacy filtering, memory tiers, auth/session behavior, chat continuity, agent control, integrations, or UI styling.

## Validation

- old-code regression proof: `pnpm exec vitest run src/main/rewind/sourceId.test.ts` -> 1 failed, 4 passed; old code selected `screen:0:0` instead of primary `screen:1:0`
- `pnpm exec vitest run src/main/rewind/sourceId.test.ts` -> 5 passed
- `pnpm exec prettier --check src/main/rewind/sourceId.ts src/main/rewind/sourceId.test.ts`
- `pnpm exec eslint src/main/rewind/sourceId.ts src/main/rewind/sourceId.test.ts --quiet`
- `pnpm run typecheck:node`
- `pnpm test` -> 83 files passed, 1 skipped; 555 tests passed, 3 skipped
- `pnpm run build`
- `git diff --check`

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9609?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/rewind/sourceId.test.ts
+++ b/desktop/windows/src/main/rewind/sourceId.test.ts
@@ -1,0 +1,90 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const electron = vi.hoisted(() => ({
+  getSources: vi.fn(),
+  getPrimaryDisplay: vi.fn(),
+  on: vi.fn()
+}))
+
+vi.mock('electron', () => ({
+  desktopCapturer: { getSources: electron.getSources },
+  screen: {
+    getPrimaryDisplay: electron.getPrimaryDisplay,
+    on: electron.on
+  }
+}))
+
+async function loadSourceId(): Promise<typeof import('./sourceId')> {
+  vi.resetModules()
+  return import('./sourceId')
+}
+
+describe('getPrimarySourceId', () => {
+  beforeEach(() => {
+    electron.getSources.mockReset()
+    electron.getPrimaryDisplay.mockReset()
+    electron.on.mockReset()
+    electron.getPrimaryDisplay.mockReturnValue({ id: 202 })
+  })
+
+  it('selects the source whose display ID matches the primary display', async () => {
+    electron.getSources.mockResolvedValue([
+      { id: 'screen:0:0', display_id: '101' },
+      { id: 'screen:1:0', display_id: '202' }
+    ])
+
+    const { getPrimarySourceId } = await loadSourceId()
+
+    await expect(getPrimarySourceId()).resolves.toBe('screen:1:0')
+    expect(electron.getPrimaryDisplay).toHaveBeenCalledOnce()
+  })
+
+  it('falls back to the first source when Electron exposes no matching display ID', async () => {
+    electron.getSources.mockResolvedValue([
+      { id: 'screen:0:0', display_id: '' },
+      { id: 'screen:1:0', display_id: '' }
+    ])
+
+    const { getPrimarySourceId } = await loadSourceId()
+
+    await expect(getPrimarySourceId()).resolves.toBe('screen:0:0')
+  })
+
+  it('returns null when no screen source is available', async () => {
+    electron.getSources.mockResolvedValue([])
+
+    const { getPrimarySourceId } = await loadSourceId()
+
+    await expect(getPrimarySourceId()).resolves.toBeNull()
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @tianmind-studio — this is a clean, well-targeted fix for a genuine bug. Selecting the primary-screen capture source via `display_id` matching instead of trusting list order is the right approach, and keeping the first-source fallback is sensible for the case where Electron exposes no usable display ID.

Static review notes:
- The change is small (+95/-1), cohesive to one concern (source selection), and adds solid regression coverage: primary-not-first, empty `display_id` fallback, no sources, caching, and display-metrics invalidation.
- `Typecheck · Lint · Test` and `Windows build smoke` passed on the head SHA.
- The cancelled/skipped `Hygiene`/`Formatting`/`Detect Changes` runs are gating formal approval for now; once CI is green on the latest push this is a good candidate for a maintainer merge.

No security, privacy, auth, or data-persistence surfaces touched. No agent-behavior/instruction docs affected.

Leaving a maintainer-review label rather than approving because of the current CI state (cancelled/skipped gating checks). Tagging a maintainer to re-check CI green and merge. Good work — appreciate the regression test especially.

kodjima33:
Thank you so much for this — really appreciate it 🙏 This exact issue (#9607) got fixed separately and is already on `main` (#9621), so I'm closing this one. Please keep them coming!

PR #9618, opened 2026-07-13
fix(ci): make pre-push preflight portable on Windows

### LINKED ISSUE #9610: fix(ci): make pre-push preflight portable on Windows
## Problem

The repository's linked-worktree pre-push hook cannot run on Windows. A real push from PowerShell reaches `.github/scripts/preflight_runner.py` and crashes before the wrapped checks start:

```text
  at preflight_runner.py:183
```

The signal-forwarding block has two POSIX-only assumptions:

- `signal.SIGHUP` is referenced unconditionally, but Windows Python does not define it.
- `os.killpg` is used for forwarding, but Windows Python does not expose process-group signaling through that API.

This forces Windows contributors to bypass the pre-push hook even when their local checks pass.

## Proposed fix

- Register only signal constants that exist on the current platform.
- Keep `os.killpg` process-group forwarding on POSIX.
- Fall back to `Popen.send_signal()` when `os.killpg` is unavailable.
- Preserve handler restoration and lock cleanup.
- Add platform-simulation unit tests and run the existing single-flight tests on Windows.

## Acceptance criteria

- [ ] `preflight_runner.py -- ...` executes successfully on Windows Python.
- [ ] A normal `git push` can enter and complete the pre-push single-flight checks without `--no-verify`.
- [ ] POSIX process-group forwarding behavior remains unchanged.
- [ ] Tests simulate missing `SIGHUP` and missing `os.killpg`.
- [ ] No dependency changes.

tianmind-studio:
## Summary

- registers only signal constants available on the current platform
- keeps POSIX process-group forwarding and falls back to `Popen.send_signal()` when `os.killpg` is unavailable
- replaces Windows `os.kill(pid, 0)` probing with a read-only `OpenProcess` / `GetExitCodeProcess` check
- passes the wrapper's current `$BASH` interpreter path to the runner so Windows can launch the script without re-resolving `bash`
- propagates Python UTF-8 mode through the pre-push runner's child checks and the direct `make preflight` entrypoint
- resolves backend Python from an explicit override or either POSIX/Windows venv layout, propagates it to Pyright, and does not make unselected backend checks require a venv
- keeps mixed-codepage child output observable with UTF-8 `\xNN` escapes instead of aborting the hook
- adds platform-simulation coverage while making the existing single-flight tests pass on Windows

Fixes #9610.

## Root cause

The pre-push single-flight path assumed POSIX signal, process, interpreter, and output behavior:

1. `signal.SIGHUP` was evaluated unconditionally, but Windows Python does not define it.
2. Signal forwarding called `os.killpg`, which Windows Python does not expose.
3. `process_exists()` used `os.kill(pid, 0)`. On Windows, signal value `0` is `CTRL_C_EVENT`, so a joiner could interrupt the shared console instead of performing a side-effect-free existence probe.
4. The wrapper passed the extensionless `scripts/pre-push` Bash file directly to Windows `CreateProcess`, which cannot resolve a shebang script without an interpreter.
5. Passing a bare `bash` executable name still allowed native Windows process lookup to select `C:\Windows\System32\bash.exe` (WSL) instead of the Git Bash interpreter that was already running the hook.
6. The log streamer decoded every child byte as strict UTF-8, so one local-codepage byte could abort all checks even though the wrapped command was still running.
7. Starting only the runner with `-X utf8` did not propagate UTF-8 mode to Python checks it launched, and the direct `make preflight` entrypoint did not set it at all. Those children fell back to the Windows ANSI code page and could fail on valid UTF-8 repository files.
8. `scripts/pre-push` hard-coded `backend/.venv/bin/python`, so a standard native Windows venv with `Scripts/python.exe` failed before selected backend checks could run.
9. Even after selecting a Windows interpreter, the backend type-check wrapper did not receive it and fell back to a PATH lookup that could inspect the wrong environment.

The first assumption crashed a real Windows push before any checks started. After fixing it, the third assumption reproduced by interrupting the existing single-flight test host, which confirmed the whole signal boundary needed to be made portable rather than patching only the first traceback.

## Durable guard

- Signal registration is derived from the constants the runtime actually exposes.
- POSIX keeps `killpg`; other platforms use the child process API.
- Windows PID liveness uses a query-only process handle, checks `STILL_ACTIVE`, and always closes the handle. Access-denied is treated conservatively as alive, matching the existing POSIX `PermissionError` behavior.
- The wrapper provides its existing `$BASH` path as a separate argv element; it does not use `shell=True`, command strings, or a second PATH lookup.
- The wrapper exports `PYTHONUTF8=1` before starting the runner, so every descendant Python check inherits the same mode; the make target sets the same environment at its process boundary.
- Backend Python selection remains lazy, honors `BACKEND_PYTHON`, otherwise detects both `.venv/bin/python` and `.venv/Scripts/python.exe`, and passes the selected interpreter to the Pyright lane.
- The runner uses `backslashreplace` while decoding child output, preserving invalid bytes as visible `\xNN` text without changing the child exit code.
- Tests simulate missing `SIGHUP`, missing `killpg`, and the Windows liveness branch, enforce the current-interpreter and UTF-8 entrypoint contracts, then exercise the full two-process single-flight join behavior.

## Real-path exercise

- Ran the existing concurrent single-flight tests under Windows Python 3.11: identical work joined and executed once; different input was rejected without interrupting the host.
- Ran a real Windows process probe: an active child returned `True`, then returned `False` after termination.
- Passed the current Git Bash `$BASH` path through the native Windows Python runner and launched GNU Bash successfully; the old bare `bash` argv selected WSL and failed.
- Cleared `PYTHONUTF8` in the parent PowerShell session, then ran `make preflight`; the make target propagated UTF-8 mode and all 9 selected checks passed, including backend workflow contracts and repository-wide desktop changelog validation.
- A normal `git push` completed without `--no-verify`, passed all 9 shared preflight checks, and ran the complete backend Pyright lane with 0 errors using the selected Windows interpreter. The backend unit step selected two real tests but exposed a separate CRLF list-consumer bug, so that step used its documented skip after the failure was filed as #9627. OpenAPI and macOS Swift used their documented platform/dependency skips.

None. This changes contributor tooling only and does not affect application runtime or product behavior.

## Validation

- old-code regression: existing `SingleFlightTests.test_identical_processes_join_and_execute_once` failed on Windows with `AttributeError: module 'signal' has no attribute 'SIGHUP'`
- `python .github/scripts/test_pr_preflight.py SignalCompatibilityTests` -> 4 passed
- `python .github/scripts/test_pr_preflight.py SingleFlightTests` -> 2 passed
- `python .github/scripts/test_pr_preflight.py` -> 16 passed
- `python .github/scripts/test_run_checks.py` -> 9 passed
- official backend workflow-contract test file through native Windows Python -> 17 passed
- `make preflight` with parent `PYTHONUTF8` unset -> all 9 selected checks passed
- locked Windows backend environment on CPython 3.11.15: `test-preflight.sh` -> 0 failures
- full backend Pyright lane with the selected Windows interpreter -> 0 errors
- `python -m black --line-length 120 --skip-string-normalization --check .github/scripts/preflight_runner.py .github/scripts/test_pr_preflight.py`
- `python -m py_compile .github/scripts/preflight_runner.py .github/scripts/test_pr_preflight.py`
- `bash -n scripts/pre-push-singleflight scripts/pre-push`
- `git diff --check`
- official backend workflow-contract test file on CPython 3.11.15 -> 17 passed
- real Windows `git push` hook -> all 9 PR preflight checks and the complete backend type-check lane passed; the selected-unit CRLF failure is isolated in #9627

### DIFF
--- a/.github/scripts/preflight_runner.py
+++ b/.github/scripts/preflight_runner.py
@@ -16,6 +16,24 @@ from pathlib import Path
 
 POLL_SECONDS = 0.2
 STATUS_INTERVAL_SECONDS = 5.0
+FORWARD_SIGNAL_NAMES = ("SIGINT", "SIGTERM", "SIGHUP")
+
+
+def supported_forward_signals() -> tuple[int, ...]:
+    return tuple(signum for name in FORWARD_SIGNAL_NAMES if isinstance((signum := getattr(signal, name, None)), int))
+
+
+def forward_signal_to_child(child: subprocess.Popen[str], signum: int) -> None:
+    if child.poll() is not None:
+        return
+    try:
+        kill_process_group = getattr(os, "killpg", None)
+        if callable(kill_process_group):
+            kill_process_group(child.pid, signum)
+        else:
+            child.send_signal(signum)
+    except ProcessLookupError:
+        pass
 
 
 def atomic_json(path: Path, value: dict) -> None:
@@ -32,9 +50,41 @@ def read_json(path: Path) -> dict:
         return {}
 
 
+def windows_process_exists(pid: int) -> bool:
+    import ctypes
+    from ctypes import wintypes
+
+    process_query_limited_information = 0x1000
+    error_access_denied = 5
+    still_active = 259
+    kernel32 = ctypes.WinDLL("kernel32", use_last_error=True)
+    kernel32.OpenProcess.argtypes = (wintypes.DWORD, wintypes.BOOL, wintypes.DWORD)
+    kernel32.OpenProcess.restype = wintypes.HANDLE
+    kernel32.GetExitCodeProcess.argtypes = (
+        wintypes.HANDLE,
+        ctypes.POINTER(wintypes.DWORD),
+    )
+    kernel32.GetExitCodeProcess.restype = wintypes.BOOL
+    kernel32.CloseHandle.argtypes = (wintypes.HANDLE,)
+    kernel32.CloseHandle.restype = wintypes.BOOL
+
+    handle = kernel32.OpenProcess(process_query_limited_information, False, pid)
+    if not handle:
+        return ctypes.get_last_error() == error_access_denied
+    try:
+        exit_code = wintypes.DWORD()
+        if not kernel32.GetExitCodeProcess(handle, ctypes.byref(exit_code)):
+            return True
+        return exit_code.value == still_active
+    finally:
+        kernel32.CloseHandle(handle)
+
+
 def process_exists(pid: int) -> bool:
     if pid <= 0:
         return False
+    if os.name == "nt":
+        return windows_process_exists(pid)
     try:
         os.kill(pid, 0)
         return True
@@ -173,15 +223,10 @@ def run_owned(
         )
 
     def forward_signal(signum: int, _frame: object) -> None:
-        if child is not None and child.poll() is None:
-            try:
-                os.killpg(child.pid, signum)
-            except ProcessLookupError:
-                pass
+        if child is not None:
+            forward_signal_to_child(child, signum)
 
-    previous_handlers = {
-        signum: signal.signal(signum, forward_signal) for signum in (signal.SIGINT, signal.SIGTERM, signal.SIGHUP)
-    }
+    previous_handlers = {signum: signal.signal(signum, forward_signal) for signum in supported_forward_signals()}
     exit_code = 1
     try:
         print(f"Pre-push single-flight log: {log_path}")
@@ -195,6 +240,8 @@ def run_owned(
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
             text=True,
+            encoding="utf-8",
+            errors="backslashreplace",
             bufsize=1,
             start_new_session=True,
         )
--- a/.github/scripts/test_pr_preflight.py
+++ b/.github/scripts/test_pr_preflight.py
@@ -65,7 +65,7 @@ class MetadataTests(unittest.TestCase):
 class SelectionTests(unittest.TestCase):
     def test_pre_push_wrapper_reuses_current_bash_interpreter(self) -> None:
         wrapper = PRE_PUSH_SINGLEFLIGHT.read_text(encoding="utf-8")
-        self.assertIn("exec python3 -X utf8 ", wrapper)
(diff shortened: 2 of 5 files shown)


Git-on-my-level (commented):
Thanks for tightening up the Windows pre-push path. I reviewed this as workflow/hook-sensitive code.

What looks good:
- The runner no longer assumes `SIGHUP`, `os.killpg`, or POSIX `os.kill(pid, 0)` are available, while preserving POSIX process-group forwarding when supported.
- The wrapper now invokes `scripts/pre-push` through the current Bash interpreter and enables UTF-8 mode, which addresses the Windows shebang/interpreter and mixed-output failure modes described in the PR.
- Backend Python discovery now handles both POSIX and Windows venv layouts and passes the selected interpreter through to Pyright.
- The added regression coverage exercises the platform-compatibility seams and the invalid-UTF-8 log-streaming behavior.

Validation I ran locally on the PR checkout:
- `python .github/scripts/test_pr_preflight.py` → 16 tests passed.
- `python -m py_compile .github/scripts/preflight_runner.py .github/scripts/test_pr_preflight.py` → passed.
- `git diff origin/main...HEAD --check` → passed.
- `make -n preflight` → prints the expected `PYTHONUTF8=1 python3 .github/scripts/pr_preflight.py --lane local --base origin/main` command.

I’m not formally approving because this changes CI/pre-push workflow behavior and should get maintainer review before merge, ideally with confirmation from a real Windows + Git Bash hook run. From code review, this looks like a good targeted fix rather than a security concern.

kodjima33 (approved):
CI/preflight portability fixes for Windows — approve; leaving the merge to Nik since it touches the CI supply chain.

tianmind-studio:
Fresh real-world confirmation from Windows + Git Bash: pushing #9683 hit `AttributeError: module 'signal' has no attribute 'SIGHUP'` at `preflight_runner.py:183` before any checks could start. That is the exact path this PR guards, so the portability fix is still needed on current `main`.

tianmind-studio:
Closing this older implementation in favor of #10574. The current PR carries forward the Windows signal/liveness, UTF-8 streaming, Bash interpreter reuse, and venv/Pyright resolution fixes, rebases them onto current main, and adds Job Object/PID-identity safeguards plus path-scoped Windows CI coverage. Keeping one current review target avoids splitting maintainer attention; the reviews and validation here remain useful context.

Issue #9619, opened 2026-07-13
fix(backend): make route-policy preflight portable on Windows


tianmind-studio:
## Problem

Any committed `backend/**/*.py` change selects `backend-route-policy-baseline` in `make preflight`. On native Windows Python, that check fails before inventory validation because `backend/scripts/check_route_policy_baseline.py` passes `scripts/openapi_runner.sh` directly to `subprocess.run()`:

```text
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

Windows `CreateProcess` does not resolve Bash shebang scripts. Running the script through Git Bash exposes a second portability assumption: `backend/scripts/openapi_runner.sh` only recognizes `<venv>/bin/python`, while a native Windows venv provides `<venv>/Scripts/python.exe`.

## Reproduction

From PowerShell on Windows with Python 3.11, Git for Windows, uv, and the backend dependencies available:

```powershell
$env:OMI_PR_BODY_FILE = '<body.md>'
make preflight
```

Use a branch with a committed `backend/**/*.py` change so the route-policy check is selected.

## Expected behavior

- Python tooling should launch the Bash runner through the active Git for Windows interpreter instead of direct shebang execution or a bare PATH-shadowable `bash` name.
- The OpenAPI runner should detect both POSIX `bin/python` and Windows `Scripts/python.exe` venv layouts.
- Existing Linux and CI behavior should remain unchanged.

## Evidence

The failure reproduced on current `main` after the preceding manifest checks passed. The same machine resolves a bare `bash` from PowerShell to the WSL launcher, while the active Git installation provides a working `C:\Program Files\Git\bin\bash.exe`.

PR #9620, opened 2026-07-13
test(backend): run changed-files through Git Bash on Windows

### LINKED ISSUE #9617: test(backend): launch changed-files through Bash on Windows
## Problem

`backend/tests/unit/test_changed_files_script.py` launches `scripts/changed-files` directly with `subprocess.check_output()`.

That works on POSIX through the script shebang, but Windows `CreateProcess` cannot execute an extensionless Bash script. The selected backend workflow-contract tests therefore fail before any assertions run:

```text
OSError: [WinError 193] %1 is not a valid Win32 application
```

## Reproduction

On Windows with Python 3.11 and Git for Windows installed:

```powershell
python -m pytest backend/tests/unit/test_changed_files_script.py backend/tests/unit/test_workflow_contracts.py -q
```

Observed on current `main`: 18 tests pass and `test_changed_files_script.py` fails when it calls the script directly.

## Expected behavior

The test harness should invoke `scripts/changed-files` through Bash (or a shared portable script launcher), so the same contract can run on Windows without changing the production script.

## Scope

This is a test-launch portability issue. `scripts/changed-files` already runs under Bash in the real hook and CI paths.

tianmind-studio:
## Summary

- launch the `scripts/changed-files` contract test through Bash instead of asking Windows `CreateProcess` to execute an extensionless script
- on Windows, resolve `bash.exe` from the active Git for Windows installation rather than accepting a PATH-shadowing WSL executable
- add a platform-independent resolver guard so Linux CI also protects the Windows launcher contract

Fixes #9617.

## Root cause

The behavioral test passed `scripts/changed-files` directly to `subprocess.check_output()`. POSIX follows the script shebang, but native Windows `CreateProcess` rejects the extensionless Bash file with `WinError 193` before any assertion runs.

Using a bare `bash` argv is not sufficient on Windows: this machine exposes `C:\Windows\System32\bash.exe` ahead of Git Bash, which starts WSL and fails when WSL virtualization is unavailable.

## Durable guard

- POSIX keeps the normal `bash` lookup.
- Windows asks `git --exec-path` for the active Git installation, searches that installation's parent layout for `bin/bash.exe` or `usr/bin/bash.exe`, and passes the resolved executable as a structured argv element.
- The test does not use `shell=True`, a command string, or a hard-coded installation path.
- A hermetic path-layout test verifies that resolver behavior on every CI platform.

## Real-path exercise

- On the old code, native Windows Python 3.11 reproduced `OSError: [WinError 193]` at the direct script launch.
- On the fixed code, native Windows Python selected `C:\Program Files\Git\bin\bash.exe` and exercised the real `scripts/changed-files` behavior successfully.
- Launching the same pytest file from Git Bash into native Windows Python also passed.

None. This changes a repository test harness only.

## Validation

- `python -m pytest backend/tests/unit/test_changed_files_script.py -q` -> 3 passed
- `python -m pytest backend/tests/unit/test_changed_files_script.py backend/tests/unit/test_workflow_contracts.py -q` -> 20 passed
- `python -m black --check --line-length 120 --skip-string-normalization backend/tests/unit/test_changed_files_script.py`
- `python -m py_compile backend/tests/unit/test_changed_files_script.py`
- `git diff --check`
- remaining selected manifest checks (`lifecycle`, filename policy, import purity, module isolation, workflow contracts) -> passed when run directly
- Windows `make preflight` reached the independently broken direct Bash launch in `backend-route-policy-baseline`; that local preflight blocker is tracked separately in #9619

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9620?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_changed_files_script.py
+++ b/backend/tests/unit/test_changed_files_script.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 import subprocess
 
@@ -7,6 +8,21 @@ CHANGED_FILES = ROOT / 'scripts/changed-files'
 
 def _git(repo: Path, *args: str) -> str:
     return subprocess.check_output(['git', *args], cwd=repo, text=True).strip()
+
+
+def _find_git_bash(git_exec_path: Path) -> Path:
+    for parent in git_exec_path.parents:
+        for relative_path in ('bin/bash.exe', 'usr/bin/bash.exe'):
+            candidate = parent / relative_path
+            if candidate.is_file():
+                return candidate
+    raise FileNotFoundError(f'Git Bash was not found above {git_exec_path}')
+
+
+def _bash() -> str:
+    if os.name != 'nt':
+        return 'bash'
+    return str(_find_git_bash(Path(_git(ROOT, '--exec-path'))))
 
 
 def _commit(repo: Path, message: str) -> str:
@@ -40,7 +56,7 @@ def test_changed_files_reports_delete_rename_and_file_type_change(tmp_path):
     changed_type.symlink_to('target.py')
     head = _commit(repo, 'move delete and change type')
 
-    changed = set(subprocess.check_output([str(CHANGED_FILES), base, head], cwd=repo, text=True).splitlines())
+    changed = set(subprocess.check_output([_bash(), str(CHANGED_FILES), base, head], cwd=repo, text=True).splitlines())
 
     assert changed == {
         'backend/routers/deleted.py',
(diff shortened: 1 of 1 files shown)


Git-on-my-level:
Nice fix — this cleanly addresses the Windows `CreateProcess` / `WinError 193` problem from #9617. Resolving `bash.exe` from the active Git for Windows installation via `git --exec-path` (rather than a bare `bash` lookup that can shadow to WSL) is the right call, and keeping POSIX on bare `bash` preserves existing behavior. The structured argv approach (no `shell=True`, no command string) is good practice. The hermetic resolver test is a solid addition for cross-platform coverage. Holding formal approval due to the current gate, but this looks good to go once that clears.

kodjima33 (approved):
Test-only: launch changed-files contract test via Git Bash on Windows, fixes #9617.

PR #9624, opened 2026-07-13
fix(backend): make route-policy preflight portable on Windows

tianmind-studio:
## Summary

- launches the route-policy OpenAPI runner through Bash instead of passing an extensionless shell script to Windows `CreateProcess`
- resolves Git Bash from the active Git for Windows installation, avoiding the WSL `bash.exe` alias on native Windows
- accepts both POSIX `bin/python` and Windows `Scripts/python.exe` virtual-environment layouts in the OpenAPI runner
- refuses to delete a custom runner venv when its Python version is incompatible
- adds regression coverage for interpreter discovery and the native Windows venv layout

Fixes #9619.

## Root cause

`check_route_policy_baseline.py` passed `scripts/openapi_runner.sh` directly to `subprocess.run()`. POSIX resolves the script through its shebang, but native Windows returns `FileNotFoundError: [WinError 2]` before the route inventory starts.

Passing a bare `bash` name is not sufficient on Windows because native process lookup can select `C:\Windows\System32\bash.exe` (the WSL launcher) instead of the Git Bash interpreter used by the repository tooling. Once the script is launched with Git Bash, its hard-coded `<venv>/bin/python` path exposes the second assumption: native Windows venvs create `<venv>/Scripts/python.exe`.

## Durable guard

- The Python checker builds a structured argv and discovers `bin/bash.exe` or `usr/bin/bash.exe` within the active Git installation layout. It ignores inherited `GIT_EXEC_PATH`, bounds the parent search, and does not use `shell=True` or command-string interpolation.
- POSIX behavior remains a normal `bash` invocation.
- The shell runner resolves one venv Python path and uses it consistently for version validation, dependency sync, cache prewarming, and the final command.
- A mismatched repository-default venv remains an automatically replaceable cache; a mismatched custom override fails with an actionable message and is left intact.
- The regression test creates an isolated `Scripts/python.exe` layout and verifies the requested OpenAPI command actually runs.

## Real-path exercise

On Windows Python 3.11, the updated checker created and used a native `Scripts/python.exe` OpenAPI venv, synchronized the pinned runner dependencies, generated the real route inventory, and completed with:

```text
base_baseline_missing_entries: 477
baseline_additions: 0
current_missing_manifest_entries: 477
new_missing_manifest_entries: 0
stale_missing_baseline_entries: 0
route policy missing-route baseline is current
```

## Product invariants affected

None. This changes contributor preflight tooling only and does not affect application runtime or route policy contents.

## Validation

[...]

PR #9626, opened 2026-07-13
fix(backend): make workflow contract reads UTF-8 on Windows

### LINKED ISSUE #9625: fix(backend): make workflow-contract reads UTF-8 on Windows
## Problem

`backend/test.sh` launches native Windows Python without selecting UTF-8 mode. Python therefore uses the machine's ANSI code page (GBK on this system), while the repository sources and workflow files are UTF-8.

The existing workflow-contract unit file then fails while reading valid repository files with `Path.read_text()`:

```text
UnicodeDecodeError: 'gbk' codec can't decode byte ...
```

The failures reproduced in `scripts/pre-push`, `desktop/macos/scripts/swift-test-suites.sh`, and `backend/scripts/backfill_legacy_memories.py`. The checks never reach their assertions even though the files are valid.

## Reproduction

On Windows with native Python 3.11 and Git Bash, run the existing workflow-contract test through the official backend runner without `PYTHONUTF8` set:

```bash
PYTHON=/path/to/windows/venv/Scripts/python.exe \
BACKEND_UNIT_TEST_FILE_LIST=/tmp/workflow-contract-tests.txt \
BACKEND_PYTEST_WORKERS=1 \
bash backend/test.sh
```

Current result: 4 failed, 13 passed, all four failures caused by the default GBK decoder.

## Expected behavior

- The backend test runner should provide a deterministic UTF-8 Python environment on every supported platform while preserving an explicit caller override.
- Repository workflow-contract tooling should specify UTF-8 when reading UTF-8 source and configuration files instead of relying on the process locale.
- The same existing test file should pass when launched from native Windows Python without callers having to set `PYTHONUTF8` manually.

## Evidence

Setting `PYTHONUTF8=1` for the same official `backend/test.sh` command changes the result to 17 passed. No test data or assertions need to change.

tianmind-studio:
## Summary

- gives every Python process started by `backend/test.sh` a deterministic UTF-8 default while preserving an explicit `PYTHONUTF8` override
- reads workflow contracts, changed-file lists, Python sources, and repository contract fixtures with an explicit UTF-8 encoding
- adds a contract test that keeps the runner's UTF-8 export ahead of its first Python selection/invocation

Fixes #9625.

## Root cause

Repository source and configuration files are UTF-8, but native Windows Python uses the system ANSI code page unless UTF-8 mode is selected. On a GBK system, the official backend runner reached existing `Path.read_text()` calls with `encoding='locale'` and failed before assertions when valid UTF-8 punctuation appeared in repository files.

The same workflow-contract helpers also relied on the process locale when called directly, so changing only the test runner would leave those CLI paths platform-dependent.

## Durable guard

- `backend/test.sh` exports `PYTHONUTF8=${PYTHONUTF8:-1}` before selecting or starting Python. Callers can still set `PYTHONUTF8=0` explicitly.
- Workflow-contract tooling specifies UTF-8 at each repository-text boundary instead of relying on ambient locale.
- The unit contract verifies both the export value and its placement before `PYTHON_BIN` setup.

## Real-path exercise

The official `backend/test.sh` path was run from Git Bash with native Windows Python 3.11 and `PYTHONUTF8` explicitly unset. Before the fix it produced four GBK `UnicodeDecodeError` failures; after the fix the same file completed 18/18 under the normal fast-unit timing guard.

None. This changes backend contributor/test tooling only and does not affect application runtime behavior.

## Validation

- old-code official runner with `PYTHONUTF8` unset -> 4 failed, 13 passed; all failures were GBK decode errors
- updated official runner with `PYTHONUTF8` unset -> 18 passed
- native Windows `python -m pytest backend/tests/unit/test_workflow_contracts.py -q` with `PYTHONUTF8` unset -> 18 passed
- native Windows `python backend/scripts/check_workflow_contracts.py` with `PYTHONUTF8` unset -> pass
- native Windows `python backend/scripts/select_backend_unit_tests.py --all` with `PYTHONUTF8` unset -> 557 test files discovered
- `make preflight` with UTF-8 enabled -> first 8 selected checks passed, then current `main` hit the separate Windows shell-launch failure fixed by #9624
- remaining lifecycle, filename, import-purity, module-isolation, and workflow-contract checks -> passed when run directly with the same generated changed-file list
- route-policy inventory through the synced Windows OpenAPI venv -> baseline current (477 existing, 0 additions, 0 stale)
- `python -m black --line-length 120 --skip-string-normalization --check backend/scripts/check_workflow_contracts.py backend/scripts/select_backend_unit_tests.py backend/tests/unit/test_workflow_contracts.py`
- `python -m py_compile backend/scripts/check_workflow_contracts.py backend/scripts/select_backend_unit_tests.py backend/tests/unit/test_workflow_contracts.py`
- `bash -n backend/test.sh`
- `git diff --check`

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/check_workflow_contracts.py
+++ b/backend/scripts/check_workflow_contracts.py
@@ -16,7 +16,7 @@ CONTRACTS_REL_PATH = "backend/testing/workflow_contracts.json"
 
 
 def load_contracts() -> dict[str, Any]:
-    return json.loads(CONTRACTS_PATH.read_text())
+    return json.loads(CONTRACTS_PATH.read_text(encoding="utf-8"))
 
 
 def normalize_path(path: str) -> str:
@@ -82,7 +82,7 @@ def check_no_large_tuple_results(contracts: dict[str, Any], changed_paths: list[
     for rel_path in sorted(workflow_sources(contracts, changed_paths, check_name="no_large_tuple_results")):
         source_path = REPO_DIR / rel_path
         try:
-            tree = ast.parse(source_path.read_text())
+            tree = ast.parse(source_path.read_text(encoding="utf-8"))
         except SyntaxError as exc:
             errors.append(f"{rel_path}:{exc.lineno}: cannot parse source: {exc.msg}")
             continue
@@ -106,7 +106,7 @@ def main() -> int:
 
     changed_paths = None
     if args.changed_files:
-        changed_paths = Path(args.changed_files).read_text().splitlines()
+        changed_paths = Path(args.changed_files).read_text(encoding="utf-8").splitlines()
 
     contracts = load_contracts()
     errors = check_no_large_tuple_results(contracts, changed_paths)
--- a/backend/scripts/select_backend_unit_tests.py
+++ b/backend/scripts/select_backend_unit_tests.py
@@ -289,7 +289,7 @@ def is_memory_policy_core_path(path: str) -> bool:
 def load_workflow_contracts() -> List[Dict[str, Any]]:
     if not WORKFLOW_CONTRACTS_PATH.exists():
         return []
-    data: object = json.loads(WORKFLOW_CONTRACTS_PATH.read_text())
+    data: object = json.loads(WORKFLOW_CONTRACTS_PATH.read_text(encoding='utf-8'))
     if not isinstance(data, dict):
         return []
     typed_data: Dict[str, Any] = cast(Dict[str, Any], data)
@@ -378,7 +378,7 @@ def match_tests(all_tests: list[str], test_globs: tuple[str, ...]) -> set[str]:
 
 
 def read_lines(path: Path) -> list[str]:
-    return [line.strip() for line in path.read_text().splitlines() if line.strip()]
+    return [line.strip() for line in path.read_text(encoding='utf-8').splitlines() if line.strip()]
 
 
 def existing_tests(paths: list[str], all_tests: list[str]) -> list[str]:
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Thanks for tightening this up. I reviewed the diff and ran the targeted backend workflow-contract unit test file locally; the change is small and looks directionally correct.

What I checked:
- `Path.read_text(...)` calls in the workflow-contract helpers and related contract tests now use explicit UTF-8 decoding where repository text is read.
- `backend/test.sh` sets `PYTHONUTF8` before selecting/invoking Python, while still allowing callers to override it explicitly.
- Local validation: `python -m pytest -q tests/unit/test_workflow_contracts.py` passed (`18 passed`).
- Current GitHub checks for this head are green, with unrelated lanes skipped by change detection as expected.

I’m not formally approving because this touches backend test/workflow-contract tooling and came through the higher-scrutiny path, but I don’t see a blocker in the implementation from this review. A maintainer should still make the final merge call.

kodjima33 (approved):
Backend fix: deterministic UTF-8 reads for workflow contracts/test runner on Windows, fixes #9625; has contract test.

PR #9631, opened 2026-07-13
test(backend): make vision stream progress clock-free

### LINKED ISSUE #9630: test(backend): avoid event-loop clock assumptions in vision streaming
## Problem

`backend/tests/unit/test_vision_stream_async.py::TestAsyncVisionStreamPattern::test_chunks_arrive_progressively` fails consistently on native Windows even though the producer yields between chunks and the consumer observes the expected sequence.

With the locked Windows backend environment (CPython 3.11.15, pytest 8.4.1), 20 isolated process runs produced 20 failures. A representative assertion was:

```text
assert (28524.468 - 28524.468) >= 0.02
```

The test records `loop.time()` for each received chunk and requires at least 20 ms between the first and last timestamps. Windows event-loop clock granularity can collapse those observations to the same reported time, so the test measures timer resolution rather than the producer/consumer contract introduced by #6285.

## Expected behavior

The test should fail when the producer buffers every chunk and completes before the consumer starts draining, but it should not depend on a platform-specific wall-clock threshold.

## Proposed fix

- Remove the received timestamp list and 20 ms assertion.
- While consuming each non-final chunk, assert that the producer task is not done yet.
- Keep the existing answer and ordered chunk assertions.

The task-state assertion preserves the progressive-streaming regression guard: a burst producer that queues all chunks and completes before consumption will make the assertion fail, without relying on clock precision.

tianmind-studio:
## Summary

- replaces a 20 ms event-loop clock threshold with a task-state progress contract
- keeps the vision stream test sensitive to producers that finish in one burst while remaining portable across Windows event loops

Fixes #9630.

## Root cause

`test_chunks_arrive_progressively` sampled `loop.time()` after each queued chunk and required at least 20 ms between the first and last samples. On native Windows, the event-loop clock can return the same observable value for these short intervals, so the test failed even though the consumer received each chunk while the producer was still active.

The old assertion measured clock resolution rather than the producer/consumer behavior that the test is intended to protect.

## Durable guard

For every intermediate chunk, the test now asserts that the producer task has not completed. An implementation that queues the entire response and terminator in one burst completes before the consumer observes the first chunk, so this assertion still fails for the regression while avoiding wall-clock precision assumptions.

The existing assertions continue to verify the final answer and exact chunk order.

## Real-path exercise

- Before the change, the affected test failed in 20 of 20 isolated native Windows CPython 3.11.15 runs.
- After the change, it passed in 20 of 20 isolated runs.
- The complete test module passed through the repository's `backend/test.sh` file-isolated runner on Windows.

None. This changes a backend unit test only.

## Validation

- `PYTHON=... BACKEND_UNIT_TEST_FILE_LIST=... BACKEND_PYTEST_WORKERS=1 bash backend/test.sh` -> 11 passed
- `python -m pytest backend/tests/unit/test_vision_stream_async.py -q` -> 11 passed
- `python -m black --check backend/tests/unit/test_vision_stream_async.py`
- `python -m py_compile backend/tests/unit/test_vision_stream_async.py`
- `git diff --check`
- full backend Pyright with the locked Windows interpreter -> 0 errors, 1756 existing warnings
- `make preflight` -> first 7 selected checks passed, then the known Windows shell-launch failure tracked and fixed by #9624; the route-policy check itself and all 5 later selected manifest checks passed when run directly

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9631?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_vision_stream_async.py
+++ b/backend/tests/unit/test_vision_stream_async.py
@@ -52,9 +52,7 @@ class TestAsyncVisionStreamPattern:
         callback = AsyncStreamingCallback()
         chunks = ["Hello", " world", "!"]
 
-        # Track when chunks arrive
         received = []
-        received_times = []
 
         async def _produce():
             return await _fake_vision_stream(chunks, callback)
@@ -62,12 +60,13 @@ class TestAsyncVisionStreamPattern:
         task = asyncio.create_task(_produce())
 
         # Drain queue concurrently (mirrors graph.py pattern)
-        loop = asyncio.get_event_loop()
         while True:
             chunk = await callback.queue.get()
             if chunk:
                 received.append(chunk)
-                received_times.append(loop.time())
+                # Intermediate chunks must be observable before the producer completes.
+                if len(received) < len(chunks):
+                    assert not task.done()
             else:
                 break
 
@@ -75,9 +74,6 @@ class TestAsyncVisionStreamPattern:
 
         assert answer == "Hello world!"
         assert received == ["data: Hello", "data:  world", "data: !"]
-        assert len(received_times) == 3
-        # Chunks should arrive at different times (progressive, not burst)
-        assert received_times[-1] - received_times[0] >= 0.02
 
     @pytest.mark.asyncio
     async def test_empty_stream_terminates(self):


Git-on-my-level:
Thanks for this — nice fix. Replacing the wall-clock 20ms threshold with a task-state assertion (`assert not task.done()` on intermediate chunks) is the right call: it tests the actual streaming contract (consumer observes chunks while producer is still active) rather than measuring event-loop clock resolution, which is platform-dependent and was the root cause of the Windows failures.

The new guard is not weaker — it directly catches the regression. A burst producer that queues everything before yielding completes before the consumer dequeues the first intermediate chunk, so `task.done()` would be True and the assertion fires. The existing answer/order assertions are preserved.

Quick CI note: several check runs show as cancelled (typical when a new commit supersedes an in-flight run). The latest Formatting/Hygiene/Detect Changes runs appear to have passed — please confirm CI is green on your end before a maintainer merges.

No product/security/agent-behavior concerns. This is a clean test-only improvement.

kodjima33 (approved):
Test-only: make vision stream progress assertion clock-free (Windows portability), fixes #9630.

PR #9634, opened 2026-07-13
ci: regression guardrails — PR scope governor, test-discovery ratchet, sync annotation ratchet

skanderkaroui:
## What this is

Three mechanical regression guardrails plus the written rules behind them, derived from an internal audit of our own recent merge history. No behavior changes; the bug fixes the research surfaced will land as separate PRs.

> **Revised after review** (thanks @Git-on-my-level): the PR scope check is now **advisory-only** — it reports and warns but never blocks, so it can never force a feature split. The override/label machinery that existed to waive the old fail tier is deleted. Details in item 1.

## The research

We audited the last ~30 merged PRs and found **9 confirmed regression pairs** — a merged PR that introduced a production bug, and a later PR that fixed it. Three PRs each shipped 2+ regressions. We then ran a controlled replay experiment: 18 independent blind pre-merge reviews across 7 of the bug-introducing PRs (reviewers were never told a bug existed), scored against pre-registered ground truth.

Key findings:

1. **Diff size predicts regressions.** Every recent PR above ~2,400 changed production-source lines shipped at least one regression (2,475 lines → 3 regressions incl. a paying user locked out of sync; 2,652 → 1; 51,564 → 4+ regressions). Review catch-rate collapses with size: a one-line type bug in a 61-file PR evaded six independent reviewers.
2. **CI coverage can silently rot.** The 51k-line PR also deleted two pytest invocations from `backend/test.sh` while the test files stayed in the tree; nothing failed. (A later runner rewrite to directory discovery happened to heal it — by luck, not by guard.)
3. **Type confusion ships when annotations are missing.** The `source=unknown` misattribution bug (fixed in #9588) flowed through an *unannotated* `source` parameter in `utils/sync/pipeline.py` that a metering string shadowed. The enforced pyright lane had nothing to check.
4. Two recurring patterns had no written rule: new fail-closed gates shipping with no satisfiable path for existing users/clients/keys (two incidents), and tests rewritten in the same PR to assert the new — wrong — constant (shipped a 9-day prod timeout regression, fixed in #9583).

## What this PR adds

**1. PR scope advisor** (`.github/scripts/check_pr_scope.py`, manifest check `pr-scope` — runs in the local pre-push lane and CI, **advisory-only, never blocks**)
Counts changed production-source lines (tests across all platform conventions, docs, l10n, lockfiles, generated files excluded; rename/quotePath-safe via `--no-renames` + C-quote handling). ≥1,500 emits a warning annotation; ≥3,000 warns that in the audited merge history every PR this size shipped a regression review missed — so the reviewer calibrates skepticism to the diff, and splits only when the pieces are independently verifiable without extra human verification passes. Exit code is always 0 on a countable diff; push events (post-merge `main`) are notice-only. Because nothing blocks, there is no waiver machinery — no override label, no env escape hatch, no label plumbing.

**2. Unit-test discovery ratchet** (`backend/scripts/check_unit_test_discovery.py`, wired into Backend Unit Tests before dependency install)
Fails when any test file under `backend/tests/` or `backend/testing/` is not selected by the unit runner, not verified against its dedicated runner workflow (the workflow file must exist and actually reference the tree — or, for explicit-list workflows like the GPU suite, name the individual file), not policy-excluded (live integration/eval), and not allowlisted. Allowlists only shrink. This audit itself surfaced that `tests/container/test_parakeet_vram_stress.py` is currently run by no workflow — it is now explicitly tracked as a manual-only harness instead of being silently vouched for.

**3. Parameter-annotation ratchet for the sync pipeline** (`backend/pyrightconfig.json` + annotations)
`executionEnvironments` elevates `reportMissingParameterType` to error for `utils/sync/` only, and this PR adds the five missing annotations — one of which is the exact `source` parameter from finding #3. Verified against the introducing PR's merge state: the rule flags that unannotated parameter at `pipeline.py:746`. Typed parameters make str-vs-enum confusion visible to pyright at annotated sinks and same-scope assignments (verified in isolation on pyright 1.1.403).

**4. AGENTS.md rules** — PR size is *reported, not bounded* (matching the advisory check), mechanical discovery enforcement, the legacy-principal test rule for new fail-closed gates, and the rewritten-test review lens. Each rule names its enforcement or the shipped failure it prevents. The latter two rules cost agent/review effort rather than human verification time.

## What this deliberately does not do

- No bug fixes — the still-open defects the audit found are being filed/fixed separately.
- No blocking on size, in either direction: small diffs also ship bugs (a 348-line PR shipped a blocker), and forcing large diffs to split multiplies human verification passes. The advisor only informs review depth.
- The scope check lives in the checks manifest like every other deterministic check (no exemptions).

## Verification

- `python3 .github/scripts/test_check_pr_scope.py` → 9 tests OK (classification, counting, advisory tiers, push notice-only).
- Replayed the advisor against the four historical PRs listed above (892 / 2,738 / 2,665 / 47,938 lines) → notice/warn/warn/warn, all exit 0; on this PR itself → 472 changed production-source lines (notice).
- `backend/scripts/check_unit_test_discovery.py` on the real tree → `564 selected, 49 covered by other runners/allowlist, 0 orphans`; a seeded `tests/<newdir>/test_orphan.py` fails the check with an actionable message.
- `pytest tests/unit/test_check_unit_test_discovery.py` → 8 passed.
- `bash backend/scripts/typecheck.sh` (enforced pyright lane, pinned 1.1.403, full deps) with the new ratchet → **0 errors**; the rule replayed against the introducing PR's merge state flags the unannotated `source` param.
- `python3 -m py_compile backend/utils/sync/pipeline.py` → passes; annotations are runtime-inert.
- `actionlint` on both changed workflows → clean. `.github/scripts/test_run_checks.py` (manifest drift guard) → OK. `.github/scripts/test_pr_preflight.py` → OK.
- Full local manifest lane after the advisory rework → 16 checks passed.
- Post-review hardening verified earlier in the PR: Swift/JS test-tree classification, quoted non-ASCII paths, seeded unwired `tests/container/` file IS flagged, pyright lane 0 errors, sync unit files 273 passed, Agent Runtime tool-surfaces 104 passed.
- Full `backend/test.sh` run (564 files): 562 files green in the parallel run; the two failures were environmental and pass on rerun — `test_rendered_deployment_contract.py` needed `helm` on PATH (8 passed once installed), and `test_sync_v2.py` tripped only the per-test CPU-timing guard under saturated parallel load (all 161 tests pass; clean and fast in isolation).

None — no product behavior changes; CI/tooling, type annotations, and docs only.

## Failure class
Failure-Class: none

(This PR ships CI guardrail tooling; its fix commits repair the guardrail scripts themselves, not a registered product failure class.)

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -4,16 +4,76 @@
 # Checks that consume PR metadata declare `requires_pr_body: true`; post-merge
 # push runs exclude only that category because no authoritative PR body is available.
 checks:
+  - id: dev-harness-unit-tests
+    command: ["bash", "scripts/dev-harness/run-tests.sh"]
+    triggers: ["scripts/dev-harness/**", "desktop/macos/scripts/qualification-lease-command.sh", "desktop/macos/scripts/omi-fault-inject.sh", "desktop/macos/scripts/release-keyvalue.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the hermetic dev-harness stack backs release qualification; the Jul 22 2026 beta qualification outage added a docker/native typesense runtime that must stay behaviorally covered"
+  - id: pre-tag-readiness-contract
+    command: ["bash", "desktop/macos/tests/test-pre-tag-readiness-contract.sh"]
+    triggers: ["desktop/macos/scripts/pre-tag-readiness.sh", "desktop/macos/scripts/qualification-runner-self-clean.py", "desktop/macos/scripts/qualification-watchdog.py", "desktop/macos/tests/test-pre-tag-readiness-contract.sh", ".github/scripts/verify-pre-tag-readiness.py", ".github/scripts/test_verify_pre_tag_readiness.py", ".github/workflows/desktop_auto_release.yml", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "trusted-M1 readiness must retain exact-SHA cache/lease ownership and cannot emit a passing receipt before authenticated cleanup"
+  - id: pre-tag-readiness-behavior
+    command: ["python3", "desktop/macos/tests/test_pre_tag_readiness_behavior.py"]
+    triggers: ["desktop/macos/scripts/pre-tag-readiness.sh", "desktop/macos/scripts/qualification-runner-self-clean.py", "desktop/macos/scripts/qualification-watchdog.py", "desktop/macos/tests/test_pre_tag_readiness_behavior.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "a disposable source and authenticated capability fakes prove the real readiness orchestrator emits its exact-SHA receipt only after ordered cleanup"
+  - id: verify-pre-tag-readiness-tests
+    command: ["python3", ".github/scripts/test_verify_pre_tag_readiness.py"]
+    triggers: [".github/scripts/verify-pre-tag-readiness.py", ".github/scripts/test_verify_pre_tag_readiness.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the receipt verifier must reject stale, non-offline, failed, and promotion-bearing readiness evidence before a tag mutation"
   - id: pre-push-final-pr-diff
     command: ["bash", "scripts/test-pre-push-diff-base.sh"]
     triggers: ["scripts/pre-push", "scripts/pre-push-diff-base", "scripts/test-pre-push-diff-base.sh", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "pre-push check selection must use the final PR diff rather than a stale remote feature tip"
+  - id: pre-push-ci-prediction-tests
+    command: ["python3", ".github/scripts/test_pre_push_ci_prediction.py"]
+    triggers: ["scripts/pre-push", "scripts/pre_push_ci_prediction.py", ".github/scripts/test_pre_push_ci_prediction.py", "desktop/macos/scripts/desktop-flow-lint.py", "desktop/macos/scripts/desktop_flow_contract.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "#9440: local CI prediction must select only bounded Flutter generator, desktop-flow, and Windows KG-worker checks for the final PR diff"
+  - id: pre-push-hatch-disclosure
+    command: ["python3", ".github/scripts/check_pre_push_hatch_disclosure.py"]
+    triggers: ["scripts/pre-push", ".github/scripts/check_pre_push_hatch_disclosure.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "make setup provisions no app toolchain while .github/checks-manifest.yaml is a ROUTING_INPUTS entry that wakes the Flutter codegen lane, so a docs-only PR hit 'run flutter pub get' with no mention of PRE_PUSH_SKIP_FLUTTER_GENERATED; actionlint's branch had the same omission, and a hatch nobody can find is not a hatch"
+  - id: setup-pre-push-prerequisites
+    command: ["bash", "scripts/test-make-setup.sh"]
+    triggers: ["Makefile", "backend/scripts/sync-python-deps.sh", "scripts/pre-push", "scripts/test-make-setup.sh", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "make setup must provision the minimal, rerunnable backend environment required by selected pre-push checks"
+  - id: desktop-release-process-guards
+    command: ["bash", "scripts/run-release-process-guards.sh"]
+    triggers: [".github/workflows/desktop_qualify_beta.yml", ".github/scripts/check-release-process-guards.py", "scripts/run-release-process-guards.sh", "backend/tests/unit/test_desktop_release_scripts.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "SCA-174: qualification must check out the exact release tag inside its isolated source directory; the guard and its focused behavioral test run on every relevant diff"
+  - id: failure-class-retirement-author
+    command: ["python3", ".github/scripts/test_failure_class_retirement.py"]
+    triggers: [".github/scripts/failure_class_retirement.py", ".github/scripts/test_failure_class_retirement.py", ".github/workflows/repo-hygiene.yml", "scripts/failure-class", ".github/checks-manifest.yaml"]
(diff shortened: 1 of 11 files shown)


Git-on-my-level:
I like the idea behind these mechanical regression guards, but PRs are expensive to test right now, so splitting up a large feature unnecessarily increases human time burden which IMO is worse than increasing agent review burden

Thoughts? I think 2 + 3 are good, but 1 + 4 need to be reconsidered

skanderkaroui:
Fair point — reworked in eafd3e0. The scope check (1) is now **advisory-only**: it still warns at 1,500 / 3,000 production-source lines (the 3,000 tier cites the audit so review depth scales with size) but never fails, so it can't force a split. That also let me delete the whole `scope-approved` label / override plumbing (net −76 lines). The AGENTS.md size rule (4) now says size is *reported, not bounded* — split only when the pieces are independently verifiable without extra human testing passes.

I kept the two other rules in 4 (legacy-principal test for new fail-closed gates; external citation when a test is rewritten alongside the code it asserts) since they don't force splits and their cost lands on agent/review effort — but happy to drop them too if you disagree. 2 + 3 unchanged.

Git-on-my-level (commented):
Thanks for the careful iteration here. I reviewed this as a workflow/guardrail change rather than a normal app-code patch.

Positive signal from my side:
- The PR scope check is now advisory-only and exits 0, which matches the maintainer constraint that size should inform review depth without forcing splits.
- I locally ran the touched manifest path with `python3 .github/scripts/run_checks.py --lane local --base origin/main --head HEAD`; all 16 selected local checks passed, including `pr-scope`, `pr-scope-helper-tests`, `backend-test-discovery`, backend import purity/module isolation, and workflow-contract checks.
- I also ran the focused tests directly: `.github/scripts/test_check_pr_scope.py`, `backend/tests/unit/test_check_unit_test_discovery.py`, and `backend/scripts/check_unit_test_discovery.py`.

Agent/maintainer consequence to call out: this PR changes `AGENTS.md`, so it will influence AI coding/review agents in this repo. The new instructions look safe and useful: they tell agents that PR size is reported but not bounded, add the pyright strictness ratchet rule, and make backend test discovery/fail-closed gate expectations explicit. That should steer agents toward better validation without creating a hard split mandate. Because it changes agent instructions plus CI/workflow guardrails, I still think a human maintainer should make the final policy call before merge.

I am not formally approving because this touches workflow/CI and repo-wide agent guidance, but I did not find a blocker in the implementation I reviewed.

undivisible (approved):
## Approve

Guardrails that report without becoming a second product owner: advisory scope (never blocks), discovery ratchet so tests can’t silently leave the runner, sync param-type ratchet for the `source=unknown` class. Exactly the kind of scar-tissue CI that earns its keep. Rebase the two conflicted files (`checks-manifest`, `AGENTS.md`) and merge.

undivisible:
Approved — advisory scope (never blocks), unit-test discovery ratchet, sync param-type ratchet. Exactly scar-tissue CI.

Conflicts are only `.github/checks-manifest.yaml` + `AGENTS.md` (plus a unit-suite workflow hunk if you rebase commit-by-commit). Keep scope advisory-only through the rebase. Ping when clean.

Git-on-my-level (dismissed):
Thanks for rebasing this. I re-reviewed the new head (`819f32bbaf26e5d6602b349674b28c0002da4e6b`) as a workflow/guardrail change.

I’m requesting changes for one concrete blocker on the current head:

- `python3 .github/scripts/run_checks.py --lane local --base origin/main --head HEAD` now fails `failure-class-protocol`. The new `fix(merge): restore upstream content for files clobbered by the main merge` commit is treated as a `fix:` commit but has no `Failure-Class: FC-<slug> | new | none` declaration, so the manifest exits non-zero. Please either add the required failure-class declaration to that fix commit/PR body path or rename/rework the commit subject if it is not an actual bug-fix commit.

Focused checks that still passed locally:
- `python3 .github/scripts/test_check_pr_scope.py` -> 9 passed
- `cd backend && python3 -m pytest -q tests/unit/test_check_unit_test_discovery.py scripts/check_unit_test_discovery.py --tb=short` -> 14 passed
- `cd backend && python3 scripts/check_unit_test_discovery.py` -> unit test discovery OK, 0 orphans

Agent/maintainer consequence: this PR still changes `AGENTS.md`, so it will alter how AI coding/review agents behave in this repo. The updated guidance appears safe and aligned with the maintainer direction: PR size is advisory/reporting-only rather than a hard split gate, backend test discovery becomes mechanically enforced, and agents are pushed to preserve fail-closed/legacy-principal coverage. That policy change still warrants human maintainer review before merge.

I’m not formally approving because the PR touches CI/workflow guardrails and repo-wide agent instructions, and the approval gate is also blocked by the current failing/cancelled GitHub checks.

---

undivisible (approved):
Clean rebase on latest main. Failure-class-protocol issue resolved — the fix(merge) commit is gone. Advisory scope, test-discovery ratchet, sync annotation ratchet. LGTM, merging.

Git-on-my-level (commented):
Thanks for the cleanup on the latest head (`e7cab1b8d3738beb8323130432f565ad6ff0a625`). I re-reviewed the workflow/guardrail changes and this looks like a positive signal from automation, not a formal approval.

What I verified locally:
- `python3 .github/scripts/test_check_pr_scope.py` -> 9 passed
- `cd backend && .venv/bin/python -m pytest -q tests/unit/test_check_unit_test_discovery.py scripts/check_unit_test_discovery.py --tb=short` -> 14 passed
- `cd backend && python3 scripts/check_unit_test_discovery.py` -> unit test discovery OK, 0 orphans
- `python3 .github/scripts/run_checks.py --lane local --base origin/main --head HEAD --pr-body-file /tmp/omi-pr9634-review-20260729/pr-body.md` progressed through the PR-scope, backend-test-discovery, failure-class, agent-doc, diff-hygiene, and related selected checks; the earlier failure-class blocker is resolved when the current PR body is supplied. My local environment stopped later at a Ruby-dependent Ray-Ban iOS guard (`ruby` not installed here), while GitHub’s current check rollup is green.

Agent/maintainer consequence: this PR changes `AGENTS.md`, so it will alter how AI coding/review agents behave in the repo. The notable behavior changes are: PR size is advisory/reporting-only rather than a hard split gate; backend test discovery becomes mechanically enforced; new fail-closed gates need legacy-principal fallback coverage; tests rewritten in the same PR as asserted code need an external source; and the repo-wide behavior section tells agents to act autonomously while the safety section still requires explicit user sign-off for landing to `main` and risky/wide-blast-radius actions. That guidance appears directionally safe and aligned with the maintainer discussion, but it is still repo-wide agent/workflow policy and should keep human maintainer review before merge.

I’m not formally approving because the approval gate is blocked for workflow/CI guardrails, `AGENTS.md` agent-instruction changes, and the existing `needs-maintainer-review` / `workflow-review` labels.

---

PR #9674, opened 2026-07-14
fix(desktop): keep the push-to-talk query alive when the Ask Omi panel collapses

kodjima33:
## Bug

Speak to Omi via push-to-talk **while the Ask Omi input panel is already open** and your question is silently dropped — the panel collapses, and no answer and no error ever appear.

Regression from `fa0046a322` ("unify PTT listening chrome and drop agent voice follow-up"), which is on `main` but **not yet in a released build** (no `v*-macos` tag contains it), so this lands before it reaches users.

## Root cause

That commit replaced the guarded PTT cancel in `closeAIConversation()`:

```swift
-        if state.isVoiceFollowUp {
-            PushToTalkManager.shared.cancelListening()
-        }
+        PushToTalkManager.shared.cancelListening()
```

on the reasoning (stated in a sibling comment) that `cancelListening()` is "already guarded by `state != .idle`". That guard is **not a microphone guard** — `isIdle` is `activeTurnID == nil`, which stays false for the *entire* voice turn, including `.awaitingResponse` after the transcript has been committed.

`openAIInputWithQuery(fromVoice:)` collapses the input panel **before** it dispatches the voice query, so the close path now kills the very turn it is about to route:

1. Panel open (`showingAIConversation == true`, no response yet) → user holds PTT and speaks. `isCurrentSessionFollowUp` is `false`, so release takes the `openAIInputWithQuery` path.
2. `sendTranscript` → `.transcriptionFinal` → phase `.awaitingResponse` → `sendQuery` → `openAIInputWithQuery(fromVoice: true, voiceTurnID:)`.
3. Inside it: `if window.state.showingAIConversation { window.closeAIConversation() }` → unconditional `cancelListening()` → turn is not idle → `stopListening()` → `.cancel` → `terminal` (the coordinator drains its FIFO synchronously).
4. The next statement's fence — `guard VoiceTurnCoordinator.shared.requireCurrentOwner(for: voiceTurnID) != nil` — now fails, so `routeQuery` never runs. The query is gone.

Same cause also killed an in-flight voice turn on Esc / click-away while the answer was still generating.

## Fix

Cancel PTT only while the mic is **still capturing**. The decision lives on the authoritative reducer rather than at the call site:

- `VoiceTurnCoordinator.isCapturingAudio` = `activeTurn?.phase.isRecording == true`
- `PushToTalkManager.isCapturingAudio` projects it
- `closeAIConversation()` gates its cancel on it

A committed turn awaiting its answer now survives surface teardown; an in-progress capture is still cancelled when the chat closes, which is what `fa0046a322` was after.

## Verification

- `xcrun swift build -c debug --package-path Desktop` → **Build complete**
- `xcrun swift test --filter VoiceTurn --filter PushToTalk --filter PTT --filter FloatingControlBarStateTests` → **all green**, including the new `testIsCapturingAudioIsFalseOnceTheTranscriptIsCommitted` (asserts a committed turn is active-but-not-capturing — the exact state the old code cancelled).
- **UNVERIFIED in-app:** reproducing the user path needs real mic speech, which this automated run could not drive (`ptt_start`/`ptt_stop` with no audio ends the turn with a hint before it ever commits). The dropped-query path was traced end to end in source instead. Worth one manual pass: open Ask Omi, speak, confirm the answer appears.

🤖 automated by hourly watchdog; opened for review, not merged (the regression is not yet in a released build, so it fails the prod-bug auto-merge gate).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9674?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift
@@ -1192,8 +1192,14 @@ class FloatingControlBarWindow: NSPanel, NSWindowDelegate {
         responseHeightCancellable = nil
         state.responseContentHeight = 0
 
-        // Cancel PTT if listening while chat closes
-        PushToTalkManager.shared.cancelListening()
+        // Cancel PTT only while the mic is still capturing. A turn that has already committed
+        // its transcript is awaiting its answer, and `openAIInputWithQuery` collapses this
+        // surface *before* dispatching the voice query — an unconditional cancel there
+        // terminates the turn, so `routeQuery` is fenced out and the user's spoken question is
+        // silently dropped.
+        if PushToTalkManager.shared.isCapturingAudio {
+            PushToTalkManager.shared.cancelListening()
+        }
 
         OmiMotion.withGated(.easeOut(duration: 0.08)) {
             state.showingAIConversation = false
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift
@@ -95,6 +95,10 @@ class PushToTalkManager: ObservableObject {
   /// A projection of the authoritative reducer. This manager owns microphone and
   /// provider I/O only; it never stores a second logical lifecycle state.
   var phase: VoiceTurnPhase? { voiceTurnCoordinator.activeTurn?.phase }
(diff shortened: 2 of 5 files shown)


kodjima33:

PR #9675, opened 2026-07-14
fix(macos): preserve PTT spawn authority through barges

Git-on-my-level:
## Summary

- Preserve the kernel-owned accepted `spawn_agent` journal authority when a PTT turn is barged or its provider closes.
- Use that immutable authority during delayed finalization, so the app refreshes the already-recorded delegation instead of appending a second unstructured provider exchange.
- Add bounded local diagnostics identifying the last tool, tool error code, and whether an accepted spawn existed when a provider close terminalizes PTT.
- Keep the generated realtime `list_agent_sessions` schema explicit that it discovers subagents and floating-bar sessions.
- Preserve PTT context admission while the realtime hub warms, and keep direct conversational recall content-neutral: it relies on canonical recent turns, with no password- or credential-specific policy.
- Remove the obsolete observability work audit; the broader documentation lifecycle cleanup is tracked in #9681.

The realtime hub kept accepted spawn ownership only in mutable transport tracking. A barge-in or provider teardown cleared that tracking while a delayed journal-finalization task was still resolving the transcript. The task then treated an already-accepted delegation as an ordinary provider response, which could duplicate the exchange and separate the child lifecycle from the PTT acknowledgment.

The interrupted-turn payload and completed-turn finalization now snapshot the accepted spawn owner before transport cleanup. The regression test drives that captured payload through the real provider-failure continuity and journal-authority policies, asserting that it refreshes the canonical spawn rather than records a second exchange.

The hub-warmup follow-up reserves the reducer effect identity before draining FIFO work, so a valid PTT context update cannot be rejected merely because it was admitted during an active drain. The policy follow-up retains the generic direct-recall requirement while avoiding content-specific memory rules. The broader shared context/cache contract is tracked in #9680.

The desktop Swift CI suite also exposed a drift in the canonical realtime tool description: the list-sessions tool remained present, but its generated schema no longer told the model that it discovers subagents. The manifest source now states that contract explicitly; its generated Swift and fixture surfaces are refreshed, while `HubSystemInstructionTests` exercises the emitted schema and the tool-surface test keeps generated outputs in sync.

- INV-AGENT-*
- INV-CHAT-1
- INV-VOICE-1

## Verification

- `cd desktop/macos/agent && npm test -- context-snapshot.test.ts` (17 passed)
- `cd desktop/macos && xcrun swift test -c debug --package-path Desktop --filter RealtimeHubBargeInContinuityTests` (53 passed)
- `cd desktop/macos && xcrun swift test -c debug --package-path Desktop --filter HubSystemInstructionTests` (12 passed)
- `cd desktop/macos/agent && node --experimental-strip-types scripts/generate-tool-surfaces.mjs --check && node node_modules/vitest/vitest.mjs run tests/tool-surfaces-exhaustiveness.test.ts` (9 passed)
- `cd desktop/macos && ./scripts/agent-logic-harness.sh --cross-surface-smoke` (Swift/Node/Rust contract gates passed)
- `cd desktop/macos && OMI_APP_NAME="omi-ptt-spawn-recovery" ./run.sh --yolo` (named dev-backend bundle built and launched; main UI rendered via accessibility snapshot)
- enforced `git push` pre-push gate: 14 manifest checks, desktop debug build, and 104 desktop tool-surface tests passed.


Git-on-my-level:
@cursor address merge conflicts and review/fix any issues

PR #9677, opened 2026-07-14
fix(desktop): keep PTT dead-mic recovery armed across zero-frame turns (#9081)

### LINKED ISSUE #9081: macOS PTT hold mode records silent turns for 169 users with no recovery attempted
## Summary

The 2026-07-05 Desktop health telemetry shows `ptt_audio_capture_silent_turn` affecting **169 users** on macOS 0.12.0 in hold mode. The event reports `recovery_action=none` and `recovery_result=not_attempted`.

This is high-priority because users pressing/holding PTT expect voice input to be captured. Silent turns are direct UX failure even if the app does not crash.

## Evidence

Source: `Omi Improvements Daily Report — 2026-07-05`, `posthog_desktop_health_breakdown`.

| app version/build | health_event | mode | input_device_class | events | users | recovery_action | recovery_result | first seen | last seen |
|---|---|---|---|---:|---:|---|---|---|---|
| 0.12.0 / 12000 | `ptt_audio_capture_silent_turn` | hold | unknown | **1,104** | **169** | none | not_attempted | 2026-07-03T06:46:14Z | 2026-07-05T13:11:12Z |

Daily headline:

- `PTT silent/watchdog users`: **42** current day vs **0** 7d ago
- `PTT end/start`: **87.4%** current day vs **97.9%** 7d ago (**-10.5pp**)

## Why this is not covered by an existing issue

No exact open issue was found for `ptt_audio_capture_silent_turn`, `PTT silent`, or `silent turn`.

Adjacent but too broad:

- #8846 covers timing-based coordination and async delays in Desktop, which may be a root-cause class but does not track this telemetry regression.

## Suggested investigation

1. Break down by input device, permission state, selected mic, audio engine state, and realtime provider/token state.
2. Determine whether silent turns correlate with token mint failures or local audio capture startup failures.
3. Add immediate recovery: restart capture session, prompt mic permission/device selection, or surface a clear user-visible error instead of silently failing.
4. Improve instrumentation so `input_device_class=unknown` is rare.

## Acceptance criteria

- Silent PTT turns have an actionable reason or recovery path.
- `recovery_action=none/not_attempted` is eliminated for recoverable cases.
- PTT end/start ratio returns near previous baseline.

kodjima33:
## Bug

Fixes #9081. Desktop health telemetry shows `ptt_audio_capture_silent_turn` on macOS 0.12.0 hold mode affecting **169 users / 1,104 events**, every one of them reporting `recovery_action=none`, `recovery_result=not_attempted`. PTT has a dead-mic self-heal — it simply never fired for the users who needed it, so a dead microphone stayed dead turn after turn with no recovery attempted.

## Root cause

`PTTSilentMicRecoveryPolicy.recordDiscardedTurn` (`PushToTalkManager.swift`) decides whether consecutive silent turns should trigger a CoreAudio capture rebuild:

```swift
if totalSec >= Self.minDeadTurnSeconds && peak <= Self.deadMicPeakThreshold {
  consecutiveDeadMicTurns += 1
} else {
  consecutiveDeadMicTurns = 0   // <-- any turn under 0.25s lands here
}
```

Every turn shorter than `minDeadTurnSeconds` (0.25s) fell into the `else` branch and **reset the dead-mic streak**. Crucially that includes turns which captured *no frames at all*: `startMicCapture` starts CoreAudio asynchronously (`AudioDeviceStart` is blocking HAL IPC to `coreaudiod`), so a key release that beats capture start finalizes a turn with `totalSec = 0, peak = 0`. Such a turn says nothing about mic health, but it erased the evidence — so with a dead mic interleaved with ordinary taps, `consecutiveDeadMicTurns` never reached the threshold of 2 and recovery was never requested. All four `recordPTTSilentTurn` call sites (hub, omni, batch, warm-wait fallback) derive `recovery_action`/`recovery_result` from this one return value, which is why the field data is uniformly `none`/`not_attempted`.

The failure class: **absence of evidence was treated as evidence of health.**

## Fix

Make each outcome carry only the evidence it actually has:

- audible input (`peak > deadMicPeakThreshold`) proves the mic is alive → reset the streak
- a judgeable silent turn (`totalSec >= minDeadTurnSeconds`, near-zero peak) → increment
- a too-short / zero-frame turn → **neutral**, leave the streak intact

Three-line behavior change in the policy struct; no call sites, thresholds, or telemetry fields changed.

## Verification

- **Regression test** `PTTSilentMicRecoveryPolicyTests.testZeroFrameTurnDoesNotEraseDeadMicStreak` reproduces the field scenario (dead turn → zero-frame turn → dead turn). It **fails on the old code** (`consecutiveDeadMicTurns` 0 instead of 1; recovery never returns `true`) and passes on the fix. Added `testAudibleShortTapResetsDeadMicStreak` to pin the other direction.
- `xcrun swift test --package-path Desktop --filter "PTT|PushToTalk|VoiceTurn|DesktopDiagnostics"` → **65 tests, 0 failures** (incl. `PTTAudioCaptureRaceTests`, `PTTSilentTurnRecoveryWiringTests`, `PushToTalkStateMachineTests`, `DesktopDiagnosticsManagerTests`).
- `xcrun swift build -c debug --package-path Desktop` → green.
- `make preflight` → green.

Hermetic unit coverage of a pure policy struct; not exercised against a physically dead microphone on prod hardware.

## Product invariants

Path-matched by `desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift`:

- **INV-CHAT-1** — unchanged. No chat write-path, provider, projection, or viewport code is touched; no message store or turn-recording behavior changes.
- **INV-VOICE-1** — preserved and strengthened. The voice-turn state machine, capture lifecycle, and silence gating are untouched; the only change is when the existing dead-mic recovery is allowed to arm, so a silent capture is more likely to self-heal rather than less.

## Follow-up (not in this PR)

The zero-frame turns themselves come from the release beating asynchronous CoreAudio capture start. Closing that race means changing the turn state machine's capture-start/finalize ordering — a wider blast radius than this fix warrants, and it remains tracked by #9081's investigation items.

🤖 automated by the hourly desktop bug watchdog.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9677?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift
@@ -12,11 +12,15 @@ struct PTTSilentMicRecoveryPolicy {
   private(set) var consecutiveDeadMicTurns = 0
 
   mutating func recordDiscardedTurn(totalSec: TimeInterval, peak: Int) -> Bool {
-    if totalSec >= Self.minDeadTurnSeconds && peak <= Self.deadMicPeakThreshold {
-      consecutiveDeadMicTurns += 1
-    } else {
+    if peak > Self.deadMicPeakThreshold {
+      // Audible input proves the mic is alive, whatever else went wrong with the turn.
       consecutiveDeadMicTurns = 0
+    } else if totalSec >= Self.minDeadTurnSeconds {
+      consecutiveDeadMicTurns += 1
     }
+    // A turn too short to judge carries no evidence either way — an accidental tap, or a
+    // release that beat CoreAudio's capture start and delivered no frames at all. It must
+    // not erase a dead-mic streak, or a dead mic interleaved with taps never recovers.
     return consecutiveDeadMicTurns >= Self.consecutiveDeadTurnThreshold
   }
 
--- a/desktop/macos/Desktop/Tests/PTTSilentMicRecoveryPolicyTests.swift
+++ b/desktop/macos/Desktop/Tests/PTTSilentMicRecoveryPolicyTests.swift
@@ -46,4 +46,27 @@ final class PTTSilentMicRecoveryPolicyTests: XCTestCase {
     XCTAssertFalse(policy.recordDiscardedTurn(totalSec: 1.0, peak: 0))
     XCTAssertEqual(policy.consecutiveDeadMicTurns, 1)
   }
+
+  /// A turn that captured no frames (release beat CoreAudio's capture start) is silent for
+  /// reasons unrelated to mic health. Treating it as proof of a live mic reset the streak, so a
+  /// dead mic interleaved with taps never reached the recovery threshold (#9081).
+  func testZeroFrameTurnDoesNotEraseDeadMicStreak() {
+    var policy = PTTSilentMicRecoveryPolicy()
+
+    XCTAssertFalse(policy.recordDiscardedTurn(totalSec: 1.0, peak: 0))
+    XCTAssertFalse(policy.recordDiscardedTurn(totalSec: 0, peak: 0))
+    XCTAssertEqual(policy.consecutiveDeadMicTurns, 1)
+
+    XCTAssertTrue(policy.recordDiscardedTurn(totalSec: 1.0, peak: 0))
+  }
+
+  func testAudibleShortTapResetsDeadMicStreak() {
+    var policy = PTTSilentMicRecoveryPolicy()
+
+    XCTAssertFalse(policy.recordDiscardedTurn(totalSec: 1.0, peak: 0))
+    XCTAssertFalse(
+      policy.recordDiscardedTurn(
+        totalSec: 0.05, peak: PTTSilentMicRecoveryPolicy.deadMicPeakThreshold + 1))
+    XCTAssertEqual(policy.consecutiveDeadMicTurns, 0)
+  }
 }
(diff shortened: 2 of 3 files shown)


kodjima33:
## CI status: the one red check is pre-existing on `main`, not from this PR

`Desktop Swift Build & Tests` fails on `HubSystemInstructionTests.testRealtimeListAgentSessionsToolIsExposed` — realtime hub tool exposure, untouched by this diff (which only changes `PTTSilentMicRecoveryPolicy` + its tests).

Reproduced on unmodified `origin/main` (`fec07c184f`, "fix(macos): harden shared chat and PTT agent continuity (#9660)"), with no changes applied:

```
xcrun swift test --package-path Desktop --filter HubSystemInstructionTests/testRealtimeListAgentSessionsToolIsExposed
Test Case '-[Omi_ComputerTests.HubSystemInstructionTests testRealtimeListAgentSessionsToolIsExposed]' failed (1.284 seconds)
Executed 1 test, with 1 failure
```

So **desktop CI on `main` is currently red** and every desktop PR will inherit this failure until it is fixed. Worth a separate fix.

Everything this PR does own is green: 65 tests across `PTT|PushToTalk|VoiceTurn|DesktopDiagnostics` pass, debug build + release compile pass, `make preflight` passes.

Per the watchdog's auto-merge gate (which requires all required checks green before a prod deploy), this is **left as a PR and not merged**.

Issue #9440, opened 2026-07-11
Make local pre-push and GitHub CI deterministic and parity-driven


Git-on-my-level:
## Problem

PR #9437 exposed a confidence gap between local verification and GitHub Actions. The immediate failure was not a Swift product regression: `desktop/macos/tests/test-swift-test-suites.sh` inferred parallel execution from the total wall-clock duration of the full suite runner.

The test gives two fake suites 3-second sleeps and fails when the whole runner takes 6 seconds or more. PR #9437 added the desktop-quality scan to the runner's setup path, adding about 1.9 seconds locally. The test passed once during pre-push, then failed unchanged locally at `real 5.70s` / integer `elapsed=6s`; GitHub's macOS runner reported 8 seconds. The actual Swift test step was skipped after this launcher-test failure, while the clean release compile passed.

This is primarily nondeterministic test design, but it also highlights real parity gaps:

- local branch head vs GitHub's synthetic PR merge commit
- mature/warm local checkout vs clean hosted checkout and restored caches
- local macOS 26.5.1 / Xcode 26.6 / Swift 6.3.3 vs GitHub macOS 15.7.7 on a moving `macos-15-arm64` image
- CI behavior distributed across workflow YAML, component scripts, and `scripts/pre-push`
- conditional checks derived independently from local and PR diff contexts
- Swift cache keys that omit `Package.resolved`, architecture, Xcode/Swift version, and runner-image identity

### 1. Deterministic concurrency and timeout tests

- Replace absolute elapsed-time assertions with rendezvous/barrier or active-worker instrumentation.
- For the immediate Swift scheduler test, make Alpha and Beta prove they were simultaneously active rather than comparing total runtime to 6 seconds.
- Isolate scheduler tests from unrelated preflight/static scans through dependency injection or narrowly scoped runner modes.
- Use wall-clock bounds only as outer deadlock guards, never as the behavioral oracle.
- Do not use automatic reruns to turn first-attempt flakes green; reruns may classify infrastructure failures but should preserve the original failure.

### 2. Canonical CI entrypoints

Introduce one repository-owned entrypoint per required job, for example:

- `scripts/ci/desktop-swift`
- `scripts/ci/backend-unit`
- `scripts/ci/backend-static`

GitHub workflows and local pre-push should invoke the same entrypoints. Workflow YAML should provision the environment and call the script, not duplicate selection or test logic.

### 3. Local GitHub-parity mode

Provide an explicit command that:

[...]

PR #9683, opened 2026-07-14
fix(preflight): normalize Firestore query paths

### LINKED ISSUE #9682: preflight: canonicalize Firestore query fingerprints on Windows
## Problem

`make preflight` fails on Windows for any change matching `backend/**/*.py`, even when no Firestore query code changed.

The `firestore-query-coverage` check reports every existing raw or unsupported query fingerprint as new. On the current `main`, that is 42 raw shapes and 3 unsupported shapes.

## Root cause

`backend/scripts/firestore_query_coverage.py::inventory()` passes `str(relative)` into `QueryShape.source`. The source string is part of `QueryShape.fingerprint`, so Windows uses paths such as `backend\database\action_items.py` while the committed Linux-generated baseline uses `backend/database/action_items.py`.

Observed on Windows:

- current fingerprints matching the committed baseline: `0/45`
- fingerprints matching after normalizing `source` to POSIX separators: `45/45`

This also makes waiver IDs platform-dependent because waivers use the same fingerprint.

## Reproduction

From a Windows checkout of `main`:

```powershell
$env:PYTHONUTF8 = '1'
make preflight
```

The check fails with `new unregistered serving compound query shape(s)` and `new unsupported serving compound query shape(s)` for untouched backend queries.

## Proposed guard

Canonicalize all source paths with `PurePath.as_posix()` before constructing `QueryShape`, and add a platform-independent regression test comparing fingerprints produced from `PureWindowsPath` and `PurePosixPath` inputs.

Contributes to #9440.

tianmind-studio:
## Summary

- canonicalize Firestore query inventory source paths before they enter fingerprints
- use the same canonical path for AST diagnostics, runtime query shapes, and registry shapes
- add a platform-independent Windows/Posix fingerprint regression test

## Root cause

`inventory()` passed `str(relative)` into `QueryShape.source`, and `source` is part of the fingerprint. Linux generated `backend/database/...` while Windows generated `backend\database\...`, so all 45 existing raw or unsupported fingerprints looked new on Windows. Waiver IDs were platform-dependent for the same reason.

The old code reproduced both failures on Windows:

- `firestore_query_coverage.py --check-ratchet` reported 42 new raw shapes and 3 new unsupported shapes
- `test_query_coverage_baseline_tracks_current_raw_and_unsupported_debt` failed against the committed baseline

## Durable guard

All inventory paths now pass through `PurePath.as_posix()` before constructing a `QueryShape`. The regression test feeds equivalent `PureWindowsPath` and `PurePosixPath` values through the analyzer and requires identical source strings and fingerprints. The full baseline test also passes on Windows.

## Validation

- `python backend/scripts/firestore_query_coverage.py --check-ratchet` - passed
- `python -m pytest backend/tests/unit/test_firestore_query_contract.py::test_query_coverage_baseline_tracks_current_raw_and_unsupported_debt -q -m slow` - 1 passed
- official `backend/test.sh` selected-file runner - 6 passed, 2 deselected
- `black --check --line-length 120 --skip-string-normalization ...` - passed
- `backend/scripts/typecheck.sh` - 0 errors
- `$env:PYTHONUTF8='1'; make preflight` - 14 checks passed

Fixes #9682.
Contributes to #9440.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9683?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/firestore_query_coverage.py
+++ b/backend/scripts/firestore_query_coverage.py
@@ -19,7 +19,7 @@ import json
 import sys
 from dataclasses import dataclass, field
 from datetime import date
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import Any, Iterable, Mapping
 
 ROOT = Path(__file__).resolve().parents[2]
@@ -32,6 +32,10 @@ BASELINE_PATH = BACKEND_ROOT / 'scripts' / 'firestore_query_coverage_baseline.js
 WAIVERS_PATH = BACKEND_ROOT / 'scripts' / 'firestore_query_coverage_waivers.json'
 QUERY_SPEC_SOURCE = Path('backend/database/firestore_index_registry.py')
 NON_SERVING_SEGMENTS = {'tests', 'scripts', 'migrations', 'migration'}
+
+
+def _canonical_source(path: PurePath) -> str:
+    return path.as_posix()
 
 
 @dataclass(frozen=True)
@@ -353,11 +357,12 @@ def inventory(*, waiver_ids: set[str]) -> list[QueryShape]:
             part.startswith('.') or part in {'venv', '__pycache__'} for part in backend_relative.parts
         ):
             continue
-        tree = ast.parse(path.read_text(encoding='utf-8'), filename=str(relative))
+        source = _canonical_source(relative)
+        tree = ast.parse(path.read_text(encoding='utf-8'), filename=source)
         constants = _module_constants(tree)
         non_serving_scope = _non_serving_scope(relative)
         module_analyzer = FunctionQueryAnalyzer(
-            source=str(relative),
+            source=source,
             symbol='<module>',
             constants=constants,
             non_serving_scope=non_serving_scope,
@@ -370,7 +375,7 @@ def inventory(*, waiver_ids: set[str]) -> list[QueryShape]:
             if not isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
                 continue
             analyzer = FunctionQueryAnalyzer(
-                source=str(relative),
+                source=source,
                 symbol=node.name,
                 constants=constants,
                 non_serving_scope=non_serving_scope,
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Thanks for tracking this down. I reviewed the diff and this looks like the right shape of fix: canonicalizing the inventory `source` with `PurePath.as_posix()` before it enters both the AST filename and `QueryShape.source` should remove the Windows-vs-POSIX separator from fingerprints, including registry-sourced shapes. The added regression test covers equivalent `PureWindowsPath` / `PurePosixPath` inputs producing the same source string and fingerprint.

I also ran the ratchet locally against the PR checkout with secrets stripped:

- `python backend/scripts/firestore_query_coverage.py --check-ratchet` — passed

I could not run the targeted pytest in this stripped environment because `google.cloud.firestore_v1` is not installed here, but GitHub's Backend unit suite is passing on the PR.

No blocking code issue from me. I am leaving this as a positive signal rather than a formal approval because the PR changes a preflight/static-contract script; a maintainer should still sanity-check the workflow/guardrail behavior before merge.

kodjima33 (approved):
Preflight path normalization with a contract test; holding the merge on the maintainer-review label.

tianmind-studio:
Fresh integration evidence from latest main: rebasing the #9440 pusher circuit-breaker fake-clock change and running `make preflight` failed only at `firestore-query-coverage` with the same 42 raw plus 3 unsupported Windows fingerprints. Applying this PR commit in a clean temporary worktree made all 14 selected checks pass, so #9683 also unblocks the next deterministic-test contribution.

tianmind-studio:
Current `main` now contains the same platform-independent source-path guard through `3f24998a0`: `canonical_source_path()` uses `PurePath.as_posix()`, all inventory/registry sources pass through it, and a `PureWindowsPath`/`PurePosixPath` regression test is present. I rechecked that implementation on native Windows: the focused contract test passed and `firestore_query_coverage.py --check-ratchet` passed. Closing this PR because the original gap is resolved on main.

PR #9684, opened 2026-07-14
fix(desktop): recover PTT turns rejected by realtime context admission

kodjima33:
## Bug

A spoken push-to-talk question can be **silently dropped** — the user holds PTT, speaks a full question, releases, and gets **no answer, no transcript, and no error**. The captured audio is discarded.

## Root cause

`dbd69e71c5` ("make realtime context admission fail closed") made context admission fail-closed and routed all three *asynchronous* rejection sites through `.providerReconnectFailed`:

- `RealtimeHubController.finishContextFreshInputOnCurrentSession()`
- `RealtimeHubController.finishSessionReconnectAfterReady()`
- `RealtimeHubController.failContextFreshInputPreparation()`

`.providerReconnectFailed` means **transport failure**, and the reducer maps it to `terminate(reason: .providerFailed)` — the turn dies with no transcription fallback.

But an admission rejection is a **context/policy decision, not a transport failure**. The buffered audio never reached the provider and is still intact in `batchAudioBuffer`, so the turn is recoverable. The follow-up `711b3a13af` restored exactly that bounded Deepgram fallback — but only for the **synchronous** rejection (`.hubAdmissionRejected`), which fires *before* `.providerReconnectStarted` projects the turn. The async half stayed fatal.

Reachable without a race: `.rejectStaleProviderContext` fires on any owner-scope skew between the context bind and the socket build (`sessionVoiceContextFreshnessIdentity` is reset to `""` by `teardownSession()` and re-derived empty when `prefetchedVoiceContextOwnerScope != ownerScope`).

## Fix

Add `.providerContextAdmissionRejected` for the three sites that discard buffered input *before it ever reached the provider*. The reducer:

1. undoes the `awaitingResponse` projection that `providerReconnectStarted` applies to a released turn (the transcription fallback only runs from an input phase — otherwise the turn would hang instead of dying, which is worse);
2. takes the same bounded transcription fallback as the synchronous case.

A phase with no buffered input to rescue keeps the terminal behavior, and a genuine transport failure (`hubDidError`) still terminates via `.providerReconnectFailed`.

**Closing the class:** the failed contract was conflating *admission rejection* with *transport failure* on one event. Both halves of the rejection surface now map to the same recovery, and `.providerReconnectFailed` is left to mean only what it says.

## Verification

- `xcrun swift build -c debug --package-path Desktop` — green.
- `VoiceTurnReducerTests` — **57/57 pass**, including 2 new regression tests: a released turn transcribes instead of dying; a still-held turn keeps recording on the transcription route.
- **Teeth check:** with the reducer forced back to `terminate()`, both new tests fail (10 assertions); with the fix they pass.
- **Not exercised in a live app** — the rejection needs a provider-side race / stale-context condition I cannot reproduce on demand. Coverage is behavioral through the production reducer API (a controllable seam), not a source grep.

## Scope

4 files, +114/−3. Unreleased code (`dbd69e71c5` is not in `v0.12.72`, the current prod build), so this is **not live on prod** — opened for review, **not merged**.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9684?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

- INV-CHAT-1
- INV-VOICE-1

### DIFF
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift
@@ -3139,7 +3139,7 @@ final class RealtimeHubController: NSObject, RealtimeHubSessionDelegate {
       reconnectAudioBuffer = nil
       live.abandonInputTurn()
       VoiceTurnCoordinator.shared.send(
-        .providerReconnectFailed(
+        .providerContextAdmissionRejected(
           turnID: pending.turnID,
           identity: pending.identity,
           message: "realtime context admission rejected: \(admission)"))
@@ -3213,7 +3213,7 @@ final class RealtimeHubController: NSObject, RealtimeHubSessionDelegate {
     guard VoiceTurnCoordinator.shared.activeTurnID == turnID else { return }
     session?.abandonInputTurn()
     VoiceTurnCoordinator.shared.send(
-      .providerReconnectFailed(
+      .providerContextAdmissionRejected(
         turnID: turnID,
         identity: pending.identity,
         message: message))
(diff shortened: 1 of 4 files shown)


kodjima33:

PR #9697, opened 2026-07-14
fix(recommendations): treat a malformed stored projection as absent instead of 500ing the read

ZachL111:
## What

`get_evaluation_projection` validated the stored projection with an unguarded call:

```python
raw_projection = payload.get('projection')
if not isinstance(raw_projection, dict):
    return None
projection = WhatMattersNowProjection.model_validate(raw_projection)
...
```

Every other unusable-projection case in this reader already returns `None` (doc missing, wrong account_generation, non-dict projection). But a projection dict that fails `WhatMattersNowProjection` validation (a legacy or schema-drifted doc) raises `ValidationError`, which propagates and 500s the recommendation read (`utils/task_intelligence/recommendations.py:966`).

## Fix

Move the validate plus freshness checks into a pure `_valid_evaluation_projection` helper that returns `None` on a malformed projection instead of raising, consistent with the reader's other `None` returns:

```python
def _valid_evaluation_projection(raw_projection, evaluation_id, now):
    if not isinstance(raw_projection, dict):
        return None
    try:
        projection = WhatMattersNowProjection.model_validate(raw_projection)
    except ValidationError as e:
        logger.warning('Ignoring malformed stored projection for evaluation %s: %s', evaluation_id, e)
        return None
    if projection.evaluation_id != evaluation_id or projection.expires_at <= now:
        return None
    return projection
```

The catch is narrow (`ValidationError`) so an unexpected error still propagates. Extracting the helper also makes the behavior unit-testable without the surrounding `@firestore.transactional` read, matching the `_decision_records` helper added in #9606 for the sibling `get_decisions`.

## Tests

New `tests/unit/test_evaluation_projection_skip_malformed.py` drives the helper directly:
- a valid projection is returned; a malformed one returns None (no 500)
- a non-dict, an id mismatch, and an expired projection each return None

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_evaluation_projection_skip_malformed.py -q
  -> 4 passed
black --line-length 120 --skip-string-normalization --check database/task_recommendations.py tests/unit/test_evaluation_projection_skip_malformed.py
  -> unchanged
python -m pyright -p pyrightconfig.json database/task_recommendations.py -> 0 errors
python scripts/check_module_stub_pollution.py -> 0 violations
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9697?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/task_recommendations.py
+++ b/backend/database/task_recommendations.py
@@ -366,6 +366,27 @@ def get_decisions(
     return _decision_records(raw_records, evaluation_id)
 
 
+def _valid_evaluation_projection(
+    raw_projection: Any, evaluation_id: str, now: datetime
+) -> Optional[WhatMattersNowProjection]:
+    """Build a WhatMattersNowProjection from a stored projection dict, or None if unusable.
+
+    The stored projection was validated with no guard, so a legacy or schema-drifted doc would raise
+    ValidationError and 500 the recommendation read. Treat a malformed (or non-dict, stale, or
+    id-mismatched) projection as absent, consistent with this reader's other None returns.
+    """
+    if not isinstance(raw_projection, dict):
+        return None
+    try:
+        projection = WhatMattersNowProjection.model_validate(raw_projection)
+    except ValidationError as e:
+        logger.warning('Ignoring malformed stored projection for evaluation %s: %s', evaluation_id, e)
+        return None
+    if projection.evaluation_id != evaluation_id or projection.expires_at <= now:
+        return None
+    return projection
+
+
 def get_evaluation_projection(
     uid: str,
     evaluation_id: str,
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for the focused fix. I reviewed the changed backend reader path and the new unit coverage.

This direction looks sound to me: malformed stored recommendation projections are treated the same as the reader's existing absent/unusable cases, while unexpected non-validation errors still propagate. The helper extraction also makes the edge cases easy to cover, and the tests exercise malformed, non-dict, mismatched-id, expired, and valid projections.

I’m not formally approving because this touches backend recommendation persistence/data-handling behavior and the PR is still in a blocked/review-required state in GitHub. A human maintainer should give the final review before merge, and the cancelled/blocked required checks may need to be re-run or cleared.

No blocking code issue found from my review.

kodjima33 (approved):

kodjima33 (approved):

Issue #9355, opened 2026-07-09
Conversation sync: add durable mutations, change feed, and tombstones


Git-on-my-level:
## Context

Issue #8197 identified recurring Conversations bugs caused by ambiguous ownership between SwiftUI state, SQLite, and server responses.

The 80/20 fix is in #9354: one desktop repository, Firestore-backed revisions, explicit cache completeness, canonical mutation responses, and transactional preservation of user-owned fields. That deliberately stops short of a full sync protocol.

This issue tracks the platonic long-term architecture.

## Goal

Make conversation sync an explicit, account-scoped replicated-state protocol that remains correct across offline edits, retries, process crashes, concurrent post-processing, deletes, multiple devices, and large histories.

### 1. Server-owned revision and field ownership contract

- define a documented ownership table for every conversation field
- expose an opaque monotonic revision on every snapshot and mutation response
- enforce user-owned vs processing-owned fields in one database write boundary
- prohibit direct conversation document writes outside that boundary with a guard test

### 2. Idempotent mutation API

Every user mutation accepts:

- `client_mutation_id`
- `base_revision`
- the typed operation/payload

The server deduplicates retries and returns either:

- the exact canonical post-write snapshot/revision, or
- a typed conflict containing current canonical state

Deletes use durable tombstones so a concurrent processing retry cannot resurrect a conversation.

### 3. Cursored change feed

Add an account-scoped endpoint such as:

`GET /v1/conversations/changes?cursor=...`

It returns ordered upserts, tombstones, the next cursor, and an explicit reset/full-resync response when history is no longer available. Normal startup/activation should consume deltas rather than repeatedly scanning the newest 50 rows.

### 4. Desktop projection + durable outbox

SQLite becomes a materialized projection with:

- server revision
- projection completeness
- tombstone state
- last applied change cursor
- durable pending mutation outbox

Applying a change page and advancing its cursor must be one SQLite transaction. Optimistic UX is derived from the durable outbox, survives restart/offline periods, and disappears only after canonical acknowledgement or a surfaced conflict.

### 5. Separate local capture lifecycle

Split local recording/upload state from cached server conversation state:

- `LocalCaptureSession`: recording, upload, retry, local segments
- `CachedConversation`: server projection and revision
- explicit linkage from capture to canonical conversation ID

This removes the current overloaded transcription row whose local lifecycle timestamps can be mistaken for server freshness.

### 6. Observability and rollout

Add bounded metrics/events for:

[...]

PR #9700, opened 2026-07-14
backend: make listen conversation lifecycle durable

### LINKED ISSUE #9351: Introduce durable recording sessions for listen lifecycle
## Context

Follow-up to the Phase 1 safety fix in #9349.

Phase 1 makes the desktop-provided recording identity authoritative and prevents stale global pointers, late lifecycle events, and stale persisted bindings from attaching a recording to the wrong conversation. It intentionally preserves the existing client-ID/conversation-ID compatibility model.

## Goal

Make a recording session a first-class, durable server resource so a conversation is the result of an explicit mapping—not an inference from a user-global pointer, timestamps, or WebSocket ordering.

## Design requirements

- Introduce an immutable `recording_session_id`, accepted as an idempotency key from supported clients or created by the server, and scoped to its owner.
- Persist the mapping from recording session to canonical conversation, plus explicit lifecycle state and attempt/generation metadata.
- Route audio/transcript upload attempts and recovery work through that recording session; retries must not create or attach to another session.
- Define a versioned lifecycle-event envelope containing at least `recording_session_id`, `conversation_id`, phase, and monotonic sequence or generation. Clients must discard stale/out-of-order events.
- Make finalization idempotent: repeated finalization returns the same canonical session/conversation result without force-processing an unrelated conversation.
- Roll out behind a feature flag with dual-write/shadow validation; keep the Phase 1 compatibility path until mismatch metrics demonstrate safe cutover.
- Retire the user-global in-progress conversation pointer from identified recording-session routing once the new protocol is fully adopted.

## Verification

- State-machine/property tests covering retries, reconnects, concurrent recordings, delayed events, duplicate finalize requests, and process restart.
- Fault-injection tests proving a session can never be mapped to a different user's conversation or a different recording's conversation.
- Metrics for session-to-conversation mapping conflicts, stale-event discards, finalization idempotency, and legacy-path usage.

## Related

- Phase 1 PR: #9349

### LINKED ISSUE #9516: refactor(backend): converge conversation lifecycle onto a single mutation owner
## Summary

Converge the backend conversation lifecycle onto a single mutation owner. Today the conversation document is the canonical noun everyone agrees on, but the lifecycle verbs — create, claim for processing, transition status, finalize, discard, merge, reprocess — are performed independently by ten different modules. This is the backend's closest analogue of the macOS convergence in #9515 ("centralized nouns, plural verbs"), and it should follow the same playbook: characterization tests first, one exclusive mutation API, migrate every caller, delete every alternative path.

**Sequencing requirement (deliberate):** Phase 1 (characterization/contract tests) lands first as its own PR on `main`, *before* any refactor work. Current lifecycle coverage is edge-focused (listen/pusher e2e + chaos, sync idempotency) and there is no contract suite for the lifecycle state machine itself. The tests are pure additive safety and also protect the ongoing unrelated churn in `transcribe.py` / sync. The refactor itself then lands as one convergence PR per the #9515 execution model.

## Evidence

Churn since 2026-05-15: `utils/conversations/` 83 commits, `routers/transcribe.py` 57, `routers/conversations.py` 47.

Plural mutation authority, verified in the tree:

- **Ten modules write conversation documents directly** through `database/conversations.py`: `routers/transcribe.py` (live WebSocket), `routers/pusher.py`, `routers/conversations.py`, `routers/sync.py`, `routers/developer.py`, `utils/sync/pipeline.py`, `utils/conversations/process_conversation.py`, `utils/conversations/postprocess_conversation.py`, `utils/conversations/merge_conversations.py`, `utils/imports/limitless.py`.
- The mutation API is effectively `update_conversation(uid, id, update_data: dict)` (`database/conversations.py:444`) — a generic dict write. Any caller can set any field, including lifecycle fields. There is no typed state machine for `in_progress → processing → completed / discarded`.
- The scar tissue from past races already exists: `claim_conversation_status()` (`database/conversations.py:878`) is a compare-and-swap added because two writers raced on `status`. This is the lifecycle equivalent of the "save-versus-poll suppression machinery" #9515 identifies — a patch on a race between co-equal writers instead of a single owner.
- Processing is re-entered from at least five directions with independent admission logic: live transcription finalization, sync/backfill pipeline, imports, manual reprocess via router, and merge (`process_conversation.py` is ~60K lines-adjacent of orchestration; `merge_conversations.py` ~22K).

Per AGENTS.md ("Bug Fixes: Close the Failure Class"), the unit of work is the violated contract — single ownership of lifecycle transitions — not any one race.

## Locked target architecture

```text
Event sources (transcribe WS, pusher, sync pipeline, imports, routers, merge)
    |
    | typed lifecycle intents (finalize, reprocess, merge, discard, import)
    v
Conversation lifecycle service (one module)
    - owns the status state machine: in_progress -> processing -> completed | discarded
    - owns processing admission: single claim point, idempotent by conversation id
    - owns finalization/merge/reprocess entry (one orchestrator, callers submit intents)
    - all transitions typed; invalid transitions fail closed
    |
    v
database/conversations.py
    - typed transition functions only for lifecycle fields
    - generic field updates (title, starred, visibility, events, ...) remain
      plain writes — they are content, not lifecycle
```

Field-level content writes stay as they are. The convergence is scoped to **lifecycle verbs**: status transitions, processing admission/claim, finalization, discard, merge, and reprocess.

## Ownership matrix after the refactor

| Concern | Exclusive owner | Allowed projections/drivers |
|---|---|---|
| Status transitions & claim | lifecycle service | readers everywhere |
| Processing admission (all entry directions) | lifecycle service | transcribe/pusher/sync/imports/routers as intent submitters |
| Finalization / merge / reprocess orchestration | lifecycle service | postprocess steps as called stages |
| Content fields (title, starred, events, apps_results, ...) | unchanged — plain updates | — |

### Phase 1 — characterization & contract tests (separate PR, lands first)

- Behavioral contract tests for the current lifecycle state machine through controllable seams (fake Firestore/Redis per `backend/docs/test_isolation.md`): every legal transition, every illegal transition, double-claim, concurrent finalize-vs-reprocess, sync-vs-live collision, merge during processing, import of an already-processed conversation.
- Register the lifecycle workflow in `backend/testing/workflow_contracts.json` (the mechanism exists; this flow is not registered).
- Add counters/traces for each current write path so the convergence PR can demonstrate zero legacy-path use before deletion.
- These tests assert **current** behavior, including current quirks — they are the lock, not the spec.

### Phase 2 — convergence PR (one PR, #9515 execution model)

1. Introduce the lifecycle service and typed transition API behind existing behavior.
2. Migrate every caller by direction: live finalization, pusher, sync pipeline, imports, manual reprocess, merge, developer router.
3. Delete before merge:
   - direct lifecycle-field writes through `update_conversation` / `conversation_ref.update` outside the service;
   - `claim_conversation_status` as a public seam (its CAS survives *inside* the service);
   - per-caller admission/dedup logic superseded by the single claim point.
4. Guard: CI check failing on new lifecycle-field writes (`'status'`, `'discarded'`, processing markers) outside the service module. Ratchet to zero, no nonzero baseline.

If scope must shrink, drop an entire direction (e.g. imports) from both implementation and issue — never land half a direction with dual authorities (#9515's rule).

## Acceptance matrix (behavioral)

- Exactly one processing run per finalized conversation under concurrent finalize triggers (live + sync + manual).
- Reprocess during active processing either queues or rejects deterministically — never two concurrent processors.
- Merge during processing has one deterministic outcome; no orphaned `processing` status.
- Discard is terminal; subsequent processing intents are rejected and observable.
- Sync/backfill of an already-live-processed conversation is idempotent (no duplicate side effects: memories, notifications, integrations).
- Crash between claim and completion is recoverable — no conversation stuck in `processing` forever.

## Verification

- `backend/test.sh` green locally; new contract tests hermetic and discovered by the runner.
- Live exercise: real transcription session end-to-end plus a sync upload of overlapping audio; inspect logs for single-claim behavior.
- Legacy-path counters at zero during exercise before deletion commit.
- `make preflight`; invariant IDs via `scripts/pr-preflight --suggest`.

## Definition of done

- [ ] Phase 1 contract tests merged first and green in CI.
- [ ] Lifecycle verbs have exactly one mutation owner; matrix above holds.
- [ ] All ten current writer modules migrated; deletion inventory removed, not deprecated.
- [ ] Guard ratchet at zero.
- [ ] Acceptance matrix passes behaviorally.
- [ ] `docs/doc/developer/backend/listen_pusher_pipeline.mdx` and AGENTS.md service map updated if boundaries move.
- [ ] PR body records root cause, recurrence evidence, durable guard, real-path evidence, rollback plan.

## Non-goals

- Changing conversation content processing quality/prompts (`utils/llm` work is tracked separately — see the LLM gateway exit-criteria issue).
- Changing the sync admission architecture (`backend-sync` lanes are the model to emulate, not a patient).
- Firestore schema migration; the document shape stays readable by current clients.
- Touching macOS; the desktop analogue is #9515.

## Related

- #9515 — macOS convergence; same governing rule and execution model, source of the ownership-matrix/deletion-inventory format.
- `backend/testing/workflow_contracts.json` — where the lifecycle workflow gets registered.

### LINKED ISSUE #9687: Epic: durable listen→conversation lifecycle — single mutation owner + durable recording sessions (parent of #9516, #9351)
## What this is

Parent/executor issue for the backend listen→conversation lifecycle hardening. It binds together #9516 (converge conversation lifecycle onto a single mutation owner) and #9351 (durable recording sessions for listen lifecycle), adds the meta-machinery neither child specifies (failure-class evidence, ordering fuzzer, anti-regrowth guards, checks-manifest registration), and defines the closure criteria for all of it.

**An agent should be able to execute from this issue alone**: read the two children for their locked designs, follow the phases here, and open **one PR** whose body contains `Closes #9516`, `Closes #9351`, and `Closes` this issue.

This parent **supersedes #9516's "Phase 1 lands as its own PR" sequencing** (deliberate: one reviewable PR was chosen over two landings). The *intent* of that sequencing is preserved as a hard commit-ordering rule: characterization/contract tests are the first commit(s), refactor commits come after, and a reviewer must be able to check out the test commit alone and watch it pass against unmodified production code.

## Why: the recurring failure classes

Churn since 2026-04-01: `routers/transcribe.py` 71 commits, `utils/conversations/process_conversation.py` 67, `routers/sync.py` 60, `routers/pusher.py` 30. The fixes cluster; every work item below maps to one of these classes — if you are tempted to cut an item, check which class it leaves open.

| Class | Example fix commits | What allowed it |
|-------|--------------------|-----------------|
| **A. Plural mutation authority** | `claim_conversation_status()` CAS scar (`database/conversations.py:878`), ten modules writing conversation docs via a generic dict API (full census in #9516) | No single owner of lifecycle verbs; any caller can set any field |
| **B. Finalization races on disconnect/restart** | `8d55fc0cfd` (guard disconnect finalization races), `25227df431` (desktop recording finalization reconciliation), `2d67863cad` (make listen finalization durable) | Finalization decisions interleave client disconnect, STT callbacks, and background finalizers with no stated ordering contract |
| **C. Identity/binding drift** | `3367796271` (bind recordings by client identity), #9349 (stale global pointer attaching recordings to wrong conversation) | Conversation attachment inferred from user-global pointers/timestamps/WS ordering instead of an explicit durable mapping |
| **D. Stream teardown edge cases** | `c69a7484c4` (flush multi-channel tail audio before pusher close), `d4252a6738` (retain failed audio, surface live STT errors) | Teardown paths hand-ordered per call site |

Class A is #9516's core. Classes B/C are #9351's core. Class D is closed as a byproduct of the ordering fuzzer below.

## Rules of engagement (binding for the whole PR)

1. **A test that would not have caught one of the fix commits above is decoration.** Name the commit or class each new test would have caught.
2. **Characterization before movement.** The Phase 1 tests assert *current* behavior including current quirks — they are the lock, not the spec. If a quirk is found that must change, that change is its own later commit with its own justification, never silently folded into the refactor.
3. **Guards must be class-closing.** Hand-enumerated interleavings are tripwires; properties that hold under *any* interleaving are guards. Prefer guards.
4. **Follow the repo's backend rules exactly**: import purity, no in-function imports, executors via `run_blocking` (never bare `asyncio.to_thread`), `asyncio.wait(FIRST_COMPLETED)` supervisor patterns in WS handlers, `record_fallback` on any provider/mode/fail-open branch, `sanitize()`/`sanitize_pii()` in logs. See `AGENTS.md` and `backend/docs/test_isolation.md`.
5. **Hermetic or out of CI.** Every new CI test uses fake Firestore/Redis seams per `backend/docs/test_isolation.md` — no live services, no sleeps, no ordering dependence.

### Phase 1 — characterization & contract lock (first commit(s))

Execute #9516's Phase 1 as specified there, plus:

- Register the conversation lifecycle workflow in `backend/testing/workflow_contracts.json` (mechanism exists; flow is unregistered).
- **Ordering fuzzer (the meta-addition):** extract the finalization *decision* logic reachable from `routers/transcribe.py` / `utils/conversations/process_conversation.py` into a pure function (inputs: session/conversation state + one event; output: decision), then property-test it over randomized legal event sequences — disconnect-before-final, final-after-disconnect, duplicate finals, concurrent finalize-vs-reprocess, restart mid-finalization, merge during processing. Seeded PRNG (`OMI_FUZZ_SEED`), fixed corpus in CI, failing sequences checked in as named regression tests. Invariants: exactly one terminal lifecycle outcome per conversation, no side-effect fanout executes twice for the same idempotency key, no transition out of a terminal state. This is #9351's "state-machine/property tests" requirement made concrete and shared with #9516's state machine.
- Gate: full `backend/test.sh` green, new tests pass against **unmodified** production code.

### Phase 2 — single mutation owner (#9516)

Execute #9516's locked target architecture: one lifecycle service owning status transitions, processing admission (single claim point, idempotent by conversation id), finalization/merge/reprocess orchestration; typed transitions failing closed; migrate all ten writer modules; delete every alternative lifecycle-write path (content-field writes stay plain). Keep the ownership matrix from #9516 as the acceptance table.

- **Anti-regrowth guard (meta-addition):** a small static checker (registered in `.github/checks-manifest.yaml`, both lanes) that fails when a module outside the lifecycle service writes lifecycle fields (`status`, claim fields) through `database/conversations.py`. Label it a static tripwire; include a self-test fixture proving it fails on a pasted-back direct write. Follow the conventions of existing `.github/scripts/check_*_ratchet.py` checkers.
- Gate: zero legacy-path use demonstrated via the Phase 1 counters/traces before any deletion commit.

### Phase 3 — durable recording sessions (#9351)

Execute #9351's design requirements as written: immutable `recording_session_id` idempotency key, persisted session→conversation mapping with lifecycle state and generation metadata, versioned lifecycle-event envelope with monotonic sequence, idempotent finalization, feature-flagged rollout with dual-write/shadow validation, retire the user-global in-progress pointer only after mismatch metrics demonstrate safe cutover. Route the new session state machine's transitions **through the Phase 2 lifecycle service** — do not create a second mutation owner.

- The rollout flag / shadow-validation scaffolding gets `LIFECYCLE:` headers per AGENTS.md (one-time files need `DELETE-AFTER:`).
- Fallback/shadow-mismatch branches call `record_fallback` (component from the closed enum, e.g. `sync_dispatch`/`other`) — no new one-off counters.

### Phase gates

If Phase 1 characterization reveals a behavior change that requires redesign beyond the two children's locked architectures, stop and surface it on this issue before proceeding. Phase 3 starts only when Phase 2's guard shows zero legacy writers.

## Verification (cite evidence in the PR body)

1. `backend/test-preflight.sh`, then full `backend/test.sh` after each phase.
2. `make preflight` — new manifest checks present and green.
3. Fuzzer seed corpus + run count/runtime recorded in the PR body.
4. Update `docs/doc/developer/backend/listen_pusher_pipeline.mdx` in the same PR (this changes conversation lifecycle — mandatory per AGENTS.md).
5. PR body includes: the failure-class table above as recurrence evidence, root cause + durable guard per class, the #9516 ownership matrix as shipped, per-child closure checklist (below), and how any non-hermetic validation (shadow-mode metrics) was run.

## Closure criteria

- **#9516 closes when:** lifecycle verbs have exactly one owner, typed transitions fail closed, all ten writers migrated, legacy paths deleted, contract suite + workflow registration in CI, anti-regrowth checker live.
- **#9351 closes when:** recording sessions are durable server resources with idempotent finalization, property tests cover its listed scenarios, rollout flag + shadow validation shipped, metrics for mapping conflicts/stale-event discards/legacy-path usage emitting.
- **This issue closes when:** both children close via the single PR and the meta-machinery (fuzzer, guards, manifest entries, doc update) is verified in CI.

## Explicitly out of scope

- #9355 (sync change feed / tombstones) — related but needs its own rollback plan; do not fold in.
- Any STT-provider work (Parakeet/Deepgram issues), diarization, translation.
- App/desktop client changes beyond what #9351's envelope versioning strictly requires (compat path stays until cutover).

## Related

Parent of: #9516, #9351. Sibling playbook: #9515 (macOS convergence), the PTT stability plan (desktop). Phase 1 precedent: #9349.

Git-on-my-level:
## Summary

Closes #9687
Closes #9516
Closes #9351

This is the one-PR landing requested by #9687. It deliberately supersedes #9516's earlier two-PR landing sequence while retaining the tests-before-refactor contract as strict commit order:

1. `66f13d02f` / `a825ffbbd` characterize current admission behavior and add the deterministic finalization-decision ordering fuzzer against unmodified production writes.
2. `513e25f04` centralizes lifecycle mutation ownership and registers the lifecycle-write static guard in the deterministic checks manifest.
3. `2ea541143` persists durable recording sessions and routes all session lifecycle operations through that service.
4. `d0347f2ad` / `c2529047f` / `8f42b84f0` / `0d8f8e9c9` / `8e8fbc7d0` / `8a18d57fe` / `166cd7d34` / `b82e6b568` / `6fbb8a5ad` / `3443cd43f` / `ffd55aa95` / `9040edec6` / `a435d3746` / `321f9161b` / `89e0855` close the independent-review race, fanout, rollover, dual-write, envelope-isolation, static-guard, isolation-boundary, fenced-job, atomic fanout-admission, deleted-row, mismatch-envelope, create-vs-processor-write, fenced-completion-signal, legacy-admission-leak, post-fanout deletion-replay, discarded-reconnect, empty-session tombstone-reuse, post-persistence legacy fanout, late-content cleanup, photo-only finalization, and lifecycle-storage API boundary findings without weakening the three phases above.

## Failure-class closure

| Failure class | Root cause | Durable guard and recurrence evidence |
| --- | --- | --- |
| Plural mutation authority | Routers, pusher, finalizers, and storage helpers could independently write lifecycle fields or conflate an initial create with a stale processor update. | `513e25f04` creates the exclusive lifecycle service; the manifest-enforced AST guard rejects direct lifecycle writes, including generic DB and Firestore transaction forms. `c2529047f` hardens the checker with bypass regressions. `6fbb8a5ad` separates explicit idempotent initial creation from fail-closed processing-result persistence. `89e0855` exposes the storage primitives only as the lifecycle service’s guard-tracked boundary, so strict type checking rejects neither an intentional private escape nor a second owner. |
| Disconnect finalization races | Late disconnects, stale processors, or a concurrent deletion could re-admit, overwrite, resurrect, or falsely report completion for a terminal conversation. | `a825ffbbd` supplies the ordering fuzzer; `c2529047f` evaluates its reducer on the transaction snapshot and fences processing/terminal writes with CAS contracts. `8e8fbc7d0` preserves the transaction result through write preparation, and `8a18d57fe` moves durable acceptance ahead of all derived writes. `b82e6b568` makes a missing row a fail-closed no-resurrection fence, including the initial finalizer read. `3443cd43f` preserves terminal `fenced` through the pusher protocol so a listener retires the request without issuing a false completed envelope. `a435d3746` also reports a later core-persistence fence back to each legacy endpoint before it can trigger direct integrations. |
| Identity and binding drift | Redis/global in-progress pointers and reconnects were not a durable recording identity. | `2ea541143` introduces immutable `(uid, recording_session_id)` bindings and ordered envelopes; `d0347f2ad` adds rollover generations, terminal immutability, delayed completion routing, and `0d8f8e9c9` keeps a dual-write legacy route from borrowing the canonical envelope. `b82e6b568` ensures that same legacy route keeps delivering processing/completed envelopes on a dual-write mapping conflict, while the canonical binding remains untouched. `8e8fbc7d0` immediately rolls a terminal reconnect before it can accept audio. `ffd55aa95` suppresses a discarded terminal reconnect rather than replaying it as completed. `9040edec6` tombstones an empty session before deleting its row, and any missing bound row forces a fresh recording generation instead of recreating the old conversation ID. |
| Teardown and fanout edges | A lease could expire between conversation completion and external integration fanout, while an empty cleanup could race with late segments or photos. | `d0347f2ad` persists a fenced fanout state and stable idempotency key on the finalization job, so replay completes the same fanout boundary before job completion. `8a18d57fe` adds an explicit, lease-fenced terminal no-fanout job disposition for a discarded or superseded processor. `166cd7d34` moves the conversation’s completed, non-discarded, job-id, and revision checks into the claim transaction itself; interleaving regressions cover both discard and newer-binding fences. `3443cd43f` gives the legacy finalization endpoint the same fail-closed admission branch, returning its durable winner before processor or integration side effects. `ffd55aa95` recognizes a durable fanout acknowledgement before fencing a deleted-row replay, allowing the worker to atomically finish without a duplicate delivery. `a435d3746` atomically tombstones/deletes only an empty parent while segment and photo writes share the parent transaction and set `has_content`; a late writer wins the retry and routes durable finalization instead of losing content. `321f9161b` admits that durable marker so a photo-only conversation reaches its finalization job. |

## Child closure criteria

- **#9516:** one lifecycle mutation authority owns initial creation, admission, transitions, discard protection, and terminal processing; raw lifecycle writes are prevented by a CI-manifest static contract and runtime regressions cover late failed/discarded/deleted finalization plus fresh direct/deferred creation. The legacy endpoints also observe a core persistence refusal and return the durable row without integration fanout.
- **#9351:** a recording session now owns immutable conversation binding plus ordered lifecycle envelopes; reconnect retries return the canonical binding, silence rollover starts a new generation, terminal phases cannot change, and shadow/dual-write/enforce provide a compatibility ramp. An empty generation is terminally tombstoned before deletion; reconnecting a missing bound row starts a new recording/conversation ID, never recreates the tombstoned generation. Cleanup and segment/photo writes share a parent transaction, so late content cancels cleanup and follows durable finalization. Dual-write preserves the legacy route on a mismatch without emitting the canonical envelope for a different conversation, while still delivering its processing/completed compatibility envelopes.
- **#9687:** every finalization handoff is transactionally admitted, lease-fenced, replayable, and only marked complete after idempotency-keyed external fanout completes. A lost fanout race or deleted row is atomically recorded as terminal `fenced`, except a previously acknowledged fanout is normally completed on deletion replay; neither case can duplicate integration, retry forever, dead-letter, or emit a false listener-completed transition. Both finalization endpoints return the durable winner before derived side effects, including when core persistence loses its later fence.

### Shipped #9516 ownership matrix

| Concern | Exclusive owner | Drivers/projections |
| --- | --- | --- |
| Status transitions and claims | `utils.conversations.lifecycle` | transcribe, pusher, sync/import/router/merge paths submit intents or read state |
| Processing admission | `utils.conversations.lifecycle.admit_processing` | live finalization, pusher, imports, manual router and merge paths |
| Finalization, merge, and reprocess handoff | `utils.conversations.lifecycle` | durable finalization worker and postprocess stages |
| Content fields | Existing narrow/generic content writes | title, starred, events, integrations, audio metadata, and other non-lifecycle content |

The static guard scans production backend sources with a zero baseline and rejects both the lifecycle persistence primitives and generic/raw writes of lifecycle fields outside the exclusive owner. Its self-tests include pasted-back direct, anonymized Firestore, generic-API, and finalization-admission escape hatches.

## Scope fence

#9355 (sync change feed) is explicitly excluded. It has a separate data/rollback plan and must land independently; this PR neither changes its feed protocol nor relies on it for lifecycle correctness.

## Verification

- `cd backend && PYTHON=.venv/bin/python bash test-preflight.sh` — 18 checks passed, 0 failed (the remaining optional integration-provider/service checks are warnings only).
- `cd backend && ./test.sh` — 584 backend unit-test files passed after the final review-fix commit.
- `cd backend && .venv/bin/python -m pytest tests/unit/test_conversation_finalization_jobs.py tests/unit/test_listen_finalization_cloud_tasks.py tests/unit/utils/test_listen_pusher_session.py tests/unit/test_conversation_search_date_validation.py -q` — 77 targeted finalization, protocol, and legacy-admission tests passed after `3443cd43f`.
- `cd backend && .venv/bin/python -m pytest tests/unit/test_transcribe_decisions.py tests/unit/test_listen_finalization_cloud_tasks.py -q` — 63 reconnect and finalization-replay tests passed after `ffd55aa95`.
- `cd backend && .venv/bin/python -m pytest tests/unit/test_recording_sessions.py -q` — 22 durable binding/tombstone tests passed after `9040edec6`, including empty-session deletion and reconnect rollover.
- `cd backend && .venv/bin/python -m pytest tests/unit/test_conversation_finalization_jobs.py tests/unit/test_recording_sessions.py tests/unit/test_conversation_search_date_validation.py tests/unit/test_process_conversation_usage_context.py tests/unit/test_transcribe_conversation_cache.py -q` — 92 targeted lifecycle and legacy-route tests passed after `321f9161b`, including photo-only finalization admission, persistence-fence fanout suppression, and late-content empty-cleanup races.
- `cd backend && .venv/bin/python -m pytest tests/unit/test_conversation_lifecycle_contract.py tests/unit/test_process_conversation_usage_context.py tests/unit/test_developer_from_segments_idempotency.py tests/unit/test_listen_finalization_cloud_tasks.py tests/unit/test_recording_sessions.py -q` — 91 lifecycle contracts passed, including deleted-row fencing, no derived work, terminal finalization, dual-write compatibility, and fresh/deferred creation.
- `cd backend && bash scripts/typecheck.sh` — 0 errors (existing strict-Pyright warnings only) after `89e0855`.
- `make preflight` — deterministic manifest checks passed (existing architecture warnings only).
- `cd desktop/macos && xcrun swift build -c debug --package-path Desktop` — passed.
- `cd desktop/macos && xcrun swift test --package-path Desktop --filter StopReconciliationTests` — 40 tests passed.
- Built and launched named test bundle `omi-durable-recording-sessions` through `./run.sh --yolo`; the non-production automation bridge responded. The seeded named bundle was signed out, so no authenticated live listen session was available; durable lifecycle behavior is exercised by the backend contracts and fuzzer above.

The finalization fuzzer runs four checked-in named edge corpus sequences plus 250 seeded randomized sequences (1–32 events each) on every test run. The default seed is `OMI_FUZZ_SEED=9687`; failures preserve the seed through the environment variable for deterministic replay.

## Invariants

No product invariants affected (`scripts/pr-preflight --suggest`).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9700?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
## Monitor run summary

Fixed 5 hermetic backend E2E test failures caused by the lifecycle guard introduced in this PR.

**Root cause:** The lifecycle guard in `update_conversation` blocks direct writes to lifecycle fields (`status`, `discarded`). This broke:

1. **Reprocessing completed conversations** — `persist_processed_conversation` only accepted existing status in `{processing, merging}`, fencing out reprocess over already-completed conversations.
2. **E2E test fakes** — fake `process_conversation` callbacks called `update_conversation` with `status`, triggering the guard.
3. **Sync test fake** — called `upsert_conversation` which was renamed to `upsert_conversation_with_lifecycle`.

**Fixes:**
- `lifecycle.py`: added `completed` to `expected_statuses` in `persist_processed_conversation` so reprocess overwrites are not fenced.
- `test_canonical_memory_pipeline.py`, `test_conversation_processing_deterministic.py`: removed `status` from `update_conversation` calls in fake functions (conversations were already completed).
- `test_core_flow_expansion.py`: removed `status` from `update_conversation`, added explicit `lifecycle.complete()` call to match the production finalize flow.
- `test_sync_lifecycle.py`: updated `upsert_conversation` → `upsert_conversation_with_lifecycle` (renamed by this PR).

**Verification:** full hermetic e2e suite passes locally (104 passed, 3 skipped). CI: all checks green including Hermetic Backend E2E.

Git-on-my-level:
@cursor address merge conflicts and do a review + fix round

PR #9701, opened 2026-07-14
fix(candidates): return None on a malformed candidate doc in get_candidate instead of 500ing

ZachL111:
## What

`get_candidate` built a `CandidateRecord` from a stored doc with no guard:

```python
snapshot = _candidate_ref(uid, candidate_id).get()
if not snapshot.exists:
    return None
return CandidateRecord.from_storage(_snapshot_dict(snapshot))
```

`CandidateRecord.from_storage` is `model_validate`, so one legacy or schema-drifted candidate doc (a missing subject discriminator, a drifted `status` enum, a missing required field) raises `ValidationError`. `GET /v1/candidates/{candidate_id}` returns this directly, and the staged-task read paths (`DELETE /v1/staged-tasks/{id}`, `POST /v1/staged-tasks/{id}/promote`) also fetch a candidate through it, so a single bad doc 500s all of them.

## Fix

Mirror the sibling `list_candidates`, which already wraps the identical construction in `try/except ValidationError` and skips the malformed doc. `get_candidate` now returns `None` on `ValidationError` (the router maps `None` to 404, exactly as if the list had skipped it):

```python
try:
    return CandidateRecord.from_storage(_snapshot_dict(snapshot))
except ValidationError as e:
    errors = e.errors(include_input=False, include_url=False)
    bounded_error_types = [str(error.get('type', 'unknown')) for error in errors[:5]]
    logger.warning('Ignoring malformed candidate record %s: error_count=%s validation_types=%s', getattr(snapshot, 'id', None), e.error_count(), bounded_error_types)
    return None
```

The catch is narrow (`ValidationError`) so unexpected errors still surface, and the log keeps only bounded structural error types (no `input_value`) since the rendered pydantic error can contain private task text -- matching `list_candidates`.

## Tests

New `tests/unit/test_get_candidate_skip_malformed.py` drives it with an injected fake db proxy:
- a malformed doc returns None (no 500)
- a missing doc returns None (unchanged)
- an unexpected non-ValidationError still propagates

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_get_candidate_skip_malformed.py -q
  -> 3 passed
black --line-length 120 --skip-string-normalization --check database/candidates.py tests/unit/test_get_candidate_skip_malformed.py
  -> unchanged
python -m pyright -p pyrightconfig.json database/candidates.py -> 0 errors
python scripts/check_module_stub_pollution.py -> Checked 638 file(s); 0 violations
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9701?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/candidates.py
+++ b/backend/database/candidates.py
@@ -547,7 +547,22 @@ def get_candidate(uid: str, candidate_id: str) -> Optional[CandidateRecord]:
     snapshot = _candidate_ref(uid, candidate_id).get()
     if not snapshot.exists:
         return None
-    return CandidateRecord.from_storage(_snapshot_dict(snapshot))
+    try:
+        return CandidateRecord.from_storage(_snapshot_dict(snapshot))
+    except ValidationError as e:
+        # A malformed/legacy candidate doc must not 500 the read; return None (the router maps that to
+        # 404), mirroring the skip in list_candidates. Catch only ValidationError so unexpected runtime
+        # errors still surface, and log bounded structural error types only since the rendered pydantic
+        # error can contain private task text.
+        errors = e.errors(include_input=False, include_url=False)
+        bounded_error_types = [str(error.get('type', 'unknown')) for error in errors[:5]]
+        logger.warning(
+            'Ignoring malformed candidate record %s: error_count=%s validation_types=%s',
+            getattr(snapshot, 'id', None),
+            e.error_count(),
+            bounded_error_types,
+        )
+        return None
 
 
 def list_candidates(
--- a/backend/tests/unit/test_get_candidate_skip_malformed.py
+++ b/backend/tests/unit/test_get_candidate_skip_malformed.py
@@ -1,0 +1,77 @@
+"""database.candidates.get_candidate returns None on a malformed candidate doc instead of 500ing.
+
+get_candidate built CandidateRecord.from_storage(_snapshot_dict(snapshot)) with no try/except, so one
+legacy or schema-drifted candidate doc raised ValidationError and 500'd GET /v1/candidates/{id} (and
+the staged-task read paths that fetch a candidate). The sibling list_candidates already skips such a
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
Thanks for the focused fix here. I reviewed the candidate read path and the new unit coverage.

This looks like the right direction: `get_candidate` now mirrors `list_candidates` by swallowing only Pydantic `ValidationError`, returning `None` so callers can map the malformed/legacy record to not-found behavior, while still letting unexpected runtime errors surface. I also like that the warning logs bounded validation error types without including Pydantic input values, which avoids leaking private task/candidate content into logs.

I ran the targeted backend unit test locally in a clean venv with a sanitized environment:

```text
pytest -q tests/unit/test_get_candidate_skip_malformed.py
3 passed, 1 warning in 0.31s
black --check --diff database/candidates.py tests/unit/test_get_candidate_skip_malformed.py
2 files would be left unchanged
```

No blocking code issues found from my pass. Since this touches malformed user-data handling / logging behavior and this repo currently requires human maintainer review before formal approval, I’m leaving this as a positive review signal rather than an approval.

ZachL111:
Thanks for the review and for running the tests. Agreed on mirroring list_candidates and keeping the catch to ValidationError so unexpected errors still surface. Ready for a human maintainer whenever the malformed-record handling is good to sign off on.

kodjima33 (approved):

kodjima33 (approved):

PR #9702, opened 2026-07-14
Guard task workflow control reads against a malformed control doc

ZachL111:
## What

`get_task_workflow_control` loads the per-user `task_intelligence_control/state` doc and validates it:

```python
payload = snapshot.to_dict()
if not isinstance(payload, dict):
    return TaskWorkflowControl()
return TaskWorkflowControl.model_validate(cast(dict[str, Any], payload))
```

`TaskWorkflowControl` is `model_config = ConfigDict(extra='forbid', ...)`, so any drift in the stored doc (a removed or renamed field left behind by an older writer, a bad enum value) fails to load with `ValidationError`. Every candidate and staged-task read gates on this function:

- `routers/candidates.py` (list, get, resolve, promote paths) calls it 6 times
- `routers/staged_tasks.py` calls it 4 times

So a single corrupt control document 500s all of those endpoints for that user, not just one.

## Fix

Fall back to the model default on `ValidationError`:

```python
try:
    return TaskWorkflowControl.model_validate(cast(dict[str, Any], payload))
except ValidationError as e:
    logger.warning(
        'Ignoring malformed task workflow control for %s: %s',
        uid,
        [str(err.get('type', 'unknown')) for err in e.errors(include_input=False, include_url=False)[:5]],
    )
    return TaskWorkflowControl()
```

The default is legacy-safe by construction: `workflow_mode=off` (not in the `{write, read}` set that `_validate_write_control` requires) and `account_generation=0`. So a corrupt control document disables candidate writes rather than silently enabling them. The model docstring already states "defaults preserve legacy behavior."

The catch is narrow (`ValidationError`), so an unexpected failure still surfaces. The log keeps only bounded pydantic error type strings, not `input_value`, since a rendered pydantic error can contain private task text.

## Tests

New `tests/unit/test_task_workflow_control_skip_malformed.py`, driven through the module `db` seam (`monkeypatch.setattr` on the module reference, no `sys.modules` mutation):

- a malformed doc (extra forbidden field) falls back to the default, with `workflow_mode` off and `account_generation` 0
- a valid doc still parses (`workflow_mode` write, `account_generation` 2)
- an unexpected non-`ValidationError` still propagates

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_task_workflow_control_skip_malformed.py -q
  -> 3 passed
black --line-length 120 --skip-string-normalization --check database/task_intelligence_control.py tests/unit/test_task_workflow_control_skip_malformed.py
  -> unchanged
python -m pyright -p pyrightconfig.json database/task_intelligence_control.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9702?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/task_intelligence_control.py
+++ b/backend/database/task_intelligence_control.py
@@ -1,12 +1,16 @@
 """Firestore persistence for per-user task workflow migration controls."""
 
+import logging
 from typing import Any, cast
 
 from google.api_core.exceptions import AlreadyExists, Conflict
+from pydantic import ValidationError
 
 from config.what_matters_now_smoke_fixture import WHAT_MATTERS_NOW_SMOKE_UID, is_development_smoke_fixture
 from database._client import db
 from models.task_intelligence import TaskWorkflowControl, TaskWorkflowMode
+
+logger = logging.getLogger(__name__)
 
 CONTROL_COLLECTION = 'task_intelligence_control'
 CONTROL_DOCUMENT = 'state'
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for the focused fix — this looks like the right failure mode for a drifted task workflow control document.

I reviewed the change and did a local sanitized verification of the new backend unit test plus formatting/type checks:

- `python -m pytest tests/unit/test_task_workflow_control_skip_malformed.py -q` → 3 passed
- `black --line-length 120 --skip-string-normalization --check database/task_intelligence_control.py tests/unit/test_task_workflow_control_skip_malformed.py` → unchanged
- `pyright -p pyrightconfig.json database/task_intelligence_control.py` → 0 errors

The narrow `ValidationError` catch is good: malformed persisted control docs fall back to the legacy-safe default (`workflow_mode=off`, `account_generation=0`), while unexpected failures still propagate. I also like that the warning logs only bounded Pydantic error types and avoids including input values that could contain private task text.

Because this touches the task workflow/user-data control path, I’m not treating this as a formal approval; it should still get maintainer/security review before merge. But from the implementation and regression-test angle, this is a positive signal.

ZachL111:
Thanks for the detailed review. Agreed the legacy-safe default (workflow_mode off, generation 0) is the right failure mode for a drifted control doc, and keeping the log to bounded error types keeps private task text out of the logs. Ready for maintainer and security sign-off whenever it fits.

kodjima33 (approved):

kodjima33 (approved):

PR #9703, opened 2026-07-14
Guard long-term memory patch against a malformed stored timestamp

ZachL111:
## What

`apply_long_term_patch_transaction` (models/memory_apply.py) pulls four optional timestamp fields off the patch and coerces them from stored ISO strings:

```python
for timestamp_key in ("last_corroborated_at", "captured_at", "updated_at", "expires_at"):
    if timestamp_key in extra_item_updates and isinstance(extra_item_updates[timestamp_key], str):
        extra_item_updates[timestamp_key] = datetime.fromisoformat(
            extra_item_updates[timestamp_key].replace("Z", "+00:00")
        )
```

`datetime.fromisoformat` raises `ValueError` on any non-ISO string. A single drifted value (a legacy or hand-edited stored timestamp, a format left behind by an older writer) aborts the entire patch apply for that memory item. This runs in the background memory-maintenance path, so the failure is a stuck item rather than an HTTP 500, but the whole patch is lost either way.

## Fix

Extract a small helper that returns `None` on `ValueError`, and drop the one malformed field instead of coercing it:

```python
def _coerce_iso_timestamp(value: str, *, field: str) -> Optional[datetime]:
    try:
        return datetime.fromisoformat(value.replace("Z", "+00:00"))
    except ValueError:
        logger.warning("Dropping malformed timestamp field %s in long-term memory patch", field)
        return None
```

```python
coerced = _coerce_iso_timestamp(extra_item_updates[timestamp_key], field=timestamp_key)
if coerced is None:
    extra_item_updates.pop(timestamp_key)
else:
    extra_item_updates[timestamp_key] = coerced
```

Dropping (rather than writing `None`) means the item keeps its prior value in both consumers:

- update path: `_apply_update_memory_item` does `updates.update(extra_updates)` then `existing.model_copy(update=updates)`, so a dropped key leaves the existing item's value untouched
- create path: `MemoryItem(**{**memory_item.dict(), **extra_item_updates})`, so a dropped key keeps the materialized value

The except is narrow (`ValueError` only). The log names the field only, never the raw value, since a stored memory timestamp field can carry private text if the data is malformed in an unexpected way.

## Close the failure class

This is the last unguarded `datetime.fromisoformat` on stored input in `models/memory_apply.py` (one site). The same class in the task-integration due-date path was already guarded; two other model deserializers (`models/app.py`, `models/calendar_mutation.py`) parse stored timestamps and are separately scoped, not touched here to keep this fix confined to one file.

## Tests

New `tests/unit/test_memory_apply_timestamp_coerce.py` (imports the light models module directly, no stubbing):

- the helper parses a valid `Z` suffix to an aware datetime, and returns `None` on a malformed string
- a malformed `captured_at` is dropped and the patch still commits (before the guard this raised `ValueError` out of the whole apply); the committed item keeps its materialized datetime
- valid `captured_at` and `updated_at` overrides are still applied as aware datetimes

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_memory_apply_timestamp_coerce.py -q
  -> 4 passed
black --line-length 120 --skip-string-normalization --check models/memory_apply.py tests/unit/test_memory_apply_timestamp_coerce.py
  -> unchanged
python -m pyright -p pyrightconfig.json models/memory_apply.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9703?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/models/memory_apply.py
+++ b/backend/models/memory_apply.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from datetime import datetime, timezone
 from enum import Enum
 from typing import Any, Dict, List, Optional
@@ -28,6 +29,8 @@ from models.product_memory import (
     MemoryItem,
 )
 from utils.memory.short_term_lifecycle import default_short_term_expiry
+
+logger = logging.getLogger(__name__)
 
 
 class ApplyStatus(str, Enum):
@@ -397,6 +400,20 @@ def _barrier_outbox_events(
     ]
 
 
+def _coerce_iso_timestamp(value: str, *, field: str) -> Optional[datetime]:
+    """Parse a stored ISO timestamp string, tolerating a trailing 'Z'.
+
+    Returns None on a malformed value so the caller can drop just that one field instead of
+    letting a single drifted string abort the whole patch. Logs the field name only, never the
+    raw value, which can carry memory text.
+    """
+    try:
+        return datetime.fromisoformat(value.replace("Z", "+00:00"))
+    except ValueError:
+        logger.warning("Dropping malformed timestamp field %s in long-term memory patch", field)
+        return None
+
+
 def apply_long_term_patch_transaction(
     *, control_state: MemoryControlState, operation: MemoryOperation, patch_payload: Dict[str, Any]
 ) -> ApplyResult:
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

Git-on-my-level:
Thanks @ZachL111 — this is a clean, well-scoped defensive hardening. The fix correctly narrows the exception to `ValueError`, preserves the existing/materialized timestamp value on both the update and create paths by popping the malformed key, and deliberately logs only the field name (never the raw value, which could carry private memory text). Good test coverage on the create path.

Two notes for a maintainer:

- **Update path test gap**: The regression test covers the `add` (create) path where a malformed `captured_at` is dropped and the materialized value is kept. The `update` path (`existing.model_copy(update=updates)`) is arguably the more common real-world case for drifted timestamps on existing items, but isn't directly exercised by a test here. Consider adding one if convenient, though the logic is symmetric and sound.
- **Checks**: all relevant CI (Backend unit suite, Hermetic Backend E2E, Formatting, Hygiene, contracts) is green on the latest runs. The cancelled runs are stale workflow re-dispatch artifacts.

This touches user-data persistence (the long-term memory subsystem), so I'm flagging it for human maintainer review before merge rather than formally approving.

PR #9707, opened 2026-07-14
Guard goal storage normalization against malformed legacy metric rows

ZachL111:
## What

`database.goals.normalize_goal_storage` projects a stored goal row into the canonical shape. It runs on every goal read (`get_user_goal`, `get_all_goals`, `get_goal_by_id`) and before progress-event writes, and it delegates the metric to `_metric_from_storage`:

```python
return GoalMetric(
    type=GoalType(data.get('goal_type', GoalType.scale.value)),
    current=float(data.get('current_value', 0)),
    target=float(data.get('target_value', 0)),
    min=float(data.get('min_value', 0)) if data.get('min_value') is not None else None,
    max=float(data.get('max_value', 10)) if data.get('max_value') is not None else None,
    unit=data.get('unit'),
)
```

`current`/`target` use a raw `float()`, `goal_type` a raw `GoalType()`, and `latest_progress_sequence` (line 267) a raw `int()`. A drifted legacy row with a null or non-numeric `current_value`/`target_value`/`latest_progress_sequence`, or a `goal_type` that is not a valid enum value, raises `TypeError`/`ValueError` out of the read, so one bad row 500s every goal endpoint for that user.

The tell: `min`/`max` on the very next lines already None-guard (the author knew these stored numbers can be null), but `current`/`target` do not.

## Fix

Add small `_safe_float`/`_safe_int` helpers (try/except returning a default) and apply them to all four metric numbers plus `latest_progress_sequence`. Guard `goal_type` by set membership, exactly like the `status` and `source` fields already do in the same function. Wrap the metric construction in a `ValidationError` backstop so a still-inconsistent metric (a `min > max` pair that trips the model bounds validator, or a malformed nested `metric` dict) degrades to no metric rather than 500ing:

```python
goal_type = data.get('goal_type', GoalType.scale.value)
if goal_type not in {member.value for member in GoalType}:
    goal_type = GoalType.scale.value
try:
    return GoalMetric(
        type=GoalType(goal_type),
        current=_safe_float(data.get('current_value'), 0.0),
        target=_safe_float(data.get('target_value'), 0.0),
        min=_safe_float(data.get('min_value'), 0.0) if data.get('min_value') is not None else None,
        max=_safe_float(data.get('max_value'), 10.0) if data.get('max_value') is not None else None,
        unit=data.get('unit'),
    )
except ValidationError:
    return None
```

Good fields are preserved; only a genuinely bad field is defaulted, and the goal itself always loads.

## Close the failure class

This closes the numeric/enum coercion of stored goal fields in the storage-normalization layer. The response-projection layer (`utils/goals_response.py`) already guards its float fields with a `response_float` helper but still has one raw `int(latest_progress_sequence)`; that sibling gap is addressed in a separate follow-up so each layer is independently reviewable.

## Tests

New `tests/unit/test_goal_storage_coerce_guard.py` (pure functions, no Firestore):

- `_safe_float`/`_safe_int` fall back on `None`/non-numeric and pass valid values through
- a malformed legacy row yields a defaulted metric (bad enum falls back to `scale`, bad numbers to 0.0)
- a valid row is preserved unchanged
- an inconsistent `min > max` pair degrades to no metric instead of raising
- `normalize_goal_storage` survives a malformed `latest_progress_sequence` + metric

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_goal_storage_coerce_guard.py -q
  -> 5 passed
black --line-length 120 --skip-string-normalization --check database/goals.py tests/unit/test_goal_storage_coerce_guard.py
  -> unchanged
python -m pyright -p pyrightconfig.json database/goals.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9707?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/goals.py
+++ b/backend/database/goals.py
@@ -184,20 +184,47 @@ def _goal_created_at_sort_key(goal: Dict[str, Any]) -> datetime:
     return _coerce_created_at(goal.get('created_at'))
 
 
+def _safe_float(value: Any, default: float) -> float:
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return default
+
+
+def _safe_int(value: Any, default: int) -> int:
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return default
+
+
 def _metric_from_storage(data: dict[str, Any]) -> Optional[GoalMetric]:
     metric = data.get('metric')
     if isinstance(metric, dict):
-        return GoalMetric.model_validate(metric)
+        try:
+            return GoalMetric.model_validate(metric)
+        except ValidationError:
+            return None
     if data.get('target_value') is None and data.get('goal_type') is None:
         return None
-    return GoalMetric(
-        type=GoalType(data.get('goal_type', GoalType.scale.value)),
-        current=float(data.get('current_value', 0)),
-        target=float(data.get('target_value', 0)),
-        min=float(data.get('min_value', 0)) if data.get('min_value') is not None else None,
-        max=float(data.get('max_value', 10)) if data.get('max_value') is not None else None,
-        unit=data.get('unit'),
-    )
+    # A drifted legacy row can carry a null/non-numeric current/target/min/max, a bad goal_type, or an
+    # inconsistent min/max pair. Coerce each field defensively (the sibling min/max already None-guard) and
+    # guard goal_type by set membership like status/source below; the ValidationError backstop degrades a
+    # still-inconsistent metric to absent instead of 500ing every goal read.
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9711, opened 2026-07-14
Guard paid-app price coercion so a bad price does not 500 the app update

ZachL111:
## What

`utils.apps.upsert_app_payment_link` builds a Stripe recurring price:

```python
payment_price = stripe.create_app_monthly_recurring_price(app.payment_product_id, int(price * 100))
```

`routers.apps.update_app` (PATCH `/v1/apps/{app_id}`) passes the raw request price into it:

```python
upsert_app_payment_link(data.get('id'), data.get('is_paid', False), data.get('price'), data.get('payment_plan'), data.get('uid'), previous_price=app.get("price", 0))
```

`create_app` validates the paid-app price up front (422 if missing/negative), but `update_app` does not. So PATCHing an app to `is_paid=true` without a `price` (`data.get('price')` is `None`), or with a raw numeric string, reaches `int(None * 100)` / `int("5.0" * 100)` and raises `TypeError`/`ValueError`, returning HTTP 500. The existing `if price == 0` guard only covered the zero case.

## Fix

Extend that guard to skip link creation for any non-positive or non-numeric price, matching the graceful `price == 0` path already there:

```python
if not isinstance(price, (int, float)) or isinstance(price, bool) or price <= 0:
    logger.error(f"App price is missing or not a positive number, app_id: {app_id}")
    return app
```

The `price` parameter was annotated `float`, but the caller passes untyped request data, so the annotation is corrected to `Any` (that annotation is also what made the guard necessary). Neither caller uses the return value (`update_app` returns `{'status': 'ok'}`, `create_app` ignores it), so returning early without a link is safe.

Scope: this stops the 500. The related ordering issue -- `update_app` persists `is_paid` before the payment link is confirmed, so a paid app can still end up without a link on any upsert miss -- predates this crash (the 500 fired after the persist too) and is a separate follow-up; it is not widened into this PR.

## Tests

New `tests/unit/test_app_paid_price_guard.py` (drives `upsert_app_payment_link` with the Stripe client and db mocked):

- a `None` / non-numeric / negative / zero / bool price skips Stripe entirely without raising
- a valid `5.0` price still creates the recurring price with `int(5.0 * 100) == 500` cents

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_app_paid_price_guard.py -q
  -> 7 passed
black --line-length 120 --skip-string-normalization --check utils/apps.py tests/unit/test_app_paid_price_guard.py
  -> unchanged
python -m pyright -p pyrightconfig.json utils/apps.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9711?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_app_paid_price_guard.py
+++ b/backend/tests/unit/test_app_paid_price_guard.py
@@ -1,0 +1,61 @@
+"""Regression: a paid-app update with a missing or non-numeric price must not 500.
+
+routers.apps.update_app passes the raw request price straight into
+utils.apps.upsert_app_payment_link, which does int(price * 100) to build a Stripe recurring
+price. A null price (is_paid toggled on without a price) or a non-numeric value used to raise
+TypeError/ValueError. The guard treats any non-positive or non-numeric price like the existing
+price==0 case: it skips link creation and returns without crashing.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+import utils.apps as apps_mod
+from utils.apps import upsert_app_payment_link
+
+
+def _app_data(**over):
+    data = {
+        'id': 'app1',
+        'name': 'Test App',
+        'category': 'productivity',
+        'author': 'Zach',
+        'description': 'desc',
+        'image': 'https://storage.googleapis.com/x/y.png',
+        'capabilities': [],
+        'uid': 'u1',
+        'is_paid': True,
+        'payment_plan': 'monthly_recurring',
+        'price': 5.0,
+    }
+    data.update(over)
+    return data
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9712, opened 2026-07-14
Reject an unsupported listen codec/sample_rate before building the decoder

ZachL111:
## What

`routers.transcribe._stream_handler` builds the native audio decoders before its main `try` block:

```python
# multi-channel
multi_opus_decoders = [_get_opuslib().Decoder(sample_rate, 1) for _ in channel_configs]
# single-channel
opus_decoder = _get_opuslib().Decoder(sample_rate, 1)
lc3_decoder = _get_lc3().Decoder(lc3_frame_duration_us, sample_rate)
```

`sample_rate` and `codec` come straight from the connect query string. `opuslib.Decoder` only accepts the standard opus rates, so a device opening `/v4/listen?codec=opus&sample_rate=44100` raises `OpusError`. A bare `codec=lc3` normalizes to a `None` frame duration (only `lc3_fs1030` carries one), so `lc3py.Decoder(None, sample_rate)` raises `TypeError`. Both raise before the handler's `try`, so the exception escapes to Starlette as an unhandled ASGI error and the socket drops as an unclean 1006 instead of closing cleanly with a reason. The decode path already knows only specific formats are valid (it gates `codec == 'opus' and sample_rate == 16000`), but the constructor was ungated.

## Fix

Add a pure `validate_audio_format(codec, sample_rate)` in `utils.transcribe_decisions` and call it at connect, before any decoder is constructed:

```python
audio_format_error = validate_audio_format(codec, sample_rate)
if audio_format_error is not None:
    logger.warning(f"Rejecting unsupported audio format: {audio_format_error} {uid} {session_id}")
    await websocket.close(code=1003, reason=audio_format_error)
    return
```

`opus`/`opus_fs320` are checked against opuslib's supported set `{8000, 12000, 16000, 24000, 48000}`; a bare `lc3` is rejected in favor of the `lc3_fs1030` variant that carries a frame duration. The guard sits ahead of all three construction sites, so an unsupported request now closes with 1003 (unsupported data) and a human-readable reason instead of crashing the handler.

The rate set is opuslib's constructor set, not narrowed to 16 kHz, so this only rejects formats that cannot build a decoder at all; it does not change which rates the decode path accepts.

## Tests

New `tests/unit/test_validate_audio_format.py` (pure helper, no heavy import):

- supported pairs pass (`opus`/`opus_fs320` at valid rates, `lc3_fs1030`, `pcm8`/`pcm16`/`aac`)
- an opus rate opuslib cannot decode (44100) is rejected, and every advertised opus rate is accepted
- bare `lc3` is rejected while `lc3_fs1030` passes

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_validate_audio_format.py -q
  -> 3 passed
black --line-length 120 --skip-string-normalization --check utils/transcribe_decisions.py routers/transcribe.py tests/unit/test_validate_audio_format.py
  -> unchanged
python -m pyright -p pyrightconfig.json utils/transcribe_decisions.py
  -> 0 errors    (routers/transcribe.py is pyright-excluded; py_compile clean)
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9712?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/transcribe.py
+++ b/backend/routers/transcribe.py
@@ -150,6 +150,7 @@ from utils.transcribe_decisions import (  # async-blockers: no-import-scope; asy
     normalize_codec_frame,
     normalize_language,
     person_id_for_client,
+    validate_audio_format,
     resolve_photo_conversation_source,
     select_translation_language,
     should_enable_speaker_identification,
@@ -379,6 +380,12 @@ async def _stream_handler(
     logger.info(
         f'_stream_handler {uid} {session_id} {language} {sample_rate} {codec} {include_speech_profile} {stt_service} {conversation_timeout} custom_stt={custom_stt_mode} onboarding={onboarding_mode} client_conversation_id={bool(client_conversation_id)}'
     )
+
+    audio_format_error = validate_audio_format(codec, sample_rate)
+    if audio_format_error is not None:
+        logger.warning(f"Rejecting unsupported audio format: {audio_format_error} {uid} {session_id}")
+        await websocket.close(code=1003, reason=audio_format_error)
+        return
 
     use_custom_stt = custom_stt_mode == CustomSttMode.enabled
     is_multi_channel = channels >= 2
--- a/backend/tests/unit/test_validate_audio_format.py
+++ b/backend/tests/unit/test_validate_audio_format.py
@@ -1,0 +1,38 @@
+"""Regression: an unsupported codec/sample_rate must be rejected before the decoder is built.
+
+routers.transcribe._stream_handler constructs the native opus/lc3 decoders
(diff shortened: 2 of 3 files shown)


kodjima33 (approved):

PR #9713, opened 2026-07-14
Cap the app-tool status-message cache so it cannot grow unbounded

ZachL111:
## What

`utils.retrieval.tools.app_tools` keeps a process-global cache of tool status messages:

```python
# Global mapping of tool names to status messages
_tool_status_messages: Dict[str, str] = {}
```

`create_app_tool` writes into it on every load, keyed by `{app_id}_{tool}`:

```python
tool_name = f"{app_id}_{app_tool.name}"
if app_tool.status_message:
    _tool_status_messages[tool_name] = app_tool.status_message
```

`load_app_tools` runs on every agentic chat turn, so distinct `{app_id}_{tool}` keys accumulate for the entire process lifetime. Nothing deletes, caps, or expires them, so the map grows without bound. This is exactly the case the repo's own guidance calls out: "Module-level dicts: add TTL-based eviction or cap size - they grow forever otherwise" (`backend/CLAUDE.md`).

## Fix

Route the write through a small bounded helper that evicts the oldest entry once a cap is reached:

```python
_MAX_TOOL_STATUS_MESSAGES = 2048

def _remember_tool_status(tool_name: str, status_message: str) -> None:
    # Re-inserting refreshes recency; evict the oldest entry once the cap is reached.
    _tool_status_messages.pop(tool_name, None)
    while len(_tool_status_messages) >= _MAX_TOOL_STATUS_MESSAGES:
        _tool_status_messages.pop(next(iter(_tool_status_messages)), None)
    _tool_status_messages[tool_name] = status_message
```

Dict insertion order (Python 3.7+) makes `next(iter(...))` the oldest key. Re-inserting a key refreshes its recency, so an actively-used tool stays cached. The cap of 2048 comfortably covers any realistic set of active app-tools while bounding worst-case memory. The getter (`get_tool_status_message`) is unchanged.

## Tests

New `tests/unit/test_app_tools_status_cap.py`:

- writing past the cap keeps the map at exactly the cap, evicts the oldest key, and retains the most recent (still readable via `get_tool_status_message`)

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_app_tools_status_cap.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check utils/retrieval/tools/app_tools.py tests/unit/test_app_tools_status_cap.py
  -> unchanged
python -m pyright -p pyrightconfig.json utils/retrieval/tools/app_tools.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9713?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_app_tools_status_cap.py
+++ b/backend/tests/unit/test_app_tools_status_cap.py
@@ -1,0 +1,41 @@
+"""Regression: the app-tool status-message cache must stay bounded.
+
+utils.retrieval.tools.app_tools keeps a process-global _tool_status_messages map keyed by
+{app_id}_{tool}. Every agentic chat turn that loads app tools writes new keys, and nothing evicted
+them, so the map grew for the whole process lifetime (the repo's own rule requires module-level dicts
+to cap or TTL). _remember_tool_status caps the map and evicts the oldest entry.
+"""
+
+import pytest
+
+import utils.retrieval.tools.app_tools as app_tools
+
+
+@pytest.fixture(autouse=True)
+def _clear_status_cache():
+    app_tools._tool_status_messages.clear()
+    yield
+    app_tools._tool_status_messages.clear()
+
+
+def test_status_cache_is_bounded_and_evicts_oldest():
+    cap = app_tools._MAX_TOOL_STATUS_MESSAGES
+    for i in range(cap + 50):
+        app_tools._remember_tool_status(f"app_{i}_tool", f"status {i}")
+
+    assert len(app_tools._tool_status_messages) == cap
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9715, opened 2026-07-14
Fail open in three vector_db functions when no Pinecone index is configured

ZachL111:
## What

`database.vector_db.index` is the Pinecone index, and it is `None` on any deploy without Pinecone configured (self-hosted vector search, the `PROVIDER_MODE=offline` harness, local dev). The module treats this as a first-class state: more than twenty functions guard it, for example

```python
def query_vectors(...):
    if index is None:
        return []
    ...

def delete_vector(...):
    if index is None:
        return
    ...
```

Three functions missed that guard and dereferenced `index` directly:

- `query_vectors_by_metadata` -> `index.query(...)`
- `upsert_vector2` -> `index.upsert(...)`
- `update_vector_metadata` -> `index.update(...)`

With no index these raise `AttributeError: 'NoneType' object has no attribute ...` instead of the intended empty/skip. `query_vectors_by_metadata` is called from the app-integration memory lookup (`utils/app_integrations.py`), where the failure silently drops the lookup; `upsert_vector2`/`update_vector_metadata` run during conversation processing (`utils/conversations/process_conversation.py`), where the structured-vector save dies as an unlogged `AttributeError` rather than the intended skip.

## Fix

Add the same `if index is None` guard the sibling functions already use, returning the type-appropriate empty result:

```python
def query_vectors_by_metadata(...) -> List[str]:
    if index is None:
        return []

def upsert_vector2(...) -> None:
    if index is None:
        return

def update_vector_metadata(...) -> Dict[str, Any]:
    if index is None:
        return {}
```

This restores the module's fail-open contract uniformly: without a vector index, reads return empty and writes are skipped, instead of raising.

## Tests

New `tests/unit/test_vector_db_index_none_guard.py` (forces `index` to None via monkeypatch):

- `query_vectors_by_metadata` returns `[]`
- `upsert_vector2` is a no-op (previously `AttributeError` on `index.upsert`)
- `update_vector_metadata` returns `{}`

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_vector_db_index_none_guard.py -q
  -> 3 passed
black --line-length 120 --skip-string-normalization --check database/vector_db.py tests/unit/test_vector_db_index_none_guard.py
  -> unchanged
python -m pyright -p pyrightconfig.json database/vector_db.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9715?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/vector_db.py
+++ b/backend/database/vector_db.py
@@ -120,6 +120,8 @@ def upsert_vector(uid: str, conversation_id: str, vector: List[float]) -> None:
 
 
 def upsert_vector2(uid: str, conversation_id: str, vector: List[float], metadata: Dict[str, Any]) -> None:
+    if index is None:
+        return
     data: VectorRecordDoc = _get_data(uid, conversation_id, vector)
     typed_metadata: Dict[str, Any] = data['metadata']
     typed_metadata.update(metadata)
@@ -128,6 +130,8 @@ def upsert_vector2(uid: str, conversation_id: str, vector: List[float], metadata
 
 
 def update_vector_metadata(uid: str, conversation_id: str, metadata: Dict[str, Any]) -> Dict[str, Any]:
+    if index is None:
+        return {}
     metadata['uid'] = uid
     metadata['memory_id'] = conversation_id
     result: Dict[str, Any] = index.update(f'{uid}-{conversation_id}', set_metadata=metadata, namespace="ns1")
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9718, opened 2026-07-14
fix(preflight): read desktop changelog as UTF-8

### LINKED ISSUE #9717: Windows desktop changelog validation uses the host locale encoding
## Problem

On native Windows Python with `PYTHONUTF8` unset and a GBK default locale, desktop changelog validation crashes before it can report a validation result.

## Reproduction

```powershell
$env:PYTHONUTF8 = $null
python .github/scripts/desktop-changelog.py validate
```

## Actual result

```text
UnicodeDecodeError: 'gbk' codec can't decode byte 0x94 in position 126
```

The traceback ends in `read_json()` at `path.read_text()`.

## Expected result

Repository changelog JSON should be decoded and written as UTF-8 regardless of the contributor's host locale, matching CI and the encoding used by the checked-in files.

## Root cause

`desktop-changelog.py` calls `Path.read_text()` and `Path.write_text()` without an explicit encoding. Native Windows Python therefore uses the process ANSI code page instead of UTF-8.

tianmind-studio:
## Summary

- reads and writes desktop changelog JSON with an explicit UTF-8 encoding
- adds a behavioral encoding contract for hosts whose default locale is not UTF-8
- registers the contract in the shared local/CI check manifest

Fixes #9717.

## Root cause

`desktop-changelog.py` relied on the implicit encoding of `Path.read_text()` and `Path.write_text()`. Native Windows Python uses the process ANSI code page when no encoding is supplied, so valid UTF-8 changelog content could raise `UnicodeDecodeError` under a GBK locale before validation started. The write path had the same host-dependent contract and could reject changelog text not representable in the active code page.

## Durable guard

The changelog boundary now always uses UTF-8 for both reads and writes. The regression tests provide path doubles that reject any omitted or non-UTF-8 encoding, and the shared check manifest runs those tests whenever the implementation or contract changes.

## Real-path exercise

On native Windows Python, I cleared `PYTHONUTF8` and ran `python .github/scripts/desktop-changelog.py validate`. The old code failed with a GBK `UnicodeDecodeError`; the fixed command completed successfully against the repository's real changelog data.

None. This changes repository tooling only and does not affect product behavior.

## Validation

- old-code regression: `test_desktop_changelog.py` failed both encoding tests before the production change
- `python .github/scripts/test_desktop_changelog.py` -> 2 passed
- `python .github/scripts/desktop-changelog.py validate` with parent `PYTHONUTF8` unset -> passed
- `python .github/scripts/test_run_checks.py` -> 9 passed
- `python -m py_compile .github/scripts/desktop-changelog.py .github/scripts/test_desktop_changelog.py` -> passed
- `make preflight` with parent `PYTHONUTF8` unset -> all 9 selected checks passed
- `git diff --check origin/main...HEAD` -> passed

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9718?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -37,6 +37,11 @@ checks:
     triggers: ["all"]
     lanes: ["local", "ci"]
     reason: "repository-wide desktop changelog schema contract"
+  - id: desktop-changelog-encoding-contract
+    command: ["python3", ".github/scripts/test_desktop_changelog.py"]
+    triggers: [".github/scripts/desktop-changelog.py", ".github/scripts/test_desktop_changelog.py"]
+    lanes: ["local", "ci"]
+    reason: "desktop changelog JSON uses UTF-8 across host locales"
   - id: desktop-changelog-entry
     command: ["python3", ".github/scripts/check-desktop-changelog.py", "--base", "{base}", "--head", "{head}", "{skip_changelog}"]
     triggers: ["desktop/macos/**"]
--- a/.github/scripts/desktop-changelog.py
+++ b/.github/scripts/desktop-changelog.py
@@ -24,14 +24,16 @@ class ChangelogError(Exception):
 
 def read_json(path: Path) -> object:
     try:
-        return json.loads(path.read_text())
+        return json.loads(path.read_text(encoding="utf-8"))
     except json.JSONDecodeError as exc:
         raise ChangelogError(f"{path} is not valid JSON: {exc}") from exc
 
 
 def write_json(path: Path, data: object) -> None:
     path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(json.dumps(data, indent=2, ensure_ascii=False) + "\n")
+    path.write_text(
+        json.dumps(data, indent=2, ensure_ascii=False) + "\n", encoding="utf-8"
+    )
 
 
 def normalize_changes(raw: object, path: Path) -> list[str]:
--- a/.github/scripts/test_desktop_changelog.py
+++ b/.github/scripts/test_desktop_changelog.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+"""Behavioral contract tests for desktop changelog JSON encoding."""
+
+from __future__ import annotations
+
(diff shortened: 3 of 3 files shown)


tianmind-studio:
Fresh downstream proof from reviewing #9720 on native Windows: with `PYTHONUTF8` unset, Python reported `cp936` / `utf8_mode=0`, and current `main` failed `desktop-changelog.py validate` with `UnicodeDecodeError` while reading #9720's valid changelog fragment (it contains an em dash). Cherry-picking this PR onto #9720 made the same validation pass; the two encoding contract tests also passed. This is now affecting a real concurrent contributor workflow, not only the synthetic regression fixture.

Git-on-my-level (commented):
Thanks for tightening this up. I reviewed the diff and also exercised the PR head locally at `69bb6fab325b76db01cd48cfaaf6d9782fc9e376`:

- `python3 .github/scripts/test_desktop_changelog.py` passed (2 tests)
- `python3 -m py_compile .github/scripts/desktop-changelog.py .github/scripts/test_desktop_changelog.py` passed
- `python3 .github/scripts/desktop-changelog.py validate` passed

The change is narrow and looks directionally correct: explicitly pinning the desktop changelog read/write boundary to UTF-8 fixes the host-locale failure mode without changing product behavior, and the new contract test covers both the read and write paths. Registering it in the manifest is also appropriate for preventing regression.

I’m leaving this as a positive maintainer signal rather than a formal approval because it touches repo check/preflight manifest/tooling paths, which this automation treats as workflow-sensitive and leaves for a human maintainer to approve/merge.

---

undivisible:
Closing as superseded by merged PR #9885, which fixed #9717 with the same explicit UTF-8 boundary and regression coverage.

Issue #4643, opened 2026-02-06
Fix AI chat timezone confusion by allowing users to choose or auto-detect the timezone, preventing incorrect event time mentions. 


hugoaap-code:
Every single time, no matter how many times I ask AI chat to remember my timezone or adding it to my memories, When engaging in conversations with AI chat, I have noticed that the AI does not account for my current time zone. 

For instance, it might refer to "tonight" when I am discussing an event that occurred in the afternoon, or "this afternoon" when I am referring to the morning. 

It would be great if we had the option to set our preferred time zone for the chat to operate within, which would help resolve these discrepancies in time references. 

Alternatively, an automatic detection of the user's time zone by the system could also address this issue.

Pizzacutioner:
Experiencing the same in the Android app. Notifications are happening an hour early from the time they are set.

CarlBalansag:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll start by reproducing the problem locally this week. Any pointers on where to start in the codebase would be appreciated!

hugoaap-code:
go ahead

Att.

Hugo Pereira

aryanorastar:
VERIFY pass (2026-08-06): this looks **FIXED_OPEN** — core chat timezone confusion is already addressed on `main`; the GitHub issue was never closed.

**Already on main**
- Auto-detect: device IANA tz → Firestore `users/{uid}.time_zone` via FCM registration (`FlutterTimezone` / web `Intl` / etc.)
- Chat prompt: `get_user_timezone` + `get_current_datetime_block` inject live local time
- Tool/render path: `resolve_display_tz` + merged work (#8483 #8495 #8538 #8588 #8575 #9719 #10012 …)

**Residual (separate product ask, not a re-do of this bug)**
- No manual timezone picker in settings
- Tz sync is FCM-coupled (denied/missing push → UTC fallback)

Recommend closing unless maintainers want a new issue scoped only to “optional timezone setting + decouple from FCM.” I won’t open a duplicate fix PR for #4643.

PR #9719, opened 2026-07-14
fix(backend): render chat tool timestamps in the user's timezone (#6214)

### LINKED ISSUE #6214: Chat assistant gives wrong times due to timezone confusion
## Problem
When chatting with Omi AI, it confuses hours and gives wrong times because it's not properly handling the user's timezone.

User asks "what time did X happen?" → AI answers with UTC time (off by 3 hours for UTC-3 user) instead of converting to user's local timezone (America/Sao_Paulo).

## Expected Behavior
AI should always use the user's configured timezone when discussing times in chat. Conversation timestamps should be converted to user's local time before being included in AI context.

## Impact
Wrong information, user loses trust in AI responses about timing.

## Fix Needed
Ensure chat AI context includes timestamps already converted to user timezone, not raw UTC.

kodjima33:
## What

Chat/agent tool results that still emitted **bare UTC timestamps** now render in the user's configured timezone, with an explicit timezone label.

Three surfaces were leaking UTC:

| Surface | Before | After |
|---|---|---|
| `get_action_items_text` (`/v1/tools/action-items` — desktop, web, MCP agent clients) | `Due: 2026-06-26 22:00:00` | `Due: 2026-06-26 19:00:00 America/Sao_Paulo` |
| `search_memories_text` (same REST surface) | `date: 2026-06-27` | `date: 2026-06-26` |
| `search_memories_tool` (agentic chat) | `date: 2026-06-27` | `date: 2026-06-26` |

## Why

From #6214:

> When chatting with Omi AI, it confuses hours and gives wrong times because it's not properly handling the user's timezone.
> User asks "what time did X happen?" → AI answers with UTC time (**off by 3 hours** for a UTC-3 user) instead of converting to the user's local timezone (America/Sao_Paulo).
>
> **Expected:** AI should *always* use the user's configured timezone when discussing times in chat.

The conversations half of this was fixed in #7736 (`conversations_to_string` localizes), but the **action-item and memory tool results were never converted** — so the issue's "always" is still not true today, and the reporter's exact symptom survives on the surfaces above.

The failure mode is specifically the **missing timezone label**: an unlabelled UTC wall clock reads as local time to the model, so it says "tonight" for a due date that is actually mid-afternoon for the user. Memory dates are worse in a quieter way — the UTC calendar date rolls over at a different instant than the user's, so a memory captured at 21:00 in São Paulo is reported a day late.

## Root cause / bug class

The agentic action-item tool already fixed exactly this in #4643, but the fix was applied at one call site rather than the shared boundary. Its REST sibling (`tool_services/action_items.py`) and both memory paths kept formatting raw `strftime` output.

To close the class rather than patch one more site, the formatting now has **one owner**: `_format_local` is promoted out of `action_item_tools.py` into `utils/conversations/render.py` as `format_local_time` (next to the existing `resolve_display_tz`), plus a new `format_local_date` for the date-only case. Both treat a naive timestamp as UTC defensively, so no path can hand the model an unlabelled wall clock.

A timezone lookup failure falls back to **labelled** UTC and never aborts retrieval.

## Verified

Backend env synced with `./scripts/sync-python-deps.sh`; `ENCRYPTION_SECRET` / `OPENAI_API_KEY` per `test.sh`.

**1. Regression tests fail without the fix.** New `backend/tests/unit/test_tool_services_timezone.py` (15 tests) imports the real modules and swaps collaborators with `monkeypatch`, so it exercises the production formatters. Reverting only the three production files and re-running:

```
8 failed, 10 passed      <- pre-fix (behavioural assertions on the rendered string)
15 passed                <- with the fix
```

**2. Real rendered tool output** (the string the model actually receives), user tz `America/Sao_Paulo`, stored `due_at = 2026-06-26T22:00:00Z`:

```
BEFORE:   Due: 2026-06-26 22:00:00                        <- no tz; model reads it as 22:00 local
AFTER:    Created: 2026-06-26 19:00:00 America/Sao_Paulo
          Due:     2026-06-26 19:00:00 America/Sao_Paulo

memory stored 2026-06-27T01:30Z
BEFORE:   date: 2026-06-27                                 <- a day late for the user
AFTER:    date: 2026-06-26
```

That is the reporter's 3-hour skew, gone.

**3. Full backend suite** — `bash test.sh`: no collection errors, no regressions. The single failure, `test_vad_gate.py::test_concurrent_sessions_no_deadlock`, is a **pre-existing flake unrelated to this change** (it touches none of the changed modules); A/B'd against pristine `origin/main`, where it fails 3/3 runs in isolation while this branch passes 2/3.

**4. `make preflight`** — `PR preflight passed: 8 checks`.

Covered: conversion correctness, naive-timestamp handling, the label on every emitted timestamp, unset-timezone fallback, and timezone-lookup failure.

## Not covered

`MemoryDB.get_memories_as_str` still renders `... UTC` — it is explicitly labelled, so the model can convert it correctly and it is not the mislabelling bug; threading a timezone into that shared model method is a wider change and is left out deliberately.

fixes #6214

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9719?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_action_item_date_validation.py
+++ b/backend/tests/unit/test_action_item_date_validation.py
@@ -245,7 +245,7 @@ _stub_package("utils.retrieval.tools")
 _stub_package("utils.llm")
 _stub_package("utils.conversations")
 
-# Stub utils.conversations.render (action_item_tools imports resolve_display_tz)
+# Stub utils.conversations.render (action_item_tools imports resolve_display_tz, format_local_time)
 _render_stub = _stub_module("utils.conversations.render")
 
 
@@ -258,7 +258,14 @@ def _real_resolve_display_tz(tz):
     return timezone.utc, "UTC"
 
 
+def _real_format_local_time(dt, display_tz, tz_label):
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    return f"{dt.astimezone(display_tz).strftime('%Y-%m-%d %H:%M:%S')} {tz_label}"
+
+
 _render_stub.resolve_display_tz = _real_resolve_display_tz
+_render_stub.format_local_time = _real_format_local_time
 
 # Stub utils.retrieval.agentic
 import contextvars
--- a/backend/tests/unit/test_action_item_tool_result_bound.py
+++ b/backend/tests/unit/test_action_item_tool_result_bound.py
@@ -89,7 +89,7 @@ for _name, _attrs in {
         "send_action_item_data_message",
         "sync_action_item_reminder",
     ],
-    "utils.conversations.render": ["resolve_display_tz"],
+    "utils.conversations.render": ["resolve_display_tz", "format_local_time"],
     "utils.retrieval.agentic": ["agent_config_context"],
 }.items():
     _m = _mod(_name)
--- a/backend/tests/unit/test_action_items_timezone.py
+++ b/backend/tests/unit/test_action_items_timezone.py
@@ -154,7 +154,14 @@ def _real_resolve_display_tz(tz):
     return timezone.utc, "UTC"
 
 
+def _real_format_local_time(dt, display_tz, tz_label):
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    return f"{dt.astimezone(display_tz).strftime('%Y-%m-%d %H:%M:%S')} {tz_label}"
+
+
 _render_stub.resolve_display_tz = _real_resolve_display_tz
+_render_stub.format_local_time = _real_format_local_time
 
 # Stub utils.retrieval.agentic (agent_config_context)
 import contextvars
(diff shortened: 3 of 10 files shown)


tianmind-studio (changes requested):
I found two integration gaps on `d1dc6b11` that should be addressed before merge.

First, the default-memory path still bypasses the new timezone conversion. Both `search_memories_text` and `search_memories_tool` compute `display_tz`, but when `search_memory_default_chat_memories_vector_decision_text` returns `USE_MEMORY`, they immediately return its preformatted `text`. That adapter still formats `updated_at` with a raw `strftime('%Y-%m-%d')` in `utils/memory/chat_memory_adapter.py`, so a `2026-06-27T01:30:00Z` memory is still rendered as `2026-06-27` for `America/Sao_Paulo`. I reproduced this through both public functions; each returned `date: 2026-06-27`. Please thread the display timezone (or a date formatter) into that adapter path and cover the `USE_MEMORY` branch in the regression tests.

Second, the new test file is not order-safe with the existing suite. On Windows, `pytest tests/unit/test_action_item_date_validation.py tests/unit/test_tool_services_timezone.py -q` fails during collection with `AttributeError: module 'google.auth' has no attribute 'credentials'`. `test_action_item_date_validation.py` snapshots `sys.modules` before its real `google.auth` imports, then restores the package while leaving imported children such as `google.auth.credentials` cached. Moving `_SYS_MODULES_SNAPSHOT = snapshot_sys_modules(...)` below the three `importable_mod_name` imports fixes the leak; the same two-file run then passes all 41 tests. Please include that isolation fix (or an equivalent hermetic approach) so the test can coexist in a single pytest process.

Issue #9724, opened 2026-07-14
Pre-push single-flight runner crashes on Windows without SIGHUP


tianmind-studio:
**Product impact**

- Primary loop layer: Docs-Tooling
- [ ] Involves data loss, failed capture, or broken sync/trust

**Describe the bug**

The installed pre-push hook routes through `.github/scripts/preflight_runner.py`, which assumes POSIX signal APIs are available. On Windows, every `git push` fails before the underlying pre-push checks start.

**To reproduce**

1. On Windows, install/use the repository Git hooks.
2. Push any branch that invokes `scripts/pre-push-singleflight`.
3. Observe `AttributeError: module 'signal' has no attribute 'SIGHUP'` from `run_owned()`.

**Current behavior**

`run_owned()` builds its handler map from `(signal.SIGINT, signal.SIGTERM, signal.SIGHUP)` unconditionally. Windows Python does not expose `SIGHUP`, so the wrapper exits with code 1 and Git rejects the push. The forwarding callback also uses POSIX-only `os.killpg`, which would be another portability hazard if invoked.

**Expected behavior**

The single-flight wrapper should register only signals supported by the host and forward them with a platform-compatible process API, allowing the same pre-push checks to run on Windows.

**Environment**

- Windows 11
- Python 3.11.15
- Git for Windows 2.50+

**Additional context**

Running `scripts/pre-push` directly succeeds; only the single-flight wrapper fails. A regression test should exercise signal selection without requiring a real signal to interrupt the test process.

PR #9725, opened 2026-07-14
test(desktop-backend): bypass proxies for loopback tests

### LINKED ISSUE #9722: Rust proxy cancellation tests fail behind a Windows system proxy
**Product impact**

**Describe the bug**

The Rust desktop backend's loopback cancellation tests use `reqwest::Client::new()`. On Windows, reqwest honors the enabled system proxy, so requests for the test server at `127.0.0.1:<ephemeral-port>` can be sent to that proxy instead of the local `TcpListener`.

**To reproduce**

1. Enable a Windows user proxy, for example `127.0.0.1:7897`.
2. From `desktop/macos/Backend-Rust`, run `cargo test routes::proxy::tests::`.
3. Observe the four local cancellation/disconnect tests time out waiting for `request_seen`.

**Current behavior**

The local fake upstream never receives the request. Diagnostic output shows the reqwest task finishing early with a proxy-generated `502` response, while `netsh winhttp show proxy` can still report direct access because the active proxy is configured in the current user's Internet Settings.

Affected tests:

- `hanging_upstream_is_cancelled_at_injected_deadline`
- `dropping_proxy_work_cancels_the_upstream_socket`
- `downstream_http_disconnect_cancels_upstream_work`
- `one_budget_includes_pre_dispatch_work_and_upstream_wait`

**Expected behavior**

Tests whose upstream is an in-process loopback listener should connect directly and behave deterministically regardless of host proxy configuration.

**Environment**

- Windows 11
- Rust 1.88.0, `x86_64-pc-windows-msvc`

**Additional context**

A test-only reqwest client built with `.no_proxy()` keeps production proxy behavior unchanged and makes the full 360-test suite pass under the same enabled Windows proxy.

tianmind-studio:
## What changed and why

Closes #9722.

Use a shared test-only reqwest client with system proxies disabled for the four proxy cancellation tests backed by an in-process loopback listener. This keeps production proxy behavior unchanged while making the tests deterministic on Windows hosts with a user proxy enabled.

none

## How it was verified

Verified on Windows 11 with the user proxy still enabled at `127.0.0.1:7897`:

- Before the fix, the four cancellation/disconnect tests consistently failed because reqwest received a proxy-generated `502` and the loopback listener never saw the request.
- `cargo test routes::proxy::tests::` -> 87 passed
- `cargo test` -> 360 passed
- `cargo fmt --check` -> passed
- `cargo clippy -- -D warnings` -> passed
- `scripts/pre-push fork` -> passed all selected checks with `PYTHONUTF8=1` and the internal-only changelog skip

The actual push used `--no-verify` only after the equivalent pre-push command passed because the single-flight hook wrapper currently crashes on Windows; #9724 tracks that separate issue.

## Tests

The four existing loopback cancellation tests are the regression coverage. They fail on the old code when a Windows system proxy is enabled and pass with the test client isolated from host proxy settings.

The violated contract was that an in-process loopback test server must receive the test request directly. `reqwest::Client::new()` honors the Windows user proxy, and that proxy can intercept `127.0.0.1` requests before they reach the test listener.

All four affected tests now obtain their client from one `local_test_client()` helper that calls `.no_proxy()`. The production Gemini clients are untouched, and future loopback cancellation coverage in this module has a single portable client factory to reuse.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9725?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Backend-Rust/src/routes/proxy.rs
+++ b/desktop/macos/Backend-Rust/src/routes/proxy.rs
@@ -1317,6 +1317,10 @@ mod tests {
     use tokio::net::{TcpListener, TcpStream};
     use tokio::sync::oneshot;
 
+    fn local_test_client() -> reqwest::Client {
+        reqwest::Client::builder().no_proxy().build().unwrap()
+    }
+
     async fn spawn_hanging_upstream() -> (
         String,
         oneshot::Receiver<()>,
@@ -1457,7 +1461,7 @@ mod tests {
     #[tokio::test]
     async fn hanging_upstream_is_cancelled_at_injected_deadline() {
         let (url, request_seen, disconnected, server) = spawn_hanging_upstream().await;
-        let client = reqwest::Client::new();
+        let client = local_test_client();
         let (deadline_tx, deadline_rx) = oneshot::channel();
         let work = tokio::spawn(async move {
             race_gemini_deadline(
@@ -1489,7 +1493,7 @@ mod tests {
     #[tokio::test]
     async fn dropping_proxy_work_cancels_the_upstream_socket() {
         let (url, request_seen, disconnected, server) = spawn_hanging_upstream().await;
-        let client = reqwest::Client::new();
+        let client = local_test_client();
         let proxy_work = tokio::spawn(async move {
             within_gemini_deadline(
                 Duration::from_secs(30),
@@ -1515,7 +1519,7 @@ mod tests {
     async fn downstream_http_disconnect_cancels_upstream_work() {
         let (upstream_url, request_seen, disconnected, upstream_server) =
             spawn_hanging_upstream().await;
-        let client = reqwest::Client::new();
+        let client = local_test_client();
         let app = Router::new().route(
             "/",
             post(move || {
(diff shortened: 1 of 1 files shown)


tianmind-studio:
This is an internal test-harness-only change, so it intentionally has no user-facing changelog fragment. Please add the `no-changelog-needed` label when triaging.

Git-on-my-level:
Nice, well-scoped fix. The `local_test_client()` helper with `.no_proxy()` is the right pattern for loopback-backed tests — `reqwest::Client::new()` honoring system proxies on Windows is a real source of flakiness, and isolating the test client from host proxy settings while leaving `gemini_client()`/`gemini_stream_client()` untouched is exactly the correct separation. The shared helper also makes future loopback tests in this module more consistent.

One note on CI: the context shows a cancelled `Hygiene` run, but the latest `Hygiene` run on this head actually completed as success, and `Rust Backend` passed — so the flaky/cancelled check doesn't reflect the current state. Not blocking.

---

kodjima33 (approved):

kodjima33 (approved):
test(desktop-backend): bypass proxies for loopback tests; CI green

kodjima33 (approved):
test(desktop-backend): bypass proxies for loopback tests; CI green

PR #9730, opened 2026-07-14
fix(ci): make pre-push single-flight runner start on hosts without POSIX signals (#9724)

### LINKED ISSUE #9724: Pre-push single-flight runner crashes on Windows without SIGHUP

kodjima33:
Fixes #9724.

## Bug
On Windows, **every `git push` fails before any pre-push check runs**:

```
```

`scripts/pre-push-singleflight` routes through `.github/scripts/preflight_runner.py`, whose `run_owned()` built its handler map from `(signal.SIGINT, signal.SIGTERM, signal.SIGHUP)` unconditionally. Windows Python does not define `SIGHUP`, so the wrapper exits 1 and Git rejects the push.

## Root cause / failure class
The runner assumes POSIX process APIs are always present. Fixing only the signal tuple would leave two more landmines on the same code path, so this PR closes the class:

1. **Signal registration** — `owned_signals()` registers only signals the host actually defines, and picks up `SIGBREAK` where it exists (Windows' console-interrupt signal).
2. **Signal forwarding** — `signal_child()` uses `os.killpg` where process groups exist (POSIX children get their own session via `start_new_session`), and falls back to `child.terminate()` elsewhere.
3. **Liveness probe** — `process_exists()` used `os.kill(pid, 0)`. On Windows `os.kill` does not send a signal; it calls `TerminateProcess`. The moment the runner became reachable on Windows, the stale-lock probe would have **killed the live lock holder**. Windows now probes with `OpenProcess`/`GetExitCodeProcess` (access-denied ⇒ alive, mirroring the POSIX `PermissionError` branch).

## Regression test
`.github/scripts/test_preflight_runner.py` (9 tests) exercises signal selection against Windows- and POSIX-shaped signal modules, forwarding on hosts with and without process groups, and asserts `process_exists` never reaches `os.kill` on Windows — no real signal needed to interrupt the test process, per the issue. Registered in `.github/checks-manifest.yaml` (`preflight-runner-portability`, local + ci lanes), so it is diff-scoped-selected whenever the runner changes.

## Verification
- `python3 .github/scripts/test_preflight_runner.py` → 9 tests OK
- `python3 .github/scripts/test_run_checks.py` → 9 tests OK (manifest contract guard)
- `make preflight` on this branch → 9 checks passed, with `preflight-runner-portability` SELECTED by the diff
- Real run: `preflight_runner.py --name wd -- sh -c 'echo "==> phase-one"; echo hello; exit 3'` streamed the phase line, wrote `result.json`, exited 3
- **A/B against `HEAD`:** SIGINT to the runner kills the child process group in both the old and the new code — POSIX forwarding behavior is unchanged
- Old expression against a Windows-shaped signal module raises `AttributeError: 'SIGHUP'`; `owned_signals()` returns `(SIGINT, SIGTERM, SIGBREAK)`

**Not verified on a real Windows host** (no Windows machine available) — the Windows paths are covered by the hermetic tests above and by CPython's documented `os.kill`/`OpenProcess` semantics.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9730?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -17,6 +17,11 @@ checks:
     triggers: [".github/schemas/desktop-release-evidence-v1.schema.json", ".github/scripts/desktop_release_doctor.py", ".github/scripts/desktop_release_doctor_report.py", ".github/scripts/test_desktop_release_doctor.py", ".github/workflows/desktop_release_doctor.yml"]
     lanes: ["local", "ci"]
     reason: "advisory desktop release evidence and drift report"
+  - id: preflight-runner-portability
+    command: ["python3", ".github/scripts/test_preflight_runner.py"]
+    triggers: [".github/scripts/preflight_runner.py", ".github/scripts/test_preflight_runner.py"]
+    lanes: ["local", "ci"]
+    reason: "single-flight pre-push runner must start on hosts without POSIX signal APIs"
   - id: diff-hygiene
     command: ["python3", ".github/scripts/check_diff_hygiene.py", "--changed-files", "{changed_files}", "--base", "{base}", "--head", "{head}"]
     triggers: ["all"]
--- a/.github/scripts/preflight_runner.py
+++ b/.github/scripts/preflight_runner.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import argparse
+import ctypes
 import hashlib
 import json
 import os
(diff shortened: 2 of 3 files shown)


tianmind-studio (changes requested):
Blocking on a real Windows host: `bash scripts/pre-push-singleflight fork` still exits before any pre-push check starts. After the signal fix succeeds, the wrapper hands the extensionless Bash script `scripts/pre-push` directly to Python's `subprocess.Popen`; Windows `CreateProcess` cannot launch it, and CPython 3.11.15 raises `FileNotFoundError: [WinError 2]` at `run_owned`.

Please pass the active Bash executable explicitly (for example, `-- "$BASH" scripts/pre-push "$@"` in `scripts/pre-push-singleflight`) and cover that wrapper contract. As written, this changes the first Windows failure but does not restore `git push`.

tianmind-studio:
Thanks for the signal/process work here. I reproduced the next Windows failure on this head and opened #9804 directly against your branch; it passes the active Bash executable to the runner, makes the existing POSIX simulations runnable on Windows, and adds the wrapper to the portability check trigger.

kodjima33:

PR #9731, opened 2026-07-14
test(backend): reuse workflow selector discovery

### LINKED ISSUE #9729: Workflow contract selector test exceeds Windows fast-unit budget
**Product impact**

**Describe the bug**

`backend/tests/unit/test_workflow_contracts.py::test_workflow_contract_sources_select_adjacent_tests` repeatedly exceeds the local 0.12-second fast-unit CPU budget on Windows even though every assertion passes.

**To reproduce**

1. On Windows, select `tests/unit/test_workflow_contracts.py` through `backend/test.sh` with the default local duration guard.
2. Observe all 19 tests pass.
3. Observe the session still exits 1 because `test_workflow_contract_sources_select_adjacent_tests` measures approximately 0.12-0.14 seconds of call-phase CPU.

**Current behavior**

The workflow-contract test file loads `select_backend_unit_tests.py` and scans the full discovered test inventory independently in seven test calls. The slow test then evaluates 19 source-path cases on top of that repeated setup. Windows filesystem/path processing pushes the call just over the strict local threshold, blocking unrelated pre-push runs.

**Expected behavior**

Shared selector loading and test discovery should run once as module-scoped test setup. Individual contract assertions should remain under the existing strict budget without adding a global allowlist exception or raising the threshold.

**Environment**

- Windows 11
- Python 3.11.15
- pytest 8.4.1

**Additional context**

The duration guard intentionally measures only the call phase so shared setup is not misattributed as a per-test regression. Reusing a module-scoped fixture follows that contract and also reduces total file runtime.

tianmind-studio:
## What changed and why

Reuse one module-scoped workflow-selector discovery across the contract tests instead of reloading the selector and rescanning every backend unit test in each test. This keeps the Windows fast-unit lane below its per-file timing budget without changing the contract assertions. Closes #9729.

none

## How it was verified

- Reproduced the baseline twice on Windows: all 19 tests passed, but the fast-unit duration guard rejected the file at about 0.14s and 0.12s against the 0.12s budget.
- Repeated the strict `backend/test.sh` run five times after the change. All 19 tests passed each time, with total runs of 0.57s, 0.47s, 0.54s, 0.52s, and 0.60s and no duration warning.
- `black --check --line-length 120 --skip-string-normalization backend/tests/unit/test_workflow_contracts.py`
- Passed lifecycle-header, version-filename, import-purity, module-isolation, workflow-contract, diff-hygiene, release-policy, and deployment-concurrency checks.
- The repository-wide Firestore query ratchet still fails on Windows with 40 unregistered and 3 unsupported fingerprints. The same command reports the identical fingerprints from an unchanged detached `origin/main` worktree, so that baseline is unrelated to this test-only diff.

## Tests

The changed file is the regression coverage. Its existing 19 contract tests still exercise the same assertion set, now under the same strict fast-unit timing guard that exposed the Windows failure.

Each of seven tests independently imported `select_backend_unit_tests.py` and walked the complete unit-test tree. Windows filesystem startup variance pushed that repeated discovery over the fast-unit per-file budget even though every assertion passed. A module-scoped fixture performs the immutable discovery once, and the existing duration guard continues to enforce the budget.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9731?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_workflow_contracts.py
+++ b/backend/tests/unit/test_workflow_contracts.py
@@ -3,6 +3,8 @@ import json
 import re
 from pathlib import Path
 
+import pytest
+
 BACKEND_DIR = Path(__file__).resolve().parents[2]
 
 
@@ -15,10 +17,15 @@ def _load_script(name: str):
     return module
 
 
-def test_memory_policy_core_change_selects_inv_mem_guard():
-    """Narrow memory policy PRs always pull INV-MEM guard tests."""
+@pytest.fixture(scope="module")
+def selector_and_all_tests():
     selector = _load_script("select_backend_unit_tests")
-    all_tests = selector.discover_all_tests()
+    return selector, selector.discover_all_tests()
+
+
+def test_memory_policy_core_change_selects_inv_mem_guard(selector_and_all_tests):
+    """Narrow memory policy PRs always pull INV-MEM guard tests."""
+    selector, all_tests = selector_and_all_tests
 
     selected, reason = selector.tests_for_changed_paths(
         ["backend/utils/memory/chat_memory_adapter.py"],
@@ -28,9 +35,8 @@ def test_memory_policy_core_change_selects_inv_mem_guard():
     assert reason == "selected backend unit tests from changed paths and workflow contracts"
 
 
-def test_workflow_contract_sources_select_adjacent_tests():
-    selector = _load_script("select_backend_unit_tests")
-    all_tests = selector.discover_all_tests()
+def test_workflow_contract_sources_select_adjacent_tests(selector_and_all_tests):
+    selector, all_tests = selector_and_all_tests
 
     full_run_cases = {
         "backend/database/memory_vector_repair_outbox_worker.py": "tests/unit/test_vector_repair_outbox_worker.py",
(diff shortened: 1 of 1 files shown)


Git-on-my-level:
Thanks for this refactor, @tianmind-studio — clean approach.

**What I checked**
- The module-scoped `selector_and_all_tests` fixture correctly replaces per-test `_load_script("select_backend_unit_tests")` + `discover_all_tests()` across all seven tests, performing the immutable filesystem discovery once instead of seven times.
- No assertion logic changed — the fixture returns `(selector, all_tests)` and each test unpacks it exactly as before.
- The `@pytest.fixture(scope="module")` scope is the right choice: discovery is deterministic and side-effect-free within a module run, so caching it is safe.

**Checks**
All check-runs on this head are green (`Formatting`, `Hygiene`, `Detect Changes`, `Backend unit suite`, `Hermetic Backend E2E`, etc.). The cancelled runs in the CI timeline are superseded re-runs, not failures.

**Verdict**
Good test-infrastructure improvement that directly addresses #9729. The Windows timing root cause (repeated discovery) is well explained in the PR body, and the fix is idiomatic pytest. Leaving this for a maintainer to merge.

---

kodjima33 (approved):

kodjima33 (approved):
test(backend): reuse workflow selector discovery; CI green

kodjima33 (approved):
test(backend): reuse workflow selector discovery; CI green

PR #9751, opened 2026-07-14
fix(firestore): index active attention overrides

Git-on-my-level:
## Summary
- add the explicit `__name__ ASC` trailing field required by the active `task_attention_overrides` Firestore query
- keep the query registry and generated `firestore.indexes.json` manifest in lockstep
- add regression coverage for the exact three-field signature and reconciliation of the live index shape

## Why
The latest development acceptance smoke reached the freshly deployed backend but surfaced a Firestore `FailedPrecondition`: the active attention-override query requires `account_generation ASC`, `expires_at ASC`, and `__name__ ASC`. The prior checked-in two-field index was insufficient.

## Scope and safety
- source-of-truth Firestore index manifest/registry and tests only
- no production deployment, Cloud resource mutation, workflow, traffic, secret, or environment change
- the existing development Firestore reconciliation gate will provision and wait for this manifest entry before the next backend candidate deploy

## Verification
- independent review approved
- focused index/query/reconciliation tests: 12 passed
- generated-manifest check, diff hygiene, typecheck (0 errors), selected backend tests, OpenAPI contract, formatting, and full repository pre-push checks passed

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9751?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/firestore_index_registry.py
+++ b/backend/database/firestore_index_registry.py
@@ -196,7 +196,7 @@ ACTIVE_ATTENTION_OVERRIDE_QUERY = FirestoreQuerySpec(
         FirestoreQueryFilter('account_generation', '==', 'account_generation'),
         FirestoreQueryFilter('expires_at', '>', 'now'),
     ),
-    index_fields=(_asc('account_generation'), _asc('expires_at')),
+    index_fields=(_asc('account_generation'), _asc('expires_at'), _asc('__name__')),
 )
 
 QUERY_SPECS = (ACTIVE_ATTENTION_OVERRIDE_QUERY,)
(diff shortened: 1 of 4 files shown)


tianmind-studio:
Follow-up from the next deployment: the corrected three-field index reached `READY`, but the acceptance smoke still returned HTTP 500 because the workflow reconciled indexes in `based-hardware-dev` while the deployed backend serves Firestore through `GOOGLE_CLOUD_PROJECT=based-hardware`.

I reproduced the empty fixture/query/transaction path successfully with the real Python SDK against the Firestore emulator, then traced the project mismatch in the deployment log and checked-in runtime contract. #9753 aligns reconciliation with `RUNTIME_GCP_PROJECT_ID` and adds a validator regression guard.

Failed follow-up run: https://github.com/BasedHardware/omi/actions/runs/29365024038

PR #9759, opened 2026-07-14
fix(openapi): make client generators portable on Windows

tianmind-studio:
## Summary
- read and write all Dart, TypeScript, and Swift OpenAPI generator files explicitly as UTF-8
- emit slash-stable source labels in generated TypeScript and Swift headers on every OS
- handle TypeScript and Swift output paths outside the repository without `relative_to()` failures
- add subprocess regression coverage with Python UTF-8 mode disabled, plus Windows path-label tests

## Root cause
The client generators used `Path.read_text()` and `Path.write_text()` without an encoding. On Windows, Python therefore inherited the active ANSI code page. The UTF-8 app-client contract failed to decode under GBK before generation started.

After forcing UTF-8 manually, TypeScript `--check` still reported all four committed outputs as stale because the generated header embedded a Windows backslash path while the committed Linux-generated header used forward slashes. The Swift generator had the same path-label contract, and both generators assumed every explicit output lived under the repository root.

This was visible in the contributor path because `scripts/pre-push` invokes all three generator checks directly; it does not inherit the UTF-8 export used by `backend/test.sh`.

## Scope and safety
- generator file I/O, generated-file labels, diagnostics, and regression tests only
- no OpenAPI schema, generated client API, backend route, application runtime, or deployment behavior changes
- generated outputs remain byte-stable on Linux and use explicit LF newlines when written on Windows

None.

## Verification
- real Windows checks with `PYTHONUTF8` unset:
  - `python backend/scripts/generate_dart_models.py --all --check`
  - `python backend/scripts/generate_ts_openapi_types.py --check`
  - `python backend/scripts/generate_swift_openapi_types.py --check`
- official backend runner for the three generator test files: 43 passed
- forced non-UTF-8 locale and Windows path regression subset: 5 passed
- Black 24.4.2 and Ruff 0.4.4: passed
- diff hygiene, lifecycle headers, versioned filenames, import purity, test isolation, workflow contracts, async blocker, route policy, and conversation lifecycle guards: passed

The full 601-file backend run had one unrelated Windows CPU-time guard at the exact `0.12s` threshold; all four assertions in that file passed, and the same file passed its isolated official-runner retry in `0.10s`. The aggregate preflight also encounters the existing Windows Firestore path-normalization false positive fixed by #9683; checks after that gate were run directly and passed.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9759?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/generate_dart_models.py
+++ b/backend/scripts/generate_dart_models.py
@@ -1236,18 +1236,18 @@ def main() -> int:
     if args.all and args.output:
         raise SystemExit('--output cannot be used with --all')
 
-    spec = json.loads(Path(args.spec).read_text())
+    spec = json.loads(Path(args.spec).read_text(encoding='utf-8'))
     groups = tuple(SCHEMA_GROUPS) if args.all else (args.group,)
     for group in groups:
         output_path = Path(args.output) if args.output else SCHEMA_GROUPS[group]['output']
         generated = build_output(spec, group)
         if args.check:
-            if not output_path.exists() or output_path.read_text() != generated:
+            if not output_path.exists() or output_path.read_text(encoding='utf-8') != generated:
                 raise SystemExit(f'{output_path} is stale; run backend/scripts/generate_dart_models.py --group {group}')
             print(f'{output_path} is up to date')
             continue
         output_path.parent.mkdir(parents=True, exist_ok=True)
-        output_path.write_text(generated)
+        output_path.write_text(generated, encoding='utf-8', newline='\n')
         print(f'wrote {output_path}')
     return 0
 
--- a/backend/scripts/generate_swift_openapi_types.py
+++ b/backend/scripts/generate_swift_openapi_types.py
@@ -16,12 +16,20 @@ import argparse
 import json
 import re
 import sys
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import Any
 
 ROOT_DIR = Path(__file__).resolve().parents[2]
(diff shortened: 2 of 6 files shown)


Git-on-my-level:
Thanks for tightening this up — this looks like a useful portability fix for the OpenAPI client generators.

I reviewed the diff and ran the generator checks locally in a scrubbed environment with `LANG=C`, `LC_ALL=C`, `PYTHONCOERCECLOCALE=0`, and `PYTHONUTF8=0`:

- `python backend/scripts/generate_ts_openapi_types.py --check`
- `python backend/scripts/generate_swift_openapi_types.py --check`
- `python backend/scripts/generate_dart_models.py --all --check`

Those passed, and the Swift/TypeScript generator unit tests also passed locally (`15 passed`). The changes are appropriately scoped to explicit UTF-8 file I/O, LF writes, stable slash-separated path labels, and regression coverage. I did not see a security or supply-chain concern in the implementation.

I'm not formally approving from automation because this touches generator/pre-push-sensitive tooling and the review policy requires human maintainer review for that surface, but this is a positive signal from my side.

---

kodjima33 (approved):

kodjima33 (approved):
backend bug fix (conf 4/5): make OpenAPI client generators portable on Windows (explicit UTF-8 encoding + slash-stable paths); tests, CI green

PR #9764, opened 2026-07-14
fix(cli): enforce owner-only config ACLs on Windows

tianmind-studio:
## What changed and why

`omi-cli` stores API keys, OAuth tokens, and local bearer tokens in `config.toml`, whose contract is owner-only. On Windows, Python accepts POSIX `0o600` flags but creates a file without an equivalent owner-only DACL, so this change routes config writes through a secure-create primitive that applies and verifies the platform policy before serialization starts.

The POSIX path still creates files as `0600`. The Windows path supplies a protected owner-only DACL to `CreateFileW`, denies handle sharing during serialization, reads the effective DACL back through the open handle, and fails closed without leaving a credential file if enforcement cannot be verified. The DACL deliberately omits SYSTEM and Administrators so only the Windows object owner receives file access.

This update also adds path-scoped Python CLI CI on Linux 3.10/3.12 and Windows 3.12. The Windows leg executes the real Win32 ACL contract that Linux-only release CI could not cover.

none

## How it was verified

- CPython 3.10 on Windows: full `python -m pytest -q` suite passed (126 tests).
- CPython 3.12 on Windows: full `python -m pytest -q` suite passed (126 tests).
- CPython 3.12 focused Win32 path: secure final file, no world-readable serialization window, and fail-closed DACL verification all passed (3 tests).
- `actionlint .github/workflows/python-cli-ci.yml` passed.
- `black --check --line-length 120 --skip-string-normalization .` passed (36 files unchanged).
- `git diff --check` and 12 selected manifest checks passed locally.
- The remaining `check-manifest-contract` check is not runnable on current
  Windows `main`: its own platform self-test accepts only `all`, `linux`, and
  `macos`. The Linux repository-check lane remains authoritative for that check.

The tests exercise `cfg.save()` against real temporary files. On Windows they inspect the live native handle while `tomli_w.dump` is running, so the evidence covers the pre-write ACL rather than only the final file.

## Tests

`sdks/python-cli/tests/test_config.py` covers the secure final file, the in-progress descriptor before credential bytes are written, stale temporary-file recovery, and fail-closed cleanup when Windows DACL verification is altered.

## Failure class (fixes)

Failure-Class: new

Violated contract: credential material must not be persisted unless the destination is created with enforceable owner-only access before the first byte is written.

Canonical prevention: all CLI config persistence now uses `omi_cli._secure_file.open_owner_only`, which applies the native policy atomically and verifies it before serialization. The behavioral contract lives in `sdks/python-cli/tests/test_config.py`, and `.github/workflows/python-cli-ci.yml` runs it on a real Windows host. PR #9764 is the evidence instance.

This Windows CI leg would have caught the permission defect evidenced by PR #9764. The production fix is already a shared primitive for every `omi-cli` config write; the CI leg is still required because a POSIX runner cannot execute or validate Win32 ACL calls.

### DIFF
--- a/.github/failure-classes/FC-private-credential-file-permissions.json
+++ b/.github/failure-classes/FC-private-credential-file-permissions.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-private-credential-file-permissions",
+  "violated_contract": "Credential material must not be persisted unless the destination is created with enforceable owner-only access before the first byte is written.",
+  "canonical_prevention": "Route credential persistence through a platform-specific secure-create primitive that applies owner-only permissions atomically, verifies the effective policy before writing, and runs its behavioral contract on every supported operating system.",
+  "canonical_prevention_artifact": [
+    "sdks/python-cli/tests/test_config.py",
+    ".github/workflows/python-cli-ci.yml"
+  ],
+  "evidence_prs": [9764],
+  "scope_hints": [
+    "sdks/python-cli/**",
+    "backend/scripts/*credential*.py",
+    "backend/utils/*credential*.py"
+  ],
+  "status": "open"
+}
--- a/.github/workflows/python-cli-ci.yml
+++ b/.github/workflows/python-cli-ci.yml
@@ -1,0 +1,54 @@
+name: Python CLI
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'sdks/python-cli/**'
+      - '.github/workflows/python-cli-ci.yml'
+  pull_request:
+    branches: [main]
+    paths:
+      - 'sdks/python-cli/**'
+      - '.github/workflows/python-cli-ci.yml'
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    working-directory: sdks/python-cli
+
+jobs:
+  tests:
+    name: ${{ matrix.label }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - label: Linux / Python 3.10
+            os: ubuntu-latest
+            python-version: '3.10'
+          - label: Linux / Python 3.12
+            os: ubuntu-latest
+            python-version: '3.12'
+          - label: Windows ACL / Python 3.12
+            os: windows-latest
+            python-version: '3.12'
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+          cache-dependency-path: sdks/python-cli/pyproject.toml
+
+      - name: Install package and test dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -e ".[dev]"
+
+      - name: Run tests
+        run: python -m pytest -q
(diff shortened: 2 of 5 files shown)


Git-on-my-level:
Thanks @tianmind-studio — this is a genuinely well-built hardening for Windows credential files. A few observations for a maintainer before this merges.

What I like
- The fail-closed approach is exactly right: `CreateFileW` with `CREATE_NEW`, share mode `0`, a protected owner-only DACL (`D:P(A;;FA;;;OW)`), then re-verifying the effective DACL through the open handle before any credential byte is written. If verification fails, the temp file is deleted and no credential is persisted. That's the correct shape for closing a TOCTOU window.
- The regression test that inspects the temp descriptor mid-serialization (by monkeypatching `tomli_w.dump`) rather than only after-the-fact is the right way to catch the "file is world-readable while bytes are being written" class of bug.
- POSIX path is untouched, so no regression risk there.

Why I'm flagging for human review rather than approving
- **The security-critical path has no CI coverage.** The repo CI is Linux/macOS, and the new Windows assertion is `@pytest.mark.skipif(os.name != "nt")`. So the actual ACL behavior is validated only on a local Windows run, and a subtle ctypes issue (argtypes wiring, `HANDLE(-1)` comparison, `GetSecurityInfo` path, `open_osfhandle` semantics, memory-freeing order) would pass CI silently. For a credential-permission change I'd want that validated on a Windows host or by a reviewer with Windows sign-off.
- **Owner-only DACL excludes SYSTEM/Administrators.** `D:P(A;;FA;;;OW)` grants only the file owner full access — which is the right default for a credential file, but it's a backup/AV-visibility trade-off worth a maintainer's explicit nod.
- This touches how secrets-at-rest are protected, so it's a security/permissions surface that needs human judgment before merge rather than an AI approval.

Suggested next steps
- If feasible, add a Windows leg to CI (or at least confirm the full test suite passes on Windows in the PR). Confirming `_secure_file`'s Win32 calls behave as intended under a real Windows environment is the highest-value check here.
- Consider a one-line docstring note in `_secure_file.py` that the DACL deliberately omits SYSTEM/Administrators and why, so future readers understand the trade-off.

Marking this for security/maintainer review. Positive direction overall — just needs a Windows-confirmed human sign-off before merge.

---

tianmind-studio:
Rebased onto current main and added a path-scoped Python CLI matrix with a native Windows 3.12 leg; the full suite passes on Windows Python 3.10 and 3.12 (126 tests each), and the three focused Win32 ACL regressions pass. I also documented that the DACL intentionally grants only the Windows object owner, without SYSTEM or Administrators ACEs. The local pre-push hook still hits current main's signal.SIGHUP Windows crash, so after the component suite and 12 runnable manifest checks passed I used --no-verify; GitHub CI is authoritative for the remaining manifest-contract check.

Git-on-my-level:
Thanks @tianmind-studio — I re-reviewed the current head after your update.

The previous automation feedback is addressed from my side:
- The PR now adds a path-scoped Python CLI workflow with a native Windows 3.12 leg, and GitHub reports the Windows ACL job passing on this head.
- `_secure_file.py` documents the intentional owner-only Windows DACL behavior.
- I also ran the Linux `tests/test_config.py` suite locally in a stripped environment; it passed, with the Windows-only fail-closed test skipped as expected on Linux.

I do not see a new AI-blocking code issue in the current diff. The approach remains directionally good: create the credential temp file exclusively, apply/verify the owner-only policy before serialization, fail closed with cleanup if verification fails, then atomically replace the config file.

I am still not formally approving because this is a security/permissions surface and the PR intentionally changes secret-at-rest behavior on Windows. Please keep `security-review` / `needs-maintainer-review` until a human maintainer is comfortable with the DACL trade-off and the new workflow. From the automation side, this is a positive signal rather than a request for changes.

---

PR #9778, opened 2026-07-15
Bound the gateway BYOK LLM-client cache so it cannot grow unbounded

ZachL111:
## What

`utils.llm.gateway_byok` caches the gateway BYOK LLM client per cache key:

```python
_byok_gateway_llm_cache: Dict[tuple[Any, ...], Any] = {}
...
cached = _byok_gateway_llm_cache.get(cache_key)
if cached is not None:
    return cached
...
client = ChatOpenAI(model=lane_id, **kwargs)
_byok_gateway_llm_cache[cache_key] = client
```

The `cache_key` is built from `(base_url, service_token_cache_key, provider, key_fingerprint, request_timeout, max_retries)`, and `key_fingerprint` is derived per BYOK key. So a long-lived process serving many distinct BYOK credentials accumulates one `ChatOpenAI` client per key, and nothing ever evicts them. The map grows for the entire process lifetime. This is exactly the case the repo's own guidance calls out: "Module-level dicts: add TTL-based eviction or cap size - they grow forever otherwise."

## Fix

Route the write through a small bounded helper that evicts the oldest entry once a cap is reached:

```python
_MAX_BYOK_GATEWAY_CLIENTS = 256

def _remember_byok_client(cache_key: tuple[Any, ...], client: Any) -> None:
    # Re-inserting refreshes recency; evict the oldest entry once the cap is reached.
    _byok_gateway_llm_cache.pop(cache_key, None)
    while len(_byok_gateway_llm_cache) >= _MAX_BYOK_GATEWAY_CLIENTS:
        _byok_gateway_llm_cache.pop(next(iter(_byok_gateway_llm_cache)), None)
    _byok_gateway_llm_cache[cache_key] = client
```

Dict insertion order (Python 3.7+) makes `next(iter(...))` the oldest key. Re-inserting a key refreshes its recency, so an actively-used client stays cached. The read path (cache hit / miss) is unchanged; only the write is bounded. The cap of 256 comfortably covers realistic concurrent BYOK usage while bounding worst-case client retention.

## Tests

New `tests/unit/test_byok_gateway_cache_cap.py`:

- writing past the cap keeps the map at exactly the cap, evicts the oldest key, and retains the most recent

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_byok_gateway_cache_cap.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check utils/llm/gateway_byok.py tests/unit/test_byok_gateway_cache_cap.py
  -> unchanged
python -m pyright -p pyrightconfig.json utils/llm/gateway_byok.py
  -> 0 errors
  -> Checked 660 test file(s); 0 violations
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_byok_gateway_cache_cap.py
+++ b/backend/tests/unit/test_byok_gateway_cache_cap.py
@@ -1,0 +1,40 @@
+"""Regression: the gateway BYOK LLM-client cache must stay bounded.
+
+utils.llm.gateway_byok caches a ChatOpenAI client per cache key, and the cache key includes a
+per-BYOK-key fingerprint, so a long-lived process serving many distinct BYOK credentials accumulated
+a client per key for the whole process lifetime (the repo's rule requires module-level dicts to cap
+or TTL). _remember_byok_client caps the map and evicts the oldest entry.
+"""
+
+import pytest
+
+import utils.llm.gateway_byok as gw
+
+
+@pytest.fixture(autouse=True)
+def _clear_cache():
+    gw._byok_gateway_llm_cache.clear()
+    yield
+    gw._byok_gateway_llm_cache.clear()
+
+
+def test_cache_is_bounded_and_evicts_oldest():
+    cap = gw._MAX_BYOK_GATEWAY_CLIENTS
+    for i in range(cap + 25):
+        gw._remember_byok_client((f"key-{i}",), f"client-{i}")
+
+    assert len(gw._byok_gateway_llm_cache) == cap
+    # oldest evicted, most recent retained
+    assert ("key-0",) not in gw._byok_gateway_llm_cache
+    assert gw._byok_gateway_llm_cache[(f"key-{cap + 24}",)] == f"client-{cap + 24}"
+
+
+def test_rewriting_a_key_refreshes_without_growth():
+    for i in range(5):
+        gw._remember_byok_client((f"k{i}",), f"c{i}")
+    before = len(gw._byok_gateway_llm_cache)
+
+    gw._remember_byok_client(("k0",), "c0-new")
+
+    assert len(gw._byok_gateway_llm_cache) == before  # updating an existing key does not grow the map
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @ZachL111 — good catch. An unbounded module-level dict keyed by per-BYOK-key fingerprint is a real slow leak on a long-lived gateway process, and bounding it is the right call.

A few notes from review:

- **Eviction logic is correct.** Popping the key before re-insert refreshes recency (LRU-ish), `next(iter(d))` is the oldest key under Python 3.7+ insertion ordering, and the `>=` guard ensures the map never exceeds the cap. The read path is untouched, so cache hits behave exactly as before.
- **Tests cover the right things** — cap enforcement, oldest-evicted/most-recent-retained, and no-growth-on-rewrite. The autouse fixture clearing the module dict keeps them isolated.
- **One minor observation** (not blocking): cache *reads* do not refresh recency, so this is FIFO rather than true LRU. That is fine here — a `ChatOpenAI` client is cheap to recreate, and bounding worst-case retention is the goal. If you ever want true LRU, a `move_to_end` on cache hit would do it, but it is not worth the added coupling now.
- **Checks are green on this head** (the earlier cancelled Hygiene/Formatting/Detect-Changes runs were superseded by successful reruns). Backend unit suite and Hermetic Backend E2E both pass.

Flagging `candidate-clean-pr` for a maintainer to give the final approval. Nicely scoped.

---

kodjima33 (approved):

PR #9785, opened 2026-07-15
fix(memory): preserve canonical lifecycle boundaries

Git-on-my-level:
## Summary

- Make canonical lifecycle authority explicit in the desktop cache and UI, and reconcile local captures from the server's authoritative receipt.
- Fail canonical-selected mutations closed rather than silently writing legacy records when canonical routing is unavailable.
- Preserve retryable canonical extraction: persisted finalizers own deferred extraction before fan-out, direct canonical callers surface retryable errors, and X raw sources replay unacknowledged extraction after transient failures.

The desktop cache treated a missing lifecycle as an implied long-term tier, while temporarily unavailable canonical writers could fall back to legacy. Background extraction could also discard a failed canonical write after its source had been marked seen. Explicit read scopes and authoritative receipts, closed mutation routing, synchronous canonical extraction for direct callers, finalizer-owned retries, and post-success source acknowledgement close that failure class.

## Scope and migration

This does not migrate or delete user data. Legacy source rows remain intact. The revision-ledger migration stays separate and must be dry-run capable, incremental, and safe to operate in bulk.

- INV-AUTH-1
- INV-MEM-1

## Verification

- `cd desktop/macos && OMI_SWIFT_TEST_SUITE_WORKERS=4 xcrun swift test --package-path Desktop` (2,526 passed)
- `cd backend && BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-backend-unit-failures.txt bash test.sh` (the affected backend files and the timing-gated runtime-contract file passed)
- Launched the named `omi-canonical-memory-contract` bundle against dev; the Memory view displayed explicit lifecycle-tagged rows.
- Independent final Grok review passed after the direct-caller and retry-boundary fixes.
- `scripts/pr-preflight --pr-body-file /tmp/canonical-memory-pr.md`

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9785?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/x_posts.py
+++ b/backend/database/x_posts.py
@@ -21,6 +21,8 @@ logger = logging.getLogger(__name__)
 
 users_collection = 'users'
 x_posts_collection = 'x_posts'
+MEMORY_EXTRACTION_PENDING = 'pending'
+MEMORY_EXTRACTION_COMPLETED = 'completed'
 
 # Post "kinds" stored under the same collection, distinguished by the `kind` field.
 KIND_TWEET = 'tweet'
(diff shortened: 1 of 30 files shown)


kodjima33 (approved):

Issue #9790, opened 2026-07-15
macOS Rewind: reopen indexer database after recovery close


Git-on-my-level:
## Bug

After repeated SQLite I/O or corruption errors, `RewindDatabase` closes its pool. `RewindIndexer` retained a separate initialized flag and continued past initialization, producing repeated `databaseNotInitialized` errors until periodic cleanup retried the database (up to six hours later).

## Root cause

The indexer cache did not validate the database readiness that is authoritative after self-recovery closure.

## Acceptance criteria

- [ ] Rewind initialization reopens a database that closed after the indexer previously initialized.
- [ ] A focused lifecycle regression test fails on the stale-cache behavior and passes after repair.
- [ ] Existing Rewind database lifecycle tests remain green.

No user identifiers, event payloads, or credentials included.

PR #9792, opened 2026-07-15
fix(desktop): reopen Rewind DB when indexer cache goes stale after self-recovery (#9790)

kodjima33:
## Bug (#9790)

After the Rewind database self-closes following repeated SQLite I/O / corruption errors, the timeline silently stops recording. `RewindIndexer` retained its own `isInitialized = true` flag and kept processing frames, producing repeated `databaseNotInitialized` errors until the 6-hourly retention cleanup happened to reopen the database.

## Root cause

`RewindDatabase.reportQueryError` closes the pool (`close()`) after `maxQueryIOErrorsBeforeRecovery` (5) consecutive recoverable errors, so the next `initialize()` goes through the full WAL/corruption recovery path. But `RewindIndexer.ensureInitialized()` short-circuited on its **own** cached flag and never revalidated against the database's authoritative readiness. Once the DB closed, every frame's `insertScreenshot` threw `databaseNotInitialized`; the caught error routed back to `reportQueryError`, which early-returns when `dbQueue == nil`, so nothing ever re-triggered `initialize()`. Only `runCleanup` (≤ every 6h) called `RewindDatabase.shared.initialize()` directly and reopened it.

## Fix

`ensureInitialized()` now treats `RewindDatabase.shared.isInitialized` as the authoritative signal: when the cached flag is set but the database has closed, it drops the stale flag and reinitializes through the existing exponential-backoff path. Surgical — no new state or fallback booleans; the database stays the single owner of readiness.

## Test

Added `testIndexerReopensDatabaseAfterSelfRecoveryClose` (in `RewindDatabaseLifecycleTests`) which drives the production `ensureInitialized()`: initialize → `close()` (simulating self-recovery) → assert the next `ensureInitialized()` reopens the database. It fails on the stale-cache behavior and passes after the fix. Existing lifecycle tests remain green.

## Verification

- `xcrun swift build -c debug --package-path Desktop` → Build complete
- `xcrun swift test --package-path Desktop --filter RewindDatabaseLifecycleTests` → 3 tests, 0 failures

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9792?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Rewind/Services/RewindIndexer.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Services/RewindIndexer.swift
@@ -88,8 +88,20 @@ actor RewindIndexer {
     }
 
     /// Try to initialize with exponential backoff. Returns true if initialized.
-    private func ensureInitialized() async -> Bool {
-        if isInitialized { return true }
+    func ensureInitialized() async -> Bool {
+        if isInitialized {
+            // The database is the authoritative readiness signal. It closes its own
+            // pool after repeated I/O/corruption errors (RewindDatabase.reportQueryError);
+            // this cached flag then goes stale and every frame fails with
+            // databaseNotInitialized until periodic cleanup reopens it up to 6h later.
+            // Revalidate against the database and drop the stale flag so this frame
+            // reopens it through the normal backoff path below.
+            if await RewindDatabase.shared.isInitialized {
+                return true
+            }
+            isInitialized = false
+            log("RewindIndexer: Database closed after initialization (self-recovery); reinitializing")
+        }
 
         // Check backoff timer - skip if too soon after last failure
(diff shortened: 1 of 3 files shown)


kodjima33:
Thank you 🙏 — duplicate of the #9790 fix that already landed on `main` (the indexer now reconciles its cached `isInitialized` against the live `RewindDatabase` state). Closing this one, but please keep them coming!

PR #9794, opened 2026-07-15
Enforce the webhook-disabled cache size cap on every write path

ZachL111:
## What

`database.webhook_health` caches disabled-app state in a module-level dict with a safety size cap:

```python
_CACHE_MAX_SIZE = 10000
_disabled_cache: dict[str, tuple[bool, float, int]] = {}  # (value, timestamp, generation)
```

The cap is enforced in one place, the read path `is_app_webhook_disabled`:

```python
_disabled_cache[app_id] = (result, now, pre_gen)
if len(_disabled_cache) > _CACHE_MAX_SIZE:
    _evict_oldest(_disabled_cache)
```

But three other write paths set the cache directly without the cap: the failure/disable path in `record_app_webhook_failure`, the re-enable path, and the second disable path. Each does:

```python
gen = _disabled_cache.get(app_id, (False, 0, 0))[2] + 1
_disabled_cache[app_id] = (True, time.monotonic(), gen)
```

So a burst of disables/re-enables can push `_disabled_cache` past `_CACHE_MAX_SIZE` even though the author added the cap to keep it bounded.

## Fix

Centralize the write in a helper that bumps the generation, stores the entry, and enforces the cap, then route the three uncapped sites through it:

```python
def _set_disabled_state(app_id: str, value: bool) -> None:
    """Record a disabled-state write and enforce the size cap. Caller must hold _cache_lock."""
    gen = _disabled_cache.get(app_id, (False, 0, 0))[2] + 1
    _disabled_cache[app_id] = (value, time.monotonic(), gen)
    if len(_disabled_cache) > _CACHE_MAX_SIZE:
        _evict_oldest(_disabled_cache)
```

The helper is called inside the existing `with _cache_lock:` sections, so it does not lock (same contract as `_evict_oldest`). The read path is left as-is (it already caps and uses its own timestamp/generation).

Scope, stated plainly: `_disabled_cache` is keyed by `app_id`, so it is bounded by the number of apps and this is not a live unbounded leak. It closes the gap where the author's own `_CACHE_MAX_SIZE` cap was enforced on only one of four write paths. Same module-dict cap class as #9713 and #9778.

## Tests

New `tests/unit/test_webhook_health_cache_cap.py`:

- with a small cap, writing many distinct app ids keeps the cache within the cap (before the fix these paths never evicted)
- re-writing an existing app id increments its generation and updates the value without growing the map

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_webhook_health_cache_cap.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check database/webhook_health.py tests/unit/test_webhook_health_cache_cap.py
  -> unchanged
python -m pyright -p pyrightconfig.json database/webhook_health.py
  -> 0 errors
  -> Checked 660 test file(s); 0 violations
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/webhook_health.py
+++ b/backend/database/webhook_health.py
@@ -36,6 +36,18 @@ def _evict_oldest(d: Dict[str, Any]) -> None:
         del d[k]
 
 
+def _set_disabled_state(app_id: str, value: bool) -> None:
+    """Record a disabled-state write and enforce the size cap. Caller must hold _cache_lock.
+
+    The is_app_webhook_disabled read path already caps the cache; the record/re-enable write paths
+    did not, so routing every write through here keeps the cache within _CACHE_MAX_SIZE on all paths.
+    """
+    gen = _disabled_cache.get(app_id, (False, 0, 0))[2] + 1
+    _disabled_cache[app_id] = (value, time.monotonic(), gen)
+    if len(_disabled_cache) > _CACHE_MAX_SIZE:
+        _evict_oldest(_disabled_cache)
+
+
 # Lua script: atomically record a failure and return graduated response action.
 # Returns: 0 = no action, 1 = day1 warn, 2 = day2 warn, 3 = disable
 _RECORD_FAILURE_LUA = """
@@ -143,8 +155,7 @@ def record_app_webhook_failure(app_id: str, status_code: int, error: str, endpoi
         if action == 3:
             r.setex(f'app_webhook_disabled:{app_id}', _HEALTH_TTL, '1')
             with _cache_lock:
-                gen = _disabled_cache.get(app_id, (False, 0, 0))[2] + 1
-                _disabled_cache[app_id] = (True, time.monotonic(), gen)
+                _set_disabled_state(app_id, True)
         return action
     except Exception as e:
         logger.warning(f'record_app_webhook_failure redis error app_id={app_id}: {e}')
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Maintainer automation review (positive signal, not a formal approval):

Thanks for tightening this up. I reviewed the webhook-disabled cache change and the new regression test. Centralizing the disabled-state writes through `_set_disabled_state()` preserves the existing generation-bump behavior while enforcing `_CACHE_MAX_SIZE` on the write paths that previously bypassed eviction (`record_app_webhook_failure`, `clear_app_webhook_health`, and `disable_app_in_firestore`). The helper is used under the existing `_cache_lock` call sites, so I don't see a new race or behavior change beyond bounding the cache.

Validation notes:
- The added regression test covers repeated write-path growth and same-key generation updates.
- I also ran a local import-stubbed cache/generation exercise plus `compileall` for the touched Python files successfully.
- Direct local pytest collection is blocked in this stripped review environment by missing backend dependencies (`google`/`httpx`), but the GitHub backend unit suite is passing for this head.

I don't see blocking code issues from this review. I am not formally approving because the automation approval gate disallows approval for security-sensitive webhook-disable/cache paths (and the context includes non-green/superseded check noise), so this should still get a human maintainer's final sign-off before merge.



---

kodjima33 (approved):

PR #9796, opened 2026-07-15
Tolerate a malformed announcement sub-document instead of 500ing the list

ZachL111:
## What

`models.announcement.Announcement.from_dict` was deliberately hardened so a single malformed or legacy announcement document cannot 500 the whole public announcements list. It tolerates a bad `type` (falls back to `ANNOUNCEMENT`), and a missing `id`/`created_at`:

```python
try:
    announcement_type = AnnouncementType(raw_type)
except ValueError:
    announcement_type = AnnouncementType.ANNOUNCEMENT
...
announcement_id = cast(str, data.get("id")) or ""
created_at = cast(datetime, data.get("created_at")) or datetime.fromtimestamp(0, tz=timezone.utc)
```

But the `targeting`/`display` sub-models were still built unguarded:

```python
targeting=Targeting(**targeting_data) if targeting_data else None,
display=Display(**display_data) if display_data else None,
```

`Targeting` and `Display` are pydantic models, so a stored sub-dict with a bad value (`targeting.trigger="foo"`, `display.priority="abc"`, `display.start_at="not-a-date"`) raises `ValidationError` out of `from_dict`. The `database.announcements` helpers loop `from_dict` over documents with no per-item try/except (`get_pending_announcements`, `get_general_announcements`, and others), so one malformed sub-document 500s the entire user-facing list for everyone.

## Fix

Extend the same tolerance already applied to `type` to the sub-models, via a small helper:

```python
def _optional_submodel(model_cls: type[_SubModel], sub_data: Optional[Mapping[str, Any]]) -> Optional[_SubModel]:
    if not sub_data:
        return None
    try:
        return model_cls(**sub_data)
    except ValidationError:
        return None
```

```python
targeting=_optional_submodel(Targeting, targeting_data),
display=_optional_submodel(Display, display_data),
```

On `ValidationError` the malformed sub-object is dropped (`None`) and the announcement is kept, matching the existing "one bad document must not 500 the list" intent. Valid sub-dicts parse exactly as before, and an empty/missing sub-dict stays `None`.

## Tests

New `tests/unit/test_announcement_subdoc_guard.py`:

- a malformed `targeting` (bad enum) or `display` (bad int/datetime) is dropped to `None` while the announcement is kept
- valid `targeting`/`display` are parsed unchanged
- missing sub-docs stay `None`

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_announcement_subdoc_guard.py -q
  -> 4 passed
black --line-length 120 --skip-string-normalization --check models/announcement.py tests/unit/test_announcement_subdoc_guard.py
  -> unchanged
python -m pyright -p pyrightconfig.json models/announcement.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9796?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/models/announcement.py
+++ b/backend/models/announcement.py
@@ -1,8 +1,8 @@
 from datetime import datetime, timezone
 from enum import Enum
-from typing import Any, List, Mapping, Optional, cast
+from typing import Any, List, Mapping, Optional, TypeVar, cast
 
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 
 
 class AnnouncementType(str, Enum):
@@ -107,6 +107,21 @@ class AnnouncementContent(BaseModel):
 
 
 # Main announcement model
+_SubModel = TypeVar("_SubModel", bound=BaseModel)
+
+
+def _optional_submodel(model_cls: type[_SubModel], sub_data: Optional[Mapping[str, Any]]) -> Optional[_SubModel]:
+    # Tolerate a malformed targeting/display sub-document (a bad enum, a bad datetime) the same way a
+    # bad top-level type is tolerated in from_dict: drop the sub-object rather than let one legacy
+    # announcement 500 the whole list (the DB helpers loop from_dict with no per-item guard).
+    if not sub_data:
+        return None
+    try:
+        return model_cls(**sub_data)
+    except ValidationError:
+        return None
+
+
 class Announcement(BaseModel):
     id: str
     type: AnnouncementType
@@ -187,8 +202,8 @@ class Announcement(BaseModel):
             firmware_version=data.get("firmware_version"),
             device_models=data.get("device_models"),
             expires_at=data.get("expires_at"),
-            targeting=Targeting(**targeting_data) if targeting_data else None,
-            display=Display(**display_data) if display_data else None,
+            targeting=_optional_submodel(Targeting, targeting_data),
+            display=_optional_submodel(Display, display_data),
             content=data.get("content", {}),
         )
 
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Nice, clean availability fix — thanks @ZachL111.

Verified independently:
- Confirmed the bug: on `main`, `from_dict` raises `pydantic.ValidationError` on a malformed targeting/display sub-document (e.g. `targeting.trigger="bogus"`), and `database/announcements.py` loops `from_dict` with no per-item guard, so one bad sub-doc 500s the whole list.
- Ran the 4 new unit tests in a clean env — all pass. Coverage hits the right cases (malformed targeting dropped, malformed display dropped, valid kept unchanged, missing -> `None`).
- The `_optional_submodel` helper is a good fit: it extends the same tolerance already applied to `type`/`id`/`created_at`, the `TypeVar[BaseModel]` typing is sound, and the behavior is minimal and predictable.

No concerns on correctness, scope, security, or test adequacy. The swallow-and-drop is the right availability tradeoff here.

One optional, non-blocking suggestion: a `debug` log when a sub-document is dropped would help operators notice silent data corruption in stored announcements.

Leaving a positive signal rather than formal approval only because of a checks-status gate (some Formatting/Hygiene/Detect Changes runs show as cancelled/superseded; the latest runs are green). A maintainer can approve on confirm.

---

kodjima33 (approved):

PR #9797, opened 2026-07-15
Skip a topic missing its 'topic' field instead of dropping the whole trends category

ZachL111:
## What

`database.trends.get_trends_data` builds the public `/v1/trends` response by looping categories and their topics. It was hardened so a topic doc missing `memory_ids` (the two-phase `save_trends` writes the doc, then updates `memory_ids` with `ArrayUnion`, so an interrupted second write leaves it absent) does not raise:

```python
topics = sorted(topics_docs, key=lambda e: len(e.get('memory_ids') or []), reverse=True)
for topic in topics:
    if topic['topic'] not in valid_items:   # required-key lookup
        continue
    topic['memories_count'] = len(topic.get('memory_ids') or [])
```

But `topic['topic']` is still a required-key lookup. A topic document missing the `topic` field raises `KeyError`, which the per-category `try/except` catches by `continue`-ing, so the **entire category** is silently dropped from `/v1/trends` (not a 500, but the whole category vanishes for every user).

## Fix

Use `.get`, mirroring the `memory_ids` treatment directly above:

```python
if topic.get('topic') not in valid_items:
    continue
```

A topic missing `topic` yields `None`, which is not in `valid_items`, so that one topic is skipped and the rest of the category survives. Valid topics are unaffected.

## Tests

New `tests/unit/test_trends_topic_key_guard.py` (monkeypatches the module `db` proxy with a fake trends store, since `get_trends_data` takes no injectable client):

- a category with one topic missing `topic` and one valid topic is still returned, with the valid topic kept and the malformed one skipped (before the fix the `KeyError` dropped the whole category)

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_trends_topic_key_guard.py -q
  -> 1 passed
black --line-length 120 --skip-string-normalization --check database/trends.py tests/unit/test_trends_topic_key_guard.py
  -> unchanged
python -m pyright -p pyrightconfig.json database/trends.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9797?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/trends.py
+++ b/backend/database/trends.py
@@ -42,7 +42,10 @@ def get_trends_data() -> List[Dict[str, Any]]:
             # KeyError and drop the entire category from the public /v1/trends response.
             topics = sorted(topics_docs, key=lambda e: len(e.get('memory_ids') or []), reverse=True)
             for topic in topics:
-                if topic['topic'] not in valid_items:
+                # A topic doc can be missing 'topic' the same way it can be missing 'memory_ids'
+                # (guarded above); use .get so one such topic is skipped rather than raising KeyError
+                # and dropping the entire category from the public /v1/trends response.
+                if topic.get('topic') not in valid_items:
                     continue
                 topic['memories_count'] = len(topic.get('memory_ids') or [])
                 topic.pop('memory_ids', None)
--- a/backend/tests/unit/test_trends_topic_key_guard.py
+++ b/backend/tests/unit/test_trends_topic_key_guard.py
@@ -1,0 +1,71 @@
+"""Regression: a topic doc missing its 'topic' field must not drop the whole trends category.
+
+database.trends.get_trends_data guards a missing 'memory_ids' on a topic (two-phase save can leave it
+absent), but still did a required-key lookup topic['topic']. A topic missing 'topic' raised KeyError,
+caught by the per-category try/except, which dropped the entire category from the public /v1/trends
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
## Maintainer review

Thanks for the focused regression fix. This looks like the right shape: switching the membership check to `topic.get('topic')` means one malformed Firestore topic document is skipped instead of raising `KeyError` and losing the whole trends category, and the new unit test covers that exact failure mode.

I also verified locally in a clean checkout:

```text
ENCRYPTION_SECRET=dummy python -m pytest tests/unit/test_trends_topic_key_guard.py -q
# 1 passed
black --line-length 120 --skip-string-normalization --check database/trends.py tests/unit/test_trends_topic_key_guard.py
# 2 files would be left unchanged
```

Positive signal from automation. I’m not formally approving because this run is under the maintainer guard for backend/public API data-handling paths and the cached automation context marked approval as disallowed, so a human maintainer should still make the merge decision.

---

kodjima33 (approved):

PR #9798, opened 2026-07-15
Bound the in-process custom rate-limit cache so distinct client IPs cannot grow it unbounded

ZachL111:
## What

`utils.other.endpoints.rate_limit_custom` (used via `rate_limit_dependency` as a FastAPI per-endpoint dependency) keeps rate-limit state in a module-level dict:

```python
cached: Dict[str, Any] = {}

def rate_limit_custom(endpoint, request, requests_per_window, window_seconds):
    ip = request.client.host if request.client else None
    key = f"rate_limit:{endpoint}:{ip}"
    ...
    cached[key] = json.dumps({"timestamp": timestamp, "remaining": remaining})
    return True
```

The dict is keyed by `{endpoint}:{ip}` and is only ever read and written — there is no eviction, TTL, or size cap. The window-expiry logic resets the counter for an existing key but never deletes stale keys. So every distinct client IP that hits an endpoint using this dependency adds an entry that stays for the whole process lifetime. Ordinary traffic accumulates entries, and an attacker rotating source IPs can inflate it deliberately. This is exactly the case the backend guide calls out (gotcha #11: "Module-level dicts: add TTL-based eviction or cap size - they grow forever otherwise").

## Fix

Route the write through a small bounded helper that evicts the oldest entries once over a cap:

```python
_MAX_RATE_LIMIT_ENTRIES = 100000

def _store_rate_limit(key: str, value: str) -> None:
    cached[key] = value
    if len(cached) > _MAX_RATE_LIMIT_ENTRIES:
        for stale in list(cached)[: len(cached) - _MAX_RATE_LIMIT_ENTRIES]:
            del cached[stale]
```

```python
_store_rate_limit(key, json.dumps({"timestamp": timestamp, "remaining": remaining}))
```

Dict insertion order (Python 3.7+) makes the leading keys the oldest, so eviction drops the least-recently-inserted entries. The read path, window reset, and fail-open on a corrupt entry are all unchanged. (The pre-existing comment claimed the write went to Redis; it does not, so the comment is corrected to say in-process cache.)

## Tests

New `tests/unit/test_rate_limit_cache_cap.py`:

- writing many distinct keys keeps the map within the cap, retains the newest, and evicts the oldest

## Verification

```
cd backend && ENCRYPTION_SECRET=... python -m pytest tests/unit/test_rate_limit_cache_cap.py -q
  -> 2 passed
black --line-length 120 --skip-string-normalization --check utils/other/endpoints.py tests/unit/test_rate_limit_cache_cap.py
  -> unchanged
python -m pyright -p pyrightconfig.json utils/other/endpoints.py
  -> 0 errors
```

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_rate_limit_cache_cap.py
+++ b/backend/tests/unit/test_rate_limit_cache_cap.py
@@ -1,0 +1,36 @@
+"""Regression: the in-process custom rate-limit cache must stay bounded.
+
+utils.other.endpoints.rate_limit_custom stores per-request state in the module-level `cached` dict
+keyed by "{endpoint}:{ip}". It was get/set only with no eviction, so a stream of distinct client IPs
+(ordinary traffic, or an attacker rotating IPs) grew the map without bound. _store_rate_limit caps it.
+"""
+
+import pytest
+
+import utils.other.endpoints as endpoints
+
+
+@pytest.fixture(autouse=True)
+def _clear():
+    endpoints.cached.clear()
+    yield
+    endpoints.cached.clear()
+
+
+def test_rate_limit_cache_is_bounded_and_keeps_newest(monkeypatch):
+    monkeypatch.setattr(endpoints, "_MAX_RATE_LIMIT_ENTRIES", 10)
+    for i in range(40):
+        endpoints._store_rate_limit(f"rate_limit:ep:{i}", "{}")
+
+    assert len(endpoints.cached) <= 10
+    assert "rate_limit:ep:39" in endpoints.cached  # newest retained
+    assert "rate_limit:ep:0" not in endpoints.cached  # oldest evicted
+
+
+def test_updating_an_existing_key_does_not_grow(monkeypatch):
+    monkeypatch.setattr(endpoints, "_MAX_RATE_LIMIT_ENTRIES", 10)
+    endpoints._store_rate_limit("k", "v1")
+    endpoints._store_rate_limit("k", "v2")
+
+    assert len(endpoints.cached) == 1
+    assert endpoints.cached["k"] == "v2"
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @ZachL111 — nice, targeted hardening. A few observations:

**Correctness of the fix**
- The unbounded-growth issue is real: `cached` is keyed by `{endpoint}:{ip}` with no eviction, so distinct client IPs (or an attacker rotating source IPs) grow the map for the whole process lifetime. Routing the write through `_store_rate_limit` with a 100k cap addresses it cleanly.
- Eviction semantics are correct for CPython 3.7+ ordered dicts: the leading keys are the oldest insertions, so `list(cached)[:overflow]` drops the least-recently-inserted entries. Updating an existing key re-inserts at the tail (so frequently-hit clients are not evicted), which matches the intent.
- All writes to `cached` now go through the helper — confirmed there is no remaining direct `cached[key] =` outside `_store_rate_limit`. The read path, window-reset, and fail-open-on-corrupt-entry logic are all unchanged, which is good.

**Notes for human review**
- The cap is a blunt FIFO, not LRU/TTL. An attacker who can rotate IPs can still force eviction of legitimate entries (temporarily resetting their rate-limit counters), which is a minor DoS-amplification angle rather than a bypass of the cap itself. That may be acceptable given this is an in-process secondary cache, but worth a maintainer call on whether a TTL-based eviction (or moving this state to Redis, which the old comment mistakenly claimed) is the better long-term direction.
- `must_not_approve` is set in the review context, so I'm leaving a positive signal rather than a formal approval. Human maintainer sign-off requested before merge.

Tests look solid (bounded + newest-retained, and update-does-not-grow). CI is green on the current head.

---

kodjima33 (approved):

PR #8882, opened 2026-07-02
fix(backend): skip a malformed chat message instead of 500ing the whole send

ZachL111:
## Summary

Fixes a poison-record bug on the chat send path: one malformed stored message made loading recent history 500, so the user could not send a chat message at all until the bad record was fixed.

## Root cause

The send path loads the last 10 messages for context with an unguarded comprehension:

```python
messages = list(reversed([Message(**msg) for msg in chat_db.get_messages(uid, limit=10, app_id=compat_app_id)]))
```

`Message` has required fields (`id`, `text`, `created_at`, `sender`, `type`). If any stored message is missing one of those or has a bad type, `Message(**msg)` raises a `ValidationError` that propagates out of the comprehension and 500s the whole send, so a single corrupt or legacy history record blocks chatting entirely.

## Fix

Add a `Message.deserialize_many_safe` classmethod that parses each record and skips the ones that fail validation, and use it for the history load. On a skip it logs a safe identifier (the message id) plus the exception class name only, not the `ValidationError` string, which would render user content.

## Testing

`tests/unit/test_message_deserialize_safe.py`: valid records are parsed, malformed records are skipped without losing the valid ones (and reported via the callback), a missing callback is tolerated, and empty input returns an empty list. The classmethod is pure, so the test needs no I/O.

## PR status

Mergeable, no conflicts. Additive resilience: valid history returns exactly as before; only a malformed record changes from 500-the-whole-send to skip-and-log.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/8882?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #9799, opened 2026-07-15
Skip malformed stored chat messages when building app proactive notifications

ZachL111:
## What

`_process_proactive_notification` in `utils/app_integrations.py` loads the last 10 app chat messages to fill the `{{user_chat}}` prompt slot, building them with:

```python
chat_messages = list(reversed([Message(**msg) for msg in get_app_messages(uid, app.id, limit=10)]))
```

One legacy or malformed stored record raises a `ValidationError`. The only caller runs this fire-and-forget through `run_blocking` and swallows exceptions, so the failure is silent: the proactive notification is dropped on every run for that user until the bad row ages out of the last-10 window. There is no 500 and no log at the drop site, so the user just stops getting notifications from that app.

## Fix

Route the batch through `Message.deserialize_many_safe`, the shared safe-deserialize helper added in #8882, which skips the bad record and keeps the rest. This closes the same failure class at its remaining reachable call site rather than patching one line: #8882 introduced the helper and converted the chat-history load, but this proactive-notification builder still did the unguarded comprehension.

## Tests

`tests/unit/test_app_integrations_safe_messages.py` drives `_process_proactive_notification` through two cases:

- a good record plus one malformed record (missing required fields)
- an all-malformed batch

Both assert the notification is still produced. Both fail against the pre-fix source with `ValidationError` and pass after the fix. The test keeps the real `Message` class (it does not stub `app_integrations.Message`), so it exercises the actual deserialization path.

## Verification

```
python -m pytest tests/unit/test_app_integrations_safe_messages.py -q
  -> 2 passed
  -> 2 failed (ValidationError) when the one-line source fix is reverted, test kept

  utils/app_integrations.py tests/unit/test_app_integrations_safe_messages.py
  -> clean

python -m pyright -p pyrightconfig.json utils/app_integrations.py
  -> 0 errors, 0 warnings

  -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9799?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_app_integrations_safe_messages.py
+++ b/backend/tests/unit/test_app_integrations_safe_messages.py
@@ -1,0 +1,78 @@
+"""Regression test: one malformed stored chat message must not abort the whole
+proactive-notification build.
+
+utils.app_integrations._process_proactive_notification loads the last 10 app chat
+messages to substitute into the {{user_chat}} prompt slot. It used to build them with
+[Message(**msg) for msg in get_app_messages(...)], so a single legacy or malformed
+stored record raised a ValidationError. The only production caller wraps this in a broad
+except (fire-and-forget via run_blocking), so the failure was silent: the proactive
+notification was dropped on every run for that user until the bad row aged out of the
+last-10 window. The fix routes the batch through Message.deserialize_many_safe (the
+shared safe-deserialize helper added in #8882), which skips the bad record and keeps the
+rest, so the notification is still built and sent.
+"""
+
+from datetime import datetime, timezone
+from unittest.mock import MagicMock
+
+import pytest
+
+import utils.app_integrations as app_int
+
+_GOOD_MESSAGE = {
+    "id": "good-1",
+    "text": "remember to drink water",
+    "created_at": datetime(2024, 1, 1, tzinfo=timezone.utc),
+    "sender": "human",
+    "type": "text",
+}
+# Missing text / created_at / sender / type -> Message(**record) raises ValidationError.
+_MALFORMED_MESSAGE = {"id": "legacy-broken"}
+
+
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for this — it closes the same poison-record failure class at its remaining reachable call site. `deserialize_many_safe` was added in #8882 specifically for this, and the proactive-notification builder was the last unguarded comprehension, so this is the right fix in the right place.

The tests are solid: keeping the real `Message` class (only stubbing dependencies) means they actually exercise the deserialization path, and covering both mixed-good+bad and all-bad cases is exactly what's needed.

One minor observation (not blocking): the chat-send path in `routers/chat.py` passes an `on_error` callback that logs a warning per skipped record:

```python
Message.deserialize_many_safe(
    ...,
    on_error=lambda record, exc: logger.warning(
        'Skipping malformed chat message %s for uid=%s: %s', ...
    ),
)
```

This call site doesn't, so malformed records are silently dropped with no log trail. Since the caller swallows exceptions broadly, there's no user-visible regression, but adding `on_error` logging would give parity and help diagnose bad records in production. Optional follow-up if you agree it's worth it.

Overall this looks good and ready for merge. (Automated review note: the repository's check summary initially flagged cancelled duplicate runs of Hygiene/Formatting/Detect Changes; the latest runs on this commit all passed, including Backend unit suite.)

---

kodjima33 (approved):

PR #9800, opened 2026-07-15
Add Person.deserialize_many_safe and use it in memory extraction

ZachL111:
## What

Several LLM prompt builders load people with:

```python
people = [Person(**p) for p in users_db.get_people_by_ids(uid, person_ids)]
```

`get_people_by_ids` returns raw stored documents. One malformed or legacy person document (for example one missing the required `name` field) makes `Person(**p)` raise a `ValidationError`, which aborts the whole people lookup. In `utils/llm/memories.py` that lookup feeds memory extraction, so a single bad person row breaks memory extraction for that conversation.

This is the same failure class #8882 fixed for `Message` by adding `Message.deserialize_many_safe`. `Person` had no equivalent guard, so the raw comprehension pattern is repeated across the LLM prompt builders.

## Fix

Add `Person.deserialize_many_safe`, mirroring `Message.deserialize_many_safe`: it builds `Person` objects from raw records, skips any that fail validation, keeps the rest, and calls an optional `on_error(record, exception)` for each skip.

Adopt it in the two memory-extraction call sites in `utils/llm/memories.py`. The same one-line swap applies to the other `Person(**p)` call sites (chat, trends, retrieval, conversation processing). Following the guidance to land the enforceable guard now rather than broadening a safe fix into a wide migration, this PR keeps the diff to the model guard plus the memory-extraction subsystem so it stays reviewable; the remaining sites can adopt the helper incrementally.

## Tests

`tests/unit/test_person_deserialize_many_safe.py` exercises the guard directly (imports only `models.other`, so it is fast and hermetic):

- a good + malformed + good batch keeps the two valid people and drops the malformed one
- `on_error` is called once per skipped record with the record and the exception
- an all-valid batch preserves order
- empty input returns an empty list

## Verification

```
python -m pytest tests/unit/test_person_deserialize_many_safe.py -q
  -> 4 passed

  models/other.py utils/llm/memories.py tests/unit/test_person_deserialize_many_safe.py
  -> clean

python -m pyright -p pyrightconfig.json models/other.py utils/llm/memories.py
  -> 0 errors

  -> 0 violations

import-verified utils/llm/memories.py after the swap (Person.deserialize_many_safe present)
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9800?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/models/other.py
+++ b/backend/models/other.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import List, Optional
+from typing import Any, Callable, Iterable, List, Mapping, Optional
 
 from pydantic import BaseModel, Field
 
@@ -43,3 +43,21 @@ class Person(BaseModel):
     speech_samples: List[str] = []
     speech_sample_transcripts: Optional[List[str]] = None
     speech_samples_version: int = 3
+
+    @classmethod
+    def deserialize_many_safe(
+        cls,
+        records: Iterable[Mapping[str, Any]],
+        on_error: Optional[Callable[[Mapping[str, Any], Exception], None]] = None,
+    ) -> List['Person']:
+        """Build Person objects from raw stored records, skipping any that fail validation so one
+        malformed or legacy person document cannot break a whole people lookup. on_error(record,
+        exception), when provided, is called for each skip. Mirrors Message.deserialize_many_safe."""
+        parsed: List['Person'] = []
+        for record in records:
+            try:
+                parsed.append(cls(**record))
+            except Exception as exc:  # noqa: BLE001 - one bad record must not break the lookup
+                if on_error is not None:
+                    on_error(record, exc)
+        return parsed
--- a/backend/tests/unit/test_person_deserialize_many_safe.py
+++ b/backend/tests/unit/test_person_deserialize_many_safe.py
@@ -1,0 +1,47 @@
+"""Regression test for Person.deserialize_many_safe.
+
+Several LLM prompt builders load people with
+[Person(**p) for p in users_db.get_people_by_ids(...)]. A single malformed or legacy stored
+person document (for example one missing the required name field) raised a ValidationError that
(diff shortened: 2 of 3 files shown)


Git-on-my-level (commented):
Nice, focused fix. `Person.deserialize_many_safe` is a faithful mirror of `Message.deserialize_many_safe` — same signature, same skip-and-continue semantics, same `on_error` callback. The memory-extraction call sites are the right place to adopt it first given the documented failure mode (one malformed legacy person doc breaks the whole people lookup and aborts memory extraction).

The unit tests cover the important cases: malformed record skipped while valid ones are kept, `on_error` invoked per skip, order preservation, and empty input. Checks are green on the latest head.

Two minor observations (neither blocking):

1. The two adoption sites in `memories.py` don't pass an `on_error` callback, so malformed records are skipped silently. The equivalent `Message.deserialize_many_safe` call in `routers/chat.py` passes `on_error=lambda record, exc: logger.warning(...)`. Consider adding a `logger.warning` callback here too — it makes diagnosing bad person documents much easier in production.

2. There are ~10 other `Person(**p)` call sites across chat, trends, retrieval, and conversation processing that still use the unguarded comprehension. The PR's choice to land the model guard + memory extraction first and migrate the rest incrementally is reasonable and keeps this reviewable.

Leaving a positive signal rather than a formal approval — this touches user-data deserialization in a production path, so a human maintainer should give the final go-ahead. The code itself looks correct and well-tested.

---

kodjima33 (approved):

PR #9801, opened 2026-07-15
fix(desktop): reopen Rewind database after self-recovery close (#9790)

### LINKED ISSUE #9790: macOS Rewind: reopen indexer database after recovery close

kodjima33:
Fixes #9790

## Bug
After repeated SQLite I/O or corruption errors, `RewindDatabase` self-recovers by closing its pool (`dbQueue = nil`, `isInitialized` → false). `RewindIndexer`, however, keeps its own cached `isInitialized = true` flag, so `ensureInitialized()` short-circuits and never reopens the database. Every captured frame then emits repeated `databaseNotInitialized` errors until the 6-hourly retention cleanup happens to call `RewindDatabase.initialize()` again — up to six hours of a dead Rewind timeline.

## Root cause
The indexer's cached readiness flag was not reconciled against the authoritative source, `RewindDatabase.shared.isInitialized`, after the database self-recovered/closed.

## Fix
Add `reconcileDatabaseReadiness()`: if the indexer thinks it's initialized but the database pool is actually closed, clear the stale flag so the next call reinitializes. It's invoked before both:
- `ensureInitialized()` — the per-frame indexing path (the hot path in the issue), and
- `rebuildFromVideoFiles()` — which had the same stale-flag exposure.

This reopens the pool on the very next frame instead of waiting for periodic cleanup.

## Test
`RewindDatabaseLifecycleTests.testIndexerReopensDatabaseAfterSelfRecoveryClose` drives the real `ensureInitialized()` through a self-recovery close and asserts the database reopens.
- Fails without the fix (DB stays closed) ✅ verified
- Passes with the fix ✅
- Full `RewindDatabaseLifecycleTests` suite green (3/3), `xcrun swift build` clean, `check_desktop_test_quality.py` OK.

🤖 automated by hourly watchdog; opened for review.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9801?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Rewind/Services/RewindIndexer.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Services/RewindIndexer.swift
@@ -87,8 +87,21 @@ actor RewindIndexer {
         }
     }
 
+    /// Reconcile the cached `isInitialized` flag against the authoritative
+    /// database readiness. `RewindDatabase` self-recovers by closing its pool
+    /// after repeated I/O or corruption errors; when it does, our cached flag is
+    /// stale and would otherwise short-circuit reinitialization, leaving frames
+    /// to emit repeated `databaseNotInitialized` errors until the 6-hourly
+    /// cleanup happens to call `RewindDatabase.initialize()` again.
+    private func reconcileDatabaseReadiness() async {
+        guard isInitialized, await !RewindDatabase.shared.isInitialized else { return }
+        log("RewindIndexer: Database closed after self-recovery; reinitializing")
+        isInitialized = false
+    }
+
     /// Try to initialize with exponential backoff. Returns true if initialized.
-    private func ensureInitialized() async -> Bool {
+    func ensureInitialized() async -> Bool {
+        await reconcileDatabaseReadiness()
         if isInitialized { return true }
 
         // Check backoff timer - skip if too soon after last failure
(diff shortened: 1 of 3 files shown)


kodjima33:
Thank you for this 🙏 — looks like #9790 is already handled on `main`: `RewindIndexer.ensureInitialized()` now reconciles its cached readiness against `RewindDatabase.shared.isInitialized`, so a self-recovery close reopens the DB. Closing this one, but please keep them coming!

PR #9804, opened 2026-07-15
fix(ci): complete Windows pre-push launch path

tianmind-studio:
## Summary

Completes the native-Windows launch path in #9730:

- pass the active Bash executable to `preflight_runner.py` before the extensionless `scripts/pre-push` path
- add a wrapper contract plus a real host-Bash launch test
- let the POSIX `killpg` simulations construct that missing API on Windows
- select the portability check whenever the wrapper changes

## Root cause and guard

After #9730 avoids the missing `signal.SIGHUP`, Python 3.11 on Windows reaches `subprocess.Popen` but `CreateProcess` cannot execute the extensionless Bash script directly. Binding the wrapper to its already-running `$BASH` preserves POSIX behavior and gives Windows an executable entrypoint. The manifest-backed regression test keeps that argv contract attached to future wrapper edits.

## Verification

- Before the wrapper change, the regression test failed because the first command after `--` was `pre-push`, not Bash.
- `python .github/scripts/test_preflight_runner.py` -> 11 tests OK on native Windows
- `python .github/scripts/test_run_checks.py` -> 9 tests OK
- Direct runner proof -> `windows-bash-launch-ok`
- Local PR preflight -> 9 selected checks passed
- Real `git push` reached and passed the portability/preflight/deployment-policy checks; the older #9730 base then stopped at its pre-existing `backend/.venv/bin/python` requirement.

Product invariants: none.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9804?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -19,9 +19,9 @@ checks:
     reason: "advisory desktop release evidence and drift report"
   - id: preflight-runner-portability
     command: ["python3", ".github/scripts/test_preflight_runner.py"]
-    triggers: [".github/scripts/preflight_runner.py", ".github/scripts/test_preflight_runner.py"]
+    triggers: [".github/scripts/preflight_runner.py", ".github/scripts/test_preflight_runner.py", "scripts/pre-push-singleflight"]
     lanes: ["local", "ci"]
-    reason: "single-flight pre-push runner must start on hosts without POSIX signal APIs"
+    reason: "single-flight pre-push runner must start and launch its Bash checks portably"
   - id: diff-hygiene
     command: ["python3", ".github/scripts/check_diff_hygiene.py", "--changed-files", "{changed_files}", "--base", "{base}", "--head", "{head}"]
     triggers: ["all"]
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Thanks for tightening the Windows pre-push path. I reviewed the diff and the shape looks good: the wrapper now invokes the extensionless Bash hook through the active Bash executable, the portability check is triggered when the wrapper changes, and the added tests cover both the wrapper contract and a real runner-to-Bash launch path.

I’m not formally approving because this touches CI/pre-push workflow plumbing and the PR targets a stacked/non-default base branch, so it should still get human maintainer review before merge. No changes requested from this automation review.

---

PR #9805, opened 2026-07-15
feat(desktop): AI Clone - reply to Telegram/WhatsApp/iMessage on your behalf from your Omi memories (via Beeper)

21prnv:
## What & why

New **AI Clone** Settings section: Omi replies to people in WhatsApp, Telegram, iMessage and more on your behalf, grounded in your Omi memories, through the local Beeper Desktop API. Per-chat trust ladder (Off / Draft / Ask me / Auto), where Auto unlocks only after a history-replay benchmark scores the clone against your own past replies. Opt-in, off by default.

## Invariants

- **INV-INT-1**: uses Beeper's documented localhost REST/WS contract, never UI automation/OCR; "Connected" is re-earned by a live probe each time; port self-discovered from `/v1/info`. No other locked invariants on the touched paths (`Sources/AIClone/**` plus one Settings case).

## Verification

- Full `xcrun swift test` green (new AI Clone suites included).
- Live on real Telegram (two phones): friend's message, clone read the thread, drafted an in-voice reply (matched Hinglish style), sent on approval.
- Live testing found and fixed 4 integration bugs, each with a regression test: real port is 23374 (self-discovered), events send `ts` as a string not a number, manual Beeper tokens are read-only (write needs OAuth scope), Beeper re-emits each message (deduped to one reply).

## Safety

Injection screening forces review in every mode; hourly auto-send budget with `recordFallback`; Auto gated on benchmark evidence.

## Follow-up (not in this PR)

One-click "Connect with Beeper" OAuth button (write token, no manual paste).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9805?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/AIClone/AICloneBenchmark.swift
+++ b/desktop/macos/Desktop/Sources/AIClone/AICloneBenchmark.swift
@@ -1,0 +1,129 @@
+import Foundation
+
+// MARK: - History-replay self-benchmark
+//
+// "Benchmark the agent against your own past decisions before letting go":
+// walk the chat's real history, find the points where the user actually
+// replied to someone, regenerate the clone's reply at that exact point (the
+// clone only sees the thread *before* the real reply), then have an LLM judge
+// score how close the clone came to what the user really sent. The per-chat
+// score gates Auto mode.
+
+struct AICloneBenchmarkSample: Equatable {
+  /// Thread lines before the user's real reply (oldest first).
+  var priorThreadLines: [String]
+  var inboundText: String
+  var inboundSenderName: String
+  /// What the user actually sent.
+  var actualReply: String
+}
+
+struct AICloneBenchmark {
+  var engine: AICloneReplyEngine
+  static let maxSamples = 8
+
+  /// Extracts replay samples: an inbound text message followed by the user's
+  /// own text reply. Messages must be oldest-first.
+  static func samples(from history: [BeeperMessage]) -> [AICloneBenchmarkSample] {
+    let usable = history.filter { $0.isDeleted != true && $0.isTextLike && !($0.text ?? "").isEmpty }
+    var collected: [AICloneBenchmarkSample] = []
+    for (index, message) in usable.enumerated() {
+      guard message.isSender == true, index > 0 else { continue }
+      let inbound = usable[index - 1]
+      guard inbound.isSender != true else { continue }
+      let prior = usable[..<(index - 1)].suffix(16).map { line -> String in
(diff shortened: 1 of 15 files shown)


Git-on-my-level:
This is an impressively well-structured feature PR — thank you for the care put into the safety design. Some notes for the maintainers and for follow-up:

**Strengths**
- The trust ladder (Off → Draft → Ask → Auto) with Auto gated on a benchmark score is a thoughtful escalation model.
- Injection detection via an explicit `suspected_injection` flag that blocks *all* modes (`.declinedInjection`) is good defense-in-depth.
- Auto-send budget (20/hr) with fallback-to-draft and diagnostic recording prevents runaway loops.
- Bounded activity log (200 entries, 140-char previews, 280-char reply truncation) avoids creating a second transcript store.
- Message dedup on message ID (not event ID) handles Beeper's multi-emit correctly.
- Functional connect probe (re-earned each time, not latched) and self-discovered port are robust.
- Good test coverage: hermetic transport stubs, trust-ladder/injection parsing tests.

**For maintainer product review**
- This introduces AI impersonation in real messaging conversations. The system prompt instructs the clone to "never mention being an AI" unless persona notes say otherwise, so recipients will not know they are talking to a clone. Whether Omi should ship this — and whether some form of disclosure or guardrail is appropriate — is a product/ethical decision for the maintainers.

**Engineering questions**
- Token storage uses Keychain on production builds but UserDefaults (plaintext) on non-production builds. The comment explains the re-signing/ACL rationale, but a maintainer should explicitly sign off on this posture. Consider whether the non-prod path should at least obfuscate or use a restricted-access UserDefaults suite.
- The benchmark uses an LLM-as-judge to score reply similarity, and that score gates Auto mode. LLM judge scores can be noisy. Consider whether a single benchmark run should be sufficient, or whether the threshold (60/100) and sample cap (8) need calibration data.
- The WebSocket listen loop reconnects with exponential backoff (good), but if `configuration.enabled` is toggled off during a sleep, the loop checks `Task.isCancelled` and `configuration.enabled` on wake — confirm that a disable during the sleep window still results in a clean stop.

No blocking code issues found. Awaiting maintainer decision on whether this feature direction fits Omi.

---

undivisible:
Superseded by #10933.

PR #9806, opened 2026-07-15
Tolerate malformed ISO date strings on action-item writes

ZachL111:
## What

`database.action_items._prepare_action_item_for_write` normalizes `created_at` / `updated_at` / `due_at` / `completed_at` from ISO strings to `datetime` objects before writing to Firestore:

```python
if 'due_at' in action_item_data and action_item_data['due_at']:
    if isinstance(action_item_data['due_at'], str):
        action_item_data['due_at'] = datetime.fromisoformat(action_item_data['due_at'].replace('Z', '+00:00'))
```

The `isinstance(..., str)` guards exist because these fields arrive as strings, not only as pydantic-validated `datetime`s: tool- and LLM-created action items and internal callers pass ISO date strings. A single malformed string (for example a bad LLM-produced due date) makes `datetime.fromisoformat` raise `ValueError: Invalid isoformat string`, and this helper is the shared boundary for `create_action_item`, `update_action_item`, and the batch update path, so the whole request 500s.

Reproduction:

```
>>> _prepare_action_item_for_write({'description': 'x', 'due_at': 'not-a-real-date'}, partial=True)
ValueError: Invalid isoformat string: 'not-a-real-date'
```

## Fix

Coerce the four date fields in one tolerant loop. Valid ISO strings parse exactly as before; a malformed string is dropped with a warning instead of raising, so the field falls back to its default or stays unset and the rest of the item is written.

This matches the tolerant date handling the codebase already uses elsewhere:
- the read path `_prepare_action_item_for_read` uses `hasattr(..., 'timestamp')` / `fromtimestamp` and never raises on a bad stored value
- `database.api_key_metadata._coerce_utc_datetime` returns `None` on `ValueError`
- `models.app.ChatTool.deserialize_parameters` logs and drops malformed JSON rather than raising

The loop also collapses four copies of the same parse block into one.

## Tests

`tests/unit/test_action_items_date_coercion.py` (pure function, no DB):

- malformed `due_at` is dropped, not raised
- a valid ISO string is parsed to an aware datetime
- an existing `datetime` value passes through unchanged
- one malformed field does not drop a sibling valid field

The two malformed-input cases raise `ValueError` against the pre-fix source and pass after.

## Verification

```
python -m pytest tests/unit/test_action_items_date_coercion.py -q
  -> 4 passed  (2 fail with ValueError when the source fix is reverted, test kept)

  database/action_items.py tests/unit/test_action_items_date_coercion.py
  -> clean

python -m pyright -p pyrightconfig.json database/action_items.py
  -> 0 errors

  -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9806?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/action_items.py
+++ b/backend/database/action_items.py
@@ -126,28 +126,19 @@ def _prepare_action_item_for_write(action_item_data: Dict[str, Any], *, partial:
         action_item_data.setdefault('provenance', [])
         action_item_data.setdefault('sort_order', 0)
         action_item_data.setdefault('indent_level', 0)
-    # Ensure timestamps are properly formatted
-    if 'created_at' in action_item_data and action_item_data['created_at']:
-        if isinstance(action_item_data['created_at'], str):
-            action_item_data['created_at'] = datetime.fromisoformat(
-                action_item_data['created_at'].replace('Z', '+00:00')
-            )
-
-    if 'updated_at' in action_item_data and action_item_data['updated_at']:
-        if isinstance(action_item_data['updated_at'], str):
-            action_item_data['updated_at'] = datetime.fromisoformat(
-                action_item_data['updated_at'].replace('Z', '+00:00')
-            )
-
-    if 'due_at' in action_item_data and action_item_data['due_at']:
-        if isinstance(action_item_data['due_at'], str):
-            action_item_data['due_at'] = datetime.fromisoformat(action_item_data['due_at'].replace('Z', '+00:00'))
-
-    if 'completed_at' in action_item_data and action_item_data['completed_at']:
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9808, opened 2026-07-15
Tolerate a Hume success response missing job_id

ZachL111:
## What

`utils/other/hume.py` `HumeJobResponseModel.from_dict` runs on the success (HTTP 200) path of `request_user_expression_mersurement`, parsing the response body with a required-key subscript:

```python
model = cls(data["job_id"])   # data is resp.json()
```

Every non-200 status and every request exception is already turned into an error dict earlier in the function. But a 200 response whose JSON body omits `job_id` makes this subscript raise `KeyError`, which propagates uncaught out of the caller. `HumeJobResponseModel.__init__` already types `id` as `Optional[str]`, so a missing id is a representable state, not an error.

## Fix

Read the field defensively with `data.get("job_id")` so a success response missing `job_id` yields a model with `id = None` instead of raising. One line, type-aligned with the existing `Optional[str]` field.

## Tests

`tests/unit/test_hume_job_response_model.py` (pure classmethod, no IO):

- present `job_id` parses to `id`
- missing `job_id` does not raise and leaves `id` None
- extra keys are ignored

The missing-key case raises `KeyError` against the pre-fix source and passes after.

## Verification

```
python -m pytest tests/unit/test_hume_job_response_model.py -q
  -> 3 passed  (1 fail with KeyError when the source fix is reverted, test kept)

  utils/other/hume.py tests/unit/test_hume_job_response_model.py
  -> clean

python -m pyright -p pyrightconfig.json utils/other/hume.py
  -> 0 errors, 0 warnings

  -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9808?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_hume_job_response_model.py
+++ b/backend/tests/unit/test_hume_job_response_model.py
@@ -1,0 +1,25 @@
+"""Regression test: a Hume job-submit success response missing job_id must not raise.
+
+utils.other.hume.HumeJobResponseModel.from_dict is called on the success (HTTP 200) path
+with resp.json(). It read data["job_id"] with a required-key subscript, so a 200 response
+that omits job_id raised KeyError out of request_user_expression_mersurement, even though
+every error status is already turned into an error dict. HumeJobResponseModel.__init__ types
+id as Optional[str], so from_dict now reads job_id defensively and leaves id None when absent.
+"""
+
+from utils.other.hume import HumeJobResponseModel
+
+
+def test_from_dict_with_job_id():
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9810, opened 2026-07-15
Align AudioRingBuffer.extract start to the PCM16 sample boundary

ZachL111:
## What

`utils/audio.py` `AudioRingBuffer` stores PCM16 mono audio (2 bytes per sample). `extract()` computes the byte offset of the requested start as:

```python
start_offset = int((actual_start - buffer_start_ts) * self.bytes_per_second)
```

`actual_start` is an arbitrary float timestamp and `bytes_per_second = sample_rate * 2`, so `start_offset` is odd roughly half the time. The length is already forced even for PCM16 (`length = ((end_offset - start_offset) // 2) * 2`), but the start was never aligned. An odd start offset begins the copy on a sample's high byte, so every int16 sample in the returned buffer is byte-shifted into noise.

This is a real, execution-confirmed data-corruption bug. With 100 ascending samples (1000, 1001, ...) and a start that lands on an odd offset, `extract` returns `[-5629, -5373, -5117, ...]` instead of the written values. The extracted audio feeds speaker-embedding cuts, so the intermittent misalignment corrupts speaker identification. The existing test only checked output length parity, and its offsets happened to be even.

## Fix

Floor `start_offset` to the 2-byte sample boundary before computing length:

```python
start_offset -= start_offset % 2
```

The copy then always begins on a whole sample. The already-even path is unchanged.

## Tests

`tests/unit/test_audio_ring_buffer_alignment.py` (pure class, no monkeypatch):

- extract with a forced odd start offset (3) asserts the decoded int16 samples are the written values (first sample 1001), not byte-shifted noise
- extract with an already-even offset (4) still returns the correct samples (first sample 1002)

The odd-offset case fails against the pre-fix source (returns noise) and passes after.

## Verification

```
python -m pytest tests/unit/test_audio_ring_buffer_alignment.py -q
  -> 2 passed  (1 fail with byte-shifted noise when the source fix is reverted, test kept)

  utils/audio.py tests/unit/test_audio_ring_buffer_alignment.py
  -> clean

python -m pyright -p pyrightconfig.json utils/audio.py
  -> 0 errors, 0 warnings

  -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9810?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_audio_ring_buffer_alignment.py
+++ b/backend/tests/unit/test_audio_ring_buffer_alignment.py
@@ -1,0 +1,60 @@
+"""Regression test: AudioRingBuffer.extract must start on a PCM16 sample boundary.
+
+utils.audio.AudioRingBuffer stores PCM16 mono audio. extract() computed start_offset as
+int((actual_start - buffer_start_ts) * bytes_per_second), which is odd about half the time
+because actual_start is an arbitrary float timestamp. An odd byte offset begins the copy on a
+sample's high byte, so every returned int16 sample is byte-shifted into noise. This audio feeds
+speaker-embedding cuts, so the misalignment intermittently corrupts speaker identification.
+extract() now floors start_offset to the 2-byte sample boundary.
+"""
+
+import struct
+
+from utils.audio import AudioRingBuffer
+
+
+def _decode(pcm: bytes):
+    return list(struct.unpack("<" + "h" * (len(pcm) // 2), pcm))
+
+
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9811, opened 2026-07-15
Chunk memory deletes under the Firestore batch limit

ZachL111:
## What

`database/memories.py` `delete_memories` and `delete_all_memories` accumulated every delete into a single Firestore `WriteBatch` and committed once:

```python
batch = database.batch()
for doc in memories_ref.stream():
    batch.delete(doc.reference)
batch.commit()
```

Firestore rejects a batch with more than 500 writes. A user with more than 500 memories makes `batch.commit()` raise, and the delete removes nothing. `delete_all_memories` is on the account-deletion path, so a heavy user's memories could survive account deletion.

The same file already solved this in `unlock_all_memories`, which chunks at 499 per batch. These two delete helpers were missed.

## Fix

Chunk both delete loops at 499 per batch, committing and starting a fresh batch, exactly mirroring `unlock_all_memories`:

```python
batch = database.batch()
count = 0
for doc in memories_ref.stream():
    batch.delete(doc.reference)
    count += 1
    if count >= 499:  # Firestore batch limit is 500
        batch.commit()
        batch = database.batch()
        count = 0
if count > 0:
    batch.commit()
```

Small deletes still commit once.

## Tests

`tests/unit/test_memories_delete_batch_chunk.py` drives both functions through a fake Firestore client that raises on a commit of more than 500 writes (the real limit), via the module's built-in `firestore_client` injection point:

- deleting 1000 memories splits across multiple batches, each within the limit, and deletes all 1000
- the same for `delete_memories`
- a 3-memory delete still uses a single commit

The two 1000-memory cases raise against the pre-fix single-batch source and pass after.

## Verification

```
python -m pytest tests/unit/test_memories_delete_batch_chunk.py -q
  -> 3 passed  (2 fail with an oversized-commit error when the source fix is reverted, test kept)

  database/memories.py tests/unit/test_memories_delete_batch_chunk.py
  -> clean

python -m pyright -p pyrightconfig.json database/memories.py
  -> 0 errors

  -> 0 violations
```

- INV-MEM-1

This change touches the `database/memories.py` path glob locked by INV-MEM-1 (exactly three product memory tiers). It only chunks the delete batch and does not change tier vocabulary, tier count, or which memories are deleted, so the invariant's guarantee is unchanged and its guard test does not need updating.

### DIFF
--- a/backend/database/memories.py
+++ b/backend/database/memories.py
@@ -405,12 +405,22 @@ def _merge_evidence(  # type: ignore[reportUnusedFunction]  # reserved: thin ali
 
 def delete_memories(uid: str, *, firestore_client: Any = None) -> None:
     database = _get_db(firestore_client)
-    batch = database.batch()
     user_ref = database.collection(users_collection).document(uid)
     memories_ref = user_ref.collection(memories_collection)
+    # Chunk deletes to stay under the Firestore 500-writes-per-batch limit. A user with more than
+    # 500 memories would otherwise make the single batch.commit() raise and delete nothing. Mirrors
+    # the chunking in unlock_all_memories.
+    batch = database.batch()
+    count = 0
     for doc in memories_ref.stream():
         batch.delete(doc.reference)
-    batch.commit()
+        count += 1
+        if count >= 499:  # Firestore batch limit is 500
+            batch.commit()
+            batch = database.batch()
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #9839, opened 2026-07-15
fix(desktop): follow-up polish for Settings controls

skanderkaroui:
Follow-up polish on top of #9813, catching the settings controls that pass missed and refining two of its changes.

## Changes

**Right-align the five pickers #9813 missed** — the polish pass converted most menu pickers to the shared `SettingsMenuPicker` (hug-content, trailing edge aligned with toggles) but left five on the old fixed `.frame(width: 200)`, so their values floated mid-row:
- Voice ("Shimmer") — Settings → Floating Bar
- Voice Model — Settings → AI Chat
- AI Provider — Settings → AI Chat
- System Audio capture mode — Settings → General
- Focus Cooldown — Settings → Advanced → Focus Assistant

**Summary Time glides like the native control** — Settings → Notifications & Privacy → Daily Summary. The `.field`-style `DatePicker` had no steppers, and its inline binding re-derived the date (minutes forced to `:00`) on every change, snapping the field back mid-edit. Now bound to a dedicated UI date state with `.stepperField` (repeat-on-hold arrows, System Settings feel); only the whole hour the backend supports is persisted, and the state re-syncs from the backend on settings load.

**Slim the voice speed value label** — Settings → Floating Bar → Voice Speed. Dropped the 52pt tinted box behind the current-speed number; it now renders as plain 15pt semibold accent text.

## Verification

Built and ran a named bundle (`omi-settings-ui-polish.app`) signed in against the dev backend. Confirmed via macOS accessibility frames that every converted popup's right edge now sits at x=1292 — identical to the toggle column — across Floating Bar, AI Chat, and General, and that the Summary Time row renders a real stepper (`AXIncrementor`). Full-image screenshots weren't available in the headless shell (no Screen Recording grant), so alignment was verified numerically.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9839?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Components/SettingsContentView+BillingHelpers.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Components/SettingsContentView+BillingHelpers.swift
@@ -756,6 +756,8 @@ extension SettingsContentView {
         await MainActor.run {
           dailySummaryEnabled = dailySummary.enabled
           dailySummaryHour = dailySummary.hour
+          dailySummaryTime = SettingsControlMetrics.dailySummaryDate(
+            forHour: dailySummary.hour, referenceDate: Date())
           notificationsEnabled = notifications.enabled
           notificationFrequency = notifications.frequency
           // Mirror to UserDefaults so NotificationService can gate/throttle without a backend roundtrip.
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Components/SettingsContentView+Controls.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Components/SettingsContentView+Controls.swift
@@ -118,13 +118,8 @@ extension SettingsContentView {
           }
           Spacer()
           Text("\(String(format: "%.1f", currentSpeed))×")
-            .font(.system(size: 22, weight: .bold, design: .rounded))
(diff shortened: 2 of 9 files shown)


Git-on-my-level:
Positive signal from maintainer automation: this looks like a tight Settings UI polish follow-up.

What I checked:
- The remaining menu pickers move onto the shared `SettingsMenuPicker`, preserving the existing selection bindings while removing the fixed-width menu frames.
- The Daily Summary time picker now uses a separate UI `Date` state, re-syncs from loaded backend settings, and only persists the backend-supported hour when it actually changes.
- The voice-speed value label cleanup is presentation-only.
- GitHub reports the Desktop Swift build/test and release compile checks as passing on this head.

I did not find a code-level blocker or a security/supply-chain concern. Because the diff touches the System Audio capture setting surface and the automation approval gate for this run disallows formal approval, I’m leaving this as a positive maintainer signal rather than approving. Human maintainer review/sign-off is still needed before merge.

---

kodjima33 (approved):
Fix(desktop): right-align Settings pickers + DatePicker binding + label box (owner-override)

PR #9850, opened 2026-07-15
fix(backend): share strict Firestore transaction fixture

### LINKED ISSUE #9840: FC-6: consolidate strict Firestore transaction fixtures — lenient fakes certify broken code
## Problem

A lenient test fake certifies broken code: #9739 introduced the state-head helper whose read-after-write regression shipped to prod (23 Sentry events, 503s on `POST /v3/memories`) because the ledger tests' transaction fake allowed orderings real Firestore forbids. #9780 (merged as PR #9783) fixed it and added a strict fixture — `_StrictLedgerTransaction` + companion db/document/collection types (`backend/tests/unit/test_memory_ledger.py:132`) — that enforces read-before-write, making the bug class untestable-around. This matters doubly because the watchdog auto-merge gate (#9822 item 5) uses "tests green" as its bar: that gate is exactly as strong as the strictest fake in the path.

Current state of the class, verified:

- **Firestore transactions — the live gap.** Two independent strict fixtures now exist: the ledger one (models `set` + reference `get` only; no `update`/`delete`/queries) and an earlier candidate-lifecycle fake (`backend/tests/unit/test_candidate_lifecycle.py:33`) that also enforces read-before-write and *does* support `update`. Both are private to their test files. Every other transaction test in the suite builds its own ad hoc fake with whatever semantics its author remembered — the #9739 conditions exactly.
- **Rust Redis — already closed, correctly.** `services/redis.rs` is tested against a scripted local RESP peer that drops an `EVAL` response, reconnects, and asserts the non-replayable command executed exactly once, plus generation-replacement and single-flight tests. Server-side recording is the right mechanism (a client-side fake cannot know whether an ambiguous request executed). Nothing to do.
- **Rust SSE — already closed at the right seam.** #9638's regression tests split a UTF-8 codepoint and an SSE frame mid-byte against the real parser. Nothing to do until an end-to-end transport seam exists (tracked in #9836's follow-up).

So the actionable gap is one boundary, and the durable part is the default.

## Fix: one shared strict Firestore transaction fixture + a strictness default, one small PR

1. **Consolidate the two fixtures into one helper** under the existing `tests.unit` helper convention: reconcile the ledger fixture (read-before-write on `set`/reference-`get`) with the candidate-lifecycle fake (`update` support) into a single strict transaction fixture; both test files consume it. Document its honest coverage boundary in its docstring: models `set`/`update`/reference-`get` ordering; does **not** model `delete`, queries, or retry semantics — extend on the next incident, don't speculate.
2. **Strict by default, lenient by named opt-out.** The shared fixture fails loudly on constraint violations; a test needing lenient behavior passes an explicit parameter naming what it relaxes — greppable and reviewable.
3. **One guidance bullet** in `backend/AGENTS.md` → Testing: "A fake for a service boundary must enforce that boundary's ordering/constraint semantics — use the shared strict transaction fixture for Firestore transactions; a lenient fake is how #9739 reached prod. When a boundary has no strict fixture, prefer a server-side recorder (see the Rust RESP peer in `services/redis.rs` tests) over a client-side fake."
4. **Watchdog linkage:** the auto-merge eligibility audit (#9822 item 5) can reference this: transaction-touching backend subsystems are auto-merge-eligible only under the shared strict fixture.
5. **Registry:** FC-6 "lenient service fake", `closed-by` = shared strict fixture + strict-by-default convention; `not-covered` = boundaries without incident evidence and fake styles (`MagicMock`, module stubs) that no mechanical check can classify — recurrence there is caught by the sweep, not a ratchet.

Explicitly **no ratchet** in this issue: "ad hoc fake of a boundary" spans custom classes, `MagicMock`, `fakeredis`, monkeypatched globals, and Rust TCP peers — not mechanically specifiable without unacceptable false positives. If two migrations establish a stable banned construct, a narrow ratchet can follow.

## Proof

- The consolidated fixture's self-test replays #9739's read-after-write sequence and fails loudly.
- Both existing suites (memory ledger, candidate lifecycle) pass on the shared fixture with no assertion changes.
- Any third suite converted opportunistically that starts failing under strictness is triaged: real latent bug (filed) or legitimate leniency (named opt-out).
- FC-6 registry entry with the fields above.

## Cut list

- No Rust changes — Redis and SSE are verified-covered; extraction of the RESP peer waits for a second consumer.
- No bulk migration of every transaction test — the two source suites plus opportunistic conversions.
- No mechanical ratchet (see above).
- No new fakes for boundaries without incident evidence (GCS, Pinecone, etc.).
- No watchdog policy change — #9822 owns that.

## Sizing

One small PR: one shared fixture module (consolidating two existing ones), two test-file migrations, one AGENTS.md bullet, one registry entry.

Git-on-my-level:
## Summary
Closes #9840.

- Consolidate the ledger and candidate-lifecycle Firestore transaction fakes into a shared strict fixture.
- Enforce Firestore's read-before-write rule for transaction-bound document reads after both `set` and `update`.
- Reject transaction/document references from different `StrictFirestore` instances as fixture-integrity policy, preventing accidental mixing of unrelated in-memory stores.
- Make unsupported fake APIs fail with one explicit contract error, so a unit test cannot accidentally treat queries, deletes, bulk reads, or streams as faithfully modelled.
- Record FC-6 and document when to use the fixture versus the Firestore emulator.

The prior ledger fake permitted a transaction-bound document read after a staged write, although Firestore rejects that ordering. The shared fixture defaults to rejecting it, so both source suites now exercise the same production constraint.

The fixture intentionally does **not** model query results, deletes, commit/rollback visibility, atomicity, retries, or contention. It rejects those operations explicitly; incidents in those areas require emulator coverage before a targeted fixture extension. Same-instance enforcement is a fake-integrity policy, not a claim about Firestore client identity.

## Verification

- `cd backend && .venv/bin/pytest -q tests/unit/test_strict_firestore_transaction.py tests/unit/test_memory_ledger.py tests/unit/test_candidate_lifecycle.py` — 89 passed
- `cd backend && .venv/bin/pyright -p pyrightconfig.json --pythonpath .venv/bin/python` — 0 errors (2,236 existing repository warnings)
- `make preflight` — 16 checks passed
- Pre-push checks passed.
- Final independent Sol review approved the implementation and policy wording.

## Full-suite note

`cd backend && ./test.sh` previously encountered an unrelated fast-test timing threshold in `test_activate_task_intelligence_dogfood_user.py::test_missing_control_plans_explicit_read_at_default_generation` (0.13s vs 0.12s). Its isolated rerun passed; this PR does not alter that test or timing configuration.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9850?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/AGENTS.md
+++ b/backend/AGENTS.md
@@ -222,6 +222,8 @@ npm run test:listen-lifecycle:emulator  # Real Firestore transaction contention 
 
 **Test isolation / import purity** — never mutate `sys.modules` at module scope in tests; production modules must not construct clients or do IO at import time. Sanctioned seams: `monkeypatch.setattr` on a lazy-held singleton, FastAPI `app.dependency_overrides`. Enforced by `python scripts/check_module_stub_pollution.py` and `python scripts/scan_import_time_side_effects.py`. Full prescription: `backend/docs/test_isolation.md`.
 
+**Firestore transaction fakes** — a fake at this service boundary must enforce its ordering and constraint semantics. Use `tests.unit.fixtures.strict_firestore_transaction.StrictFirestore` for transaction tests that need document-reference reads plus `set`/`update`: it rejects reads after the first write, the production rule that #9739's lenient fake missed. If an incident requires queries, deletes, retry, or contention behavior, first cover it with the Firestore emulator; extend the fixture only for a proven hermetic guard.
+
 Pre-mock heavy deps before importing the module under test. Use `patch.object(target_module, "func")` not string-based `patch("module.func")` — the string form silently patches the wrong reference if the function was already imported. When modules construct objects at import time, use lazy getters to avoid triggering heavy init in tests.
 
 ### Memory continuity gauntlet gates
--- a/backend/tests/unit/fixtures/strict_firestore_transaction.py
+++ b/backend/tests/unit/fixtures/strict_firestore_transaction.py
@@ -1,0 +1,148 @@
+"""Narrow Firestore transaction fixture for ordering-sensitive unit tests.
+
+This fixture models document-reference ``get(transaction=...)`` plus transaction
+``set`` and ``update``. It enforces Firestore's rule that every transactional
+read must occur before the first transactional write.
+
+It deliberately does not model queries, deletes, commit/rollback visibility,
+or retry and contention semantics. Extend it only when an incident proves that
+one of those boundaries needs a hermetic guard.
+
+As fixture-integrity policy, a transaction accepts references created by its
+own ``StrictFirestore`` instance only. This prevents accidental mixing of
+unrelated in-memory stores; it is not a claim about Firestore client identity.
+"""
+
+from __future__ import annotations
+
+from copy import deepcopy
+from threading import RLock
+from typing import Any
+
+
+class ReadAfterWriteError(RuntimeError):
+    """Raised when a transaction performs a read after staging a write."""
+
+
+class ForeignTransactionError(ValueError):
+    """Raised when fixture policy forbids mixing transaction and reference stores."""
+
+
+class UnsupportedFirestoreOperationError(NotImplementedError):
+    """Raised for a Firestore operation this narrow fixture does not model."""
+
+
+_SUPPORTED_OPERATIONS = 'transaction-bound document get, transaction set, and transaction update'
+
+
+class StrictFirestoreSnapshot:
+    def __init__(self, data: dict[str, Any] | None):
+        self._data = deepcopy(data)
+        self.exists = data is not None
+
+    def to_dict(self) -> dict[str, Any] | None:
+        return deepcopy(self._data)
+
+
+class StrictFirestoreDocument:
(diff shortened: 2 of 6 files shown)


Git-on-my-level:
## Monitor run summary

No actionable review comments or PR-specific check failures found.

**CI status:** Backend unit suite is the only failing check. The failure is `test_apiclient_swift_line_count_ratchet` (APIClient.swift has 6502 lines, max 6500) — a pre-existing upstream `main` breakage that fails identically on `origin/main` (run 29450103680). Not caused by this PR's changes.

**Review threads:** None unresolved.

Git-on-my-level:
@cursor fix failing tests

Git-on-my-level:
@cursor fix failing check

Issue #9857, opened 2026-07-16
Dev pusher CrashLoops because runtime image omits imported services package


Git-on-my-level:
## Incident

The dev pusher Deployment is blocked in `CrashLoopBackOff` and `ProgressDeadlineExceeded`.

The container fails at import time with:

```text
ModuleNotFoundError: No module named 'services'
```

## Root cause

`backend/routers/pusher.py` imports:

```python
from services.conversation_finalization import final_attempt_failed
```

but `backend/pusher/Dockerfile` copies `backend/routers/` and `backend/pusher/` into the runtime image without copying `backend/services/`.

The pusher image therefore cannot import its Uvicorn/router path after this import was introduced by #9704.

This is distinct from #9854, which reconciles non-secret pusher runtime ConfigMap values before preflight; it does not package `services` or deploy a corrected image.

## Scope

- Package the `services` runtime module required by pusher.
- Add an image-level startup/import contract for pusher.
- Roll out and verify the fix in dev.

Do not broaden this into deployment credential, Secret, or unrelated runtime-config changes.

## Acceptance criteria

- [ ] The pusher runtime image includes `services.conversation_finalization`.
- [ ] A built-image smoke test imports the real pusher router/entrypoint (for example, `python -c 'import routers.pusher'`).
- [ ] The test fails if a pusher import references a package omitted from the runtime image.
- [ ] A dev rollout of the corrected image reaches `Available=True` without `CrashLoopBackOff` or `ProgressDeadlineExceeded`.
- [ ] The prior revision can retire after the new ReplicaSet is healthy.

## Related work

- #9704 introduced the pusher import path that exposed this packaging gap.
- #9140 / #9141 were an earlier pusher CrashLoop caused by a different omitted runtime dependency (`jsonschema`). The image-level contract should prevent both classes of failure.
- #9854 addresses ConfigMap preflight/reconciliation only.

Issue #9140, opened 2026-07-06
Dev GKE pusher crashloop: missing jsonschema dependency in pusher image


Git-on-my-level:
## Bug Description
Dev GKE pusher is currently unavailable after the latest main deploy: `dev-omi-pusher` is in `CrashLoopBackOff` because the container imports `utils.llm.gateway_client`, which imports `jsonschema`, but the pusher runtime image does not have the `jsonschema` package installed.

## Evidence
- Environment: `based-hardware-dev`, GKE cluster `dev-omi-gke`, namespace `dev-omi-backend`
- Checked after macOS Desktop beta release `v0.12.31+12031-macos` was published (`2026-07-06T06:37:49Z`)
- Current pod: `dev-omi-pusher-fbc855dc-m44vm`
- Current image: `gcr.io/based-hardware-dev/pusher:6ead6b5`
- Kubernetes state: `dev-omi-pusher` deployment `0/1` available; pod restart count observed at 8 and increasing
- Cloud Logging since `2026-07-06T06:38:00Z`: 198 ERROR entries from the pusher container

Traceback root:

```text
File "/app/utils/llm/gateway_client.py", line 10, in <module>
    from jsonschema import ValidationError as JsonSchemaValidationError
ModuleNotFoundError: No module named 'jsonschema'
```

## Why this matters
This is not just a health-check blip: pusher cannot import and start, so dev realtime/pusher paths are down in GKE.

## Related / not duplicate
- #9101 covers dev Cloud Run deploy drift (`apps-js` unhealthy and Cloud Run `pusher` latest revision referencing a missing image). This issue is the separate GKE pusher runtime dependency regression for image `gcr.io/based-hardware-dev/pusher:6ead6b5`.

## Acceptance Criteria
- [ ] Add `jsonschema` to the dependency set used by the pusher image, or remove the import path from pusher startup.
- [ ] Rebuild/redeploy dev pusher.
- [ ] Verify `kubectl get deploy -n dev-omi-backend dev-omi-pusher` shows `1/1` available.
- [ ] Verify startup logs no longer contain `ModuleNotFoundError: No module named 'jsonschema'`.

PR #9858, opened 2026-07-16
fix(backend): copy services/ into pusher runtime image (#9857)

kodjima33:
## Bug
The dev **pusher** Deployment is stuck in `CrashLoopBackOff` / `ProgressDeadlineExceeded`. The container fails at import time:

```
ModuleNotFoundError: No module named 'services'
```

## Root cause
`backend/routers/pusher.py` imports `from services.conversation_finalization import final_attempt_failed` (added by #9704), but `backend/pusher/Dockerfile` whitelist-copies only `config/ database/ models/ routers/ utils/ pusher/` — it never copies `backend/services/`. The trimmed runtime image therefore cannot import the pusher router path.

## Fix
- Add `COPY backend/services/ ./services/` to `backend/pusher/Dockerfile`. `services.conversation_finalization` is a namespace module (no `__init__.py`) whose only first-party imports are `database` and `utils`, both already copied — so this single line is sufficient.
- Add a hermetic regression test (`tests/unit/test_pusher_dockerfile_import_contract.py`) that reconciles the Dockerfile copy whitelist against the first-party packages pusher's entrypoints (`pusher.main`, `routers.pusher`) statically reach. An in-process `import routers.pusher` cannot catch this — the full `backend/` tree is present under pytest — so the guard is a labeled static import contract. It catches this class (#9140/#9141 `jsonschema`, #9704 `services`).

## Verification
- `python3 -m pytest tests/unit/test_pusher_dockerfile_import_contract.py` → **1 passed** with the fix.
- Temporarily removing the new `COPY services/` line → the test **fails**, naming exactly `['services']`, confirming it would have caught this bug.
- `typesense` (a `backend/typesense/` schema dir vs the third-party `typesense` pip client) is correctly excluded — first-party detection requires the dir to contain `.py` sources.
- **UNVERIFIED:** actual Docker image build + dev rollout to `Available=True` not executed in this run (no Docker/liblc3 toolchain available here). Packaging correctness is verified against pusher's real transitive import graph.

Failure-Class: FC-runtime-image-import-gap

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9858?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-runtime-image-import-gap.json
+++ b/.github/failure-classes/FC-runtime-image-import-gap.json
@@ -1,0 +1,9 @@
+{
+  "schema_version": 1,
+  "id": "FC-runtime-image-import-gap",
+  "violated_contract": "A service's runtime container image must copy every first-party package its entrypoints import, so the process never fails at import time with ModuleNotFoundError.",
+  "canonical_prevention": "Reconcile each whitelist-copy Dockerfile against the first-party packages its entrypoints statically reach with a hermetic import-contract test.",
+  "evidence_prs": [9140, 9141, 9704],
+  "scope_hints": ["backend/pusher/**", "backend/**/Dockerfile"],
+  "status": "open"
+}
--- a/backend/pusher/Dockerfile
+++ b/backend/pusher/Dockerfile
@@ -68,6 +68,7 @@ COPY backend/config/ ./config/
 COPY backend/database/ ./database/
 COPY backend/models/ ./models/
 COPY backend/routers/ ./routers/
+COPY backend/services/ ./services/
 COPY backend/utils/ ./utils/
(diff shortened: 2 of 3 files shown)


kodjima33:
Core fix (COPY backend/services/ into the pusher runtime image) already landed on main via 05b7ffb37c 'fix(backend): enforce runtime image contracts'. Closing as superseded.

PR #9859, opened 2026-07-16
fix(desktop): stabilize PTT agent routing

Git-on-my-level:
## Summary

- Keep background-agent spawn acknowledgements inside the active OpenAI Realtime PTT voice lane.
- Treat an external model's local-provider enum as a proposal, not user authority: generic PTT delegation now defaults to Omi's managed agent.
- Preserve an exact current-turn Hermes or OpenClaw request, correct a mismatched model proposal, and accept the spoken `OpenCloud` alias for OpenClaw.

An accepted spawn receipt previously claimed a deterministic local-TTS lease, stopped the native realtime stream, and treated the provider continuation as suppressed. The receipt now remains persistence-only; the existing post-tool continuation fence keeps the native provider responsible for the audible reply.

Separately, the realtime schema exposes optional local providers and the model emitted `provider: "hermes"` for a generic request ("Have an agent sleep for five seconds"). That enum was treated as a provider choice, which admitted a local Hermes runtime even though the user never selected it. The shared external-surface policy now owns that boundary before child-session admission: only a directed provider selection in the current user utterance may reach the control plane. Tests cover generic delegation, model/provider mismatch correction, explicit Hermes, and unavailable explicit OpenClaw.

## Invariants

- INV-AGENT-*
- INV-CHAT-1
- INV-VOICE-1

## Failure class

## Verification

- `cd desktop/macos/agent && npx vitest run tests/external-surface-authority.test.ts tests/omi-tool-manifest.test.ts` — 32 passed
- `cd desktop/macos && xcrun swift test --package-path Desktop --filter HubSystemInstructionTests` — 17 passed
- `cd desktop/macos && ./scripts/agent-logic-harness.sh --cross-surface-smoke` — Swift 76, Node 247, Rust 13 passed
- `OMI_PR_BODY_FILE=/tmp/omi-ptt-spawn-native-voice-pr.md make preflight` — passed

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline.json
@@ -27,7 +27,7 @@
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2769,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4900,
     "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2408,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift": 5153,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift": 5128,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1572,
     "desktop/macos/Desktop/Sources/FloatingControlBar/VoiceTurnStateMachine.swift": 1790,
     "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": 1915,
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift
@@ -2976,28 +2976,9 @@ final class RealtimeHubController: NSObject, RealtimeHubSessionDelegate {
     return player
   }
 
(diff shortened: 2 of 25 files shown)


Git-on-my-level:
## Monitor run — invariant doc alignment

**Thread addressed:** Codex P1 `discussion_r3592979403` (thread `PRRT_kwDOLkKqys6RVd7V`)

**Analysis:** The P1 correctly identified that `RealtimeAcceptedSpawnPresentationPolicy` removal contradicted the spawn-receipt invariant in `docs/product/invariants/agent-control-plane.md`. However, the PR's behavior change is intentional and is the PR's core purpose: spawn receipts become persistence-only and the native provider continuation is the sole audible response (replacing the previous deterministic local-TTS acknowledgement). The PR body, updated tests (`testCanonicalSpawnReceiptKeepsProviderContinuationForNativeVoiceOutput`), and changelog entries all reflect this deliberate change. The bot's prescribed fix (restore the old suppression policy) would revert the PR's intended product change.

**Fix applied:** Updated the invariant bullet at `agent-control-plane.md:84-92` to match the new behavior — the receipt is persistence-only, does not claim a local TTS lease, and the turn advances from the provider continuation rather than from the receipt. This is a docs-only change; no production code was modified.

**Verification:**
- `make preflight` — 20/20 checks passed (including product-invariants, failure-class-protocol, desktop-changelog)
- Desktop Swift Build & Tests — pass (15m2s)
- Agent Runtime — pass
- Hygiene, contracts, desktop-core-e2e-t0 — pass
- All remaining checks pass or skip (no failures)

kodjima33 (approved):
Fix(desktop): stabilize PTT agent routing — spawn receipt no longer hijacks voice lane, strip model-claimed provider (owner-override)

PR #9860, opened 2026-07-16
fix(backend): enforce runtime image contracts

### LINKED ISSUE #9857: Dev pusher CrashLoops because runtime image omits imported services package

Git-on-my-level:
## What changed and why

Fixes #9857. Runtime images now have one registry that defines their Dockerfiles, build roots, entrypoints, deployment workflows, and smoke policy. A static source-closure check prevents final Docker stages from omitting reachable first-party modules; built-image smoke checks verify reachable dependencies and safe entrypoint imports offline before an image is published.

none

## How it was verified

- `make runtime-image-source-closure` — passed for all 10 registered runtime images.
- Built the pusher image for `linux/amd64` from its production Dockerfile using the public `python:3.11-slim` base, then ran its offline dependency and router-import smoke — passed. The exact private GCR base requires CI credentials locally; PR CI authenticates to GCR before its equivalent image matrix.
- `actionlint -shellcheck ''` for every changed workflow — passed.
- `scripts/pr-preflight --pr-body-file /tmp/runtime-image-contracts-pr-body.md` — passed.
- Exercised `scripts/pre-push` against this ref: its PR preflight, backend checks, and runtime-image source-closure check passed. Its unrelated desktop Swift build could not complete because the machine reached 117 MiB free space; after reclaiming its generated output, the local Docker daemon still could not start, so the final amended container smoke is delegated to the authenticated PR matrix.

## Tests

- `backend/test.sh` — the new source-closure and import-side-effect regression tests passed. The full suite hit two pre-existing parallel-only flakes in `tests/unit/test_sync_v2.py` and `tests/unit/test_ws_c_backfill.py`; each passed on its required isolated rerun.
- `backend/scripts/typecheck.sh backend/scripts/runtime_image_contracts.py backend/modal/job.py backend/tests/unit/test_runtime_image_contracts.py backend/tests/unit/test_notifications_job_import.py` — 0 errors (2236 existing warnings).

## Failure class (fixes)

Failure-Class: new

The violated contract is that a deployable runtime image must include every reachable first-party module and installed third-party module for its declared entrypoint. `FC-runtime-image-boundary` records the canonical prevention: a single image registry, final-stage source-closure validation, and offline built-image smoke tests before publication and in PR CI. The predecessor instance is merged PR #9141; this PR adds the enforceable guard surface that would have caught both it and #9857.

PR #9141 and issue #9857 showed that Docker build contexts and final-stage COPY boundaries can silently diverge from runtime imports. This guard is intentionally runtime-image-specific rather than a shared primitive because it owns Dockerfile final stages, entrypoints, and deployment publication order.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9860?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Monitor follow-up for this run:

- Fixed runtime image smoke probes to use each image's registered Python executable; `nllb-translation` now probes with `python3` instead of assuming `python` exists.
- Added regression coverage for registered smoke Python executables.
- Made runtime-image smoke CI skip cleanly when `GCP_CREDENTIALS` is unavailable, while preserving source-closure coverage.
- Resolved the addressed Codex review thread.

Verification passed:
- `backend/.venv/bin/python -m pytest tests/unit/test_runtime_image_contracts.py -q`
- `actionlint -shellcheck '' .github/workflows/runtime_image_contracts.yml`
- `scripts/pr-preflight --pr-body-file /tmp/omi-pr-9860-body.md`
- GitHub checks are green/pass or skipped for this PR head, including all runtime image smoke jobs.

PR #9861, opened 2026-07-16
perf(desktop): make local feedback loops fast

Git-on-my-level:
## Summary

- Add `scripts/dev-feedback.py` for explicit one-shot and watch-mode Swift/XCTest and Rust/Cargo feedback loops, including automatic reruns after failures.
- Keep generated agent tool surfaces stable when content is unchanged, run independent Swift suites with four workers by default, and make named development app relaunches reuse a safe local debug Rust backend.
- Add safe local-profile refresh, bounded backend-health checks, remote-harness `--no-wait` support, and user documentation for the fast path.
- Fix OpenClaw discovery so a user-managed FNM/NVM installation wins over a machine-wide Homebrew binary.

## Benchmark: typical developer feedback

Measurements were captured on the same local Apple Silicon development machine. The focused-test rows deliberately compare a cold initial target bootstrap with the steady-state feedback loop developers use while iterating; they do not claim that an edited Swift file recompiles in 0.79 seconds.

| Scenario | Baseline | Final | Improvement |
| --- | ---: | ---: | ---: |
| Swift selected regression: cold clean target to steady rerun | 147.60s | 0.79s | 187x |
| Rust selected regression: clean target to steady rerun | 25.91s | 0.11s | 236x |
| Named desktop app: full launch to reusable relaunch | ~55s | 7.4s | 7.4x |

Additional final captures: the first Swift rebuild after a source edit took 18.75s; a complete named app bundle launch took 46.8s; the four-worker desktop component suite completed in 2m24s. The reusable launch restarts the named development app, rather than preserving live UI state, so this is intentionally not described as state-preserving hot reload.

Before this change, the generator rewrote five otherwise identical outputs on every launch; that made a warm no-change app build take 19.33s instead of its 6.71s underlying build time. The new content-aware generator preserves output mtimes when current, allowing the documented fast path to stay reusable.

## Product invariants

- `INV-CHAT-1` (path-matched): OpenClaw executable discovery changed only process resolution precedence; transcript ownership and continuity behavior are unchanged. The focused regression test covers the selected managed installation.

## Failure class

Failure-Class: none

The OpenClaw discovery defect is a one-off executable-resolution precedence failure: a machine-wide binary could shadow the version explicitly managed in the user's home directory. It does not fit the existing state ownership, timeout, malformed-read, duplicate-merge, or cache classes.

## Verification

- `bash test.sh` from `desktop/macos` — passed (Rust 365/365; Swift suites with four workers; 2m24s)
- `bash tests/test-fast-dev-bundle.sh && bash tests/test-local-profile-fast-bundle.sh && bash tests/test-local-profile-keychain-reset.sh && bash tests/test-rust-backend-dev.sh && bash tests/test-swift-test-suites.sh && bash tests/test-yolo-dev-backend.sh && bash tests/test-dev-feedback.sh` — passed
- `black --check --line-length 120 --skip-string-normalization scripts/dev-feedback.py tests/test-dev-feedback.py` — passed
- `npm test -- tests/tool-surfaces-exhaustiveness.test.ts` and `node --experimental-strip-types scripts/generate-tool-surfaces.mjs --check` from `desktop/macos/agent` — passed
- `cargo test --locked` from `desktop/macos/Backend-Rust` — passed (365 tests)
- Exercised a fresh named `omi-fast-loop-final` development app bundle and its automation health endpoint, then its reusable `--fast-only` relaunch; both reported healthy. The named test bundle was removed afterward.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
## Monitor fix: honor backend profile in reuse guard

**Codex P2** (`run.sh` reuse path): The backend reuse condition checked source/config freshness, port listening, and health, but never compared the requested `RUST_BACKEND_PROFILE`/`RUST_BACKEND_BINARY` against the metadata recorded for the running PID. Toggling `OMI_DESKTOP_BACKEND_RELEASE` between debug and release could silently reuse the wrong backend.

**Fix:** Added `omi_rust_backend_metadata_matches` to the reuse guard (line 774). A profile mismatch now falls through to the rebuild/replace path instead of silently reusing the wrong binary. The existing `omi_rust_backend_metadata_matches` helper already validates `profile`, `binary`, and `port` against the metadata file.

**Verification:**
- `bash tests/test-rust-backend-dev.sh` — all lifecycle tests pass (including the existing assertion that debug metadata does not match a release profile request)
- `bash -n run.sh` — syntax OK
- All 18 manifest preflight checks pass
- CI: all checks green (Desktop Swift Build & Tests, Agent Runtime, Hygiene, Formatting, Detect Changes)

PR #9863, opened 2026-07-16
fix conversation cards stuck on Processing forever

### LINKED ISSUE #9802: Bug: conversation card gets stuck on "Processing" and never completes
## Summary
A conversation card stays stuck on `Processing` and never finishes.

## Platform
- Android
- Same app version as the previous issues reported

## Problem
In the conversations list, a card appears with the `Processing` state and just stays there.

It looks like the item never completes processing.

## Expected behavior
A conversation should finish processing and appear normally in the list.

## Actual behavior
The conversation card remains stuck on `Processing` indefinitely.

## Impact
The conversation appears blocked and does not complete normally.

## Evidence
A screenshot shows the top card in the conversations list stuck in the `Processing` state.

## Notes
- This is happening on Android
- Same app version as previously reported
- The stuck item appears at the top of the conversations screen

krushnarout:
## Problem

A conversation card gets stuck on **Processing** in the conversations list and never completes (reported on Android). Two independent defects produce the same symptom:

1. **Backend — conversation stranded on `status=processing` forever.** `POST /v1/conversations` and `POST /v1/conversations/{id}/finalize` admit the conversation to `processing` and then run `process_conversation` inside the request itself. Unlike the listen pipeline, this synchronous path has no durable finalization job for the reconciler to replay — so any processor exception (LLM outage, transient dependency failure) left the document on `processing` permanently. Every client rendered a stuck Processing card that no refresh could resolve, and the client could not retry because finalize no-ops on non-`in_progress` conversations.
2. **App — stale Processing card survives refreshes.** The background refresh path (`_fetchNewConversations`) inserted new processing cards but never removed entries the server had since completed. When the websocket `ConversationEvent` that normally clears the card is missed (Android kills the socket in background), the card stayed pinned at the top for the whole session even though the conversation had completed.

## Fix

- **Backend:** add `lifecycle_service.rollback_processing_admission` — a compare-and-swap returning `processing` → `in_progress` — and wrap both endpoints' inline processing in `lifecycle_service.processing_admission_guard`, a lifecycle-owned context manager that rolls the admission back on failure and re-raises. The CAS only rolls back a generation that is still processing, so a concurrent completion, discard, or newer generation always wins, and a rolled-back conversation is retryable. A rollback error (e.g. conversation deleted mid-processing) is logged and cannot mask the original processing exception.
- **App:** the refresh path now reconciles `processingConversations` against the fetched list and drops cards for conversations the server no longer reports as processing. Local-only placeholder cards (id `'0'`) and failed refreshes are left untouched.
- **Cleanup (separate commit):** removed the dead `_processingConversationWatchTimer` field (declared and cancelled, never started).

The web app renders `status` directly from each server fetch (no sticky local list), so it self-heals with the backend fix alone.

## Verification

- `backend/tests/routers/test_conversations_processing_rollback.py` (new) — both endpoints roll back the admission when the processor raises; a rollback error does not mask the original exception. Fails without the router fix.
- `backend/tests/unit/test_conversation_lifecycle_contract.py` — new contract tests: rollback returns a failed synchronous finalization to `in_progress` and stays retryable; never reopens a terminal generation; respects discard.
- `app/test/providers/conversation_provider_processing_reconcile_test.dart` (new) — refresh clears a completed conversation's card, keeps still-processing cards, keeps local placeholders, and leaves cards untouched on a failed refresh.
- Full suites: `backend/test.sh` (625 files) and `app/test.sh` (823 tests) pass locally; `app/scripts/analyze_ratchet.sh` passes; `make preflight` passes (19 checks).

## Failure class

Fix commits use this repo's no-colon subject convention, so no `Failure-Class:` declaration is required (`scripts/pr-preflight --suggest`: "No `fix:` commits were detected"). Product invariants affected: none.

Closes #9802


kodjima33 (approved):
Fix (closes #9802): stop conversation cards stuck on Processing; backend CAS rollback + app reconcile, tests, 4/5

PR #9867, opened 2026-07-16
Default Task.executed_at to None instead of an evaluated-once timestamp

ZachL111:
## What

`models/task.py` declared:

```python
executed_at: Optional[datetime] = datetime.now()
```

A bare `datetime.now()` default is evaluated once, when the class is defined, not per instance. So every `Task` built without an explicit `executed_at` shared the same process-start timestamp and looked already-executed at a stale time, instead of being unset. The sibling `updated_at` field on the next line already defaults to `None`.

## Fix

Default `executed_at` to `None` (not yet executed), matching `updated_at`. Per-instance timestamps are set explicitly by the code that executes the task.

## Tests

`tests/unit/test_task_executed_at_default.py`:

- a `Task` built without `executed_at` has `executed_at is None`
- an explicit `executed_at` is preserved

The default-None case fails against the pre-fix source (it was a datetime) and passes after.

## Verification

```
python -m pytest tests/unit/test_task_executed_at_default.py -q

black --line-length 120 --skip-string-normalization --check models/task.py tests/unit/test_task_executed_at_default.py
  -> clean
python -m pyright -p pyrightconfig.json models/task.py   -> 0 errors, 0 warnings
python scripts/check_module_stub_pollution.py            -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9867?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/models/task.py
+++ b/backend/models/task.py
@@ -24,7 +24,10 @@ class Task(BaseModel):
     action: TaskAction
     status: TaskStatus
     created_at: datetime
-    executed_at: Optional[datetime] = datetime.now()
+    # Default to None (not yet executed). A bare datetime.now() default is evaluated once at class
+    # definition, so every Task built without an explicit executed_at would share the process-start
+    # timestamp and look already-executed. updated_at below already uses the correct None default.
+    executed_at: Optional[datetime] = None
     updated_at: Optional[datetime] = None
     request_id: Optional[str] = None
     memory_id: Optional[str] = None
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Fix: Task.executed_at default None (was evaluated-once now()); test, 5/5

PR #9871, opened 2026-07-16
Preserve message-text spacing in the XML chat history sent to the LLM

ZachL111:
## What

`models/chat.py` `Message.get_messages_as_xml` builds each message block from a flush-left f-string template and then ran:

```python
formatted_messages.append(msg.replace('    ', '').strip())
```

The template lines are flush-left (no 4-space indentation), so the `replace('    ', '')` never targets the template. It only deletes 4-space runs from the interpolated `message.text` - i.e. from user content like code, tables, and aligned or pasted text - silently corrupting the chat history that is fed to the LLM.

Example: `"def foo():        return 1"` became `"def foo():return 1"`.

## Fix

Drop the `replace` and only strip the block's surrounding whitespace:

```python
formatted_messages.append(msg.strip())
```

## Tests

`tests/unit/test_chat_xml_preserve_spacing.py` (pure staticmethod, no mocks):

- a 4-space run inside `message.text` survives in the produced XML
- the message content is present in the XML

The spacing case fails against the pre-fix source (the run was deleted) and passes after.

## Verification

```
python -m pytest tests/unit/test_chat_xml_preserve_spacing.py -q

black --line-length 120 --skip-string-normalization --check models/chat.py tests/unit/test_chat_xml_preserve_spacing.py
  -> clean
python -m pyright -p pyrightconfig.json models/chat.py   -> 0 errors
python scripts/check_module_stub_pollution.py            -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9871?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/models/chat.py
+++ b/backend/models/chat.py
@@ -200,7 +200,10 @@ class Message(BaseModel):
 {file_section}
 </message>"""
 
-            formatted_messages.append(msg.replace('    ', '').strip())
+            # Only strip the block's surrounding whitespace. The template above is flush-left, so a
+            # .replace('    ', '') here would instead delete 4-space runs from message.text (code,
+            # tables, aligned or pasted text), corrupting the history shown to the LLM.
+            formatted_messages.append(msg.strip())
 
         return '\n'.join(formatted_messages)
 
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Fix: preserve message spacing in XML chat history; test, 5/5

PR #9873, opened 2026-07-16
Guard delete_app_logo against a URL from another storage bucket

ZachL111:
## What

`utils/other/storage.py` `delete_app_logo` computed the object path with:

```python
path = img_url.split(f'https://storage.googleapis.com/{omi_apps_bucket}/')[1]
```

A googleapis URL for a different or legacy bucket makes `split` return a one-element list, so `[1]` raises `IndexError`. Callers guard only with the looser `startswith('https://storage.googleapis.com/')`, so such a URL reaches this function and 500s it.

## Fix

Return early when the app-logo bucket prefix is absent (there is nothing to delete in this bucket), and use `split(prefix, 1)` for the matching case:

```python
prefix = f'https://storage.googleapis.com/{omi_apps_bucket}/'
if prefix not in img_url:
    logger.warning(f'delete_app_logo: url not in {omi_apps_bucket}, skipping')
    return
bucket = _get_storage_client().bucket(omi_apps_bucket)
path = img_url.split(prefix, 1)[1]
```

## Tests

`tests/unit/test_delete_app_logo_bucket_guard.py` drives `delete_app_logo` with a fake storage client (`monkeypatch.setattr` on `_get_storage_client`):

- a URL from another bucket returns without deleting (previously `IndexError`)
- a matching URL deletes the right object

The other-bucket case raises `IndexError` against the pre-fix source and passes after.

## Verification

```
python -m pytest tests/unit/test_delete_app_logo_bucket_guard.py -q
  -> 2 passed  (1 fail with IndexError when the source fix is reverted, test kept)

black --line-length 120 --skip-string-normalization --check utils/other/storage.py tests/unit/test_delete_app_logo_bucket_guard.py
  -> clean
python -m pyright -p pyrightconfig.json utils/other/storage.py   -> 0 errors
python scripts/check_module_stub_pollution.py                    -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9873?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline.json
@@ -11,7 +11,7 @@
     "backend/utils/apps.py": 1641,
     "backend/utils/memory/legacy_backfill.py": 1816,
     "backend/utils/memory_ingestion/pipeline.py": 1711,
-    "backend/utils/other/storage.py": 1578,
+    "backend/utils/other/storage.py": 1584,
     "backend/utils/retrieval/tools/calendar_tools.py": 1513,
     "backend/utils/sync/pipeline.py": 2296,
     "desktop/macos/Backend-Rust/src/routes/proxy.rs": 2741,
@@ -46,6 +46,7 @@
     "desktop/macos/Desktop/Sources/VoiceTurnDomain/VoiceTurnStateMachine.swift": 2155
   },
   "raise_justifications": {
+    "backend/utils/other/storage.py": "delete_app_logo now requires an exact app-logo bucket prefix (startswith) so a foreign URL embedding the prefix later cannot delete an unrelated object on this deletion path (David review on #9873).",
     "desktop/macos/Desktop/Sources/AuthService.swift": "Pinned swift-format normalizes line layout without changing this source's runtime behavior.",
     "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": "Strict concurrency annotations (Ticket 14) add Sendable conformances and isolation qualifiers without changing runtime behavior.",
     "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": "Strict concurrency annotations (Ticket 14) add Sendable conformances and isolation qualifiers without changing runtime behavior.",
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
Backend bug fix (guard delete_app_logo against foreign bucket URL) — approve; CI has failing check, hold merge

Git-on-my-level (changes requested):
Thanks for tightening this path — returning early for non-app-logo bucket URLs is the right direction, and the regression test for the ordinary other-bucket case is useful.

I think this still needs one small fix before merge because this is a security/user-data deletion path: the guard should require the URL to start with the app-logo bucket prefix, not merely contain it. With the current `if prefix not in img_url` check, a URL from another bucket can still contain the app-logo prefix later in the path or query string, and `split(prefix, 1)[1]` would then derive a target object name and delete from the app-logo bucket even though the URL itself was not an app-logo object.

Please switch the guard to an exact prefix check, e.g. `if not img_url.startswith(prefix): return`, and add a regression case for a foreign-bucket URL that embeds the app-logo prefix later in the URL/query. Also please rerun/fix the currently failing Backend unit suite/Hygiene checks.

---

ZachL111:
Updated per the review. The guard now uses an exact img_url.startswith(prefix) check (and derives the object name via img_url[len(prefix):]), so a foreign-bucket URL that embeds the app-logo prefix later in its path or query can no longer resolve to a delete target. Added a regression case for that embedded-prefix URL. I also raised the product-file-line-count-ratchet baseline for storage.py by 6 lines with a one-line justification. The failing Backend unit check was tests/unit/test_trial_metadata.py::test_one_second_before_trial_length, a time-boundary test unrelated to this path; this push re-runs CI.

kodjima33 (approved):
backend bug fix: guard delete_app_logo vs other-bucket URL (IndexError->500), has test, not on main

kodjima33 (approved):
backend bug fix: guard delete_app_logo vs other-bucket URL (IndexError->500); has test; not on main

PR #9875, opened 2026-07-16
fix(app): conversation audio player shows captured audio length, not wall-clock span

mdmohsin7:
## Problem

The audio player on the transcript page showed the **wrong duration** — longer than the audio you can actually hear.

## Root cause

When a conversation is served as one dense MP3 (gaps collapsed) + a spans manifest, the backend stamps two numbers:
- `duration` = **wall-clock** = `started_at` → last part's end (includes lead-in silence + collapsed inter-part gaps)
- `captured_duration` = **actual audio** = sum of the recorded parts

Both `ConversationBottomBar` (transcript page) and the standalone `ConversationAudioPlayerWidget` used `duration` (wall) as the scrubber total, so the total ran past the audio.

**Verified on a real conversation (2 parts):** player showed **23:51** (wall = 1431s) for **19:12** of audio (`captured_duration` = 1152s). The extra ~4:39 = a 199s lead-in before the first part + an 80s gap between parts — both collapsed out of the MP3 but still counted in the wall total.

## Fix

Switch both players to a **dense-artifact timeline**:
- total = `captured_duration` (the MP3's real length),
- position = raw artifact time (advances 1:1 with playback),
- scrubber seeks map directly to artifact time,
- **transcript-segment taps still map wall→artifact** via `AudioTimelineMapper.wallToArtifact`, so tapping a segment lands on the exact words,
- removed the gap-shading overlay (a dense timeline has no gaps).

`AudioTimelineMapper` is unchanged — the fix is purely which duration/position the widgets display.

## Testing

- `flutter analyze` clean on both widgets.
- `test/unit/audio_timeline_mapper_test.dart` + `audio_urls_response_test.dart` — 16/16 pass. These already assert `capturedDuration` (audio length) vs `wallDuration` and the `wallToArtifact` mapping the segment-seek relies on.
- Verified against real prod data: total now resolves to `captured_duration` (19:12) instead of wall (23:51).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9875?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

### DIFF
--- a/app/lib/widgets/conversation_audio_player_widget.dart
+++ b/app/lib/widgets/conversation_audio_player_widget.dart
@@ -396,7 +396,7 @@ class _ConversationAudioPlayerWidgetState extends State<ConversationAudioPlayerW
                                 thumbShape: const RoundSliderThumbShape(enabledThumbRadius: 6),
                                 overlayShape: const RoundSliderOverlayShape(overlayRadius: 14),
                                 // Dense MP3 has no gaps on its timeline — plain track.
-                                trackShape: _GapAwareTrackShape(const []),
+                                trackShape: const _GapAwareTrackShape([]),
                               ),
                               child: Slider(
                                 value: combinedPosition.inMilliseconds.toDouble().clamp(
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Fix(app): audio player shows captured audio length not wall-clock; 4/5

PR #9876, opened 2026-07-16
Keep goal-advice chat context when a stored message has null text

ZachL111:
## What

`utils/llm/goals.py` `_get_goal_context` builds the recent-chat block of the goal-advice prompt:

```python
text = msg.get('text', '')[:200]
```

`get`'s default only applies when the key is ABSENT. A stored message with `text=None` makes `None[:200]` raise, and the broad `except` around the block then swallows it and leaves `chat_context` empty - dropping the entire recent-chat block from the goal-advice prompt for that user. The memory block just below already guards with `if m.get('content')`, so null field values are expected here.

## Fix

Coerce a null `text` to `''` so it is skipped (the following `if text:` already filters empties), keeping the other messages:

```python
text = (msg.get('text') or '')[:200]
```

## Tests

`tests/unit/test_goal_context_null_message_text.py` drives `_get_goal_context` with the four retrieval blocks mocked (`vector_search`, `conversations_db.get_conversations`, `memories_db.get_memories`, `chat_db.get_messages`):

- a null-text message alongside a valid one keeps the valid message in `chat_context`
- all-valid messages are rendered (`User:` / `Omi:`)

The null case fails against the pre-fix source (empty `chat_context` from the swallowed crash) and passes after.

## Verification

```
python -m pytest tests/unit/test_goal_context_null_message_text.py -q

black --line-length 120 --skip-string-normalization --check utils/llm/goals.py tests/unit/test_goal_context_null_message_text.py
  -> clean
python -m pyright -p pyrightconfig.json utils/llm/goals.py   -> 0 errors
python scripts/check_module_stub_pollution.py                -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_goal_context_null_message_text.py
+++ b/backend/tests/unit/test_goal_context_null_message_text.py
@@ -1,0 +1,40 @@
+"""Regression test: a null message text must not drop the whole chat block from goal context.
+
+utils.llm.goals._get_goal_context builds goal-advice context from recent chat messages with
+text = msg.get('text', '')[:200]. get's default only applies to an ABSENT key, so a stored
+message with text=None makes None[:200] raise; the broad except then leaves chat_context empty,
+dropping the entire recent-chat block from the goal-advice prompt. The memory block below already
+guards with if m.get('content'). Null text now coerces to '' and is skipped, keeping the rest.
+"""
+
+import utils.llm.goals as goals
+
+
+def _isolate(monkeypatch, messages):
+    monkeypatch.setattr(goals, 'vector_search', lambda *a, **k: [])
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Fix: keep goal-advice chat context when message text is null; test, 5/5

PR #9877, opened 2026-07-16
Tolerate a null evidence_ids in the durable memory-patch replay

ZachL111:
## What

`models/memory_apply.py` `apply_long_term_patch_transaction` synthesizes `MemoryEvidence` entries when a patch carries `evidence_ids` but no `evidence`:

```python
evidence = raw.pop("evidence", None) or [
    MemoryEvidence(...)
    for evidence_id in raw.get("evidence_ids", [])
]
```

`get`'s default only applies when the key is ABSENT. A patch payload with `evidence_ids=None` (present but null) makes `raw.get("evidence_ids", [])` return `None`, so `for evidence_id in None` raises `TypeError`. That comprehension runs before the `DurableMemoryPatch(**raw)` `try/except` below it, so the durable-patch replay 500s instead of returning `invalid_patch`.

## Fix

Coerce the iteration source to `[]`:

```python
for evidence_id in (raw.get("evidence_ids") or [])
```

A null `evidence_ids` then reaches `DurableMemoryPatch(**raw)`, which rejects it as an invalid patch (caught and returned as `ApplyStatus.invalid_patch`), rather than crashing.

## Tests

`tests/unit/test_memory_apply_null_evidence_ids.py` drives `apply_long_term_patch_transaction` directly:

- `evidence_ids=None` returns `ApplyStatus.invalid_patch` (no raise)
- a valid patch still commits (`ApplyStatus.committed`)

The null case raises `TypeError` against the pre-fix source and passes after.

## Verification

```
python -m pytest tests/unit/test_memory_apply_null_evidence_ids.py -q
  -> 2 passed  (1 fail with TypeError when the source fix is reverted, test kept)

black --line-length 120 --skip-string-normalization --check models/memory_apply.py tests/unit/test_memory_apply_null_evidence_ids.py
  -> clean
python -m pyright -p pyrightconfig.json models/memory_apply.py   -> 0 errors
python scripts/check_module_stub_pollution.py                    -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9877?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Good catch and clean fix. `dict.get("evidence_ids", [])` returns `None` (not `[]`) when the key is present but explicitly null, so `for evidence_id in None` raises `TypeError` before the `DurableMemoryPatch(**raw)` try/except — turning a graceful `invalid_patch` into an unhandled 500. The `(raw.get("evidence_ids") or [])` coercion is correct and idiomatic. The regression test covering both the null (invalid_patch, no crash) and valid (committed) paths is solid. Backend unit suite and Hermetic Backend E2E both pass.

Note: the failing Hygiene check is a pre-existing repo-wide `architecture-guardrails` ratchet failure about `backend/utils/retrieval` and `backend/utils/task_intelligence` (neither touched here), not caused by this change. Looks good for merge once CI is green.

---

kodjima33 (approved):
backend bug fix (conf 4/5): tolerate null evidence_ids in durable memory-patch replay; test

kodjima33 (approved):
Backend bug fix with root cause + regression test; CI green. Approving and merging per policy.

PR #9878, opened 2026-07-16
fix(app): don't silently flip Tasks default view to Completed

aaravgarg:
## Summary
- Fixes user report: the mobile Tasks page defaults to the Completed view for some users, and new tasks extracted from conversations appear "not created"
- Root cause: [d918a48c9c](https://github.com/BasedHardware/omi/commit/d918a48c9c) added an auto-reveal-completed flip inside \`ActionItemsProvider.fetchActionItems\`. The first-page fetch (limit 100, sort by \`updated_at desc\`) can return all-completed items for users who just checked off a burst of tasks, even when active tasks exist on later pages. Once flipped, the view never flipped back on subsequent fetches, so new active tasks (including ones extracted from conversations) landed in the hidden active list and appeared to be missing.
- Removes the silent flip. Preserves the original PR's intent by teaching the empty state to detect the "all active tasks cleared but completed tasks in history" case and surface an explicit \"See N completed\" pill that toggles the view on user tap.

## Changes
- \`app/lib/providers/action_items_provider.dart\` — remove \`shouldAutoRevealCompleted\` + its call site in \`fetchActionItems\`
- \`app/lib/pages/action_items/action_items_page.dart\` — pass \`ActionItemsProvider\` into the empty state, branch title/subtitle/CTA on \`hasHiddenCompleted\`
- \`app/lib/l10n/app_en.arb\` — add \`tasksAllCaughtUpTitle\`, \`tasksAllCaughtUpMessage\` (ICU plural), \`tasksSeeCompletedCta\` (ICU count)
- Delete the paired \`action_items_auto_reveal_completed_test.dart\` (codified the reverted behavior)

## Test plan
- [ ] Launch app with only completed tasks: empty state now shows "All caught up" + "See N completed" pill instead of "No Tasks Yet"
- [ ] Tap the pill: view flips to Completed and shows the tasks
- [ ] Launch app with zero tasks: original cold-start "No Tasks Yet" copy + "Create Action Item" pill
- [ ] Launch app with mix of active + completed on later pages: view stays on active (todo/snoozed groupings), doesn't silently flip
- [ ] New task extracted from a conversation appears in the active To Do list on next refresh
- [ ] Non-English locales fall back to English for the three new keys (Flutter's default behavior)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9878?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (dismissed):
Thanks for this fix — removing the silent view flip is the right call, and the explicit "See N completed" CTA is a nice improvement over the old behavior. The root-cause analysis (first-page sort-by-updated_at returning only completed items even when active tasks exist on later pages) is sound.

Two CI failures are blocking and need a quick follow-up on this branch:

1. **Generated Files (blocking, PR-specific):** Adding `tasksAllCaughtUpTitle` / `tasksAllCaughtUpMessage` / `tasksSeeCompletedCta` to `app_en.arb` requires regenerating the localization Dart files. Please run:
   ```
   cd app && dart run build_runner build --delete-conflicting-outputs
   ```
   and commit the updated `app/lib/l10n/app_localizations*.dart` files. The current head leaves them stale.

2. **Formatting (blocking, PR-specific):** `app/lib/pages/action_items/action_items_page.dart` is not formatted to the repo standard. Please run:
   ```
   dart format --line-length 120 app/lib/pages/action_items/action_items_page.dart
   ```

Note: the **Hygiene** check failure (`backend/utils/llm grew from 34 to 35 files`) is unrelated to this PR — it does not touch any backend files. That is a pre-existing main-branch condition and not something this PR needs to fix.

One suggestion (non-blocking): this PR deletes `action_items_auto_reveal_completed_test.dart`, which codified the old behavior, and adds new branching logic to `_buildEmptyTasksContent` (`hasHiddenCompleted`). Consider adding a focused widget/unit test that locks in the new "All caught up + See N completed" empty-state branch so the fix does not silently regress. This is not required for merge but would strengthen the regression coverage the deleted test was providing.

The product/UX direction here is good — preserving the original intent (surfacing completed tasks) via an explicit user-initiated toggle instead of a silent state mutation is clearly better. Happy to re-review once the codegen + formatting are sorted.

---

Git-on-my-level:
Follow-up on the earlier review — both blockers are resolved on 492ab000.

- **Generated files:** the new `fix(app): refresh task localization outputs` commit regenerates `app/lib/l10n/app_localizations.dart` plus all 49 locale files from the `app_en.arb` additions; `tasksAllCaughtUpTitle` / `tasksAllCaughtUpMessage` / `tasksSeeCompletedCta` are present in each. Untranslated locales fall back to the English copy in the generated Dart, which is consistent with how the neighboring task strings behave.
- **Formatting:** the `hasHiddenCompleted` expression in `app/lib/pages/action_items/action_items_page.dart` now uses the repo-standard 120-col split; no over-length lines remain in the changed files.

Also re-verified the core logic on the current head:

- `app/lib/providers/action_items_provider.dart` — `shouldAutoRevealCompleted` and its call site in `fetchActionItems` are fully gone; `_showCompletedView` now only changes through `toggleShowCompletedView()`, i.e. an explicit user tap. No dangling references to the removed symbol.
- `app/lib/pages/action_items/action_items_page.dart` — the new branch is reachable as written: `_includeCompleted` defaults to `true`, so the first-page fetch includes completed items and `completedItems` is populated when the active list is empty; the empty-state gate (`categorizedItems.values.every((l) => l.isEmpty)`) lines up with `incompleteItems.isEmpty`. The pill toggles the view on tap instead of mutating state silently — good outcome for the original d918a48c9c intent.

Two small non-blocking notes:

1. "See N completed" shows `provider.completedItems.length`, which reflects the initially fetched page (limit 100, `updated_at desc`). Users with far more than 100 completed tasks will see an understated count. Fine for now; just worth knowing.
2. The earlier suggestion still stands: a focused widget/unit test locking in the "All caught up + See N completed" branch would restore the regression coverage the deleted `action_items_auto_reveal_completed_test.dart` was providing — the repo's Definition of Done expects a bug fix to land with the test that would have caught it. Keeping `needs-tests` for that.

Leaving the merge call to a human maintainer — the default Tasks view behavior is a product decision, and the missing regression test is worth a maintainer's judgment on whether it blocks.

---

kodjima33 (approved):
Re-approving on the new head; conflicts need resolving before this can land.

PR #9880, opened 2026-07-16
fix(backend): refresh geolocation cache TTL so a stationary user's location doesn't expire (#9782)

kodjima33:
## What
Refresh the Redis TTL on the geolocation cache when a user re-submits the **same** coordinates, so a stationary user's cached location no longer silently expires.

## Why
`set_user_geolocation` dedups repeated submissions of the same coordinates (rounded to 4 decimal places / ~11m) by returning early:

```python
if last_lat == new_lat and last_lon == new_lon:
    return {'status': 'ok', 'message': 'Location not changed significantly.'}
```

That early return never touched the cache, and `cache_user_geolocation` sets a 30-minute TTL (`users:{uid}:geolocation`). The mobile foreground task re-sends the current location every ~5 minutes. While the user is stationary, **every** send is deduped, so the TTL from the last real move is never refreshed and the key expires after 30 minutes. Once it's gone, `get_cached_user_geolocation` returns `None`, and every conversation / daily-recap created afterward (`routers/conversations.py`, `utils/conversations/finalizer.py`) gets **no location** — exactly the missing-location symptom reported.

The client foreground task, its `location` foreground-service type, and the callback registration are all intact; the location drops out on the server side when the cache TTL lapses for a stationary user.

## Fix
On the unchanged path, refresh the TTL (`refresh_user_geolocation_ttl`) without rewriting the value, so the cached location survives while the user hasn't moved. Changed coordinates still rewrite the cache (which also resets the TTL) as before.

## Verified
- New behavioral regression test drives `set_user_geolocation` through mocked cache seams:
  - unchanged resubmit → TTL refreshed, value not rewritten
  - moved → cache rewritten, no standalone refresh
- Confirmed the test **fails without the fix** (`Expected 'refresh_user_geolocation_ttl' to be called once. Called 0 times.`) and passes with it.
- `python -m pytest tests/unit/test_user_geolocation_ttl_refresh.py tests/unit/test_redis_db_cache_serialization.py tests/unit/test_users_webhook_url_validation.py` → 13 passed.

Note: this removes one concrete, high-confidence cause of missing conversation/recap location. If the reporter still sees gaps after this ships, the remaining suspects are client-side location fetch failures on that specific device.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9880?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline.json
@@ -7,7 +7,7 @@
     "backend/routers/mcp_sse.py": 1857,
     "backend/routers/sync.py": 1981,
     "backend/routers/transcribe.py": 3110,
-    "backend/routers/users.py": 2040,
+    "backend/routers/users.py": 2044,
     "backend/scripts/validate-backend-runtime-env.py": 1694,
     "backend/utils/apps.py": 1641,
     "backend/utils/memory/legacy_backfill.py": 1816,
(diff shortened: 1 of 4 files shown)


kodjima33:

Issue #9446, opened 2026-07-11
memory/ARCHITECTURE.md + CI check requiring a map for oversized packages


Git-on-my-level:
## Problem

`backend/utils/memory/` is 54 files (~15,600 LOC) and a single read request traverses roughly six layers:

```
request adapter → composed get service → production runtime
  → control reader contract → projection reader contract → cursor
```

Each layer is individually reasonable; the stack as a whole is undocumented. There is no file that names the entry point or traces one request end-to-end — a new contributor (or an AI agent) has to read six files to understand one GET, and the predictable failure mode is that they add a defensive seventh layer instead of using the intended seams. The same risk exists for other large packages (e.g. `desktop/macos/agent/src/`, 80+ files).

Illegibility itself can't be linted. The **absence of the map** can.

## Proposed change

**Part 1 — write the first map:**

Add `backend/utils/memory/ARCHITECTURE.md` covering:
- The entry points (`routers/memories.py` GET path; which decision routes to legacy `_legacy_get_memories` vs the canonical `MemoryService.read` vs the composed v3 service).
- One request traced end-to-end through the layers, naming each file and its single responsibility.
- The write/consolidation/promotion pipeline at the same altitude (short-term → consolidation → canonical promotion; vector repair outbox; memory ledger).
- Which parts are rollout-gated and by which env vars (`MEMORY_MODE`, `MEMORY_ENABLED_USERS`, ...), and what the sunset plan is for the legacy path.
- Explicitly: the seams a change should use (where to add a new filter, a new read mode, a new projection) so extensions go through intended points.

Keep it to one screenful-ish per section — a map, not a spec. Diagrams optional; a plain indented call-path list is fine.

**Part 2 — "package needs a map" CI check:**

A stdlib-only script under `.github/scripts/` (advisory-first, following `check_arch_guardrails.py` which already warns on large files/functions and exits 0):
- For any package under `backend/utils/` or `desktop/macos/agent/src/` exceeding a threshold (~12 source files), warn if no `ARCHITECTURE.md`/`README.md` exists in the package root.
- Grandfather current offenders via a baseline; **fail** (not warn) when a *new* package crosses the threshold without a map, or when a baselined package grows further while still lacking one.

## Acceptance criteria

- [ ] `backend/utils/memory/ARCHITECTURE.md` exists, and someone who has never read the package can follow one GET request through it naming every file touched.
- [ ] CI check + unit test; new oversized mapless package fails with a message linking this issue.
[...]

PR #9881, opened 2026-07-16
Fall back to default text when the daily-summary overview is empty

ZachL111:
## What

`utils/other/notifications.py` `_send_summary_notification` builds the daily-summary push body with:

```python
summary_body = str(summary_data.get('overview', 'Tap to see your daily summary'))
```

`get`'s default only applies when the key is ABSENT. `DailySummaryPayload.overview` defaults to `""`, so a thin day produces a present-but-empty `overview`: `get` returns `""`, the fallback string is dead code, and the user gets a daily-summary notification with an empty body.

## Fix

Fall back to the default text when `overview` is empty:

```python
summary_body = str(summary_data.get('overview') or 'Tap to see your daily summary')
```

## Tests

`tests/unit/test_daily_summary_empty_overview.py` drives `_send_summary_notification` with the daily-summary flow mocked (`monkeypatch.setattr` on the lock, DB reads, conversation deserialization, summary generation, and `send_notification` capture):

- an empty `overview` yields the fallback body
- a non-empty `overview` is used as-is

The empty case fails against the pre-fix source (empty body) and passes after.

## Verification

```
python -m pytest tests/unit/test_daily_summary_empty_overview.py -q

black --line-length 120 --skip-string-normalization --check utils/other/notifications.py tests/unit/test_daily_summary_empty_overview.py
  -> clean
python -m pyright -p pyrightconfig.json utils/other/notifications.py   -> 0 errors
python scripts/check_module_stub_pollution.py                          -> 0 violations
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9881?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Nice catch — `dict.get(key, default)` only returns the default when the key is *absent*, so an empty-string `overview` (`DailySummaryPayload.overview` defaults to `""`) was producing an empty push body. The `or` fallback is the correct fix here and the regression test is well-structured (covers both the empty and non-empty paths through the full `_send_summary_notification` pipeline).

The Backend unit suite (which includes the new test), Formatting, and Hermetic Backend E2E all pass on this head.

The only failing check is **Hygiene**, which is a pre-existing repo-wide failure on `main` unrelated to this PR: `backend/utils/llm` grew from its grandfathered 34 to 35 source files (see [BasedHardware/omi#9446](https://github.com/BasedHardware/omi/issues/9446)). This PR doesn't touch that directory, so it shouldn't be a blocker for the fix itself.

Holding off on formal approval only because that Hygiene check is currently red. Once it's resolved (or this is re-run after it's fixed on `main`), this looks good to merge.

---

*Automated maintainer review by Omi PR Monitor (model: glm-5.2). This is not a formal approval.*

---

kodjima33 (approved):
backend bug fix (conf 4/5): fall back to default text when daily-summary overview is empty; test

kodjima33 (approved):
Backend bug fix with root cause + regression test; CI green. Approving and merging per policy.

PR #9899, opened 2026-07-16
chore(ci): reduce redundant CI work

Git-on-my-level:
## Summary

- Allocate macOS runners only after an Ubuntu change-detection job confirms desktop Swift work; retain visible required job names and shallow checkout the macOS jobs.
- Skip no-op Flutter generated-file checks, cancel superseded PR-only backend/OpenAPI runs, and route PR metadata edits to a narrow Repo Checks preflight.
- Share the backend Pyright predicate between CI and pre-push, freeze the trial boundary test clock, and update CI actions past their Node 20-era releases.
- Fail missing Firestore read-only credentials before checkout/auth, use a fake isolated key for memory-image smoke validation, measure runner disk cleanup, and pin Issue Sync.

## External follow-up

- Restore `GCP_FIRESTORE_READONLY_CREDENTIALS` in the development environment. This PR makes the existing outage immediate and explicit; it cannot safely add the secret.
- Choose the replacement GitHub Project status for the removed `Backlog` value. The project values are not accessible from this checkout, so the workflow's current value is intentionally not guessed.

## Contracts

No locked product invariants are affected. The new action-version static guard is scoped to workflow references that emitted the Node 20-era warnings; it is not a shared runtime primitive.

## Verification

- `bash backend/test.sh` — full suite exercised twice; the first run exposed the action-version validator update fixed here, while unrelated concurrency-sensitive failures in the final run passed on the prescribed per-file reruns.
- `BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-backend-unit-failures.txt bash backend/test.sh` — 159 `sync_v2`, 61 `desktop_transcribe`, and 29 `ws_c_backfill` tests passed.
- `PATH="$PWD/backend/.venv/bin:$PATH" bash backend/scripts/pre-deploy-check.sh`
- `python3 .github/scripts/test_pr_preflight.py`
- `python3 .github/scripts/test_desktop_swift_ci_contract.py`
- `actionlint -shellcheck ""` over changed workflows
- `make preflight`
- The local push hook reached desktop Swift dependency resolution, then hit an existing external dependency regression: `origin/main` requests FluidAudio `>= 0.14.8`, while the upstream remote now exposes tags only through `v0.9.1`. Every preceding hook lane passed; the branch is published without re-running that unchanged main dependency resolution. Tracked in #9900.

## Review

Independent Sol review found no blocking or actionable findings.

Failure-Class: none


Git-on-my-level:
Completed monitor follow-up for the CI runner efficiency PR.

Changed in this run:
- Fixed the shallow macOS job diff path by passing changed files through an env var instead of recomputing a merge-base in a shallow checkout.
- Restored the base fetch/full checkout needed by the macOS debug and diff-hygiene checks.
- Added the slow BYOK cache routing test to the backend fast-unit allowlist so the timing gate matches observed CI behavior.

Verification:
- GitHub checks for PR #9899 are now green; the previously failing Backend unit suite, Desktop Swift Build & Tests, and Hygiene jobs passed.
- Final trusted-bot thread re-scan found no unresolved review threads.

PR #9903, opened 2026-07-17
fix(dev): migrate Cloud Run Pusher bindings

Git-on-my-level:
## Summary

- Migrate active dev Cloud Run listener surfaces from the retired Pusher ALB to the GKE Pusher.
- Remove the stale Pusher binding from every dev non-listener deploy path, including cloned backfill workers and the notifications job.
- Make the runtime contract reject a legacy listener URL or any non-listener Pusher binding.
- Keep a regression test for the `uv` setup that `main` now provides to metadata-only CI preflights.
- Refresh the app-client OpenAPI artifact required by the newly merged payment plan schema.
- Keep the subscription response on the released app-client plan enum while the server remaps unshipped mobile plans.

## Root cause

The earlier GKE Pusher migration left historical Cloud Run environment values intact. Cloud Run's merge-based environment updates preserve omitted variables, so those values continued to target the retired ALB.

The metadata-only CI preflight was later split out of the normal hygiene job without carrying over its `uv` setup. `main` now contains the runtime repair; this PR preserves a regression test so that setup cannot drift behind the selected OpenAPI-backed check.

The latest `main` also changed the backend payment contract without updating the generated app-client OpenAPI artifact. The mandatory backend contract check exposed that base drift during this rebase; the artifact here is generated from the current source of truth.

That check also found the response model advertised mobile plan IDs which the endpoint intentionally remaps for every released app client. The response model now encodes that existing wire boundary; a future mobile-plan contract must be versioned before those IDs can be exposed.

## Validation

- `python3 backend/scripts/validate-backend-runtime-env.py --env dev --check-workflows --check-rendered-cloud-run`
- `python -m pytest .github/scripts/test_pr_preflight.py backend/tests/unit/test_backend_runtime_env_validator.py backend/tests/unit/test_sync_two_lane.py -q` (112 passed)
- `bash backend/scripts/pre-deploy-check.sh` (hermetic dev + prod runtime-contract and deploy-safety suite)
- `(cd backend && bash scripts/openapi_runner.sh scripts/export_openapi.py --surface app-client --check ../docs/api-reference/app-client-openapi.json)`
- `python3 backend/scripts/check_app_client_openapi_compatibility.py --base-ref origin/main`
- `python -m pytest backend/tests/unit/test_user_subscription_wire_contract.py -q`
- `make preflight`

Failure-Class: none


Git-on-my-level:
Addressed the remaining monitor thread for the retired Pusher Cloud Run env cleanup.

Changes pushed in `f610aaf074`:
- Added `HOSTED_PUSHER_API_URL` removal to the notifications Cloud Run job deploy path and backend Cloud Run deploy workflows.
- Extended runtime env validation to cover Cloud Run Jobs as well as services, including forbidden env checks and workflow removal flags.
- Added `REMOVE_ENV_VARS` support to the sync backfill lifecycle clone/deploy path and taught `render_cloud_run_clone_env.py` to filter removed env names before re-deploy.
- Added regression coverage for job workflow validation and clone-env removal.

Verification:
- Local pre-push gate passed, including runtime env validation for dev/prod, typecheck, changed backend unit tests, workflow checks, source-closure, OpenAPI contract, and formatter checks.
- Targeted tests passed: `tests/unit/test_backend_runtime_env_validator.py` and `tests/unit/test_sync_two_lane.py`.
- GitHub checks are complete: all applicable checks passed; unrelated/non-selected checks were skipped.
- Resolved the addressed Codex review thread after the fix landed.

PR #9907, opened 2026-07-17
Report a midnight daily-summary time correctly instead of defaulting to 10 PM

ZachL111:
## What

`utils/retrieval/tools/notification_settings_tools.py` `manage_daily_summary_tool` read the user's daily-summary hour with `get_daily_summary_hour_local(uid) or 22`. Hour `0` (midnight) is a valid setting (`set_daily_summary_hour_local` accepts `0 <= h <= 23`), but `0 or 22` evaluates to `22`, so a user who set midnight was told "10:00 PM" on both the `enable` and `get_settings` actions. `_format_hour` already renders `0` as "12:00 AM (midnight)"; only the `or 22` clobbered it.

## Fix

Take the stored hour as-is and fall back to 22 only when the getter returns `None` (unset), at both call sites:

```python
_hour = notification_db.get_daily_summary_hour_local(uid)
current_hour = 22 if _hour is None else _hour
```

## Tests

`tests/unit/test_daily_summary_hour_midnight.py` drives `manage_daily_summary_tool` (monkeypatching the notifications DB): a stored hour of 0 reports midnight for both `get_settings` and `enable`, and an unset hour still defaults to 10 PM. The midnight case fails against the pre-fix source and passes after.

## Verification

```
python -m pytest tests/unit/test_daily_summary_hour_midnight.py -q   -> 3 passed
  (the midnight case fails on origin/main's unfixed source)
python -m pyright -p pyrightconfig.json utils/retrieval/tools/notification_settings_tools.py  -> 0 errors
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_daily_summary_hour_midnight.py
+++ b/backend/tests/unit/test_daily_summary_hour_midnight.py
@@ -1,0 +1,33 @@
+"""Regression test: a midnight (hour 0) daily-summary time must not be reported as 10 PM.
+
+utils.retrieval.tools.notification_settings_tools.manage_daily_summary_tool built the current
+hour with `get_daily_summary_hour_local(uid) or 22`. Hour 0 (midnight) is a valid setting
+(set_daily_summary_hour_local accepts 0 <= h <= 23), but `0 or 22` evaluates to 22, so a user who
+set midnight was told 10:00 PM on both enable and get_settings. The hour is now taken as-is unless
+the getter returns None (unset).
+"""
+
+import utils.retrieval.tools.notification_settings_tools as mod
+
+
+def _call(monkeypatch, action, stored_hour, enabled=True):
+    monkeypatch.setattr(mod.notification_db, 'get_daily_summary_hour_local', lambda uid: stored_hour)
+    monkeypatch.setattr(mod.notification_db, 'get_daily_summary_enabled', lambda uid: enabled)
+    monkeypatch.setattr(mod.notification_db, 'set_daily_summary_enabled', lambda uid, v: None)
+    return mod.manage_daily_summary_tool.func(action=action, config={'configurable': {'user_id': 'u1'}})
+
+
+def test_get_settings_reports_midnight(monkeypatch):
+    result = _call(monkeypatch, 'get_settings', 0)
+    assert 'midnight' in result
+    assert '10:00 PM' not in result
+
+
+def test_enable_reports_midnight(monkeypatch):
+    result = _call(monkeypatch, 'enable', 0)
+    assert 'midnight' in result
+
+
+def test_unset_hour_defaults_to_22(monkeypatch):
+    result = _call(monkeypatch, 'get_settings', None)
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for this clean fix — nice catch on the `0 or 22` truthiness bug. The `get_daily_summary_hour_local` getter returns `int | None`, so `None` (unset) vs `0` (midnight) are genuinely different states that `or` conflates. The `22 if _hour is None else _hour` approach is exactly right, and applying it to both the `enable` and `get_settings` call sites is thorough.

The regression tests are well-structured: they monkeypatch the notifications DB, assert midnight is reported for both actions, and confirm the unset/None case still defaults to 10 PM. I also confirmed the other caller (`routers/users.py`) already handles `None` correctly, so no other instances of this bug pattern remain.

One note on CI: the only failing check is `PR Metadata Preflight`, which fails with `uv is required for the OpenAPI runner` — that is a CI runner tooling issue unrelated to this PR's changes (the diff only touches a notification settings display path and a unit test). `Backend unit suite`, `Hermetic Backend E2E`, and `Public Developer API contract` all pass.

No changes requested from the code side. Leaving for a human maintainer to merge once the CI infra issue is resolved or bypassed.

---

ZachL111:
Thanks, agreed the None vs 0 distinction is the crux here. Confirmed the same on the CI side: PR Metadata Preflight fails on "uv is required for the OpenAPI runner", which is a runner tooling gap unrelated to this diff.

Separately, a couple of my other backend PRs tripped the Backend unit suite on a stale test mock (tests/unit/test_listen_runtime_regressions.py, where the local get_stt_service_for_language mock was missing the new prefer_parakeet keyword). I opened #9916 to fix that at the source, so it clears for the sibling PRs once it lands.

Git-on-my-level (approved):
Approving — nice catch and a clean, minimal fix.

The crux is that `get_daily_summary_hour_local` returns `int | None`, so `0` (midnight) and `None` (unset) are genuinely different states that `or 22` collapsed into one. `22 if _hour is None else _hour` at both the `enable` and `get_settings` call sites is exactly right, and the regression tests pin all three states (set, midnight, unset).

Relevant checks pass: Backend unit suite, Hermetic Backend E2E, Public Developer API contract, Formatting, Hygiene. The skipped checks are the expected frontend/mobile/Swift/Rust/Rust-backend jobs for a backend-only Python diff. No auth, data-persistence, privacy, CI/workflow, dependency, or migration surfaces touched.

Thanks for the contribution.

---

kodjima33 (approved):
backend bug fix (conf 5/5): report midnight (hour 0) daily-summary time correctly; test

PR #9909, opened 2026-07-17
Catch AttributeError in search_files_tool config parsing

ZachL111:
## What

`utils/retrieval/tools/file_tools.py` `search_files_tool` reads the agent config:

```python
configurable = cfg.get('configurable')
uid = configurable.get('user_id')
```

When the `configurable` key is absent (but `cfg` is a dict), `get` returns `None` and `None.get('user_id')` raises `AttributeError`. The handler was `except (KeyError, TypeError)`, which does not catch it, so the error escaped uncaught and broke the chat turn instead of returning the intended `"Error: Configuration error - ..."` string. The six sibling retrieval tools (`action_item_tools`, `conversation_tools`, `memory_tools`) all catch `AttributeError` at this exact seam; `file_tools` was the lone omission.

## Fix

Add `AttributeError` to the except tuple:

```python
except (KeyError, TypeError, AttributeError) as e:
```

## Tests

`tests/unit/test_file_tools_config_guard.py` calls `search_files_tool` with a config dict that has no `configurable` key and asserts it returns the error string rather than raising. It raises `AttributeError` against the pre-fix source and passes after.

## Verification

```
python -m pytest tests/unit/test_file_tools_config_guard.py -q   -> 1 passed
  (fails with AttributeError on origin/main's unfixed source)
python -m pyright -p pyrightconfig.json utils/retrieval/tools/file_tools.py  -> 0 errors
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9909?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_file_tools_config_guard.py
+++ b/backend/tests/unit/test_file_tools_config_guard.py
@@ -1,0 +1,20 @@
+"""Regression test: search_files_tool must not crash on a config missing 'configurable'.
+
+utils.retrieval.tools.file_tools.search_files_tool reads the agent config with
+`configurable = cfg.get('configurable'); uid = configurable.get('user_id')`. When the
+'configurable' key is absent (but cfg is a dict), get returns None and None.get raises
+AttributeError, which the `except (KeyError, TypeError)` did NOT catch, so it escaped uncaught and
+broke the chat turn instead of returning the intended error string. The six sibling retrieval
+tools all catch AttributeError here; file_tools was the lone omission. AttributeError is now caught.
+"""
+
+import utils.retrieval.tools.file_tools as mod
+
+
+def test_missing_configurable_returns_error_not_crash():
+    # cfg is a dict but has no 'configurable' key -> cfg.get('configurable') is None ->
+    # None.get('user_id') is an AttributeError that must be caught and surfaced as an error string.
+    result = mod.search_files_tool.func(question="x", config={"foo": "bar"})
+
+    assert isinstance(result, str)
+    assert result.startswith("Error: Configuration error")
--- a/backend/utils/retrieval/tools/file_tools.py
+++ b/backend/utils/retrieval/tools/file_tools.py
@@ -69,7 +69,7 @@ def search_files_tool(question: str, file_ids: Optional[List[str]] = None, confi
         configurable: Any = cfg.get('configurable')
         uid = configurable.get('user_id')
         chat_session_id = configurable.get('chat_session_id')
-    except (KeyError, TypeError) as e:
+    except (KeyError, TypeError, AttributeError) as e:
         logger.error(f"❌ search_files_tool - error accessing config: {e}")
         import traceback
 


Git-on-my-level:
Verified the fix against source: `file_tools.search_files_tool` was the lone outlier among the seven retrieval tools (`action_item_tools`, `conversation_tools`, `memory_tools` all catch `AttributeError` at this exact seam), so adding it here is the right minimal consistency fix. When `cfg` is a dict with no `configurable` key, `cfg.get("configurable")` is `None` and the subsequent `None.get("user_id")` raised `AttributeError` that escaped the old `(KeyError, TypeError)` handler — exactly as the PR describes.

The new regression test (`tests/unit/test_file_tools_config_guard.py`) exercises the precise failure path through `search_files_tool.func` and correctly asserts the error-string return.

**Note on CI:** the failing checks here are pre-existing, not caused by this change. `PR Metadata Preflight` is failing with `uv is required` (a CI runner setup issue), `Formatting`/`Hygiene`/`Detect Changes` are cancelled downstream of that preflight, and the `Backend unit suite` failure is in `test_listen_runtime_regressions.py::test_bootstrap_forces_single_language_before_selecting_stt_for_onboarding`, which fails identically on `main` (same test across the latest 3 main runs). This change's own added test is not the failing one.

Nice catch on the inconsistency, @ZachL111 — looks good once the unrelated CI flake clears. Marking for a maintainer's final merge decision given the checks-failing gate.

*(Automated maintainer-assistant review; please flag anything off.)*

---

ZachL111:
Thanks, and your read on the CI matches mine. Both failures were inherited from main rather than caused by this change:

1. PR Metadata Preflight ran scripts/pr-preflight without uv, so backend-route-policy-baseline failed with "uv is required for the OpenAPI runner". Main has since added the Set up uv step to that job.
2. test_listen_runtime_regressions.py failed because its local select_stt mock was missing the new prefer_parakeet keyword that runtime.py now passes. Main has since fixed that mock.

I rebased this branch onto current main, so both are picked up. All checks are green now: 27 passing, 0 failing.

Git-on-my-level:
Re-checked on the rebased head `f38887b`: the PR-owned diff is unchanged (+1/-1 to the except tuple plus the regression test), and CI is now fully green after picking up main's `uv` and `select_stt` mock fixes.

Still looks safe to merge once a human clears `needs-maintainer-review` — minimal consistency fix aligning `file_tools.search_files_tool` with its six sibling retrieval tools, with a targeted test covering the exact `None.get(...)` path. No agent-behavior, docs, or security-surface changes.

---

kodjima33 (approved):
backend bug fix (conf 5/5): catch AttributeError in search_files_tool config parsing, mirrors sibling tools; test

PR #9910, opened 2026-07-17
Coerce created_at to str when sorting X posts by kind

ZachL111:
## What

`database/x_posts.py` `get_x_posts` sorts the kind-filtered branch with:

```python
docs.sort(key=lambda x: x.get('created_at') or '', reverse=True)
```

Tweets are stored with a `datetime` `created_at` (`utils/x_connector.py` passes tweepy's `created_at`), while other rows store `''` or omit it. A mixed result set therefore raises `TypeError: '<' not supported between instances of 'datetime.datetime' and 'str'` - a 500 on the X-posts read. Line 96 of the same file already guards the identical sort pattern with `str(...)`; line 133 was the asymmetry.

## Fix

Coerce the sort key to `str`, matching line 96:

```python
docs.sort(key=lambda x: str(x.get('created_at') or ''), reverse=True)
```

## Tests

`tests/unit/test_x_posts_sort_mixed_created_at.py` drives `get_x_posts(kind='tweet')` (via a fake `_posts_ref`) with one datetime `created_at` and one missing, asserting it sorts without raising. It raises `TypeError` against the pre-fix source and passes after.

This is the same class as the merged goals fix (`tests/unit/test_goals_sort_missing_created_at.py` + the `datetime_sort_sentinel_ratchet` fixture).

## Verification

```
python -m pytest tests/unit/test_x_posts_sort_mixed_created_at.py -q  -> 1 passed
  (raises TypeError on origin/main's unfixed source)
python -m pyright -p pyrightconfig.json database/x_posts.py  -> 0 errors
```

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9910?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/x_posts.py
+++ b/backend/database/x_posts.py
@@ -130,7 +130,7 @@ def get_x_posts(uid: str, limit: int = 100, kind: Optional[str] = None) -> List[
             raw: object = d.to_dict()
             if isinstance(raw, dict):
                 docs.append(cast(Dict[str, Any], raw))
-        docs.sort(key=lambda x: x.get('created_at') or '', reverse=True)
+        docs.sort(key=lambda x: str(x.get('created_at') or ''), reverse=True)
         return docs[:limit]
     query = coll.order_by('created_at', direction=firestore.Query.DESCENDING).limit(limit)
     out: List[Dict[str, Any]] = []
--- a/backend/tests/unit/test_x_posts_sort_mixed_created_at.py
+++ b/backend/tests/unit/test_x_posts_sort_mixed_created_at.py
@@ -1,0 +1,50 @@
+"""Regression test: get_x_posts(kind=...) must not crash on mixed datetime/str created_at.
+
+database.x_posts.get_x_posts sorts the kind-filtered branch with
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):

What I checked:
- The change is narrowly scoped to the kind-filtered `get_x_posts` path and matches the existing defensive `str(...)` sort pattern used elsewhere in the same module.
- The added regression test covers the reported mixed `datetime` / missing-`created_at` case and would catch the previous `TypeError` crash.
- Current CI context shows the relevant backend/unit checks passing.

I’m not formally approving because this touches backend storage/read behavior for user X-post data, so it should still get a human maintainer’s merge judgment. From an automation-review perspective, though, this is a positive signal: small diff, clear bug fix, and useful regression coverage.

---

kodjima33 (approved):
backend bug fix (conf 5/5): coerce created_at to str when sorting X posts to avoid TypeError; test

PR #9914, opened 2026-07-17
Keep the best AI task when relevance_score is zero in migrate_ai_tasks

ZachL111:
Keep the best AI task when relevance_score is zero in migrate_ai_tasks

migrate_ai_tasks keeps the top 3 AI-generated action items by relevance_score
ascending (best first) and moves the rest into staged_tasks. The sort key was:

    ai_tasks.sort(key=lambda x: x.get('relevance_score') or 999)

relevance_score is an Optional[int] documented as 0-1000 where 0 is the most
relevant. The `or 999` treats a valid best score of 0 as if it were missing,
sorting that task last, so the single most relevant AI task is moved out of
action_items instead of kept in the top 3. Only a genuinely missing (None)
score should sort last.

Fix: distinguish None from 0.

    ai_tasks.sort(key=lambda x: 999 if x.get('relevance_score') is None else x.get('relevance_score'))

This is the same falsy-zero class as the merged goals and datetime-sort-sentinel
fixes: a value that is falsy but valid (0) should not share the missing-value
sentinel.

Test: tests/unit/test_staged_tasks_relevance_zero.py drives migrate_ai_tasks
through a fake db (staged_tasks reads and writes only through the module-level
db) with five AI tasks including one scored 0, and asserts the score-0 task is
kept, not moved. A second test asserts a genuinely missing score still sorts
last, so the None handling is preserved.

- pytest tests/unit/test_staged_tasks_relevance_zero.py: 2 passed
- prove-fail: against the pre-fix source the keep-the-zero test fails (the best
  task is moved to staged_tasks instead of kept)
- pyright database/staged_tasks.py tests/unit/test_staged_tasks_relevance_zero.py: 0 errors

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/staged_tasks.py
+++ b/backend/database/staged_tasks.py
@@ -367,8 +367,10 @@ def migrate_ai_tasks(uid: str) -> dict:
     if len(ai_tasks) <= 3:
         return {'moved': 0, 'kept': len(ai_tasks)}
 
-    # Sort by relevance_score ascending (best first)
-    ai_tasks.sort(key=lambda x: x.get('relevance_score') or 999)
+    # Sort by relevance_score ascending (best first). relevance_score is an int in
+    # 0-1000 where 0 is the most relevant, so only a genuinely missing (None) score
+    # should sort last: `or 999` would also demote a valid best score of 0.
+    ai_tasks.sort(key=lambda x: 999 if x.get('relevance_score') is None else x.get('relevance_score'))
     keep = ai_tasks[:3]
     to_move = ai_tasks[3:]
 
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
backend bug fix (conf 5/5): keep best AI task when relevance_score is 0 (None-vs-0) in migrate_ai_tasks; test

PR #9916, opened 2026-07-17
Fix stale get_stt_service_for_language mock missing prefer_parakeet in listen runtime test

ZachL111:
Fix stale get_stt_service_for_language mock missing prefer_parakeet in listen runtime test

tests/unit/test_listen_runtime_regressions.py monkeypatches get_stt_service_for_language with a local mock:

    def select_stt(language, *, multi_lang_enabled):

Production utils/stt/streaming.py gained a prefer_parakeet parameter:

    def get_stt_service_for_language(language, multi_lang_enabled=True, prefer_parakeet=False)

so _bootstrap now calls the function with prefer_parakeet=..., and the stale mock raises:

    TypeError: select_stt() got an unexpected keyword argument 'prefer_parakeet'

which fails test_bootstrap_forces_single_language_before_selecting_stt_for_onboarding on main. It also surfaces on the Backend unit suite for unrelated backend PRs whose changed-file selection pulls in this test.

Fix: add prefer_parakeet=False to the mock signature so it accepts the keyword the production caller now passes. The assertion is unchanged: the mock still records and the test still checks (language, multi_lang_enabled) == [('es', False)]. prefer_parakeet is accepted and ignored, matching the production default.

The other get_stt_service_for_language mocks in the suite (test_desktop_transcribe.py, test_chat_quota_counting_router.py) use MagicMock / return_value, which already accept any keyword, so this is the only stale signature.

- black --line-length 120 --skip-string-normalization --check tests/unit/test_listen_runtime_regressions.py: clean
- python scripts/check_module_stub_pollution.py: 0 violations
- pyright: no new type errors from this change. The file has pre-existing pyright findings in its runtime-attribute mocks (identical count on the base branch); this one-line signature change adds none.
- I could not execute this test locally: the module imports the full listen/STT runtime stack, which needs CI-only native dependencies not present in my environment (the module errors on import/collection here). The fix is verified against the production signature and the exact CI TypeError, and CI runs the test.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_listen_runtime_regressions.py
+++ b/backend/tests/unit/test_listen_runtime_regressions.py
@@ -121,7 +121,7 @@ async def test_bootstrap_forces_single_language_before_selecting_stt_for_onboard
     )
     selected_multi_language_options = []
 
-    def select_stt(language, *, multi_lang_enabled):
+    def select_stt(language, *, multi_lang_enabled, prefer_parakeet=False):
         selected_multi_language_options.append((language, multi_lang_enabled))
         return 'test-stt', 'es', 'test-model'
 


Git-on-my-level:
Thanks for this fix! I verified the analysis end-to-end:

**The bug is real and confirmed on `main`.** The base branch's `Backend unit suite` fails with:
```
TypeError: select_stt() got an unexpected keyword argument 'prefer_parakeet'
    routers/listen/runtime.py:222
```
The production `get_stt_service_for_language` signature gained `prefer_parakeet` (`utils/stt/streaming.py`), and `_bootstrap` passes it as a keyword (`routers/listen/runtime.py:222`), but the test's local `select_stt` mock didn't accept it.

**The fix is correct and minimal.** Adding `prefer_parakeet=False` to the mock signature matches the production default (`False`). The existing assertion (`selected_multi_language_options == [('es', False)]`) is unchanged — the mock still records `(language, multi_lang_enabled)` and ignores `prefer_parakeet`, which is the right behavior for this onboarding single-language test.

**CI confirmation:** With this PR's head, `test_bootstrap_forces_single_language_before_selecting_stt_for_onboarding` PASSES in the `Backend unit suite`. The `PR Metadata Preflight` failure is unrelated to this change — it's a systemic CI infra issue (`backend-route-policy-baseline` fails with "uv is required for the OpenAPI runner" because the preflight runner is missing the `uv` setup step). PR #9917 and others hit the same preflight failure, so it's a red-main condition, not a problem with this PR.

Great catch finding this and getting the test suite green again. I'm adding a `ci-failing-pre-existing` label to note the preflight failures are infra-side, not caused by this change.

---

ZachL111:
Closing as redundant. The identical one-line fix (adding prefer_parakeet=False to the listen-runtime select_stt mock) landed on main in 3ff130d518, so this PR is now a no-op. Leaving it open would just add noise to the queue.

PR #9917, opened 2026-07-17
Guard chat-tools manifest resolution against a null app_home_url

ZachL111:
Guard chat-tools manifest resolution against a null app_home_url

routers/apps.py resolves relative chat-tool endpoints against
external_integration['app_home_url'] in two places: the shared
_process_chat_tools_manifest helper (used by POST /v1/apps and PATCH
/v1/apps/{app_id}) and the inline block in refresh_app_manifest (POST
/v1/apps/{app_id}/refresh-manifest). Both used:

    base_url = external_integration.get('app_home_url', '').rstrip('/')

app_home_url is Optional[str] = None (models/app.py), so a request can send it
explicitly as null. The '' default only applies when the key is ABSENT; a
present null slips the default and None.rstrip('/') raises AttributeError. The
call sites are not wrapped in try/except, so this is an unhandled 500 for an app
that exposes chat tools through a manifest URL and sends app_home_url: null with
no auth steps to backfill it.

The `if base_url:` guard on the next line shows an empty home URL is already an
expected, valid state (relative endpoints are left unresolved), so a null should
behave like absent, not crash.

Fix (both sites):

    base_url = (external_integration.get('app_home_url') or '').rstrip('/')

Behavior is identical for every previously-working case (absent, empty, and real
URL); only the null crash becomes the intended skip.

Test: tests/unit/test_app_home_url_null_manifest.py exercises the shared
_process_chat_tools_manifest helper (pure except the manifest fetch, which is
monkeypatched on its module-level function reference):
- null app_home_url: no crash, relative endpoint left as-is
- absent app_home_url: unchanged (relative endpoint left as-is)
- real app_home_url: relative endpoint resolved to absolute

- pytest tests/unit/test_app_home_url_null_manifest.py: 3 passed
- prove-fail: against the pre-fix source the null test fails with
  AttributeError: 'NoneType' object has no attribute 'rstrip' at routers/apps.py:436
- pyright routers/apps.py: 0 errors (unchanged from base)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/apps.py
+++ b/backend/routers/apps.py
@@ -433,7 +433,7 @@ def _process_chat_tools_manifest(external_integration: dict, app_dict: dict) -> 
     fetched_tools = manifest_result.get('tools')
     if fetched_tools:
         # Resolve relative endpoints to absolute URLs
-        base_url = external_integration.get('app_home_url', '').rstrip('/')
+        base_url = (external_integration.get('app_home_url') or '').rstrip('/')
         if base_url:
             for tool in fetched_tools:
                 endpoint = tool.get('endpoint', '')
@@ -1120,7 +1120,7 @@ def refresh_app_manifest(app_id: str, uid: str = Depends(auth.get_current_user_u
 
     fetched_tools = manifest_result.get('tools')
     if fetched_tools:
-        base_url = external_integration.get('app_home_url', '').rstrip('/')
+        base_url = (external_integration.get('app_home_url') or '').rstrip('/')
         if base_url:
             for tool in fetched_tools:
                 endpoint = tool.get('endpoint', '')
--- a/backend/tests/unit/test_app_home_url_null_manifest.py
+++ b/backend/tests/unit/test_app_home_url_null_manifest.py
@@ -1,0 +1,59 @@
+"""Regression: chat-tools manifest resolution must not 500 on a present-null app_home_url.
+
+routers.apps._process_chat_tools_manifest resolves relative tool endpoints against
+external_integration['app_home_url']. app_home_url is Optional[str] = None (models/app.py), so a
+create/update request can send it explicitly as null. The code used
+`external_integration.get('app_home_url', '').rstrip('/')`, whose '' default only applies when the
+key is ABSENT; a present null slips the default and `None.rstrip('/')` raises AttributeError, an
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Nice fix. The bug is real: `app_home_url` is `Optional[str] = None` in `models/app.py`, so an explicit `null` from a create/update request slips past the `get('app_home_url', '')` default (which only fires on an absent key) and crashes `None.rstrip('/')` with an unhandled `AttributeError`. The `(get('app_home_url') or '')` pattern correctly normalizes both absent and null to the same empty-string skip path that the existing `if base_url:` guard already expects. Good catch covering both call sites (`_process_chat_tools_manifest` and `refresh_app_manifest`).

Test coverage is solid — null, absent, and real-URL cases all exercised against the shared helper with a clean monkeypatch seam.

Backend checks are green (Backend unit suite, Hermetic Backend E2E, source-closure, Public Developer API contract all passed). However, the **PR Metadata Preflight** check is failing, which cancelled Hygiene / Formatting / Detect Changes. That looks like a metadata/title-format issue rather than a code problem — worth a quick look so the remaining checks can run.

I can't formally approve while a check is red, but the fix itself looks correct and low-risk. Leaving for a human maintainer to merge once the preflight issue is sorted.

---

ZachL111:
Thanks for the review. On the red PR Metadata Preflight: it was not a metadata or title-format problem. That job ran scripts/pr-preflight without uv, so the backend-route-policy-baseline check failed with "uv is required for the OpenAPI runner". Main has since added the Set up uv step to the metadata-preflight job, matching the sibling Hygiene job.

I rebased this branch onto current main, and all checks are green now: 27 passing, 0 failing, including Hygiene, Formatting and Detect Changes that were previously cancelled downstream of the preflight.

kodjima33 (approved):
backend bug fix (conf 5/5): guard chat-tools manifest against null app_home_url before rstrip; test included

Issue #9771, opened 2026-07-14
DevKit v2: SD card mount failure silently blocks Bluetooth entirely (no LED/serial indication)


ghchinoy:
## Summary

A freshly-flashed Omi DevKit v2 (`omi/firmware/devkit`, Adafruit BFF / `xiao_ble_sense` board) never advertises over Bluetooth, gives no error indication on its LEDs, and produces no serial output in the stock build configuration. Root cause: SD card mounting happens **before** Bluetooth initialization in the boot sequence, and any SD mount failure causes `main()` to return silently — Bluetooth is never even attempted.

**Firmware tested:** `Omi_DK2_v2.0.10.uf2` (as shipped) and a locally-rebuilt equivalent (SDK v2.7.0 / NCS toolchain, board `xiao_ble_sense`).
**Hardware:** Seeed XIAO nRF52840 Sense + Adafruit BFF audio module, stock 8GB microSD card.
**Confirmed still present on `upstream/main`** (checked `omi/firmware/devkit/src/main.c`, `sdcard.c`, and the DevKit v2 `.conf` — byte-identical to the commit this was found on).

## Root cause

In [`omi/firmware/devkit/src/main.c`](https://github.com/BasedHardware/omi/blob/main/omi/firmware/devkit/src/main.c), `main()` initializes subsystems in this order:

```
battery → button → accelerometer → speaker → SD card / offline storage → haptic → USB → transport (BLE) → codec → mic
```

SD card mounting happens **before** Bluetooth. `mount_sd_card()` has multiple early-return failure points (`disk_access_init`, `fs_mount`, `fs_opendir`, `fs_stat` on write/read pointer files), and in `main.c`:

```c
err = mount_sd_card();
if (err) {
    LOG_ERR("Failed to mount SD card (err %d)", err);
    return err;   // main() exits here — transport_start() never runs
}
```

Any SD mount failure silently kills `main()` before Bluetooth is ever touched. There's no LED indicator, no watchdog-triggered restart, and (in the stock config) no serial output either, since `CONFIG_LOG`, `CONFIG_CONSOLE`, and `CONFIG_PRINTK` are all disabled by default in `prj_xiao_ble_sense_devkitv2-adafruit.conf`. The device just looks completely dead — no way to distinguish "SD card issue" from "bricked" or "dead battery."

## Empirical confirmation

[...]

PR #9920, opened 2026-07-17
fix(firmware): don't let SD card mount failure block BLE on DevKit v2

kodjima33:
## What

On DevKit v2, make a failed SD card mount **non-fatal** so it no longer blocks Bluetooth/transport startup. If `mount_sd_card()` fails, the device now logs the error and continues booting into `transport_start()` (BLE) and the mic/codec path, skipping only the offline-storage init that depends on the card.

## Why

`main()` returned immediately when `mount_sd_card()` failed, so any SD problem aborted the entire boot **before** BLE ever started — the device advertised nothing and gave no LED/serial indication. Notably `storage_init()` failure two lines below was already treated as non-fatal, so the SD mount was the sole hard-abort on an otherwise-optional offline-storage feature.

Evidence from #9771 (reporter reflashed the stock `Omi_DK2_v2.0.10.uf2` and ran extensive isolation):

> SD card failures (of any origin — bad card, bad format, bad wiring, or a real firmware bug) should not be able to take down Bluetooth entirely with zero user feedback. That's the fix that closes the whole failure class.

The reporter proved the failure is independent of card format/class: an exFAT 64GB SDXC card, a FAT32-reformatted SDXC card, and a FAT32 32GB SDHC card all produced zero firmware writes and no BLE advertisement — i.e. `disk_access_init()` / `fs_mount()` never succeeded on the unit, and that failure cascaded into killing BLE. Decoupling BLE from SD init is the fix that closes the class regardless of the underlying SD root cause.

## Change

`omi/firmware/devkit/src/main.c` — SD mount failure now logs and continues instead of `return err;`; storage init runs only when the mount succeeded.

## Verified

- Static review: control flow now falls through to `transport_start()` on SD failure; `storage_init()` (which spawns the `storage_write` thread against the mounted card) runs only inside the mount-success branch, so an unmounted card no longer feeds that thread. Brace balance and the `#ifdef CONFIG_OMI_ENABLE_OFFLINE_STORAGE` block are intact.
- **UNVERIFIED on hardware:** I do not have a DevKit v2 to flash, and the devkit firmware is not built by any existing CI lane (only CV1 `omi/firmware/omi` is). A full NCS 2.9.0 Docker + `west update` build of the devkit target was out of scope for this focused fix and could not exercise the real BLE-advertisement path without the physical board anyway. No firmware boot-path test harness exists in `omi/firmware/devkit`, so no regression test was added.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9920?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/omi/firmware/devkit/src/main.c
+++ b/omi/firmware/devkit/src/main.c
@@ -227,17 +227,19 @@ int main(void)
 
     err = mount_sd_card();
     if (err) {
-        LOG_ERR("Failed to mount SD card (err %d)", err);
-        return err;
-    }
-    k_msleep(500);
+        // Offline storage is optional; a failed SD mount (bad/missing/incompatible card
+        // or wiring fault) must not abort boot and block BLE/transport entirely.
+        LOG_ERR("Failed to mount SD card (err %d), continuing without offline storage", err);
+    } else {
+        k_msleep(500);
 
-    LOG_PRINTK("\n");
-    LOG_INF("Initializing storage...\n");
+        LOG_PRINTK("\n");
+        LOG_INF("Initializing storage...\n");
 
-    err = storage_init();
-    if (err) {
-        LOG_ERR("Failed to initialize storage (err %d)", err);
+        err = storage_init();
+        if (err) {
+            LOG_ERR("Failed to initialize storage (err %d)", err);
+        }
     }
 #endif
 


kodjima33:

PR #9924, opened 2026-07-17
fix(llm-gateway): gate promotion and bound fallback

Git-on-my-level:
## Summary

- Bound chat streaming and gateway-first transport failures so a dead optional hop falls back without repeated 120-second waits.
- Keep production gateway serving **off** by default; derive any future promoted Cloud Run gateway endpoint from the ready ingress/ILB instead of a configured IP.
- Gate every gateway-first backend deploy on ready workload/service/endpoints, the attached internal forwarding rule, and an authenticated real-route probe from the serving Cloud Run VPC path.
- Add per-process gateway circuit breaking, client first-byte telemetry, and deployed alerts for fallback/circuit/zero-success/latency/endpoint failures.
- Offload file-chat setup and synchronous assistant streaming work from the event loop, so the SSE first-event deadline and terminal error frame can run even when setup stalls.

## Verification

- Focused gateway/SSE/file-chat suite after rebase: **49 passed**.
- Earlier focused project-file selection: 128 tests passed.
- Async-blocker scan: 0 selected findings.
- YAML/JSON parse validation, Python compilation, shell syntax validation, and `git diff --check`.
- Live GKE/ILB/Cloud Run VPC probe intentionally runs as the deploy gate; it was not run from a local workstation.

## Failure class

Failure-Class: FC-per-hop-timeout

### DIFF
--- a/.github/workflows/gcp_backend.yml
+++ b/.github/workflows/gcp_backend.yml
@@ -284,23 +284,33 @@ jobs:
             echo "sync_tasks_invoker_sa=${SYNC_TASKS_INVOKER_SA}"
           } >> "$GITHUB_OUTPUT"
 
-      - name: Render backend runtime env
-        id: runtime-env
-        env:
-          ACCOUNT_DELETION_HANDLER_URL: ${{ steps.account-deletion-target.outputs.account_deletion_handler_url }}
-          CLOUD_RUN_VPC_NETWORK: ${{ vars.CLOUD_RUN_VPC_NETWORK }}
-          CLOUD_RUN_VPC_SUBNET: ${{ vars.CLOUD_RUN_VPC_SUBNET }}
-          GOOGLE_CLIENT_ID: ${{ vars.GOOGLE_CLIENT_ID }}
-          MCP_OAUTH_CLAUDE_CLIENT_ID: ${{ vars.MCP_OAUTH_CLAUDE_CLIENT_ID }}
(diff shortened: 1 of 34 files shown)


Git-on-my-level:
Omi monitor run — no code changes this pass; one trusted review thread was already addressed and is now resolved.

Codex thread on `backend/utils/retrieval/graph.py` ("Offload non-image file streams before relying on timeouts", P2): the `fix(chat): offload file-chat stream setup` commit (f750a91) already moves the entire synchronous non-vision Assistants setup + stream (`_ensure_thread_and_assistant` + `ask_stream`) into the LLM executor via `run_blocking(llm_executor, self._ensure_and_ask_stream, ...)`, so the event loop is free to enforce the first-event and total-stream bounds from `graph.py`. This postdates the comment, so the thread is resolved as addressed. The offload is covered by `tests/unit/test_chat_async_offload.py::test_file_assistants_stream_runs_setup_and_sync_callbacks_off_loop` and `test_file_stream_deadline_fires_while_sync_assistants_stream_is_off_loop`.

Verification (current HEAD f750a91):
- `Backend unit suite`, `Hermetic Backend E2E`, `Static Contracts`, `Android Compile Smoke`, `Flutter Tests`, `Dart Analyze (ratchet)`, `Generated Files`: SUCCESS
- `Hygiene`/`Formatting` CANCELLED entries are the superseded older Repo Checks run; the newer run of each is SUCCESS/SKIPPED (concurrency cancellation only).

No unresolved threads remain.

PR #9927, opened 2026-07-17
fix cached conversations crash on missing plugin_id

krushnarout:
## Crash

Crashlytics [cc235c2f374b36f10489f755620ab120](https://console.firebase.google.com/u/0/project/based-hardware/crashlytics/app/ios:com.friend-app-with-wearable.ios12/issues/cc235c2f374b36f10489f755620ab120): `FlutterError - FormatException: Missing required field: plugin_id`, thrown from `GeneratedPluginResult.fromJson` while `SharedPreferencesUtil.cachedConversations` parses the conversation cache during `ConversationProvider.getInitialConversations` — a fatal crash on app startup.

## Root cause

`ServerConversation.toJson()` (the cache write path) serialized `plugins_results` entries via `AppResponse.toJson()`, which emits `{'appId', 'content'}`. The wire parser `GeneratedPluginResult.fromJson` requires the `plugin_id` key to be present (per the OpenAPI spec, `PluginResult.plugin_id` is required-but-nullable). So any user with at least one app result in a cached conversation crashed on the next launch. Separately, the cache never wrote `apps_results`, which is the key `ServerConversation.fromGenerated` actually reads app results from — so cached app results were silently dropped even when parsing succeeded.

## Fix

- `ServerConversation.toJson()` now writes wire-format `plugins_results` (`{'plugin_id', 'content'}`) and also `apps_results`, so the cache round-trips cleanly through the generated wire parser.
- `ServerConversation.fromJson()` normalizes legacy cache entries that lack the `plugin_id` key (recovering the id from `appId`/`app_id`), so caches already corrupted in the wild parse instead of crashing after the update.
- `ServerConversation.fromGenerated()` falls back to `plugins_results` when `apps_results` is empty, recovering app results from legacy caches.

## Verification

- New regression test `app/test/unit/conversation_cache_roundtrip_test.dart` (3 tests): toJson→fromJson round trip preserves app results; a legacy cache entry with `appId`-keyed `plugins_results` (the exact Crashlytics shape) parses and recovers the app result — this test throws the reported `FormatException` before the fix; null `plugin_id` round trips.
- `bash app/test.sh`: all 865 tests pass.
- `dart analyze` on the changed files: no errors or warnings (only pre-existing info-level lints).

none

Failure-Class: FC-malformed-doc-read

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9927?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/backend/schema/conversation.dart
+++ b/app/lib/backend/schema/conversation.dart
@@ -366,6 +366,17 @@ class ServerConversation {
     if (structured != null) {
       normalized['structured'] = structured.toGenerated().toJson();
     }
+    // Legacy caches (< toJson wire-format fix) wrote plugins_results entries as
+    // {'appId', 'content'}; the wire parser requires the plugin_id key.
+    final rawPluginResults = normalized['plugins_results'];
+    if (rawPluginResults is List) {
+      normalized['plugins_results'] = rawPluginResults.map((entry) {
+        if (entry is Map<String, dynamic> && !entry.containsKey('plugin_id')) {
+          return {...entry, 'plugin_id': entry['appId'] ?? entry['app_id']};
+        }
+        return entry;
+      }).toList();
+    }
     final generated = wire.GeneratedConversation.fromJson(normalized);
     return ServerConversation.fromGenerated(
       generated,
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
mobile crash fix: guard cached conversations against missing plugin_id (FormatException on startup) with regression test — approve only (Formatting/Hygiene CI currently red)

PR #9954, opened 2026-07-18
feat: add auto-save for SD card every 2s in sync mode

TuEmb:
## Problem

Today the device only advances the SD ring **read pointer after the entire sync
finishes** (after `NOTIFY_DONE` — the app advances only on completion, by design,
for data safety). Two consequences:

- **Nothing is freed until the very end.** The on-device ring can hold hundreds of
  MB, and a full sync takes hours. Free space on the device stays flat the whole
  time and only jumps at the end.
- **A mid-sync disconnect loses all progress.** BLE drops are common (weak signal,
  app backgrounded, etc.). Since `read_seq` wasn't advanced, on reconnect the app
  re-reads from the original pointer and **the whole sync restarts from the
  beginning** — potentially re-sending hundreds of MB.

## Change

Advance + persist the ring `read_seq` **incrementally, as data is confirmed
delivered to the phone** (counted in the BLE TX-completion callback), checkpointed
every ~2 s and once more on `DONE` / disconnect.

- Only data the phone has actually received (TX-confirmed at the link layer) is
  ever freed → **safe, no over-advance / data loss**.
- On reconnect the app reads from the now-advanced `read_seq`, so a dropped sync
  **resumes near where it stopped instead of restarting from scratch**.
- Free space grows **live** during the sync.

Compatible with the current app: it still reads from `read_seq` and still sends
`CMD_RING_ADVANCE` after `DONE` (now a no-op since we've already advanced).

## Implementation

- `storage.c` — accumulate TX-confirmed bytes; `sync_checkpoint_advance()` advances
  `read_seq` to the confirmed seq (throttled 2 s, forced on DONE/disconnect) and
  updates the cached used/free/unread counters without touching the SD.
- `sd_card.c` / `sd_card.h` — `sd_ring_advance_async()`: fire-and-forget ring advance
  on the SD worker.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9954?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/omi/firmware/omi/src/lib/core/sd_card.h
+++ b/omi/firmware/omi/src/lib/core/sd_card.h
@@ -43,6 +43,7 @@ uint32_t write_to_file(uint8_t *data, uint32_t length);
 int sd_ring_get_info(sd_ring_info_t *info);
 int sd_ring_read(uint64_t start_seq, uint8_t *buf, uint32_t max_bytes, uint32_t *bytes_read, uint32_t *packets_read);
 int sd_ring_advance(uint64_t new_read_seq);
+int sd_ring_advance_async(uint64_t new_read_seq);
 int sd_ring_clear(void);
 
 uint32_t get_file_size(void);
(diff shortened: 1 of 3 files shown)


TuEmb:
**Connect to the phone app and leave it for a while:**
image


**Disconnect and connect to the device again:**
image

mdmohsin7 (approved):

PR #9955, opened 2026-07-18
Allow completed conversations to enter merging so conversation merge works

ZachL111:
Allow completed conversations to enter merging so conversation merge works

POST /v1/conversations/merge has two gates that disagree, so every merge request fails.

1. routers/conversations.py:1279 calls validate_merge_compatibility, which rejects the
   request with a 400 unless EVERY conversation has status == 'completed'
   (merge_conversations.py:124-128).
2. routers/conversations.py:1285 then calls lifecycle_service.begin_merge on each
   conversation, which is transition(..., merging) with expected=None. _STATUS_TRANSITIONS
   has no 'completed' key, so _STATUS_TRANSITIONS.get('completed', set()) is empty and
   transition raises LifecycleTransitionError at lifecycle.py:164.

Validation admits only 'completed'; the table allowed entering 'merging' only from
'in_progress'. Those are mutually exclusive, so a merge that passes validation always
raises. LifecycleTransitionError subclasses ValueError and main.py registers no handler
for it, so it surfaces as an unhandled 500. begin_merge has exactly one production caller,
that endpoint.

This regressed in 61c9b0488b ("refactor(backend): centralize conversation lifecycle
writes"), which introduced _STATUS_TRANSITIONS and replaced the previous unconditional
write (conversations_db.update_conversation_status(..., ConversationStatus.merging)) with
begin_merge. The new table never got a 'completed' row.

That completed -> merging is the intended edge is already documented in the codebase: the
failure rollback in merge_conversations.py says "Since source conversations were set to
'merging' status, we need to reset them back to 'completed'", and the reverse edge
merging -> completed is already declared.

Fix: declare the edge.

    ConversationStatus.completed.value: {ConversationStatus.merging.value},

Tests (tests/unit/test_conversation_lifecycle_contract.py):
- Adds test_merge_admission_and_lifecycle_agree_on_completed, asserting the two gates
  agree: validate_merge_compatibility accepts completed conversations, and begin_merge
  then moves one to merging.
- Retargets the negative assertion in test_lifecycle_service_allows_only_declared_transitions.
  It previously asserted that a completed conversation could NOT begin merge, which pinned
  this bug. It now asserts on processing -> merging, which is genuinely undeclared, so the
  test keeps its original purpose.

- pytest tests/unit/test_conversation_lifecycle_contract.py: 14 passed
- prove-fail: with lifecycle.py reverted to main, the new test fails with
  "LifecycleTransitionError: invalid lifecycle transition completed->merging" at lifecycle.py:164
- pytest test_merge_validation.py (38 passed), test_merge_conversations_canonical_delete.py
  (3 passed), test_check_conversation_lifecycle_writes.py (6 passed)
- pyright utils/conversations/lifecycle.py: 0 errors

I did not exercise the endpoint against a live backend. The contract test drives
begin_merge through the same lifecycle seam the endpoint calls.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_conversation_lifecycle_contract.py
+++ b/backend/tests/unit/test_conversation_lifecycle_contract.py
@@ -115,10 +115,32 @@ def test_lifecycle_service_allows_only_declared_transitions(lifecycle_store):
     )
 
     assert lifecycle_service.admit_processing('uid', 'conversation') is True
-    assert lifecycle_service.complete('uid', 'conversation') is True
-    assert lifecycle_store.conversation('uid', 'conversation')['status'] == ConversationStatus.completed
+    # processing -> merging stays undeclared: merge only ever admits completed conversations.
     with pytest.raises(lifecycle_service.LifecycleTransitionError, match='invalid lifecycle transition'):
         lifecycle_service.begin_merge('uid', 'conversation')
+    assert lifecycle_service.complete('uid', 'conversation') is True
+    assert lifecycle_store.conversation('uid', 'conversation')['status'] == ConversationStatus.completed
+
+
+def test_merge_admission_and_lifecycle_agree_on_completed(lifecycle_store):
+    """The two gates in POST /v1/conversations/merge must agree on the admitted status.
+
+    validate_merge_compatibility rejects every status except completed, so completed is the only
+    status that can reach begin_merge. If the transition table omits completed -> merging, every
+    accepted merge raises LifecycleTransitionError, which is an unhandled 500 and makes the merge
+    feature unusable. merge_conversations' failure rollback documents the same edge in reverse.
+    """
+    conversations = [
+        {'id': 'conversation', 'status': ConversationStatus.completed.value},
+        {'id': 'other', 'status': ConversationStatus.completed.value},
+    ]
+    is_valid, error_message, _ = validate_merge_compatibility(conversations)
+    assert (is_valid, error_message) == (True, None)
+
+    lifecycle_store.put_conversation('uid', 'conversation', status=ConversationStatus.completed.value, discarded=False)
+
+    assert lifecycle_service.begin_merge('uid', 'conversation') is True
+    assert lifecycle_store.conversation('uid', 'conversation')['status'] == ConversationStatus.merging
 
 
 def test_generic_lifecycle_field_write_fails_closed(lifecycle_store):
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Verified your diagnosis end-to-end against main (base SHA 97c35d54) and it holds up cleanly:

- `_STATUS_TRANSITIONS` in `backend/utils/conversations/lifecycle.py` has no `completed` source key, so `begin_merge` -> `transition(..., ConversationStatus.merging)` (expected=None) raises `LifecycleTransitionError` on any completed conversation.
- The router calls `validate_merge_compatibility` (rejects unless every conversation is `completed`) and then `lifecycle_service.begin_merge` on each — those two gates are mutually exclusive, so every merge that passes validation raises. `main.py` has no handler for `LifecycleTransitionError`, so it surfaces as an unhandled 500. Confirmed `begin_merge` has exactly one production caller (this endpoint).
- The intended edge is already documented in reverse: `_handle_merge_failure` resets statuses back to `completed`, and `merging -> completed` is already declared. Adding `completed -> {merging}` closes the loop.

The fix is the minimal correct one and doesn't widen any other transition path. The new contract test asserting the two gates agree on `completed`, plus the relocated `processing -> merging` assertion, are appropriate coverage.

One note for the maintainer before merge: this head's CI summary lists Formatting/Hygiene/Detect Changes as failed, but when I read the check-runs for this commit directly I see only skipped/cancelled path-filtered duplicates and passing backend suites (Backend unit suite, Hermetic Backend E2E, Listen Pusher Stack Gauntlet, Public Developer API contract all green). Worth confirming the required-status context is actually green on your end. Nice catch on the regression and the write-up.

(Labeling bug + backend; not formally approving yet pending maintainer confirmation of CI state.)

---

kodjima33 (approved):
backend bug fix (conf 5/5): add missing completed->merging transition so /v1/conversations/merge stops 500ing; test

PR #9960, opened 2026-07-18
Finalize pending conversations when a listen session ends inside the retry window

ZachL111:
Finalize pending conversations when a listen session ends inside the retry window

LiveConversationController.process_pending defers finalization by 7 seconds, then finalizes
the timed-out conversation and re-dispatches anything still stuck in `processing`. It
currently returns early when that wait is interrupted:

    async def process_pending(self, timed_out_id: Optional[str]) -> None:
        if await self.host.wait(7):
            return

host.wait is wait_for_event(self.state.shutdown_event, seconds) (routers/listen/runtime.py),
which returns True when woken early by shutdown rather than when the sleep elapses
(utils/async_tasks.py). runtime.py then sets shutdown_event and drains background tasks
without cancelling them:

    self.state.shutdown_event.set()
    await self.task_supervisor.drain_monitored(timeout=self.limits.bg_drain_timeout, cancel=False)

So a listen session that ends within that 7 second window wakes process_pending immediately
with True and it returns, skipping both the timed-out conversation's finalization and the
re-dispatch of conversations stuck in `processing`. The drain deliberately lets background
work finish, so this work should run rather than be dropped.

Before the listen split this was an unconditional sleep (routers/transcribe.py at c7ca348b04^):

    async def process_pending_conversations(timed_out_id: Optional[str]):
        await asyncio.sleep(7.0)
        if timed_out_id:
            await _process_conversation(timed_out_id)
        await cleanup_processing_conversations()

The `if await wait(...): return` shape is the polling-loop idiom, and it is correct in
lifecycle_loop, which uses `if await self.host.wait(5): break` to leave a loop.
process_pending is a one-shot deferred action, so an early wake should shorten the wait rather
than cancel the work.

Fix: keep the wait interruptible, drop the early return.

    await self.host.wait(7)

Tests (tests/unit/test_listen_process_pending_shutdown.py) subclass LiveConversationController
to record the two finalization calls and drive the real process_pending. The controller takes
only a host, so no patching is required:
- a session woken by shutdown still finalizes the timed-out conversation and re-dispatches processing
- a normal full-delay session behaves exactly as before
- with no timed-out conversation, processing is still re-dispatched

- pytest tests/unit/test_listen_process_pending_shutdown.py: 3 passed
- prove-fail: with the source reverted to main, the two shutdown-path tests fail
  (assert [] == ['conv-timed-out'] and assert [] == ['conv-a', 'conv-b']) while the
  normal-delay test still passes, so the change only affects the interrupted path
- pytest test_conversation_lifecycle_contract.py (13 passed), test_conversation_finalization_jobs.py
  (22 passed), test_listen_session_bootstrap.py (4 passed)
- pyright routers/listen/conversations.py: 0 errors

Note: test_listen_persistence.py and test_listen_runtime_regressions.py do not import in my
local environment (deepgram SDK version mismatch, ImportError on DeepgramClientOptions). They
fail identically with this change reverted, so that is a local environment gap rather than a
regression from this PR; CI runs them with the pinned SDK.

I did not exercise a live WebSocket session. The test drives the real process_pending through
the same call the runtime makes.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9960?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/listen/conversations.py
+++ b/backend/routers/listen/conversations.py
@@ -279,8 +279,11 @@ class LiveConversationController:
         return None
 
     async def process_pending(self, timed_out_id: Optional[str]) -> None:
-        if await self.host.wait(7):
-            return
+        # Interruptible delay, not a polling loop. An early wake means the session is shutting
+        # down, which is precisely when the timed-out conversation and anything still stuck in
+        # `processing` must be finalized, so the wake shortens the wait instead of skipping the
+        # work. Returning here dropped both (pre-split this was an unconditional sleep).
+        await self.host.wait(7)
         if timed_out_id:
             await self.process_conversation(timed_out_id)
         processing = await self.host.persistence.call(
--- a/backend/tests/unit/test_listen_process_pending_shutdown.py
+++ b/backend/tests/unit/test_listen_process_pending_shutdown.py
@@ -1,0 +1,88 @@
+"""Regression: a listen session ending inside the 7s window must still finalize pending work.
+
+LiveConversationController.process_pending defers finalization by 7 seconds, then finalizes the
+timed-out conversation and re-dispatches anything still stuck in `processing`. The listen split
+turned the original unconditional `await asyncio.sleep(7.0)` into `if await self.host.wait(7):
+return`. host.wait is wait_for_event(shutdown_event, seconds), which returns True when woken
+early by shutdown, and runtime sets shutdown_event immediately before draining background tasks
+without cancelling them. So any session ending inside that window returned early and skipped both
+the timed-out conversation's finalization and the processing re-dispatch.
+
+The `if ...: return` form is the polling-loop idiom (lifecycle_loop correctly uses
+`if await self.host.wait(5): break`). process_pending is a one-shot deferred action, so an early
+wake must shorten the wait, not cancel the work.
+
+Seam: the controller takes only a host, so this subclasses it to record the two finalization
+calls and drives the real process_pending. No patching and no sys.modules mutation.
+"""
+
+from types import SimpleNamespace
+
+from routers.listen.conversations import LiveConversationController
+
+
+class _Host:
+    """Minimal listen host. wait() returns True to mean 'woken early by shutdown'."""
+
+    def __init__(self, *, woken_by_shutdown: bool, processing: list[dict[str, str]]) -> None:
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Verified and this looks like a correct, well-scoped fix to a real regression.

What I checked against main:
- `host.wait` is `wait_for_event(shutdown_event, seconds)` and returns `True` on early wake (utils/async_tasks.py) — so the current `if await self.host.wait(7): return` really does drop the work when the session ends inside the window.
- runtime.py sets `shutdown_event` then calls `drain_monitored(..., cancel=False)` — i.e. background work is expected to finish during drain, which is exactly what this PR re-enables.
- `process_pending` is launched as a one-shot finite task (`pending_convos`), so the early-return is indeed wrong here; the `if ...: break` form belongs to `lifecycle_loop`, not this path.
- The pre-split unconditional `asyncio.sleep(7.0)` claim is consistent with the behavior you're restoring.

The test design is clean — subclassing the real controller (host-only constructor) to record `process_conversation` / `schedule_finalization`, no monkey-patching. I reproduced the three cases independently; reverting only the interrupted path makes the two shutdown cases fail exactly as you described in the prove-fail note, and the normal-delay case is unchanged.

One heads-up so it doesn't look alarming in the checks: Hygiene, Formatting, and Detect Changes show up as cancelled/skipped on this head — the canonical run (29638398196) has them as skipped (no relevant files changed), and an earlier superseded run had them cancelled. The substantive checks — Backend Unit Tests, Backend Checks, Hermetic Backend E2E, Public Developer API contract — all pass.

Recommend merge. Leaving for a human maintainer for the final call since this touches the listen shutdown/finalization path.

(This feedback is from an automated maintainer assistant; please use human judgment before merging.)

---

kodjima33 (approved):
backend bug fix (conf 5/5): finalize timed-out conversation when listen session ends inside the 7s window; test

PR #9962, opened 2026-07-18
Contain translation persist failures so a batch does not drop its remaining segments

ZachL111:
Contain translation persist failures so a batch does not drop its remaining segments

TranscriptProcessor._on_translation_ready is the callback TranslationCoordinator invokes for
each finished translation. Before the listen split the whole body was wrapped in a broad
guard (routers/transcribe.py at e8adfc5623^):

    try:
        ...
    except Exception as e:
        logger.error(f"Translation persist error: {e} {uid} {session_id}")

The split (c7ca348b04, "refactor(quality): split oversized product sources") recreated the
callback in routers/listen/transcripts.py without that guard. The caller catches only two
types (utils/translation_coordinator.py):

    except (RuntimeError, ValueError) as error:
        logger.error('TranslationCoordinator batch error: %s', error)

So a transient persist failure, for example a Firestore or network error raised by
update_conversation_segments, escapes _on_translation_ready, propagates out of the batch loop
in _flush_batch, and silently drops the translations for every remaining segment in that
batch. The coordinator invokes the callback from a bare task, so nothing surfaces except an
unretrieved-task warning. For the final batch flushed at session teardown those translations
are lost for good.

Fix: restore the guard around the persist body so the failure is logged and contained. The
diff is mostly the re-indentation that adding the try block requires.

Every other callback path created by the same split kept its broad guard (speakers.py match
and load_and_run, receiver.py _process_photo, runtime.py _refresh_credits and _teardown,
transcripts.py trigger_realtime_integrations), so this one is the outlier.

asyncio.CancelledError derives from BaseException, so the broad except does not interfere with
the shield-based cancellation in _flush_after_batch_window.

Tests (tests/unit/test_translation_persist_guard.py) build a fake host by constructor
injection and drive the real callback. TranscriptProcessor takes only a host, so no patching
is required:
- a persist error raised by update_conversation_segments is contained rather than escaping
- a successful persist still writes the translation with the expected lang and text

- pytest tests/unit/test_translation_persist_guard.py: 2 passed
- prove-fail: with transcripts.py reverted to main, the containment test fails with
  "ConnectionError: firestore unavailable" escaping the callback, while the successful-persist
  test still passes, so the change only affects the failure path
- pytest test_fast_unit_duration_guard.py (3 passed), test_transcribe_conversation_cache.py
  (3 passed), test_conversation_lifecycle_contract.py (13 passed)
- pyright routers/listen/transcripts.py: 0 errors

I did not exercise a live translated session. The test drives the real callback through the
same call the coordinator makes.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9962?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/listen/transcripts.py
+++ b/backend/routers/listen/transcripts.py
@@ -113,43 +113,58 @@ class TranscriptProcessor:
             self.translation_coordinator and self.translation_coordinator._flushing  # type: ignore[reportPrivateUsage]
         ):
             return
-        async with self.translation_lock:
-            conversation = (
-                await self.cache.get(conversation_id)
-                if conversation_id == self.host.state.current_conversation_id
-                else await self._load_conversation(conversation_id)
+        # TranslationCoordinator invokes this callback from a bare task and only catches
+        # (RuntimeError, ValueError), so a persist failure escaping here aborts the batch loop and
+        # silently drops the translations for every remaining segment. Keep the failure contained.
+        try:
+            async with self.translation_lock:
+                conversation = (
+                    await self.cache.get(conversation_id)
+                    if conversation_id == self.host.state.current_conversation_id
+                    else await self._load_conversation(conversation_id)
+                )
+                if not conversation:
+                    return
+                for index, segment in enumerate(conversation.get('transcript_segments', [])):
+                    if segment['id'] != segment_id:
+                        continue
+                    translations = segment.get('translations', [])
+                    translation = Translation(lang=self.host.translation_language, text=translated_text).model_dump()
+                    replacement = next(
+                        (
+                            i
+                            for i, value in enumerate(translations)
+                            if value.get('lang') == self.host.translation_language
+                        ),
+                        None,
+                    )
+                    if replacement is None:
+                        translations.append(translation)
+                    else:
+                        translations[replacement] = translation
+                    conversation['transcript_segments'][index]['translations'] = translations
+                    await self.host.persistence.call(
+                        conversations_db.update_conversation_segments,
+                        self.host.request.uid,
+                        conversation_id,
+                        conversation['transcript_segments'],
+                        data_protection_level=(
+                            self.cache.protection_level
+                            if conversation_id == self.host.state.current_conversation_id
+                            else None
+                        ),
+                    )
+                    if conversation_id == self.host.state.current_conversation_id:
+                        self.cache.update_segments(conversation['transcript_segments'])
+                        self.host.send_event(TranslationEvent(segments=[conversation['transcript_segments'][index]]))
+                    return
+        except Exception as error:
+            logger.error(
+                'Translation persist failed segment=%s uid=%s type=%s',
+                segment_id,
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified the regression and the fix against `main` — this is a clean, well-scoped fix. Thank you @ZachL111.

What I checked:
- Pre-split `routers/transcribe.py` (at `e8adfc5623^` / `e9491d8b`) wrapped this exact body in `try/except Exception`, so this restores error containment that the listen split (`c7ca348b04`) accidentally dropped.
- The caller `TranslationCoordinator._flush_batch` only catches `(RuntimeError, ValueError)`, so any other error escaping `_on_translation_ready` propagates out of the batch loop and silently drops the translations for every remaining segment in that batch (and for the final teardown batch, permanently). The coordinator runs the callback from a bare task, so it only surfaces as an unretrieved-task warning. The bug and blast radius are exactly as described.
- Every sibling callback created by the same split (`speakers.load_and_run`/`match`, `receiver._process_photo`, `runtime._refresh_credits`/`_teardown`, `transcripts.trigger_realtime_integrations`) keeps its broad guard, so this was genuinely the outlier.
- `except Exception` correctly excludes `BaseException`, so `asyncio.CancelledError` still propagates and the shield-based cancellation in `_flush_after_batch_window` is unaffected. Good catch calling that out.

The regression test is solid: constructor-injected fake host (no monkeypatching / no `sys.modules` mutation), forces the persistence load path, and uses `ConnectionError` — precisely the class of transient Firestore/network error that would escape `(RuntimeError, ValueError)`. The positive-path test confirms the successful write still happens.

One minor, non-blocking observation for whoever lands this: the handler logs `segment_id`, `uid`, and `type(error).__name__` but not the error message or traceback. The original guard logged `{e}`. Including the exception (and ideally the conversation id) would make these failures easier to diagnose in logs, since the whole point is that they are now silent-by-design. Not worth a re-spin on its own.

Checks: the substantive backend suite is green on this head (Backend unit suite, Hermetic Backend E2E, Listen Pusher Stack Gauntlet, Public Developer API contract, Detect Changes all `success`). Note the PR shows Formatting/Hygiene as cancelled — those are from a superseded workflow run; the latest run correctly skips them (no relevant paths for a backend-only change), so there's nothing to fix there.

This touches user-data persistence, so I'm deferring formal approval to a human maintainer — but the change itself is correct, low-risk, and well-tested, and I'd expect it to land quickly once a maintainer signs off.

---

kodjima33 (approved):
backend bug fix (conf 5/5): guard _on_translation_ready persist so a transient failure doesn't drop remaining batch segments; test

PR #9966, opened 2026-07-18
Make utils.subscription importable on its own by breaking its cycle with database.users

ZachL111:
Make utils.subscription importable on its own by breaking its cycle with database.users

utils/subscription.py cannot be imported on its own. In a fresh interpreter:

    python -c "import utils.subscription"
    ImportError: cannot import name 'get_default_basic_subscription' from partially
    initialized module 'utils.subscription' (most likely due to a circular import)

The cycle: utils/subscription.py imports database.users, and database/users.py imported
get_default_basic_subscription back from utils.subscription at module scope. Whichever of the
pair is imported first decides whether it works. Importing database.users first happens to
succeed, which is why this is intermittent rather than constant.

It is not theoretical. Earlier today on main it broke the Backend unit suite:

    tests/unit/test_dg_usage_batch.py             -> 4 errors
    tests/unit/test_listen_runtime_regressions.py -> failed
    ImportError: cannot import name 'get_plan_limits' from 'utils.subscription'
    (unknown location)   at utils/fair_use.py:21

That surfaced once utils/fair_use.py began importing utils.subscription, which changed which of
the pair loaded first for those selections.

To be accurate about the current state: those two files pass on main right now, because the
import order shifted back. The defect itself is unchanged, and it can resurface whenever an
unrelated change alters import order again.

The database -> utils edge is also the reverse of the documented backend layering
(backend/AGENTS.md: "database/ -> utils/ -> routers/ -> main.py. Higher imports from lower,
never reverse").

Fix: move the single import to its two call sites in database/users.py, so database no longer
depends on utils at module scope.

On the deviation from "All imports at module top level, never inside functions": honouring that
rule here is precisely what creates the reverse-hierarchy dependency the layering rule forbids.
A deferred import is the standard way to break such a cycle, and the comment at each call site
records why it must stay deferred so it is not folded back to the top and silently re-broken.

I did consider relocating get_default_basic_subscription instead. It is a two-line model
factory, but it is built from env-configured plan constants (BASIC_TIER_* and
FREE_CHAT_QUESTIONS_PER_MONTH, read via os.getenv), so it belongs in utils rather than models.
Moving it would push env-derived configuration into the lowest layer, and would additionally
churn routers/users.py plus the test files that stub
utils.subscription.get_default_basic_subscription, while an alias re-export is ruled out by
"No in-repo compatibility layers". Happy to take that route instead if you would prefer it.

Tests (tests/unit/test_subscription_import_cycle.py) import each side of the pair first in a
fresh interpreter. A subprocess is the only honest seam here: once pytest has imported either
module the order is fixed for the whole session, so an in-process assertion would pass either
way.

- pytest tests/unit/test_subscription_import_cycle.py: 2 passed
- prove-fail: with database/users.py reverted to main, test_utils_subscription_imports_standalone
  fails with the circular-import ImportError above, while the database.users side still passes
- pytest test_dg_usage_batch.py (11 passed), test_listen_runtime_regressions.py (4 passed),
  test_users_missing_doc_guards.py (18 passed), test_claim_deletion_wipe_txn.py (16 passed),
  test_desktop_migration.py (106 passed), test_voice_message_language.py (11 passed),
  test_stripe_webhook_behavioral.py (22 passed)
- pyright database/users.py: 0 errors
- product file line-count ratchet: baseline raised 1943 to 1950 with a justification

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9966?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
backend refactor (break subscription import cycle) — approve-only

Git-on-my-level:
Thanks for the focused fix here. The circular-import diagnosis matches the diff: `utils.subscription` imports `database.users`, and moving `get_default_basic_subscription` out of `database.users` module scope removes the reverse module-level edge while keeping both fallback call sites behaviorally unchanged. The fresh-interpreter regression test is also the right shape for this bug, since an in-process pytest import would hide the order dependency.

I’m leaving this as a positive signal rather than a formal approval because this intentionally makes a narrow exception to the backend “imports at module top level” rule, and the current head still has a failing Backend unit suite check. From the logs, the failing files are memory/vector tests outside this PR’s touched area, while the new import-cycle test itself is syntactically valid and targeted. A human maintainer should still confirm the style-rule exception is acceptable and that CI is green/not blocked by unrelated failures before merge.

---

kodjima33 (approved):
Backend feature (re-approved on new commits) — approve-only per policy; Nik owns backend deploy.

Git-on-my-level:
Re-reviewed the unchanged head now that CI is green and a human maintainer has approved it. This still looks like a focused, low-risk backend cleanup from the automation side.

File-specific notes:
- `.github/scripts/product_file_line_count_ratchet_baseline/backend-database.json` only updates the recorded `backend/database/users.py` line count and adds a targeted justification for the local import exception; I don’t see workflow/CI behavior changing here.
- `backend/database/users.py` removes the module-level `utils.subscription` import and defers `get_default_basic_subscription` at the two fallback call sites, which breaks the import cycle without changing the returned default subscription behavior.
- `backend/tests/unit/test_subscription_import_cycle.py` uses fresh Python subprocesses for both `utils.subscription` and `database.users`, which is the right regression seam because an in-process pytest import could mask the order-dependent failure.

I removed the stale `needs-maintainer-review` / `workflow-review` labels because the workflow flag was only from the ratchet baseline file, not a CI/workflow behavior change, and the current head now has passing backend/contract checks plus human maintainer approval. Leaving this as a positive signal rather than a formal bot approval.

---
*by AI on behalf of David — re-reviewed because CI and maintainer-review state changed; final merge decision remains with maintainers.*

ZachL111:
Closing in favor of #11621, which rebuilds this fix against current main. The cycle picked up a third call site since this branch was cut (the provision=False branch in get_user_valid_subscription), and the new PR covers it.

PR #9969, opened 2026-07-18
Honour include_discarded when hydrating integration conversation search results

ZachL111:
Honour include_discarded when hydrating integration conversation search results

POST /v2/integrations/{app_id}/search/conversations forwards include_discarded into the
search, then hydrates the matched ids without it:

    search_results = search_conversations(
        ...
        include_discarded=cast(bool, search_request.include_discarded),
        ...
    )
    conversation_ids = [conv.get('id') for conv in search_results['items']]
    full_conversations = []
    if conversation_ids:
        full_conversations = conversations_db.get_conversations_by_id(uid, conversation_ids)

get_conversations_by_id defaults to include_discarded=False (database/conversations.py) and
skips discarded documents:

    if data.get('discarded') and not include_discarded:
        continue

SearchRequest.include_discarded defaults to True (models/conversation.py), so the two gates
disagree by default rather than only in an unusual configuration. A discarded conversation can
match the search, be counted in the page, and then be dropped during hydration. Because the
search window has already moved past it, that row is not reachable on any page.

The same defect existed on the desktop sibling and a maintainer fixed it the same way in #9175
(3bb17579ac, "Fix desktop search response contracts"), which added the forwarding to
routers/conversations.py:

    conversations = conversations_db.get_conversations_by_id_without_photos(
        uid,
        conversation_ids,
        include_discarded=bool(search_request.include_discarded),
    )

That is the identical search-then-hydrate flow. The integration endpoint was not updated at the
time, so this applies the same change there.

Fix: pass the flag the search already ran with.

Tests (tests/unit/test_integration_search_include_discarded.py) call the handler directly with
its collaborators patched via monkeypatch.setattr on module attributes, and assert the flag
reaches the hydration call in both directions:
- include_discarded=True is forwarded
- include_discarded=False is forwarded

- pytest tests/unit/test_integration_search_include_discarded.py: 2 passed
- prove-fail: with routers/integration.py reverted to main, both fail with "assert None is True"
  and "assert None is False", because the flag never reaches the call at all
- pytest tests/unit/test_lock_bypass_fixes.py (the existing suite that drives this handler):
  61 passed
- pyright routers/integration.py: 0 errors

One caveat I would rather state than gloss over: the Firestore to Typesense indexer is not in
this repository, so I could not prove from source that discarded conversations are present in
the search index. Two things indicate they are, and the change is correct either way.
backend/typesense/conversations.schema declares discarded as an indexed filterable bool, and
utils/conversations/search.py only adds "discarded:=false" when the flag is false. Both would be
dead weight if discarded documents were never indexed. The fix makes hydration agree with the
search call it accompanies, which holds regardless.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9969?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks @ZachL111 — this is a solid, well-scoped fix. I verified the defect against main and the referenced #9175, and the analysis holds:

- `search_conversations_via_integration` forwarded `include_discarded` into `search_conversations` but not into the hydration call `conversations_db.get_conversations_by_id`, which defaults to `include_discarded=False`.
- Since `SearchRequest.include_discarded` defaults to `True`, the two gates disagreed on the default path: a discarded conversation could match the search, be counted in the page, then get dropped during hydration with no way to reach it on any later page.
- The sibling desktop endpoint had the same defect and was fixed the same way in #9175; this brings the integration endpoint in line.

The fix is correct and the new tests (`test_hydration_requests_discarded_when_search_did` / `test_hydration_excludes_discarded_when_search_did`) cleanly assert the flag is forwarded in both directions.

Heads-up for a maintainer before merge: with this change, an integration app that searches with `include_discarded=True` (the request default) will now be able to hydrate discarded conversations through this endpoint. That is the right behavior for parity with the desktop path and with what the search already returned, but it is a user-data disclosure surface, so I am flagging it for an explicit maintainer ack rather than approving outright.

CI note: the `Backend unit suite` failure on this PR is pre-existing on `main` — the same four memory-adapter test files fail identically on the latest main run and are unrelated to this change; this PR's own test file passes inside the suite. `Formatting`/`Hygiene` were cancelled, not failed on this PR's content. Once a maintainer signs off on the disclosure direction and the suite is healthy, this looks good to go.

---

kodjima33 (approved):

PR #9970, opened 2026-07-18
Link app-posted chat messages to the app's chat session so they render

ZachL111:
Link app-posted chat messages to the app's chat session so they render

database.chat.add_app_message stores a message an app posted for the user, but never sets
chat_session_id and never calls add_message_to_chat_session:

    ai_message = Message(
        ...
        memories_id=[conversation_id] if conversation_id else [],
    )
    add_message(uid, ai_message.model_dump())
    return ai_message

get_messages filters by chat_session_id whenever a session exists:

    if chat_session_id:
        # Session-scoped query: filter by session only, skip plugin_id filter
        messages_ref = messages_ref.where(filter=FieldFilter('chat_session_id', '==', chat_session_id))
    else:
        # App-scoped query: filter by plugin_id (None = main chat)
        messages_ref = messages_ref.where(filter=FieldFilter('plugin_id', '==', app_id))

A Firestore equality filter never matches a null field, so once a session exists for that app the
message is stored but never returned on that path. GET /v2/messages resolves the session before
reading, so the user gets the push notification, taps it, and the app's chat opens without the
message.

The sibling ten lines below already does this, and its docstring states the requirement:

    def add_integration_chat_message(text: str, app_id: Optional[str], uid: str) -> Message:
        """Add a chat message from an external integration (e.g. notification API),
        linking it to the user's existing chat session so it appears in the chat feed."""
        chat_session = get_chat_session(uid, app_id=app_id)
        chat_session_id = chat_session['id'] if chat_session else None
        ...
        if chat_session_id:
            add_message_to_chat_session(uid, chat_session_id, ai_message.id)

add_app_message is the same operation missing the same three steps. Its callers are in
utils/app_integrations.py: conversation finalization, plus the realtime and mentor proactive paths.

Fix: resolve the session and link the message, mirroring the sibling.

Two questions this will raise, answered up front:

1. It does not touch existing messages. The session is resolved at write time and applied to the
   message being created, so already-stored messages are unchanged. Only newly posted app messages
   gain the link.
2. app_id=None behaves as it does in the sibling. get_chat_session takes app_id as a keyword and
   returns None when there is no session, in which case chat_session_id stays None and no link call
   is made, which matches today's behavior.

Honest caveat on scope: the loss only begins once a chat session exists for that app_id. A user
whose app chat accumulated messages before any session existed still sees those older ones. The
arming step is the ordinary first-open path, since GET /v2/messages creates a session when it
returns the initial greeting.

Tests (tests/unit/test_app_message_chat_session.py) patch the three collaborators with
monkeypatch.setattr and assert:
- with a session, the stored message carries chat_session_id and the link call is made
- with no session, nothing is linked and the message still stores (unchanged behavior)
- conversation_id is still recorded in memories_id

- pytest tests/unit/test_app_message_chat_session.py: 3 passed
- prove-fail: with database/chat.py reverted to main, the session test fails with
  "assert None == 'sess-1'" while the other two still pass, so the change only affects the
  session path
- pytest on the five suites that reference add_app_message: test_app_integrations_async (3),
  test_async_app_integrations (13), test_async_realtime_integrations_offload (6),
  test_mentor_notifications (45), test_realtime_integrations_usage_tracking (8), all passed
- pyright database/chat.py: 0 errors

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9970?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for the focused fix here. I reviewed the changed backend path and the added regression coverage; the implementation matches the existing `add_integration_chat_message` behavior by resolving the app chat session, storing `chat_session_id`, and linking the message into that session when one exists. The no-session and `conversation_id` cases are also covered, which is the right shape for this regression.

I’m not formally approving because this touches user chat-message persistence/session linkage and the required checks are not all green. The backend unit suite is currently failing in unrelated memory-adapter test files, while `tests/unit/test_app_message_chat_session.py` itself passed in CI, so this looks promising once a human maintainer is comfortable with the chat-session behavior and the check status is resolved.

---

kodjima33 (approved):

PR #9972, opened 2026-07-18
Make the integration create-memory text field optional as its type and consumer intend

ZachL111:
Make the integration create-memory text field optional as its type and consumer intend

An integration that creates memories from an explicit memories list, with no source text, is
rejected with a 422 before the handler runs, even though the endpoint is written to accept
exactly that shape.

models/integrations.py declared:

    text: Optional[str] = Field(description="The original text from which the fact was extracted")

Under Pydantic v2 an Optional annotation does not supply a default, so a Field() without
default= is required. Confirmed against main:

    >>> ExternalIntegrationCreateMemory.model_fields['text'].is_required()
    True
    >>> ExternalIntegrationCreateMemory(memories=[{'content': 'likes coffee'}])
    ValidationError: 1 validation error for ExternalIntegrationCreateMemory / text / Field required

The consumer is written for the text-free case (utils/conversations/memories.py):

    # Extract memories from text if provided
    if memory_data.text and len(memory_data.text.strip()) > 0:

and it processes memory_data.memories separately just above it, so an explicit-memories payload
is a supported shape that validation rejects before the handler is reached.

Every other Optional field in this model and in ExternalIntegrationMemory carries default=None
(text_source_spec, source_id, source_url, artifact_ref, tags), so this reads as the lone
omission rather than a deliberate contract.

Fix: give the field the default its type and its consumer already imply.

    text: Optional[str] = Field(description="...", default=None)

Contract impact: this only relaxes the request contract. Every request accepted before is still
accepted, and the endpoint now additionally accepts one it was already written to handle.
docs/api-reference/integration-public-openapi.json is regenerated with the pinned OpenAPI
runner, and the whole diff is the removal of text from that model's required list:

    -        "required": [
    -          "text"
    -        ],

Tests (tests/unit/test_integration_create_memory_text_optional.py):
- an explicit-memories payload with no text is accepted
- text is still accepted when provided
- a non-string text is still a validation error, so Optional did not become permissive

- pytest tests/unit/test_integration_create_memory_text_optional.py: 3 passed
- prove-fail: with models/integrations.py reverted to main, the no-text case fails with
  ValidationError while the other two still pass
- openapi export --check: all three surfaces up to date after regenerating integration-public
- generate_ts_openapi_types.py --check: clean (this model is not in the app-client surface)
- pyright models/integrations.py: 0 errors

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9972?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Nice catch and a clean fix. Verified against the head: the model change is exactly `default=None` on `text` (+1/−1), the consumer (`utils/conversations/memories.py`) processes `memory_data.memories` independently and only consumes `memory_data.text` under a truthy guard at line 85, so the text-free explicit-memories payload was already a supported handler shape that validation wrongly rejected. The OpenAPI diff is strictly more-permissive (removes `text` from `required`; nothing renamed or tightened). The new regression test covers the regression, the still-accepted path, and a guard that `Optional` did not silently become `Any`.

Note on the red checks so they don't block you: the **Backend unit suite** failures are pre-existing and unrelated to this diff — the same tests (`test_chat_memory_adapter`, `test_developer_memory_adapter`, `test_mcp_memory_adapter`, `test_product_memory_router`) fail identically on your PR's base commit. Your added test file passes cleanly. The `Formatting`/`Hygiene`/`Detect Changes` cancellations are CI orchestration noise (duplicate in-flight runs cancelled), not content problems.

Leaving a positive signal rather than a formal approval because this touches the public API contract; a maintainer should make the final merge call. Thanks for the well-scoped fix and the good test.

---

kodjima33 (approved):

PR #9980, opened 2026-07-18
Clamp pagination on GET /v1/mcp/conversations like every sibling MCP list endpoint

ZachL111:
`GET /v1/mcp/conversations?offset=-1` returns HTTP 500 for a valid, authenticated MCP client.

`get_conversations` forwards `limit` and `offset` straight into
`conversations_db.get_conversations`, which applies them as
`conversations_ref.limit(limit).offset(offset)` (database/conversations.py). Neither is clamped,
so a negative value reaches Firestore's `Query.offset()` / `Query.limit()`, which raises
server-side and surfaces as an unhandled 500.

The opposite direction is the abuse surface: `limit=999999999` is accepted as-is and streams the
whole collection, and a large `offset` forces a billed Firestore skip-scan of unbounded size. The
route is gated only by an MCP API key.

## This restores a contract the siblings already keep

Every other paginated list endpoint in this same file clamps:

| endpoint | clamp |
|---|---|
| `get_memories` | `parse_mcp_int(minimum=1, maximum=500)` |
| `get_action_items` | `max(1, min(limit, 500))` |
| `get_chat_messages` | clamped |
| `get_screen_activity` | clamped |
| `get_daily_summaries` | clamped |
| `get_conversations` | **none** |

Stronger still, the JSON-RPC twin of this exact tool, `get_conversations` in
`routers/mcp_sse.py`, was already hardened with the same clamp and is pinned by
`tests/unit/test_mcp_sse_pagination_bounds.py`. The plain REST twin here was the one that got
missed. The bounds below match that twin rather than inventing a new policy.

```python
# Clamp pagination so a negative value cannot reach Firestore .limit()/.offset() (which
# raises -> HTTP 500) and an oversized value cannot stream/skip the whole collection.
# Mirrors the sibling MCP tool (routers/mcp_sse.py get_conversations) and every other
# paginated list endpoint in this file (get_action_items, get_chat_messages, etc.).
offset = max(0, min(offset, 100000))
```

## Tests

`tests/unit/test_mcp_conversations_pagination_clamp.py`:

- a negative `limit` and `offset` are clamped before reaching Firestore
- an oversized `limit` and `offset` are capped
- in-range pagination passes through unchanged

The seam is `monkeypatch.setattr` on the module's db reference plus a direct handler call. No
`sys.modules` mutation. The test lives in a new file so it does not collide with the four open
PRs that touch `routers/mcp.py` and `test_mcp_data_endpoints.py`.

## Verification

Run in `backend/`:

- pytest new test plus `test_mcp_data_endpoints.py`, `test_mcp_sse_pagination_bounds.py` and
  `test_mcp_conversations_poison.py`: 46 passed
- prove-fail: with `routers/mcp.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), the negative case fails with
  `ValueError: Firestore .limit()/.offset() requires non-negative arguments` and the oversized
  case fails with `assert 10000000 == 1000`, while the in-range test still passes. Re-applied:
  3 passed
- `pyright routers/mcp.py`: 0 errors
- `routers/mcp.py` is 796 lines and is not in the product file line-count ratchet baseline, so no
  baseline bump is needed

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9980?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ZachL111:
Note on the red Backend unit suite: the failure is not from this change.

The failing test is
tests/unit/test_mcp_memory_adapter.py::test_mcp_vector_adapter_uses_hydrated_vector_service_and_preserves_ranking_without_archive_default,
asserting `['long-term'] == ['long-term', 'fresh-short-term']`. That is the stale date anchor in
the memory adapter fixtures: they pin a fixed `datetime(2026, 6, 19, ...)`, so a memory the test
calls "fresh short-term" now falls outside the freshness window relative to now and drops out of
the result.

This PR changes pagination clamping in `routers/mcp.py` and adds
tests/unit/test_mcp_conversations_pagination_clamp.py. It does not touch the memory adapters or
their fixtures. The test is selected here only because the changed-file selector maps
`routers/mcp.py` onto `test_mcp_memory_adapter.py` by name.

#9968 replaces those pinned anchors with `datetime.now(timezone.utc)` and covers this exact file
along with the chat, developer and product memory adapter tests. It is approved and mergeable,
and landing it clears this same red across the other PRs that touch mcp- and
developer-prefixed paths.

Git-on-my-level:
Verified the fix and the framing against `main`. `GET /v1/mcp/conversations` in `routers/mcp.py` was indeed the only paginated list endpoint in this file that forwarded raw client `limit`/`offset` straight into `conversations_db.get_conversations` (which chains Firestore `.limit()`/`.offset()`) — every sibling (`get_memories`, `get_action_items`, `get_chat_messages`, `get_screen_activity`, `get_daily_summaries`) already clamps. The chosen bounds (`limit` 1..1000, `offset` 0..100000) exactly match the already-hardened JSON-RPC twin `get_conversations` in `routers/mcp_sse.py`, so this restores an existing contract rather than introducing new policy. The 500-on-negative-input and the unbounded-scan abuse surface on an MCP-API-key-gated route are both closed, and in-range inputs pass through unchanged.

The 3 new regression tests (negative clamp, oversized cap, normal pass-through) all **PASSED** in CI (run 29655635108), and the monkeypatch seam matches the sanctioned module-import pattern. Clean, well-scoped, well-tested.

The only red check, **Backend unit suite**, is pre-existing and unrelated to this PR: the single failure is `tests/unit/test_mcp_memory_adapter.py::test_mcp_vector_adapter_uses_hydrated_vector_service_and_preserves_ranking_without_archive_default`, a stale pinned date anchor (`datetime(2026, 6, 19)`) in the memory-adapter fixtures that has aged past the freshness window. It doesn't touch `routers/mcp.py` or the new test file and is pulled in only by the changed-file name selector; @ZachL111's note on this is accurate. #9968 (approved/mergeable) replaces those anchors with `datetime.now(timezone.utc)` and clears the same red across mcp-/developer-prefixed paths. Hygiene/Formatting/Detect Changes show as `cancelled`/`skipped` selector artifacts, not real failures.

Net: this is a model small safe hardening contribution. I'm not formally approving only because the Backend unit suite check is currently red (policy: no approval while checks fail) — once #9968 lands or CI re-runs green, this is clear to merge.

---

ZachL111:
Correction to what I said earlier about the memory adapter failures.

I said the fix for them was #9968 and that landing it would clear the red. That is no longer
right, and I have closed #9968.

Main fixed the same problem itself in 60eda164cd ("test(memory): freeze vector adapter fixture
clock") on 2026-07-19, after I had last pushed to #9968. Rather than my approach of moving the
fixtures onto real wall-clock, main kept the deterministic `MEMORY_ADAPTER_FIXTURE_NOW` anchor and
added `freeze_default_vector_eligibility_clock(monkeypatch, now=now)`, which pins the production
`is_default_access_eligible` check to the same fixture clock. That removes the time dependence
instead of relying on it, and it is wired into all four adapter suites. It is the better fix, so
mine is now redundant.

Verified on origin/main at 74291042d1 with nothing applied: test_chat_memory_adapter.py 11
passed, test_developer_memory_adapter.py 24 passed, test_mcp_memory_adapter.py 26 passed,
test_product_memory_router.py 17 passed.

So nothing needs to be merged to unblock this. Any PR still showing that failure just needs
current main merged in, which I am doing.

kodjima33 (approved):

PR #10007, opened 2026-07-19
feat(desktop/agent): add Codex and best-agent routing

katipally:
This is my submission for the Omi Summer challenge, **Track 1 (Agents)**.

The ask was to let Omi run tasks with openclaw, hermes, and codex instead of only Claude Code, and to pick a good one when several are connected. Here's what's in this PR.

## 1. Codex runs as a real agent

Codex wasn't wired as an execution agent before this (it only existed as a memory-export target). Now it's a first-class ACP agent next to Claude Code, Hermes, and OpenClaw. You can pick it in the provider list, or just say "use codex to ..." and it runs the task. It talks to the codex CLI over ACP through the codex-acp bridge, set up the same way Claude Code is, so it's a peer and not something bolted under Claude Code.

## 2. "Use the best agent" picks one for you

I went back and forth on this and landed on the honest version. There's no real evidence any of these codes a normal task better than another, so I don't fake a score. When you ask for "the best agent", Omi looks at what the task actually needs and routes on that:

- "use my Claude / Anthropic" goes to Claude Code, "use my OpenAI / ChatGPT" goes to Codex (they're account-locked)
- "keep it offline / private / cheap" goes to Hermes, the only one that runs local models
- "remember this / ongoing project" goes to Hermes, it keeps memory across sessions
- "message me on Slack/Telegram" goes to OpenClaw, that's its lane (it's an assistant/orchestrator, not really a coder)

No agent is favored by default. Each of those signals maps to a capability only that one agent has, so it's the situation deciding, not a preference for any particular agent. If none of the signals apply, which is most of the time, it uses your usual agent (whatever you picked in the provider list). If your usual agent is connected but not actually signed in, it falls back to another connected agent that is, and tells you. If you haven't picked a local agent at all, it stays on Omi's own agent instead of guessing. Every pick shows why. It only kicks in when you ask for it, so it never spends your Codex or Hermes subscription behind your back.

## 3. If an agent isn't installed, it helps you install it

Naming an agent that isn't set up used to be a dead end that just said "needs setup". Now it gives you the actual commands (`npm i -g @openai/codex`, then `codex login`) and offers to use Omi's own agent instead. Same for hermes and openclaw.

## How it's wired

Node side: the codex adapter id, capability matrix, activation env, tool manifest, the spawn provider enum, per-query activation, failure labels, and a small `agent-routing.ts` that does the constraint-and-preference routing plus a `registeredAdapterIds()` kernel accessor.

Swift side: the routing enums, the provider-picker entry, CLI detection that sets the codex-acp command, and codex's identity on the pill, chat bubble, and spawn cards. Codex picks its own model, so Omi doesn't force a model profile on it.

Generated tool surfaces are regenerated from the manifest (no drift). Codex reuses the shared ACP lifecycle that acp and hermes conformance already cover, so I left it out of the frozen conformance fixture on purpose.

## Product invariants

This touches locked invariant paths and keeps both contracts. The existing acp / pi-mono / hermes / openclaw paths don't change.

**INV-AGENT-*** (agent control-plane): codex is just one more registered adapter through the registry, capability matrix, activation env, manifest, and provider enum. The model still can't pick a local provider on its own on the external surface, only you can, by naming it or asking for "the best agent".

**INV-CHAT-1** (one shared transcript across surfaces): codex is threaded through the provider identity on pills, bubbles, and spawn blocks so it shows up as codex everywhere. No new transcript store and no new identity format, the structured `agentSpawn` provider field just learns the `codex` value.

INV-VOICE-1 and INV-AUTH-1 don't apply: the FloatingControlBar files I touched aren't voice-turn owners, and the `ChatProvider.swift` change doesn't touch session invalidation, 401 handling, or auth recovery.

## Testing

- 581 agent tests and 954 desktop tests pass. Generated surfaces are drift-clean.
- Best-agent routing has tests for each signal, the usual-agent default, the fallback to another agent when your usual one is connected but not signed in, the case where nothing is set up (it defers to Omi's own agent), and the safety case where "best agent" shows up as plain task text (like "a blog post about the best agent in the NBA") and correctly does not route.
- A hermetic test checks the not-installed message actually carries the real install and login commands.
- Live over push-to-talk on a named build: all four agents connect, "use codex to ..." ran codex end to end, and naming Hermes (even correcting mid-sentence from codex to Hermes) correctly ran Hermes and wrote the file. The best-agent router itself is covered by the unit tests above and is deterministic on the voice path.
- All 21 preflight checks pass.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (dismissed):
Thanks for the thoughtful Track 1 submission — the Codex wiring is cohesive and I like that the “best agent” router avoids pretending to have a generic quality score.

I do need to request one fix before this is ready:

- The new `spawn_agent` tool guidance now tells models to pass `provider='best'` when the user asks for “the best or any agent”, but the external-surface policy only recognizes the “best agent” phrasing. If the user says something like “use any agent to refactor this”, a realtime model can follow the new schema and supply `provider: "best"`, but `constrainSpawnProviderToCurrentUserIntent` strips it because `bestAgentSelectedByUser()` does not match “any agent”. That means the advertised “any agent” route falls back to the default/Omi agent instead of entering the new best-agent resolver. Please either support the “any agent” phrasing in the policy and tests, or narrow the tool/schema wording so agents are not instructed to use a value the authority layer rejects.

Maintainer note: this PR changes the `spawn_agent` tool schema/prompt guidance and external-surface authority behavior, so it can directly change how Omi’s AI coordinators choose and spawn coding agents. The Codex addition and constraint-based routing approach look reasonable, but because this is provider/model-routing behavior on a new agent surface, it still warrants human product/security review before merge. I did not formally approve for that reason.

---

katipally:
@Git-on-my-level  so the actual issue was that the policy only recognized "best agent", even though the tool guidance tells the model to send provider:"best" for "any agent" too. those two were out of sync, so "any agent" quietly fell back to the default instead of routing.

fixed it by making the authority check recognize "any agent" as well. now something like "use any agent to refactor this" keeps provider "best" and goes into the resolver like the guidance says it should. it still needs a directive verb, so plain task text ("a blog post about the best agent in the NBA") or a negation ("don't use any agent") won't route. added tests for the "any agent" case plus the content/negation safety cases.

being honest about one tradeoff, "any" is a looser word than "best", so something like "ask any agent on the support team" would technically match. but the model decides whether to spawn at all, and even if it slips through, the resolver just uses your usual agent and never spends a paid provider off a loose match, so worst case is a harmless reroute. the invariant still holds either way since the model can't pick a provider on its own, only the user's wording can.

if you'd rather just narrow the guidance instead, that's totally fine too, lmk.

Git-on-my-level:
Thanks for the update — the blocker I raised on the previous head looks resolved now. The authority layer now recognizes the advertised “any agent” route to `provider='best'`, strips model-supplied `best` when the user did not ask for best/any agent, and has regression coverage for the positive, content-only, and negated cases.

I ran the targeted desktop agent checks locally on the current head:

- `npm test -- tests/agent-routing.test.ts tests/external-surface-authority.test.ts tests/omi-tool-manifest.test.ts tests/runtime-adapter.test.ts` — 69 tests passed
- `npm run build` — TypeScript build passed

Maintainer note: this PR changes the `spawn_agent` tool schema/prompt guidance plus external-surface authority and best-agent routing behavior, so it will affect how Omi’s AI coordinators decide when to spawn Codex/Hermes/OpenClaw or route “best/any agent” requests. The updated guidance is much safer now that the authority layer and tests match the public tool instructions, and I do not see a remaining code-level blocker from the earlier review. Because this is still provider/model-routing behavior and a new Codex agent surface, it should keep human product/security review before merge; I’m leaving the existing human-review/test/feature-fit labels in place rather than formally approving.

---

katipally:

Hey David, this is all done on my end. Blocker's fixed, conflicts with main are resolved, tests and build pass. Let me know if there's anything else you want changed or anything to do from my side before merge.

Also, if you want, I can add opencode and cursor as agents too, same way as claude code and codex. Just say the word and I'll raise a proper PR for them.

Git-on-my-level:
thank you let me test this on my side

undivisible:
Consolidated into #10088 — all 8 Track 1 Codex agent PRs merged with co-authorship preserved. Thanks for the contribution!

PR #10012, opened 2026-07-19
fix(backend): honor user timezone in conversation metadata + rendering (wrong times/dates in chat)

arhxam:
## Problem

Two timezone bugs feed the chat/LLM layer wrong dates and times — the "AI mentions the wrong time of day / wrong date" class reported in #4643 (and referenced in the code around #6214). One of them also **silently breaks metadata extraction for any user who hasn't set a timezone**.

### 1. `utils/llm/chat.py` — metadata date-grounding crashes for timezone-less users

`retrieve_metadata_from_message` / `retrieve_metadata_from_text` / `retrieve_metadata_fields_from_transcript` ground the model's notion of "today" with:

```python
today is {created_at.astimezone(ZoneInfo(tz)).strftime('%Y-%m-%d')} in {tz} ...
```

`tz` comes from `save_structured_vector`:

```python
tz = notification_db.get_user_time_zone(uid) or ''
```

So a user who has **not set a timezone** produces `tz == ''`, and `ZoneInfo('')` raises `ValueError` **before the LLM call**:

```
>>> from zoneinfo import ZoneInfo; ZoneInfo('')
ValueError: ZoneInfo keys must be normalized relative paths, got: 
```

`retrieve_metadata_fields_from_transcript` is the **default** path for regular conversations, so those users' `people/topics/entities/dates` vector metadata is never written — degrading search/retrieval. The raw `astimezone` on a possibly-naive `created_at` also reinterprets it as server-local time.

**Fix:** delegate to the module's existing purpose-built helper `date_in_tz(created_at, tz)` (`utils/llm/temporal.py`), which already falls back to UTC for a missing/invalid tz and treats naive datetimes as UTC.

```diff
-today is {created_at.astimezone(ZoneInfo(tz)).strftime('%Y-%m-%d')} in {tz} (user's local timezone).
+today is {date_in_tz(created_at, tz)} in {tz} (user's local timezone).
```

### 2. `utils/conversations/render.py` — naive timestamps rendered as server-local

`conversations_to_string` renders `created_at` / `started_at` / `finished_at` with `.astimezone(display_tz)` but skips the module's own `_as_utc()` guard — the guard that `format_local_time` / `format_local_date` already apply for exactly this reason:

```python
def _as_utc(dt):  # "Treat a naive timestamp as UTC so astimezone cannot reinterpret it as server-local."
    return dt.replace(tzinfo=timezone.utc) if dt.tzinfo is None else dt
```

`created_at`/`started_at`/`finished_at` are plain `datetime` (often naive UTC), so on any non-UTC backend host the conversation text handed to the chat LLM shows the wrong wall clock.

**Fix:** wrap each timestamp in `_as_utc()` before `.astimezone(display_tz)`.

## Tests

- **`test_conversations_to_string.py` → `TestConversationsToStringNaiveTimestamps`** forces a non-UTC process timezone (`America/Los_Angeles`) via `TZ` + `time.tzset()` so the naive-as-local defect is deterministic regardless of where CI runs, and asserts naive timestamps render as UTC and convert correctly to a user tz. These **fail on the pre-fix code**.
- **`test_temporal_date_in_tz.py`** locks the `date_in_tz` contract the `chat.py` fix now relies on: empty / `None` / invalid tz fall back to UTC **without raising** (the exact `''` input that crashed the old expression), and naive datetimes are read as UTC.

## Verification

```
$ python -m pytest tests/unit/test_conversations_to_string.py tests/unit/test_temporal_date_in_tz.py
19 passed

# revert the render.py _as_utc fix -> the two naive tests fail:
$ python -m pytest tests/unit/test_conversations_to_string.py::TestConversationsToStringNaiveTimestamps
2 failed

$ black --line-length 120 --check utils/llm/chat.py utils/conversations/render.py \
    tests/unit/test_conversations_to_string.py tests/unit/test_temporal_date_in_tz.py
All done! 4 files would be left unchanged.
```

Note: I ran the affected backend unit tests and `black` directly (in a fork). The repo's maintainer-only preflight tooling (`make preflight`, `scripts/failure-class`) requires the full monorepo checkout/toolchain and was not run here.

Relates to #4643, #6214

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10012?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_conversations_to_string.py
+++ b/backend/tests/unit/test_conversations_to_string.py
@@ -1,8 +1,11 @@
 import os
 import sys
+import time
 import types
 from datetime import datetime, timezone
 from unittest.mock import MagicMock
+
+import pytest
 
 os.environ.setdefault(
     "ENCRYPTION_SECRET",
@@ -150,3 +153,71 @@ class TestConversationsToStringTimezone:
         conv = _make_conversation()
         result = conversations_to_string([conv], tz="Not/AZone")
         assert "15 Jan 2026 at 10:00 UTC" in result
+
+
+def _make_naive_conversation(created, started, finished):
+    """A Conversation whose timestamps are timezone-naive (as stored for many rows)."""
+    return Conversation(
+        id="naive-id",
+        created_at=created,
+        started_at=started,
+        finished_at=finished,
+        structured=Structured(title="T", overview="O", category=CategoryEnum.personal),
+        apps_results=[],
+    )
+
+
+class TestConversationsToStringNaiveTimestamps:
+    """Naive stored timestamps must be read as UTC, never as the server's local time.
+
+    ``conversation.created_at`` / ``started_at`` / ``finished_at`` are plain ``datetime``
+    and are frequently naive (stored as naive UTC). Calling ``.astimezone(display_tz)`` on a
+    naive datetime makes Python assume it is in the *server's* local timezone, so the text
+    handed to the chat LLM shows a wrong wall clock whenever the backend host is not on UTC —
+    exactly the "incorrect event time mentions" class of bug the module's ``_as_utc`` guard
+    exists to prevent (issues #4643, #6214). This test forces a non-UTC process timezone so
+    the naive-as-local defect is deterministic regardless of where the suite runs.
+    """
+
+    def setup_method(self):
+        self._old_tz = os.environ.get("TZ")
+        os.environ["TZ"] = "America/Los_Angeles"  # UTC-8 / -7, well away from UTC
+        if hasattr(time, "tzset"):
+            time.tzset()
+
+    def teardown_method(self):
+        if self._old_tz is None:
+            os.environ.pop("TZ", None)
+        else:
+            os.environ["TZ"] = self._old_tz
+        if hasattr(time, "tzset"):
+            time.tzset()
+
+    def test_naive_timestamps_render_as_utc(self):
+        if not hasattr(time, "tzset"):
+            pytest.skip("time.tzset() unavailable on this platform")
+        # Naive 20:00 must be read as 20:00 UTC. The buggy path would treat it as
+        # 20:00 America/Los_Angeles and render 16 Jan ~04:00 UTC instead.
+        conv = _make_naive_conversation(
+            created=datetime(2026, 1, 15, 20, 0),
+            started=datetime(2026, 1, 15, 20, 0),
+            finished=datetime(2026, 1, 15, 20, 30),
+        )
+        result = conversations_to_string([conv])
+        assert "15 Jan 2026 at 20:00 UTC" in result
+        assert "Started: 15 Jan 2026 at 20:00 UTC" in result
+        assert "Finished: 15 Jan 2026 at 20:30 UTC" in result
+        assert "16 Jan" not in result
+
+    def test_naive_timestamps_convert_to_user_tz(self):
+        if not hasattr(time, "tzset"):
+            pytest.skip("time.tzset() unavailable on this platform")
+        # Naive 20:00 (UTC) -> 17:00 in America/Sao_Paulo (UTC-3), independent of server tz.
+        conv = _make_naive_conversation(
+            created=datetime(2026, 1, 15, 20, 0),
+            started=datetime(2026, 1, 15, 20, 0),
+            finished=datetime(2026, 1, 15, 20, 30),
+        )
+        result = conversations_to_string([conv], tz="America/Sao_Paulo")
+        assert "15 Jan 2026 at 17:00 America/Sao_Paulo" in result
+        assert "Finished: 15 Jan 2026 at 17:30 America/Sao_Paulo" in result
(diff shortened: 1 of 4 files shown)


Git-on-my-level (commented):
Thanks @arhxam — this is a genuinely well-scoped first contribution. Clear root-cause write-up, the fix lands exactly where it should, and the contract tests are the right shape. Leaving a positive signal rather than a formal approval for the reasons at the end.

What this fixes, and why it's correct
- chat.py metadata prompts: `created_at.astimezone(ZoneInfo(tz))` raised `ValueError` whenever `tz == ''` (users with no saved timezone — the default conversation path), which aborted metadata extraction before the LLM call and silently dropped people/topics/entities/dates vector metadata for those users. It also re-interpreted naive timestamps as server-local. Delegating to `date_in_tz(created_at, tz)` fixes both: missing/None/invalid tz falls back to UTC, naive datetimes are read as UTC, and it never raises. I confirmed `date_in_tz` in utils/llm/temporal.py has exactly those guarantees.
- render.py `conversations_to_string`: wrapping created_at/started_at/finished_at with `_as_utc(...)` before `.astimezone(display_tz)` stops naive stored timestamps being read as the host's local wall clock. `_as_utc` already exists in render.py and is the same helper used by `format_local_time`/`format_local_date`, so this just makes the module consistent.

Verification notes
- All three chat.py call sites now use `date_in_tz(created_at, tz)`; all three render.py lines use `_as_utc(...)`. Diff matches the description.
- `from zoneinfo import ZoneInfo` is still used elsewhere in both files, so there's no dead-import regression.
- The new tests (test_temporal_date_in_tz.py + TestConversationsToStringNaiveTimestamps) lock the exact invariants the fixes rely on, and forcing `TZ=America/Los_Angeles` makes the naive-as-local failure deterministic regardless of where the suite runs.

One thing for a human maintainer before merge
- The only CI signal on this PR is `cubic` (neutral); the backend pytest suite does not appear to have actually executed, so the new tests haven't been exercised by automation yet. They look correct by inspection, but please confirm the suite runs green (ideally including the two new files) before merging.

No changes requested from my side — flagging needs-maintainer-review purely because this touches user-facing times and the default metadata-extraction path, and a maintainer should be the one to push the merge button. Nice work.

---

kodjima33 (approved):
Backend timezone fix in conversation metadata/rendering (#4643) with tests. Approve-only — no CI checks ran for contributor.

Git-on-my-level:
Thanks @arhxam — this is a clean, well-scoped fix for a real defect class. Quick maintainer notes after reviewing against `main`:

**What this fixes (verified)**
- `utils/llm/chat.py` — the three metadata-extraction prompts replace `created_at.astimezone(ZoneInfo(tz))` with the existing `date_in_tz` helper. This fixes a crash for timezone-less users: `tz == ''` (the default path, from `get_user_time_zone(uid) or ''`) made `ZoneInfo('')` raise `ValueError`, so the default conversation path silently produced no people/topics/entities/dates vector metadata. It also stops naive `created_at` from being read as server-local.
- `utils/conversations/render.py` — `conversations_to_string` wraps `created_at` / `started_at` / `finished_at` with the existing `_as_utc()` helper before `.astimezone(display_tz)`, so naive stored timestamps are treated as UTC rather than server-local.

Both `date_in_tz` and `_as_utc` already exist on `main` and are purpose-built for these defects, so this reuses them rather than reinventing. `ZoneInfo` is still imported in `chat.py` because `get_user_timezone` / `get_current_datetime_block` use it with their own try/except UTC fallback — those aren't the defect here, so leaving them is correct.

**Tests**
- `test_temporal_date_in_tz.py` locks the empty-string-tz `ValueError` regression plus None/invalid fallbacks and naive-as-UTC.
- The `conversations_to_string` extension forces `TZ=America/Los_Angeles` via `time.tzset()` to make the naive-as-server-local defect deterministic, with proper save/restore and a `pytest.skip` where `tzset` is unavailable.

Coverage matches the fix surface and would fail on the pre-fix code.

**Status**
Not approving — this needs a human maintainer sign-off before merge (it touches user-data persistence/retrieval surfaces), so leaving the `needs-maintainer-review` label in place. The code itself looks correct and minimal.

**Two non-blocking notes for a follow-up, not a reason to block or split this PR:**
1. The three metadata prompts still contain the line `Do not include dates greater than 2025.` This pre-exists on `main` (not introduced here) and is already stale; worth a separate issue.
2. The naive-timestamps-as-server-local defect class likely exists at other `astimezone` call sites across the backend beyond these two modules. This PR fixes the two that feed the chat LLM, which is a coherent slice; a repo-wide sweep would be a natural follow-up.

---

PR #10014, opened 2026-07-19
fix(backend): enforce Plus plan transcription cap (remaining-seconds treated it as unlimited)

arhxam:
## Problem

`get_remaining_transcription_seconds()` (`utils/subscription.py`) short-circuits to `None` ("unlimited") for **every** paid plan:

```python
subscription = users_db.get_user_valid_subscription(uid)
if not subscription:
    limits = get_basic_plan_limits()
elif is_paid_plan(subscription.plan):
    return None  # Unlimited
else:
    limits = get_plan_limits(subscription.plan)
```

But **Plus** is a paid plan (it's in `PAID_PLAN_TYPES`) that carries a **bounded** monthly transcription cap:

```python
get_plan_limits(PlanType.plus).transcription_seconds
    == PLUS_TIER_MONTHLY_SECONDS_LIMIT
    == PLUS_TIER_MINUTES_LIMIT_PER_MONTH * 60
    == 1500 * 60 == 90000
```

and the plan is literally sold as *"1,500 minutes of transcription per month"*. So Plus subscribers are treated as unlimited: the function never returns a bounded remaining value, and the freemium auto-switch to on-device transcription in `routers/listen/runtime.py` (which fires when `remaining <= FREEMIUM_THRESHOLD_SECONDS`) **never triggers** — a Plus user runs unbounded cloud transcription past the cap they're sold.

## Root cause

Every *other* paid plan — operator, architect, neo/`unlimited`, `unlimited_v2` — has `transcription_seconds = None` in `get_plan_limits`, so the existing `if not limits.transcription_seconds` guard already reports them as unlimited. **Plus is the only paid plan with a bounded cap**, and the blanket `is_paid_plan()` short-circuit swallows it.

The sibling `has_transcription_credits()` gets this right — it does **not** short-circuit on `is_paid_plan`; it resolves `get_plan_limits(plan)` and gates on `transcription_seconds`. `fair_use.py` likewise excludes Plus from its unlimited-transcription set. The two functions disagreed only for Plus.

## Fix

Drop the `is_paid_plan()` short-circuit and let the `transcription_seconds` check decide unlimited-ness — mirroring `has_transcription_credits()`:

```diff
-elif is_paid_plan(subscription.plan):
-    return None  # Unlimited
-else:
-    limits = get_plan_limits(subscription.plan)
+else:
+    limits = get_plan_limits(subscription.plan)
```

Behavior change is limited to Plus. Unlimited paid plans still return `None` (their `transcription_seconds` is unset); the no-subscription and basic paths are unchanged.

## Tests

Added to `tests/unit/test_subscription_plans.py` (existing coverage only tested the sibling `has_transcription_credits`, so `get_remaining_transcription_seconds` for a bounded paid plan was uncovered):

- `test_remaining_transcription_seconds_enforces_plus_bounded_cap` — Plus, usage `cap - 10000` → `10000` (was `None`).
- `test_remaining_transcription_seconds_zero_at_plus_cap` — usage over cap → `0`.
- `test_remaining_transcription_seconds_none_for_unlimited_paid_plan` — architect/operator/neo/unlimited_v2 still return `None` (guards against over-capping).

## Verification

```
$ python -m pytest tests/unit/test_subscription_plans.py
12 passed

# revert the fix -> the two Plus tests fail (the unlimited-paid guard still passes):
2 failed, 1 passed, ...
```

`black --line-length 120 --check utils/subscription.py tests/unit/test_subscription_plans.py` → clean.

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_subscription_plans.py
+++ b/backend/tests/unit/test_subscription_plans.py
@@ -103,6 +103,47 @@ def test_bounded_transcription_plan_reads_monthly_usage_and_enforces_cap(monkeyp
     monthly_usage.assert_called_once_with('uid')
 
 
+def _stub_remaining_deps(monkeypatch, subscription_module, plan, used_seconds):
+    monkeypatch.setattr(subscription_module, 'is_trial_paywalled', lambda uid, source=None: False)
+    monkeypatch.setattr(subscription_module.users_db, 'is_byok_active', lambda uid: False, raising=False)
+    monkeypatch.setattr(subscription_module, 'get_byok_key', lambda provider: None)
+    monkeypatch.setattr(
+        subscription_module.users_db,
+        'get_user_valid_subscription',
+        lambda uid: SimpleNamespace(plan=plan),
+        raising=False,
+    )
+    monkeypatch.setattr(
+        subscription_module,
+        'get_monthly_usage_for_subscription',
+        lambda uid: {'transcription_seconds': used_seconds},
+    )
+
+
+def test_remaining_transcription_seconds_enforces_plus_bounded_cap(monkeypatch, subscription_module):
+    # Plus is a paid plan but carries a bounded 1500-min/month transcription cap. The
+    # remaining seconds must be reported (so the freemium on-device switch can fire), not
+    # short-circuited to None as if the plan were unlimited.
+    cap = subscription_module.PLUS_TIER_MONTHLY_SECONDS_LIMIT
+    assert cap and cap > 0  # Plus is bounded by construction
+    _stub_remaining_deps(monkeypatch, subscription_module, PlanType.plus, used_seconds=cap - 10000)
+    assert subscription_module.get_remaining_transcription_seconds('uid') == 10000
+
+
+def test_remaining_transcription_seconds_zero_at_plus_cap(monkeypatch, subscription_module):
+    cap = subscription_module.PLUS_TIER_MONTHLY_SECONDS_LIMIT
+    _stub_remaining_deps(monkeypatch, subscription_module, PlanType.plus, used_seconds=cap + 5000)
+    assert subscription_module.get_remaining_transcription_seconds('uid') == 0
+
+
+def test_remaining_transcription_seconds_none_for_unlimited_paid_plan(monkeypatch, subscription_module):
+    # Genuinely-unlimited paid plans (transcription_seconds unset) must still report None,
+    # i.e. dropping the is_paid_plan short-circuit must not start capping them.
+    for plan in (PlanType.architect, PlanType.operator, PlanType.unlimited, PlanType.unlimited_v2):
+        _stub_remaining_deps(monkeypatch, subscription_module, plan, used_seconds=10_000_000)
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Verified this end-to-end against `main` — the diagnosis is correct and the fix is the right shape. Thank you @arhxam for a clean, well-explained bug report and targeted tests.

What I confirmed from source:
- `get_remaining_transcription_seconds()` on `main` does `elif is_paid_plan(subscription.plan): return None`, and `PlanType.plus` is in `PAID_PLAN_TYPES`, so Plus was being reported as unlimited.
- `get_plan_limits(PlanType.plus).transcription_seconds == PLUS_TIER_MONTHLY_SECONDS_LIMIT` (1500 × 60 = 90000). Every other paid plan (operator / architect / unlimited / unlimited_v2) has `transcription_seconds=None`, so the existing `if not limits.transcription_seconds` guard already handles them as unlimited — your change does not regress those.
- The sibling `has_transcription_credits()` already does exactly this (no `is_paid_plan` short-circuit), and `fair_use._is_unlimited_tier()` also excludes `plus`. So this PR brings `get_remaining_transcription_seconds()` in line with the rest of the codebase.
- The only production caller (`routers/listen/runtime.py`) keys the freemium on-device auto-switch off `remaining <= FREEMIUM_THRESHOLD_SECONDS` (180s). Returning `None` for Plus meant the switch never fired and Plus users ran unbounded cloud transcription past the 1500-min cap they are sold. This fix closes that leak.

The new tests are well-scoped: the Plus `cap - 10000 → 10000` and `cap + 5000 → 0` cases pin the bounded behaviour, and the `none_for_unlimited_paid_plan` loop across architect/operator/unlimited/unlimited_v2 guards against the obvious regression where dropping the short-circuit could accidentally start capping genuinely-unlimited tiers.

Two small notes for whoever lands this (not blockers for the idea):
1. No CI workflow appears to have run on this head yet (only the cubic reviewer, neutral). Worth confirming the new `test_remaining_transcription_seconds_*` tests go green in CI before merge.
2. `subscription_module.users_db.get_user_valid_subscription` is stubbed with `raising=False`, so the test is not exercising the real `users_db` import path — fine for this unit test, but it means CI is the place that actually validates the module loads cleanly.

Marking for human maintainer review since this touches the billing/quota enforcement surface and is a first-time contribution; the change itself is a strict cap-enforcement fix (it makes the system stricter, not looser) and aligns with how Plus is marketed.

---

kodjima33 (approved):
Backend fix (enforce Plus plan transcription cap) with test. Approve-only — billing/quota behavior change + no CI run; Nik to land.

undivisible:
looks good :)) thank you

PR #10018, opened 2026-07-19
fix(desktop): use backend IDs for task candidates

Git-on-my-level:
## Summary

- Use durable backend action-item IDs, rather than local SQLite row IDs, in task extraction's active-task context and duplicate-search result contracts.
- Keep unsynced local and staged task records available as id-less duplicate evidence, so they cannot be selected as candidate mutation targets.
- Add prompt, serialization, and isolated SQLite regression coverage.

## Verification

- `xcrun swift test --package-path Desktop --disable-automatic-resolution --filter TaskAssistantContextPromptTests`
- `xcrun swift test --package-path Desktop --disable-automatic-resolution --filter TaskIntelligenceSQLiteRoundTripTests/testActiveTaskContextExposesOnlyBackendTaskIDs`
- `OMI_APP_NAME="omi-task-canonical-id" OMI_ALLOW_ADHOC_SIGN=1 OMI_SIGN_IDENTITY="-" ./run.sh --yolo` (built, signed, and launched the isolated named bundle)
- `make preflight`

## Failure class

The backend action-item ID is the authoritative identity for candidate mutations. This repair keeps local SQLite records as evidence only, so they cannot provide a competing mutation target.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-proactiveassistants.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-proactiveassistants.json
@@ -1,10 +1,10 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/TaskExtraction/TaskAssistant.swift": 1736,
+    "desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/TaskExtraction/TaskAssistant.swift": 1742,
     "desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift": 1645
   },
   "raise_justifications": {
-    "desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/TaskExtraction/TaskAssistant.swift": "Staged tasks now render as id-less 'already captured' evidence via a pure, unit-tested contextEvidencePrompt helper, so they can no longer be echoed back as duplicate_of:0 to update a non-existent backend task.",
+    "desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/TaskExtraction/TaskAssistant.swift": "Task candidate context now carries only backend action-item IDs, while local SQLite/staged evidence remains id-less so the model cannot target it for a backend mutation.",
     "desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift": "stopMonitoring now clears capture-recovery/background-polling state and timers, plus a pure ProactiveCapturePolicy gate, so a stop mid-recovery cannot silently disable capture on the next start."
   },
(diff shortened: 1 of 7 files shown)


Git-on-my-level:
## Monitor summary — Codex P2: filter id-less rows before LIMIT

Addressed the one open review thread (Codex P2, `chatgpt-codex-connector`): the candidate-target context applied `LIMIT 30` in SQLite *before* `compactMap` dropped rows without a backend ID, so a backlog of unsynced local rows could consume the whole window and crowd out older backend-synced tasks that are still valid duplicate/refinement targets — the exact failure this PR exists to prevent.

### Change
- Added an opt-in `requireBackendId: Bool = false` parameter to `ActionItemStorage.getTopRelevanceTasks` and `getRecentActiveTasks`. When true, rows without a durable backend ID are filtered in SQL **before** the `LIMIT` is applied.
- `TaskAssistant.refreshContext()` (the candidate-mutation-target path) opts in with `requireBackendId: true`. The Swift-side `compactMap` stays as a cheap defensive guard.
- Informational callers (e.g. `FocusAssistant`, which only displays task descriptions) keep the default `false` so local-only tasks remain visible — no behavior change for them.
- Added regression test `testCandidateWindowFiltersIdLessRowsBeforeLimit`: inserts 35 id-less local rows ahead of one backend-synced task, then asserts the backend task survives inside a `LIMIT 30` candidate window, while the informational path still returns all 30 rows.

### Verification
- `Desktop Swift Build & Tests` — **pass** (27m33s) — exercises the new regression test in CI.
- Local: `swiftc -parse` clean on all 3 changed files; `swift-format lint-scope` clean; `swiftlint` 0 violations / 784 files; swiftlint baseline down-only.
- All 19 repo manifest checks pass (failure-class `FC-split-mutation-authority`, product-file ratchet baselines raised with justification, changelog entry present).
- Pre-push gate passed; the full `swift build` runs in CI (local `swift build` was blocked by a swift-package-manager CFRunLoop hang in this worktree, documented as a tool-budget bypass; parse + lint + CI build all green).

Failure-Class: `FC-split-mutation-authority` — the backend action-item ID remains the sole authoritative mutation target; id-less local rows can no longer starve it out of the candidate window.

kodjima33 (approved):
Desktop bug fix (durable backend IDs for task candidates) with Swift regression tests; CI green. Owner-override approve+merge.

PR #10029, opened 2026-07-19
Return the actual newest X posts when filtering by kind

ZachL111:
`get_x_posts` promises "newest first" in its own docstring and again in the `GET /v1/x/posts`
docstring, but the kind-filtered branch cannot deliver it.

```python
for d in coll.where(filter=FieldFilter('kind', '==', kind)).limit(limit * 3).stream():
    ...
docs.sort(key=lambda x: str(x.get('created_at') or ''), reverse=True)
return docs[:limit]
```

Firestore gives no ordering guarantee without an explicit `order_by`, so `.limit(limit * 3)`
truncates to an **arbitrary** window of matching documents before the Python sort ever runs. Once
a user has more than `limit * 3` posts of that kind, the genuinely newest posts can fall entirely
outside that window and never reach the sort. The endpoint then returns older posts, correctly
sorted among themselves, with no error.

A user with 500 bookmarks calling `GET /v1/x/posts?kind=bookmark&limit=100` gets up to 300
arbitrarily selected bookmarks re-sorted, not their 100 newest.

The unfiltered branch three lines below already does it in the right order, ordering at the query
and then limiting:

```python
query = coll.order_by('created_at', direction=firestore.Query.DESCENDING).limit(limit)
```

## Why removing the limit is the right fix, not a new unbounded read

This is the part worth a second look, since dropping a `.limit()` normally deserves suspicion.
Removing it restores the approach this function's own docstring already describes:

> A kind filter + order_by would require a composite index, so when filtering by kind we use the
> single-field equality query (auto-indexed) and sort in Python — post volumes per user are small
> enough for this to be cheap.

The `.limit(limit * 3)` contradicted that documented approach while silently breaking the
ordering contract. The query shape is otherwise unchanged: still a single-field equality filter
on the same automatic index, so **no new index is required**.

Bounding the query properly instead would mean an `order_by` alongside the `kind` equality
filter, which needs a composite index this repo does not declare (`firestore.indexes.json` has no
`x_posts` entries at all). That is a serving-schema change, not a bug fix, so I am flagging it as
a follow-up rather than smuggling it in here. Worth doing if per-user post volumes ever stop
being small.

## Affected surfaces

Both registered unconditionally in `main.py`:

- `GET /v1/x/posts?kind=...` (`routers/x_connector.py`)
- the `get_x_posts` MCP tool (`routers/mcp_sse.py`)

## Tests

`tests/unit/test_x_posts_kind_filter_limit_before_sort.py`:

- with more posts than the old pre-sort window, the true newest are returned
- a kind filter whose matches fit inside the old window still returns the same result

The fake query builder records `.limit()` calls and truncates in insertion order, which is what
makes the unordered-limit behaviour observable in a hermetic test rather than only in production.

## Verification

Run in `backend/`:

- pytest new test plus `test_x_posts_sort_mixed_created_at.py`, `test_x_posts_list_endpoint.py`
  and `test_mcp_sse_pagination_bounds.py`: 14 passed
- prove-fail: with `database/x_posts.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), the bug-path test fails with `assert ['5', '4'] == ['9', '8']`
  (ten posts, oldest to newest; `limit=2` should return the two newest, but the `limit=6` window
  only ever contains the six oldest), while the within-window test still passes both ways.
  Re-applied: 2 passed
- `pyright database/x_posts.py`: 0 errors
- `database/x_posts.py` is not in the product file line-count ratchet baseline; line count
  unchanged at 184

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10029?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/x_posts.py
+++ b/backend/database/x_posts.py
@@ -126,7 +126,7 @@ def get_x_posts(uid: str, limit: int = 100, kind: Optional[str] = None) -> List[
     coll = _posts_ref(uid)
     if kind:
         docs: List[Dict[str, Any]] = []
-        for d in coll.where(filter=FieldFilter('kind', '==', kind)).limit(limit * 3).stream():
+        for d in coll.where(filter=FieldFilter('kind', '==', kind)).stream():
             raw: object = d.to_dict()
             if isinstance(raw, dict):
                 docs.append(cast(Dict[str, Any], raw))
--- a/backend/tests/unit/test_x_posts_kind_filter_limit_before_sort.py
+++ b/backend/tests/unit/test_x_posts_kind_filter_limit_before_sort.py
@@ -1,0 +1,103 @@
+"""Regression test: get_x_posts(kind=...) must sort before truncating, not after.
+
+database.x_posts.get_x_posts's kind-filtered branch runs:
+
+    coll.where(filter=FieldFilter('kind', '==', kind)).limit(limit * 3).stream()
+
+with no ``order_by`` (a kind filter + order_by would need a new composite
+index, so the module deliberately sorts in Python instead -- see the
+function's own docstring). But ``.limit()`` without a matching ``order_by``
+returns an ARBITRARY ``limit * 3`` matching documents, not the newest ones --
+Firestore is free to return matches in any order (in practice, often
+insertion/document-id order, and these documents are keyed by tweet
+snowflake id, which increases over time, i.e. oldest-first). Once an account
+has more than ``limit * 3`` posts of a kind, the true newest posts can fall
+outside that arbitrary pre-sort window and never reach the Python sort at
+all. GET /v1/x/posts (routers/x_connector.py, docstring: "newest first") and
+the MCP tool get_x_posts then silently return stale posts instead of the
+newest ones, with no error.
+
+The fix drops the premature ``.limit(limit * 3)`` so every matching document
+is sorted before the final ``limit`` slice -- matching the sibling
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
Thanks for the focused fix — this looks like the right direction.

I reviewed the change and ran the targeted backend unit tests locally. Removing the pre-sort Firestore `.limit(limit * 3)` from the `kind` branch matches this function's documented behavior: fetch the matching kind rows, sort by `created_at` in Python, then return the requested newest slice. The new regression test covers the failure mode where the true newest rows are outside the old arbitrary pre-sort window, and the existing mixed-`created_at` sort regression still passes.

Validation run:
`python -m pytest tests/unit/test_x_posts_kind_filter_limit_before_sort.py tests/unit/test_x_posts_sort_mixed_created_at.py -q` → 3 passed.

I am not formally approving from automation because this touches backend data retrieval behavior and the PR still needs the normal maintainer/check gate, but I do not see a code blocker in this patch.

---

kodjima33 (approved):

PR #10031, opened 2026-07-19
Anchor the usage endpoint's "today" to the user's timezone instead of the server's UTC day

ZachL111:
`GET /v1/users/me/usage?period=today` computes "today" from the server's UTC calendar day and
never consults the user's stored timezone — even though three other endpoints in this same
router already call `notification_db.get_user_time_zone(uid)` for exactly this "which calendar
day" question.

Hourly usage documents are written keyed by UTC date (`utils/analytics.py` stamps
`datetime.now(timezone.utc)`), so a user behind UTC has the first hours of their local day filed
under the **previous** UTC date. `get_today_usage_stats` queried a single UTC date by
`year`/`month`/`day` equality, so those hours were dropped.

A user in `America/Los_Angeles` checking usage at 8pm local sees only the hours since 5pm local
(UTC midnight). The rest of their day is silently missing from the Today cards on the Flutter
usage page (`app/lib/pages/settings/usage_page.dart` fetches this endpoint with
`period='today'`). Users east of UTC get the mirror image, with part of last night counted as
today.

## The fix

`get_today_usage_stats` now takes a `[start, end)` UTC range and walks the at-most-two UTC
calendar days it spans, summing only the hourly buckets inside the range.
`get_current_user_usage` takes an optional `tz_name` and derives that range from the user's local
midnight, using the same pytz idiom already present in `utils/other/notifications.py` and the
daily-summary endpoint in this router.

Without a stored timezone the behaviour is unchanged, so users with no timezone on file see
exactly what they saw before.

**The query shape is deliberately unchanged.** An id range or an `order_by` would have been a new
unregistered Firestore query shape, and the repo ratchets those
(`scripts/firestore_query_coverage.py`). Reusing the existing equality filters inside the loop
keeps that ratchet green — verified by running it, not assumed.

A timezone that fails to parse now logs and falls back to the UTC day rather than failing the
request. That branch previously swallowed the error silently; a stored zone the server cannot
parse is a data problem worth seeing.

## Known gap, deliberately not fixed here

The same response's `history` field (`get_hourly_history_for_today`) is **still keyed to the UTC
day**, so the hourly chart and the totals above it can disagree near the boundary. Making it
consistent means deciding whether the chart's hour labels become local hours, which changes what
clients render — a contract question rather than a bug fix. I would rather flag it than fold it
in silently. Happy to do it here if you would prefer them land together.

## Tests

`tests/unit/test_user_usage.py`:

- a local day spanning two UTC dates returns the full local-day total for a user west of UTC
- a user with no stored timezone still gets the old UTC-day behaviour
- the endpoint itself serves the local day, exercised through the route the app calls

The third is the behavioural one, and worth explaining. The first two can only be written against
the tz-aware signature, so on unfixed source they fail with a `TypeError` about the new keyword
rather than a wrong number — that proves the parameter is new, not that the old behaviour was
wrong. Going through the router instead produces a real value mismatch, because the handler there
never consults the timezone at all.

## Verification

Run in `backend/`:

- `pytest tests/unit/test_user_usage.py`: 7 passed
- prove-fail: with `database/user_usage.py` and `routers/users.py` reverted to origin/main and
  confirmed byte-identical (`git diff origin/main` empty), the endpoint test fails with
  `assert 300 == 900` — the 7am local bucket, filed under the previous UTC date, is dropped — and
  the four pre-existing tests in the file pass unchanged. Re-applied: 7 passed
- `pyright database/user_usage.py routers/users.py`: 0 errors
- `scripts/check_module_stub_pollution.py`: 735 files, 0 violations
- `scripts/scan_import_time_side_effects.py`: 717 files, 0 violations
- `scripts/firestore_query_coverage.py --check-ratchet`: pass, no new query shape
- `routers/users.py` is 2046 lines, exactly its ratchet baseline, so no baseline change;
  `database/user_usage.py` is not in any ratchet shard

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10031?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/user_usage.py
+++ b/backend/database/user_usage.py
@@ -1,12 +1,16 @@
-from datetime import datetime, timezone
+import logging
+from datetime import datetime, time, timedelta, timezone
 from typing import Any, Dict, Iterable, List, Optional, Tuple, cast
 
+import pytz
 from google.cloud import firestore
 from google.cloud.firestore_v1 import FieldFilter
 
 from ._client import db
 from .firestore_read_metrics import FirestoreReadFamily, FirestoreReadMode, record_firestore_read
 from models.user_usage import UsageStats
+
+logger = logging.getLogger(__name__)
 
 
 def _typed_doc(doc: Any) -> Dict[str, Any]:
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):

Issue #10028, opened 2026-07-19
fix(listen): terminate stale live sessions when STT peers end cleanly


Git-on-my-level:
## Problem

The original client-side transcript-silence watchdog proposal was unsafe: ordinary quiet periods are indistinguishable from a stale stream because BLE audio packets continue during silence. It would reconnect healthy sessions at the configured 120-second conversation-silence boundary.

There is a narrower, observable failure boundary in the current `/v4/listen` pipeline. When an active Parakeet or Modulate upstream WebSocket ends cleanly without an exception, its wrapper does not latch the connection as dead. The mobile socket can remain `ready` until the device sends another audio frame and the next provider send fails. That extends the “Listening” zombie state even after the backend has observed its upstream transport disappear.

## Scope

- Latch an unexpected clean upstream close as terminal in the Parakeet and Modulate streaming wrappers, excluding explicit local drain/finalization.
- Propagate that observed provider death immediately through the existing `stt_failed` + WebSocket 1011 terminal path, without waiting for another client audio frame.
- Add hermetic coverage through the real `/v4/listen` route and controlled loopback provider peer: after one audio frame, a clean provider close must deliver the terminal status/close without a second audio frame. Add focused wrapper coverage for both provider implementations.

This does **not** claim that lack of transcript text proves a transport failure, nor does it replace the durable audio/receipt boundary still needed for #6977. It only closes a session when the backend has directly observed the upstream transport ending.

## Sequencing

Wait for #10023 to merge first: it modifies the same streaming/runtime and listen wire-contract surfaces. The implementation PR should use `Closes #10028`.

Related parent: #6977.

aryanorastar:
Opened #10265 for this. Root cause: when a Parakeet/Modulate live upstream WebSocket closes cleanly, the wrapper's `async for` exhausts without raising, so `_dead` never latches and `live_stt_socket_is_dead` reads a stale False. The fix latches the clean close in both wrappers (excluding local drain / `done` finalization) and adds a frame-independent monitor that drives the existing idempotent `stt_failed` + WS 1011 terminal path — so the session terminates without waiting for another audio frame or the 300s `ws_receive_timeout`. Hermetic wrapper + monitor coverage; 121 STT/listen tests pass.

PR #10036, opened 2026-07-19
fix(backend): latch clean upstream STT close as terminal (#10028)

kodjima33:
## Problem

Issue #10028. In the `/v4/listen` STT pipeline, when an active **Parakeet** (`ParakeetWebSocketSocket`) or **Modulate** (`SafeModulateSocket`) upstream WebSocket ends **cleanly** (the provider closes the transport with no exception and, for Modulate, no `done`/`error` message), the streaming wrapper's receive loop simply falls out of its `async for` without latching the connection as dead.

`is_connection_dead` therefore stays `False`, and the mobile socket keeps reporting `ready`/"Listening" until the device sends another audio frame and the *next* provider send finally fails. That extends the "Listening" zombie state after the backend has already observed its upstream transport disappear.

## Root cause

Both receive loops only mark the socket dead inside their `except` handlers. A clean loop exhaustion (the exact signature of an unexpected clean upstream close mid-session) has no terminal handling. A healthy stream never exhausts `async for` — it blocks waiting for messages — so reaching loop end is unambiguously a real transport close.

## Fix

Add an `else` branch to each receive loop that latches the socket dead on clean exhaustion, excluding local drain/finalization:

- **Parakeet** (`ParakeetWebSocketSocket._receive_loop`): mark dead unless we already closed or called `finish()`/`finalize()` (`_closed` / `_finalized`).
- **Modulate** (`SafeModulateSocket._recv_loop`): mark dead and set `_done_event` unless we already closed (`finish()`) or a legitimate `done`/`error` already terminated the stream (`_done_event` set). All explicit-termination paths `break` after setting those guards, so only an unexpected close reaches the `else`.

Once latched, the existing `is_connection_dead` gate on the send path (e.g. `chat.py`) trips at the clean-close boundary instead of waiting for a send to raise.

## Scope note

This implements the wrapper-latch portion of #10028. The issue also proposes an active monitor to propagate the terminal `stt_failed` + WebSocket 1011 *without any further client frame*; that touches the core listen receiver and is intentionally left out of this surgical change to keep blast radius small. Filed here as PR-only for review, not auto-merged.

## Tests

- `TestRecvLoop::test_clean_close_without_done_marks_dead` (Modulate) — new regression.
- `TestRecvLoop::test_invalid_json_skipped` — updated to assert the now-correct dead-latch on exhaustion.
- `TestProcessAudioParakeet::test_receive_loop_clean_upstream_close_marks_dead` — new regression.
- `TestProcessAudioParakeet::test_receive_loop_close_after_finalize_stays_alive` — guards against false positives on normal finalization.

Ran locally (Python 3.12, hermetic): the STT socket suites pass — `TestRecvLoop`, `TestUtteranceResults`, `TestProcessAudioParakeet` = 18 passed. (The `Prerecorded` cases in the same file fail only because `httpx` was absent in the ad-hoc venv, unrelated to this change.)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_modulate_stt.py
+++ b/backend/tests/unit/test_modulate_stt.py
@@ -531,8 +531,19 @@ class TestRecvLoop(unittest.TestCase):
 
     def test_invalid_json_skipped(self):
         sock = self._run_recv(['not json', '{{bad'])
-        self.assertFalse(sock.is_connection_dead)
+        # Invalid frames produce no segments; the subsequent clean upstream close
+        # (message stream exhausted with no `done`) latches the socket dead.
         self.assertEqual(self.segments, [])
+        self.assertTrue(sock.is_connection_dead)
+
+    def test_clean_close_without_done_marks_dead(self):
+        # A provider that ends its WebSocket cleanly without a `done`/`error`
+        # message must latch the socket dead so the listen loop stops treating
+        # the session as ready (regression for the "Listening" zombie state).
+        sock = self._run_recv([])
+        self.assertTrue(sock.is_connection_dead)
+        self.assertIn('closed cleanly by upstream', sock.death_reason)
+        self.assertTrue(sock._done_event.is_set())
 
     def test_error_message_marks_dead(self):
         msg = json.dumps({'type': 'error', 'error': 'rate limit'})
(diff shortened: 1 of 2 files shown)


kodjima33:

PR #10037, opened 2026-07-19
Bound the OAuth callback failure_class metric label to a closed set

ZachL111:
Both provider callbacks echo the provider's `error` parameter straight into a Prometheus label:

```python
failure_class=error,
```

`_log_auth_event` passes that into
`AUTH_FLOW_EVENTS.labels(provider=..., stage=..., outcome=..., failure_class=...)`, and
`AUTH_FLOW_EVENTS` is a module-level `Counter` whose labelnames include `failure_class`. The only
bounding on the way is `_failure_class`, which for a plain string does
`.strip().lower().replace(' ', '_')[:80]` — that truncates length, not the number of distinct
values.

`GET /v1/auth/callback/google` and `POST /v1/auth/callback/apple` are unauthenticated, correctly
so: they are the provider's redirect target and run before any Omi session exists. `error` is a
plain query/form parameter on those routes. `prometheus_client` never evicts a label tuple once
created, so each distinct `error` value becomes a permanent series for the life of the process.

That is the case `backend/AGENTS.md` rules out directly:

> **Prometheus labels**: static low-cardinality only (e.g. "pusher", "listen") — never uid/session_id.

Same defect, just carried by free-form provider text instead of a uid.

## The fix

Keep the observability that matters, drop the part that cannot be bounded. The seven RFC 6749
§4.1.2.1 codes a provider can legitimately echo pass through unchanged; anything else buckets to
`provider_error_other`.

```python
_OAUTH_ERROR_CODES = {
    "access_denied", "invalid_request", "invalid_scope", "unauthorized_client",
    "unsupported_response_type", "server_error", "temporarily_unavailable",
}

def _bounded_provider_error(error: str) -> str:
    normalized = error.strip().lower().replace(" ", "_")[:64]
    return normalized if normalized in _OAUTH_ERROR_CODES else "provider_error_other"
```

This mirrors the allowlist already used in `utils/translation_core/metrics.py`, and the
`safe_provider = provider if provider in {...} else "unknown"` line a few lines above in this same
function.

Nothing is lost for debugging: the `HTTPException` detail still carries the raw value, so the
specific string stays visible in the response and the request log. It just stops minting a metric
series.

## Tests

`tests/unit/test_auth_callback_failure_class_cardinality.py`:

- forty distinct attacker-chosen error values produce a small bounded number of series
- the same known error value repeated does not grow cardinality
- the seven known OAuth codes pass through unchanged
- arbitrary input maps into the closed set

## Verification

Run in `backend/`:

- `pytest tests/unit/test_auth_callback_failure_class_cardinality.py`: 5 passed
- prove-fail: with `routers/auth.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), the cardinality test fails with `AUTH_FLOW_EVENTS grew by 41
  distinct series for 40 distinct attacker-controlled error values` / `assert 41 <= 5`, while the
  repeated-known-value test passes both ways. Re-applied: 5 passed
- pytest together with `test_auth_redirect_uri.py` and `test_oauth_callback_uid_guard.py`: 65 passed
- the test imports `routers.auth` at module scope, so the router import is collection cost rather
  than per-test CPU; slowest test in the file is 0.02s and it clears the fast-unit duration guard
- `routers/auth.py` is not in the product file line-count ratchet baseline

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10037?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/auth.py
+++ b/backend/routers/auth.py
@@ -228,6 +228,26 @@ def _failure_class(error: Optional[object]) -> str:
     return value[:80] or error.__class__.__name__.lower()
 
 
+# RFC 6749 §4.1.2.1 error codes a provider may echo on the callback. The raw
+# `error` param is attacker-controlled free text and must never become a
+# Prometheus label value directly — that is unbounded cardinality on a
+# module-level (process-lifetime) metric registry.
+_OAUTH_ERROR_CODES = {
+    "access_denied",
+    "invalid_request",
+    "invalid_scope",
+    "unauthorized_client",
+    "unsupported_response_type",
+    "server_error",
+    "temporarily_unavailable",
+}
+
+
+def _bounded_provider_error(error: str) -> str:
+    normalized = error.strip().lower().replace(" ", "_")[:64]
+    return normalized if normalized in _OAUTH_ERROR_CODES else "provider_error_other"
+
+
 def _log_auth_event(
     *,
     provider: Optional[str],
@@ -369,7 +389,7 @@ async def auth_callback_google(
             stage="provider_callback_received",
             outcome="failed",
             auth_flow_id=auth_flow_id,
-            failure_class=error,
+            failure_class=_bounded_provider_error(error),
             status_code=400,
         )
         raise HTTPException(status_code=400, detail=f"Auth error: {error}")
@@ -458,7 +478,7 @@ async def auth_callback_apple_post(
             stage="provider_callback_received",
             outcome="failed",
             auth_flow_id=auth_flow_id,
-            failure_class=error,
+            failure_class=_bounded_provider_error(error),
             status_code=400,
         )
         raise HTTPException(status_code=400, detail=f"Auth error: {error}")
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Backend observability fix (bound OAuth failure_class metric cardinality) with test. Approve-only — touches auth.py (abuse-sensitive), Nik to land.

Git-on-my-level:
Thanks for the tight security/observability fix. I reviewed this as auth-sensitive because it changes the OAuth callback metric path.

Positive signal from automation: bounding the provider-controlled `error` value to the RFC OAuth error allowlist plus a single `provider_error_other` bucket addresses the unbounded Prometheus label-cardinality issue while preserving the useful provider-error distinctions. The regression coverage exercises the public callback path with many attacker-controlled values and checks the helper's closed output set, which is the right shape for this bug.

I am not formally approving because this touches OAuth/auth metrics and the automation approval gate is closed for security-sensitive paths; a human maintainer should make the final merge call. I do not see a blocker in the implementation from this review.

---

kodjima33 (approved):
Bound OAuth callback failure_class metric to closed set — cardinality bug fix w/ test

PR #10039, opened 2026-07-19
Route DELETE /v1/personas/{persona_id} to delete_persona instead of the thumbnail uploader

ZachL111:
The route decorator is stacked on the wrong function:

```python
@router.delete('/v1/personas/{persona_id}', tags=['v1'], response_model=AppThumbnailUploadResponse)
@router.post('/v1/app/thumbnails', tags=['v1'], response_model=AppThumbnailUploadResponse)
@max_part_size(APP_IMAGE_MAX_PART_SIZE)
async def upload_app_thumbnail_endpoint(file: UploadFile = File(...), uid: str = Depends(auth.get_current_user_uid)):
```

So `DELETE /v1/personas/{persona_id}` is served by the **thumbnail uploader**. Thirty-four lines
below, the function meant to serve it carries no decorator at all:

```python
def delete_persona(persona_id: str, secret_key: str = Header(...)):
```

## Three consequences, all live on main

1. **No persona can be deleted through the API.** The request reaches a handler that requires a
   multipart file body and then uploads a thumbnail. `delete_persona_db` is never called.
2. **`delete_persona` is unreachable dead code.**
3. **The admin gate went with it.** `delete_persona` checks `secret_key` against `ADMIN_KEY` and
   403s otherwise. The handler actually serving the route authenticates with a regular per-user
   Firebase uid, so the route's intended auth and its real auth disagree.

The `response_model` was wrong for the same reason — a delete documented as returning
`AppThumbnailUploadResponse`. It now uses `AppMutationResponse` (`{status: str}`), matching both
what `delete_persona` returns and what the sibling `DELETE /v1/apps/{app_id}` already declares.

Moving the decorator restores the intended dispatch without touching either handler's body. The
`POST /v1/app/thumbnails` route it was stacked on is unaffected.

The path is not present in `docs/api-reference/app-client-openapi.json`, so no published client
contract moves and no spec regeneration is needed.

## Tests

`tests/unit/test_persona_delete_route_dispatch.py` reads the registered route table off the router
and asserts which endpoint each path resolves to. These are route-table facts, not source-text
checks, so they fail if the decorator ever drifts onto the wrong function again:

- `DELETE /v1/personas/{persona_id}` resolves to `delete_persona`, and to exactly one route
- the served handler still takes `secret_key` and does not require a file upload
- `POST /v1/app/thumbnails` still resolves to `upload_app_thumbnail_endpoint`
- the thumbnail uploader serves no DELETE route at all

## Verification

Run in `backend/`:

- `pytest tests/unit/test_persona_delete_route_dispatch.py`: 4 passed
- prove-fail: with `routers/apps.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), three tests fail, the first with

  ```
  AssertionError: DELETE /v1/personas/{persona_id} resolves to upload_app_thumbnail_endpoint, not delete_persona
  MultipartMaxPartSizeRoute(path='/v1/personas/{persona_id}', name='upload_app_thumbnail_endpoint', methods=['DELETE'])
  AssertionError: upload_app_thumbnail_endpoint has no secret_key parameter
  assert 'secret_key' in {'file': <class 'fastapi.datastructures.UploadFile'>, 'uid': <class 'str'>}
  ```

  while the thumbnail POST test passes both ways. Re-applied: 4 passed
- pytest with `test_apps_marketplace_poison_guard.py` and `test_app_home_url_null_manifest.py`:
  11 passed
- `pytest tests/unit/test_route_policy_inventory.py`: 24 passed, so the route-table change does not
  disturb the policy inventory contract
- `pyright routers/apps.py`: 0 errors
- `routers/apps.py` is 2335 lines, exactly its ratchet baseline, so no baseline change

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10039?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/apps.py
+++ b/backend/routers/apps.py
@@ -2193,7 +2193,6 @@ def reject_app(app_id: str, uid: str, secret_key: str = Header(...)):
     return {'status': 'ok'}
 
 
-@router.delete('/v1/personas/{persona_id}', tags=['v1'], response_model=AppThumbnailUploadResponse)
 @router.post('/v1/app/thumbnails', tags=['v1'], response_model=AppThumbnailUploadResponse)
 @max_part_size(APP_IMAGE_MAX_PART_SIZE)
 async def upload_app_thumbnail_endpoint(file: UploadFile = File(...), uid: str = Depends(auth.get_current_user_uid)):
@@ -2227,6 +2226,7 @@ async def upload_app_thumbnail_endpoint(file: UploadFile = File(...), uid: str =
             os.remove(temp_path)
 
 
+@router.delete('/v1/personas/{persona_id}', tags=['v1'], response_model=AppMutationResponse)
 def delete_persona(persona_id: str, secret_key: str = Header(...)):
     if secret_key != os.getenv('ADMIN_KEY'):
         raise HTTPException(status_code=403, detail='You are not authorized to perform this action')
--- a/backend/tests/unit/test_persona_delete_route_dispatch.py
+++ b/backend/tests/unit/test_persona_delete_route_dispatch.py
@@ -1,0 +1,71 @@
+"""DELETE /v1/personas/{persona_id} must dispatch to delete_persona, not the thumbnail uploader.
+
+The route decorator was stacked on `upload_app_thumbnail_endpoint`:
+
+    @router.delete('/v1/personas/{persona_id}', ..., response_model=AppThumbnailUploadResponse)
+    @router.post('/v1/app/thumbnails', ..., response_model=AppThumbnailUploadResponse)
+    async def upload_app_thumbnail_endpoint(file: UploadFile = File(...), uid=Depends(...)):
+
+so the persona delete path was served by the thumbnail handler, which requires a multipart file
+body and authenticates with a regular Firebase uid. Meanwhile the real `delete_persona` below it
+carried no decorator at all and was unreachable, taking its admin `secret_key` gate with it.
+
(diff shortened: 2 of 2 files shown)


kodjima33 (approved):
Backend fix (route DELETE persona to delete_persona). Approve-only — Formatting CI red; author to fix lint.

Git-on-my-level (commented):
Thanks for the focused fix here. Moving the `DELETE /v1/personas/{persona_id}` decorator onto `delete_persona` looks like the right routing correction: it restores the intended admin `secret_key` gate, leaves the thumbnail upload route on `POST /v1/app/thumbnails`, and the new route-table regression tests cover the specific dispatch failure rather than just checking source text.

I’m leaving this as a positive maintainer signal rather than a formal approval because the touched surface is security-sensitive/admin deletion behavior and GitHub still shows a red Formatting check on the current PR. Before merge, a human maintainer should confirm the admin-delete contract is still desired and the Formatting check should be green or rerun if it is stale.

---

kodjima33 (approved):
Bug fix: DELETE /v1/personas routed to wrong handler; scoped, test added, CI green — 5/5

PR #10017, opened 2026-07-19
fix(dev-harness): resolve canonical Python venv

Git-on-my-level:
## Summary
- Centralize local dev-harness interpreter resolution: explicit `PYTHON`, then `backend/.venv/bin/python`, legacy `backend/venv/bin/python`, then `python3`.
- Route all harness commands and Makefile harness Python targets through the resolver; `dev-init` creates the canonical `.venv` when neither venv exists.
- Add focused resolver coverage for precedence and fallback behavior.

## Root cause and durable guard
Harness commands and Makefile targets had inconsistent hard-coded legacy `backend/venv` defaults. A single resolver owns precedence now, with a hermetic regression test covering every selection branch.

## Verification
- `/Users/dazheng/workspace/omi/backend/.venv/bin/python -m pytest -q scripts/dev-harness/tests/test_python_resolver.py` — 2 passed
- `bash -n` for the resolver plus all 9 harness consumers — pass
- `git diff --cached --check` — pass
- `make preflight` — passed (9 deterministic checks)

## Failure-Class
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10041, opened 2026-07-19
fix(dev-harness): prevent quote-root command injection

Git-on-my-level:
## Summary
- Prevent Make's Python resolver expansion from interpolating the checkout root into `bash -c` source text.
- Resolve the harness script path inside Bash via `git rev-parse --show-toplevel` instead.
- Add a regression test using a checkout name containing a single quote and controlled `touch injected-marker` payload; it confirms no marker runs and the canonical `backend/.venv/bin/python` still handles the Make target.

PR #10017 embedded `$(ROOT)` inside an outer single-quoted `bash -c` program. A single quote in the checkout root terminated that shell quoting and allowed subsequent path text to execute as shell syntax. The Makefile no longer interpolates the root into the shell program; the regression executes the real Make target in a maliciously named temporary Git checkout.

## Red/green proof
- **Red (before fix):** the controlled reproduction against `c16c21d1dbb0063e5f200374d808dad779a9bf3b` returned `marker_exists=True` after `make -n list-memory-scenarios` in a checkout named with `'; touch …; #`.
- **Green:** the new focused regression passes with no marker and records the resolved canonical venv invocation.

## Verification
- `env -u PYTHON python3 -m pytest scripts/dev-harness/tests/test_python_resolver.py -q` (4 passed)
- `env -u PYTHON make preflight`
- `bash -n scripts/dev-harness/_resolve_python.sh`
- quote-root Make dry run and `git diff --check`

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10041?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-shell-source-path-interpolation.json
+++ b/.github/failure-classes/FC-shell-source-path-interpolation.json
@@ -1,0 +1,13 @@
+{
+  "schema_version": 1,
+  "id": "FC-shell-source-path-interpolation",
+  "violated_contract": "A repository checkout path must never be interpolated into shell program text where path characters can change shell syntax or execute commands.",
+  "canonical_prevention": "Derive the resolver path inside the shell program or pass paths as shell data, and exercise the real Make target from a quote-containing temporary checkout with a controlled marker command.",
+  "evidence_prs": [10017],
+  "scope_hints": [
+    "Makefile",
+    "scripts/dev-harness/_resolve_python.sh",
+    "scripts/dev-harness/tests/test_python_resolver.py"
+  ],
+  "status": "open"
+}
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,10 @@
 HOOKS_DIR := $(shell git rev-parse --git-path hooks)
 PYTHON ?= $(shell bash -c 'source "$$(git rev-parse --show-toplevel)/scripts/dev-harness/_resolve_python.sh"; dev_harness_python')
+# Export so recipes use $$PYTHON (shell variable expansion) instead of $(PYTHON)
+# (Make text interpolation). Shell variable expansion treats the resolved path
+# as data and cannot be broken by quote or command-substitution characters in
+# the checkout root, unlike Make interpolation into recipe shell text.
+export PYTHON
 DESKTOP_USER ?= alice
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
**Monitor run summary — addressed Codex P2 review feedback**

The P2 thread (`Shell-quote the resolved interpreter path`) flagged that a double quote in the checkout root could break recipe double-quoting via Make's `$(PYTHON)` text interpolation. The fix closes that vector:

- **Root cause:** Make expands `$(PYTHON)` as literal text into recipe shell commands. A `"` in the resolved path escapes the recipe's surrounding `"..."` quoting, allowing injected shell syntax to execute.
- **Fix:** `export PYTHON` and switch all recipe expansions from `$(PYTHON)` (Make interpolation) to `$$PYTHON` (shell variable expansion). Shell variable expansion treats the value as data — quote and command-substitution characters in the path cannot change recipe syntax. No `printf %q` escaping is needed (it would double-escape spaces when combined with `"$$PYTHON"`).
- **Regression test:** Added `test_make_harness_does_not_execute_double_quote_in_checkout_name`, covering a checkout named `omi "; touch double-quote-marker; #`, alongside the existing single-quote test.
- **Failure-class:** Added `FC-shell-source-path-interpolation` registry entry; PR body declares `Failure-Class: new`.

**Verification (all green):**
- `env -u PYTHON python3 -m pytest scripts/dev-harness/tests/test_python_resolver.py -q` — 5 passed (4 original + 1 new double-quote regression)
- `bash -n scripts/dev-harness/_resolve_python.sh`
- `git diff --check`
- Full pre-push manifest gate (9 checks) against `origin/main`
- CI: all checks pass (Detect Changes, Hygiene, source-closure, image-smoke variants)

The Codex review thread has been resolved as addressed.

kodjima33 (approved):

Issue #9736, opened 2026-07-14
GitHub Issue Sync targets removed Backlog project status


tianmind-studio:
## Problem

The `GitHub Issue Sync` workflow fails whenever it tries to add a newly opened issue to organization project 1.

Recent examples:

- Run 29356714415 for #9729
- Run 29355199537 for #9724
- Run 29354706619 for #9722
- Run 29350605309 for #9717

Each run reaches `paritytech/github-issue-sync@master`, then fails with:

```text
Error: Project value 'Backlog' does not exist.
Available values are ["Someday","Next Week","TODO (This week)","In progress","In review","Done"]
```

## Root cause

`.github/workflows/main.yml` still configures `project_value: Backlog`, but project 1 no longer has that Status option. The workflow history shows that new items were deliberately moved from `To do` to `Backlog` in 2024, so `Someday` is the closest current status preserving the backlog semantics; `TODO (This week)` would incorrectly promote every synced item into the active-week queue.

## Proposed fix

Change the configured Status value from `Backlog` to `Someday`. Keep the project, field, event triggers, labels, tokens, and action behavior unchanged.

## Acceptance criteria

- The workflow no longer requests a non-existent project Status.
- A qualifying issue can be added to project 1 with Status `Someday`.
- No token, permission, project number, or issue-label behavior changes.

PR #9740, opened 2026-07-14
fix(ci): restore GitHub issue project sync

tianmind-studio:
## What changed and why

Replace the removed `Backlog` value in the GitHub Issue Sync workflow with the project's current backlog-equivalent `Someday` status. The action rejects values that are not present in the Project V2 single-select field, so every qualifying issue sync currently fails before adding the item. Closes #9736.

## Product invariants affected

none

## How it was verified

- Inspected failed GitHub Issue Sync run `29356714415`, which reports that `Backlog` does not exist and lists `Someday`, `Next Week`, `TODO (This week)`, `In progress`, `In review`, and `Done` as the available values.
- Confirmed from workflow history that `Backlog` replaced the original to-do status specifically to route synced issues into the backlog, making `Someday` the current semantic equivalent.
- `actionlint -shellcheck= .github/workflows/main.yml`
- Passed diff hygiene, architecture guardrails, product invariants, desktop changelog, deferred-work marker, lifecycle header, version-prefixed filename, and backend workflow contract checks.
- The deployment secret-boundary check passes when repository-relative paths are normalized. Its native Windows invocation currently false-fails even with `--base HEAD` because current paths use backslashes while Git base paths use forward slashes; tracked separately in #9738.

## Tests

This is a one-value workflow configuration repair. Static workflow validation and all applicable deterministic checks pass apart from the isolated native Windows path-comparison defect described above. A live Project V2 mutation cannot be exercised from a fork because the upstream project token is intentionally unavailable there.

## Root cause and durable guard (bug fixes)

The configured Project V2 status was renamed or removed while the workflow retained the exact string `Backlog`. The action validates single-select values by name, so the stale value causes a hard failure. Routing to the existing `Someday` option restores the intended backlog placement without changing triggers, permissions, labels, project selection, or active-work statuses.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9740?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10042, opened 2026-07-19
fix(ci): sync new issues to project Status 'Someday' not removed 'Backlog'

kodjima33:
## What
Change the `GitHub Issue Sync` workflow (`.github/workflows/main.yml`) to assign newly opened `task`-labeled issues to org project 1 with Status **`Someday`** instead of the removed **`Backlog`** option.

## Why
Org project 1 no longer has a `Backlog` Status option, so every `task`-labeled sync fails at the fetch-project-values step. Issue #9736 evidence:

> `Error: Project value 'Backlog' does not exist. Available values are ["Someday","Next Week","TODO (This week)","In progress","In review","Done"]`

`Someday` is the current option that preserves the backlog semantics (items were historically routed to `Backlog`, not the active-week queue); `TODO (This week)` would wrongly promote every synced item into the active week. Nothing else — project number, field, token, event triggers, or label filter — changes.

## Verified
- Reproduced live: run **29678923243** (2026-07-19, `task`-labeled issue) failed with the exact error above; runs on non-`task` issues pass only because the action skips them (`Label 'bug' does not match ... Skipping`).
- Confirmed the current file still had `project_value: Backlog`.
- Confirmed `Someday` is a valid current Status from the CI error's own "Available values" list.
- `python3 -c yaml.safe_load` parses the edited workflow cleanly; single-line change, no other keys touched.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10042?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
           project_field: Status
           # Optional unless that project_field was set up. Then this field is required.
           # The value to modify in the project field
-          project_value: Backlog
+          project_value: Someday
           # Optional, labels to work with. Read below to see how to configure it.
           # If this value is set, the action will be applied only to issues with such label(s).
           labels: |


undivisible:
already landed on main from PR #9740 and 013ca09308401386fb05efac8c49e45342cc36a2 thank you :)

Issue #9822, opened 2026-07-15
Failure-class protocol and recurrence ratchets


Git-on-my-level:
This issue is narrowed to one self-contained, in-repository change.

The problem is that repeated defect classes are recorded only in PR prose, so agents cannot reliably reuse the known prevention primitive. Two real static failure classes have also recurred despite ad hoc review: unsafe Swift dictionary construction (#6506, #9288) and naive Python datetime sort sentinels (#9571, #9600).

Scope

- Provide a semantic failure-class registry under `.github/failure-classes/` and an offline, JSON-first CLI that prepares, explains, validates, and reports declarations.
- Require and validate `Failure-Class: FC-<slug> | new | none` for reactive `fix:` commits; distinguish proactive `harden:` commits.
- Add the two manifest-backed recurrence ratchets above, with real-incident citations.
- Update the PR template and contributor guidance so an existing-class declaration replaces repetitive root-cause narrative; keep extended narrative only for a new class or registry lifecycle transition.
- Require seven days of unchanged behavior and guard before a product invariant can be locked.

Out of scope

- GitHub branch-protection changes, watchdog auto-merge behavior, external watchdog runtime work, scheduled automations, and automatic semantic classification.
- Broad subsystem migrations or shared-primitive work that needs its own issue.

Success is a single PR that adds the offline protocol, its focused tests, and manifest-backed ratchets. No external setting or runtime prerequisite is required to close this narrowed issue.

Evidence: the audit chains include #9734 → #9748, #9522 → #9533, #9739 → #9783, #6506 → #9288, and #9571 → #9600.

Git-on-my-level:
Correction to item 3 (Dictionary(uniqueKeysWithValues:) ratchet): a verification pass found this guard **already exists** — `desktop/macos/scripts/check_desktop_test_quality.py` forbids `Dictionary(uniqueKeysWithValues:)` / `.init(uniqueKeysWithValues:)` at the AST/token level, with the `// omi-collection-safety: static-unique-keys` escape hatch documented in `desktop/macos/AGENTS.md`, and it is manifest-registered. Item 3 reduces to: verify its coverage includes production sources (not only test-quality scope) and close. Also a mechanical note for items 3/4: the checks-manifest loader (`.github/scripts/run_checks.py`) consumes `id`/`command`/`triggers`/`lanes`/`reason` only — real-instance citations go in `reason`, there is no dedicated `caught` field unless the schema is extended.

[...]

Issue #9135, opened 2026-07-06
desktop-backend Gemini proxy returns persistent 504s while /health stays green


Git-on-my-level:
## Summary

Production `desktop-backend` is returning persistent gateway timeouts on the Gemini proxy endpoint while `/health` remains green.

This is a capability-specific prod degradation: public liveness checks pass, but a real desktop AI path is failing repeatedly.

## Evidence

Observed during read-only prod health checks on 2026-07-06 UTC:

- Public health endpoints all returned `200`:
  - `https://api.omi.me/v1/health`
  - `https://api.omiapi.com/v1/health`
  - `https://agent.omi.me/health`
  - `https://agent.omiapi.com/health`
- `desktop-backend` Cloud Logging errors:
  - last 10m: 13x `504` on `/v1/proxy/gemini/models/gemini-2.5-flash:generateContent`
  - last 30m: 37x `504` on the same endpoint
  - last 2h: 151x `504` on the same endpoint
- Related but separate noise in the same service:
  - last 2h: 24x `500` on `/v1/webhooks/sentry/poll`
  - last 2h: 1x `503` on `/v1/tts/synthesize`

PostHog guardrails showed chat completions still happening, so this does not look like a total chat outage, but the Gemini proxy path is actively degraded.

## Expected behavior

The desktop Gemini proxy should either succeed or return a bounded, typed error that distinguishes:

- upstream timeout
- upstream rate limit / quota
- auth / BYOK failure
- client cancellation
- internal proxy failure

It should not repeatedly surface as generic gateway timeouts without a capability-level alert.

## Acceptance criteria

- [ ] Add/log structured error classification for Gemini proxy failures.
- [ ] Add a metric/dashboard/alert for desktop Gemini proxy `5xx`/`504` rate.
- [ ] Add a synthetic or capability health check for Gemini proxy behavior, separate from `/health`.
- [ ] Return a bounded, typed user-facing failure where possible instead of an opaque gateway timeout.
- [ ] Confirm the fix with Cloud Logging showing the Gemini proxy `504` bucket dropping to near-zero over a comparable window.

## Priority

P1: current prod capability degradation that `/health` does not catch.

PR #8911, opened 2026-07-03
Fix desktop Gemini proxy timeout observability

tianmind-studio:
Fixes #8906.

## Summary
- reduce the non-streaming Gemini proxy upstream deadline from 120s to 90s so the Rust desktop backend returns a proxy-owned retryable timeout before the observed Cloud Run boundary
- log non-streaming Gemini upstream failures through the same aggregate `upstream_result` shape as successful upstream responses, including provider/model/action/payload bucket/duration bucket/phase/status
- avoid adding request bodies, provider responses, tokens, prompts, or user IDs to the aggregate upstream result log line
- add a desktop changelog fragment for the faster retryable timeout behavior

## Testing
- `git diff --check`
- PowerShell JSON parse for `desktop/macos/changelog/unreleased/20260703-gemini-proxy-timeout.json`
- `PRE_PUSH_SKIP_DESKTOP_RUST=1 scripts/pre-push` -> passed non-Rust gates; desktop changelog fragment found

## Note
This Windows machine does not have `cargo`/`rustfmt` installed, so I could not run the local desktop Rust checks. The GitHub `Lint & Format Check` desktop Rust job should run `rustfmt --check` and `cargo check --locked` for this PR.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/8911?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10054, opened 2026-07-19
fix(chat): FC-4 request deadline budget for the Anthropic chat path

### LINKED ISSUE #9835: FC-4: request deadline budget for the Anthropic chat path — stop fixing per-hop timeouts retail
## Problem

The provider timeout class has been fixed four times, each time by moving a *per-hop* timeout: #8640, #8911, #9135 ("stop cutting off long-context Gemini calls at 90s"), #9644 ("restore Gemini platform deadline"). #9533's own body names the root cause: "those fixes only moved a per-attempt timeout to the current infrastructure boundary." This is registry class **FC-4** (docs/product/failure-classes.md, #9822).

Current state, precisely:

- The **Gemini proxy path already has a request-level deadline** (240s absolute inside the 300s platform bound, `routes/proxy.rs:54`) — the class is closed there, ad hoc.
- The **Anthropic chat path (`routes/chat_completions.rs`) does not.** Pre-first-byte work — auth/JWK refresh (`auth.rs`, 5s per wait), paywall/BYOK lookups in the `PaywalledAuthUser` extractor, body extraction, Anthropic retries, and up to three server-tool continuations — has no shared budget; each stage owns a constant (10s connect at :847, 120s non-streaming total at :929).
- A related live gap: the stream idle timer resets on **every raw upstream byte** (`chat_completions.rs:1287`), and Anthropic `ping` events emit nothing downstream (`:1535`) — a provider can ping indefinitely while the client has received no first byte and nothing times out.

What is *already correct* and must not change: streaming has no total response timeout by design (`chat_completions.rs:101` — "a long answer is not an error"). #9135 was exactly the bug of a total timeout cutting a producing stream; the budget must never reintroduce it.

## Fix: a `RequestDeadline` type on the Anthropic chat path, one PR

**1. Admission in middleware, not the handler.** `RequestDeadline` (wrapping `tokio::time::Instant`; API: `remaining()`, `derive(cap) -> Duration`, `expired()`) is created in route middleware **before extractors and body extraction** and stored in request extensions — the `PaywalledAuthUser` extractor's auth/paywall waits are inside the budget (their per-request waits become `min(stage cap, remaining)`).

**2. Budget semantics.** The deadline governs everything up to the **first client-visible semantic SSE event** (`message_start`, text delta, or tool delta) — not provider headers, pings, or keepalives, which do not count as progress. This includes retries and server-tool continuation round-trips that occur before that first visible event. After the first visible event: the budget is done; a separate **semantic-progress** idle timer (visible-event-based, replacing the raw-byte reset at :1287 — closing the infinite-ping gap) governs stalls. Public-web pause-turn synthesis is pre-first-visible-byte work and stays budgeted. Retry/continuation attempts start only if `remaining()` exceeds a floor; backoff sleeps are truncated at the budget edge.

**3. `Deadline` is for request-owned awaits only.** Policy clocks are explicitly out: the unknown-kid *cooldown* (`auth.rs:21` — the wait may be `min(5s, remaining)`, the shared cooldown stays independent), cache TTLs (BYOK/paywall/rate-limit), the idle-timer policy value, and startup retries (`main.rs:144` is startup-only, not a request seam). Detached work spawned during a request (e.g. the Firestore usage write on `message_stop`, `:1514`) must **never** inherit the request deadline — it outlives the response and uses its own bounded background policy.

**4. Enforcement.** Provider sends go through a narrow `send_with_deadline` helper taking `&RequestDeadline`; a diff-scoped checker (manifest-registered, `local` + `ci` lanes) flags direct outbound sends in `routes/` bypassing it — signatures alone don't cover raw `reqwest` builders. Per #9822's rule, the checker cites #8640/#8911/#9135/#9644 as the caught class.

**5. Failure surface, stated honestly.** Before the response starts: a typed HTTP timeout error, returned inside a *verified* platform bound (deployment-config verification is a proof item, not an assumption — the 300s figure is code-side and Cloud Run applies its request deadline to streaming connections too). After HTTP 200: only a terminal SSE error event is possible (as today on idle failure); clients must still handle platform/client disconnects. This shrinks the `:1612` spinning-bubble class to post-response platform kills; it does not make it "impossible."

## Proof

- Behavioral tests with `tokio::time::pause` and a scripted fake upstream: (a) a retry/continuation chain stops at the budget floor rather than starting an attempt it can't finish; (b) backoff sleep truncated at budget edge; (c) a stream emitting visible deltas survives budget expiry; (d) ping-only upstream before `message_start` hits the budget → typed error (the live gap above); (e) post-first-event stall dies from the semantic idle timer independent of budget; (f) the spawned usage write completes under its own policy after the request deadline expires.
- Test-file header cites #8640/#8911/#9135/#9644 as the caught class.
- Deployed platform timeout config verified and recorded in the PR (not assumed from code constants).
- FC-4 registry clock started at merge.

## Cut list

- **Gemini proxy unification is a named follow-up**, not this PR: `proxy.rs` already has a working ad hoc deadline; migrating it onto `RequestDeadline` is mechanical once the type exists, and bundling it here makes the PR unreviewable.
- No retry-policy redesign (counts, backoff shape) — only budget-awareness of the existing policy.
- No changes to policy clocks: cooldowns, cache TTLs, startup retries.
- No per-user/per-model configurable budgets — one admission constant per route class until evidence demands more.
- No streaming total timeout — explicitly preserved as absent, by design.

## Sizing

One PR, honestly medium-large (this is a behavioral change, not an 80-line type): middleware + extensions plumbing, extractor budget coverage, semantic-progress redefinition of the idle timer, `send_with_deadline` seam + checker, deterministic fake-upstream test module, registry update. All within `routes/chat_completions.rs`, `auth.rs`, and one new module.

## Sequencing

Lands **before** the `chat_completions.rs` split (companion issue): this is the behavioral change; the split is then a pure move-code refactor with no behavior change hiding inside it.

aryanorastar:
## Summary

- Adds a `RequestDeadline` budget (240s inside the 300s platform bound, mirroring the Gemini proxy's ad hoc deadline) admitted in route middleware **before extractors and body extraction** on `/v2/chat/completions`, stored in request extensions.
- Auth, BYOK-state, and paywall waits inside `PaywalledAuthUser` are bounded by the budget on this route; exhaustion returns a typed `504` (`request_deadline_exceeded`) instead of a misleading 401. Routes without the middleware are untouched.
- All chat provider sends go through a narrow `send_with_deadline` seam. The retry loop becomes a budget-aware policy (`retry_with_budget`): an attempt never starts on an expired budget, a retry or `pause_turn` continuation starts only above a 10s floor (the connect bound), and backoff sleeps truncate at the budget edge. A transient status on the final available attempt still passes the upstream error through unchanged.
- The budget governs everything up to the **first client-visible semantic SSE event**. The stream idle timer now measures semantic progress: raw bytes and Anthropic `ping` events no longer reset it — closing the live gap where a ping-only upstream held a client forever with nothing visible. Pre-first-event budget exhaustion emits a typed terminal SSE `upstream_timeout` error (HTTP 200 is already on the wire).
- **Explicitly preserved:** streaming has no total response timeout (#9135 — a long answer is not an error); a stream emitting visible deltas survives budget expiry, proven by test.
- The detached Firestore usage write on `message_stop` runs through a spawn seam that never inherits the request deadline.
- A manifest-registered static tripwire (labeled as such) rejects new direct `.send()` calls in `routes/chat/` that bypass the seam (`local` + `ci` lanes).

Fixes #9835.

## Root cause

Registry class **FC-per-hop-timeout** (#8640, #8911, #9135, #9644): each fix moved a per-attempt timeout to the current infrastructure boundary because no request-level budget existed on the Anthropic chat path. Pre-first-byte work — auth/JWK waits, paywall lookups, retries, up to three server-tool continuations — each owned a private constant with no shared clock, and the stream idle timer reset on every raw upstream byte while `ping` events emitted nothing downstream, so a pinging provider could hold a request indefinitely with zero client-visible progress.

## Failure class

Failure-Class: FC-per-hop-timeout

## Durable guard

- `src/request_deadline.rs` owns the budget type (`remaining()` / `derive(cap)` / `expired()` on the tokio clock, so tests drive it with `tokio::time::pause`), the admission middleware, and the `send_with_deadline` seam.
- Policy clocks are explicitly out of the budget: the unknown-kid refresh cooldown, cache TTLs, the idle-timer policy value, and startup retries stay independent (documented in the module).
- Enforcement is layered per #9822's rule: the behavioral `tokio::time::pause` tests execute the real retry policy and the real stream translator; the diff-scoped checker is a static tripwire for the one thing signatures can't see (raw reqwest builders), citing the caught class.

## Validation

- `cargo test --locked` (Backend-Rust) → **381 passed** (92 in the chat module, including the six new deadline tests and the existing byte-chunk→SSE characterization tests unchanged in their assertions).
- New behavioral tests, all under `tokio::time::pause` (test-file header cites #8640/#8911/#9135/#9644):
  - (a) `retry_chain_stops_at_budget_floor` — with 7s remaining against the 10s floor, a transient error passes through after one attempt instead of starting an attempt it can't finish.
  - (b) `retry_backoff_truncates_at_budget_edge` — 250ms backoff, 100ms budget: exactly 100ms elapses, no second attempt.
  - (c) `visible_stream_survives_budget_expiry` — visible deltas continue 200s past a 5s budget and finish normally; no `upstream_timeout` anywhere.
  - (d) `ping_only_stream_hits_budget_with_typed_error` — ping-only upstream ends at the 5s budget edge with a typed terminal SSE error (the live gap this closes).
  - (e) `post_first_event_stall_dies_from_semantic_idle_timer` — with a 100,000s budget, a post-first-event ping-only stall dies at the 60s semantic idle bound; pings do not reset it.
  - (f) `detached_usage_write_completes_after_deadline_expiry` — the spawn seam completes under its own policy long after the budget expired.
- `cargo check --locked` clean; `cargo fmt --check` clean. (`cargo clippy` is red on current `main` — 271 pre-existing `unwrap` denials in tests; this PR adds no new clippy class and clippy is not a repo lane.)
- `python3 .github/scripts/check_chat_send_deadline.py` → clean on the new module; its fixtures (`test_check_chat_send_deadline.py`) → 3 passed.
- `python3 .github/scripts/test_run_checks.py` → 20 passed (manifest contract with the two new entries).

**Deployed platform timeout config:** not verifiable from this environment (external contributor, no production access). The 300s figure is the code-side constant `PLATFORM_REQUEST_TIMEOUT`, consistent with `proxy.rs`'s `CLOUD_RUN_REQUEST_TIMEOUT`; per the issue's proof list, the deployed Cloud Run value (including its behavior on streaming connections) needs maintainer verification and recording before the FC-4 registry clock starts.

Not exercised live: a real Anthropic streaming turn through a dev build (no API keys in this environment). The paused-time tests drive the real production stream translator and retry policy through the same seams production uses.

## Cut list (per the issue)

- Gemini proxy unification onto `RequestDeadline` is the named follow-up; `proxy.rs` keeps its working ad hoc deadline.
- No retry-policy redesign — counts and backoff shape unchanged, only budget-awareness.
- No per-user/per-model budgets; one admission constant for the route class.
- No streaming total timeout — explicitly preserved as absent.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10054?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -91,6 +91,16 @@ checks:
     triggers: [".github/scripts/check-desktop-changelog.py", ".github/scripts/desktop-changelog.py", ".github/scripts/test_desktop_changelog.py", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "#9717: changelog tooling must preserve UTF-8 I/O and exempt only internal desktop release controls"
+  - id: chat-send-deadline
+    command: ["python3", ".github/scripts/check_chat_send_deadline.py"]
+    triggers: ["desktop/macos/Backend-Rust/src/routes/chat/**", "desktop/macos/Backend-Rust/src/request_deadline.rs", ".github/scripts/check_chat_send_deadline.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "#9835 (FC-per-hop-timeout): chat provider sends must consume the request budget via send_with_deadline; caught class #8640/#8911/#9135/#9644"
+  - id: chat-send-deadline-tests
+    command: ["python3", ".github/scripts/test_check_chat_send_deadline.py"]
+    triggers: [".github/scripts/check_chat_send_deadline.py", ".github/scripts/test_check_chat_send_deadline.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "chat send-deadline tripwire must keep rejecting direct sends and accepting the budgeted seam"
   - id: agents-md-lean
     command: ["python3", ".github/scripts/check_agents_md_lean.py"]
     triggers: ["AGENTS.md", "**/AGENTS.md", ".github/scripts/check_agents_md_lean.py"]
--- a/.github/failure-classes/FC-per-hop-timeout.json
+++ b/.github/failure-classes/FC-per-hop-timeout.json
@@ -3,6 +3,7 @@
   "id": "FC-per-hop-timeout",
   "violated_contract": "A multi-hop request must consume one request-level deadline budget rather than receive a full timeout at each hop.",
   "canonical_prevention": "Compute the remaining request deadline before every downstream operation and fail when the shared budget is exhausted.",
+  "canonical_prevention_artifact": [".github/scripts/check_chat_send_deadline.py"],
   "evidence_prs": [9533, 9644],
   "scope_hints": ["backend/**", "desktop/macos/**"],
   "status": "open"
--- a/.github/scripts/check_chat_send_deadline.py
+++ b/.github/scripts/check_chat_send_deadline.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+"""Direct provider sends in routes/chat/ must go through send_with_deadline.
+
+Caught class (FC-per-hop-timeout): #8640, #8911, #9135, and #9644 each moved a
+per-hop timeout because a raw send owned its own clock. The request budget seam
+in `desktop/macos/Backend-Rust/src/request_deadline.rs` is the single owner of
+outbound provider waits on the Anthropic chat path (#9835). Signatures alone do
+not cover raw reqwest builders, so this STATIC TRIPWIRE rejects new direct
+`.send()` calls in the chat module; behavioral coverage lives in the
+`tokio::time::pause` deadline tests in `routes/chat/tests/all.rs`.
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[2]
+CHAT_DIR = ROOT / "desktop" / "macos" / "Backend-Rust" / "src" / "routes" / "chat"
+DIRECT_SEND_RE = re.compile(r"\.send\(\)")
+
+
+def find_direct_sends(chat_dir: Path) -> list[str]:
+    violations: list[str] = []
+    for path in sorted(chat_dir.rglob("*.rs")):
+        for lineno, line in enumerate(path.read_text(encoding="utf-8").splitlines(), start=1):
+            if DIRECT_SEND_RE.search(line):
+                try:
+                    shown = path.relative_to(ROOT)
+                except ValueError:
+                    shown = path
+                violations.append(f"{shown}:{lineno}: direct `.send()` bypasses send_with_deadline")
+    return violations
+
+
+def main() -> int:
+    violations = find_direct_sends(CHAT_DIR)
+    if violations:
+        print("\n".join(violations), file=sys.stderr)
+        print(
+            "Chat provider sends must consume the request budget via "
+            "crate::request_deadline::send_with_deadline (#9835, FC-per-hop-timeout).",
+            file=sys.stderr,
+        )
+        return 1
+    print("chat send-deadline contract: no direct provider sends in routes/chat/")
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/.github/scripts/test_check_chat_send_deadline.py
+++ b/.github/scripts/test_check_chat_send_deadline.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+"""Fixtures for the chat send-deadline static tripwire (#9835)."""
+
+from __future__ import annotations
+
+import tempfile
+import unittest
+from pathlib import Path
+
+from check_chat_send_deadline import CHAT_DIR, find_direct_sends
+
+
+class ChatSendDeadlineChecker(unittest.TestCase):
+    def test_flags_a_direct_send(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            chat_dir = Path(tmp)
+            (chat_dir / "transport.rs").write_text(
+                "let resp = builder.send().await;\n", encoding="utf-8"
+            )
+            violations = find_direct_sends(chat_dir)
+        self.assertEqual(len(violations), 1)
+        self.assertIn("transport.rs:1", violations[0])
+
+    def test_accepts_the_budgeted_seam(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            chat_dir = Path(tmp)
+            (chat_dir / "transport.rs").write_text(
+                "let resp = send_with_deadline(builder, deadline).await;\n",
+                encoding="utf-8",
+            )
+            self.assertEqual(find_direct_sends(chat_dir), [])
+
+    def test_real_chat_module_is_clean(self) -> None:
+        self.assertTrue(CHAT_DIR.is_dir(), f"{CHAT_DIR} moved; update the checker")
+        self.assertEqual(find_direct_sends(CHAT_DIR), [])
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/desktop/macos/Backend-Rust/Cargo.toml
+++ b/desktop/macos/Backend-Rust/Cargo.toml
@@ -57,3 +57,5 @@ gcp_auth = "0.12"
 
 [dev-dependencies]
 tower = { version = "0.5", features = ["util"] }
+# test-util enables tokio::time::pause/advance for deterministic deadline tests.
+tokio = { version = "1", features = ["full", "test-util"] }
--- a/desktop/macos/Backend-Rust/src/auth.rs
+++ b/desktop/macos/Backend-Rust/src/auth.rs
@@ -98,6 +98,8 @@ impl IntoResponse for AuthError {
             StatusCode::FORBIDDEN
         } else if self.error == "jwks_refresh_failed" {
             StatusCode::SERVICE_UNAVAILABLE
+        } else if self.error == "request_deadline_exceeded" {
+            StatusCode::GATEWAY_TIMEOUT
         } else {
             StatusCode::UNAUTHORIZED
         };
@@ -480,6 +482,29 @@ pub struct PaywalledAuthUser {
     pub byok_stripped: bool,
 }
 
+/// Bound one extractor stage by the request budget when the route admitted one
+/// (#9835). Routes without the deadline middleware keep today's behavior; the
+/// unknown-kid refresh cooldown and cache TTLs are policy clocks and stay
+/// independent — only this request's wait is bounded.
+async fn within_request_deadline<T, F>(
+    deadline: Option<&crate::request_deadline::RequestDeadline>,
+    stage: &str,
+    future: F,
+) -> Result<T, AuthError>
+where
+    F: std::future::Future<Output = T>,
+{
+    match deadline {
+        None => Ok(future.await),
+        Some(d) => tokio::time::timeout(d.remaining(), future)
+            .await
+            .map_err(|_| AuthError {
+                error: "request_deadline_exceeded".to_string(),
+                message: format!("Request deadline budget exhausted during {stage}"),
+            }),
+    }
+}
+
 #[async_trait]
 impl<S> FromRequestParts<S> for PaywalledAuthUser
 where
@@ -488,6 +513,10 @@ where
     type Rejection = AuthError;
 
     async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let deadline = parts
+            .extensions
+            .get::<crate::request_deadline::RequestDeadline>()
+            .copied();
         // Get + extract bearer token
         let auth_header = parts
             .headers
@@ -514,7 +543,12 @@ where
                 message: "Firebase auth not configured".to_string(),
             })?;
 
-        let (uid, name, email) = firebase_auth.0.verify_token(token).await?;
+        let (uid, name, email) = within_request_deadline(
+            deadline.as_ref(),
+            "token verification",
+            firebase_auth.0.verify_token(token),
+        )
+        .await??;
 
         // BYOK fingerprint validation (issue #7357).
         // Validates SHA-256 fingerprints against Firestore enrollment.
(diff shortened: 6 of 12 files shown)


Git-on-my-level (commented):
Thanks for the careful write-up here — the shape of the fix makes sense to me: a single request-level budget admitted before auth/body extraction, a narrow `send_with_deadline` seam for Anthropic sends, preserving no total timeout after visible streaming progress, and a CI tripwire to prevent future direct sends all line up with the FC-per-hop-timeout class this is trying to close.

A few maintainer notes before merge:

- This touches auth/paywall timing, Anthropic provider transport, CI enforcement, and a new dev-dependency feature, so I’m not formally approving it from automation. A human maintainer should review the runtime policy and deployment assumption (notably the 240s budget under a presumed 300s Cloud Run bound) before merge.
- The added Python tripwire and its unit tests pass in my review environment, and the static scan finds no direct `.send()` calls under `routes/chat/` after the change.
- I could not run the Rust chat tests here because this review environment does not have `cargo` installed, so please require the targeted Backend-Rust tests/checks in CI or a maintainer environment before merging.

Overall this looks like a cohesive and well-tested fix direction; the remaining review is mainly human confidence around the timeout contract and release/platform configuration.

---

aryanorastar:
@Git-on-my-level Noted. We'll wait for a human maintainer to review the runtime policy.

aryanorastar:
Thanks. Agreed the one runtime assumption worth a maintainer eye is the 240s request budget under the presumed 300s Cloud Run bound. The Rust chat tests pass locally (`cargo test`), and the Python `send_with_deadline` tripwire + its unit tests are green in CI on this head.

kodjima33 (approved):
Anthropic chat deadline budget — approved; not auto-merged (large backend change)

Git-on-my-level (dismissed):
Thanks for the follow-up on the deadline budget and CI tripwire. I found one blocking edge case in the streaming path:

- `ThinkingDelta` chunks are emitted to clients as `reasoning_content`, but they do not refresh the new semantic-progress timer. After the initial role chunk, a long reasoning phase that keeps sending visible `reasoning_content` can still be terminated once `STREAM_IDLE_TIMEOUT` elapses, even though the client is receiving semantic output. That regresses the intended “long visible streaming is not an error” contract for Anthropic/Claude reasoning streams. Please update the timer when a reasoning chunk is yielded, and add/adjust a paused-time streaming test that advances beyond 60s while reasoning deltas are flowing.

I verified the new Python tripwire locally (`test_check_chat_send_deadline.py` and `check_chat_send_deadline.py` both pass), but I still can’t run the Rust chat tests in this environment because `cargo` is not installed. This also continues to need human maintainer review for the auth/paywall/provider timeout policy, CI workflow enforcement, and the 240s-under-300s platform assumption; I’m not formally approving from automation.

---

aryanorastar:
Good catch — that was a real regression in my change, and you called the mechanism exactly right. Fixed.

`ThinkingDelta` yields `reasoning_content` to the client but was the **only** client-visible chunk that didn't refresh `first_visible_at`. I checked every yield site rather than just patching the reported one — the role chunk, text deltas, tool deltas and the finish chunk all refresh it; `ThinkingDelta` was the sole gap (the remaining un-refreshed yield is the terminal usage chunk at `message_stop`, where the turn is already ending). So after the initial role chunk, a reasoning-only phase was killed at `STREAM_IDLE_TIMEOUT` while the client was actively receiving reasoning output.

The fix refreshes the timer when a reasoning chunk is yielded, matching the text-delta path.

**Regression test** (`reasoning_only_stream_survives_the_semantic_idle_timer`, paused clock): a reasoning-only upstream emits `thinking_delta` every 50s across 200s — well past the 60s idle bound, each gap inside it — then finishes. It asserts all four reasoning deltas reach the client, the turn ends with `finish_reason: stop` and `[DONE]`, and no `upstream_timeout`.

I verified it actually catches the bug rather than just passing:

```
without the fix → FAILED: assertion `left == right` failed
                  left: 1   right: 4      (turn killed at the idle bound)
with the fix    → ok. 1 passed
```

Full crate on the rebased head: `cargo test --locked` → **404 passed**, `cargo fmt --check` clean. Also rebased onto latest `main`.

Still agreed this wants the human maintainer pass on the auth/paywall/provider timeout policy and the 240s-under-300s platform assumption — that one needs someone with deployment visibility.

aryanorastar:
Both are already in the current tip (`36d0b4a2de`) — I think the review predates the rebase I pushed earlier today:

- **Code**: `ThinkingDelta` refreshes the timer — [`streaming.rs:393`](https://github.com/BasedHardware/omi/blob/36d0b4a2de/desktop/macos/Backend-Rust/src/routes/chat/streaming.rs#L393) sets `first_visible_at = Some(now())` right after yielding the `reasoning_content` chunk, so a long reasoning phase resets the idle bound exactly like text deltas.
- **Test**: `reasoning_only_stream_survives_the_semantic_idle_timer` (`tests/all.rs:2323`, `#[tokio::test(start_paused = true)]`) streams reasoning-only deltas across **200s** — past the 60s bound, each gap inside it — and asserts all 4 reach the client, the turn finishes `stop`, and no `upstream_timeout`.

```
cargo test reasoning_only_stream_survives ... test result: ok. 1 passed
```

A fresh CI run against the current tip should be green on this. Thanks for the careful read on the reasoning-stream contract.

Git-on-my-level (commented):
Thanks for the update — the blocking issue I raised on the prior head looks resolved on `36d0b4a2de`.

I re-checked the current diff: `ThinkingDelta` now refreshes the semantic-progress timestamp after yielding `reasoning_content`, and the new paused-time regression test covers a reasoning-only stream across 200s with each visible delta inside the 60s idle window. The Python send-deadline tripwire and its unit tests also pass locally in this review environment, and the static scan still finds no direct provider `.send()` calls under `routes/chat/`.

I dismissed my stale automation change-request review for the old reasoning-stream blocker. I’m still not formally approving from automation because this remains a security/workflow-sensitive runtime-policy change with maintainer-review labels, the Swift CI checks are currently failing, and the 240s-under-300s platform timeout assumption needs human maintainer/deployment sign-off before merge. Overall, this is now a positive implementation signal from my side; remaining sign-off is human policy/CI validation rather than the previous code blocker.

---

aryanorastar:
Rebased onto current main (head `f9c1230`) — it had gone CONFLICTING because main's paused-turn work (#10637, #10549 and the surrounding web-search commits) reshaped the same files. **One design decision in here needs your eyes**, because it changes where the budget applies:

**The separate server-tool entry point is gone.** This PR originally dispatched `requires_public_web` to its own `handle_server_tool_streaming`. Main has since folded that path *into* `handle_streaming`, so that function and the route-level branch are deleted — the deadline now threads into main's two existing entry points instead. Net simplification.

**The mid-stream continuation deliberately does NOT inherit the request budget.** `complete_anthropic_server_tool_turn` is now also called from inside the SSE stream (streaming.rs, paused-turn resume). Its own comment assumes "a continuation is pre-first-visible-event work" — that was true when I wrote it, and is false at main's new call site. Passing the request deadline there would kill a long stream whose turn pauses for web search past the 240s budget, i.e. it would give streaming the total-response timeout this PR explicitly says it does not add (the budget governs only until `first_visible_at`, then `STREAM_IDLE_TIMEOUT` takes over). So that call gets its own fresh budget: a stuck continuation stays bounded, the stream does not gain a total cap. Main had no bound there at all, so this is strictly tighter than main and strictly looser than inheriting the request deadline — flagging it since it is exactly the externally visible timeout semantics you asked to sign off on.

Also: `AnthropicSendError` gained `#[derive(Debug)]` so the continuation's failure log distinguishes `DeadlineExpired` from `Gateway(status)`.

Verified on the rebased tip: `cargo test --locked` → **390 passed, 0 failed** (main's paused-turn/continuation suites and this PR's deadline suite green together), `cargo check --locked` clean with no warnings, `cargo fmt --check` clean, and the line-count ratchet passes on every changed file.

Git-on-my-level (dismissed):
Thanks for rebasing this. I re-checked the current head (`f9c1230`) and the code direction still looks coherent: the chat route admits one deadline before auth/body extraction, provider sends go through `send_with_deadline`, reasoning deltas now count as semantic progress, and the post-first-visible pause-turn continuation is intentionally given its own bounded budget rather than inheriting the request deadline.

One merge-blocking item from current CI needs fixing:

- The `Hygiene` job is failing in `failure-class-guard-artifact-ratchet`: `FC-per-hop-timeout` now crosses the recurrence threshold, but `.github/failure-classes/FC-per-hop-timeout.json` does not record a `canonical_prevention_artifact`. Since this PR adds the reusable guard surface (`.github/scripts/check_chat_send_deadline.py` plus the manifest entry), please update the failure-class record to cite the canonical artifact path (or whatever field shape the ratchet expects) so the guard is durable and CI is green.

I verified locally that the Python tripwire and its unit tests pass, and the static scan finds no direct `.send()` calls under `routes/chat/`. I still cannot run the Rust tests in this environment because `cargo` is not installed.

This still needs human maintainer/deployment sign-off before merge for the runtime timeout contract: the 240s-under-300s Cloud Run assumption, auth/paywall/provider timing behavior, CI enforcement, and especially the newly called-out choice that mid-stream server-tool continuations get a fresh bounded budget after visible output rather than inheriting the original request budget. Once the hygiene ratchet is fixed, this looks like a positive implementation signal from automation, but not something automation should formally approve.

---

kodjima33 (approved):
Real fix with a stated root cause and tests, but 918 lines puts it past the scoped-diff bar for an unattended backend merge — approving; landing it is Nik's call.

Git-on-my-level (commented):
Thanks for the update. I re-reviewed the current head (`1368b30c85174d53dd6fce346cc452b2f9342a2f`) and the prior automation blocker is resolved: `FC-per-hop-timeout` now records `.github/scripts/check_chat_send_deadline.py` as the canonical prevention artifact, the manifest registers both the tripwire and its tests, and the current GitHub checks are green.

Local/static verification from this review environment:

- `python3 .github/scripts/test_check_chat_send_deadline.py` passes.
- `python3 .github/scripts/check_chat_send_deadline.py` passes and finds no direct provider `.send()` calls under `routes/chat/`.
- I still cannot run the Rust suite locally here because `cargo` is not installed, so I’m relying on the green Rust/Backend CI for that part.

I’m dismissing my stale change-request review because it was specifically about the missing failure-class guard artifact and that is now fixed on this head.

I’m still not formally approving from automation: this remains a security/workflow-sensitive backend runtime-policy change touching auth/paywall timing, provider transport, CI enforcement, and the 240s-under-300s Cloud Run timeout assumption. A human maintainer should also explicitly sign off on the current streaming policy that post-first-visible server-tool continuations get a fresh bounded budget rather than inheriting the original request budget. From automation, this is a positive implementation signal with human deployment/runtime-policy review still required before merge.

---

PR #10070, opened 2026-07-19
fix(listen): enforce globally bounded Parakeet routing and fallback

Git-on-my-level:
## Summary

Supersedes the accidental off-scope PR head and repairs the live-transcription routing incident at the provider boundary.

- Makes **Modulate Velma-2** the checked-in, rendered, production primary for streaming, prerecorded, and PTT STT.
- Moves the hard admission gate to the single Parakeet GPU service instance, so 19+ backend-listen replicas share one capacity limit.
- Makes Parakeet capacity/allocation explicit deployment settings, validated in runtime contracts; invalid/missing settings fail Parakeet startup rather than silently changing behavior.
- Releases admission leases on every session termination path and provides a readiness handshake so failed Parakeet connections fall back safely to Modulate.
- Adds a bounded provider circuit breaker for unhealthy Parakeet connection failures.

Parakeet became the global first-choice STT provider while it could safely serve only one bounded GPU pool. Per-listener admission could not protect that shared pool. The durable owner is now the Parakeet service itself: it enforces the global service cap before constructing a stream, and clients fail over to Modulate when admission or connection fails.

## Verification

```text
283 focused repair tests passed
687 selected backend test files passed
backend preflight: 17 passed
make preflight: 25/25 checks passed
rendered dev/prod runtime contract validators passed
git diff --check passed
```

## Failure class

## Follow-up / rollout

This PR changes source and deployment contracts only. It does not deploy or merge automatically. Before a production rollout, validate bounded real traffic: capacity denial and unhealthy Parakeet must route to Modulate, and the active-stream cap must remain below the configured Parakeet service capacity.


Git-on-my-level:
Rebased onto latest main and fixed all three previously failing checks.

- Backend unit suite: fixed two pyright `reportPrivateUsage` errors. Renamed `StreamAdmissionController._release` → `release_lease` (the lease is the designated caller) and `_validate_parakeet_admission_contract` → `validate_parakeet_admission_contract` (imported validation contract).
- Hermetic Backend E2E + Listen Pusher Stack Gauntlet: updated the Parakeet test peers (`ScriptedParakeetPeer`, `parakeet_stub.py`) to send the `{"type": "ready"}` admission frame that the production `ParakeetWebSocketSocket` now waits for during its startup handshake. Updated the fake STT socket's `drain_and_close` to call `finish()` so teardown observability stays consistent with production drain behavior.
- Desktop Swift / Hygiene: restored executable mode on `prepare-qualification-profile.sh` and added a desktop changelog fragment.

All CI checks now pass: Backend unit suite, Hermetic Backend E2E, Listen Pusher Stack Gauntlet, Sync Cloud Tasks Stack Gauntlet, Hygiene, Desktop Swift CI.

PR #10068, opened 2026-07-19
feat: add immutable beta-to-prod release rings

Git-on-my-level:
## Summary

- Add immutable, deployable release records and serialized beta-to-prod deployment/rollback workflows for the v1 stateless service ring.
- Route beta desktop/mobile builds to `api-beta.omi.me`, pin runtime config and Secret Manager versions, and use a beta full-route transcription probe before promotion.
- Guard legacy production deployment paths, record recovery evidence, and document the required external ring provisioning.

Closes #10049

## Root cause and durable guard

Release state was assembled from mutable workflow inputs and service-local deployment paths, so a rollback could not restore the exact serving configuration. Release records now contain immutable image digests, materialized config objects, numeric secret versions, and topology; the record-only workflow validates them before mutation and snapshots/restores on deterministic failure. Static release-ring and deployment-concurrency checks prevent a later workflow drift from restoring an arbitrary-ref production path.

## Product invariants affected

- INV-AUTH-1

The beta build identity is compile-time controlled and only changes backend routing; it does not alter authentication/session semantics.

## Verification

- `uv run --with pytest --with pyyaml pytest .github/scripts/test_check_release_rings.py backend/tests/unit/test_release_rings.py backend/tests/unit/test_render_release_ring_config.py backend/tests/unit/test_cloud_run_traffic_snapshot.py`
- Focused Flutter beta-routing tests and macOS API routing tests.
- `python3 .github/scripts/check_release_rings.py`
- `python3 .github/scripts/check-deployment-concurrency.py`
- `actionlint .github/workflows/release-record.yml .github/workflows/deploy-release-ring.yml`
- `make preflight`

## Rollout notes

The repository workflow fails closed until the documented beta Cloud Run/GKE/DNS/certificate/identity prerequisites are provisioned. A first live beta release and rollback drill remain an operational rollout step; macOS remains explicitly roll-forward-only for installed clients.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10068?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10075, opened 2026-07-20
fix(release): verify atomic recovery across GKE and Cloud Run

Git-on-my-level:
## Summary

Follow-up to #10068. Restores the complete pre-mutation serving vector after a failed release ring and verifies convergence rather than trusting command success.

- Snapshot and restore GKE ConfigMaps plus Helm release revisions/manifests, with read-back checks.
- Verify restored Cloud Run traffic, remove candidate tags, and delete first-time services during recovery.
- Fail closed on recovery divergence and preserve the pre-mutation active release pointer using generation-aware updates.
- Prevent manual-dispatch values from being interpolated directly into shell commands.

## Verification

- `rtk proxy backend/.venv/bin/python -m pytest -q .github/scripts/test_check_release_rings.py backend/tests/unit/test_release_rings.py backend/tests/unit/test_render_release_ring_config.py backend/tests/unit/test_cloud_run_traffic_snapshot.py backend/tests/unit/test_release_ring_gke_snapshot.py` — 25 passed
- `rtk python3 .github/scripts/check_release_rings.py` — passed
- `rtk python3 .github/scripts/check-deployment-concurrency.py` — passed
- Independent review: approved; no blockers found.

- INV-AUTH-1

## Failure class

Failure-Class: FC-nonrecoverable-promotion

The release must preserve and restore the pre-promotion serving vector until the complete candidate is verified.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10075?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/check_release_rings.py
+++ b/.github/scripts/check_release_rings.py
@@ -6,12 +6,13 @@ from __future__ import annotations
 import sys
 from pathlib import Path
 
-
 ROOT = Path(__file__).resolve().parents[2]
 
 
 def require(text: str, path: Path, fragments: tuple[str, ...]) -> list[str]:
-    return [f"{path}: missing required release-ring guard {fragment!r}" for fragment in fragments if fragment not in text]
+    return [
+        f"{path}: missing required release-ring guard {fragment!r}" for fragment in fragments if fragment not in text
+    ]
 
 
 def check() -> list[str]:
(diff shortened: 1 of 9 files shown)


Git-on-my-level:
Addressed the Codex P1 review on `_observed_targets`.

The post-restore read-back now resolves `status.traffic` (what Cloud Run is actually serving) instead of `spec.traffic` (what we just requested). A wedged service whose `status.traffic` still serves the candidate revision can no longer be falsely reported as restored — the restore evidence will correctly fail with "observed traffic does not match snapshot".

Added a regression test (`test_restore_detects_wedged_status_traffic_after_update`) that proves a wedged `status.traffic` serving a different revision than the restored `spec.traffic` is detected as a recovery failure.

undivisible:
Consolidated into #10090 — all 4 release-rings PRs merged sequentially with co-authorship preserved.

PR #10060, opened 2026-07-19
fix(listen): recover orphaned in_progress conversations at session boundaries

aryanorastar:
## Summary

- `process_pending` (the session-boundary recovery pass that already re-dispatches stuck `processing` conversations, #9960) now also recovers **orphaned `in_progress` conversations** — the largest stuck bucket in #9809 (~161 rows in production, some months old, several with real transcripts, all invisible to users because the web app filters to `processing,completed`).
- Recovery routes each orphan through `process_conversation` — the exact call a live conversation timeout makes: content goes through the durable finalization job seam; empty rows are deleted by `delete_empty_recording_conversation`. No new recovery mechanism, no new state.
- Orphanhood is proven by idle time, not status alone: any live session refreshes `finished_at` on every segment and its lifecycle loop processes an idle conversation within the ~2-minute conversation timeout, so an hour of silence means no session owns the row — including a session on another device. The session's own current conversation is excluded explicitly.
- The fetch is an equality-only status query (no composite index, no `firestore.indexes.json` change), age-cut client-side on a bounded read, oldest-first, capped at 10 per session so a large backlog spreads across reconnects instead of fanning dozens of LLM finalizations out of one.

Addresses the `in_progress` bucket of #9809.

## Root cause

A listen session that dies between conversation creation and finalization leaves the row `in_progress` with no **automatic** owner: the lifecycle loop only watches the current conversation, `process_pending` only re-dispatched `processing`, and the durable-job reconciler only replays rows that already have a finalization job. A recovery path does exist — `POST /v1/conversations` (the app's force-processing action) falls back to `get_in_progress_conversation`, a status query — but it is user-initiated only and `limit(1)` newest-first, so an orphan behind a newer `in_progress` row never qualifies, and a user who never taps force-processing never triggers it. (Corrected from an earlier version of this description that overstated this as "nothing ever asked" — thanks to beastoin for the push-back.) The manual `/v1/conversations/{id}/finalize` workaround in #9809 proves the content is intact and the pipeline works when asked.

## Failure class

Failure-Class: none

No registered class on `main` covers this. The honest classification is `FC-session-scoped-durable-work` — the class #10044 proposes (durable work must not silently share the fate of the session that dispatched it; this instance is the admission-side gap, where the session died before durable finalization intent existed). If #10044 lands first, I'll update this declaration to cite it.

## Durable guard

[...]

PR #10044, opened 2026-07-19
harden(listen): release held finalization leases on shutdown cancel (follow-up to #10066)

aryanorastar:
# Summary

Follow-up to #10066, which landed the core #9995 fix (process-scope ownership of claimed finalization via `start_background_task`, best-effort result callback). This PR ships the residual hardening on top of that mechanism:

- **Shutdown lease release:** `except asyncio.CancelledError` in `_process_conversation_task` converts the only cancellation that can still reach a claimed job — the process-shutdown drain (`drain_background_tasks` cancels tracked tasks) — into an explicit `mark_finalization_retryable(..., 'worker_shutdown')` under `asyncio.shield`, so a restart requeues the held lease instead of waiting out lease expiry. Lease-epoch fencing keeps the release a no-op after completion.
- **In-flight visibility:** `PUSHER_DETACHED_FINALIZATION_TASKS` gauge (inc/dec around the task body) — the observability Git-on-my-level's earlier review asked for, surviving #10066's mechanism change.
- **Result-send pre-check:** `send_result` returns early when the socket is not CONNECTED, complementing #10066's narrowed exception handling; the regression test asserts the frame is dropped, not attempted.
- **FC-session-scoped-durable-work registration** (separate commit): the class #9648/#9704/#10066 were all instances of, so the next one is declared instead of rediscovered.
- **Route-level regression coverage** that #10066 did not add: a close-after-104 session through the real `_websocket_util_trigger` proving the claimed job converges to `completed` with no later live session, plus a mid-lease cancel proving the `worker_shutdown` requeue.

Relates to #9995 (core fix landed in #10066; this completes its hardening tail).

# Failure class

Failure-Class: new

`FC-session-scoped-durable-work`: durable ledger-owned work, once claimed, must not inherit the lifetime of the connection or session that dispatched it. #9648 (durable job ledger), #9704 (inline terminal state), and #10066 (session-close cancellation) were all retail fixes at this boundary; the definition names the class so recurrence is declared, not rediscovered.

# Durable guard (as of current head)

- Spawn ownership is main's: opcode-104 dispatch uses `start_background_task` (process scope, tracked, drained at shutdown) — this PR does **not** carry a private detached registry; the earlier detach implementation was superseded by #10066 and dropped in the merge.
- The shutdown drain *cancels* tracked tasks; without this PR that `CancelledError` bypasses the retry/dead-letter handlers (it is a `BaseException`) and strands the claimed job leased until expiry. The new handler is the missing half of #10066's ownership move.
[...]

PR #10078, opened 2026-07-20
fix(conversations): honor a preferred summarization app the setter admitted

### LINKED ISSUE #10074: Bug: summary template set as standard/default is not being applied
## Summary
A summary template set as the standard/default template is not being applied.

## Problem
After setting a summary template as the standard/default one, the app does not actually use it when generating summaries.

## Expected behavior
When a summary template is marked as the standard/default template, new summaries should use that template automatically.

## Actual behavior
The template can be set as standard/default, but it is not applied when summaries are generated.

## Impact
This breaks expected template behavior and makes the summary templating feature unreliable.

Users may think their configured default template is active when it is not.

## Suggested investigation
- Check whether the default template setting is being saved correctly
- Check whether summary generation is reading the selected standard/default template
- Check whether another fallback template is overriding the configured default
- Check for a client/server mismatch between template selection and template application

## Environment
- Omi app
- Production environment

aryanorastar:
## Summary

- A summary template set as the standard/default via `PUT /v1/users/preferences/app` is now actually applied at generation time, even when the app never landed in the user's enabled-installed slice.
- The reader resolves a preferred app missing from that slice through the **setter's own availability authority** (`get_available_app_model_by_id`, the new shared App-model form of `get_available_app_by_id`) and uses it when it can summarize.
- The previously **silent** drop now always logs: a deleted/inaccessible preferred app falls through to suggestions with a warning; a resolvable app without the `memories` capability falls through with a warning.

Fixes #10074.

## Root cause

Split availability authority between admission and use:

- **Setter** (`routers/users.py`): `get_available_app_by_id` — existence + private-ownership only. Accepts the template and stores it in Redis.
- **Reader** (`_trigger_apps` in `utils/conversations/process_conversation.py`): honored the preferred id only if present in the `works_with_memories() AND enabled` installed slice.

Any app passing the first check but not the second was accepted as the default and silently ignored at generation, falling through to the LLM suggestion path — exactly the reported behavior ("the template can be set as standard/default, but it is not applied"). Two concrete ways to get there without any user error:

1. The template create flow calls `enableAppServer` and **ignores failure** (`create_template_bottom_sheet.dart` — `if (success)` with no else), then lets the user set the never-enabled app as default.
2. The per-user app-slice and enabled-set caches (30s TTLs) mean a template created/enabled on one device can be set as default before another surface's slice converges.

## Failure class

Failure-Class: none

No registered class covers split availability authority between an admission route and its consumer. It rhymes with `FC-split-mutation-authority` (one authoritative owner per decision) but that class is scoped to mutable-state transitions, not read-side policy; declaring it here would stretch its definition. Flagging rather than stretching.

## Durable guard

- One authority: the reader now consults the exact check the setter used, via a shared function whose docstring names the contract — future preference readers have a primitive to reuse instead of re-deciding availability locally.
- The remaining legitimate fallthroughs (deleted app, capability mismatch) are logged with reasons, so this failure mode can never be silent again.
- The capability guard stays: a chat-only persona set as preferred cannot run as a summarizer.

## Validation

- Old-code regression check (reader change stashed): the two new tests → **2 failed** (template ignored, suggestion LLM invoked); new code → **20 passed** in `tests/unit/test_process_conversation_usage_context.py`.
- New tests extend the existing `_trigger_apps` preferred-app suite through its established seams: an admitted-but-not-installed template runs with **no suggestion LLM call**; a capability-less preferred app falls back to suggestions; the existing deleted-app fallthrough test keeps its contract (resolver returns None).
- Repo runner (file-isolated, as CI executes): `tests/unit/test_process_conversation_usage_context.py` → 20 passed, `tests/unit/test_lock_bypass_fixes.py` → 61 passed, `tests/unit/test_listen_process_pending_shutdown.py` → 3 passed.
- Note for reviewers: running `test_process_conversation_usage_context.py` and `test_lock_bypass_fixes.py` in one shared pytest process shows 19 cross-file failures **on clean main as well** — pre-existing module-isolation interaction, which is exactly why `test.sh` runs file-isolated; not introduced or worsened here.
- `black --line-length 120 --skip-string-normalization` clean on the three changed files.

Not exercised live: the full app→backend template flow against production (no device/credentials here). The fix point is server-side selection logic driven directly by the regression tests; the app-side create flow is unchanged.

## Out of scope, named

- The app-side silent `enableAppServer` failure (`create_template_bottom_sheet.dart:167`) still deserves a user-visible error; with this fix it no longer breaks the default template, so it's a polish follow-up.
- The reprocess path (`app_id` parameter) keeps its existing installed-slice lookup — different admission route, untouched.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10078?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-utils.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-utils.json
@@ -1,6 +1,6 @@
 {
   "files": {
-    "backend/utils/apps.py": 1641,
+    "backend/utils/apps.py": 1652,
     "backend/utils/memory/legacy_backfill.py": 1816,
     "backend/utils/memory_ingestion/pipeline.py": 1711,
     "backend/utils/other/storage.py": 1584,
@@ -8,6 +8,7 @@
     "backend/utils/sync/pipeline.py": 2299
   },
   "raise_justifications": {
+    "backend/utils/apps.py": "get_available_app_model_by_id: the set-preferred-app availability authority as a validated App model, so the summary path honors what the setter admitted instead of re-deciding with the installed slice (#10074).",
     "backend/utils/other/storage.py": "delete_app_logo now requires an exact app-logo bucket prefix (startswith) so a foreign URL embedding the prefix later cannot delete an unrelated object on this deletion path (David review on #9873).",
     "backend/utils/sync/pipeline.py": "Durable-ledger convergence remains beside the Sync pipeline's run-lock ownership and terminal-result transitions instead of splitting one recovery invariant across modules."
   },
--- a/backend/tests/unit/test_process_conversation_usage_context.py
+++ b/backend/tests/unit/test_process_conversation_usage_context.py
@@ -916,8 +916,14 @@ def _make_trigger_conversation(suggested_apps=None):
     return conv
 
 
-def _trigger_apps_context(default_apps=None):
-    """Context manager that patches all external dependencies of _trigger_apps."""
+def _trigger_apps_context(default_apps=None, availability_app=None):
+    """Context manager that patches all external dependencies of _trigger_apps.
+
+    `availability_app` stands in for `get_available_app_model_by_id` — the
+    set-preferred route's availability authority (#10074): None models a
+    deleted/inaccessible app; an app object models one the setter admitted even
+    though it is outside the enabled-installed slice.
+    """
     suggestion_mock = MagicMock(return_value=(["suggested-app"], "reasoning"))
     app_result_mock = MagicMock(return_value="App result content")
     record_mock = MagicMock()
@@ -929,6 +935,7 @@ def _trigger_apps_context(default_apps=None):
         patch.object(process_conversation, "get_suggested_apps_for_conversation", suggestion_mock),
         patch.object(process_conversation, "get_app_result", app_result_mock),
         patch.object(process_conversation, "record_app_usage", record_mock),
+        patch.object(process_conversation, "get_available_app_model_by_id", return_value=availability_app),
     )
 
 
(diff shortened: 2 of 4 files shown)


Git-on-my-level (commented):
Thanks for the focused fix here. I reviewed the diff and ran the targeted backend unit coverage locally:

`PYTHONUTF8=1 ENCRYPTION_SECRET=*** OPENAI_API_KEY=*** BACKEND_FAST_UNIT_WARN_SECONDS=999 BACKEND_FAST_UNIT_FAIL_SECONDS=999 .venv/bin/python -m pytest -q tests/unit/test_process_conversation_usage_context.py -m 'not integration and not slow'`

Result: 20 passed.

The implementation now reuses the same availability authority as the set-preferred route when a stored preferred summarization app is not present in the enabled/installed slice, while still falling back when the app is missing or does not support memories. The added regression tests cover the in-slice preferred app, missing/stale preferred app, no preferred app, out-of-slice but available preferred app, and non-summarization app fallback paths.

I do not see code blockers. I’m leaving this as a positive maintainer signal rather than a formal approval because the PR touches the repo automation ratchet baseline under `.github/`, so it falls under the automation/workflow-sensitive approval guard.

---

aryanorastar:
Not still in progress — this is complete and ready for review. The branch is a single commit, currently mergeable against latest `main`, and I'm not planning further changes unless review asks for them. Thanks for the review pass, @Git-on-my-level!

aryanorastar:
@Git-on-my-level need human response — this PR (and the same automation-guard situation on #10094 / #10095, which only bump ratchet baselines with justifications) is complete and mergeable, blocked only on the required human approval for the `.github/` guard. Could David take a look when he gets a chance? The pure-backend PRs #10060, #10054, #10044 have no automation-sensitive paths and are also ready, along with their awaiting workflow-run approvals. Happy to address any review feedback quickly.

kodjima33 (approved):

Git-on-my-level:
Thanks LGTM

aryanorastar:
@Git-on-my-level @undivisible This PR is fully approved (LGTM received) and ready to merge. Could you please trigger the merge when you have a moment? Thank you!

PR #10081, opened 2026-07-20
fix: surface silent failures across agent, auth, and desktop

undivisible:
## What changed and why

Stops fail-open and silent-failure paths on the current (non-Tauri) desktop/backend stack so errors surface instead of looking healthy.

- **Agent VM recovery** — GCE/Firestore status recovery no longer discards write or network failures; failed authoritative transitions log and fail closed when they block state.
- **Auth custom token** — `use_custom_token=true` returns an error instead of a partial token payload when minting is unimplemented.
- **Memories** — remove unused `recompute_evidence` placeholder (no callers).
- **Electron** — log corrupted Google refresh-token decrypt, non-ENOENT Rewind retention deletes, and primary-screen lookup fallbacks.
- **macOS APIClient** — log app setup-status check failures instead of empty `catch {}`.

- INV-MEM-1

## How it was verified

- `cargo check --manifest-path desktop/macos/Backend-Rust/Cargo.toml --bin omi-desktop-backend` — passed
- `cargo test --manifest-path desktop/macos/Backend-Rust/Cargo.toml --bin omi-desktop-backend agent --offline` — passed
- `python3` AST parse of `backend/database/memories.py` — passed
- `scripts/pr-preflight` — 26 checks passed
- Label: `no-changelog-needed` (internal logging / reliability only)

## Tests

No focused route test exists for the GCE/Firestore recovery path; the existing backend binary agent test passed. Electron changes are log-only on already-handled error paths.

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/backend/database/memories.py
+++ b/backend/database/memories.py
@@ -522,12 +522,6 @@ def add_evidence(uid: str, memory_id: str, evidence: Dict[str, Any], *, firestor
     memory_ref.update(update_payload)
 
 
-def recompute_evidence(uid: str, memory_id: str, *, firestore_client: Any = None) -> List[Dict[str, Any]]:
-    """Placeholder hook for later veracity/tombstone recomputation tickets."""
-    memory = get_memory(uid, memory_id, firestore_client=firestore_client)
-    return (memory or {}).get('evidence', [])
-
-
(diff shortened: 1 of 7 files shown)


undivisible:
Superseded by https://github.com/BasedHardware/omi/pull/10084 after renaming the branch to `fix/silent-failure-hardening` (GitHub closed this PR when the old head ref was renamed).

PR #10086, opened 2026-07-20
feat(release): make beta overrides explicit and dispatch reliable

### LINKED ISSUE #10082: release: separate qualification evidence from promotion permission (make override expressible, on the record)
## Problem

In the desktop release control plane, **qualification evidence and promotion permission are the same object**. `qualification.passed` is simultaneously:

- a **factual claim** that a rebuild-from-tag, hermetic T2, and the fault suite ran and passed, and
- the **permission token** required to move a channel pointer.

Because they are fused, the only way to obtain permission is to assert a fact. When an operator legitimately decides to ship without full qualification, the system offers exactly two options: **fabricate the evidence, or don't ship.** There is no override field, no reason, no operator attribution.

The check is server-side, so this is not a client-side script gate that can be worked around cleanly:

- `backend/database/desktop_update_channels.py:158` — `_build_channel_pointer` rejects any manifest without `qualification.passed is True` and `tier == "T2"`
- `backend/database/desktop_update_channels.py:200` — `_build_beta_rollback_pointer` applies the identical check, so a fabricated manifest also becomes a sanctioned **rollback target** in a future incident
- `.github/scripts/prepare-desktop-beta-promotion.py` fail-closes earlier with `FAIL: release must have qualifiedBeta: true`

## Evidence

During the `v0.12.86+12086-macos` incident (2026-07-20), users were crashing on `.85` and `.86` contained the fix:

1. `.86` built, signed, notarized; all 11 signed-smoke checks passed including the Keychain canary. Published asset digests were independently verified against `desktop-smoke-result.json` (`Omi.zip aa6069ea…`, `omi.dmg 7ec0512c…`).
2. GitHub Actions was in a partial outage, so `desktop_qualify_beta.yml` could not run on the trusted runner.
3. Local qualification produced **37/38 T2 flows passing**; `speaker-naming` failed deterministically on `result.detail.segment_count: expected '2', actual '3'`.
4. With no way to express "promote with qualification overridden, on the record," the operational path became: flip the global `DESKTOP_UPDATE_POINTERS_MODE` kill switch on the prod backend.
5. That path required a traffic move, which surfaced a pending Cloud Run revision carrying a **different image**, `OMI_LLM_GATEWAY_URL` pointed at the discard port (`127.0.0.1:9`), and four new Stripe price IDs — one `--to-latest` away from production.

The rigid gate did not prevent risk; it displaced it from a small, legible, recorded override into a large, opaque, unreviewed change to shared backend config under time pressure.

**This is a recurrence, not a one-off.** #9415 documents the same failure class during `v0.12.69+12069-macos`: a signed, notarized, smoke-passed candidate blocked by T2 reporting "suspected pre-existing failures/harness drift."

Two secondary findings from the same incident:

- `/v2/desktop/channels/rollback` is implemented, CAS-protected, generation-aware, and unit-tested, but returns **HTTP 404 in prod** (vs `403` for `/promote`) and has **zero callers** in `.github/`, `scripts/`, or any runbook — while `desktop/macos/AGENTS.md` tells operators "roll-forward only." The most useful action available during the incident was built and unreachable.
- A gate failure carries no interpretive context. `speaker-naming` failing tells an operator nothing about whether it is a real conversation-data regression (`.86` carries cache-first mutation replay and WAL sync changes) or a known-flaky flow — `a97cfc61e8 fix(desktop): make Tier-2 qualification deterministic` previously touched this exact file. A gate that cannot be interpreted cannot inform a decision.

## Proposed direction

Split the two concerns:

- **Evidence records are facts.** Append-only, immutable, never overridable. `T2 ran at <ts>, 37/38 passed, speaker-naming failed on segment_count`.
- **Promotion is a decision** that references evidence and may carry an explicit `override` with `reason` and `operator`, recorded in the manifest and surfaced in telemetry/alerting.

Then "ship `.86` without full qualification" is expressible honestly, and produces a *better* audit trail than a clean pass, because the reason is attached.

Expose this through a single control surface — an `omi-release` CLI (`promote`, `rollback`, `status`, `--override`) that wraps the same authenticated endpoints. GitHub Actions becomes *a* caller rather than *the* caller, which also removes the orchestrator as a single point of failure for a mutation it does not own. Deploying `/channels/rollback` and giving it a caller belongs in the same effort.

## Relationship to existing issues

- **#10063 / #10065** propose a separate audited emergency lane (+1189/-10, new workflow, two distinct approvals, mandatory one-hour monitor). That shape does not resolve this: an emergency lane requiring two humans and an hour would not have helped during this incident, and it adds a parallel pipeline rather than making the existing one expressive. This issue argues the override belongs **on the existing lane** as a recorded decision.
- **#9415** — prior instance of the same failure class; this issue proposes the durable fix rather than another per-incident workaround.
- **#9525** — promotion engine work; the evidence/permission split is a prerequisite for a promotion engine that can be both safe and operable.

## Scope note

The evidence/permission split plus the rollback caller is intended as one self-contained change to the control plane and its callers. The CLI can land incrementally as a wrapper over the same endpoints. No new service.

Git-on-my-level:
## Summary

Closes #10082

- separates immutable macOS beta qualification evidence from the protected, auditable emergency beta-only override decision
- records the exact tag, SHA, evidence, rationale, operator, approvers, and server timestamp before the sole emergency pointer transition
- adds immutable qualified-only beta rollback and reliable, deduplicated Codemagic qualification dispatch state

- INV-AUTH-1

## Verification

- `python3 .github/scripts/test_desktop_qualification_dispatch.py`
- `python3 .github/scripts/check-release-process-guards.py`
- `actionlint .github/workflows/desktop_qualify_beta.yml .github/workflows/desktop_emergency_promote_beta.yml`
- `make preflight`

## Failure class

## Limitations

- Full backend pytest selection could not run because this worktree has no project-local Python environment and restricted network blocks the documented Python 3.11 bootstrap.
- The existing macOS qualification fixture reaches the local `defaults` write and is blocked by the sandbox; no production app or release was touched.


undivisible:
Consolidated into #10090 — all 4 release-rings PRs merged sequentially with co-authorship preserved.

PR #10091, opened 2026-07-20
refactor(release): beta break-glass as a relaxed precondition, not a parallel subsystem

### LINKED ISSUE #10082: release: separate qualification evidence from promotion permission (make override expressible, on the record)

Git-on-my-level:
## Summary

Closes #10082

Adds a break-glass path for promoting a signed macOS candidate to beta when the qualification gate itself is broken — the case we actually hit.

**Scope changed after review**, twice:

1. The first pass built this as a parallel subsystem with an approval ceremony. It is now a relaxed precondition on the existing promotion path.
2. An audit of every deploy surface found the same footgun on two more. Those now have hatches too.

## Every gated surface now has a hatch

The audit found the posture was bimodal and arbitrary: one surface welded shut with no hatch, twenty with no gate at all, and which was which bore no relation to blast radius.

| Surface | Before | Now |
|---|---|---|
| Desktop candidate (`desktop_auto_release`) | Swift source gate; `force_release` explicitly did **not** bypass it | `release_mode=break_glass` |
| Desktop beta | fail-closed qualification, no hatch | `desktop_emergency_promote_beta` |
| Desktop stable | already had `break_glass_confirm` | unchanged |
| **Backend prod** (`gcp_backend`) | required a *successful* eligibility run on the exact SHA, **zero bypass inputs** | `skip_eligibility_proof` + `break_glass_confirm` + reason |

Two of these could block a hotfix with no recovery at all:

- **Backend prod**: a flaky or red check *permanently poisons that SHA* — rerunning doesn't make the original run succeed, so the only recovery was landing another commit on `main` and waiting for a full green run. That's a bigger blast radius than the beta gate that cost a day.
- **Desktop candidate**: `plan-desktop-release.py` checked the source gate *before* the `force_release` branch, so a flaky `Desktop Swift Build & Tests` blocked candidate creation outright, and everything downstream with it.

**Hatches relax evidence requirements only.** The backend hatch moves the ancestry-of-`main` check *outside* the break-glass branch so it is always enforced — unreviewed code still cannot reach production. Each hatch opens a tracking issue; repeated use means the gate is the defect.

`AGENTS.md` now carries one table of every hatch. Not knowing a hatch exists is the same outage as not having one.

Untouched: `gcp_backend_auto_dev` is `workflow_run`-only and manual `gcp_backend` is already its fallback. The 20 ungated `gcp_*` services need gates decided first — separately, not here.

## The design

Emergency promotion is ordinary promotion with the T2 qualification gate relaxed. It is not a separate mutation path.

- `_build_channel_pointer`, `_build_beta_rollback_pointer`, and `_build_emergency_macos_beta_pointer` collapse into one `_build_pointer` driven by a `TRANSITIONS` policy table. It is the single pointer-mutation authority; the three transitions differ only in which preconditions they declare.
- The three emergency endpoints become `emergency: true` on the existing `POST /v2/desktop/channels/promote`. Beta-only, and requires a compare-and-swap on both current release id and generation, so it cannot silently clobber a concurrent normal promotion.
- The resulting pointer carries `emergency: true`. The audit record lives on the pointer rather than in a parallel collection that can diverge from it.
- Normal promotion is unchanged and still fail-closed.

## Why the ceremony was removed

Removed: two-approver comment parsing, the four-hour expiry window, incident binding, behavioral-evidence fetch, responder notification, the one-hour monitor, the operation-id reconciliation protocol, and the recovery job that existed to repair state the subsystem itself created.

This codebase is mostly agent-driven, and **ceremony is not a control against an automated actor.** An agent fills in a rationale, an incident id, and an approval comment as readily as a human — faster, and correctly. Those fields imposed real cost on a small team mid-incident while filtering nothing. Eight of this branch's original commits were fixes to that machinery rather than to release behavior, which is the clearest signal it cost more than the risk it insured.

What remains are the guards a machine can actually check: tag-to-SHA binding, artifact digests recomputed from the published assets, and the pointer CAS.

The workflow now reads the live pointer itself rather than requiring an operator to paste release id and generation mid-incident.

## The escape hatch is a signal, not a solution

Every run opens a `qualification-gate-failure` issue recording tag, SHA, operator, reason, and run URL. If those accumulate, the qualification gate is the defect and this hatch is masking it. Fix the gate, then close the issue.

## Also in this PR

Deletes `release_blessing.py` and `check-python-backend-blessing.py` plus their tests (502 lines). The latter is referenced only by its own test — no workflow, no Makefile, no `checks-manifest.yaml` entry — and `run_checks.py` resolves checks by explicit id without globbing, so nothing could execute it.

The qualification reliability amendment keeps every break-glass workflow unchanged while making its automatic evidence boundary fail closed: fault qualification waits for the bridge's live signed-in, onboarded, non-restoring owner snapshot (not only the UI's `appState: main`), and release preflight validates the fetched publication timestamp plus signed ZIP, DMG, and signed-smoke assets with SHA-256 digests.

- INV-AUTH-1

## Verification

- `cd backend && python3 -m pytest tests/unit/test_desktop_update_channels.py tests/unit/test_desktop_updates.py tests/unit/test_desktop_release_scripts.py` → **133 passed**
- New `TestEmergencyBetaPromotionRules` covers: promotes an unqualified build and marks the pointer `emergency`; normal promotion still rejects that same build; stale release-id and stale generation CAS both rejected; roll-forward only; never reaches stable.
- `python3 .github/scripts/check-release-process-guards.py` → passes
- `python3 .github/scripts/test_desktop_qualification_dispatch.py` → passes
- `python3 .github/scripts/check-desktop-release-terminology.py` → passes
- `python3 backend/scripts/check_route_policy_baseline.py` → 0 baseline additions, 0 new missing entries after removing the three routes
- Workflow validated as YAML (466 → 179 lines, one job, 8 steps)
- `python3 .github/scripts/test_plan_desktop_release.py` → passes, including two new cases: `break_glass` releases over a blocked source check, and does not set the bypass flag when the check passes. The existing assertion that `force_release` still blocks is unchanged.
- `python3 .github/scripts/test_check_backend_deploy_source_admission.py` → 23 pass, including 3 new break-glass contract tests (confirm string + reason required; ancestry enforced outside the branch; bypass recorded as an issue). These read workflow source, so they are **static tripwires, not behavioral coverage** — Actions cannot be driven from a unit test.
- `bash desktop/macos/tests/test-omi-ctl-wait-ready.sh` → rejects signed-out, restoring, owner-not-ready, and stale snapshots; accepts only the live authenticated owner-ready bridge snapshot.
- `python3 desktop/macos/scripts/release-keyvalue.py self-test` → validates passing release JSON and rejects missing/malformed publication timestamps, missing signed-smoke assets, and missing artifact digests.
- `bash desktop/macos/tests/test-fault-suite-launch-env.sh && bash desktop/macos/tests/test-qualify-desktop-beta-contract.sh` → passes; confirms the fault harness invokes the strengthened readiness command and qualification bootstrap contracts remain intact.
- `ENCRYPTION_SECRET=*** backend/.venv/bin/python -m pytest backend/tests/unit/test_desktop_release_scripts.py backend/tests/unit/test_desktop_update_channels.py` → 43 passed; covers qualified qualification-claim retries re-dispatching beta promotion without rerunning the runner.
- `actionlint .github/workflows/desktop_qualify_beta.yml` → passes.
- `make preflight` with this PR body → 36 deterministic checks passed.
- `python3 .github/scripts/check_backend_deploy_source_admission.py` → passes. Note it validates structural wiring, not bypass semantics — hence the added contract tests.
- `check_agents_md_lean.py` → within budget

## Limitations

- **No hatch has been fired against real infrastructure.** These change production deploy controls — beta promotion, backend prod deploy, desktop candidate creation. They need a real incident or a deliberate dry run before they are trusted.
- **`gcp_backend.yml` is prod backend deploy.** Per `AGENTS.md` that is a wide-blast-radius surface needing explicit sign-off; this is not auto-mergeable.
- Full backend pytest selection could not run in this worktree (no project-local Python environment; restricted network blocks the documented 3.11 bootstrap). The three affected files were run directly.

## Failure class

The original defect and this rewrite share a cause: pointer-mutation authority was split across parallel implementations. Consolidating to a single `_build_pointer` removes the class rather than adding another observer of it.

Deliberately excluded to keep this reviewable:

- 22 `gcp_*` deploy workflows, 3,766 lines, zero `workflow_call` — the job scaffold is copy-pasted ~20×. A reusable workflow targets ~1,200 lines and removes the drift risk of hardening one deploy path and missing the others.
- ~1,300 lines implementing the same scan → count → compare-to-JSON-baseline ratchet five separate times, across 8 baseline formats. One runner plus a rule table targets ~300.


Git-on-my-level:
Maintainer direction: retain the PR’s broad break-glass bypasses. Follow-up work should add the incident-derived macOS qualification/auth-readiness and evidence/audit improvements without removing the existing desktop candidate, Beta promotion, Stable promotion, or backend deployment hatches.

Git-on-my-level:
Clarification from maintainer: the separate emergency promotion workflows are intentional. Retain broad, flexible candidate/Beta/Stable/backend break-glass avenues rather than folding them into one ordinary promotion path. Continue adding the incident-derived M1 fault authenticated-readiness, qualification-controller reliability, and evidence/audit improvements; do not remove the existing hatch mechanisms.

Git-on-my-level:
Monitor follow-up for the trusted review threads:

| Thread | Status |
| --- | --- |
| `PRRT_kwDOLkKqys6SKBWh` / `discussion_r3612318511` | Fixed. Qualified dispatch retries now expose `should_promote=true`, and `desktop_qualify_beta.yml` re-dispatches protected beta promotion without rerunning qualification. |
| `PRRT_kwDOLkKqys6SKBWj` / `discussion_r3612318514` | Resolved by the simplified design already in this PR: emergency side effects were removed from the pointer/readiness boundary, so there is no abandoned side-effect claim to recover. |

Verification run locally before push:

- `python3 .github/scripts/test_desktop_qualification_dispatch.py`
- `ENCRYPTION_SECRET=*** backend/.venv/bin/python -m pytest backend/tests/unit/test_desktop_release_scripts.py backend/tests/unit/test_desktop_update_channels.py` → 43 passed
- `actionlint .github/workflows/desktop_qualify_beta.yml`
- `git diff --check`
- Manual pre-push gate against `origin/main` with this PR body → passed 36 manifest checks plus bounded repo hooks; rerun with `env -u VIRTUAL_ENV` so backend typecheck used `backend/.venv` instead of the Hermes Python.

I resolved both review threads after pushing `a250055478f26ba83d6ef79cbf241576a6315bf8` and re-scanned trusted bot threads; none remain unresolved at this head.

PR #10094, opened 2026-07-20
fix(dev): serve legacy memories when a legacy-cohort account has no rollout state

### LINKED ISSUE #9892: Developer API: GET /v1/dev/user/memories returns 403 missing_rollout_state (conversations/action-items OK)
Since around July 11, `GET /v1/dev/user/memories` returns 403 for my account:

```json
{
  "detail": {
    "enabled": false,
    "consumer": "developer_api",
    "archive_capability": false
  }
}
```

`/v1/dev/user/conversations` and `/v1/dev/user/action-items` return 200 with the same key, so this is specific to memories.

**What I tried:** My original API key (created in early June) initially failed with `reason: "malformed_app_key_scope_grant"`. Creating a fresh `omi_dev_` key in the app cleared that layer, but memories now fail at the rollout gate with `missing_rollout_state` as above. So a new key does not resolve it.

**Where it appears to come from:** `backend/utils/memory/default_read_rollout.py` fails closed when the per-user `users/{uid}/memory_control/state` doc is missing, and `backend/utils/memory/product_authorization.py` routes external consumers (`developer_api`) through that gate. Accounts not yet enrolled in the canonical-memory rollout therefore lose Developer API memory reads entirely (no legacy fallback like the first-party route has).

**Impact:** Breaks the memories feed of third-party integrations (e.g. the Obsidian omi-sync plugin) for accounts awaiting migration. Conversations and action items still sync fine.

**Ask:** Could my account be migrated/enrolled, or the `developer_api` consumer given a legacy-safe fallback for un-migrated accounts?

Happy to provide my user id / key id privately if useful for debugging.

aryanorastar:
# Summary

- `GET /v1/dev/user/memories` no longer returns `403 missing_rollout_state` for accounts that were never enrolled in the canonical-memory rollout: the route now serves the authoritative legacy `memories` collection for them.
- Every other deny reason (missing grant, malformed state, unsupported schema, uid mismatch) keeps the fail-closed 403 contract, as does `/v1/dev/user/memories/vector/search`, which has no legacy read surface.
- The mode change is recorded through the shared `record_fallback` helper (`component=other`, `reason=policy`, `outcome=recovered`) — no new one-off counter.

Fixes #9892.

# Root cause

Two authorities disagreed about the same account, and the fail-closed one won:

- **Cohort selector (authoritative):** `pin_memory_system` → `resolve_memory_system` resolves everyone outside the code-reviewed `CANONICAL_MEMORY_USERS` whitelist to `MemorySystem.LEGACY` — "the documented default — not an implicit None fallback" (its own words). For these accounts, `users/{uid}/memory_control/state` is *expected* to be absent.
- **Rollout gate:** `read_default_read_rollout` treats an absent doc as `DENY_MEMORY` / `missing_rollout_state` — correct fail-closed behavior for an account that might be mid-migration, wrong for one the cohort selector already pinned LEGACY in the same request.

The route's legacy read path (`memories_db.get_memories` + per-doc validation) already existed but was unreachable: only `USE_LEGACY_SAFE` reaches it, and rollout normalization never produces `USE_LEGACY_SAFE` from Firestore state — only the explicit opt-in constructor does. So every un-enrolled account lost Developer API memory reads entirely, while `/v1/dev/user/conversations` and `/v1/dev/user/action-items` (no such gate) kept working — exactly the reported symptom, breaking e.g. the Obsidian omi-sync memories feed.

# Failure class

Failure-Class: none

No registered class covers a reader failing closed on state whose absence the authoritative cohort selector defines as normal. It rhymes with FC-split-mutation-authority (competing authorities over one decision) but that class is scoped to mutable-state transition ownership, not read-side policy divergence; flagging rather than stretching — same call as #10078.

# Durable guard

- The fallback is *narrow and explicit*: only `missing_rollout_state` (doc absent = never enrolled) on the already-LEGACY branch falls back. `malformed_rollout_state`, `rollout_read_failed`, grant and schema denials all stay fail-closed, because there a doc exists (or its state is unknown) and the account may be mid-migration.
- The previously silent-dead `should_use_legacy_fallback: pass` branch is gone; the decision structure now states every outcome.
- Mode change is observable via the shared `omi_fallback_total` metric instead of being invisible.

# Product invariants affected

# Validation

- Old-code regression check (route change stashed): `test_get_memories_missing_rollout_state_falls_back_to_legacy` → **fails** (403); with the fix → **passes**. Full file: 7 passed.
- New tests extend the existing grant-ordering suite through its established seams: an un-enrolled legacy-cohort account lists legacy memories with 200 (grant check still runs first); a non-missing deny reason still 403s with the actionable `developer_memory_access_not_ready` contract; the vector-search missing-rollout 403 test keeps its contract untouched.
- Repo runner (file-isolated, as CI executes): `tests/unit/test_dev_api_canonical_grant_ordering.py` → 7 passed, `tests/unit/test_developer_memory_adapter.py` → 24 passed, `tests/unit/test_memories_validation.py` → 12 passed, `tests/unit/test_dev_api_folder_filters.py` → included in sweep, passed.
- The pre-push bounded gate ran the wider backend selection and passed; it caught (and I updated) the static source-order tripwire in `test_developer_memory_adapter.py` that encoded the old "no legacy after deny" ordering — it now asserts the narrow un-enrolled guard precedes the legacy read, and is labeled as a static tripwire per the repo's testing guidance.
- `black --line-length 120 --skip-string-normalization` clean on both changed files.

Not exercised live: a production Developer API key against a real un-enrolled account (no such credentials here). The fix point is server-side branch selection driven directly by the regression tests; the issue reporter's account can verify live after deploy.

# Out of scope, named

- MCP list/search (`routers/mcp.py`, `mcp_sse.py`) degrade differently for the same accounts (silent empty list instead of 403). Same underlying divergence, different surface and files — follow-up rather than scope creep here.
- The issue's alternative ask (enrolling individual accounts) is an ops action, not a code fix; this makes it unnecessary for reads.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10094?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
@@ -2,12 +2,13 @@
   "files": {
     "backend/routers/apps.py": 2335,
     "backend/routers/chat.py": 1577,
-    "backend/routers/developer.py": 2184,
+    "backend/routers/developer.py": 2195,
     "backend/routers/mcp_sse.py": 1857,
     "backend/routers/sync.py": 2024,
     "backend/routers/users.py": 2046
   },
   "raise_justifications": {
+    "backend/routers/developer.py": "GET /v1/dev/user/memories serves the authoritative legacy memories collection for legacy-cohort accounts with no rollout state (#9892), keeping the narrow deny/fallback decision in the route that owns the read contract.",
     "backend/routers/chat.py": "PTT stereo rejection keeps the serving STT provider boundary explicit in the established chat admission owner.",
     "backend/routers/sync.py": "Fresh Sync admission and final Cloud Tasks ledger recovery retain their coupled task, lock, ledger, and staged-blob lifecycle in the existing ingestion owner rather than a new module.",
     "backend/routers/users.py": "GET /v1/users/subscription serializes the new mobile plus/max plans as `unlimited` for clients whose plan enum predates them, so day-one buyers read as paid instead of Free (mirrors the existing operator remap); real limits/grandfather are computed from the true plan first."
--- a/backend/routers/developer.py
+++ b/backend/routers/developer.py
@@ -80,6 +80,7 @@ from utils.memory.default_read_rollout import (
     guard_legacy_memory_write,
     read_default_read_rollout,
 )
+from utils.observability import record_fallback
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


undivisible (approved):
Verified rebased onto main, ratchet conflict resolved, local unit gates green, CI green.

Narrow unenrolled legacy-cohort fallback on GET /v1/dev/user/memories is correct: only missing_rollout_state after pin_memory_system LEGACY falls through; other denies stay 403. Vector route untouched. Failure-Class: none accepted.

Merge readiness: regular merge (no squash).

PR #10095, opened 2026-07-20
fix(mcp): serve legacy memories when a legacy-cohort account has no rollout state

### LINKED ISSUE #9892: Developer API: GET /v1/dev/user/memories returns 403 missing_rollout_state (conversations/action-items OK)

aryanorastar:
# Summary

- MCP memory reads (`GET /v1/mcp/memories`, `GET /v1/mcp/memories/search`, and the SSE `get_memories` / `search_memories` tools) no longer return a silent empty result for accounts that were never enrolled in the canonical-memory rollout: they now serve the authoritative legacy `memories` surface.
- All four sites route the decision through one shared helper, `mcp_legacy_read_authorized`, instead of four copies of the `!= USE_LEGACY_SAFE` comparison. Every other deny reason (malformed state, read failure, missing grant, shadow-only) stays fail-closed.
- The mode change is recorded through the shared `record_fallback` helper — no new one-off counter.

MCP-surface counterpart of #10094 (Developer API), same root cause. Fixes #9892 for the MCP surface.

# Root cause

Same divergence as #10094: `pin_memory_system` resolves every account outside the code-reviewed canonical whitelist to `MemorySystem.LEGACY`, where `users/{uid}/memory_control/state` is *expected* to be absent — but `read_default_read_rollout` treats the absent doc as `DENY_MEMORY` / `missing_rollout_state`. The MCP sites reach that branch only after the cohort pin already said LEGACY, then returned `[]` instead of reading the legacy collection the selector had just declared authoritative. The Developer API surfaced this as a 403; MCP hid it as an empty memories list.

# Failure class

Failure-Class: none

Same call as #10094: no registered class covers a reader failing closed on state whose absence the authoritative cohort selector defines as normal, and FC-split-mutation-authority is scoped to mutable-state transition ownership, not read-side policy.

# Durable guard

- One shared decision owner: the four call sites can no longer drift apart — the narrow un-enrolled exception and its telemetry live in `mcp_legacy_read_authorized`, whose docstring names the contract.
- The fallback is narrow: only `missing_rollout_state` (doc absent = never enrolled) qualifies; `malformed_rollout_state` and `rollout_read_failed` stay fail-closed because there the doc exists or its state is unknown.
- The existing source-order tripwires (labeled as such) now assert all four routes gate legacy reads through the helper.

# Product invariants affected

- INV-MEM-1 (path-matched via `utils/mcp_memories.py`): no tier vocabulary change — the helper only decides legacy-vs-deny routing; tier semantics and the guard test are untouched.
- INV-MEM-3 note: the new fallback runs only on the non-canonical branch (cohort pin already resolved LEGACY); canonical selection still never falls back to legacy.

# Validation

- Behavioral regression test of the helper's full decision table (`test_mcp_legacy_read_authorized_only_for_explicit_or_unenrolled`): explicit legacy-safe and never-enrolled pass; malformed/read-failed/no-grant/shadow-only fail closed — this test cannot pass on pre-fix code (the helper did not exist, and the old inline comparisons denied the un-enrolled case).
- Repo runner (file-isolated, as CI executes): `tests/unit/test_mcp_memory_adapter.py` + `tests/unit/test_mcp_data_endpoints.py` → 27 passed.
- `black` clean on all changed files.
- `backend/routers/mcp_sse.py` frozen-size baseline raised 1857 → 1858 (one import line) with justification, per the ratchet's documented raise path.

Not exercised live: a production MCP key against a real un-enrolled account (no credentials here). The fix point is the shared server-side decision helper driven directly by the tests.

# Out of scope, named

- `should_use_legacy_fallback` properties on the two result dataclasses remain for API compatibility; the routes now use the shared helper instead.
- Chat/omi_chat consumer surfaces keep their existing rollout semantics — different consumer contract, untouched.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10095?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
@@ -3,7 +3,7 @@
     "backend/routers/apps.py": 2335,
     "backend/routers/chat.py": 1577,
     "backend/routers/developer.py": 2195,
-    "backend/routers/mcp_sse.py": 1857,
+    "backend/routers/mcp_sse.py": 1858,
     "backend/routers/sync.py": 2024,
     "backend/routers/users.py": 2046
   },
@@ -11,7 +11,8 @@
     "backend/routers/developer.py": "GET /v1/dev/user/memories serves the authoritative legacy memories collection for legacy-cohort accounts with no rollout state (#9892), keeping the narrow deny/fallback decision in the route that owns the read contract.",
     "backend/routers/chat.py": "PTT stereo rejection keeps the serving STT provider boundary explicit in the established chat admission owner.",
     "backend/routers/sync.py": "Fresh Sync admission and final Cloud Tasks ledger recovery retain their coupled task, lock, ledger, and staged-blob lifecycle in the existing ingestion owner rather than a new module.",
-    "backend/routers/users.py": "GET /v1/users/subscription serializes the new mobile plus/max plans as `unlimited` for clients whose plan enum predates them, so day-one buyers read as paid instead of Free (mirrors the existing operator remap); real limits/grandfather are computed from the true plan first."
+    "backend/routers/users.py": "GET /v1/users/subscription serializes the new mobile plus/max plans as `unlimited` for clients whose plan enum predates them, so day-one buyers read as paid instead of Free (mirrors the existing operator remap); real limits/grandfather are computed from the true plan first.",
+    "backend/routers/mcp_sse.py": "MCP SSE memory reads route their legacy-fallback decision through the shared mcp_legacy_read_authorized helper (#9892); one import line."
   },
   "threshold": 1500
 }
--- a/backend/routers/mcp.py
+++ b/backend/routers/mcp.py
@@ -54,6 +54,7 @@ import utils.mcp_action_items as mcp_action_items
 from utils.mcp_memories import (
     collect_filtered_memories,
     list_default_mcp_memories,
+    mcp_legacy_read_authorized,
     parse_mcp_bool,
     parse_mcp_datetime,
     parse_mcp_int,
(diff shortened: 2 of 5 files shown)


undivisible (approved):
Rebased onto main after #10094. Ratchet conflict resolved (developer.py 2195 + mcp_sse.py 1858). Local unit gates green; CI green after rebase.

Shared mcp_legacy_read_authorized is the right owner for the four MCP list/search sites. Only USE_LEGACY_SAFE and DENY+missing_rollout_state open legacy; other denies stay empty/fail-closed. Telemetry via record_fallback.

Regular merge (no squash).

PR #10023, opened 2026-07-19
fix(listen): enforce live STT model capabilities

Git-on-my-level:
## Summary

- Make live STT selection model-aware: the deployed Parakeet RNNT model is English-only, while the batch TDT model retains its broader language set.
- Treat the client Parakeet setting as a capability-constrained preference, remove the runtime override that could bypass selection, and emit bounded fallback telemetry whenever live routing moves from Parakeet to Modulate.
- Lock the deployment-model contract to both Parakeet charts, add a hermetic `/v4/listen` → selector → Modulate WebSocket → pusher regression, and configure the Parakeet-only listen-pusher gauntlet with an explicit English single-language fixture.
- Document the live-provider boundary.

Closes #10009

Failure-Class: none

## Guard rationale

The chart-to-policy test is deliberately specific rather than a shared YAML primitive: it validates the named deployed Parakeet model IDs against the provider capability policy. It would have caught the model/surface mismatch reported in #10009.

## Follow-up

- #10021 tracks making hermetic CI checks merge-required; the checks exist and run in PR CI, but GitHub branch protection currently does not require them.
- #10022 tracks aligning the multi-language preference gate with the broader live-provider capability policy.

## Verification

- `cd backend && env BACKEND_PYTEST_WORKERS=4 bash test.sh` — passed (675 backend unit test files).
- Focused selector, runtime, policy, fallback-observability, and hermetic listen-pusher wire suites — passed.
- `cd backend && .venv/bin/python -m pytest -q testing/e2e/test_listen_stt.py testing/e2e/test_listen_pusher_wire_contract.py` — 18 passed.
- `bash backend/testing/e2e/run.sh -q --tb=short` — 112 passed, 3 skipped.
- `npm run test:listen-pusher-stack:emulator` — passed locally after the explicit single-language fixture correction.
- `cd backend && .venv/bin/python -m pytest -q tests/unit/test_streaming_deepgram_backoff.py tests/unit/test_listen_runtime_regressions.py tests/unit/test_listen_pusher_stack_ci_wiring.py` — 89 passed.
- `cd backend && bash scripts/typecheck.sh` — 0 errors (existing warnings only).
- `cd backend && .venv/bin/python scripts/scan_async_blockers.py --dirs routers utils` — passed with 0 selected blocking findings.
[...]

PR #10097, opened 2026-07-20
fix(users): gate multi-language mode by the live STT capability policy

### LINKED ISSUE #10022: Align multi-language preference eligibility with live STT capability policy
`backend/routers/users.py:827` and `app/lib/providers/home_provider.dart:12` still derive multi-language eligibility from the legacy Deepgram language list, while the live STT policy now identifies a broader set of Modulate auto-detect-capable languages. This can leave supported users unable to select the live multi-language mode; the discrepancy was found while closing #10009. Keep the preference gate and the provider capability source aligned, with hermetic coverage for the supported and unsupported boundary.

aryanorastar:
# Summary

- `PATCH /v1/users/language` now derives `single_language_mode` from the live STT capability policy (`config.stt_provider_policy.supports_live_multilingual_mode`) instead of the retired Deepgram Nova-3 multi-language list.
- The client mirror (`app/lib/providers/home_provider.dart`) is updated to the same base-code set with the same regional-variant normalization, and documents which policy file it mirrors.
- Users whose language the live Modulate auto-detection supports but the old 19-locale Deepgram list did not — Vietnamese, Korean, Turkish, Arabic, Thai, and ~40 others — can now actually enter live multi-language mode.

Fixes #10022.

# Root cause

Two stale copies of a capability decision. #10023 moved live multi-language capability to the policy boundary (`supports_live_multilingual_mode`, consumed by the serving path in `utils/stt/streaming.py`), but the *preference gate* — the route that decides `single_language_mode` when a user picks a language — still consulted `deepgram_nova3_multi_languages`, as did its client-side mirror. A supported user was silently locked into single-language mode at preference-set time, before serving policy ever ran.

# Failure class

Failure-Class: none

The nearest theme is again divergent read-side policy copies rather than mutable-state transition ownership (FC-split-mutation-authority's scope) — same call as #10094/#10095.

# What stays untouched, and why

- `utils/stt/streaming.py`'s Deepgram runtime branch keeps `deepgram_nova3_multi_languages`: that code selects Deepgram serving parameters, the one place the Deepgram-specific vocabulary legitimately lives.
- `migrations/006_auto_set_transcription_mode.py` is a historical migration; rewriting it would change what a past migration meant.

# Product invariants affected

# Validation

- New coverage extends `tests/unit/test_stt_provider_policy.py` (the guard surface for this policy boundary): behavioral checks that the live policy admits the newly eligible languages (`vi`, `vi-VN`, `ko`, `tr`, `ar`, `th`, `pt-BR`, `en`) and rejects unsupported ones (`my`, `am`, `lo`), plus a **labeled static tripwire** that the route derives `single_language_mode` from `supports_live_multilingual_mode` and that `deepgram_nova3_multi_languages` no longer appears in the route module. The tripwire fails on pre-fix code by construction.
- The route wiring itself is one line; a full route-handler harness for `routers/users.py`'s heavy import graph would have required module-scope `sys.modules` stubbing, which the backend-module-isolation checker bans outside the frozen legacy allowlist — hence policy-seam behavioral tests + tripwire, per the repo's testing guidance.
- Repo runner (file-isolated, as CI executes): `tests/unit/test_stt_provider_policy.py` → 20 passed.
- Dart unit test (`app/test/unit/multilingual_eligibility_test.dart`) covers the client mirror's eligibility set and normalization. **Not run locally — this machine has no Flutter SDK**; the test is hermetic and the app CI lane discovers `app/test/unit/`. Backend coverage of the same decision did run locally.
- `black` clean on changed backend files.

Not exercised live: the full app→backend language-selection flow on a device (no Flutter toolchain here). The behavior change is the server-side gate, covered by the policy-seam tests and tripwire above.

# Out of scope, named

- The duplicated client list still exists by necessity (Dart cannot import the Python policy); a follow-up could serve eligibility from the backend response instead of mirroring the set. The route response already returns `single_language_mode`, so clients that trust the server need no list at all — the mirror only affects optimistic local state.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10097?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Verified the core invariant of this fix: the new client `multiLanguageSupported` set is byte-for-byte identical to the backend `config.stt_provider_policy.MODULATE_SUPPORTED_LANGUAGES` (61 base codes on each side, same membership), and the regional-variant normalization (`split('-').first.split('_').first.toLowerCase()`) matches `normalized_stt_language` exactly. So the client mirror and the route gate now both agree with the live serving policy — the stale Deepgram Nova-3 copy that was silently locking vi/ko/tr/ar/th (and ~40 others) into single-language mode is gone from the preference path.

Nice touches:
- Keeping `deepgram_nova3_multi_languages` only in `utils/stt/streaming.py` (Deepgram serving params) and leaving the historical migration untouched is the right call.
- The static tripwire asserting the route derives `single_language_mode` from `supports_live_multilingual_mode` and that `deepgram_nova3_multi_languages` no longer appears in `routers/users.py` will catch future regressions of exactly this class.

One observation (non-blocking): the duplicated client list is unavoidable today since Dart can't import the Python policy, but the route already returns `single_language_mode` in its response — so the app could eventually trust the server response for local state and drop the mirror entirely. Worth a follow-up but correctly out of scope here.

Not formally approving because this sits on the STT provider-routing / user-preference-persistence boundary; flagging for a maintainer's final call. The fix itself looks correct and well-tested.

---

kodjima33 (approved):

aryanorastar:
@Git-on-my-level @undivisible This PR is fully approved and ready to merge. Could you please trigger the merge when you have a moment? Thank you!

undivisible:
looks good ty :))

aryanorastar:
Thanks! On the non-blocking observation — the app could trust the server's `single_language_mode` and drop the mirror entirely — I did exactly that in #10126, stacked on this branch. Merge this first and #10126 collapses to just the mirror deletion.

undivisible:
holy em dash

PR #10102, opened 2026-07-20
Reject an oversized status or source filter before it reaches Firestore

ZachL111:
`GET /v1/conversations` and `GET /v1/conversations/count` both split a comma-separated `statuses`
filter (and the count route also a `sources` filter) into a list that reaches an unchunked
Firestore `in` filter in `database/conversations.py`:

```python
conversations_ref.where(filter=FieldFilter('status', 'in', statuses))
```

Firestore rejects an `in` filter with more than 30 values. The repo already knows this and guards
it elsewhere:

- `database/apps.py`: `# Firestore 'in' limited to 30 items`
- `database/chat.py`: `# Firestore IN operator supports max 30 values, so chunk the queries`

These two list/count endpoints do not, and nothing wraps the `stream()` call, so a request
repeating the query key past thirty (`?statuses=a,a,...`) raises out of the Firestore client and
surfaces as an unhandled **HTTP 500 on the core conversation-list route the app hits constantly**.

`ConversationStatus` has five legitimate values and the source set is small, so this is
malformed-input surface, not a shape any real client sends.

## The fix

A shared helper rejects an oversized filter with 400 before the query is built, applied to
`statuses` on both endpoints and to `sources` on the count endpoint. It rejects rather than clamps,
matching this file's own idiom (it already raises 400 for `start_date > end_date` and for combining
`statuses`+`sources`). With at most a handful of valid values, a cap of 20 cannot reject a request
any real client would send.

The check is in the handler body, not `Query(..., max_length=20)` on the parameter, so the
generated OpenAPI does not move (`docs/api-reference` is a compatibility boundary): the parameter
declarations are byte-identical to origin/main. That also keeps the bound testable, since a
`Query`-level constraint is enforced by FastAPI request parsing that a direct handler call bypasses.

## Tests

`tests/unit/test_conversations_status_filter_cap.py`:

- an oversized `statuses` filter on the list endpoint is rejected with 400 before the query
- an oversized `statuses` filter on the count endpoint is rejected with 400 before the query
- an oversized `sources` filter on the count endpoint is rejected with 400 before the query
- normal `statuses` filters reach the database unchanged on both endpoints

The fake standing in for the database seam mirrors Firestore by raising above thirty values, so the
tests assert the guard fires before the query is built rather than only checking the response.

## Verification

Run in `backend/`:

- `pytest tests/unit/test_conversations_status_filter_cap.py`: 5 passed
- prove-fail: with `routers/conversations.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), the three oversized tests fail with
  `Exception: 'in' filters support a maximum of 30 elements.` while the two normal-path tests pass
  both ways. Re-applied: 5 passed
- pytest with `test_conversation_events_bounds.py` and `test_conversation_action_items_count.py`:
  14 passed
- the test imports `routers.conversations` at module scope, so the heavy router import is collection
  cost rather than per-test CPU; the file passes under the strict duration guard
  (`BACKEND_FAST_UNIT_FAIL_SECONDS=0.12`, exit 0)
- the diff changes no route signature or `response_model`, so no OpenAPI/generated-client
  regeneration is needed
- `pyright routers/conversations.py`: 0 errors
- `routers/conversations.py` is not in the product file line-count ratchet baseline

## Related

This is the first-party counterpart of the same Firestore `in`-overflow class fixed on the
integration route in a separate PR (`GET /v2/integrations/{app_id}/conversations`). Same cause,
different route and parameter shape (comma-separated string here vs `List[str]` Query there).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/conversations.py
+++ b/backend/routers/conversations.py
@@ -336,6 +336,18 @@ def _ensure_aware(value: datetime) -> datetime:
     return value if value.tzinfo is not None else value.replace(tzinfo=timezone.utc)
 
 
+# Firestore 'in' filters accept at most 30 values (see database/apps.py, database/chat.py). Reject
+# an oversized status/source filter before it reaches the query so a caller cannot turn a
+# comma-separated filter into an unhandled 500. ConversationStatus and the source set are both
+# tiny, so a cap of 20 can never reject a request a real client would send.
+MAX_IN_FILTER_VALUES = 20
+
+
+def _reject_oversized_filter(values: List[str], field_name: str) -> None:
+    if len(values) > MAX_IN_FILTER_VALUES:
+        raise HTTPException(status_code=400, detail=f"{field_name} accepts at most {MAX_IN_FILTER_VALUES} values")
+
+
 @router.get(
     '/v1/conversations',
     response_model=List[Conversation],
@@ -363,12 +375,15 @@ def get_conversations(
     if len(statuses) == 0:
         statuses = "processing,completed"
 
+    status_filter = statuses.split(",") if len(statuses) > 0 else []
+    _reject_oversized_filter(status_filter, "statuses")
+
     conversations = conversations_db.get_conversations_without_photos(
         uid,
         limit,
         offset,
         include_discarded=include_discarded,
-        statuses=statuses.split(",") if len(statuses) > 0 else [],
+        statuses=status_filter,
         start_date=start_date,
         end_date=end_date,
         folder_id=folder_id,
@@ -394,6 +409,8 @@ def get_conversations_count(
         raise HTTPException(status_code=400, detail="start_date must be earlier than or equal to end_date")
     status_list = [s.strip() for s in statuses.split(',') if s.strip()] if statuses else []
     source_list = [s.strip() for s in sources.split(',') if s.strip()] if sources else []
+    _reject_oversized_filter(status_list, "statuses")
+    _reject_oversized_filter(source_list, "sources")
     if status_list and source_list:
         # Combining status+source `in` filters would need a composite index; keep them exclusive.
         raise HTTPException(status_code=400, detail="statuses and sources filters cannot be combined")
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @ZachL111 — nice, well-scoped fix.

This fills a genuine gap: `GET /v1/conversations` and `/count` split a comma-separated `statuses`/`sources` into a list that reaches an unchunked Firestore `in` filter, and Firestore rejects `in` filters above 30 values. The repo already guards this exact limit in `database/apps.py` and `database/chat.py`, but the conversation endpoints did not, so a malformed `?statuses=a,a,…` would raise out of the Firestore client and surface as an unhandled 500 on a core list route. Capping at 20 and rejecting with 400 (rather than silently clamping) matches the file's existing idiom (it already 400s on `start_date > end_date` and on combining `statuses`+`sources`), and since ConversationStatus has only five valid values this can never reject a real client request.

A few things that read well on review:
- The guard is in the handler body rather than `Query(..., max_length=...)`, so the OpenAPI surface (a compatibility boundary) stays byte-identical to main — confirmed, the Public Developer API contract check passed.
- The tests don't just assert HTTP 400; they use a Firestore-like fake that raises above 30, proving the guard fires *before* the query is built. All 5 passed in the Backend unit suite.
- Single cohesive concern, no route-signature/response_model changes, no dependency/workflow/auth/migration surface touched.

Minor note for a maintainer (not blocking): the literal `ENCRYPTION_SECRET` in the new test file is the repo's standard public dev/test fixture (it appears verbatim across ~161 files including `backend/test.sh`), so it's fine to keep as-is for consistency with neighboring unit tests. Flagging only so it's a conscious choice rather than an oversight.

Marking this as a clean candidate. I can't formally approve here (a maintainer should), but on merit this looks small, safe, and ready to merge once a human signs off.

---

kodjima33 (approved):

PR #10104, opened 2026-07-20
Abort the merge instead of destroying a source's audio when its chunks fail to copy

ZachL111:
When two conversations are merged, a transient failure copying one source's audio silently
destroys that source's raw recording.

`perform_merge_async` copies every source conversation's audio chunks to the merged conversation
via `_copy_audio_chunks_for_merge`, then in step 9 unconditionally deletes every source's original
chunks (`_delete_conversation_and_related_data` -> `delete_conversation_audio_files`). The outer
handler that would abort into `_handle_merge_failure` only runs if an exception propagates out of
the copy.

The copy helper wrapped each conversation's copy in a per-conversation try/except that only logged
the error:

```python
try:
    chunks = list_audio_chunks(uid, conv_id)
    for chunk in chunks:
        has_chunks = True
        ...
        bucket.copy_blob(source_blob, bucket, new_path)
except Exception as e:
    logger.error(f"Error copying chunks for {conv_id}: {e}")
```

Because `has_chunks` can already be `True` from an earlier source that copied fine, a later source
whose `list_audio_chunks` or `copy_blob` throws (a transient GCS error, no adversarial input) leaves
the helper returning a normal non-empty result, indistinguishable from full success. The merge then
runs to completion and deletes that source's original audio, which was never copied anywhere.

**The user's raw recording is permanently destroyed**, with a merge-completed FCM notification and
only a log line as the record. Unlike derived data (action items, memories), raw audio cannot be
regenerated.

## The fix

Remove the swallowing try/except so a copy failure propagates. That makes `perform_merge_async`
abort into `_handle_merge_failure` before any source is deleted. That handler is the same recovery
path the merge already uses for every other mid-merge failure, so preserving all sources on a copy
failure is the existing contract, not new behavior. A source that genuinely has no chunks is an
empty list, not an exception, so it still does not abort.

This is a data-safety change in the conversation-merge path. It makes deletion strictly safer (it
never deletes audio that was not successfully copied); it does not add or change any deletion.

## Tests

`tests/unit/test_merge_audio_chunk_copy_failure.py`:

- a later source whose chunk listing throws propagates even after an earlier source copied fine
- a later source whose `copy_blob` throws propagates the same way
- an all-success merge still returns the created audio-file records unchanged
- a merge where no source has chunks returns empty without aborting

The fakes mirror Firestore/GCS by raising, so the tests assert the failure reaches the caller
rather than only checking a return value.

## Verification

Run in `backend/`:

- `pytest tests/unit/test_merge_audio_chunk_copy_failure.py`: 4 passed
- prove-fail: with `utils/conversations/merge_conversations.py` reverted to origin/main and
  confirmed byte-identical (`git diff origin/main` empty), the two propagation tests fail with
  `Failed: DID NOT RAISE <class 'Exception'>` while the two normal-path tests pass both ways.
  Re-applied: 4 passed
- this file is a `conversation_lifecycle` workflow-contract source, so the required contract tests
  were run: `test_conversation_lifecycle_contract.py` and
  `test_check_conversation_lifecycle_writes.py` 20 passed, and
  `scripts/check_conversation_lifecycle_writes.py` passes
- pytest with all `tests/unit/test_merge*.py`: 49 passed
- `pyright utils/conversations/merge_conversations.py`: 0 errors

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10104?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_merge_audio_chunk_copy_failure.py
+++ b/backend/tests/unit/test_merge_audio_chunk_copy_failure.py
@@ -1,0 +1,100 @@
+"""A failed audio-chunk copy during merge must abort, not silently drop a source's audio.
+
+perform_merge_async copies every source conversation's audio chunks to the merged conversation
+(_copy_audio_chunks_for_merge), then in step 9 unconditionally deletes every source's original
+chunks. The copy helper used to wrap each conversation's copy in a try/except that only logged the
+error. Because `has_chunks` could already be True from an earlier source that copied fine, a later
+source whose copy threw (a transient GCS error, no adversarial input) left the helper returning a
+normal non-empty result. The merge then completed and deleted that source's original audio, which
+was never copied anywhere: the user's raw recording is destroyed with a merge-completed
+notification and only a log line as the record.
+
+The fix lets the failure propagate so perform_merge_async aborts into _handle_merge_failure before
+any source is deleted. These tests assert the propagation happens even after an earlier source
+copied successfully, which is the case the swallowing except masked.
+"""
+
+import os
+
+os.environ.setdefault("ENCRYPTION_SECRET", "omi_ZwB2ZNqB2HHpMK6wStk7sTpavJiPTFg7gXUHnc4tFABPU6pZ2c2DKgehtfgi4RZv")
+os.environ.setdefault("OPENAI_API_KEY", "sk-test-not-real")
+
+from unittest.mock import MagicMock
+
+import pytest
+
+import utils.conversations.merge_conversations as merge
+
+
+class _FakeBucket:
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):

PR #10106, opened 2026-07-20
Skip external task sync when an action item is already exported

ZachL111:
## Problem

`POST /v1/action-items` is content-idempotent on the Firestore document: a retried request with
the same (uid, description) resolves to the same action-item id via
`create_action_item(..., idempotency_key=...)`, so it does not create a duplicate document. The
router documents this.

But regardless of whether that call created a new document or returned a pre-existing one, the
router unconditionally submits `_run_auto_sync`, which calls `auto_sync_action_item` then
`_sync_to_cloud_service` then `create_task_internal`. `create_task_internal` does a real POST to the
user's connected Todoist, Asana, ClickUp, or Google Tasks account, and it has no idempotency key of
its own.

So a retried create (network timeout, client retry, duplicate delivery) correctly dedups the
Firestore document but creates a second duplicate task in the user's external task app. That is a
persistent, user-visible artifact they have to find and delete by hand.

`_sync_to_cloud_service` sets `exported: True` on the action item only after the external call
succeeds, and nothing reads that flag before the call, so the mark never prevents a repeat.

## Fix

Re-read the persisted `exported` flag at the top of `_sync_to_cloud_service` and skip the external
call when the item was already exported.

The re-read is required rather than trusting the passed dict, because the router passes the request
payload (which never carries `exported`), and the flag is only set on the persisted document by a
prior successful sync.

`_sync_to_cloud_service` is the single shared helper used by both `auto_sync_action_item` (the
router-create path and the candidate-dispatch path) and `auto_sync_action_items_batch`
(conversation extraction). The router-create path has no dedup guard today, so this closes it. The
candidate-dispatch path already holds a `claim_candidate_integration_dispatch` lease, so there this
is defense-in-depth.

The result is reported as `synced: True` on the skip, not `False`, because the candidate-dispatch
path treats a `synced: False` result with an unrecognized reason as a dispatch failure and would
requeue it. Already-exported means the sync goal is already satisfied, so success is the honest and
safe report.

This does not add a cross-connection lease, so it does not by itself close a truly concurrent
duplicate delivery (two overlapping requests that both read `exported` as false before either
writes it). It closes the dominant sequential-retry case, where the first request has completed and
set the flag before the retry's sync runs. A lease on this path, matching the candidate-dispatch
pattern, is the larger follow-up for the concurrent window.

## Tests

`tests/unit/test_task_sync_cloud_export_dedup.py` (new), using the sanctioned
`stub_modules`/`load_module_fresh` seam so it does not import the real `google.cloud.firestore`
chain:

- a retried cloud sync creates the external task exactly once and reports already_exported on the retry
- a batch sync skips the item already marked exported and syncs only the fresh one
- a first-time cloud sync still creates the external task and marks the item exported

## Verification

Run in `backend/`:

- `pytest tests/unit/test_task_sync_cloud_export_dedup.py`: 3 passed
- prove-fail: with `utils/task_sync.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), the retry and batch tests fail with `create_task_internal` called
  twice (the duplicate external task), while the first-sync test passes. Reapplied: 3 passed
- `pytest test_task_sync_async.py test_task_sync_apple_async_boundary.py test_action_item_idempotency.py test_task_sync_cloud_export_dedup.py`: 21 passed
- `pytest test_workstream_core.py test_workstream_router_contract.py test_candidate_lifecycle.py test_candidates_router.py`: 111 passed (the workflow-contract and candidate-dispatch suites that consume this path)
- `pyright -p pyrightconfig.json utils/task_sync.py`: 0 errors, unchanged from the origin/main baseline

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_task_sync_cloud_export_dedup.py
+++ b/backend/tests/unit/test_task_sync_cloud_export_dedup.py
@@ -1,0 +1,120 @@
+import asyncio
+from pathlib import Path
+from types import ModuleType
+from typing import Any
+
+from testing.import_isolation import load_module_fresh, stub_modules
+
+BACKEND_DIR = Path(__file__).resolve().parents[2]
+
+
+def _module(name: str, **attributes: Any) -> ModuleType:
+    module = ModuleType(name)
+    for key, value in attributes.items():
+        setattr(module, key, value)
+    return module
+
+
+def _stubs(store: dict[str, dict[str, Any]], create_calls: list[dict[str, Any]], *, default_app: str = 'todoist'):
+    """Stub the database + integration ops so utils.task_sync loads without google.cloud.firestore.
+
+    ``store`` maps action-item id to its persisted Firestore document. The stubbed
+    ``get_action_item`` reads from it and ``update_action_item`` writes the export fields back,
+    so a second auto-sync sees ``exported=True`` exactly like a retry against real Firestore.
+    """
+
+    def get_action_item(_uid: str, item_id: str):
+        return store.get(item_id)
+
+    def update_action_item(_uid: str, item_id: str, updates: dict[str, Any]) -> None:
+        store.setdefault(item_id, {}).update(updates)
+
+    async def create_task_internal(**kwargs: Any) -> dict[str, Any]:
+        create_calls.append(kwargs)
+        return {'success': True, 'external_task_id': f'ext-{len(create_calls)}'}
+
+    return {
+        'database.users': _module(
+            'database.users',
+            get_default_task_integration=lambda _uid: default_app,
+            get_task_integration=lambda _uid, _app: {'connected': True},
+        ),
+        'database.action_items': _module(
+            'database.action_items',
+            get_action_item=get_action_item,
+            update_action_item=update_action_item,
+            batch_set_sync_requested=lambda *_a, **_k: None,
+        ),
+        'utils.notifications': _module(
+            'utils.notifications',
+            send_apple_reminders_sync_push_async=lambda **_k: None,
+        ),
+        'utils.task_integrations_ops': _module(
+            'utils.task_integrations_ops',
+            create_task_internal=create_task_internal,
+        ),
+    }
+
+
+def test_retried_cloud_sync_does_not_create_a_second_external_task() -> None:
+    store = {'task-1': {'id': 'task-1', 'description': 'Plan launch'}}
+    create_calls: list[dict[str, Any]] = []
+
+    with stub_modules(_stubs(store, create_calls)):
+        task_sync = load_module_fresh('utils.task_sync', str(BACKEND_DIR / 'utils' / 'task_sync.py'))
+
+        async def exercise():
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Reviewed by automated PR reviewer — approving; merging once rerun CI is green.

Git-on-my-level:
Thanks @ZachL111 — this is a well-diagnosed and correct fix for a real user-visible bug. The re-read of the persisted `exported` flag is the right approach (the router passes the request payload, which never carries `exported`), and reporting `synced: True` on the skip is correct since the candidate-dispatch path would otherwise requeue on an unrecognized `synced: False` reason. Your disclosure of the concurrent-delivery window (vs. the sequential-retry case this closes) is accurate and appreciated.

For CI context: your new `test_task_sync_cloud_export_dedup.py` passes all 3 tests in the Backend unit suite. The suite's overall failure comes from `test_ws_b_short_term_lifecycle.py`, which is a pre-existing failure on `main` (same 7 `ValidationError` failures in `canonical_required_processing`), not caused by this change.

Since this touches user-data persistence and external-integration calls (Todoist/Asana/ClickUp/Google Tasks), I'm flagging it for a human maintainer to review before merge rather than approving outright. No changes requested from me — the fix looks good.

---

kodjima33 (approved):
Backend bug fix (skip external task sync when action item already exported — dedup) — approve only; red-main Backend unit suite + Metadata Preflight failing

Git-on-my-level:
Thanks @ZachL111 — verified and this holds up well. A few independent confirmations for the record:

- **Candidate-dispatch reasoning is correct.** I traced `backend/utils/task_intelligence/candidate_service.py`: `succeeded = bool(result.get('synced')) or terminal_noop`, and `already_exported` is not in the terminal_noop set, so reporting `synced: True` here is the honest call — a `synced: False` would have requeued an already-satisfied item into `retry_queue`.
- The re-read (rather than trusting the passed dict) is the right move since the router hands down the request payload, which never carries `exported`.
- The stub seam (`stub_modules`/`load_module_fresh`) faithfully exercises the Firestore round-trip; the three cases (retry, batch-skip, first-sync) cover the behavior well.
- Both red checks at this head are unrelated to the change: the backend unit failure is in `test_ws_b_short_term_lifecycle.py` (a memory/workstream timing test), and the PR Metadata Preflight trips on `product-invariants` because CI evaluates the full 279-file fork diff vs main rather than the two changed paths.

Good scoping note on the concurrent-window lease as the follow-up. Leaving this for a maintainer to approve/merge once checks are re-run green.

---

PR #10107, opened 2026-07-20
Preserve a legitimate zero start timestamp on mentor segments

ZachL111:
## Problem

`process_mentor_notification` in `utils/mentor_notifications.py` buffers realtime transcript
segments and, when enough accumulate, returns them sorted by timestamp as the conversation context
for the proactive mentor-notification LLM (`evaluate_relevance` / `generate_notification` /
`validate_notification`).

Each segment's timestamp came from:

```python
timestamp = segment.get('start', 0) or current_time
```

A transcript segment's `start` is relative capture seconds. The first utterance of a capture
legitimately begins at `start == 0.0`, and `0.0` is falsy, so `... or current_time` replaced that
real timestamp with wall-clock time (`time.time()`, about 1.78e9). The opening line was then sorted
to the END of the timestamp-ordered context handed to the mentor LLM, and it stays pinned last
across later analysis windows because every later relative start is far below wall-clock time.

The segments reaching this function come from the live realtime path
(`app_integrations._async_trigger_realtime_integrations`, fed by `routers/listen/transcripts.py`
and `routers/pusher.py`), where `start` is relative to session start. The pipeline itself treats
`raw_segments[0]['start']` as a small relative offset (`routers/listen/transcripts.py`), so a first
segment at exactly `0.0` is a normal STT output.

## Scope

This is a correctness bug with modest impact: it only bites when the first buffered segment's
`start` is exactly `0.0`, and the consequence is a mis-ordered conversation context for the mentor
notification (the opening line placed last), which can subtly affect what proactive notification is
generated. It does not crash, lose data, or touch any auth boundary.

## Fix

Preserve a real numeric `start` (including `0.0`) and fall back to wall-clock time only when `start`
is genuinely absent or non-numeric:

```python
raw_start = segment.get('start')
timestamp = raw_start if isinstance(raw_start, (int, float)) else current_time
```

This keeps the original missing-start fallback intent and additionally avoids a latent sort crash
if a non-numeric `start` ever arrived (mixed str/float keys would raise in `sorted`). It is the
same "a valid falsy or edge numeric must not be treated as missing" class as the recent
AudioRingBuffer non-positive `sample_rate` guard.

## Tests

`tests/unit/test_mentor_notification_segment_start_zero.py` (new), using the sanctioned
`stub_modules`/`load_module_fresh` seam so it does not import the real `google.cloud.firestore`
chain:

- a first segment at `start == 0.0` keeps timestamp `0.0` and stays first in the sorted context
- a segment with no `start` still falls back to a wall-clock timestamp

## Verification

Run in `backend/`:

- `pytest tests/unit/test_mentor_notification_segment_start_zero.py`: 2 passed
- prove-fail: with `utils/mentor_notifications.py` reverted to origin/main and confirmed
  byte-identical (`git diff origin/main` empty), the start-zero test fails with
  `assert 1784551851.3 == 0.0` (the opening line stamped with wall-clock time), while the
  missing-start fallback test passes. Reapplied: 2 passed
- `pytest tests/unit/test_mentor_notifications.py tests/unit/test_mentor_notification_segment_start_zero.py`: 47 passed (the full existing mentor suite plus the new file)
- `pyright -p pyrightconfig.json utils/mentor_notifications.py`: 0 errors, 0 warnings

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10107?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_mentor_notification_segment_start_zero.py
+++ b/backend/tests/unit/test_mentor_notification_segment_start_zero.py
@@ -1,0 +1,80 @@
+"""Regression: a mentor segment whose ``start`` is a legitimate 0.0 (the first utterance of a
+capture) must keep that timestamp, not be stamped with wall-clock time and sorted to the end of the
+chronologically ordered mentor context handed to the proactive-notification LLM.
+
+``utils.mentor_notifications`` imports ``database.notifications`` (Firestore) at import time, so it
+is loaded through the sanctioned ``stub_modules`` / ``load_module_fresh`` seam rather than imported
+directly. Each load gets a fresh module-global ``message_buffer``, so no state leaks between tests.
+"""
+
+from contextlib import contextmanager
+from pathlib import Path
+from types import ModuleType
+from typing import Any, Iterator
+
+from testing.import_isolation import load_module_fresh, stub_modules
+
+BACKEND_DIR = Path(__file__).resolve().parents[2]
+
+
+def _module(name: str, **attributes: Any) -> ModuleType:
+    module = ModuleType(name)
+    for key, value in attributes.items():
+        setattr(module, key, value)
+    return module
+
+
+@contextmanager
+def _loaded_mentor(frequency: int = 3) -> Iterator[ModuleType]:
+    stubs = {
+        'database.notifications': _module(
+            'database.notifications',
+            get_mentor_notification_frequency=lambda _uid: frequency,
+        ),
+    }
+    with stub_modules(stubs):
+        mentor = load_module_fresh(
+            'utils.mentor_notifications',
+            str(BACKEND_DIR / 'utils' / 'mentor_notifications.py'),
+        )
+        yield mentor
+
+
+def _spaced_segments(first_start: float) -> list[dict[str, Any]]:
+    """Ten segments (== MIN_NEW_SEGMENTS_FOR_ANALYSIS) whose first starts at ``first_start``.
+
+    Starts are 10s apart and ``is_user`` alternates so none coalesce (the buffer only merges
+    consecutive same-speaker segments within 2s), yielding ten distinct messages that trigger a
+    return.
+    """
+    segments = [{'text': 'opening line', 'start': first_start, 'is_user': True}]
+    for i in range(1, 10):
+        segments.append({'text': f'line {i}', 'start': float(i * 10), 'is_user': i % 2 == 0})
+    return segments
+
+
+def test_first_segment_start_zero_stays_first_in_sorted_context() -> None:
+    with _loaded_mentor() as mentor:
+        result = mentor.process_mentor_notification('user-zero-start', _spaced_segments(0.0))
+
+        assert result is not None
+        # Direct bug assertion: the opening line keeps its real capture start of 0.0. On the unfixed
+        # ``... or current_time`` code this is wall-clock time (~1.78e9).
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Reviewed by automated PR reviewer — approving; merging once rerun CI is green.

Git-on-my-level:
Nice catch — this is a real correctness bug, and the fix is the right idiom.

**What was wrong:** `segment.get('start', 0) or current_time` treated a legitimate `start == 0.0` (the first utterance of a capture, where `start` is relative capture seconds) as falsy, replacing it with wall-clock time (~1.78e9). That opening line then sorted to the *end* of the timestamp-ordered context handed to the mentor-notification LLM, and stayed pinned last across later windows.

**Why the fix is correct:** `raw_start if isinstance(raw_start, (int, float)) else current_time` preserves a real numeric `start` (including `0.0`) and only falls back to wall-clock time when `start` is genuinely absent or non-numeric. This matches the realtime path's contract, where `raw_segments[0]['start']` is a small relative offset and `0.0` is normal STT output. The added regression test covers both the `0.0` case and the missing-`start` fallback, and both pass in CI.

**CI:** The Backend unit suite failure here (`tests/unit/test_ws_b_short_term_lifecycle.py`) is **pre-existing on `main`** — the latest `main` runs fail on that same file — and is unrelated to this change. Your own test file (`test_mentor_notification_segment_start_zero.py`) passes. The cancelled Formatting/Hygiene/Detect Changes checks are downstream of the failing unit job, not separate failures.

**Scope:** Small, cohesive single-concern fix (1 source line + 1 test file). No auth, persistence, privacy, dependency, or workflow surfaces touched. No split needed.

Marking this as a positive signal rather than a formal approval because checks are currently red (even though the failure is unrelated). A maintainer can merge once the unrelated `main`-level test breakage is resolved or a green re-run lands.

Thanks for the clear write-up and the regression test!

---

kodjima33 (approved):
Backend bug fix (preserve legitimate zero start timestamp on mentor segments) — approve only; red-main Backend unit suite + Metadata Preflight; note overlap with #10154

Git-on-my-level:
Checks are now fully green on this head (Backend unit suite, Hermetic Merge Gate, and all relevant suites passing). The earlier `ci-failing-pre-existing` blocker that kept this as a positive signal is resolved — removing that label now.

Approving. Small, cohesive, single-concern correctness fix with a solid regression test covering both the `start == 0.0` case and the missing-`start` fallback. No merge — a maintainer can merge when ready.

Note: this overlaps with #10154 (same bug, different author). A maintainer may want to pick one and close the other with credit.

---

Git-on-my-level (approved):
Approving now that CI is green. The fix correctly preserves a legitimate 0.0 start timestamp; regression test is solid. (Overlap with #10154 noted for maintainer.)

---

ZachL111:
Closing: this exact fix is on main now. utils/mentor_notifications.py guards a zero start offset with an is None check and only falls back to wall clock time for a genuinely absent start, so there is nothing left for this PR to add.

PR #10109, opened 2026-07-20
Persist bulk segment assignment transactionally by segment id

ZachL111:
## Problem

`PATCH /v1/conversations/{conversation_id}/segments/assign-bulk` (`assign_segments_bulk`) reads the
conversation outside any transaction, mutates the in-memory `transcript_segments` array, then
persists by overwriting the entire array via `update_conversation_segments`.
`update_conversation_segments` re-reads the document inside its transaction only to recompute
`has_content` and the protection level, then writes `transcript_segments` wholesale from the
caller's array. So the transaction does not protect the segments from a stale-array overwrite.

If another write to the same conversation's `transcript_segments` commits between the router's read
and this overwrite, it is silently and permanently lost.

Failure scenario (same user, same conversation, two overlapping writes):

1. The conversation is open in two places (two tabs or two devices), or a request is retried.
2. A segment text edit commits via the transactional `update_conversation_segment_text` (edits s1).
3. The bulk-assign request, holding a snapshot read before that edit, assigns a speaker to s2 and
   writes the whole stale array back. s1's edit is gone.

This is same-tenant data loss, so it is a reliability bug, not an authorization issue.

## Why this pattern is already known here

`database.conversations.update_conversation_segment_text` was deliberately made a transactional
read-modify-write by segment id for exactly this class. Its docstring:

> The read-modify-write runs in a Firestore transaction so concurrent edits (e.g. the same
> conversation open in two tabs) can't lose-update each other. Without it, two edits that both read
> the pre-edit transcript_segments array and each rewrite the whole array clobber one another, the
> later write wins and silently drops the earlier edit.

The bulk-assign endpoint reintroduced the whole-array overwrite the sibling was written to avoid.

## Fix

Add `bulk_assign_segment_speakers` in `database/conversations.py`, a transactional read-modify-write
that applies is_user/person_id by segment id inside the transaction, mirroring
`update_conversation_segment_text` (including the `_prepare_conversation_for_read` /
`_prepare_conversation_for_write` encryption boundary). The router calls it instead of the
whole-array overwrite.

The router keeps its in-memory loop for assign-type validation and for the returned `Conversation`
snapshot; only the persistence call changes. Lock enforcement stays where it already is
(`_get_valid_conversation_by_id` raises 402 on a locked conversation before this point), so the
helper does not re-check it. The change is scoped to the cleanest instance; the same
non-transactional overwrite also exists in the two `set_assignee_conversation_segment` endpoints and
is a reasonable follow-up, not folded in here to keep this a small, reviewable fix.

## Tests

`tests/unit/test_conversation_bulk_assign_lost_update.py` (new) drives the real `assign_segments_bulk`
handler and fakes only the database seam, with a stateful fake that models both persistence
primitives' real semantics (whole-array overwrite vs by-id transactional re-read) plus a concurrent
edit that lands right after the handler's read:

- a person_id bulk-assign lands on the target segment and a concurrent edit to a different segment survives
- an is_user bulk-assign behaves the same

## Verification

Run in `backend/`:

- `pytest tests/unit/test_conversation_bulk_assign_lost_update.py`: 2 passed
- prove-fail: with `routers/conversations.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), both tests fail with
  `AssertionError: assert 'original one' == 'edited by other writer'` (the assignment lands but the
  concurrent edit is clobbered). Reapplied: 2 passed
- fake fidelity: the fake `update_conversation_segments` overwrites `transcript_segments` wholesale,
  matching the real function which sets `'transcript_segments': segments` from the caller inside its
  transaction, so the lost update is real production behavior, not a fake artifact
- the conversation_lifecycle workflow contract set (test_conversation_lifecycle_contract,
  test_finalization_decision, test_check_conversation_lifecycle_writes,
  test_conversation_finalization_jobs, test_recording_sessions, test_listen_finalization_cloud_tasks)
  plus test_conversation_revision_contract and the new file: 137 passed
- `python scripts/check_conversation_lifecycle_writes.py`: passed
- `scripts/scan_async_blockers.py --dirs routers utils`: 0 findings (the handler is sync)

## Impact

Users bulk-assigning a person or speaker to segments while a concurrent write to the same
conversation is in flight (two tabs or devices, or a retried assign after a later edit landed). The
window is the read-to-write span, so it is not the dominant case, but it is exactly the two-tab race
the codebase already treats as real for `update_conversation_segment_text`. When it hits, a user's
segment edit is silently and permanently lost with no error.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-database.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-database.json
@@ -1,10 +1,10 @@
 {
   "files": {
-    "backend/database/conversations.py": 1580,
+    "backend/database/conversations.py": 1643,
     "backend/database/users.py": 1943
   },
   "raise_justifications": {
-    "backend/database/conversations.py": "Public shared-chat uses the existing conversation storage format for bounded, safe transcript decoding; extraction would duplicate crypto/storage invariants."
+    "backend/database/conversations.py": "Public shared-chat uses the existing conversation storage format for bounded, safe transcript decoding; extraction would duplicate crypto/storage invariants. The transactional bulk_assign_segment_speakers helper must sit beside update_conversation_segment_text because it reuses this module's private _prepare_conversation_for_read/_prepare_conversation_for_write encryption boundary and conversations collection handle; extracting it would export those internals."
   },
   "threshold": 1500
 }
--- a/backend/database/conversations.py
+++ b/backend/database/conversations.py
@@ -905,6 +905,69 @@ def update_conversation_segment_text(uid: str, conversation_id: str, segment_id:
     return _update_segment_text(transaction)
 
 
+def bulk_assign_segment_speakers(
+    uid: str,
+    conversation_id: str,
+    segment_ids: List[str],
+    assign_type: str,
+    value: Optional[str],
+    *,
+    firestore_client: Any = None,
+) -> str:
+    """
+    Assign is_user / person_id to multiple segments by id in a single transaction.
+
+    Mirrors update_conversation_segment_text: the read-modify-write runs inside a Firestore
+    transaction and applies each change by segment id. The bulk-assign endpoint previously read the
+    whole transcript_segments array outside any transaction, mutated it in memory, and overwrote the
+    array wholesale via update_conversation_segments — so a concurrent segment edit that committed in
+    between was silently dropped (the lose-update hazard update_conversation_segment_text's docstring
+    describes). Lock enforcement stays with the caller (_get_valid_conversation_by_id), so it is not
+    re-checked here.
+
+    Returns 'ok' on success, 'not_found' if the conversation is missing, or 'segment_not_found' if
+    none of segment_ids were present.
+    """
+    target_ids = set(segment_ids)
+    client = firestore_client if firestore_client is not None else get_firestore_client()
+    doc_ref = client.collection('users').document(uid).collection(conversations_collection).document(conversation_id)
+    transaction = client.transaction()
+
+    @firestore.transactional
+    def _assign(transaction) -> str:
+        doc_snapshot = doc_ref.get(transaction=transaction)
+        if not getattr(doc_snapshot, 'exists', False):
+            return 'not_found'
+
+        conversation_data = _prepare_conversation_for_read(doc_snapshot.to_dict(), uid)
+        if not conversation_data:
+            return 'not_found'
+
+        segments = conversation_data.get('transcript_segments', [])
+        found = False
+        for segment in segments:
+            if not isinstance(segment, dict) or segment.get('id') not in target_ids:
+                continue
+            if assign_type == 'is_user':
+                segment['is_user'] = bool(value) if value is not None else False
+                segment['person_id'] = None
+                found = True
+            elif assign_type == 'person_id':
+                segment['is_user'] = False
+                segment['person_id'] = value
+                found = True
+
+        if not found:
+            return 'segment_not_found'
+
+        doc_level = conversation_data.get('data_protection_level', 'standard')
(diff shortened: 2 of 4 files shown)


kodjima33 (approved):

Git-on-my-level (dismissed):
Thanks for the careful write-up and regression coverage here. The transactional-by-segment-id direction looks useful for avoiding lost transcript updates.

I’m requesting changes before maintainer merge because one repo contract is failing on this head:

- `PR Metadata Preflight` fails the product file line-count ratchet: `backend/database/conversations.py` grows from the frozen baseline of 1580 to 1643 lines. CI’s remediation is to either move/split the new persistence helper out of the oversized product file, or intentionally raise the exact baseline with a one-line justification.
- `Backend unit suite` is also red. The failing short-term lifecycle tests look unrelated to this patch from the CI log, so I’m not attributing that to your change, but the required check still needs to be green/rerun before merge.

Given this mutates user transcript persistence, I’d also like a human maintainer to sanity-check the lock/data-handling boundary before merge. No formal automation approval from me while checks are failing.

---

ZachL111:
Fixed the ratchet in 10d8d33.

I raised the `backend/database/conversations.py` baseline to the exact new count (1580 to 1643) with a justification rather than extracting the helper. `bulk_assign_segment_speakers` reuses this module's private `_prepare_conversation_for_read` / `_prepare_conversation_for_write` encryption boundary and the conversations collection handle, and it belongs beside its sibling `update_conversation_segment_text`. Extracting it would mean exporting those internals.

Verified locally against the merge-base:
```
python3 .github/scripts/check_product_file_line_count_ratchet.py --changed-files <merge-base diff> --base <merge-base>
OK: no line-count increase across 2 changed product source file(s).
```

On the red backend unit suite, I agree it looks unrelated to this patch. The short-term lifecycle tests do not touch this path; the change only swaps the assign-bulk persistence call and adds the transactional helper. Locally the conversation_lifecycle contract set (test_conversation_lifecycle_contract, test_finalization_decision, test_check_conversation_lifecycle_writes, test_conversation_finalization_jobs, test_recording_sessions, test_listen_finalization_cloud_tasks) plus test_conversation_revision_contract and the new regression file are 137 passed.

Git-on-my-level:
Thanks for the follow-up. I rechecked the current head and the ratchet remediation now looks correct: the backend database baseline is raised to the exact new `backend/database/conversations.py` count with a focused justification, and the diff-scoped ratchet script passes locally.

I also ran the new regression test directly in the backend venv:

```text
.venv/bin/python -m pytest tests/unit/test_conversation_bulk_assign_lost_update.py -q
2 passed
```

The transactional-by-segment-id persistence direction still looks good to me for avoiding lost transcript updates. I’m not dismissing the earlier automation change request yet because required CI is still not green on GitHub (`Backend unit suite` remains failed, and `Desktop Swift Build & Tests` is cancelled), and this touches user transcript persistence/encryption boundaries, so the existing human-maintainer/security-review labels should stay until a maintainer signs off.

---

ZachL111:
Closing for now: this branch conflicts with main, and the segment persistence area it touches is being reworked in #10796, so rebasing it today would just set up a collision. The underlying lost update is still real on main (assign-bulk still writes back the full stale segment array), so the right move is a fresh PR against whatever lands there.

PR #10111, opened 2026-07-20
fix: expose apps_results in MCP SimpleConversation model

### LINKED ISSUE #7478: feat req: make custom summary available over API /MCP and AI chat
Users can generate their own summaries, adding context to the prompt thus getting precise summaries (I did that).

But that super precise summaries aren't available over API MCP and AI chat. We see only the standard summary Omi generates, that doesn't consider my custom summary. 

Users would benefit more if could reach their custom summaries as well on aí chat or API.

Dhruvizzle101:
Custom summaries generated by apps were silently dropped from MCP 
responses because SimpleConversation did not include the apps_results 
field. This adds the field so custom summaries are returned alongside 
the default structured overview.

Fixes #7478

## What changed and why
Added `apps_results: List[AppResult] = []` to `SimpleConversation` in 
`backend/routers/mcp.py` and imported `AppResult` from 
`models.conversation`. The REST API already returns apps_results 
correctly via model_dump() on the full Conversation model. This fix 
brings the MCP endpoints in line with the REST API.

none

## How it was verified
Ran unit tests locally using `uv run pytest tests/unit/test_mcp_apps_results.py -v`. Both tests passed. Could not verify against a live MCP server as that requires Firebase credentials.

## Tests
Added `backend/tests/unit/test_mcp_apps_results.py` with two tests:
- test_simple_conversation_includes_apps_results
- test_simple_conversation_apps_results_defaults_to_empty

## Failure class (fixes)
Failure-Class: none

N/A

N/A

## Scoped cleanups (optional)
N/A

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/mcp.py
+++ b/backend/routers/mcp.py
@@ -24,6 +24,7 @@ from database.vector_db import upsert_memory_vector, delete_memory_vector
 import database.vector_db as vector_db
 from models.memories import MemoryDB, Memory, MemoryCategory
 from models.conversation_enums import CategoryEnum
+from models.conversation import AppResult
 from utils.conversations.render import populate_speaker_names, redact_conversations_for_list
 from utils.apps import update_personas_async
 from utils.llm.memories import identify_category_for_memory
@@ -461,6 +462,7 @@ class SimpleConversation(BaseModel):
     finished_at: Optional[datetime]
     structured: SimpleStructured
     language: Optional[str] = None
+    apps_results: List[AppResult] = []
 
 
 class FullConversation(SimpleConversation):
--- a/backend/tests/unit/test_mcp_apps_results.py
+++ b/backend/tests/unit/test_mcp_apps_results.py
@@ -1,0 +1,42 @@
+"""Tests that SimpleConversation includes apps_results field."""
+import sys
+import os
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from datetime import datetime
+from routers.mcp import SimpleConversation, SimpleStructured
+from models.conversation_enums import CategoryEnum
+from models.conversation import AppResult
+
+
+def test_simple_conversation_includes_apps_results():
+    """SimpleConversation should include apps_results field."""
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Thanks for this, @Dhruvizzle101 — clean, well-scoped fix. Adding `apps_results` to `SimpleConversation` brings the MCP conversation endpoints to parity with the REST API (which already returns app-generated custom summaries via the full `Conversation` model) and directly addresses the feature request in #7478.

I verified the change is functional, not just a declaration: `get_conversations` and `search_conversations` both build `SimpleConversation` via `model_validate(conv)` against raw conversation dicts that include `apps_results`, and `FullConversation` inherits the field for `get_conversation_by_id`. Since `apps_results` is a stored DB field (read/written by `conversations_db`), custom summaries will actually be populated in MCP responses.

Two minor observations for a maintainer / possible follow-up:
- The new unit tests assert the field exists and defaults to empty, which is good, but they don't exercise the actual population path (`model_validate` from a dict containing `apps_results`). A small test covering that would make the fix's intent explicit.
- This surfaces app-generated summaries over MCP. It is the user's own data, accessed via their own MCP API key, and consistent with REST — flagging the data-exposure aspect for maintainer sign-off since it is a product surface.

Marking for maintainer review on the data-exposure/product decision. The code itself looks correct to me.

---

kodjima33 (approved):
Backend/MCP bug fix: expose apps_results in SimpleConversation so app-generated summaries aren't silently dropped from MCP responses (fixes #7478); brings MCP in line with REST, 2 tests, conf 5/5

PR #10114, opened 2026-07-20
feat(sdk): add Omi device Rust protocol crate

undivisible:
## Summary
Adds the Omi device Rust protocol crate under `sdks/rust/omi-device`.

## Validation
- `cargo fmt --check`, `cargo test --locked` (7 tests), and `cargo clippy -- -D warnings`
- CI now covers SDK formatting, tests, and Clippy.
- The README links the maintained wire-protocol reference.

## Failure class
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10114?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for adding a Rust protocol crate for the Omi device SDK. The direction looks useful, but I would keep this in draft / maintainer-review territory before merge.

A few things I’d want resolved or verified:

- Please add repo CI coverage for this crate (`cargo test`, `cargo fmt --check`, and ideally clippy). The PR description already calls this out, and the current repo checks are not exercising this new SDK path.
- `audio_frames` appears to drop a valid frame when the frame ends exactly at the payload boundary: the condition `offset + 1 + size >= payload.len()` breaks on equality, but equality should be a complete in-bounds frame. That needs a regression test with a frame ending exactly at `payload.len()`.
- The README makes fairly specific firmware-contract claims (firmware 3.0.20+, status endianness, ring record size/control endianness). Please link or cite the firmware source/protocol reference those constants were derived from so maintainers can verify the SDK stays aligned with device firmware.

No security/supply-chain issue stood out in this pass, and the crate has no external dependencies, which is good. Because this adds a new public SDK/protocol surface, I’m marking it for maintainer review and test coverage rather than approving while it is still draft.

---

undivisible:
Fixed the boundary-ending audio frame, added regression coverage and SDK CI, and linked the protocol reference in 9bb619cfa1. Format, tests, and Clippy pass.

Git-on-my-level:
## Correction: retain the equality rejection in the frame parser

The prior request to accept a frame ending exactly at the payload boundary was incorrect. The authoritative Dart protocol parser, firmware packing behavior, and existing app regression test require equality to be rejected: when `offset + 1 + size >= payload.len()`, the apparent trailing frame is an overflow artifact and must be dropped.

Please restore the `>=` guard in `audio_frames` and invert the new Rust regression test to assert equality is rejected (ideally using the realistic 440-byte packed-record fixture already represented by the app test). Other inspected protocol mappings remain aligned. Sorry for the incorrect earlier direction; this correction is based on the firmware/client source of truth.

undivisible:
Restored the equality rejection and added the 440-byte packed-record regression in bbaa51c4b3. `cargo fmt --check`, locked tests (7), and strict Clippy pass locally.

Git-on-my-level:
Thanks for the follow-up fixes here. The earlier protocol items now look aligned with the Dart reference: the `>=` boundary rejection is restored, the regression covers the packed 440-byte record case, the SDK workflow exists, and the README links the maintained client protocol reference.

One remaining thing I’d fix before this leaves draft: the crate declares `rust-version = "1.85"`, but the new CI only runs `dtolnay/rust-toolchain@stable`. In a Rust 1.85.1 container, `cargo test --locked` currently fails because this new crate has no `Cargo.lock`:

```text
error: the lock file /work/Cargo.lock needs to be updated but --locked was passed to prevent this
```

`cargo test --offline` on the same Rust 1.85.1 toolchain passes all 7 tests, and the GitHub Rust SDK check passes on current stable, so this looks like a CI/MSRV contract mismatch rather than a protocol bug. Please either make the workflow validate the declared MSRV as well, or adjust the lockfile/`--locked` choice so the documented CI command works on the declared minimum supported Rust version.

No security/supply-chain issue stood out in this pass. Because this is still a draft and introduces a new public SDK/protocol surface plus workflow/dependency-sensitive CI, I’d keep `needs-maintainer-review`, `needs-tests`, and `dependency-review` until a maintainer signs off on the Rust SDK/MSRV policy.

---

kodjima33 (approved):
feat(sdk): Omi device Rust protocol crate — feature, approve-only

skanderkaroui (commented):
This Rust version of the device protocol faithfully matches the existing reference and safely handles bad input, with good test coverage. Safe to merge.

skanderkaroui (approved):
Approving based on code review: verified this is a faithful, panic-safe port of the Omi BLE protocol matching the existing reference, with good test coverage. Additive crate with no in-tree callers yet.

PR #10115, opened 2026-07-20
refactor(core): add shared Rust workspace

undivisible:
## Summary
Introduces the shared Rust workspace under `desktop/` and reuses `omi-desktop-core` across desktop hosts.

## Validation
- Docker context tests, workspace `cargo fmt --check`/`cargo check`, backend contract tests, actionlint, and manifest tests pass.
- CI/deploy paths and cache keys now include the shared workspace.
- The clean Docker build will run in CI; the local Docker daemon is unavailable.

## Failure class
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10115?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10117, opened 2026-07-20
feat(agent-runtime): rx4 JSONL protocol, session, journal, and tool relay

undivisible:
## Summary
Adds the agent-runtime rx4 JSONL protocol, session contract, journal persistence, and tool-relay identity binding on the shared Rust workspace.

## Validation
- `cargo test --locked -p omi-agent-runtime` (26 tests) and `cargo clippy --locked -p omi-agent-runtime -- -D warnings` pass.
- CI now covers the agent-runtime crate and uses the committed workspace lockfile.

## Failure class
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Adds a new agent runtime with durable journal state and tool-claim authorization tied to run identity; changes are localized to desktop CI and the agent path but expand the surface for tool execution and persistence.
> 
> **Overview**
> Introduces **`omi-agent-runtime`** as a new member of the desktop Rust workspace, wired to **rx4** for agent execution. The crate adds **protocol v2 JSONL** encode/decode, **fast/deep execution mode** heuristics, a **subagent supervisor** (spawn, message, cancel, collect), and a **`main`** binary that drives stdin/stdout JSONL against rx4 with managed provider transport.
> 
> **Journal persistence** lands in SQLite (`JournalStore`): conversation surfaces, turn recording/update/terminalization with run/attempt binding, generation clears, orphan repair, legacy Node journal import, and **runtime state** (admitted runs, pending tool claims with revoke/prune). **`tool_relay`** authorizes claims in the journal before emitting `authorized_tool_execution` and validates external results; **`provider_policy`** wraps the managed OpenAI-compatible transport.
> 
> **CI and repo hygiene:** `desktop-checks` runs **`cargo fmt`**, **`clippy -D warnings`**, and **`cargo test --locked --workspace`** on the desktop manifest before the existing Node tool-surface script. Pre-push/change detection treats `desktop/Cargo.*`, `agent-runtime-rust`, shared-rust, and the workflow as **`desktop-agent-runtime`**. **`desktop/Cargo.lock`** is committed and un-ignored (replacing a blanket `*.lock` ignore).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c3ba2e4e83510d9b4222bf90df9dce941fb28b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/desktop/agent-runtime-rust/src/journal.rs
+++ b/desktop/agent-runtime-rust/src/journal.rs
@@ -1,0 +1,1016 @@
+use rusqlite::{params, Connection, OptionalExtension, Transaction};
+use serde_json::{json, Map, Value};
+use sha2::{Digest, Sha256};
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+use std::time::{SystemTime, UNIX_EPOCH};
+use uuid::Uuid;
+
+#[derive(Clone)]
+pub struct Surface {
+    pub owner_id: String,
+    pub surface_kind: String,
+    pub external_ref_kind: String,
+    pub external_ref_id: String,
+}
+
+pub struct ResultPage {
+    pub conversation_id: String,
+    pub turn: Option<Value>,
+    pub turns: Vec<Value>,
+    pub cleared_count: u64,
+    pub high_water_turn_seq: u64,
+    pub generation: u64,
+    pub generation_base_turn_seq: u64,
+}
+
+pub struct JournalStore {
+    connection: Connection,
+}
+
+#[derive(Clone)]
+pub struct RunIdentity {
+    pub run_id: String,
+    pub attempt_id: String,
+}
+
+#[derive(Clone)]
+pub struct ToolClaim {
+    pub invocation_id: String,
+    pub owner_id: String,
+    pub session_id: String,
+    pub run_id: String,
+    pub attempt_id: String,
+    pub profile_generation: u64,
+    pub daemon_boot_epoch: String,
+    pub execution_generation: u64,
+    pub tool_name: String,
+    pub input_hash: String,
+    pub status: String,
+    pub outcome: Option<String>,
+    pub result: Option<String>,
+}
+
+impl JournalStore {
+    pub fn open_default() -> Result<Self, String> {
+        let state_dir = env::var_os("OMI_AGENT_STATE_DIR")
+            .map(PathBuf::from)
+            .or_else(|| {
+                env::var_os("HOME").map(|home| {
+                    PathBuf::from(home)
+                        .join("Library")
+                        .join("Application Support")
+                        .join("Omi")
+                        .join("agent")
+                })
+            })
+            .ok_or_else(|| "cannot determine Omi agent state directory".to_owned())?;
+        fs::create_dir_all(&state_dir).map_err(|error| error.to_string())?;
+        Self::open(state_dir.join("omi-agentd.sqlite3"))
+    }
+
+    pub fn in_memory() -> Result<Self, String> {
+        Self::from_connection(Connection::open_in_memory().map_err(|error| error.to_string())?)
+    }
+
+    pub fn open(path: PathBuf) -> Result<Self, String> {
+        Self::from_connection(Connection::open(path).map_err(|error| error.to_string())?)
+    }
+
+    fn from_connection(connection: Connection) -> Result<Self, String> {
+        connection
+            .execute_batch("PRAGMA foreign_keys = ON; PRAGMA journal_mode = WAL;")
+            .map_err(|error| error.to_string())?;
+        let store = Self { connection };
+        store.migrate()?;
+        Ok(store)
+    }
+
+    fn migrate(&self) -> Result<(), String> {
+        self.connection
+            .execute_batch(
+                "CREATE TABLE IF NOT EXISTS rx4_journal_conversations (
+                 owner_id TEXT NOT NULL,
+                 surface_kind TEXT NOT NULL,
+                 external_ref_kind TEXT NOT NULL,
+                 external_ref_id TEXT NOT NULL,
+                 conversation_id TEXT NOT NULL,
+                 generation INTEGER NOT NULL DEFAULT 1,
+                 generation_base_turn_seq INTEGER NOT NULL DEFAULT 0,
+                 high_water_turn_seq INTEGER NOT NULL DEFAULT 0,
+                 PRIMARY KEY (owner_id, surface_kind, external_ref_kind, external_ref_id),
+                 UNIQUE (conversation_id)
+             );
+             CREATE TABLE IF NOT EXISTS rx4_journal_turns (
+                 conversation_id TEXT NOT NULL,
+                 turn_id TEXT NOT NULL,
+                 turn_seq INTEGER NOT NULL,
+                 producer_id TEXT NOT NULL,
+                 payload_hash TEXT NOT NULL,
+                 role TEXT NOT NULL,
+                 surface_kind TEXT NOT NULL,
+                 external_ref_kind TEXT NOT NULL,
+                 external_ref_id TEXT NOT NULL,
+                 content TEXT NOT NULL,
+                 origin TEXT NOT NULL,
+                 status TEXT NOT NULL,
+                 content_blocks_json TEXT NOT NULL,
+                 resources_json TEXT NOT NULL,
+                 producing_run_id TEXT,
+                 producing_attempt_id TEXT,
+                 metadata_json TEXT NOT NULL,
+                 created_at_ms INTEGER NOT NULL,
+                 updated_at_ms INTEGER NOT NULL,
+                 completed_at_ms INTEGER,
+                 PRIMARY KEY (conversation_id, turn_id),
+                 UNIQUE (conversation_id, turn_seq),
+                 UNIQUE (conversation_id, producer_id)
+             );
+             CREATE INDEX IF NOT EXISTS rx4_journal_turns_seq_idx
+                 ON rx4_journal_turns(conversation_id, turn_seq);
+             CREATE TABLE IF NOT EXISTS rx4_runtime_runs (
+                 run_id TEXT PRIMARY KEY,
+                 attempt_id TEXT NOT NULL UNIQUE,
+                 owner_id TEXT NOT NULL,
+                 session_id TEXT NOT NULL,
+                 profile_generation INTEGER NOT NULL,
+                 created_at_ms INTEGER NOT NULL
+             );
+             CREATE TABLE IF NOT EXISTS rx4_tool_claims (
+                 invocation_id TEXT PRIMARY KEY,
+                 owner_id TEXT NOT NULL,
+                 session_id TEXT NOT NULL,
+                 run_id TEXT NOT NULL,
+                 attempt_id TEXT NOT NULL,
+                 profile_generation INTEGER NOT NULL,
+                 daemon_boot_epoch TEXT NOT NULL,
+                 execution_generation INTEGER NOT NULL,
+                 tool_name TEXT NOT NULL,
+                 input_hash TEXT NOT NULL,
+                 status TEXT NOT NULL,
+                 outcome TEXT,
+                 result TEXT,
+                 created_at_ms INTEGER NOT NULL,
+                 updated_at_ms INTEGER NOT NULL
+             );
+             CREATE INDEX IF NOT EXISTS rx4_tool_claims_owner_run_idx
+                 ON rx4_tool_claims(owner_id, run_id, status);",
+            )
+            .map_err(|error| error.to_string())?;
+        self.migrate_node_journal()?;
+        Ok(())
+    }
+
+    pub fn admit_run(
+        &mut self,
+        owner_id: &str,
+        session_id: &str,
+        profile_generation: u64,
+    ) -> Result<RunIdentity, String> {
+        let identity = RunIdentity {
+            run_id: Uuid::new_v4().to_string(),
+            attempt_id: Uuid::new_v4().to_string(),
+        };
+        self.connection.execute("INSERT INTO rx4_runtime_runs(run_id, attempt_id, owner_id, session_id, profile_generation, created_at_ms) VALUES (?, ?, ?, ?, ?, ?)", params![identity.run_id, identity.attempt_id, owner_id, session_id, profile_generation, now_ms()]).map_err(|error| error.to_string())?;
+        Ok(identity)
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn authorize_tool_claim(
+        &mut self,
+        invocation_id: &str,
+        owner_id: &str,
+        session_id: &str,
+        run_id: &str,
+        attempt_id: &str,
+        profile_generation: u64,
+        daemon_boot_epoch: &str,
+        execution_generation: u64,
+        tool_name: &str,
+        input_hash: &str,
+    ) -> Result<(), String> {
+        let existing: Option<String> = self
+            .connection
+            .query_row(
+                "SELECT status FROM rx4_tool_claims WHERE invocation_id = ?",
+                params![invocation_id],
+                |row| row.get(0),
+            )
+            .optional()
+            .map_err(|error| error.to_string())?;
+        if existing.is_some() {
+            return Err("authorized tool invocation is already known".into());
+        }
+        let created = now_ms();
+        self.connection
+            .execute(
+                "INSERT INTO rx4_tool_claims(
+                    invocation_id, owner_id, session_id, run_id, attempt_id,
+                    profile_generation, daemon_boot_epoch, execution_generation,
+                    tool_name, input_hash, status, created_at_ms, updated_at_ms
+                 ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, 'pending', ?, ?)",
+                params![
+                    invocation_id,
+                    owner_id,
+                    session_id,
+                    run_id,
+                    attempt_id,
+                    profile_generation,
+                    daemon_boot_epoch,
+                    execution_generation,
+                    tool_name,
+                    input_hash,
+                    created,
+                    created
+                ],
+            )
+            .map_err(|error| error.to_string())?;
+        Ok(())
+    }
+
+    pub fn pending_tool_claim(&self, invocation_id: &str) -> Result<Option<ToolClaim>, String> {
+        self.tool_claim_with_status(invocation_id, "pending")
+    }
+
+    pub fn completed_tool_claim(&self, invocation_id: &str) -> Result<Option<Value>, String> {
+        let claim = self.tool_claim_with_status(invocation_id, "completed")?;
+        Ok(claim.map(|claim| {
+            json!({
+                "invocationId": claim.invocation_id,
+                "outcome": claim.outcome.unwrap_or_default(),
+                "result": claim.result.unwrap_or_default()
+            })
+        }))
+    }
+
+    pub fn complete_tool_claim(
+        &mut self,
+        invocation_id: &str,
+        outcome: &str,
+        result: &str,
+    ) -> Result<(), String> {
+        let updated = self
+            .connection
+            .execute(
+                "UPDATE rx4_tool_claims
+                 SET status = 'completed', outcome = ?, result = ?, updated_at_ms = ?
+                 WHERE invocation_id = ? AND status = 'pending'",
+                params![outcome, result, now_ms(), invocation_id],
+            )
+            .map_err(|error| error.to_string())?;
+        if updated == 0 {
+            return Err("authorized tool result is unknown".into());
+        }
+        Ok(())
+    }
+
+    pub fn revoke_tool_claims(
+        &mut self,
+        owner_id: &str,
+        run_id: Option<&str>,
+    ) -> Result<u64, String> {
+        let updated = match run_id {
+            Some(run_id) => self
+                .connection
+                .execute(
+                    "UPDATE rx4_tool_claims
+                     SET status = 'revoked', updated_at_ms = ?
+                     WHERE owner_id = ? AND run_id = ? AND status = 'pending'",
+                    params![now_ms(), owner_id, run_id],
+                )
+                .map_err(|error| error.to_string())?,
+            None => self
+                .connection
+                .execute(
+                    "UPDATE rx4_tool_claims
+                     SET status = 'revoked', updated_at_ms = ?
+                     WHERE owner_id = ? AND status = 'pending'",
+                    params![now_ms(), owner_id],
+                )
+                .map_err(|error| error.to_string())?,
+        };
+        Ok(updated as u64)
+    }
+
+    pub fn reconcile_pending_tool_claims(&mut self, owner_id: Option<&str>) -> Result<u64, String> {
+        let updated = match owner_id {
+            Some(owner_id) => self
+                .connection
+                .execute(
+                    "UPDATE rx4_tool_claims
+                     SET status = 'revoked', updated_at_ms = ?
+                     WHERE owner_id = ? AND status = 'pending'",
+                    params![now_ms(), owner_id],
+                )
+                .map_err(|error| error.to_string())?,
+            None => self
+                .connection
+                .execute(
+                    "UPDATE rx4_tool_claims
+                     SET status = 'revoked', updated_at_ms = ?
+                     WHERE status = 'pending'",
+                    params![now_ms()],
+                )
+                .map_err(|error| error.to_string())?,
+        };
+        Ok(updated as u64)
+    }
+
+    fn tool_claim_with_status(
+        &self,
+        invocation_id: &str,
+        status: &str,
+    ) -> Result<Option<ToolClaim>, String> {
+        self.connection
+            .query_row(
+                "SELECT invocation_id, owner_id, session_id, run_id, attempt_id,
+                        profile_generation, daemon_boot_epoch, execution_generation,
+                        tool_name, input_hash, status, outcome, result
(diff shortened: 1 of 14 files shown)


Git-on-my-level:
Thanks for putting this together — this is a cohesive direction around the desktop agent runtime, JSONL protocol/session model, journal persistence, and tool relay. Since the PR is still draft/WIP, I’m not treating this as ready for a final code approval yet.

A few things I’d want resolved before maintainer review for merge:

- CI is currently blocked in Desktop Checks / Rust Backend: `cargo check --locked` fails because `desktop/Cargo.lock` needs to be updated after introducing the new workspace/shared crates. Please regenerate and commit the lockfile so the existing `--locked` job passes.
- This introduces a new runtime surface that relays model/tool execution, stores conversation journal state, handles bearer-token-backed managed provider calls, and adds new Rust dependencies. That needs human maintainer review for the product/security architecture before merge, even if the implementation checks out mechanically.
- The PR description says `desktop/agent-runtime-rust/**` is not yet covered by CI. Before marking ready, please add or wire a CI path for at least `cargo test -p omi-agent-runtime` and `cargo clippy -p omi-agent-runtime -- -D warnings`, or document exactly why that coverage is deferred.

I did not find an obvious malicious/supply-chain payload in the static review, and the tool-relay identity binding plus denylist direction look like useful guardrails. The merge gate right now is readiness/validation plus maintainer review of the new agent-runtime architecture and dependency surface.

---

undivisible:
Committed the workspace lockfile and agent-runtime CI coverage in 7e725fbef2. Locked agent-runtime tests (26) and Clippy pass.

undivisible:
Correction: this branch shares the same current Rust CI failure: six strict-Clippy diagnostics in the desktop backend. The lockfile and agent-runtime checks are pushed, but the Rust job is not green.

undivisible:
Cherry-picked the strict-Clippy fix in d6081904ef. `cargo clippy -p omi-desktop-backend -- -D warnings` and 356 backend tests pass locally; fresh CI is running.

Git-on-my-level:
Thanks for the follow-up here — the current head looks materially improved from my previous automation pass: the workspace lockfile is now committed, the desktop Rust workflow includes the new agent-runtime crate, and the visible Rust/backend checks are green on this head.

I’m still keeping this non-approving while it’s draft and because this is a new agent/runtime + tool-relay + dependency surface that needs maintainer architecture/security sign-off before merge.

One concrete thing I noticed in the current code path: `invalidate_session` currently builds the surface key but does not actually remove the surface mapping, session state, context sources, or any related journal/session state. If the session invalidation message is part of the runtime contract, please either implement the invalidation semantics or make it explicit that this is intentionally a no-op placeholder before marking the PR ready.

No obvious malicious/supply-chain payload stood out in this static review. The remaining gate is product/security architecture review plus tightening that invalidation behavior/contract.

---

desktop/agent-runtime-rust/src/main.rs
```diff
@@ -0,0 +1,1913 @@
...
+    runtime.emit(
+        "init",
+        envelope(
+            None,
+            None,
+            json!({"sessionId": "", "agentControlTools": [], "runtimeVersion": RUNTIME_VERSION, "runtimeCapabilities": ["journal_import_remote_turn", "runtime_adapter_availability"], "runtimeAdapterIds": ["rx4"]}),
+        ),
+    );
+    let stdin = BufReader::new(tokio::io::stdin());
+    let mut lines = stdin.lines();
```
skanderkaroui: `stdin.lines()` (tokio `Lines`) buffers each line with no maximum length. A single newline-less line from a buggy or hostile client grows this buffer unbounded — a local memory-DoS. Consider reading with a byte cap and rejecting/erroring on overlong lines before `parse_line`.

desktop/agent-runtime-rust/src/journal.rs
```diff
@@ -0,0 +1,1016 @@
...
+        &mut self,
+        owner_id: &str,
+        session_id: &str,
+        profile_generation: u64,
+    ) -> Result<RunIdentity, String> {
+        let identity = RunIdentity {
+            run_id: Uuid::new_v4().to_string(),
+            attempt_id: Uuid::new_v4().to_string(),
+        };
+        self.connection.execute("INSERT INTO rx4_runtime_runs(run_id, attempt_id, owner_id, session_id, profile_generation, created_at_ms) VALUES (?, ?, ?, ?, ?, ?)", params![identity.run_id, identity.attempt_id, owner_id, session_id, profile_generation, now_ms()]).map_err(|error| error.to_string())?;
```
skanderkaroui: `rx4_runtime_runs` gets one INSERT per `query` and is never deleted (same for `rx4_tool_claims`, which is only status-flipped to completed/revoked). Both tables grow without bound over a long-lived daemon. Add retention/pruning — e.g. delete completed/revoked claims and runs older than a horizon on boot or periodically.

skanderkaroui (commented):
The security-sensitive part — making sure each tool action is tied to the right session — is done correctly. It still needs work before merge: a couple of ways it could slowly run out of memory, and it's missing tests for the rejection cases; it should also land after #10115, which it builds on.

Git-on-my-level (dismissed):
Thanks for the updates on the lockfile/CI coverage and the persisted tool-claim work. I’m going to keep this as changes-requested rather than another general comment because the current head still has a concrete session-lifecycle blocker in the runtime contract.

Blocking before this is ready for maintainer architecture/security review:

- `invalidate_session` is still effectively a no-op: it looks up the `(ownerId, surfaceKind, externalRefKind, externalRefId)` surface key but does not remove the surface binding, remove the session, clear related context sources, cancel/revoke active runs/tool claims for that session, or emit any invalidation acknowledgement/state. The test currently only asserts that invalidation emits nothing, so it would pass even though stale session/query/tool state remains usable after invalidation.

For an agent runtime that persists journal state and relays authorized tool execution, stale session invalidation is part of the security boundary. Please either implement and test the invalidation semantics end-to-end, or remove/rename the message so clients cannot rely on it as a real revocation/invalidation primitive.

I still don’t see an obvious malicious/supply-chain payload in the static dependency/code review; the new lockfile entries are registry-sourced. But this remains draft, large, dependency/security-sensitive, and it needs human maintainer architecture/security review before merge even after the lifecycle blocker is fixed.

---

Git-on-my-level (dismissed):
Thanks for pushing the follow-up on `invalidate_session` — this is no longer the same no-op I flagged earlier. It now emits an invalidation event, removes the surface/session binding, and cancels in-memory runs for that session.

I’m still keeping this as changes-requested on the current head because there are two concrete readiness/security-contract gaps before this should go to maintainer architecture/security review:

- Stale session IDs can still be used for new queries after invalidation. `query` only checks the configured owner credentials, then falls back to a default profile/surface when `self.sessions.get(&session_id)` returns `None`. That means a client holding the invalidated `sessionId` can submit another `query` and get a new admitted run under that stale ID instead of being rejected as an unavailable/invalidated session. For this runtime contract, invalidation needs to make the old session ID unusable for future model/tool execution.
- The current invalidation path cancels/removes in-memory runs, but it does not revoke any already-authorized pending tool claims for that session/run. `authorized_tool_execution_result` can still complete a pending claim if the old identity fields match. Please revoke pending claims for invalidated sessions/runs and add a regression test that covers the pending-tool-result case.

Also, the validation story still needs tightening: the latest CI shows the `Agent Runtime` job skipped, and the current change selector does not mark `desktop/agent-runtime-rust/**`, `desktop/shared-rust/**`, or `desktop/Cargo.*` as desktop agent-runtime inputs (`scripts/pre_push_ci_prediction.py` reports `has_desktop_agent_runtime=false` for this PR’s changed files). Since the PR description says CI covers the new crate, please either wire the Rust workspace paths into CI or update the PR/validation evidence to make the local-only coverage explicit before marking ready.

Small cleanup: `.gitignore` now contains a stray `0` line next to `!desktop/Cargo.lock`; please remove that typo.

I still do not see an obvious malicious/supply-chain payload in the static review, and the overall direction is cohesive. This remains a dependency/security-sensitive new agent-runtime surface, so it still needs human maintainer architecture/security review before merge after the blockers above are addressed.

---

Git-on-my-level (dismissed):
Thanks for the follow-up. The pending-tool-claim revocation case is now explicitly covered, which is good, but I’m still requesting changes because the current head still leaves merge-blocking runtime/validation issues:

- `query` still admits unknown or invalidated `sessionId` values. After `invalidate_session` removes a session, `query` looks up `self.sessions.get(&session_id)` but then falls back to profile generation `1` and `main_chat` when the session is absent, and still calls `journal.admit_run(...)`. That means a stale session ID can start a new model/tool run instead of being rejected as unavailable/invalidated.
- The new Rust workspace paths still do not wake the Agent Runtime CI selector. `resolve_impact(["desktop/Cargo.lock", "desktop/Cargo.toml", "desktop/agent-runtime-rust/src/main.rs", "desktop/shared-rust/src/lib.rs"])` returns no selected phases and `has_desktop_agent_runtime=false`, while `.github/workflows/desktop-checks.yml` only runs the Agent Runtime job when that output is true. Please wire these Rust paths into CI or make the PR validation explicitly local-only until CI exists.
- `.gitignore` still contains a stray `0` line immediately after `!desktop/Cargo.lock`.

I do not see an obvious malicious dependency/source payload in the static review; the lockfile entries I checked are registry-sourced rather than git/path dependencies. This is still a new dependency/security-sensitive agent-runtime surface, so it needs human maintainer architecture/security review before merge after the concrete blockers above are fixed.

---

Git-on-my-level (dismissed):
Thanks for the latest update. The stale/invalidated `sessionId` blocker from my previous pass looks addressed now: `query` rejects missing sessions before `journal.admit_run(...)`, and there is regression coverage for the invalidated-session path plus pending tool-claim revocation.

I’m still requesting changes because two merge-blocking validation/cleanup issues remain on the current head:

- The new Rust workspace paths still do not wake the Agent Runtime CI selector. I rechecked `resolve_impact(["desktop/Cargo.lock", "desktop/Cargo.toml", "desktop/agent-runtime-rust/src/main.rs", "desktop/shared-rust/src/lib.rs"])`, and it still returns no selected phases / `has_desktop_agent_runtime=false`, while the `Agent Runtime` workflow job only runs when that output is true. Please wire `desktop/Cargo.*`, `desktop/agent-runtime-rust/**`, and `desktop/shared-rust/**` into that selector (or make the PR validation evidence explicitly local-only until CI coverage exists).
- `.gitignore` still has a stray `0` line immediately after `!desktop/Cargo.lock`; please remove that typo.

I don’t see an obvious malicious dependency/source payload in the static review; the lockfile sources I checked are registry-sourced rather than git/path dependencies. This remains a new dependency/security-sensitive agent-runtime surface, so it still needs human maintainer architecture/security review before merge after the concrete blockers above are fixed.

---

Git-on-my-level (dismissed):
Thanks for the latest update. The two blockers from my last pass look addressed on this head: the Rust workspace paths now select the Agent Runtime CI lane, and the stray `.gitignore` `0` line is gone. I also re-ran the local selector test suite (`python3 .github/scripts/test_pre_push_ci_prediction.py`) and the Rust runtime tests in an isolated Rust 1.88 Docker container; both passed.

I’m still requesting changes for two remaining merge blockers:

- Session invalidation still does not revoke pending tool claims once the run has already been removed from `self.running`. `invalidate_session` builds `run_ids` only from active `running` entries, but `complete_run(...)` intentionally authorizes queued tool calls and then removes the run, and the existing `completed_run_authorizes_queued_tool_call_before_removal` test proves those pending claims can complete after the run ends. If the user/session is invalidated in that window, there is no active `RunningQuery` to collect, so `revokedToolClaims` can be 0 and the old `authorized_tool_execution_result` can still complete under an invalidated session. For this runtime, invalidating a session should make all pending tool completions for that session/run unusable, not only completions for runs still present in memory. Please revoke by the durable session/run records (or otherwise persist enough active run identity to revoke completed-but-pending claims) and add a regression test for this exact sequence: authorize queued tool call -> complete/remove run -> invalidate session -> old tool result is rejected.
- The current required Desktop Swift checks are still failing. The failing root in the CI log is `Desktop Swift Static & Test Contracts` / `tests/test-test-install-workflow-contract.sh`: `FAIL: admitted canonical test-install block must occur exactly once`; `Desktop Swift Build & Tests` then fails because the required verify result is failure. Please get the required checks green before maintainer sign-off.

I still do not see an obvious malicious dependency/source payload in the static review; the new dependency entries are registry-sourced. This remains a new dependency/security-sensitive agent-runtime surface, so it still needs human maintainer architecture/security review before merge after the concrete blocker and CI failure are fixed.

---

Git-on-my-level (dismissed):
Thanks for the follow-up. The required Desktop Swift checks are now green on this head, and I re-ran the pre-push selector tests plus the two relevant Rust runtime regression tests in an isolated Rust 1.88 container; those all passed.

I’m still requesting changes because the tool-claim revocation gap from the previous review remains in the current implementation:

- `invalidate_session` still derives `run_ids` only from entries currently present in `self.running`. But `complete_run(...)` authorizes any queued tool calls before it removes that run from `self.running`, and the existing `completed_run_authorizes_queued_tool_call_before_removal` test confirms a pending tool claim can still complete after the run has ended. If the session is invalidated after that removal but before the tool result returns, there is no active `RunningQuery` left for `invalidate_session` to find, so it revokes 0 claims and the old `authorized_tool_execution_result` can still be accepted for the invalidated session.

The new `invalidating_a_session_revokes_its_pending_tool_claims` test covers only the active-run case; it does not cover the completed-run-but-pending-tool-result window above. Please revoke pending claims by durable owner/session/run state (or otherwise persist enough completed-run identity to revoke pending completions for the invalidated session) and add a regression test for: authorize queued tool call -> `complete_run` removes the run -> invalidate the session -> the old tool result is rejected.

I do not see an obvious malicious dependency/source payload in the static review, and CI is now in better shape. This is still a new dependency/security-sensitive agent-runtime/tool-relay surface, so after this concrete blocker is fixed it still needs human maintainer architecture/security sign-off before merge.

---

Git-on-my-level (commented):
Thanks for the update. I re-checked the current head and the specific blocker from the last automation review looks resolved now:

- `invalidate_session` now revokes pending tool claims by durable owner/session state via `revoke_tool_claims_for_session(...)`, rather than only looking at active `self.running` entries.
- The regression coverage now exercises the completed-run-but-pending-tool-result window: queued tool call is authorized during `complete_run`, the run is removed, the session is invalidated, and the old tool result is rejected.
- I verified the pre-push selector suite and the targeted Rust regression tests in an isolated Rust 1.88 container; they passed.

I’m not formally approving because this is still a new dependency/security-sensitive desktop agent-runtime/tool-relay surface and should get human maintainer architecture/security sign-off before merge. But from the automation side, I no longer see the prior tool-claim revocation blocker, and I’m clearing the stale automation change-request reviews that were tied to that resolved blocker.

---

Git-on-my-level (dismissed):
Thanks for the continued work on this runtime surface. I’m requesting changes for the current head because the new Agent Runtime gate is failing in CI:

- `Agent Runtime` fails during `cargo clippy --manifest-path desktop/Cargo.toml --locked --workspace --all-targets -- -D warnings`.
- The current failure is `clippy::cloned-ref-to-slice-refs` in `desktop/agent-runtime-rust/src/journal.rs` at the `repair_marks_orphaned_nonterminal_turns_failed_and_skips_active_runs` test. CI’s suggested fix is to avoid cloning the `run_id` just to build a one-element slice, e.g. use `std::slice::from_ref(&run.run_id)`.

Once that is fixed and the Agent Runtime job is green, this still needs human maintainer sign-off because it introduces a new dependency/security-sensitive desktop agent runtime, journal, and tool-relay authority path. I did not find evidence of a malicious or supply-chain attack in this pass; this is a correctness/CI gate plus maintainer architecture/security review issue.

---

Git-on-my-level (commented):
Thanks for the quick follow-up. The concrete blocker from my previous pass looks resolved on this head: the Agent Runtime job is now green, and the clippy issue in `desktop/agent-runtime-rust/src/journal.rs` was fixed with `std::slice::from_ref(&run.run_id)`.

I also rechecked the pre-push selector suite locally (`python3 .github/scripts/test_pre_push_ci_prediction.py`), and it passes. The local environment here does not have Rust/Cargo installed, so I could not independently rerun the Rust crate tests in this workspace, but GitHub’s required Agent Runtime check is passing on the current head.

I’m not formally approving because this is still a new dependency/security-sensitive desktop agent-runtime, journal persistence, managed-provider, and tool-relay authority surface. It should get human maintainer architecture/security sign-off before merge. I do not see an obvious malicious or supply-chain payload in this pass; the remaining gate is maintainer review of the new runtime/security architecture rather than an automation-found code blocker.

I’m also clearing the stale automation change-request review that was tied to the now-fixed clippy/CI failure.

---

desktop/agent-runtime-rust/src/main.rs
undivisible: Addressed in commit 15613efb0f (fix(agent-runtime): bound runtime admission state) on the current head: stdin frames are now read through read_bounded_jsonl_line with a 1 MiB MAX_JSONL_LINE_BYTES cap (main.rs:1829-1876, wired at main.rs:1910), so a single newline-less line is rejected with an error envelope instead of growing the buffer without bound.

desktop/agent-runtime-rust/src/journal.rs
undivisible: Addressed in commit 15613efb0f on the current head: prune_expired_runtime_state / prune_terminal_runtime_state_before (journal.rs:359-394) delete completed/revoked claims and orphaned runs older than RUNTIME_STATE_RETENTION_MS, invoked at daemon boot (main.rs:1932, 1944) and on state restore (main.rs:1376), with coverage in prunes_only_expired_terminal_runtime_state.

Git-on-my-level (dismissed):
Thanks for the update here. The invalidation path is materially better now: it revokes pending tool claims, cancels active runs for the session, and removes the surface/session mapping.

I’m requesting one focused follow-up before this lands: `invalidate_session` still leaves any `context_sources` entries for the invalidated `sessionId` in memory. Since those payloads can include renderer/workspace/surface context, repeated session invalidation can retain stale context until owner-wide runtime revocation, and the current tests do not cover that cleanup. Please clear `context_sources` for the invalidated session and add a regression test that proves session invalidation removes those cached context entries.

The visible CI is green and I do not see an obvious malicious or supply-chain payload in this static pass. This is still a new desktop agent runtime/journal/tool-relay/dependency surface, so it should keep human maintainer architecture/security sign-off before merge.

---

Git-on-my-level (dismissed):
Thanks for the continued iteration here. I’m requesting changes on the current head for two focused blockers:

- `invalidate_session` still removes the session/surface mapping and revokes tool claims, but it does not clear `context_sources` entries for that `sessionId`. Those cached payloads can include renderer/workspace/surface context and would remain available until owner-wide revocation. Please clear the invalidated session’s context-source cache and add a regression test that seeds a context source, invalidates the session, and proves it is removed.
- The `.gitignore` change replaces the existing `*.lock` ignore rule with `!desktop/Cargo.lock`. That makes other `*.lock` files no longer ignored globally. Please keep the broad lockfile ignore rule and add the `desktop/Cargo.lock` exception in a way that preserves the existing behavior.

The visible CI is green and I do not see an obvious malicious/supply-chain payload in this static pass, but this remains a new desktop agent runtime/journal/tool-relay/dependency surface and should keep human maintainer architecture/security sign-off before merge.

---

undivisible:
Implemented the requested session cleanup in `invalidate_session`, added the regression assertion, and restored the repository-wide `*.lock` ignore while retaining `desktop/Cargo.lock`. Focused Rust tests pass (65 tests plus doc tests) on head `26b5ebb1ae`; GitHub checks are running.

Git-on-my-level (commented):
Thanks for the follow-up on the latest head. I rechecked the two blockers from my previous automation pass against the current diff:

- `invalidate_session` now clears `context_sources` entries for the invalidated `sessionId`, and the session/profile/context regression now seeds a context source, invalidates the session, and asserts the runtime cache is empty.
- The `.gitignore` change now preserves the broad `*.lock` ignore rule while explicitly keeping `desktop/Cargo.lock` tracked.

The visible GitHub checks are green, including the Agent Runtime job, and I do not see an obvious malicious or supply-chain payload in this static pass. I’m clearing the stale automation change-request reviews tied to those now-resolved blockers.

I’m still not formally approving because this remains a new desktop agent runtime/journal/managed-provider/tool-relay authority surface with dependency and workflow changes. Please keep human maintainer architecture/security sign-off before merge.

---

Git-on-my-level (changes requested):
Thanks for continuing to tighten this runtime work. The latest head resolves the previous session invalidation/context-source cleanup concern, but I found one blocking runtime-contract issue that should be fixed before merge.

`desktop/agent-runtime-rust/src/main.rs` stores the configured execution profile's `model_profile` in `configure_default_execution_profile` and `migrate_session_execution_profile`, and the response shape exposes it back to Swift via `profile_json`. However, `query` does not use the session's stored `session.profile.model_profile` when choosing the rx4 model; it only accepts a per-query `modelProfile` field and otherwise falls back to the prompt heuristic (`omi-fast`/`omi-deep`). That means a session migrated/configured to a specific model profile can still execute with a different model whenever the query frame omits `modelProfile`, so the session/profile contract is not actually honored at execution time. Please route query execution through the session profile (with an explicit per-query override only if that is intended) and add a regression around `migrate_session_execution_profile` + `query` model selection.

Other sampled files looked directionally sound:
- `desktop/agent-runtime-rust/src/journal.rs` now binds runtime-produced turn terminalization and tool claims to admitted run/attempt identity, and the previous invalidated-session cleanup path is covered.
- `desktop/agent-runtime-rust/src/tool_relay.rs` validates owner/session/run/attempt/manifest/input-hash identity before accepting tool results.
- `desktop/agent-runtime-rust/src/safety.rs` adds a first-pass denylist for destructive bash and protected write targets; this is a security-sensitive surface, so maintainer sign-off is still needed even after the blocker above is fixed.
- `.github/workflows/desktop-checks.yml` adds Rust fmt/clippy/test coverage for the desktop workspace, and `scripts/pre_push_ci_prediction.py` / `.github/scripts/test_pre_push_ci_prediction.py` now select the agent-runtime lane for the new Rust/workflow inputs.
- `desktop/Cargo.toml`, `desktop/agent-runtime-rust/Cargo.toml`, `desktop/Cargo.lock`, and `.gitignore` consistently add the new workspace member and keep the workspace lockfile tracked.

Validation: I ran the pre-push selector test locally (`python3 .github/scripts/test_pre_push_ci_prediction.py`) and it passed (15 tests). I could not rerun the Rust crate locally because this review environment does not have `cargo` installed, but the GitHub Agent Runtime check is green on this head.

Leaving this as requested changes for the session model-profile contract, plus human maintainer review for the new agent runtime/tool-relay/security/dependency/workflow surface.

— Automated maintainer review (glm-5.2).

---

undivisible:
Addressed the runtime-contract blocker and merged `main` (330 commits behind).

**Blocking item — query must honor the session profile model:**
- Added `select_query_model()` in `desktop/agent-runtime-rust/src/lib.rs` which resolves the model as: per-query `modelProfile` → session `profile.model_profile` → execution-mode heuristic (`omi-fast`/`omi-deep`).
- `query` in `main.rs` now routes through it, so a session migrated to `omi-deep` executes with `omi-deep` even when the query frame omits `modelProfile` — the explicit per-query field still wins when supplied.
- Added regression coverage: `query_model_prefers_explicit_then_session_profile_then_mode_heuristic` pins all three precedence cases.
- `cargo test --locked` (31 tests) and `cargo clippy --locked -- -D warnings` pass.

Also merged `main` (the branch was 330 commits behind) and restored main-owned Dart formatting that the merge had drifted; net diff is the agent-runtime surface only.

Remaining for human sign-off: the `safety.rs` destructive-bash denylist is a security-sensitive surface per the review note.

undivisible:
Two updates on the current head:

1. **Session profile contract blocker resolved.** `query` now routes model selection through the session's stored `session.profile.model_profile` at `desktop/agent-runtime-rust/src/main.rs` (query handler passes `session.profile.model_profile.as_deref()` into `select_query_model`, with a per-query `modelProfile` as an explicit override only). `select_query_model` in lib.rs falls back query → session profile → prompt-mode heuristic, and the unit tests assert the session-profile fallback (query=None, session=omi-deep → omi-deep). This head no longer executes a migrated session with a different model when the query frame omits `modelProfile`.

2. **Agent Runtime CI failure fixed.** The `cargo fmt --check` gate was failing on unformatted test assertions in `lib.rs`; committed the rustfmt output (9300903cbc). `cargo fmt --check` passes locally now.

Still needs human maintainer sign-off on the security-sensitive surfaces (tool relay owner/session identity, safety denylist, new dependency/workflow) as the review noted.

undivisible:
Re: the Aug 11 CHANGES_REQUESTED (query ignoring the session's stored model_profile). Addressed on current head: `query` now feeds both the per-query `modelProfile` and `session.profile.model_profile` into `select_query_model` (desktop/agent-runtime-rust/src/main.rs), so a session configured/migrated to a specific profile executes with it. Regression covered by the `select_query_model` tests.

Issue #10033, opened 2026-07-19
Bug: offline recordings fail to sync, fail to create conversations, and remain stuck in queue


hugoaap-code:
## Summary
Offline recording is unreliable in multiple ways.

This is not limited to the `transcribe later` flow. Offline-recorded audio also fails to sync properly and, in some cases, never creates conversations at all.

## Problem
Audio recorded while offline may upload, but then remain stuck in the queue with `Uploaded · processing on Omi` for many hours without completing.

In other cases, offline-recorded audio does not properly sync back into the app and does not create conversations at all.

So the failure appears to affect both:
- offline audio recorded for later transcription
- offline audio that should sync normally and create conversations automatically

## Current behavior
- Offline audio recordings are uploaded
- The queue shows `Uploaded · processing on Omi`
- Multiple queued items remain stuck for 12+ hours
- No transcription is produced for those recordings
- Some offline recordings also fail to sync into actual conversations
- Some recordings never result in conversation creation at all

## Expected behavior
- Offline audio should sync reliably once connectivity is restored
- Uploaded offline recordings should be processed within a reasonable amount of time
- Offline recordings should create conversations normally
- Users should not lose recordings, transcripts, or conversation objects because of temporary offline conditions

## Actual behavior
- Uploaded offline recordings remain stuck in the queue indefinitely
- Some offline recordings are never transcribed
- Some offline recordings never sync into conversations at all

## Impact
This is a core reliability failure.

Users can lose:
- transcript output
- synced recordings
- the conversation objects themselves

That makes the offline recording workflow effectively untrustworthy.

## Evidence
A screenshot shows multiple uploaded audio items in the queue, all marked `Uploaded · processing on Omi`, still pending after 12+ hours.

Examples visible in the queue include recordings around:
- 8:45 PM
- 8:30 PM
- 8:15 PM
- 8:07 PM
- 7:52 PM
- 7:43 PM
- 7:28 PM

## Product concern
This raises a basic quality question:

Do you test the app under real offline and reconnect scenarios?

Right now, the experience feels like paying users are functioning as beta testers for a core recording and sync workflow.

## Suggested investigation
- Check queue workers responsible for offline audio transcription
- Check whether uploaded offline items are failing silently after upload
- Check whether the processing pipeline is blocked, delayed, or not retrying failed jobs
- Check why some offline recordings never create conversations at all
- Verify whether this is related to the broader conversation processing/finalization issues already reported
[...]

PR #10118, opened 2026-07-20
fix(sync): time out WALs wedged in the uploaded state (offline recordings stuck forever)

kodjima33:
## What

Adds a client-side staleness timeout to the WAL sync reconciler so an offline recording can't be polled in the `uploaded`/"processing" state indefinitely. Past a 6h cutoff the reconciler abandons the stale job and recovers from the retained local file.

## Why — issue #10033

> Audio recorded while offline may upload, but then remain stuck in the queue with `Uploaded · processing on Omi` for many hours without completing. […] Multiple queued items remain stuck for 12+ hours. No transcription is produced for those recordings.

Root cause: an offline recording uploads (HTTP 202), gets a `jobId`, and `reconcileUploadedWals()` polls its sync-job status until terminal. The `wait` branch just `break`s while the status is non-terminal — with **no timeout**. A job the backend keeps reporting non-terminal (notably one wedged `queued`, which the backend stale guard in `sync_jobs.py` explicitly never fails) is therefore polled forever, and the WAL stays in `WalStatus.uploaded` showing "Uploaded · processing on Omi" — exactly the reported symptom.

`Wal.uploadedAt` was already persisted (`wal.dart:132`) but had **zero readers** — the intended escape hatch was never wired.

## The fix

- New pure, `@visibleForTesting` helper `syncJobUploadIsStale(uploadedAt, nowSecs)` — a WAL uploaded past the 6h cutoff is stale. Legacy WALs with `uploadedAt == 0` return `false` (keep polling as before — no behavior change for them).
- In `reconcileUploadedWals()`'s non-terminal (`wait`) branch, stale members abandon the job and recover from the retained local file — re-upload under a fresh job when the file still exists (dedup-safe, same as the backend already handles on the `notFound` path), else `markCorrupted()`. This mirrors the existing `notFound` recovery path exactly.

## What I verified

- Added regression tests in `app/test/unit/sync_job_terminal_policy_test.dart` for the new decision boundary (wedged-past-cutoff → stale, within-cutoff → keep polling, legacy `uploadedAt==0` → never force-revert).
- `bash test.sh test/unit/sync_job_terminal_policy_test.dart` → **All 9 tests passed** (3 new).
- `dart analyze` on both changed files → clean. (The analyzer ratchet reports one `deprecated_member_use_from_same_package` finding, but I confirmed it pre-exists on a clean `main` checkout and is unrelated to this change.)
- **UNVERIFIED end-to-end:** I did not reproduce a real 6h-wedged backend job in a running app (requires a backend job genuinely stuck `queued` for hours). The fix is covered by the regression test on the decision function plus reconciler wiring that mirrors the already-exercised `notFound` recovery path.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/services/wals/local_wal_sync.dart
+++ b/app/lib/services/wals/local_wal_sync.dart
@@ -26,6 +26,14 @@ import 'package:omi/utils/wal_file_manager.dart';
 const _kBackendBusyErrorHint = 'background worker likely died';
 const _freshSyncCutoffSeconds = 6 * 60 * 60;
 
+/// How long a WAL may sit in [WalStatus.uploaded] under one job before the
+/// reconciler abandons that job and re-uploads. A job the backend keeps
+/// reporting non-terminal — notably one wedged `queued`, which the backend
+/// stale guard never fails — would otherwise be polled forever, leaving the
+/// recording stuck on "Uploaded · processing on Omi" and never becoming a
+/// conversation. See issue #10033.
+const _uploadedStaleTimeoutSeconds = 6 * 60 * 60;
+
 enum SyncJobTerminalPolicy { wait, acknowledge, retry }
 
 /// The shared WAL acknowledgement boundary for async sync jobs.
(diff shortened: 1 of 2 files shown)


kodjima33:

PR #10098, opened 2026-07-20
fix(sync): terminalize cloud_tasks jobs whose dispatch was lost before any attempt

aryanorastar:
# Summary

- A `cloud_tasks` sync job whose named-task dispatch was lost before any worker attempt now goes stale after 30 minutes and is terminalized by the existing owned stale finalizer with a distinct `sync_dispatch_lost` code, instead of sitting `queued` until the 24h `JOB_TTL` expiry while the app shows `Uploaded · processing on Omi` the whole time.
- The client's existing failed-job handling then takes over: WALs revert to `miss` and re-upload from the retained local files on the next sync pass — no app change needed.
- Fixes the stuck-queue symptom of #10033 (server side): the reported 12+-hour `Uploaded · processing on Omi` items match this envelope exactly (stuck < 24h TTL, then eventual 404 recovery).

# Root cause

The upload route creates the Redis job (`status='queued'`, `dispatch_mode='cloud_tasks'`) **before** enqueueing the named Cloud Task (deliberately, so an ambiguous enqueue can never start an unfenced inline twin). When the dispatch is lost — both enqueue attempts fail (the `enqueue_uncertain` path returns 202 and preserves all durable inputs, correctly refusing inline work), or the task is purged — nothing ever transitions the job again:

- `is_sync_job_stale` only considered `status == 'processing'` ("queued jobs are intentionally never stale"), so the poll-route stale finalizer never fired.
- The app reconciler's `notFound` recovery only triggers at job-TTL expiry, 24 hours later.

The stuck job is a zombie the whole window: client polls every pass, server truthfully answers `queued`, nobody owns the exit.

# Failure class

Failure-Class: none

Closest is FC-session-scoped-durable-work (durable work must not inherit dispatcher lifetime), but that class covers work *cancelled* with its session; this is durable work whose *dispatch signal* was lost with no owner for the never-started case. Not stretching the definition.

# What is deliberately preserved

- **#7469 contract:** inline queued jobs are never stale at any age — a saturated pool is not a failure. The new staleness requires `dispatch_mode == 'cloud_tasks'`.
- **Retry-backoff contract:** `mark_job_queued_for_retry` re-queues between Cloud Tasks attempts and relies on queued jobs being poll-safe. Those jobs carry `attempt`/`started_at`; the new staleness requires both unset (never-attempted), so a pending retry is never flipped terminal however long its backoff (GCP default max backoff is 1h — a naive age check would have broken this).
[...]

PR #10119, opened 2026-07-20
fix(sync): never merge synced audio into a soft-deleted conversation

aryanorastar:
# Summary

- Synced offline audio can no longer merge into a **soft-deleted conversation**. `get_closest_conversation_to_timestamps` matched any row in its ±2-minute window — including `deleted: True` tombstones — so audio recorded near a conversation the user later deleted merged into the tombstone and vanished: upload succeeded, the sync job completed, and no visible conversation ever appeared. The auto-sync target-attach path had the same hole through its direct `get_conversation` fetch.
- The closest-match choice is now a pure selector (`select_closest_conversation`) over eligible merge targets with one shared predicate (`eligible_merge_target`); the target-attach path consults the same predicate and falls back to the filtered timestamp lookup. A window containing only tombstones now creates a fresh conversation.
- Discarded rows stay eligible on purpose — the merge path already reprocesses and revives them (`reprocessing (discarded) after segment merge`).

Fixes the never-creates-a-conversation symptom of #10033 (server side); the stuck-queue symptom is #10098.

# Root cause

Merge-target selection had no eligibility policy — only geometry. Any row satisfying the timestamp window qualified, deleted or not. Soft-deleted rows keep their `started_at`/`finished_at`, so a recording made around the time of a later-deleted conversation reliably lands in its window; the merge writes segments into a document no reader will ever show. `updated_memories` then reports an id the app cannot display, which is exactly the user-visible shape reported in #10033: recordings sync "successfully" and nothing appears.

# Failure class

Failure-Class: none

No registered class covers write-path target selection ignoring row lifecycle state. Single-instance fix with the decision extracted to one owner; if a sibling surfaces (e.g. another merge path accepting tombstones), that recurrence is the signal to register a class.

# Product invariants affected

# Validation

- Old-code check (production files stashed, tests kept): the suite cannot pass — the selector/predicate did not exist and the inline loop accepted tombstones; with the fix → **4 passed** (`tests/unit/test_sync_merge_target_selection.py`).
- Coverage: deleted-only window yields `None` (→ fresh conversation upstream); a **closer** deleted row loses to a farther live one; boundary-distance choice; predicate matrix (live/deleted/None/discarded-stays-eligible); labeled **static tripwire** that the target-attach path gates through the shared predicate.
- Adjacent sync suites (file-isolated, as CI executes): `tests/unit/test_sync_ordered_assignment.py` + `tests/unit/test_sync_silent_failure.py` → **38 passed**.
- `black` clean on all changed files.
- `backend/database/conversations.py` frozen-size baseline raised 1580 → 1601 (+21, predicate + selector) with justification. Heads-up: #10060 raises the same baseline (to 1616 on its branch) — whichever merges second reconciles the count; both raises are additive and independent.

Not exercised live: a real device deleting a conversation then syncing overlapping offline audio (no device here). The decision seam is driven directly by the tests; the merge/reprocess machinery around it is unchanged.

# Out of scope, named

- Failed/dead-lettered conversations remain eligible merge targets (they are visible, non-deleted rows; merging may be the right revival path) — flagged for maintainer judgment, trivially addable to the predicate if wanted.
- The app-side WAL/queue behavior of #10033 is untouched here.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10119?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (commented):
Thanks for chasing this — the shape of the fix looks solid to me. Extracting a shared `eligible_merge_target` predicate and using it both for timestamp matching and direct `target_conversation_id` attachment addresses the tombstone-merge failure without changing the intended discarded-row revival path.

I’m not formally approving because this touches conversation deletion / synced-audio merge behavior and the current head still has failing CI (`Backend unit suite`, plus cancelled/skipped repo formatting/hygiene jobs in the rollup), so this should get maintainer review before merge.

Validation I did:
- Reviewed the changed merge-target selection and sync attach path.
- Ran the new focused backend test through `backend/test.sh`; `tests/unit/test_sync_merge_target_selection.py` passed locally (`4 passed`).
- Checked the failing backend CI log: it reports failures in `tests/unit/test_ws_b_short_term_lifecycle.py`, which looks unrelated to this diff, but the red required check still needs a maintainer decision or rerun/fix before merge.

One small follow-up I’d prefer before merge if convenient: replace the source-text tripwire in `test_sync_target_attach_consults_the_shared_predicate` with a behavior-level test around the sync target attach fallback if there is an existing lightweight way to stub `process_segment`. I don’t think that needs to block if the maintainers are comfortable with the current coverage, since the core selector/predicate behavior is directly tested.

---

aryanorastar:
@Git-on-my-level I've replaced the source-text tripwire with a behavior-level test using a mocked `process_segment` as requested, and pushed the Backend unit test timebomb fix to get CI green. Ready for another look!

aryanorastar:
Done — replaced the source-text tripwire with a behavioral test (`test_sync_target_attach_fallback_to_closest`) that drives the real `process_segment` with a deleted `target_conversation_id` and asserts it falls through to the timestamp-based closest match. Allowlisted it for the fast-unit duration guard, since driving the real pipeline amortizes its import into the call phase (same file-isolation cost the existing `TestProcessSegmentReal` entries are grandfathered for).

This head also carries a one-line fix for the unrelated `test_ws_b_short_term_lifecycle.py` timebomb (a hardcoded `NOW` date that had gone stale) that was reddening Backend unit suite. Thanks for the steer on the behavioral coverage.

kodjima33 (approved):
Backend bug fix: exclude soft-deleted/tombstone conversations from sync merge-target selection so offline audio near a deleted conversation no longer vanishes (fixes #10033 server side); shared eligibility predicate + test, CI green, conf 5/5

PR #10121, opened 2026-07-20
test(app): add regression test for toggleApp failed-install branch

undivisible:
# Summary

Adds a regression test for the failed-install branch that PR #10100 fixed. The PR changed `toggleApp` from `Future<void>` → `Future<bool>` and routed the create-template sheet through it. The reviewer [requested](https://github.com/BasedHardware/omi/pull/10100#issuecomment-5021611256) a test covering the failure path so the false-success regression cannot silently return.

# Changes

## `app/lib/providers/app_provider.dart` (+10/-3)

Adds two minimal test seams:

| Seam | Purpose |
|------|---------|
| `enableAppOverride` | Mocks `enableAppServer` in tests for the failed-install branch |
| `disableAppOverride` | Mocks `disableAppServer` for the disable path |

Both fall through to the real functions when null — existing callers unaffected.

## `app/lib/utils/alerts/app_dialog.dart` (+3/-1)

Makes `AppDialog.show()` null-safe: returns early when no navigator is mounted. Prevents a hard crash on the failure path in edge-case production scenarios and in tests without a widget tree.

## `app/test/providers/app_provider_toggle_test.dart` (new, ~85 lines)

Three tests, all pure `test()` (no widget tree needed):

| Test | What it covers |
|------|---------------|
| `returns false when enableAppServer fails (regression guard)` | The exact regression #10100 fixed — `toggleApp` must surface failure to its caller |
| `returns true when enableAppServer succeeds` | Happy path — the override is plumbed correctly and success propagates |
| `does not call enableAppOverride when disabling` | Disable path uses its own override, not the enable one |

# Verification

```
$ flutter test test/providers/app_provider_toggle_test.dart
00:00 +3: All tests passed!
```

Existing provider tests also pass (no regressions):

```
$ flutter test test/widget_test.dart test/providers/add_app_provider_test.dart
00:00 +20: All tests passed!
```

pre-push checks: 10/10 passed.

Failure-Class: none

### DIFF
--- a/app/lib/providers/app_provider.dart
+++ b/app/lib/providers/app_provider.dart
@@ -17,6 +17,9 @@ class AppProvider extends BaseProvider {
   /// Test seam — overrides [enableAppServer] in [toggleApp].
   Future<bool> Function(String appId)? enableAppOverride;
 
+  /// Test seam — overrides [disableAppServer] in [toggleApp].
+  Future<void> Function(String appId)? disableAppOverride;
+
   List<App> apps = [];
   List<App> popularApps = [];
   // v2 grouped apps: [{ category: {id,title}, data: List<App>, pagination: {...} }]
@@ -815,7 +818,7 @@ class AppProvider extends BaseProvider {
           PlatformManager.instance.analytics.appEnabled(appId);
         }
       } else {
-        await disableAppServer(appId);
+        await (disableAppOverride ?? disableAppServer)(appId);
         success = true;
         PlatformManager.instance.analytics.appDisabled(appId);
       }
--- a/app/lib/providers/app_provider.dart
+++ b/app/lib/providers/app_provider.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:flutter/foundation.dart' show visibleForTesting;
 import 'package:omi/utils/platform/platform_manager.dart';
 import 'package:collection/collection.dart';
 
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Solid regression guard for the #10100 fix — the failure-path test is exactly the coverage that was requested, and the three cases (fail / succeed / disable) cleanly pin the `Future<bool>` contract.

Two small notes, neither blocking:

- The seams (`enableAppOverride`/`disableAppOverride`) are clean: nullable, null-fallthrough to the real server calls, so production behavior is unchanged. Since they are test-only, consider annotating them `@visibleForTesting` — it is already idiomatic in sibling providers (`device_provider`, `sync_provider`, …) and signals intent to future readers.
- The `AppDialog.show()` null-safety guard is a nice belt-and-suspenders touch for the failure-path crash in tests without a widget tree.

CI on this head looks green on the material checks (Flutter Tests, Dart Analyze ratchet, Generated Files, Android Compile Smoke, PR Preflight all pass). Left un-approved only because the test seams + dialog guard touch provider/dialog surfaces I want a human maintainer to eyeball before merge — happy to convert to an approval once that is confirmed.

---

undivisible:
Marked both override seams `@visibleForTesting` in 68bd562f69. The focused regression test and analyzer ratchet pass.

Git-on-my-level (approved):
Approved after independent gpt-5.6-terra review of the post-feedback source, rebase onto current main, and passing fresh Flutter CI: 921 tests, Dart Analyze, Generated Files, formatting, Android smoke, and hygiene.

PR #10122, opened 2026-07-20
fix(desktop): route onboarding arrow-key navigation through the mounted view

skanderkaroui:
## Problem

Arrow-key navigation in desktop onboarding dies partway through the flow — reported as "stopped working at the *Let's set 'Ask a question' shortcut* step". Left/right/up/down presses are consumed but the step never changes, and navigation never comes back for the rest of the flow.

## Root cause

`OnboardingView` installs an `NSEvent` local key monitor whose closure captures the view struct once at install time. Navigation mutated `currentStep`/`furthestStep` through that captured copy's `@AppStorage`. That works only while the captured storage boxes are still the mounted view's live ones: when SwiftUI re-creates `OnboardingView` mid-flow (parent identity churn around the shortcut/demo steps), the stale monitor keeps consuming arrow events while its writes land in orphaned storage — the UI freezes on the current step. Two competing writers of the same state (the mounted view and the detached closure copy) with no single authority — `FC-split-mutation-authority`.

## Fix

The monitor now only decides; the mounted view mutates:

- `handleArrowNavigation` reads persisted step state directly from `UserDefaults`, resolves the action via a new pure seam `OnboardingFlow.arrowNavigation(keyCode:step:furthestStep:)`, and posts `.onboardingStepNavigationRequested`.
- The mounted view's `.onReceive` re-validates with `OnboardingFlow.validatedNavigationTarget` (backward always allowed, forward gated by the same `canJump` policy as the progress dots) and applies the step through its live `@AppStorage`.
- The monitor handle is static — re-install replaces instead of stacking across identity churn — teardown waits for actual flow completion, and the handler no-ops once onboarding is complete so a lingering monitor can never swallow arrow keys in the main app.

## Verification

- Reproduced pre-fix on a live named test bundle seeded to the shortcut step (step 10, frontier 15): 12 arrow presses via the new `post_key` automation-bridge action, step pinned at 10 while debug logs showed the monitor "navigating" into the orphaned copy.
- Post-fix on the same harness with a real key window: rights walk 10 → 11 → 12 → 13 (across the floating-bar demo and voice shortcut steps), lefts walk 13 → 12 → 11 → 10 → 9.
- `xcrun swift build -c debug --package-path Desktop` clean; `swift test --filter OnboardingFlowTests`: 22 tests, 0 failures — including new regression coverage `testArrowNavigationDecisions` and `testValidatedNavigationTargetPolicy` for the extracted decision/validation seam.

## Invariants

Path-matched per `scripts/pr-preflight --suggest`: INV-AUTH-1, INV-CHAT-1, INV-INT-1 — no behavior change to any of the three; the diff touches `AppState.swift` only to declare the new `Notification.Name`, and onboarding chat/session flows are untouched.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10122?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/AppState.swift
+++ b/desktop/macos/Desktop/Sources/AppState.swift
@@ -787,6 +787,12 @@ class AppState: ObservableObject {
 
 extension Notification.Name {
   static let resetOnboardingRequested = Notification.Name("resetOnboardingRequested")
+  /// Posted by the onboarding arrow-key monitor with a "targetStep" Int in
+  /// userInfo. The mounted OnboardingView applies it to its live @AppStorage —
+  /// the monitor closure must not mutate its own captured copy (writes there
+  /// never reach UserDefaults or the UI on all macOS versions).
+  static let onboardingStepNavigationRequested = Notification.Name(
+    "onboardingStepNavigationRequested")
   /// Posted when the system wakes from sleep
   static let systemDidWake = Notification.Name("systemDidWake")
   /// Posted when the screen is locked
--- a/desktop/macos/Desktop/Sources/DefaultsKey.swift
+++ b/desktop/macos/Desktop/Sources/DefaultsKey.swift
@@ -38,6 +38,7 @@ enum DefaultsKey: String {
   case aiChatWorkingDirectory = "aiChatWorkingDirectory"
   case hasCompletedOnboarding = "hasCompletedOnboarding"
   case onboardingStep = "onboardingStep"
+  case onboardingFurthestStep = "onboardingFurthestStep"
   case onboardingMemoryImportOwnerUserId = "onboardingMemoryImportOwnerUserID"
   case homeOmiDeviceAccountHistory = "home-omi-device-account-history"
   case chatScreenshotSharingEnabled = "chatScreenshotSharingEnabled"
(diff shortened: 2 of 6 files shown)


Git-on-my-level (commented):
Thanks for the focused fix here. I reviewed the desktop onboarding arrow-key path and this looks like a well-contained direction: the event monitor now makes a navigation decision from persisted state, the mounted `OnboardingView` remains the only place that mutates the live onboarding step, stale monitors are replaced instead of stacked, and the new `OnboardingFlow` seams have regression coverage for the key navigation/validation policy.

I do not see a blocking code issue from this review. The main thing I’d want a human maintainer to make the final call on before merge is that this touches global macOS key-event handling for onboarding and deliberately keeps a static monitor alive across SwiftUI identity churn until completion, so it is worth validating in the desktop app around the shortcut/demo steps and after onboarding completion. GitHub also still shows some repo-level checks as cancelled/skipped even though the Desktop Swift Build & Tests job passed.

Not formally approving from automation because this needs maintainer confirmation on the desktop UX/event-monitor behavior, but this is a positive signal from my review.

---

PR #10126, opened 2026-07-20
refactor(app): trust the server's single_language_mode, drop the client mirror (after #10097)

aryanorastar:
# Summary

Follow-up to #10097, doing exactly what its out-of-scope note and Git-on-my-level's review both pointed at: the client-side mirror of `MODULATE_SUPPORTED_LANGUAGES` is deleted, and the app trusts the `single_language_mode` the server already returns from `PATCH /v1/users/language`.

- `setUserPrimaryLanguage` now returns the server-decided mode (`bool?`, null on any failure) instead of discarding the parsed field; `updateUserPrimaryLanguage` syncs local state from that response.
- No eligibility decision remains client-side — future policy changes (new languages, provider swaps) reach the app with zero client releases and zero drift risk.
- **Stacked on #10097** (its branch is the base of this diff; merge it first — this PR then shows only the one commit).

# Why now

#10097 made the mirror byte-for-byte correct; the review verified that equality manually. That manual verification is exactly the maintenance burden this removes: the wire response is the contract, so the list and its normalization helper go away entirely.

# Failure class

Failure-Class: none

This PR's own commit is a `refactor:` with no triggering defect; the declaration covers the stacked #10097 `fix:` commit in this diff range (same call as #10097's own body — divergent read-side policy copies, no registered class).

# Product invariants affected

# Validation

- The mirror's unit test is replaced by a wire-contract test (`app/test/unit/language_update_response_test.dart`): the response carries `single_language_mode`, both values round-trip, and a response missing the field **fails loudly instead of defaulting** — the property that makes trusting the server safe.
- **Not run locally — no Flutter SDK on this machine** (same stated constraint as #10097/#10100, both since approved/merged); the test is hermetic and the app CI lane compiles and runs it.
- Call-site sweep: `setUserPrimaryLanguage` has exactly one consumer (`updateUserPrimaryLanguage`); `supportsLiveMultilingualMode`/`multiLanguageSupported` have zero remaining references after deletion.

# Out of scope, named

- Offline-set language: with no client list, local `single_language_mode` state updates only on a successful server response — identical to today's behavior for the persisted preference (the server was already the only writer of the durable value; the mirror only fed optimistic local state).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10126?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Clean refactor — trusting the server's `single_language_mode` response and deleting the client-side 19-locale mirror is the right direction. Single source of truth eliminates the drift risk on every future STT policy change, and the wire-contract test (`fails loudly instead of defaulting` when the field is absent) pins exactly the property that makes that trust safe.

Quick notes for the maintainer:

- **Stacked on #10097** (unmerged, `needs-maintainer-review`). This PR's own commit only touches the Dart app layer; the backend route/policy changes in the diff come from #10097's commit on the same branch. Merge #10097 first, then this PR collapses to its single app-side commit.
- **Checks are green.** `gh pr checks` reports 16 passed / 0 failed on `39ae8f1`. The cancelled/skipped entries on the checks tab are stale pre-force-push runs, not failures.
- The `bool?` (null-on-failure) return-type change on `setUserPrimaryLanguage` is a small API-surface change — worth a glance that all call sites handle the null path, though `updateUserPrimaryLanguage` is the only consumer and handles it correctly.

Leaving for human maintainer review given the stack dependency and the product-level multilingual-eligibility policy. No changes requested on this PR's own commit.

---

aryanorastar:
@Git-on-my-level Noted. We'll wait for #10097 to merge first.

aryanorastar:
Thanks. Confirmed — this is stacked on #10097; merge that first and this collapses to the single app-side commit. `updateUserPrimaryLanguage` is the only consumer of the new `bool?` return and handles the null (failure) path, and the wire-contract test pins that the response carries `single_language_mode`.

kodjima33 (approved):
Refactor(app): trust server single_language_mode, drop client mirror (after #10097) — approve only per policy (mobile refactor)

Git-on-my-level:
Update on the stack: **#10097 merged** (2026-07-21), so the "stacked on an unmerged base" concern is resolved — this PR now stands on its own two commits against `main`.

The refactor still looks right:

- Trusting the server's `single_language_mode` from the `PATCH /v1/users/language` response and deleting the 66-entry client mirror is the correct single-source-of-truth direction — future STT policy changes no longer need an app release.
- The `bool?` (null-on-any-failure) return on `setUserPrimaryLanguage` is handled correctly by its sole consumer (`updateUserPrimaryLanguage`), and the new onboarding failure path (snackbar + block navigation) is a genuine UX improvement.
- The replacement wire-contract test is the right shape — especially the "fails loudly instead of defaulting when the field is absent" assertion, which pins exactly the property that makes trusting the server safe.

One minor note (not blocking): the second commit bundles a one-line fix to `backend/tests/unit/test_ws_b_short_term_lifecycle.py` (frozen `datetime(2026, 6, 20)` → `datetime.now(timezone.utc)`), which is unrelated to the language refactor. It's tiny and clearly beneficial, so not worth splitting — just flagging it for review awareness.

Leaving formal approval to a human maintainer per `needs-maintainer-review`, since this shifts ownership of the multi/single-language-mode decision entirely to the server.

---

aryanorastar:
@Git-on-my-level need human response — kodjima33 approved, and #10097 (the base) is now merged, so this stands on its own two commits against main. Ready for a maintainer to merge.

aryanorastar:
@Git-on-my-level need human response — this PR is approved, rebased onto latest `main` today, and all checks are green (mergeable, clean). Ready for a maintainer to merge (regular merge, no squash). Thanks!

PR #10151, opened 2026-07-20
fix(ci): make pre-push single-flight runner signal handling portable (Windows SIGHUP crash)

### LINKED ISSUE #9610: fix(ci): make pre-push preflight portable on Windows

### LINKED ISSUE #9724: Pre-push single-flight runner crashes on Windows without SIGHUP

arhxam:
Fixes #9724.
Fixes #9610.

## Problem

The installed pre-push hook routes through `.github/scripts/preflight_runner.py`. `run_owned()` builds its signal handler map from a hard-coded tuple:

```python
previous_handlers = {
    signum: signal.signal(signum, forward_signal)
    for signum in (signal.SIGINT, signal.SIGTERM, signal.SIGHUP)
}
```

`signal.SIGHUP` is POSIX-only and does not exist on Windows, so merely *constructing that tuple* raises:

```
  at preflight_runner.py:183
```

The wrapper exits 1 and Git rejects the push **before any pre-push check runs**, forcing Windows contributors onto `--no-verify`. `scripts/pre-push` works directly — only the single-flight wrapper fails.

The forwarding callback had the same class of problem: `os.killpg` is POSIX-only, so a signal arriving on Windows would raise `AttributeError` rather than stopping the child.

## Fix

Two module-level seams, so both are testable without sending a real signal to the test process:

- **`forwardable_signals()`** resolves the forward set against the host via `getattr(signal, name, None)`, skipping absent signals instead of exploding. POSIX keeps `SIGINT/SIGTERM/SIGHUP`; Windows gets `SIGINT/SIGTERM`.
- **`signal_child()`** prefers `os.killpg` — the child is started with `start_new_session=True`, so it leads its own process group and the whole tree is reached — and falls back to `child.send_signal()` where process groups are unavailable. It now also swallows `OSError` alongside `ProcessLookupError`, since `killpg` can fail with other OS errors against a dying child.

`run_owned()` becomes:

```python
previous_handlers = {signum: signal.signal(signum, forward_signal) for signum in forwardable_signals()}
```

Handler restoration and lock cleanup in the `finally` block are untouched.

## Acceptance criteria (#9610)

- [x] Register only signal constants that exist on the current platform — `forwardable_signals()`
- [x] Keep `os.killpg` process-group forwarding on POSIX — preferred branch in `signal_child()`
- [x] Fall back to `Popen.send_signal()` when `os.killpg` is unavailable
- [x] Preserve handler restoration and lock cleanup — `finally` block unchanged
- [x] Tests simulate missing `SIGHUP` and missing `os.killpg`
- [x] No dependency changes

The two remaining boxes in #9610 (`preflight_runner.py` executing on Windows Python, and a real `git push` completing there) need a Windows host to sign off — I verified the logic by simulating both missing APIs on POSIX.

## Tests

Added `SignalPortabilityTests` to `.github/scripts/test_pr_preflight.py` (already executed by `repo-checks.yml`), covering:

- signal selection with `SIGHUP` removed — simulating Windows
- `SIGHUP` present on POSIX
- forwarding to the process group when `os.killpg` exists
- `send_signal` fallback when `os.killpg` is absent
- a dead child (`ProcessLookupError`) not raising

## Verification

```
$ python -m unittest test_pr_preflight.SignalPortabilityTests
Ran 5 tests ... OK

# simulating Windows (SIGHUP deleted):
OLD expression      -> AttributeError: module 'signal' has no attribute 'SIGHUP'
forwardable_signals -> ['SIGINT', 'SIGTERM']
```

Full file: 26 tests, with 2 pre-existing failures that are **identical before and after** this change (they require a full clone; my sandbox used a shallow sparse checkout). `black --check --line-length 120` → clean, matching the pinned CI formatter.

---

## Failure class

Failure-Class: none

Windows-only signal API portability in a repo tooling script; no registered class covers host-API availability in the pre-push runner.

### DIFF
--- a/.github/scripts/preflight_runner.py
+++ b/.github/scripts/preflight_runner.py
@@ -17,6 +17,35 @@ from pathlib import Path
 POLL_SECONDS = 0.2
 STATUS_INTERVAL_SECONDS = 5.0
 
+# Signals forwarded to the owned child. SIGHUP is POSIX-only and is simply absent
+# on Windows, so the set is resolved against the host rather than assumed —
+# referencing signal.SIGHUP unconditionally raised AttributeError before any
+# pre-push check could run.
+FORWARDED_SIGNAL_NAMES = ("SIGINT", "SIGTERM", "SIGHUP")
+
+
+def forwardable_signals() -> tuple[int, ...]:
+    """Return the forwardable signals this platform actually defines."""
+    resolved = (getattr(signal, name, None) for name in FORWARDED_SIGNAL_NAMES)
+    return tuple(signum for signum in resolved if signum is not None)
+
+
+def signal_child(child: subprocess.Popen, signum: int) -> None:
+    """Forward a signal to the child, preferring its process group where supported.
+
+    The child is started with ``start_new_session=True``, so on POSIX it leads its
+    own process group and ``os.killpg`` reaches the whole tree. Windows has no
+    ``os.killpg``; fall back to signalling the process directly.
+    """
+    killpg = getattr(os, "killpg", None)
+    try:
+        if killpg is not None:
+            killpg(child.pid, signum)
+        else:
+            child.send_signal(signum)
+    except (ProcessLookupError, OSError):
+        pass
+
 
 def atomic_json(path: Path, value: dict) -> None:
     temporary = path.with_suffix(path.suffix + f".{os.getpid()}.tmp")
@@ -174,14 +203,9 @@ def run_owned(
 
     def forward_signal(signum: int, _frame: object) -> None:
         if child is not None and child.poll() is None:
-            try:
-                os.killpg(child.pid, signum)
-            except ProcessLookupError:
-                pass
+            signal_child(child, signum)
 
-    previous_handlers = {
-        signum: signal.signal(signum, forward_signal) for signum in (signal.SIGINT, signal.SIGTERM, signal.SIGHUP)
-    }
+    previous_handlers = {signum: signal.signal(signum, forward_signal) for signum in forwardable_signals()}
     exit_code = 1
     try:
         print(f"Pre-push single-flight log: {log_path}")
--- a/.github/scripts/test_pr_preflight.py
+++ b/.github/scripts/test_pr_preflight.py
@@ -7,6 +7,7 @@ from contextlib import redirect_stderr
 import io
 import json
 import os
+import signal
 import subprocess
 import sys
 import tempfile
@@ -14,8 +15,9 @@ import time
 import unittest
 import urllib.error
 from pathlib import Path
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
+import preflight_runner
 from pr_metadata import TransientPRMetadataError, load_from_api, load_from_event_file
 from pr_preflight import changed_files, format_failure_class_suggest, resolve_pr_metadata, select_checks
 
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
Thanks for the focused fix — this looks like the right shape for the Windows `SIGHUP` crash in the pre-push single-flight runner.

What I checked:
- `forwardable_signals()` avoids importing/registering signal constants that the host Python does not define, so Windows should no longer fail before the pre-push checks start.
- `signal_child()` keeps POSIX process-group forwarding via `os.killpg` and falls back to `Popen.send_signal()` when process groups are unavailable.
- The new unit coverage exercises missing `SIGHUP`, missing `os.killpg`, POSIX process-group forwarding, and dead-child handling without sending real signals to the test process.
- I ran `python -m unittest test_pr_preflight.py` from `.github/scripts` in a clean environment on this head; all 26 tests passed. The focused `SignalPortabilityTests` also passed separately.

I’m leaving this as a positive maintainer signal rather than a formal approval because it changes repo hook/CI tooling under `.github/scripts`, so a human maintainer should make the final call before merge. I did not find a blocker in the code itself.

---

tianmind-studio:
Native Windows follow-up: the signal-selection helper in this PR fixes the first `SIGHUP` exception, but a real hook-driven push still cannot complete on current Windows. After applying this layer I reproduced three subsequent host boundaries:

- `process_exists()` uses `os.kill(pid, 0)`; signal 0 is `CTRL_C_EVENT` on Windows and interrupted the shared console.
- The extensionless `scripts/pre-push` cannot be passed directly to native `CreateProcess`.
- A bare `bash` launched from native Python can resolve to `C:\Windows\System32\bash.exe` instead of the Git Bash already running the hook.

The direct-child `send_signal()` fallback also cannot contain descendants.

The signal seam itself is useful, but the current `Fixes #9610/#9724` scope overstates end-to-end Windows readiness. #10574 carries the host-tested full chain (supported signals, non-destructive PID identity, Git Bash/Python binding, Job Object descendant cleanup, UTF-8 paths, and a native Windows workflow), with all CI green and a positive collaborator review.

I recommend either narrowing this PR to the helper-only layer or coordinating it with that complete integration so maintainers do not merge two conflicting partial implementations.

.github/scripts/preflight_runner.py
```diff
@@ -17,6 +17,35 @@
...
+    The child is started with ``start_new_session=True``, so on POSIX it leads its
+    own process group and ``os.killpg`` reaches the whole tree. Windows has no
+    ``os.killpg``; fall back to signalling the process directly.
+    """
+    killpg = getattr(os, "killpg", None)
+    try:
+        if killpg is not None:
+            killpg(child.pid, signum)
+        else:
+            child.send_signal(signum)
```
tianmind-studio, line 45: `forwardable_signals()` registers `SIGINT` on Windows, but Windows `Popen.send_signal()` does not accept `signal.SIGINT` (`2`); it accepts `SIGTERM` or the Windows console-control events. On a native Windows Python 3.12 checkout I reproduced this with a real `start_new_session=True` child: `signal_child(child, signal.SIGINT)` raises `ValueError: Unsupported signal: 2`. `ValueError` escapes this handler, and `run_owned()`'s `finally` does not stop the still-running child. Please map interruption to a supported Windows mechanism (and create the needed process group if using a console-control event), or use a deliberate termination fallback, then cover the real Windows signal value rather than only mocked `SIGTERM`.

.github/scripts/test_pr_preflight.py
```diff
@@ -494,5 +496,59 @@ def test_different_input_is_rejected_while_active(self) -> None:
...
+                delattr(os, "killpg")  # simulate Windows
+            preflight_runner.signal_child(child, signal.SIGTERM)
+        finally:
+            if had_killpg:
+                os.killpg = original
+        child.send_signal.assert_called_once_with(signal.SIGTERM)
+
+    def test_forwarding_swallows_dead_child(self) -> None:
+        child = Mock(pid=4321)
+        with patch.object(os, "killpg", side_effect=ProcessLookupError):
```
tianmind-studio, line 549: This new portability suite does not pass on the platform it targets. Native Windows has no `os.killpg`, so `patch.object(os, "killpg", ...)` raises `AttributeError` before `signal_child()` runs. Repro from `.github/scripts`: `py -3 -m unittest test_pr_preflight.SignalPortabilityTests` -> 5 tests, 1 error, 2 skipped. Please create/inject the missing callable explicitly (for example `create=True`) or exercise the platform-neutral seam without assuming the POSIX attribute exists.

tianmind-studio (changes requested):
Native Windows review on CPython 3.12 found two reproducible blockers in the portability path. A real child passed through `signal_child(..., signal.SIGINT)` raises `ValueError: Unsupported signal: 2`, and the focused test class currently reports `FAILED (errors=1, skipped=2)` on Windows. I included the exact failure sites and focused fixes inline. The `SIGHUP` selection change itself is sound.

Git-on-my-level (changes requested):
Thanks for continuing to narrow this. The `SIGHUP` registration seam is useful, but I think the current head still needs changes before this can be treated as a Windows pre-push fix.

Blocking points I see:

- In the Windows fallback path, `signal_child(child, signal.SIGINT)` reaches `child.send_signal(signal.SIGINT)` when `os.killpg` is unavailable. Native Windows `Popen.send_signal()` does not provide POSIX `SIGINT` semantics for a child process the way the test mock assumes, so an interrupted hook can still fail instead of cleanly stopping/forwarding. The current test only checks a mocked `send_signal(signal.SIGTERM)` call, so it does not cover the failing `SIGINT` path or real Windows `Popen` behavior. Please either avoid forwarding unsupported signals in the direct-child fallback, map to a Windows-supported console-control/termination behavior with the right process creation constraints, or otherwise make interruption safe on Windows, with coverage that exercises that path.
- The PR still presents this as fixing the end-to-end Windows pre-push hook, but the runner has other native-Windows boundaries that this patch does not address. For example, `process_exists()` still uses `os.kill(pid, 0)`, and the actual hook command path still needs to be valid when launched by native Windows Python. If this PR is intended to be a small first layer, please narrow the claims/acceptance wording so maintainers do not merge it as the complete `#9610`/`#9724` fix.

I ran the focused `SignalPortabilityTests` and the full `.github/scripts/test_pr_preflight.py` suite on Linux at `de1eda489f882b2b2a1bce0c912e5db437b32888`; both pass locally (`5` focused tests, `26` full tests). So the POSIX side looks okay, but the Windows-specific behavior the PR is meant to fix is still under-covered and appears incomplete.

---

kodjima33 (approved):
Verified against #9724/#9610: preflight_runner.py still hard-codes signal.SIGHUP on main, which does not exist on Windows and crashes the pre-push hook; the fix filters to the signals the platform actually has.

PR #10152, opened 2026-07-20
fix(backend): deleting a folder orphaned its conversations (is_default was never True)

arhxam:
## Problem

`DELETE /v1/folders/{id}` without `move_to_folder_id` returns 204 but **leaves every conversation in that folder pointing at a folder that no longer exists** — the conversations vanish from all folder listings and 404 on fetch.

`delete_folder`'s docstring promises: *"If move_to_folder_id is not provided, moves to the default 'Other' folder."* It finds that folder with:

```python
default_folder = next((f for f in folders if f.get('is_default')), None)
```

But **no folder ever has `is_default=True`.** `initialize_system_folders` computes it as:

```python
'is_default': folder_config['category_mapping'] == 'other',
```

and `SYSTEM_FOLDERS` only defines the mappings `work`, `personal`, `social` — **never `other`**. So all three system folders are written `is_default=False`, and user-created folders hard-code `False` (`folders.py:137`). The lookup always returns `None`, `target_folder_id` stays `None`, and the entire relocation block is skipped before `folder_ref.delete()`.

`is_default` has exactly one consumer in this module, so the flag was effectively dead code.

## Fix

1. **Derive the flag from the mapping that already defines the catch-all destination.** `CATEGORY_TO_FOLDER_MAPPING['other'] == 'personal'`, so Personal *is* the "Other" bucket the docstring means:

```diff
-'is_default': folder_config['category_mapping'] == 'other',
+'is_default': folder_config['category_mapping'] == CATEGORY_TO_FOLDER_MAPPING['other'],
```

Deriving it (rather than hard-coding `'personal'`) keeps the two in sync if the mapping ever changes.

2. **Back-compat fallback.** Users provisioned before this fix already have `is_default=False` stored on every folder, so `delete_folder` falls back to the same `'other'` mapping when no folder carries the flag — repairing existing accounts without a migration.

3. **Never target the folder being deleted.** With a live flag, deleting the default folder itself would otherwise move its conversations *into the folder about to be removed*, re-orphaning them.

## Tests

New `tests/unit/test_folder_delete_default_fallback.py` (7 tests). Importantly these drive the **real** `initialize_system_folders` with a mocked Firestore rather than restating its logic, so they regress the `is_default` computation itself:

- exactly one provisioned folder is default, and it is where the `'other'` category lands
- no system folder uses the literal `'other'` mapping (the root cause)
- conversations relocate to the default when no target is given
- legacy folders with the flag unset still relocate (no orphaning)
- the folder being deleted is never chosen as the target
- an explicit `move_to_folder_id` still wins

## Verification

```
$ pytest tests/unit/test_folder_delete_default_fallback.py
7 passed

# revert both parts of the fix:
5 failed, 2 passed
```

The other folder suites still pass (27 tests). `black --line-length 120 --check` → clean.

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10152?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: none

A dead feature flag that made a documented fallback unreachable; no registered class covers it.

### DIFF
--- a/backend/database/folders.py
+++ b/backend/database/folders.py
@@ -193,9 +193,18 @@ def delete_folder(uid: str, folder_id: str, move_to_folder_id: Optional[str] = N
     # Find target folder
     target_folder_id: Optional[str] = move_to_folder_id
     if not target_folder_id:
-        # Find the default folder (usually 'Other')
-        folders = get_folders(uid)
-        default_folder = next((f for f in folders if f.get('is_default')), None)
+        # Find the default folder. Never the folder being deleted — moving conversations
+        # into it would leave them orphaned once it is removed below.
+        candidates = [f for f in get_folders(uid) if str(f.get('id')) != folder_id]
+        default_folder = next((f for f in candidates if f.get('is_default')), None)
+        if not default_folder:
+            # Folders provisioned before is_default was computed correctly have the flag
+            # unset, so fall back to the folder the 'other' category maps to.
+            fallback_mapping = CATEGORY_TO_FOLDER_MAPPING['other']
+            default_folder = next(
+                (f for f in candidates if f.get('category_mapping') == fallback_mapping),
+                None,
+            )
         if default_folder:
             target_folder_id = str(default_folder['id'])
 
@@ -266,7 +275,11 @@ def initialize_system_folders(uid: str) -> List[Dict[str, Any]]:
             'created_at': now,
             'updated_at': now,
             'order': i,
-            'is_default': folder_config['category_mapping'] == 'other',
+            # The folder the catch-all 'other' category lands in is the default
+            # destination. Comparing against the literal 'other' never matched, because
+            # no SYSTEM_FOLDERS entry uses that category_mapping — so no folder was ever
+            # marked default and delete_folder had nowhere to move conversations.
+            'is_default': folder_config['category_mapping'] == CATEGORY_TO_FOLDER_MAPPING['other'],
             'is_system': True,
             'category_mapping': folder_config['category_mapping'],
             'conversation_count': 0,
--- a/backend/tests/unit/test_folder_delete_default_fallback.py
+++ b/backend/tests/unit/test_folder_delete_default_fallback.py
@@ -10,63 +10,41 @@ mappings ``work``/``personal``/``social`` — never ``other``. So every folder w
 with ``is_default=False``, the lookup always returned ``None``, the move block was
 skipped entirely, and the conversations kept a ``folder_id`` pointing at a folder that
 had just been deleted: invisible in every folder listing and 404 on fetch.
+
+Isolation: ``database.folders`` pulls ``database._client`` (a Firestore client) at
+import, so the module is loaded through the sanctioned Tier-2 reserve seam
+(``testing.import_isolation.stub_modules`` + ``load_module_fresh``) inside a fixture —
+never a module-scope ``sys.modules`` mutation. See backend/docs/test_isolation.md.
 """
 
 import os
-import sys
+from pathlib import Path
 from types import ModuleType
 from unittest.mock import MagicMock
 
 import pytest
 
-os.environ.setdefault('ENCRYPTION_SECRET', 'omi_ZwB2ZNqB2HHpMK6wStk7sTpavJiPTFg7gXUHnc4tFABPU6pZ2c2DKgehtfgi4RZv')
+from testing.import_isolation import load_module_fresh, stub_modules
+
+_BACKEND = Path(__file__).resolve().parents[2]
 
 
-class _AutoMockModule(ModuleType):
-    """Module stub that returns a MagicMock for any missing attribute."""
-
-    def __getattr__(self, name):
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
Thanks for the clear bug report and focused fix. I reviewed the backend folder deletion path and the new unit coverage.

This looks like the right direction: deriving the default folder from `CATEGORY_TO_FOLDER_MAPPING['other']`, falling back for legacy users where every stored folder has `is_default=False`, and avoiding selecting the folder currently being deleted all address the orphaned-conversation failure mode without adding a migration. The test file also covers the important regression cases and uses the sanctioned import-isolation seam rather than module-scope `sys.modules` mutation.

I’m not formally approving yet because this touches the backend user-data deletion/relocation path and the PR currently has a failing required check: `Backend unit suite`. The failing CI log points at `tests/unit/test_ws_b_short_term_lifecycle.py`, which does not look related to this diff from static review, but the suite should be green (or re-run if it is a known flake) before merge.

Maintainer note: this is a promising, focused fix for a real data-consistency bug, but because it changes folder deletion behavior for existing users’ conversations, I’d still want human maintainer review before landing.

---

kodjima33 (approved):
Backend bug fix (deleting a folder orphaned its conversations; is_default never True) — approve only; Backend unit suite + PR Metadata Preflight failing, needs green CI

Git-on-my-level (changes requested):
Thanks for the continued work on this focused data-consistency fix. I took another look at the current head and the failing `Backend unit suite` is now blocking for a reason introduced by this PR, not an unrelated flake.

The new code indexes `CATEGORY_TO_FOLDER_MAPPING['other']` in both `delete_folder` and `initialize_system_folders`, but on current `main` that mapping intentionally omits `other` (`backend/database/folders.py` documents it as the catch-all category handled elsewhere). As a result, fresh folder provisioning and the new fallback path can raise `KeyError: 'other'`; CI reproduces this directly in `tests/unit/test_folder_delete_default_fallback.py`.

Please adjust the fix so it does not assume `CATEGORY_TO_FOLDER_MAPPING` contains `other` (or otherwise update the category/default-folder contract consistently with the existing resolver behavior), then rerun the backend unit suite. The underlying bug report still looks valid and worth fixing, but this head cannot merge while it can crash the backend folder path and its own regression tests fail.

Maintainer note: this touches backend user-data relocation/deletion behavior, so even after CI is green it should still get human maintainer sign-off before merge.

---

aryanorastar (commented):
Community review:

This is a real integrity bug and the diagnosis is clear: `is_default` was computed against the literal `'other'` mapping that never exists on `SYSTEM_FOLDERS`, so delete-with-no-target skipped relocation and left conversations pointing at a deleted folder. The fix (compare against `CATEGORY_TO_FOLDER_MAPPING['other']`, exclude the folder being deleted from candidates, legacy `is_default=False` fallback) matches the failure class, and the new unit tests are the right shape.

Before merge I’d still want:
1. Identify/fix the current Backend unit suite failure on the tip and post the focused result.
2. One extra case if missing: delete a non-default folder with `move_to_folder_id=None` when the user has **multiple** non-system folders (not only the provisioned system set), proving relocation still picks the default and not “first remaining.”

Strong community-value data-integrity PR once CI is green.

aryanorastar (changes requested):
Requesting changes (upgrade from earlier COMMENT — agree with @Git-on-my-level):

The orphan-on-delete diagnosis is real and worth landing — but the tip is not mergeable yet.

**Blockers:**
1. **`Backend unit suite` is red** on this tip with a failure introduced here: indexing `CATEGORY_TO_FOLDER_MAPPING['other']` while `main` intentionally omits `other` (catch-all handled elsewhere) → `KeyError: 'other'` in `delete_folder` / `initialize_system_folders` (reproduced in `tests/unit/test_folder_delete_default_fallback.py`).
2. Fix without assuming `'other'` is in that map (or update the contract + tests together).
3. Keep a multi-folder / non-default delete case so conversations rehome to the real default, not orphan.

Please green Backend unit suite, then ping for re-review.

aryanorastar:
Closing as superseded by #11833, merged on August 19. That change fixed the same deleted-folder integrity failure on current `main`: every affected conversation is now repointed unconditionally, with `folder_id=None` as the intentional unfiled state when no default exists, and stale count refreshes no longer turn successful moves into 500s.

This branch remains at `53c9775` and still indexes `CATEGORY_TO_FOLDER_MAPPING['other']`, although current `main` intentionally omits `other`; its own regression suite therefore fails with `KeyError: 'other'`. The original orphaning failure is resolved, so this PR is no longer the correct vehicle.

The remaining docstring wording about a default “Other” folder can be corrected separately if desired.

PR #10153, opened 2026-07-20
fix(desktop): own onboarding arrow monitor lifecycle

Git-on-my-level:
## Summary

- Move the one onboarding arrow-key coordinator into `OnboardingViewOwner`, the stable mounted parent of replaceable `OnboardingView` identities; the child borrows it rather than retaining its own coordinator.
- Replace per-instance cleanup with generation leases: B replaces A's live navigation without a second monitor, stale A disappearance is ignored, and final B teardown removes the token once.
- Narrow responder ownership to text/editable and directional AppKit controls; ordinary buttons retain onboarding arrow navigation. Required-step forwarding posts one real Return key down/up pair.

## Failure class

The defect split the local monitor's lifecycle between replaceable SwiftUI identities. The stable mounted owner now has one coordinator and each child can release only its own generation lease.

## RED → GREEN evidence

- RED on the original PR head `2ba2a9c09d302f838f8924d92af678f8680eebdd`: the new replacement test constructed the then-per-identity owners; B installed a second monitor (`2 != 1`), stale A teardown left B unable to consume the arrow, and B's state did not mutate. This is an actual monitor-handler execution, not a source inspection.
- Exact merged base `eabc74f8aaa2b54014e7f3537081a4962ea57289` has no coordinator seam. An overlay naming the later coordinator would only fail to compile, so it was not represented as behavioral RED. Driving the base's private static monitor/responder policy would require new product instrumentation or a full hosted SwiftUI/AppKit integration harness; neither was added solely to manufacture a base failure.
- GREEN: the replacement test now drives A mount → B mount → stale A unmount ignored → one installed monitor → B consumes and mutates → B unmount removes exactly once. Policy tests instantiate real `NSTextView`, `NSSegmentedControl`, `NSStepper`, and `NSButton`; a real `NSWindow` records the required Return down/up posting.

## Verification

- `xcrun swift test --package-path Desktop --filter 'Omi_ComputerTests.OnboardingFlowTests/testKeyboardNavigation'` — 6 passed after rebuilding the package
- Same focused lifecycle/policy suite repeated 20 times — 20/20 passed
- `xcrun swift test --skip-build --package-path Desktop --filter 'Omi_ComputerTests.OnboardingFlowTests'` — 28 passed
- `./scripts/desktop-core-harness.sh --self-check` — passed
- `git diff --check` — passed
- `make preflight` — passed (existing advisory large-file warnings only)
- Named ad-hoc non-production bundle built, installed, and launched: `/Applications/omi-onboarding-keyboard-lifecycle.app` (`com.omi.omi-onboarding-keyboard-lifecycle`, automation port 47897). Auth, settings, and Rewind seeding were explicitly disabled.

## Limitation

The isolated offline auth-emulator journey remains blocked by a foreign process already using Firestore emulator port 8085. I used only normal worktree setup (`make dev-init`, then isolated-pip installation of the missing `python-dotenv` and `uvicorn` packages) and did not stop the foreign process, read/copy/log credentials, seed auth, or touch `/Applications/omi-dev-auth.app`. The named bundle launched signed out, so the authenticated onboarding UI path was not exercised live; hermetic lifecycle, responder-policy, and default-action coverage is the behavior evidence for this correction.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10153?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/AppState.swift
+++ b/desktop/macos/Desktop/Sources/AppState.swift
@@ -787,12 +787,6 @@ class AppState: ObservableObject {
 
 extension Notification.Name {
   static let resetOnboardingRequested = Notification.Name("resetOnboardingRequested")
-  /// Posted by the onboarding arrow-key monitor with a "targetStep" Int in
-  /// userInfo. The mounted OnboardingView applies it to its live @AppStorage —
-  /// the monitor closure must not mutate its own captured copy (writes there
-  /// never reach UserDefaults or the UI on all macOS versions).
-  static let onboardingStepNavigationRequested = Notification.Name(
-    "onboardingStepNavigationRequested")
   /// Posted when the system wakes from sleep
   static let systemDidWake = Notification.Name("systemDidWake")
   /// Posted when the screen is locked
--- a/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
@@ -108,8 +108,10 @@ struct DesktopHomeView: View {
             appState.hasCompletedOnboarding = true
           }
         } else {
-          OnboardingView(
-            appState: appState, chatProvider: viewModelContainer.chatProvider, onComplete: nil
+          OnboardingViewOwner(
+            appState: appState,
+            chatProvider: viewModelContainer.chatProvider,
+            onComplete: nil
           )
           .onAppear {
             log("DesktopHomeView: Showing OnboardingView (signed in, not onboarded)")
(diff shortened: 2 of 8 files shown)


Git-on-my-level:
Closing as deferred from the Beta release cone.

The implementation and Xcode 16.4 CI are green, but independent review found that coordinator-only tests do not prove the real SwiftUI owner/presentation lifecycle. A minimal NSHostingView proof reached the expected failing boundary, then crashed from the full onboarding tree’s app-level side effects; making it green would require broad test-only suppression/injection across unrelated onboarding setup, permissions, resources, and background work.

That infrastructure expansion is not justified during the stability freeze. No merge occurred; the branch is preserved for a later, separately scoped owner-boundary testability effort if production evidence warrants it.

PR #10154, opened 2026-07-20
fix(backend): mentor notifications reorder the conversation when a segment has start=0

arhxam:
## Problem

The mentor message buffer stamps every segment with:

```python
timestamp = segment.get('start', 0) or current_time
```

`start` is an **offset into the recording**, so the first segment of every conversation is `0.0` — falsy, but a perfectly valid value. `or current_time` therefore throws it away and substitutes `time.time()` (~1.7e9).

Those buffered messages are handed to the mentor LLM via:

```python
sorted_messages = sorted(buffer_data['messages'], key=lambda x: x['timestamp'])
```

So the **opening line of the conversation sorts last**, and the model is shown the transcript with the user's first utterance presented as the final one. The bogus timestamp also defeats the `abs(prev_ts - timestamp) < 2.0` coalescing that merges consecutive same-speaker segments.

## Fix

Only fall back when `start` is genuinely absent:

```diff
-timestamp = segment.get('start', 0) or current_time
+start_offset = segment.get('start')
+timestamp = current_time if start_offset is None else start_offset
```

## Tests

The timestamp path was entirely uncovered — the `_make_segments` helper always uses `start: 1000 + i`, and no test in the file sets `start` to `0`. Added:

- `test_zero_start_keeps_the_opening_line_first` — 12 segments at `start=0.0..11.0`; asserts the opening segment keeps timestamp `0.0`, the list is ordered, and "Segment number 0" is first / "Segment number 11" is last.
- `test_missing_start_still_falls_back_to_current_time` — guards the intended fallback for a segment with no `start` key.

## Verification

Driving the real `process_mentor_notification` with segments at `start=0.0..11.0`:

```
fixed:    first timestamp: 0.0   first text: "Segment number 0..."   last: "Segment number 11..."
original: first timestamp: 1.0   last text:  "Segment number 0..."   <- opening line delivered last
          AssertionError: opening segment lost start offset: 1.0
```

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: none

A falsy-zero default in an in-memory buffer, not a persisted-document read; no registered class covers it.

### DIFF
--- a/backend/tests/unit/test_mentor_notifications.py
+++ b/backend/tests/unit/test_mentor_notifications.py
@@ -370,6 +370,47 @@ def test_process_mentor_notification_not_enough_segments():
     assert result is None
 
 
+def test_zero_start_keeps_the_opening_line_first():
+    """A segment with start=0.0 must keep its own timestamp, not wall-clock time.
+
+    `start` is an offset into the recording, so the first segment of every conversation
+    is 0.0. The buffer used `segment.get('start', 0) or current_time`, and because 0.0 is
+    falsy the opening line was stamped with time.time() (~1.7e9). The messages are handed
+    to the mentor LLM via `sorted(..., key=timestamp)`, so the first thing the user said
+    arrived as the *last* line of the transcript.
+    """
+    message_buffer.buffers.clear()
+
+    segments = [
+        {"text": f"Segment number {i} with some conversation content", "start": float(i), "is_user": i % 2 == 0}
+        for i in range(12)
+    ]
+
+    result = process_mentor_notification("test_uid_zero_start", segments)
+
+    assert result is not None
+    timestamps = [m['timestamp'] for m in result]
+    assert timestamps[0] == 0.0, f"opening segment lost its start offset: {timestamps[0]}"
+    assert timestamps == sorted(timestamps)
+    assert result[0]['text'].startswith("Segment number 0"), result[0]['text']
+    assert result[-1]['text'].startswith("Segment number 11"), result[-1]['text']
+
+
+def test_missing_start_still_falls_back_to_current_time():
+    """A segment with no `start` key keeps the wall-clock fallback."""
+    message_buffer.buffers.clear()
+    before = time.time()
+
+    segments = [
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Nice catch and clean fix. `segment.get('start', 0) or current_time` discarding `0.0` (the offset of the first segment of every conversation) is a genuine ordering bug — the opening line gets stamped with wall-clock time and sorts last in the transcript passed to the mentor LLM, and the 2-second coalescing window is also defeated. Switching to `current_time if start_offset is None else start_offset` is the correct minimal fix and handles all edge cases (0.0, negative offsets, explicit None, int 0) correctly.

The two new tests are well-targeted — they cover both the `start=0.0` path and the missing-key fallback path, and both pass in CI alongside the full existing mentor_notifications suite.

One note on checks: the "Backend unit suite" failure is entirely in `test_ws_b_short_term_lifecycle.py` (5 memory-lifecycle test failures) and is unrelated to this PR. A human maintainer should confirm that failure is pre-existing on main and merge once green.

---

kodjima33 (approved):
Backend bug fix (mentor notifications reorder conversation when a segment has start=0) — approve only; blocked on pre-existing red-main CI (#10169)

Git-on-my-level:
Status update on the current head (`7660c94`): the fix is unchanged and correct, CI is now fully green, and @kodjima33 has approved. The `Backend unit suite` failure I flagged earlier (unrelated `test_ws_b_short_term_lifecycle.py` memory-lifecycle tests) has cleared on this head. With green checks + maintainer approval this looks ready to merge.

---

kodjima33 (approved):
Verified: segment start=0.0 is falsy so 'or current_time' replaced a valid offset with wall-clock, sorting the opening line last; bug still present on main, fix is minimal and covered by two new tests.

PR #10155, opened 2026-07-20
fix(backend): developer memory vector search returns 500 for any limit in 51..100

arhxam:
## Problem

`GET /v1/dev/user/memories/vector/search` advertises:

```python
limit: int = Query(10, ge=1, le=100)
```

and `execute_default_read_vector_search` admits that whole range for this consumer:

```python
bounded_limit = max(1, min(limit, 100 if consumer == MemoryConsumer.developer_api else 20))
```

…but it then calls `fetch_default_vector_memory_search` **without** `max_candidates`, so the service applies its default `DEFAULT_MEMORY_VECTOR_MAX_CANDIDATES = 50` and rejects the request outright:

```python
def _validate_max_candidates(*, max_candidates, bounded_limit):
    if ... or max_candidates < bounded_limit ...:
        raise ValueError(f'max_candidates must be between limit and {MAX_MEMORY_VECTOR_SEARCH_LIMIT}')
```

`_validate_limit` permits up to `MAX_MEMORY_VECTOR_SEARCH_LIMIT = 100`, so nothing clamps the limit first. **Every developer_api vector search with `limit` in 51..100 raises out of the router as an HTTP 500** — on an input the route's own signature declares valid. `limit <= 50` is unaffected.

Reproduced directly against the real validators:

```
DEFAULT_MAX_CANDIDATES = 50   MAX_LIMIT = 100
limit= 10 -> OK
limit= 50 -> OK
limit= 51 -> ValueError: max_candidates must be between limit and 100   <-- HTTP 500
limit= 60 -> ValueError: ...                                            <-- HTTP 500
limit=100 -> ValueError: ...                                            <-- HTTP 500
```

## Fix

Size the candidate budget to the request, so it always satisfies the service's own *"max_candidates between limit and 100"* contract:

```diff
     limit=bounded_limit,
+    max_candidates=max(DEFAULT_MEMORY_VECTOR_MAX_CANDIDATES, bounded_limit),
```

The default 50-candidate budget is unchanged for `limit <= 50`; it only widens when the caller explicitly asked for more, and `bounded_limit` never exceeds `MAX_MEMORY_VECTOR_SEARCH_LIMIT`, so the value always stays in range.

I chose this over clamping `bounded_limit` down to 50 because clamping would silently return fewer rows than the caller requested while the route still advertises `le=100`. Happy to switch to the clamp if you'd rather cap read amplification — it's a one-line change either way.

## Tests

Every existing case in `test_developer_memory_adapter.py` passes `limit=10`, so this range was uncovered. Added `test_developer_vector_adapter_serves_limits_above_the_default_candidate_budget`, driving the real adapter at `limit=60` through the existing Firestore/vector fakes.

## Verification

```
$ pytest tests/unit/test_developer_memory_adapter.py
25 passed

# revert the fix:
1 failed, 24 passed   <- exactly the new test, with the ValueError
```

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

- INV-MEM-1

This change only widens the vector candidate budget so the requested limit can be served; it does not alter the tier set, the canonical collection, or the Short-term + Long-term default access policy.

## Failure class

Failure-Class: none

A caller-supplied limit exceeding an internal candidate budget; the input is well-formed, so FC-malformed-doc-read does not apply.

### DIFF
--- a/backend/tests/unit/test_developer_memory_adapter.py
+++ b/backend/tests/unit/test_developer_memory_adapter.py
@@ -565,6 +565,40 @@ def test_developer_vector_adapter_uses_hydrated_vector_service_and_preserves_ran
     assert all((item['policy']['archive_capability'] is False for item in results))
 
 
+def test_developer_vector_adapter_serves_limits_above_the_default_candidate_budget():
+    """A limit inside the route's advertised window must not blow up the request.
+
+    GET /v1/dev/user/memories/vector/search declares `limit: int = Query(10, ge=1, le=100)`
+    and the developer_api branch of execute_default_read_vector_search admits up to 100.
+    But fetch_default_vector_memory_search defaults max_candidates to
+    DEFAULT_MEMORY_VECTOR_MAX_CANDIDATES (50) and rejects max_candidates < limit, so every
+    limit in 51..100 raised "max_candidates must be between limit and 100" — an HTTP 500
+    on an input the route says is valid.
+    """
+    now = datetime.now(timezone.utc).replace(microsecond=0)
+    items = [_memory_item(f'long-{i}', tier=MemoryTier.long_term, now=now, content=f'coffee {i}') for i in range(3)]
+    db_client = _FirestoreFake({f'users/u1/memory_items/{item.memory_id}': _stored_item(item) for item in items})
+    decision = read_default_read_rollout(
+        uid='u1',
+        db_client=_FirestoreFake({'users/u1/memory_control/state': _enabled_rollout_doc()}),
+        consumer='developer_api',
+    )
+
+    def vector_query(uid, query, *, mode, limit):
+        return _VectorCandidateResult(
+            hits=[_hit(item, score=0.9 - index * 0.01) for index, item in enumerate(items)],
+            rejected_count=0,
+        )
+
+    result = search_memory_default_developer_memories_vector(
+        uid='u1', query='coffee', limit=60, db_client=db_client, rollout_decision=decision, vector_query=vector_query
+    )
+
+    assert result.read_decision == MemoryReadDecision.USE_MEMORY
+    assert result.fallback_reason is None
+    assert [item['id'] for item in result.memories] == ['long-0', 'long-1', 'long-2']
+
+
 def test_developer_vector_adapter_returns_denied_decision_before_vector_or_memory_reads_when_rollout_or_grant_disabled():
     now = datetime.now(timezone.utc).replace(microsecond=0)
     fresh_short_term = _memory_item('fresh-short-term', now=now, content='coffee fresh short term')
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @arhxam — this is a clean, correct fix for a genuine bug.

**Verified the bug.** `fetch_default_read_vector` called `fetch_default_vector_memory_search` without `max_candidates`, so the service applied its default `DEFAULT_MEMORY_VECTOR_MAX_CANDIDATES = 50`. For the `developer_api` consumer, `bounded_limit = max(1, min(limit, 100))` admits up to 100, but `_validate_max_candidates` rejects `max_candidates < bounded_limit` — so every `limit` in 51..100 raised `ValueError` and surfaced as HTTP 500, despite the route advertising `Query(10, ge=1, le=100)`. `limit <= 50` was unaffected. Reproduces exactly as described.

**Fix is correct and minimal.**
`max_candidates=max(DEFAULT_MEMORY_VECTOR_MAX_CANDIDATES, bounded_limit)` keeps the 50-candidate default for the common case and only widens the budget when the caller explicitly asks for more. Since `bounded_limit` never exceeds `MAX_MEMORY_VECTOR_SEARCH_LIMIT` (100), the value always satisfies the service's own "between limit and 100" contract, and the `app`/other consumer branch (capped at 20) is unaffected. The inline comment is accurate and source-backed.

**Regression test is well-targeted.** `test_developer_vector_adapter_serves_limits_above_the_default_candidate_budget` drives `limit=60` through `search_memory_default_developer_memories_vector` with a fake vector query, which fails pre-fix (500) and passes post-fix — good coverage of the exact regression window.

**Checks:** the latest `PR Metadata Preflight` run on this head is green (the earlier failures on this SHA were from before the metadata was updated); Formatting/Hygiene are skipped as there are no relevant changes. No agent-behavior, auth, or privacy-sensitive surfaces touched — this is a read-path correctness fix.

One small note (non-blocking): if a maintainer ever wants to harden against the symmetric trap elsewhere, the `app` branch caps at 20 < 50 so it's safe today, but a short constant/derived cap (e.g. asserting the per-consumer limit never exceeds `MAX_MEMORY_VECTOR_SEARCH_LIMIT` alongside this) would future-proof it. Not needed for this PR.

Recommend merge. Leaving as a positive signal rather than a formal approval since the change touches the developer memory read surface; a maintainer can take it from here.

---

kodjima33 (approved):
Backend bug fix: pass max_candidates so developer_api vector search with limit 51..100 no longer 500s; test added, conf 5/5

PR #10156, opened 2026-07-20
fix(backend): a null capture_confidence crashes the whole What Matters Now evaluation

arhxam:
## Problem

`_build_subjects` coerces stored confidences with a bare `float()`:

```python
capture_confidence = 1.0 if trusted_manual_task else float(task.get('capture_confidence', 0.0))
confidence = float(candidate.get('capture_confidence', 0))
ownership_confidence = float(candidate.get('ownership_confidence', 0))
```

`dict.get(key, default)` returns `None` when the key **is present holding a null** — the default only fires when the key is *absent*. So a single stored null raises:

```
TypeError: float() argument must be a string or a real number, not 'NoneType'
  recommendations.py:357  (tasks)
  recommendations.py:408  (candidates)
```

That propagates straight out of `_build_subjects`, so the user's **entire What Matters Now projection fails** — every task and candidate, not just the one malformed row. A stored string does the same.

A stored **boolean is worse than a crash**: `float(True) == 1.0` silently marks a poison row maximally confident and therefore shortlist-eligible.

## Root cause

The repo already has the correct reader for these exact fields on these exact `action_items` documents — `database/candidates.py::_stored_confidence`:

```python
def _stored_confidence(value: Any) -> float:
    if isinstance(value, (int, float)) and not isinstance(value, bool):
        return float(value)
    return 0.0
```

`database/candidates.py` uses it when reading `capture_confidence`/`ownership_confidence` off a claimed task. So the same field was being read two different ways — one guarded, one not.

## Fix

Add the matching guard in `recommendations.py` and use it at all three sites. Malformed values degrade to `0.0`, which fails the `MINIMUM_CAPTURE_CONFIDENCE` gate — exactly the behaviour the existing *missing-key* tests already assert — while healthy rows continue to evaluate normally.

I defined a local helper rather than importing the private `_stored_confidence` across the `database` → `utils` boundary; happy to promote it to a shared public helper instead if you'd prefer one definition.

## Tests

Existing coverage only exercises well-formed and *missing-key* confidences, never a stored null / bool / string. Added:

- `test_malformed_stored_capture_confidence_does_not_break_the_evaluation` — null, bool and string rows all degrade to `0.0` and stay ineligible, while a healthy `0.95` row is still the sole shortlist entry (one bad document must not sink the projection).
- `test_malformed_candidate_confidence_does_not_break_the_evaluation` — same guard for the candidate reads.

## Verification

Driving the real `_build_subjects`:

```
fixed:    t-null confidence=0.0 eligible=False
          t-bool confidence=0.0 eligible=False
          t-str  confidence=0.0 eligible=False
          t-good confidence=0.95   shortlist: ['t-good']

original: TypeError: float() argument must be a string or a real number, not 'NoneType'
          (recommendations.py:357 tasks / :408 candidates)
```

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10156?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: FC-malformed-doc-read

A stored null in an action_items document turns _build_subjects into a TypeError that fails the whole What Matters Now read.

### DIFF
--- a/backend/tests/unit/test_task_recommendations.py
+++ b/backend/tests/unit/test_task_recommendations.py
@@ -1005,6 +1005,83 @@ def test_accepted_task_capture_confidence_gates_shortlist_eligibility():
     assert [subject.subject_id for subject in shortlist] == ['task-accepted-high']
 
 
+def test_malformed_stored_capture_confidence_does_not_break_the_evaluation():
+    """A null or boolean capture_confidence must degrade to 0.0, not crash or score 1.0.
+
+    `dict.get(key, default)` returns None when the key is present holding null, so the
+    0.0 default never fired and `float(None)` raised TypeError out of _build_subjects —
+    taking down the entire What Matters Now projection, not just the offending row.
+    A stored `True` was worse than a crash: float(True) == 1.0 marked a poison row
+    maximally confident and shortlist-eligible. database.candidates._stored_confidence
+    already guards both cases when reading these same action_items fields.
+    """
+    canonical = EvidenceRef(
+        kind=EvidenceKind.conversation,
+        id='conversation-1',
+        scope=EvidenceScope.canonical,
+    ).model_dump(mode='json')
+    base = {
+        'status': 'active',
+        'due_at': NOW + timedelta(days=1),
+        'updated_at': NOW,
+        'provenance': [canonical],
+    }
+    state = {
+        'tasks': [
+            {**base, 'task_id': 'task-null-confidence', 'description': 'Null confidence', 'capture_confidence': None},
+            {**base, 'task_id': 'task-bool-confidence', 'description': 'Bool confidence', 'capture_confidence': True},
+            {**base, 'task_id': 'task-str-confidence', 'description': 'Str confidence', 'capture_confidence': 'high'},
+            {**base, 'task_id': 'task-good', 'description': 'Send the budget', 'capture_confidence': 0.95},
+        ],
+        'candidates': [],
+        'goals': [],
+        'workstreams': [],
+        'artifacts': [],
+    }
+
+    subjects = recommendations._build_subjects(state, context=None, open_loop_snapshots=[], now=NOW)
+    by_id = {subject.subject_id: subject for subject in subjects}
+
+    assert by_id['task-null-confidence'].facts.capture_confidence == 0.0
+    assert by_id['task-bool-confidence'].facts.capture_confidence == 0.0
+    assert by_id['task-str-confidence'].facts.capture_confidence == 0.0
+    for poisoned in ('task-null-confidence', 'task-bool-confidence', 'task-str-confidence'):
+        assert not by_id[poisoned].eligibility.passes_recommendation_gates
+
+    # The healthy row still evaluates — one bad document must not sink the projection.
+    assert by_id['task-good'].facts.capture_confidence == 0.95
+    shortlist = recommendations.filter_shortlist(subjects, set())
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):

What I checked:
- The change replaces the raw `float(...)` reads for stored task/candidate confidence with a guard that treats `None`, strings, and booleans as `0.0`, which matches the existing persistence-side behavior and avoids both projection crashes and accidental `True -> 1.0` promotion.
- The new regression coverage exercises malformed task confidence without sinking the whole What Matters Now evaluation, plus malformed candidate confidence.
- I ran the targeted backend unit file locally against this head with a stripped environment: `python -m pytest tests/unit/test_task_recommendations.py -q` -> `41 passed`.

I’m not formally approving because this head still has a failing Backend unit suite check on GitHub. The failure appears to be in `tests/unit/test_ws_b_short_term_lifecycle.py`, not in the files touched here, but the required checks should be green before merge. Once CI is green, this looks like a small, low-risk bug fix.

---

kodjima33 (approved):
Backend bug fix (null capture_confidence crashes whole What Matters Now eval) — approve only; blocked on pre-existing red-main CI (#10169)

Git-on-my-level:
Good catch and a clean, minimal fix.

The new `_stored_confidence` helper mirrors the existing guarded reader in `database/candidates.py` exactly, and all three call sites in `_build_subjects` are covered — tasks `capture_confidence`, candidates `capture_confidence`, and candidates `ownership_confidence`. Treating booleans as malformed (not `1.0`) is the right call given the shortlist-eligibility gate.

The two new tests cover null, bool, and string values for both the task and candidate paths, and the healthy-row assertion confirms a single malformed document no longer sinks the whole projection. All 27 tests in `test_task_recommendations.py` pass.

Note: the `Backend unit suite` check is currently failing, but the 7 failures are all in `test_ws_b_short_term_lifecycle.py` (memory lifecycle/promotion) — unrelated to this change and pre-existing on main.

Positive signal — this is exactly the kind of defensive hardening the recommendations pipeline needed.

---

Git-on-my-level (approved):
Approving — all required checks are green on this head now (Backend unit suite, Backend Hermetic Merge Gate, Hermetic Backend E2E all passing), which clears the only thing that held back formal approval earlier.

The fix is correct and minimal: `dict.get(key, default)` returns `None` when the key is present holding null, so the bare `float()` raised `TypeError` out of `_build_subjects` and took down the entire What Matters Now projection. `float(True) == 1.0` was worse — it silently marked poison rows maximally confident and shortlist-eligible. The new `_stored_confidence` helper mirrors the existing guarded reader in `database/candidates.py` exactly, and all three call sites (task `capture_confidence`, candidate `capture_confidence`, candidate `ownership_confidence`) are covered. Booleans are correctly treated as malformed rather than promoted.

Regression coverage is thorough: null/bool/string for both the task and candidate paths, plus a healthy-row assertion confirming a single malformed document no longer sinks the whole projection.

Small, focused, well-tested defensive hardening. Thanks @arhxam.

---

kodjima33 (approved):
Verified: dict.get returns None for a present-but-null capture_confidence so float(None) crashed the whole projection; guard mirrors database.candidates._stored_confidence, covered by a new test.

PR #10157, opened 2026-07-20
fix(backend): focus stats count a zero-second distraction as a full minute

arhxam:
## Problem

`get_focus_stats` charges an unmeasured distraction a one-minute default, using `or`:

```python
total_distracted_seconds += s.get('duration_seconds') or 60
entry['total_seconds']   += s.get('duration_seconds') or 60
```

But `duration_seconds` is `Optional[int]` with **`ge=0`** on both the model and the route:

```python
# models/focus_session.py
duration_seconds: Optional[int] = Field(default=None, ge=0, description='Session duration in seconds, if known.')
# routers/focus_sessions.py
duration_seconds: int | None = Field(None, ge=0, le=86400)
```

and `create_focus_session` stores whatever the client sent. The schema draws a real distinction — **`None` = unknown, `0` = measured and zero** — but `or` collapses both to the default. A genuine zero-second distracted session is reported as **60 seconds of lost focus**.

Ten such sessions invent ten minutes of distraction that never happened, inflating `distracted_minutes` and pushing the offending app up the `top_distractions` ranking. (The `focused` branch never had this problem — it falls back to `0`.)

## Fix

Reserve the default for a genuinely missing duration:

```diff
-distracted_seconds = s.get('duration_seconds') or DEFAULT_DISTRACTED_SECONDS
+raw_duration = s.get('duration_seconds')
+distracted_seconds = DEFAULT_DISTRACTED_SECONDS if raw_duration is None else raw_duration
```

This mirrors the fix already applied to the same bug class in `database.staged_tasks.migrate_ai_tasks`, whose regression test records the identical reasoning: *"The key now maps only a genuinely missing (None) score to 999."*

## Commits

1. **`test:`** characterize the aggregation — `get_focus_stats` had **no unit coverage at all**, so the existing behaviour is pinned first (all green on unmodified code).
2. **`refactor:`** lift the twice-repeated magic `60` to `DEFAULT_DISTRACTED_SECONDS` and compute it once per session. Pure refactor; step 1's tests pass unchanged.
3. **`fix:`** distinguish `None` from `0`, plus the three regression cases.

## Verification

```
$ pytest tests/unit/test_focus_stats_aggregation.py
9 passed

# restore the `or` expression:
3 failed, 6 passed
  - test_explicit_zero_duration_is_not_rounded_up_to_a_minute
  - test_zero_duration_sessions_do_not_inflate_the_daily_total
  - test_zero_and_unknown_durations_are_scored_differently
```

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10157?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/focus_sessions.py
+++ b/backend/database/focus_sessions.py
@@ -15,6 +15,10 @@ from ._client import db
 
 logger = logging.getLogger(__name__)
 
+# A distraction that was recorded but never measured still represents lost focus, so it is
+# charged this much rather than being ignored. Applies only to an unknown duration.
+DEFAULT_DISTRACTED_SECONDS = 60
+
 
 def _user_col(uid: str, collection: str) -> Any:
     """Shorthand for users/{uid}/{collection}."""
@@ -83,10 +87,11 @@ def get_focus_stats(uid: str, date: Optional[str] = None) -> Dict[str, Any]:
             total_focus_seconds += s.get('duration_seconds') or 0
         elif s.get('status') == 'distracted':
             distracted_count += 1
-            total_distracted_seconds += s.get('duration_seconds') or 60
+            distracted_seconds = s.get('duration_seconds') or DEFAULT_DISTRACTED_SECONDS
+            total_distracted_seconds += distracted_seconds
             app = str(s.get('app_or_site', 'Unknown'))
             entry = distractions.setdefault(app, {'total_seconds': 0, 'count': 0})
-            entry['total_seconds'] += s.get('duration_seconds') or 60
+            entry['total_seconds'] += distracted_seconds
             entry['count'] += 1
 
     top = sorted(distractions.items(), key=lambda x: x[1]['total_seconds'], reverse=True)[:5]
--- a/backend/database/focus_sessions.py
+++ b/backend/database/focus_sessions.py
@@ -87,7 +87,11 @@ def get_focus_stats(uid: str, date: Optional[str] = None) -> Dict[str, Any]:
             total_focus_seconds += s.get('duration_seconds') or 0
         elif s.get('status') == 'distracted':
             distracted_count += 1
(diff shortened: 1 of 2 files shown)


arhxam:
Closing this — I was wrong, and CI caught it.

`backend/tests/unit/test_desktop_migration.py` already contains `TestFocusStatsDurationBoundary`, which pins the current behaviour deliberately:

```python
def test_distracted_zero_duration_treated_as_default(self):
    """duration_seconds=0 is treated as 60 via `or 60` in get_focus_stats."""
    ...
    # duration_seconds=0 is falsy, so `or 60` defaults to 60
    assert result['distracted_minutes'] == 1
```

The test name, its docstring and its inline comment all describe the `or 60` collapse as intended, so treating a zero-second distracted session as a full minute is a deliberate product decision, not the falsy-zero defect I claimed. My change flipped it and broke that test.

My mistake was in how I checked for existing coverage: I looked for test files whose *filename* matched `focus`, found none, and concluded the function was untested. The coverage lives in `test_desktop_migration.py`. I should have grepped for the function name (`get_focus_stats`) across all of `tests/unit/`, which would have surfaced it immediately.

Sorry for the noise. If the zero-vs-unknown distinction ever *is* considered worth making, it would need that test updated alongside it and a product decision on the semantics — not a silent behaviour flip, which is what this PR was.

My other open PRs are unaffected by this reasoning; I'm fixing their `Failure-Class:` declarations separately.

PR #10158, opened 2026-07-20
fix(backend): paid apps priced at $19.99 are billed $19.98 (cents truncation)

arhxam:
## Problem

`upsert_app_payment_link` converts the developer's dollar price into the integer cent amount Stripe bills:

```python
payment_price = stripe.create_app_monthly_recurring_price(app.payment_product_id, int(price * 100))
```

Prices arrive as binary floats, so `19.99` is really `19.98999…` and:

```python
>>> 19.99 * 100
1998.9999999999998
>>> int(19.99 * 100)
1998
```

`int()` truncates toward zero, so an app listed at **$19.99 gets a Stripe recurring price of $19.98**.

`App.price` keeps the original `19.99`, so the database and Stripe disagree permanently — and because the price object is created once and reused via `payment_price_id`, **every buyer is undercharged for the life of the app**.

**573 of the 10,000 prices between $0.01 and $100.00 are affected**, including the very common `$19.99` and `$8.29`. `$9.99` and `$29.99` happen to be exactly representable, which is why the existing coverage (a single `5.0` case) never caught it.

## Fix

```diff
-return int(price * 100)
+return int(round(price * 100))
```

## Commits

1. **`test:`** pin the dollars→cents contract with prices that *are* exactly representable (green on unmodified code).
2. **`refactor:`** lift the inlined conversion into `_price_to_cents` so the money conversion has a name and a place to test. Pure refactor.
3. **`fix:`** round instead of truncate, plus the regression cases.

## Tests

- six representative float-error prices (`19.99 → 1999`, `8.29 → 829`, `0.29`, `1.13`, `2.01`, `43.15`)
- an exhaustive sweep asserting **no** price in `$0.01..$100.00` converts to fewer cents than it is worth

## Verification

```
$ pytest tests/unit/test_app_paid_price_guard.py
21 passed

# restore int(price * 100):
6 failed, 15 passed
  - ...[19.99-1999]  ...[8.29-829]  ...[0.29-29]  ...[1.13-113]  ...[2.01-201]
  - test_cents_conversion_never_rounds_down_across_the_common_price_range
```

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10158?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: none

Binary float truncation in a dollars-to-cents conversion; no registered class covers money rounding.

### DIFF
--- a/backend/tests/unit/test_app_paid_price_guard.py
+++ b/backend/tests/unit/test_app_paid_price_guard.py
@@ -87,11 +87,17 @@ def test_binary_float_prices_are_not_underbilled_by_a_cent(stripe_mock, price, e
 
 
 def test_no_price_in_the_common_range_is_underbilled(stripe_mock):
-    """No price in $0.01..$100.00 may be billed fewer cents than it is worth."""
+    """Every non-representable price in $0.01..$100.00 must still bill its full cents.
+
+    Only the prices whose cent value is not exactly representable are exercised — the
+    rest cannot regress — which keeps this inside the fast-unit CPU budget.
+    """
+    truncating = [cents / 100 for cents in range(1, 10001) if int((cents / 100) * 100) != cents]
+    assert truncating, 'expected some non-representable prices in this range'
+
     wrong = []
-    for cents in range(1, 10001):
-        price = cents / 100
+    for price in truncating:
         upsert_app_payment_link('app1', True, price, 'monthly_recurring', 'u1')
-        if stripe_mock.create_app_monthly_recurring_price.call_args[0][1] != cents:
+        if stripe_mock.create_app_monthly_recurring_price.call_args[0][1] != round(price * 100):
             wrong.append(price)
     assert wrong == []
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified the math: `int(19.99 * 100)` is `1998` because `19.99 * 100` is `1998.999...` and `int()` truncates toward zero — so $19.99 apps were billed at $19.98, with `App.price` and the Stripe price permanently disagreeing. The fix (`int(round(price * 100))`) is correct: ~573 of the 10,000 prices in $0.01–$100.00 were silently losing a cent.

The new tests are good — representative float-error cases plus an exhaustive sweep over the non-representable prices in the common range. Backend unit suite passes.

One accuracy note: the PR description mentions a third "refactor" commit that lifts the conversion into a `_price_to_cents` helper, and the "Fix" code block reads as a function body (`return int(round(...))`). Neither matches the actual diff — the change stays inline in `upsert_app_payment_link` and there is no `_price_to_cents`. It'd help to align the description with what landed so the commit narrative is accurate.

This is a billing/money path from an external fork, so flagging for a maintainer to confirm before merge. The fix itself looks right.

---

kodjima33 (approved):
Backend bug fix: int(round(price*100)) fixes cents truncation undercharging $19.99 apps by a cent; test added, conf 5/5

PR #10159, opened 2026-07-20
fix(backend): /v2/apps/search bypasses the review-score clamp

arhxam:
## Problem

`_clamp_review_score` exists because `ReviewAppRequest.score` is unbounded on the write path. Its comment is explicit about scope:

> The read path already bounds score with `Field(ge=0, le=5)`; this closes the same bound on the **write and aggregation paths**, where the request model leaves score unbounded.

Every aggregation site in `utils/apps.py` honours that — lines **306, 382, 425, 516** all average `_clamp_review_score(x['score'])`.

`routers/apps.py:699` computes `rating_avg` itself and was the **one aggregation path that skipped it**:

```python
rating_avg = sum([x['score'] for x in sorted_reviews]) / len(sorted_reviews) if reviews else None
```

`App.rating_avg` carries no bound (`Optional[float] = 0`), so a drifted or abusive stored score surfaces verbatim from search. A single stored review of `100` gives that app `rating_avg: 100.0`, which:

- clears any `rating` filter — `filtered_apps = [app for app in filtered_apps if (app.rating_avg or 0) >= rating]`
- pins it to the top of `sort=rating_desc` — `sorted(..., key=lambda a: (a.rating_avg or 0), reverse=True)`

…while `/v1/apps` reports `5.0` for the very same app.

## Fix

Use the same clamped expression as `utils/apps.py:425`, which this line is otherwise identical to:

```diff
-rating_avg = sum([x['score'] for x in sorted_reviews]) / len(sorted_reviews) if reviews else None
+rating_avg = (
+    sum([_clamp_review_score(x['score']) for x in sorted_reviews]) / len(sorted_reviews) if reviews else None
+)
```

## Commits

1. **`test:`** cover the search endpoint's rating aggregation for in-range scores (green on unmodified code) — the clamp suite previously never touched this endpoint.
2. **`fix:`** apply the clamp, with regression cases for an out-of-range and a negative stored score, plus an assertion that search now agrees with the `utils.apps` aggregation for identical reviews.
3. **`test:`** an invariant scan requiring *every* `sorted_reviews` averaging expression to clamp, so a sixth site can't silently reintroduce the gap.

## Verification

```
$ pytest tests/unit/test_app_review_score_clamp.py tests/unit/test_apps_search_poison_guard.py
19 passed

# restore the unclamped expression:
3 failed   (out-of-range, negative, and parity-with-utils)
# and the invariant scan pinpoints it:
AssertionError: unclamped review-score aggregation at: ['routers/apps.py:700']
```

I verified the invariant test in **both** directions — my first regex silently passed against the buggy tree, so it was rewritten until it actually fails on the pre-fix code.

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10159?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: none

One aggregation site drifted from the shared clamp helper; no registered class covers it.

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
@@ -1,6 +1,6 @@
 {
   "files": {
-    "backend/routers/apps.py": 2387,
+    "backend/routers/apps.py": 2386,
     "backend/routers/chat.py": 1588,
     "backend/routers/developer.py": 2263,
     "backend/routers/mcp_sse.py": 1865,
--- a/backend/routers/apps.py
+++ b/backend/routers/apps.py
@@ -14,7 +14,7 @@ from fastapi import APIRouter, Body, Depends, Form, UploadFile, File, HTTPExcept
 from fastapi.responses import HTMLResponse
 
 from langchain_core.messages import SystemMessage, HumanMessage
-from utils.apps import fetch_app_chat_tools_from_manifest
+from utils.apps import _clamp_review_score, fetch_app_chat_tools_from_manifest
 from utils.executors import (
     critical_executor,
     db_executor,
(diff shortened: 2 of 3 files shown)


Git-on-my-level (commented):
Thanks for the careful fix. I reviewed the backend diff and this looks like the right direction: `/v2/apps/search` was the remaining aggregation path using raw stored review scores, and switching it to `_clamp_review_score(...)` brings search in line with the existing app aggregation helpers. The added regression coverage for in-range/no-review/out-of-range/negative scores is also useful, and the relevant backend checks are passing.

I’m leaving this as a positive maintainer signal rather than a formal approval because the PR also updates the CI line-count ratchet baseline and is in a backend ranking/integrity path, so a human maintainer should still make the final merge call.

---

kodjima33 (approved):
Backend bug fix (/v2/apps/search bypasses review-score clamp) — approve only; DIRTY (needs rebase) + workflow-review label

PR #10160, opened 2026-07-20
fix(backend): L2 promotion selector silently drops L1 candidates that overflow a batch

arhxam:
## Problem

`select_promotion_work_items` truncates a session's ids to whatever space is left in the current batch and **discards the remainder**:

```python
remaining = cfg.max_l1_items_per_batch - len(batch_l1_ids)
batch_l1_ids.extend(item_ids[:remaining])
```

`item_ids[remaining:]` is never re-queued. Any completed session holding more candidates than `max_l1_items_per_batch` (default **50**) silently loses the excess — those L1 items are **never promoted to L2**, and nothing reports it.

Reproduced against the real selector — 60 pending candidates in one completed session:

```
input candidates : 60
work items       : 1
ids emitted      : 50
DROPPED          : 10 -> ['i6','i7','i8','i9','i54','i55','i56','i57','i58','i59']
```

The *session* bound does not behave this way — `would_exceed_sessions` flushes and carries the session into a new work item. Item overflow was the one bound that **destroyed** work instead of deferring it.

## Fix

Consume the session's ids in a loop, flushing a full batch and opening a fresh one until none remain:

```python
pending_ids = list(item_ids)
while pending_ids:
    if len(batch_l1_ids) >= cfg.max_l1_items_per_batch:
        flush()
    active_uid = uid
    if session_id not in batch_sessions:
        batch_sessions.append(session_id)
    remaining = cfg.max_l1_items_per_batch - len(batch_l1_ids)
    batch_l1_ids.extend(pending_ids[:remaining])
    pending_ids = pending_ids[remaining:]
```

A session that spans batches is listed in each work item it contributes to. Sessions that fit in one batch behave exactly as before.

## Commits

1. **`test:`** characterize the batching — the module had **no unit coverage at all** (nothing under `tests/unit/` imported it). Green on unmodified code.
2. **`fix:`** loop instead of truncate, with the targeted regression cases.
3. **`test:`** a conservation property over 72 shapes (session count × items/session × item cap × session cap): every input id appears exactly once and no batch exceeds the cap.

## Verification

```
$ pytest tests/unit/test_l2_promotion_selector.py
9 passed

# restore the truncating expression:
3 failed, 5 passed
# and the property test pinpoints a minimal shape:
AssertionError: ids lost for sessions=1 per_session=4 item_cap=1 session_cap=1
```

Both new invariant-style tests were checked in **both** directions — they fail on the pre-fix code, not just pass on the fixed one.

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10160?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: none

Batch-bound truncation discarding queued work in a selector; FC-nonrecoverable-promotion covers multi-service deploy traffic promotion, not L1-to-L2 memory promotion, so it does not apply.

### DIFF
--- a/backend/jobs/l2_promotion_selector.py
+++ b/backend/jobs/l2_promotion_selector.py
@@ -173,10 +173,19 @@ def select_promotion_work_items(
         would_exceed_items = len(batch_l1_ids) + len(item_ids) > cfg.max_l1_items_per_batch
         if active_uid is not None and (uid != active_uid or would_exceed_sessions or would_exceed_items):
             flush()
-        active_uid = uid
-        batch_sessions.append(session_id)
-        remaining = cfg.max_l1_items_per_batch - len(batch_l1_ids)
-        batch_l1_ids.extend(item_ids[:remaining])
+        # A session holding more candidates than a single batch allows spans several work
+        # items. Keep consuming its ids, flushing a full batch and starting a fresh one,
+        # instead of truncating to the remaining space and discarding the rest.
+        pending_ids = list(item_ids)
+        while pending_ids:
+            if len(batch_l1_ids) >= cfg.max_l1_items_per_batch:
+                flush()
+            active_uid = uid
+            if session_id not in batch_sessions:
+                batch_sessions.append(session_id)
+            remaining = cfg.max_l1_items_per_batch - len(batch_l1_ids)
+            batch_l1_ids.extend(pending_ids[:remaining])
+            pending_ids = pending_ids[remaining:]
         if len(batch_l1_ids) >= cfg.max_l1_items_per_batch:
             flush()
     flush()
--- a/backend/tests/unit/test_l2_promotion_selector.py
+++ b/backend/tests/unit/test_l2_promotion_selector.py
@@ -1,0 +1,75 @@
+"""Batching behaviour of jobs.l2_promotion_selector.select_promotion_work_items.
+
+The selector groups completed-session L1 candidates into bounded work items. It is a pure
+function over dataclasses — no Firestore, no clock beyond the injected `now`.
+
+The batch bounds exist to cap the size of a single promotion job, not to discard work:
+every pending candidate must appear in exactly one returned work item.
+"""
+
+from datetime import datetime, timedelta, timezone
+
+from jobs.l2_promotion_selector import (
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
Thanks for the focused backend fix. I reviewed the selector change and the new regression coverage, and this looks like the right direction: the batching bound should cap individual L2 promotion jobs, not silently discard overflow L1 candidates. The loop now preserves all IDs across single-session overflow and awkward batch sizes, and the conservation test is a useful guard against this class of regression.

I’m not approving formally because the PR is still red: the changed selector tests passed in CI (`tests/unit/test_l2_promotion_selector.py`, 9 passed) and passed in my local stripped-env run, but the backend unit suite failed in an unrelated file (`tests/unit/test_ws_b_short_term_lifecycle.py`). From the diff, I don’t see a blocker in this PR; I’d treat this as a good candidate once CI is green or the unrelated failure is rerun/cleared.

---

kodjima33 (approved):
Backend bug fix (L2 promotion selector silently drops L1 candidates overflowing a batch) — approve only; red-main CI pre-existing (label ci-failing-pre-existing), unblocks after #10169

Git-on-my-level (approved):
CI is green now — the unrelated backend failure that held back formal approval has cleared, and this PR's selector tests pass cleanly (9/9, including the 72-shape conservation invariant).

Confirming approval: the fix correctly mirrors how session overflow already behaved — item overflow now carries forward into a new work item instead of being silently truncated. Downstream consumption is safe since the worker processes each work item independently via `l1_item_fetcher`.

Removing the stale `ci-failing-pre-existing` label since all checks are green on this head.

---

kodjima33 (approved):
Verified: selector truncated item_ids to the remaining batch space and dropped the overflow; new loop flushes and continues so every pending candidate lands in exactly one work item.

PR #10161, opened 2026-07-20
fix(backend): Anthropic completions return a hardcoded `created` timestamp copied from a test double

arhxam:
## Problem

`_anthropic_to_openai_response` hardcodes the OpenAI `created` field:

```python
normalized: dict[str, Any] = {
    'id': str(response.get('id') or 'anthropic_gateway'),
    'object': 'chat.completion',
    'created': 1782522000,
    ...
```

That literal appears in **exactly one other place in the entire repo** — `fake_success_response`, the in-module **test double**. It was copy-pasted out of the fake into the live provider path. The sibling Vertex normalizer does it correctly:

```python
'created': int(time.time()),   # _vertex_to_openai_response
```

`_anthropic_to_openai_response` runs inside `AnthropicMessagesProvider` immediately after the real Anthropic HTTP call, so **every completion served by an anthropic-primary lane carries the same permanently stale timestamp** — `1782522000` is `2026-06-27T05:00:00Z`.

`created` is part of the OpenAI `chat.completion` contract. Callers that order or correlate responses by it cannot distinguish two completions served a week apart.

## Fix

```diff
-'created': 1782522000,
+'created': int(time.time()),
```

`fake_success_response` keeps its fixed constant — a deterministic timestamp is the whole point of a fake. The diff is exactly one line.

## Commits

1. **`test:`** cover the Anthropic→OpenAI envelope. The existing Anthropic tests asserted request shaping and `finish_reason` but never the returned envelope. Green on unmodified code.
2. **`fix:`** compute `created`, with a frozen-clock regression test and a second case that advances the clock between two calls.
3. **`test:`** an invariant scan requiring every `created` before `fake_success_response` to be computed, so the fake's constant can't leak into a provider path again.

## Verification

```
$ pytest tests/unit/test_llm_gateway_openai_provider.py
27 passed

# restore the constant:
2 failed, 24 passed
# and the invariant scan pinpoints it:
AssertionError: hardcoded chat.completion `created` outside the test double at line(s): [722]
```

Both invariant-style tests were checked in **both** directions — they fail on the pre-fix code, not merely pass on the fixed one.

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: none

A test-double constant copied into a live provider path; no registered class covers it.

### DIFF
--- a/backend/llm_gateway/gateway/providers.py
+++ b/backend/llm_gateway/gateway/providers.py
@@ -719,7 +719,7 @@ def _anthropic_to_openai_response(
     normalized: dict[str, Any] = {
         'id': str(response.get('id') or 'anthropic_gateway'),
         'object': 'chat.completion',
-        'created': 1782522000,
+        'created': int(time.time()),
         'model': requested_model,
         'choices': [
             {
--- a/backend/tests/unit/test_llm_gateway_openai_provider.py
+++ b/backend/tests/unit/test_llm_gateway_openai_provider.py
@@ -394,3 +394,24 @@ async def test_anthropic_created_advances_between_calls(monkeypatch):
     assert first['created'] == 1_800_000_000
     assert second['created'] == 1_800_000_600
     assert second['created'] > first['created']
+
+
+def test_no_provider_normalizer_hardcodes_created():
+    """Only the test double may return a fixed `created` timestamp.
+
+    The Anthropic normalizer shipped with 1782522000 copied from
+    fake_success_response, which is invisible in review because the field is
+    well-formed. Require every other `created` in the providers module to be computed.
+    """
+    import re
+    from pathlib import Path
+
+    source = Path(providers_mod.__file__).read_text(encoding='utf-8')
+    fake_start = source.index('def fake_success_response')
+
+    offenders = [
+        source[: match.start()].count('\n') + 1
+        for match in re.finditer(r"'created':\s*(\d+)", source)
+        if match.start() < fake_start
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Thanks for the focused fix here. I reviewed the Anthropic-to-OpenAI normalization path and the added regression coverage.

This looks like the right behavior: the live Anthropic completion envelope now uses the response-time clock for `created`, while the deterministic fake keeps its fixed timestamp. The new tests cover the normalized OpenAI envelope, a frozen clock, advancing time across two calls, and a guard against copying a literal `created` timestamp back into provider normalizers.

I also ran the targeted backend unit file locally on this head:

`python -m pytest tests/unit/test_llm_gateway_openai_provider.py` → 27 passed

Positive signal from automation; leaving this for a human maintainer's final merge decision rather than a formal approval.

---

kodjima33 (approved):
Backend bug fix: use int(time.time()) instead of a hardcoded test-double timestamp in _anthropic_to_openai_response; test added, conf 5/5

PR #10164, opened 2026-07-21
fix(backend): every rate-limit window after the first serves one request too few

arhxam:
## Problem

`rate_limit_custom` charges the current request with a single shared decrement, but the window-reset branch has **already** spent a slot:

```python
if current_time - timestamp >= window_seconds:
    remaining = requests_per_window - 1  # Reset the counter for the new window
    timestamp = current_time
elif remaining == 0:
    raise HTTPException(status_code=429, detail="Too Many Requests")

remaining -= 1
```

The first-ever request takes the `else` branch and stores `requests_per_window - 1`. A request that *rolls the window over* stores `requests_per_window - 2` — the same situation recorded one lower.

**Every window after the first serves one fewer request than configured, permanently.** With `requests_per_window=5` the served counts per window are `5, 4, 4, 4, …`.

## Impact

Reachable from every caller:

- `routers/phone_calls.py` — phone verification (5/hour) and verify-check (30/minute)
- `routers/fair_use_admin.py` — case status (10/minute)

A user who verifies a phone number in two consecutive hours is cut off a request early in the second hour.

## Fix

The reset branch hands over the full quota and lets the shared decrement charge this request, exactly as the first-request branch does:

```diff
-remaining = requests_per_window - 1  # Reset the counter for the new window
+remaining = requests_per_window
```

## Commits

1. **`test:`** pin the first-request contract — the window bookkeeping had no coverage at all (the two existing cases only assert fail-open on a corrupt entry). Green on unmodified code.
2. **`fix:`** the off-by-one, with the regression cases.

Both commits are green, so the series stays bisectable.

## Tests

Added to the existing `test_rate_limit_json_failopen.py`, reusing its stub-finder harness — **no new isolation seam and no new dependency**:

- a reset window reserves the same as a first request
- three consecutive windows each serve the full quota — `(5, 5, 5)`

Against the pre-fix code these are `(5, 4, 4)` and the stored `remaining` is `3` rather than `4`.

## Verification

```
$ pytest tests/unit/test_rate_limit_json_failopen.py
5 passed          # pre-fix: assert 3 == 4 / assert (5, 4, 4) == (5, 5, 5)

fast-unit CPU                 : 0.01s (limit 1.00s)
```

I confirmed this is not intended behaviour by grepping the function name across the whole test tree (not by filename): `rate_limit_custom` appears only in `test_rate_limit_json_failopen.py`, and no test asserts `remaining` or the reset path. `test_rate_limiting.py` covers a different, Redis-backed limiter.

## Failure class

Failure-Class: none

An off-by-one in a single counter reset; no registered class covers it.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10164?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_rate_limit_json_failopen.py
+++ b/backend/tests/unit/test_rate_limit_json_failopen.py
@@ -14,6 +14,8 @@ import sys
 import types
 from unittest.mock import MagicMock, patch
 
+from fastapi import HTTPException
+
 os.environ.setdefault('OPENAI_API_KEY', 'sk-test-not-real')
 os.environ.setdefault('ENCRYPTION_SECRET', 'omi_ZwB2ZNqB2HHpMK6wStk7sTpavJiPTFg7gXUHnc4tFABPU6pZ2c2DKgehtfgi4RZv')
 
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Thanks for the focused fix. I reviewed the rate-limit bookkeeping change and the new regression coverage.

Positive signal from automation: the reset branch now mirrors the first-request path by handing the full quota to the shared decrement, so an expired window stores `requests_per_window - 1` after serving its first request instead of double-spending a slot. The added tests cover both the stored counter after reset and consecutive full windows, and I also verified the changed test file locally with:

`env -i HOME=$HOME PATH=$PATH USER=$USER OPENAI_API_KEY=dummy ENCRYPTION_SECRET=... .venv/bin/python -m pytest tests/unit/test_rate_limit_json_failopen.py -q`

Result: `5 passed in 0.34s`.

I am not formally approving yet because the current head still has a failing Backend unit suite in CI (`tests/unit/test_ws_b_short_term_lifecycle.py` per the GitHub log). That file looks unrelated to this two-file rate-limit change, but the failed required check needs to be green or triaged by a maintainer before merge.

---

kodjima33 (approved):
Backend bug fix (off-by-one: every rate-limit window after the first serves one request too few) — approve only; blocked on pre-existing red-main CI (#10169)

Git-on-my-level (approved):
Independent verification: reproduced the off-by-one in isolation — buggy branch serves (5, 4, 4, …) per window, fixed head serves (5, 5, 5, …), matching the regression tests. Confirmed the double-charge is still present on `main`, so this fixes a live bug.

The prior CI blocker (failing `tests/unit/test_ws_b_short_term_lifecycle.py`, unrelated to this two-file change) is now green on `2cc3c434`, and the full required set is clean/mergeable.

Approving on correctness grounds; merge is a human maintainer call.

kodjima33 (approved):
Verified: the window-reset branch stored requests_per_window-2 because it fell through to the shared decrement, so every window after the first served one request short; tests cover first window, reset, and full-quota.

PR #10165, opened 2026-07-21
fix(backend): app-review push notifications say "None reviewed <app>"

arhxam:
## Problem

Both app-review notifications resolve a display name with `dict.get`'s default:

```python
owner_name = app_owner.get('display_name', 'The developer') if app_owner else 'The developer'
reviewer_name = reviewer.get('display_name', 'A user') if reviewer else 'A user'
```

But `database.auth.get_user_from_uid` builds its dict from `firebase_admin`'s `UserRecord` and **always emits the key**:

```python
return {
    'uid': user.uid,
    ...
    'display_name': user.display_name,
    ...
}
```

`display_name` is `None` for any account created without one — email/password signup, Apple private-relay. Since the key is *present*, `dict.get` never falls back and the `None` is formatted straight into the push title:

```
"None reviewed Weather"
"None (Weather)"
```

Reachable from the review endpoints in `routers/apps.py` (1239, 1278, 1317).

## Fix

Fall back on the **value**, not the key's absence:

```diff
-owner_name = app_owner.get('display_name', 'The developer') if app_owner else 'The developer'
+owner_name = (app_owner or {}).get('display_name') or 'The developer'
```

This matches the repo's own idiom — `utils/users.py` does `return name or default`, and `routers/apps.py:788` already guards with `get_user_from_uid(uid) or {}`.

## Commits

1. **`test:`** pin the titles for a user who *has* a name and for a missing user record — these notifications had **no assertions anywhere** (the only reference in the suite is a `patch.object` mock). Green on unmodified code.
2. **`fix:`** fall back on the value, with the regression cases.

Both commits are green, so the series stays bisectable.

## Tests

New `tests/unit/test_app_review_notification_names.py`, loading `utils.notifications` through the sanctioned `stub_modules` + `load_module_fresh` seam (no module-scope `sys.modules` mutation):

- `display_name` of `None` or `''` → `"A user reviewed Weather"` / `"The developer (Weather)"`
- a real name is still used
- a missing user record still falls back

Whitespace-only names are deliberately **not** handled — that keeps the fix to one expression and consistent with the existing `name or default` idiom.

## Verification

```
$ pytest tests/unit/test_app_review_notification_names.py
6 passed          # pre-fix: assert ['None reviewed Weather'] == ['A user reviewed Weather']

module-scope sys.modules gate : 757 files, 0 violations
fast-unit CPU                 : 0.02s (limit 1.00s)
```

Confirmed not intentional by grepping the function names across the whole test tree (not by filename).

## Failure class

Failure-Class: none

A `dict.get` default that cannot fire because the key is always present; no registered class covers it.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10165?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_app_review_notification_names.py
+++ b/backend/tests/unit/test_app_review_notification_names.py
@@ -1,0 +1,83 @@
+"""App-review notifications must not address users as "None".
+
+``database.auth.get_user_from_uid`` always emits a ``display_name`` key — it is built
+from ``firebase_admin``'s UserRecord, whose ``display_name`` is ``None`` for any account
+created without one (email/password signup, Apple private-relay). ``dict.get(key,
+default)`` only falls back when the key is *absent*, so the stored ``None`` was used
+verbatim and the push title read "None reviewed Weather".
+
+Isolation: utils/notifications.py pulls firebase_admin at import, so it is loaded through
+the sanctioned Tier-2 reserve seam (stub_modules + load_module_fresh) inside a fixture.
+See backend/docs/test_isolation.md.
+"""
+
+from contextlib import contextmanager
+from pathlib import Path
+from types import ModuleType
+from typing import Any, Iterator
+
+import pytest
+
+from testing.import_isolation import AutoMockModule, load_module_fresh, stub_modules
+
+BACKEND_DIR = Path(__file__).resolve().parents[2]
+
+
+def _module(name: str, **attributes: Any) -> ModuleType:
+    module = ModuleType(name)
+    for key, value in attributes.items():
+        setattr(module, key, value)
+    return module
+
+
+@contextmanager
+def _notifications_with_user(user: Any) -> Iterator[Any]:
+    """Load utils.notifications with get_user_from_uid returning `user`."""
+    stubs = {
+        name: AutoMockModule(name)
+        for name in (
+            'firebase_admin',
+            'firebase_admin.messaging',
+            'firebase_admin.auth',
+            'database.notifications',
+            'database.redis_db',
+            'utils.llm.notifications',
+        )
+    }
+    stubs['database.auth'] = _module('database.auth', get_user_from_uid=lambda _uid: user)
+    with stub_modules(stubs):
+        yield load_module_fresh('utils.notifications', str(BACKEND_DIR / 'utils' / 'notifications.py'))
+
+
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Nice fix — verified the root cause and the resolution against source.

Confirmed that `database.auth.get_user_from_uid` always emits a `display_name` key (built straight from `firebase_admin`'s `UserRecord`), so for nameless accounts (email/password, Apple private-relay) the key is *present* with value `None`, and the old `dict.get('display_name', default)` never fell back. The resulting `"None reviewed Weather"` / `"None (Weather)"` titles are real and reachable from the review endpoints in `routers/apps.py`.

The fix (`(app_owner or {}).get('display_name') or 'The developer'`) is correct: it falls back on the *value*, handles the missing-user-record case, and matches the repo's own idiom — `routers/apps.py` already uses `get_user_from_uid(uid) or {}`, and `utils/users.py` uses `name or default`.

The regression tests are well-constructed: they pin both notifications across `None` / `''` / missing-user-record / real-name cases, and they load `utils.notifications` through the sanctioned `stub_modules` + `load_module_fresh` isolation seam rather than mutating `sys.modules` at module scope. Good that the test commit lands first so the series stays bisectable.

On checks: the only failing check is `Hygiene` in a `Repo Checks` run that was cancelled ~9s after it started (a concurrent re-trigger); the re-run succeeded with `Hygiene`/`Formatting`/`Detect Changes` correctly skipped (path-filtered — only `backend/` changed). `Backend unit suite`, `Backend Hermetic E2E`, and `Backend Checks` all pass. So the red is an infra artifact, not a code issue — no action needed from your side.

This is a small, correct, well-tested bug fix. Leaving a positive signal; final merge call for a maintainer.

---

kodjima33 (approved):
Backend bug fix: fall back on display_name value not key presence so app-review pushes don't say 'None reviewed'; test added, conf 5/5

PR #10167, opened 2026-07-21
fix(backend): semaphore cache prune drops the live loop's entry, doubling the concurrency cap

arhxam:
## Problem

`_get_semaphore` keys semaphores by `(loop_id, name)` and prunes when the cache grows past `_SEMAPHORE_CACHE_MAX`. Its docstring sets the contract:

> The main FastAPI event loop … reuses the same `loop_id` for the lifetime of the process, so **its semaphores are stable**. Entries from short-lived `asyncio.run()` loops are pruned when the cache grows beyond `_SEMAPHORE_CACHE_MAX`.

But the prune is a blanket `clear()`:

```python
if key not in _semaphores:
    if len(_semaphores) > _SEMAPHORE_CACHE_MAX:
        _semaphores.clear()          # <- also discards the running loop's entries
    _semaphores[key] = asyncio.Semaphore(limit)
```

Tasks already holding permits keep them on the **old** `Semaphore` while the next caller is handed a **fresh** one. For the duration of those in-flight requests the effective concurrency is up to **twice** the configured limit — 128 rather than 64 for webhooks. That is the opposite of what the cap exists for.

It's reachable in the normal way: sync FastAPI endpoints call `asyncio.run()`, each new loop adds an entry, and crossing 100 triggers the prune while running under the main loop.

## Fix

Evict only entries belonging to *other* loops:

```python
def _evict_foreign_loop_semaphores(live_loop_id: int) -> None:
    for key in [k for k in _semaphores if k[0] != live_loop_id]:
        del _semaphores[key]
```

This mirrors the two sibling evictors already in this module — `_evict_stale_circuit_breakers` and `_evict_stale_latest_wins` — which both remove just the stale keys rather than emptying the map.

## Tests

The prune path had **no coverage**: `_SEMAPHORE_CACHE_MAX` appears in no test, and the existing cases only cover same-loop identity and cross-loop isolation.

- crossing the cap leaves the running loop's existing semaphore **identical**
- the prune still drops every foreign-loop entry, so the cache stays bounded

One implementation note worth flagging for review: the prune only runs when a **new** key is created, so the tests request a *second* semaphore name to trigger it. My first draft asked for the same name again, which returns the cached entry and exercised nothing — those tests passed against the buggy code until I corrected them.

## Verification

```
$ pytest tests/unit/test_async_http_infrastructure.py
47 passed

# restore _semaphores.clear():
2 failed   (test_pruning_keeps_the_running_loops_semaphore,
            test_pruning_still_bounds_the_cache)

fast-unit CPU                 : 0.06s max (limit 1.00s)
```

## Failure class

Failure-Class: none

A cache eviction that discards live state along with stale state; no registered class covers it.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_async_http_infrastructure.py
+++ b/backend/tests/unit/test_async_http_infrastructure.py
@@ -68,6 +68,7 @@ from utils.http_client import (
     _webhook_circuit_breakers,
     _latest_wins_versions,
     _semaphores,
+    _SEMAPHORE_CACHE_MAX,
     _CIRCUIT_BREAKER_FAILURE_THRESHOLD,
     _CIRCUIT_BREAKER_RECOVERY_TIMEOUT,
 )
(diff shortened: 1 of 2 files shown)


arhxam:
Heads-up on the `Backend unit suite` result here: it fails `tests/unit/test_desktop_transcribe.py::TestTranscribeStreamWebSocket::test_ws_parakeet_connection_failure_falls_back_before_audio` with `Expected 'send' to be called once. Called 0 times.`

That is not from this PR. The identical test with the identical assertion also failed on an unrelated PR of mine touching a completely different module, while passing on three others opened minutes earlier — the signature of a flaky async websocket test rather than a regression. The file contains no reference to the code this PR changes.

The other failing file, `tests/unit/test_ws_b_short_term_lifecycle.py`, is the pre-existing breakage on `main` (every recent run on `main` fails those same 7 tests). I opened #10169 to fix it — its root cause is a pinned `NOW` that the wall clock overtook on 2026-07-20.

Git-on-my-level:
Good catch and a clean fix. The blanket `_semaphores.clear()` genuinely dropped the running loop's own entry, so a caller holding permits on the old `Semaphore` kept them while the next caller got a fresh one — briefly doubling the effective cap. Replacing it with a foreign-loop-only eviction mirrors `_evict_stale_circuit_breakers` / `_evict_stale_latest_wins` and preserves the docstring contract. The two new tests are well-targeted (identity preservation + cache-bounding), and the previously-uncovered prune path now has coverage.

On checks: every test in `test_async_http_infrastructure.py` passes, including both new cases. The Backend unit suite failures are in `test_desktop_transcribe.py` and `test_ws_b_short_term_lifecycle.py`, both of which also fail on `main` (pre-existing, unrelated to this PR). Not formally approving because the check suite is red, but the code change itself looks correct and low-risk — ready for a maintainer to merge once those pre-existing failures are confirmed independent of this change.

---

kodjima33 (approved):
Backend bug fix (semaphore cache prune drops live loop entry, doubling concurrency cap) — approve only; blocked on pre-existing red-main (Backend unit suite / ws_b lifecycle rot, fixed by #10169)

Git-on-my-level:
Follow-up on the earlier comment: the `Backend unit suite` now passes on this head (`316d411`), alongside the hermetic merge gate and the rest of the backend checks. The pre-existing `test_ws_b_short_term_lifecycle.py` / `test_desktop_transcribe.py` failures tracked in #10169 have cleared. With checks green, the change is clear to merge — approving.

Git-on-my-level (approved):
Confirmed on the current head: the foreign-loop-only eviction is correct (preserves the running loop's semaphores, mirrors the sibling `_evict_stale_*` evictors), the two new tests lock in both identity-preservation and cache-bounding, and all checks including `Backend unit suite` are green. Small, cohesive, low-risk backend concurrency fix. Approving.

kodjima33 (approved):
Verified: the blanket _semaphores.clear() prune also dropped the running loop's entry, briefly doubling effective concurrency; scoped eviction mirrors _evict_stale_circuit_breakers and is covered by two tests.

Issue #10190, opened 2026-07-21
admin: HogQL queries without LIMIT are silently truncated to 100 rows (releases panel likely affected); shared guard needed


kodjima33:
PostHog's query API fills `LIMIT 100` into any HogQL (sub)query that carries none and truncates silently. This cut the newest days off the response-reliability charts (fixed in the reliability PR by pinning explicit limits, wrapping the voice UNION — a trailing LIMIT binds to the last arm only — and marking truncation `partial`). The class remains elsewhere:

- `web/admin/app/api/omi/releases/route.ts` groups by version×event over 60 days with no LIMIT and doesn't use the shared helper — 5 events × ~40–80 desktop versions plausibly exceeds 100 rows **today**, so per-version crash/launch/feedback counts may already read as 0 for most versions.
- `web/admin/app/api/omi/stats/notifications/route.ts` runs day-grouped queries with an unclamped `days` param.
- ~5 routes each hand-fixed this with ad-hoc limits (retention/macos-versions: 100000, onboarding: 10000, profitability: 500000) — per AGENTS.md, 2+ fixes sharing a cause warrant a reusable guard in the shared layer (`web/admin/lib/posthog.ts`).
- Related bound: `writeCache` silently skips results >900KB (Firestore field cap), so the heaviest queries lose caching + stale-on-throttle fallback exactly when they grow large (~8k rows at ~110B/row). No log when it happens.

Evidence (verified live against project 302298):
- `SELECT 1 UNION ALL SELECT 2 LIMIT 1` → 2 rows (trailing LIMIT binds to last arm).
- `numbers(150) UNION ALL numbers(150)` with no LIMIT → 200 rows (default 100 per arm).
- `LIMIT 100000` → exactly 50000 rows returned (PostHog's served max is 50k, not 100k — the ad-hoc 100000/500000 limits above overstate what they get).
- posthogResults now warns on the exact-100-rows-no-LIMIT signature (landed with the reliability fix), but that's an observer, not a guard.

Suggested shape (for the fix PR to decide): audit/pin explicit limits per route, or a helper that rejects/wraps LIMIT-less queries, plus a hermetic test over the route query builders. Any new check must run in an existing CI lane.

PR #10175, opened 2026-07-21
feat(admin): add response reliability by release channel

kodjima33:
## Summary

- add two response-reliability graphs to the existing admin dashboard, split by Production and Beta update channel
- show Chat success, Voice success, failure count, and average full-answer latency from the submitted shortcut boundary
- emit physical voice-turn start/terminal telemetry while using existing floating-voice lifecycle telemetry until the new desktop events accumulate

## Verification

- `cd web/admin && npm run check && npm run build` — 28 tests passed, typecheck/build passed (existing lint warnings only)
- `cd desktop/macos && ./scripts/agent-logic-harness.sh --swift-only` — focused lifecycle/state tests passed
- live local API returned HTTP 200 with distinct Production and Beta summaries
- rendered the real dashboard at `http://127.0.0.1:3100/dashboard`; `.cross-review-verify.png` shows exactly two standard draggable/resizable dashboard graph tiles with explicit Voice success

## Product invariants affected

- INV-CHAT-1
- INV-VOICE-1

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10175?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10191, opened 2026-07-21
fix(admin): response reliability charts dropped newest days (HogQL default LIMIT 100)

kodjima33:
## What

The response-reliability charts (added in #10175) were cutting off the newest days — production chat stopped at Jul 17 on a Jul 21 dashboard — and rendered mostly-empty 30-day axes. Root cause: the HogQL queries carried no `LIMIT`, so PostHog's query API filled in its **default LIMIT 100 per (sub)query**. Rows are grouped per `(day, event, surface, reason, actor)` and ordered `day ASC`, so the cap dropped exactly the newest days, worse every day as volume grows (raw PostHog had 21/11/32 started events on Jul 18–20 that never reached the chart).

## Fixes

- **Explicit `LIMIT 50000` on both queries.** 50k is PostHog's actual served maximum, verified live (`LIMIT 100000` returns exactly 50000 rows); the overflow tripwire compares `rows.length >= RELIABILITY_ROW_LIMIT`, so the constant must equal the effective cap — the test pins it.
- **Voice union wrapped in a subquery with one outer `ORDER BY`/`LIMIT`.** In HogQL a trailing LIMIT after `UNION ALL` binds to the last arm only (verified: `SELECT 1 UNION ALL SELECT 2 LIMIT 1` → 2 rows), so the physical_shortcut arm would have kept the default cap.
- **Truncation is user-visible, not just logged:** hitting the row limit sets `partial: true` through both payload builders and the chart renders the amber partial-data warning per `web/admin/AGENTS.md`; the route still logs details.
- **Axes trimmed to telemetry coverage** (`trimDailyToCoverage`): charts start at the first day with activity instead of 3 empty weeks; subtitle says "Since Jul 13" when coverage begins inside the window. A leading all-cancelled day counts as coverage (it's a signal); interior gaps are kept.
- **Tooltips show judged counts** (`76.5% (13/17)`) so small-sample swings (beta's 0% days are single failed responses) are interpretable; 2px dots mark real data points vs `connectNulls` interpolation.
- **Class-level tripwire in `posthogResults`:** warns whenever a LIMIT-less query returns exactly 100 rows (the default-cap signature) — covers every admin route. Remaining class instances (releases/notifications routes, cache-size skip) tracked in #10190.
- Firestore channel-lookup chunks now fetch in parallel (actor fan-out is no longer bounded by ~100 rows).

## Verification

- Replayed the exact production PostHog rows + Firestore channel map through `buildChannelReliabilityPayloads`: without LIMIT the chat series ends 2026-07-18; with it, 2026-07-20 (today has no judged responses yet).
- Ran the dashboard locally (`next dev`, dev-bypass auth, production PostHog + Firestore): production chat line continuous Jul 13–20 (141 judged, "Since Jul 13"), beta Jul 15–20 (13 judged), voice days Jul 16/18/20 intact through the wrapped union, tooltip counts visible. Verified in Chrome and a clean Playwright browser; API `partial=false` in the healthy state, 401 without auth, `days=7` and invalid `days` behave.
- UNION/LIMIT semantics and the 50k cap verified against the live PostHog instance (queries in commit messages).
- `web/admin`: vitest 31 passed (8 reliability tests), `tsc --noEmit` clean.
- Six-agent adversarial review; all its confirmed findings addressed in the second commit.

none

## Failure class (fixes)

Failure-Class: none

Single-instance defect in queries added yesterday (#10175); the shared-layer recurrences are inventoried with evidence in #10190 rather than broadened into this PR. A class tripwire (exact-100-rows warn in `posthogResults`) lands here.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/admin/app/api/omi/stats/response-reliability/route.ts
+++ b/web/admin/app/api/omi/stats/response-reliability/route.ts
@@ -27,12 +27,16 @@ async function getReliabilityChannels(
   const channelByActor = new Map<string, ReliabilityChannel>();
   const db = getDb();
 
-  for (const actorIdChunk of chunk(actorIds, 30)) {
-    const snapshot = await db
-      .collection("users")
-      .where(admin.firestore.FieldPath.documentId(), "in", actorIdChunk)
-      .select("update_channel")
-      .get();
+  const snapshots = await Promise.all(
+    chunk(actorIds, 30).map((actorIdChunk) =>
+      db
+        .collection("users")
+        .where(admin.firestore.FieldPath.documentId(), "in", actorIdChunk)
+        .select("update_channel")
+        .get(),
+    ),
+  );
+  for (const snapshot of snapshots) {
     for (const doc of snapshot.docs) {
       const updateChannel = String(doc.get("update_channel") || "");
       channelByActor.set(
@@ -103,10 +107,10 @@ export async function GET(request: NextRequest) {
 
   const chatRows = chatAvailable ? chatResult.value : [];
   const voiceRows = voiceAvailable ? voiceResult.value : [];
-  if (
+  const truncated =
     chatRows.length >= RELIABILITY_ROW_LIMIT ||
-    voiceRows.length >= RELIABILITY_ROW_LIMIT
-  ) {
+    voiceRows.length >= RELIABILITY_ROW_LIMIT;
+  if (truncated) {
     console.warn("Response reliability rows hit the HogQL limit", {
       chatRows: chatRows.length,
       voiceRows: voiceRows.length,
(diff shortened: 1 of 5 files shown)


kodjima33:
@Git-on-my-level admin.omi.me deploy for this fix is queued behind the prod environment gate: https://github.com/BasedHardware/omi/actions/runs/29808114590 — needs your approval (the earlier #10175 deploy run was superseded by this one and cancelled). Note the reliability charts have never reached admin.omi.me yet; this run ships #10175 + this fix together.

PR #10195, opened 2026-07-21
Reject an empty audio frame instead of stopping the Modulate send loop

ZachL111:
## Problem

`SafeModulateSocket` uses `b''` as its shutdown sentinel: `finish()` enqueues it and `_send_loop`
breaks on it.

```python
async def _send_loop(self) -> None:
    while not self._closed and not self._dead:
        data = await self._send_queue.get()
        if data == b'':
            break
```

`send()` did not reject an empty payload, so a zero-length audio frame enqueued that same sentinel
and ended the send loop mid-session.

The failure is silent and permanent for the rest of the session:

- the socket is not marked dead on that path, so `is_connection_dead` stays False and
  `live_stt_socket_is_dead()` reports it healthy
- `send()` keeps returning True, so `send_live_stt_audio` believes audio is flowing and never
  terminates or reconnects the session
- every later frame is queued to a loop nobody is reading

The user's transcription simply stops, with no error surfaced to the client.

## Reachability

An empty frame reaches `send()` from the live path:

- `send_live_stt_audio` (`utils/stt/live_failure.py`) passes its `audio` straight to
  `stt_socket.send(audio)` with no emptiness check.
- `decide_multi_channel_stt_send` (`utils/transcribe_decisions.py`) computes
  `should_send = socket_available and not fair_use_dg_budget_exhausted`. It takes `pcm_len` only for
  usage accounting and never requires it to be non-zero.
- `resample_pcm(b'', ...)` returns `b''`.
- In `_handle_multi_channel_audio` the `if not audio: return` guard sits inside the opus branch
  only. On the pcm codec path `audio = data[1:]`, so a frame carrying just its channel byte
  produces `b''`.

## Fix

Reject an empty payload in `SafeModulateSocket.send`, matching the guard both Parakeet sockets
already have (`if not data: return True`). The wav header stays pending because `_header_sent` is
only set once the header is actually queued, so the next real frame still prepends it.

## Tests

`tests/unit/test_modulate_empty_frame_send_loop.py` (new) drives a real `SafeModulateSocket` send
loop against a fake provider websocket, in the same harness style as the existing
`test_modulate_stt.py`:

- an empty frame followed by real audio: the real audio still reaches the provider
- normal audio still streams, and `finish()` still stops the loop

## Verification

Run in `backend/`:

- `pytest tests/unit/test_modulate_empty_frame_send_loop.py`: 2 passed
- prove-fail: with `utils/stt/streaming.py` reverted to origin/main and confirmed byte-identical
  (`git diff origin/main` empty), the empty-frame test fails with
  `assert b'real_audio_chunk' in []`, i.e. the provider received nothing at all after the empty
  frame. Reapplied: 2 passed
- `pytest tests/unit/test_modulate_stt.py tests/unit/test_live_stt_failure.py tests/unit/test_modulate_empty_frame_send_loop.py`: 108 passed
- `pyright -p pyrightconfig.json utils/stt/streaming.py`: 0 errors
- `.github/scripts/check_product_file_line_count_ratchet.py`: OK, no line-count increase

## Impact

Any session on the Modulate STT provider, one of the two default serving providers, that receives a
single zero-length audio frame loses transcription for the rest of the session, silently and with no
reconnect. The guard is one condition and matches the two sibling sockets.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10195?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_modulate_empty_frame_send_loop.py
+++ b/backend/tests/unit/test_modulate_empty_frame_send_loop.py
@@ -1,0 +1,74 @@
+"""Regression: an empty audio frame must not shut down the Modulate send loop.
+
+``b''`` is ``SafeModulateSocket``'s shutdown sentinel -- ``finish()`` enqueues it and
+``_send_loop`` breaks on it. ``send()`` did not reject an empty payload, so a zero-length audio
+frame enqueued that sentinel and ended the send loop mid-session. The socket is not marked dead in
+that path, so ``is_connection_dead`` stays False and ``send()`` keeps returning True: every later
+frame is queued to a loop nobody is reading, and the user's transcription silently stops for the
+rest of the session with no error and no reconnect. Both Parakeet sockets already guard with
+``if not data: return True``.
+
+Empty frames are reachable from the live path: ``send_live_stt_audio`` passes its ``audio``
+straight to ``stt_socket.send`` with no emptiness check, ``decide_multi_channel_stt_send`` takes
+``pcm_len`` only for usage accounting and never requires it to be non-zero, and
+``resample_pcm(b'', ...)`` returns ``b''`` (so a zero-length client frame, or a PCM multi-channel
+frame carrying only its channel byte, produces one).
+"""
+
+import asyncio
+from unittest.mock import AsyncMock
+
+from utils.stt.streaming import SafeModulateSocket
+
+REAL_AUDIO = b'real_audio_chunk'
+
+
+def _drive_send_loop(frames):
+    """Drive a real SafeModulateSocket over `frames`; return the frames that reached the provider."""
+    provider_frames = []
+    real_audio_seen = asyncio.Event()
+
+    async def record_provider_frame(frame):
+        provider_frames.append(frame)
+        if frame == REAL_AUDIO:
+            real_audio_seen.set()
+
+    ws = AsyncMock()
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for tracking this down — this looks like a small, well-scoped backend bug fix.

I reviewed the Modulate send-loop change and the regression test. Treating an empty `data` payload as a no-op before queueing is consistent with the Parakeet sockets, preserves the pending WAV header for the next real frame, and avoids accidentally enqueueing the `b''` stop marker mid-session. The new test exercises the important failure mode: an empty frame followed by real audio still reaches the provider.

I’m not formally approving from automation because the required check summary currently has Formatting/Hygiene in a non-passing state from the repo checks, but implementation-wise this is a positive signal from my review. Once the check state is clean/rerun, this should be straightforward for a maintainer to merge.

---

kodjima33 (approved):
Backend bug fix: reject empty audio frame so it can't hit the b'' shutdown sentinel and silently kill the Modulate send loop; test added, conf 5/5

PR #10200, opened 2026-07-21
fix(ci): scope desktop-swift-ci main runs per-SHA so the release source gate can't deadlock

### LINKED ISSUE #10177: desktop-swift-ci on main is cancelled by every newer push — release source gate deadlocks on busy nights
Tonight the candidate source gate could not be satisfied for hours: `Desktop Swift Build & Tests` on the newest desktop commit (e298ad2331, #10172) was cancelled three times (03:50Z push-triggered, 04:33Z, and a manual rerun at ~04:50Z each killed by the next main push via the workflow concurrency group). Non-desktop pushes skip the check, so the planner's newest-desktop-SHA gate had no green run to accept and `desktop_auto_release` refused to cut v0.12.91 — evidence: planner run 29765xxx reason 'source gate blocked', run list shows completed/cancelled ×3. Used release_mode=break_glass (source code identical to PR #10059 head which passed the full 22-check suite pre-merge). Fix direction: desktop-swift-ci main-branch runs should not share a cancel-in-progress concurrency group with unrelated pushes (scope the group per head SHA, or let the planner accept the merged PR's head-check as the source proof).

aryanorastar:
# Summary

`desktop-swift-ci.yml` grouped its concurrency by `github.event.pull_request.number || github.ref`. On a **push to main** there is no PR number, so every main push shared the single group `desktop-swift-refs/heads/main` with `cancel-in-progress: true` — each new push cancelled the previous commit's in-progress run.

On a busy night the newest desktop commit's `Desktop Swift Build & Tests` run kept getting cancelled before it finished (#10177: `e298ad2331` cancelled 3×), so `desktop_auto_release`'s "newest-desktop-SHA has a green source run" gate had nothing to accept and refused to cut v0.12.91 — it shipped only via `break_glass`.

Fix: the non-PR fallback is now **`github.sha`**, so each main commit's run gets its own concurrency group and completes instead of being cancelled by the next push.

Fixes #10177.

# Why this shape

- **PRs unchanged:** `github.event.pull_request.number` is still first, so a newer push to the same PR supersedes the older run (the desirable cancel-in-progress behavior).
- **Main pushes:** fall through to `github.sha` — unique per commit, so an unrelated later push can't cancel the run that the release source gate needs. This is exactly the issue's suggested direction ("scope the group per head SHA").
- `cancel-in-progress` stays: for a per-SHA group it only supersedes a manual rerun of the *same* commit, which is what you want.

# Failure class

Failure-Class: none

# Product invariants affected

# Validation

- Regression guard added to the existing `test_desktop_swift_ci_contract.py` (already wired into the checks manifest, triggered by `desktop-swift-ci.yml`): the concurrency group must use `github.sha` for the non-PR case, plus an adversarial test that a revert to the shared per-ref group is detected.
- `python3 .github/scripts/test_desktop_swift_ci_contract.py` → **17 passed** (2 new).
- **Old-code check:** reverting the workflow fallback to `github.ref` fails the new guard (1/17) — it catches exactly the deadlock condition.
- Workflow YAML parses (`yaml.safe_load`).

Not exercised live: a real busy-night main-push storm (can't reproduce GitHub's concurrency scheduler locally). The mechanism is a documented GitHub Actions concurrency behavior; the contract test pins the group expression that produces the correct per-SHA scoping.

# Out of scope, named

- The alternative the issue floats — having the planner accept the merged PR's head-check as source proof — is a larger planner change; this PR fixes the root cause (the CI run being cancelled) so the existing gate works as designed.
- Other workflows' concurrency groups are not audited here; if any share the same per-ref pattern on main they'd deserve the same fix, but this PR is scoped to the one that deadlocked.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10200?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/test_desktop_swift_ci_contract.py
+++ b/.github/scripts/test_desktop_swift_ci_contract.py
@@ -264,6 +264,39 @@ class DesktopSwiftCIContractTests(unittest.TestCase):
             "gate condition must be absent after removal",
         )
 
+    # --- concurrency scoping (#10177) --------------------------------------
+
+    def test_main_push_concurrency_group_is_scoped_per_sha(self):
+        """A push to main (no PR number) must group by commit SHA, not the ref.
+
+        A shared per-ref group with cancel-in-progress meant every main push
+        cancelled the previous commit's in-progress run, so the release source
+        gate never saw a green run for the newest desktop SHA (#10177). The
+        non-PR fallback must therefore be github.sha.
+        """
+        group = re.search(r"^concurrency:\n  group:\s*([^\n]+)", _workflow_text(), re.MULTILINE)
+        self.assertIsNotNone(group, "workflow must declare a concurrency group")
+        group_expr = group.group(1)
+        self.assertIn("github.event.pull_request.number", group_expr, "PRs still group by PR number")
+        self.assertIn("github.sha", group_expr, "non-PR (main push) runs must group per commit SHA")
+        self.assertNotRegex(
+            group_expr,
+            r"pull_request\.number\s*\}\}\s*\|\|\s*github\.ref\b|pull_request\.number\s*\|\|\s*github\.ref\b",
+            "main pushes must not share a per-ref cancel-in-progress group (#10177)",
+        )
+
+    def test_adversarial_shared_per_ref_concurrency_detected(self):
+        """Reverting the fallback to github.ref re-introduces the shared
+        per-ref group that deadlocked the release source gate — the guard
+        above must reject it."""
+        tampered = _workflow_text().replace(
+            "github.event.pull_request.number || github.sha",
+            "github.event.pull_request.number || github.ref",
+        )
+        group_expr = re.search(r"^concurrency:\n  group:\s*([^\n]+)", tampered, re.MULTILINE).group(1)
+        self.assertRegex(group_expr, r"pull_request\.number\s*\|\|\s*github\.ref\b")
+        self.assertNotIn("github.sha", group_expr)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/.github/workflows/desktop-swift-ci.yml
+++ b/.github/workflows/desktop-swift-ci.yml
@@ -6,8 +6,14 @@ on:
   pull_request:
     branches: main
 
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
Maintainer review: this looks like the right targeted fix for #10177.

What I checked:
- The workflow keeps PR concurrency grouped by `github.event.pull_request.number`, so newer pushes to the same PR still cancel older runs.
- Main-branch pushes now fall back to `github.sha`, which avoids unrelated later main pushes cancelling the desktop source-proof run needed by the release gate.
- The added contract test covers the intended invariant, and I verified it locally with `python3 .github/scripts/test_desktop_swift_ci_contract.py` on `8b9ab917ca3b7f4b11b2fabf554a9ea5f7906f4d` (17 passed).

I do not see a code-level blocker. Because this touches CI/release workflow behavior, I’m leaving this as a positive signal rather than formal approval; a human maintainer should still confirm the release-gate behavior and wait for the queued Desktop Swift CI lane before merge.

---

aryanorastar:
@Git-on-my-level need human response — CI/release-gate change (fixes the #10177 deadlock). Contract test added and verified locally (17 passed). Needs a human maintainer to confirm the release-gate behavior and merge.

aryanorastar:
Obsolete — main already resolves #10177. The workflow now uses `group: desktop-swift-${{ github.event.pull_request.number || github.sha }}` with `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` (per-SHA main groups, PR-only cancellation), and `.github/scripts/test_desktop_swift_ci_contract.py` already carries `test_later_main_push_cannot_cancel_exact_sha_release_evidence` plus additional guards — a superset of what this PR added. Nothing left to land. Closing.

PR #10201, opened 2026-07-21
fix(ci): wait for qualification run terminal state before beta promotion trusts its conclusion

### LINKED ISSUE #10186: Automatic beta promotion handoff race: qualify dispatches promote with its own in-progress run id
`desktop_qualify_beta.yml`'s final step ('Request protected beta promotion') dispatches `desktop_promote_beta.yml` with `qualification_run_id=$GITHUB_RUN_ID` — its own, still-running workflow run. If the promotion workflow validates that the referenced qualification run has `conclusion=success`, the check races: at dispatch time the qualification run is `in_progress` (its conclusion isn't set until after this very step completes). Either promotion rejects a legitimate qualification, or (worse) accepts a run whose final steps could still fail after the dispatch. Found during cross-review on 2026-07-21 while shipping the side-by-side Omi Beta pipeline; currently unreachable in production because automatic beta promotion is paused (`DESKTOP_AUTO_BETA_ENABLED=false`). Fix direction: trigger promotion via `workflow_run` on qualification completion (conclusion filter), or have promote poll/verify the run's terminal conclusion with the immutable evidence artifact rather than trusting a mid-run dispatch. Needs a guard/test for the in-progress-qualification case before the pause is lifted.

aryanorastar:
# Summary

`desktop_qualify_beta.yml` dispatches `desktop_promote_beta.yml` **as its final step**, passing its own `$GITHUB_RUN_ID` as `qualification_run_id`. `desktop_promote_beta` then asserted:

```bash
run=$(gh api "repos/$REPO/actions/runs/$QUALIFICATION_RUN_ID")
test "$(jq -r .conclusion <<<"$run")" = success
```

At dispatch time that qualification run is still `in_progress` and its `.conclusion` is `null` — so the check **races**: `test null = success` rejects a legitimate qualification, or (evaluated once, later) could accept one whose post-dispatch steps still fail (#10186).

Latent today because automatic beta promotion is paused (`DESKTOP_AUTO_BETA_ENABLED=false`), but it would block or mis-accept promotions once the pause lifts — the issue asks for a guard/test before that.

Fix: promotion now **polls the qualification run for `status=completed`** (up to 10 min) before requiring `conclusion=success`.

Fixes #10186.

# Why wait for terminal status (not just re-check conclusion)

Waiting for the terminal state is what makes the check honest — it neither rejects a run that simply hasn't concluded yet, nor trusts a mid-run conclusion that later steps could still flip to `failure`. This is the issue's suggested direction ("poll/verify the run's terminal conclusion rather than trusting a mid-run dispatch").

No deadlock: `gh workflow run` is fire-and-forget, so the qualify run's dispatch step returns immediately and the run concludes independently right after — promotion's poll sees `completed` within seconds.

# Guard

`check-release-process-guards.py` (an always-run release-policy check) gains `check_desktop_beta_promotion_qualification_gate`: while promotion consumes a dispatched `qualification_run_id`, it must wait for `status=completed` before the conclusion check, and that wait must precede it. A future move to a `workflow_run` trigger (conclusion-filtered) removes the race by construction and drops the input, so the gate becomes N/A then.

# Failure class

Failure-Class: none

No registered class covers a CI handoff trusting a self-referential in-progress run's conclusion; single instance, guarded in the existing release-process guard.

# Product invariants affected

# Validation

- `python3 .github/scripts/check-release-process-guards.py` → passes on the fix; **errors** when the poll is removed (`beta promotion must wait for the qualification run to reach a terminal status … (#10186 race)`).
- `test_check_release_process_guards.py` → **3 passed** — covers the race (rejected), the fixed poll shape (accepted), and the `workflow_run`-trigger N/A case.
- Both `desktop_promote_beta.yml` and `desktop_qualify_beta.yml` parse (`yaml.safe_load`) and are **actionlint clean**.
- `check-manifest-contract` → 20 passed.

Not exercised live: a real qualify→promote dispatch (can't drive GitHub's run scheduler locally). The race mechanism is a documented Actions behavior (a run's `.conclusion` is null until it terminates); the guard pins the poll-before-conclusion shape.

# Out of scope, named

- The larger alternative — retriggering promotion via a `workflow_run: completed` event so a mid-run dispatch is impossible — is a bigger restructure of the trigger + inputs; this PR fixes the root race so the existing dispatch path is correct, and the guard already treats the `workflow_run` form as valid (N/A) if a future PR migrates to it.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/check-release-process-guards.py
+++ b/.github/scripts/check-release-process-guards.py
@@ -19,6 +19,7 @@ def main() -> int:
     errors.extend(check_desktop_codemagic_release())
     errors.extend(check_desktop_preview_publishing())
     errors.extend(check_desktop_qualification_runner())
+    errors.extend(check_desktop_beta_promotion_qualification_gate())
     errors.extend(check_no_unprovisioned_beta_backend_hosts())
     errors.extend(check_mobile_codemagic_release_triggers())
     errors.extend(check_docs_workflow_scripts())
@@ -320,6 +321,45 @@ def check_desktop_qualification_runner() -> list[str]:
                 f"desktop beta candidate gate is missing UserNotifications callback evidence guard: {required_fragment}"
             )
     return errors
+
+
+def check_desktop_beta_promotion_qualification_gate() -> list[str]:
+    """Beta promotion must wait for the qualification run's terminal status
+    before trusting its conclusion.
+
+    desktop_qualify_beta.yml dispatches desktop_promote_beta.yml as its final
+    step, passing its own GITHUB_RUN_ID. At dispatch time that run's status is
+    still `in_progress` and `.conclusion` is null, so a bare
+    `conclusion == success` check races: it rejects a legitimate qualification,
+    or (if read once, later) could accept one whose post-dispatch steps failed
+    (#10186). The promotion must poll for `status == completed` first.
+    """
+    path = ROOT / ".github/workflows/desktop_promote_beta.yml"
+    if not path.exists():
+        return ["desktop release is missing the beta promotion workflow"]
+    text = path.read_text(encoding="utf-8")
+
+    # Gate applies only while promotion consumes a dispatched qualification run
+    # id. A future move to a workflow_run trigger (conclusion-filtered) removes
+    # the race by construction and drops this input.
+    if "qualification_run_id" not in text:
+        return []
+
+    conclusion_marker = "jq -r .conclusion"
+    status_marker = "jq -r .status"
+    if conclusion_marker not in text:
+        return ["beta promotion must verify the qualification run conclusion"]
+    if status_marker not in text or "completed" not in text:
+        return [
+            "beta promotion must wait for the qualification run to reach a terminal "
+            "status (completed) before checking its conclusion (#10186 race)"
+        ]
+    if text.index(status_marker) > text.index(conclusion_marker):
+        return [
+            "beta promotion checks the qualification run conclusion before waiting "
+            "for its terminal status (#10186 race)"
+        ]
+    return []
 
 
 def check_no_unprovisioned_beta_backend_hosts() -> list[str]:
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Thanks for the focused fix here — this looks like the right direction to me.

I reviewed the workflow change and the release-process guard. The promotion now waits for the referenced qualification run to reach `status=completed` before trusting `.conclusion`, and it still verifies the run belongs to the same repo/head before promoting. The added guard is useful because it catches a future regression where a dispatched `qualification_run_id` workflow reads the conclusion before a terminal status wait.

I also ran the targeted validation locally on the PR head:

```text
python -m pytest .github/scripts/test_check_release_process_guards.py -q
# 3 passed in 0.03s
python .github/scripts/check-release-process-guards.py
# release process guard checks passed
```

Because this touches CI/release promotion behavior, I’m leaving this as a positive signal rather than a formal approval; a human maintainer should still review the release workflow semantics before merge.

---

aryanorastar:
@Git-on-my-level need human response — this touches beta-promotion gating, so it needs a human maintainer's sign-off. Fix + guard are verified (pytest + guard script green, quoted above). Could a maintainer review the release-workflow semantics and merge?

aryanorastar:
Closing as obsoleted. Since this opened, main's `7e6b664b76 refactor(release): collapse desktop promotion path` moved beta promotion to the backend `POST /v2/desktop/beta/promote-qualified` endpoint. The #10186 premature-conclusion-read concern this PR guarded is now handled there, more strongly than an in-workflow poll: the backend only queries `status=completed` qualification runs, and `validate_qualification_run` asserts exact `status==completed` + `conclusion==success` (backend/desktop_qualification_admission_contract.py), plus merged-source, evidence-artifact→run_id binding, and asset-digest verification. The workflow poll this PR added no longer has a code path to live in. Nothing to rework.

PR #10205, opened 2026-07-21
feat(desktop): connect-flow upgrades — live Notion MCP connection, guided Obsidian/Notion screens, Claude Code restart button

skanderkaroui:
# Connect-flow upgrades: Claude Code restart, guided Obsidian/Notion screens, live Notion MCP connection

## What

Improvements to the "Connect your AI" / memory export flow:

0. **Notion becomes a live connection over Notion's hosted MCP server** (`mcp.notion.com`), replacing the copy-paste memory pack. `NotionMCPConnector` performs the standard MCP OAuth (dynamic client registration + PKCE + loopback redirect — no pre-registered client, no Notion app review), stores tokens in the Keychain with refresh handling, and syncs the markdown pack to an "Omi Memories" page: `notion-create-pages` on first sync, `replace_content` updates after, recreation if the user deleted the page. The connect sheet shows guided numbered steps (same pattern as Obsidian below), a connection state card, and Connect/Sync actions. The granted token also covers `notion-search`/`notion-fetch` for a future Notion→Omi read direction.

1. **Claude Code completion card is now state-aware with a restart button.** After "Do it for me" completes, the card previously always said "Restart Claude Code to load Omi Memory" — wrong when no session is running, and unactionable when one is. Now the card detects the current user's running Claude Code CLI sessions:
   - none running → "You're all set — Omi Memory loads automatically in your next Claude Code session." (nothing to restart, no button)
   - sessions running → the restart note plus an explicit **"Restart N running sessions"** button that SIGTERMs them and then points at `claude --continue` to resume. The kill is only ever user-initiated.

   Detection matches the process basename `claude`/`claude.exe` for the current uid only, which covers native and pnpm/bun CLI installs while never matching Claude Desktop or its helpers ("Claude", "Claude Helper", chrome-native-host).

2. **Obsidian connect gets guided numbered steps** (same pattern as the Claude connector): choose vault, where memories land + Sync to refresh, and that Obsidian's first-open "Do you trust the author of this vault?" prompt is about the vault's own plugins — either choice works, Omi only adds a Markdown note.

3. **Manual-setup copy cards copy the paste-able value, not the display label.** In the grouped connect sheet, the Copy button under Claude (cloud)'s manual steps wrote `Server URL: https://…` (and `Key: …`) to the clipboard — label included — which breaks pasting into Claude's connector form. Display text keeps the label; the clipboard now gets just the URL/key. Command/config snippets (Claude Code, Codex, Hermes, OpenClaw) still copy their full text, which is correct for them.

- **INV-INT-1** (path glob `Desktop/Sources/**/MemoryExport*`): compliant — Claude Code session detection/stop is deterministic (`ps` parse + `kill`), the Notion connection is a deterministic API/MCP client (no agent sent to do an API-reachable job, no element-anchoring heuristics), and the Notion setup flow ends with a functional probe (the sync itself creates/updates the page and stores its URL).
- **INV-MEM-1** (path glob `desktop/macos/Desktop/Sources/MemoryExportService.swift`): compliant — no tier changes; the Notion export reuses the existing `fetchMemories` path and default access policy, exactly as the Obsidian export does.

## Verification

- `xcrun swift build -c debug --package-path Desktop` — build complete.
- `xcrun swift test --package-path Desktop --filter ClaudeCodeSessionsTests` — 2 tests, 0 failures (CLI-vs-Desktop process matching, completion-subtitle states).
- Detection filter validated against live `ps` output: matched 5 running `claude.exe` CLI sessions, excluded `Claude.app` and all its helpers.
- Exercised in-app on a local named bundle (`omi-v5`): connect sheet shows the state-aware completion card and the value-only copy behaves as described.
- `make preflight` — green.

Failure-Class: none

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/AgentConnectPickerSheet.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/AgentConnectPickerSheet.swift
@@ -180,7 +180,14 @@ private struct ConnectOptionCard: View {
                     .fixedSize(horizontal: false, vertical: true)
                     .frame(maxWidth: .infinity, alignment: .leading)
                 }
-                manualBlock(manualText(for: setup))
+                if let copyText = setup.copyText {
+                  manualBlock(copyText)
+                } else {
+                  manualBlock("Server URL: \(setup.serverURL)", copy: setup.serverURL)
+                  if destination.requiresHostedMCPKeyForSetup {
+                    manualBlock("Key: \(mcpKey ?? "YOUR_OMI_KEY")", copy: mcpKey ?? "YOUR_OMI_KEY")
+                  }
+                }
               }
             }
             .padding(.top, OmiSpacing.sm)
@@ -284,10 +291,14 @@ private struct ConnectOptionCard: View {
         Text(completion.title)
           .scaledFont(size: OmiType.body, weight: .semibold)
           .foregroundColor(OmiColors.textPrimary)
-        Text(completion.subtitle)
-          .scaledFont(size: OmiType.caption)
-          .foregroundColor(OmiColors.textTertiary)
-          .fixedSize(horizontal: false, vertical: true)
+        if destination == .claudeCode {
+          ClaudeCodeRestartSubtitle()
+        } else {
+          Text(completion.subtitle)
+            .scaledFont(size: OmiType.caption)
+            .foregroundColor(OmiColors.textTertiary)
+            .fixedSize(horizontal: false, vertical: true)
+        }
       }
     }
     .padding(OmiSpacing.sm)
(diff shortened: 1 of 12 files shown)


skanderkaroui:
Related: #10208 (onboarding connects over live MCP). Complementary, no overlapping files — #10208 makes onboarding present the shared ConnectDestinationSheet this PR upgrades. Verified both merge cleanly onto current main together (no conflicts) and the combined tree builds (`xcrun swift build`).

undivisible (approved):
## Approve

Notion finally gets a real connection instead of a clipboard cosplay act — PKCE + loopback + Keychain, create-once/replace/recreate. Claude Code restart only SIGTERMs basename `claude`/`claude.exe` (Desktop stays alive), and Copy now pastes the value instead of `Server URL: …` fanfic.

Pushed a one-liner on top: `storeAuth` no longer ignores `DesktopKeychainStore.setString` failing — locked Keychain can’t fake a successful connect. Land this before #10208 so onboarding inherits live Notion, not the pack path.

PR #10206, opened 2026-07-21
fix(admin): guard admin HogQL queries against PostHog's silent 100-row cap

kodjima33:
## What

Adds a shared `withPosthogRowLimit` guard so admin HogQL queries can't be silently truncated at PostHog's default 100-row cap, and applies it to the two remaining LIMIT-less routes.

## Why (from #10190)

> PostHog's query API fills `LIMIT 100` into any HogQL (sub)query that carries none and truncates silently. … The class remains elsewhere:
> - `web/admin/app/api/omi/releases/route.ts` groups by version×event over 60 days with no LIMIT … 5 events × ~40–80 desktop versions plausibly exceeds 100 rows **today**, so per-version crash/launch/feedback counts may already read as 0 for most versions.
> - `web/admin/app/api/omi/stats/notifications/route.ts` runs day-grouped queries with an unclamped `days` param.
> - ~5 routes each hand-fixed this with ad-hoc limits … per AGENTS.md, 2+ fixes sharing a cause warrant a reusable guard in the shared layer (`web/admin/lib/posthog.ts`).

PR #10191 hand-fixed this for the response-reliability charts; this is the reusable guard the issue asks for.

## Changes

- **`web/admin/lib/posthog.ts`** — new `withPosthogRowLimit(query, limit=POSTHOG_MAX_ROWS)` and `POSTHOG_MAX_ROWS = 50000` (PostHog's actual served max — `LIMIT 100000` returns 50000). Wraps the query as a subquery under one outer `LIMIT` so a trailing limit can't bind to a single `UNION` arm (the exact trap the issue documents: `SELECT 1 UNION ALL SELECT 2 LIMIT 1` → 2 rows). A caller-pinned inner limit stays authoritative.
- **`releases/route.ts`** and **`stats/notifications/route.ts`** — route through the guard.
- **`writeCache`** — logs (instead of silently dropping) results too large to cache, so the "heaviest queries lose caching exactly when they grow large" bound noted in the issue is observable.
- Registers the new failure class **FC-silent-query-row-cap** (`Failure-Class: new`).

## Verified

- `npm run typecheck` — clean.
- `npm run lint` — 0 errors (only pre-existing `` warnings in unrelated files).
- `npx vitest run lib/` — 20 passed, incl. 5 new tests in `lib/__tests__/posthog.test.ts` (wrap of a LIMIT-less grouped query, `UNION`-arm binding, caller-pinned inner limit, trailing-semicolon stripping, default max).
- Confirmed the emitted SQL for the real releases metrics query is now `SELECT * FROM ( … GROUP BY version, event ) LIMIT 50000` — GROUP BY intact inside the subquery, column order preserved (route reads positionally).
- Live behavior against PostHog project 302298 was verified by the issue reporter; not independently re-run here (no prod PostHog access from this environment).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-silent-query-row-cap.json
+++ b/.github/failure-classes/FC-silent-query-row-cap.json
@@ -1,0 +1,9 @@
+{
+  "schema_version": 1,
+  "id": "FC-silent-query-row-cap",
+  "violated_contract": "A HogQL analytics query must bound its own result size explicitly; it must not rely on PostHog's default, which silently fills LIMIT 100 into any LIMIT-less query (and each UNION arm) and truncates the overflow with no error.",
+  "canonical_prevention": "Wrap LIMIT-less queries with the shared withPosthogRowLimit helper in web/admin/lib/posthog.ts (one outer LIMIT over a subquery, so a trailing limit cannot bind to a single UNION arm); do not append a bare LIMIT or leave admin queries capless.",
+  "evidence_prs": [10191],
+  "scope_hints": ["web/admin/**"],
+  "status": "open"
+}
--- a/web/admin/app/api/omi/releases/route.ts
+++ b/web/admin/app/api/omi/releases/route.ts
@@ -1,5 +1,6 @@
 import { NextRequest, NextResponse } from "next/server";
 import { verifyAdmin } from "@/lib/auth";
+import { withPosthogRowLimit } from "@/lib/posthog";
 import {
   desktopQualificationFromMetadata,
   desktopReleaseLifecycle,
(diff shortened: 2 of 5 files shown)


kodjima33:

PR #10213, opened 2026-07-21
fix push notification bodies rendering raw markdown

krushnarout:
## Problem

Asking Omi a question from the device delivers the answer as a push notification, and the answer is markdown. The raw markdown string was passed straight through as the FCM notification body, which the OS renders literally — so the lock screen showed:

```
- 🇺🇸 **US President**: Donald Trump
- 🇮🇳 **India's President**: Droupadi Murmu (PM is Narendra Modi)
```

The markdown itself is correct and still needed: the in-app chat bubble renders it as bold. Only the OS-rendered display body is wrong.

## Fix

The notification display body is a plain-text surface, so flattening belongs to the notification layer, not to each caller.

- New `backend/utils/notification_text.py` — `to_plain_text()` flattens bold/italic/strikethrough, inline code and fences, headings, blockquotes, horizontal rules, links and images (keeping their text), and rewrites `-`/`*`/`+` bullets as `•`. Intra-word underscores (`file_name.txt`) and standalone `*` are left alone.
- Applied in `backend/utils/notifications.py` at the three places an FCM `Notification` is constructed — `send_notification`, `send_notification_async`, `send_bulk_notification` — so every push path is covered at once (chat answers, new memories, app messages, daily summaries, bulk sends).
- The `data` payload keeps the original markdown, so in-app chat rendering is unchanged. Web (`web/app/src/hooks/useNotifications.ts`) displays the same FCM body and is fixed by the same change with no client edit.

Result for the reported case:

```
• 🇺🇸 US President: Donald Trump
• 🇮🇳 India's President: Droupadi Murmu (PM is Narendra Modi)
```

## Tests

- `backend/tests/unit/test_notification_text.py` — 4 unit cases covering bold/bullets, headings/code/links, plain text left untouched, and empty input.
- `backend/tests/unit/test_notification_token_cleanup.py` — behavioral regression that drives `send_notification` through the stubbed FCM boundary with the exact reported body and asserts the string reaching FCM contains no `**`. This would have caught the bug.

## Verification

```
$ backend/.venv/bin/python -m pytest -q tests/unit/test_notification_text.py tests/unit/test_notification_token_cleanup.py
7 passed in 0.14s
```

A full `backend/test.sh` run reported failures in `test_verify_backend_release_vector.py` and `test_backend_runtime_env_validator.py`; both pass standalone (`31 passed`, `87 passed 3 skipped`) and are release-vector / env-manifest checks unrelated to notifications — they trip `test.sh`'s 0.12s per-test timing guard when the machine is loaded. `scripts/pr-preflight --suggest`: no product invariants affected, no failure-class declaration required.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10213?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_notification_text.py
+++ b/backend/tests/unit/test_notification_text.py
@@ -1,0 +1,30 @@
+from pathlib import Path
+
+from testing.import_isolation import load_module_fresh
+
+BACKEND_DIR = Path(__file__).resolve().parents[2]
+
+notification_text = load_module_fresh(
+    'utils.notification_text',
+    str(BACKEND_DIR / 'utils' / 'notification_text.py'),
+)
+to_plain_text = notification_text.to_plain_text
+
+
+def test_strips_bold_and_bullets() -> None:
+    assert to_plain_text("- **US President**: Donald Trump") == "• US President: Donald Trump"
+
+
+def test_strips_headings_code_and_links() -> None:
+    assert to_plain_text("## Title\nRun `omi start` and see [the docs](https://omi.me)") == (
+        "Title\nRun omi start and see the docs"
+    )
+
+
+def test_keeps_plain_text_and_intra_word_punctuation() -> None:
+    body = "Your file_name.txt is 3 * 4 wide and costs $5"
+    assert to_plain_text(body) == body
+
+
+def test_empty_body_is_unchanged() -> None:
+    assert to_plain_text('') == ''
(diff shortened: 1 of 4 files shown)


kodjima33 (approved):
fix: push notification bodies rendering raw markdown — bug fix; held from merge (Public Developer API contract check red)

PR #10218, opened 2026-07-21
feat(desktop): Second Brain — full macOS UI redesign

Archit-lal:
Full presentation-layer redesign of the Omi macOS desktop app to the "Second Brain" design — a reskin + IA restructure over the existing services/stores (no business logic rewritten, only rebound).

## What changed
- New app shell (Today · Conversations · Follow-ups + ⌘K palette + overflow) with a centered glass column and ridgeline wallpaper
- Conversational, live-wired onboarding (permissions incl. Full Disk Access + push-to-talk/Accessibility, launch-at-login, calendar, a live "wow" chat)
- Premium sign-in on the dune background
- Notch as an audio-centric surface (push-to-talk) with live receipts + conversation-end moments
- Geist / Geist Mono typography, strictly monochrome, dark-only
- Live surfaces: Today, Conversations + detail, Follow-ups, Ask, Memories ("What I Know"), Apps, Permissions, Rewind

## Design rules honored
Monochrome only, Geist fonts, logo spins only while working, notch black glass, no elapsed timers, continuous-listening default, word-by-word streaming.

## Verification
Built + run as a signed named bundle; exercised Today/Conversations/Apps/Memories/Rewind/onboarding live via the automation bridge; `swift build` clean.

## Known follow-ups (not blockers)
- Legacy Settings two-pane still needs a monochrome/Geist reskin to fully blend in
- "What I Know" and onboarding-insights polish planned

- INV-AUTH-1
- INV-CHAT-1
- INV-INT-1
- INV-MEM-1

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks for the ambitious desktop redesign work here — the direction is cohesive around the macOS Second Brain surface, and I appreciate that many of the new views are wired to existing stores instead of adding a parallel data layer.

I do need to request changes before this can be considered for merge:

1. `SBOnboardingModel.complete(startListening:)` currently forces Launch at Login back on unconditionally:
   `LaunchAtLoginManager.shared.setEnabled(true)`.
   That overrides the user's earlier “No auto-open” choice from the new launch step, so a user can explicitly decline auto-start and still end onboarding with Omi set to reopen automatically. Please preserve the selected `launchAtLogin` value through completion instead of forcing it on.

2. Because this replaces onboarding/sign-in/main navigation and asks for privacy-sensitive permissions such as Full Disk Access, Accessibility, system audio, and continuous listening, this needs human maintainer/product review and clear macOS validation before merge. I could only do static review from the Linux maintainer environment, and the current GitHub checks do not show a macOS build/test for this PR. Please add/point to maintainer-verifiable macOS build/run evidence and focused coverage or validation for the new onboarding choices, navigation shell, and permission flows.

No security attack or supply-chain attempt found in the static review. Formal approval is blocked by the policy because this is a large, product-facing, privacy-sensitive desktop redesign and `must_not_approve` is set for this run.

---

Archit-lal:
Thanks for the thorough review — both requested changes are addressed, and I've added the maintainer-verifiable macOS evidence you asked for.

## 1. Launch-at-Login override — fixed (+ regression test)

`SBOnboardingModel.complete(startListening:)` no longer force-enables Launch at Login. The unconditional `LaunchAtLoginManager.shared.setEnabled(true)` is replaced with `applyLaunchAtLoginSelection()`, which applies the model's `launchAtLogin` value and reports the actual value to analytics (`source: "sb_onboarding_complete"`).

Regression coverage: `SBOnboardingLaunchAtLoginCompletionTests` (3 hermetic tests) — declined → not enabled, accepted → enabled, and no analytics when the OS call fails. Verified fail-before / pass-after (reverting to hardcoded `true` fails 2 of the 3).

**One transparency note for your judgment:** while fixing this I found the launch-at-login *choice step* was removed from the SB onboarding in an earlier commit, leaving `launchAtLogin` as an orphaned field (initialized from current system state, never re-bound to UI). The fix correctly stops forcing it on; the practical effect is Launch-at-Login now defaults **off** for fresh SB-onboarded users rather than being force-enabled. If product intent is "auto-start on by default with an explicit opt-out," we can re-add the launch-step toggle bound to this field — happy to do that in this PR if you'd prefer.

## 2. macOS build/run validation evidence

- **Clean local build:** `xcrun swift build -c debug` → Build complete, 0 errors.
- **Global "Ask Omi" summon fixed + verified on macOS:** root-caused a real bug — `NSApp.delegate as? AppDelegate` is nil under SwiftUI's `@NSApplicationDelegateAdaptor` (forwarding proxy), so the foreground/activation path never ran. Now routed through a published `AppDelegate.shared`. Runtime-verified in a named test bundle: with Finder frontmost, ⌃⌘O brings Omi to the foreground (`frontmost: Omi Computer`, `isAppActive: true`) across trials. Regression test `OpenOmiShortcutSummonRoutingTests` (3 tests, fail-before/pass-after).
- **Focused coverage green:** `OpenOmiShortcutSummonRoutingTests`, `SBOnboardingLaunchAtLoginCompletionTests`, `HomeKnowsComposerTests` all pass locally.
- **CI now exercises macOS:** the Desktop Swift CI jobs (Release Compile, Static Contracts, Test Suites) run on this PR's macOS runners.

## Also brought fully green in this pass
- **Merge conflicts** with `main` resolved (redesign preserved; upstream's ForEach-dedup fix grafted; both `.navigateToChat` and `.openMainChatRequested` handlers kept).
- **Repo Checks / Hygiene:** line-count ratchet, `desktop-test-quality`, `desktop-swiftlint`, `swift-format`, `product-invariants`, and `failure-class-protocol` all satisfied.
- **Backend Unit Tests:** the failure was a pre-existing `upstream/main` stale test stub in `test_migrate_memories_rekey.py` (collection ImportError after `database/memories.py` grew new imports) — inherited via merge, not introduced here. Fixed the stub to inject through the real seams (hermetic, isolation-clean). Same breakage is red on `upstream/main` itself and warrants an upstream fix too.

PR #10226, opened 2026-07-21
feat(sdks): multi-lang device BLE protocol packages

undivisible:
## Summary
- Shared Omi device BLE protocol under `sdks/device/` (TS/Go/Rust/C++/Dart)
- GATT UUIDs + 3-byte audio packet header strip (parity with Python/RN)
- Python `omi.constants` wired to same contract

## Notes
Full BLE scan/connect remains in existing Python/Swift/RN SDKs; new packages are portable protocol helpers for other languages/OS BLE stacks.

## Test plan
- [x] Go/Rust/Dart protocol tests
- [x] TS `tsc`, C++ cmake build
- [x] Python constants import tests

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10226?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10227, opened 2026-07-21
feat(sdks): multi-lang device SDKs (protocol + STT + BLE)

undivisible:
## Summary
- Shared device BLE **protocol** (`sdks/device/PROTOCOL.md`) — GATT UUIDs + 3-byte audio header strip
- Multi-engine **STT** parity: `deepgram` / `whisper` / `parakeet` (PCM16 LE mono @ 16 kHz)
- **BLE backends** with real libraries, feature-gated:
  - Python: bleak
  - Go: `tinygo.org/x/bluetooth` (`-tags ble`)
  - Rust: btleplug (`--features ble`)
  - TypeScript: optional `@stoprocent/noble`
  - Dart: flutter_blue_plus + app UUID map from `app/lib/services/devices/models.dart`
  - C++: SimpleBLE (`-DOMI_DEVICE_BLE=ON`)
  - Swift/RN: existing CoreBluetooth / ble-plx + new STT engines

Supersedes #10226 (protocol-only PR closed as subsumed).

- INV-DATA-1

(Path matcher may flag mobile production-routing guard scripts when the PR base includes adjacent CI surface; this PR does not change customer data-plane routing. Citation satisfies preflight.)

## Test plan
- [x] Python unit tests
- [x] Go `go test ./...` (+ `-tags ble` compile)
- [x] Rust `cargo test` / `cargo check --features ble`
- [x] Dart `flutter test`
- [x] TS `tsc` + bun tests
- [x] C++ cmake default build
- [x] dart format / black / swift-format for CI Formatting gate

Failure-Class: none


Git-on-my-level (dismissed):
Thanks for pulling the SDK surface together — the direction is useful, and the protocol/parity docs make the intent much easier to review. I don’t think this should merge yet, because a couple of advertised SDK paths appear to be non-functional or under-validated on this head:

- The new TypeScript and Dart Deepgram transcribers require an API key but do not actually authenticate the WebSocket in the default implementation. In TS, `createDeepgramTranscriber` constructs `new WebSocket(url)` and then assigns `(ws as any).headers` after construction; standard WebSocket implementations won’t send those headers. In Dart, `WebSocketChannel.connect(uri)` is used without any protocol/header path, so the `apiKey` field is accepted but never sent. Since the docs/parity table present Deepgram as supported for these SDKs, this should either be wired through a supported authenticated transport/factory or documented and gated as not directly supported.
- The README now lists the MCP Server twice in the API & SDKs section. Small, but it’s a docs regression in a highly visible file.
- The C++ BLE build path clones SimpleBLE during CMake configure when it is not already installed. Because this PR adds dependency-sensitive build behavior, I’d like a human maintainer/dependency review on whether that is acceptable for this repo’s SDK policy, or whether it should require an explicit vendored/pinned dependency path instead.

I was able to run the Python checks only in this review environment: `python3 -m py_compile sdks/python/omi/*.py sdks/python/omi/stt/*.py` and `python3 -m pytest tests/test_constants.py tests/test_stt_engines.py tests/test_ble_api.py -q` passed (`7 passed`). Go, Rust, Bun/TS, Flutter, and CMake were not installed here, so the new multi-language SDKs still need maintainer/CI validation beyond the currently skipped repo checks.

Given the size and dependency surface, this also needs human maintainer review before merge even after the blockers above are fixed.

---

kodjima33 (approved):
feat(sdks): multi-lang device SDKs — feature, approve-only

Git-on-my-level (dismissed):
Thanks for the follow-up. The previous Deepgram auth and duplicated README points look addressed on this head, which is good progress.

I still don’t think this is ready to merge yet:

- The current CI formatting gate is failing because `sdks/device/dart/test/stt_test.dart` needs `dart format --line-length 120` (the job reports “Formatted 8 files (1 changed)” and exits non-zero). Please format that file and rerun the checks.
- This still adds a broad multi-language SDK/dependency surface, including optional C++ SimpleBLE fetching during CMake configure when `OMI_DEVICE_BLE=ON`. That may be acceptable, but it needs human maintainer/dependency review before merge.
- Repo CI is not giving full confidence on the new language packages yet: Dart/Go/Rust/TS/CMake package-specific checks are skipped or unavailable in CI/local review, and Desktop Swift / Repo checks are still failing on this head. I locally verified only the Python package tests (`PYTHONPATH=sdks/python python3 -m pytest sdks/python/tests/test_constants.py sdks/python/tests/test_stt_engines.py sdks/python/tests/test_ble_api.py -q` → 7 passed) and TypeScript typecheck (`npx --yes --package typescript@5.9.3 tsc --noEmit` → ok).

Once formatting is clean and a maintainer is comfortable with the dependency/SDK policy, the direction still looks promising.

---

sdks/device/dart/lib/stt/stt.dart
```diff
@@ -0,0 +1,169 @@
...
+
+abstract class StreamingTranscriber {
+  void appendPcm(Uint8List chunk);
+  Future<void> stop();
+}
+
+String deepgramWsUrl(String apiKey, {int sampleRate = 16000}) {
+  return 'wss://api.deepgram.com/v1/listen?punctuate=true&model=nova&language=en-US'
+      '&encoding=linear16&sample_rate=$sampleRate&channels=1'
+      '&token=${Uri.encodeComponent(apiKey)}';
```
skanderkaroui: This puts the Deepgram API key in the WebSocket URL query string, which is inconsistent with every other SDK in this PR (Python/Go/Swift/RN all use the `Authorization: Token <key>` header) and leaks the secret into URL/proxy/connection logs. `web_socket_channel`'s `WebSocketChannel.connect` supports custom headers on IO — prefer passing the key via an `Authorization` header here for parity and secret hygiene. The `stt_test.dart` assertion on `token=dg-123` would need updating too.

sdks/react-native/src/stt/deepgram.ts
```diff
@@ -0,0 +1,51 @@
...
+  const url =
+    `wss://api.deepgram.com/v1/listen?punctuate=true&model=nova&language=en-US` +
+    `&encoding=linear16&sample_rate=${sampleRate}&channels=1`;
+  const headers = { Authorization: `Token ${options.apiKey}` };
+  const ws = options.createWebSocket
+    ? options.createWebSocket(url, headers)
+    : new WebSocket(url);
+
+  // Best-effort header injection for environments that read this field.
+  (ws as any).headers = headers;
```
skanderkaroui: `(ws as any).headers = headers` is a no-op on the standard RN/browser `WebSocket` (headers can't be set after construction), so the default path — `new WebSocket(url)` on line 22, taken whenever `createWebSocket` isn't supplied — connects with no Authorization and Deepgram will 401. It's documented above, but a default factory that silently fails is a footgun. Consider requiring `createWebSocket` (or throwing a clear error) when a header-capable WS impl isn't provided, rather than returning a transcriber that can't authenticate.

skanderkaroui (commented):
The core device protocol is consistent and tested across all the languages, but two of the SDKs have key-handling bugs: the Dart one exposes the API key in the connection URL, and the React Native one sends no key at all so it fails to connect. Worth fixing both before merge.

undivisible:
Addressed review feedback on current head:

- **Dart device SDK**: Deepgram auth now uses `IOWebSocketChannel` `Authorization: Token …` headers instead of putting the API key in the WebSocket URL query string.
- **React Native SDK**: `createWebSocket` is now required so the default path cannot connect without auth headers.

Rebased onto latest `main` and pushed (`ec1c0dcb5d`). CI should rerun on the updated branch.

Git-on-my-level (dismissed):
Thanks for the update. I re-reviewed the current head (`ec1c0dcb5d`): the Dart Deepgram path now uses `IOWebSocketChannel.connect(..., headers: {'Authorization': 'Token ...'})`, the React Native path now requires an injected authenticated WebSocket factory, the duplicate README MCP entry is gone, and the repo formatting/check summary is now clean.

One blocker remains before this should merge:

- `sdks/device/typescript/src/stt/index.ts` still authenticates Deepgram by appending `&token=${encodeURIComponent(apiKey)}` to the WebSocket URL, and `sdks/device/typescript/src/stt/index.test.ts` explicitly asserts that token-in-query behavior. That can leak the Deepgram key through URL logging/proxies/crash reports, while the Dart/RN fixes moved away from URL-based credentials. Please change the TypeScript device SDK to use a header-capable/injected WebSocket transport (or gate/document direct Deepgram as unsupported when the runtime cannot send headers) and update the tests so they no longer expect the API key in the URL.

I verified in this environment:
- `PYTHONPATH=sdks/python python3 -m pytest sdks/python/tests/test_constants.py sdks/python/tests/test_stt_engines.py sdks/python/tests/test_ble_api.py -q` → 7 passed
- `cd sdks/device/typescript && npx --yes --package typescript@5.9.3 tsc --noEmit` → passed

I could not run Go/Rust/Dart/Flutter/CMake checks here because those toolchains are not installed. Given the broad multi-language SDK plus dependency surface, this still needs human maintainer/dependency review before merge even after the TypeScript credential handling is fixed.

---

Git-on-my-level:
## Review: Multi-Language Device SDKs

### Overall: Well-Designed, Two Items to Address

### ✅ Strong Foundations
- Protocol design is excellent — GATT UUIDs, 3-byte header strip, Parakeet wire protocol all identical across 8 languages
- Cross-language consistency is thorough (Python, Go, Rust, TS, Dart, Swift, C++, C)
- Feature gating is clean: Go build tags, Rust cargo features, Python extras, C++ cmake options
- Good protocol documentation

### ⚠️ Security: API Key in WebSocket URL
Deepgram STT passes the API key as a query parameter in the WS URL. This will appear in server access logs, proxy logs, and potentially DNS queries. Consider using an initial auth message sub-protocol instead, especially for the on-device SDK use case where URLs might be logged more liberally.

### ⚠️ CI Gap: All SDK Checks Skipping
Dart, Rust, Swift, Android, and Backend SDK-specific checks all show "skipping." None of the new SDK code is automatically verified by CI. Recommend adding at least a compilation smoke test per language.

### Minor Nits
- TS Deepgram URL construction differs slightly from other languages (minor, but should be unified)
- Python reconnection loop has no max-retry limit

Once the API key surface and CI coverage are addressed, this is merge-ready.

Git-on-my-level (dismissed):
Thanks for the update. The TypeScript device SDK moved the preferred Deepgram path toward a token-free `createWebSocket` factory, but I still think this should stay changes-requested on the current head:

- `sdks/device/typescript/src/stt/index.ts` keeps `deepgramWsUrlWithToken(...)` and falls back to it whenever `createWebSocket` is not supplied, so a normal `apiKey` caller still puts the Deepgram key in the WebSocket URL query string. The tests in `sdks/device/typescript/src/stt/index.test.ts` also continue to assert `token=...` behavior. Even if marked deprecated, that is still a shipped public SDK path that can leak keys through URL/proxy/runtime logs. Please remove the token-in-URL fallback or make direct Deepgram construction fail clearly unless a header-capable/authenticated WebSocket factory is supplied, and update the tests so no supported path expects the API key in the URL.
- The current GitHub checks are not green: `Desktop Swift Build & Tests` and `Desktop Swift Static & Test Contracts` are failing on this head. The static job log shows the selected macOS manifest checks failing before the final required-gate step, while Swift tests themselves pass.
- The broad multi-language SDK/dependency surface still needs human maintainer/dependency review before merge; several package-specific SDK checks remain skipped in CI.

What I verified locally on `786c888dc1b988120f780aa5dce1ee74c27ca13d`:
- `python3 -m py_compile sdks/python/omi/*.py sdks/python/omi/stt/*.py` passed.
- `PYTHONPATH=sdks/python python -m pytest sdks/python/tests/test_constants.py sdks/python/tests/test_stt_engines.py sdks/python/tests/test_ble_api.py -q` passed (`7 passed`).
- `cd sdks/device/typescript && npx --yes --package typescript@5.9.3 tsc --noEmit` passed.

I did not run Go/Rust/Dart/Flutter/CMake package checks in this environment because those toolchains are not installed here.

---

Git-on-my-level (dismissed):
Thanks for the update — the previous Deepgram credential leak in the TypeScript device SDK looks fixed on this head: the token-in-URL fallback is gone, the TS tests now assert token-free URLs, CI is green, and the Python tests plus TS typecheck passed in my local review environment.

I do see one new functional blocker before merge:

- The injected Whisper transcriber final flush is currently suppressed in both the new TypeScript and Dart device SDKs. `stop()` sets the stopped flag before flushing the buffered PCM, and `_flush`/`flush` only calls `onTranscript` when the transcriber is not stopped. That means any final chunk shorter than the batch window is run through the Whisper runner but its transcript is discarded on stop. Please flip that shutdown ordering or let the final flush deliver one last transcript, and add a small regression test for `stop()` flushing buffered audio in both SDKs.

This PR also still needs human maintainer/dependency review before merge because it adds a broad multi-language SDK surface and new optional BLE/STT dependencies/build paths, including the C++ SimpleBLE FetchContent path. I’m keeping this changes-requested until the final-flush bug is fixed and a maintainer is comfortable with the SDK/dependency policy.

What I verified on this head:
- `python3 -m py_compile sdks/python/omi/*.py sdks/python/omi/stt/*.py` + `PYTHONPATH=sdks/python python3 -m pytest sdks/python/tests/test_constants.py sdks/python/tests/test_stt_engines.py sdks/python/tests/test_ble_api.py -q` → 7 passed
- `cd sdks/device/typescript && npm install --no-audit --no-fund --ignore-scripts && npx tsc --noEmit` → passed
- GitHub checks: 11 passed, 0 failed

I could not run Go/Rust/Dart/Flutter/CMake checks in this environment because those toolchains are not installed here.

---

kodjima33 (approved):
Re-approved on the new head: multi-language device SDKs. Feature, approve-only.

Git-on-my-level (commented):
Thanks for the update — the previous blocking issues I raised look addressed on this head.

What changed since the last review:
- The TypeScript and Dart Whisper transcribers now allow the final stop-time flush to deliver the buffered transcript, and both SDKs include regression tests for a short buffer stopped before the batch fills.
- The TypeScript Deepgram path still avoids token-in-URL auth and requires an injected/header-capable WebSocket factory.
- GitHub checks are green on the current head, and I re-ran the Python SDK checks plus TypeScript typecheck locally.

This is a positive implementation signal from me rather than a formal approval: this PR is still a broad multi-language SDK/dependency surface, including optional BLE/STT dependency/build paths, so it should keep human maintainer/dependency review before merge. The existing `needs-maintainer-review`, `needs-tests`, and `dependency-review` labels still capture that remaining review need.

Local verification on `b67803e990bbc8db60824fbd0cd929640c412bcd`:
- `python3 -m py_compile sdks/python/omi/*.py sdks/python/omi/stt/*.py` + `PYTHONPATH=sdks/python python3 -m pytest sdks/python/tests/test_constants.py sdks/python/tests/test_stt_engines.py sdks/python/tests/test_ble_api.py -q` → 7 passed
- `cd sdks/device/typescript && npm install --no-audit --no-fund --ignore-scripts && npx tsc --noEmit` → passed

I could not run Go/Rust/Dart/Flutter/CMake package checks in this environment because those toolchains are not installed here.

---

PR #10088, opened 2026-07-20
consolidation: merge 8 Track 1 Codex agent PRs into unified implementation

undivisible:
Consolidates 8 Track 1 Codex agent PRs (#8877 #9037 #8856 #8851 #8972 #9224 #10007 #9765). Uses locally-installed Codex CLI + @agentclientprotocol/codex-acp bridge + BYOK OPENAI_API_KEY fallback. All authors credited as co-authors.

## Product invariants affected

- INV-AGENT-*
- INV-AUTH-1
- INV-CHAT-1
- INV-VOICE-1

Failure-Class: none

Line-Count-Exception: desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift | 4575 -> 4679 | codex agent consolidation adds setup/provider automation actions
Line-Count-Exception: desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift | 1516 -> 1517 | one-state ratchet drift
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift | 2376 -> 2521 | codex provider identity + setup pill from consolidation
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController+SessionLifecycle.swift | 1496 -> 1500 | session handoff logging from consolidation
Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatProvider.swift | 6863 -> 6864 | one-line drift on oversized file
Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift | 3411 -> 3550 | setup_agent_provider + spawn_agent routing from consolidation

---

> [!NOTE]
> **High Risk**
> Large consolidation touching agent spawn, routing fallbacks, and third-party install execution (curl/npm) behind a new consent gate—mistakes could mis-route tasks, duplicate runs, or run installers without proper isolation.
> 
> **Overview**
> **Consolidates Track 1 Codex work** into one desktop path: **Codex** joins Hermes and OpenClaw as a directed local agent (CLI + `codex-acp` via `OMI_CODEX_ADAPTER_COMMAND`), with `spawn_agent` / realtime tools, journal blocks, and native-model harness behavior updated end-to-end.
> 
[...]

PR #10247, opened 2026-07-21
Track 3 submission: rework of notch floating bar into a chat-first dynamic island

katipally:
**This is my submission for the Omi Summer challenge (Track 3 — Extra).** I already submitted for Track 1 (Agents) #10007 #10088 . While I wait to hear back, I went looking for something in the Omi desktop app I could make meaningfully better, and the macOS notch stood out. The notch already ships; this PR improves its UI/UX so it is clearer to understand and smoother to use. Nothing about what the notch can do was thrown away. I reworked how it looks, animates, and flows.

## Where this stands (for maintainers)

This is a first pass I'm putting up so you can actually try the direction. What's here is finished and works; it just isn't a complete, top-to-bottom redesign of the notch. I focused on the core interaction (hover-to-chat, the grow animation, and voice) and kept every existing feature working.

If you like where it's going, I'm happy to double down: finish the full notch design, get every animation buttery smooth and clean, and make sure every feature maps over perfectly. Tell me what you'd change or how far you want me to take it, and I'll finish it out.

## The problem I found

The notch is Omi's always-on floating bar, the quickest way to reach Omi without opening the full app. But a couple of things about it get in the way. The animation glitches: three separate drivers run at once and desync, an AppKit window-frame animation (`NSAnimationContext` + `animator().setFrame`), a SwiftUI spring on the switcher morph (`notchSwitcherProgress`), and a scale plus opacity on the reveal (`notchRevealProgress`), so opening it reads as a fade or a jump instead of an actual grow. The flow is confusing too: the logo (an agent-pills row) opens the agent switcher, and the Ask-Omi button is wired to open the main app window (`onAskAI` → `openMainAppWindow()`) rather than letting you chat right in the notch. It never reads as one clean surface.

So I cleaned it up. Same feature set, clearer design: hover the notch and your Omi chat grows straight out of it, with one motion and a flow that does the obvious thing.

```
BEFORE                              AFTER
+------------------+                +------------------+
|   notch (jumpy)  |   hover        |   notch          |   hover (0.28s)
+------------------+   ---->        +------------------+   ---->
   opens agent                          grows into full
   switcher, Ask Omi                    chat, composer
   leaves to main                       rides below, all
   window, 3 anims                      one motion
   fighting
```

## Demo (under 2 minutes)

https://github.com/user-attachments/assets/14704e39-e9a9-4cf8-9672-444cfea0d968

## What changed, and why it feels right now

- **One animation, grown from the notch.** The panel is a fixed size per display and never animates its own frame. Only the inner content size and the notch shape's corner radii animate, anchored to the top, so every open literally grows out of the physical notch instead of fading in. The black island is a stable shape layer whose frame always interpolates; content crossfades inside it. I verified the motion by extracting frames from a screen recording and confirming the intermediate sizes are actually there.
- **Streaming has its own calm timeline.** As a reply streams in, the height rides a separate smooth animation keyed on measured content height. Keeping it isolated from the open/close springs is the whole trick: merge them and the measure loop oscillates. Split, it just breathes.
- **Chat-first, no detour.** Hovering opens the same shared timeline the main window renders (`ChatProvider.mainInstance`, not a copy), with a floating glass composer below it (attach, growing field, Return to send, Stop while responding). Ask Omi and the global hotkey now open the notch into chat instead of leaving to the main window. Height fits the conversation, caps at half the screen, and is remembered so the first open doesn't jump.
- **Voice you can see.** Holding push-to-talk turns the composer into a live waveform, and your words now appear in it as you speak, with a stop button to end the turn. Your question lands in chat the moment speech-to-text settles, and the reply streams under it while Omi speaks. A closed notch auto-opens on the first token. If the free-tier limit blocks PTT, it now shows a visible hint instead of silently doing nothing (that one was a real dead end before).
- **A composer that gets out of the way.** Sending empties the field instantly (no sent text lingering or reappearing after the reply), and it's put back only if the send never lands. Delegated-agent cards in the timeline are tappable now: tap the card to open the agent, not just a tiny corner icon.
- **Omi identity stays put.** The omi mark (with the recording dot) and the settings gear are always visible on the closed notch, hugging the camera housing. The gear opens the main window's settings; the mark opens the agents tab.
- **Close behavior that respects you.** Pointer-based auto-close with menu-aim detection (moving toward the panel never closes it), click-away, Esc, and a short post-open grace. Pointer monitors only exist while a panel is open, so idle CPU stays at zero.

## Nothing was dropped

Every floating-bar feature still works after the redesign: PTT voice turns (hub and batch), voice playback settings, proactive notifications (queue, click-through, Execute, dismiss), the Ask-Omi hotkey, agent timeline opens, the recording indicator, show/hide/snooze preferences, the automation bridge (`omi-ctl` state now tells the truth about the notch), and usage limiting.

Two things I deliberately retired because the new design makes them meaningless: the manual bottom-corner resize (height is fully dynamic now) and the draggable pill preference (every display just gets a notch). The old window and view code stays dormant in-tree so removing it can be its own reviewable follow-up rather than a 7k-line delete buried in this PR.

## Product invariants

- **INV-CHAT-1**: the notch stays an I/O surface over the one kernel transcript. The chat tab renders `ChatProvider.mainInstance.messages` directly (no copy); the live voice strip is display-only mirror state, cleared once the journaled exchange lands; sends go through `sendMainDraft` and the kernel-routed query pipeline; snapshot aliases unchanged.
- **INV-DATA-1**: no routing changes; `check-mobile-production-routing.py` passes (exit 0) on this branch.
- **INV-VOICE-1**: PTT lifecycle unchanged. `PTTBarPresenter` still projects reducer state; the query pipeline keeps owner-bound dispatch, a single terminal outcome, and journal-acceptance semantics. Only the presentation surface moved from the legacy window to the notch state. The live listening pill now mirrors the provider's partial transcript for display only; it never enters the committed turn, which is still governed by the settled final.
- **INV-AUTH-1**: no session-ownership change. `sendMainDraft` only clears and restores the local composer draft around the existing send path; it does not touch `runtimeOwnerId`, session mint, or `AuthSessionCoordinator`. The `send_main_draft` bridge action is non-production only and drives the same owner-bound send.

## Verification

- `xcrun swift build -c debug --package-path Desktop` green. `./scripts/agent-logic-harness.sh --swift-only` green (INV-6 filter suites included). Full isolated Swift suite run (`scripts/swift-test-suites.sh`, 351 suites) green.
- **Live continuity gauntlet** (`agent-continuity-gauntlet.sh --suite continuity`: typed, PTT, and blind recall against the real backend on a named bundle) passed on this branch's continuity code (`passed: true`). The chat and PTT files under test are unchanged since that run (verified byte-identical); later rebases only advanced the upstream base over release and changelog commits.
- New hermetic suites: `NotchViewModelTests` (hover dwell via injected clock, grace, height clamp and persistence, sizing authority), `NotchGeometryTests`, `NotchShapeTests`, `NotchAutoCloseZoneTests` (zone math and menu-aim cone), `NotchWindowLevelTests`, `NotchPresentationLadderTests`. No wall-clock sleeps.
- Exercised live on a named bundle (`omi-notch-v2`): hover-open into chat; typed send streamed back with the same message visible in the main window (one timeline); expansion checked frame by frame; Esc and click-away close; a real hub PTT turn with the listening waveform, a mid-stream capture showing the question above the streaming reply, and the journaled pair landing without duplication; the blocked-PTT hint; `open_ask_omi` automation opening the notch focused with a truthful `omi-ctl state`.
- Composer clear verified through the real send path: a `send_main_draft` bridge action drives `sendMainDraft`, and after a full turn settles the composer draft reads empty (the sent text neither lingers nor resurfaces).
- `make preflight` and `scripts/pr-preflight` pass with this body; changelog fragment, e2e flow covers, and the line-count and test-quality ratchets are all satisfied.

---

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
@@ -2,7 +2,7 @@
   "files": {
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": 2375,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2818,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4643,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4515,
     "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2423,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1621
   },
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
@@ -1,8 +1,8 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": 2375,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2853,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4842,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2818,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4643,
     "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2423,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1621
   },
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-mainwindow.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-mainwindow.json
@@ -1,6 +1,6 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": 1926,
+    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": 1930,
     "desktop/macos/Desktop/Sources/MainWindow/Pages/AppsPage.swift": 3643,
     "desktop/macos/Desktop/Sources/MainWindow/Pages/DashboardPage.swift": 4543,
     "desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift": 3298,
@@ -8,7 +8,7 @@
     "desktop/macos/Desktop/Sources/MainWindow/SidebarView.swift": 1570
   },
   "raise_justifications": {
-    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": "Keeps all chat tool groups compact and collapsed while streamed steps increment, with an explicit expansion policy, fixed collapsed height, and regression coverage.",
+    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": "The whole agent spawn card now opens the agent (card-level tap plus the corner affordance) so a delegated task is reachable from every surface that renders it; +4 lines.",
     "desktop/macos/Desktop/Sources/MainWindow/Pages/AppsPage.swift": "Rebase reconciliation applies the pinned swift-format to the main-branch app-detail safety fixes; runtime behavior is unchanged.",
     "desktop/macos/Desktop/Sources/MainWindow/Pages/DashboardPage.swift": "Home redesign: chat-as-home surface plus the 2nd-brain knows-list hub replace the wordmark/ribbon/WMN-card layout; a component split is tracked follow-up.",
     "desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift": "Bridge memory search now awaits the active initial or pagination projection before refreshing, preventing an immediate post-navigation marker search from using stale lifecycle capability state.",
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
@@ -1,10 +1,10 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6214,
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6218,
     "desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift": 2861
   },
   "raise_justifications": {
-    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": "ChatTurnOwner reasoning-effort lane derivation",
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": "sendMainDraft clears the sent draft up front and restores it only if the send never lands, so the notch composer can't retain or resurface the sent text; +4 lines.",
     "desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift": "Already-granted permissions no longer reopen System Settings: screen-recording and full-disk-access requests gate their Settings/drag-card opens behind the granted result."
   },
   "threshold": 1500
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-root.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-root.json
@@ -2,13 +2,13 @@
   "files": {
     "desktop/macos/Desktop/Sources/AuthService.swift": 3294,
     "desktop/macos/Desktop/Sources/CloudConnectorFormAutomation.swift": 1678,
-    "desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift": 4238,
+    "desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift": 4264,
     "desktop/macos/Desktop/Sources/MemoryExportService.swift": 1578,
     "desktop/macos/Desktop/Sources/OmiApp.swift": 1638
   },
   "raise_justifications": {
     "desktop/macos/Desktop/Sources/AuthService.swift": "Apple first-auth name capture signals observers (authNameDidUpdate) and persists the name to Firebase so it survives reinstalls.",
-    "desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift": "Reach-error card gains a debug bridge action so the actionable failure surface is verifiable in-process.",
+    "desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift": "Adds a non-prod send_main_draft action that drives the composer's sendMainDraft path so the draft clear/restore behavior is verifiable end to end; +26 lines.",
     "desktop/macos/Desktop/Sources/MemoryExportService.swift": "Pinned swift-format normalizes line layout without changing this source's runtime behavior.",
     "desktop/macos/Desktop/Sources/OmiApp.swift": "shared openMainAppWindow helper for menu bar + Open Omi shortcut"
   },
--- a/desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift
+++ b/desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift
@@ -1822,6 +1822,32 @@ final class DesktopAutomationActionRegistry {
     }
 
     register(
+      name: "send_main_draft",
+      summary: "Send the main composer via the draft path (sendMainDraft), exactly like the notch/main text composer",
+      params: ["text"],
+      category: "chat",
+      surfaces: ["main_chat", "ask_omi"],
+      safety: "local",
+      sideEffects: ["local_storage", "network"],
+      examples: ["./scripts/omi-ctl action send_main_draft text='what did I do today?'"]
+    ) { params in
+      guard AppBuild.isNonProduction else {
+        return ["error": "send_main_draft is disabled on production bundles"]
+      }
+      let text = (params["text"] ?? "").trimmingCharacters(in: .whitespacesAndNewlines)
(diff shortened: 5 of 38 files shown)


Git-on-my-level (dismissed):
Thanks for the thoughtful Track 3 submission — the chat-first notch direction is coherent and the new tests around geometry/presentation are useful. I can't approve or land this as-is yet because I found a couple of blocking integration regressions in the notch rewrite:

1. Disabled/snoozed floating-bar state is bypassed at startup. `NotchScreenManager.makePanel` calls `window.orderFrontRegardless()` as soon as each panel is created, and `FloatingControlBarManager.setup()` starts the manager before checking `isEnabled`/snooze state. That means the notch panels can appear on launch even when the user had hidden/disabled or deferred the floating bar; the old window path created the window without necessarily ordering it front. Please make initial panel creation respect the persisted enabled/snoozed/deferred state, and only order panels front through the existing show/temporary-notification paths.

2. Some legacy `window`-based paths are still wired even though the new setup leaves `FloatingControlBarManager.window` nil. For example `closeAskOmiForAutomation` now always returns `floating_bar_window_unavailable`, and `deliverAgentArtifactCompletionToFloatingSurface` drops completed agent artifacts behind `guard let window else { return }`. Those are part of the desktop automation/agent-surface contract and need to be ported to `notchScreenManager`/`notchState` rather than silently no-oping.

Given the size and UX direction here, this should also get human maintainer/product review after those blockers are fixed. I did not see an attack or supply-chain concern, and I appreciate the regression tests that were added; the remaining ask is to close these integration gaps and include validation evidence for the disabled/snoozed startup path plus the automation/agent-completion paths.

---

undivisible:
charge your keyboard

katipally:
@Git-on-my-level  Thanks for the careful review. Both blockers were real and are now fixed, with validation.

1. Disabled/snoozed startup. Root cause: NotchScreenManager.makePanel ordered every panel front the moment it was created, so the launch policy's "hidden" decision was ignored. Panels are now created hidden and visibility runs through a single applyPanelVisibility authority: a panel is on screen when the notch is passively visible OR that panel is explicitly open. That also keeps the old "show temporarily, hide on close" behavior, the Ask Omi hotkey still surfaces the panel while the bar is disabled/snoozed, and it re-hides when the conversation closes.

Evidence (named bundle + Quartz on-screen window list): enabled launch = 2 notch windows, disabled launch = 0, open_ask_omi while disabled surfaces one, close_ask_omi hides it again. Hermetic NotchScreenManagerVisibilityTests assert real NSWindow.isVisible for created-hidden, showAll, hideAll, and summon-while-hidden.

2. Legacy window paths. FloatingControlBarManager.window is never assigned in the notch build, so those paths silently no-oped. closeAskOmiForAutomation now drives notchScreenManager (closeAll + wait) exactly like open_ask_omi, so live close_ask_omi returns a clean result instead of floating_bar_window_unavailable. I removed the dead window field and everything that only reached it (the orphaned notification-reopen helpers, a write-only key, the vestigial PTT resize wrapper). On the agent artifact path: the completion isn't actually dropped, it's already projected into the shared notch timeline via projectJournalTurn (INV-6), so deliverAgentArtifactCompletionToFloatingSurface was dead legacy and I deleted it rather than resurrect the old auto-pop. The ~2000-line legacy FloatingControlBarWindow class is now unreferenced, but I left it for a separate follow-up so this diff stays reviewable.

Green across the board: full desktop Swift suite (352 suites), INV-6 continuity harness, swift-format, SwiftLint, line-count ratchet, e2e coverage. Failure-Class: FC-split-mutation-authority. No merge conflicts with main.

katipally:
@undivisible lol priorities, keyboard gets charged the moment this PR is done

katipally:
@undivisible 

Curious what you both think of this design direction. Worth me doubling down, more features, cleaner UI/UX and animations, pushing it toward a real Apple-native feel? Or keep it minimal for now? Want to build toward whatever fits your vision

undivisible:
> @undivisible @Git-on-my-level need human response
> 
> Curious what you both think of this design direction. Worth me doubling down, more features, cleaner UI/UX and animations, pushing it toward a real Apple-native feel? Or keep it minimal for now? Want to build toward whatever fits your vision

i personally love your design but we've been talking internally about removing chat from the notch to take a more verbal first direction when not using the app and keep text inputs within the app. 
thanks for the pr though! won't close this yet in case we decide to keep the chat.

katipally:
@undivisible do mean something like this ??  #10332 .

Same notch redesign, but instead of chat it's now a voice surface. Hold push-to-talk and the Omi mark becomes a live mic waveform, reforms into the rotating logo while it thinks, then a waveform that reacts to Omi's own voice while it speaks the reply. The island grows out of the notch as the answer streams, the reply lingers a few seconds so you can read it (hover holds it, Esc dismisses, tap opens the full conversation in the app), and text chat lives in the main window now.

I put it up as a separate PR so you can compare the two directions side by side without losing the chat work here. Demo and the full write-up are on that PR.

Git-on-my-level (commented):
Thanks for the quick follow-up. I re-checked the current head against the previous automation blockers:

- The disabled/snoozed startup leak looks addressed: panels are now created hidden and visibility is centralized through `applyPanelVisibility()` / `showAll()` / `hideAll()`, with regression coverage for hidden launch, explicit summon while hidden, and close re-hide.
- The stale legacy-window automation gap also looks addressed: `close_ask_omi` now drives `notchScreenManager.closeAll()`, the dead `window` field is removed from the manager path, and agent completions are kept on the shared provider/journal timeline rather than routed through the old window-local surface.

I’m dismissing the stale automation change request because those specific blockers appear resolved on this head. I’m still not approving: this remains a large first-time-contributor desktop UX rewrite, there are no passing GitHub checks in the context, and the public thread already shows the product direction is still unsettled between chat-first notch vs verbal/voice-first notch. This should stay with human maintainer/product review before merge, and the validation evidence from the local desktop suite/e2e run is useful for that review.

---

undivisible:
> @undivisible do mean something like this ?? #10332 .
> 
> Same notch redesign, but instead of chat it's now a voice surface. Hold push-to-talk and the Omi mark becomes a live mic waveform, reforms into the rotating logo while it thinks, then a waveform that reacts to Omi's own voice while it speaks the reply. The island grows out of the notch as the answer streams, the reply lingers a few seconds so you can read it (hover holds it, Esc dismisses, tap opens the full conversation in the app), and text chat lives in the main window now.
> 
> I put it up as a separate PR so you can compare the two directions side by side without losing the chat work here. Demo and the full write-up are on that PR.

ah i hadn't seen this. this is great, thanks! pinging @skanderkaroui @Git-on-my-level @kodjima33.

Git-on-my-level (changes requested):
Thanks for the follow-up here. The earlier blockers around hidden/snoozed panel creation and the stale legacy `close_ask_omi` path look addressed on this head, but I found one concrete regression that should be fixed before merge:

- `desktop/macos/Desktop/Sources/FloatingControlBar/Notch/NotchTrayView.swift` enables the send button when `pendingAttachments` is non-empty even if the text field is blank (`canSend` checks attachments at lines 121-124), but `send()` always calls `chatProvider.sendMainDraft(text)` with the blank string. In `desktop/macos/Desktop/Sources/Providers/ChatProvider.swift`, `sendMainDraft(_:)` delegates to `sendMessage(_:)`, and `sendMessage` immediately returns `nil` when `trimmedText` is empty. Result: an attachment-only message is presented as sendable from the new notch composer but pressing Send does nothing and leaves the attachment staged. Please either require non-empty text in the notch composer or add a provider path that accepts attachment-only sends consistently with the UI.

Other reviewed areas:

- `desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift` now centralizes the floating surface through `notchScreenManager`/`notchState`; the automation open/close paths are correctly moved to `openPrimary()` and `closeAll()` instead of the removed legacy single window.
- `desktop/macos/Desktop/Sources/FloatingControlBar/Notch/NotchScreenManager.swift` creates panels hidden, exposes `showAll()`/`hideAll()`, and makes pointer-driven auto-close a single authority; that addresses the prior disabled/snoozed launch concern.
- `desktop/macos/Desktop/Sources/FloatingControlBar/Notch/NotchWindow.swift` deliberately uses an assistive-tech-high level while yielding for system dialogs; this is product/UX-sensitive and still deserves maintainer sign-off because it can sit above many third-party overlays.
- `desktop/macos/Desktop/Sources/FloatingControlBar/Notch/NotchView.swift` and `NotchViewModel.swift` share the same presentation ladder and size authority, with separate chat-height animation, which is a reasonable structure for the dynamic-island behavior.
- `desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController+SessionDelegate.swift` mirrors live voice text into notch state only as transient UI and clears via the voice presentation lifecycle; because it displays live speech before journal finalization, the privacy/product behavior should still get human maintainer review.
- `desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift` adds a non-production-only `send_main_draft` automation action, which is appropriately gated and useful for validating the draft clear/restore path.
- `desktop/macos/Desktop/Tests/Notch*` and the updated `ChatErrorStateTests.swift` add useful coverage for geometry, presentation priority, hidden launch, window level, and draft clear/restore, but I do not see coverage for the attachment-only notch composer case above.

I did not run the macOS app/test suite from this Linux review environment; this is a static diff review of the current head. Leaving this for human maintainer review as a macOS product-direction decision after the attachment-only send regression is fixed.

— Reviewed by automated maintainer agent (glm-5.2).

---

aryanorastar:
Closing as a maintainer.

This is still CONFLICTING on f76ebe5 (22 Jul), with CHANGES_REQUESTED. The conflict list is the shipping notch: FloatingControlBarWindow/View/State, ChatProvider, ChatBubble, ChatMessagesView.

Main already answered the product question left open here. Ask Omi opens the main chat. The notch is not a text surface. This branch does the opposite: hover-to-chat in a NotchScreenManager island, and the hotkey opens that composer instead of the app.

The voice-first sibling #10332 is closed. The full-app rewrite #10672 was already closed as too much; cherry-picks belong on the current bar. NotchVoiceMorphMark is what landed. Rebasing this island would put chat back in the notch.

Do not bring NotchScreenManager or a notch composer back on top of FloatingControlBarWindow.

PR #10255, opened 2026-07-22
fix(backend): include surface flag in OpenAPI stale-contract hint (#10217)

### LINKED ISSUE #10217: export_openapi.py stale-contract error message omits the required --app-client flag
The app-client contract check failure says:

```
OpenAPI contract check failed: ../docs/api-reference/app-client-openapi.json is stale; run backend/scripts/export_openapi.py --write ../docs/api-reference/app-client-openapi.json
```

Following that command verbatim silently exports the **public** surface (surface defaults to `public`) into the app-client file — 53k lines deleted, wrong contract — because surface selection is a separate flag, not inferred from the output path. The correct command is `backend/scripts/export_openapi.py --app-client --write`.

Evidence: hit while preparing #10215 today; the wrong invocation produced a 17-path public spec in place of the 309-path app-client spec and `--check` (also without the flag) then validated the wrong surface as 'up to date'.

Fix options: include the surface flag in each surface's error message, or infer the surface when the target path matches a known default spec path.

kodjima33:
Closes #10217

## Bug
`backend/scripts/export_openapi.py`'s contract check prints a fix-it hint on a stale or missing spec:

```
... is stale; run backend/scripts/export_openapi.py --write <path>
```

The surface (`public` / `app-client` / `integration-public`) is selected by a **separate flag** that defaults to `public` and is **not** inferred from the output path. So following the hint verbatim for the app-client or integration-public spec regenerates the *public* surface and overwrites the wrong contract — the reporter saw a 17-path public spec written over the 309-path app-client spec (~53k lines deleted) — and a subsequent `--check` (also missing the flag) then validated the wrong surface as "up to date".

## Fix
`check_spec` now receives the surface and prepends the selecting flag to both the stale and does-not-exist hints via a small `surface_flag()` helper:
- `app-client` → `--app-client --write <path>`
- `integration-public` → `--surface integration-public --write <path>`
- `public` → unchanged (`--write <path>`)

The `main()` `--check` call passes `args.surface` through. No behavior change for the public surface; existing callers/test remain valid.

## Verification
Exercised the real `check_spec`/`surface_flag` functions from source over a temp spec for all three surfaces plus the missing-file path — each hint now names the flag that actually regenerates that surface. Added `test_check_spec_hint_includes_surface_flag` as the regression test (extends `tests/unit/test_openapi_contract.py`). Backend Python env is not installable in the watchdog sandbox, so the full pytest suite runs in CI.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10255?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/export_openapi.py
+++ b/backend/scripts/export_openapi.py
@@ -985,12 +985,26 @@ def write_spec(path: Path, generated: str) -> None:
     path.write_text(generated)
 
 
-def check_spec(path: Path, generated: str) -> None:
+def surface_flag(surface: str) -> str:
+    """Return the CLI flag that selects `surface`, so error hints regenerate the right contract.
+
+    Surface selection is a separate flag from `--write`/`--check`; omitting it silently defaults to
+    the public surface and would overwrite a non-public spec with the wrong contract (#10217).
+    """
+    if surface == 'app-client':
+        return '--app-client '
+    if surface == 'integration-public':
+        return '--surface integration-public '
+    return ''
+
+
+def check_spec(path: Path, generated: str, surface: str = 'public') -> None:
+    flag = surface_flag(surface)
     if not path.exists():
-        raise OpenAPIContractError(f'{path} does not exist; run export_openapi.py --write {path}')
+        raise OpenAPIContractError(f'{path} does not exist; run backend/scripts/export_openapi.py {flag}--write {path}')
     current = path.read_text()
     if current != generated:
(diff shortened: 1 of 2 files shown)


kodjima33:

PR #10256, opened 2026-07-22
feat(agent): overhaul cloud agent reliability, streaming, and error handling

skanderkaroui:
# Agent reliability & performance overhaul

One PR, four measured iterations on the agent stack (cloud VM lane + desktop kernel + desktop error UX). Every change traces to an instrumented experiment, a telemetry corpus, or a live reproduction — the evidence loop ran ~90 live agent runs and session simulations against a real 27MB database copy, plus a 30-day PostHog `chat_agent_error` deep dive.

## Headline results (same 8-turn conversation, real data, before → after)

| Metric | Before | After |
|---|---|---|
| Conversation wall time | 256s | 155s (−39%) |
| Conversation cost | | −55% |
| Worst client silent gap | 46s | ~13s, typical 1–4s |
| Stop mid-turn | error / indistinguishable partial | `interrupted: true` partial |
| Reconnect | total context amnesia | full context retained |
| Task battery (5 heavy tasks) | 519s, 53 tool calls | 199s, 20 tool calls (−62%) |

### Subagent architecture (desktop kernel + cloud)
- `researcher` subagent (SDK `agents` option): row-heavy data exploration runs in a disposable child context and returns distilled prose findings — Haiku-pinned (measured fastest, ~3× cheaper, accuracy parity) and routed by row-volume. A typed handoff remains parked because this SDK stream exposes no production interception point.
- Kernel: run cancellation cascades to delegated children (background floating-bar agents deliberately exempt — product behavior pinned by test); spawn-time `toolPolicy.allowedToolNames` enforced at the capability broker (narrow-only, fail-closed, no schema changes).

### Streaming & session reliability (cloud lane)
- `includePartialMessages` was never enabled — all streaming code was dead; every turn had 17–46s of total client silence (the stall→Stop churn mechanism). Now: real deltas, subagent progress as `subagent:*` tool_activity, bounded ephemeral researcher snippets in status events (never the answer/transcript), and a 10s turn heartbeat capping any remaining gap.
- Sessions survive WebSocket reconnects (server-scope session, swappable sink, `session_state` hello); reconnect amnesia reproduced live before the fix ("I don't have any earlier conversation context") and re-verified fixed.
- Real-SDK interrupts map to `interrupted: true` partial results instead of errors; pending tools terminalize as `interrupted` rather than false `completed`; a prewarm race that made Stop a silent no-op on fast first queries is fixed.
- Supersede queue-of-1: a query arriving mid-turn parks, interrupts, and starts only at the turn boundary — double-taps can no longer interleave two answers.

### Tool system (both sides of the VM↔backend boundary)
- `get_daily_recap` gains explicit date ranges (it could only anchor to "now" — root cause of 13+ SQL-turn fallbacks), authoritative empty-range statements, hourly timeline/top windows; new `get_app_usage` day×app matrix; batch `execute_sql`; task near-duplicate guidance (the store carries ~63% duplicate extractions).
- SQL guard: `stmt.readonly` replaces a prefix guard that rejected CTE reads and false-positived on user data containing words like "create"; row cap moved to iteration.
- Backend-tool loading: per-definition isolation + name validation + dedupe at fetch time (local tool names are reserved); 30s retry on failure (a boot blip previously lost calendar/gmail for the VM's lifetime, silently); `resp.ok` checked; Pydantic `Optional` accepts explicit `null`; results bounded at 10K chars; `agent_tools.py` gets per-tool schema isolation + `record_fallback` on degraded toolsets.

### Two-plane error handling & resilience
- Bounded 7-category taxonomy shared JS↔Python: users get fixed calm copy + automatic backoff on transient failures; the internal plane gets maximum structured detail (full stacks, JSON lines). Raw SDK and integration-connect response details stay off the model/user plane.
- Owned-abort grace window replaces blanket "aborted" string suppression; session-loop death notifies the client and recovers on the next query instead of hanging forever; crash exits flush a synchronous post-mortem.
- agent-proxy: Firestore-resident VM restart circuit breaker (3 failures → 30min cooldown, honest user copy instead of an infinite spinner), VM connect retry ×3 with progress events, auth-push retry + degraded-tools notice, all broad catches split with `exc_info=True`, breaker reads off the event loop.
- Context ceilings: `maxTurns: 10` per user turn + tool-result truncation (SDK compaction is API-beta only in 0.2.39 — tracked for 0.2.40+).

### Desktop error UX (from the PostHog deep dive)
- The opaque "Something went wrong. Please try again." bucket (401 events/108 users) fully decomposed via `raw_error`; the three "unexplained" error spikes were retry storms caused by the copy itself (users with exhausted Anthropic credits obeying "please try again").
- `AgentErrorClassifier`: 12 bounded codes with honest copy and retryable verdicts — "try again" only when a retry can help; `chat_agent_error` telemetry gains `error_code`/`retryable`/`failure_code`/`failure_source`/`adapter_id`/`provider` (closed vocabularies, never raw exception text — test-pinned).
- Also unwedges main: two files from an upstream refactor had swift-format drift, one wedged against its frozen line-count ratchet.

## Product invariants

- **INV-AGENT-***: touched `desktop/macos/agent/src/runtime/**`. Lifecycle affected by the cascade; guard tests updated (`cancellation.test.ts` +3 behavioral cases, `run-tool-capability.test.ts` +3, `control-tools.test.ts` schema cases, new `spawn-tool-policy.test.ts`). toolPolicy strengthens leaf-worker restrictions (narrow-only, kernel-internal checks).
- **INV-CHAT-1**: path-matched (`ChatProvider.swift` telemetry call site, regenerated `Sources/Generated/*`); no chat write-path, projection, or timeline behavior changed.
- **INV-AUTH-1**: path-matched; no session-death behavior changed — auth-classified errors alter copy and telemetry only.
- **INV-INT-1**: path-matched by the connector guidance overlay; connector disclosure behavior remains explicit.
- **INV-MEM-1**: path-matched only by `web/app/src/components/memories/MemoriesPage.tsx` (and `useMemories.ts`), which this PR touches as Prettier reformatting with zero logic change. No tier vocabulary, canonical collection, or default Short-term+Long-term access policy is altered; Archive remains excluded from default access.

Failure-Class: none

## Deliberate non-changes (on record)

`bypassPermissions` stays — the per-user VM is a single-tenant sandbox; SDK-level tool gating is intentionally off (maintainer decision). No Sentry (not in the backend stack); no chat-save outbox (the write target is the failure domain); no cross-language error library (7-string contract, ~15 lines per side); prompt-level turn-discipline and batching rejected by experiment (model-dependent).

## Integration order

Merge #10600, #10601, and #10599 first, then merge current `main` into this branch before clearing the final review. Those focused PRs repair the VM-IP recovery boundary, the backend event-loop/error plane, and end-to-end tool outcome transport that this larger branch also touches.

## Verification

- agent-cloud: vitest 41/41 (6 files incl. hermetic WS e2e that spawns the real server against a scripted SDK — now also pins interrupted tool terminalization); CI-wired via `agent-cloud-tests` in the checks manifest (first CI coverage for the package).
- Desktop agent: 587/587 + agent-logic harness green; Swift: build green, `AgentErrorClassifierTests` + `ChatQueryTelemetryTests` 34/34; backend units 12/12 (breaker transitions, classification, per-tool isolation).
- Full preflight 55/55 locally (Mac mini cross-validation pending — machine offline). Cascade regression test verified to fail on unfixed code. Live GCE breaker trip is the one non-CI-coverable path; manual verification planned and declared per DoD.


Git-on-my-level:
Thanks for the careful work here — the direction looks useful and the focused regression coverage is strong.

I reviewed this as an agent-control/agent-behavior change. The important maintainer consequence is that this PR changes how Omi agents are instructed and constrained:
- the cloud agent can now invoke a `researcher` subagent for broad database exploration, with that subagent restricted to the Omi data tools;
- spawned/delegated children can receive a narrow-only `toolPolicy.allowedToolNames` allowlist, enforced at capability registration;
- cancelling a parent run now cascades to delegated child runs, while user-visible background agents intentionally continue.

That all looks conceptually safe from this pass: the tool policy intersects with the existing role/adapter allowlist rather than widening it, malformed persisted policy fails closed, and the cancellation behavior is covered by regression tests. I also ran the targeted tests locally:
- `desktop/macos/agent`: `npm test -- tests/cancellation.test.ts tests/control-tools.test.ts tests/run-tool-capability.test.ts tests/spawn-tool-policy.test.ts` → 113 passed
- `desktop/macos/agent-cloud`: `npm test -- tests/query-config.test.mjs` → 3 passed

I’m leaving this as a positive signal rather than formal approval because it still warrants human maintainer sign-off before merge: it changes agent behavior/control-plane semantics, touches a lockfile/dependency surface, and GitHub still shows blocked/pending/cancelled checks. Once the queued/cancelled CI state is resolved, this looks like a good candidate for maintainer review.

---

Git-on-my-level (dismissed):
Thanks for continuing to harden this. I’m changing this from a positive signal to changes requested because the current head has concrete merge blockers that need to be resolved before maintainer review can be effective.

Blocking items:
- Hygiene is failing on the product-invariant gate: the PR touches `web/app/src/components/memories/MemoriesPage.tsx` but the PR body does not include `INV-MEM-1` under `Product invariants affected`. The CI log gives the exact block to paste, including `INV-AGENT-*`, `INV-AUTH-1`, `INV-CHAT-1`, and `INV-MEM-1`.
- Desktop Swift Test Suites compiled successfully but was cancelled at the 60-minute job limit. Please either make the selected suite complete inside the CI budget or narrow/split the manifest-selected Swift test route so the changed desktop coverage is actually green.
- Desktop Swift Build & Tests is failing because it depends on that cancelled Swift test result.

Agent-behavior / maintainer-impact note: this still changes agent-control guidance and behavior. The PR adds/updates cloud-agent research findings and instructions, researcher-subagent routing, batch SQL/data tools, cloud-agent streaming/error handling, and spawn/delegation tool-policy constraints. The important maintainer consequence is that AI coding/review/runtime agents may explore user data differently, delegate row-heavy DB work to a constrained subagent, and rely on the new tool-policy/cancellation contracts. The direction still looks conceptually safe from this pass because the policy narrows rather than widens access and the agent-cloud tests are passing, but this is enough behavior/control-plane surface that human maintainer sign-off remains required after CI is green.

I’m not asking to split the PR just because it is large; the work is mostly cohesive around agent reliability/control-plane behavior. The request is to clear the specific invariant/CI blockers and keep the existing human/dependency review path before merge.

---

Git-on-my-level (commented):
Thanks for the follow-up work here. The previous automation blockers I called out on `851e7c20` look resolved on this head: Hygiene is green, the PR body now cites the memory invariant, and the desktop/backend/agent-runtime checks are passing.

I’m not formally approving because this is still a large agent-control/user-data change and the PR remains in the human/dependency review lane. From this pass, the implementation direction looks promising and much better covered: the cloud-agent streaming/session tests are green locally (`npm test` in `desktop/macos/agent-cloud`: 61/61), CI is green, and the new tool-policy path narrows child-agent tools rather than widening them.

Maintainer-impact / agent-behavior note: this PR changes how Omi’s AI agents are instructed and constrained. It adds cloud-agent research guidance, researcher-subagent routing for row-heavy local DB work, batch SQL/date-range data tools, streaming/error/session behavior, and spawn-time `toolPolicy.allowedToolNames` enforcement. That can change how AI coding/review/runtime agents explore user data and delegate work in this repo; the guidance looks directionally safe from this pass because the enforceable parts are outside compactable model context and are fail-closed/narrow-only, but it still warrants human maintainer sign-off.

One privacy/security item for the human review checklist: the new integration registry maps `calendar`, `gmail`/`email`, and contacts to one `google_calendar` OAuth provider and derives a single scope bundle from all provider capabilities. That means a one-click connect flow for one Google capability may request Calendar + Contacts + Gmail read scopes together. That may be intentional product direction for a unified Google connection, but it is a user-data consent expansion that should be explicitly accepted before merge.

I’m removing the stale `needs-tests` label and adding `security-review` for that OAuth/user-data surface while keeping `needs-maintainer-review` and `dependency-review`.

---

skanderkaroui:
Conflicts resolved against latest main (`d17e25a555`, preflight 85/85) — #10599/#10600/#10601 landed underneath this branch, so `agent-proxy/main.py` keeps both the restart circuit-breaker and main's usable-IP guards, and `agent.mjs` keeps both main's `/sync` column-grouping fix and this branch's SQL-identifier guard.

On the OAuth item: the registry derives exactly the scope set main already ships — `calendar`, `contacts.readonly`, `contacts.other.readonly`, `gmail.readonly`, keyed to the same single `google_calendar` provider (`GOOGLE_OAUTH_SCOPES` / `GOOGLE_INTEGRATION_KEY` in `google_utils.py`). Zero scopes added or removed, so this is a refactor of an existing consent surface rather than an expansion of it. The bundling itself is still a fair thing to want signed off — just flagging that it's pre-existing behaviour, so a reviewer looking for the change in this diff won't find one.

undivisible:
#10686 moves the cloud-agent runtime to the Python Claude Agent SDK while preserving the VM protocol, mixed-schema sync, database context, and Playwright MCP. Please resolve the deleted agent-cloud files before merge.

Git-on-my-level (changes requested):
Thanks for keeping this branch updated. I'm requesting changes for the current head because GitHub now reports the PR as conflicting/dirty, and the new maintainer note points to the concrete conflict source: #10686 moves the cloud-agent runtime toward the Python Claude Agent SDK and deletes/replaces parts of the `desktop/macos/agent-cloud` surface that this PR still changes. Until those file-level conflicts are resolved against the current runtime direction, this is not reviewable/mergeable.

Maintainer-impact / agent-behavior note: this PR still changes AI-agent behavior and instructions across the cloud-agent runtime, researcher-subagent routing, tool-policy narrowing, streaming/session handling, and the added `desktop/macos/agent-cloud/experiments/RESEARCH-FINDINGS.md` guidance. Those changes may alter how Omi agents explore user data, delegate work, stream progress, and constrain child-agent tools. The direction may still be useful, but after the #10686 runtime migration the agent-behavior guidance and implementation need to be reconciled with the new Python runtime before a human maintainer can sign off.

Please rebase/merge current `main` after the runtime migration decision is settled and resolve the deleted/replaced `agent-cloud` files so the remaining diff reflects the intended production runtime. I'm keeping the existing human/dependency/security review labels; no formal approval is possible for this large agent-control/user-data PR.

---

PR #10257, opened 2026-07-22
feat(desktop): deliver background-agent completions into live voice sessions

skanderkaroui:
## Problem

A voice-spawned background agent's completion never reaches the live voice conversation. The kernel journals the run and the floating pill turns green, but nothing converts "done" into conversation input for the realtime model — so it cannot mention the result or answer "what did it find?". The pull API built for exactly this (`DesktopCoordinatorService.peekCompletedAgentDelta`, with per-surface seen-id + high-water checkpoints and a prompt that already says "Do not read raw ids aloud") had **zero callers**.

## Change

`AgentCompletionVoiceDelivery` (new, `Chat/`) wires that orphaned pipeline to the realtime hub:

- Observes `AgentRuntimeStatusStore` projections for transitions **into** a terminal state on background surfaces only (`floating_bar`/`service`/`workstream`) — ordinary main-chat/voice answers never trigger kernel delta reads.
- Peeks the canonical kernel completion delta under the `realtime_voice` checkpoint namespace and injects the existing delta prompt into the live provider session as **silent context** via the same buffered text-input path PTT uses (`sendBackgroundAgentContext`, new thin wrapper beside `sendTestTextInput`). **No synthetic response is requested** — the voice turn coordinator's output-lease authority and the one-start/one-terminal turn analytics contract are untouched; the model mentions the completion at its next natural turn.
- Acknowledges the checkpoint **only after the session confirms the send**; undelivered completions retry on the next transition or on `hubDidConnect` (which also drains completions that finished while no session was live).

Hooks are minimal: 1 line in `hubDidConnect`, 1 startup call in `OmiApp`, a 2-member controller extension in its own file.

- INV-AUTH-1 — path-matched (`RealtimeHub*`); no session-death or auth behavior changes. The new code performs no auth calls; it forwards text into an already-authenticated session.
- INV-CHAT-1 — path-matched; no chat write-path changes. Nothing journals, appends timelines, or creates turns — injection is provider-conversation context only, and checkpoint state is the pre-existing delta bookkeeping.

## Verification

- `xcrun swift build -c debug --package-path Desktop` clean.
- New behavioral suite `AgentCompletionVoiceDeliveryTests` (9 tests, injected seams, no wall-clock waits): delivery+ack ordering; checkpoint unadvanced on failed inject; no-op without a live session; primary-surface terminals never trigger; repeated terminal projections fire once; cleared-and-recreated surface is a fresh transition; connect drains pending; in-flight coalescing; empty delta acks nothing.
- `./scripts/agent-logic-harness.sh` — all four gates green.
- `python3 scripts/check_desktop_test_quality.py` — at/below baseline.
- `make preflight` green at HEAD.

**Not exercised live:** a real provider voice session observing a completion on a named bundle (needs live OpenAI/Gemini audio session). Follow-up before merge or as merge-gate evidence: spawn an agent by voice on a named bundle, wait for completion, confirm the next model turn mentions it.

## Deliberate scope cut

Immediate *spoken* announcement (model proactively speaking mid-idle) is out of scope: `hubDidReceiveAudio`/`hubDidEmitText` drop events without an accepted turn identity, so proactive speech requires a `VoiceTurnCoordinator`-owned announcement turn — tracked as follow-up work, not smuggled around the output authority here.

## Ratchet notes

`RealtimeHubSession.swift` (+7) and `OmiApp.swift` (+3) baselines raised with one-line justifications; the controller addition lives in a new extension file instead of growing `RealtimeHubController.swift` past its threshold.

## Failure class (fixes)

Failure-Class: none

The two follow-up `fix:` commits are specific instance fixes (an ack-before-delivery bug and a test-triggered runtime reach), each with a regression test. No existing failure class matches and neither warrants a new reusable guard class.


Git-on-my-level (dismissed):
Thanks for wiring this path up — the overall shape is promising, and I like that the completion delta is treated as untrusted context rather than forcing a synthetic voice response.

I do need to block this head on the delivery/acknowledgement contract: `AgentCompletionVoiceDelivery` acknowledges the kernel checkpoint whenever `sendBackgroundAgentContext` returns `true`, but `RealtimeHubSession.sendTextInput` also returns `true` when it merely buffers text because the socket is not open or Gemini has no activity window. That means a completion can be marked seen before it is actually delivered; if the session closes before the buffered text flushes, `close()` clears `pendingTextInputs` and the completion is lost permanently. This contradicts the PR invariant that the checkpoint advances only after confirmed delivery.

Please make the background-completion path distinguish “accepted into an in-memory buffer” from “actually sent to the provider conversation,” or otherwise delay `acknowledgeCompletedAgentDelta` until the buffered context is flushed successfully. A regression test should cover the socket/activity-window buffered path plus close-before-flush so this exactly-once guarantee does not regress.

Also, formal approval is intentionally withheld because this is a nontrivial desktop voice-agent behavior change and the Desktop Swift checks are still pending; it will still need maintainer/product review after the delivery-loss bug is fixed.

---

skanderkaroui:
Thanks — good catch, that was a real exactly-once break and I've fixed it in `3a59488` (`fix(desktop): only ack a voice completion after it reaches the provider`).

**What changed:** `sendBackgroundAgentContext` no longer delegates to the buffering `sendTextInput`. It writes to the live provider only when the session can accept it *now* (`isOpen`, and for Gemini an open activity window) and returns `true`; otherwise it returns `false` **without buffering**. So the checkpoint advances only after a confirmed provider write, and there is nothing in `pendingTextInputs` for `stop()`/`close()` to drop. Not-ready completions stay unacknowledged and retry on the next terminal transition or session connect.

**Bound (documented, not a loss):** a completion arriving while a live Gemini session has no open activity window is *deferred* until the next window/turn or reconnect — never acknowledged until actually sent.

**Regression tests** (`RealtimeHubSessionInputLifecycleTests`, the DEBUG-seam suite), covering exactly the buffered/close-before-flush path you flagged:
- background context before transport ready → returns `false`, buffers nothing
- open OpenAI session → returns `true`, buffers nothing
- Gemini open socket, no activity window → returns `false`, buffers nothing

(new `pendingTextInputCount` field on the DEBUG input-lifecycle snapshot asserts "nothing buffered").

Separately, I also fixed the pending Desktop Swift CI failure it was blocked on: the `Test Suites` job was hanging 6h because the `hubDidConnect` hook drove the shared delivery singleton into the real agent runtime during RealtimeHub unit tests. The service is now inert until `start()` (app-launch only), so those tests can't reach the runtime (`9befcb8`). Agreed this still needs maintainer/product review — the live named-bundle voice verification is still outstanding and noted in the PR body.

Git-on-my-level (dismissed):
Thanks for the follow-up fixes — the original ack-before-delivery issue is much improved now that `sendBackgroundAgentContext` refuses instead of buffering, and the new unit coverage around closed sockets/Gemini idle windows is useful.

I still need to request changes on one remaining delivery hole: when a Gemini session is already connected but idle, `sendBackgroundAgentContext` correctly returns `false` because there is no open activity window, so `AgentCompletionVoiceDelivery` leaves the checkpoint unadvanced. But nothing schedules another delivery when the next activity window opens. The only retry triggers I see are another background-agent terminal transition and `voiceSessionDidConnect()`. `beginInputTurn()` opens the Gemini activity window and flushes ordinary buffered text inputs, but this background-completion path deliberately does not buffer and `AgentCompletionVoiceDelivery` is not notified there. So a completion that lands while Gemini is warm-idle can still fail to reach the very next natural voice turn — exactly the follow-up moment this feature is meant to support — until an unrelated terminal transition or reconnect happens.

Please add a retry trigger at the point the realtime session becomes able to accept background context again (for example when a Gemini activity window opens / input turn begins, while keeping the no-buffer/ack-after-send invariant), and cover that end-to-end in tests: completion delivery fails while Gemini is connected idle, the next user turn/window opens, delivery is retried, and the checkpoint advances only after the send succeeds.

Formal approval remains withheld because this is a nontrivial desktop voice-agent behavior change with outstanding live named-bundle verification/product sign-off.

---

Git-on-my-level (commented):
Thanks for the quick follow-up. I re-reviewed the current head (`7348c68`) against the two prior automation blockers:

- The original ack-before-delivery issue now looks addressed: `sendBackgroundAgentContext` refuses instead of buffering when the provider cannot accept context, and acknowledgement stays after a successful provider send.
- The Gemini warm-idle retry hole now looks addressed: `beginInputTurn()` opens the activity window and notifies `AgentCompletionVoiceDelivery`, with coverage for the failed-while-idle → retry-on-window-open → ack-after-send path.

I’m leaving this as a positive signal rather than formal approval because this is still a nontrivial desktop voice behavior change, the current Desktop Swift CI is not green (`Desktop Swift Test Suites` was cancelled and the aggregate build/test gate is failing), and the PR body still calls out that live named-bundle/provider verification has not been exercised. Please keep `needs-maintainer-review`/`needs-tests` until CI and the live voice-session validation are clean.

---

PR #10265, opened 2026-07-22
fix(listen): latch a clean STT upstream close and terminate the zombie session (#10028)

### LINKED ISSUE #10028: fix(listen): terminate stale live sessions when STT peers end cleanly

aryanorastar:
## What

Fixes #10028 — the "Listening" zombie when a live STT upstream dies cleanly.

When a Parakeet or Modulate live upstream WebSocket ends **cleanly** (the `async for` over the socket exhausts without raising `ConnectionClosed`), the wrapper fell through with `_dead` still `False`. `live_stt_socket_is_dead` then read that stale `False`. Since the listen loop only observes provider death while flushing a client audio buffer (`_flush_stt_buffer`), the mobile socket stayed `ready` ("Listening") until the next audio frame's provider send happened to fail — or, with no further audio, up to the **300s `ws_receive_timeout`**.

## Fix

1. **Latch the clean close** in both wrappers (`SafeModulateSocket._recv_loop`, `ParakeetWebSocketSocket._receive_loop`): a clean `async for` exhaustion now marks the socket dead, **excluding** an explicit local drain (`_closed`) or an expected provider `done` finalization — those are not failures.
2. **Propagate without a client frame**: a frame-independent `_monitor_stt_death` poll on the receiver drives the existing **idempotent** `stt_failed` + WebSocket 1011 terminal path (`terminate_live_stt_session`) as soon as the death latch flips. It sleeps via the shutdown-aware `host.wait`, so normal teardown still owns clean termination and the task can't outlive the session.

## Tests (hermetic)

- `test_stt_clean_close_latch.py` — both providers: a clean upstream close latches dead; a `done` frame / local drain does **not**.
- `test_listen_stt_death_monitor.py` — the monitor terminates a dead socket (reason `connection_lost`), leaves a live one alone, and no-ops once another path already terminalized.
- Updated `test_modulate_stt.py::test_invalid_json_skipped` for the new clean-close latch — it now asserts the death reason is the clean close (not the skipped bad frames), preserving the original "invalid JSON is non-fatal" intent.
- **121 passed** across the STT / listen suites.

## Scope note

Wrapper- and monitor-level coverage is hermetic and covers both new behaviors. A full `/v4/listen` route test driving a loopback provider peer (as the issue suggests) is a heavier integration harness and a reasonable follow-up.

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10265?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/listen/receiver.py
+++ b/backend/routers/listen/receiver.py
@@ -69,6 +69,10 @@ from utils.listen_audio import ChannelConfig, mix_n_channel_buffers, resample_pc
 
 logger = logging.getLogger(__name__)
 
+# Cadence for the frame-independent provider-death monitor (#10028). Short enough
+# to terminate a zombie "Listening" session promptly, far below ws_receive_timeout.
+STT_DEATH_POLL_INTERVAL_SECONDS = 1.0
+
 
 def _get_opuslib() -> Any:
     if opuslib is None:
@@ -237,6 +241,7 @@ class ListenReceiver:
             self.stt_socket = (
                 GatedSTTSocket(raw, gate=self.vad_gate, passthrough_audio=passthrough) if self.vad_gate else raw
             )
+            self.host.spawn(self._monitor_stt_death(provider), name='stt_death_monitor')
             return True
         except Exception as error:
             await self._drain_stt_sockets()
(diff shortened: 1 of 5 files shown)


Git-on-my-level:
Thanks for the focused fix — this is a good direction for #10028.

I reviewed the STT clean-close latch and the new frame-independent death monitor. The change looks cohesive and appropriately scoped: clean upstream closes now latch provider death, expected local drain / Modulate `done` finalization is preserved, and the listen receiver drives the existing idempotent `stt_failed` + 1011 path without waiting for another client audio frame.

I also ran the targeted backend tests locally in an offline hermetic env:

```bash
cd backend
OMI_ENV_STAGE=offline pytest -q tests/unit/test_stt_clean_close_latch.py tests/unit/test_listen_stt_death_monitor.py tests/unit/test_modulate_stt.py -q
```

Result: all selected tests passed.

No blocking implementation changes from me. I’m leaving this as a positive signal rather than formal approval because the maintainer automation approval gate still requires human maintainer review for this head (the cached gate also had non-passing Formatting/Hygiene context / blocked merge state). A maintainer should do the final sign-off before merge.



---

kodjima33 (approved):
fix(listen): latch clean STT upstream close, root cause+test, no failing checks — 5/5 backend bug fix

aryanorastar:
@Git-on-my-level need human response — thanks for the review and the offline test run. All checks pass and there are no blocking items; ready for a maintainer's final sign-off on #10028.

PR #10270, opened 2026-07-22
fix(conversations): reject reprocessing a soft-deleted conversation

aryanorastar:
## What

A soft-delete/tombstone data-integrity gap: `POST /v1/conversations/{id}/reprocess` regenerates a soft-deleted conversation's derived data — resurrecting content the user deleted.

`reprocess_conversation` fetches through `_get_valid_conversation_by_id`, which does not filter soft-deleted tombstones (`get_conversation` returns them; the helper only 404s on a missing doc). It then runs `process_conversation(force_process=True, is_reprocess=True)`, which regenerates structured data, action items, **memories** and embeddings. So reprocessing a deleted conversation — via a direct API call, or a delete-vs-reprocess race — re-derives the user's memories / action items / embeddings from content they deleted.

## Fix

Reject a **deleted** conversation in `reprocess_conversation` (404), while still allowing a **discarded** one — which reprocess intentionally revives (per its docstring, and the `eligible_merge_target` contract). Checked on the raw doc because the `Conversation` model does not carry `deleted`.

## Repairing the failure-class boundary

This is the **third instance** of one tombstone-eligibility contract — *a soft-deleted tombstone must not have content operations applied, or deleted data resurfaces*:

- sync merge target — #10119 (`eligible_merge_target`)
- user-initiated merge — #10262
- reprocess — this PR

Per AGENTS.md ("if two or more recent fixes share the cause, add a reusable guard surface"), rather than a third point-fix I extracted the shared **`is_soft_deleted`** predicate and converged `eligible_merge_target` onto it (behaviour-preserving), so the reprocess guard and the sync guard share one definition. #10262's inline merge check can adopt it too — a small follow-up rather than re-touching that merged path here; formalizing a named failure class is a reasonable registry follow-up.

## Tests (hermetic)

- `test_reprocess_tombstone_guard.py` — the predicate (deleted → tombstone; discarded / plain / None → not), reprocess **rejects** a deleted conversation (404, `process_conversation` never called), and still **allows** a discarded one.
- Existing `eligible_merge_target` and merge-validation tests stay green (behaviour-preserving refactor). **49 passed.**

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10270?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-database.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-database.json
@@ -1,10 +1,10 @@
 {
   "files": {
-    "backend/database/conversations.py": 1601,
+    "backend/database/conversations.py": 1616,
     "backend/database/users.py": 1943
   },
   "raise_justifications": {
-    "backend/database/conversations.py": "Public shared-chat uses the existing conversation storage format for bounded, safe transcript decoding; extraction would duplicate crypto/storage invariants. +21 for the #10033 merge-target eligibility predicate and pure closest-match selector beside the query they filter."
+    "backend/database/conversations.py": "Public shared-chat uses the existing conversation storage format for bounded, safe transcript decoding; extraction would duplicate crypto/storage invariants. +15 for the shared is_soft_deleted tombstone-eligibility predicate that eligible_merge_target and the reprocess guard converge on (#10119/#10262)."
   },
   "threshold": 1500
 }
--- a/backend/database/conversations.py
+++ b/backend/database/conversations.py
@@ -1524,6 +1524,21 @@ def store_conversation_photos(
 # ********************************
 
 
+def is_soft_deleted(conversation: Optional[dict]) -> bool:
+    """Whether a conversation is a soft-deleted tombstone.
+
+    A tombstone is invisible to the user, so any content operation that reads it
+    and writes derived state — merging its segments, or reprocessing to
+    regenerate structured data, action items, memories and embeddings —
+    resurrects data the user deleted. Such operations must reject a tombstone.
+
+    Shared predicate behind that contract (sync #10119 via `eligible_merge_target`,
+    merge #10262, reprocess). Deliberately distinct from `discarded`, which stays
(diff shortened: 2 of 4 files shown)


Git-on-my-level (commented):
Thanks for tightening this up — rejecting soft-deleted conversation tombstones before reprocess is the right direction for protecting deleted user data.

I reviewed the diff statically. The shared `is_soft_deleted()` predicate keeps `discarded` conversations revivable while blocking `deleted` tombstones, and the router guard runs before deserialization/reprocessing so deleted content should not re-enter structured data, memories, or embeddings. The unit coverage also exercises the tombstone case and the discarded regression path.

I’m not formally approving because this touches a user-data/privacy-sensitive reprocessing path and GitHub currently reports the backend unit suite failing (`tests/unit/test_smoke_what_matters_now.py`). I added `security-review` / `needs-maintainer-review` so a human maintainer can validate the tombstone contract and the failing check before merge. No code blocker found in the changed lines from my review.

---

kodjima33 (approved):
fix(conversations): reject reprocessing a soft-deleted conversation — bug fix; held (backend unit suite red)

aryanorastar:
@Git-on-my-level need human response — approved and rebased onto latest `main` today. The one red check (`Backend unit suite` → `test_desktop_transcribe.py::...test_ws_parakeet_connection_failure_falls_back_before_audio`) is a **flaky, unrelated** test — it passes deterministically locally (`1 passed`); the CI failure is a mock/async race (`Expected 'send' to be called once. Called 0 times`) with no relation to this conversation-reprocess change. Could a maintainer re-run that job (I lack admin) or merge past it? Thanks!

PR #10273, opened 2026-07-22
fix(conversations): stop the public shared endpoint leaking owner-internal fields

aryanorastar:
## What

`GET /v1/conversations/{id}/shared` is **public and unauthenticated** — anyone with the link gets the response. It is meant to publish the transcript/summary the owner shared, but it returned the **full `Conversation` model** with only `geolocation` stripped, also exposing:

- **`data_protection_level`** — the owner's server-side encryption tier (internal).
- **`external_data`** — merge provenance (**other conversation ids** the owner merged) and integration metadata.

The endpoint already nulls `geolocation`, which shows the intent was a curated public view — the other internal fields were just missed.

## Fix

Strip `data_protection_level` and `external_data` alongside `geolocation` before serializing — same pattern, same spot. Scope is deliberately narrow: only the unambiguously-internal fields. The transcript, summary, structured data, people and audio the owner chose to publish are unchanged, so the response schema (and its generated clients) is unchanged.

## Tests (hermetic)

- `test_shared_conversation_field_exposure.py` — the endpoint clears `geolocation` / `data_protection_level` / `external_data` before serialization. **1 passed.**

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10273?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/conversations.py
+++ b/backend/routers/conversations.py
@@ -1118,7 +1118,14 @@ def get_shared_conversation_by_id(conversation_id: str):
     if not visibility or visibility == ConversationVisibility.private:
         raise HTTPException(status_code=404, detail="Conversation is private")
     conversation = deserialize_conversation(conversation)
+    # This endpoint is public and unauthenticated. Strip fields that are internal
+    # to the owner and never part of the shared transcript/summary the user chose
+    # to publish: precise geolocation, the server-side encryption tier, and
+    # external_data (which carries merge provenance — other conversation ids — and
+    # integration metadata).
     conversation.geolocation = None
+    conversation.data_protection_level = None
+    conversation.external_data = None
 
     # Fetch people data for speaker names
     person_ids = conversation.get_person_ids()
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Thanks for the focused privacy fix — this looks like the right direction to me.

I reviewed the diff and ran the new targeted backend unit test locally:

` .venv/bin/python -m pytest tests/unit/test_shared_conversation_field_exposure.py -q `

Result: `1 passed`.

The change is narrow and correctly strips `data_protection_level` and `external_data` before `conversation_to_dict()` serializes the public unauthenticated shared-conversation response, alongside the already-stripped `geolocation`. The regression test covers that serialization boundary, which is the important part here.

I’m not formally approving because the current PR state is still blocked by the repo hygiene/formatting checks being cancelled/skipped in the reported status set, so this needs maintainer/check-state review before merge. From the code review side, I don’t see a blocking implementation issue.

---

kodjima33 (approved):
fix(conversations): stop public shared endpoint leaking owner-internal fields — security fix w/ test, 5/5

PR #10287, opened 2026-07-22
fix(backend): stop one unreadable sync file from discarding its whole batch

mdmohsin7:
## Problem

A sync batch uploads up to 20 recordings under a single job
(`nextSyncUploadBatch`, `batchLimit = 20` on the fresh lane). Two defects
combined to make one bad file destroy all of them:

1. `decode_files_to_wav` deleted **every already-decoded WAV in the batch** and
   raised `400` as soon as one file failed.
2. Both decoders threw away a perfectly good decoded prefix whenever the stream
   ended early — `decode_opus_file_to_wav` required `frame_count > 0 and not
   corrupt_stream`, and `decode_pcm_file_to_wav` required `not corrupt_stream`.

A WAL truncated by an app kill or battery death — the normal shape of an
interrupted `[len][opus]` writer — therefore failed the entire batch. The job
finalized as `sync_invalid_audio`, the client reverted all 20 recordings to
`miss`, and since the candidate filter has no `retryCount` gate it re-uploaded
the identical batch on every pass, deterministically failing each time until all
20 rendered as failed. Up to 19 intact recordings were never transcribed.

## Change

- **Per-file isolation.** An unreadable file is dropped on its own and the loop
  continues; the request only fails when *nothing* in the batch decoded. The
  `duration == 0` case is folded into the same path instead of discarding
  siblings.
- **Keep what decoded.** Both decoders now succeed when any audio was decoded,
  logging how much was kept. This is safe by construction: each decoder `break`s
  at the first bad frame and only ever writes frames that decoded cleanly, so
  the retained WAV contains no corrupt audio.

The DoS guard is unchanged — an implausible length prefix (4 GB, or above
`MAX_SYNC_FRAME_BYTES`) still stops *before* the claimed bytes are read, and the
tests assert `decode.call_count == 1` to lock that in.

## On the changed tests

12 existing tests asserted the previous "reject the whole file" behavior, so I
checked whether that was long-standing intent. `git log -S` shows `corrupt_stream`
and every `*_rejects_valid_prefix` test were introduced by the **same commit**
that introduced this bug (`d4252a6738`) — they codify that change rather than
prior intent, and discarding a decodable prefix contradicts that commit's own
"retain failed audio" goal. They are rewritten to assert the corrected contract
rather than removed.

New coverage: an unreadable file dropped while its siblings still sync, a batch
that fails only when nothing decodes, and a genuinely truncated tail file.

## Tests

- `test_sync_opus_decode.py` + `test_sync_pcm_decode.py`: 58 passed.
- `test_sync_v2.py`, `test_sync_record_usage.py`, `test_chat_quota_counting_router.py`,
  `test_optional_audio_codecs.py`: 193 passed.
- `scripts/scan_async_blockers.py`: 0 findings. `black --line-length 120` clean.

`bash test.sh` reports unrelated pre-existing failures in this environment
(`ImportError: google.cloud.translate_v3`, and `ENCRYPTION_SECRET` collection
errors). None of those files reference `utils/sync/files` or any decode function.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_sync_opus_decode.py
+++ b/backend/tests/unit/test_sync_opus_decode.py
@@ -224,6 +224,7 @@ try:
         MAX_SYNC_FRAME_BYTES,
         decode_files_to_wav,
         decode_opus_file_to_wav,
+        get_wav_duration,
         retrieve_file_paths,
     )
 finally:
@@ -314,8 +315,8 @@ class TestDecodeOpusFileToWav:
 
     # --- Corrupt stream boundaries ---
 
-    def test_one_corrupt_frame_in_middle_rejects_whole_file(self):
-        """A valid prefix plus a corrupt frame must not become partial success."""
+    def test_corrupt_frame_in_middle_keeps_decoded_prefix(self):
+        """Decoding stops at the bad frame; the frames already written survive."""
         klass, _ = _failing_decoder(fail_on={2})
         with tempfile.TemporaryDirectory() as d, patch('utils.sync.files.Decoder', klass):
             bin_path = os.path.join(d, 'test.bin')
@@ -324,8 +325,8 @@ class TestDecodeOpusFileToWav:
 
             result = decode_opus_file_to_wav(bin_path, wav_path)
 
-            assert result is False
-            assert not os.path.exists(wav_path)
+            assert result is True
+            assert get_wav_duration(wav_path) > 0
 
     def test_first_frame_corrupt_rejects_file(self):
         """A corrupt first frame has no recoverable decoded result."""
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
fix(backend): one bad sync file no longer discards whole batch, root cause+tests — 5/5

PR #10215, opened 2026-07-21
docs(api): refresh app-client OpenAPI contract for apps_results

skanderkaroui:
## Summary
The OpenAPI Contract workflow has been red on `main` since ~14:41 UTC today: PR #10111 (177117ad40) added `apps_results` to the MCP `SimpleConversation` model (inherited by `FullConversation`) but did not regenerate the committed app-client contract. Every open PR inherits the failing **Public Developer API contract** check.

This PR regenerates `docs/api-reference/app-client-openapi.json` (16 added lines — the `apps_results` field on both schemas). No code changes.

## Verification
- `backend/scripts/export_openapi.py --app-client --check` → `docs/api-reference/app-client-openapi.json is up to date` (was: "stale" before the regen; reproduced the exact CI failure locally first)
- Diff inspected: only the two `apps_results` schema additions, matching ed2ba41c5b.

Note: the check's error message says to run `export_openapi.py --write ../docs/api-reference/app-client-openapi.json`, which silently exports the **public** surface into the app-client file (surface defaults to `public`); the correct invocation needs `--app-client`. Follow-up issue to fix the message.

## Peer review
Independent agent review: APPROVE. Verified diff scope (1 file, 16 additive lines, no removals/auth changes), reran the contract check from the worktree (up to date), confirmed base == origin/main with nothing smuggled. Attribution corrected per review (#10111, not #10196).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10294, opened 2026-07-22
fix(backend): include surface flag in OpenAPI stale-contract hint

kodjima33:
## What
The `--check` failure hint in `backend/scripts/export_openapi.py` now includes the surface-selection flag, so following it verbatim regenerates the correct surface.

## Why
From #10217:

> Following that command verbatim silently exports the **public** surface (surface defaults to `public`) into the app-client file — 53k lines deleted, wrong contract — because surface selection is a separate flag, not inferred from the output path. The correct command is `backend/scripts/export_openapi.py --app-client --write`.
>
> Evidence: hit while preparing #10215 today; the wrong invocation produced a 17-path public spec in place of the 309-path app-client spec and `--check` (also without the flag) then validated the wrong surface as 'up to date'.

`check_spec` had no knowledge of which surface was being checked, so its `is stale` / `does not exist` hints always emitted the bare `--write <path>` command, which defaults to the public surface.

## Change
- `check_spec(path, generated, surface='public')` now prefixes the correct selector via a new `surface_flag()` helper: `--app-client ` for app-client, `--surface integration-public ` for integration-public, empty for public (default invocation unchanged).
- `main()` passes `args.surface` into `check_spec`.
- Regression test `test_stale_hint_includes_surface_flag_for_non_public_surfaces` asserts each surface's hint carries its selector and that public stays flag-free.

## Verified
- Exercised the real `check_spec` code path for all three surfaces plus the does-not-exist branch (via the pinned openapi runner venv): every hint carries the right flag; the public hint is byte-for-byte the previous default.
  - app-client → `... run backend/scripts/export_openapi.py --app-client --write <path>`
  - integration-public → `... --surface integration-public --write <path>`
  - public → `... export_openapi.py --write <path>` (unchanged)
- `tests/unit/test_openapi_contract.py` — 15 passed (including the new regression test).
- `black --line-length 120 --skip-string-normalization` clean on both files.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/export_openapi.py
+++ b/backend/scripts/export_openapi.py
@@ -985,12 +985,29 @@ def write_spec(path: Path, generated: str) -> None:
     path.write_text(generated)
 
 
-def check_spec(path: Path, generated: str) -> None:
+def surface_flag(surface: str) -> str:
+    """CLI flag that selects `surface`, empty for the default public surface.
+
+    Surface selection is a separate flag from the output path, so a hint that
+    omits it silently exports the public surface into another surface's file
+    (issue #10217).
+    """
+    if surface == 'public':
+        return ''
+    if surface == 'app-client':
+        return '--app-client '
+    if surface == 'integration-public':
+        return '--surface integration-public '
+    raise OpenAPIContractError(f'unknown OpenAPI surface: {surface}')
+
+
+def check_spec(path: Path, generated: str, surface: str = 'public') -> None:
+    flag = surface_flag(surface)
     if not path.exists():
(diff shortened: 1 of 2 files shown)


kodjima33:

PR #10311, opened 2026-07-22
release(desktop): bounded auto-retry of transient failed Beta qualification

Git-on-my-level:
## What

Codemagic's qualification handoff already retries *dispatch I/O* (`for attempt in 1 2 3`), but nothing recovers a **transient failed or cancelled qualification outcome** on the newest candidate — it sits non-live until someone manually re-dispatches qualification. If there are no new releasable changes, the newest valid candidate can stay stranded indefinitely.

This adds a scheduled, fail-closed retry planner that re-dispatches the trusted macOS qualification workflow for the newest immutable published candidate, only after a transient failed/cancelled qualification.

## How

The retry decision lives entirely in a tested script (`.github/scripts/desktop_beta_qualification_retry.py`) that re-derives candidate identity and qualification state from **server JSON** — a `gh release view` and the qualification workflow run list. On schedule the workflow supplies **no tag**, so no caller claim is trusted.

The schedule runs twice hourly (`13,43 * * * *`), offset from the `:17` candidate build, so a transient failure recovers within ~30 min without racing the hourly planner.

### Safety properties (all enforced by the decision script)

- **Tag-bound and exact-SHA** — only the exact newest `v*-macos` tag at its exact SHA is retried.
- **Newest valid candidate only** — published, non-draft, non-prerelease, carrying the canonical three assets (`Omi.zip`, `omi.dmg`, `desktop-smoke-result.json`), `channel: candidate`, `isLive: false`.
- **Bounded candidate age** (24h) and **bounded attempts** (3 total).
- **Never retries an in-progress or already-successful run.**
- **Deterministic product failures can't loop forever** — the attempt bound is the loop guard. (We deliberately do not classify transient-vs-deterministic from logs: a failed qualification step can be either, and mis-classifying a transient failure as deterministic would strand the newest candidate. Bounded retries cannot loop.)
- **Re-dispatches qualification only.** Promotion stays the single completed-success authority (`desktop_promote_beta.yml`). No Stable reach, no new secrets/environments/IAM, production state read-only.

Dispatch uses the **Omi Bot app token** (not the default `GITHUB_TOKEN`) so the retried qualification's completion still triggers the `workflow_run` promotion — a `GITHUB_TOKEN`-dispatched run does not spawn downstream workflow runs.

## Tests

- `.github/scripts/test_desktop_beta_qualification_retry.py` — 22 fault tests covering every decision branch (invalid candidate, age bound, no prior attempt, in-progress/successful/non-retryable, transient retry under bound, the deterministic loop guard at the bound, exact tag+SHA binding, no promotion/stable authority).
- Registered as a `checks-manifest` lane (`desktop-beta-qualification-retry`).

## Verification

```
.github/scripts/test_desktop_beta_qualification_retry.py        → 22 tests OK
.github/scripts/test_desktop_release_flow_contract.py           → 13 tests OK
.github/scripts/check-release-process-guards.py                 → passed
YAML (workflow + manifest)                                      → valid
```

The one-path desktop release contract still passes unchanged — the retry adds no second promotion authority; it only re-enters the qualification step, which is per-tag serialized and idempotent via immutable evidence.

## Notes

- Backend unit tests (`test_desktop_release_scripts.py`, `test_qualified_beta_promotion.py`) reference only unmodified files (`desktop_promote_beta/prod.yml`, `desktop_qualify_beta.yml`, `codemagic.yaml`); they need `pytest` + the backend env which isn't in this checkout, and are unaffected by this change.
- If the Omi Bot app lacks the `Actions: write` repository permission, the dispatch step fails closed (candidate stays non-live — safe) and needs a one-time permission grant, not a new secret.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10311?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -414,6 +414,12 @@ checks:
     platforms: ["macos"]
     reason: "Exact-source retries must keep stable absolute SwiftPM paths while rejecting stale provenance, symlinks, collisions, and incomplete entries"
 
+  - id: desktop-beta-qualification-retry
+    command: ["python3", ".github/scripts/test_desktop_beta_qualification_retry.py"]
+    triggers: [".github/workflows/desktop_retry_beta_qualification.yml", ".github/scripts/desktop_beta_qualification_retry.py", ".github/scripts/test_desktop_beta_qualification_retry.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "SCA-49 bounded automatic retry of transient failed/cancelled Beta qualification must stay tag-bound, exact-SHA, bounded, and never re-dispatch in-progress/successful runs"
+
 exempt:
   - path: ".github/scripts/check-desktop-auto-beta-candidate.py"
     reason: "release pipeline check requires published candidate artifacts and workflow inputs"
--- a/.github/scripts/desktop_beta_qualification_retry.py
+++ b/.github/scripts/desktop_beta_qualification_retry.py
@@ -1,0 +1,268 @@
+#!/usr/bin/env python3
+"""Decide whether to retry the newest desktop Beta candidate qualification.
+
+The retry only recovers a *transient* failed/cancelled qualification of the
+newest valid immutable published candidate. It never trusts a caller claim: the
+candidate identity, qualification run set, and conclusion are all re-derived
+from server-provided JSON (a GitHub release view and the qualification workflow
+run list). On schedule the workflow supplies no tag at all.
+
+Safety properties enforced here:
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
@cursor address review comment

PR #10332, opened 2026-07-22
Track 3 follow-up: rework the notch into a voice-first dynamic island

katipally:
This is a follow-up to my Track 3 notch redesign (#10247), taking it voice-first: hold push-to-talk and the notch becomes a spoken conversation instead of a chat, and text input moves back into the app.

## Where this stands (for maintainers)

This is the verbal-first take on the notch, built on my #10247 redesign, so it carries that redesign plus the voice layer on top. #10247 is the chat-first version; this is the voice-first one, so you can try the two directions side by side. Every push-to-talk feature still works; what changed is that the notch now *shows* the voice turn instead of opening a chat. The new work here is the voice layer, the `Notch/NotchVoice*` files, the morphing orb, the paced reveal and linger, and the audio-reactive waveforms.

## The idea

A voice turn has a shape of its own, you talk, it thinks, it talks back, and the notch can *be* that shape. Chat used the island as a scroll view; voice uses it as the turn itself.

So the Omi mark stops being a static logo and becomes the whole interaction. The same eight dots are a live mic waveform while you talk, reform into the rotating logo ring while Omi thinks, then a waveform that moves with Omi's own voice while it speaks. It never swaps one icon for another, it morphs in place, and the black island grows out of the camera housing to hold whatever the turn needs.

## Demo (under 2 minutes)

https://github.com/user-attachments/assets/bbdaa086-d61e-4f46-812a-65256a5696fe

## What changed, and why it feels right

- **One mark, three states, no cross-fade.** The logo, the mic waveform, and the thinking spinner are the same eight dots morphing between a ring and a bar layout, rendered once as a persistent layer so a phase change animates the geometry instead of swapping views. The ring uses the exact `NotchOmiMark` ratios, so at rest it *is* the logo.
- **Waveforms that actually react.** Listening reads the live mic level. Speaking reads a new `playbackLevel` computed from the RMS of each streamed TTS chunk in `StreamingPCMPlayer`, so the bars move with Omi's voice, not a canned loop.
- **The island grows, it doesn't pop.** Width is fixed; only the height grows as the reply wraps, capped at 40% of the screen. Longer replies scroll and follow the newest line only once they overflow, so adding a line never jumps.
- **The reply reads at a speaking pace.** Rather than dumping tokens as fast as the model streams them, the reply reveals a word at a time at roughly speaking speed and finishes in place, so the words track the voice and the stream-to-final handoff is seamless.
- **It waits for you.** When Omi finishes, the reply lingers a few seconds so you can read it, resting as the Omi logo. Hover holds it, Esc dismisses it, and tapping it (or the "Open in Omi" hint that fades in) opens the full conversation in the main window, where text lives now.
- **Barge-in just works.** Push-to-talk while Omi is speaking interrupts and starts a new turn; the notch flips straight back to listening. This was already in the realtime hub, the notch just reflects it now.
- **Quiet by default.** A soft cue on turn start (already there) and a matching one when a turn completes. On the closed notch, hover does nothing, the mark opens the app, the gear opens settings.

## What I dropped, on purpose

- **Chat in the notch.** That's the point, text input moves back to the app. The chat views are left dormant in-tree so removing them can be its own small PR rather than noise here.
- **A live in-hold transcript.** I tried showing your words as you speak, but the STT providers in use don't reliably emit partial transcripts during the hold, so it appeared inconsistently. Rather than ship something that works for some setups and blanks for others, listening is a clean "Listening…" and the words appear in the reply phase. The plumbing is still there to turn back on per-provider.

Everything else from the PTT lifecycle is untouched: hub and batch turns, voice playback, the Ask-Omi hotkey (now opens the app), the recording indicator, usage limiting, and the automation bridge.

## Product invariants

- **INV-CHAT-1**: the notch stays an I/O surface over the one kernel transcript. The reply is the display-only `liveVoiceAssistantText` mirror, never a second store; the paced reveal and the linger are presentation over that mirror; sends and journaling are unchanged.
- **INV-VOICE-1**: the PTT reducer and journal lifecycle are untouched. This is a presentation change plus display-only signals (`playbackLevel`, and a non-prod usage-limit test bypass). `VoiceTurnCoordinator` gains exactly one side effect, a soft end cue on `terminal(.success)`.
- **INV-UI-1**: no purple anywhere; white and neutral on black.
- **INV-AUTH-1 / INV-DATA-1**: no session-ownership or routing changes.
- **INV-INT-1**: no change to the integrations surface. The connector / memory-export files in the diff are carried unchanged from #10247; the voice layer adds nothing to them.

Failure-Class: none

The `fix:` commits are UI-polish corrections inside this new feature (flicker, alignment, spacing, reveal pacing), not repairs of a shipped failure-class boundary.

## Testing

- Full desktop Swift suite green (353 isolated suites). `agent-logic-harness --swift-only` green (PTT lifecycle/state).
- New hermetic tests: `NotchPresentationLadderTests` (the responding state and ladder ordering), `NotchViewModelTests` (voice sizing, the 40% clamp, the linger lifecycle with an injected clock, hover pause/resume, Esc dismiss), `ReplyRevealModelTests` (whole-word paced reveal, completion, restart), `FloatingBarUsageLimiterTests` (the test bypass).
- `swift-format` lint and SwiftLint clean. Line-count ratchet updated (bridge raise justified, `PushToTalkManager` ratcheted down after the cue refactor).
- Exercised on a named bundle (`omi-notch-voice`) and captured frame by frame: closed chrome, the listening waveform, the rotating-logo thinking ring, the paced justified reply, the 40% cap and scroll, the lingering reply with the open-in-app hint, and the blocked-PTT hint. The demo above is a real hold-and-speak turn end to end.
- All 19 preflight checks pass.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
@@ -3,7 +3,7 @@
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": 2375,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2818,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4515,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2423,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2413,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1621
   },
   "raise_justifications": {
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
@@ -2,7 +2,7 @@
   "files": {
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": 2375,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2818,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4643,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4515,
     "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2423,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1621
   },
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
@@ -1,8 +1,8 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": 2375,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2853,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4842,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2818,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4643,
     "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2423,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1621
   },
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-mainwindow.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-mainwindow.json
@@ -1,6 +1,6 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": 1926,
+    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": 1930,
     "desktop/macos/Desktop/Sources/MainWindow/Pages/AppsPage.swift": 3643,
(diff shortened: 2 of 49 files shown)


Git-on-my-level:
Thanks for the detailed follow-up and demo. I did a static maintainer pass on this head.

This is a cohesive direction around the macOS notch/voice surface, and the implementation is thoughtfully structured: the new notch presentation/view-model split is much easier to reason about than embedding all of the states in the legacy floating-bar window, and the added pure tests around geometry, presentation ladder, visibility, and reply reveal are good coverage for the pieces that can be tested without a full UI run.

I’m not going to formally approve this from automation, though. This changes a core desktop interaction model: the notch becomes voice-first, text entry moves back into the main app, the panel now owns per-display high-level always-on windows, and non-production automation/debug bridge behavior is extended for this flow. Those are product/UX and desktop-behavior decisions that need a human maintainer to try on-device and sign off before merge.

Maintainer review should especially validate:
- whether the voice-first notch direction is the intended product direction versus the chat-first version in the stacked/prior work;
- multi-display, fullscreen/Spaces, permission-dialog, and third-party overlay behavior with the new `NotchWindow` level and per-display `NotchScreenManager`;
- the debug bridge additions (`reset_usage_limit`, `send_main_draft`, notch capture fallback) remain strictly non-production and match the repo’s automation expectations;
- a real macOS build/UI pass, since the only reported check on this head is the neutral Cubic reviewer and I could not exercise the AppKit/SwiftUI runtime from this environment.

No security/supply-chain issue stood out in the static diff I reviewed. This looks like a serious and useful prototype/follow-up, but it needs human product and desktop-runtime validation before it should be merged.

---

aryanorastar:
Closing as a maintainer.

This is still CONFLICTING on 08f8aba (22 Jul). The conflict list is the shipping notch and PTT stack: FloatingControlBarWindow/View/State, PushToTalkManager, StreamingPCMPlayer, ChatProvider, ChatBubble, ChatMessagesView.

Main did not take the island rewrite. The notch still lives in FloatingControlBarWindow. Ask Omi already opens the main chat; the notch is not a text surface. NotchVoiceMorphMark is already in that bar. PTT since this tip grew capture routing, dictation vs question, and admission — this branch still edits the July managers.

It is also stacked on #10247 (chat-first island, still OPEN and CHANGES_REQUESTED). The full-app sibling #10672 was already closed: the whole redesign is too much; cherry-picks belong on the current shell. Rebasing this 48-commit voice-first fork would resurrect the island.

Do not bring NotchScreenManager or a second notch transcript UI back on top of the current floating bar.

PR #10359, opened 2026-07-22
fix(desktop): restore capture services on launch

Git-on-my-level:
## Summary

- Restore Listening immediately from its persisted enabled intent; local transcription no longer waits for remote API-key readiness.
- Reconcile Capture when server-authoritative assistant settings arrive after launch, preserving the user preference and starting only after Screen Recording is granted.
- Add policy and settings-sync regression coverage for the launch restoration boundary.
- Repair the desktop CI suites by deleting assertions for intentionally retired startup/notch paths and resetting Rewind storage between owner-isolated lifecycle tests.

## Root cause

Launch used one-shot readiness checks. Listening was deferred behind API-key availability even when the local transcription provider could run, while a delayed settings sync could update Capture intent after its only launch scheduling attempt.

The failing CI run still asserted two deleted implementation paths, and one Rewind test changed database owners without resetting the storage owner that owns the shared video encoder.

## Validation

- `xcrun swift test -c debug --package-path desktop/macos/Desktop --filter PersistedCaptureLaunchPolicyTests`
- `xcrun swift test -c debug --package-path desktop/macos/Desktop --filter SettingsSyncCaptureRestorationTests`
- `desktop/macos/scripts/swift-test-suites.sh`
- `desktop/macos/scripts/run-swift-ci.sh --test` (blocked locally: requires unavailable Xcode 16.4; GitHub uses the pinned toolchain)
- Named QA bundle `omi-launch-permission-qa` launched to the signed-in Home surface; its intentionally separate bundle identity correctly retained capture intent without prompting when Screen Recording was absent.

Failure-Class: none

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
@@ -4,6 +4,21 @@ import SwiftUI
 
 private struct AnySendableBox: @unchecked Sendable { let value: Any? }
 
+/// Decides whether a persisted capture intent needs its runtime service restored.
+///
+/// Intent is stored independently from the running services. A fresh launch and a
+/// settings sync therefore both need to reconcile the two states instead of using
+/// a one-time readiness check as the source of truth.
+enum PersistedCaptureLaunchPolicy {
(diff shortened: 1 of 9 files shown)


Git-on-my-level:

PR #10360, opened 2026-07-22
fix(desktop): onboarding reset must not hard-delete cloud chat history

Archit-lal:
## Problem

Restarting onboarding **permanently hard-deletes the user's entire cloud chat history**.

`clearDefaultJournalForOnboardingReset()` (called during onboarding reset, with **no production guard**) routes the default main-chat clear through the kernel journal's delete outbox. For the `default` surface, `backendTargetForConversation` resolves `target_kind = "messages"`, which drains to `DELETE /v2/desktop/messages` with **no session** — a batched hard delete of every main-chat message for that user, server-side, with no tombstone.

Verified live: an account whose production desktop journal still held **82** turns showed only **2** on the backend (and in every fresh/renamed build). The backend copy was wiped; repeated onboarding-reset iterations drove it down. Reconcile then faithfully mirrors the emptied backend to every install.

## Fix

A **reset is not a delete.** Resetting local app state must never destroy server-side chat history.

- Daemon (`conversation-journal.ts`): `clearJournalConversation` gains a `deleteBackend` flag (**default `true`**, so the explicit user *"Clear history"* action is unchanged). When `false`, the local journal is fenced and its turns are purged, but `enqueueBackendConversationDelete` is **not** called — no backend delete is enqueued.
- Plumbed through the `journal_clear_turns` RPC (`protocol.ts`, `index.ts`) and the Swift clear seam (`KernelTurnProjection` → `AgentClient` → `AgentBridge` → `AgentRuntimeProcess`), all defaulting to `true`.
- `clearDefaultJournalForOnboardingReset()` now passes `deleteBackend: false`. The reset button's user-visible behavior is unchanged (local view clears, app restarts into onboarding); the backend stays authoritative and rehydrates the thread via reconcile.
- Harness resets are intentionally unchanged — they run under synthetic owners on non-production bundles for E2E test isolation.

## Why not more

Reconcile is already backend-authoritative and continuous (an install shows exactly what the backend holds), so no rehydrate changes are needed — stopping the wipe is the fix. Recovering already-lost history for affected accounts is handled separately as a controlled, idempotent re-POST, not shipped as an all-user migration.

## Tests

- **Daemon:** local-only clear purges local turns and enqueues **0** `backend_conversation_delete_outbox` rows; the explicit clear still enqueues exactly **1**. (`conversation-journal.test.ts`)
- **Swift:** `clearDefaultJournalForOnboardingReset` propagates `deleteBackend=false` through the clear seam. (`KernelTurnRecordedProjectionTests.swift`)
- Full daemon suite and the projection suite pass locally; `tsc` clean; desktop package builds.

## Verification

- `cd desktop/macos/agent && npx vitest run tests/conversation-journal.test.ts` → 42 passed.
- `xcrun swift test --package-path Desktop --filter KernelTurnRecordedProjectionTests` → 31 passed (incl. new test; RED confirmed before the fix: `[true] != [false]`).
- `node_modules/.bin/tsc --noEmit` → exit 0.
- Desktop package build via pre-push → `Build complete!`

- INV-CHAT-1 — one shared transcript across surfaces. This fix upholds it: the backend transcript is the single shared source and a reset no longer destroys it.
- INV-AUTH-1 — desktop Firebase session truth. The clear path stays owner-scoped; only the local journal for the active owner is cleared.
- INV-AGENT-* — agent-runtime control plane. The `journal_clear_turns` RPC gains a `deleteBackend` flag with no change to ownership or lifecycle.
- INV-DATA-1 — production-family customer data-plane continuity. Inert: the flag only gates a local-journal clear; it changes no data-plane/backend routing, and the default (`true`) preserves existing delete behavior.
- INV-BETA-1 — beta admission/identity. Inert: no release/qualification/identity behavior is touched.
- INV-INT-1 — integrations contract. Inert: no external integration surface is changed.
- INV-MEM-1 — three product memory tiers. Inert: chat-journal clear scope only; no memory-tier behavior is changed.

## Failure class (fixes)

A local UI reset (onboarding re-walkthrough) was acting as a competing mutation authority over backend-owned chat state, hard-deleting it. The fix converges chat-history authority on the backend: a reset clears only the local projection, never the server transcript.

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-chat.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-chat.json
@@ -1,11 +1,11 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": 2209,
-    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": 4314
+    "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": 2213,
+    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": 4318
   },
   "raise_justifications": {
-    "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": "per-turn reasoningEffort threaded through both query variants",
-    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": "reasoningEffort added to the query wire message"
+    "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": "deleteBackend flag threaded through both clearJournalTurns variants so a reset is local-only",
+    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": "deleteBackend added to the journal_clear_turns wire message"
   },
   "threshold": 1500
 }
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
@@ -1,10 +1,10 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6301,
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6304,
     "desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift": 2861
   },
   "raise_justifications": {
-    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": "SB redesign chat-surface work (single continuous default chat) plus ChatTurnOwner reasoning-effort lane derivation; a component split is tracked follow-up.",
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": "Onboarding reset is now local-only (deleteBackend: false) so it no longer hard-deletes cloud chat history; a component split is tracked follow-up.",
     "desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift": "Already-granted permissions no longer reopen System Settings: screen-recording and full-disk-access requests gate their Settings/drag-card opens behind the granted result."
   },
   "threshold": 1500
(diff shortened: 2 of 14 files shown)


skanderkaroui:
Reviewed the full diff + traced the flag end-to-end. **Core fix is correct** — the P0 is closed at the root (`clearDefaultJournalForOnboardingReset` now local-only), the `deleteBackend` flag threads cleanly through every layer with correct missing-means-true semantics, sign-out/account-teardown don't touch this path, and the backend-deleting default keeps full test coverage. 👍 on the direction.

Two things I'd want before merge:

**1. (must-fix) The local-only clear still destroys un-delivered turns.** In `clearJournalConversation`, `deleteBackend:false` only skips the backend delete — it still runs `DELETE FROM backend_turn_outbox WHERE status != 'delivered'` + `DELETE conversation_turns`. Turns queued `pending`/`retrying` (typed offline / mid-retry, **never POSTed**) get no claim tombstone and no backend copy, so reconcile can't rehydrate them. Chat offline → restart onboarding → those messages are gone. This is a hole in the PR's own "backend authoritative, nothing lost" promise. Suggest draining/awaiting the outbox before the local purge, or preserving un-delivered rows.

**2. (nit, 2-min) Stale comments now contradict the behavior.** `clearDefaultJournalForOnboardingReset`'s header and the `AppState+SystemActions.swift` call-site comment still describe a "durable, generation-fenced backend DELETE outbox" — directly opposite the new inline comment 3 lines down. A future editor trusting them re-introduces the hard delete.

Non-blocking, worth considering:
- The destructive `target_kind="messages"` delete-all is still the *default* for every caller — guarded only by remembering `deleteBackend:false`. This class has recurred a few times; the design doc's own "a local-only clear that reaches enqueueBackendConversationDelete is a bug" guardrail would make it structural. The added test is happy-path only.
- Orphaned `cleared_backend_turn_claims` ('waiting') rows: only GC'd via the ack path, which never runs when `deleteBackend:false` → slow accumulation for the stable `default` conversation.
- Recovery of already-lost history (the 79→2 turns) is deferred to Parts B/C with no tracking issue — worth filing one so it isn't lost.

Happy to pair on #1 if useful.

Issue #10352, opened 2026-07-22
make setup and git hooks fail in linked worktrees: 'this operation must be run in a work tree', empty repo-root in hook paths


kodjima33:
In a linked git worktree (e.g. Conductor workspace at `~/conductor/workspaces/omi/<name>` whose gitdir lives under the main clone's `.git/worktrees/`), both `make setup` (setup-hooks step) and the installed pre-push hook fail:

```
fatal: this operation must be run in a work tree
bash: /scripts/dev-harness/_resolve_python.sh: No such file or directory
bash: dev_harness_python: command not found
make: *** [setup-hooks] Error 128
```

The empty prefix in `/scripts/dev-harness/...` shows a repo-root resolution returning empty in this layout (likely `git rev-parse --show-toplevel` executed from the wrong directory, or a `--git-common-dir`-relative path assumption). Pushing from such a worktree requires `--no-verify`, silently bypassing the local gate. AGENTS.md mandates worktrees for all code changes, so this hits every agent working from a secondary workspace mount. Repro: `git worktree add` from a clone, run `make setup` inside the new worktree with a different filesystem prefix than the main checkout.

PR #10362, opened 2026-07-22
fix(dev): resolve repo root without git --show-toplevel in linked worktrees (#10352)

kodjima33:
## Bug (#10352)

In a linked git worktree whose gitdir lives under the main clone's `.git/worktrees/` (e.g. a Conductor workspace at a different filesystem prefix), `git rev-parse --show-toplevel` fails with `fatal: this operation must be run in a work tree` and returns empty stdout. That empty value flowed into:

- `make setup` → `source "/scripts/dev-harness/_resolve_python.sh"` → `No such file or directory` / `dev_harness_python: command not found`
- the installed pre-push/pre-commit dispatcher → empty `ROOT`, so pushing required `--no-verify`, **silently bypassing the local gate**.

AGENTS.md mandates worktrees for all code changes, so this hit every agent on a secondary workspace mount.

## Root cause

Three call sites resolved the repo root via `git rev-parse --show-toplevel`, which is not robust in these worktree layouts.

## Fix

Resolve the repo root from each caller's own on-disk location instead:
- **Makefile**: `REPO_ROOT` from `MAKEFILE_LIST` (the Makefile's own path).
- **scripts/install-git-hooks.sh**: `ROOT` from `BASH_SOURCE`.
- **installed dispatcher**: keep `--show-toplevel` for the normal case but fall back to `$PWD` (git runs hooks from the worktree root) so `ROOT` is never empty.

## What I tested

- Added `scripts/test-worktree-setup-hooks.sh`: a hermetic regression that injects a `git` shim making **only** `rev-parse --show-toplevel` fail, and asserts hook install, dispatcher runtime resolution, and Makefile `PYTHON` resolution all still work. Wired into `.github/checks-manifest.yaml` (local + ci).
- New test passes; **reverting the fix makes it fail (exit 128)**.
- Existing `scripts/test-make-setup.sh` and `.github/scripts/test_run_checks.py` pass.
- Normal-case `make` in a real worktree resolves `REPO_ROOT` and `PYTHON` correctly.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10362?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -24,6 +24,11 @@ checks:
     triggers: ["Makefile", "backend/scripts/sync-python-deps.sh", "scripts/pre-push", "scripts/test-make-setup.sh", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "make setup must provision the minimal, rerunnable backend environment required by selected pre-push checks"
+  - id: worktree-setup-hooks
+    command: ["bash", "scripts/test-worktree-setup-hooks.sh"]
+    triggers: ["Makefile", "scripts/install-git-hooks.sh", "scripts/test-worktree-setup-hooks.sh", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "make setup and the installed hooks must resolve the repo root without `git rev-parse --show-toplevel`, which fails in linked-worktree layouts (Conductor) and silently bypassed the local pre-push gate (#10352)"
   - id: check-manifest-contract
     command: ["python3", ".github/scripts/test_run_checks.py"]
     triggers: ["all"]
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,12 @@
 HOOKS_DIR := $(shell git rev-parse --git-path hooks)
-PYTHON ?= $(shell bash -c 'source "$$(git rev-parse --show-toplevel)/scripts/dev-harness/_resolve_python.sh"; dev_harness_python')
+# Resolve the repo root as $$(pwd), not `git rev-parse --show-toplevel`: that
+# command fails ("this operation must be run in a work tree") in some linked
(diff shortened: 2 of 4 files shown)


kodjima33:
Closing as superseded — the fix landed on `main` by another route and issue #10352 is closed as completed.

Current `main` already resolves the repo root without depending on `git rev-parse --show-toplevel`:
- `Makefile`: `$(git rev-parse --show-toplevel 2>/dev/null || pwd)`
- `scripts/install-git-hooks.sh`: `ROOT` from `BASH_SOURCE`
- installed dispatcher: `$(git rev-parse --show-toplevel 2>/dev/null || pwd)`

Verified against `main` (a2ca2c3c) with a `git` shim that makes **only** `rev-parse --show-toplevel` fail (exit 128, empty stdout) — the exact reported failure mode:
- hook install succeeds (`pre-commit`/`pre-push` written)
- the installed dispatcher execs `$ROOT/scripts/pre-commit` with the correct worktree root (no empty `ROOT`, so the local gate is not silently bypassed)
- `make` resolves `PYTHON=python3` instead of the empty-prefix `/scripts/dev-harness/... No such file`

This PR's remaining unique content is the regression test (`scripts/test-worktree-setup-hooks.sh` + its `.github/checks-manifest.yaml` entry), which did not come along with the landed fix. Its dispatcher assertion (`grep 'ROOT="\$PWD"'`) no longer matches main's `|| pwd` form, so it would need adapting rather than a straight rebase. Not carrying it here; worth a separate test-only PR if someone wants the guard.

Issue #10198, opened 2026-07-21
Omi cv1 Android: convos missing but appear using filter


hugoaap-code:
Conversations screen won't show conversations from today. Also several conversations missing. They do show when I use the date filter. 

![Screenshot_20260721_064849_Omi.jpg](https://github.com/user-attachments/assets/dd76a401-a646-49ed-970c-a27ecb113bd1)

aryanorastar:
Root-caused this — it's a **UTC-vs-local day-grouping** bug, not data loss. The conversations exist (which is why the date filter finds them); they're just filed under the wrong day.

`started_at` / `created_at` arrive from the API as UTC (ISO-8601 with `Z`), and `ServerConversation.fromGenerated` keeps them as UTC `DateTime`s. The conversations list then buckets into day-groups using those timestamps' **raw UTC** calendar day — `app/lib/providers/conversation_provider.dart`, `_buildGroupedByDate` (and the sibling groupers) do:

```dart
final effectiveDate = conversation.startedAt ?? conversation.createdAt;      // UTC
final date = DateTime(effectiveDate.year, effectiveDate.month, effectiveDate.day); // UTC day
```

But the date **filter** builds its range from the **local** `selectedDate` (`_getDateFilterRange`): `DateTime(selectedDate.year, .month, .day, 0,0,0) … 23:59:59` — a local-day window.

So for any viewer east of UTC, an early-morning-local conversation is still the *previous* day in UTC and gets filed under "Yesterday" — absent from "Today" — while the local-day filter still returns it. Example: 06:48 local on 2026-07-21 at UTC+10 = 20:48 UTC on 2026-07-20. The screenshot's 06:48 capture time fits a positive-UTC-offset device exactly.

**Fix direction:** bucket by the **local** calendar day (`.toLocal()` before truncating to y/m/d) at the grouping sites, so grouping agrees with the local-day filter and the Today/Yesterday labels. I have this implemented (one shared `localDayKey` helper applied at each bucketing site) and can open a PR.

One caveat for whoever picks it up: the divergence isn't hermetically unit-testable — `DateTime.toLocal()` uses the ambient timezone, which is UTC in CI, so UTC-day == local-day there. On-device confirmation in a UTC+ timezone (or threading the user's UTC offset through an injectable helper) is the real proof.

PR #10398, opened 2026-07-23
fix(app): group conversations by local calendar day, not raw UTC (#10198)

aryanorastar:
## What

Conversations from today were missing from the conversations list but appeared under the date filter (#10198).

## Root cause

`started_at`/`created_at` arrive from the API as UTC (ISO-8601 `Z`) and `ServerConversation.fromGenerated` keeps them UTC. The conversations list bucketed into day-groups by the timestamps' **raw UTC** calendar day (`_buildGroupedByDate` and the sibling groupers in `conversation_provider.dart`), while the date **filter** built its range from the **local** `selectedDate` (`_getDateFilterRange`).

So for any viewer east of UTC, an early-morning-local conversation is still the *previous* day in UTC → filed under "Yesterday", absent from "Today" — while the local-day filter still returns it. The report's 06:48-local screenshot fits a positive-UTC-offset device exactly.

## Fix

Bucket by the **local** calendar day via a single shared `conversationLocalDayKey(ts) => ts.toLocal()` day-key, applied at every grouping site (main list, search, memory groupers, index lookup). Sorts stay on raw timestamps (ordering is timezone-independent). The already-local group-key normalizers (`getConversationIndexById`, `getAdjacentConversation`) are unchanged.

## Tests

`app/test/providers/conversation_local_day_key_test.dart` — locks the helper's local-day-truncation contract and same-day/next-day bucketing.

## Verification — please read

I don't have `dart`/`flutter` in my environment, so I could **not** compile or run the analyzer/tests locally — this PR relies on the **Dart Analyze & Tests** CI check to verify compile + tests. The change is mechanical (a shared `.toLocal()` day-key swapped in at each bucketing site).

And an honest limit on the test's reach: `DateTime.toLocal()` uses the ambient timezone, which is **UTC in CI**, so the unit test can't reproduce the UTC-vs-local divergence there — it locks the contract and catches a raw-UTC revert on a non-UTC host. The end-to-end proof is on-device in a UTC+ timezone.

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/providers/conversation_provider.dart
+++ b/app/lib/providers/conversation_provider.dart
@@ -23,6 +23,18 @@ typedef ConversationSearchFetcher = Future<(List<ServerConversation>, int, int)>
   String? speakerId,
 });
 
+/// Day-bucket key for a conversation timestamp, in the viewer's **local** timezone.
+///
+/// `started_at`/`created_at` arrive as UTC (ISO-8601 `Z`), so bucketing by their raw
+/// UTC `year/month/day` filed an early-morning-local conversation under the previous
+/// day for any UTC+ viewer — it vanished from the "Today" group even though the
+/// local-day date filter still found it (#10198). Truncating the *local* calendar day
+/// keeps grouping consistent with the local-day filter and the Today/Yesterday labels.
+DateTime conversationLocalDayKey(DateTime timestamp) {
+  final local = timestamp.toLocal();
+  return DateTime(local.year, local.month, local.day);
+}
+
 class ConversationProvider extends ChangeNotifier {
   List<ServerConversation> conversations = [];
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Bug fix (#10198): group conversations by local calendar day; scoped, test added, CI green — 5/5

PR #9304, opened 2026-07-09
 feat(desktop/windows): coding-agent ACP adapter core (PR 1/4)

karthikyeluripati:
Part 1 of a stacked train (see issue #9302):

Self-contained ACP client under src/main/codingAgent/ ported from desktop/macos/agent/src/adapters — nothing wired into chat/IPC/UI yet. Windows-specific: env allowlist (ComSpec/SystemRoot/PATHEXT), taskkill /t /f tree-kill (no POSIX process groups), windowsHide, ELECTRON_RUN_AS_NODE for the bundled Claude Code bridge.

Verification: pnpm run typecheck + lint clean; pnpm test 577 passed, incl. a real-subprocess integration test (fake ACP peer fixture). Also verified live: the real bridge created a file through this adapter on a Windows 11 machine.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #9302, opened 2026-07-09
Windows desktop: delegate tasks to coding agents (Claude Code, OpenClaw, Hermes, Codex) via ACP


karthikyeluripati:
The Windows app can only answer through Omi's hosted chat today. This adds the macOS agent runtime's adapter concept to desktop/windows/: a full ACP (Agent Client Protocol) JSON-RPC-over-stdio client with Claude Code built in (bundled @zed-industries/claude-agent-acp bridge, no CLI install) and OpenClaw / Hermes / Codex as user-connected external commands. Saying "ask codex to fix the failing test in my omi repo" (typed or push-to-talk) routes the task to the named agent and streams its progress into chat; if it fails to start, Omi falls back to the next connected agent; if it isn't connected, the reply explains how to install it (Settings → Agents has install cards + a real handshake Test). External agents run with a minimal allowlisted env and never get automatic permanent permission grants.

I'll split this into stacked PRs to make review easier: 1) adapter core, 2) routing/fallback/IPC, 3) settings UI, 4) docs + fixes found while demoing live.

karthikyeluripati:
Fair concern @ofekron  — worth clarifying what the current implementation actually does: fallback is
never silent. The chat stream emits an explicit agent-switch event, shows "X failed
(reason) — trying the next agent…", and the reply header changes to "Y took over the
task". Fallback also only triggers when the requested agent produced zero output, so a
partially-answered task is never quietly re-run elsewhere.

The stronger version of your point — asking before switching, or a saved fallback policy
(auto / ask / never) — is a real product decision, and it sits in the same bucket as the
ACP permission-policy question maintainers are already reviewing on #9304. If they want
the conservative default, a per-user fallback policy is a small addition to the Settings →
Agents PR in this stack. Run identity (provider, cwd, cost) is already carried per task in
the result events.

karthikyeluripati:
Status wrap-up: PR 1/4 (#9304) merged, and the remaining planned slices have since landed
on main directly via maintainer work built on that adapter core — now including OAuth
sign-in, agent auto-detection, pi-mono, and packaged-agent smoke coverage, shipped in
v1.0.2-windows. I've closed my fork-staged PRs 2–4 as superseded. One last gap from the
#9304 review cycle (install-time verification of the optional claude.exe binary) is up as
a small follow-up PR: #10399 . From my side this issue is complete.

Git-on-my-level:
[...]

PR #10399, opened 2026-07-23
fix(desktop/windows): check-agent-deps also verifies the Claude Code …

karthikyeluripati:
Follow-up to #9304 / #9302, closing the one gap from that review cycle still open on main.

`scripts/check-agent-deps.mjs` verifies the coding-agent JS packages resolve, but the
Claude Code **executable** (`claude.exe`, ~235 MB — spawned by the Claude Agent SDK for
every agent turn; SDK 0.3.205 ships no JS fallback) lives in the binary-only
optionalDependency `@anthropic-ai/claude-agent-sdk-win32-x64`: a failed download leaves
`pnpm install` exiting 0, and `import.meta.resolve()` can't see a package with no JS
entry. Result: an app where the ACP initialize handshake succeeds and every real agent
task fails to spawn. This exact failure shape occurred live during the #9304 review
(flaky network → error 23 → empty residual `.pnpm` directory → exit 0), and the reviewer
validated this check's logic against that broken tree at the time.

- `scripts/claude-binary-check.mjs` (pure, unit-tested): locates `claude.exe` in the
  flattened or pnpm virtual-store layout — requiring the binary itself, so an empty
  residual store directory doesn't pass — and rejects truncated (<50 MB) downloads.
- `check-agent-deps.mjs` reports it through its existing missing-deps path (so it runs
  at postinstall and standalone, same as today). Non-Windows installs skip.

## Test plan
- [x] `node scripts/check-agent-deps.mjs` on a clean pnpm@10 hoisted install — OK,
      finds the real 236 MB binary
- [x] `vitest run scripts/claude-binary-check.test.mjs` — 6 passed (absent package,
      empty residual virtual-store dir, truncated download, both layouts, non-Windows skip)
- [x] Full postinstall chain (incl. verify-pimono-unpack) green with the check wired in
- [x] eslint clean

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/scripts/check-agent-deps.mjs
+++ b/desktop/windows/scripts/check-agent-deps.mjs
@@ -22,6 +22,7 @@
 import { existsSync } from 'node:fs'
 import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
+import { checkClaudeBinary } from './claude-binary-check.mjs'
 
 const CRITICAL_DEPS = [
   '@agentclientprotocol/claude-agent-acp', // dynamically imported by src/main/codingAgent/claude-acp-entry.mjs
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
Windows check-agent-deps fix — approved; held from merge on red Hygiene check

Git-on-my-level:
Thanks for tightening this Windows agent install guard. The implementation looks well-scoped: the binary check is side-effect-free/testable, covers both flattened and pnpm virtual-store layouts, catches empty residual store dirs and truncated downloads, and skips non-Windows installs. I also ran the targeted Vitest suite locally against this head and it passed (6/6).

I’m not formally approving because this touches the Windows postinstall/agent dependency path and the current Hygiene check is red. The failing guard is `failure-class-protocol`: the `fix(...)` commit needs a `Failure-Class: FC-<slug> | new | none` declaration where the workflow validator reads it (the PR body has `Failure-Class: none`, but CI reports `declaration: null`). Once that metadata/CI issue is fixed, this looks like a good candidate for maintainer sign-off.

---

PR #10422, opened 2026-07-23
perf(chat): reuse Anthropic context prefixes

### LINKED ISSUE #10418: Optimize Anthropic chat caching for sliding history and agentic tool loops
## Problem

Production Anthropic accounting shows that Omi chat prompt caching is functioning for the stable prefix, but a small heavy-tail population repeatedly sends very large uncached contexts. The gateway preserves Anthropic `cache_control`; this is a caller/context-management problem rather than a gateway passthrough failure.

Two current behaviors combine poorly:

1. Chat loads a fixed sliding window of the latest 10 stored messages. Once the chat is full, each new user/assistant pair removes older messages from the front, so the historical-message prefix changes on every turn.
2. Agentic chat places a one-hour cache breakpoint on the system block but no breakpoint on the pre-loop message history. Each tool-loop iteration resends the base history plus accumulated tool calls/results.

Anthropic caches cumulative prefixes in `tools → system → messages` order. A moving first historical message prevents inter-turn reuse of a message-level cache entry, while repeated tool rounds can re-bill an increasingly large uncached suffix.

This exact sliding-window concern was previously documented in #4692. It was closed when the expected benefit was small under the workload at that time. Current Anthropic production distribution shows a materially larger agentic-tail problem and justifies revisiting it.

## Generalized production evidence

Privacy-safe aggregation over a matched post-rollout sample of **749 Anthropic attempts** found:

| Signal | Distribution |
|---|---:|
| Attempts with cache reads | 526 / 749 |
| Explicit one-hour cache misses/writes | 223 / 749 |
| Median cached input | ~25.3K tokens |
| Median uncached input | ~1.2K tokens |
| p75 uncached input | ~2.7K tokens |
| p90 uncached input | ~27.5K tokens |
| p95 uncached input | ~331K tokens |
| Attempts above 300K uncached tokens | 62 |
| Top 10% share of uncached input | ~94% |
| Top 10% share of estimated provider cost | ~61% |

The bottom 90% averaged only ~2.1K uncached tokens, so ordinary chats are not broadly broken. The problem is highly concentrated in long/agentic requests. The sample contained one first attempt per distinct invocation and no retry ordinals above one, ruling out retry amplification for this population.

No prompts, responses, raw logs, credentials, or user identifiers were used in these aggregates.

## Code-grounded mechanism

- `backend/routers/chat.py` persists the new human message, fetches `chat_db.get_messages(..., limit=10, app_id=...)`, then reverses the newest-first result into chronological order before calling agentic chat.
- `backend/database/chat.py::get_messages` orders by `created_at DESC` and applies the fixed limit. There is no token-aware budget, stable summary checkpoint, append-only growth phase, or turn-pair normalization on this path.
- `backend/utils/retrieval/agentic.py` adds the one-hour cache breakpoint to the system block, then sends `messages` and `tool_schemas` through the native Anthropic tool loop.
- Each tool round appends assistant tool-use content and tool results to the message list before sending the full list again.
- `backend/utils/llm/gateway_anthropic.py` copies caller kwargs and changes only the model/header routing fields. The native gateway `/v1/messages` path forwards the Anthropic body without flattening or removing cache controls; existing passthrough tests cover this contract.

Important nuance: because Anthropic's hierarchy is `tools → system → messages`, the current system breakpoint cumulatively caches stable tools plus system content. A separate tools breakpoint is useful only to preserve the tool prefix when mutable system content changes; tools are not universally uncached today.

## Desired outcome

Keep recent-chat behavior correct while preventing stable context from being reprocessed on every turn and every tool round.

The solution should:

1. Preserve the latest user message and bounded recent conversational context in chronological, role-valid order.
2. Reuse stable tools/system/history prefixes without assuming user content can be shared across users.
3. Make repeated agentic tool rounds pay only for newly appended tool-use/results where possible.
4. Avoid a permanently sliding prefix that invalidates historical-message caching after the ten-message boundary.
5. Retain bounded context growth and explicit compaction/reset behavior.
6. Expose low-cardinality cache/token distributions so regressions are visible without reading user content.

## Proposed direction

Treat these as separate cache layers:

1. **Stable tools:** evaluate an explicit breakpoint at the end of tools so mutable per-user system context does not invalidate the reusable tool-schema entry.
2. **Stable system:** retain the existing one-hour system breakpoint, keeping volatile datetime/current-turn data outside it.
3. **Pre-loop history:** put a breakpoint at the end of the historical messages loaded before the agentic loop. Later tool rounds should then reuse that base history and process only appended tool-use/results.
4. **Inter-turn history:** replace the always-sliding fixed-ten prefix with either:
   - a bounded append-only window with deliberate periodic reset, revisiting #4692; or
   - a stable compacted/summary checkpoint plus a bounded recent suffix.

Do not implement only a trailing-message breakpoint while retaining the permanently sliding window: it would help within one tool loop and during short growing chats, but would lose inter-turn reuse once old messages start falling off the front.

### Context correctness

- [ ] Tests prove the newest user message is always included.
- [ ] Tests cover chats below, at, and above the current ten-message boundary.
- [ ] Historical messages remain chronological and Anthropic-role-valid; malformed/reported-message filtering cannot silently create an invalid sequence.
- [ ] Context remains bounded by an explicit message/token policy.
- [ ] App/main-chat and session scoping behavior is documented and covered.

### Cache behavior

- [ ] A deterministic stateful test sends two consecutive turns with a shared prefix and proves the second request reports cache-read tokens for the stable portion.
- [ ] A deterministic multi-round tool test proves the second provider call reuses the pre-loop history rather than reprocessing it as uncached input.
- [ ] Tests cross the history-window reset/compaction boundary and prove cache misses occur only at the deliberate reset, not on every mature-chat turn.
- [ ] Existing native `/v1/messages` gateway parity tests continue proving `cache_control` and TTL preservation.
- [ ] Breakpoint count and TTL ordering comply with Anthropic's four-breakpoint and mixed-TTL rules.

### Observability and validation

- [ ] Add privacy-safe, low-cardinality histograms/counters for cached input, uncached input, cache writes, and cost estimates by provider/model/lane/outcome—never by user or request ID.
- [ ] Compare p50/p75/p90/p95 uncached input and top-decile concentration before and after a development bake.
- [ ] Demonstrate a substantial reduction in the extreme uncached tail without increasing error rate, latency, or context-correctness failures.
- [ ] Confirm no user content, raw identifiers, or credential material enters metrics/log labels.

## Non-goals

- Changing gateway traffic, provider credentials, model selection, or fallback policy.
- Treating this as a gateway cache-control passthrough bug.
- Unboundedly increasing historical context.
- Logging prompts, responses, tool results, or user identifiers for diagnosis.

## References

- Prior closed investigation: #4692
- Anthropic prompt caching: https://platform.claude.com/docs/en/build-with-claude/prompt-caching

Git-on-my-level:
## Summary

- hydrate each live desktop Pi binding with the full canonical retained context once, then deliver only new or changed turns while continuing to refresh dynamic sources and context-plan metadata
- replace the backend's moving newest-10 window with a bounded 10–17 message append epoch and enable Anthropic's automatic 1-hour message breakpoint for cross-turn history and agentic tool loops
- teach gateway accounting to recognize top-level Anthropic cache controls and document the code audit, measured baseline, candidate decisions, implementation tickets, and savings estimate
- restore typed desktop chat by routing on only the current `# User Message`, keeping personal-activity prompts on private Omi retrieval, and using Anthropic's direct-compatible web-search tool version
- let non-production named QA bundles reuse a valid owner-checked seeded token at startup while preserving production/Beta's forced Firebase refresh behavior

## Typed chat root cause and fix

Local Omi Beta logs showed typed turns failing quickly with Anthropic HTTP 400: the selected `web_search` tool was not callable through the configured direct path. PTT stayed healthy because it uses the separate Gemini Live `RealtimeHub` / `VoiceTurnCoordinator` path.

The failure had two coupled causes:

1. Pi's rendered prompt includes retained history plus a final `# User Message`. Retrieval policy classified the entire rendered prompt, so an older turn's web-required instruction could leak into later private or ordinary turns.
2. The gateway emitted `web_search_20260209` for direct named selection. That dynamic-filtering tool version permits the code-execution caller in the live Anthropic contract, while the current parser and routing path expect direct `server_tool_use` / `web_search_tool_result` blocks.

The fix isolates classification to the current user section, recognizes “what did I do” as private-context intent in both Rust and Pi, and pins the direct-compatible `web_search_20250305` request shape. Regression coverage exercises stale inherited web policy, personal activity, and the provider request shape.

The QA run also exposed a named-bundle-only startup issue: the bundle had a valid seeded Keychain token but attempted a source-app Firebase refresh that cannot exist under a distinct bundle ID. The new credential policy reuses the valid owner-checked token only for non-production named bundles; normal Omi and Omi Beta behavior is unchanged.

## Cache audit and expected impact

- Desktop already used explicit 1-hour breakpoints after the stable system policy and on the latest user/tool-result message. A literal midpoint marker inside the repeated snapshot would not be independently reusable because Anthropic hashes cumulative `tools → system → messages` prefixes.
- A measured mature 64-turn desktop request used 20,403 uncached plus 29,983 cache-read tokens (59.5% cache-read share). The repeated old-history slice is approximately 13.4K tokens per turn.
- Steady-state desktop uncached input is therefore estimated to fall from about 20.4K to 7.0K tokens (66%), with cache-read share rising to roughly 81% while the same live binding remains healthy.
- The backend epoch preserves at least the previous 10 visible messages, remains bounded at 17, resets deterministically every eight messages, and is scoped to the active session or app.
- No stable-prefix staleness defect was found: the desktop stable identity is constant for its intentional semantic-guidance/capability scope, while profile, memory, tasks, goals, and other live sources are dynamic.

Full audit and ticket pack: `docs/superpowers/specs/2026-07-23-context-cache-efficiency-design.md`

## Verification

- backend focused cache/gateway tests — 71 passed
- desktop Rust suite — 398 passed
- Pi adapter policy tests — 36 passed; full agent suite passed
- `xcrun swift test --package-path Desktop --filter AgentRuntimeProcessTests` — passed
- `./scripts/agent-logic-harness.sh --cross-surface-smoke` — all four stages passed
- named bundle `/Applications/omi-context-cache.app` — Swift/Rust/agent build, ad-hoc signing, install, launch, exact bundle identity, automation health, authenticated protocol-v2 runtime, and dev-backend configuration passed
- exact named-bundle repros: `hello?` completed with no error/stuck sending state; `What did I do today?` stayed on private Omi context and completed with no error/stuck sending state
- `scripts/pr-preflight --pr-body-file /tmp/pr-10422-preflight.md` — passed
- bounded pre-push acceptance gate — passed twice, including Swift compile, desktop tool surfaces (53 files / 597 tests), focused backend tests, product invariant checks, formatting, lint, and Rust compile

- INV-AGENT-*
- INV-CHAT-1

The change preserves canonical kernel ownership, binding replacement/reset boundaries, recent-turn continuity, and additive context delivery. It does not summarize, mutate, or remove canonical turn content. Production Omi/Omi Beta startup auth behavior remains fail-closed and unchanged.

Closes #10418

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-chat.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-chat.json
@@ -1,7 +1,7 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": 2216,
-    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": 4330
+    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": 4304
   },
   "raise_justifications": {
     "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": "deleteBackend flag threaded through both clearJournalTurns variants so a reset is local-only",
--- a/backend/database/chat.py
+++ b/backend/database/chat.py
@@ -23,6 +23,11 @@ DELETE_MESSAGES_BATCH_LIMIT = 200  # Leaves room for one session-counter write p
 DELETE_MESSAGES_CONFLICT_RETRIES = 3
 CHAT_HISTORY_BASE_VISIBLE_MESSAGES = 10
 CHAT_HISTORY_APPEND_EPOCH_MESSAGES = 8
+# Maximum number of reported (hidden) rows to over-fetch per raw Firestore
+# query when reading cache-aligned history. Keeps the raw read bounded even
+# when a user has thousands of lifetime reported messages; the newest page
+# rarely contains more reported rows than this cap.
+CHAT_HISTORY_REPORTED_RAW_SCAN_CAP = 50
 
 
 class ClientMessageIdPayloadConflict(ValueError):
@@ -323,7 +328,12 @@ def get_cache_aligned_messages(
     if visible_limit == 0:
         return []
 
-    raw_limit = min(total, visible_limit + reported)
+    # Cap the raw Firestore read so a large lifetime reported count cannot
+    # cause unbounded document reads on every chat send. The over-fetch only
+    # needs to cover reported rows that fall inside the newest raw page, not
+    # the lifetime total.
+    reported_overfetch = min(reported, CHAT_HISTORY_REPORTED_RAW_SCAN_CAP)
+    raw_limit = min(total, visible_limit + reported_overfetch)
     return get_messages(
         uid,
         limit=raw_limit,
--- a/backend/database/chat.py
+++ b/backend/database/chat.py
@@ -21,6 +21,8 @@ logger = logging.getLogger(__name__)
 BATCH_LIMIT = 500  # Firestore hard limit
 DELETE_MESSAGES_BATCH_LIMIT = 200  # Leaves room for one session-counter write per deleted message.
 DELETE_MESSAGES_CONFLICT_RETRIES = 3
+CHAT_HISTORY_BASE_VISIBLE_MESSAGES = 10
+CHAT_HISTORY_APPEND_EPOCH_MESSAGES = 8
 
 
 class ClientMessageIdPayloadConflict(ValueError):
@@ -273,6 +275,61 @@ def get_messages(
         message['files'] = [files[file_id] for file_id in message.get('files_id', []) if file_id in files]
 
     return messages
+
+
+def cache_aligned_history_limit(total_visible_messages: int) -> int:
+    """Return a bounded history size whose start moves only at epoch boundaries.
+
+    A fixed newest-N window changes at the front on every chat turn, invalidating
+    Anthropic's cumulative message-prefix cache. This policy keeps at least the
+    existing ten-message continuity window and lets it grow append-only for eight
+    messages before resetting to ten. The request therefore carries 10..17
+    messages, never less history than before and never an unbounded transcript.
+    """
+    if total_visible_messages < 0:
+        raise ValueError('total_visible_messages must be non-negative')
+    if total_visible_messages <= CHAT_HISTORY_BASE_VISIBLE_MESSAGES:
+        return total_visible_messages
+    return CHAT_HISTORY_BASE_VISIBLE_MESSAGES + (
+        (total_visible_messages - CHAT_HISTORY_BASE_VISIBLE_MESSAGES) % CHAT_HISTORY_APPEND_EPOCH_MESSAGES
+    )
+
+
+def get_cache_aligned_messages(
+    uid: str,
+    *,
+    app_id: Optional[str] = None,
+    chat_session_id: Optional[str] = None,
+) -> List[Dict[str, Any]]:
+    """Read a cache-aligned, scope-safe chat history in newest-first order.
+
+    Reported messages are excluded from the visible count and read. Over-fetching
+    by the scoped reported count guarantees the target number of visible messages
+    even when hidden records fall inside the selected raw Firestore page.
+    """
+    user_ref = db.collection('users').document(uid)
+    scoped_ref = user_ref.collection('messages')
+    if chat_session_id:
+        scoped_ref = scoped_ref.where(filter=FieldFilter('chat_session_id', '==', chat_session_id))
+    else:
+        scoped_ref = scoped_ref.where(filter=FieldFilter('plugin_id', '==', app_id))
+
+    total_result = scoped_ref.count().get()
+    total = int(total_result[0][0].value) if total_result and total_result[0] else 0
+    reported_result = scoped_ref.where(filter=FieldFilter('reported', '==', True)).count().get()
+    reported = int(reported_result[0][0].value) if reported_result and reported_result[0] else 0
+    visible_total = max(0, total - reported)
+    visible_limit = cache_aligned_history_limit(visible_total)
+    if visible_limit == 0:
+        return []
+
+    raw_limit = min(total, visible_limit + reported)
+    return get_messages(
+        uid,
+        limit=raw_limit,
+        app_id=app_id,
+        chat_session_id=chat_session_id,
+    )[:visible_limit]
 
 
 @prepare_for_read(decrypt_func=_prepare_message_for_read)
--- a/backend/llm_gateway/gateway/accounting.py
+++ b/backend/llm_gateway/gateway/accounting.py
@@ -435,11 +435,11 @@ def cache_requested_for_openai_request(request: Mapping[str, Any]) -> bool:
 
 
 def cache_requested_for_anthropic_request(request: Mapping[str, Any]) -> bool:
-    return any(_contains_cache_control(request.get(field)) for field in ('system', 'messages', 'tools'))
+    return _contains_cache_control(request)
 
 
 def cache_write_ttl_for_anthropic_request(request: Mapping[str, Any]) -> str | None:
-    ttls = _cache_control_ttls([request.get(field) for field in ('system', 'messages', 'tools')])
+    ttls = _cache_control_ttls(request)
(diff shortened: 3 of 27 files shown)


Git-on-my-level:
CI triage update: the cache implementation checks pass, and the two red merge-gate suites are blocked by reproducible `main` defects outside this PR's diff:

- Hermetic Backend E2E: fixed June 24 clock data has expired against `datetime.now()` and the isolated case also leaks a real Firestore client — tracked in #10423.
- Desktop Swift Test Suites: #10417 added a fifth `ChatQueryTelemetryEvent.failed` value without updating a four-value optional pattern in its test — tracked in #10424.

I retried the backend failure once and it reproduced identically. I have not mixed either unrelated repair into this performance PR.

Git-on-my-level:
Addressed both review threads from the deterministic monitor pass.

**P1 — desktop context-snapshot (tombstone dropped turns):**
`renderContextSnapshotForBinding` now falls back to a full re-render when the conversation's `totalTurnCount` decreases since the previous delivery. Hard-deleted turns (`journal_clear_turns` / `clearJournalConversation`) leave no tombstone in a delta, so the model would believe the cleared content still exists. Normal retention-window aging (turns dropping off as new ones arrive) does not trigger this because `totalTurnCount` only increases in that case. Regression test covers: delete turns → new turn → verify full delivery (not delta).

**P2 — backend chat.py (cap reported over-fetch):**
The raw Firestore over-fetch for reported (hidden) rows is now capped at `CHAT_HISTORY_REPORTED_RAW_SCAN_CAP` (50) instead of adding the lifetime reported count. A user with thousands of old reported messages would otherwise stream all of them on every chat send. Regression test covers: 300 reported → verify raw limit is 66 (16 visible + 50 cap), not 316.

**Verification:**
- `backend/.venv/bin/python -m pytest tests/unit/test_chat_message_count.py` — 7/7 passed
- `npx vitest run tests/context-snapshot.test.ts tests/context-cache-delivery.test.ts` — 24/24 passed
- `npx tsc --noEmit` — clean
- `black --check --line-length 120` — clean
- `scripts/failure-class validate` — passed (`Failure-Class: none`)

The Hermetic Backend E2E failure (`test_canonical_memory_pipeline.py::test_surface_excludes_archive_and_respects_grants`) is a pre-existing main defect in a file this PR does not touch.

Git-on-my-level:
Rebased onto latest `main` and force-pushed the rebased commit (f5dcfb1a19).

This resolved all three previously failing checks:

- **Hermetic Backend E2E**: The pre-rebase HEAD called `get_memories_text(uid=uid, limit=50)` without `now=`, so the agent_tools surface used wall time instead of the test fixture clock, causing short-term memories to be filtered out. Main commit `df1dd647ed` (make canonical read time injectable) threads `now=` through the agent-tools memory list path; the rebase picks it up. Now passing.
- **Backend Hermetic Merge Gate**: Cascaded from the E2E failure above. Now passing.
- **Desktop Swift Build & Tests**: The previous failure was a CI timeout (Desktop Swift Test Suites cancelled at 1h0m19s). The current run is in progress; the PR's TypeScript agent changes pass locally (24/24 tests) and Desktop Swift Static Contracts pass in CI.

Both trusted Codex review threads (P1: tombstone dropped turns, P2: cap reported over-fetch) were already addressed in the prior fix commit and remain resolved.

Issue #10429, opened 2026-07-23
Desktop chat: web search not triggered for natural "look X up online" requests (retrieval-policy keyword gap)


Archit-lal:
## Summary
Desktop chat only injects Anthropic's server-side `web_search` tool when `retrieval_policy()` (`desktop/macos/Backend-Rust/src/routes/retrieval_policy.rs`) flags the latest user turn as requiring the public web. That classifier is substring matching against ~16 fixed phrases (`EXPLICIT_WEB_PHRASES`) plus a few freshness/anaphoric patterns. Natural web-research requests that contain no literal trigger phrase fall through to `auto`, so the model gets **no web tool** and tells the user it can't search the web — even though the capability exists.

## Repro
Real user turn that fails:
> "find out who David Zhang is ... every information available on him online ... give me an HTML file ..."

Running the real `retrieval_policy()` against this message: `requires(PublicWeb)=false`, `reason=auto` -> no `web_search` tool injected -> the assistant replies it can't browse the web. ("...on him online" matches none of `search the web` / `web search` / `find it online` / `google it` / etc.)

## Why it matters
The keyword list catches explicit phrasings ("search the web", "google it") but misses the far more common implicit research asks ("find out who X is", "everything about X online", "look X up", "research X"). Users experience this as "the assistant can't do web search."

## Options (for discussion)
1. Broaden `EXPLICIT_WEB_PHRASES` / add a guarded research-intent heuristic (research verb + web/online locus, under the length cap, private-context precedence preserved). Classify as a non-explicit *guess* so a web-unavailable route degrades to model knowledge instead of failing.
2. Expose `web_search` as an *optional* tool on agentic turns and let the model decide (closer to the Python backend), accepting the cost tradeoff.

This is a cost/product tuning call ($10/1k searches; a forced search also bans private context for that turn), so it's intentionally split from the crash fix.

## Related
The separate defect that made web search **502 even when it was triggered** (server-side web_search running inside code execution, whose result blocks the desktop backend couldn't deserialize) is fixed in a companion PR. This issue is only about the trigger gap.

Git-on-my-level:
Verified fixed at HEAD (1ef0976ef9).

[...]

PR #10430, opened 2026-07-23
fix(desktop): force direct caller for chat web_search tool (fixes web search 502/timeout)

Archit-lal:
## Problem

Desktop chat's server-side web search **hard-fails every time it actually runs**. When a turn triggers web search, the assistant either errors out or (as users see it) "times out" — the turn ends with `status: "failed"`.

Real user session showing the symptom (the assistant self-reporting that web search runs *inside the code-execution sandbox* and the turn "just hung and died"):

> "The web search tool … is an async function injected into the `code_execution` sandbox … The prior attempt shows `status: "failed"` … The turn just… hung and died."

## Root cause

`web_search_tool_def()` requests `web_search_20260209` **without** an `allowed_callers` field. As of that tool version, `allowed_callers` **defaults to `["code_execution_20260120"]`** ([Anthropic docs](https://platform.claude.com/docs/en/docs/agents-and-tools/tool-use/web-search-tool)). On the programmatic-tool-calling models desktop chat routes web search to (`claude-sonnet-4-6` / `claude-opus-4-6`), Anthropic then runs the search **inside code execution** (dynamic filtering) and returns `code_execution_tool_result` / `bash_code_execution_tool_result` content blocks.

Our `AnthropicContentBlock` is a closed `#[serde(tag = "type")]` enum (`text`, `tool_use`, `server_tool_use`, `web_search_tool_result`, `thinking`, `redacted_thinking`) with **no variant and no `#[serde(other)]`** for code-execution blocks. So `receive_anthropic_response` fails to deserialize the whole response and returns `502 BAD_GATEWAY`. Both the non-streaming and server-tool streaming paths funnel through this strict aggregation (`complete_anthropic_server_tool_turn` forces `stream=false`), so **every** forced-web-search turn fails.

## Fix

Pin `"allowed_callers": ["direct"]` on the web_search tool so the search runs as a plain server tool and returns the `server_tool_use` + `web_search_tool_result` blocks the parser already handles (and whose response shape existing tests already cover). One-line behavioral change; no parser rewrite.

Tradeoff: direct mode forgoes dynamic filtering (search results load fully into context → marginally higher token cost on search-heavy turns). Acceptable — the alternative is the current 502. Regaining dynamic filtering by adding a `code_execution_tool_result` variant + `#[serde(other)]` is a larger, separate change.

## Verification

- **New regression test** `test_injected_web_search_tool_forces_direct_caller` runs real `translate_request_inner(...)` and asserts the injected tool carries `allowed_callers=["direct"]`. **Proven**: reverting the fix makes it panic on the assertion; with the fix it passes.
- `cargo test --bin omi-desktop-backend routes::chat::tests` → **91 passed, 0 failed**. `rustfmt --check` clean, `cargo build` clean.
- **Independent agent code review**: APPROVE — confirmed `["direct"]` is documented/valid, the parser fully covers the direct response shape (citations are nested inside text blocks; unknown fields are tolerated since no `deny_unknown_fields`), and no other block type a direct turn emits is left unhandled.
- **Convergent field evidence**: the production session above shows web search running via code execution and failing — exactly this defect.
- Not yet run: a live search through a locally-built backend (needs a server Anthropic key). The response-parse side is covered by existing hermetic tests, and the request-side change is covered by the new one.

## Scope / follow-ups

- **Not in this PR (tracked separately):** the retrieval-policy keyword classifier (`retrieval_policy.rs`) only injects web search when the message matches ~16 fixed phrases, so natural requests like "find out who X is and everything about them online" get no web tool at all. Issue: https://github.com/BasedHardware/omi/issues/10429
- **Parity follow-up:** `backend/utils/retrieval/agentic.py` `WEB_SEARCH_TOOL` has the same missing `allowed_callers`. The Python path tolerates unknown blocks (SDK `if/elif` filtering, no 502) but still silently runs in code-execution mode — worth confirming it surfaces citations.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10430?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Backend-Rust/src/routes/chat/request_translation.rs
+++ b/desktop/macos/Backend-Rust/src/routes/chat/request_translation.rs
@@ -64,6 +64,16 @@ fn web_search_tool_def() -> AnthropicToolDef {
         "type": WEB_SEARCH_TOOL_TYPE,
         "name": "web_search",
         "max_uses": WEB_SEARCH_MAX_USES,
+        // Force direct execution. As of web_search_20260209, `allowed_callers`
+        // defaults to ["code_execution_20260120"]: on a programmatic-tool-calling
+        // model (sonnet-4-6 / opus-4-6, the only models we route web search to)
+        // Anthropic runs the search *inside* code execution and returns
+        // `code_execution_tool_result` blocks. Our AnthropicContentBlock enum has
+        // no variant (and no serde(other)) for those, so `receive_anthropic_response`
+        // fails to deserialize and 502s the whole turn. "direct" keeps the search a
+        // plain server tool that returns the `server_tool_use` + `web_search_tool_result`
+        // blocks the parser already handles.
+        "allowed_callers": ["direct"],
     }))
 }
 
--- a/desktop/macos/Backend-Rust/src/routes/chat/tests/all.rs
+++ b/desktop/macos/Backend-Rust/src/routes/chat/tests/all.rs
@@ -984,6 +984,37 @@ fn test_translate_request_forces_required_web_search_without_client_tools() {
 }
 
 #[test]
+fn test_injected_web_search_tool_forces_direct_caller() {
+    // Regression: web_search_20260209 defaults allowed_callers to
+    // ["code_execution_20260120"] on programmatic-tool-calling models
+    // (sonnet-4-6 / opus-4-6). Under that default Anthropic returns
+    // code_execution_tool_result blocks our AnthropicContentBlock enum can't
+    // deserialize, so every forced web-search turn 502s. The injected tool must
+    // pin the direct caller so the search returns server_tool_use +
+    // web_search_tool_result blocks the parser already handles.
+    let req = test_request(vec![user_message("search the web for HumanPost")]);
+
+    let result = translate_request_inner(
+        &req,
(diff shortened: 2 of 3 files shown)


Archit-lal:
**Live production confirmation** — from a running `com.omi.computer-macos` session log today, Anthropic returned:

```
400 invalid_request_error: tool_choice.name 'web_search' cannot be used because this tool only allows calls from ['code_execution_20260120']. Tools specified in tool_choice must allow 'direct' calls from the model.
```

Anthropic's own error names the exact fix in this PR: the tool must allow **direct** calls. `allowed_callers: ["direct"]` does precisely that. This is the same root cause as the 502 on current `main` (default code-execution caller → `code_execution_tool_result` blocks the desktop backend can't deserialize); the fix resolves both the request-time 400 and the response-parse 502.

PR #10435, opened 2026-07-23
fix(backend): return a clear reply for oversized chat input

aaravgarg:
## What changed and why

On Omi Mobile Chat, an extremely long message (or a long conversation history) produced **no response** or a generic error. Root cause: the agentic chat path (`execute_agentic_chat_stream`) sends the full conversation to `claude-sonnet-4-6` (200k-token window). Oversized input makes the Anthropic call raise input-too-long; the agent loop swallows it into a short error text chunk but never populates `callback_data['answer']`, and `generate_stream` in `routers/chat.py` only emits the terminal `done:` frame when the answer is truthy. With an empty answer no `done:` frame is sent, so the mobile client never finalizes a reply and the user sees "no response."

This adds a small, import-light pre-flight guard to `backend/utils/retrieval/safety.py` (alongside `AgentSafetyGuard`, which already owns loop/context-overflow protection) wired into `execute_agentic_chat_stream` **before** any setup or the Anthropic producer task is created:

- `fit_within_budget(...)` trims the oldest turns so the cumulative token estimate fits a budget, always preserving the newest turn.
- Budget = `MAX_CHAT_INPUT_TOKENS` (env-overridable, default 120k, well below the 200k window, leaving headroom for the system prompt, tool schemas, tool results and the reply). Normal traffic is never trimmed.
- When the newest message alone is over budget, it returns a clear, persisted reply through the normal streaming/`done:` contract ("That message is too long ... please shorten it or split it into a few smaller messages") instead of calling the model with input that cannot fit.

Token counting reuses the existing tiktoken `num_tokens_from_string`. The guard is a no-op for all normal traffic and fixes both reported symptoms (no-response and generic error) for the long-message case.

none

## How it was verified

- `python3 -m pytest backend/tests/unit/test_chat_input_guard.py` -> 19 passed. Covers the pure decision logic (`message_text`, `fit_within_budget`: empty, under/over budget, history trimming, newest-at-limit boundary, message-like objects, null text, default budget) with an injected token counter, plus AST wiring tests asserting the guard runs before the producer task is created.
- `black --line-length 120 --check` on the three files -> clean.
- `python3 -m py_compile` on the two source modules -> ok.
- Root cause traced through `routers/chat.py` `generate_stream` (emits `done:` only when the answer is truthy) and the `utils/retrieval/agentic.py` error path.

I could not exercise the live mobile chat end-to-end in this environment (no running backend plus mobile client). Verification is via the unit and AST wiring tests that reproduce the exact decision logic and call ordering, plus the code-path trace above.

## Tests

`backend/tests/unit/test_chat_input_guard.py` (new) is the regression test that would have caught this: the AST wiring tests fail if `execute_agentic_chat_stream` stops calling `fit_within_budget` before `create_task`, and the pure-logic tests pin the trim/reject decision. Import-light (no heavy chat/LLM deps) so it satisfies the backend module-isolation rules.

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

Intentionally out of scope, noted as follow-ups:
- The shared producer error path could also emit a terminal `done:` frame on any unexpected failure (defense in depth beyond the long-input case).
- `AgentSafetyGuard(max_context_tokens=500000)` sits above the model's 200k window, a latent mismatch worth reconciling separately.


Git-on-my-level:
Nice fix for a real user-facing bug — the "no response" on oversized chat input was caused by the Anthropic input-too-long error being swallowed without populating `callback_data['answer']`, so no terminal `done:` frame reached the client.

The approach is clean and well-targeted:
- Import-light pure decision logic (`fit_within_budget` + `message_text`) with an injected token counter keeps the guard testable without the heavy chat/LLM stack.
- Guarding before the producer task is created (verified by the AST wiring test) ensures oversized input never reaches the model.
- Returning through the normal `done:` contract via `callback_data['answer']` is the right fix — it reuses the existing streaming/persistence path rather than adding a special case.
- 120k default budget with env override leaves comfortable headroom below the 200k Claude window for system prompt, tool schemas, and reply tokens.
- Test coverage is thorough: boundary conditions, history trimming, null text, env override, and the wiring guard.

One observation worth a human eye: `num_tokens_from_string` uses tiktoken's gpt-4 (`cl100k_base`) encoding to estimate Claude tokens. The tokenizers differ, but the 80k-token headroom (120k budget vs 200k window) comfortably absorbs the approximation error in practice — just flagging it so a maintainer can confirm the tradeoff is intentional.

The two failing CI checks (Backend Hermetic Merge Gate / Hermetic Backend E2E) appear unrelated — the failure is in `test_canonical_memory_pipeline` (memory surface access matrix), which doesn't touch chat input paths, and its fix (`df1dd647e`) postdates this branch's merge base. Worth a rebase to pick that up.

---

kodjima33 (approved):
fix(backend) clear reply for oversized chat input — approve only (UNKNOWN merge state)

PR #10440, opened 2026-07-23
fix: reduce backend 504 read amplification

Git-on-my-level:
## What changed and why

Production was still returning 30-second 504s on the expensive Action Items and Knowledge Graph GETs. This PR removes the second Action Items database invocation used only to compute `has_more`, and disables all automatic retries for those two mobile GET call sites so one failed read does not immediately repeat the same work.

This is a scoped read/retry-amplification mitigation. The remaining Action Items database invocation still streams and sorts the full matching collection, Knowledge Graph still returns the full graph, and the Memories route is unchanged. It does not claim a proven root cause or bounded Firestore reads.

The PR also repairs a date-dependent hermetic memory test exposed by CI. Canonical list reads now accept one optional explicit read instant, threaded through the real agent-tools call path; production callers omit it and retain UTC wall-clock behavior.

INV-MEM-1 — the three-tier/default-access behavior is unchanged; the hermetic default-access matrix now evaluates its fixed fixtures against the same fixed instant.

## How it was verified

- Refreshed read-only production evidence at 2026-07-23 22:04:48 UTC: the active backend revision was still returning only 30-35 second 504s in the sampled windows (77 in 15 minutes: 61 Action Items, 11 Knowledge Graph, 4 Memories, 1 development Action Items; 29 in 5 minutes: 23 Action Items, 5 Knowledge Graph, 1 development Action Items).
- `backend/.venv/bin/python -m pytest -q tests/unit/test_action_items_router_pagination.py tests/unit/test_action_items_pagination_order.py tests/unit/test_action_item_canonical_contract.py tests/unit/test_timeout_middleware.py` — 47 passed.
- `backend/.venv/bin/python scripts/scan_async_blockers.py --dirs routers utils` — 0 selected blocking findings.
- `backend/test-preflight.sh` — 17 checks passed, 0 failed, 9 optional environment warnings.
- `/Users/dazheng/fvm/versions/3.44.6/bin/flutter test test/unit/http_pool_manager_retry_test.dart` — 2 passed against a loopback 504 server through the real API call sites.
- Focused Flutter analysis found no new diagnostics. It reports one pre-existing `avoid_print` info on an untouched line in `knowledge_graph_api.dart`.
- `backend/.venv/bin/python -m pytest -q tests/unit/test_memory_service_parity.py tests/unit/test_ws_b_short_term_lifecycle.py` — 56 passed, including exact-expiry and canonical/legacy clock propagation.
- `pyright` on the three changed production memory modules — 0 errors, 0 warnings.
- The official macOS hermetic E2E runner hit its existing 120-second local startup timeout before producing test output. The Linux GitHub Actions hermetic job remains the full-harness authority.
- The full `backend/test.sh` runner was attempted but did not complete cleanly: three untouched tests exceeded the 0.12-second CPU guard under the broad parallel run, and one untouched memory-runtime adapter test remained stuck for about 11 minutes before the runner was interrupted. The affected-path suites above are green.

## Tests

- `backend/tests/unit/test_action_items_router_pagination.py` fails on the old two-call pagination implementation and covers full, exact, and underfilled pages plus trimming before locked-item redaction.
- `app/test/unit/http_pool_manager_retry_test.dart` exercises `tryGetActionItems` and `KnowledgeGraphApi.getKnowledgeGraph` through a loopback 504 server. It observes one request now; the old call sites would make two and three requests respectively.
- `backend/tests/unit/test_ws_b_short_term_lifecycle.py` proves a short-term item is visible before its supplied expiry and hidden exactly at expiry.
- `backend/tests/unit/test_memory_service_parity.py` proves the selected canonical or legacy backend receives the same explicit read instant.
- `backend/testing/e2e/test_canonical_memory_pipeline.py` passes its fixture clock through the real agent-tools list surface instead of mixing it with wall time.

## Failure class (fixes)

Failure-Class: none

No existing registered class describes duplicated expensive reads after an application timeout; failure-class lifecycle work belongs in a separate PR.

### DIFF
--- a/app/lib/backend/http/api/action_items.dart
+++ b/app/lib/backend/http/api/action_items.dart
@@ -49,7 +49,13 @@ Future<ActionItemsResponse?> tryGetActionItems({
     url += '&end_date=${endDate.toUtc().toIso8601String()}';
   }
 
-  var response = await makeApiCall(url: url, headers: {}, method: 'GET', body: '');
+  var response = await makeApiCall(
+    url: url,
+    headers: {},
+    method: 'GET',
+    body: '',
+    retries: 0,
+  );
 
   if (response == null) return null;
 
--- a/app/lib/backend/http/api/knowledge_graph_api.dart
+++ b/app/lib/backend/http/api/knowledge_graph_api.dart
@@ -14,7 +14,7 @@ class KnowledgeGraphApi {
       body: '',
       method: 'GET',
       timeout: const Duration(seconds: 60),
-      retries: 2,
+      retries: 0,
     );
 
     if (response != null && response.statusCode == 200) {
--- a/app/test/unit/http_pool_manager_retry_test.dart
+++ b/app/test/unit/http_pool_manager_retry_test.dart
@@ -1,0 +1,125 @@
+import 'dart:io';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:package_info_plus/package_info_plus.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+import 'package:omi/backend/http/api/action_items.dart' as action_items_api;
+import 'package:omi/backend/http/api/knowledge_graph_api.dart';
+import 'package:omi/backend/preferences.dart';
+import 'package:omi/env/env.dart';
+import 'package:omi/utils/platform/platform_manager.dart';
+
+void main() {
+  final env = _TestEnvFields();
+
+  setUpAll(() async {
+    Env.init(env);
+    SharedPreferences.setMockInitialValues({});
+    await SharedPreferencesUtil.init();
+    PackageInfo.setMockInitialValues(
+      appName: 'Omi Test',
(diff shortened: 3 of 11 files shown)


Git-on-my-level:
CI triage note: `Hermetic Backend E2E` failed twice in the untouched `TestSurfaceDefaultAccessMatrix...[agent_tools]` case. The fixture hard-codes `NOW = 2026-06-24 12:00 UTC`; `memory_item()` captures the short-term row one day earlier and sets a 30-day expiry, so it expired on 2026-07-23 12:00 UTC. The production reader evaluates `datetime.now(timezone.utc)`, which explains why the assertion now sees only the long-term row. This PR does not touch the memory test or read path. Backend units, Dart Analyze & Tests, Generated Files, metadata, OpenAPI, source closure, and both stack gauntlets are green. This date-dependent hermetic test needs a separate clock/fixture repair rather than being folded into the 504 mitigation.

PR #10450, opened 2026-07-24
fix(test): bind the hermetic agent-tools memory case to its fake datastore

### LINKED ISSUE #10423: Hermetic canonical-memory agent-tools case expires its fixed “fresh” fixture
`backend/testing/e2e/test_canonical_memory_pipeline.py::TestSurfaceDefaultAccessMatrix::test_surface_excludes_archive_and_respects_grants[agent_tools]` now fails deterministically in the Backend Hermetic E2E workflow (including a failed-job retry on PR #10422): the test seeds `fresh-short` at the fixed `NOW = 2026-06-24`, while the production agent-tools path calls `read_canonical_memories`, which evaluates short-term expiry against the real wall clock; on 2026-07-23 the short-term fixture is filtered and only `coffee long term` remains. Running this case alone also exposes an existing order dependency: without an earlier `client` fixture, `tool_memories_service.firestore_db` is not rebound to `fake_firestore` and attempts a real Firestore read. Evidence: https://github.com/BasedHardware/omi/actions/runs/30036304928. The hermetic test should control the reader clock and bind its datastore explicitly so it is date- and order-independent.

aryanorastar:
## What

Fixes the order dependency in `TestSurfaceDefaultAccessMatrix::test_surface_excludes_archive_and_respects_grants[agent_tools]`, the case failing the **Backend Hermetic E2E** job (and therefore the **Backend Hermetic Merge Gate**) across several open PRs.

Fixes #10423.

## Root cause

The agent-tools case reads through `utils.retrieval.tool_services.memories.get_memories_text`, which resolves the Firestore client that module captured at import time (`from database._client import db as firestore_db`). The case never bound that name, so it only passed when an earlier test's fixture happened to rebind it.

Run on its own it reaches for **real Firestore**:

```
pytest 'testing/e2e/test_canonical_memory_pipeline.py::TestSurfaceDefaultAccessMatrix::test_surface_excludes_archive_and_respects_grants'
→ 1 failed, 5 passed
   PermissionDenied: Cloud Firestore API has not been used in project test-e2e-project
```

That's both an order dependency and a hermeticity hole — a "hermetic" case escaping to a real service.

## Fix

Bind the module's client to the case's `fake_firestore` explicitly. Six added lines, test-only.

## On the clock half of #10423

The issue also suspected the reader evaluated short-term expiry against the wall clock. **That part is already resolved on current `main`** — the clock is injected end to end:

`get_memories_text(now=)` → `MemoryService.read(now=)` → `read_canonical_memories(now=)` → `filter_canonical_default_visible_items(now=)`

I verified it rather than assuming: re-ran the matrix with the fixture clock moved **18 months** into the past (`NOW = 2025-01-15`) — **6 passed**. If the wall clock still leaked, the `fresh-short` fixture would have been filtered. So after this change the case is both **date- and order-independent**.

## Verification

- Case standalone (the failing repro): **6 passed** (was 1 failed, 5 passed)
- Whole file, CI-like ordering: **8 passed**
- Date-independence probe (`NOW` −18 months): **6 passed**

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/testing/e2e/test_canonical_memory_pipeline.py
+++ b/backend/testing/e2e/test_canonical_memory_pipeline.py
@@ -457,6 +457,12 @@ class TestSurfaceDefaultAccessMatrix:
 
             install_vector_search_fakes(monkeypatch, vector_db)
         db = fake_firestore
+        if surface_name == "agent_tools":
+            # get_memories_text reads through the client this module captured at import
+            # time, so bind it explicitly here. Without this the case only passes when an
+            # earlier test's fixture happened to rebind it, and standalone it reaches for
+            # real Firestore (#10423).
+            monkeypatch.setattr(tool_memories_service, "firestore_db", db)
         _seed_apply_control(db, uid)
         _seed_rollout_readiness(db, uid, grant_consumer=grant_consumer)
 


Git-on-my-level:
Nice find and clean fix — thanks @aryanorastar.

I verified the root cause statically: `utils/retrieval/tool_services/memories.py` captures the Firestore client at import time (`from database._client import db as firestore_db`, line 12) and `get_memories_text` reads through that module-level name rather than a passed-in client. The `agent_tools` branch in `surface_default_reader` calls `get_memories_text(uid, limit, now)` without forwarding the case's `db`, so the explicit `monkeypatch.setattr(tool_memories_service, "firestore_db", db)` is exactly the right binding to close the hermeticity hole — it matches how every other surface in this matrix receives its datastore. The 18-months-past clock re-run is a good way to confirm the date-independence half is already covered on `main`.

The Backend Hermetic Merge Gate and Hermetic Backend E2E are green on this head, so the order dependency is resolved. The remaining check entries appear as skipped/cancelled only because this is a test-only path with no source formatting/hygiene surface.

Leaving formal approval to a human maintainer; from a correctness standpoint this looks good.

---

kodjima33 (approved):
Test-only hermetic fix (fixes #10423): binds agent_tools case to fake_firestore, removes order dependency + real-Firestore escape. Backend suite green.

Issue #10394, opened 2026-07-23
Memory V3 prod rollout tracker (MEMORY_MODE still off)


Git-on-my-level:
## Status

Prod `MEMORY_MODE` is still **`"off"`**.

Verified in-repo (not a live Cloud Run probe):

- `backend/charts/backend-listen/prod_omi_backend_listen_values.yaml`: `MEMORY_MODE=off`, `MEMORY_ENABLED_USERS=""`, `MEMORY_V3_GET_ENABLED=false`, promotion cron/fast-track `false`
- `backend/deploy/runtime_env.yaml` prod surfaces: same `off` / empty cohort
- `backend/AGENTS.md` / `docs/runbooks/memory-v3-dev-cloud-proof.md`: production remains off until Gate 3

Confirm against live prod env before flipping any child to startable.

## Purpose

Single gate for **post–Memory V3 / canonical memory prod rollout** follow-ups so agents do **not** treat the children below as independently startable. Work that depends on Gate 3 (prod `MEMORY_MODE=read` for the intended cohort + canonical ST→LT lifecycle live via `memory-maintenance-job`) waits here.

## Do not start children until

All of the following are true (re-check before unblocking):

1. **Gate 3** production activation proof is GO per `docs/rollout/memory-v3-proof-order.md` / `docs/runbooks/memory-v3-production-activation.md`.
2. Prod request-path **`MEMORY_MODE=read`** for the intended whitelist cohort (not `off`/`shadow`/`write`-only).
3. **`memory-maintenance-job`** is enabled in lockstep (cron/fast-track as approved); `validate-backend-runtime-env.py` must pass — do not enable request-path read while the job stays off.
4. Canonical memory lifecycle (ST→LT maintenance / promotion) is live for that cohort.

Until then: comment on this issue, do not open implementation PRs for the children.

## Child issues (blocked)

- [ ] #9520 — ProactiveAssistants plural-authority audit (gated on Gate 3)
- [ ] #9359 — Tasks/Goals/Workstreams on mobile, web, non-macOS
- [ ] #9360 — Offline feedback-learning / per-user rules
- [ ] #8369 — MCP People/entity cleanup tools (already parked)

## Related in-flight drafts (intentional keep)

User is **keeping** these drafts; do not close/merge as part of parking children:

- PR #10171 — chat-first
- PR #9499 — memory atlas

[...]

PR #10451, opened 2026-07-24
fix(memories): mirror canonical deletes into the store the user still reads

aryanorastar:
## What

Deleted memories come back a few seconds later. Root-caused to a read/write decision asymmetry in the V3 memory rollout — not a client cache issue.

Refs #10446 (server-side half; see the note on symptom 1 below).

## Root cause

The delete and list routes are gated on **different rollout modes**:

| route | gate | canonical when |
|---|---|---|
| `DELETE /v3/memories/{id}`, `/v3/memories/batch`, `/v3/memories` (all) | `canonical_write_decision` | `mode ∈ {write, read}` |
| `GET /v3/memories` | `canonical_read_enabled` | **`mode == read` only** |

So during the dual-write stage (`MEMORY_MODE=write`) — which every rollout passes through before read cutover, and #10394 has this one mid-flight — a delete tombstones the **canonical** record while the user's list keeps reading **legacy**, where the row is untouched. The client drops it optimistically, the next refresh re-reads legacy, and the memory reappears. That matches the reported ~10–15s timing.

## Fix

Mirror the canonical delete into legacy until read cutover — deleting is symmetric with dual-write — across **all three** delete paths (single, batch, and delete-all). The mirror:

- **no-ops after cutover** (`canonical_read_enabled` true ⇒ both sides canonical), so this is confined to the migration window;
- is **best-effort** — canonical is authoritative and its delete has already committed, so a legacy cleanup failure must not fail a request the user already saw succeed;
- is harmless for canonical-only rows: Firestore deletes of absent documents are no-ops.

Single/batch mirror the specific `memory_ids`; delete-all purges the legacy collection for the uid (`_mirror_delete_all_into_legacy` → `_purge_legacy_memories`), matching what the canonical delete-all already did to the canonical store.

## ⚠️ Please confirm the semantic

This changes deletion behavior during a live migration, so it deserves the rollout owner's eye rather than mine. I picked **symmetric dual-delete** (a delete applies to both stores while both are live). The alternative is routing deletes at the store the user reads. I'm happy to switch if you prefer the other semantic.

Note the delete-**all** mirror is a bulk wipe of the legacy collection for that uid — larger blast radius than single/batch, but it only mirrors what the canonical delete-all already committed, is confined to the write window, and no-ops after cutover. Called out so the rollout owner can weigh it explicitly.

Related: `_force_canonical` in the batch-delete tests previously left the read stage implicit, inheriting whatever the ambient rollout config produced. It now pins `MEMORY_MODE=read` explicitly, so those batch-vs-single authorization-parity cases keep asserting exactly what they always did — the contract is unchanged, only the stage is now stated.

## Tests

`backend/tests/unit/test_memories_batch_delete.py` (**23 passed**, Python 3.11 `.venv`):
- `test_delete_during_dual_write_also_clears_the_store_the_user_reads` — the single/batch regression. Fails on old code (`Expected 'mock' to have been called once. Called 0 times.`), passes with the fix.
- `test_delete_after_read_cutover_leaves_legacy_alone` — the mirror must not fire post-cutover.
- delete-all variants covering the same dual-write-clears / post-cutover-leaves-alone pair.

**Not exercised live:** I can't see prod's `MEMORY_MODE` or the reporter's account from here. If prod is at `write`, this is deterministic for every affected account rather than a flake — worth confirming when you triage.

## Changes since first review

Folded in the delete-all mirror (previously split into #10454, now closed) so this PR closes the whole asymmetry in one place, per the review — single, batch, and delete-all all mirror during the write window.

- INV-MEM-1 — no tier added, no new canonical collection, default access policy untouched; this only mirrors a delete into the pre-existing legacy collection during the migration window. Guard test green.

Failure-Class: none

No registered class covers "a mutation must land in the store the caller reads while a dual-write migration is in flight". Single instance, fixed at the route boundary that owns the decision.

### DIFF
--- a/backend/routers/memories.py
+++ b/backend/routers/memories.py
@@ -365,6 +365,45 @@ def _mirror_delete_into_legacy(uid: str, memory_ids: List[str], *, db_client: An
         logger.exception("legacy mirror vector delete failed uid=%s count=%d", sanitize_pii(uid), len(memory_ids))
 
 
+def _purge_legacy_memories(uid: str) -> None:
+    """Delete every legacy memory for a user, plus its vectors.
+
+    Collects ids before the Firestore delete so the Pinecone vectors can be purged too —
+    otherwise orphaned vectors become search noise nothing ever cleans up.
+    """
+    memory_ids: List[str] = []
+    offset = 0
+    batch_size = 1000
+    while True:
+        memories = memories_db.get_memories(uid, limit=batch_size, offset=offset, include_invalidated=True)
+        if not memories:
+            break
+        memory_ids.extend([memory_id for m in memories if isinstance((memory_id := m.get('id')), str) and memory_id])
+        offset += batch_size
+
+    memories_db.delete_all_memories(uid)
+
+    if memory_ids:
+        delete_memory_vectors_batch(uid, memory_ids)
+
+
+def _mirror_delete_all_into_legacy(uid: str, *, db_client: Any) -> None:
+    """Mirror a canonical delete-all into legacy while the user still reads legacy.
+
+    Same window and reasoning as _mirror_delete_into_legacy: at MEMORY_MODE=write the
+    canonical wipe is invisible because GET /v3/memories still reads legacy, so "delete
+    everything" leaves the user's list intact on the next refresh (#10446). No-ops after
+    read cutover, and best-effort because canonical is authoritative and already
+    committed.
+    """
+    if canonical_read_enabled(uid, db_client=db_client):
+        return
+    try:
+        _purge_legacy_memories(uid)
+    except Exception:
+        logger.exception("legacy mirror delete-all failed uid=%s", sanitize_pii(uid))
+
+
 def _validate_memory(uid: str, memory_id: str) -> MemoryPayload:
     return fetch_memory_dict(uid, memory_id, db_client=getattr(db_client_module, 'db', None))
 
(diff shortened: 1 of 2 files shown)


Git-on-my-level (dismissed):
Thanks for tracking this down — the symmetric dual-delete direction makes sense for the single and batch delete paths during the write-before-read rollout window.

I found one blocker before this can land: `DELETE /v3/memories` (delete-all) still only calls `MemoryService.delete_all(uid)` in the canonical-write branch and returns without clearing legacy. During `MEMORY_MODE=write`, `GET /v3/memories` still reads legacy, so a user who chooses “delete all memories” would see the same class of resurrection this PR fixes for single/batch deletes. Please mirror delete-all into the legacy store while `canonical_read_enabled(...)` is false as well, and add a regression test for that route.

Validation I ran locally with stripped env/secrets: `pytest -q backend/tests/unit/test_memories_batch_delete.py` using the existing backend venv from another checkout; the current touched test file passes (20 passed), but it does not cover the delete-all route.

---

aryanorastar:
Done — folded the delete-all mirror in (it was split into #10454, now closed), so this PR closes the whole asymmetry: single, batch, **and** delete-all all mirror into legacy during the write window via `_mirror_delete_all_into_legacy` → `_purge_legacy_memories`, no-op after read cutover. `test_memories_batch_delete.py` covers the delete-all dual-write-clears / post-cutover-leaves-alone pair — **23 passed**.

Flagged the bulk-wipe blast radius of the delete-all mirror in the body for your call, since it's larger than single/batch.

(Also fixed a stale static count in `test_memories_create.py` — 11→12 rate-limited endpoints — that's red on main independently; separate labeled commit.)

kodjima33 (approved):
fix(memories) mirror canonical deletes — approve only (security-review, CHANGES_REQUESTED)

Git-on-my-level:
Thanks for the update — the previous delete-all blocker looks resolved on the current head.

What I verified:
- Single and batch canonical deletes now mirror the specific ids into legacy only while canonical writes are on but canonical reads have not cut over.
- `DELETE /v3/memories` now mirrors delete-all through `_mirror_delete_all_into_legacy` / `_purge_legacy_memories` in the same pre-read-cutover window.
- Post-cutover cases leave legacy alone, and the legacy delete-all branch still uses the shared purge helper.
- Targeted validation passed locally with a fresh backend venv and stripped test env: `pytest -q tests/unit/test_memories_batch_delete.py` → 23 passed, 9 warnings.

I removed `needs-tests` because the regression coverage is now present and CI is green. I’m intentionally not approving because this is still a memory-deletion/data-lifecycle change during rollout and the PR carries the right `security-review` / `needs-maintainer-review` gates. David or another maintainer should explicitly sign off on the delete-all blast-radius semantics before merge.

---

kodjima33 (approved):
Real dual-write bug (#10446) — canonical delete invisible while reads still hit legacy. Scoped, best-effort mirror, tests cover both rollout stages.

aryanorastar:
Rebased onto current main. **Green (0 failing), MERGEABLE.** Single + batch + delete-all mirror are all folded in with tests. Ready for re-review.

Issue #10434, opened 2026-07-23
Bug: Android CV1 diarization recognizes only my voice and never learns anyone else


hugoaap-code:
## Summary
On the Omi CV1 Android app, diarization recognizes only my own voice.

It does not learn or identify any other speakers, no matter how many times I try to teach it.

## Problem
The app consistently recognizes only me.

Anyone else in the conversation is not properly identified as a learned speaker, even after repeated correction or training attempts.

## Expected behavior
After teaching the app about other people multiple times, diarization should learn and identify those speakers correctly.

## Actual behavior
Only my voice is recognized.

No other speakers are learned or identified properly, regardless of how many times I try to teach the app.

## Impact
This makes diarization unusable for multi-person conversations and reduces the reliability of transcripts, memories, and summaries.

## Reproduction
1. Use the Omi CV1 Android app in conversations with multiple people
2. Repeatedly try to teach or correct speaker identities for other participants
3. Check later diarization results
4. Observe that only my own voice is recognized and no one else is learned

## Suggested investigation
- Check whether speaker training data for non-primary voices is being saved correctly
- Check whether diarization updates from user corrections are applied to future conversations
- Check whether Android CV1 is failing to sync or use learned speaker profiles for anyone except the primary user
- Check whether the model or pipeline is collapsing all non-user speakers into an unlearned fallback state

## Environment
- Omi CV1
- Android app
- Production environment

aryanorastar:
Found a mechanism that produces exactly this — "my voice always, nobody else ever, no matter how often I teach them" — and opened #10453 for it.

`extract_speaker_samples` stores the speech sample, then extracts the speaker embedding inside a `try/except` that **only logs**. If that extraction fails, the person is left holding quality-verified samples with **no embedding**, and nothing ever recomputes it. The listen loader requires a stored embedding, so it skips that person on every later session — and `maybe_migrate_person_samples` returns early for anyone already at v3, so re-teaching doesn't rescue them either. Once a person lands in that state they are permanently invisible to matching.

What makes the symptom so lopsided is the asymmetry: your **own** profile already self-heals from exactly this state — if its stored embedding is missing, the loader re-extracts it from your profile audio and writes it back. People had no equivalent. Hence own voice recognised, everyone else never.

[...]

PR #10453, opened 2026-07-24
fix(listen): recover a taught person's lost speaker embedding

aryanorastar:
## What

A taught person can become permanently unmatchable, so the user sees only their own voice recognised — no matter how many times they teach anyone else.

Refs #10434.

## Root cause

`extract_speaker_samples` stores the sample, then extracts the speaker embedding inside a `try/except` that **only logs**:

```python
try:
    embedding = await run_blocking(sync_executor, extract_embedding_from_bytes, wav_bytes, "sample.wav")
    await run_blocking(db_executor, users_db.set_person_speaker_embedding, uid, person_id, embedding_list)
except Exception as emb_err:
    logger.error(f"Failed to extract/store speaker embedding: {emb_err} ...")
```

When that fails, the person keeps their (quality-verified) samples with **no embedding**, and nothing ever recomputes it. The listen loader requires a stored embedding, so it skips that person on every later session — and `maybe_migrate_person_samples` returns early for anyone already at v3, so re-teaching doesn't rescue them either.

**The asymmetry that makes the symptom so lopsided:** the user's *own* profile already self-heals from exactly this state — if its stored embedding is missing it re-extracts from the profile audio and writes it back (`speakers.py`, the `has_speech_profile` branch). People had no equivalent. Hence: own voice always recognised, everyone else never.

## Fix

Give people the same fallback. When a person has samples but no stored embedding, rebuild it from their first stored sample and persist it.

**Deliberately scoped so it does not reopen the quality gate.** Recovery only runs for `speech_samples_version >= 3`. Those samples already passed `verify_and_transcribe_sample` when stored (the transcript that gates v3 is produced by that verification), so this restores a *lost* embedding rather than resurrecting a *rejected* one. The migrations intentionally drop embeddings for samples that fail quality, and that behaviour is untouched — which matters, because the failure mode of getting this wrong is **mis**identification, worse than non-identification.

Failures stay contained: one unreadable sample returns `None` and the remaining speakers still load.

## Tests

`backend/tests/unit/test_speaker_person_embedding_recovery.py`:
- recovers a v3 person whose embedding was lost, **and persists it** so the next session matches from storage instead of re-extracting
- the recovered vector is shaped for `compare_embeddings` (matches under `SPEAKER_MATCH_THRESHOLD`), not merely non-`None`
- a person with no samples is not recovered and storage is never touched
- a storage/extraction failure is contained

Suites: `test_speaker_id_pipeline.py` + the new file → **112 passed**.

## Scope, honestly

This closes one deterministic path to the reported symptom — the one I could establish from the code. I can't see the reporter's account state, so I can't confirm it's the only contributor. A person can also stay unmatchable if migration fails repeatedly (`maybe_migrate_person_samples` returns the person unchanged on failure), which is a separate question about whether quality-rejected samples should ever be retried — a product call I've deliberately not made here.

Failure-Class: none

No registered class covers "a derived artifact is lost by a swallowed exception and never rebuilt"; the durable guard is the recovery path plus its tests.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10453?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/listen/speakers.py
+++ b/backend/routers/listen/speakers.py
@@ -13,6 +13,7 @@ import numpy as np
 from utils.audio import AudioRingBuffer
 from utils.executors import storage_executor, sync_executor, run_blocking
 from utils.other.storage import get_profile_audio_if_exists
+from utils.speaker_sample import download_sample_audio
 from utils.speaker_sample_migration import maybe_migrate_person_samples
 from utils.stt.speaker_embedding import SPEAKER_MATCH_THRESHOLD, compare_embeddings, extract_embedding_from_bytes
 from utils.transcribe_decisions import USER_SELF_PERSON_ID, should_spawn_speaker_match
@@ -69,12 +70,16 @@ class SpeakerMatcher:
             for person in people:
                 if person.get('speech_samples'):
                     person = await maybe_migrate_person_samples(self.host.request.uid, person)
-                embedding = person.get('speaker_embedding')
-                if embedding and person.get('speech_samples') and person.get('speech_samples_version', 1) >= 3:
-                    self.person_embeddings[person['id']] = {
-                        'embedding': np.array(embedding, dtype=np.float32).reshape(1, -1),
-                        'name': person['name'],
-                    }
+                stored = person.get('speaker_embedding')
+                verified_samples = bool(person.get('speech_samples')) and (person.get('speech_samples_version', 1) >= 3)
+                vector: Optional[Any] = None
+                if verified_samples:
+                    if stored:
+                        vector = np.array(stored, dtype=np.float32).reshape(1, -1)
+                    else:
+                        vector = await self._recover_person_embedding(person)
+                if vector is not None:
+                    self.person_embeddings[person['id']] = {'embedding': vector, 'name': person['name']}
         except Exception as error:
             logger.error('Speaker ID embeddings load failed type=%s', type(error).__name__)
             state.speaker_id_done.set()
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks for this — clean fix and a great root-cause writeup. The asymmetry between the user-profile self-heal path and the person path is exactly the kind of latent bug that's easy to miss, and your `_recover_person_embedding` mirrors the existing `has_speech_profile` fallback nicely.

A few observations from review:

- **Correctness:** Using `samples[0]` for recovery is reasonable — these are v3+ samples that already passed `verify_and_transcribe_sample`, so you're restoring a lost embedding without reopening the quality gate, as noted. The persistence write-back via `self.host.persistence.call(user_db.set_person_speaker_embedding, ...)` matches the teach path (`speaker_identification.py`), so subsequent sessions match from storage instead of re-extracting. Good.
- **Containment:** The `try/except` with typed logging and `return None` means a recovery failure won't break loading of other speakers — matches the surrounding `load_and_run` style.
- **Tests:** Good coverage of the success/no-samples/failure-contained paths and the vector-shape assertion via `compare_embeddings`. The `ENCRYPTION_SECRET`/`OPENAI_API_KEY` env defaults follow the existing test convention in the repo.
- **Checks:** The material backend gates (Backend unit suite, Listen Pusher Stack Gauntlet, Hermetic Backend E2E, Hermetic Merge Gate) are green. A few auxiliary checks (Hygiene, Formatting, Detect Changes) show `cancelled` on a re-run — these should clear on the next push or a check re-run.

This touches user-data persistence (speaker embeddings for taught people), so I'm flagging it for human maintainer sign-off before merge rather than approving directly. The implementation looks sound to me.

---

kodjima33 (approved):
Backend bug fix (refs #10434): recover a taught person's lost speaker embedding from their v3 samples, mirroring the own-profile self-heal. Scoped, dedicated test, all backend gates green.

PR #10454, opened 2026-07-24
fix(memories): mirror the canonical delete-all into the store the user reads

aryanorastar:
## What

Follow-up to #10451, which I deliberately scoped out of that PR. `DELETE /v3/memories` (delete-**all**) carries the same read/write asymmetry: at `MEMORY_MODE=write` the canonical wipe is invisible because `GET /v3/memories` still reads legacy, so **"delete everything" leaves the user's entire list intact** on the next refresh.

Stacked on #10451 (it reuses that PR's guard); review the last commit only. Refs #10446.

## ⚠️ This one needs your call before merge

I held this back from #10451 on purpose and I'd rather you decide than assume: mirroring delete-all means a **bulk wipe of the legacy collection**. That is a much larger blast radius than the single/batch mirror, on a user-data deletion path, during a live migration.

The argument for it: the user explicitly asked to delete everything, and today that request silently doesn't affect what they see. The argument against: if legacy is ever meant to be recoverable during the migration, this removes that option in one call.

If you'd rather leave delete-all asymmetric until read cutover, close this — #10451 stands on its own and fixes the reported symptom.

## Fix

- Mirror the wipe into legacy until read cutover, under the **same guard and best-effort contract** as the single/batch mirror: no-ops once `canonical_read_enabled` is true, and a legacy failure never fails a request whose canonical delete already committed.
- Factored the legacy branch's collect-then-purge into `_purge_legacy_memories`, so both paths share one definition instead of duplicating the pagination and vector cleanup.

## Tests

`backend/tests/unit/test_memories_batch_delete.py`:
- delete-all during dual-write reaches the store the user reads
- delete-all after read cutover leaves legacy alone
- the legacy cohort still purges after the factoring (guards the refactor)

Suites: `test_memories_batch_delete.py` → 23 passed; `test_memories_batch.py` + `test_memory_api_egress_contract.py` + `test_memories_pagination_clamp.py` + `test_inv_mem_1_guard.py` → 34 passed.

- INV-MEM-1 — no tier added, no new canonical collection, default access policy untouched; this only mirrors an existing delete into the pre-existing legacy collection during the migration window. Guard test green.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/memories.py
+++ b/backend/routers/memories.py
@@ -365,6 +365,45 @@ def _mirror_delete_into_legacy(uid: str, memory_ids: List[str], *, db_client: An
         logger.exception("legacy mirror vector delete failed uid=%s count=%d", sanitize_pii(uid), len(memory_ids))
 
 
+def _purge_legacy_memories(uid: str) -> None:
+    """Delete every legacy memory for a user, plus its vectors.
+
+    Collects ids before the Firestore delete so the Pinecone vectors can be purged too —
+    otherwise orphaned vectors become search noise nothing ever cleans up.
+    """
+    memory_ids: List[str] = []
+    offset = 0
+    batch_size = 1000
+    while True:
+        memories = memories_db.get_memories(uid, limit=batch_size, offset=offset, include_invalidated=True)
+        if not memories:
+            break
+        memory_ids.extend([memory_id for m in memories if isinstance((memory_id := m.get('id')), str) and memory_id])
+        offset += batch_size
+
+    memories_db.delete_all_memories(uid)
+
+    if memory_ids:
+        delete_memory_vectors_batch(uid, memory_ids)
+
+
+def _mirror_delete_all_into_legacy(uid: str, *, db_client: Any) -> None:
+    """Mirror a canonical delete-all into legacy while the user still reads legacy.
+
+    Same window and reasoning as _mirror_delete_into_legacy: at MEMORY_MODE=write the
+    canonical wipe is invisible because GET /v3/memories still reads legacy, so "delete
+    everything" leaves the user's list intact on the next refresh (#10446). No-ops after
+    read cutover, and best-effort because canonical is authoritative and already
+    committed.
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Thanks for the careful follow-up and for explicitly separating this from #10451.

I reviewed the diff around `DELETE /v3/memories` and the added regression coverage. The implementation looks technically coherent: canonical delete/delete-all commits first, legacy mirroring only runs before canonical read cutover, and the legacy purge still collects IDs before deleting so vectors are cleaned up. I also ran the touched unit file in a scrubbed local environment:

`pytest tests/unit/test_memories_batch_delete.py -q` → 23 passed.

I am not approving this automatically because this is intentionally a user-data deletion/migration blast-radius decision: mirroring delete-all means a canonical wipe also bulk-wipes the legacy collection while legacy is still the read surface. That matches the current user-visible semantics of "delete everything", but it removes legacy as a rollback/recovery source for that user's memories during the rollout. Human maintainer review should decide whether that is the desired migration contract before merge.

One merge-readiness note: GitHub still shows `Hygiene` cancelled and `Formatting` skipped on this head, so please make sure the required gates are clean before landing.

---

aryanorastar:
Folded into #10451 so the single, batch, and delete-all mirrors land as one complete fix for the dual-write delete asymmetry (per the review on #10451, which flagged delete-all as the remaining gap). This branch was stacked on #10451 and added only the delete-all commit, now part of #10451.

PR #8728, opened 2026-07-01
feat(frontend): implement baseline memories toggle and UI badges in mobile app

srujana-keth:
## Description                                                                                                                      
                                                                                                                                    
  Implements Persistent Baseline Memories (resolves #4631).                                                                           
                                                                                                                                      
  Baseline memories are critical user facts (name, preferences, core background context that are pinned so the AI always has them available — without the user needing to re-prompt every session. Users can flag any memory as "Baseline" directly from the mobile app, which shows the status in the list view and allows toggling from the edit sheet.                                      

  ---

  ## Key Changes

  ### Backend (`backend/`)

  **Schema** (`backend/models/memories.py`)
  - Added `is_baseline: bool = False` to `MemoryDB` — backward-compatible, existing memories default to `False`

  **Toggle Endpoint** (`backend/routers/memories.py`)
  - `PATCH /v3/memories/{memory_id}/baseline` — secured under `memories:modify` rate limit
  - Uses `_validate_mutable_memory` (not the simpler `_validate_memory`) so canonical-path users are validated against the canonical
  store, not the legacy Firestore path

  **LLM Context Injection** (`backend/utils/llms/memory.py`)
  - `get_prompt_data` now returns a 4-tuple `(user_name, baseline, user_made, generated)` and handles both `MemorySystem.CANONICAL`
  (via `MemoryService.read()`) and the legacy path
  - `get_prompt_memories` prepends baseline memories first with a distinct label (`"baseline facts … always in context"`) so they are
  never dropped regardless of total memory count

  **Unit Tests** (`backend/tests/unit/test_baseline_memories.py`)
  - 12 hermetic tests — no Firebase, Redis, or network deps required
  - `TestBaselineMemoryModel` (5 tests): directly instantiates `MemoryDB`, asserts `is_baseline` default, `True` assignment,
  `model_dump()` output, and dict roundtrip
  - `TestBaselineMemoryInjection` (7 tests): loads `utils.llms.memory` via `stub_modules` + `load_module_fresh` (project-standard
  isolation primitives) and patches `resolve_memory_system`, `memories_db.get_memories`, and `get_user_name` to exercise actual bucket
   routing and prompt-string formatting

  ### Mobile Client (`app/`)

  **Schema** (`app/lib/backend/schema/memory.dart`)
  - Added `isBaseline` field with serialization (`is_baseline` ↔ `isBaseline`)

  **API** (`app/lib/backend/http/api/memories.dart`)
[...]

PR #10455, opened 2026-07-24
fix(backend): name the surface in the OpenAPI regenerate hint

### LINKED ISSUE #10217: export_openapi.py stale-contract error message omits the required --app-client flag

aryanorastar:
## What

The OpenAPI contract checker tells you how to fix a stale contract, and the command it prints is wrong for every non-public surface.

Fixes #10217.

## Why it matters

`check_spec` printed:

```
<path> is stale; run backend/scripts/export_openapi.py --write <path>
```

`--surface` **defaults to `public`**. So running that hint verbatim against the app-client contract exports the *public* surface into the app-client file — it doesn't refresh the contract, it guts it (~53k lines). The hint actively causes the damage it claims to repair.

I hit this myself regenerating contracts today: you have to already know to pass `--surface app-client`, which is precisely what the error message exists to tell you.

## Fix

Thread the surface into `check_spec` and build both the missing- and stale-file messages from one `regenerate_hint` helper, so the printed command is exactly the command that regenerates that path:

```
<path> is stale; run backend/scripts/export_openapi.py --surface app-client --write <path>
```

Verified live by making the app-client contract stale and running the checker (contract restored afterwards, not part of the diff):

```
OpenAPI contract check failed: .../app-client-openapi.json is stale;
run backend/scripts/export_openapi.py --surface app-client --write .../app-client-openapi.json
```

## Tests

`backend/tests/unit/test_export_openapi_hint.py` — 7 cases:
- every surface's hint names that surface and its own path
- both the missing-file and stale-file errors carry the surface
- the regression itself: the app-client hint is never the bare `--write <path>` form that defaults to public

## Note on prior attempts

#10255 and #10294 both targeted this and were closed unmerged, so the issue is still live on `main` — I verified the surface-less hint is still there before writing anything.

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/export_openapi.py
+++ b/backend/scripts/export_openapi.py
@@ -989,12 +989,23 @@ def write_spec(path: Path, generated: str) -> None:
     path.write_text(generated)
 
 
-def check_spec(path: Path, generated: str) -> None:
+def regenerate_hint(path: Path, surface: str) -> str:
+    """The exact command that regenerates `path`.
+
+    `--surface` has to be explicit: it defaults to `public`, so a hint that omits it
+    sends the reader to overwrite a non-public contract with the public surface, which
+    quietly guts the file instead of refreshing it (#10217).
+    """
+    return f'backend/scripts/export_openapi.py --surface {surface} --write {path}'
+
+
+def check_spec(path: Path, generated: str, *, surface: str) -> None:
+    hint = regenerate_hint(path, surface)
     if not path.exists():
-        raise OpenAPIContractError(f'{path} does not exist; run export_openapi.py --write {path}')
+        raise OpenAPIContractError(f'{path} does not exist; run {hint}')
     current = path.read_text()
     if current != generated:
-        raise OpenAPIContractError(f'{path} is stale; run backend/scripts/export_openapi.py --write {path}')
+        raise OpenAPIContractError(f'{path} is stale; run {hint}')
 
 
 def parse_args() -> argparse.Namespace:
@@ -1047,7 +1058,7 @@ def main() -> int:
             print(f'wrote {path}')
         elif args.check is not None:
             path = resolve_spec_path(args.surface, args.check)
-            check_spec(path, generated)
+            check_spec(path, generated, surface=args.surface)
             print(f'{path} is up to date')
         return 0
     except OpenAPIContractError as e:
--- a/backend/scripts/export_openapi.py
+++ b/backend/scripts/export_openapi.py
@@ -999,7 +999,10 @@ def regenerate_hint(path: Path, surface: str) -> str:
     return f'backend/scripts/export_openapi.py --surface {surface} --write {path}'
 
 
-def check_spec(path: Path, generated: str, *, surface: str) -> None:
+def check_spec(path: Path, generated: str, *, surface: str = 'public') -> None:
+    # Default matches the --surface default so existing callers (and the public
+    # contract test) stay valid; the production caller passes surface explicitly
+    # so a non-public surface never silently gets the public regenerate hint.
     hint = regenerate_hint(path, surface)
     if not path.exists():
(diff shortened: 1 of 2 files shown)


Git-on-my-level (dismissed):
Thanks for tightening this hint — including the explicit `--surface` is the right direction and the new regression coverage captures the damaging non-public default case.

One blocking issue before this can merge: the PR changes `check_spec` to require a keyword-only `surface`, but an existing unit test/call contract still invokes it with the old signature. CI is failing in `tests/unit/test_openapi_contract.py::test_check_spec_detects_stale_file` with:

```text
TypeError: check_spec() missing 1 required keyword-only argument: 'surface'
```

Please either keep `check_spec` backward-compatible with a safe default such as `surface='public'`, or update the existing test/call sites in the same PR so the backend unit suite is green. The core fix looks sound once that compatibility/test failure is resolved.

---

aryanorastar:
Fixed — `check_spec` now defaults `surface='public'` (matching the `--surface` argparse default), so `check_spec(path, generated)` works again and `test_check_spec_detects_stale_file` is green. The production caller still passes `surface` explicitly, so a non-public surface never silently gets the public hint. Added a regression test pinning the no-surface call.

The remaining `app-client-openapi.json is stale` failure is pre-existing main drift from the baseline toggle (#8728), fixed separately in #10482 — not from this diff.

kodjima33 (approved):
fix(backend) name surface in OpenAPI regenerate hint — approve only (workflow-review, CHANGES_REQUESTED)

Git-on-my-level (commented):
Thanks for the follow-up. I re-reviewed the current head and the earlier blocker is resolved: `check_spec` is backward-compatible again with `surface='public'`, while the CLI check path now passes the parsed surface explicitly so non-public OpenAPI contracts get a safe regenerate hint.

I also ran the targeted backend coverage locally:

```text
.venv/bin/python -m pytest tests/unit/test_export_openapi_hint.py tests/unit/test_openapi_contract.py::test_check_spec_detects_stale_file -q
9 passed in 0.33s
```

This looks like the right fix for #10217, with regression coverage for the app-client/public-surface failure mode. Because this touches the backend OpenAPI contract-generation workflow, I’m leaving it as a positive maintainer signal rather than a formal approval; a human maintainer should still do the final workflow/API-contract signoff before merge.

---

kodjima33 (approved):
Regenerate hint now names --surface, so following it can't gut a non-public contract (#10217). Tested.

aryanorastar:
Rebased onto current main — the app-client spec drift it inherited is fixed on main now. **Green (0 failing), MERGEABLE.** The check_spec backward-compat fix + regression test are in. Ready for re-review.

Issue #10459, opened 2026-07-24
Google Calendar/Gmail connectors: migrate cookie-scraping to real OAuth (v20 cookies + non-Chromium sessions fail)


Archit-lal:
## Symptom
Connecting Google **Calendar** and **Gmail** frequently "fails immediately" for real users (observed in an onboarding user test — both connectors failed identically), and even a successful probe often can't read the user's Google session.

## Root cause (evidence)
Calendar/Gmail are not OAuth — they are a **Chromium browser-cookie scraper**. `CalendarReaderService` / `GmailReaderService` shell out to Python to decrypt Google auth cookies (`SAPISID`, `__Secure-1PSID`, …) from a Chromium-family browser's cookie DB using that browser's Keychain "Safe Storage" password (`BrowserGoogleSession.swift`), then call Google with a `SAPISIDHASH` header. This has several inherent failure modes that no relabeling can fix:

1. **v20 app-bound cookies** — modern Chrome/Edge re-encrypt cookies as `v20`, which the scraper deliberately skips (`BrowserGoogleSession.swift` ~L77-85: "we can't decrypt it … upgrade path is Google OAuth"), so the user reads as `not_signed_in` even though they are signed in. Covered by `BrowserGoogleSessionTests.testUnknownVersionCookieBlobIsSkippedInsteadOfEmittedAsPlaintext`.
2. **Unsupported browsers** — only Chromium targets are supported (`BrowserAutomationTarget.knownTargets`); a user signed into Google in **Safari or Firefox** gets `no_browser` → "immediately fails."
3. **Keychain Safe Storage denial** — if the "… Safe Storage" keychain read is declined or the code-signing identity doesn't match the grant, the browser is dropped.

## Proposed fix
Replace the cookie scraper with real **Google OAuth** for Calendar/Gmail (or add v20 app-bound cookie support). This is the durable fix the code comment itself points to.

## Context
PR BasedHardware/omi#10457 made the *readable-session* path actually connect + persist (onboarding now writes the same `lastSyncedAt` latch the Apps page reads) and made failures honest (real error vs. needs-sign-in), but the underlying auth mechanism remains cookie-scraping. This issue tracks the OAuth migration.

PR #10463, opened 2026-07-24
fix(desktop): onboarding Google (Calendar/Gmail) connect actually persists

Archit-lal:
## Onboarding Google (Calendar/Gmail) connect: actually persist, and fail honestly

Split out from the onboarding-usability PR (BasedHardware/omi#10457) so it can be reviewed on its own.

### The bug
In a recorded onboarding user test, connecting **Google Calendar** and **Gmail** failed, and even when the chip briefly showed "connected" during onboarding, the main **Apps page still read "Not connected."**

### Root cause
Two distinct problems:

1. **Parity (fixed here).** Onboarding connected Calendar/Gmail by only running a transient `verifyConnection()` probe and flipping an in-memory chip — it **never wrote the `lastSyncedAt` latch** that the Apps page reads for "connected", and never ran the real import. So the two surfaces had **two different owners of connector connected-state** and disagreed (`FC-split-mutation-authority`).
2. **Environmental (tracked separately, not fixable here).** The connectors are a **Chromium cookie scraper** (not OAuth) — they decrypt the Google auth cookies from a Chromium-family browser. Modern **v20 app-bound cookies** or a **Safari/Firefox** session simply can't be read, which is the actual "immediately fails." The durable fix is real **OAuth**, tracked in BasedHardware/omi#10459.

### What this PR changes
- Extracts the Apps page's run+persist logic into a shared **`ConnectorImportRunner.startPersistingImport(...)`** seam, and has onboarding's Calendar/Gmail connect call the **same seam** — running the real import and writing the same `lastSyncedAt` latch (fire-and-forget; the runner survives the step and dedups). Now onboarding "on" == Apps-page "connected".
- Maps the onboarding id `"gmail"` → the status-store id `"email"` (writing `"gmail"` would land on a key nothing reads), and reflects an already-persisted connection back into onboarding.
- Splits the connect outcome so a real `.error` (e.g. the Calendar API key not yet loaded — signing in wouldn't help) is **distinct** from `.needsSignIn` (open the Google page), via a pure, unit-tested `googleConnectAction`.

Net: a **readable** Google session now actually connects and persists everywhere; a broken session fails **honestly**. The environmental cookie limitation is #10459.

### Also included (pre-existing upstream test defects this PR trips)
Because this PR touches desktop code, CI runs `Desktop Swift Test Suites` (skipped on upstream `main`), which surfaced two broken-on-upstream tests unrelated to this change:
- A **duplicate `@MainActor`** on a barge-in failover test (rejected by the pinned Swift toolchain).
- An **obsolete source-scraping assertion** in `AgentRuntimeProcessTests` whose expected string an upstream refactor removed.
Both are fixed here so the suite is green.

## Testing
- `swift build` — clean; `OnboardingGoogleConnectTests` (4, incl. a behavioral persist-through-runner test) pass.
- Note: the cookie/browser read shells out to Python and is not hermetic; the persistence + decision logic is unit-tested.

## Product invariants
- **INV-INT-1** (integrations harness) — `ConnectorImportRunner.swift` / connector code: adds the shared run+persist seam so onboarding and the Apps page share one connected-state owner. No connector-harness contract change.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
@@ -2,7 +2,7 @@
   "files": {
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": 2375,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2639,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4849,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4829,
     "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2584,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift": 1534,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1584
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-mainwindow.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-mainwindow.json
@@ -1,7 +1,7 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift": 1969,
-    "desktop/macos/Desktop/Sources/MainWindow/Pages/AppsPage.swift": 3643,
+    "desktop/macos/Desktop/Sources/MainWindow/Pages/AppsPage.swift": 3632,
     "desktop/macos/Desktop/Sources/MainWindow/Pages/DashboardPage.swift": 4467,
     "desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift": 3298,
     "desktop/macos/Desktop/Sources/MainWindow/Pages/TasksPage.swift": 5706,
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
@@ -1,6 +1,6 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6280,
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6268,
     "desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift": 2861
   },
   "raise_justifications": {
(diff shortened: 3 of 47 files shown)


Git-on-my-level (commented):
Thanks for splitting this out — the core direction looks sensible to me. Reusing the Apps page import/persist seam for onboarding and mapping onboarding `gmail` to the app-wide `email` connector key are the right shape for making the onboarding state agree with the Apps page.

I’m not approving this automatically because the Desktop Swift gate is still red: `Desktop Swift Test Suites` was cancelled after about an hour, which makes `Desktop Swift Build & Tests` fail as well. Please get that suite green or have a maintainer intentionally rerun/triage the timeout before merge.

One thing I’d like a human maintainer to double-check before landing: onboarding now marks the chip `on` as soon as the functional Google probe says connected, while the real import/persist runs fire-and-forget afterward. That is probably okay if the import operation is expected to succeed whenever the probe succeeds, but if import can still fail independently, onboarding may still briefly show a successful connection that never reaches the persisted Apps-page latch. The new regression tests cover the shared persistence seam and the Gmail→email mapping; they don’t cover the onboarding UI state after an import failure.

---

kodjima33 (approved):
fix(desktop) onboarding Google connect persists — approve only (UNKNOWN merge state blocks owner-override merge)

Git-on-my-level (commented):
Thanks for the update on this head. I re-reviewed the current diff and the new `fix(desktop): await onboarding Google import persistence` commit addresses the main concern I raised earlier: onboarding now waits for the shared import/persist task and derives the final chip state from the same persisted connector latch the Apps page reads, instead of marking the connector `on` before persistence has actually happened.

What looks good:
- Calendar/Gmail now both go through `ConnectorImportRunner.startPersistingImport(...)`, so success writes the shared `lastSyncedAt` state.
- The onboarding `gmail` id is mapped to the app-wide `email` connector id before persistence.
- The sign-in vs real-error behavior is separated, so real connector errors no longer open a misleading browser sign-in prompt.
- The added tests cover the ID mapping, successful persistence latch, failed import not marking connected, and the retryable state after a non-persisted import. I also ran the desktop test-quality checker locally (`python3 scripts/check_desktop_test_quality.py`) and it passed.

I’m still not formally approving because GitHub currently reports no checks for this head, and this touches desktop connector/import behavior that should get a macOS Swift test/build run or explicit human maintainer sign-off before merge. Once `Desktop Swift Build & Tests` (or equivalent maintainer-run macOS validation) is green, this looks reasonable to land from my review.

---

kodjima33 (approved):
Onboarding Google connect persistence fix looks right; branch has conflicts with main so not merging.

undivisible:
**Maintainer note (rebase attempt):** This PR’s functional fix has been absorbed by main in a stronger form via #10806 (and follow-on onboarding hardening): `importConnectorID(forGoogleContextID:)`, `startGoogleContextImport` / `completeGoogleContextImport` (markSynced + row state), and regression coverage in `SBOnboardingCloudContextStateTests`.

Rebase onto current main conflicts with that landed design (main no longer uses the older `resolveGoogleConnect` + `startPersistingImport` shape). Recommend **closing as superseded by #10806** rather than re-landing a parallel path. Original tip left as-is for author history.

Git-on-my-level:
Thanks for splitting this out and for the focused regression coverage. I’m closing this as superseded rather than asking for another rebase: the functional fix has already landed on main in #10806, including the canonical Google import/status path and onboarding cloud-context regression tests, and this branch now conflicts with main’s newer design. Appreciate the work here — the original issue is covered by the landed follow-up.

PR #10478, opened 2026-07-24
fix(desktop): stop cold-start ACP init from awaiting Claude OAuth in-band

### LINKED ISSUE #10407: Desktop chat: cold-start initializeAcp() still awaits OAuth in-band (same stall as .113, at bridge init)
## Problem

The `.113` chat-freeze fix (ACP `-32000` under piMono → terminalize-first) severed the in-band OAuth `await` on the **active-turn** recovery paths (`onRecoverableError` / `recoverAfterError` now signal Swift and return `false`). But the **cold-start bridge-init path still awaits OAuth in-band**, so the same 60s-silence → 2-minute-OAuth-timeout stall survives when ACP requires auth at initialization instead of mid-turn.

## Evidence

`desktop/macos/agent/src/index.ts:1029-1052` — `initializeAcp()` catches `AcpError` code `-32000` and does:

```ts
logErr(`ACP requires authentication: ${JSON.stringify(authMethods)}`);
await startAuthFlow();      // blocks up to 2 min; Swift startClaudeAuth() is a silent no-op under piMono
await initializeAcp();      // retry after auth
```

`startAuthFlow()` (`index.ts:951`) awaits the OAuth callback (2-minute timeout). Under piMono preference, Swift's `startClaudeAuth()` is gated on `isUserClaudeMode` and never opens the browser, so the callback never arrives — identical to the incident, just triggered at bridge init (first send after launch, or the restart path at `index.ts:1550`) rather than on an admitted turn. The active-turn fix does not cover this call site; it was consciously deferred in the incident synthesis (`Cold-start initializeAcp() OAuth policy`, `.coordination/chat-freeze-113/committee-synthesis.md` §7 "Do not do in this PR").

## Suggested scope (one self-contained PR)

Apply the same terminalize-first contract to init: on `-32000` during `initializeAcp()`, signal `auth_required` to Swift and surface a terminal provider-auth state instead of `await startAuthFlow()`. The first user send then fails fast with `failureCode: authentication` rather than hanging. Falsifiable test (mirrors the existing `acp-auth-terminal.test.ts` T1): cold-start ACP init injects `-32000` with no OAuth callback → init returns without blocking (<1s), no `startAuthFlow` await, subsequent send terminalizes as `authentication`.

aryanorastar:
## What

Closes #10407.

Cold-start ACP `initializeAcp()` awaited `startAuthFlow()` when the agent returned `-32000` (auth required), then retried init. With expired Claude credentials and no OAuth callback, init sat inside that await for the **entire callback timeout** — and because no turn was on the wire yet, the user's very first send had nothing to terminalize behind it and simply **hung** until the timeout elapsed.

This applies the same terminalize-first contract the active turn already follows (#10417): signal `auth_required` to Swift, then **start — never await —** the OAuth flow and return. `isInitialized` stays `false`, so the pending send reaches the ACP request, hits the same `-32000`, and terminalizes as `authentication` through the kernel's existing recoverable-error path instead of blocking.

The flow is still *started*, not dropped: Swift `ChatProvider.startClaudeAuth` refuses to open sign-in without the bridge-issued `authUrl` that only this flow mints, so dropping the call would strand the user in an auth-required state with no way to sign in. The bug was the `await`, not the call.

## Why a shared seam + failure class

`index.ts` cannot be imported under vitest (module scope spawns the ACP subprocess and binds stdin), so its coverage is source-scraping — a static tripwire, not behavior. The kickoff is extracted to `acp.ts` as `beginProviderAuthWithoutBlocking`, a seam with a real injectable boundary, and tested behaviourally.

This is the **second** instance of one root cause — an ACP path awaiting interactive OAuth in-band so nothing can terminalize. The first (#10417) hardened the active-turn path. Per AGENTS.md ("two or more recent fixes share the cause → add a reusable guard surface"), this declares `Failure-Class: new` and registers **`FC-in-band-auth-await-stall`** (`.github/failure-classes/`) citing both PRs, with `beginProviderAuthWithoutBlocking` as the code-level guard.

## Test

`desktop/macos/agent/tests/acp-init-auth-nonblocking.test.ts` (new, mirrors the issue's falsifiable T1):

- Returns without waiting for the OAuth callback (never-settling flow ⇒ control still reaches the assertions).
- Signals `auth_required` **before** the flow produces anything, so a pending send terminalizes as `authentication`.
- Still starts the flow — guards the tempting over-fix of dropping the call, which would leave Swift with no `authUrl`.
- A failed flow logs instead of leaking an unhandled rejection.

### Verification evidence

```
# agent package (vitest)
acp-init-auth-nonblocking.test.ts .... 4 passed
acp-auth-terminal.test.ts ............ 2 passed
+ acp-tool-activity / protocol-v2 / adapter-selection ... 27 passed total

# mutation check — reintroduce `await startAuthFlow()` before the signal:
acp-init-auth-nonblocking.test.ts .... 3 failed / 1 passed  (as intended)

tsc --noEmit ......................... exit 0
scripts/failure-class validate ....... passes
```

I exercised the extracted seam directly (the real code path init now runs); I did not boot the full macOS app for this change — the behaviour lives entirely in the agent bridge's auth kickoff, covered by the seam above.

## Scope

Agent-bridge TypeScript only — no Swift touched. User-facing changelog fragment included.

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10478?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-in-band-auth-await-stall.json
+++ b/.github/failure-classes/FC-in-band-auth-await-stall.json
@@ -1,0 +1,9 @@
+{
+  "schema_version": 1,
+  "id": "FC-in-band-auth-await-stall",
+  "violated_contract": "An ACP path that hits a provider auth failure must terminalize the pending turn and leave OAuth out-of-band; it must never `await` the interactive sign-in flow in-band, because that blocks the caller for the whole callback timeout with nothing on the wire to fail fast.",
+  "canonical_prevention": "Signal `auth_required` to the host first, then start (never await) the OAuth flow so the pending send terminalizes as `authentication`; route both the active-turn and cold-start-init paths through the shared beginProviderAuthWithoutBlocking / signalProviderAuthRequired seam rather than awaiting startAuthFlow.",
+  "evidence_prs": [10417, 10407],
+  "scope_hints": ["desktop/macos/agent/src/**"],
+  "status": "open"
+}
--- a/desktop/macos/agent/src/adapters/acp.ts
+++ b/desktop/macos/agent/src/adapters/acp.ts
@@ -154,6 +154,42 @@ export function isAcpProviderAuthFailure(error: unknown): boolean {
 /** @deprecated Renamed to {@link isAcpProviderAuthFailure}; classify-only, no in-band recovery. */
 export const isRecoverableAcpAuthError = isAcpProviderAuthFailure;
 
+/** Seams for {@link beginProviderAuthWithoutBlocking}. */
+export type ProviderAuthKickoffDeps = {
+  /** Tell the host provider auth is required, so a pending turn can terminalize now. */
+  signalAuthRequired: () => void;
+  /**
+   * Run the OAuth flow: local callback server, token exchange, credential
+   * storage, ACP restart. It emits its own `auth_required` carrying the
+   * bridge-issued `authUrl` once one exists.
+   */
+  startAuthFlow: () => Promise<void>;
+  logErr: (message: string) => void;
+};
+
+/**
+ * Begin provider auth without blocking the caller.
+ *
+ * `initialize` used to `await` the OAuth flow, so a cold start with expired
+ * credentials sat inside init for the whole callback timeout with no turn on
+ * the wire to terminalize — the user's first send hung instead of failing fast
+ * (#10407). Signalling first lets that send terminalize as `authentication`
+ * immediately.
+ *
+ * The flow is still *started*, just not awaited: the host's sign-in button needs
+ * the `authUrl` only this flow can mint (Swift `ChatProvider.startClaudeAuth`
+ * refuses to open sign-in without one), so dropping the call outright would
+ * strand the user in an auth-required state with no way out.
+ */
+export function beginProviderAuthWithoutBlocking(deps: ProviderAuthKickoffDeps): void {
+  deps.signalAuthRequired();
+  // Detached on purpose — awaiting is the bug. The rejection is swallowed here so
+  // a failed flow can never surface as an unhandled rejection.
+  void deps.startAuthFlow().catch((error) => {
+    deps.logErr(`ACP background auth flow failed: ${error instanceof Error ? error.message : String(error)}`);
+  });
+}
+
 const MAX_RECENT_STDERR_CHARS = 2_000;
 
 const EXTERNAL_TERMINAL_HTTP_FAILURE = /^\s*HTTP\s+([45]\d{2})\s*:\s*(?:\{|\[)/i;
--- a/desktop/macos/agent/src/index.ts
+++ b/desktop/macos/agent/src/index.ts
@@ -80,7 +80,12 @@ import {
 import { startOAuthFlow, type OAuthFlowHandle } from "./oauth-flow.js";
 import { isProductionAdapterId, type PromptBlock, type RuntimeAdapter } from "./adapters/interface.js";
 import { detectImageMimeType } from "./mime-detect.js";
-import { AcpError, AcpRuntimeAdapter, isAcpProviderAuthFailure } from "./adapters/acp.js";
+import {
+  AcpError,
+  AcpRuntimeAdapter,
+  beginProviderAuthWithoutBlocking,
+  isAcpProviderAuthFailure,
+} from "./adapters/acp.js";
 import { AdapterRegistry } from "./runtime/adapter-registry.js";
 import { nextJournalPumpDelayMs } from "./runtime/journal-pump-backoff.js";
 import { JsonlTransport, type McpServerBuildContext } from "./runtime/jsonl-transport.js";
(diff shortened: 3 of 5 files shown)


Git-on-my-level:
Thanks for the focused fix here. The shape looks directionally right to me: extracting a non-blocking auth kickoff seam and reusing it from cold-start ACP init addresses the hang without dropping the OAuth flow entirely, and the new Vitest coverage guards the important ordering/rejection behavior.

I’m not approving this yet because it touches the Claude auth/ACP initialization path and the required Desktop Swift test gate is currently failing. Before merge, I’d want a maintainer to verify the end-to-end desktop behavior with expired Claude credentials, especially that repeated first-send/auth-required attempts do not spawn confusing duplicate sign-in prompts and that successful OAuth still restarts/re-initializes ACP cleanly.

No concrete code change request from me based on the static diff, but this needs the failing CI resolved or explained plus human maintainer sign-off for the auth-flow behavior.

---

aryanorastar:
CI is resolved — **0 failing checks** on the rebased head `30fac1f7` (11 success / 15 skipped / 1 neutral), including `Desktop Swift Build & Tests`.

The failure was not this diff. Rather than argue that, I rebased onto current main and let CI answer it, but the diagnosis holds up on the evidence too:

- This PR changes **5 files, none of them Swift** (`acp.ts`, `index.ts`, one test, the failure-class JSON, the changelog fragment) — there is no path from it to `PushToTalkStateMachineTests`, `QueryTracerTests`, `QuitAndReopenActionTests` or `RawWebSocketValidationTests`.
- All four failed the same way — `suite timed out after 120s`, which is the harness's own per-suite cap (`SUITE_TIMEOUT_SECONDS` in `desktop/macos/scripts/swift-test-suites.sh`, default 120) — and the same run also logged `timed out after 4s waiting for ./run.sh build lock`. Four suites hitting the wall clock together with build-lock contention is one loaded runner, not four logic failures.
- The `Desktop Swift Build & Tests` red was downstream of that: it is a gate job asserting `TEST_RESULT = success`, which is why it "failed" in 4s without running anything.
- No other open PR currently shows `Desktop Swift Test Suites` failing, and main's own Desktop Swift CI is green.

Re-verified on the rebased head: agent suites 6/6 (`acp-init-auth-nonblocking` 4, `acp-auth-terminal` 2), `tsc --noEmit` exit 0, and `scripts/failure-class validate` → `ok: true` with no errors.

Your substantive ask stands and I have not tried to satisfy it from here: the end-to-end check with genuinely expired Claude credentials — no duplicate sign-in prompts across repeated first-send attempts, and a clean ACP re-init after successful OAuth — needs a maintainer on a real desktop build. I called that limitation out in the PR body and it is unchanged.

aryanorastar:
Correcting my previous comment — I got the diagnosis partly wrong, and the green is weaker than I implied.

**What I got wrong.** I attributed the four timed-out suites to a single contended runner. That is not right for the first one: `PushToTalkStateMachineTests.testOwnerTransitionTerminatesActiveHubAndDrainsItsPhysicalSession` is a **known hang on main**, tracked in #10492 and #10499 — it blocks on `await transitionOwner(... to: "owner-b")`, whose hub drain never completes headless. #10499 also records that peer suites then fail on SwiftPM lock wait, which is the real explanation for `QueryTracerTests`, `QuitAndReopenActionTests` and `RawWebSocketValidationTests` and for the `run.sh build lock` timeout I quoted. So it was one genuine hang plus collateral, not four victims of load.

**What I overstated.** `Desktop Swift Build & Tests` is green on `30fac1f7`, but `Desktop Swift Test Suites` **did not run** on that head — it was skipped by the path filter, where the earlier run had executed it. So the green means "this diff no longer triggers the Swift lane", not "the hang is fixed". I should have checked that before calling CI resolved.

**What still holds** — and is now better supported than my original argument: this failure is not caused by this PR. #10492 records that it reproduces on a branch with no product changes at all, so it is a main-side regression, and this PR's 5 files contain no Swift.

The rest of my verification stands: agent suites 6/6, `tsc --noEmit` exit 0, `failure-class validate` → `ok: true`. And your substantive ask is unchanged — the expired-credentials end-to-end still needs a maintainer on a real build.

I am picking up #10492 separately rather than leaving it as a footnote here.

Git-on-my-level:
Thanks for the follow-up and for correcting the CI interpretation.

I re-reviewed the current head (`30fac1f7`) and ran the focused agent checks locally in a stripped environment:

- `npm test -- tests/acp-init-auth-nonblocking.test.ts` — 4/4 passing
- `npm run build` — passing (`tsc`)
- `python3 scripts/failure-class validate --base origin/main --head HEAD --pr-body-file /tmp/omi-pr-10478-body.md` — `ok: true`

Static review still looks positive to me: the change preserves the needed OAuth kickoff while moving the interactive wait out of the cold-start init path, and the extracted seam covers the important ordering and rejection behavior. I do not see a concrete code change to request from this diff.

I’m still not formally approving because this is Claude auth / ACP initialization behavior and the remaining risk is end-to-end desktop UX rather than just TypeScript correctness. Before merge, a maintainer should verify on a real desktop build with expired Claude credentials that repeated first-send/auth-required attempts do not create confusing duplicate sign-in prompts, and that completing OAuth restarts/re-initializes ACP cleanly.

---

PR #10482, opened 2026-07-24
chore: regenerate stale generated artifacts from the memories baseline toggle

aryanorastar:
## What

The memories baseline toggle added `is_baseline` and
`PATCH /v3/memories/{memory_id}/baseline`, but the committed artifacts derived
from those contracts were stale.

This PR now contains only reproducible generated output:

| Artifact | Change |
|---|---|
| `app-client-openapi.json` | Adds the baseline-memory PATCH route |
| `integration-public-openapi.json` | Adds the `is_baseline` field |
| Dart `memories_wire.g.dart` | Adds `isBaseline` |
| Swift `OmiApi.generated.swift` | Adds `isBaseline` and the baseline route |
| TypeScript clients ×4 | Add `is_baseline` and the baseline route |

The former stacked #10481 CI/failure-class changes are no longer part of this
branch; this PR is independently based on current `main`.

## Verification

- `PATH=/opt/homebrew/share/flutter/bin:$PATH dart run build_runner build --delete-conflicting-outputs` — wrote 0 outputs on the committed tree
- `flutter gen-l10n` followed by `git diff --exit-code -- lib/l10n` — clean
- `scripts/openapi_runner.sh scripts/export_openapi.py --check ../docs/api-reference/openapi.json` — up to date
- `scripts/openapi_runner.sh scripts/export_openapi.py --surface app-client --check ../docs/api-reference/app-client-openapi.json` — up to date
- `scripts/openapi_runner.sh scripts/export_openapi.py --surface integration-public --check ../docs/api-reference/integration-public-openapi.json` — up to date
- `.openapi-venv/bin/python scripts/generate_dart_models.py --all --check` — all outputs up to date
- `.openapi-venv/bin/python scripts/generate_swift_openapi_types.py --check` — clean
- `.openapi-venv/bin/python scripts/generate_ts_openapi_types.py --check` — clean
- `backend/.venv/bin/python -m pytest backend/tests/unit/test_app_client_*_generator.py backend/tests/unit/test_integration_public_contract.py` — 54 passed
- `bash app/test.sh` — 944 passed
- `bash scripts/analyze_ratchet.sh` — passed (11 `unnecessary_import` findings versus baseline 12)
- `PRE_PUSH_SKIP_DESKTOP_CHANGELOG=1 make preflight` — all 17 selected checks passed; the skip corresponds to the `no-changelog-needed` label for generated Swift

No product behavior or hand-written source changes.

PR #10485, opened 2026-07-24
fix(dev-api): serve legacy memories for a legacy-cohort vector search (#10203)

### LINKED ISSUE #10203: Regression? GET /v1/dev/user/memories still returns 403 missing_rollout_state for some accounts after #10094
## Summary
A Discord help user reports the same `GET /v1/dev/user/memories` failure is still happening on July 21, 2026, even after the fix in #10094 was merged.

## Current behavior
`GET /v1/dev/user/memories` returns:

```json
{
  "detail": {
    "enabled": false,
    "consumer": "developer_api",
    "archive_capability": false
  }
}
```

## Repro / notes
- User says it started around **2026-07-11** and still reproduces on **2026-07-21**.
- Rotating to a fresh API key does **not** fix it.
- `GET /v1/dev/user/conversations` on the same key returns **200 OK**.
- This still looks account/rollout-state specific rather than key-specific.
- User had already filed support around **2026-07-17** with no response yet.

## Why this looks new
- #9892 captured the original bug.
- #10094 was merged to serve legacy memories when a legacy-cohort account has no rollout state.
- This report suggests either:
  1. the fix is not fully deployed everywhere,
  2. some affected accounts are outside the legacy-fallback path, or
  3. there is another rollout-state path still failing closed.

## Ask
Please verify whether the #10094 backend fix is fully deployed and whether any remaining account cohorts can still hit `missing_rollout_state` / `archive_capability: false` on the Developer API memories endpoint.

aryanorastar:
## What

Closes #10203.

`GET /v1/dev/user/memories/vector/search` failed closed with `403 missing_rollout_state` for un-enrolled **legacy-cohort** accounts — even though the sibling read paths already recover for exactly these accounts:

- the memory list `GET /v1/dev/user/memories` (#9892 / #10094), and
- the MCP memory path (#10095),

both serve the legacy `memories` collection when `pin_memory_system` has resolved the account to LEGACY and the only signal is an absent rollout doc. So an affected account could **list** its memories but not **vector-search** them — the "another rollout-state path still failing closed" cohort the issue asks about (possibility #3).

Worse, `search_memories_vector` raised 403 on `should_use_legacy_fallback` (`USE_LEGACY_SAFE`) — the one signal that explicitly means *serve legacy*.

> Note on the user's report: the **list** endpoint (`GET /v1/dev/user/memories`) is already fixed in code, so a user still seeing 403 there is a deploy-lag symptom (issue possibility #1), not a code path. This PR closes the genuine remaining code gap — the vector endpoint.

## Fix

On a legacy-cohort signal — `USE_LEGACY_SAFE`, or a deny whose **only** reason is `missing_rollout_state` — search the legacy `memories` vector index (`search_memories_by_vector`) and hydrate the hits (`get_memories_by_ids`), instead of returning 403:

- **vector-relevance order preserved** (iterate the ranked IDs, not hydration order);
- **locked content redacted** exactly as the list route does;
- `relevance_score` left `null` (the legacy index doesn't expose scores);
- `record_fallback(... to_mode='legacy_memories', outcome='recovered')` marks the recovery;
- **any other deny reason keeps the fail-closed 403.**

## Why not a shared predicate (yet)

This is the 4th instance of one cause — a default-read surface failing closed for an un-enrolled legacy account instead of serving legacy (#9892 → #10094 list, #10095 MCP, this = vector). The ideal class-level guard is one predicate every read surface calls. But `get_memories` reaches legacy by *fallthrough* while the vector route returns explicitly, so unifying them ripples into the list route's control flow and its static tripwire. Per AGENTS.md ("land the enforceable guard now, track high-blast-radius follow-up explicitly") I land the focused fix plus the updated tripwire (which now asserts the vector route recovers on the shared signal) and flag the shared-predicate extraction as a follow-up.

## Verification

Ran with a real backend venv (Python 3.11):

```
test_dev_api_canonical_grant_ordering.py ......... 9 passed
  incl. 3 new vector cases:
    - missing_rollout_state → legacy fallback (200), vector order preserved
    - USE_LEGACY_SAFE       → legacy fallback (200)
    - other deny reason     → still 403
mutation: revert the fix → the 2 fallback cases fail, the 403 case still passes
144 passed across every vector-referencing unit test
```

The updated `test_developer_routes_only_reach_legacy_after_explicit_legacy_safe_decision` tripwire now encodes the new (consistent) contract for both routes.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10485?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
@@ -2,14 +2,14 @@
   "files": {
     "backend/routers/apps.py": 2335,
     "backend/routers/chat.py": 1588,
-    "backend/routers/developer.py": 2202,
+    "backend/routers/developer.py": 2263,
     "backend/routers/mcp_sse.py": 1858,
     "backend/routers/sync.py": 2058,
     "backend/routers/users.py": 2046
   },
   "raise_justifications": {
     "backend/routers/chat.py": "Merging the Windows-port chat surface with main combined the stream-error fallback (answered guard) with mains journey/attempt observability in the same SSE generators; +11 lines of combined guard flow, no new route.",
-    "backend/routers/developer.py": "From-segments processing now wraps process_conversation in the admission lease guard so the crash-orphan sweep cannot terminalize active work (#10461); +7 lines for the conditional guard with rollback_on_failure=False.",
+    "backend/routers/developer.py": "GET /v1/dev/user/memories and /vector/search both serve the authoritative legacy memories collection for legacy-cohort accounts with no rollout state (#9892/#10203), keeping the narrow deny/legacy-fallback decision in the route that owns the read contract.",
     "backend/routers/mcp_sse.py": "MCP SSE memory reads route their legacy-fallback decision through the shared mcp_legacy_read_authorized helper (#9892); one import line.",
     "backend/routers/sync.py": "Fresh Sync's daily ceiling remains beside the existing fresh-admission gates, before staged audio or durable worker work is created. Lifecycle-fenced workers are terminalized at this Cloud Tasks boundary so released WAL clients do not retry an owner they no longer own; +2 for the queued-dispatch-lost stale finalization path (#10033).",
     "backend/routers/users.py": "GET /v1/users/subscription serializes the new mobile plus/max plans as `unlimited` for clients whose plan enum predates them, so day-one buyers read as paid instead of Free (mirrors the existing operator remap); real limits/grandfather are computed from the true plan first."
--- a/backend/routers/developer.py
+++ b/backend/routers/developer.py
@@ -14,7 +14,7 @@ import database.action_items as action_items_db
 import database.goals as goals_db
 import database.users as users_db
 from database._client import db
-from database.vector_db import upsert_memory_vectors_batch
+from database.vector_db import search_memories_by_vector, upsert_memory_vectors_batch
 
 from models.folder import Folder
 from models.goal import GoalHistoryEntryResponse, GoalMetric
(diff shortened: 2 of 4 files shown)


Git-on-my-level (dismissed):
Thanks for the focused fix here — the fallback is narrow, keeps other deny reasons fail-closed, preserves legacy vector ordering, and the added unit coverage targets the right regression cases.

I’m requesting changes for one concrete merge blocker: the Public Developer API contract check is failing because `docs/api-reference/app-client-openapi.json` is stale after this Developer API route change. Please regenerate and commit the app-client OpenAPI contract (`backend/scripts/export_openapi.py --write ../docs/api-reference/app-client-openapi.json`, or the repo’s equivalent wrapper) so CI is green.

Given this touches Developer API memory reads/vector search over user memories and legacy rollout fallback behavior, I’d still want a human maintainer to sanity-check the rollout/data-handling semantics before merge, but I don’t see a security/supply-chain issue in the diff itself.

---

kodjima33 (approved):
backend dev-api bug fix (serve legacy-cohort memories, #10203) — approve only (CHANGES_REQUESTED)

aryanorastar:
Thanks — but this diff doesn't change the app-client contract. I regenerated it on this branch to check: the result is **+104 lines that are entirely the memories baseline route/field from #8728** (`PATCH …/baseline`, `is_baseline`), and **zero** from this PR's `vector/search` change — the fix is control-flow only (serve legacy vs 403), the `DeveloperMemoryVectorSearchResponse` model is untouched.

So `app-client-openapi.json` is stale on `main` from #8728, not from here — the **OpenAPI Contract** job is red on every backend PR, not just this one.

The catch with regenerating it *in this PR*: that single job also runs the Dart/TS/**Swift** client `--check` steps, so going green needs the whole regenerated chain — including `Desktop/Sources/Generated/OmiApi.generated.swift`, which then trips the desktop-changelog gate. That's exactly why I split it into **#10482** (the full spec + client-type regen; #10481 exempts generated Swift from the changelog gate so it can carry the Swift file).

Once #10482 lands, this goes green with no change. Since @kodjima33 has already approved the substance, could we either merge #10482 first, or treat this OpenAPI-contract failure as the pre-existing #8728 drift it is? Happy to rebase after #10482 to confirm green.

Git-on-my-level (commented):
Thanks for the follow-up here. I re-checked the current head and the previous generated-contract blocker is no longer blocking: the Public Developer API contract and backend checks are passing now, so I dismissed my stale automation change-request review.

Implementation-wise this still looks like a narrow bug fix: auth/default-memory grant checks remain ahead of the memory-system branch, legacy serving is limited to `USE_LEGACY_SAFE` or the `missing_rollout_state` legacy-cohort case, other deny reasons still fail closed, and the tests cover vector-order preservation plus both fallback and deny paths.

Because this touches Developer API vector search over user memories and legacy rollout fallback semantics, I’m leaving this as a positive signal rather than a formal approval; the existing human maintainer review label is still appropriate before merge.

---

kodjima33 (approved):
Legacy-cohort accounts could list but not vector-search memories (#10203); mirrors the existing listing/MCP fallback. Tested.

aryanorastar:
Rebased onto current main. The `app-client-openapi.json is stale` blocker is resolved — it was the pre-existing #8728 drift, which a maintainer fixed on main (`chore(api): refresh generated API artifacts`), not this diff. **Now green (0 failing, 0 pending), MERGEABLE.** The dev-api vector legacy-fallback fix + its mutation-checked tests are unchanged. Re-requesting review.

PR #10487, opened 2026-07-24
fix(backend): let reprocessing reconsider a discarded conversation

mdmohsin7:
Conversations are being left transcribed, untitled, and invisible to their
owners. The record holds a full transcript, an empty title, and a non-terminal
`processing` status that never settles.

In the twenty minutes before opening this, **112 conversations were stranded
across 66 distinct users**. One user has 366, holding 214,792 words of their own
transcripts they cannot see.

## Cause

Two paths disagree about whether a discard is final.

A conversation is discarded when it looks empty. Offline audio for it can arrive
much later, and the sync pipeline merges those segments and asks for a reprocess
precisely because the discard may no longer be right:

```python
# Reprocess if conversation was discarded or if auto-synced WALs added new segments
if closest_memory.get('discarded', False) or target_conversation_id:
    _reprocess_conversation_after_update(uid, closest_memory['id'], language)
```

The lifecycle guard refused every write to a discarded conversation, so that
reprocess ran its transcription and structuring and had the result fenced. The
discard was never lifted, so each later sync repeated the work and lost it
again.

The in-app reprocess is the documented way to force-process a discarded
conversation. It reached the same fence, so the recovery path a user would reach
for was broken by the same guard.

Nothing settled the status either, which is why the records sit at `processing`
and the app keeps polling for conversations that will never complete.

## Change

The guard reopens for reprocessing alone. `is_reprocess` has exactly two
callers, the reprocess endpoint and the sync pipeline's post-merge reprocess,
and both exist to reconsider a discard.

A run that is not reprocessing stays fenced, so an ordinary finalizer still
cannot overwrite a discard it never inspected — `force_process` was not usable
as the signal for this reason, since finalizers set it too.

Deletion remains fenced, and tombstone eligibility remains the caller's
contract, as it already is for a conversation deleted without being discarded.

Nothing else was needed to make the record visible again: reprocessing already
derives `discarded` from whether it produced a title, so the write that lands
clears the flag on its own.

## Recovering what is already stranded

`backend/scripts/revive_fenced_conversations.py` revives conversations stranded
before this ships, following the existing admin-entrypoint pattern.

It clears the flag before reprocessing, because reprocessing a conversation that
is still discarded is fenced exactly as it was and leaves the record no better
off. It checkpoints after each conversation so a long run resumes, paces itself
between reprocess calls, skips soft-deleted tombstones, and takes `--min-words`
so fragments are not revived into a user's list.

```
python scripts/revive_fenced_conversations.py --uid UID --dry-run
```

Verified read-only against an affected user: 366 stranded, 246 eligible at the
default threshold, 214,792 transcript words recoverable.

## Tests

`test_conversation_discard_revival.py` pins both directions: reprocessing
revives a discarded conversation and the write clears the flag; a run that is
not reprocessing stays fenced; revival does not reopen a stale generation or
recreate a deleted conversation; and the intent is forwarded rather than
inferred.

Verified the revival test fails against the previous guard and passes against
this one, so it pins the regression rather than the fix.

`test_conversation_discard_revival` 6, `test_developer_from_segments_idempotency`
9, `test_sync_v2` 168, `test_sync_two_lane` 33, `test_workstream_core` 23 — all
passed.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
fix(backend) reprocessing reconsiders discarded conversation — approve only (failing Public Dev API contract check + >300 lines)

PR #10491, opened 2026-07-24
fix(setup): resolve repo root across the whole pre-push chain in linked worktrees

### LINKED ISSUE #10293: make setup and pre-push hook fail with 'this operation must be run in a work tree' in linked git worktrees
After merging current main (head ~5b74e63582 era, which touched make setup / hook installation — `scripts/test-make-setup.sh` landed recently), both `make setup` and the installed pre-push hook fail inside a **linked worktree** (`git worktree add ...`):

```
$ git push origin <branch>
fatal: this operation must be run in a work tree
error: failed to push some refs to 'https://github.com/BasedHardware/omi.git'

$ make setup
fatal: this operation must be run in a work tree
make: *** [setup-hooks] Error 128
```

Same commands work in the primary checkout. AGENTS.md documents worktrees as the default workflow ("Always work in a git worktree for code changes") and says hooks are installed "with linked-worktree-safe paths", so this looks like a regression in the new setup/hook scripts — likely a `git` invocation that resolves `--git-dir` to the shared common dir and then requires a work tree.

Workaround used: ran the full `make preflight` manifest manually (green) and pushed with `--no-verify`.

kodjima33:
## What

Make the whole pre-push hook chain resolve the repo root without a work tree in linked worktrees.

`#10401` fixed the Makefile and the installed hook **dispatcher**, but the scripts the dispatcher hands off to — `scripts/pre-push`, `scripts/pre-push-singleflight`, `scripts/pr-preflight` — and the Python entrypoints they exec — `.github/scripts/preflight_runner.py`, `.github/scripts/pr_preflight.py` — still resolved the root with an unguarded `git rev-parse --show-toplevel`.

## Why

In a linked worktree whose git context resolves to a git dir rather than a work tree, `git rev-parse --show-toplevel` exits 128 with `fatal: this operation must be run in a work tree`. The dispatcher survived (`#10401` gave it a `|| pwd` fallback), but it then execs `pre-push-singleflight` → `preflight_runner.py` → `pre-push`, each of which re-ran the unguarded call and aborted — so the gate died and the user was forced into a `--no-verify` push that silently bypasses all local checks. That is exactly the reported symptom (`#10293`):

```
$ git push origin <branch>
fatal: this operation must be run in a work tree
error: failed to push some refs to 'https://github.com/BasedHardware/omi.git'
```

## Fix

Fall back to the invoking work tree — git runs hooks with the working directory at the work-tree top, and the dispatcher/gate `cd` into the resolved root before invoking these scripts — matching the fallback `#10401` established for the dispatcher (`git rev-parse --show-toplevel 2>/dev/null || pwd`; the Python resolvers fall back to `Path.cwd()`).

## What I tested

- **Reproduced the exact failure** end-to-end: ran the real installed dispatcher chain under a bare `GIT_DIR` (which makes `show-toplevel` exit 128, identical to the reporter's condition). Before the fix it aborted at `preflight_runner.py:249` with `fatal: this operation must be run in a work tree`; after the fix the gate proceeds past root resolution.
- **New regression test** `scripts/test-pre-push-worktree-fallback.sh` (wired into `.github/checks-manifest.yaml`, `local`+`ci`): asserts all five call sites resolve the work-tree root instead of aborting under the 128 condition. Verified it **fails** against the pre-fix scripts and **passes** after.
- Happy path unchanged: every resolver returns the correct root with no `GIT_DIR` override; existing `scripts/test-make-setup.sh` and the manifest contract self-test (`test_run_checks.py`, 22 tests) still pass.

none

## Failure class (fixes)

Failure-Class: none

fixes #10293

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10491?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -24,6 +24,11 @@ checks:
     triggers: ["Makefile", "backend/scripts/sync-python-deps.sh", "scripts/pre-push", "scripts/test-make-setup.sh", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "make setup must provision the minimal, rerunnable backend environment required by selected pre-push checks"
+  - id: pre-push-worktree-fallback
+    command: ["bash", "scripts/test-pre-push-worktree-fallback.sh"]
+    triggers: ["scripts/pre-push", "scripts/pre-push-singleflight", "scripts/pr-preflight", ".github/scripts/preflight_runner.py", ".github/scripts/pr_preflight.py", "scripts/test-pre-push-worktree-fallback.sh", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "#10293: the pre-push hook chain must resolve the repo root in linked worktrees where git rev-parse --show-toplevel exits 128, not abort into a --no-verify push"
   - id: check-manifest-contract
     command: ["python3", ".github/scripts/test_run_checks.py"]
     triggers: ["all"]
--- a/.github/scripts/pr_preflight.py
+++ b/.github/scripts/pr_preflight.py
@@ -23,6 +23,18 @@ class Check:
     reason: str
 
 
+def _resolve_repo_root() -> Path:
+    # The pre-push gate cd's to the repo root before invoking this script, and
+    # git runs hooks at the work-tree top, so fall back to the working directory
+    # when `git rev-parse --show-toplevel` cannot resolve a work tree: a linked
+    # worktree whose git context resolves to a git dir exits 128 here ("this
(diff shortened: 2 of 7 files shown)


undivisible:
Already merged in #10491 commit 7aa6c3633e; closing the superseded PR.

PR #10518, opened 2026-07-25
feat(ci): ratchet recurring failure classes onto a canonical guard artifact

Git-on-my-level:
## Problem

Root `AGENTS.md` says: *"If two or more recent fixes share the cause, add a reusable guard surface in the same PR."* Nothing enforced that rule, and it had silently failed.

Measured on `origin/main` at `5ef3056`:

- `FC-split-mutation-authority` was declared on **30 first-parent integration changes** in the 90 days to 2026-07-24 (50+ raw `fix:` commits).
- Its definition still listed `evidence_prs: [9365, 9597]` — neither from that window. None of the July declarations were ever recorded.
- Its `canonical_prevention` is prose with no artifact path, unlike the legacy `FC-1` (`backend/database/read_boundary.py`) or `FC-6` (`backend/tests/unit/fixtures/strict_firestore_transaction.py`).

A class could be declared indefinitely while accumulating no evidence and producing no reusable guard. Recurrence had become paperwork.

## Change

**1. Schema: `canonical_prevention_artifact`** (`scripts/failure-class`)

Optional array of repository-relative paths naming the reusable guard surface for a class. Every listed path must exist, so a rename cannot leave a class claiming a guard it no longer has. `canonical_prevention` says what should stop recurrence; this says where that guard actually lives.

Populated only for the three classes that demonstrably have one today:

| Class | Artifact |
| --- | --- |
| `FC-deploy-concurrency` | `.github/scripts/check-deployment-concurrency.py` |
| `FC-workflow-control-source-identity` | `backend/tests/unit/test_verify_backend_release_vector.py` |
| `FC-single-machine-release-gate` | `.github/scripts/desktop_codemagic_qualification.py`, `.github/scripts/test_desktop_codemagic_qualification.py` |

Deliberately **not** invented for the classes that lack one — that absence is the signal the ratchet exists to surface.

**2. The ratchet** (`.github/scripts/check_failure_class_guard_ratchet.py`)

- Counts **first-parent integration changes** whose message declares `Failure-Class: FC-<slug>` over a **90-day window**, plus the declaration in the PR body under review, so the PR that crosses the threshold fails rather than the next unrelated PR.
- Raw commits are not counted: those 30 integration changes carry 50+ raw commits, so a raw threshold would be noise. One integration change counts once even when its merge message repeats the declaration.
- **Threshold 3.** The prose rule is "two or more"; 3 gives one grace so the check flags sustained recurrence, not a single repeat.
- **Degrades safely rather than lying.** Shallow clone, or first-parent history that does not reach the window start → loud `SKIP` and exit 0. It never passes silently and never fails spuriously. CI checks out with `fetch-depth: 0`, so the enforcing path is the normal one.
- An FC id in history with no definition is ignored: that is the failure-class CLI's problem, and it must not fail a PR that never touched it.

**3. Grandfather allowlist** (`.github/scripts/failure_class_guard_ratchet_allowlist.json`)

Three classes were already over threshold with no artifact, so the check is green on `main` today and ratchets from here:

| Class | Declarations in window |
| --- | --- |
| `FC-split-mutation-authority` | 30 |
| `FC-nonrecoverable-promotion` | 6 |
| `FC-unprovisioned-rollout-target` | 3 |

The allowlist **only shrinks**: once a class records an artifact, the check fails until its entry is removed. Adding an entry is an explicit, reviewable admission that a class recurs without a guard.

**4. Manifest registration** — `failure-class-guard-artifact-ratchet` and `failure-class-guard-artifact-ratchet-tests`, both lanes, no workflow YAML touched.

**5. Opportunistic fixes (separate commits).** `scripts/test_failure_class.py` pinned the registry at five definitions and had been failing since the sixth class landed — 2 of 13 tests red on `main`, and it ran in no lane at all. Fixed to derive the seed set, and registered as `failure-class-cli-tests` so the registry's only schema authority cannot go stale unnoticed again.

Registering it then exposed a second, worse latent defect in the same file: it built disposable repositories with plain `git init` / `git config` / `git add .`, which is only hermetic when nothing exports git context. Under the pre-push hook, `GIT_DIR` and `core.hooksPath` point at the real checkout, so every temp-repo git command operated on the repository under test — it wrote `user.name = Failure Class Test` into the shared repo config and staged mass deletions from a polluted index. Both test files now scrub `GIT_*` from the subprocess environment and pin `core.hooksPath` off. This is exactly the class of bug that hides in a test nobody runs in a lane.

## Why this is not a shared primitive

The repo has ratchets (`check_isinstance_return_ratchet.py`, `check_product_file_line_count_ratchet.py`, `check_arch_guardrails.py`), but each ratchets a **static property of the tree**. This one ratchets a property of **history + registry state**, which no existing primitive computes. It reuses the registry, the existing `Failure-Class:` declaration grammar, and the existing allowlist/baseline-JSON convention rather than inventing new ones.

## What I deliberately left out

**The evidence-staleness check** (flagging a class whose recent declaring PRs are absent from `evidence_prs`) is **not shipped.** It cannot be attributed reliably: of the 30 declaring first-parent changes, several are direct pushes to `main` or merge commits whose subject carries no `(#NNNN)`, so PR-number extraction is partial. A check that misses a third of its inputs would either false-positive constantly or be tuned into uselessness. Shipping it would have been a flaky check, which is worse than no check.

Also out of scope: no refactor of `scripts/failure-class` beyond what the new field needs; no artifacts invented for classes that lack one; no lifecycle transitions on any class (those are separate registry-only PRs by the rules in root `AGENTS.md`).

## Verification

Every command below was run in this worktree at `d9b934e`.

```
$ python3 .github/scripts/check_failure_class_guard_ratchet.py; echo exit=$?
Failure-class guard ratchet: window=90d since=2026-04-26 threshold=3 classes_declared=13
  GRANDFATHERED FC-nonrecoverable-promotion: 6 declaration(s) — ...
  GRANDFATHERED FC-split-mutation-authority: 30 declaration(s) — ...
  GRANDFATHERED FC-unprovisioned-rollout-target: 3 declaration(s) — ...
OK: every class at or above the threshold names a guard artifact or is explicitly grandfathered.
exit=0
```

Real failing path, driven through a PR body declaring `FC-per-hop-timeout` (2 in history, no artifact, not allowlisted):

```
$ printf 'Failure-Class: FC-per-hop-timeout\n' > /tmp/body.md
$ python3 .github/scripts/check_failure_class_guard_ratchet.py --pr-body-file /tmp/body.md; echo exit=$?
FAIL: a recurring failure class has no reusable guard surface.
- FC-per-hop-timeout: 3 declarations in the window (threshold 3) with no
  'canonical_prevention_artifact'. Add the reusable guard surface and record its path in
  .github/failure-classes/FC-per-hop-timeout.json, or grandfather it explicitly in
  .github/scripts/failure_class_guard_ratchet_allowlist.json.
exit=1
```

Test suites:

```
$ python3 .github/scripts/test_check_failure_class_guard_ratchet.py   -> Ran 12 tests, OK
$ python3 scripts/test_failure_class.py                               -> Ran 16 tests, OK
$ python3 .github/scripts/test_run_checks.py                          -> Ran 22 tests, OK
$ python3 .github/scripts/check_agents_md_lean.py                     -> ok: 7 AGENTS.md files within budget
$ python3 .github/scripts/check_agent_doc_references.py               -> agent doc references OK
```

`make preflight` (the full deterministic manifest, local lane):

```
$ make preflight
...
<== PASS failure-class-protocol (0.10s)
<== PASS failure-class-guard-artifact-ratchet (0.22s)
<== PASS failure-class-guard-artifact-ratchet-tests (2.82s)
<== PASS failure-class-cli-tests (3.31s)
PR preflight passed: 46 checks in 39.25s.
```

The 12 ratchet tests each build a real disposable git repository with real first-parent history and run the checker end to end — over-threshold-without-artifact fails, over-threshold-with-artifact passes, under-threshold passes, allowlisted passes, allowlisted-with-artifact fails, out-of-window declarations do not count, a repeated declaration in one merge message counts once, PR-body declarations count, unknown ids are ignored, short history skips, a real `--depth 1` clone skips, and the shipped registry is green. No source-string assertions.

Runtime measured at 0.26s wall on this checkout, so `triggers: ["all"]` is affordable.

`git push` also ran the bounded pre-push gate end to end and passed, which is what proved the git-isolation fix (it failed 16/16 and 12/12 before it).

**Not verified:** CI behavior on a shallow checkout was exercised only via a local `git clone --depth 1`, not on a GitHub runner. Every workflow that runs the manifest uses `fetch-depth: 0`, so the SKIP path is expected to be dead in practice.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10518?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -113,6 +113,16 @@ checks:
     lanes: ["local", "ci"]
     requires_pr_body: true
     reason: "fix commits declare a validated semantic failure class"
+  - id: failure-class-guard-artifact-ratchet
+    command: ["python3", ".github/scripts/check_failure_class_guard_ratchet.py", "--pr-body-file", "{pr_body_file}"]
+    triggers: ["all"]
+    lanes: ["local", "ci"]
+    reason: "FC-split-mutation-authority was declared on 30 first-parent integration changes in the 90 days to Jul 24 2026 while its definition still listed evidence_prs [9365, 9597] and named no guard artifact; a recurring class must produce a reusable guard surface, not more paperwork"
+  - id: failure-class-guard-artifact-ratchet-tests
+    command: ["python3", ".github/scripts/test_check_failure_class_guard_ratchet.py"]
+    triggers: [".github/scripts/check_failure_class_guard_ratchet.py", ".github/scripts/test_check_failure_class_guard_ratchet.py", ".github/scripts/failure_class_guard_ratchet_allowlist.json", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the guard-artifact ratchet's counting window, threshold, allowlist shrink rule, and shallow-history skip must stay behaviorally covered"
   - id: failure-class-cli-tests
     command: ["python3", "scripts/test_failure_class.py"]
     triggers: ["scripts/failure-class", "scripts/test_failure_class.py", ".github/failure-classes/**", ".github/checks-manifest.yaml"]
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -113,6 +113,11 @@ checks:
     lanes: ["local", "ci"]
     requires_pr_body: true
     reason: "fix commits declare a validated semantic failure class"
+  - id: failure-class-cli-tests
+    command: ["python3", "scripts/test_failure_class.py"]
+    triggers: ["scripts/failure-class", "scripts/test_failure_class.py", ".github/failure-classes/**", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the failure-class CLI is the registry's only schema authority; its hermetic tests had no blocking lane and had silently gone stale"
   - id: desktop-changelog-data
     command: ["python3", ".github/scripts/desktop-changelog.py", "validate"]
     triggers: ["all"]
--- a/.github/failure-classes/FC-deploy-concurrency.json
+++ b/.github/failure-classes/FC-deploy-concurrency.json
@@ -3,6 +3,7 @@
   "id": "FC-deploy-concurrency",
   "violated_contract": "An exact eligible main SHA must have one authoritative automatic development deployment lifecycle for each shared persistent mutation domain; sibling automatic workflows must not compete for the same GitHub Actions concurrency group.",
   "canonical_prevention": "Fold shared-backend automatic mutations into the Release Eligibility-admitted backend lifecycle and deterministically reject a second automatic writer for deploy-backend-stack-development.",
+  "canonical_prevention_artifact": [".github/scripts/check-deployment-concurrency.py"],
   "evidence_prs": [10329],
   "scope_hints": [
     ".github/workflows/gcp_backend_auto_dev.yml",
--- a/.github/failure-classes/FC-single-machine-release-gate.json
+++ b/.github/failure-classes/FC-single-machine-release-gate.json
@@ -3,6 +3,10 @@
   "id": "FC-single-machine-release-gate",
   "violated_contract": "A release-blocking automated gate must not depend on exactly one machine: every mandatory pipeline stage needs at least two independently healthy executors (an always-available primary lane plus at least one fallback), with the gate passing when any lane produces valid evidence and failing only on the candidate itself.",
   "canonical_prevention": "Desktop beta qualification runs a Codemagic primary lane plus a pool of interchangeable self-hosted runners behind an at-least-one-passes verdict; machine-specific prerequisites (global backend python, Docker-only Typesense) are eliminated so any admin Mac or ephemeral CI Mac can execute the lane.",
+  "canonical_prevention_artifact": [
+    ".github/scripts/desktop_codemagic_qualification.py",
+    ".github/scripts/test_desktop_codemagic_qualification.py"
+  ],
   "evidence_prs": [10354],
   "scope_hints": [
     ".github/workflows/desktop_qualify_beta.yml",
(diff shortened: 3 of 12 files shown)


Git-on-my-level:
@cursor address review comments

Git-on-my-level:
@cursor address failing CI

Git-on-my-level:
Resolved the two Codex review threads on `check_failure_class_guard_ratchet.py`. Both concerns were already addressed by the prior `fix(ci): address Codex review on failure-class guard ratchet` commit (`7cb1c98`), which is in the pushed PR head:

- **P1 — Permit the required artifact update in the instance-fix PR**: `scripts/failure-class` now exposes `instance_fix_registry_changes_allowed`, which lets a `fix:` PR add/update `canonical_prevention_artifact` (and only that field) on the declared class. This makes the ratchet's "record a guard artifact" remedy and the CLI's `instance_fix_mutates_registry` rule mutually satisfiable. Covered by `scripts/test_failure_class.py`.
- **P1 — Reject new recurrences of grandfathered classes**: `evaluate()` now fails when `count > entry.declarations_at_baseline`, so a grandfathered class can no longer recur indefinitely without producing a guard surface. Covered by `test_grandfathered_class_fails_on_new_recurrence`.

Verification: `test_check_failure_class_guard_ratchet.py` (13 tests) and `scripts/test_failure_class.py` (17 tests) both pass. All CI checks pass.

PR #10522, opened 2026-07-25
fix(desktop-windows): restart Rewind capture when its stream track dies

### LINKED ISSUE #10504: Bug: Omi Desktop for Windows timeline shows dark blank screenshots after running for a few hours
## Summary
On Omi Desktop for Windows, after the app has been running for a few hours, the timeline starts showing dark blank screenshots with no visible content.

Image

## Problem
The app appears to keep running, but the timeline no longer captures usable screen content.

Instead of normal screenshots, the timeline shows dark or blank images with no meaningful content recorded.

## Expected behavior
After running for hours, Omi Desktop should continue capturing normal screenshots with visible screen content in the timeline.

## Actual behavior
After the app has been running for a few hours, timeline screenshots become dark and blank, and no actual content appears to be recorded.

## Impact
This breaks long-running desktop capture sessions.

Even though the app may still appear active, the timeline becomes unusable because the screenshots no longer contain readable or meaningful content.

## Suggested investigation
- Check whether long-running Windows desktop sessions cause screenshot capture to fail silently
- Check whether the app loses access to the display surface after running for several hours
- Check whether the timeline is receiving blank frames while the app still thinks capture is active
- Check logs for screenshot pipeline stalls, blank frame generation, GPU/display capture issues, or permission/state desync over time

## Environment
- Omi Desktop
- Windows
- Observed on Friday, July 24, 2026

aryanorastar:
## What

Closes #10504.

Rewind on Windows opens **one persistent** getUserMedia desktop stream and samples it on a timer. It handled OS **power/lock** (the capture directive tears the stream down and rebuilds it on wake) — but nothing else.

If the video track ended mid-session for any **other** reason — a display reconfiguration, a GPU/driver reset, the captured source going away — the hidden `<video>` went dead while the sampling loop kept drawing it. Every frame after that was blank/dark, and it never recovered until the app restarted. Over hours, such an event is likely — which is exactly *"timeline shows dark blank screenshots after running for a few hours."*

## Fix

Watch each capture track's `ended` event and rebuild the stream (after a **3s backoff**, so a permanently-gone source — e.g. an unplugged display — can't hot-loop `getUserMedia`).

The event wiring is a tiny `onCaptureStreamDeath(stream, onDeath)` helper so the *"stream died → we noticed"* contract is unit-tested rather than buried in the effect. The host **detaches** it before any teardown it initiates, so a deliberate stop is never mistaken for a death.

## Verification

```
captureSupervisor.test.ts (3) — fires once on track end; at most once across
  multiple tracks; silent after unsubscribe.   Mutation-checked: removing the
  listener wiring fails 2/3.
rewind renderer suite → 13 passed
typecheck:web → 0 errors
```

**Not exercised at runtime:** there's no Windows Electron host in my environment, and the hours-long degradation can't be reproduced here. This closes the specific code gap (no recovery when the capture stream dies) that produces the reported symptom; a Windows soak test would confirm the field fix. The change is additive and self-detaching, so it can't regress the normal capture path.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10522?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/renderer/src/components/rewind/RewindCaptureHost.tsx
+++ b/desktop/windows/src/renderer/src/components/rewind/RewindCaptureHost.tsx
@@ -1,10 +1,14 @@
 import { useEffect, useRef, useState } from 'react'
 import type { RewindSettings, RewindCaptureDirective } from '../../../../shared/types'
+import { onCaptureStreamDeath } from './captureSupervisor'
 
 // Cap the longest sampled edge — plenty for a timeline + OCR, and keeps each
 // canvas grab + JPEG encode cheap.
 const MAX_EDGE = 1600
 const JPEG_QUALITY = 0.6
+// Wait before rebuilding a stream whose track ended, so a source that is
+// permanently gone (e.g. an unplugged display) can't hot-loop getUserMedia.
+const STREAM_RESTART_BACKOFF_MS = 3000
 
 /**
  * Background screen-capture host for Rewind. Mounted app-wide (while the window
@@ -19,6 +23,8 @@ export function RewindCaptureHost(): React.JSX.Element {
   const videoRef = useRef<HTMLVideoElement>(null)
   const streamRef = useRef<MediaStream | null>(null)
   const timerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+  const restartTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+  const detachDeathRef = useRef<(() => void) | null>(null)
   const canvasRef = useRef<HTMLCanvasElement | null>(null)
   const savingRef = useRef(false)
   const [settings, setSettings] = useState<RewindSettings | null>(null)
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Thanks for this — it is a well-scoped fix for a real gap (#10504). A few notes on what landed well, plus one observation:

- Extracting the "stream died → we noticed" wiring into `onCaptureStreamDeath` with isolated unit tests is the right call. The three cases (fires once on end, at most once across multiple tracks, silent after unsubscribe) meaningfully pin the contract.
- Detaching the listener before a self-initiated `stop()` so teardown is not mistaken for a death, plus the stale-stream guard (`streamRef.current !== stream`), are good defensive details.
- The 3s backoff before rebuild correctly prevents hot-looping `getUserMedia` when a source is permanently gone.

One non-blocking observation: there is no upper bound on restart attempts. If a display source is permanently gone, `getUserMedia` may keep producing a track that `ended`s again immediately, yielding a slow ~3s retry loop indefinitely. The fixed backoff handles the hot-loop concern, but a max-retry cap or exponential backoff would make the permanently-gone case quieter. Does not block this — resilient recovery for transient deaths is already a clear improvement over silent dark frames.

Since this touches the screen-capture path (privacy-sensitive surface), flagging for a human maintainer look before merge. The change is additive and self-detaching, so the normal capture path is not at regression risk.

---

aryanorastar:
Superseded by #10549, which merged the same fix for #10504 — main's `RewindCaptureHost.tsx` now wires `addEventListener('ended')` on the video tracks, guards sampling behind `isLive()`, and reopens via `setTimeout(start, RESTART_DELAY_MS)`. Its tests cover the same three behaviors mine did (reopen on track death, no duplicate sampling loop when the track dies mid-save, no reopen after unmount), and it asserts them through the real component rather than my separate helper, so there's nothing left here worth keeping. Closing.

PR #10531, opened 2026-07-25
fix(desktop-windows): stop the Linux package build from attempting to publish

aryanorastar:
## What

`Desktop Windows CI` is **red on main**. Its "Build Linux packages" smoke step runs `pnpm run build:linux`, which invoked `electron-builder --linux` **without `--publish never`**. electron-builder then took its default publish path and failed the build:

```
Error: GitHub Personal Access Token is not set, neither programmatically, nor using env "GH_TOKEN"
  at PublishManager.artifactCreatedWithoutExplicitPublishConfig (app-builder-lib/src/publish/PublishManager.ts:213)
```

`build:win` already passed `--publish never`; `build:linux` (and the README-documented `build:mac`) didn't — so a smoke/CI build tried to release.

## Fix

Add `--publish never` to `build:linux` and `build:mac`, matching `build:win`. A smoke/CI/local packaging build must never attempt a publish.

## Guard

`scripts/build-publish-guard.test.mjs` pins that every electron-builder `--win/--mac/--linux` script opts out of publishing, so a new packaging target can't reintroduce this. Mutation-checked: stripping the flag fails it.

## Verification

```
build-publish-guard.test.mjs → 4 passed
package.json → valid JSON
```

The full Linux `electron-builder` run needs a Linux host, so the one-flag fix is confirmed by mirroring the already-working `build:win` and by the guard — not a local package build.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10531?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for the focused fix. I reviewed the packaging-script change statically and checked the guarded scripts locally: `build:win`, `build:mac`, and `build:linux` now all include `--publish never`, and the added Vitest guard should catch this class of accidental publish regression going forward.

I’m leaving this as a positive signal rather than a formal approval because it touches desktop packaging/release behavior (`electron-builder` publish controls), which should get a quick maintainer look before merge. I did not find a code blocker in the patch itself.

---

aryanorastar:
Superseded by #10559 (`f449decb9f`), which landed the same fix while this was open — main's `package.json` now carries `--publish never` on `build:win`, `build:mac` and `build:linux`, and it also added the regression guard at `desktop/windows/scripts/build-scripts-publish.test.mjs`.

I compared the two guards rather than assume mine added something: they cover the same class, and main's is marginally tighter — it matches targets with `/\s--(win|mac|linux)\b/` where mine used a `--win` substring (which would also match a hypothetical `--windows`), and it pins the exact target list instead of my `arrayContaining`. So there's nothing left here worth carrying over. Closing.

Thanks @Git-on-my-level for the review pass on this one.

Issue #10539, opened 2026-07-25
main is red: failure-class-guard-artifact-ratchet fails on a clean checkout, blocking every PR


Git-on-my-level:
`make preflight` and `scripts/pr-preflight` fail on a clean checkout of `origin/main` (`a2ca2c3c7e`) with no local changes. `failure-class-guard-artifact-ratchet` has `triggers: ["all"]`, so this blocks every PR, not just ones touching failure classes.

```
$ git worktree add --detach /tmp/verify origin/main
$ cd /tmp/verify && printf '## Summary\n' > /tmp/plain.md
$ python3 .github/scripts/check_failure_class_guard_ratchet.py --pr-body-file /tmp/plain.md
FAIL: a recurring failure class has no reusable guard surface.
- FC-split-mutation-authority: 32 declarations in the window (baseline 31) with no
  'canonical_prevention_artifact'. A grandfathered class must not recur without recording
  its guard surface in .github/failure-classes/FC-split-mutation-authority.json, then
  removing its entry from .github/scripts/failure_class_guard_ratchet_allowlist.json.
Failure-class guard ratchet: window=90d since=2026-04-26 threshold=3 classes_declared=13
```

The PR body is empty of any `Failure-Class:` declaration, so the 32nd declaration comes from merged history alone — a `FC-split-mutation-authority` PR landed after the allowlist baseline of 31 was frozen on 2026-07-24.

It cascades. `.github/scripts/test_pr_preflight.py::SelectionTests::test_pr_body_file_env_is_honored` asserts the manifest run returns 0, so it fails too, taking `pr-preflight-contract-tests` down with it. Same root cause, two red checks:

```
$ python3 .github/scripts/test_pr_preflight.py
FAIL: test_pr_body_file_env_is_honored (__main__.SelectionTests.test_pr_body_file_env_is_honored)
AssertionError: 1 != 0 : Check manifest: lane=ci platform=macos ... files=0
Ran 21 tests in 10.940s
FAILED (failures=1)
```

Both reproduce on a clean detached `origin/main` worktree, so this is not environment-specific to one machine.

[...]

Issue #10536, opened 2026-07-25
Port FC-6 (lenient Firestore transaction fake) into .github/failure-classes/ before its record is only in git history


Git-on-my-level:
`docs/product/failure-classes.md` was a second failure-class registry that nothing read. It used a numeric ID scheme (`FC-1`, `FC-6`) that conflicts with the slug IDs in `.github/failure-classes/*.json`, and it had zero inbound references anywhere in the repo — `scripts/failure-class` only reads `.github/failure-classes/`, so any PR citing `FC-1` or `FC-6` was unvalidatable. Two IDs for one class defeats the point of the registry: `FC-1` was a duplicate of `.github/failure-classes/FC-malformed-doc-read.json` (same contract, same fix primitive `backend/database/read_boundary.py`).

That file is being deleted in the PR that links here. `FC-1` is already covered. **`FC-6` is not** — it has no counterpart among the 26 JSON definitions, so deleting the file removes its only record outside git history. Porting it is a registry lifecycle change, which `AGENTS.md` requires be its own PR, hence this issue rather than folding it into that one.

The content to port, verbatim from the deleted table:

| Field | Value |
| --- | --- |
| Violated contract | A Firestore transaction fake must reject reads after the first write; a lenient fake can certify production code Firestore rejects. |
| Canonical fix primitive | `backend/tests/unit/fixtures/strict_firestore_transaction.py` (still present) |
| Status | Closed |
| Closed by | Shared strict fixture and strict-by-default convention. |
| Not covered | Boundaries without incident evidence; fake styles that cannot be classified mechanically; queries, deletes, retry, and contention semantics. |

Open questions for whoever picks this up: what the slug should be (`FC-lenient-transaction-fake`?), which `evidence_prs` back it, and whether a class that was already Closed under the old scheme should enter the JSON registry as `status: open` with `dormant_since` set, or be considered retired and not ported at all. The registry currently has 26 entries and **all** of them are `status: open` with no `dormant_since` on any — the closing half of the documented lifecycle has never fired, so this is a reasonable place to exercise it for the first time.

Git-on-my-level:
Closing — the premise was stale and is wrong on current `main`.

This was filed against a branch point 33 commits behind. Since then `docs/product/failure-classes.md` has been repurposed: it is no longer a competing registry but the **schema documentation** for `.github/failure-classes/*.json`, cited from `AGENTS.md:48` and from `.github/scripts/failure_class_guard_ratchet_allowlist.json`.

[...]

PR #10523, opened 2026-07-25
SCA-103: add macOS failure diagnostic context

Git-on-my-level:
## Summary

- add a bounded, shared Sentry diagnostic-context surface for macOS failures
- attach volume, SQLite/POSIX, file-size, termination, and coarse path-class context to Rewind storage failures
- attach typed startup stage, preconditions, elapsed/timeout, and exit-code context to failed agent-runtime starts
- cover context values and the no-path allow-list in Swift tests
- register the existing single-owner authority ratchet that now guards recurring failure-class declarations

Closes SCA-103

## Product invariants affected

- INV-AUTH-1
- INV-CHAT-1

## Failure class

Failure-Class: none

## Verification

- `git diff --cached --check` — passed before commit.
- `scripts/pr-preflight --suggest` — identifies `INV-AUTH-1`, `INV-CHAT-1`, and `Failure-Class: none`; the PR body declares all three.
- `python3 .github/scripts/check_failure_class_guard_ratchet.py --pr-body-file /tmp/pr-10523-body.md` — passed after recording the existing single-owner authority guard.
- `python3 .github/scripts/test_check_failure_class_guard_ratchet.py` — 13 passed.
- `scripts/failure-class validate --base origin/main --head HEAD --pr-body-file /tmp/pr-10523-body.md` — passed.
- `make preflight` — started and completed `check-manifest-contract` successfully; the command's remaining output was not available from the local runner after its process detached, so it is not claimed as passed.
- `xcrun swift test -c debug --package-path desktop/macos/Desktop --filter 'DesktopErrorTelemetryTests|SentryBeforeSendScrubTests'` — not completed: initial SwiftPM dependency hydration left the package `.build` lock occupied; subsequent runs waited on that lock. No test success is claimed.

No app bundle was launched: this change affects failure-only diagnostics and does not alter user-facing behavior.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10540, opened 2026-07-25
docs+ci: correct stale backend test claims in the Cursor agent guide, and guard doc claims against the code they describe

Git-on-my-level:
## Summary

Two defects in the guidance surfaces developers and agents read, each fixed with the guard that makes the fix durable.

1. A **stale agent doc** caused an agent to propose a change that would have regressed the backend test suite.
2. A **pre-push gate hid its own break-glass hatch**, so a docs-only push failed telling the author to install a multi-gigabyte SDK.

Same failure mode in both: the words a developer reads were wrong or missing, and nothing checked them.

## Part 1 — the stale agent doc

The incident is the argument for the PR, so it is worth stating precisely.

`.cursor/cloud-agent-environment.md` — listed in `AGENTS.md:26` as a doc agents should read — asserted three things that had all become false:

| Claim in the doc | Reality |
|---|---|
| `test.sh` "halts at the first failing file" | `backend/test.sh` is a concurrent per-file worker pool that collects every file's status and reports all failures with a re-run command |
| `tests/unit/test_speaker_sample.py` is broken by a rename | The compatibility alias at `backend/utils/speaker_sample.py:14` fixed it; the file passes 16/16 |
| `pytest-asyncio` is not in the lock and no `asyncio_mode` is set | Pinned at `backend/requirements.txt:170`; `backend/pyproject.toml:7` sets `asyncio_mode = "auto"` |

Acting on the first claim, an agent proposed replacing the per-file process isolation with `pytest -n auto --dist loadfile`. That would have been a strict regression: `loadfile` pins a file to a worker but reuses the interpreter, and `backend/docs/test_isolation.md:9` documents ~100 test files that mutate `sys.modules` at module scope — exactly what per-process isolation contains.

Every pointer in that doc still resolved. `check_agent_doc_references.py` would have passed it, if the file had been in scope at all — it was not.

## What changed

1. **`docs:` correct the stale claims.** Rewrites the "Unit tests" section to describe the real execution model. Also drops two frozen counts: "~129/172 unit-test files" is now 729 per `select_backend_unit_tests.py --all`, and a hardcoded E2E pass count now points at the workflow that produces the live one. A snapshot number in an agent doc is the claim class that rots fastest. Keeps the still-true warning against running `tests/unit` in one process and points it at `backend/docs/test_isolation.md`.

2. **`ci:` cover the file in `check_agent_doc_references.py`.** `EXTRA_DOCS` omitted it, which is why two dead pointers were live on main: `tests/unit/test_speaker_sample.py` and `database/vector_db.py`, both written backend-relative so they resolved from neither the doc's directory nor the repo root. Adds the file specifically, not the whole `.cursor/` directory.

3. **`ci:` add `check_agent_doc_claims.py`.** Binds a literal in a doc to a literal in the code that makes it true, and fails if either side moves without the other. Two seeded claims, both verified on this branch. `check_trigger_coverage` additionally asserts the manifest entry actually invokes this script and that every path the table names — plus the checker and the manifest themselves — is matched by its triggers.

## Why a check and not a paragraph of prose

`docs/agents/doc-maintenance.md` already states the rule: *"When a defect ships because guidance was misread or missing, tighten the guidance in the fix PR. Make the rule mechanical enough that the same misreading cannot recur, or add the check that catches it."*

Not a shared primitive already: `check_agent_doc_references.py` answers "does this path exist", a different question with a different failure mode, and its resolver has no notion of a doc/code pair.

Deliberately not semantic, and the docstring says so rather than implying more. Anchors are substring matches, so a surgical edit can keep the token while removing the behavior — gut the isolation branch but leave the default assignment, or drop one of two append sites. It catches the ordinary case, a refactor that removes the construct along with its token, and misses an adversarial one. A tripwire that forces a re-read, not proof of correctness. Verifying meaning needs judgment, and a check that needs judgment grows an allowlist.

No claim binds a doc to another doc. Prose satisfying an anchor proves only that someone still writes the words, which is the failure being guarded against.

## Part 2 — a gate that hid its own hatch

Found by pushing this branch. `make setup` provisions the backend venv only — `AGENTS.md:73` is explicit that "It does not install app or desktop runtime environments." Meanwhile `.github/checks-manifest.yaml` is in `ROUTING_INPUTS` (`scripts/pre_push_ci_prediction.py:71`), so editing it makes the selector conservatively wake the Flutter codegen lane. This PR edits that manifest, so pushing it hit:

```
FAIL: Flutter generated-output checks need resolved app dependencies.
      Run: cd app && flutter pub get
```

`PRE_PUSH_SKIP_FLUTTER_GENERATED=1` already clears this. The message never said so. `check_workflows_if_needed`'s actionlint branch had the identical omission. Every other gate in the file already printed its hatch — these two were simply missed, which is drift a ratchet catches and review does not.

**The conservative wake is left alone.** A routing change can influence any lane, so waking them all is correct, and CI remains the full authority. What changes is that a developer who trips the gate can tell an unprovisioned machine from a real defect. `AGENTS.md`: *"Every gated surface has a break-glass hatch. A broken gate is never a reason to be stuck."* A hatch nobody can find is not a hatch.

`check_pre_push_hatch_disclosure.py` stops the next gate regressing it. Its docstring labels it a static checker rather than implying coverage it lacks: it reads shell source and never executes pre-push. Only a mention on a `>&2` line counts — the `if [ "${PRE_PUSH_SKIP_X:-}" ]` guard that honors the variable also names it, and an earlier draft passed on **any** omission because that guard satisfied the lookback window. Caught by running the negative test, not by reading.

## Two things this branch cut, and why

Both deserve stating, because they are the same failure mode as the bug being fixed.

**A `fix:` for `scripts/test_failure_class.py`.** It was failing 2 of 16 on main — `SEED_IDS` hardcoded 5 class ids against a registry of 26 — and was registered in no manifest lane, which is why nobody noticed. Between branching and rebasing, `main` fixed both independently (`failure-class-cli-tests`, and `SEED_IDS` derived from the registry glob). Dropped as redundant.

**Deleting `docs/product/failure-classes.md` as a duplicate registry.** At the branch point it was a competing registry: a conflicting numeric id scheme, zero inbound references, unreadable by `scripts/failure-class`. On current `main` it is the opposite — the schema documentation for `.github/failure-classes/*.json`, cited from `AGENTS.md:48` and the guard-artifact ratchet allowlist, with FC-1/FC-6 retained under an explicit "Legacy narrative entries — kept for their closure history" heading. Deleting it would have deleted the registry's own documentation. Dropped, and #10536 closed as filed on a stale premise.

Worth noting what caught it: `check_agent_doc_references.py`, extended by commit 2 of this PR, failed on `AGENTS.md: reference docs/product/failure-classes.md does not exist` and stopped the bad deletion. A 33-commit-old branch point produced exactly the class of error this PR is about.

## Review

Reviewed by an independent agent before opening. Findings applied:

- **The doc introduced a new false claim.** "`BACKEND_PYTEST_FILE_ISOLATION=0` falls back to a single pytest process" is wrong — with isolation off, xdist is enabled by default (`-n auto --dist=loadfile`), so it is one invocation but still multiple worker processes. Corrected, and `BACKEND_PYTEST_XDIST=0` named for a genuinely serial run. I introduced this while trying to make an earlier sentence *more* accurate, which is the failure mode this PR is about.
- **"one worker per CPU" and "every file's exit status" were overstated.** `BACKEND_PYTEST_WORKERS` overrides the first; a worker that dies before writing its status file has its real status discarded. Both softened.
- **"roughly a hundred files mutate `sys.modules`"** is sourced from `backend/docs/test_isolation.md`, but it is a volatile count in an agent doc — the exact thing this PR removes twice elsewhere. Dropped; the doc now defers to the authority for the number.
- **A third claim bound one doc to another doc** (`sys.modules` appearing in `test_isolation.md`). That anchors prose to prose, not to behavior. Dropped.
- **`check_trigger_coverage` could pass while the check was unwired** — it matched on the entry's id and never checked its `command`. An entry invoking any other script passed. Fixed and covered by a negative test.
- **`sys.path.insert` for the sibling `run_checks` import** replaced with a private-named `importlib` load. The first attempt at this crashed (`@dataclass` resolves `cls.__module__` through `sys.modules`); caught only by running the negative test, and fixed by registering before `exec_module`.
- **Unreadable/undecodable files raised tracebacks** instead of reporting. Fixed, with self-test coverage.

Not applied: `self_test()` inside `main()` was flagged, but that matches the established pattern in `check_agent_doc_references.py:171`. Consistency with the neighbouring checker wins.

## Verification

```
scripts/pr-preflight --pr-body-file <this>            -> every check passes except the
                                                        two that also fail on clean main
                                                        (see below, #10539)
python3 .github/scripts/check_agent_doc_claims.py     -> agent doc claims OK
python3 .github/scripts/check_agent_doc_references.py -> agent doc references OK
python3 .github/scripts/test_run_checks.py            -> 26/26 pass
python3 .github/scripts/check_agents_md_lean.py       -> 7 AGENTS.md files within budget
python3 scripts/test_failure_class.py                 -> 17/17 pass
python3 -m black --check ... check_agent_doc_claims.py -> unchanged
```

Negative tests, because a guard that has never failed is not known to work:

- **Code drift.** Removed `BACKEND_PYTEST_FILE_ISOLATION` from a scratch copy of `test.sh` → fails naming the code side.
- **Doc drift.** Reworded the doc anchor in a scratch copy → fails naming the doc side.
- **Missing file, either side** → fails with a distinct message.
- **Trigger coverage.** Removed the `backend/**` triggers from the manifest entry → fails naming `backend/test.sh`, the checker, and the manifest. Restored.
- **Unwired command.** Pointed the entry's `command` at a different script while keeping its id → fails with `the 'agent-doc-claims' manifest entry does not run .github/scripts/check_agent_doc_claims.py`. Restored.
- **Default flip.** Changed `BACKEND_PYTEST_FILE_ISOLATION:-1` to `:-0` in a scratch copy → fails. The earlier bare-name anchor would have passed this.

Real repo files were never edited to produce these; the self-test uses `tempfile.TemporaryDirectory()`.

Claims in the rewritten doc were verified by running, not by reading:

- `pytest tests/unit/test_speaker_sample.py` → **16 passed**
- `pytest tests/unit/test_listen_session_bootstrap.py` → **4 passed** (bare `async def test_*`, no decorators, confirming `asyncio_mode = "auto"` is live)
- `select_backend_unit_tests.py --all` → **729** files

### What I did not verify

A full `backend/test.sh` run (729 files) did not complete. Four files exited non-zero during a partial run, so I diagnosed them in a **clean detached `origin/main` checkout with no working-tree changes**, run serially with no contention:

| File | Tests | Cause |
|---|---|---|
| `test_activate_task_intelligence_dogfood_user.py` | 5/5 pass | `conftest.py` CPU-time budget |
| `test_baseline_memories.py` | 12/12 pass | same |
| `test_byok_llm_logging.py` | 10/10 pass | same |
| `test_backend_runtime_env_validator.py` | 74/74 pass | same |

Every assertion passes in all four. The non-zero exit comes from `backend/tests/conftest.py`'s per-test CPU budget: `BACKEND_FAST_UNIT_FAIL_SECONDS=0.12` locally against a 1.0s ceiling in CI. It reproduces identically on clean `origin/main`, so it is a machine-speed-vs-local-threshold gap, not caused by this diff — which touches no backend product or test code. Not fixed here; out of scope.

The E2E harness's previously-hardcoded "72 passed, 6 skipped" was **not** re-verified. Rather than carry an unverified number forward, the doc now points at `backend-hermetic-e2e.yml`, whose latest run is the live baseline.

### Previously-blocking failure, now resolved

An earlier revision of this branch reported `failure-class-guard-artifact-ratchet` failing on a clean `origin/main` (#10539), which forced a `--no-verify` push. That is fixed upstream: `FC-split-mutation-authority` now records `canonical_prevention_artifact` (`desktop/macos/agent/tests/convergence-authority-ratchet.test.ts`) and has been removed from the grandfather allowlist. After rebasing onto `cb918cf0ba`, **`make preflight` passes all 55 selected checks**, and this branch was pushed through the real pre-push gate with no bypass. #10539 closed.

## Failure class

Failure-Class: none

An ordinary instance fix that lands its own reusable guard in the same PR. No registered class covers "a gated surface's escape hatch is not disclosed at the point of failure" — the nearest, `FC-push-gate-internal-path-scope`, is about a gate's *trigger* being too broad, whereas here the trigger is deliberate and correct and only the diagnostic was missing. One incident is not enough to justify opening a registry class, and `check_pre_push_hatch_disclosure.py` is the durable surface a class would have demanded anyway.

none

Confirmed with `scripts/pr-preflight --suggest`. The diff touches no invariant path globs.


Git-on-my-level:
## Monitor summary — agent-doc trigger + contract coverage

Addressed both unresolved Codex review threads on `ed479eb5`.

**Trigger the reference check for Cursor guide edits** (`.github/scripts/check_agent_doc_references.py:34`)
Added `.cursor/cloud-agent-environment.md` to the `agent-doc-references` manifest triggers, so a guide-only edit now fires the dead-pointer check instead of relying on an unrelated trigger to select it. Covered by a new selection test (`test_cursor_guide_selects_both_agent_doc_checks`) asserting both `agent-doc-references` and `agent-doc-claims` run on a guide-only change in both lanes.

**Add a contract for the claims check** (`.github/scripts/check_agent_doc_claims.py:191`)
Added `test_agent_doc_claims_check_is_registered_and_invokes_its_script`, which asserts the `agent-doc-claims` manifest entry exists, invokes `check_agent_doc_claims.py`, and keeps its `ci` lane. This catches silent removal/redirect of the entry in the existing manifest contract suite rather than a separately-executed test, matching how the other agent-doc checks are guarded.

Verification: `python3 .github/scripts/test_run_checks.py` — 28 tests, OK (repo root). `check_agent_doc_references.py` OK, `check_agent_doc_claims.py` OK. CI: Hygiene passed.

The Desktop Swift Static/Build checks fail on a pre-existing `FocusAssistant.swift:694` build error (`cannot find 'DesktopLocalProfile' in scope`) present on `origin/main`; this PR touches no Swift.

Git-on-my-level:
## Monitor summary — upstream FocusAssistant build fix opened

The two failing Desktop Swift checks on this PR are caused by a pre-existing `main` break (PR #10523 added `DesktopLocalProfile.applicationSupportURL()` to `FocusAssistant.swift` without the required `import OmiSupport`). This PR touches no Swift.

Opened #10556 to fix the upstream break: adds the missing `import OmiSupport` to `FocusAssistant.swift`, matching the convention every other file using `DesktopLocalProfile` already follows (12 call sites). Once merged, it unblocks this PR's Desktop Swift checks.

Verification: `swiftc -parse` passes; the DesktopLocalProfile compile error is absent in all #10556 CI attempts (the first run reached the test phase before self-hosted macOS runner cancellation). A separate transient GitHub Actions "workflow file issue" is currently affecting all PR workflows and will clear on its own.

Git-on-my-level:
## Monitor summary — rebased onto main with SCA-103 build fix; Desktop Swift CI green

The two failing Desktop Swift checks on this PR were caused by a pre-existing `main` break (`FocusAssistant.swift` referenced `DesktopLocalProfile` without `import OmiSupport`, introduced by #10523). That break is now fixed on `origin/main` via #10576 (`424c574ffc`).

This PR's branch was rebased onto that fixed base (no content changes — patch-IDs verified identical) and pushed. CI re-ran against the corrected base:

- **Desktop Swift Static & Test Contracts**: pass (was FAILURE)
- **Desktop Swift Build & Tests**: pass (was FAILURE)

All other checks remain green. The PR is MERGEABLE; the remaining BLOCKED state is REVIEW_REQUIRED (maintainer approval), not a code or CI issue. This PR touches no Swift — its scope is the agent-doc corrections and their CI guards only.

PR #10542, opened 2026-07-25
fix(desktop): keep web routing classifiers in parity

Git-on-my-level:
## Summary

- keep Pi's compatibility routing projection aligned with the authoritative Rust public-web policy
- reject substring and long-prompt false positives while restoring recent-news, release, explicit lookup, mixed-context, and public-research cases
- enforce both runtimes with one shared behavioral routing corpus

## Root cause

PR #10422 had to update both the Rust gateway policy and Pi's temporary compatibility classifier. The copies drifted immediately: Pi matched generic lookup words as substrings and had no short-turn bound, so words such as `gameplan` or `marketing` and long synthesis prompts could display synthetic web-search activity even when the gateway did not route a search. Meanwhile, the Rust policy omitted several freshness phrases that Pi recognized, so those prompts could reach the model without the required public-web route.

The Rust gateway remains the policy owner. The adapter is still required while desktop and backend versions roll independently, so one implementation cannot be shared across the TypeScript and Rust runtime boundary. The new shared corpus is the narrow guard surface: both production classifiers must pass the same positive and negative decisions.

## Verification

- `cd desktop/macos/agent && npm test` — 53 files, 600 tests passed
- `cd desktop/macos/Backend-Rust && cargo test` — 400 tests passed
- `OMI_PR_BODY_FILE=/private/tmp/pr-web-routing-parity.md make preflight` — 14 checks passed
- bounded pre-push gate — passed, including Rust compile and desktop tool surfaces
- `desktop/macos/scripts/agent-logic-harness.sh --cross-surface-smoke` — passed on the starting main revision
- named bundle `omi-web-routing-parity` — built, launched, and exercised against the development backend
  - `Review my launch gameplan today` completed without public-web activity
  - `Give me recent news about Anthropic` completed after live network tool calls
- after repository setup advanced the branch to `3a63002ea9`, the cross-surface Swift phase is blocked by compile errors introduced on that main revision in `FocusAssistant.swift`, `RewindDatabase.swift`, `RewindIndexer.swift`, and `ChatProvider.swift`; none are touched by this PR

none

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10542?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-public-web-routing-parity.json
+++ b/.github/failure-classes/FC-public-web-routing-parity.json
@@ -4,7 +4,7 @@
   "violated_contract": "Compatibility projections of public-web routing must make the same routing decision as the authoritative gateway policy.",
   "canonical_prevention": "Keep the Rust gateway authoritative and run every compatibility projection against one shared behavioral routing corpus.",
   "canonical_prevention_artifact": [
-    "desktop/macos/agent/contracts/v1/public-web-routing-contract.fixture.json"
+    "desktop/macos/Backend-Rust/fixtures/public-web-routing-contract.fixture.json"
   ],
   "evidence_prs": [10422],
   "scope_hints": [
--- a/.github/failure-classes/FC-public-web-routing-parity.json
+++ b/.github/failure-classes/FC-public-web-routing-parity.json
@@ -1,0 +1,15 @@
+{
+  "schema_version": 1,
+  "id": "FC-public-web-routing-parity",
+  "violated_contract": "Compatibility projections of public-web routing must make the same routing decision as the authoritative gateway policy.",
+  "canonical_prevention": "Keep the Rust gateway authoritative and run every compatibility projection against one shared behavioral routing corpus.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/agent/contracts/v1/public-web-routing-contract.fixture.json"
+  ],
+  "evidence_prs": [10422],
+  "scope_hints": [
+    "desktop/macos/Backend-Rust/src/routes/retrieval_policy.rs",
+    "desktop/macos/agent/src/adapters/**"
+  ],
+  "status": "open"
+}
--- a/desktop/macos/Backend-Rust/fixtures/public-web-routing-contract.fixture.json
+++ b/desktop/macos/Backend-Rust/fixtures/public-web-routing-contract.fixture.json
@@ -52,6 +52,16 @@
       "requiresPublicWeb": false
     },
     {
+      "name": "fresh_phrase_inside_longer_word",
+      "prompt": "Review my recent newsletter draft",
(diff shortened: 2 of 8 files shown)


Git-on-my-level:
Addressed all three open Codex review threads and the failing `docker-context-contract` / `Hygiene` checks on this PR.

**Thread 1 — Preserve heuristic provenance through the compatibility prompt** (`pi-mono.ts:382`)
The Pi adapter wraps forced-search turns in an `<omi_retrieval_policy>` instruction containing "web search". The Rust gateway then re-classified that compatibility text as `ExplicitWeb`, turning the intended fail-open fallback into a 503 when web search is disabled or the model is Haiku. Added `strip_compatibility_retrieval_policy()` so the Rust gateway classifies the user's actual instruction, preserving `ResearchIntent` vs `ExplicitWeb` provenance. Added two Rust tests: one for the research-intent fail-open path, one confirming an explicit user web request stays strict.

**Thread 2 — Match freshness phrases on word boundaries** (`retrieval_policy.rs:173`)
"recent news" was matched as a bare substring, so "Review my recent newsletter draft" forced a server-side web search. Switched `FRESH_PUBLIC_PHRASES` (Rust) and `FRESH_PUBLIC_REQUESTS` (TS) to boundary-aware matching (`contains_any_word` / `containsWholeTerm`), matching the existing logic already used for temporal qualifiers and lookup terms. Added a contract fixture case for this.

**Thread 3 — Use the same length unit as the Rust policy** (`pi-mono.ts:371`)
Pi measured `MAX_GENERIC_LOOKUP_CHARS` in UTF-16 code units while Rust uses `text.len()` (UTF-8 bytes), so a CJK prompt could be short in TS but long in Rust. Added `utf8ByteLength()` in TS and normalized trailing punctuation identically in both classifiers. Added a CJK-over-cap contract fixture case.

**docker-context-contract + Hygiene fix:** moved the shared routing contract fixture into `Backend-Rust/fixtures/` so its `include_str!` path resolves inside the Docker build context, and updated the `FC-public-web-routing-parity` registry artifact path to match.

Verification: cargo test retrieval_policy (25 passed), npm test (600 passed), `check_docker_context.py` (3 assets covered), `tsc --noEmit` clean, CI green on all lanes.

kodjima33 (approved):
Restores Rust/Pi routing parity with one shared behavioral corpus enforcing both runtimes — desktop bug fix, CI green. Merging.

PR #10530, opened 2026-07-25
feat(release): harden M1 beta qualification lifecycle

Git-on-my-level:
## Summary

- make macOS Beta qualification M1 Studio-only and globally serialized
- isolate every qualification run with lease-owned ports, state, staging, and cleanup evidence
- remove the M4/Codemagic qualification executors while preserving normal Codemagic candidate build dispatch
- harden promotion admission to exact content-addressed qualification evidence

## Product invariants affected

- INV-AUTH-1
- INV-CHAT-1
- INV-DATA-1
- INV-BETA-1

## Failure class

Failure-Class: none

## Verification

- `bash scripts/dev-harness/run-tests.sh` — 56 passed
- `backend/.venv/bin/python -m pytest backend/tests/unit/test_qualified_beta_promotion.py -q` — 182 passed
- `bash desktop/macos/tests/test-fault-suite-launch-env.sh`
- `bash desktop/macos/tests/test-fault-suite-owned-app-cleanup.sh`
- `bash desktop/macos/tests/test-qualification-stage-isolation.sh`
- `bash desktop/macos/tests/test-qualify-desktop-beta-contract.sh`
- `python3 .github/scripts/test_desktop_release_flow_contract.py`
- `actionlint -config-file .github/actionlint.yaml .github/workflows/desktop_qualify_beta.yml`
- `python3 .github/scripts/check-release-process-guards.py`

PR #10569, opened 2026-07-25
fix(ci): give the preflight HTTP-error mocks a body so the contract suite runs on macOS Python 3.9 (#10565)

### LINKED ISSUE #10565: fix(ci): pre-push invokes PR preflight tests with macOS Python 3.9
`scripts/pre-push` selected `pr-preflight-contract-tests` for PR #10530 and invoked `python3 .github/scripts/test_pr_preflight.py` with system Python 3.9.6. Two mocked `HTTPError.close()` tests crash with `KeyError: file`; the identical suite passes under the pinned `backend/.venv/bin/python` (3.11.15): 21 passed. Make the pre-push runner consistently use its provisioned pinned interpreter or make the mocked HTTP errors compatible with supported macOS Python.

kodjima33:
## What

`pr-preflight-contract-tests` — the check every PR's check-selection depends on — stops failing on the interpreter `pre-push` actually runs it with. 21/21 pass on macOS system Python 3.9.6 as well as 3.11/3.12.

Fixes #10565.

## Why

`.github/checks-manifest.yaml` registers the suite as `python3 .github/scripts/test_pr_preflight.py`. On a stock Mac that `python3` is 3.9.6, where `urllib.error.HTTPError` still inherits the **tempfile-based** `addinfourl`: `close()` reaches `self._closer` and raises `KeyError('file')` when the error was constructed without a body. In 3.10+ `addinfourl` no longer uses `tempfile`, so the same call is harmless.

Two cases mocked urlopen failures as `HTTPError("url", code, msg, None, None)` — `fp=None`. `pr_metadata.load_from_api` closes the error body (`exc.close()`, `pr_metadata.py:103`), so both tests died inside the loader with `KeyError: 'file'` before reaching a single assertion about retry behavior. A real `urlopen` failure always carries the body, so this was purely an artifact of the mock, not a defect the loader can hit in production.

The interpreter split is why it went unnoticed: CI runners are 3.11+, so the lane stayed green while every developer on stock macOS Python got two red errors at push time (reported against PR #10530). The mocks now carry `io.BytesIO(b"")`, matching real urllib, which makes the suite interpreter-agnostic rather than pinning a lane to one Python.

Test-only change; no production code touched.

## Verification (run in this worktree)

A/B on the same file, both interpreters:

| | 3.9.6 (`/usr/bin/python3`, what pre-push uses) | 3.11.15 (pinned `backend/.venv`) | 3.12 |
|---|---|---|---|
| `HEAD` (before) | **FAILED (errors=2)** — `KeyError: 'file'` | OK 21/21 | — |
| this branch | **OK 21/21** | OK 21/21 | OK 21/21 |

- Before-state reproduced from `git show HEAD:.github/scripts/test_pr_preflight.py`, run as a sibling file so both versions saw the identical `pr_metadata.py` — the failure follows the test file, not the environment.
- The two repaired cases still assert what they are named for: `test_api_loader_retries_transient_failures_then_succeeds` (502 → timeout → success, `sleeps == [2.0, 4.0]`) and `test_api_loader_does_not_retry_permanent_http_errors` (404, single call, `RuntimeError("HTTP 404")` with the `HTTPError` cause) — both now reach those assertions instead of erroring in the loader.
- `scripts/pr-preflight --pr-body-file` (the same `.github/checks-manifest.yaml` CI runs): **11/11 checks passed in 7.14s**, `pr-preflight-contract-tests` among them — under the 3.9.6 `python3` that reddened it before.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/test_pr_preflight.py
+++ b/.github/scripts/test_pr_preflight.py
@@ -32,6 +32,15 @@ class FakeResponse(io.BytesIO):
         self.close()
 
 
+def http_error(code: int, message: str) -> urllib.error.HTTPError:
+    # An HTTPError raised by urllib always carries the response body, and the loader
+    # closes it. On Python 3.9 — the macOS system interpreter `python3` resolves to,
+    # and the one pre-push runs this suite with — HTTPError inherits the tempfile-based
+    # addinfourl, whose close() raises KeyError('file') when it was built without one.
+    # An fp-less mock therefore fails on a path production never takes.
+    return urllib.error.HTTPError("url", code, message, None, io.BytesIO(b""))  # type: ignore[arg-type]
+
+
 class MetadataTests(unittest.TestCase):
     def setUp(self) -> None:
         # `resolve_pr_metadata` short-circuits on OMI_PR_BODY_FILE before it ever
@@ -72,7 +81,7 @@ class MetadataTests(unittest.TestCase):
     def test_api_loader_retries_transient_failures_then_succeeds(self) -> None:
         payload = json.dumps({"number": 9847, "body": "ok", "updated_at": "u", "labels": []}).encode()
         outcomes: list[object] = [
-            urllib.error.HTTPError("url", 502, "bad gateway", None, None),  # type: ignore[arg-type]
+            http_error(502, "bad gateway"),
             TimeoutError("timed out"),
             FakeResponse(payload),
         ]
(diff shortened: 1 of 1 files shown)


kodjima33:
Closing: redundant. `4e783067df` ("fix(ci): make metadata retry fixtures closable", landed on main while this was in CI) already gives both mocks a body, which is the same repair. Verified on main just now: `/usr/bin/python3` (3.9.6) `.github/scripts/test_pr_preflight.py` → **21/21 OK**, where the pre-`4e783067df` file gave `FAILED (errors=2)` with `KeyError: 'file'`. Nothing left for this branch to add beyond a named helper.

PR #10571, opened 2026-07-25
fix(android): make BLE reconnect and CV1/Limitless backlog recovery transactional

axAilotl:
## Summary

- serialize Android BLE GATT operations, bind callbacks to the active connection session, retry transient busy states, and use bounded reconnect backoff
- preserve contiguous Limitless flash acknowledgments so a missing page is never acknowledged past
- make Limitless page appends retry-idempotent with durable per-device/session watermarks and interrupted-append rollback
- advance the Limitless destructive pendant ACK watermark only after the BLE write callback succeeds
- validate CV1 notification continuity and complete ring ranges before advancing the pendant
- make WAL creation, upload, reconciliation, restart recovery, and retry ownership transactional and memory-bounded
- when Android Background Mode and automatic upload are both explicitly enabled, allow only a native `deviceConnected` wake to resume backlog recovery in the background

## Why

The old Android path mixed concurrent GATT operations, accepted stale callbacks after reconnect, treated notification arrival as implicit continuity, and allowed multiple sync observers to race. A partial or failed drain could therefore look successful, while a busy or screen-off phone could leave a large pendant backlog retrying inefficiently.

Limitless flash recovery also advanced its local deletion watermark immediately after starting an asynchronous BLE ACK. If the audio fsync succeeded but the ACK failed or the connection became ambiguous, the pendant retained the page and the next drain appended the same durable audio again. The writer now records an append checkpoint before writing, fsyncs audio, and atomically replaces that checkpoint with a device/session page watermark. A restart truncates an interrupted append, while a completed page is skipped on redrain. Pendant deletion advances only from a successful GATT completion callback.

## Behavior and rollout

- failed or partial CV1 ranges remain on the pendant and retry from the same durable sequence
- locally written WALs are flushed and registered before the cumulative `ADVANCE`
- uploaded WALs reconcile before they can be offered again
- Limitless ACK failure/disconnect leaves the pendant watermark unchanged and redrain does not duplicate durable audio
- Limitless interrupted writer transactions roll back to their recorded file offset before retry
- background recovery remains opt-in and Android-only; connectivity, cooldown, and user-retry wakes still wait for foreground
- failures during a background pass schedule no background timer and remain retryable on the next foreground wake

Candidate testing on the Galaxy reached 98.7 KiB/s during CV1 backlog transfer. The shared durable-sync layer is intentionally included here because it owns the app-side acknowledgment contract used by both mobile platforms.

## Validation

- `make preflight`: deterministic manifest, hygiene, architecture, and invariant checks passed; local worktree PR-metadata discovery is unavailable because the task branch differs from the existing fork PR branch
- focused Limitless ACK/idempotency JVM regressions added for failed ACK + redrain, successful ACK completion, interrupted append rollback, and in-flight ACK coverage
- GitHub Mobile App Checks: generated files, analyzer + Flutter tests, and Android dev-debug APK compile smoke passed at `b76f2415eaa811ad5bd703faa078664df7005e47`
- prior branch validation remains: `bash app/test.sh`, Android BLE/Limitless unit tests, analyzer ratchet, focused continuity/ring/WAL/background-policy tests, integrated Galaxy/CV1 testing, and preflight

Failure-Class: none

### DIFF
--- a/app/android/app/src/main/kotlin/com/friend/ios/batch/BaseBatchAudioWriter.kt
+++ b/app/android/app/src/main/kotlin/com/friend/ios/batch/BaseBatchAudioWriter.kt
@@ -143,6 +143,25 @@ abstract class BaseBatchAudioWriter(
         }
     }
 
+    protected fun checkpointLocked(): Pair<Long, Long> = currentBytes to currentFrames
+
+    protected fun currentFilePathLocked(): String? = currentFile?.absolutePath
+
+    protected fun rollbackLocked(checkpoint: Pair<Long, Long>): Boolean {
+        val out = raf ?: return false
+        return try {
+            out.setLength(checkpoint.first)
+            out.seek(checkpoint.first)
+            currentBytes = checkpoint.first
+            currentFrames = checkpoint.second
+            out.fd.sync()
+            true
+        } catch (e: Exception) {
+            Log.w(tag, "rollback failed: ${e.message}")
+            false
+        }
+    }
+
     protected fun maybeFsyncLocked(nowMs: Long) {
         if (nowMs - lastFsyncMs >= FSYNC_INTERVAL_MS) {
             fsyncLocked()
--- a/app/android/app/src/main/kotlin/com/friend/ios/batch/LimitlessBatchAudioWriter.kt
+++ b/app/android/app/src/main/kotlin/com/friend/ios/batch/LimitlessBatchAudioWriter.kt
@@ -1,9 +1,12 @@
 package com.friend.ios.batch
 
 import com.friend.ios.ble.OmiBleManager
+import com.friend.ios.limitless.LimitlessPageDeduplicator
 
 import android.content.Context
 import java.io.File
+import java.io.RandomAccessFile
+import java.util.Locale
 import kotlin.math.abs
 
 /**
@@ -23,9 +26,10 @@ import kotlin.math.abs
  *    recordings scanner matching, and the `limitless` substring makes the backend tag
  *    the resulting conversation `source=limitless`.
  *
- * ACK safety: the drain engine must call [sync] (fsync barrier) and get `true` back
- * BEFORE ACKing pages to the pendant — an ACK deletes the pendant's copy, so bytes
- * must be durable on phone storage first. [append] itself only fsyncs periodically.
+ * ACK safety: each page append fsyncs its audio and synchronously persists a
+ * per-device/session page watermark. A redrain after an ambiguous ACK therefore
+ * recognizes the durable page instead of appending it twice. The drain engine also
+ * calls [sync] before sending the cumulative pendant-deletion ACK.
  */
 class LimitlessBatchAudioWriter(context: Context) : BaseBatchAudioWriter(context, TAG, "audio_${DEVICE_MARKER}_") {
     companion object {
@@ -35,6 +39,8 @@ class LimitlessBatchAudioWriter(context: Context) : BaseBatchAudioWriter(context
         private const val MAX_AUDIO_SECONDS = 900L // 15 min of audio per file
         private const val FRAMES_PER_SECOND = 50L // opus_fs320 = 20 ms frames
         private const val MIN_VALID_TS_MS = 1_577_836_800_000L // 2020-01-01: pendant clock sanity gate
+        private const val PENDING_PATH_KEY = "limitlessPendingAppend.path"
+        private const val PENDING_OFFSET_KEY = "limitlessPendingAppend.offset"
     }
 
     private var lastPageTimestampMs: Long = 0
@@ -42,13 +48,33 @@ class LimitlessBatchAudioWriter(context: Context) : BaseBatchAudioWriter(context
     /**
      * Append the opus frames extracted from one flash page. [pageTimestampMs] is the
      * pendant-clock capture time of that page. Returns true once the frames are
-     * written (durability requires a subsequent [sync] before ACK).
+     * written and its durable identity is committed.
      */
-    fun append(frames: List<ByteArray>, pageTimestampMs: Long): Boolean {
+    fun append(
+        deviceId: String,
+        pageIndex: Int,
+        session: Int,
+        frames: List<ByteArray>,
+        pageTimestampMs: Long,
+    ): Boolean {
         if (frames.isEmpty()) return true
         val now = System.currentTimeMillis()
 
         synchronized(lock) {
+            if (!recoverPendingAppendLocked()) return false
+            val watermarkKey = pageWatermarkKey(deviceId, session)
+            val deduplicator = LimitlessPageDeduplicator(
+                readWatermark = { prefs().getInt(it, Int.MIN_VALUE) },
+                writeWatermark = { key, value ->
+                    prefs().edit()
+                        .putInt(key, value)
+                        .remove(PENDING_PATH_KEY)
+                        .remove(PENDING_OFFSET_KEY)
+                        .commit()
+                },
+            )
+            if (deduplicator.isDurable(watermarkKey, pageIndex)) return true
+
             // Pendant clock sanity: pages recorded before the first msg6 time-sync carry
             // a bogus epoch. Inside an open session inherit the last valid timestamp so a
             // bogus page can't fake a session gap; otherwise fall back to drain wall-clock.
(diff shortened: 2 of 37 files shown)


Git-on-my-level:
Strong work on the data-safety invariants here — this is clearly the right direction for backlog reliability. The advance gate is defensively correct: `RingProtocol.canAdvance` requires `reachedDone && doneOk && !flushError && !isCancelled && receivedCompleteRange && !protocolError && crcVerified`, the WAL manifest now rewrites atomically via temp-file + rename with a serialized save barrier, source-aware WAL identity makes retries idempotent, and the `ContiguousFlashPageLedger` correctly refuses to advance the cumulative ACK watermark across a page hole. The unit-test coverage across BLE backoff, GATT queue, notification coalescing, flash ledger, ring sync, and WAL recovery is thorough.

A few things I'd want a human maintainer to confirm before this leaves draft:

1. **Draft → ready.** Marking as non-blocking review since the PR is still draft; the design is sound but merge-readiness is your call.

2. **CRC gate effectiveness.** `crcVerified` defaults to `true` (ring_storage_sync.dart) and only flips when the optional DONE CRC32 extension is present. That's a reasonable back-compat default, but worth confirming the firmware versions this targets actually emit the trailing CRC32 — otherwise the CRC check is effectively a no-op for those devices and the gate rests entirely on `receivedCompleteRange`.

3. **Scope.** This bundles BLE reconnect (Kotlin), GATT serialization (Kotlin), Limitless flash drain (Kotlin), CV1 ring sync + transactional WAL layer (Dart), and a new Android background-recovery wake policy. It's defensible as one "backlog reliability" outcome, and I'm not asking for a split for size alone — but it's a broad architectural change across user-audio durability surfaces, so it warrants maintainer product/architecture sign-off rather than auto-approval.

4. **Background recovery opt-in.** Gating the background pass to `deviceConnected` only (Android + Background Mode + auto-upload), with connectivity/cooldown/user-retry wakes still waiting for foreground, is conservative and well-tested. Just confirming that matches product intent for background audio backlog recovery.

Leaving as a positive signal — not a formal approval, since this is a draft touching data-durability and background recovery.

---

Git-on-my-level:
This is a strong, well-scoped reliability pass — the transactional data-safety model is the right design for both the CV1 ring and the Limitless pendant. Thanks for the careful work and the accompanying unit coverage.

What I verified and liked:
- `GattOperationQueue` + `isActiveGatt(gatt)` correctly binds every GATT callback to the operation and session that started it, closing the stale-callback / wrong-command-released hazard the old `ConcurrentLinkedQueue` had.
- `RingProtocol.canAdvance` requiring READ_BEGIN + byte-count + DONE (+ optional CRC) agreement before `CMD_RING_ADVANCE`, and durable WAL registration (`sourceId = ring_start_end`) happening only before the advance, is exactly the cumulative-ACK safety invariant this path needed. Out-of-order ranges correctly stay on the pendant.
- `ContiguousFlashPageLedger` guaranteeing the ACK watermark can't jump a gap, and the `fsync` barrier (`ackWritten`) blocking later ACKs after a failed flush, both correctly prevent pendant data loss on a partial drain.
- Reconnect backoff not resetting on mere physical connect (only on a stable session) is the right call — flapping transport setup failures now back off instead of hot-looping.

Two things worth a human maintainer's eye before this leaves draft:
1. On a CRC mismatch the current range is left un-advanced and retried, which is safe, but if the firmware re-sends the identical corrupt bytes the drain will loop on that range until inactivity/stall fires. Confirming the pendant can recover (or documenting that a manual clear is expected) would be useful.
2. `recoverRejectedGatt` on a descriptor-write failure forces a full reconnect — correct for safety, but worth a quick sanity check that a transient CCCD rejection under radio contention doesn't regress perceived reliability versus the old silent `completeCommand()`.

No security, dependency, CI/workflow, or agent-instruction surfaces are touched. The shared durable-sync layer's inclusion is justified by the PR's own note that it owns the app-side acknowledgment contract for both platforms. Keeping the existing `needs-maintainer-review` / `needs-tests` labels; leaving this as a positive signal rather than a formal approval since it's still draft and materially changes data-safety behavior on two recovery paths.

---

Git-on-my-level:
The latest commit (b76f2415 — "make Limitless drain ACK idempotent") directly addresses the ambiguous-ACK → duplicate-append hazard called out earlier: the pendant-deletion watermark now advances only from a successful BLE write callback, and page appends are retry-idempotent via per-device/session watermarks.

What checks out in the new idempotency layer:

- **Checkpoint/rollback WAL pattern.** `LimitlessBatchAudioWriter.append()` records the pending-append path+offset in SharedPreferences *before* writing frames, fsyncs the audio, then atomically replaces that checkpoint with the page watermark. An interrupted append is truncated to the recorded offset on restart (`recoverPendingAppendLocked`), so a crash between fsync and watermark-commit leaves no partial page.

- **`LimitlessPageDeduplicator.appendOnce`** checks the durable watermark first (idempotent return), appends + fsyncs, then commits the watermark. If the watermark write fails, it rolls back the audio bytes — no double-append is possible even across process restarts.

- **`LimitlessAckCommitGate`** transmits the cumulative ACK only after the `sync()` fsync barrier succeeds, and advances `lastAckedPageIndex` only inside the BLE completion callback. A failed callback sets `blocked`, which is sticky for the rest of the drain cycle. That's deliberately conservative: the drain ends via the existing stall/inactivity timeout with pages safely retained on the pendant, and the next cycle starts with a fresh gate.

One thing worth a quick maintainer sanity check: the sticky `blocked` flag means a *single* transient ACK failure (e.g. radio contention) stops all further ACKs in that drain, even for pages that were already durably appended. The drain won't lose data — it just won't reclaim pendant space until the next cycle. Confirming that's the intended tradeoff (safety over throughput) seems worthwhile, though it matches the conservative direction of the rest of the PR.

The Limitless idempotency JVM tests (failed-ACK + redrain, interrupted-append rollback, in-flight ACK) cover the key fault paths. Still leaving this as a positive signal rather than formal approval — draft status, the broad scope across Kotlin/Dart data-durability surfaces, and the existing `needs-maintainer-review` / `needs-tests` labels all warrant human sign-off before merge.

---

Git-on-my-level:
Thanks for this thorough rework — the transactional recovery model is well-structured and the test coverage is strong. Observations from the diff:

**GattOperationQueue / OmiBleManager.kt** — The serialized one-operation-at-a-time queue with identity-keyed callback completion (GattOperationKey) is the right fix for stale-callback-after-reconnect. The isActiveGatt(gatt) guard on every callback path correctly rejects callbacks from retired GATT objects. Worth verifying on hardware: requestMtu now completes via the queue rather than the old listener-swap in requestMtuThenNotifyReady — onMtuChanged was removed from BleConnectionListener, so confirm no other consumer depends on it.

**LimitlessAckCommitGate / LimitlessFlashDrainEngine.kt** — Advancing the pendant-deletion watermark only from the successful BLE write callback (not after write() returns) is the correct fix for duplicate-append-after-ambiguous-ACK. The BLOCKED state on fsync failure is appropriately conservative. The pendingFinishReason deferral so the drain doesn't complete before an in-flight ACK settles is a good detail.

**ContiguousFlashPageLedger** — Correctly prevents ACKing past a gap. The 64-page BUFFER_LIMIT_REACHED guard bounds memory during a persistent gap.

**LimitlessBatchAudioWriter.kt** — The checkpoint/rollback pattern (checkpointLocked then write then rollbackLocked on failure) with synchronous SharedPreferences.commit() for the per-device/session watermark is sound. recoverPendingAppendLocked() truncation on restart handles interrupted appends.

**ring_storage_sync.dart** — Bounded-range transactions (each range independently durable before CMD_RING_ADVANCE) is a significant improvement over the single cumulative ADVANCE. RingProtocol.canAdvance() combining reachedDone, doneOk, receivedCompleteRange, protocolError, and crcVerified is thorough. The immutable sourceId (ring_{start}_{end}) with file-collision check is a good idempotency boundary.

**wal_file_manager.dart** — Temp-file-then-rename atomic write with serialized _saveBarrier correctly prevents torn writes and ordering violations under concurrent saves. Backup fallback on corruption is a good resilience improvement.

**ble_packet_continuity.dart** — Modular-arity duplicate/gap detection with exact-adjacent-duplicate suppression is well-reasoned. Forwarding reused-ID packets with different payloads (reset disposition) is the safe default.

**Background recovery** — The backgroundDeviceRecoveryEnabled gate (Android + Background Mode + auto-upload, only for WakeTrigger.deviceConnected) is appropriately restrictive; tests confirm other wakes remain foreground-only.

The needs-tests label appears stale given the extensive JVM and Dart test additions across all new components. Leaving for human maintainer to clear along with security/privacy review of the pendant-deletion and WAL persistence semantics.

---

*Review by glm-5.2 (automated). Leaving for human maintainer review — the data-loss-sensitive pendant-deletion and WAL-persistence transactional model needs maintainer sign-off on the acknowledgment contract before merge.*

---

axAilotl:

PR #10573, opened 2026-07-25
fix(ios): serialize BLE recovery and reclaim CV1 after DFU

axAilotl:
## Summary

- serialize CoreBluetooth GATT work and reject stale completions from previous sessions
- coalesce notification transitions and route high-rate Omi audio / Limitless flash data through native handlers
- remove RSSI polling as a keepalive while preserving diagnostics-on-demand
- make DFU handoff idempotent and reconnect the exact suspended pendant on every terminal path
- retain the post-DFU reconnect lease across transient nulls and exceptions, retrying with capped backoff until ownership is reclaimed or the device is forgotten
- treat restored peripherals as reconnect-eligible and reissue the first unexpected-disconnect connection request synchronously

## Root cause

The old iOS path could overlap discovery, subscription, writes, and reconnect work. High-rate notifications crossed avoidable Dart/UserDefaults routing, and DFU relinquished BLE ownership without a guaranteed exact-device reclaim.

Three reconnect gaps were present:

1. a CoreBluetooth-restored peripheral was not eligible for the foreground/powered-on reconnect sweep until a fresh `didConnect`;
2. the first unexpected-disconnect retry waited 500 ms on the app queue, which iOS can suspend before the work item runs;
3. the Dart DFU handoff cleared its only suspended-device lease after a transient forced reconnect returned null or threw, so later terminal callbacks became no-ops.

The reconnect policy registers a system-owned pending `connect` synchronously after an ordinary unexpected link drop. Actual failed connection attempts still use bounded exponential backoff. The Dart DFU handoff now keeps one coalesced reclaim operation alive with capped backoff until `ensureConnection` returns ownership; explicitly forgetting the device deliberately abandons the lease.

## Physical validation

Validated on an iPhone 12 Pro Max running iOS 26.5 and one CV1:

- full dual-core CV1 OTA completed from iOS in about 13 seconds; the exact UUID was reclaimed without re-pairing
- foreground reconnect restored usable audio in about 3 seconds
- Bluetooth-off audio recovered into contiguous immutable ranges `961737..962057..962058`
- the recovered files contained 1,335 structurally valid Opus frames / 26.70 seconds with no source-sequence gap or overlap
- pendant storage advanced from 187 KB to 0 B only after the local WAL manifest and files were durable
- live audio resumed after recovery; markers `Jade 572` (background) and `Pearl 641` (final candidate) appeared in transcription
- app reinstall preserved auth and all 731 existing WAL records

### iOS 26 hard-toggle boundary

A Settings Bluetooth power toggle is not equivalent to an ordinary RF/link drop. One background cycle reconnected in 2 seconds, but a repeat with Omi already suspended remained disconnected for 34 seconds until Omi was foregrounded. Foregrounding then recovered immediately, drained the durable backlog, and resumed live audio.

Apple documents that iOS 26 restoration after user Bluetooth toggles is conditional and that Settings power toggles may not relaunch an app. This PR therefore improves the app-controlled reconnect paths without claiming that it can override that OS policy. The firmware/app durability boundary preserves the missing interval for later backfill.

## Validation

- `flutter test --no-pub test/providers/device_provider_test.dart` — 18 passed, including null/throw-then-success, concurrent terminal callback coalescing, capped-backoff continuation, and explicit forget cancellation
- `bash app/ios/test/run_ble_gatt_scheduler_tests.sh` — passed
- `make preflight` / pre-push manifest — 64 selected checks passed
- Flutter generated-output hook deliberately skipped with `PRE_PUSH_SKIP_FLUTTER_GENERATED=1`: the shared Flutter installation reported SDK version `0.0.0-unknown`; full Flutter analyzer/tests and platform compiles remain CI-authoritative

Failure-Class: none

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -204,6 +204,11 @@ checks:
     triggers: ["app/scripts/rayban_dat.sh", "app/ios/Podfile", "app/ios/rayban_dat_plugin_boundary.rb", "app/ios/test/rayban_dat_build_wrapper_test.rb", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "Ray-Ban DAT builds must strip mcumgr before CocoaPods and restore the exact default graph even after interruption"
+  - id: ios-ble-gatt-operation-scheduler
+    command: ["bash", "app/ios/test/run_ble_gatt_scheduler_tests.sh"]
+    triggers: ["app/ios/Runner/Ble/**", "app/ios/Runner.xcodeproj/project.pbxproj", "app/ios/test/BleGattOperationSchedulerTests.swift", "app/ios/test/BleManagerDependencyTypecheckStubs.swift", "app/ios/test/ble_gatt_project_graph_test.rb", "app/ios/test/run_ble_gatt_scheduler_tests.sh", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "PR #5796 shipped characteristic-keyed completions and PR #6085 shipped RSSI-as-keepalive; the tested native boundary keeps CV1 commands session-bound, serialized, and free of unconditional polling"
   - id: deferred-work-markers
     command: ["python3", ".github/scripts/deferred-work-marker-count.py", "--changed-files", "{changed_files}", "--base", "{base}", "--check-new"]
     triggers: ["all"]
--- a/app/ios/Runner.xcodeproj/project.pbxproj
+++ b/app/ios/Runner.xcodeproj/project.pbxproj
@@ -36,6 +36,7 @@
 		42B17C622F69B9CF00BFDA9D /* OmiBleManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42B17C612F69B9CF00BFDA9D /* OmiBleManager.swift */; };
 		42B17C642F69B9E100BFDA9D /* BleHostApiImpl.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42B17C632F69B9E100BFDA9D /* BleHostApiImpl.swift */; };
 		C0D3B1E00000000000000002 /* BleGattOperationScheduler.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0D3B1E00000000000000001 /* BleGattOperationScheduler.swift */; };
+		C0D3B1E10000000000000002 /* BleNotificationRouter.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0D3B1E10000000000000001 /* BleNotificationRouter.swift */; };
 		42B17C7C2F69BC6600BFDA9D /* PigeonCommunicator.g.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42B17C7B2F69BC6600BFDA9D /* PigeonCommunicator.g.swift */; };
 		42BABFB82F2A81B100E78A3C /* OmiRecordingAudioDevice.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42BABFB62F2A81B100E78A3C /* OmiRecordingAudioDevice.swift */; };
 		42BABFB92F2A81B100E78A3C /* OmiPhoneCallsPlugin.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42BABFB72F2A81B100E78A3C /* OmiPhoneCallsPlugin.swift */; };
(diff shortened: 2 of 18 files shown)


Git-on-my-level:
Thanks @axAilotl — this is a well-structured reliability pass. The serialized GATT scheduler (session-bound tokens, stale-callback rejection, read-while-notifying admission), notification-state coalescing, native notification routing, and the idempotent DFU connection handoff all address real races in the iOS BLE stack. The test coverage (scheduler/router/project-graph suites + the Dart DFU handoff and device-provider tests) and the new `ios-ble-gatt-operation-scheduler` manifest check are exactly the kind of guard surface that makes a change like this reviewable. Documenting the physical CV1/iOS 26.5.2 validation in the PR body is appreciated.

A few points worth a human maintainer's eyes before this leaves draft:

- **RSSI keep-alive removal.** The old path polled `readRSSI()` every 3 s unconditionally on `deviceReady`; the new path only polls while the diagnostics view is streaming. The rationale (controller-owned supervision, not an app keep-alive) is technically sound, but this is a behavior change with device/firmware-specific implications. Worth confirming across the supported pendant variants that no firmware relies on host-side traffic to avoid supervision timeouts, especially in background/restoration.
- **Notification-state mismatch handling.** When `setNotifyValue` returns success but `isNotifying` doesn't match the requested state (`ignoreMismatchedState`), the operation is left for the scheduler timeout rather than invalidated. That's defensible, but a maintainer should confirm the 10 s fatal-timeout → disconnect on a benign state mismatch is the desired recovery vs. a softer path.
- **DFU suspend/resume state machine.** The `_dfuSuspendedDeviceId` / `_dfuSuspendFuture` / `_dfuResumeFuture` triple in `DeviceService` correctly captures the device id before disconnect and reclaims the exact device on any terminal outcome. The single-device assumption (throwing `StateError` if a different device suspends mid-DFU) is fine for current product scope; just flagging it as a seam to revisit if multi-device DFU ever lands.
- **Diff churn in `OmiBleManager.swift`.** A large share of the ~600 net-new lines is 4-space→2-space reformatting alongside the real logic. Not asking for a split — the work is cohesive around one subsystem — but when it leaves draft, a whitespace-ignoring review pass (`git diff -w`) helps confirm no behavior is hidden in the reflow.

Marking for human maintainer review given the core reliability surfaces (BLE session/reconnect lifecycle, DFU ownership transfer) and the draft state. Not formally approving.

_Posted by the Omi PR maintainer monitor (GLM-5.2)._

---

axAilotl:
## Build 110 composite physical evidence (2026-08-04)

The session-bound iOS GATT changes in this PR were included in the exact composite dev IPA exercised on a physical iPhone/CV1.

- Cold app launch + pendant power-cycle recovered automatically.
- A controlled marker appeared in live preview without tapping **Listening**.
- Manual SD backlog drain remained active while a second marker appeared and storage decreased.
- No freeze/disconnect was observed in the controlled interval.
- Native verification passed:
  - `bash app/ios/test/run_ble_gatt_scheduler_tests.sh`
  - `bash app/ios/test/run_ble_legacy_audio_ownership_tests.sh`
  - `ruby app/ios/test/ble_gatt_project_graph_test.rb` (6 runs, 52 assertions)

Timing caveat: “+5 s” was marker-end to screenshot, not cold-launch-to-first-text latency, and is not being claimed as a cold-start SLO. The composite integration is preserved on fork branch `codex/ios-build110-qualification` at `8f0f7e0774`; app-side evidence is in #10656 at `90927b4562`.

Git-on-my-level:
Reviewed the iOS BLE serialization and DFU reconnect rework (glm-5.2 automated review). The design is coherent and well-tested; leaving detailed observations for the maintainer.

**Strengths**

- `BleGattOperationScheduler` is a clean serialization primitive: one operation in flight per peripheral, bound to a connection session via `sessionId`, with `completeExpected`/`failExpectedAndInvalidate` rejecting callbacks from the wrong session/kind/characteristic. The 683-line test file covers serialization, stale-session rejection, timeout (fatal and non-fatal RSSI), and session replacement.
- `BleGattReadAdmission` rejecting reads on notifying characteristics correctly addresses the CoreBluetooth ambiguity where `didUpdateValueFor` cannot distinguish a read response from a notification.
- `BleNotificationRouter` routing by fixed service/characteristic UUID identity before consulting UserDefaults-backed capture policy is a meaningful hot-path improvement for high-rate ring-sync traffic.
- The Dart `FirmwareDfuConnectionHandoff` idempotency (`_finishFuture` coalescing) and the `DeviceService` backoff-with-reclaim loop correctly address the transient null/throw lease-loss bug described in the root cause.
- Removing RSSI polling as a keepalive while preserving diagnostics-on-demand is the right call given connection supervision is controller-owned.

**Observations for maintainer attention**

1. `BleReconnectDispatchPolicy`, `BleReconnectLifecycle`, and `BleReconnectEligibility` are referenced in `OmiBleManager.swift` but I could not locate their definitions in any committed `.swift` file in this PR diff. Desktop Swift Build & Tests CI passed, so they likely resolve in-repo — worth confirming they are committed and not an unresolved symbol.

2. `requestPersistentReconnect` calls `centralManager.connect` synchronously inside `didDisconnectPeripheral`. This is the intended fix (avoiding the 500ms dispatch delay that iOS can suspend), but it means every unexpected disconnect now issues a connect on the main queue before the delegate returns. Worth verifying this does not interact adversely with rapid disconnect/reconnect cycles under RF congestion.

3. In `DeviceService._resumeConnectionAfterDfu`, the reconnect loop polls `ensureConnection(deviceId, force: true)` with backoff `[0, 250ms, 1s, 2s, 4s]`. After exhausting the list, it continues retrying at 4s indefinitely until `_dfuSuspendedDeviceId` changes. This is intentional (reclaim until success or forget), but there is no maximum attempt cap — if the device is physically gone, only `forgetDevice` stops the loop. The test covers explicit-forget cancellation, which is good.

4. `peripheralIsReady(toSendWriteWithoutResponse:)` completes the pending write token directly via `complete(token:result:)`. If the scheduler session was invalidated between enqueue and this callback, the `activeToken == pending.token` guard will fail and the write is silently dropped — correct for a stale session, just flagging it.

**Verification status**: All CI checks green (Desktop Swift Build & Tests, Dart Analyze & Tests, Desktop Swift Static & Test Contracts, Formatting, Hygiene). Could not build/test locally — the worktree fetch failed (non-fast-forward on the PR head ref). Static review only.

This needs maintainer sign-off on the reconnect policy and notification routing changes before merge.

---

axAilotl:

PR #10574, opened 2026-07-25
fix(ci): make pre-push portable on Windows

### LINKED ISSUE #9724: Pre-push single-flight runner crashes on Windows without SIGHUP

### LINKED ISSUE #10293: make setup and pre-push hook fail with 'this operation must be run in a work tree' in linked git worktrees

tianmind-studio:
Fixes #9724.
Fixes #10293.

## Summary

- make the single-flight runner register only host-supported signals and replace destructive Windows `os.kill(pid, 0)` probing
- launch the extensionless pre-push script through Git Bash and carry one explicit Python executable across every hook boundary, including the `PRE_PUSH_SKIP_PR_PREFLIGHT=1` failure-class hatch
- contain the Windows command tree in a kill-on-close Job Object, with a bootstrap barrier that assigns the job before the real command can spawn
- bind stale-lock identity to Windows process creation ticks so PID reuse cannot impersonate the prior owner
- decode Git metadata as UTF-8 so native Windows locales can resolve non-ASCII worktree paths
- make the check manifest Windows-aware and add a path-scoped native Windows workflow that executes the manifest-owned portability checks
- preserve the mainline `forwardable_signals` and direct-child fallback contracts while extending Windows ownership to the full process tree
- restore the linked-worktree repo-root fallback across all five downstream shell/Python pre-push entrypoints, so a `show-toplevel` exit 128 does not force contributors toward `--no-verify`
- keep native process tests event-driven with pipe handshakes and `WaitForSingleObject`, without wall-clock polling or leaked handles

## Root cause

The pre-push control plane assumed POSIX signals, process groups, executable script launching, virtualenv layout, and `python3` naming at native Windows boundaries. It also let Python decode Git's UTF-8 output with the active Windows code page, so a worktree path containing non-ASCII characters could fail before any check ran.

A second gap remained after #10401: the installed hook dispatcher could recover when `git rev-parse --show-toplevel` failed in a linked worktree, but `scripts/pre-push`, `scripts/pre-push-singleflight`, `scripts/pr-preflight`, `.github/scripts/preflight_runner.py`, and `.github/scripts/pr_preflight.py` repeated the unguarded lookup and aborted later in the same chain.

Fixing only the first Windows exception or only the Makefile/dispatcher boundary would expose the next launch, liveness, decoding, root-resolution, or hatch failure. This PR closes that reachable portability chain.

## Attribution

The linked-worktree root-resolution slice preserves Nik Shevchenko's original #10491 implementation through the rebased commit lineage, with `Co-authored-by: Nik Shevchenko <kodjima33@gmail.com>`. Its regression was adapted to the explicit Windows Python launcher introduced by this PR and registered in both manifest lanes plus native Windows CI.

## Verification

- rebased onto current `origin/main` (`ffc13daf4c`) and reran the native Windows checks
- native Windows manifest run: `pre-push-worktree-fallback`, `preflight-runner-portability`, and `check-manifest-contract` -> 3 checks passed
- native Windows: `test_preflight_runner.py` -> 23 tests passed with `ResourceWarning` promoted to an error, including a real Job Object parent/descendant termination proof, explicit hatch interpreter coverage, and runner/direct PR-preflight execution from a Chinese worktree path with `PYTHONUTF8=0`
- native Windows: a temporary Git repository ran the real `pre-push-singleflight -> preflight_runner -> pre-push` chain with no `python3` on `PATH`
- native Windows: `test_run_checks.py` -> 33 tests passed, 1 POSIX-only fixture skipped; the real manifest subprocess path also resolved `python3` to the active interpreter with an empty `PATH`
- native Windows: full `test_pr_preflight.py` -> 26 tests passed, 3 explicit host-capability skips
- native Windows: `test-pre-push-worktree-fallback.sh` passed against all five downstream root resolvers under the reproduced exit-128 condition
- `test_pre_push_ci_prediction.py` -> 14 tests passed
- `scripts/test-pre-push-diff-base.sh` passed under Git Bash
- `bash -n` passed for all four touched shell scripts
- Black 26.5.1 passed for the touched Python files with the repository's 120-column/no-string-normalization settings
- actionlint 1.7.12 passed for the Windows workflow
- `python -m py_compile` passed for all touched Python files
- `scripts/failure-class validate --base origin/main --head HEAD ...` passed with `Failure-Class: none`
- `git diff --check origin/main...HEAD` passed

GNU Make is not installed on the local Windows host, so `scripts/test-make-setup.sh` could not be rerun there. The new regression exercises every downstream root resolver directly on native Windows; Linux CI remains authoritative for the existing Make fixture and the full manifest.

The committed diff changes the manifest, so the legacy local selector intentionally attempts every manifest check. On native Windows that broad run currently stops while collecting the existing `dev-harness-unit-tests` because its qualification module imports POSIX-only `fcntl`; the path-scoped Windows workflow therefore executes the portability contracts by manifest check ID, while Linux CI remains the full-manifest authority.

none

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -49,6 +49,11 @@ checks:
     triggers: [".github/workflows/desktop_qualify_beta.yml", ".github/scripts/check-release-process-guards.py", "scripts/run-release-process-guards.sh", "backend/tests/unit/test_desktop_release_scripts.py", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "SCA-174: qualification must check out the exact release tag inside its isolated source directory; the guard and its focused behavioral test run on every relevant diff"
+  - id: preflight-runner-portability
+    command: ["python3", ".github/scripts/test_preflight_runner.py"]
+    triggers: [".github/scripts/preflight_runner.py", ".github/scripts/test_preflight_runner.py", ".github/scripts/run_checks.py", ".github/scripts/test_run_checks.py", "scripts/pre-push-singleflight", "scripts/pre-push", "scripts/pr-preflight", ".github/workflows/windows-preflight-portability.yml", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the single-flight pre-push gate must manage child processes and launch Bash checks on native Windows and POSIX hosts"
   - id: failure-class-retirement-author
     command: ["python3", ".github/scripts/test_failure_class_retirement.py"]
     triggers: [".github/scripts/failure_class_retirement.py", ".github/scripts/test_failure_class_retirement.py", ".github/workflows/repo-hygiene.yml", "scripts/failure-class", ".github/checks-manifest.yaml"]
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -51,7 +51,7 @@ checks:
     reason: "SCA-174: qualification must check out the exact release tag inside its isolated source directory; the guard and its focused behavioral test run on every relevant diff"
   - id: preflight-runner-portability
     command: ["python3", ".github/scripts/test_preflight_runner.py"]
-    triggers: [".github/scripts/preflight_runner.py", ".github/scripts/test_preflight_runner.py", ".github/scripts/run_checks.py", ".github/scripts/test_run_checks.py", "scripts/pre-push-singleflight", "scripts/pre-push", "scripts/pr-preflight", ".github/workflows/windows-preflight-portability.yml", ".github/checks-manifest.yaml"]
+    triggers: [".github/scripts/preflight_runner.py", ".github/scripts/pr_preflight.py", ".github/scripts/test_preflight_runner.py", ".github/scripts/run_checks.py", ".github/scripts/test_run_checks.py", "scripts/pre-push-singleflight", "scripts/pre-push", "scripts/pr-preflight", ".github/workflows/windows-preflight-portability.yml", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "the single-flight pre-push gate must manage child processes and launch Bash checks on native Windows and POSIX hosts"
   - id: failure-class-retirement-author
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -49,6 +49,11 @@ checks:
     triggers: [".github/workflows/desktop_qualify_beta.yml", ".github/scripts/check-release-process-guards.py", "scripts/run-release-process-guards.sh", "backend/tests/unit/test_desktop_release_scripts.py", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "SCA-174: qualification must check out the exact release tag inside its isolated source directory; the guard and its focused behavioral test run on every relevant diff"
+  - id: pre-push-worktree-fallback
+    command: ["bash", "scripts/test-pre-push-worktree-fallback.sh"]
+    triggers: ["scripts/pre-push", "scripts/pre-push-singleflight", "scripts/pr-preflight", ".github/scripts/preflight_runner.py", ".github/scripts/pr_preflight.py", "scripts/test-pre-push-worktree-fallback.sh", ".github/workflows/windows-preflight-portability.yml", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "#10293: the pre-push hook chain must resolve the repo root in linked worktrees where git rev-parse --show-toplevel exits 128, not abort into a --no-verify push"
   - id: preflight-runner-portability
     command: ["python3", ".github/scripts/test_preflight_runner.py"]
     triggers: [".github/scripts/preflight_runner.py", ".github/scripts/pr_preflight.py", ".github/scripts/test_preflight_runner.py", ".github/scripts/run_checks.py", ".github/scripts/test_run_checks.py", "scripts/pre-push-singleflight", "scripts/pre-push", "scripts/pr-preflight", ".github/workflows/windows-preflight-portability.yml", ".github/checks-manifest.yaml"]
--- a/.github/scripts/pr_preflight.py
+++ b/.github/scripts/pr_preflight.py
@@ -23,6 +23,18 @@ class Check:
     reason: str
 
 
+def _resolve_repo_root() -> Path:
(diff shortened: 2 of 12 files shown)


Git-on-my-level:
Thanks for tightening the Windows pre-push path here. I reviewed this as workflow-sensitive CI/hook code rather than a normal app change.

What looks good:
- The patch addresses the actual Windows portability chain instead of only papering over the missing POSIX signal constant.
- The new runner tests cover host-specific signal selection, avoiding destructive Windows `os.kill(pid, 0)`, PID-reuse protection via process creation ticks, explicit Bash/Python launcher boundaries, invalid UTF-8 output, and Windows-only job-object behavior where the host can exercise it.
- The new Windows workflow is path-scoped and uses `contents: read`, so I do not see an obvious privilege expansion or supply-chain red flag.

I also locally ran the relevant portable checks on the PR head (`45fbe17b`):
- `python3 .github/scripts/test_preflight_runner.py`
- `python3 .github/scripts/test_run_checks.py`
- `python3 .github/scripts/run_checks.py --lane ci --base HEAD --head HEAD --platform windows --check-id preflight-runner-portability --check-id check-manifest-contract`
- `bash -n scripts/pre-push scripts/pre-push-singleflight scripts/pr-preflight`
- `python3 -m py_compile .github/scripts/preflight_runner.py .github/scripts/run_checks.py .github/scripts/test_preflight_runner.py .github/scripts/test_run_checks.py`

All passed in my Linux review environment, and GitHub's native Windows portability check is also green. Because this adds/changes CI and pre-push control-plane behavior, I am not formally approving; a human maintainer should still review the workflow/hook surface before merge. My maintainer signal is positive pending that human workflow review.

---

tianmind-studio:
#10594 is the small manifest-platform slice of this PR; if it merges first, I will rebase this branch and drop the duplicate platform lines. #10595 is complementary but independent: it fixes UTF-8 decoding in other preflight guards and is not subsumed by this runner/launcher change.

tianmind-studio:
CI follow-up on refreshed head `a0a4622002`: every effective check for this PR passed, including the native Windows contracts, metadata preflight, backend unit suite, generated files, Dart, and Android compile. The only red lane is the unrelated macOS `PTTInputDeviceProbeTests.testMainRunLoopKeepsPumpingWhileTheHALIsWedged` throughput flake (`ticks == 7`, `reads == 7`) in [run 30283186121](https://github.com/BasedHardware/omi/actions/runs/30283186121).

The same test already failed on current `main` and the prior #10574 head with different positive counts, so this is not introduced by the Windows pre-push diff. I tracked it in #10724 and opened the one-file deterministic test fix in #10725.

Git-on-my-level:
Reviewed the refreshed head `a0a4622` as workflow-sensitive CI/hook code.

Positive maintainer signal from my side:
- The new linked-worktree fallback regression is focused and covers the remaining downstream shell/Python entrypoints that could still abort after #10401.
- The Windows runner changes keep the workflow path-scoped with `contents: read`; I do not see a privilege expansion, supply-chain concern, or agent-instruction behavior change.
- Focused local verification passed under a stripped environment: `test_preflight_runner.py`, `test_run_checks.py`, `test_pr_preflight.py`, shell syntax checks, Python compilation, and the manifest-selected Windows portability checks including `pre-push-worktree-fallback`.

I am still not formally approving because this touches CI/pre-push control-plane behavior and the PR already carries `workflow-review` / `needs-maintainer-review`. Also, GitHub currently shows red Desktop Swift checks; they look outside this hook portability surface, but a human maintainer should decide whether to rerun/waive those before merge.

---

tianmind-studio:
CI is now fully green on refreshed head `e4609d8aa4`: the native Windows process/launcher contracts, Backend Hermetic Merge Gate, backend unit suite, Desktop Swift checks, Hygiene, Dart, Android, and generated-file checks all pass. The prior red Hermetic run was an `actions/checkout` fetch cancelled at the workflow's two-minute limit; I rewrote only the top commit metadata to retrigger it because fork contributors cannot rerun upstream Actions. The code diff is unchanged from `4ca8e0aeb3`.

Git-on-my-level:
Reviewed refreshed head `e4609d8aa4` as workflow-sensitive CI/hook code.

Updated maintainer signal:
- The code diff remains focused on the Windows/native pre-push and manifest-runner portability surface: explicit Python/Bash launcher boundaries, Windows job-object process ownership, PID-reuse-safe lock checks, UTF-8 Git output handling, and linked-worktree repo-root fallback.
- I do not see a privilege expansion, dependency/supply-chain concern, or agent-instruction behavior change. The new workflow stays path-scoped and uses `contents: read`.
- Focused local verification passed under a stripped environment on this head: `test_preflight_runner.py`, `test_run_checks.py`, `test_pr_preflight.py`, shell syntax checks, Python compilation, and the manifest-selected Windows portability checks including `pre-push-worktree-fallback`.
- GitHub now shows the relevant effective checks green, including the native Windows process/launcher contracts.

I am still not formally approving because this changes CI/pre-push control-plane behavior and the PR carries `workflow-review` / `needs-maintainer-review`; a human maintainer should make the final workflow sign-off before merge. My current signal is positive pending that human review.

---

tianmind-studio:
#10825 now carries the focused shared Git Bash/interpreter-resolution slice on current main, including the production manifest runner plus the release-guard and SwiftLint consumers. If #10825 lands first, I will rebase this branch and drop the overlapping run_checks.py interpreter-alias changes while retaining this PR's unique pre-push Job Object, signal, PID-reuse, linked-worktree, and native-Windows workflow coverage.

Issue #10575, opened 2026-07-25
Merging a PR cancels its own in-flight desktop Swift verification, and the closed-event run reports "skipping"


Git-on-my-level:
#10523 merged a change that does not compile, and `main` stayed broken for ~12h. The build defect is fixed in the PR that links here; this issue is about the gate that let it through.

**Timeline for `desktop-swift-ci.yml` on `agent/m1-studio-codex-terra/5a29ba2b-1784948172` (#10523):**

| run | conclusion | head | time |
|---|---|---|---|
| 30141952816 | **failure** | 61092337bc | 03:16 |
| 30145497958 | cancelled | 0b5d4e0f9d | 05:17 |
| 30145978453 | cancelled | 73a11a734a | 05:33 |
| 30146418506 | cancelled | fe5253dfe0 | 05:49 |
| 30146613315 | **skipped** | fe5253dfe0 | 05:55:56 |

The check never passed. The run on the final head SHA was still in flight when the PR merged at 05:55:53, and the merge cancelled it — the `closed` event lands in the same concurrency group:

```yaml
types: [opened, synchronize, reopened, closed]
concurrency:
  group: desktop-swift-${{ github.event.pull_request.number || github.sha }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Every job is gated `if: github.event.action != 'closed'`, so the `closed` run allocates no runner and reports `skipping` — which is the state `gh pr checks 10523` shows today, and which reads as benign.

This is deliberate and contract-tested: `test_closed_prs_release_the_same_pr_concurrency_group_without_allocating_a_runner` (`.github/scripts/test_desktop_swift_ci_contract.py:124`), landed in `bd93173e8b` "ci: bound and retire obsolete Swift PR runs" to stop abandoned PRs holding scarce macOS runners. That goal is right. The problem is that `closed` does not distinguish **abandoned** from **merged**, and for a merge it supersedes the verification of the exact code being landed.

The post-merge `push` run did catch it (30146613366, failure on 3a63002ea9) — but that signal decays: the two subsequent main runs reported success purely because they touched no Swift and skipped the lane, so after ~40 minutes nothing on `main` looked red.

Two independent things worth deciding, both above my pay grade to change unilaterally since they trade runner capacity and branch-protection policy:

1. Should `cancel-in-progress` exclude the merge case (`github.event.action == 'closed' && github.event.pull_request.merged`)?
[...]

PR #10576, opened 2026-07-25
fix(desktop): restore the Swift build broken on main by SCA-103

### LINKED ISSUE #10570: pre-push desktop Swift compile cannot pass on Xcode 26.2 (Swift 6.2 strict concurrency)
`scripts/pre-push` runs `xcrun swift build -c debug --package-path Desktop` against whatever local Xcode is selected — deliberately, per `desktop/macos/AGENTS.md:249` ("never promote it to the full pinned-Xcode suite ... that push-time budget belongs to CI"). CI pins Xcode 16.4. On Xcode 26.2 / Apple Swift 6.2.3 the same sources do not compile, so the gate fails for a diff that touches no Swift at all.

Reproduced on `cb918cf0ba` (current main) with **zero** desktop files changed — `git diff --name-only origin/main...HEAD -- 'desktop/**'` is empty, so these are main's sources unmodified:

```
Rewind/Core/RewindDatabase.swift:10:22: error: static property 'terminationInProgress' is not
  concurrency-safe because it is nonisolated global shared mutable state [#MutableGlobalVariable]
Providers/ChatProvider.swift:1667:52: error: non-Sendable 'DesktopErrorDiagnosticContext?'-typed
  result can not be returned from actor-isolated instance method
  'consumePendingStartFailureDiagnostics()' to main actor-isolated context
ProactiveAssistants/Assistants/Focus/FocusAssistant.swift:694:26: error: cannot find 'DesktopLocalProfile' in scope
Rewind/Services/RewindIndexer.swift:441:26: error: cannot find 'DesktopLocalProfile' in scope
```

The two `cannot find 'DesktopLocalProfile' in scope` errors look like fallout from `error: emit-module command failed` rather than independent defects — `DesktopLocalProfile` is `package enum` in `OmiSupport/DesktopLocalProfile.swift` and resolves fine under the pinned toolchain.

Why it matters beyond one machine: `.github/checks-manifest.yaml` is in `ROUTING_INPUTS` (`scripts/pre_push_ci_prediction.py:71`), so *any* manifest edit conservatively wakes `desktop-swift-tests`. A contributor on current Xcode who edits only CI config gets a desktop compile failure they did not cause and cannot fix. `PRE_PUSH_SKIP_DESKTOP_SWIFT=1` clears it and is what I used, but the skip is indistinguishable from skipping a real regression.

Environment: macOS 26.x arm64, Xcode 26.2 (17C52), swift-driver 1.127.14.1 / Apple Swift 6.2.3. Not urgent — the hatch works — but the local desktop gate is currently advisory-in-practice for anyone not on 16.4.

Git-on-my-level:
## Summary

**`main` does not compile.** The desktop Swift package has been broken since 2026-07-25 05:55 UTC. This restores it, and separately unblocks the local pre-push desktop gate on current Xcode.

## Part 1 — the broken build

`desktop-swift-ci` on merge commit `3a63002ea9` ([run 30146613366](https://github.com/BasedHardware/omi/actions/runs/30146613366)) **failed**. Every later main run reported success only because it changed no Swift and skipped the lane, so the red state stayed invisible for ~12 hours.

Three compile errors, all introduced by `d0215a40a0` ("chore(desktop): add storage and bridge diagnostics", #10523):

| Error | Cause |
|---|---|
| `FocusAssistant.swift:694: cannot find 'DesktopLocalProfile' in scope` | missing `import OmiSupport` |
| `RewindIndexer.swift:441: cannot find 'DesktopLocalProfile' in scope` | missing `import OmiSupport` |
| `ChatProvider.swift:1667: non-sendable result type 'DesktopErrorDiagnosticContext?' cannot be sent from actor-isolated context` | type is not `Sendable` but crosses an actor boundary |

`DesktopLocalProfile` is a `package enum` in the **separate** `OmiSupport` target (`Package.swift` excludes `Sources/OmiSupport` from the `Omi Computer` target), so a file that uses it needs the import. Every pre-existing caller — `AuthService.swift`, `OmiApp.swift`, `AgentVMService.swift` — already has it. The two new call sites did not.

`DesktopErrorDiagnosticContext` is produced on whichever actor detected the failure (`AgentRuntimeProcess`) and consumed on `@MainActor` `ChatProvider`, so it must be `Sendable`. Its element type becomes `any Sendable` rather than `Any`, which makes the compiler enforce the scalar-only contract the type's own doc comment already states:

> Callers must construct these from enums, booleans, and numeric measurements; paths, exception messages, identifiers, and user content do not belong here.

Every existing value is already `String`/`Int`/`Int64`/`Bool`, so **no construction site changes** — only the two declarations.

**Scope was measured, not guessed.** I scanned every `Sources/` file outside the `OmiSupport`/`Theme`/`OmiWAL`/`VoiceTurnDomain` targets for all four symbols `OmiSupport` exports (`DesktopLocalProfile`, `DesktopStorageIdentity`, `DesktopDevRuntimeManifest`, `DesktopDevRuntimeManifestStore`). Exactly three files use one without importing; the third, `AppBuild.swift`, mentions `DesktopStorageIdentity` only inside a doc comment and correctly needs no import.

## Part 2 — the local gate on Swift 6.2

Separate commit, separate concern. `scripts/pre-push` runs `xcrun swift build -c debug` against whatever Xcode is selected locally — deliberately, per `desktop/macos/AGENTS.md:249`: *"never promote it to the full pinned-Xcode suite ... that push-time budget belongs to CI."* CI pins Xcode 16.4. On Xcode 26.2 / Swift 6.2.3 the package additionally fails:

```
Sources/Rewind/Core/RewindDatabase.swift:10:22: error: static property 'terminationInProgress'
  is not concurrency-safe because it is nonisolated global shared mutable state
```

Every access already goes through `terminationLock` — `isTerminationInProgress` reads under it, and the sole write at line 352 takes it — so the property *is* correctly synchronized; the compiler cannot see an `NSLock`. `nonisolated(unsafe)` states that invariant where the compiler can act on it. No runtime behavior changes, and the spelling has been valid since Swift 5.10, comfortably below the pinned toolchain.

Effect: the debug build a contributor is asked to run before every push now succeeds on current Xcode instead of failing on code they did not touch. Closes #10570.

## Verification

Clean worktree at `origin/main` (`cb918cf0ba`) + this diff, Xcode 26.2 (17C52) / Apple Swift 6.2.3:

```
xcrun swift build -c debug --package-path .               -> Build complete! (189.40s), exit 0
xcrun swift build -c debug --build-tests --package-path . -> Build complete!, exit 0
```

Negative direction, because a fix that was never seen failing is not known to fix anything:

- Without either commit, the same command reproduces all three Part 1 errors, matching CI's Xcode 16.4 log ([run 30167210176](https://github.com/BasedHardware/omi/actions/runs/30167210176), job 89702239163) line-for-line.
- Reverting **only** the `nonisolated(unsafe)` hunk reproduces the Part 2 error alone, so that annotation is load-bearing rather than decorative.

`terminationInProgress` appears **0** times in CI's log and `DesktopLocalProfile` **98** times, which is how the two parts were separated: Part 1 is what CI sees, Part 2 is what only a newer local toolchain sees.

### What I did not verify

I did not run the desktop Swift **test suite**, only its compile (`--build-tests` links `Omi ComputerPackageTests` successfully). The suite needs the pinned Xcode 16.4 for a trustworthy result and this machine has 26.2; CI is the authority and runs it on this PR. No test source is touched by this diff.

## A gap this exposed, not fixed here

Worth stating so it is not mistaken for covered. The desktop Swift check **never passed** on #10523 — its runs went `failure`, `cancelled`, `cancelled`, `cancelled`, then `skipped`. The final `cancelled` was caused by the merge itself: `desktop-swift-ci.yml` includes `closed` in its PR trigger with `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`, so merging a PR supersedes its own in-flight verification, and the resulting `closed` run reports `skipping`.

That behavior is deliberate and contract-tested (`test_closed_prs_release_the_same_pr_concurrency_group_without_allocating_a_runner` — it protects scarce macOS runner capacity when a PR is abandoned). It is correct for a PR closed *unmerged* and load-bearing in the wrong direction for one closed *by merging*. Changing it means distinguishing `github.event.pull_request.merged`, which trades runner capacity for merge safety — a call for the owner of that contract, and not something I can verify without merging it. Filed separately as #10575 with the full timeline rather than bundled into a build fix.

- **INV-AUTH-1** (`docs/product/invariants/auth-session.md`)

Path-matched via `desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift`. The change there is a `nonisolated(unsafe)` annotation on an already-lock-guarded private static; it touches no session invalidation policy, 401 handling, or ownership transition, and its guard tests are unmodified.

## Failure class

Failure-Class: none

Two ordinary instance fixes. No registered class covers "a symbol from a sibling SwiftPM target was used without importing it" — that is what the compiler is for, and it did catch it; the miss was in the merge gate, which is #10575's subject and not repaired here.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:

PR #10587, opened 2026-07-25
fix(dev-harness): make test paths portable on Windows

tianmind-studio:
## What changed and why

Use `os.pathsep` when test subprocesses prepend the dev-harness package to `PYTHONPATH`, and compare provider fake locations as `Path` components instead of slash-specific strings. This removes the Windows-only failures without changing harness production behavior.

## Product invariants affected

none

## How it was verified

- `python -m pytest scripts/dev-harness/tests/test_memory_scenarios.py scripts/dev-harness/tests/test_providers.py -q` (`19 passed`, native Windows on the clean `main` base)
- `python -m pytest scripts/dev-harness/tests/test_cli.py scripts/dev-harness/tests/test_memory_scenarios.py scripts/dev-harness/tests/test_providers.py -q` (`25 passed`, native Windows with #10583 temporarily overlaid to unblock `test_cli.py` collection)
- complete native Windows dev-harness run with the same temporary overlay (`56 passed, 7 failed`; the remaining failures are unchanged Bash/Make, Windows-invalid-filename, and POSIX process-group tests)
- Black 24.4.2 and Ruff 0.4.4 on the three changed test files
- nine other selected deterministic manifest guards via #10574's explicit Git Bash runner
- `git diff --check`

The standard push gate stops before running this diff's checks on current `main` because `preflight_runner.py` references the unavailable Windows `signal.SIGHUP`; executable PR preflight also resolves the manifest's bare `bash` to the disabled WSL shim. #10574 carries those independent runner fixes. This branch was pushed with `--no-verify` after the checks above passed.

## Tests

The existing subprocess tests in `test_cli.py` and `test_memory_scenarios.py` now exercise the platform-native `PYTHONPATH` separator. The provider tests in `test_providers.py` retain the same relative-location contract using platform-native path components.

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups

Removed the stale `subprocess` import from `test_providers.py` in a separate commit after Ruff surfaced it.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10587?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10589, opened 2026-07-25
test(dev-harness): run resolver coverage on Windows

tianmind-studio:
## What changed and why

Run the dev-harness resolver and Makefile coverage through the actual shell/tool names available on each platform. On Windows, the tests now resolve Git Bash from the Git installation instead of invoking the WSL `bash.exe` shim, accept `mingw32-make` as GNU Make, and compare paths in Bash's path representation.

The double-quote checkout-name case remains covered on POSIX and is skipped only on Windows, where the filesystem does not permit `"` in a path component.

none

## How it was verified

- `python -m pytest scripts/dev-harness/tests/test_python_resolver.py -q -rs` (`4 passed, 1 skipped`, native Windows)
- complete native Windows dev-harness with #10583 and #10587 temporarily overlaid (`62 passed, 1 skipped`)
- Black 24.4.2 and Ruff 0.4.4 on `test_python_resolver.py`
- `git diff --check`
- nine applicable deterministic PR-manifest checks through the explicit preflight runner

The repository pre-push hook was invoked with the project Python available, but current `main` exits on Windows before running its selected checks because `signal.SIGHUP` is unavailable. The branch was therefore pushed with `--no-verify`; remote Linux CI is the authoritative full preflight.

Remote Linux CI completed with `58 passed` in the dev-harness suite and all 11 selected manifest checks passing.

## Tests

The existing resolver ordering, explicit `PYTHON`, checkout-with-spaces, and single-quote injection regressions now execute on Windows. Only the impossible Windows double-quote filename fixture is platform-gated.

## Failure class (fixes)

Not applicable; this PR contains no `fix:` commit.

### DIFF
--- a/scripts/dev-harness/tests/test_python_resolver.py
+++ b/scripts/dev-harness/tests/test_python_resolver.py
@@ -13,6 +13,57 @@ RESOLVER = HARNESS_ROOT / "_resolve_python.sh"
 MAKEFILE = HARNESS_ROOT.parents[1] / "Makefile"
 
 
+def _bash_command() -> str:
+    if os.name == "nt":
+        git = shutil.which("git")
+        if git:
+            git_root = Path(git).resolve().parent.parent
+            for relative_path in (Path("bin/bash.exe"), Path("usr/bin/bash.exe")):
+                candidate = git_root / relative_path
+                if candidate.is_file():
+                    return str(candidate)
+    else:
+        bash = shutil.which("bash")
+        if bash:
+            return bash
+    pytest.skip("Bash is required for dev-harness resolver tests")
+
+
+def _make_command() -> str:
+    for name in ("make", "mingw32-make"):
+        executable = shutil.which(name)
+        if executable:
+            return executable
+    pytest.skip("GNU Make is required for dev-harness Makefile tests")
+
+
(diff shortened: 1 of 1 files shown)


Git-on-my-level:
Thanks for tightening up the dev-harness resolver coverage here. I reviewed the diff and also ran the resolver test file on this head on Linux:

`python3 -m pytest scripts/dev-harness/tests/test_python_resolver.py -q -rs` → `5 passed`

The change looks narrowly scoped to the test harness: it resolves Git Bash explicitly on Windows instead of relying on a `bash` shim, allows `mingw32-make`, converts test paths into Bash path form before crossing into Git Bash, and skips only the double-quote checkout-name case that Windows cannot represent as a filename. I do not see a blocker in the test changes.

Because this is in dev-harness/Makefile execution coverage and the key value is native Windows behavior, I am leaving this as a positive signal rather than a formal approval. A maintainer with Windows CI/context should make the final merge call, but this looks directionally good to me.

---

kodjima33 (approved):
Chore: dev-harness test-only portability fix (resolve Git Bash / mingw32-make on Windows, compare paths in Bash representation). Test files only.

PR #10594, opened 2026-07-25
fix(ci): support Windows in manifest preflight

tianmind-studio:
## What changed and why

Treat native Windows as a supported check-manifest platform. `detect_platform()` already returned `windows`, but manifest validation rejected that value, so every Windows `scripts/pr-preflight` run failed its own contract test. The POSIX Firestore runner harness is now skipped on Windows, where `bash` resolves to the WSL launcher rather than the POSIX environment the test requires.

none

## How it was verified

- Reproduced both failures on unmodified `main`: `windows` was rejected, and the POSIX runner test launched Windows' WSL `bash.exe`.
- `python .github/scripts/test_run_checks.py`: 27 tests passed with the single POSIX-only case skipped.
- `scripts/pr-preflight --pr-body-file ...`: all 11 selected checks passed on `platform=windows`.

This exercises the real Windows preflight path end to end, not only the platform helper.

## Tests

`test_detect_platform_maps_windows` pins the supported platform mapping. The existing host-platform test now runs successfully on Windows, while the Firestore shell harness remains covered on POSIX CI.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/run_checks.py
+++ b/.github/scripts/run_checks.py
@@ -16,7 +16,7 @@ from dataclasses import dataclass
 from pathlib import Path, PurePath
 from typing import Any
 
-VALID_PLATFORMS = {"all", "macos", "linux"}
+VALID_PLATFORMS = {"all", "macos", "linux", "windows"}
 
 
 def detect_platform() -> str:
@@ -26,6 +26,8 @@ def detect_platform() -> str:
         return "macos"
     if system == "Linux":
         return "linux"
+    if system == "Windows":
+        return "windows"
     return system.lower()
 
 
--- a/.github/scripts/test_run_checks.py
+++ b/.github/scripts/test_run_checks.py
@@ -16,6 +16,7 @@ from collections import Counter
 from contextlib import redirect_stderr, redirect_stdout
 from io import StringIO
 from pathlib import Path
+from unittest.mock import patch
 
 from run_checks import (
     VALID_PLATFORMS,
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Thanks for the focused CI fix. I reviewed the manifest preflight changes statically and ran `.github/scripts/test_run_checks.py` on the PR head locally; it passed all 27 tests.

This looks like the right direction: `windows` is now a valid manifest platform, `detect_platform()` maps native Windows explicitly, the invalid-platform test still covers rejection with `plan9`, and the POSIX-only Firestore runner harness is skipped on Windows rather than trying to exercise a Unix shell through the WSL launcher.

Because this touches CI/preflight behavior, I’m not formally approving from automation; a maintainer should still do the final workflow review before merge. I don’t see a blocker in the patch as submitted.

---

kodjima33 (approved):
12-line CI manifest-preflight Windows fix. Approve-only — .github/ tooling stays out of agent merge authority.

kodjima33 (approved):
Verified: detect_platform() already returned 'windows' but VALID_PLATFORMS rejected it, so every Windows pr-preflight failed its own contract test; 12-line fix, no check is weakened.

PR #10595, opened 2026-07-25
fix(preflight): decode Git content as UTF-8 on Windows

tianmind-studio:
## What changed and why

Decode Git repository content explicitly as UTF-8 in three deterministic preflight guards instead of using the Windows host locale:

- deployment secret-boundary reads from `git ls-tree` and `git show`;
- deferred-work marker reads from `git diff` and `git show`;
- product file line-count ratchet reads baseline paths and both sharded and legacy JSON blobs from `git ls-tree` and `git show`.

Git repository paths and blobs are UTF-8 data, but Python's `text=True` used the active Windows code page (GBK in the reproduced environment). A valid Unicode source comment or baseline justification could therefore crash these guards before they enforced policy.

none

## How it was verified

- Before the deployment fix on native Windows, `python .github/scripts/check_deployment_secret_boundary.py --base HEAD` failed with `UnicodeDecodeError: 'gbk' codec can't decode byte ...`; the same production command passed after the fix.
- Before the deferred-marker fix, native Windows `scripts/pr-preflight` crashed while decoding a real Unicode TypeScript diff.
- Before the line-ratchet fix, native Windows preflight crashed in `baseline_at_ref()` while decoding a real baseline shard; the same production command now passes without `PYTHONUTF8`.
- `python .github/scripts/test_check_deployment_secret_boundary.py` -> 9 passed.
- `python .github/scripts/test_run_checks.py DeferredMarkerTests` -> 3 passed.
- `python .github/scripts/test_check_product_file_line_count_ratchet.py` -> 9 passed.
- Native Windows `scripts/pr-preflight --pr-body-file ...` in an integration worktree with #10594 applied -> all 15 selected checks passed without `PYTHONUTF8`.
- Black 26.5.1 with the repository's 120-column configuration, Ruff 0.4.4, and `git diff --check` passed on all changed Python files.

## Tests

The deployment fixture commits UTF-8 workflow content, exercising the real `git show` path. Mocked subprocess contracts require all seven Git content reads across the three guards to pass `encoding="utf-8"`; the line-ratchet contract covers both sharded and legacy baseline paths, so Linux CI also rejects a future return to locale-driven decoding.

## Failure class (fixes)

Failure-Class: none

Python inherited the Windows locale when `subprocess.run(..., text=True)` decoded Git output. All affected guards now declare strict UTF-8 decoding at the Git boundary, with production-path and configuration-contract regression coverage.

### DIFF
--- a/.github/scripts/check_deployment_secret_boundary.py
+++ b/.github/scripts/check_deployment_secret_boundary.py
@@ -18,7 +18,6 @@ from dataclasses import dataclass
 from pathlib import Path, PurePath
 from typing import Iterable
 
-
 ROOT = Path(__file__).resolve().parents[2]
 DEFAULT_POLICY = ROOT / "config" / "deployment-setting-classification.json"
 EXPRESSION = re.compile(r"\$\{\{\s*(secrets|vars)\.([A-Z][A-Z0-9_]*)\s*}}")
@@ -62,11 +61,7 @@ def _deployment_paths(root: Path) -> set[str]:
             paths.add(relative)
     charts = root / "backend" / "charts"
     if charts.is_dir():
-        paths.update(
-            repository_relative_path(path, root)
-            for path in charts.glob("*/*_values.yaml")
-            if path.is_file()
-        )
+        paths.update(repository_relative_path(path, root) for path in charts.glob("*/*_values.yaml") if path.is_file())
     return paths
 
 
@@ -80,7 +75,7 @@ def _base_paths(root: Path, base: str) -> set[str]:
         ["git", "ls-tree", "-r", "--name-only", base],
         cwd=root,
         check=True,
-        text=True,
+        encoding="utf-8",
         stdout=subprocess.PIPE,
         env=_git_environment(),
     )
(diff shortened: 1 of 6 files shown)


Git-on-my-level (commented):
Thanks for the focused Windows preflight fix. This looks like the right direction: the changed Git subprocess reads now decode repository content as UTF-8 explicitly, and the added tests cover the deployment boundary, deferred-work marker, and line-ratchet paths that were still relying on the host locale.

I also ran the touched tests in a stripped environment:
- `python .github/scripts/test_check_deployment_secret_boundary.py` -> 9 passed
- `python .github/scripts/test_check_product_file_line_count_ratchet.py` -> 9 passed
- `python .github/scripts/test_run_checks.py DeferredMarkerTests` -> 3 passed

No blocking code issues found from my review. Since this touches GitHub/preflight guard scripts that enforce security/workflow policy, I am leaving this as a positive maintainer signal rather than formal approval; a human maintainer should still sign off before merge.

---

kodjima33 (approved):
CI check-script portability fix (UTF-8 decode of Git content on Windows). Approve-only — .github/ tooling stays out of agent merge authority.

kodjima33 (approved):
Verified: text=True decoded git output with the Windows ANSI code page; the change is encoding='utf-8' on three preflight guards plus formatter reflow — no check is weakened, new regression test asserts the encoding kwarg.

PR #10596, opened 2026-07-25
SCA-127: discover all exact-SHA planner checks

Git-on-my-level:
## Summary
- query every page of exact-SHA check runs instead of using the suppressive `filter=latest` lookup
- select the latest matching run deterministically and fail closed on malformed API data
- cover gh command construction and slurped-response parsing in the planner regression suite

## Validation
- `python3 .github/scripts/run_checks.py --lane local --base origin/main --head HEAD --skip-pr-body-checks`
- exact incident SHA lookup returns `Release Eligibility`: `completed/success`

Closes SCA-127

Failure-Class: FC-release-build-config-source-binding

### DIFF
--- a/.github/scripts/plan-desktop-release.py
+++ b/.github/scripts/plan-desktop-release.py
@@ -126,9 +126,9 @@ def github_check_status(repository: str, sha: str, check_name: str) -> tuple[str
         [
             "gh",
             "api",
-            f"repos/{repository}/commits/{sha}/check-runs?filter=latest",
-            "--jq",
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Addressed by the updated fail-closed parser: a matching check run without `started_at` is rejected as invalid API data. `github_check_status` returns an error, so `required_source_checks_gate` stays `waiting` and ultimately blocks; it cannot fall back to an older success or create a candidate. The review fix also compares validated numeric IDs for timestamp ties.

PR #10599, opened 2026-07-25
fix(desktop): record an outcome on every terminal chat tool call

skanderkaroui:
## Problem

`chat_tool_call_completed` fired only on success. The call site removed the tool's start time for any terminal status, then gated the event on `toolStatus == .completed`, so a failed, cancelled, or interrupted tool call consumed its start time and emitted nothing.

Found by querying the live PostHog corpus (project 302298, Omi desktop macOS, 30 days): **23,682 `chat_tool_call_completed` events containing only successes**, and no tool-failure event anywhere in the project. Tool failure rate was not merely unmeasured but structurally unmeasurable — no dashboard could ever show a tool regression, and the `tool_stall` errors (p50 102s) had no per-tool attribution.

## Change

`chat_tool_call_completed` now fires for every terminal status and carries a bounded `outcome`.

The subtlety worth reviewing: `ToolCallStatus.fromBridgeStatus` maps `failed`, `cancelled`, and `interrupted` all to `.failed`. Emitting naively would have counted user Stops as tool failures — the same mistake that made the legacy `chat_agent_error` corpus unreadable, where 530 of 1,893 events over 30 days were user Stops recorded as errors. `ChatTelemetryDimension.toolOutcome` keeps the three apart and is deliberately independent of `ToolCallStatus`, with a test pinning that independence.

The bridge now preserves `cancelled` and `interrupted` from ACP and local subprocess adapters through the Swift callback. Previously ACP collapsed cancellation into failure and the subprocess adapter converted cancellation/interruption back to `started`, making the new outcome dimension inaccurate. Unknown future statuses are bounded to `unknown` rather than inflating successful completions.

**Event semantics change:** the event now counts all terminal calls, not just successes. Filter `outcome == "completed"` for the previous meaning. No in-repo consumer reads this event (`web/admin/lib/response-reliability.ts` reads `chat_agent_error` only), but an external PostHog dashboard counting it raw will see a superset.

## Product invariants

- **INV-AGENT-***: adapter lifecycle status is preserved through the existing outbound protocol; canonical cancelled/interrupted events remain failure-terminal in the runtime journal.
- **INV-CHAT-1**: path-matched via `Providers/ChatProvider.swift`. Telemetry call site only — no chat write-path, projection, or timeline behavior changed.
- **INV-AUTH-1**: path-matched. No auth behavior changed.

Failure-Class: none

## Verification

- `npm test -- --run tests/acp-tool-activity.test.ts tests/local-subprocess-adapter.test.ts` — 13/13 pass
- `npm run build` — TypeScript green
- `xcrun swift test --package-path Desktop --filter ChatQueryTelemetryTests` — 27/27 pass
- Failure mode pinned at both adapter boundaries and at the telemetry mapping.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10599?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-providers.json
@@ -1,6 +1,6 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6269,
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift": 6268,
     "desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift": 2861
   },
   "raise_justifications": {
--- a/desktop/macos/Desktop/Sources/AnalyticsManager.swift
+++ b/desktop/macos/Desktop/Sources/AnalyticsManager.swift
@@ -713,11 +713,16 @@ class AnalyticsManager {
     String(value.trimmingCharacters(in: .whitespacesAndNewlines).prefix(128))
   }
 
-  /// Track individual tool calls made by the Claude agent
-  func chatToolCallCompleted(toolName: String, durationMs: Int) {
+  /// Track individual tool calls made by the Claude agent.
+  ///
+  /// Fires for every terminal status, not only success — a failed tool call
+  /// previously emitted nothing, so tool reliability was unmeasurable. Filter
+  /// on `outcome == "completed"` for the pre-existing success-only meaning.
+  func chatToolCallCompleted(toolName: String, durationMs: Int, outcome: String) {
     let props: [String: Any] = [
       "tool_name": ChatTelemetryDimension.toolName(toolName),
       "duration_ms": durationMs,
+      "outcome": ChatTelemetryDimension.toolOutcome(outcome),
     ]
     PostHogManager.shared.track("chat_tool_call_completed", properties: props)
   }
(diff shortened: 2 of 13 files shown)


Git-on-my-level (commented):
Maintainer automation review: positive signal, not a formal approval because this touches desktop agent/runtime telemetry surfaces that need human maintainer sign-off.

What looks good:
- The change fixes the original measurement gap by emitting `chat_tool_call_completed` for every terminal tool-call outcome, while preserving `completed`, `failed`, `cancelled`, and `interrupted` as separate bounded values.
- The Swift telemetry mapping avoids counting unknown statuses as successful completions, and the tests pin the user-Stop-vs-failure distinction.
- The ACP and local subprocess bridge changes preserve cancellation/interruption through the runtime boundary, so the Swift-side `outcome` dimension should receive the intended terminal status instead of the prior collapsed/started values.
- I ran the two affected agent adapter test files and the TypeScript build locally: 13/13 tests passed and `npm run build` succeeded. CI also shows the desktop Swift and agent-runtime checks passing.

No blocking issue found from the static/code review. A human maintainer should still review the telemetry semantics and any external dashboard impact before merge, since the event now counts all terminal calls rather than success-only calls.

---

kodjima33 (approved):
Desktop bug fix: chat_tool_call_completed only fired on success, making tool failures structurally unmeasurable; now emits on every terminal status with a bounded outcome dimension, cancelled/interrupted preserved through the bridge. Tests green.

PR #10600, opened 2026-07-25
fix(agent-proxy): stop persisting an unresolved VM IP as a ready address

skanderkaroui:
## Problem

When the GCE external-IP poll timed out, `_start_vm_and_wait` returned the literal string `"unknown"` and the caller wrote it to Firestore alongside `status: "ready"`.

`"unknown"` is truthy, so the poisoned record satisfied every reader:

1. connect takes the fast path (`status == "ready" and vm_ip`)
2. the health probe dials `http://unknown:8080/health` and fails
3. `_ensure_vm_running(health_failed=True)` finds GCE reporting `RUNNING`
4. so it **hard-resets a healthy VM**, then writes `vm.get("ip")` — still `"unknown"` — back with `status: "ready"`

No writer anywhere repaired the field, so every later reconnect reset the user's VM again and never recovered the real address. The field was already known to be dangerous: both copies guard `candidate != "unknown"` when *reading* the GCE response, but nothing guarded the *write*.

## Change

Repairs the ownership boundary rather than the call site, per the failure-class guidance:

- `_update_firestore_vm` owns `agentVm.ip`, so it now refuses any value that is not dialable. A future caller cannot reintroduce the placeholder.
- The poll failure raises instead of returning a sentinel, which reuses the callers' existing `except -> status: "error"` path rather than adding a second one.
- **Already-poisoned records self-heal:** the ready fast path and both vm-ensure `RUNNING` branches require a usable IP. `_start_vm_and_wait` first checks whether GCE is already running; if so it resolves the external IP without issuing a second start request.

Both copies are fixed (`agent-proxy/main.py` and `routers/agent_tools.py`); they were byte-identical.

## Product invariants

None matched by these paths.

Failure-Class: none

## Verification

- `pytest tests/unit/test_agent_proxy_async_boundaries.py tests/unit/test_agent_vm_async.py tests/unit/test_tools_agent_route_response_shape.py tests/unit/test_agent_tool_geolocation_async.py` — 23 passed
- Both new tests confirmed to **fail** on the unfixed source: `'unknown' assigned to ['ip'] at line 213`, plus the writer-rejection `ValueError` case
- The GCE-timeout path itself is not reachable in a hermetic test; it is exercised only against live GCE, so it is covered here by the boundary guard plus the static assertion that the literal is only ever bound to the named constant

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10600?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/agent-proxy/main.py
+++ b/backend/agent-proxy/main.py
@@ -161,38 +161,43 @@ async def _start_vm_and_wait(vm_name: str, zone: str) -> str:
 
     t0 = time.monotonic()
     token = await run_blocking(critical_executor, _get_gce_access_token)
+    instance_url = f"https://compute.googleapis.com/compute/v1/projects/{GCE_PROJECT}/zones/{zone}/instances/{vm_name}"
     start_url = (
         f"https://compute.googleapis.com/compute/v1/projects/{GCE_PROJECT}/zones/{zone}/instances/{vm_name}/start"
     )
 
     async with httpx.AsyncClient(timeout=180) as client:
-        resp = await client.post(start_url, headers={"Authorization": f"Bearer {token}"}, content=b"")
-        if resp.status_code not in (200, 204):
-            raise Exception(f"GCE start failed: {resp.status_code} {resp.text}")
-        t_start = time.monotonic() - t0
-        logger.info(f"[vm-start] {vm_name} start API call: {t_start:.1f}s")
+        instance_resp = await client.get(instance_url, headers={"Authorization": f"Bearer {token}"})
+        already_running = instance_resp.status_code == 200 and instance_resp.json().get("status") == "RUNNING"
+        if already_running:
+            logger.info(f"[vm-start] {vm_name} is already RUNNING; resolving its external IP")
+        else:
+            resp = await client.post(start_url, headers={"Authorization": f"Bearer {token}"}, content=b"")
+            if resp.status_code not in (200, 204):
+                raise Exception(f"GCE start failed: {resp.status_code} {resp.text}")
+            t_start = time.monotonic() - t0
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Nice fix — this addresses a genuinely nasty production loop cleanly and at the right boundary.

The root cause (truthy sentinel `'unknown'` persisted as a ready address, causing permanent VM reset loops with no self-healing) is well understood and the repair is correct: making `_update_firestore_vm` the single owner that refuses non-dialable IPs, raising instead of returning a sentinel, and adding `_is_usable_vm_ip` gates on every ready/RUNNING fast path so already-poisoned records recover on the next connect. The RUNNING-before-start check is a good efficiency win too.

Tests look solid — the behavioral test mocks httpx and exercises `_start_vm_and_wait` directly, and the static AST tripwire is correctly labeled as a static checker. Both copies are fixed identically.

Leaving a positive signal rather than formal approval since this touches Firestore user-data persistence (`agentVm` fields). Flagging for maintainer sign-off before merge.

_(automated maintainer review — glm-5.2)_

---

kodjima33 (approved):
Backend bug fix: agent-proxy persisted the literal 'unknown' as a ready VM IP, causing repeated hard-resets of healthy VMs; write path now rejects non-dialable IPs and poisoned records self-heal. Tests added.

PR #10601, opened 2026-07-25
fix(backend): keep execute_tool off the event loop

skanderkaroui:
## Problem

`POST /v1/agent/execute-tool` is `async def`, but ran two blocking calls directly on the event loop:

- `load_app_tools(uid)` — a Redis read plus one Firestore document read **per enabled app**
- `target.invoke(params, config=config)` — every `CORE_TOOLS` entry is a sync `@tool` that fans out to Firestore/Pinecone, and `get_conversations_tool` accepts `limit` up to 5000

One agent VM calling this endpoint parked the whole backend event loop for the duration of that scan: every other connection, `/health`, and the HPA metrics endpoint froze with it. This is the "silent poison" class `backend/AGENTS.md` → Common Gotchas #3 describes.

The canonical path one file over already does both correctly — `utils/retrieval/agentic.py:707` (`await run_blocking(db_executor, load_app_tools, uid)`) and `:331` (`await tool_obj.ainvoke(...)`). This router was the drift.

Not caught by `scan_async_blockers.py`: its DB detector is import-aware over `database.*` names, and neither `load_app_tools` nor `t.invoke` is one.

## Change

Both calls now go through `run_blocking(db_executor, ...)`, which copies the current context so `agent_config_context` still resolves inside the worker thread (verified: `utils/executors.py:89` uses `contextvars.copy_context()`).

Also stops echoing raw exception text. Pydantic renders `input_value=<...>`, so a `ValidationError` from caller-supplied params wrote user content verbatim into backend logs and returned it over the wire under HTTP 200. The log line now carries `error_type` only and the response returns a fixed `Tool execution failed` category. Sanitizing free-form text was insufficient because ordinary user prose contains no token pattern to redact.

## Product invariants

None matched by these paths.

Failure-Class: none

## Verification

- `pytest tests/unit/test_tools_agent_route_response_shape.py tests/unit/test_agent_vm_async.py tests/unit/test_agent_tool_geolocation_async.py tests/unit/test_chat_memory_tool_caller.py` — 18 passed
- The new test is **behavioral, not a static tripwire**: a sync tool blocks inside `invoke` while the test coroutine tries to regain control. Confirmed to fail on the unfixed router with `assert 3.006059459061362 < 1.0`; the whole file runs in 0.08s with the fix.
- Note the shared fixture stubs `run_blocking` to call `fn()` inline, which would erase the seam under test, so this case restores a real `ThreadPoolExecutor` offload including the contextvar copy.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10601?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/agent_tools.py
+++ b/backend/routers/agent_tools.py
@@ -302,13 +302,15 @@ async def execute_tool(
     }
     agent_config_context.set(config)
 
-    # Find the tool
+    # Find the tool. `load_app_tools` reads Redis plus one Firestore document
+    # per enabled app, so it must not run on the event loop — see the canonical
+    # path in utils/retrieval/agentic.py.
     all_tools = list(CORE_TOOLS)
     try:
-        app_tools = load_app_tools(uid)
+        app_tools = await run_blocking(db_executor, load_app_tools, uid)
         all_tools.extend(app_tools)
-    except Exception as e:
-        logger.error(f"⚠️ Error loading app tools: {e}")
+    except Exception as error:
+        logger.error("⚠️ Error loading app tools error_type=%s", type(error).__name__)
 
     target = None
     for t in all_tools:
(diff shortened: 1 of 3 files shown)


Git-on-my-level (approved):
Reviewed and approved. This correctly moves the two blocking calls in POST /v1/agent/execute-tool (load_app_tools and target.invoke) onto db_executor via run_blocking, matching the canonical path already used in utils/retrieval/agentic.py. The contextvar copy in run_blocking keeps agent_config_context resolving in the worker thread, and app tools with a coroutine still take the async path (no double-offload).

The exception-text hardening is a good catch: pydantic renders input_value=..., so a ValidationError could previously echo caller-supplied content into logs and the HTTP 200 body. Logging error_type only and returning a fixed category is the right call.

Verified: synced the backend venv and ran the touched suites locally — test_tools_agent_route_response_shape.py (incl. the new event-loop-blocking and exception-sanitization cases) and test_chat_memory_tool_caller.py all pass, plus test_agent_vm_async.py and test_agent_tool_geolocation_async.py (18 total).

---

kodjima33 (approved):
Backend bug fix: /v1/agent/execute-tool ran blocking Firestore/Pinecone work on the event loop; now routed through run_blocking like the canonical agentic.py path, with a behavioral regression test.

PR #10572, opened 2026-07-25
fix(firmware): make CV1 capture durable and 3.2× faster on iPhone

axAilotl:
## Summary

- preserve every accepted Opus frame across transient BLE, SD queue, remount, and disconnect failures
- make the SD ring authoritative before live GATT enqueue, retaining a recoverable copy until an explicit durable app `ADVANCE`
- keep the live and storage paths on the same stable record identity so replay is duplicate-safe
- reject unsafe live-only delivery when durable storage is unavailable
- add CRC-covered ring completion, bounded bulk-transfer link policy, right-sized dual-core ACL buffers, and a one-shot 15 ms interval request during contiguous backlog drains
- bind the CV1 DIS and MCUboot app-core versions and reject dual-core OTA version drift in the release gate

## Root cause

The live path treated a successful `bt_gatt_notify_cb` enqueue as ownership transfer. The controller can accept a frame and then lose the RF link or app before durable receipt, so `transport.c` skipped storage and permanently dropped the only recoverable copy.

The corrected ownership rule is: the pendant SD/ring owns the record until the app explicitly advances the durable watermark. GATT completion is flow control, not durable acknowledgment.

## Physical findings

Tested on a physical CV1 and iPhone 13 Pro Max running iOS 26.5.2:

- stock durable backlog baseline: 10,166,673 bytes in 622 s = 15.96 KiB/s
- candidate steady window: 18,162,224 bytes in 348 s = 50.97 KiB/s (3.19x)
- audited 56 consecutive durable ranges beginning at sequence 125,361 with zero gaps, overlaps, or duplicate source IDs; subsequent metadata confirmed 224,361 → 226,161
- Bluetooth off/on, app termination/relaunch, and a complete MCUboot OTA all resumed at the exact unfinished or next durable sequence
- the first range after OTA was exactly 202,761 → 204,561; the app reclaimed the pendant without re-pairing or a hot restart

## SCA-144 regression

Head `9c052f5ffa22f9ef9c9bcdd55fdf7064837da0c2` adds host regressions proving:

- a live enqueue accepted immediately before disconnect remains recoverable from storage
- a later durable ACK reclaims the record idempotently without duplicate ownership
- storage unavailability cannot silently degrade an accepted live enqueue into live-only loss

## Companion PRs

- Android/shared durable sync: https://github.com/BasedHardware/omi/pull/10571
- iOS BLE/DFU recovery: https://github.com/BasedHardware/omi/pull/10573

## Validation

[...]

PR #10604, opened 2026-07-25
feat(firmware): add idle sleep and BLE device controls

undivisible:
## Summary

- add BLE-controlled sleep and persistent device naming, both behind an encrypted link
- ship idle auto-sleep and the capture / low-battery LED patterns as opt-in Kconfig, disabled in `omi.conf`
- ref-count `current_connection` behind `get_current_connection()` across all call sites
- expose validated sleep and device-name controls to the mobile transport layer

## Review round: changes requested

- BLE sleep (19B10014), capture-state (19B10015), and device-name (19B10016) now use `BT_GATT_PERM_*_ENCRYPT`; an unpaired central can no longer power off or rename the pendant.
- Idle auto-sleep no longer decides from `charging || audio_subscribed` alone. The policy moved to `omi/firmware/omi/src/lib/core/idle_sleep.h`: a storage sync transfer is activity, and any build that can capture offline never auto-sleeps, because the firmware cannot distinguish an idle pendant from one recording to SD. `OMI_ENABLE_IDLE_SLEEP` now `depends on !OMI_ENABLE_OFFLINE_STORAGE`.
- Idle auto-sleep and both new LED patterns are disabled in `omi.conf`, so the shipping build's LED language and power behavior are unchanged; the LED language is documented next to `set_led_state()`.
- Capture flag is cleared on disconnect (no link means no audio subscription).
- The device name is back in the primary advertisement (shortened when it does not fit the 31-byte payload) with the complete name in scan response, so passive scanners still see a name.

## Dependency

Depends on #10572. Keep this PR targeting `main`; merge only after #10572 lands.

## Validation

- `bash omi/firmware/omi/tests/run_idle_sleep_test.sh` -> `idle-sleep policy tests passed`; mutation-verified (dropping the offline-capture guard fails 2 assertions).
- Registered as `firmware-idle-sleep-tests` in `.github/checks-manifest.yaml` (local + ci lanes).
- `make preflight` run locally on this head.
- Earlier rounds: mobile analyzer/tests and Android compile smoke, NCS 2.9.0 signed dual-image CV1 build, BabbleSim run 30183037698, real CV1 HW5 OTA/rename/capture/reset checks.

Not verified on this head: no NCS toolchain or CV1 hardware was available for this review round, so the firmware was not compiled or flashed here. Physical system-off/wake, encrypted-link pairing behavior for the new characteristics, charger readings, and LED electrical output need hands-on maintainer QA.

Failure-Class: none


Git-on-my-level (dismissed):
Thanks for pushing this forward. The battery/power-control direction is useful, and the PR already includes good validation notes plus mobile coverage for the new transport methods.

I do see one blocker before this should move out of draft / toward merge:

- The new BLE control characteristics use plain `BT_GATT_PERM_WRITE` / `BT_GATT_PERM_READ | BT_GATT_PERM_WRITE` for sleep, capture state, and persisted device name. With the current permissions, a nearby BLE client that can connect can issue a system-off command or persistently rename the device without an encrypted/authenticated link. For sleep in particular, that is a denial-of-service / UX risk because waking requires physical button access. Please require an appropriate protected link for these control writes (for example encrypted/authenticated GATT permissions consistent with the rest of Omi's pairing model, or another maintainer-approved authorization gate) before enabling these controls by default.

A couple of non-blocking notes for the maintainer pass:

- This also touches `.github/scripts/check-release-process-guards.py`; the narrow guard update looks aligned with the isolated tag checkout described in the PR, and the new focused guard test passes locally.
- Because this is a draft firmware behavior change depending on #10572, and physical system-off/wake / charger / LED behavior still needs hands-on validation, it should stay under human maintainer review before merge.

Local validation I ran:

- `python -m pytest .github/scripts/test_check_release_process_guards.py::test_desktop_qualification_runner_accepts_run_isolated_tag_checkout -q` passed.
- The full release-guard pytest file has pre-existing parent-fixture failures in this shallow review checkout, so I did not treat that as a PR regression.
- Flutter/Dart tooling was not installed in this review environment, so I could not rerun the new app test locally.

---

Git-on-my-level (dismissed):
Thanks for the update. I re-reviewed the current head and the main blocker from the earlier review is still present:

- The new BLE sleep, capture-state, and persisted device-name characteristics are still registered with plain `BT_GATT_PERM_WRITE` / `BT_GATT_PERM_READ | BT_GATT_PERM_WRITE`. Because these controls are enabled by default, a connected nearby BLE client could still issue system-off or persistently rename the device without a protected link or other maintainer-approved authorization gate. Sleep is especially sensitive because recovery requires physical button access.

Please require an appropriate protected/encrypted/authenticated link or another explicit authorization gate for these control writes before this moves toward merge.

Keeping this as a requested change on the current head. Separate from that blocker, this should remain under human maintainer/security review because it is still a draft firmware power/control behavior change, depends on #10572, touches workflow-adjacent release-guard coverage, and the physical system-off/wake, charger, and LED behavior still need hands-on validation.

---

Git-on-my-level (commented):
Thanks for the follow-up update. I re-reviewed the current head and the earlier blocking BLE-permissions issue looks addressed: the sleep, capture-state, and persisted device-name characteristics are now registered with encrypted read/write permissions where applicable.

I’m not approving this automatically because this is still a firmware power/control behavior change that depends on #10572, the PR notes that physical system-off/wake, idle-timeout duration, charger readings, and LED electrical output still need hands-on observation, and the Desktop Swift CI checks are currently failing. This should stay under human maintainer/security review before merge, but from the automation pass the prior BLE authorization blocker no longer appears to be blocking on this head.

---

Git-on-my-level (dismissed):
Thanks for continuing to tighten this up. The earlier BLE control-permission blocker appears addressed on the current head: the sleep, capture-state, and device-name controls now use encrypted GATT permissions where applicable.

I found a new blocking issue before this can move forward:

- `omi/firmware/omi/src/lib/core/transport.c` still registers `audio_ccc_config_changed_handler` in the audio service CCC entries, but on this head the function is only declared and no longer defined. The base branch has the definition; this head has only the prototype/usages. With `CONFIG_OMI_ENABLE_CAPTURE_LED=y` also enabled, `transport.c` additionally declares/calls `update_advertising_name()` from the device-name work handler, but that helper is not defined either. These missing definitions should fail the firmware build/link path for the enabled configuration, so please restore/implement them before merge.

Separate from that blocker, this should remain under human maintainer/security review because it changes firmware power/control behavior, depends on #10572, and the PR still notes hands-on validation gaps for physical system-off/wake, idle timeout, charger readings, and LED output.

---

Git-on-my-level (commented):
Thanks for continuing to iterate on this.

I re-reviewed the current head. The prior automation blocker about missing `audio_ccc_config_changed_handler` / `update_advertising_name()` definitions appears resolved on this head: both helpers are now present in `omi/firmware/omi/src/lib/core/transport.c`, and the BLE sleep / capture-state / device-name controls are registered with encrypted permissions where applicable.

I’m not approving this automatically because this remains a firmware power/control change with security-sensitive BLE controls, it still depends on #10572 being merged first, and the PR’s validation notes still leave physical system-off/wake, idle-timeout duration, charger readings, and LED electrical output for hands-on maintainer validation. Please keep this under human maintainer/security review before merge.

---

Git-on-my-level (dismissed):
Thanks for continuing to iterate here.

I found one current-head blocker that should be fixed before this moves forward:

- `update_idle_sleep()` treats the device as active only when it is charging, connected with audio notifications subscribed, or in a connected storage transfer. With offline storage enabled in `omi.conf`, a disconnected pendant that is actively capturing/writing to SD still has `is_connected == false` and will accumulate idle time. After the timeout, the main loop can call `turnoff_all()` and stop a long offline recording session unexpectedly. Please either gate auto-sleep on an explicit “not recording / no offline capture active” state, or keep the feature disabled until firmware can distinguish real idle from disconnected offline recording.

The earlier BLE permission and missing-helper blockers look addressed on this head, but this remains firmware power/control behavior with security-sensitive BLE controls, an open dependency on #10572, and hands-on validation gaps for physical system-off/wake, idle timeout, charger readings, and LED output, so it still needs human maintainer/security review before merge.

---

Git-on-my-level (commented):
Thanks for the follow-up changes here.

The previous automation blocker around idle sleep interrupting disconnected offline recording looks resolved on this head: `update_idle_sleep()` now also holds the device awake when offline storage is enabled and the SD/mic path indicates active capture, in addition to connected storage transfer and subscribed audio.

I’m not approving because this is still firmware power/control behavior with encrypted BLE device controls, an open dependency on #10572, and remaining hands-on validation items for physical system-off/wake, idle timeout, charger readings, and LED output. Human maintainer/security review should still sign off before merge.

---

Git-on-my-level (commented):
Thanks for the latest fixes here.

I re-reviewed the current head. The earlier automation blockers still look addressed: the sensitive sleep / capture-state / device-name controls use encrypted GATT permissions, the missing helper definitions are present, and idle sleep now keeps active offline capture/storage awake. The newest head also preserves the low-battery LED unset guard, advertises the runtime/persisted BLE name rather than the build-time default, and uses the audio subscription check so an early BLE link is treated as active for idle-sleep purposes.

I do not see a new automation-level blocker on this pass. I’m still not approving automatically because this remains firmware power/control behavior with security-sensitive BLE controls, it depends on #10572, and the PR’s own validation notes still leave physical system-off/wake, idle-timeout duration, charger readings, and LED electrical output for hands-on maintainer validation. Please keep this under human maintainer/security review before merge.

---

Git-on-my-level (dismissed):
Thanks for the continued iteration here. I re-reviewed the current head and found two current blockers that should be fixed before this moves forward:

- The BLE sleep, capture-state, and persisted device-name characteristics are currently registered with plain read/write permissions (`BT_GATT_PERM_WRITE` / `BT_GATT_PERM_READ | BT_GATT_PERM_WRITE`). These are sensitive controls: a connected nearby client could issue system-off or persistently rename the device without a protected link. Please restore an appropriate encrypted/authenticated-link requirement or another explicit maintainer-approved authorization gate for these writes/reads where applicable.
- `update_idle_sleep()` now only treats charging or a connected audio subscription as active. With offline storage enabled, a disconnected pendant that is actively capturing/writing to storage can still accumulate idle time and call `turnoff_all()` during a long offline recording. Please keep auto-sleep gated on an explicit “not recording / no offline capture active” state, or leave it disabled until the firmware can reliably distinguish real idle from disconnected offline capture.

I’m keeping this as human maintainer/security review rather than approval because it changes firmware power/control behavior and the PR itself still notes physical system-off/wake, idle-timeout duration, charger readings, and LED output need hands-on validation.

---
*by AI on behalf of David — if you need David’s attention urgently, please @Git-on-my-level and escalate with a concise note.*

omi/firmware/omi/omi.conf
```diff
@@ -27,8 +27,8 @@ CONFIG_RING_BUFFER=y
 CONFIG_BT=y
 CONFIG_BT_SMP=y
 CONFIG_BT_PERIPHERAL=y
-CONFIG_BT_DIS=n
-CONFIG_BT_BAS=n
+# CONFIG_BT_DIS=n # configurated bellow (enabled for firmware/hardware revision)
+# CONFIG_BT_BAS=n # configurated bellow (enabled for battery service)
```
TuEmb: I think we can remove those lines instead of comments

omi/firmware/omi/src/main.c
```diff
@@ -166,15 +169,51 @@ void set_led_state()
...
 }
 
+#ifdef CONFIG_OMI_ENABLE_IDLE_SLEEP
+#define IDLE_SLEEP_TIMEOUT_SEC (CONFIG_OMI_IDLE_SLEEP_TIMEOUT_MIN * 60)
+static uint32_t idle_seconds = 0;
+
+static void update_idle_sleep(void)
+{
+    // Streaming audio to a subscribed central, or charging, holds the device awake.
+    bool active = is_charging || (is_connected && transport_is_audio_subscribed());
```
TuEmb: I don't think that's enough. In the case where the device is connected but not streaming audio and is syncing audio files from storage instead, we still need to keep the device powered on during the entire sync process.

omi/firmware/omi/src/main.c
```diff
@@ -166,15 +169,51 @@ void set_led_state()
...
+#ifdef CONFIG_OMI_ENABLE_CAPTURE_LED
+        // Blue while a central is actively capturing (audio subscribed), red otherwise
+        blue = is_connected && is_capturing;
+        red = !blue;
+#else
         blue = is_connected;
         red = !is_connected;
+#endif
+#if defined(CONFIG_OMI_ENABLE_BATTERY) && defined(CONFIG_OMI_ENABLE_BATTERY_LOW_LED)
+        if (battery_percentage > 0 && battery_percentage <= CONFIG_OMI_BATTERY_LOW_THRESHOLD) {
```
TuEmb, line 194: The blinking red LED pattern is currently used to indicate a time sync error. I think this change could confuse users, as the same LED pattern would represent two different states.

omi/firmware/omi/Kconfig
```diff
@@ -130,4 +130,72 @@ config OMI_WATCHDOG_TIMEOUT_MS
...
+    int "Low-battery LED threshold (percent)"
+    range 1 50
+    default 10
+    depends on OMI_ENABLE_BATTERY_LOW_LED
+    help
+        "Battery percentage at or below which the low-battery LED warning
+         activates."
+
+config OMI_ENABLE_CAPTURE_LED
+    bool "Capture-state LED (blue capturing, red idle)"
```
TuEmb, line 188: Red also indicates that the device is disconnected and recording offline to its local storage. We shouldn't have two different states using the same LED behavior, as this could confuse users.

TuEmb (changes requested):
### FIRMWARE REVIEWS ONLY

This is just my first impression. I haven't tested this PR on the actual device yet.

**Strengths**:

- Ref-count refactor is complete and balanced. All 14 call sites verified; no leaks, no double-unref; transport_off correctly transfers ownership out under the lock rather than borrowing.
- Writable/persisted device name — rename via 19B10016, stored in NVS, re-applied on boot without needing CONFIG_BT_SETTINGS.

**Concerns:**

- The LED patterns have changed quite a lot. I think we should review them again to make sure the LED behaviors are consistent and don't confuse users.
- `Idle timeout before auto-sleep` --> This feature is useful for saving power, but if users are not aware of it, they may think there is an issue with the device, and it could also lead to data loss. For example, if a user is in a meeting with a 15-minute break, the device may automatically go to sleep during the break. When the meeting resumes, the user may not realize that the device has turned off and may assume it is still recording. 

**FYI:** While BLE is connected with no audio streaming or file syncing, the current draw is only ~2 mA. With a 150 mAh battery, this means the device can theoretically last for around 75 hours in this state.

undivisible:
Thanks — both concerns addressed on `e8d006b8b0`.

**LED patterns:** shipping LED behaviour is now identical to before this PR. `CONFIG_OMI_ENABLE_CAPTURE_LED` and `CONFIG_OMI_ENABLE_BATTERY_LOW_LED` are off in `omi.conf`; the full LED language is documented above `set_led_state()` in `src/main.c` so it can be reviewed as a whole before anything changes.

**Idle timeout:** disabled in `omi.conf`, and `OMI_ENABLE_IDLE_SLEEP` now `depends on !OMI_ENABLE_OFFLINE_STORAGE` — a disconnected pendant can be mid-recording and the firmware cannot tell that from idle, so the combination is no longer buildable. Your meeting-break case is exactly why; it should not ship until the app warns the user first. The policy also moved to a Zephyr-free header with a host test (`omi/firmware/omi/tests/run_idle_sleep_test.sh`, wired into CI).

Also in this push: the sleep / capture-state / device-name characteristics now require an encrypted link, the capture flag is cleared on disconnect, and the device name is back in the primary advertisement (shortened when it doesn't fit) with the complete name in scan response.

Not verified here: no NCS toolchain or CV1 available, so this head was not compiled or flashed. Needs on-device QA for pairing behaviour on the three encrypted characteristics, rename + advertisement/scan-response visibility, and physical system-off/wake.

Git-on-my-level:
Thanks for the latest update — I re-reviewed the current head and the automation blockers I raised on the previous head look addressed.

- `omi/firmware/omi/src/lib/core/transport.c`: the sleep command (`19B10014`), capture-state (`19B10015`), and device-name (`19B10016`) characteristics are now registered with encrypted GATT permissions where enabled, and the write handlers validate offset/prepared writes/lengths before acting. The connection refactor also uses `get_current_connection()` to return a ref-counted handle and unreferences it at the updated call sites I sampled.
- `omi/firmware/omi/src/lib/core/idle_sleep.h` and `omi/firmware/omi/src/main.c`: the idle policy now treats charging, storage transfer, connected audio subscription, and offline-capture-capable builds as busy; `update_idle_sleep()` wires that policy in only behind `CONFIG_OMI_ENABLE_IDLE_SLEEP`.
- `omi/firmware/omi/Kconfig` and `omi/firmware/omi/omi.conf`: idle auto-sleep depends on `!OMI_ENABLE_OFFLINE_STORAGE`, and the shipping config keeps idle sleep plus the new capture/low-battery LED patterns disabled while enabling only the encrypted app-driven sleep/name controls.
- `omi/firmware/omi/src/settings.c` / `settings.h`: persisted device naming is bounded by `CONFIG_BT_DEVICE_NAME_MAX`, rejects empty values when loading/saving, null-terminates the in-memory cache, and is re-applied during transport startup.
- `app/lib/services/devices/connectors/omi_connection.dart` and `app/test/services/omi_device_controls_test.dart`: the mobile side writes the expected sleep/name characteristic UUIDs and rejects blank, over-25-byte, or NUL-containing names before sending them.
- `.github/checks-manifest.yaml` and `omi/firmware/omi/tests/run_idle_sleep_test.sh` / `test_idle_sleep.c`: the host-verifiable idle-sleep policy test is registered in the manifest and covers the offline-capture guard that was previously blocking.
- `.github/scripts/test_check_release_process_guards.py`: this diff only adds coverage for the desktop qualification runner’s isolated tag checkout behavior; it still keeps this PR in workflow-review territory because it touches release/check guard coverage.

I’m not approving this automatically because this is still firmware power-off / BLE identity-control behavior, has workflow/check-manifest changes, carries a rebase/blocking state, and needs maintainer/product sign-off on the final shipped UX for sleep/LED behavior. Leaving this for human maintainer review for firmware power-control and workflow sign-off before merge.

---

undivisible:
Thanks for the firmware review. Both concerns are addressed on the current head:

**LED patterns** — the capture and low-battery LED patterns are now opt-in Kconfig, disabled in `omi.conf`, so shipping LED behavior is identical to before this PR. The review-fix commit `3f44b12836` ("keep idle auto-sleep and new LED patterns off by default") made this explicit.

**Idle auto-sleep data loss** — two guards:
- Auto-sleep is disabled by default (`omi.conf`) and only reachable through an explicit BLE command behind an encrypted link (`f8b7a43da4` — the sleep, capture-state, and device-name characteristics now require `BT_GATT_PERM_WRITE_ENCRYPT` / `BT_GATT_PERM_READ_ENCRYPT`).
- `update_idle_sleep()` never powers off while `offline_capture_possible` is true (SD capture without a BLE link), nor during storage transfer or subscribed audio (`4d4c7ee2d1`), with regression coverage in `test_idle_sleep.c`.

Also merged `main` on this head (the branch was 330 commits behind). CI is re-running.

undivisible:
Addressed the four review points (all already resolved or fixed on the current head):

1. **Remove commented BT config lines** — done in 7989e42c7d. The commented `CONFIG_BT_DIS/BAS/DEVICE_NAME/APPEARANCE` stubs above the real config block are now removed; the effective assignments (omi.conf:86-108) are unchanged.

2. **Keep the device powered during a storage sync while connected but not streaming audio** — the idle policy (`omi_idle_is_busy`) already treats `storage_transfer_active` as busy, so a file sync holds the device awake. Verified at `omi/firmware/omi/src/lib/core/idle_sleep.h` and `update_idle_sleep` in main.c.

3. **Blinking red reused for low battery conflicts with the time-sync blink** — resolved by 3f44b12836: `CONFIG_OMI_ENABLE_CAPTURE_LED` and the low-battery LED branch are now **off by default**, so the shipped build keeps the existing LED behavior (time-sync blink red, red = not connected). The new patterns only activate when a board opts in.

4. **Red overlapping with 'disconnected + offline recording'** — same off-by-default resolution: with `CONFIG_OMI_ENABLE_CAPTURE_LED` unset, the capture-state LED never runs. Additionally, offline-storage builds never auto-sleep (4d4c7ee2d1) since the firmware can't tell idle from mid-recording.

I have not run this on hardware; the LED config changes are config-gated so the default firmware path is unchanged.

omi/firmware/omi/src/main.c
undivisible: Changed the opt-in low-battery warning to a distinct double-blink red sequence; the existing single red/blue time-sync phase and solid-red disconnected state remain unchanged.

omi/firmware/omi/Kconfig
undivisible: The opt-in low-battery behavior now uses the same distinct double-blink sequence, so it no longer shares the solid-red disconnected or time-sync pattern.

Git-on-my-level:
Thanks for the latest round on this head — re-reviewed `11aa7b5687` and the new commits look right on everything I can verify without hardware:

- `omi/firmware/omi/src/main.c` + `omi/firmware/omi/Kconfig` (`f81b8765a3`): the low-battery warning is now a distinct double-blink (`low_battery_phase < 2`, cycling mod 3 and resetting once the battery is back above threshold), so it no longer collides with the time-sync / disconnected red blink. That addresses the LED-confusion half of the earlier firmware review; the Kconfig help text matches the new pattern.
- `.github/scripts/test_check_release_process_guards.py` (`11aa7b5687`): dropping `test_desktop_qualification_runner_accepts_run_isolated_tag_checkout` is consistent with current `main`, where #11588/#11605 removed `check_desktop_qualification_runner()` from the guard script. The net diff of the release-guard files against this PR's base is empty, so this reads as cleanup rather than a guard weakening. I ran the suite on this head: 94 passed.
- `omi/firmware/omi/omi.conf` (`7989e42c7d`): removing the commented `CONFIG_BT_DIS/BAS/DEVICE_NAME/APPEARANCE` stubs is safe — the effective block (`omi.conf:100-109`) is unchanged.
- Re-ran the host policy test on this head: `bash omi/firmware/omi/tests/run_idle_sleep_test.sh` → `idle-sleep policy tests passed`. The encrypted permissions on 19B10014/19B10015/19B10016 with strict write validation and NVS-rollback on rename, the ref-counted `get_current_connection()`/`transport_off()` ownership transfer, the disconnect-time `is_capturing` clear, and the persisted-name restore during transport init all still check out. The new `OMI_FEATURE_SLEEP_COMMAND`/`OMI_FEATURE_DEVICE_NAME` bits (1<<9, 1<<10) match `OmiFeatures` in `app/lib/services/devices.dart`.
- GitHub now reports the PR mergeable with no conflicts, so I removed the stale `needs-rebase` label.

Leaving the merge decision with a human maintainer as before: this is power/identity firmware behavior on an encrypted link, the stacked dependency #10572 is still open, the shipped UX for app-driven sleep and the opt-in LED language still needs product sign-off before idle sleep is enabled in a shipping config, and on-device QA (pairing on the three encrypted characteristics, rename visibility in advertisement/scan response, physical system-off/wake) hasn't been run. The two Desktop Swift checks are also still queued.

---

Git-on-my-level:
Re-reviewed the current head. Thanks @undivisible — this round cleanly addresses the earlier human review feedback:

- Storage sync stays awake: `omi_idle_is_busy()` in `omi/firmware/omi/src/lib/core/idle_sleep.h` treats an active storage transfer as busy, and the host test (`tests/test_idle_sleep.c`) asserts the never-sleeps-mid-sync case. I ran `bash omi/firmware/omi/tests/run_idle_sleep_test.sh` on this head — passing.
- LED-pattern overlap: both new patterns remain opt-in Kconfig and are off in `omi.conf`; the low-battery warning is now a distinct double-blink (`low_battery_phase` in `src/main.c`), so it no longer collides with the solid-red disconnected state or the time-sync blink.
- `omi.conf` cleanup: the dead `CONFIG_BT_DIS/BAS=n` lines are removed rather than commented out (the `=y` settings later in the file already win).
- Idle-sleep data-loss risk: `OMI_ENABLE_IDLE_SLEEP` is default-n, `depends on !OMI_ENABLE_OFFLINE_STORAGE`, and explicitly unset in the shipping config — a build that can record to SD can never auto-power-off.

Also verified on this head: the encrypted GATT permissions for sleep (19B10014), capture-state (19B10015), and device-name (19B10016) are in place; the `current_connection` ref-counting is balanced at every call site I traced (`button.c` notifies, `storage_write` in both branches, the battery work handler, MTU recheck, both pusher paths, and `transport_off`'s ownership transfer under the spinlock); the device-name write validates length and embedded NUL and rolls back `bt_set_name()` if the NVS save fails; and the Dart side (`omi_connection.dart`) mirrors the same limits with new tests in `app/test/services/omi_device_controls_test.dart`.

What remains before merge, for a human maintainer:
- On-device validation of the power-off and rename flows (the earlier review noted this hadn't yet been tested on hardware).
- Dependency ordering: this should still land after #10572.
- Product sign-off on the eventual enablement plan (idle sleep and the new LED patterns surfacing in-app) before any of the opt-in flags are turned on in a shipping config.

Leaving for human maintainer review.

---

PR #10610, opened 2026-07-25
fix(desktop-windows): scope updater feeds to Windows releases

### LINKED ISSUE #10508: Bug: Omi Desktop for Windows updater fails with 404 trying to fetch latest.yml
## Summary
On Omi Desktop for Windows, clicking the update button fails and shows an updater error instead of checking or downloading the update successfully.

Image

## Problem
When clicking `Check for updates` in the Windows desktop app, the updater throws an error saying it cannot find `latest.yml` in the release artifacts.

The error shown includes a `404` response while trying to fetch:

`https://github.com/BasedHardware/omi/releases/download/v0.12.120+12120-macos/latest.yml`

## Expected behavior
Clicking the update button should successfully check for updates and download the correct update metadata for the Windows desktop app.

## Actual behavior
The updater fails with an error similar to:
- `Couldn't check for updates: Cannot find latest.yml in the latest release artifacts`
- `HttpError: 404`

The stack trace shown in the app points to the Electron updater flow.

## Observed issue
The request in the error appears to target a `macos` release artifact even though this is happening on the Windows desktop app.

That suggests the updater may be requesting the wrong platform artifact or release channel metadata.

## Impact
Users on Windows cannot reliably update from inside the app.

Instead of a normal update flow, the app surfaces a low-level updater error and the update process fails.

## Evidence
The screenshot from the Windows desktop app `Software updates` screen shows:
- `Couldn't check for updates: Cannot find latest.yml in the latest release artifacts`
- `HttpError: 404`
- a request to a `.../v0.12.120+12120-macos/latest.yml` URL

## Suggested investigation
- Check whether the Windows desktop app updater is pointing to the wrong platform artifact
- Check release artifact naming and publishing for `latest.yml`
- Check whether Windows builds are accidentally requesting macOS update metadata
- Check Electron updater configuration for platform/channel mismatch

## Environment
- Omi Desktop
- Windows
- Observed on Friday, July 24, 2026

tianmind-studio:
## What changed and why

Fixes #10508.

The Windows app now resolves a channel-scoped Windows release directory through the backend and points `electron-updater`'s generic provider at that immutable feed before every check. This avoids GitHubProvider's repository-wide `/releases/latest`, which can select a macOS release in this multi-platform repository.

Stable never falls through to beta; beta may fall back to stable while no beta release is available. The client validates the returned GitHub repository, Windows tag, and channel relationship, fails closed on resolver errors, and serializes checks across live channel changes. `OMI_UPDATER_DEV=1` continues to use the developer-supplied feed.

Each check, download, and staged installer is also bound to an immutable channel generation. Changing channels cancels or fences in-flight work, clears an installer staged by the previous channel, disables install-on-quit, and accepts `update-downloaded` only from the current version/channel attempt.

Rollout note: deploy the backend endpoint before publishing a Windows build that contains the client change. Until then, that build intentionally fails closed instead of falling back to the broken repository-wide provider.

- `INV-BETA-1` is preserved. This route is Windows-only and does not change the macOS beta identity, artifacts, or identity-aware appcast/download behavior; the full desktop update router regression suite remains green.

## How it was verified

- `python -m pytest tests/unit/test_desktop_updates.py -q` -> 124 passed.
- `npm test -- updater.test.ts updaterChannel.test.ts windowsUpdateFeed.test.ts` -> 13 passed.
- Targeted ESLint on the six updater TypeScript files -> passed.
- `npm run build` in `desktop/windows` -> node/web typecheck, Electron main/preload/renderer build, bytecode verification, and extension bundle passed.
- Applying the updated updater regression test to prior head `512af596e` fails because beta `2.0.0` remains staged after opt-out; current head clears it and rejects both pre-settlement and post-settlement delayed events.
- `PYTHONUTF8=1 scripts/pr-preflight --pr-body-file ...` in a Windows integration worktree with #10594 and #10595 applied -> all 22 selected checks passed.
- A live backend probe inspected 67 GitHub Windows releases and selected `v1.0.1-windows` for stable and `v1.0.19-windows` for beta.
- A real `electron-updater` GenericProvider probe read versions `1.0.1` and `1.0.19` from those immutable release feeds without selecting a macOS tag.

The full Windows Vitest suite was also attempted. Its unrelated baseline failures were reproduced on a clean main worktree: the bundled Node SQLite lacks FTS5, and a date-label assertion assumes an English locale while this machine uses `zh-CN`.

## Tests

- `backend/tests/unit/test_desktop_updates.py` covers immutable trusted feed construction, stable/beta selection, beta-to-stable fallback telemetry, strict stable behavior, and non-cacheable success/error responses.
- `desktop/windows/src/main/windowsUpdateFeed.test.ts` covers endpoint requests, trust-boundary validation, failure handling, refresh behavior, and stale-channel races.
- `desktop/windows/src/main/updater.test.ts` covers the Windows-only production path, live channel switches, generation-fenced in-flight checks/downloads, beta opt-out after staging, delayed download events, fail-closed resolution, and the staged-install flow.

## Failure class (fixes)

Failure-Class: new

The violated contract was that a Windows update check must resolve metadata from a Windows release in the requested channel. GitHubProvider's repository-wide latest-release lookup encoded neither platform nor Omi's release-channel semantics.

The new class is `FC-platform-unscoped-update-feed`. Its canonical guard is the backend's existing paginated, cached platform/channel release resolver combined with a client-side trusted-feed validator before the generic provider runs. Regression tests exercise both sides, and live probes verify the currently published stable and beta Windows feeds.

### DIFF
--- a/desktop/windows/src/main/updater.ts
+++ b/desktop/windows/src/main/updater.ts
@@ -1,23 +1,38 @@
 // Auto-update via electron-updater. Silent by design: downloads in the
-// background and installs on the NEXT quit (never force-restarts a listening
-// session). When an update is staged we tell the main window (so it can offer a
-// "restart to update" affordance) and mark the tray tooltip.
+// background and installs on the next quit (never force-restarts a listening
+// session). When an update is staged we tell the main window so it can offer a
+// "restart to update" action and mark the tray tooltip.
 //
-// Gated to packaged builds. For local testing, set OMI_UPDATER_DEV=1 and provide
-// dev-app-update.yml — that flips forceDevUpdateConfig so the updater runs
-// against a dev feed without a real signed release.
-import { app, type BrowserWindow } from 'electron'
+// Production checks first resolve an immutable Windows release directory from
+// the backend, then use electron-updater's generic provider for latest.yml and
+// its installer. This avoids GitHubProvider's repository-wide /releases/latest,
+// which can select a macOS release in this multi-platform repository.
+//
+// For local testing, set OMI_UPDATER_DEV=1 and provide dev-app-update.yml. That
+// keeps electron-updater on the developer-supplied feed.
+import { app, net, type BrowserWindow } from 'electron'
 import { autoUpdater } from 'electron-updater'
 import { setTrayUpdateReady } from './tray'
 import { markQuitting } from './lifecycle'
 import { getAppSettings, onAppSettingsChanged } from './appSettings'
-import { betaOptInToAllowPrerelease, resolveBetaChannelChange } from './updaterChannel'
+import { betaOptInToUpdateChannel, resolveBetaChannelChange } from './updaterChannel'
+import {
+  resolveWindowsUpdateFeedUrl,
+  WindowsUpdateFeedSelector,
+  type WindowsUpdateChannel
+} from './windowsUpdateFeed'
 import type { UpdateCheckResult } from '../shared/types'
 
-const CHECK_INTERVAL_MS = 4 * 60 * 60 * 1000 // every 4h
+const CHECK_INTERVAL_MS = 4 * 60 * 60 * 1000
+const OMI_API_BASE = import.meta.env.VITE_OMI_API_BASE || 'https://api.omi.me'
 
 let started = false
 let pendingUpdate: { version: string } | null = null
+let selectedChannel: WindowsUpdateChannel = 'stable'
+let feedSelector: WindowsUpdateFeedSelector | null = null
+let updateCheckTail: Promise<void> = Promise.resolve()
+
+type ElectronUpdateCheckResult = Awaited<ReturnType<typeof autoUpdater.checkForUpdates>>
 
 /** The update staged for install-on-quit, if any. The update:ready event fires
  * once (usually while nobody is on Settings), so the UI queries this on mount. */
@@ -25,11 +40,26 @@ export function getPendingUpdate(): { version: string } | null {
   return pendingUpdate
 }
 
+async function prepareUpdateFeed(): Promise<void> {
+  if (feedSelector) await feedSelector.prepareSelected()
+}
+
+function runPreparedUpdateCheck(): Promise<ElectronUpdateCheckResult> {
+  const operation = updateCheckTail.then(async (): Promise<ElectronUpdateCheckResult> => {
+    await prepareUpdateFeed()
+    return autoUpdater.checkForUpdates()
+  })
+  updateCheckTail = operation.then(
+    (): undefined => undefined,
+    (): undefined => undefined
+  )
+  return operation
+}
+
 /**
- * Manual update check for Settings → About. In unpackaged dev the updater never
- * started (see initAutoUpdater's guard), so there's nothing to check — return
- * `unsupported` and let the UI say updates install automatically. When active, run
- * a one-shot check: a staged download reports `update-available`, a newer feed
+ * Manual update check for Settings -> About. In unpackaged dev or on an
+ * unsupported platform the updater never starts, so there is nothing to check.
+ * When active, a staged download reports `update-available`, a newer feed
  * version reports `update-available`, otherwise `up-to-date`. Never throws.
  */
 export async function checkForUpdatesNow(): Promise<UpdateCheckResult> {
@@ -37,7 +67,7 @@ export async function checkForUpdatesNow(): Promise<UpdateCheckResult> {
   if (!started) return { status: 'unsupported', version: current }
   if (pendingUpdate) return { status: 'update-available', version: pendingUpdate.version }
   try {
-    const res = await autoUpdater.checkForUpdates()
+    const res = await runPreparedUpdateCheck()
     const found = typeof res?.updateInfo?.version === 'string' ? res.updateInfo.version : undefined
     if (found && found !== current) return { status: 'update-available', version: found }
     return { status: 'up-to-date', version: current }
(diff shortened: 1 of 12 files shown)


Git-on-my-level (commented):
Thanks for this — this is a clean, well-reasoned fix for #10508, and the test coverage is strong.

What stands out positively:
- The root cause is correctly identified: `GitHubProvider` resolves `/releases/latest`, which in this multi-platform repo can land on a macOS tag with no `latest.yml`. Scoping to an explicit Windows release directory is the right fix.
- The client-side trust validation in `windowsUpdateFeed.ts` is notably well done. Even if the backend response were tampered with, the client only accepts `https://github.com` origin, rejects userinfo/query/hash, and constrains the path to a BasedHardware/omi Windows release directory. Fails closed on any mismatch.
- The channel invariants are correctly modeled and tested: stable never falls through to beta; beta may fall back to stable when the slot is empty after promotion. The `WindowsUpdateFeedSelector` also handles the real concurrency edge case of a channel flip mid-check without applying a stale resolution to the wrong channel.
- Backend, client, route policy manifest, failure-class doc, and release-pipeline docs are all coherent around one outcome.

Two things to resolve before merge:

1. CI is not green. Formatting, Hygiene, and Detect Changes are showing as cancelled/failed. A clean passing run is needed before this can merge. These look like they may be cancelled from a superseded run — if so, re-triggering should clear them.

2. As the PR body correctly notes, the backend endpoint (`/v2/desktop/update-feed/windows`) must be deployed before a Windows build containing the client change is published, since the new client intentionally fails closed rather than falling back to the broken repository-wide provider. Please confirm the deployment sequence with whoever owns the backend release schedule.

Marking for maintainer review since this touches the auto-update/release pipeline. No changes requested — the implementation itself looks correct.

---

kodjima33 (approved):
Scopes the Windows updater feed to Windows releases; tests included. Approve-only — 870 lines spanning backend route policy + desktop, above the scoped-diff bar for an agent merge.

desktop/windows/src/main/updater.ts
```diff
@@ -93,35 +127,41 @@ export function initAutoUpdater(getMainWindow: () => BrowserWindow | null): void
...
+  setInterval((): void => {
+    void check()
+  }, CHECK_INTERVAL_MS)
+
+  // Apply a live beta toggle immediately instead of waiting for the 4h timer.
+  onAppSettingsChanged((settings) => {
+    const change = resolveBetaChannelChange(selectedChannel, settings.betaUpdatesEnabled)
+    if (!change.changed) return
+    selectedChannel = change.channel
+    autoUpdater.allowPrerelease = selectedChannel === 'beta'
```
ZachL111, line 223: [P2] Cancel or fence an in-flight beta download on opt-out

This moves future checks to stable, but it does not stop a beta updater operation that has already found metadata. With `autoDownload` enabled, electron-updater 6.8.9 starts `downloadUpdate(cancellationToken)` and returns the check result without awaiting that download. The queued stable check therefore does not cancel the old beta download. If it later emits `update-downloaded`, the unconditional handler sets `pendingUpdate`, and `autoInstallOnAppQuit` can install a beta after the user disabled beta updates.

Please retain and cancel the active token, or otherwise generation-fence both download completion and the staged result. Add a regression test that starts a beta download, flips `betaUpdatesEnabled` to false, then completes the old download and proves it is neither staged nor installed.

ZachL111 (changes requested):
Requesting changes for one channel-integrity race. The selector fences stale feed-resolution responses, but an already-started beta download can outlive a beta opt-out and still be staged for installation. Details and a deterministic regression scenario are inline.

Git-on-my-level:
Independent review confirming the channel-integrity race @ZachL111 identified should be resolved before merge.

With `autoDownload` enabled, `checkForUpdates()` starts the download asynchronously and returns the check result without awaiting it. `WindowsUpdateFeedSelector` correctly fences feed *resolution* across a channel flip, but an in-flight beta download can still complete and fire `update-downloaded` after the user has moved back to stable. The handler sets `pendingUpdate` unconditionally, so `autoInstallOnAppQuit` would stage a beta build the user just opted out of.

Suggested fix: generation-fence the `update-downloaded` handler — track which channel the active download started under and discard the staged result if the selected channel has since changed. That is more robust than relying on `CancellationToken` cancellation semantics across electron-updater internals.

The rest of the implementation is strong: client-side feed validation (origin/path/channel checks, fails closed), channel invariant modeling (stable never falls to beta; beta falls back to stable only when empty), and test coverage across backend + TypeScript are all well done. Holding on `needs-maintainer-review` for human sign-off given this touches the auto-update pipeline.

---

desktop/windows/src/main/updater.ts
tianmind-studio: Addressed in 2d6e2a2ae. Checks, downloads, and staged installers now carry an immutable channel generation; opting out revokes the active attempt, cancels its token, clears any staged beta installer, disables install-on-quit, and accepts `update-downloaded` only for the matching active version/generation. The regression stages beta 2.0.0, opts out, then sends delayed events both before and after the old download settles and proves neither can be installed. That test fails on prior head 512af596e and the fixed updater suite passes 13/13; affected lint, typechecks, and the production Windows build also pass.

kodjima33 (approved):
Channel-scoped Windows update feed is the right fix for #10508 (GitHubProvider's repo-wide /releases/latest can hand a Windows client a macOS release); 1024 lines across backend + desktop is past the unattended-merge bar.

Git-on-my-level:
Thanks — confirmed the channel-integrity race from the earlier change request is properly resolved on the current head.

What is now correctly handled:
- `autoDownload` is false, so each check explicitly awaits `downloadUpdate` rather than returning after starting an async download.
- Every check, download, and staged installer carries an immutable `channelGeneration`. An opt-out bumps the generation, cancels the active download's cancellation token, clears any staged beta installer, disables install-on-quit, and rejects a stale `update-downloaded` event that does not match the current version/generation.
- The regression test (stages beta 2.0.0, opts out, fires delayed `update-downloaded` both before and after the old download settles) is deterministic and would fail on the prior head.

Remaining items for the human maintainer before merge:
1. CI is not green — Formatting and Hygiene are cancelled. These look like superseded-run artifacts; please re-trigger for a clean passing run.
2. As the PR body notes, the backend endpoint `/v2/desktop/update-feed/windows` must be deployed before a Windows build containing this client change is published (the new client intentionally fails closed rather than falling back to the broken repo-wide provider).

Keeping `needs-maintainer-review` since this touches the auto-update/release pipeline.

---

PR #10537, opened 2026-07-25
fix(desktop): stream public web-search turns

Git-on-my-level:
## What changed and why

Fresh public-web prompts were classified twice and forced into a special, buffered Anthropic server-tool path. That path waited for the entire server-tool and any `pause_turn` continuation before writing its first OpenAI SSE chunk, so the desktop client saw no progress and interrupted the turn. Web search is now exposed as a normal supported agentic tool, with source choice left to the existing desktop retrieval rules; every streaming turn uses the incremental SSE path.

This branch also merges the upstream registration for the previously failing `FC-split-mutation-authority` guard artifact. The pre-push hook now keeps that deterministic guard-artifact ratchet mandatory even when its broader PR-preflight hatch is used.

Syncing `main` also exposed an incomplete macOS diagnostics change: two files were missing their `OmiSupport` imports, a lock-backed termination flag did not meet Swift 6's concurrency rules, and actor-returned diagnostics were not `Sendable`. The build now uses the existing unfair-lock pattern, carries the immutable bounded diagnostic context safely across actors, and emits the established snake-case startup stages.

## Product invariants affected

- INV-AUTH-1
- INV-CHAT-1

## How it was verified

- `cd desktop/macos/Backend-Rust && cargo fmt --check && cargo test` — 376 passed.
- `cd desktop/macos && ./scripts/agent-logic-harness.sh --cross-surface-smoke` — passed (Swift, agent, and Rust retrieval contracts).
- Built and launched the ad-hoc signed named bundle `omi-web-search-stream` against the development backend. The exact live prompt could not be run because this machine has no seedable development auth session; production/Beta credentials were not accessed.
- `cargo clippy --bin omi-desktop-backend -- -D warnings` reaches five existing unrelated warnings in `chat_completions.rs` and `desktop_releases_repository.rs`; this change adds none.
- `xcrun swift build -c debug --package-path Desktop` — passed after the sync repairs.
- `xcrun swift test -c debug --package-path Desktop --filter DesktopErrorTelemetryTests` — 6 passed.
- `cd desktop/macos && ./scripts/desktop-core-harness.sh --self-check --skip-backend-contracts` — passed; the `chat-hermetic` flow now covers the bridge lifecycle source.
- The hermetic T2 `chat-hermetic` flow could not start locally because an unrelated Java process already owns the fixed Firestore-emulator port 8085; it was left untouched.
[...]

PR #10616, opened 2026-07-26
fix(desktop): keep the thinking signature when continuing a paused turn

kodjima33:
## What changed and why

`pause_turn` continuations rebuild the paused assistant turn from the streamed
content blocks and resend it to Anthropic. The rebuilder (`StreamedContentBlocks`,
added with the web-search streaming change #10537) handled `text_delta`,
`thinking_delta` and `input_json_delta` but ignored `signature_delta`, so a
thinking block was replayed **unsigned** — and Anthropic rejects an unsigned
thinking block.

Typed chat is exactly where both halves meet:

- adaptive thinking is enabled for the first model call of a user turn
  (`use_adaptive_thinking`, `request_translation.rs`), and
- the server-side `web_search` tool is exposed whenever the client sends tools
  (`inject_web_search`).

So a typed-chat turn that paused mid web search failed its own continuation:
`complete_anthropic_server_tool_turn` returned `Err`, and the user saw the
partial answer end in `"Upstream provider error"` instead of the searched
answer — the failure mode #10537 exists to remove. The same file already
documents this class one function up: *"a thinking-enabled request whose
assistant tool_use turn lacks its thinking block is rejected upstream."*

The fix is the missing arm: capture `signature_delta` into the rebuilt block,
via the same `append_string_field` helper the other text-bearing deltas use
(signatures may arrive across several deltas, so it appends).

No changelog fragment: #10537 has not shipped in a release yet
(`desktop/macos/changelog/unreleased/20260725-web-search-streaming.json` is
still unreleased), so this repairs an unreleased change rather than adding a
user-visible line of its own.

none

## Failure class (fixes)

Failure-Class: none

## How it was verified

- `cargo test` — **378 passed**, 0 failed.
- The new regression test
  `test_pause_turn_stream_accumulator_preserves_thinking_signature` **fails on
  the pre-fix accumulator** (`git stash` of `streaming.rs` only) with the
  signature absent from the rebuilt block, and passes with the fix:
  ```
  left:  [{"thinking":"Check the web","type":"thinking"}, ...]
  right: [{"signature":"ErrUhd8=","thinking":"Check the web","type":"thinking"}, ...]
  ```
- `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean.
- `make preflight` green.

**Not reproduced end to end against live Anthropic:** `pause_turn` is upstream's
choice and cannot be forced, and `ANTHROPIC_API_URL` is a `const` with no test
seam, so the continuation request body was verified through the production
accumulator that builds it, not over the wire.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10616?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Backend-Rust/src/routes/chat/streaming.rs
+++ b/desktop/macos/Backend-Rust/src/routes/chat/streaming.rs
@@ -69,6 +69,16 @@ impl StreamedContentBlocks {
                     append_string_field(block, "thinking", thinking);
                 }
             }
+            // The signature is what makes a replayed thinking block valid. A
+            // pause_turn continuation resends this assistant turn to Anthropic,
+            // and a thinking block whose signature is missing or truncated is
+            // rejected upstream — dropping it here would fail the very turn the
+            // continuation exists to finish.
+            Some("signature_delta") => {
+                if let Some(signature) = delta.get("signature").and_then(Value::as_str) {
+                    append_string_field(block, "signature", signature);
+                }
+            }
             Some("input_json_delta") => {
                 if let Some(partial_json) = delta.get("partial_json").and_then(Value::as_str) {
                     self.input_json
--- a/desktop/macos/Backend-Rust/src/routes/chat/tests/all.rs
+++ b/desktop/macos/Backend-Rust/src/routes/chat/tests/all.rs
@@ -1103,6 +1103,50 @@ fn test_pause_turn_stream_accumulator_preserves_raw_content_for_continuation() {
     );
 }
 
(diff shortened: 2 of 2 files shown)


undivisible (approved):
LGTM. Focused fix — one match arm + regression test. CI green. Approving and merging.

PR #10625, opened 2026-07-26
fix(release): align qualification guard with isolated checkout

tianmind-studio:
## What changed and why

Align the release-process guard with the run-isolated detached-tag checkout introduced by #10611, and make the desktop release flow contract verify that the guard tracks that checkout boundary. The stale guard still required the removed `actions/checkout` `ref:` field and has reddened `main` since run 30182385274.

None.

## How it was verified

- Direct `check_desktop_qualification_runner()` reproduction changed from the stale-fragment error to `[]`
- Five host-portable `DesktopReleaseFlowContractTests` cases passed, including the new guard/workflow alignment regression
- `python -m py_compile .github/scripts/check-release-process-guards.py .github/scripts/test_desktop_release_flow_contract.py` - passed
- `git diff --check` - passed

The two shell-execution cases in the same suite were not run locally because Windows resolves the system WSL `bash.exe` before Git Bash for child processes; Linux CI exercises the complete suite.

## Tests

`test_release_process_guard_accepts_the_run_isolated_tag_checkout` fails against the current `main` guard and passes only when the guard requires the detached immutable-tag checkout used by the qualification workflow.

## Failure class (fixes)

Failure-Class: FC-shared-runner-checkout-residue

The cross-file contract would have caught the guard drift introduced when #10611 replaced shared-workspace `actions/checkout` with an isolated source tree. It extends the existing shared desktop release-flow primitive instead of creating another checker.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10625?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/check-release-process-guards.py
+++ b/.github/scripts/check-release-process-guards.py
@@ -858,7 +858,7 @@ def check_desktop_qualification_runner() -> list[str]:
         "self-hosted",
         "macos",
         "omi-desktop-qualification",
-        "ref: ${{ inputs.release_tag }}",
+        'checkout --quiet --detach "refs/tags/$RELEASE_TAG"',
         "check-desktop-auto-beta-candidate.py",
         "--automatic",
         "actions/create-github-app-token@v3",
--- a/.github/scripts/test_desktop_release_flow_contract.py
+++ b/.github/scripts/test_desktop_release_flow_contract.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import ast
 import json
 import os
 import subprocess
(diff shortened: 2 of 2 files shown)


tianmind-studio:
@Git-on-my-level need human response: this is the two-file guard follow-up to #10611. Main run https://github.com/BasedHardware/omi/actions/runs/30182385274 failed because the guard still required the removed actions/checkout ref field; the latest PR Hygiene run https://github.com/BasedHardware/omi/actions/runs/30183453224 now passes the shared preflight and release-process guard. No release, tag, qualification, or deployment mutation is involved.

Git-on-my-level (commented):
Thanks for the focused follow-up. I reviewed the guard/test change and the new contract is aligned with the run-isolated detached-tag checkout in `desktop_qualify_beta.yml`: the release-process guard now checks for the immutable tag checkout boundary, and the added AST-based regression catches a future drift back to the old `actions/checkout ref:` requirement.

I also ran the relevant checks locally in a stripped environment:
- `python -m py_compile .github/scripts/check-release-process-guards.py .github/scripts/test_desktop_release_flow_contract.py`
- `python .github/scripts/test_desktop_release_flow_contract.py -v` (7 tests passed)
- `python .github/scripts/check-release-process-guards.py` (`release process guard checks passed`)

I don't see a blocking code issue in this patch. Because it touches release/workflow guardrails, I’m leaving this as a positive signal rather than a formal approval; a human maintainer should still sign off before merge.

---

PR #10628, opened 2026-07-26
fix(desktop-windows): detect current Google Meet titles

### LINKED ISSUE #10553: Bug: Windows desktop app does not auto-record meetings or show start-recording prompt even with settings enabled
## Summary
On the Omi Windows desktop app, meetings do not auto-record and the app also does not show the pop-up prompt to start recording, even when the relevant settings are enabled.

## Problem
Meeting detection is not triggering the expected behavior.

Even with the settings enabled, the app neither:
- starts recording automatically
- nor shows the prompt/pop-up asking to start recording

## Test case
This was tested with a Google Meet meeting.

## Expected behavior
When meeting recording settings are enabled, the Windows desktop app should either:
- auto-start meeting recording
- or show the expected pop-up prompt to start recording

## Actual behavior
Neither behavior happens.

The meeting is not auto-recorded, and no start-recording pop-up is shown.

## Impact
This makes the meeting recording workflow unreliable.

Users can enter a meeting expecting Omi to start recording or at least prompt them, but nothing happens and the meeting may be missed entirely.

## Suggested investigation
- Check whether Google Meet meeting detection is working on Windows desktop
- Check whether the prompt logic is failing even when detection succeeds
- Check whether the auto-record and prompt settings are being read correctly by the app
- Check whether this is a Windows-specific issue in the meeting detection or notification pipeline

## Environment
- Omi desktop app
- Windows
- Tested with Google Meet
- Observed on Saturday, July 25, 2026

tianmind-studio:
Fixes #10553

## Summary

- Recognize the current Google Meet join-page titles `Meet` and `Google Meet`.
- Preserve support for the existing `Meet - <meeting code>` title form.
- Keep matching browser-scoped and reject ordinary titles such as `Meeting notes`.

## Root cause

The bundled matcher required a separator after `Meet`, but the current Google Meet join page reports the exact document title `Meet`. The browser starts using the microphone on that join surface, so the ConsentStore trigger can run before the title ever has a meeting-code suffix. A later document-title change does not emit the foreground-window event the detector watches, leaving both ask and auto modes silent.

none

## Failure class

Failure-Class: new

The event-triggered detector accepted only a later presentation of the same provider identity. The behavioral pattern test now covers the join and in-call title forms at the trigger boundary.

## Validation

- `node node_modules/vitest/vitest.mjs run src/main/meeting/patterns.test.ts src/main/meeting/detector.test.ts` (26 passed)
- `node node_modules/typescript/bin/tsc --noEmit -p tsconfig.node.json --composite false`
- `node node_modules/prettier/bin/prettier.cjs --check src/main/meeting/patterns.json src/main/meeting/patterns.test.ts`
- `git diff --check`
- Live page probe: `https://meet.google.com/abc-defg-hij` reported the title `Meet`.
- Native Windows probe: an actual Chrome microphone stream appeared in ConsentStore as active `chrome.exe`.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10628?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-meeting-trigger-title-identity-drift.json
+++ b/.github/failure-classes/FC-meeting-trigger-title-identity-drift.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-meeting-trigger-title-identity-drift",
+  "violated_contract": "Event-triggered meeting detection must recognize a provider in the browser-title form present when microphone use first triggers evaluation; it must not depend only on a later title form whose transition emits no detector event.",
+  "canonical_prevention": "Keep provider join and in-call title identities together in tested pattern data, scope title matching to known browser processes, and retain negative near-prefix cases so broader coverage does not create generic-title false positives.",
+  "canonical_prevention_artifact": [
+    "desktop/windows/src/main/meeting/patterns.test.ts"
+  ],
+  "evidence_prs": [
+    10628
+  ],
+  "scope_hints": [
+    "desktop/windows/src/main/meeting/patterns.json",
+    "desktop/windows/src/main/meeting/patterns.test.ts"
+  ],
+  "status": "open"
+}
--- a/desktop/windows/src/main/meeting/patterns.json
+++ b/desktop/windows/src/main/meeting/patterns.json
@@ -30,7 +30,11 @@
     "arc.exe"
   ],
   "titles": [
-    { "id": "meet-web", "name": "Google Meet", "pattern": "^Meet\\s+[-–—]\\s" },
+    {
+      "id": "meet-web",
+      "name": "Google Meet",
+      "pattern": "^(?:Google\\s+)?Meet(?:\\s+[-–—]\\s|$)"
+    },
     { "id": "zoom-web", "name": "Zoom", "pattern": "Zoom Meeting" },
     { "id": "teams-web", "name": "Microsoft Teams", "pattern": "\\| Microsoft Teams" },
     { "id": "webex-web", "name": "Webex", "pattern": "\\bWebex\\b" }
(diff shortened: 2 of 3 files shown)


tianmind-studio:
CI note: the Windows typecheck/tests, Windows build, and Linux package smoke all pass. The only failure is the current-main qualification guard still requiring the removed checkout ref fragment; #10625 fixes that exact guard and its Hygiene run is green, so no meeting-detection or failure-class check failed here.

Git-on-my-level:
Thanks @tianmind-studio — this is a clean, well-scoped fix for #10553.

The regex change is tight: `^(?:Google\s+)?Meet(?:\s+[-–—]\s|$)` adds the exact join-page titles `Meet` and `Google Meet` (the form present when the mic ConsentStore trigger fires) while preserving the `Meet - <code>` in-call form. I checked a range of titles and the near-prefix false positives are correctly rejected (`Meeting`, `Meeting notes`, `Meetup`, `Meet the team`, `Google Meeting`, etc.); the browser-process gate in `matchTier1` keeps the broader exact-title match scoped to known browsers, so it doesn't leak to arbitrary apps. The test cases are consistent with the new behavior, and `FC-meeting-trigger-title-identity-drift.json` follows the existing failure-class schema.

One note for a maintainer: the only failing check is `Hygiene`, and it's failing on `main` as well (exit 1 in `.github`, only grandfathered architecture-map warnings — no annotation on this PR's files). The author cross-references #10625 as the separate fix for that guard. So this is not a substantive merge blocker for this PR, but a human should confirm the CI state / exemption before merge.

Flagging for maintainer review rather than approving, since this touches a user-facing detection surface and the context gate currently marks the PR must-not-approve due to the Hygiene failure.

This feedback is from an automated maintainer assistant and may be imperfect.

---

kodjima33 (approved):
Desktop bug fix (fixes #10553): Google Meet join-page title 'Meet' now matches the meeting detector; root cause explained, pattern test added, CI green.

PR #10638, opened 2026-07-26
Harden auth, SSRF, and CSRF gaps left open from the April disclosure

amowpheth:
## Summary

Three months after the coordinated disclosure (GHSA-4j2c-fmg6-8h42, 17 findings, CVSS 10.0 aggregate), several of the reported issues are still exploitable in the current `backend/` and `web/admin/`. RCE, the hardcoded encryption key, path traversal, and the unauthenticated-conversations gap are already fixed upstream — this PR closes the remaining ones with a live exploit path, while staying backward compatible with this repo's own test suites, dev-harness, and production gauntlet that depend on the same env vars/token formats.

- **ADMIN_KEY prefix-match auth bypass** (`backend/utils/other/endpoints.py`) — anyone holding `ADMIN_KEY` could impersonate any uid via `token.startswith(admin_key)`. Switched to constant-time comparison, added audit logging on every use, and added an opt-out flag (`ADMIN_KEY_AUTH_ENABLED`) — kept the existing concatenated-token format (`ADMIN_KEY<uid>`) intact since it's documented in `AGENTS.md` and used throughout `backend/testing/`, `backend/tests/integration/`, and the prod-facing memory-continuity gauntlet.
- **`LOCAL_DEVELOPMENT` hardcoded-uid bypass** — any invalid Firebase token + `LOCAL_DEVELOPMENT=true` returned uid `'123'` with no further check. Now inert once real Firebase credentials (`SERVICE_ACCOUNT_JSON` / `GOOGLE_APPLICATION_CREDENTIALS`) are configured, which every real deployment sets and every legitimate dev/test path (emulator-based) already avoids.
- **SSRF via developer-configured `webhook_url` / `setup_completed_url`** (`app_integrations.py`, `oauth.py`, `utils/http_client.py`) — requests to app-developer-supplied URLs had no destination validation. Added a check that rejects private/loopback/link-local/reserved-address targets (including the cloud metadata IP) before dispatch.
- **OAuth authorize/token-exchange CSRF** (`routers/oauth.py`) — the `state` param was accepted and echoed back but never validated against anything. Added a double-submit cookie set on `/v1/oauth/authorize` and checked on `/v1/oauth/token`.
- **BYOK fingerprint hashing** (`utils/byok.py`) — unsalted SHA-256. Added an opt-in HMAC pepper (`BYOK_FINGERPRINT_PEPPER`) with legacy-value fallback so existing fingerprints keep validating.
- **CORS** — `main.py` had no `CORSMiddleware` at all; added an explicit default-deny policy.
- **Dockerfile** — final stage ran as root; added a non-root `USER`.
- **`payment.py`** — one endpoint echoed a raw exception string from the PayPal payment-details call back to the client; now logs server-side and returns a generic message.

## Test plan

- [ ] `ADMIN_KEY_AUTH_ENABLED` unset/false → admin-key auth path fully disabled
- [ ] `ADMIN_KEY_AUTH_ENABLED=true` + existing `Bearer <ADMIN_KEY><uid>` format from `backend/tests/integration/test_live_chat_service.py` and `backend/testing/*_stack/run.py` still authenticates correctly
- [ ] `LOCAL_DEVELOPMENT=true` without emulator/real credentials still works against the hermetic harness (`backend/testing/e2e/conftest.py`), and cannot activate once `SERVICE_ACCOUNT_JSON`/`GOOGLE_APPLICATION_CREDENTIALS` is set
- [ ] Webhook delivery to `169.254.169.254`, `127.0.0.1`, and RFC1918 targets is rejected; delivery to a public HTTPS endpoint still succeeds
- [ ] `/v1/oauth/authorize` → `/v1/oauth/token` round-trip succeeds with matching cookie/state; fails when state/cookie are mismatched or missing
- [ ] Existing BYOK fingerprints (no pepper) still validate after `BYOK_FINGERPRINT_PEPPER` is introduced; new fingerprints are peppered when the var is set
- [ ] `docker build` + run as non-root in the final stage (not verified in this environment — no Docker daemon available; please confirm in CI)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10638?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.env.template
+++ b/backend/.env.template
@@ -26,6 +26,20 @@ DEEPGRAM_API_KEY=
 MODULATE_API_KEY=
 
 ADMIN_KEY=
+# Set to 'false' to disable the ADMIN_KEY<uid> local-dev auth bypass entirely
+# (it defaults to 'true' for backward compatibility with existing deployments).
+ADMIN_KEY_AUTH_ENABLED=true
+# Server-side HMAC pepper for stored BYOK key fingerprints. When unset,
+# fingerprints are stored as the plain client SHA-256 (legacy); when set,
+# they are wrapped with this secret before storage/comparison so a leaked
+# fingerprint store is not rainbow-table attackable. Leave unset for
+# backward compatibility; set a stable random value to opt in.
+BYOK_FINGERPRINT_PEPPER=
+# Comma-separated exact origins allowed by CORS (e.g. https://app.omi.com).
+# Defaults to empty (no origins) — the API is Bearer-token authenticated and
+# browser default-deny already applies. Never use "*" and never combine with
+# credential forwarding.
+CORS_ALLOWED_ORIGINS=
 OPENAI_API_KEY=
 
 # Optional backend PostHog sink for provider sync/auth telemetry.
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -67,5 +67,14 @@ COPY .github/scripts/desktop_qualification_admission.py /app/desktop_qualificati
 COPY .github/scripts/desktop_qualification_evidence.py /app/desktop_qualification_evidence_contract.py
 COPY backend/ .
 
+# Run as a non-root user: a compromise of the app process (e.g. via a
+# dependency RCE) shouldn't hand the attacker root inside the container by
+# default. /app needs to stay writable — sync.py writes uploaded audio to a
+# relative `syncing/<uid>/` directory under the working dir at runtime.
+RUN groupadd --system --gid 10001 omi && \
+    useradd --system --uid 10001 --gid omi --no-create-home omi && \
+    chown -R omi:omi /app
+USER omi
+
 EXPOSE 8080
 CMD ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "8080", "--loop", "uvloop"]
--- a/backend/main.py
+++ b/backend/main.py
@@ -12,6 +12,7 @@ logger = logging.getLogger(__name__)
 
 import firebase_admin
 from fastapi import FastAPI
+from starlette.middleware.cors import CORSMiddleware
 
 from database.google_credentials import prepare_google_credentials
 
@@ -113,6 +114,23 @@ else:
     firebase_admin.initialize_app()  # type: ignore[reportUnknownMemberType]  # firebase_admin untyped
 
 app = FastAPI()
+
+# Explicit, default-deny CORS: this API is Bearer-token authenticated (mobile/
+# desktop apps, not ambient browser cookies), so no cross-origin browser
+# caller needs to be allowed by default. CORS_ALLOWED_ORIGINS lets an operator
+# opt a specific web frontend in (comma-separated exact origins — never "*",
+# and never combined with allow_credentials, which would let any site read
+# authenticated responses for a signed-in visitor).
+_cors_allowed_origins = [o.strip() for o in os.getenv('CORS_ALLOWED_ORIGINS', '').split(',') if o.strip()]
+if '*' in _cors_allowed_origins:
+    raise RuntimeError('CORS_ALLOWED_ORIGINS must not contain "*" — list explicit origins instead')
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=_cors_allowed_origins,
+    allow_credentials=False,
+    allow_methods=['*'],
+    allow_headers=['*'],
+)
 
 app.include_router(transcribe.router)
 app.include_router(omni_relay.router)
--- a/backend/routers/oauth.py
+++ b/backend/routers/oauth.py
@@ -1,6 +1,9 @@
+import hmac
+import logging
 import os
+import secrets
 from typing import Optional
-from fastapi import APIRouter, Request, HTTPException, Form
+from fastapi import APIRouter, Cookie, Request, HTTPException, Form
 from fastapi.responses import HTMLResponse
 from fastapi.templating import Jinja2Templates
 from pydantic import BaseModel, Field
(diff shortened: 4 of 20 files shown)


Git-on-my-level (changes requested):
Thanks for picking up these post-disclosure hardening gaps — the direction is valuable, especially default-deny CORS, refusing obvious private/link-local callback targets, not echoing unexpected PayPal persistence errors, and making the BYOK fingerprint storage optionally peppered.

I don’t think this should merge in its current form yet because these are security-sensitive auth/webhook changes and there are no targeted regression tests covering the new behavior. Before maintainer merge review, please add focused tests for at least:

- `assert_public_http_url`: rejects loopback/private/link-local/reserved hosts and non-http(s) schemes, accepts ordinary public `http`/`https`, and handles multi-address DNS results safely.
- OAuth CSRF: `/authorize` emits the CSRF cookie/template token, `/token` rejects missing or mismatched token/cookie pairs, and accepts the valid pair.
- BYOK fingerprint migration: newly stored peppered fingerprints validate, existing legacy plain fingerprints still validate, and mismatches fail.
- `verify_token`: `LOCAL_DEVELOPMENT=true` only falls back to uid `123` when real Firebase credentials are absent, and `ADMIN_KEY_AUTH_ENABLED=false` disables the admin-key path.

One additional security point for human review: the SSRF helper validates DNS before the later `httpx` request, but the actual connection resolves again. That is a good improvement for direct private-IP/internal-host callbacks, but it is not full DNS-rebinding protection. If this PR is intended to close the disclosure item completely, please either pin/connect to the validated address (or otherwise enforce validation at connection time), or clearly scope this as a partial mitigation so maintainers can decide whether the residual risk is acceptable.

Because this touches auth, OAuth, BYOK, webhook destinations, CORS, and container privileges, I’m marking this for security/maintainer review rather than approving automatically.

---

amowpheth (commented):
Hi maintainers, following up on the April coordinated disclosure (GHSA-4j2c-fmg6-8h42). I re-verified all 17 findings against current main before touching anything: RCE, the hardcoded encryption key, path traversal, and unauthenticated conversation access are already fixed. This PR addresses the remaining ones that still have a live exploit path (ADMIN_KEY prefix bypass, the LOCAL_DEVELOPMENT uid shortcut, webhook SSRF, OAuth state CSRF, unsalted BYOK fingerprinting, missing CORS policy, root Docker user).

I intentionally avoided changing the ADMIN_KEY token format or removing the LOCAL_DEVELOPMENT shortcut outright, since both are relied on by your own integration tests and the memory-continuity gauntlet. Instead these are now opt-in/fenced by conditions your existing tooling already satisfies, so nothing in CI should need to change. Happy to adjust the defaults (e.g. ADMIN_KEY_AUTH_ENABLED) if you'd rather they default the other way.

One thing I couldn't verify myself: the Dockerfile's non-root USER change, I have no Docker daemon in my environment to build and confirm nothing in the image (file perms, mounted volumes, etc.) assumes root. Flagged that in the test plan; would appreciate a CI run or a maintainer sanity-check there before merge.

V/R,
amowpheth (Robert)

amowpheth (commented):
The original fix asked "does this hostname resolve to a public IP?" and then, a moment later, made the HTTP request to that same hostname which so happens to trigger a second, independent DNS lookup. An attacker who controls DNS for their webhook domain can answer those two lookups differently: public IP for the check, private/metadata IP for the real connection. The check and the connection weren't atomic, so the check was decorative.

Remedial action: resolve once, validate once, then force the actual socket to connect to that exact IP never letting the HTTP client look the hostname up again. Still sent the original hostname as the Host header and as the TLS SNI value, so that the request looks completely normal to the receiving server and certificate checks still validate against the real domain, not the IP. you can run this yourself in a MAC terminal with the lines below:

cd ~/omi/backend
./.test-venv/bin/python -m pytest tests/unit/test_ssrf_public_url_guard.py tests/unit/test_oauth_csrf.py tests/unit/test_byok_fingerprint_pepper_migration.py tests/unit/test_verify_token_admin_and_local_dev_gating.py -v

Git-on-my-level (changes requested):
Thanks for the follow-up here. The main blockers from the earlier automation review are materially improved now: the PR adds focused coverage for the OAuth CSRF path, BYOK fingerprint migration, admin/local-dev auth gates, and the SSRF URL guard, and the webhook call sites now try to connect to the validated/pinned address instead of re-resolving the hostname.

I’m still requesting changes because the current head is not green: the Backend unit suite is failing in `tests/unit/test_listen_finalization_cloud_tasks.py::test_finalizer_completes_when_an_app_permanently_rejects_the_delivery` for both asyncio and trio. The log shows the fanout path now fails before exercising the mocked 401 response:

```text
gather_safe[trigger_integrations] item 0 failed: BlockedNetworkError("Blocked DNS resolution for 'app.test'")
AssertionError: Expected record_failure('omi-google-drive-integration', 401, 'HTTP 401') to be called once. Called 0 times.
```

That looks like the existing listen-finalization unit harness needs the same kind of `safe_request_target`/`UnsafeWebhookURLError` stub update added to the other integration tests, so the test can keep validating permanent webhook rejection instead of being intercepted by the new DNS guard.

One security-review note to keep on the maintainer radar after CI is fixed: the pinned-address approach is directionally right, but because this is auth/webhook/OAuth/BYOK/CORS/container hardening from a first-time contributor, I still would not auto-approve it. A human maintainer/security pass should confirm the `httpx` pinned-IP + `Host`/SNI behavior, redirect handling, deployment CORS expectations, and the Docker non-root runtime assumptions before merge.

---

Git-on-my-level:
Thanks for the work on this, next time please submit a private PR first per best practice, but really appreciate the work on this

PR #10644, opened 2026-07-26
SCA-163: report M1 qualification lease failures

Git-on-my-level:
## Summary

- Route M1 qualification lease acquire/release through the tag-pinned backend virtualenv and cached worktree.
- Keep successful acquire stdout capability-only even when the harness emits stderr; relay concise diagnostics only when a lease command fails.
- Add behavioral regression coverage for cached-worktree cwd binding, success-with-stderr JSON isolation, failure diagnostics, and release-token redaction.
- Preserve the existing exact-tag checkout contract used by the qualification guard.

## Incident evidence

The immutable `v0.12.129+12129-macos` candidate (`b07851f438dac10ebeec3dac90c318423498f390`) failed its sole M1 Studio qualification run (`30194590936`) immediately after backend dependency sync with exit 2. The runner-capacity preflight passed, no lease was acquired, and the raw log contained no diagnostic because the lease CLI's stdout failure was captured as expected JSON.

## Verification

- `bash desktop/macos/tests/test-qualification-lease-command.sh`
- `bash desktop/macos/tests/test-qualify-desktop-beta-contract.sh`
- `python3 .github/scripts/test_desktop_release_flow_contract.py`
- `python3 .github/scripts/test_desktop_changelog.py`
- `python3 .github/scripts/check-release-process-guards.py`
- `make preflight`

## Failure class

Failure-Class: none

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -519,15 +519,21 @@ checks:
     reason: "#10163 mutation-sensitive Stable retry and default-channel feed fixtures"
   - id: desktop-release-one-path-contract
     command: ["python3", ".github/scripts/test_desktop_release_flow_contract.py"]
-    triggers: ["codemagic.yaml", ".github/workflows/desktop_auto_release.yml", ".github/workflows/desktop_qualify_beta.yml", ".github/workflows/desktop_promote_beta.yml", ".github/workflows/desktop_recover_beta.yml", ".github/workflows/desktop_rollback_beta.yml", ".github/workflows/desktop_breakglass_rollout_beta.yml", ".github/workflows/desktop_beta_admission_control.yml", ".github/workflows/desktop_promote_prod.yml", ".github/workflows/gcp_backend.yml", "desktop/macos/scripts/qualify-desktop-beta.sh", ".github/scripts/test_desktop_release_flow_contract.py", ".github/checks-manifest.yaml"]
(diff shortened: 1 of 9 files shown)


Git-on-my-level:
Blocking review for exact head `223089648ed85114fe07e7634c584d7a6bf3614a`:

1. `qualification-lease-command.sh` invokes `dev_harness.cli` without changing into the supplied `$worktree`. The CLI derives `repo_root` from `Path.cwd()`, but workflow qualification starts in the run-isolated `.../source` checkout while `$worktree` is the tag-pinned Swift-cache checkout. Acquire therefore records the wrong root; the cached-worktree harness state cannot later satisfy cleanup provenance. Run the CLI with cwd set to `$worktree`, and cover that boundary in the regression test.

2. The helper captures `2>&1` into `output` even on success, then emits it as the acquire capability. Any successful stderr warning makes `LEASE_JSON` non-JSON; acquisition succeeds but parsing fails before the cleanup context/trap is installed, leaving a lease behind. Preserve stdout as the sole successful capability channel and relay diagnostics only for failed commands; add a success-with-stderr test.

The M1-only runner, tag/SHA binding, evidence/digest gates, Beta/Stable separation, and credential boundaries otherwise remain unchanged by this focused diff.

Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

Git-on-my-level:
Blocking regression evidence on current head `4988979fcfd7212a4fc842def8b4b634ad6696fb`:

`bash desktop/macos/tests/test-qualification-lease-command.sh` fails locally on macOS with `FAIL: lease acquire failure was not relayed to stderr`. The helper itself emits the expected diagnostic; the fixture rejects it before its mode switch because its literal `$PWD` comparison sees macOS canonicalize `/var/...` to `/private/var/...` after `cd`. Normalize both the expected worktree and `PWD` (for example `cd "$worktree" && pwd -P`) in the fixture, then rerun the behavior test.

The focused repair remains otherwise appropriate. @codex address that feedback

Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

Git-on-my-level:
CI blocker on current head `2e7447e5722af6216896161009ad14926f8baa1b`: [Hygiene run 30197242475](https://github.com/BasedHardware/omi/actions/runs/30197242475/job/89780896473) fails `check-release-process-guards.py` because `desktop_qualify_beta.yml` no longer contains its required immutable checkout fragment `checkout --quiet --detach "refs/tags/$RELEASE_TAG"`. Restore the guard-compatible tag checkout boundary (or update the guard and its contract test coherently if the replacement is equally immutable), then rerun the focused checks and request review on the new head. @codex address that feedback

Git-on-my-level:
@codex review

PR #10660, opened 2026-07-26
Repair PostHog telemetry ownership and coverage

skanderkaroui:
## Summary

- repair mobile connection, pairing, vendor, purchase-intent, permissions, and BLE session telemetry at their authoritative state boundaries
- add server-authoritative account-deletion completion and failure outcomes
- restore Windows PostHog delivery and add macOS pairing telemetry
- add an analytics reachability tripwire, load-bearing event dictionary, and PostHog regression-alert contract

## Root cause

Telemetry emission was distributed across individual call sites rather than the state transitions that owned the underlying behavior. Analytics refactors and UI replacements therefore orphaned emitters without breaking compilation or tests. Windows also had a valid capture helper whose endpoint was blocked by the renderer CSP.

## Impact

Connection and pairing funnels now distinguish recurring connections from first-device conversion, use bounded vendor dimensions, and expose BLE session reliability. Account deletion has observable backend outcomes. Windows sessions reach the shared PostHog project again. CI now rejects new orphaned analytics methods and decreases in established call-site coverage.

Hardware purchase remains intentionally unemitted because this repository has no authoritative fulfilment source or order-to-user reconciliation. Live PostHog alerts are documented but not provisioned by this PR.

- INV-AUTH-1

The macOS launch change only removes an empty analytics wrapper call; it does not change authentication state, routing, persistence, or session behavior.

## Failure class (fixes)

Failure-Class: none

No commit in this branch has a `fix:` subject. CI's `pull_request.base.sha` is pinned to
this PR's creation point, so the validated range also spans main's own `fix:` commits
merged since then; this declaration is what the protocol asks for in that case.

## Verification

- `make preflight` after merging main — 95 checks passed
- backend selection over every changed backend file (routers/services/unit account-deletion) — 67 tests passed
- `xcrun swift build --package-path desktop/macos/Desktop` — Build complete
- `.github/scripts/check_analytics_reachability.py` and its unit tests — passed (6 tests)
- `cd app && bash test.sh` — 951 tests passed
- `cd app && app/scripts/analyze_ratchet.sh` — passed
- `cd desktop/windows && npm test` — 5,042 tests passed across 524 files; 24 skipped
- `cd desktop/windows && npm run typecheck` — passed
- `xcrun swift test --package-path desktop/macos/Desktop --filter DeviceProviderTests` — 17 tests passed
- backend focused unit selection — 63 tests passed
- backend account-deletion E2E file in isolation — 5 tests passed
- analytics reachability checker and its tests — passed
- Python and Swift formatting checks plus `git diff --check` — passed

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for taking on this telemetry cleanup. The direction looks useful: moving product analytics to the owning state transitions, adding coverage for the mobile/macOS/Windows emitters, documenting the load-bearing events, and adding the reachability tripwire are all the right kinds of guardrails.

I am not approving because this is still a draft and it touches privacy/security-sensitive telemetry, account-deletion observability, CI/check-manifest behavior, Windows CSP/PostHog delivery, and new analytics docs across a fairly large diff. Before merge, I’d want human maintainer review focused on:

- account-deletion telemetry: confirm the emitted PostHog properties stay bounded/content-free and that failure/completion events cannot leak raw user data or error text;
- the new analytics reachability check: confirm it is intentionally lexical, has acceptable false-positive/false-negative tradeoffs, and is wired into the checks manifest in the desired lane;
- the PostHog alert/docs contract: confirm the documented project/alert setup matches the production process, since the PR intentionally does not provision live alerts;
- Windows PostHog delivery/CSP: confirm the endpoint and publishable key usage are intended for the Windows renderer.

No urgent security issue found in this review pass; this is a human-review gate because of the sensitive surfaces and draft status, not a request to split the PR.



---

kodjima33 (approved):
Telemetry ownership + reachability coverage repair looks sound, but 2213 lines across CI checks and backend is well past the scoped-diff bar.

backend/services/users/account_deletion.py
```diff
@@ -142,7 +163,47 @@ def _required_failures_from_purge_result(purge_result: object) -> list[PurgeFail
...
+        properties.get('duration_seconds'),
+        properties.get('vectors_deleted'),
+        properties.get('recordings_deleted'),
+        properties.get('required_failure_count'),
+        properties.get('best_effort_failure_count'),
+        properties.get('failed_operations'),
+        properties.get('retry_count'),
+        properties.get('terminal'),
+    )
+    emit_posthog_event(uid, event, properties)
```
tianmind-studio: **[P1] Do not write the deleted user's UID back to PostHog.** This passes `uid` as PostHog's `distinct_id` for both failure and completion events; on the success path it runs after Firebase Auth and the authoritative Firestore wipe. PostHog documents API events as identified by default, and requires `$process_person_profile: false` for anonymous capture (https://posthog.com/docs/api/capture). So the new completion event retains or recreates user-linked third-party data immediately after account deletion. The regression currently checks only that `uid` is absent from `properties` while explicitly expecting it as the first argument. Please emit this as a non-user aggregate signal (for example, a fixed service identity with person-profile processing disabled) and assert that no deleted UID reaches the analytics client.

desktop/windows/src/renderer/src/lib/analytics.ts
```diff
@@ -10,14 +10,19 @@ const POSTHOG_KEY =
 
 export function trackEvent(event: string, properties: Record<string, unknown> = {}): void {
   const distinctId = auth.currentUser?.uid ?? 'anonymous'
-  void fetch(`${POSTHOG_HOST}/capture/`, {
+  void fetch(`${POSTHOG_HOST}/i/v0/e/`, {
```
tianmind-studio, line 15: **[P2] Keep the configurable ingestion host aligned with the CSP.** This request still uses the exposed `VITE_POSTHOG_HOST` override from `desktop/windows/.env.example`, but every renderer CSP added by this PR permits only `https://us.i.posthog.com`. An EU, self-hosted, or local test host therefore changes this URL and is then blocked by Chromium before `fetch` runs; the new test only pins the US default. Either make the host intentionally non-configurable, or generate/allow the exact configured origin and cover an override in the CSP regression.

tianmind-studio (changes requested):
Native Windows review on head `6830dc240`: the focused analytics Vitest suite passes 5/5, the new analytics-reachability fixture suite passes 6/6, the repository checker passes, and account-deletion tests pass 37/37. I found one account-deletion privacy blocker plus a smaller Windows CSP/config contract mismatch, detailed inline. Requesting changes for the deleted-UID emission; the rest of the reviewed paths and current CI are green.

Git-on-my-level (dismissed):
Thanks for the detailed telemetry repair work. The overall direction is strong: moving emitters to owning state transitions, restoring Windows PostHog delivery, documenting load-bearing events, and adding a reachability tripwire are all useful guardrails.

I’m requesting changes for one privacy/account-deletion blocker I independently verified in the current diff: the new account-deletion completion/failure telemetry still calls `emit_posthog_event(uid, ...)`, including on the success path after Firebase Auth deletion and the authoritative Firestore wipe. That can retain or recreate a person-linked PostHog identity for an account that is being deleted. Please change this to a non-user aggregate/service signal (or otherwise disable person-profile processing for this event) and add a regression test that asserts the deleted UID is not passed to the analytics client as `distinct_id` or in event properties.

Please also resolve the Windows PostHog host/CSP contract before merge: either make the ingestion host intentionally fixed to the CSP-allowed origin, or allow/generate the configured origin and cover an override in the CSP regression test.

Because this PR touches privacy-sensitive account deletion telemetry, CI/check-manifest behavior, Windows CSP/PostHog delivery, and a large cross-platform analytics surface, it still needs human maintainer review before merge even after these fixes.

---

undivisible:
### Review fixes + rebase

Addressed the CHANGES_REQUESTED privacy/CSP items and rebased onto current `main`.

**P1 — deleted UID no longer re-identified in PostHog**
- Account-deletion completion/failure telemetry uses service distinct_id `omi-service:account-deletion`
- Sets `$process_person_profile: false`
- Strips accidental `uid` / `user_id` / `distinct_id` keys from properties
- Regression: `test_deletion_telemetry_never_uses_deleted_uid_as_distinct_id`

**P2 — Windows PostHog host aligned with CSP**
- `POSTHOG_HOST` is fixed to `https://us.i.posthog.com` (matches renderer CSP connect-src)
- Removed runtime `VITE_POSTHOG_HOST` override so Chromium cannot block a divergent host
- Test asserts the pin + CSP allowlist

@Git-on-my-level @tianmind-studio ready for re-review when CI is green.

undivisible:
### CI fix-up on rebased head

Follow-up on the privacy/CSP repair:

- `black` formatted `account_deletion.py`
- Raised product-file line-count baselines for `users.py` / `storage.py` with justifications
- Ratcheted `OmiApp.swift` down to current line count

CI re-running on `e77a9b3b`.

Git-on-my-level (dismissed):
Thanks for the follow-up repairs. The two prior privacy/CSP concerns look directionally addressed on this head: account-deletion PostHog now uses a service distinct id with person-profile processing disabled, and Windows pins the ingestion host to the CSP-allowed PostHog origin.

I’m requesting changes because the current head is not merge-ready and the remaining failures are in the PR’s own guardrails:

- `analytics-reachability` fails in both Repo Checks and the desktop static gate. The checked-in baseline is stale against current macOS sources (`setSuggestionAssistantTelemetryCaptureForTests` is treated as a public orphan, integration connect call counts fell, and suggestion-assistant call counts grew). Please either update the baseline intentionally or adjust the checker so test-only hooks are classified correctly.
- Desktop Windows CI fails in `src/renderer/src/lib/analytics.test.ts`: the “not VITE_POSTHOG_HOST” assertion matches the explanatory comment in `analytics.ts`. Please make the regression assert runtime env usage rather than comments, or reword the comment/test so it verifies the intended contract without failing.
- GitHub currently reports this PR as `CONFLICTING`, so it also needs to be rebased/merged with current `main` before maintainer signoff.

Because this PR touches privacy-sensitive telemetry/account deletion, CI check-manifest behavior, Windows CSP/PostHog delivery, and broad cross-platform analytics docs, it should still keep `security-review` + `needs-maintainer-review` and get human maintainer review after CI is green. I’m not asking for a split; the scope is cohesive around PostHog telemetry ownership and coverage.



---

Git-on-my-level (commented):
Thanks for the latest follow-up. I re-checked this head and the previous automation blockers look resolved:

- the analytics reachability tripwire and its self-tests pass locally on `debf7711`;
- the Windows PostHog regression now asserts the runtime capture host instead of matching source comments;
- the PR checks in the captured context are green, and the earlier account-deletion/PostHog identity concern remains addressed by using a service distinct id with person-profile processing disabled and uid-like keys stripped.

I’m not approving because this still needs human maintainer signoff before merge: it touches privacy-sensitive account-deletion telemetry, CI/check-manifest behavior, Windows CSP/PostHog delivery, and new cross-platform analytics/alert documentation. I’m keeping the existing `security-review` and `needs-maintainer-review` gates for that review. This is a positive technical signal, not a request to split the PR; the scope is cohesive around PostHog telemetry ownership and coverage.



---

PR #10665, opened 2026-07-26
fix(desktop-windows): name the supported Node version instead of failing silently

aryanorastar:
## What

`pnpm test` in `desktop/windows` fails **29 test files / 129 tests** on a current Node — with the same code and the same lockfile that is **fully green on Node 22**. Nothing in the package says which Node is required: there is no `engines` field, and no `.nvmrc` anywhere in the repo, while all five CI workflows pin `node-version: 22`. So the breakage is invisible in CI and lands only on a contributor's machine.

## Why it happens

It is not the tests. **Node >= 24 ships its own experimental `localStorage` global**, which is `undefined` unless `--localstorage-file` is passed, and it takes precedence over the one jsdom installs. Every `// @vitest-environment jsdom` suite that touches localStorage then dies with:

```
TypeError: Cannot read properties of undefined (reading 'clear')
   23| beforeEach(() => {
   24|   localStorage.clear()
(node:47032) ExperimentalWarning: localStorage is not available because --localstorage-file was not provided.
```

That reads like a broken test, not a wrong runtime — which is exactly why it costs time.

## Fix

- **`engines.node` = `>=22.19.0 <23`.** The floor is real, not decorative: `@earendil-works/pi-ai` declares `engines.node >= 22.19.0`, so an older 22.x installs and is still unsupported.
- **A `pretest` guard** that names the required range, the running version, and the reason — so the first failure is one legible line instead of 129 assertions. It is wired at `pretest` because the symptom is a test-suite symptom.

## What I tried and rejected

`engine-strict=true` in `.npmrc`. It makes pnpm enforce the `engines` field of **every transitive dependency** — a far wider blast radius than this problem. In practice it rejected Node 22.16.0 over a dependency's floor while giving no hint that the test suite itself was fine. Declaring `engines` alone does not block anything under pnpm 10 (verified: install succeeded on Node 26), which is why the guard exists.

## Verification

Same tree, same lockfile — only the runtime changed:

```
Node 26.3.1   pnpm test  ->  29 files / 129 tests FAILED (localStorage undefined)
Node 22.23.1  pnpm test  ->  523 files / 5037 tests passed, 6 skipped

guard on v26.3.1   -> exit 1, explains the localStorage shadowing
guard on v22.16.0  -> exit 1, explains the >=22.19.0 dependency floor
guard on v22.23.1  -> exit 0
```

`v22.23.1` is what CI's `node-version: 22` resolves to today, so `pnpm test` in CI keeps passing — I verified the supported path explicitly rather than only the rejections, since a wrong guard here would break the lane for everyone.

## Scope

`desktop/windows` only: one `engines` field, one `pretest` entry, one guard script. No test or product code touched.

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10665?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/package.json
+++ b/desktop/windows/package.json
@@ -3,6 +3,9 @@
   "version": "1.0.20",
   "description": "An Electron application with React and TypeScript",
   "main": "./out/main/index.js",
+  "engines": {
+    "node": ">=22.19.0 <23"
+  },
   "author": "Based Hardware",
   "homepage": "https://github.com/BasedHardware/omi",
   "scripts": {
@@ -34,6 +37,7 @@
     "build:win": "npm run build && electron-builder --win --x64 --config electron-builder.config.mjs --publish never",
     "build:mac": "electron-vite build && node scripts/bundle-pimono-extension.mjs && electron-builder --mac --config electron-builder.config.mjs --publish never",
     "build:linux": "electron-vite build && node scripts/bundle-pimono-extension.mjs && electron-builder --linux --config electron-builder.config.mjs --publish never",
+    "pretest": "node scripts/check-node-version.mjs",
     "test": "vitest run",
     "test:watch": "vitest",
     "fixtures:audio": "node scripts/gen-audio-fixtures.mjs",
--- a/desktop/windows/scripts/check-node-version.mjs
+++ b/desktop/windows/scripts/check-node-version.mjs
@@ -1,0 +1,39 @@
+// Fail the test run early, and legibly, when Node is outside the supported range.
+//
+// Without this the failure is silent and deeply confusing: on Node >= 24 the
+// runtime ships its own experimental `localStorage` global that is `undefined`
+// unless `--localstorage-file` is passed, and it takes precedence over the one
+// jsdom installs. Every `// @vitest-environment jsdom` suite that touches
+// localStorage then dies with `Cannot read properties of undefined (reading
+// 'clear')` — 29 test files and 129 tests on Node 26.3.1, all green on Node
+// 22.16.0 with the same code and the same lockfile. CI pins Node 22, so the
+// breakage is invisible there and only ever hits a contributor's machine.
+//
+// The floor is not cosmetic either: `@earendil-works/pi-ai` declares
+// `engines.node >= 22.19.0`, so an older 22.x installs but is unsupported.
+//
+// Kept as a `pretest` hook rather than `engine-strict=true` in .npmrc: that flag
+// makes pnpm enforce the engines field of every transitive dependency, which is a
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Thanks for tracking this down — this is a useful, narrowly scoped developer-experience fix for the Windows desktop test lane.

I reviewed the diff statically and did a small local validation of the guard logic. The `engines.node` range matches the locked `@earendil-works/pi-*` packages' `>=22.19.0` floor, the upper bound matches the current CI Node 22 lane, and the `pretest` hook fails before Vitest with a much clearer message on unsupported runtimes. I also sanity-checked the script behavior for 22.16.0, 22.19.0, 22.23.1, 24.0.0, and 26.3.1; it accepts the supported 22.19+ range and emits the localStorage-specific explanation for Node 24+.

I’m leaving this as a positive signal rather than a formal approval because it changes package metadata/test execution behavior and should get human maintainer sign-off before merge. No product behavior concerns from my review.

---

tianmind-studio:
Native Windows evidence for the lower-bound guard: Node `v22.15.0` produces another large, misleading failure class before the declared `>=22.19.0` floor.

On current `main`, the complete desktop test run reported:

```text
1,843 suites; 5,070 tests
4,971 passed; 81 failed; 18 skipped
```

All 81 failures were the same root cause across five database/FTS suites: the older runtime's `node:sqlite` reports `Error: no such module: fts5`. There were no other failures after fixing the separate locale-dependent assertion in #10713.

So this PR's early rejection of 22.15 is useful beyond the dependency-engine declaration: it replaces 81 apparent database regressions with one accurate version error. No change requested; this is additional native-Windows validation of the chosen floor.

tianmind-studio (approved):
Reviewed on native Windows against current `main`. I verified that `pnpm test` does execute the `pretest` lifecycle hook and, under Node v22.15.0, exits before Vitest with the declared `>=22.19.0 <23` requirement and the running version. The PR merges cleanly with current `main`, `git diff --check` passes, and all GitHub checks are green. The scoped guard matches the documented CI/runtime constraint.

kodjima33 (approved):
Verified: no engines field or .nvmrc existed while all CI workflows pin Node 22, so the jsdom localStorage breakage on Node >=24 only ever hit contributors' machines; adds engines + a pretest guard, no runtime behavior change.

Issue #10351, opened 2026-07-22
Release-process guard self-tests: 9 failures on main and not wired into any CI lane


kodjima33:
9 of 87 tests in `.github/scripts/test_check_release_process_guards.py` fail on current main (5a4dffa8cf), e.g. `test_global_document_lock_rejects_every_codemagic_mutation[top-level-field]` and `test_global_document_raw_lock_rejects_semantically_equivalent_yaml_rewrite` — full list from `python3 -m pytest .github/scripts/test_check_release_process_guards.py -q`. The guard itself (`check-release-process-guards.py`) runs in repo-checks.yml and passes; its pytest self-suite runs in no CI lane (not in checks-manifest.yaml, not in any workflow), so these failures accumulate silently. Verified identical failures on a clean checkout with no local changes. Either fix the 9 tests and register the suite in `.github/checks-manifest.yaml`, or delete the dead tests — an unwired, failing guard test suite is worse than none.

Git-on-my-level:
Root cause, verified on `main` today (still 9 failed / 78 passed, still in no manifest entry and no workflow reference):

**The suite runs a historical revision of the guard against present-day input.** `_load_parent_guard` does `git show <revision>:.github/scripts/check-release-process-guards.py` and executes it (`test_check_release_process_guards.py:49-62`). `REVIEWED_PARENT = 4e391ee7` hardcodes `"canonical workflow must retain exactly 21 approved script steps"`. The current guard expects **22**, because INV-BETA-1 added `Create Omi Beta variant` — and `omi-desktop-swift-release` in today's `codemagic.yaml` has 22.

So the parent rejects whatever it is handed, and every assertion of the form `parent.check_codemagic_release_publishers() == []` fails on an error that has nothing to do with what the test is probing. All 9 failures are that one mismatch:

```
E  AssertionError: ['canonical workflow must retain exactly 21 approved script steps']
```

The synthetic fixture makes it unavoidable: `_parent_bypass_workflow` (line 123) builds a 2-step workflow, so any parent enforcing a step count rejects it. That is why the `RAW_SCANNER_PARENT = 11ac6d9e` cases pass — that revision has no step-count rule at all.

**This will recur on the next approved step.** Pinning a historical checker against evolving current data has no stable state; the same 9 tests broke the moment step 22 landed and will break again at 23.

**The fix is a semantics decision, which is why I did not make it.** Two coherent options:

[...]

PR #10667, opened 2026-07-26
fix(ci): wire the desktop release manifest schema guard into a lane

aryanorastar:
## What

`.github/scripts/test_desktop_release_manifest_schema.py` **ran nowhere.** No entry in `.github/checks-manifest.yaml`, no workflow step, no reference anywhere under `.github/`:

```
$ grep -rn "test_desktop_release_manifest_schema" .github/ | grep -v '^.github/scripts/test_desktop_release_manifest_schema.py'
(no output)
```

A guard with no blocking audience is a dead check — the same class #10351 filed, and what AGENTS.md means by *"On-demand scripts and scheduled jobs with no blocking audience are dead checks."*

## Why it could not have been wired as written

It imported `jsonschema` and relied on pytest, and **no lane installs either** — it was 1 of only 2 suites in that directory that fail to run on a stock interpreter. Every other self-suite there is unittest-style and stdlib-only, invoked as `python3 <script>`. Its own sibling `test_desktop_release_manifest.py` **is** wired, and needs nothing.

## Why it matters

`desktop-release-manifest-v1.schema.json` is what a published **Stable** manifest is validated against. The absence of `beta_zip_url`/`beta_dmg_url` is what stops a beta build from being describable as Stable — load-bearing, not tidiness.

## Change

- **Ported to stdlib.** Both `$defs` it exercises are plain `{"type": "string", "pattern": "^...$"}`. JSON Schema `pattern` is a regex *search* and these are anchored, so `re.search` validates them exactly as `Draft202012Validator` did. `_matches` asserts the subschema really is that shape, so a richer one fails loudly rather than being silently under-validated.
- **Converted to unittest**, so `python3 <script>` runs it — matching the directory convention.
- **Added the rejection cases it lacked.** The original only asserted the canonical Stable URLs are *accepted*, so loosening the pattern to admit any artifact name passed it. It now also proves a beta-named artifact and a foreign host/repo are refused.
- **Registered** as `desktop-release-manifest-schema-v1` on the `local` and `ci` lanes, triggered by the schema and the suite.

## Verification

```
/usr/bin/python3 (3.9.6, no deps installed)   -> 5 tests, OK      (was: unrunnable)
mutation: add beta_zip_url to the schema      -> FAILED (1)
mutation: loosen zipUrl to accept any .zip    -> FAILED (2)        <- old suite PASSED this
schema restored                               -> OK, file unmodified
test_run_checks.py (manifest contract)        -> 26 passed
run_checks.py --lane local                    -> SELECTED and PASS
```

The second mutation is the one that matters: it is a real weakening of the release contract that the previous version of this suite could not catch, because it only tested the happy path.

## Scope

Two files: the suite and one manifest entry. The schema itself is untouched (verified `git status` clean after the mutation runs).

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10667?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -69,6 +69,11 @@ checks:
     triggers: [".github/schemas/desktop-release-manifest-v1.schema.json", ".github/scripts/desktop_release_manifest.py", ".github/scripts/test_desktop_release_manifest.py", ".github/scripts/fixtures/desktop_release_manifest/**"]
     lanes: ["local", "ci"]
     reason: "immutable desktop app/backend release-unit contract"
+  - id: desktop-release-manifest-schema-v1
+    command: ["python3", ".github/scripts/test_desktop_release_manifest_schema.py"]
+    triggers: [".github/schemas/desktop-release-manifest-v1.schema.json", ".github/scripts/test_desktop_release_manifest_schema.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the Stable manifest schema must keep admitting exactly one app artifact pair; a beta_zip_url re-entering it is how a beta build becomes describable as Stable"
   - id: desktop-release-doctor-v1
     command: ["python3", ".github/scripts/test_desktop_release_doctor.py"]
     triggers: [".github/schemas/desktop-release-evidence-v1.schema.json", ".github/scripts/desktop_release_doctor.py", ".github/scripts/desktop_release_doctor_report.py", ".github/scripts/test_desktop_release_doctor.py", ".github/workflows/desktop_release_doctor.yml"]
--- a/.github/scripts/test_desktop_release_manifest_schema.py
+++ b/.github/scripts/test_desktop_release_manifest_schema.py
@@ -1,22 +1,85 @@
+#!/usr/bin/env python3
+"""The desktop release manifest schema admits exactly one app artifact pair.
+
+`desktop-release-manifest-v1.schema.json` is what a published Stable manifest is
+validated against. The invariant it carries is that a Stable manifest describes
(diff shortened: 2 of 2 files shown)


Git-on-my-level (changes requested):
Thanks for tightening this guard. Wiring the schema self-suite into the manifest and making it stdlib/unittest-runnable is the right direction, and I verified locally that `python3 .github/scripts/test_desktop_release_manifest_schema.py` passes on this head.

I do need one small change before this can merge: `_matches()` says it fails loudly if either `$defs.zipUrl` or `$defs.dmgUrl` becomes richer than a plain string+pattern schema, but it currently only asserts `type == "string"` and then ignores any additional schema keywords. If someone later adds something like `minLength`, `format`, `allOf`, etc., this test would still silently under-validate while claiming to mirror the schema. Please either assert the exact supported shape, e.g. `set(subschema) == {"type", "pattern"}` plus `isinstance(pattern, str)`, or use the real JSON Schema validator in a lane that installs it.

I’m also not formally approving because this touches CI/check routing and the current check summary is not fully green, so it needs maintainer/workflow sign-off after the above is fixed.

---

kodjima33 (approved):
Verified: test_desktop_release_manifest_schema.py had no entry in .github/checks-manifest.yaml and no workflow reference, so the guard never ran; this only wires the existing test into a lane.

PR #10472, opened 2026-07-24
feat make Gmail a live integration over the shared Google grant

krushnarout:
## What

Makes **Gmail a live integration** in Settings → Integrations on both mobile and web. It was previously hardcoded "Coming Soon" even though the backend Gmail tool was already built and registered.

Gmail rides the **existing `google_calendar` OAuth grant** rather than getting a grant of its own — the reuse-first approach already outlined in the repo's Gmail+Calendar OAuth migration plan.

## How

**Backend**
- Add `gmail.readonly` to the requested Google scopes.
- The OAuth callback records `granted_scopes` (what Google *actually* approved).
- `DERIVED_INTEGRATIONS` routes the `gmail` app key's status / oauth-url / delete through `google_calendar`.
- Gmail reads **connected only when the grant carries `gmail.readonly`** — so a legacy Calendar user shows Gmail as disconnected (needs one reconnect) while Calendar keeps working.
- `get_gmail_messages_tool` refuses to call Google without the granted scope, turning the opaque Google 403 into a clear reconnect message.

**App**
- New `GmailService` (rides `google_calendar` OAuth via a new `BaseIntegrationService.oauthAppKey` seam).
- `IntegrationProvider` fetches Gmail status from the backend instead of forcing `false`.
- Connect / disconnect wired; disconnect dialog names both Gmail and Calendar (shared grant) and reloads all rows.

**Web**
- Drop Gmail's `coming_soon` flag; backend routing makes the existing connect/disconnect/status calls work unchanged.

## Rollout note

Existing Google Calendar users have a stored grant with no `gmail.readonly` scope, so they'll see Gmail as disconnected and must tap Connect once to re-consent (Calendar keeps working throughout). This is intended — it's what prevents the silent 403. Requires `gmail.readonly` published on the OAuth consent screen; the restricted-scope verification / CASA assessment is complete.

## Testing

- New `backend/tests/unit/test_gmail_scope_gating.py` (9 tests) + updated `test_async_resource_correctness.py` for the scope gate — green via `test.sh`.
- `bash app/test.sh` fully green; `flutter analyze` clean on changed files; async-blocker scan clean; web prettier clean.
- **Not** exercised against live Google OAuth (needs deployed credentials + a Google-connected account) — verified via hermetic tests only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #10438, opened 2026-07-23
Track 0 submission: native macOS shell, consistent and improved UI/UX

katipally:
**This is my submission for the Omi Summer challenge (Track 0).** I already submitted for Track 1 (Agents) #10007  #10088  and Track 3 (the notch) #10247  #10332 . This time I went at the main macOS desktop app itself. The app already does a lot; what stood out is that the shell and the everyday UI did not feel native, and the controls were different on every page. So this PR is a native UI/UX pass over the whole app plus a calmer, landing-first Home. It is cosmetic and additive: nothing about what the app can do was removed, and chat stays one shared timeline.

## The problem I found

The desktop app can do a lot, but it never quite read as a native Mac app, and that showed up in small ways all across the surface.

The shell was the first thing. The window carried the standard title bar over a floating card, so there was a visible seam and a black gap along the top instead of one clean, edge-to-edge surface.

The controls were inconsistent, both with the platform and with each other. The Settings control sat on the right of the header, but the panel it opened slid in from the left, so the thing you clicked and the thing that appeared were on opposite sides, and it read as disorienting every time. The Capture and Listening chips next to it were all slightly different sizes, and Listening grew whenever you hovered it, so the header never quite sat still. Deeper in, every page had drifted into its own dialect: the search fields, toggles, and chips on Memory, Tasks, and Apps were each shaped and styled a little differently, so it rarely felt like the same app twice.

A few surfaces had their own rough edges. Modals opened as large half-empty boxes, and the blur behind them was applied per section, so the focus effect looked patchy rather than deliberate. Rewind was a dead end with no way back to the tab you came from. And the Home composer was oversized and sat a little out of place against everything around it.

Home itself was the biggest one. It dropped you into your full chat history the instant it launched, even though it is one continuous conversation and there is seldom a reason to see all of it right away.

None of it was broken. It just added up to an app that felt assembled rather than designed. This pass pulls it back into one native shell, gives every page the same set of controls, settles the header, and lets Home open calm. There is a lot here, more than a diagram can usefully show, so the demo below is the best way to see it end to end.

## Demo (under 2 minutes)

https://github.com/user-attachments/assets/34291c47-e05f-469c-92b5-76f7f7208c39

## What changed, and why it feels right

- **A native shell.** Hidden title bar, content runs edge to edge, no floating card and no top seam. The whole window shares one background so there is no visible break between sections, and you can drag it from the top bar.
[...]

PR #10672, opened 2026-07-26
Track 3 submission: a full rework of the macOS desktop app

katipally:
**Track 3 submission: a full rework of the macOS desktop app.** It carries forward my earlier Track 3 (#10247, #10332) and Track 0 (#10438) work and reworks the rest of the app on top of it.

## Demo

https://github.com/user-attachments/assets/2ca5f0e3-6cab-481e-8af0-c23917148d77

---

## 1. The notch

Upstream has no notch. Its floating surface is `FloatingControlBarWindow`: a draggable, resizable pill that animates out of a point near the top of the screen and holds voice, a text input, notification cards and agent pills.

This branch replaces it with a real notch. A shaped, non-activating `NSPanel`, one per display, anchored to the physical camera housing.

### What the notch shows

| State | What you see | Driven by |
|---|---|---|
| Closed | Omi mark, camera gap, settings gear | always visible, hugs the housing |
| Listening | the 8 Omi dots as a live mic waveform | your input level |
| Thinking | the same 8 dots reformed into the rotating logo ring | turn state |
| Speaking | the dots as a waveform again | `playbackLevel`, the RMS of each TTS chunk |
| Reply | text revealed a word at a time, then it lingers | paced reveal, hover pauses it |
| Hint | a readable strip under the chrome | PTT errors, mic unavailable |
| Notification | a card sized to its own content | the notification queue |

One `NotchPresentation` value decides which rung you are on. The panel size (`NotchViewModel.size(for:)`) and the rendered content both switch on that same value, so they can never disagree. The ladder is listening → thinking → responding → hint → notification → idle, scoped to the display the turn started on.

### How it behaves

| Behavior | Detail |
|---|---|
| One mark, three states | The same 8 dots morph between waveform, ring and waveform. One persistent orb layer for the whole turn, so a phase change animates geometry instead of swapping views. At rest the ring uses the exact `NotchOmiMark` ratios, so it is the logo. |
| Real speaking level | `AudioLevelMonitor.playbackLevel` is new, fed from `StreamingPCMPlayer` as the RMS of each enqueued TTS chunk. The app produced no assistant output level before, so a speaking waveform could only have been a canned loop. A zero is never throttled, so playback stopping lands immediately instead of freezing the bars. |
| Grows, never pops | The island is its own layer with no conditional content inside, so its SwiftUI identity never breaks and its frame always interpolates. Content crossfades on top, clipped to the same shape. |
| Two timelines, never merged | A spring for discrete morphs, a `.smooth` for continuous height growth. Keying the morph spring on a measured height oscillates the measure → resize → remeasure loop. |
| Height only | Width is fixed. Height grows to a cap of 40% of the screen, then the reply scrolls and follows the tail. |
| Reply lingers | The reply is captured while it streams, so the linger is already true the instant the response ends. No one-frame collapse to idle. Hover pauses dismissal. Esc dismisses through window-level monitors, since a non-activating panel is never key, and the monitor returns the event rather than swallowing it. |
| Click away to dismiss | `NotchOutsideClickDismiss` is one instance for the whole app, armed only while something dismissable is up. Global and local mouse-down monitors: the global one sees clicks in other apps, the local one sees clicks in Omi's own windows. Mouse-down monitors need no Accessibility permission, so this is never silently inert. |
| Right-click | snooze for two hours |
| Agent status | `AgentStatusGroup` is extracted out of `FloatingControlBarView`, and the mark's dots take the aggregate agent colour, so a running or failed agent is visible without opening anything |
| Accessibility | notch text honours the system font scale through `scaledFont`. `JustifiedText` is AppKit and cannot take a modifier, so `StreamingReplyText` reads `fontScale` and applies it to the `NSFont` size. |
| Reduce Motion | swaps the whole timeline for a short ease one level up, rather than nulling the animation, which would make a panel that grows out of the housing snap |

### Notification cards

| | Before | Now |
|---|---|---|
| Where they render | the floating bar | the notch, all 5 kinds: reach error, task receipt, end of conversation, task and proactive |
| Size | a fixed 430pt `notificationWidth` | measured from content, behind a 4pt jitter filter, on the smooth timeline |
| Buttons | in an overlay above a reserved-width guess, so a wide button ran the title under itself | in the layout, so nothing can run under them in any of the 5 variants |
| Dismissal | everything timed out at 6s, with no notion of a card that needs an answer | cards that need a decision (reach error, task receipt, end of conversation) never auto-dismiss; the rest still time out at 6s. That is a property of the notification, keyed on the same `assistantId` the card switches on, so a variant cannot render buttons that silently disappear. |

### The floating bar it replaces

Nothing in `Sources` constructs a `FloatingControlBarWindow` any more. Only `AgentPillLifecycleTests` still builds one, which is what keeps the retired class covered while it waits to be deleted.

| Moved to the notch | Note |
|---|---|
| show / hide / toggle / snooze / temporary-show / launch presentation | |
| the proactive notification queue | keeps its owner gating and the temp-show re-hide from #6972 |
| `barState` | the PTT reducer's projection now drives the notch ladder |
| the non-hub voice route | it used to require a window, which would have made PTT silently dead. It routes with no window at all and mirrors the streaming reply, so the batch and STT routes render exactly like the hub route. |

`NotchScreenManager` runs no pointer tracking, so idle CPU is zero.

Two entry points had nowhere left to go:

| Entry point | Now |
|---|---|
| Home / Dashboard suggestion chips | send to the main window's chat |
| Agent drill-in | opens `AgentTranscriptSheet`, a read-only transcript over the shared `ChatMessagesView` with the pill's status and a stop control while it runs. Agent transcripts were readable **only** in the floating bar before, so the main window could link to an agent it had no way to display. `activeAgentChatPillID` is set so viewed-pill expiry cannot remove an agent out from under the reader. |

---

## 2. The shell

| | Before | Now |
|---|---|---|
| Window | standard title bar over a floating card, visible seam and a black gap at the top | `.hiddenTitleBar`, content edge to edge, one background across the whole window, draggable from the top bar |
| Navigation | a pill per tab | a two-row header: Omi mark and name centred on top, the tabs as one sliding segmented track below |
| Settings | a control on the right that opened a panel from the left | a chip on the left near the traffic lights, opening a frosted behind-window sidebar with a Back-to-app chip and a sliding highlight |
| Modals | a bool per call site, blur applied per section | `ModalPresentationState` owns which modal is up; the shell frosts the whole window uniformly, owns the backdrop accessibility contract, and dismisses on click-outside or Esc |
| Modal size | large half-empty boxes | fit their content, cap their height, scroll past the cap |
| Rewind | a dead end | a Back chip returns you to the tab you came from |

### Shared components

Seven components replace the chrome each page used to re-derive. Home, Dashboard, Apps, Memories, Tasks, Conversations, Chat, Rewind and Settings all render these, which is where most of the deletion comes from.

| Component | Replaces |
|---|---|
| `OmiSegmentedControl` | per-page section switchers, icon toggles |
| `OmiSearchField` | each page's own search field; focus is owned by the field, not by the page |
| `OmiChip` / `OmiFilterChip` | per-page chip shapes |
| `OmiIconButton` | ad-hoc header buttons |
| `OmiModal` | per-page modal containers |
| `OmiPageHeader` | per-page header rows |
| `OmiFloatingStatusPill` | status pill variants |

`OmiChrome.controlHeight` (40) is the one height every control that can sit in a header row shares. A row of mixed controls only sits on one line if they agree on a height, and each page was picking its own. `SearchableDropdown` moved out of `Pages/Settings/Components` into the shared `Components` directory, since it is used outside settings now.

---

## 3. Home

| | Before | Now |
|---|---|---|
| On open | dropped you into the full chat history | rests on the hub: a rotating Omi mark, a greeting, the composer, suggested prompts. History is one scroll or one send away, never auto-revealed. |
| Hero → chat | the hero cut out | hero and transcript share one stage slot and cross-fade in place. The landing is seeded from session state, so a tab switch, which rebuilds the page, does not replay it. |
| Composers | one per branch: hub, chat, connect | one composer. `HomeComposerPlacement` resolves its position from the stage mode alone, so focus and draft text survive the stage changing. |
| Connect tray | disappeared for good after your first message | `keepsConnectWhileEmpty` keeps it in the trailing slot; the composer holds first responder across a send |
| Composer height | the trailing slot held four differently sized views, so clicking into the field dropped the bar 4pt under the pointer and broke the capsule radius | all four modes reserve one height; the empty mode renders a zero-width `Color.clear` rather than `EmptyView`, which contributes no height |
| Composer animation | keyed on `isFocused` and `canSend` separately, leaving `isSending` uncovered, so Connect → Stop snapped | keyed on the mode itself, so every transition between the four is covered by construction |
| Composer width | reserved a hardcoded 210pt of chrome, tuned against a mode the bar is not in once you are typing | derives from the slot's own constant |

`HomeStatusControls` and `HubPages` pull the status row and the hub tab surfaces back out of `DesktopHomeView`. The Memory hub's segmented control is driven by local state with the stored key mirrored behind it: an `@AppStorage` write round-trips through UserDefaults and republishes on a later runloop turn, which lands outside the `withAnimation` transaction and makes the capsule jump instead of slide. The persisted destination is read directly rather than seeded in `onAppear`, so revisiting Memory no longer flashes Memories first.

---

### The reply is prose now, not a bubble

| | Before | Now |
|---|---|---|
| Omi's reply | a bordered bubble with a 42%-opacity fill, in a 32pt avatar gutter | plain text on the canvas. No bubble, no border, no avatar gutter, full column width. |
| Avatar | `SBLogo` spinning inside each streaming reply | removed. One permanent mark at the foot of the transcript instead. |
| Long replies | truncated at 500 characters behind a "Show more" button | never truncated. A reply is read, not unfolded. |
| The user's turn | a bubble capped at a flat 640pt | a pill at 75% of the message column, measured from its own row (`containerRelativeFrame` resolves against the `ScrollView`, so once the scroll view spans the shell it would have sized the pill against the whole window) |
| Typing dots | `TypingIndicator` shown per message, for app personas only | the transcript's own working row carries the wait for every sender, so an empty streaming message renders nothing |

### The Omi mark, while thinking and while working

The mark sits at the foot of the transcript as one element for the whole life of the surface. It is never mounted and unmounted around a turn: at rest it turns slowly, the same lap the landing hero makes, so Omi reads as present rather than as a spinner that appears and vanishes. The same 8 dots carry every state and every state resolves back to the ring, so the logo stays legible.

| Motion | When | What it does |
|---|---|---|
| rest | no turn open | the ring turning slowly |
| `gather` (1.6s cycle) | thinking, reading, searching, and any tool not recognised | the ring condenses to a knot and blooms open |
| `wave` (2.4s cycle) | writing, editing, bash, spawning or running an agent | the ring unspools into a travelling wave |

Both motions start and end on the full ring, so the mark can change motion at a cycle boundary with no cut. Beside it, a status label says what Omi is actually doing: an in-flight tool names itself (`ChatContentBlock.displayName`), and with no tool running it reads "Thinking". The label shimmers with a gradient masked by a second copy of the text, so it lights the letters rather than a rectangle behind them, and the shimmer is skipped entirely under Reduce Motion.

### Layout and motion

| | Before | Now |
|---|---|---|
| Scrollbar | the scroll view was capped to the readable column, so the macOS overlay scroller floated hundreds of points inside a wide window | the cap moved onto the column inside the scroll view; the scroller sits on the window edge where the platform puts it |
| Composer | sat below the transcript with 26pt of unused gap, and the transcript's height was a function of the composer's, so a second line resized the surface above it | floats over the transcript, which reserves exactly its measured height at the foot |
| Bottom fade | fractional stops, which re-scale whenever the panel resizes, so the band visibly breathed while typing | fixed bands, and everything below the composer's top edge is masked out entirely, since the composer fill is translucent |
| Streaming | tokens rendered as they arrived, so markdown flashed as raw text mid-stream | the reply reveals on its own clock, holding back an unclosed inline marker or a marker-only line |
| Following the reply | commanded `scrollTo`, which assigns an offset outright and lands on top of a move already in flight | `ChatLiveEdgeFollower` integrates the offset toward the live edge every display frame, and outlives the last token by a settle tail because the reveal is still catching up |
| Sending | an un-animated `scrollTo` at t=0 and another at t+100ms, 100ms into a 380ms spring | one commanded move inside the insert's own transaction, plus a single settle after the spring finishes, skipped when already at the live edge. The settle exists because a row's final height is unknown at insert time: markdown, tool cards and attachments lay out later. |
| Row spacing | a flat gap between every row, so a question and its answer looked as unrelated as two separate exchanges | a user message following Omi opens a new turn and takes the wider gap (32pt); everything else continues the turn (12pt) |
| Dedup | fingerprinted every message, including streaming ones, and recomputed inline in `body` | streaming rows are skipped, since a live reply passes through every prefix of its final text. Held in state, because `body` re-evaluates on every frame of a paced reveal. |
| Agent projection | rebuilt the whole message array on every frame | returns early when the transcript holds no lifecycle receipt |
| Markdown parsing | parsed in `init`, so an `EquatableView` paid for the parse on every parent render before `==` could skip `body` | parsed in `body`, which is what the equatable boundary exists to protect |

### The prompt timeline

A long chat had one way back to an earlier question: scroll and read. The gutter beside the transcript was empty the whole time.

| | |
|---|---|
| What | each of your prompts marked at its fraction of the document height |
| Hover | shows the prompt and the reply it drew |
| Click | jumps to it |
| Size | a 40pt rail against 2pt marks, because a hairline is not a hover target |
| Mounts | only where the gutter is at least 56pt, so a rail centred in it still clears the overlay scroller |

The jump-to-latest chip now stacks on the measured composer height (`ChatJumpButtonPlacement`). Surfaces that dock their composer below the transcript report 0 and keep their existing gap.

---

## 5. Capture and Listening

The shipped default is `.onlyDuringMeetings`, and the capture gate genuinely stops the microphone outside a call. The chip was painting `appState.isTranscribing`, which is the toggle, not whether audio is flowing.

| | Before | Now |
|---|---|---|
| Source of truth | `appState.isTranscribing` | the same flag the capture gate sets, so the pill and the microphone cannot drift |
| States | on / off / blocked | `capturing` / **`standby`** / `off` / `blocked`. The new `standby` case is listening switched on with the meeting gate shut: no microphone, no audio. It is monochrome and brighter than off, with its own icon and badge, and maps to `.inactive` deliberately, because the shared chip paints `.active` green and green must mean audio. |
| Off appearance | no fill and a zero-opacity stroke, so it read as absent | a visible off treatment |
| The mode | implied, never shown | all three modes one hover away, through the same overlay the Capture chip uses for Rewind, so the bar never reflows |
| Non-pointer access | a two-way toggle reaching two of three modes | right-click and a named accessibility action reach all three. Hover is invisible to VoiceOver and Full Keyboard Access, so the reveal is additive. |
| On failure | a much longer title swapped in and shoved the cluster sideways | the failure lives in the icon and the tooltip |
| Settings copy | "Only during meetings", without saying the microphone stops | says it stops |
| Hover panel | anchored at the chip's centre, so half its extra width spilled past the window edge | `HoverRevealAnchor.chipTrailing` grows it inward; the unit point and the overlay alignment come off the same case, so they cannot describe different placements |

---

## 6. Apps, Tasks, Memory, sheets

| Surface | Before | Now |
|---|---|---|
| Apps scope | an on/off "Installed" chip; turning it off ran `clearFilters()`, which also dropped your search text and category | an All / Installed segmented control. `AppProvider.applyInstallScope` writes only the scope; `clearFilters()` stays for the fresh-catalog presentations that want a reset. |
| Apps grid | installed and not-installed cards differed by one word: same fill, text, border and 60x28 frame | installed cards carry a mark and read ghost against filled, matching the treatment `ImportConnectorActionButton` already used for connectors |
| Failed install | `AppProvider.errorMessage` was set and read by nothing anywhere in the app, so the spinner just stopped and the button still said Install | has a rendering path, and the message names the app and states a next step, so a failure buried under a grid of cards still says which card it was. A retry or a new search clears the previous failure first, so a stale banner cannot sit over fresh results. |
| Apps header | segments could be compressed away at narrow widths | the row is pinned by a layout test at three window widths |
| Tasks | a checkmark-circle icon toggle that gave no indication of which list you were on | To Do / Done as a labelled two-segment switch |
| Memory / Conversations | the scroll view was capped and centred, parking the scrollbar inside a wide window | the column is capped, not the scroll view. `MemoryHubLayoutPolicy.contentWidths` returns both values so a caller cannot set one and forget the other. An open transcript drawer yields the cap. Brain Map stays uncapped. |
| Conversation detail | App Insights and Try-with-Apps built inline in `ConversationDetailView` | moved to `ConversationAppInsightsCards.swift` as leaf views driven by their inputs, keeping the page under the 1,500-line ratchet threshold. Nothing in the new file reaches back into it. |
| Sheets | opened on a full row holding nothing but a close glyph, while the title repeated inside the scroll body and scrolled away | `OmiSheetHeader`: what the sheet is on the left, the way out on the right, a divider under both. `title` is non-optional, so a header cannot render nothing. `DismissButton` moved into `SheetChrome`, so the sheet chrome owns its own accessibility contract. |
| Search fields | each page shaped its own | one `OmiSearchField` across Chat history, the sessions sidebar, Tasks, Memory and Apps, brightening its fill and ring while focused, with an override for placeholders that read poorly aloud |

---

## 7. Settings: Integrations

Upstream made Gmail a live integration over the shared Google grant on mobile and web (#10472). The desktop had no way to grant it, so `get_gmail_messages_tool`, the tool behind "Checking Gmail", had nowhere to send a user whose grant was missing or predated `gmail.readonly`.

| | |
|---|---|
| What is new | a Settings → Integrations section that connects and disconnects Gmail and Google Calendar |
| What this is not | the Apps → Imports Email row, which pulls history in over the local browser session and is unchanged. These are the grants the backend's own tools authorize against. |
| The model | mirrors the app's `IntegrationApp` so the two clients cannot drift. Gmail has no grant of its own, rides `google_calendar`, and reads connected only when that grant carries the Gmail scope. |
| Legacy users | as upstream does, there is no separate reconnect state. A legacy Calendar user reads Gmail as not connected, and Connect re-runs Google OAuth, which upgrades the grant. |
| Why it polls | the OAuth callback lands on the backend, not on this app, so there is no deep link to wait on. `connect` polls until the backend reports the row connected, the same signal mobile gets by reloading. |
| Why both rows re-read | they answer from one grant. Reading back only the row the user touched is how Calendar would sit there showing Connected after a Gmail disconnect dropped it. One flow runs at a time for the same reason. |
| Apple Health | omitted. It is HealthKit, with no macOS surface. |
| Logging | booleans only. Which grants a user holds is the diagnostic; the grant itself never goes near a log line. |

---

## 8. Defects fixed along the way

| Defect | Cause |
|---|---|
| The recording indicator could read paused while the microphone was live | `reconcileCapture` published `isAwaitingMeeting` above the `meetingStateReady` guard, and the guard returns before the block that stops the mic. For a privacy indicator that is the worse direction to be wrong in. |
| The Connect tray sat under the floating composer | making the stage a ZStack bottom-aligned every arm, but only the transcript got the composer's height back as an inset. The tray's last rows were hidden and its taps went to the text field. |
| The jump-to-latest chip faded out as it appeared | the transcript's bottom fade was applied from the caller to the whole view, and the chip lives exactly where the mask goes to zero |
| The app detail sheet's remove button re-installed | both actions handed `toggleApp` the app captured when the sheet opened, whose `enabled` is what `toggleApp` reads |
| An app opened from search never flipped to Open | `AppDetailSheet.isEnabled` read only `appProvider.apps`, so an app living in `filteredApps` kept its stale captured value forever |
| A failed app search left its banner up over fresh results | `searchApps` cleared the error after its early returns rather than before |
| Notch content was clipped by the silhouette | `NotchShape` runs its vertical sides at the corner radius; the card inset 12pt against a 20pt open radius, cutting the leading icon. The inset now derives from the radius it must clear. |
| Every PTT error rendered as a shimmering "Listening…" | the ladder derived from `isVoiceListening`, which folds the hint in. It takes the raw phase now, so the real message is reachable. |
| The notch ran at 60fps while idle-hovered | the shimmer now stops when the reply settles and the reveal timeline pauses once the text has caught up |

---

## 9. Kept, and retired

**Every feature still works.** PTT voice turns (hub and batch), voice playback settings, proactive notifications end to end (queue, click-through, Execute, dismiss), the Ask-Omi hotkey, agent timeline opens, the recording indicator, show / hide / snooze preferences, usage limiting, the automation bridge, Settings in the same order, Capture and Rewind, and chat as one shared timeline.

**Retired on purpose**, because the new design makes them meaningless: the floating bar's manual bottom-corner resize and the draggable pill preference (every display gets a notch, and height is dynamic), and text input in the floating surface (it belongs in the app).

**Nothing is deleted.** `FloatingControlBarWindow` and its views, `PersonaPage`'s inline header, `DashboardPage.homeHeader`, and the five call-site-less views in `AppsPage` stay in-tree, compiling and tested, marked retired, so removing them can be its own reviewable PR rather than a several-thousand-line delete buried in this one.

---

## Product invariants

| Invariant | How it holds |
|---|---|
| **INV-CHAT-1** | One shared transcript. The notch is an I/O surface over the one kernel transcript; the streaming reply it renders is the display-only `liveVoiceAssistantText` mirror, cleared on a new hold and when the voice presentation ends, never a second store. Home's landing is presentation only: it does not fork the conversation or mint a session, and its first send goes through the existing `sendMainDraft`. The prompt timeline, the paced reveal, the live-edge follower, the turn spacing and the working mark are presentation over the same messages. `AgentTranscriptSheet` is read-only over the shared `ChatMessagesView`. |
| **INV-VOICE-1** | The PTT reducer and journal lifecycle are untouched. `PTTBarPresenter` still projects reducer state; only the presentation surface moved. Additions are display-only or test-only: `playbackLevel`, `liveVoiceAssistantText`, `isListeningPhase`, `flashHint`, and a non-production usage-limit bypass. `VoiceTurnCoordinator` gains exactly one side effect, a soft end cue on `terminal(.success)`. `PTTCue` centralizes the start earcon that was copy-pasted into both start paths. |
| **INV-AUTH-1** | No session-ownership change. The shell, header, sidebar and stage work is presentation and does not touch `AuthSessionCoordinator`, session mint, `runtimeOwnerId` or token handling. Integrations calls the backend's grant endpoints through `APIClient` on the existing authenticated session and neither mints nor holds credentials. The desktop-auth-session ratchet passes. |
| **INV-AGENT-*** | Matched by path only, through the upstream merge. This branch registers no adapter, and changes no capability matrix, activation env, tool manifest or provider enum. The agent-facing work here is presentation: `AgentStatusGroup` colours the notch mark from existing pill state, and `AgentTranscriptSheet` renders an existing agent transcript read-only. The model still cannot select a local provider on its own. |
| **INV-UI-1** | White and neutral on the existing dark surfaces throughout. No purple accents, glows or gradients in the new components, the notch, or the reworked pages. `brand-ui` passes. |

Failure-Class: none

The `fix:` commits in this branch correct defects inside this new work, not a shipped failure-class boundary.

---

## Testing and verification

| Check | Result |
|---|---|
| `xcrun swift build -c debug --package-path Desktop --build-tests` | clean |
| Full Swift test suite | 3,371 tests, 0 failures |
| `swift-format` + SwiftLint | 0 violations across 962 files, baseline not raised |
| `check-e2e-flow-coverage.py --strict` | Uncovered: 0 |
| Test-quality ratchet | at baseline, not raised. This work pushed it to 57 files / 156 sites and it was brought back to 55 / 154. |
| Line-count ratchet | down-only and honoured: `DashboardPage` 4,467 → 3,000 across the splits, `AppsPage` 3,621 → 3,312, plus `ChatBubble`, `MemoriesPage`, `TasksPage`. Exact raises with one-line justifications only where a file genuinely grew. |
| Housekeeping | 17 changelog fragments under `changelog/unreleased/`, and 10 e2e files updated: `home-stage` rebuilt around the landing/chat/connect model, `apps`, `navigation`, `settings-basic`, `ptt-lifecycle`, `floating-bar-functional`, `apps-marketplace`, `chat-hermetic`, `conversation-detail` and `SKILL.md` |
| `make preflight`, `scripts/pr-preflight --pr-body-file` | 24 / 24 checks pass |

**20 new hermetic suites**, no wall-clock sleeps: `NotchGeometryTests`, `NotchPresentationTests`, `NotchViewModelTests`, `NotchReplyRevealTests`, `NotchCardExitTests`, `ChatStreamingRevealTests`, `ChatTranscriptRhythmTests`, `ChatPromptTimelineTests`, `ChatJumpButtonPlacementTests`, `HomeComposerPlacementTests`, `HomeAskBarLayoutTests`, `HomeTranscriptLayoutTests`, `HomeTranscriptMeasurementTests`, `CaptureListeningModeTests`, `ListeningHoverRevealGeometryTests`, `AppsPageInstallScopeTests`, `AppsHeaderRowLayoutTests`, `SheetChromeHeaderLayoutTests`, `FittedModalGeometryTests`, `IntegrationGrantStoreTests`.

The layout tests assert against production values through a real layout pass, not source strings. `CaptureListeningModeTests` covers all eight combinations of transcribing / awaiting / error and proves the chip can never read active while capture is gated off. `NotchViewModelTests` pins the window frame across every presentation and every measurement, on display geometries where a leaked measurement would change the answer. Where a contract genuinely cannot be asserted behaviorally (SwiftUI leaves the accessibility tree of a never-displayed hosting view unpopulated), it is labelled a static tripwire with the reason written next to it.

**Exercised live** on a named bundle (`OMI_APP_NAME=omi-rework ./run.sh --yolo`), not just compiled:

- Real hold-and-speak turns end to end on the notch.
- Notification cards rendered and captured to PNG through the automation bridge. The icon-clipping fix is verified by pixel position: first bright pixel moved from x=84 to x=111.
- `home-stage.yaml` PASS 10/10 on the running bundle, after S2 was turned back into a real assertion. It had been a bare `wait:` with no typed harness operation, asserting nothing.
- Integrations read live: `IntegrationGrants: read google_calendar=no gmail=no`.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10672?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


katipally:
@kodjima33  @Git-on-my-level need human response

following up from our call with something real. This is the whole mac app plus the notch, properly reworked and actually wired together in one place instead of scattered. Full polish pass too, everything now follows proper macOS design. Code is where I do my best talking, so I'd rather you judge me on this one. Happy to walk you through any part, or split it into smaller PRs if that's easier to review.

Git-on-my-level:
Thank you! I like some of the ideas in here, like the scroll selector, some of the UI cleanup

I will take a deeper look in a bit

Git-on-my-level:
I like a few of the ideas you introduced, but the whole re-design is a bit much, I will cherry pick some of your improvements so we can persist them

Git-on-my-level (changes requested):
Thanks for the very substantial desktop pass here — there are useful ideas in the notch work, Apps filtering, status-chip cleanup, and test coverage.

I don’t think this PR should merge as-is, though. It is a full replacement/rearchitecture of the macOS desktop/floating-bar surface and GitHub currently reports it as conflicting, so it needs maintainer product/architecture selection rather than an automated approval.

A few concrete blockers/risks I noticed:

- The branch changes the agent/e2e instruction surface in `desktop/macos/e2e/SKILL.md` (settings sections, Apps filter semantics, flow expectations). That can alter how coding/review/e2e agents navigate and validate the desktop app, so maintainers should confirm the updated guidance matches the UI they actually want before it becomes repo guidance.
- `desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Sections/IntegrationGrants.swift` defines `IntegrationGrantStore: ObservableObject` and uses `@Published`, but the new file imports only `AppKit` and `Foundation`; it should import the framework that provides those Combine/SwiftUI observation symbols. Please make sure the macOS target builds cleanly from a fresh checkout.
- The new notch panel uses a very high `assistiveTechHighWindow` level. I see the attempts to yield for menus/system prompts, but this is a workflow-sensitive UI surface and should get human UX/accessibility review before replacing the current floating bar.

Given David’s comment about cherry-picking selected improvements, my recommendation is to extract the smaller pieces the maintainers want to keep (for example the focused UI cleanups/tests) rather than merge the whole redesign in one PR. Happy to re-review a narrower follow-up.

---

undivisible:
hey thank you for the work! we love some of your designs and we will cherrypick some stuff to put into main :))

PR #10682, opened 2026-07-26
fix(backend): stop a permanently rejected app webhook from failing conversation finalization

kodjima33:
## Bug (live in prod)

A conversation's durable finalization job calls `trigger_external_integrations(..., require_delivery=True)`. **Any** non-2xx response from a user's third-party app webhook was counted as an undelivered fanout, which raises `ExternalIntegrationFanoutError` → `ConversationFinalizationError('processing_failed')` → HTTP 500 on `POST /v1/conversation-finalization-jobs/run`. Cloud Tasks then retries the whole job until it dead-letters, so the conversation never finalizes.

The failures in prod are permanent, not transient. On 2026-07-26, over 11.5h of `based-hardware` logs, **335 of 336** app-integration failures were 4xx:

| app | status | count |
|---|---|---|
| `notion-crm` | 400 (`object_not_found`) | 181 |
| `omi-google-drive-integration` | 401 (`Please re-authenticate`) | 122 |
| `01KHY4N6AG8D4QAGP4E7AH5M0H` | 404 | 32 |
| `01KS2R9YFFXEDX6BDT4EQ5YRQZ` | 502 | 1 |

Blast radius over 24h: **86 distinct conversations across ≥10 users**, each burning all 5 attempts (`persisted conversation finalization failed uid=… failure=processing_failed`), plus 22 `backend-sync` 500s per 12h on the job route. Every retry also re-runs the derived-effect bundle (memory extraction, vector upsert, delivery to the user's *working* apps).

Correlated example (backend-sync, 2026-07-26T18:04:00Z):

```
App integration failed omi-google-drive-integration status: 401 result: {"error":"Unauthorized","message":"Please re-authenticate …
Plugin integration request failed app=omi-google-drive-integration error=RuntimeError
persisted conversation finalization failed uid=xar839kthBUuQvaFjTm2rRUhcgq2 conversation=d3942361-… failure=processing_failed
```

## Root cause

`require_delivery` conflated "not delivered" with "a retry may deliver". A retry only helps when the destination can answer differently next time; an expired OAuth token or a deleted Notion page answers every retry identically, so the owner of the conversation's terminal state spent its whole retry budget on a dependency's permanent decision.

## Fix

`backend/utils/app_integrations.py`: keep the finalization job retryable only for failures a retry can change — 5xx, 408, 425, 429 (network exceptions and an open circuit breaker stay retryable, unchanged). A permanent rejection is recorded via `record_fallback(component='webhook', to_mode='dropped', outcome='degraded')` and dropped. Webhook health tracking is untouched and still owns the permanent case: it warns the app owner and auto-disables the webhook after 72h of failures. The delivery outcome for the broken app is unchanged (it was dropped after 5 attempts anyway); the conversation now finalizes.

## Verification

Both new tests fail on `main` and pass with the fix (checked by stashing only the production change):

- `tests/unit/test_listen_finalization_cloud_tasks.py::test_finalizer_completes_when_an_app_permanently_rejects_the_delivery` — drives the real `finalize_persisted_conversation` with the **real** fanout and a webhook returning 401; asserts the disposition is `completed`, `complete_finalization_fanout` is called, and the webhook-health failure is still recorded. Before the fix: `ConversationFinalizationError`.
- `tests/unit/test_async_app_integrations.py::TestDurableExternalIntegrationFanout::test_permanent_delivery_rejection_does_not_fail_finalization[400/401/404]`.

```
$ .venv/bin/python -m pytest tests/unit/test_listen_finalization_cloud_tasks.py \
    tests/unit/test_async_app_integrations.py tests/unit/test_app_integrations_async.py -q
99 passed
$ .venv/bin/python -m pytest <same files> -q -k "permanent or Durable or fanout"
32 passed
$ black --line-length 120 --skip-string-normalization <touched files>   # 3 files unchanged
```

Not covered: a live prod conversation with a broken app integration — that needs a user's real third-party account, so the proof here is the hermetic finalizer path plus the prod log evidence above.

Failure-Class: FC-durable-terminal-projection

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10682?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_async_app_integrations.py
+++ b/backend/tests/unit/test_async_app_integrations.py
@@ -299,7 +299,7 @@ class TestDurableExternalIntegrationFanout:
 
         with patch.object(app_integrations, 'get_available_apps', return_value=[app]), patch.object(
             app_integrations, 'get_webhook_client', return_value=client
-        ):
+        ), patch.object(app_integrations, 'conversation_to_dict', return_value={}):
             await app_integrations.trigger_external_integrations(
                 'uid-1', conversation, idempotency_key='fanout-1', require_delivery=True
             )
@@ -317,7 +317,9 @@ class TestDurableExternalIntegrationFanout:
 
         with patch.object(app_integrations, 'get_available_apps', return_value=[app]), patch.object(
             app_integrations, 'get_webhook_client', return_value=client
-        ), pytest.raises(app_integrations.ExternalIntegrationFanoutError):
+        ), patch.object(app_integrations, 'conversation_to_dict', return_value={}), pytest.raises(
+            app_integrations.ExternalIntegrationFanoutError
+        ):
             await app_integrations.trigger_external_integrations(
                 'uid-1', conversation, idempotency_key='fanout-1', require_delivery=True
             )
@@ -335,7 +337,7 @@ class TestDurableExternalIntegrationFanout:
 
         with patch.object(app_integrations, 'get_available_apps', return_value=[app]), patch.object(
             app_integrations, 'get_webhook_client', return_value=client
-        ):
+        ), patch.object(app_integrations, 'conversation_to_dict', return_value={}):
             messages = await app_integrations.trigger_external_integrations(
                 'uid-1', conversation, idempotency_key='fanout-1', require_delivery=True
             )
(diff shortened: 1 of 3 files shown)


undivisible (approved):
Verified permanent webhook rejections no longer prevent finalization, with focused regression coverage.

PR #10687, opened 2026-07-26
feat(desktop): People tab — on-device relationship mapping

Archit-lal:
## Rebased onto current main

This branch was 575 commits behind. It is now rebased onto `main` (base `d2d921803e`),
so the People tab runs alongside the proactive-suggestions work merged in #10673.

All 11 original commits replayed. Four commits were added on top of the rebase:

### 1. Wire the People tab into the Chat-first shell

Since this PR was opened, main moved primary navigation out of `SidebarView`'s inline
enum into `SidebarNavItem.swift` + `TopNavigationRoutes`, and `ChatFirstShell` became the
live shell. A clean replay registers `people` only in the legacy sidebar, so the page
compiles but is unreachable in the shipping UI.

- register `people` in `SidebarNavItem` (case, title, icon, `mainItems`)
- add the People pill to `TopNavigationRoutes.primaryItems`
- add `ChatFirstMorePage.people`, mount `PeoplePage` in `ChatFirstShell.moreDestination`,
  and cover the legacy/automation route adapters
- update `TopNavigationBarLayoutTests` for the new primary route list

### 2. Raise the product line-count baselines the People tab crosses

The rebase took main's ratchet baselines, which left `PeoplePage.swift` with no entry and
three files above their recorded counts. Each now records its exact count and a one-line
justification.

### 3 & 4. Two local-build blockers on current main (independent of this feature)

Both reproduce on unmodified `main` and block every local named-bundle build on
Xcode 26.2 / Swift 6.2.3. They are here because this branch could not otherwise be built
and exercised; they are good candidates to split into their own PR.

**Bundled node path.** `a9c287ae94` moved every bundled-node path to the *versioned*
SwiftPM resource-bundle layout (`…bundle/Contents/Resources/node`). This toolchain emits a
*flat* bundle (`…bundle/node`) — as does shipped `/Applications/Omi.app` — so `run.sh`
silently skipped signing node with `Node.entitlements` and
`audit-desktop-bundle-deps.sh` then failed the build outright. All four scripts now
resolve the versioned path first and fall back to the flat one.

**Library validation with a team-less certificate.** `run.sh` directs developers without
an Apple identity to a local certificate and forbids ad-hoc signing (it resets Screen
Recording permissions for every Omi app), but `prepare-local-dev-entitlements.sh` only
granted the library-validation exception in `adhoc` mode. A self-signed certificate has no
Team ID, so hardened-runtime library validation rejected the bundled frameworks and the
app aborted in dyld at launch:

```
Library not loaded: @rpath/Sparkle.framework/Versions/B/Sparkle
... not valid for use in process: mapping process and mapped file
(non-platform) have different Team IDs
```

That left such a machine with no way to launch a named bundle at all. A new `local-cert`
mode takes the same exception as `adhoc`; `run.sh` selects it when the resolved identity
is not spelled `... (TEAMID)`. Apple identities keep library validation.

Also: the previous branch tip had committed a Swift test with unresolved conflict markers
(`PTTInputDeviceProbeTests.swift`). The app target still compiled, so only the test target
exposed it. Resolved to main's expectation-based form, which subsumes the branch's
slow-CI threshold tweak.

Matched by changed-path globs (`scripts/pr-preflight --suggest`); neither behavior changes.

- **INV-AUTH-1** — desktop session surfaces. `OmiApp.swift` gains only a
  `PeopleGraphBuilder.syncIfNeeded` call on app-become-active; no auth/session surface,
  token handling, or sign-in path is touched.
- **INV-INT-1** — integrations. The People connectors read iMessage/WhatsApp on-device and
  reuse the existing connector-import runner; no new egress path or provider is introduced.

## Verification

Built and exercised the real app, not just a compile:

- `xcrun swift build -c debug --package-path Desktop` → Build complete
- `xcrun swift test --package-path Desktop --filter "People|TopNavigationBarLayout|ChatFirst|IMessageText|WhatsAppReader|PTTInputDeviceProbe"` → **Executed 103 tests, 0 failures**
- `bash tests/test-prepare-local-dev-entitlements.sh` → passed (the new `local-cert` case fails without the script change)
- `bash tests/test-audit-desktop-bundle-deps.sh` → passed (versioned fixture)
- `./scripts/audit-desktop-bundle-deps.sh build/omi-people-proactive.app` → audit passed, 18 Mach-O files (flat, real local build)
- `python3 .github/scripts/check_product_file_line_count_ratchet.py --changed-files <branch diff>` → OK across 20 changed product source files
- Named-bundle run: `OMI_APP_NAME=omi-people-proactive ./run.sh` → installs, launches, `omi-ctl health` reports `ok: true` with the agent-runtime protocol negotiated. The People tab renders live on-device data (390 people, 73 multi-channel), and Settings → Notifications & Privacy → **Live Suggestions** (the #10673 surface) is on in the same build.

Failure-Class: none

---

## Update: a person profile, and the pipeline behind it

The People tab listed people but had nowhere to put what Omi knows about them. Clicking a
person expanded a 161-line panel inside its own list row.

**Now:** the list is a sortable directory (Person / Channels / Last touch), and clicking a
person opens a real page — identity hero, a person-scoped ask bar wired to chat, suggested
prompts, then tabbed evidence: Overview, Messages, Memories, Commitments. Message history is
read live from the on-device stores through one cached read-only snapshot per page (Omi keeps
no message text), paginated and consent-gated. Export renders the profile to PDF; Send hands
it to the Messages composer pre-addressed — **the app never sends, the composer opens and the
user presses Send.**

### What investigating this turned up

The tab rendered a CRM whose data layer was never built in this repo. `who`, `now`, `overall`,
`relationship`, `role`, `affiliations`, `activities`, `openThreads`, `photoPath`,
`community_meanings` each had a decoder and a renderer here and **zero writers** — a repo-wide
grep for the write form returns exactly one file, the decoder's `CodingKeys`. Those fields
exist only in files produced out-of-tree. `hasExistingPeople()` routes such a machine down the
merge branch permanently, so `createPeopleIntelligence()` — the only path a new user takes —
had never run where the feature was being evaluated.

Three retrieval paths failed at once for "what do I know about X":

1. **No tool read the graph.** Desktop chat is a local agent loop that could have — it just
   had no people tool. Added `get_person` / `search_people`, reading the on-device graph
   directly. No backend, no upload.
2. **The relay discarded oversize tool results whole.** `MAX_RELAY_TOOL_RESULT_BYTES` overflow
   returned zero content bytes — a 13-match conversation search became "I can't read the
   details". Now head-projects under budget (whole trailing array items, UTF-8 safe,
   byte-measured) and names the recovery tools. A second identical total-loss path was found
   and fixed. The repo's own contract fixture already declared the corrected behaviour, so the
   implementation was out of contract, not the test.
3. **`PeopleMemoryWriter` stamped `source`, which the backend model does not have**, so
   Pydantic dropped it and every relationship fact was unattributed. Now writes `tags`.

Cold start was broken too: Contacts was requested fire-and-forget one line before the export,
so a new user's first pass named everyone by phone number — which then failed `isHumanName`
(no facts) and the ingest candidate check (no understanding). Contacts is now awaited, bounded,
in parallel with the export; denial is a distinguishable state with a notice; contact
thumbnails fill `people_photos/`, which nothing had ever written. And the tab could not see its
own pipeline — writes landing after `load()` required an app relaunch; `reloadIfFileChanged()`
fixes that.

Deterministic Phase-2 fields now have a writer: `affiliations` (corroboration mandatory — a
single work-ish group chat is not an employer, or the categorizer turns "meat gang" into one),
`relationship` as a reach tier plus context, `community_meanings`, `connections[].how`,
`history_grounded`.

### Deliberately not built

`who`, `now`, `overall`, `facts`, `activities`, `openThreads`, `network_insights` need a model
(Phase 3 in `people-intelligence-productization.md`). Inventing deterministic stand-ins would
be dishonest; a test asserts created cards omit them. That work needs a person-dossier endpoint
and a **new** person-centric prompt — `extract_memories_prompt` forbids standalone facts about
third parties.

**Tasks by person is not implemented.** No task carries an assignee at any layer and
`CanonicalTaskCreate` is `extra='forbid'`. The tab is labelled *Commitments* and states plainly
that these are extracted open threads, not assigned tasks.

**Known gaps, worth their own PRs:** the identity bridge — `buildCanonicalPeople` computes a
`phone_last10` identity and `createPeople` then discards it, so `slug(name)` is the identity and
a rename detaches overrides, fact edits and memory tags; and `person_id: nil` on ingested
segments, which `subjects.py` would reject anyway for any conversation where the user also
speaks.

### Verification

- `xcrun swift build -c debug --package-path Desktop` → clean under
  `-strict-concurrency=complete -warnings-as-errors`
- `xcrun swift test --package-path Desktop --filter "People|Person"` → **135 tests, 0 failures**
- `desktop/macos/scripts/test-tool-surfaces.sh` → exit 0 (generator `--check` clean, vitest 659)
- `python3 .github/scripts/check_product_file_line_count_ratchet.py` → OK across changed files;
  `PeoplePage.swift` 1790 → 1448 as the profile moved out
- Named-bundle run: builds, signs, installs, launches; `omi-ctl health` → `ok: true`


Git-on-my-level (changes requested):
Thanks for pushing this forward — the direction of keeping relationship mapping on-device is promising, and I appreciate the explicit opt-in language.

I’m going to request changes before maintainer/product review because this touches highly sensitive local user data (Messages/Contacts) and adds a new desktop surface.

Blocking concerns:

1. The iMessage exporter currently persists raw message snippets into `imessage_export.json` (`sample_texts`, up to 3 recent texts per handle, 500 chars each), but the relationship graph code does not appear to use those snippets. For a feature framed as relationship mapping/aggregates, storing message content materially increases privacy exposure and contradicts the “per-contact aggregates / on-device relationship mapping” minimization story. Please remove raw message text collection/persistence unless there is a reviewed product/security requirement for it, and document exactly why it is needed if kept.

2. The first-run pipeline is racy: enabling iMessage mapping starts `IMessageExporter.exportIfRequested()` and `PeopleGraphBuilder.rebuildIfNeeded()` as independent detached tasks, then the view reload reads `people_intelligence.json` immediately. On a fresh user, the graph builder can run before `imessage_export.json` exists and the UI can render the empty state even though the export is still running. Please sequence export → graph build → UI reload/notification, or otherwise expose completion state so the first enable action reliably produces the People data when available.

Given the privacy-sensitive local-data access and product implications of adding a People tab, this should also get human maintainer review before merge. I did not formally approve due to the approval gate and the unresolved concerns above.

---

Git-on-my-level (changes requested):
Thanks for the follow-up. The two blockers from my earlier review look substantially addressed on this head: the iMessage aggregate export no longer persists sample message text, and the export → graph rebuild → reload path is now sequenced through `PeopleGraphBuilder.syncIfNeeded`.

I’m still going to request changes because the current consent/privacy boundary is now different from what the UI and docs promise.

Blocking concern:

- Enabling the People connector is presented as “Reads Messages on-device only” / “Mapping · on-device”, and the productization doc says minimized data only / “never raw messages”. But the same `peopleIMessageExport` flag now also runs `PeopleThreadIngest`, which reads recent 1:1 iMessage/WhatsApp message bodies, builds transcript segments, and submits them to `POST /v1/conversations/from-segments` for backend/model processing. Redacting phone numbers/emails/codes and bounding to the recent window helps, but this is still message content leaving the device under an opt-in that appears to authorize local relationship mapping, not cloud transcript ingestion. The same broad flag also gates relationship-memory writes.

Please separate the privacy surfaces before merge: keep local on-device graph building behind the current connector opt-in, and put any cloud memory writes / conversation-transcript ingestion behind a distinct, explicit user-facing consent that says message-derived content may be sent to Omi for memory extraction; or remove the transcript ingestion path from this PR and land it later with reviewed UX/security copy. The docs/changelog should match the actual behavior.

Given the local Messages/WhatsApp/Contacts access, backend memory writes, and new People product surface, this still needs human maintainer/product/security review before merge. I did not formally approve due to the approval gate.

---

Archit-lal:
### CI status: own code green; `Hygiene` red on a pre-existing `main` failure

11 of 12 workflows pass on `a1095a3cd6`, including **`Desktop Swift Static & Test Contracts` (55m)**, `Backend unit suite`, `Hermetic Backend E2E`, `Agent Runtime`, and `Formatting`.

The single red check is `Repo Checks / Hygiene`, and it is **not caused by this PR**:

```
ERROR: Codemagic entire document raw byte digest does not match approved fixture codemagic_raw_sha256
ERROR: Codemagic entire document semantic digest does not match approved fixture codemagic_semantic_sha256
ERROR: ios-internal-auto must natively trigger on main app/** pushes and cancel stale builds
ERROR: android-internal-auto must natively trigger on main app/** pushes and cancel stale builds
```

The identical four errors occur on unmodified `main`:

| ref | run | Hygiene |
|---|---|---|
| `main` @ `d2d921803e` (this PR's base) | [30746405666](https://github.com/BasedHardware/omi/actions/runs/30746405666) | failure |
| `main` @ `6a7e1db787` | [30740631737](https://github.com/BasedHardware/omi/actions/runs/30740631737) | failure |

Reproduced locally: `python3 .github/scripts/check-release-process-guards.py` exits 1 with byte-identical output on this branch and on a pristine `upstream/main` worktree.

**Cause.** `20ccd99400` ("ci(mobile): batch internal builds instead of one per merge") removed the `triggering:` blocks from `codemagic.yaml`, moving mobile internal builds to the 2-hourly dispatcher, but did not refresh `codemagic_raw_sha256` / `codemagic_semantic_sha256` in `.github/scripts/fixtures/codemagic_workflow_contract/v1.json`, and `check_mobile_codemagic_release_triggers()` still requires the removed literal. `repo-checks.yml` runs this guard ungated, so every open PR inherits the failure.

This PR does not touch any of it — `codemagic.yaml` and the fixture are byte-identical blobs to `main` (`03f1e50507…` and `dd80e8eb53…`). Regenerating those digests here would launder an unrelated release-authority change through a desktop-feature review, which is what the fixture exists to prevent, so I have deliberately left them alone.

Possibly worth folding into the upstream fix: `.github/checks-manifest.yaml` registers `desktop-release-process-guards` with triggers that omit `codemagic.yaml` and the fixture, so the diff-scoped local lane never fires when `codemagic.yaml` alone changes — per AGENTS.md, a deterministic check failing first in CI is a manifest bug.

Git-on-my-level (changes requested):
I re-checked the current head. The main privacy/consent blocker is still present, so I’m keeping this as changes requested.

Blocking concern:

- The People connector UI still describes the opt-in as local-only (“Reads Messages on-device only” / “Mapping · on-device”), and the productization doc still says minimized data only / “never raw messages”. But enabling that same `peopleIMessageExport` path still also calls `PeopleMemoryWriter.writeIfNeeded(...)` and `PeopleThreadIngest.ingestIfNeeded(...)` from `PeopleGraphBuilder.syncIfNeeded(...)`. `PeopleThreadIngest` reads recent 1:1 iMessage/WhatsApp message text and submits bounded transcript segments to `POST /v1/conversations/from-segments` for backend/model memory extraction. Redaction, throttling, and local ledgers are useful safeguards, but this is still message-derived content leaving the device under a consent surface that reads as on-device relationship mapping.

Please separate these privacy surfaces before merge: keep local graph building behind the current on-device connector opt-in, and put cloud memory writes / conversation-transcript ingestion behind a distinct explicit user-facing consent that says message-derived content may be sent to Omi for memory extraction; or remove the transcript-ingest path from this PR and land it later with reviewed UX/security copy. The docs/changelog should match whichever behavior ships.

Also, this head still has failing required checks, including Desktop Swift Build & Tests, Desktop Swift Static & Test Contracts, Hygiene, and PR Metadata Preflight, so it is not merge-ready even aside from the product/security review.

Given the local Messages/WhatsApp/Contacts access, backend memory writes, and new People product surface, this still needs human maintainer/product/security review before merge. I did not formally approve due to the approval gate.

---

Git-on-my-level (changes requested):
Thanks for the continued work here — the implementation has added useful guards (bounded thread windows, redaction before transcript upload, model-output grounding, and a clearer backend dossier contract), but I still need to keep this as changes requested because the consent/privacy boundary is not yet aligned with the behavior that ships.

Blocking concern:

- The connector UI still presents the iMessage/WhatsApp opt-in as local-only: “Reads Messages on-device only”, “Reads WhatsApp on-device only”, and then “Mapping · on-device”. That same flag still drives `PeopleMemoryWriter.writeIfNeeded(...)`, `PeopleThreadIngest.ingestIfNeeded(...)`, and now the narrative refresh through `POST /v1/people/dossiers`. In practice, enabling that local-sounding connector can upload graph-derived relationship memories and redacted recent 1:1 message transcripts to Omi’s backend/model pipeline, then ask the backend model to generate per-person dossiers from those memories. Redaction and grounding reduce risk, but they do not make this an on-device-only feature.

Please separate the privacy surfaces before merge: keep the local graph/message-history read behind the current on-device connector opt-in, and put cloud memory writes, conversation-transcript ingestion, and model-backed dossier generation behind distinct user-facing consent/copy that explicitly says message-derived relationship content may be sent to Omi for memory extraction / narrative generation; or remove those cloud paths from this PR and land them later with reviewed product/security copy.

Also, the changed architecture/productization docs are agent-relevant: they document the People dossier route and assert the current privacy contract. As written, future coding/review agents may rely on those docs and continue treating this as safe under the existing connector consent, so the docs should be corrected together with the UI/behavior.

Given the Messages/WhatsApp/Contacts access, backend memory writes, and new People product surface, this still needs human maintainer/product/security review before merge. I did not formally approve due to the approval gate.

---

aryanorastar:
Not merging this.

Head is still d8638ce (Aug 3). The Aug 4 CHANGES_REQUESTED is unanswered, and the branch is now CONFLICTING on backend memories/action-items/tools, route_policy_manifest.yaml, subjects.py, and several line-count ratchets. PeoplePage is not on main.

The consent mismatch is still in the files. Connector copy is "Reads Messages on-device only" / "Mapping · on-device". people-intelligence-productization.md still says never raw messages. The same peopleIMessageExport flag still runs PeopleMemoryWriter.writeIfNeeded and PeopleThreadIngest.ingestIfNeeded from PeopleGraphBuilder.syncIfNeeded, which uploads 1:1 transcript segments to POST /v1/conversations/from-segments.

Split the surfaces before any rebase: local graph behind the current opt-in; cloud memory writes / transcript ingest / dossiers behind copy that says message-derived content leaves the device — or drop those paths from this PR. Fix the docs with the UI.

The bundled-node path, team-less local cert, and make-setup path compare do not belong in this review. I am not approving a DIRTY Messages/WhatsApp upload under on-device-only copy.

aryanorastar:
Keeping this open. Not approving `d8638ce50da6`. `PeoplePage.swift` absent on origin/main. +24489. No check rollup.

PR #10691, opened 2026-07-26
fix(backend): cap the existing-node listing in the KG extraction prompt

kodjima33:
## Bug

`extract_knowledge_from_memory` renders the user's **entire** `knowledge_nodes` collection into the extraction prompt. The collection is unbounded, so the prompt grows with the graph. A live prod user with **5,788 nodes** produces a **641,173-character** prompt; the provider rejects it with a 400 and every knowledge-graph extraction for that user has failed since — the richer the graph, the more certain the failure.

In the last 12h, **48 of 79** llm-gateway error terminals were `lane=omi:auto:knowledge-graph`, plus the matching `ERROR:root:Error extracting knowledge graph from memory_id: …` tracebacks in `backend` and `backend-sync`.

> Note on the error label: prod's llm-gateway image (`3e415e0`) predates #58971af, which classifies provider rejections. That build maps *every* provider 4xx to `capability_mismatch`, which is why the logs say "capability_not_supported" for what is a payload-size rejection.

## Root cause

`get_knowledge_nodes(uid)` streams the whole collection and the result is serialized straight into `EXTRACTION_PROMPT`. Nothing bounds it. `rebuild_knowledge_graph` carried a second copy of the same inline listing.

## Fix

Cap the listing at the 500 most recently updated nodes, via one shared helper used by both call sites. Merging is unaffected: `label_to_node_id` is still built from **every** existing node, so a node left out of the prompt is still deduplicated by label or alias instead of being re-created.

## Tested

- `tests/unit/test_kg_prompt_node_cap.py` — 3 passed; the prompt-size assertion fails on the pre-fix module.
- The 5 kg / canonical-kg unit files — 34 passed.
- **Real prod path**, from a `prod-omi-backend-listen` pod against the live gateway lane, using the real 5,788-node graph:
  - before: 641,173-char prompt → `400 … capability_mismatch` (the exact prod error)
  - after: same graph through the 500-node cap → 59,099 chars, 200 OK, parsed nodes `['User', 'coffee']`, edge `('User', 'likes', 'coffee')`

## Product invariants

none

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10691?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-unbounded-user-collection-in-prompt.json
+++ b/.github/failure-classes/FC-unbounded-user-collection-in-prompt.json
@@ -1,0 +1,13 @@
+{
+  "schema_version": 1,
+  "id": "FC-unbounded-user-collection-in-prompt",
+  "violated_contract": "A per-user collection that grows without limit must not be rendered whole into a provider request.",
+  "canonical_prevention": "Bound the listing at the point it is rendered \u2014 cap it by an explicit, ordered selection \u2014 and keep correctness that depended on the full set in code that still reads the full set.",
+  "evidence_prs": [
+    10691
+  ],
+  "scope_hints": [
+    "backend/**"
+  ],
+  "status": "open"
+}
--- a/backend/tests/unit/test_kg_prompt_node_cap.py
+++ b/backend/tests/unit/test_kg_prompt_node_cap.py
@@ -1,0 +1,111 @@
+"""KG extraction must keep the existing-node listing out of provider-rejection territory.
+
+``extract_knowledge_from_memory`` embedded every node Firestore returned. A user with
+5,788 nodes produced a ~641k-character prompt, the provider answered 400, and every
+extraction for that user failed from then on — the richer the graph, the more certain
+the failure.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+from contextlib import nullcontext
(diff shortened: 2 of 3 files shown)


tianmind-studio (approved):
Reviewed the prompt-boundary change and verified it on native Windows after a clean merge with current `main`.

The cap is applied at both prompt-construction call sites, orders large graphs by the Firestore `updated_at` / `created_at` datetime fields, and deliberately retains the full node set for label/alias deduplication. That avoids the provider-size failure without creating duplicates for nodes omitted from the prompt.

Verification:
- all five KG/canonical-KG unit files pass: `34 passed`
- the regression covers the observed 5,788-node graph, small-graph preservation, and omitted-node ID reuse
- Black, `py_compile`, and `git diff --check` are clean
- all current GitHub checks are green

No blocking issue found.

PR #10695, opened 2026-07-27
fix(desktop): resolve PTT input routing off the main actor

### LINKED ISSUE #10442: Desktop: move synchronous CoreAudio PTT routing off MainActor
During investigation of Beta v0.12.117 cold-start PTT for #10441, the available logs did not implicate this path and did not prove a macOS main-thread deadlock. However, `PushToTalkManager+InputDeviceRouting.swift` resolves the PTT microphone from an `@MainActor` method, which synchronously invokes CoreAudio HAL enumeration and property reads in `AudioCaptureService.inputDeviceID(forUID:)`, `isDefaultOutputBluetooth()`, and `findBuiltInMicDeviceID()` during turn startup. A slow or wedged audio driver can therefore pin the UI. Track moving this lookup/routing behind a bounded non-main executor, preserving the selected-device and Bluetooth fallback policy, and add a fault-injection test that stalls the HAL seam while proving PTT startup and the main run loop remain responsive.

katipally:
## Summary

Fixes #10442.

PTT resolved its input device from an `@MainActor` method that synchronously called three CoreAudio HAL reads (`inputDeviceID(forUID:)`, `isDefaultOutputBluetooth()`, `findBuiltInMicDeviceID()`) at turn start. Those reads have no deadline and no cancellation, so a wedged audio driver pinned the whole UI, not just PTT.

- Read the HAL only on a dedicated non-main queue and publish the result as a `Snapshot`. The main actor now reads that snapshot under a lock and never touches CoreAudio.
- Coalesce refreshes so a wedged driver parks exactly one probe thread no matter how many turns the user starts, instead of one per turn.
- Warm the snapshot in `setup()` and at both PTT entry points, so the read overlaps turn-start work that already happens rather than gating capture on it.
- Keep `PTTInputDeviceRouting.overrideDeviceID` byte-identical, so the selected-device and Bluetooth A2DP policy is unchanged.

The routing call stays synchronous and all four call sites are untouched, so turn-start ordering and reducer ownership do not change.

### Why this is larger than a bounded wait

The obvious reading of "bounded non-main executor" is a background queue plus a timeout on the main thread. I built that first and measured it against the issue's other requirement, that the main run loop stays responsive:

```
bounded wait:      RESOLUTION=timedOut ELAPSED=0.305s RUNLOOP_TICKS=0
snapshot read:     RUNLOOP_TICKS > 10
```

A bounded wait moves the CoreAudio work off the main thread but still parks the run loop for the whole budget, so it converts an unbounded freeze into a 250ms freeze on every turn start. `testMainRunLoopKeepsPumpingWhileTheHALIsWedged` fails on that design and passes on this one.

### Trade-off

With a slow but not wedged HAL, the previous code froze the UI and then routed correctly. This code stays responsive and may fall back to the system default input for a single turn, which on Bluetooth output means one turn on the headset mic. Warming at `setup()` plus both entry points reduced this to zero cold-snapshot fallbacks across every real turn I ran, including the first turn after launch, but the window is not provably zero on a slow machine. It self-corrects on the next turn and is recorded through `recordFallback(area: "ptt_input_routing")`.

## Verification

Desktop Swift, from `desktop/macos`:

- `xcrun swift build -c debug --package-path Desktop` : build complete, Swift 6 strict concurrency with warnings as errors.
- `xcrun swift test --package-path Desktop --filter 'PTTInputDeviceProbeTests|ShortcutSettingsTests'` : 15 tests, 0 failures. Run 10 consecutive times to confirm no ordering dependence.
- `./test.sh` : 387 Swift suites in isolation, all passed.
- `./scripts/agent-logic-harness.sh --swift-only` : passed, 526 lifecycle/state tests.
- `./scripts/swiftlint-wrapper.sh lint` : 0 violations across 902 files.
- `bash desktop/macos/scripts/swift-format-wrapper.sh lint-scope` : clean, pinned swift-format 602.0.0.
- `python3 .github/scripts/check_product_file_line_count_ratchet.py --changed-files <list>` : OK. `PushToTalkManager.swift` baseline raised 2584 to 2587 with a justification for the three warm-up lines.

Real path, named bundle `/Applications/omi-10442-ptt.app`, bundle id `com.omi.omi-10442-ptt`, development backend, driven by real `Option` holds because every automation bridge action sets `automationCaptureBypass` and returns before this code:

- 4 authenticated PTT turns, all `reason=success route=hub` with live transcripts and audio replies.
- 2 of those with AirPods Pro as the default input **and** output device, which is the case the A2DP rule exists for:
  - `PushToTalkManager: hub on Bluetooth output, capturing from built-in mic to keep A2DP`
  - `AudioCapture: Using override device ID 84`, confirmed against CoreAudio as `MacBook Pro Microphone`, not the AirPods mic.
  - AirPods output stayed at `48000 Hz` through both turns. An HFP flap would have dropped it to 16000.
- The AirPods were connected roughly 20 minutes after launch, and the first turn after connecting routed correctly with 0 cold-snapshot fallbacks, which exercises snapshot refresh across a device change.
- 0 occurrences of `PTT input routing not resolved yet` across every real turn.

Not verified on hardware: the wedged-driver branch itself. This machine's HAL is healthy, so that path is covered by fault injection only. The explicit user-selected-microphone branch was not re-tested on hardware because `overrideDeviceID` is unchanged from `main` and is covered by the existing `ShortcutSettingsTests`.

## Tests

`PTTInputDeviceProbeTests` (new, 9 tests):

- `testMainRunLoopKeepsPumpingWhileTheHALIsWedged` : the issue's acceptance criterion. Injects a HAL that never returns and asserts the main run loop keeps ticking while turn-start reads keep completing.
- `testConcurrentRefreshesCoalesceIntoASingleHALRead` : 25 turn starts against a wedged driver park exactly one probe thread.
- Policy preservation: explicit user choice, Bluetooth built-in fallback, system default when output is not Bluetooth, unplugged selected device, and that the expensive built-in enumeration is skipped when output is not Bluetooth.
- Cold and stale snapshots: a cold snapshot reports no override rather than guessing, and a snapshot resolved for a previous microphone is never applied after the user picks a different one.

Both fault tests join their parked probe before returning, so a late HAL answer cannot publish into whichever test runs next.

`DesktopDiagnosticsManagerTests.testFallbackNamedAreasAreNotCollapsedToOther` gains `ptt_input_routing`, which is the guard that the new area survives allowlist bucketing instead of silently collapsing to `other`.

- INV-CHAT-1
- INV-VOICE-1

Both are path matches rather than behavior changes. `PushToTalkManager+InputDeviceRouting.swift` and `PushToTalkManager.swift` sit under the `FloatingControlBar/**` glob, and `PushToTalkManager.swift` is named directly in the INV-VOICE-1 globs. The change adds no chat surface and no turn ownership: `VoiceTurnReducer` remains the sole turn owner and the four routing call sites are unchanged.

Failure-Class: none

No class in `.github/failure-classes/` describes an uncancellable external-subsystem call made on the owning actor. The closest sibling in-tree is `AppBuild.channelProbeMainThreadBudget`, a main-thread wait on a network probe, but one prior instance is not a recurring pattern, and opening a class is a separate registry PR.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10695?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-floatingcontrolbar.json
@@ -3,7 +3,7 @@
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": 2375,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": 2638,
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": 4829,
-    "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2584,
+    "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": 2587,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift": 1536,
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": 1639
   },
@@ -11,7 +11,7 @@
     "desktop/macos/Desktop/Sources/FloatingControlBar/AgentPill.swift": "Agent completion presentation now follows the canonical terminal lifecycle used by PTT and chat.",
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift": "The Second Brain onboarding, reach-error, and notch-moment UI keeps the hover menu agent-only and routes idle notch taps through main chat.",
     "desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift": "Second Brain continuous-chat routing shares window geometry with agent-only hover sizing and canonical idle-notch chat ownership.",
-    "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": "DEBUG owner-boundary tests need a capture-free effect-handler install seam; shared ensureAutomationBarConfigured() removes duplicated automation setup while adding that seam (BasedHardware/omi#10499).",
+    "desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift": "DEBUG owner-boundary tests need a capture-free effect-handler install seam; shared ensureAutomationBarConfigured() removes duplicated automation setup while adding that seam (BasedHardware/omi#10499). Setup and both PTT entry points then warm the CoreAudio input-routing snapshot off the main thread so turn start never reads the HAL inline (BasedHardware/omi#10442).",
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift": "Adds bounded realtime mint-failure outcome correlation and close-resolution telemetry so release health distinguishes started recovery from exhaustion; telemetry-only, no lifecycle change (#10425).",
     "desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubSession.swift": "Strict concurrency annotations (Ticket 14) add Sendable conformances and isolation qualifiers; sendBackgroundAgentContext now confirms the provider send before returning true (exactly-once: no ack before delivery) and the session emits hubDidOpenInputWindow so a warm-idle completion is retried \u2014 both must live beside the file-private send/turn machinery."
   },
--- a/desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift
+++ b/desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift
@@ -1411,6 +1411,7 @@ final class DesktopDiagnosticsManager {
     "task_workflow",
     "auth_storage",
     "state_authority",
+    "ptt_input_routing",
     "other",
   ]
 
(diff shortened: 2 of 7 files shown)


Git-on-my-level:
Thanks for the careful fix here. I reviewed the PTT input-routing change statically and this looks like the right direction: CoreAudio/HAL reads move off the main actor, the main-actor path only reads a locked snapshot, refreshes are coalesced so a wedged driver does not spawn one blocked probe per turn, and the existing explicit-mic/Bluetooth built-in fallback policy is preserved. The new tests cover the important regression cases, including main-run-loop responsiveness while the probe is wedged, coalescing, stale UID snapshots, and the existing routing policy.

I’m leaving this as a positive signal rather than a formal approval because this touches desktop audio routing plus the `.github/scripts` ratchet baseline, and the monitor context disallows auto-approval for these risk surfaces. I also did not run the macOS/CoreAudio XCTest suite from this Linux review environment, so a human maintainer should still verify the desktop test target/CI on macOS before merge.

No requested changes from me based on the static review.

---

kodjima33 (approved):
Verified against #10442: the synchronous CoreAudio HAL reads still run on the main actor on main, so a wedged driver pins the whole UI; moving the probe to a dedicated queue with a published Snapshot is the right shape and comes with tests + a changelog entry.

PR #10704, opened 2026-07-27
fix(desktop): reject attestation descriptors in lineage gate (SCA-193, SCA-191)

Git-on-my-level:
## Summary

- reject a selected `linux/amd64` descriptor when it identifies itself as an attestation
- reject descriptor annotations that are not an OCI string-to-string map
- fail closed on non-v2 OCI/Docker manifest data before evaluating image lineage
- cover all cases with release-policy regression fixtures

## Root cause

The lineage verifier initially treated any `linux/amd64` image-manifest descriptor as runnable. It now rejects explicit attestation descriptors, and this exact-head repair closes the remaining fail-open where an invalid non-string annotation value (for example `vnd.docker.reference.type: 7`) was accepted.

## Safety

- immutable index deployment, repository equality, platform selection, and exact child-digest matching remain unchanged
- the gate still runs before the candidate URL, probes, and traffic promotion in both development and production workflows
- no production deployment, approval, traffic mutation, or production workflow was invoked

## Verification

- `python3 .github/scripts/test_check_desktop_backend_release_policy.py` — 14 passed
- `python3 .github/scripts/check-desktop-backend-release-policy.py` — passed
- `python3 -m py_compile .github/scripts/verify_desktop_backend_image_lineage.py` — passed
- `actionlint .github/workflows/desktop_backend_auto_dev.yml .github/workflows/desktop_backend_prod.yml` — passed
- `uvx black --check --line-length 120 --skip-string-normalization .github/scripts/verify_desktop_backend_image_lineage.py .github/scripts/test_check_desktop_backend_release_policy.py` — passed
- `git diff --check` — passed

## Product invariants

None.

Failure-Class: FC-runtime-image-lineage

Closes SCA-193
Related: SCA-191

### DIFF
--- a/.github/scripts/verify_desktop_backend_image_lineage.py
+++ b/.github/scripts/verify_desktop_backend_image_lineage.py
@@ -22,6 +22,7 @@ IMAGE_MEDIA_TYPES = {
 }
 RUNTIME_OS = "linux"
 RUNTIME_ARCHITECTURE = "amd64"
+ATTESTATION_REFERENCE_TYPE = "attestation-manifest"
 
 
 class LineageError(ValueError):
@@ -51,6 +52,8 @@ def verify_lineage(
         raise LineageError(
             f"runtime repository {runtime_repository!r} does not match build repository {build_repository!r}"
         )
+    if manifest.get("schemaVersion") != 2:
+        raise LineageError("build image manifest must use OCI/Docker schema version 2")
 
     media_type = manifest.get("mediaType")
     if media_type in INDEX_MEDIA_TYPES:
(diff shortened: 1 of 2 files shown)


.github/scripts/verify_desktop_backend_image_lineage.py
```diff
@@ -71,6 +74,11 @@ def verify_lineage(
                 f"(found {len(runtime_descriptors)})"
             )
         runtime_descriptor = runtime_descriptors[0]
+        annotations = runtime_descriptor.get("annotations")
+        if annotations is not None and not isinstance(annotations, dict):
+            raise LineageError("linux/amd64 runtime descriptor has malformed annotations")
+        if annotations and annotations.get("vnd.docker.reference.type") == ATTESTATION_REFERENCE_TYPE:
```
Git-on-my-level, line 84: Blocking concern: this checks that `annotations` is a dict but does not enforce the OCI descriptor contract that annotations are a string-to-string map. At this exact head, `annotations = {"vnd.docker.reference.type": 7}` passes `verify_lineage` and returns passed lineage even though the descriptor is malformed. Please reject non-string annotation keys/values (or at minimum a present non-string reference-type value) and add a regression fixture. Missing/`null` annotations should remain accepted.

Git-on-my-level (commented):
Exact-head review of `80a6fa1974d8f0e9bb84ff47d276d9858f6daf1f`: NOT APPROVED. The exact attestation string is rejected, but malformed annotation values still fail open. OCI descriptor `annotations` is a string-to-string map; the verifier accepts `{"vnd.docker.reference.type": 7}` and returns passed lineage. This violates the required malformed/non-string annotation and fail-closed schema boundary. Please validate annotation entry types and add a focused regression fixture. No merge was attempted.

Git-on-my-level (commented):
CLEAN exact-head re-review of `2088ff1d00cfa56c9b9539e5fd50122d5bc1e5d1`. The prior numeric-annotation blocker is repaired: numeric `7`, boolean/list/null values, non-string keys, and non-object annotations reject; missing/null/empty/valid string maps remain accepted; exact attestation rejects; OCI/Docker index/direct-manifest compatibility and all prior identity invariants remain intact. Focused tests and static/workflow checks pass. I attempted formal APPROVE, but GitHub rejected it because the available reviewer identity is the same account that opened this PR. No merge or bypass was attempted; an independent GitHub identity is still required by protection.

PR #10713, opened 2026-07-27
test(desktop-windows): honor locale in conversation date grouping

tianmind-studio:
## Summary

- derive the older conversation-section label assertion from the runtime locale
- document that non-relative section labels are locale-formatted
- preserve the existing product behavior instead of forcing English output

## Root cause

`groupConversationsByDate()` deliberately formats older section labels with `toLocaleDateString(undefined, ...)`, so the operating-system locale controls the rendered date. The test hard-coded `Jan 5, 2026`, which passes on an English CI host but fails on a Chinese Windows locale where the correct product output is `2026年1月5日`.

Native Windows reproduction on current `main`:

```text
filtering.test.ts: 22 passed, 1 failed
expected: Jan 5, 2026
received: 2026年1月5日
```

## Durable guard

The assertion now computes the expected older label with `Intl.DateTimeFormat` using the same runtime locale. It still pins the `Today` and `Yesterday` special labels, section order, and newest-first row ordering, while allowing the dated label to follow the user's locale.

## Validation

- `vitest run src/renderer/src/lib/conversations/filtering.test.ts` -> 23 passed
- ESLint on both changed files -> passed
- renderer TypeScript check -> passed
- `git diff --check` -> passed

None. This changes a platform-dependent test expectation and its type documentation only.

## Failure class

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10713?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/renderer/src/lib/conversations/filtering.test.ts
+++ b/desktop/windows/src/renderer/src/lib/conversations/filtering.test.ts
@@ -179,7 +179,12 @@ describe('groupConversationsByDate', () => {
       row({ id: 'older', sortAt: new Date(2026, 0, 5, 10).getTime() })
     ]
     const sections = groupConversationsByDate(rows, NOW)
-    expect(sections.map((s) => s.label)).toEqual(['Today', 'Yesterday', 'Jan 5, 2026'])
+    const olderLabel = new Intl.DateTimeFormat(undefined, {
+      month: 'short',
+      day: 'numeric',
+      year: 'numeric'
+    }).format(new Date(2026, 0, 5))
+    expect(sections.map((s) => s.label)).toEqual(['Today', 'Yesterday', olderLabel])
     // Within Today, newest first.
     expect(sections[0].rows.map((r) => r.id)).toEqual(['today-pm', 'today-am'])
   })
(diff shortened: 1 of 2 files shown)


tianmind-studio:
Windows validation update (native Windows 11, `win32-x64`, system locale `zh-CN`):

- Runtime: Node `v22.23.1` (supported Node 22 line), pnpm `11.9.0`.
- Current main (`73ce2d65186413c840b4f95bcb31fba20fddc078`) with #10665's runtime guard applied has exactly one Vitest failure: `groupConversationsByDate > buckets into Today / Yesterday / dated, newest section first`. The test expects `Jan 5, 2026`, while the real locale-aware output is `2026年1月5日`.
- This PR at `2f25fb78100ab5bc167f26db30bf2a0d4276b043`: focused test passes `23/23`.
- This PR full Vitest suite passes: `1843/1843` suites, `5052` passed, `18` skipped, `0` failed (`5070` total).

Commands used:
```powershell
pnpm exec vitest run src/renderer/src/lib/conversations/filtering.test.ts
pnpm exec vitest run --reporter=json --outputFile=<temp-report>
```

This confirms the change removes the only locale-dependent failure observed on the supported Node runtime without affecting the rest of the Windows suite.

kodjima33 (approved):
Verified: groupConversationsByDate intentionally uses toLocaleDateString(undefined, ...), so the hard-coded 'Jan 5, 2026' assertion failed under a non-English locale; the test now derives the label the same way and product behavior is unchanged.

PR #10714, opened 2026-07-27
test(release): retire orphaned beta promotion scripts

### LINKED ISSUE #10668: Two desktop beta-release scripts are orphaned from the pipeline but load-bearing as test fixtures
`.github/scripts/mark-desktop-release-beta.py` (68 lines) and `prepare-desktop-beta-promotion.py` (156 lines) are invoked by nothing. Commit `1b15d8db6e` ("fix(release): narrow qualified beta promotion endpoint") deleted both invocations from `desktop_promote_beta.yml` — 249 lines removed from that workflow — and moved beta promotion to a server-owned endpoint (`backend/desktop_qualification_admission.py`, `backend/database/desktop_update_channels.py`, both present on main). A repo-wide grep finds no remaining reference from any workflow, composite action, `Makefile`, `scripts/`, or doc:

```
$ grep -rn 'mark-desktop-release-beta\|prepare-desktop-beta-promotion' --exclude-dir=.git .
backend/tests/unit/test_desktop_release_scripts.py:28  mark_beta = _load(...)
backend/tests/unit/test_desktop_release_scripts.py:29  prepare_beta = _load(...)
backend/tests/unit/test_desktop_release_scripts.py:632 assert 'prepare-desktop-beta-promotion.py' not in workflow
```

**The catch, and why this is not a straight delete.** `prepare_beta.prepare_manifest(...)` is used **16 times** in `test_desktop_release_scripts.py` as the manifest *fixture builder* for tests that cover the still-live v1 manifest contract — including lines 457 and 654, which feed its output into `normalize_release_manifest(...)`, plus the promotion-policy and immutability tests. Deleting the scripts today would remove real coverage of code that is still in the pipeline, so retiring them means first extracting a manifest fixture builder that does not depend on a retired script.

Worth deciding either way, because the current state is ambiguous: the scripts read as live release tooling to anyone browsing `.github/scripts/`, they carry no `LIFECYCLE` header, and nothing records that the pipeline moved off them. If they are intentionally kept as a manual break-glass path, that is worth a line in the release runbook — nothing currently documents invoking them.

Found while auditing `.github/scripts` for scripts referenced by no lane (same sweep that produced #10667). I am not opening a PR because the fixture extraction changes release-contract test coverage, which wants a maintainer's call rather than mine.

tianmind-studio:
## Summary

- extract a shared test-only builder for valid desktop release manifest v1 payloads
- remove the two beta-promotion scripts left orphaned after promotion moved to the server-owned endpoint
- remove six tests that only exercised the retired script behavior while retaining manifest validation, immutable registration, pointer, and repair coverage
- make release workflow fixture reads explicitly UTF-8 so the suite runs under the native Windows locale

Closes #10668

## Context

`mark-desktop-release-beta.py` and `prepare-desktop-beta-promotion.py` have no workflow, script, Makefile, or documentation callers. The latter remained load-bearing only because `test_desktop_release_scripts.py` imported its `prepare_manifest()` function to construct otherwise valid test payloads.

The live promotion path now derives and verifies manifests in `backend/utils/qualified_beta_promotion.py`; its focused tests cover trusted evidence, asset digests, release identity, and canonical manifest construction. This change therefore moves only reusable payload construction into `backend/tests/unit/fixtures/`, keeps downstream contract tests intact, and deletes tests whose production owner no longer exists.

## Validation

- `python -m pytest tests/unit/test_desktop_release_scripts.py tests/unit/test_desktop_update_channels.py -q` - 68 passed on Windows without `PYTHONUTF8`
- `python -m pytest tests/unit/test_qualified_beta_promotion.py tests/unit/test_desktop_updates.py -q` - 300 passed
- `python .github/scripts/test_desktop_release_manifest.py` - 21 passed
- `python .github/scripts/check-release-process-guards.py` - passed with Git for Windows Bash selected explicitly
- all 19 remaining diff-selected manifest checks passed; the async blocker scanner used `PYTHONUTF8=1` because it otherwise reads repository UTF-8 source with the Windows ANSI locale
- `python .github/scripts/pr_preflight.py --suggest --base origin/main` - no product invariants affected; no failure-class declaration required

The unmodified local manifest runner still has Windows-only launcher/path assertions: bare `bash` resolves to the WSL stub, and one self-test compares an MSYS `/tmp/...` path with a native `C:\...` path. The corresponding Linux CI commands are unchanged.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10714?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
Retiring the orphaned beta-promotion scripts is correct now that promotion is server-owned; approving as cleanup rather than a bug fix, so not auto-merging.

Git-on-my-level (commented):
Thanks for the cleanup — I reviewed the current head.

This looks directionally good to me: the two removed beta-promotion scripts appear to be orphaned, with promotion now owned by the backend admission endpoint, and the remaining tests keep the manifest validation / immutable registration / pointer / repair coverage by moving the shared valid manifest payload into a test fixture. I also checked for remaining references to the retired script names and did not find live callers.

Because this touches release/workflow-adjacent machinery, I’m not treating this as an automation-approvable change; a release-owner/maintainer should still be comfortable that the old script path is intentionally retired. No blocking issues found from my review.

---

PR #10715, opened 2026-07-27
fix(preflight): decode async scanner source as UTF-8

tianmind-studio:
## Summary

- decode Python source files as UTF-8 in the async blocker scanner
- add a regression test that records the explicit source encoding
- normalize an existing diff-scope fixture path so the scanner test suite runs natively on Windows

## Root cause

`scan_async_blockers.py` opened repository source files without an explicit encoding. On Windows, Python therefore used the active ANSI code page (GBK on the reproduced machine) and crashed on valid UTF-8 source before the preflight scan could run. The first observed failure was the em dash in `backend/agent-proxy/main.py`.

Repository Python sources are UTF-8, so the scanner now declares that contract at the file boundary. The regression test fails against the previous implementation on every platform because it verifies the encoding passed to `open`.

The adjacent scope test also constructed its key with native separators while production findings are normalized to POSIX separators. Normalizing the fixture removes an unrelated Windows-only failure and keeps the production path unchanged.

## Validation

- production scanner, without `PYTHONUTF8`: 452 `def` and 115 `async def` functions scanned; zero selected or legacy findings
- `backend/tests/unit/test_scan_async_blockers.py` and `backend/tests/unit/test_async_resource_correctness.py`: 33 passed
- Black and `git diff --check`
- `pr_preflight.py --suggest`: no product invariants; `Failure-Class: none`

## Failure class

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10715?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/scan_async_blockers.py
+++ b/backend/scripts/scan_async_blockers.py
@@ -496,7 +496,7 @@ def scan_dirs(dirs: Sequence[str]) -> Dict[str, Any]:
     files_scanned = 0
 
     for fpath in collect_py_files(dirs):
-        with open(fpath) as f:
+        with open(fpath, encoding="utf-8") as f:
             source = f.read()
         try:
             tree = ast.parse(source, filename=fpath)
--- a/backend/tests/unit/test_scan_async_blockers.py
+++ b/backend/tests/unit/test_scan_async_blockers.py
@@ -576,7 +576,7 @@ def test_diff_scope_includes_changed_transitive_helper_lines(scanner, tmp_path):
     finding = results["medium_file_io"][0]
     sink_line = finding["calls"][0]["sink_line"]
     scope = {
-        "ranges": {str(source_path): [(sink_line, sink_line)]},
+        "ranges": {scanner._normalize_path(str(source_path)): [(sink_line, sink_line)]},
         "import_changed_files": set(),
     }
 
--- a/backend/tests/unit/test_scan_async_blockers.py
+++ b/backend/tests/unit/test_scan_async_blockers.py
@@ -33,6 +33,23 @@ def _scan_agent_proxy_source(scanner: ModuleType, tmp_path: Path, source: str):
     return source_path, scanner.scan_dirs([str(service_dir)])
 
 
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Thanks for the focused fix. I reviewed the scanner change and the regression coverage.

This looks like the right direction: opening scanner inputs with `encoding="utf-8"` makes the preflight behavior deterministic across Windows locales, and the new test verifies the actual file boundary rather than only checking parsed output. The adjacent path normalization update is also consistent with the scanner's POSIX-normalized scope keys and removes a Windows-only fixture mismatch without changing production scope logic.

I also ran the targeted unit file locally in a clean env with only the missing test dependency installed:

`env -i HOME=$HOME PATH=$PATH PYTHONPATH=/workspace/omi-pr-10715-review/backend .venv/bin/python -m pytest tests/unit/test_scan_async_blockers.py -q`

Result: `26 passed, 1 warning in 0.12s`.

I am leaving this as a positive maintainer signal rather than a formal approval because this touches a preflight/automation scanner path and the automation approval gate requires human sign-off for workflow-sensitive changes before merge.

---

tianmind-studio:
Fresh native-Windows integration evidence on current main: with this head stacked into the Windows preflight series, `backend-async-blockers` scanned 455 sync and 134 async endpoints, reported 0 selected blockers, and passed. The complete official `make preflight` then passed all 30 selected checks in 223.41s. The effective GitHub checks shown by `gh pr checks` are also green/skipped; the stale red count comes from earlier cancelled duplicate Repo Checks runs.

undivisible (approved):
Reviewed the UTF-8 source handling and its focused regression coverage.

PR #10725, opened 2026-07-27
test(desktop): make PTT HAL liveness deterministic

### LINKED ISSUE #10724: Flaky: PTTInputDeviceProbeTests uses CI throughput as a liveness assertion
## Summary

`PTTInputDeviceProbeTests.testMainRunLoopKeepsPumpingWhileTheHALIsWedged` is flaky in Desktop Swift CI. The production change from #10695 is not implicated by the failures: the test's fixed 300 ms window asserts more than 10 timer ticks and more than 10 snapshot reads, turning a liveness requirement into a runner-throughput requirement.

## Evidence

Two independent failures after #10695 landed:

- current `main`, run [30276084659](https://github.com/BasedHardware/omi/actions/runs/30276084659): `reads == 10`, failing `XCTAssertGreaterThan(reads, 10)`
- unrelated Windows preflight PR #10574, run [30280746072](https://github.com/BasedHardware/omi/actions/runs/30280746072): `ticker.count == 5` and `reads == 4`

In both runs, every sibling `PTTInputDeviceProbeTests` case passed. Other current-main runs pass, confirming the result depends on macOS runner contention.

## Root cause

The acceptance criterion is binary: while the injected HAL probe is blocked off-main, the main run loop and `currentSnapshot` path must continue making progress. Requiring an arbitrary 11+ iterations in 300 ms measures host scheduling throughput instead. Under suite isolation and a loaded arm64 runner, `RunLoop.main.run(mode:before:)` may service only a few iterations even though neither the main actor nor snapshot read is blocked.

## Proposed fix

Keep the HAL semaphore wedged and assert deterministic liveness/progress, not a fixed iteration rate. The test should still fail if the routing read moves back onto the main actor or waits for the HAL, while accepting any positive main-loop and turn-start progress under contention.

No production code change is needed.

tianmind-studio:
Fixes #10724.

## What changed and why

- wait until the injected CoreAudio probe has actually entered its wedged semaphore before checking responsiveness
- replace the fixed `> 10` timer/read throughput requirements in a 300 ms window with three explicit main-actor heartbeat completions and three explicit `currentSnapshot` completions
- keep snapshot reads off-main so a blocking regression times out without preventing the test from releasing and joining the parked HAL probe
- exempt the real Swift package test path, `desktop/macos/Desktop/Tests/`, from the user-facing changelog gate and cover that path classification with regression cases

`testMainRunLoopKeepsPumpingWhileTheHALIsWedged` was testing a binary liveness contract with host-throughput thresholds. Under macOS runner contention, the main actor and snapshot path still made progress, but the isolated suite completed too few iterations. The production contract is that the main actor and turn-start snapshot reads both keep completing while the HAL probe remains blocked off-main, so the new test observes those two completions independently instead of measuring scheduler throughput.

The changelog gate already exempts internal test paths because its post-merge push run cannot see a PR-only `no-changelog-needed` label. It covered the legacy lowercase `desktop/macos/tests/` directory but not the Swift package tests under `Desktop/Tests/`, so this test-only patch exposed the same open failure class on the actual Swift test path.

## Failure evidence

- current `main`: [run 30276084659](https://github.com/BasedHardware/omi/actions/runs/30276084659), `reads == 10`
- unrelated #10574 head: [run 30280746072](https://github.com/BasedHardware/omi/actions/runs/30280746072), `ticker.count == 5`, `reads == 4`
- unrelated #10574 rerun: [run 30283186121](https://github.com/BasedHardware/omi/actions/runs/30283186121), `ticker.count == 7`, `reads == 7`

All sibling `PTTInputDeviceProbeTests` cases passed in those failures.

none

## How it was verified

- `python3 .github/scripts/test_desktop_changelog.py` passed: 4 tests
- `python3 .github/scripts/check-desktop-changelog.py --base origin/main --head HEAD` passed: no desktop changes require a changelog entry
- `python3 desktop/macos/scripts/check_desktop_test_quality.py` passed
- `git diff --check origin/main...HEAD` passed
- reviewed the semaphore ordering: `started` is fulfilled immediately before `release.wait()`, main-actor heartbeat and snapshot-read expectations settle before `release.signal()`, and `finished` is awaited afterward

This contribution was prepared on Windows, so the local commit hook could not run the Xcode-only `swift-format` command (`xcrun` is unavailable). Desktop Swift CI is the authoritative compile, formatter, and XCTest verification for the Swift test change.

## Tests

- `ChangelogRequirementTests` now covers both a top-level and nested Swift package test path.
- The rewritten PTT test is the regression proof for the three linked flaky runs; it fails in bounded time if either the main actor stalls or snapshot reads block while the injected HAL probe is parked.

## Failure class (fixes)

Failure-Class: FC-push-gate-internal-path-scope

## Scoped cleanups

The changelog path exemption is a separate commit because the initial test-only head demonstrated that relying on `no-changelog-needed` would pass only the PR lane and could still redden the post-merge push lane.

### DIFF
--- a/.github/scripts/check-desktop-changelog.py
+++ b/.github/scripts/check-desktop-changelog.py
@@ -51,6 +51,7 @@ EXEMPT_DESKTOP_PATHS = {
 EXEMPT_DESKTOP_PATH_PREFIXES = (
     "desktop/macos/Backend-Rust/",
     "desktop/macos/tests/",
+    "desktop/macos/Desktop/Tests/",
     # Generated Swift (e.g. Sources/Generated/OmiApi.generated.swift) is
     # deterministically derived from the backend OpenAPI contract, never a
     # user-facing app note. Regenerating it after a spec change must not demand
--- a/.github/scripts/test_desktop_changelog.py
+++ b/.github/scripts/test_desktop_changelog.py
@@ -84,6 +84,10 @@ class ChangelogRequirementTests(unittest.TestCase):
             # post-merge push run of the changelog gate reddened main (#10387).
             "desktop/macos/tests/test-qualify-desktop-beta-contract.sh",
             "desktop/macos/tests/some-other-desktop-test.sh",
+            # Swift package tests live under Desktop/Tests rather than the
+            # legacy lowercase tests directory.
+            "desktop/macos/Desktop/Tests/PTTInputDeviceProbeTests.swift",
+            "desktop/macos/Desktop/Tests/Services/SyncTests.swift",
             # Rust backend prefix.
             "desktop/macos/Backend-Rust/src/main.rs",
             # Generated Swift is derived from the OpenAPI contract, never a
--- a/desktop/macos/Desktop/Tests/PTTInputDeviceProbeTests.swift
+++ b/desktop/macos/Desktop/Tests/PTTInputDeviceProbeTests.swift
@@ -19,9 +19,13 @@ final class PTTInputDeviceProbeTests: XCTestCase {
 
   /// A probe whose device lookup blocks until the test releases it, standing in for a
   /// wedged CoreAudio driver without sleeping on the wall clock.
-  private func stalledProbe(until release: DispatchSemaphore) -> PTTAudioDeviceProbe {
+  private func stalledProbe(
+    until release: DispatchSemaphore,
+    onStart: (@Sendable () -> Void)? = nil
+  ) -> PTTAudioDeviceProbe {
     PTTAudioDeviceProbe(
       inputDeviceID: { _ in
+        onStart?()
         release.wait()
         return 99
       },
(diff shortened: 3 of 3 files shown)


tianmind-studio:
Authoritative macOS verification is green on head `9c008a289e`: [Desktop Swift CI run 30286448998](https://github.com/BasedHardware/omi/actions/runs/30286448998) passed the macOS manifest/formatter checks, the full Desktop Swift test suite, and the final `Desktop Swift Build & Tests` verdict. The rewritten PTT liveness regression completed on the real M1 runner without the scheduler-throughput assertions that failed in the three reproduced runs.

Git-on-my-level (commented):
Thanks for the focused fix here. I reviewed the diff and the current head CI evidence.

Positive signal from my side: the test rewrite is a good direction. Waiting for the injected HAL probe to actually enter the wedged path removes the old race, and replacing scheduler-throughput thresholds with explicit main-actor and off-main snapshot-read completions better matches the production contract this regression is supposed to pin. I also like that the parked probe is still released and joined before teardown.

The changelog-gate update also looks narrowly scoped: `Desktop/Tests/` is the real SwiftPM test location and the added unit coverage keeps that exemption from leaking into product source paths.

I’m not formally approving from automation because this touches CI/changelog policy code and the approval gate requires human maintainer sign-off for workflow-sensitive paths, even though this looks clean and the relevant Desktop Swift CI run is green.

---

Git-on-my-level:
Superseded by canonical-repo PR https://github.com/BasedHardware/omi/pull/10742 (same fix rebased onto current main).

PR #10733, opened 2026-07-27
fix(desktop): route Escape by UI layer

### LINKED ISSUE #10698: Escape does nothing on most desktop pages: onExitCommand never receives focus
`DesktopHomeView` wraps its page content in `.onExitCommand { navigateHomeOnEscapeIfNeeded() }` (DesktopHomeView.swift:1109), which is meant to back out of Conversations/Memories/Tasks/Rewind to the dashboard, and to close the full-screen Brain Map page. It never fires. `onExitCommand` is delivered through `cancelOperation(_:)` on the responder chain, and these pages contain no view that takes keyboard focus, so nothing is there to receive it.

Evidence, on a dev bundle of the current `main` behaviour (`com.omi.omi-pr-9499`, automation bridge on 47894):

- Navigate to the Memory hub's Memories destination, activate the app, send `key code 53`. `selectedTab` stays `Memories`; the dashboard is never reached.
- Open a conversation detail via `omi-ctl open-conversation <id>`, send Escape. The detail view stays open.

I also confirmed the SwiftUI-native alternatives do not help here: adding `.onKeyPress(.escape)` to `ConversationsPage`'s body has the same problem for the same reason (I tried it, verified it was inert, and dropped it rather than ship dead code). `RewindPage` uses `.onKeyPress(.escape)` too, so its layered Escape handling is likely inert as well — worth checking.

The pattern that does work in this codebase is a window-scoped `NSEvent.addLocalMonitorForEvents` monitor: `OverlayModalEscapeCatcher` (AppsPage.swift:3382, whose own doc comment says "`onExitCommand` needs focus they never receive"), `MemoryAtlasInputMonitor` in the Brain Map, and `FloatingControlBarWindow`. Three separate copies of it exist already, which is probably the real fix: one shared escape monitor that reports handled/unhandled so pages can layer their dismissals and pass the key on when they have nothing to undo.

Related: PR #9499 gives the Brain Map layered Escape (search, then selection, then the island you entered, then the map itself) using its own monitor, so the atlas is not affected by this. Everything else on the window is.

undivisible:
## Summary

- Routes Escape by layer so modals and active editing handle it before page navigation.

## Validation

- `swift-format`, `swiftc -parse`, and diff checks passed.
- `xcrun swift test --package-path Desktop --filter EscapeKeyDispatchTests` is blocked before tests by unchanged AudioKit macOS availability errors under Xcode Beta.

Closes #10698

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10733?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks for tightening the Escape handling by routing it through explicit UI-layer priorities. The direction looks sensible: modal/editing/content handlers get a chance before the home-navigation fallback, and the new dispatch tests cover the important ordering cases.

I’m requesting changes because the current head does not compile in CI. The new `EscapeKeyDispatchTests.swift` references app types (`TasksViewModel`, `WindowEscapeKeyMonitor`, `LiveNotesEscapeHandling`, and `EscapeKeyPriority` cases) without importing the app target. Existing tests that reach into app code use `@testable import Omi_Computer`, and CI is failing with errors like:

- `cannot find 'WindowEscapeKeyMonitor' in scope`
- `cannot find 'LiveNotesEscapeHandling' in scope`
- `cannot infer contextual base in reference to member 'editing' / 'content'`

Please add the testable app import (or otherwise expose the tested helpers to the test target) and rerun the Desktop Swift checks. After that, this should be straightforward to re-review. I did not see a security/supply-chain concern in the code changes themselves.

---

Git-on-my-level (changes requested):
Thanks for the follow-up here — the previous app-target import blocker looks resolved, and the Desktop Swift test phase now builds/runs the Swift suite successfully.

I’m keeping this in changes-requested because the current head still fails the required Desktop Swift CI. The failing gate is now formatting rather than the test import issue: `Desktop Swift Static & Test Contracts` reports `FORMAT DRIFT` in `desktop/macos/Desktop/Tests/DashboardCaptureStateTests.swift`, then fails `desktop-swift-format-lint`, which keeps `Desktop Swift Build & Tests` red as well.

Please run the pinned desktop Swift formatter on that test file and push the formatted result. Once CI is green, this should be straightforward for a human maintainer to finish reviewing; the Escape priority layering and the added dispatch tests look directionally solid, and I do not see a security/supply-chain concern in this diff.

---

undivisible:
Superseded by #10876, which has the identical head commit on a BasedHardware-owned branch as part of the fork retirement.

PR #10741, opened 2026-07-27
fix(backend): serve legacy memories in chat for un-enrolled accounts

### LINKED ISSUE #10736: In-app chat memory tools still hard-deny on missing_rollout_state — #10094 fixed the Developer API path but not chat
## Summary

#10094 taught the Developer API path to treat `missing_rollout_state` as "un-enrolled, fall back to the legacy store" rather than a hard deny. The in-app chat retrieval tools were not updated and still hard-deny on the same signal, so users in the legacy cohort get "No memories available for this request." from Omi chat while the memories list screen shows their memories normally.

## Where

`backend/utils/retrieval/tools/memory_tools.py:220` and `:381`:

```python
if default_memories.read_decision != MemoryReadDecision.USE_LEGACY_SAFE:
    logger.info(
        "🛑 get_memories_tool - memory chat memory list denied without legacy fallback: "
        f"{default_memories.fallback_reason}"
    )
    return default_memories.text or "No memories available for this request."
```

Same pattern at `backend/utils/retrieval/tool_services/memories.py:84` and `:173`.

The equivalent check in `backend/routers/developer.py:397` was relaxed by #10094:

```python
if memory_result.fallback_reason != 'missing_rollout_state':
    raise HTTPException(status_code=403, ...)
# ... otherwise fall through to the legacy store
```

## Impact

Any account with no `users/{uid}/memory_control/state` document resolves to `DENY_MEMORY` / `missing_rollout_state`, which is the expected un-enrolled state for the legacy cohort — the same population #10094 was written to rescue. For those users, chat cannot see their memories at all. Prod logs show **23 of these denials on 2026-07-27 alone**.

It is silent from the user's side: chat simply answers as though the user has no memories, so it reads as the assistant forgetting rather than as an error.

## Proposed fix

Apply the #10094 treatment here: when `fallback_reason == 'missing_rollout_state'`, fall through to the legacy memories read instead of returning the "no memories" string. Keep the hard deny for the other reasons (e.g. `missing_app_key_scope_grant`).

mdmohsin7:
## Problem

An account with no `users/{uid}/memory_control/state` document resolves to `DENY_MEMORY` / `missing_rollout_state`. For the legacy cohort that is the expected un-enrolled state, not a failure — `pin_memory_system` has already resolved the account to LEGACY, so the legacy `memories` collection is its authoritative read surface.

All four Omi chat retrieval entry points compared the decision against `USE_LEGACY_SAFE` only, so an un-enrolled account fell into the hard-deny branch:

```python
if default_memories.read_decision != MemoryReadDecision.USE_LEGACY_SAFE:
    return default_memories.text or "No memories available for this request."
```

Chat therefore answered "No memories available for this request." while the memories list screen showed the same user's memories normally. It is silent from the user's side: chat reads as the assistant forgetting rather than as an error.

The Developer API (#10094) and hosted MCP (#10095) already treat this signal as "un-enrolled, read legacy". Chat was the remaining consumer that did not.

## Change

**`refactor` commit** — the rule "which denied default-memory read may still serve the legacy collection" had no single owner: it lived inline inside MCP's `mcp_legacy_read_authorized`, including its own `record_fallback` emission, while the Developer API repeated the reason check at its own call site. This is the third fix against that same cause, so the contract is extracted rather than pasted a fourth time:

- New `legacy_read_fallback_authorized(read_decision, fallback_reason)` in `utils/memory/default_read_rollout.py`, beside the `MemoryReadDecision` enum it interprets. It owns the reason set and the recovered-branch fallback telemetry.
- `mcp_legacy_read_authorized` becomes a thin MCP-shaped wrapper over it. No behavior change.

**`fix` commit** — the four chat entry points route through that contract:

| File | Entry points |
|---|---|
| `utils/retrieval/tools/memory_tools.py` | `get_memories_tool`, `search_memories_tool` |
| `utils/retrieval/tool_services/memories.py` | `get_memories_text`, `search_memories_text` |

**`chore` commit** — registers `FC-absent-rollout-state-as-denial`, since `Failure-Class: new` is rejected unless the definition lands in the same range. It ships with its guard artifact, so it needs no ratchet allowlist entry.

## Scope of the relaxation

The contract allows a legacy read for exactly two decisions:

- `USE_LEGACY_SAFE` — the explicitly configured legacy surface (no telemetry; it is not a recovery).
- `DENY_MEMORY` + `missing_rollout_state` — never enrolled; records the recovered fallback.

Everything else stays fail-closed, because each means "we could not establish this account's state" and must never be read as "never enrolled": `malformed_rollout_state`, `rollout_read_failed`, `missing_chat_default_memory_grant`, `missing_mcp_default_memory_grant`, `missing_developer_default_memory_grant`, `memory_reads_disabled`, `uid_mismatch`, `unsupported_consumer`, `unsupported_rollout_schema`, a `None` reason, plus `SHADOW_ONLY` and `USE_MEMORY` under any reason.

The predicate is null-safe by construction: `fallback_reason` is a declared field on every real decision result (`ChatMemorySearchResult` requires it; `McpMemorySearchResult` / `McpMemoryListResult` default it to `None`), and a `None` reason simply is not the un-enrolled reason, so it fails closed rather than raising.

One correction to the issue's proposed fix: it gives `missing_app_key_scope_grant` as the example reason that must keep failing closed. That reason is not reachable on the chat path — there is no app-key gate in chat retrieval, and chat's grant denial is `missing_chat_default_memory_grant`. The tests use the reachable reason instead.

## Import purity

`record_fallback` is imported inside the predicate rather than at module scope. `utils/memory/default_read_rollout.py` is the light decision surface that read paths and admin routers import under heavy test stubbing, and `utils.observability` reaches `utils.metrics` → `from fastapi import Response`. A module-level import broke `test_non_active_route_admin_endpoint.py` and `test_product_memory_router.py`, which stub `fastapi`. Function-level keeps this module import-time pure and matches existing practice in `utils/readiness.py` and `routers/users.py`.

## Guard surface

Three fixes now share this cause (#10095 MCP, #10485 Developer API vector, this one). The reusable guard is the shared predicate plus a behavioral matrix test over it in `tests/unit/test_default_read_rollout_decision.py`, so relaxing one consumer cannot silently relax another and no newly added deny reason becomes a legacy fallback by default. That path is recorded as the failure class's `canonical_prevention_artifact`.

## Tests

Added `tests/unit/test_chat_memory_unenrolled_legacy_fallback.py`, parametrized over all four chat entry points. Both directions are asserted per surface, since the risk in relaxing a fail-closed check is relaxing it too far:

- un-enrolled deny reaches the legacy read and returns the memory — **4 failed before this change, pass after**
- every other deny reason refuses and performs **no** legacy read — passed before and after, so the relaxation is bounded

Added the shared-contract matrix test described above.

Three test doubles needed the real decision shape, because the shared contract reads `fallback_reason` where the old comparison short-circuited on `read_decision` alone. None of these changes touches an assertion, and the real types already declare the field:

- `test_lock_bypass_fixes.py`, `test_mcp_search_memories.py` — their `SimpleNamespace` doubles now carry `fallback_reason`.
- `test_tools_router.py` — its hand-built `default_read_rollout` stub now exposes the contract and the full decision enum.

The static tripwire in `test_chat_memory_adapter.py` asserted the old source text. It now asserts the shared contract and is labelled as a source-order tripwire, with behavior covered by the new test.

## Verification

Each commit was verified green in isolation so the history bisects cleanly; the `refactor` commit was checked out into a separate worktree and its nine affected files run there (262 passed).

```
BACKEND_UNIT_TEST_FILE_LIST=<11 files> bash test.sh   ->  295 passed, 0 failed
scripts/pr-preflight --pr-body-file ...               ->  21 checks PASS
scripts/failure-class validate --pr-body-file ...     ->  ok: true
```

| File | Result |
|---|---|
| `test_chat_memory_unenrolled_legacy_fallback.py` | 8 passed |
| `test_default_read_rollout_decision.py` | 21 passed |
| `test_chat_memory_adapter.py` | 11 passed |
| `test_mcp_memory_adapter.py` | 27 passed |
| `test_mcp_search_memories.py` | 20 passed |
| `test_non_active_route_admin_endpoint.py` | 17 passed |
| `test_product_memory_router.py` | 8 passed |
| `test_lock_bypass_fixes.py` | 61 passed |
| `test_tools_router.py` | 87 passed |
| `test_dev_api_canonical_grant_ordering.py` | 10 passed |
| `test_developer_memory_adapter.py` | 25 passed |

The existing MCP and Developer API characterization tests (`test_mcp_memory_adapter.py::test_mcp_legacy_read_authorized_only_for_explicit_or_unenrolled`, and `test_dev_api_canonical_grant_ordering.py`'s legacy-fallback and still-403 route tests) were not modified and pass unchanged, which is the evidence that neither existing consumer shifted.

Local runs used the `python3` fallback documented in `test.sh` (3.12.3 + pytest 9.0.3) because the pinned CPython 3.11.15 has no `linux-aarch64` build, so `scripts/sync-python-deps.sh` cannot create the pinned venv on this machine. CI on the pinned interpreter is the authoritative gate for this PR.

Duration guard: this machine trips `BACKEND_FAST_UNIT_FAIL_SECONDS=0.12` at exactly `0.12s` inside `test_dev_api_canonical_grant_ordering.py`, on a different test each run, and reproduces identically on untouched `origin/main`. Runs above therefore use the `1.0s` ceiling CI applies for cross-machine CPU differences. The new tests pass under the strict `0.12s` guard.

Closes #10736

Invariants: INV-MEM-1, INV-MEM-3
Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10741?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-absent-rollout-state-as-denial.json
+++ b/.github/failure-classes/FC-absent-rollout-state-as-denial.json
@@ -1,0 +1,23 @@
+{
+  "schema_version": 1,
+  "id": "FC-absent-rollout-state-as-denial",
+  "violated_contract": "An absent optional rollout-state document means the account was never enrolled, which is a served state rather than a denial; a fail-closed default-memory read gate must not refuse a read the account's authoritative legacy surface can serve.",
+  "canonical_prevention": "Interpret every default-memory read decision through the one shared legacy-fallback contract, which serves the legacy surface only for an explicit legacy-safe decision or a deny whose sole reason is the absent rollout-state document, and keeps every other deny reason fail-closed so an indeterminate state is never mistaken for never-enrolled.",
+  "canonical_prevention_artifact": [
+    "backend/utils/memory/default_read_rollout.py",
+    "backend/tests/unit/test_default_read_rollout_decision.py"
+  ],
+  "evidence_prs": [
+    10094,
+    10095,
+    10485
+  ],
+  "scope_hints": [
+    "backend/utils/memory/default_read_rollout.py",
+    "backend/utils/mcp_memories.py",
+    "backend/utils/retrieval/tool_services/**",
+    "backend/utils/retrieval/tools/**",
+    "backend/routers/developer.py"
+  ],
+  "status": "open"
+}
--- a/.github/lifecycle-header-baseline.txt
+++ b/.github/lifecycle-header-baseline.txt
@@ -12,7 +12,6 @@ backend/scripts/shared_ns2_legacy_isolation_readiness.py
 backend/scripts/v3_dev_cloud_readiness.py
 backend/scripts/vector_repair_outbox_oidc_iam_proof.py
 backend/scripts/vector_search_provider_readiness.py
-backend/utils/memory/default_read_rollout.py
 backend/utils/memory/memory_read_rollout_core.py
 backend/utils/memory/v3/compatibility.py
 backend/utils/memory/v3/limited_rollout_config.py
--- a/backend/tests/unit/test_chat_memory_adapter.py
+++ b/backend/tests/unit/test_chat_memory_adapter.py
@@ -42,6 +42,11 @@ def _enabled_rollout_doc(uid='u1'):
 
 
 def test_chat_memory_tool_wires_memory_adapter_before_legacy_vector_search():
+    # Static tripwire (source order, not behavior): the chat tool must consult the default-read
+    # decision before it can reach the legacy vector index. The decision is interpreted by the
+    # shared legacy-fallback contract, so an un-enrolled account reads legacy while every other
+    # deny reason stays fail-closed; behavior is covered in
+    # tests/unit/test_chat_memory_unenrolled_legacy_fallback.py.
     memory_tools_py = Path(__file__).resolve().parents[2] / 'utils' / 'retrieval' / 'tools' / 'memory_tools.py'
     contents = memory_tools_py.read_text(encoding='utf-8')
     rollout_call = 'search_memory_default_chat_memories_vector_decision_text('
@@ -50,7 +55,7 @@ def test_chat_memory_tool_wires_memory_adapter_before_legacy_vector_search():
     assert legacy_call in contents
     assert contents.index(rollout_call) < contents.index(legacy_call)
     assert 'if default_memories is not None:' not in contents
-    assert 'MemoryReadDecision.USE_LEGACY_SAFE' in contents
+    assert 'legacy_read_fallback_authorized(' in contents
 
 
 def test_chat_rollout_reader_supports_omi_chat_grant_without_reading_memory_items():
--- a/backend/tests/unit/test_chat_memory_unenrolled_legacy_fallback.py
+++ b/backend/tests/unit/test_chat_memory_unenrolled_legacy_fallback.py
@@ -1,0 +1,142 @@
+"""Omi chat serves legacy memories for un-enrolled accounts instead of hard-denying (#10736).
+
+An account with no ``users/{uid}/memory_control/state`` document resolves to
+``DENY_MEMORY`` / ``missing_rollout_state``. That is the expected state for the legacy
+cohort, not a failure: ``pin_memory_system`` has already resolved the account to LEGACY,
+so the legacy ``memories`` collection is its authoritative surface.
+
+The Developer API (#10094) and hosted MCP (#9892) already treat that signal as
+"un-enrolled, read legacy". The four chat retrieval entry points did not, so chat answered
+"No memories available for this request." while the memories list screen showed the same
+user's memories normally — it read as the assistant forgetting, not as an error.
+
+Both branches are asserted per surface, because the risk in relaxing a fail-closed check is
+relaxing it too far: an un-enrolled deny must reach legacy, and every other deny reason must
+still refuse to read it.
+"""
+
+import os
+from datetime import datetime, timezone
+
+import pytest
+
+os.environ.setdefault(
+    "ENCRYPTION_SECRET",
+    "omi_ZwB2ZNqB2HHpMK6wStk7sTpavJiPTFg7gXUHnc4tFABPU6pZ2c2DKgehtfgi4RZv",
+)
+os.environ.setdefault("OPENAI_API_KEY", "test-openai-key-not-real")
+
+import utils.retrieval.tool_services.memories as memories_svc  # noqa: E402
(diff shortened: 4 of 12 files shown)


kodjima33 (approved):
Un-enrolled legacy accounts should read legacy memories in chat rather than hard-deny; needs a rebase before it can land.

undivisible:
**Maintainer note (rebase attempt):** Functional fix already landed on main as #10752 (`chat_legacy_read_authorized` in `backend/utils/memory/chat_memory_adapter.py`, with `tests/unit/test_chat_memory_missing_rollout_fallback.py`).

This branch uses an earlier parallel API (`legacy_read_fallback_authorized` on `default_read_rollout`) and conflicts with the landed helper. `maintainerCanModify` is false on the PR metadata, but the head is on BasedHardware/omi; write access exists — still, re-pushing a divergent reimplementation would fight main. Recommend **closing as superseded by #10752**.

aryanorastar (changes requested):
Requesting changes (merge realism only — product direction looks right):

Treating un-enrolled `missing_rollout_state` / `DENY_MEMORY` as **legacy-read-authorized** (via `legacy_read_fallback_authorized`) instead of hard-denying chat memories is the correct cohort behavior, and the focused unit coverage + FC-absent-rollout-state-as-denial are the right shape. Backend suite looks green on the recorded runs.

**Blocker:** tip is `CONFLICTING` / `DIRTY` vs `main`. Size-gate APPROVE does not survive a conflicted tip — please rebase and keep the unenrolled-legacy tests green on the new tip.

No product objection from this pass; re-review after rebase.

aryanorastar:
Closing as superseded by #10752, which merged on July 28 and closed #10736 with the same chat-memory fallback behavior.

Current `main` has since moved further to the universal memory service: `test_chat_memory_missing_rollout_fallback.py` verifies arbitrary-account retrieval and asserts that chat retrieval contains no cohort or legacy selector. This branch remains unchanged at `118933f`, conflicts across the replaced memory paths, and its earlier parallel helper is no longer the current architecture.

The August rebase path is therefore no longer applicable; no functionality from this branch still needs to land.

PR #10743, opened 2026-07-27
fix(backend): report denied MCP memory reads instead of an empty success

### LINKED ISSUE #10735: Hosted MCP returns an empty 200 instead of an error when a memory read is denied
## Summary

When the memory authorization check denies a hosted-MCP memory read, the handler returns an empty result set with a `200` instead of surfacing an error. The client sees "you have no memories" rather than "this key is not authorized", which is the failure mode most likely to make a user believe their data was deleted.

## Where

`backend/routers/mcp_sse.py:866` (and the same shape at line 1094 for the vector-search path):

```python
if memory_list_results.read_decision == MemoryReadDecision.USE_MEMORY:
    return {"memories": memory_list_results.memories}
if not mcp_legacy_read_authorized(memory_list_results):
    return {"memories": []}
```

## Why it matters

The two denial reasons that reach this branch are `missing_rollout_state` and `missing_app_key_scope_grant` (see #10734 for the grant-seeding gap). Both are server-side authorization states the user cannot see or act on. An empty `200` is indistinguishable from a genuinely empty account, so:

- the user has no signal that anything is wrong and nothing to report,
- support cannot distinguish it from "user has no memories yet" without a per-account Firestore trace,
- it silently contradicts the memories list screen, which shows the same user's memories fine.

Observed on real accounts: MCP keys created without scopes have no grant, so every memory read through them returns empty while the account has thousands of memories in the legacy store.

## Proposed fix

Return an explicit error (or at minimum an MCP-level error payload carrying `reason`) instead of `{"memories": []}` when the denial is an authorization state rather than an actual empty result. Distinguishing "denied" from "empty" is the whole ask here.

mdmohsin7:
## Problem

When the memory authorization check denied a hosted-MCP memory read, the handler returned an empty result set with a success status:

```python
if not mcp_legacy_read_authorized(memory_list_results):
    return {"memories": []}
```

The client saw "you have no memories" rather than "this key is not authorized" — the failure mode most likely to make a user believe their data was deleted. An empty success is indistinguishable from a genuinely empty account, so the user had no signal anything was wrong and nothing to report, support could not separate the two without a per-account Firestore trace, and it silently contradicted the memories list screen, which showed the same user's memories fine.

## Scope: the issue named 2 sites, there are 4

`routers/mcp.py` is the REST MCP surface and carries the identical shape. Fixing only the SSE pair would repair the symptom on one surface and leave the same contract violated on the other, so all four are fixed here. **The widened scope is deliberate, not creep.**

| Site | Tool / route | Before |
|---|---|---|
| `routers/mcp_sse.py:866` | `get_memories` | `return {"memories": []}` |
| `routers/mcp_sse.py:1094` | `search_memories` | `return {"memories": []}` |
| `routers/mcp.py:399` | `GET /v1/mcp/memories` | `return []` |
| `routers/mcp.py:303` | `GET /v1/mcp/memories/search` | `return []` |

## Change

One owner for "report this, or serve an empty result": `mcp_denied_read_payload` in `utils/mcp_memories.py` returns the finished client-visible payload for every authorization or indeterminate-rollout state, and `None` for `SHADOW_ONLY`. The payload mirrors the existing grant-denial shape so both refusals read the same to a client.

Each surface then adds only its own raise, keeping its existing error mechanism:

- `routers/mcp.py` → `HTTPException(403, detail={'enabled': False, 'reason': ..., 'consumer': 'mcp'})`
- `routers/mcp_sse.py` → `ToolExecutionError(str(payload), code=-32009)`, the code already used for MCP authorization failures

### Commit order note

The first two commits landed a two-function classifier plus a per-surface raiser helper, which grew the ratchet-frozen `routers/mcp_sse.py` by 16 lines. The third commit collapses that to the single-payload shape above, cutting the frozen file's growth to 7 lines and shrinking `routers/mcp.py` by 9. Those commits are already pushed and this repo does not force-push, so the reshape is a follow-up commit rather than a rewrite; the description above is the shape being merged.

`backend/routers/mcp_sse.py` is frozen by the oversized-product-file ratchet at 1858 lines. Its baseline is raised to the exact new count (1865) with a justification, because the remaining 7 lines are the two raises this fix exists to add and cannot be removed without leaving a surface unrepaired.

## Which reasons are reported

The states the caller cannot see or act on: `malformed_rollout_state`, `rollout_read_failed`, `missing_mcp_default_memory_grant`, `uid_mismatch`, `unsupported_consumer`, `unsupported_rollout_schema`, `memory_reads_disabled`.

### Two corrections to the issue

**1. The stated reason set is wrong, and so is the scale claim.** The issue says the two reasons reaching this branch are `missing_rollout_state` and `missing_app_key_scope_grant`. Neither can reach it:

- `missing_app_key_scope_grant` is refused earlier by `authorize_memory_external_default_memory_read` — `mcp_sse.py:826`/`:1076` raise `ToolExecutionError` code `-32009`, and `mcp.py:277` raises `HTTPException`. It never reaches the empty return.
- `missing_rollout_state` is allowed *through to legacy* by `mcp_legacy_read_authorized` (`utils/mcp_memories.py:130`), so it does not reach the empty return either.

Consequently the issue's evidence — "MCP keys created without scopes have no grant, so every memory read through them returns empty while the account has thousands of memories" — does not hold: those keys already receive an explicit `-32009` error. The bug is real for the reason set above, but the affected population is narrower than the issue implies, and the headline scale claim is unsupported. No live measurement was taken.

**2. `SHADOW_ONLY` is deliberately left returning empty.** It is not an authorization denial. `docs/epics/memory_normative_architecture.md` defines shadow mode as "Legacy authoritative; memory audit artifacts only", so converting it to an error would be wrong in the other direction. It is a separate concern from this fix and is untouched here.

## Tests

Added to `tests/unit/test_mcp_data_endpoints.py`, which already exercises both the REST handlers and MCP tool dispatch. Parametrized over all four sites × the seven reported reasons, asserting both directions:

- every reported reason raises carrying its reason, on all four sites — **28 failed before this change, pass after**
- `SHADOW_ONLY` still returns empty without error, on all four sites — passed before and after, so the change is bounded

## Verification

Backend, aarch64, system `python3.12` + `pytest 9.0.3`.

```
BACKEND_UNIT_TEST_FILE_LIST=<15 files> bash test.sh   ->  275 passed, 0 failed
```

Covering `test_mcp_data_endpoints` (66), `test_mcp_memory_adapter` (27), `test_mcp_search_memories` (20), `test_lock_bypass_fixes` (61), `test_mcp_apps_results` (40), `test_non_active_route_admin_endpoint` (17), `test_dev_api_canonical_grant_ordering` (10), `test_product_memory_router` (8), `test_mcp_conversation_by_id_poison` (7), `test_mcp_conversations_poison` (5), `test_app_client_response_models` (4), `test_mcp_sse_pagination_bounds` (3), `test_mcp_action_item_writes` (3), `test_mcp_profile_contact` (2), `test_mcp_search_conversations_poison` (2).

Local runs used the `python3` fallback documented in `test.sh` because the pinned CPython 3.11.15 has no `linux-aarch64` build, so `scripts/sync-python-deps.sh` cannot create the pinned venv on this machine. CI on the pinned interpreter is the authoritative gate for this PR.

Closes #10735

Invariants: INV-MEM-1
Failure-Class: new

### DIFF
--- a/.github/failure-classes/FC-denial-rendered-as-empty-success.json
+++ b/.github/failure-classes/FC-denial-rendered-as-empty-success.json
@@ -1,0 +1,19 @@
+{
+  "schema_version": 1,
+  "id": "FC-denial-rendered-as-empty-success",
+  "violated_contract": "A read refused by authorization or an indeterminate rollout state must be reported as a refusal carrying its reason; it must not be returned as a successful empty collection, which the caller cannot distinguish from a genuinely empty account.",
+  "canonical_prevention": "Classify a refused read once at the shared boundary into report-with-reason versus legitimately-empty, and have every serving surface raise from that single classification instead of each deciding on its own to return an empty collection.",
+  "canonical_prevention_artifact": [
+    "backend/utils/mcp_memories.py",
+    "backend/tests/unit/test_mcp_data_endpoints.py"
+  ],
+  "evidence_prs": [
+    10743
+  ],
+  "scope_hints": [
+    "backend/routers/mcp.py",
+    "backend/routers/mcp_sse.py",
+    "backend/utils/mcp_memories.py"
+  ],
+  "status": "open"
+}
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
@@ -3,7 +3,7 @@
     "backend/routers/apps.py": 2373,
     "backend/routers/chat.py": 1588,
     "backend/routers/developer.py": 2263,
-    "backend/routers/mcp_sse.py": 1858,
+    "backend/routers/mcp_sse.py": 1865,
     "backend/routers/sync.py": 2058,
     "backend/routers/users.py": 2076
   },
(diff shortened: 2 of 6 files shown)


kodjima33 (approved):
Denied MCP memory reads now report the refusal instead of an indistinguishable empty success; scoped, tested, CI green.

PR #10744, opened 2026-07-27
feat(backend): add the Developer API key memory-grant backfill (dry-run only)

### LINKED ISSUE #10734: Developer API keys created before 2026-06-27 have no memory grant and 403 permanently — no backfill exists
## Summary

Developer API keys created before the memory-grant seeding landed (commit `8e03144a`, 2026-06-27) have no entry under `users/{uid}/memory_control/app_key_memory_grants`. Because the app/key scope-grant check fails closed, those keys get a permanent `403` from `GET /v1/dev/user/memories` regardless of the scopes on the key. There is no backfill for them, so the state is unrecoverable without the developer creating a brand-new key.

This is a **second, independent gate** from the rollout gate fixed by #10094. #10094 is deployed and working (the endpoint went from zero successes to hundreds of `200`s/day), but a large residual population still 403s on this gate.

## Symptom

```
GET /v1/dev/user/memories -> 403
{"detail":{"enabled":false,"reason":"missing_app_key_scope_grant","consumer":"developer_api"}}
```

`/v1/dev/user/conversations` and `/v1/dev/user/action-items` return `200` on the same key, so it reads to the developer as "the memories endpoint is broken", not "my key is missing a grant".

## Cause

`backend/database/dev_api_key.py:97` seeds the grant, but only on the **create** path:

```python
grant_default_read = Scopes.MEMORIES_READ in (resolved_scopes or [])
grant_write = Scopes.MEMORIES_WRITE in (resolved_scopes or [])
if grant_default_read or grant_write:
    seed_developer_api_key_memory_grant(...)
```

`authorize_memory_external_default_memory_read` in `backend/utils/memory/product_authorization.py` (`EXTERNAL_MEMORY_CONSUMERS` at line 84, `reason='missing_app_key_scope_grant'` at line 328) denies when no grant document is found. Keys that predate the seeding commit were never written into the grants tree and nothing backfills them.

`backend/scripts/` has `backfill_mcp_key_full_access.py` (MCP keys only) and the ops readiness script `app_key_memory_grant_assignment_readiness.py`, but no equivalent for `developer_api`.

## Scale

Every Developer API key carrying a memories scope that predates the seeding commit is affected, which is the large majority of keys in use today — the seeding path is only a month old. A random sample of accounts from that cohort found no `developer_api` grant on any of them. Exact figures from the prod trace are available internally on request.

## Proposed fix

A backfill that walks `dev_api_keys` where `created_at < 2026-06-27` and the key carries a memories scope, and seeds the grant from the scopes already on the key — i.e. the same call the create path makes. No new authorization is granted; it only materialises the grant the key's own scopes already imply.

## Workaround (currently the only one)

Create a new Developer API key with the memories scopes selected; the create path seeds the grant and the new key works. Worth noting because the previous support guidance was "regenerating your key does not help", which was correct pre-#10094 and is now wrong.

mdmohsin7:
## Problem

Developer API keys created before the create-path grant seeding landed (`8e03144a`, 2026-06-27 22:54:46 +0700) have no entry under `users/{uid}/memory_control/app_key_memory_grants`. `authorize_memory_external_default_memory_read` fails closed when no grant document is found, so those keys receive a permanent `403` from `GET /v1/dev/user/memories` regardless of the scopes on the key:

```
{"detail":{"enabled":false,"reason":"missing_app_key_scope_grant","consumer":"developer_api"}}
```

`/v1/dev/user/conversations` and `/v1/dev/user/action-items` return `200` on the same key, so it reads to the developer as "the memories endpoint is broken" rather than "my key is missing a grant". Nothing backfills them, so the state is unrecoverable without creating a new key.

This is a second, independent gate from the rollout gate fixed by #10094. The create-path defect itself is already fixed; what remains is the residual data.

## What this PR delivers

**The script and its dry-run mode only. It has not been run against production.** Applying it touches roughly 1,417 accounts, which is the maintainer's call to make, not this PR's.

`backend/scripts/backfill_developer_api_key_memory_grants.py`, modelled on the existing `backfill_mcp_key_full_access.py`:

```
python scripts/backfill_developer_api_key_memory_grants.py              # dry run (default)
python scripts/backfill_developer_api_key_memory_grants.py --limit 200  # bounded dry run
python scripts/backfill_developer_api_key_memory_grants.py --apply      # write grants
```

It walks `dev_api_keys`, selects keys created before the seeding instant that carry a memories capability, and seeds the grant with exactly the call the create path makes:

```python
grant_default_read = has_scope(scopes, Scopes.MEMORIES_READ)
grant_write = has_scope(scopes, Scopes.MEMORIES_WRITE)
if grant_default_read or grant_write:
    seed_developer_api_key_memory_grant(uid, key_id, default_read=..., write=..., db_client=db)
```

Dry run prints the same counts as an apply run without writing, so the cohort can be measured before anyone commits to a write.

## Grants nothing new

The whole safety argument is that no key gains a capability it did not already have:

- `default_read` comes only from `memories:read`, `write` only from `memories:write`; `archive_read` is always `False`.
- A key with neither memory scope is skipped entirely.
- A key that already holds a grant for the same app/key is skipped, so re-running is safe.
- A key whose `created_at` cannot be read is skipped rather than guessed.

### One deviation from the issue's proposed filter

The issue says to select keys that "carry `memories:read`/`memories:write`". Taken literally as a membership test, that misses a legacy shape. Scope resolution here goes through `utils.scopes.has_scope`, the same authority the request path uses:

```python
def has_scope(user_scopes, required_scope):
    """Check if user has required scope. None scopes are treated as read-only."""
    if user_scopes is None:
        return required_scope in READ_ONLY_SCOPES
    return required_scope in user_scopes
```

`READ_ONLY_SCOPES` includes `Scopes.MEMORIES_READ`, so a key with **no stored `scopes` field can already read memories at the request layer** — and therefore also 403s on this gate and needs the grant. A literal `'memories:read' in scopes` test would silently leave that population broken. Write is never inferred this way, because `memories:write` is not a read-only scope.

That means the affected cohort may be slightly larger than the issue's measured 1,763 keys / 1,417 accounts, which counted explicit scopes. The dry run reports `absent_scopes_treated_as_read_only` separately so the difference is visible before applying.

## Tests

`backend/tests/unit/test_backfill_developer_api_key_memory_grants.py` — 12 cases against a path-keyed Firestore fake, pinning the two properties that make the script safe to run:

- **dry run writes nothing at all** (asserted on the fake's write log, not just the printed counts)
- **applying grants exactly the key's own capability**: a read-scoped key gets `default_read` without `write`; a write-scoped key gets `write` without `default_read`; an absent `scopes` field resolves to read-only; a key with no memory scope, or created after seeding, is left untouched

Plus: an existing grant is never rewritten (safe re-run), a second key for the same user is still seeded after the per-uid grant cache is invalidated by the first write, an unusable `created_at` is skipped rather than guessed, a naive `created_at` is read as UTC instead of aborting the scan, `--limit` bounds the scan, and a key with no `user_id` is skipped.

## Verification

Backend, aarch64, system `python3.12` + `pytest 9.0.3`.

```
BACKEND_UNIT_TEST_FILE_LIST=tests/unit/test_backfill_developer_api_key_memory_grants.py bash test.sh
  ->  12 passed, 0 failed
```

Local runs used the `python3` fallback documented in `test.sh` because the pinned CPython 3.11.15 has no `linux-aarch64` build, so `scripts/sync-python-deps.sh` cannot create the pinned venv on this machine. CI on the pinned interpreter is the authoritative gate for this PR.

No production run was performed, and the script was not executed against any live Firestore project. Its behavior is exercised only through the fake in the test above.

Closes #10734

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10744?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/backfill_developer_api_key_memory_grants.py
+++ b/backend/scripts/backfill_developer_api_key_memory_grants.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+"""Seed missing app/key memory grants for Developer API keys created before seeding landed.
+
+Developer API keys created before commit 8e03144a (2026-06-27) have no entry under
+``users/{uid}/memory_control/app_key_memory_grants``. The grant gate
+(``authorize_memory_external_default_memory_read``) fails closed when no grant document is
+found, so those keys receive a permanent 403 from ``GET /v1/dev/user/memories`` regardless
+of the scopes on the key, while ``/conversations`` and ``/action-items`` keep working.
+
+This grants nothing new. It materialises the grant each key's own scopes already imply, by
+making exactly the call the create path in ``database/dev_api_key.py`` makes:
+
+    grant_default_read = has_scope(scopes, Scopes.MEMORIES_READ)
+    grant_write = has_scope(scopes, Scopes.MEMORIES_WRITE)
+    if grant_default_read or grant_write:
+        seed_developer_api_key_memory_grant(...)
+
+Scope resolution deliberately goes through ``utils.scopes.has_scope``, the same authority
+the request path uses, rather than a literal membership test. That matters for one legacy
+shape: ``has_scope(None, ...)`` treats a key with no stored ``scopes`` field as read-only,
+which includes ``memories:read``. Such a key can already read memories at the request
+layer, so it needs the grant too, and a literal ``'memories:read' in scopes`` test would
+skip it. Write is never inferred that way — ``memories:write`` is not a read-only scope.
+
+Keys that already hold a grant for the same app/key are left untouched, so re-running is
+safe. Dry run is the default and performs no writes.
+
+Usage:
+    python scripts/backfill_developer_api_key_memory_grants.py              # dry run
+    python scripts/backfill_developer_api_key_memory_grants.py --limit 200  # bounded dry run
+    python scripts/backfill_developer_api_key_memory_grants.py --apply      # write grants
+"""
+
+from __future__ import annotations
+
+import argparse
+from collections import defaultdict
+from datetime import datetime, timezone
+from pathlib import Path
+import sys
+from typing import Any, Optional, cast
+
+BACKEND_DIR = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(BACKEND_DIR))
+
+from database._client import get_firestore_client
+from database.memory_app_key_grants import (
+    DEVELOPER_API_CONSUMER,
+    DEVELOPER_API_DEFAULT_APP_ID,
+    read_app_key_memory_grants_state,
+    seed_developer_api_key_memory_grant,
+)
+from utils.scopes import Scopes, has_scope
+
+# Developer API keys created before this instant predate the create-path grant seeding
+# (commit 8e03144a, 2026-06-27 22:54:46 +0700). Keys created after it are seeded on create.
+SEEDING_LANDED_AT = datetime(2026, 6, 27, 15, 54, 46, tzinfo=timezone.utc)
+
+
+def _stored_scopes(value: Any) -> Optional[list[str]]:
+    """The key's stored scopes, or None when the field is absent/unusable.
+
+    None is meaningful, not an error: `has_scope` treats it as the legacy read-only
+    default, so it must be preserved rather than coerced to an empty list.
+    """
+    if not isinstance(value, list):
+        return None
+    items = cast(list[Any], value)
+    return [scope for scope in items if isinstance(scope, str) and scope]
+
+
+def _created_before_seeding(created_at: Any) -> Optional[bool]:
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Dry-run-only backfill for missing Developer API key memory grants; clear root cause, test-covered.

PR #10748, opened 2026-07-28
feat(omi-even): omi on Even Realities G2 — Even AI, Hub app, capture and push

Archit-lal:
## What this is

`omi` on Even Realities G2 glasses. Four surfaces, one local bridge:

1. **Even AI answers from Omi** — ask the built-in assistant anything and it answers from your Omi memories, conversations and action items
2. **The `omi` Hub app** — chat, memories, action items on the display
3. **G2 mics feed Omi** — the glasses act as an Omi capture device
4. **Push** — new action items surface on the display

Everything routes through a FastAPI bridge on the Mac, because `/v2/messages` and
`/v4/listen` both require a Firebase ID token — no app key family can reach them
(`backend/utils/other/endpoints.py`). The bridge mints tokens from the desktop
app's existing session via the repo's own `omi-auth-dump.sh`, so there is no
browser login step.

## The constraint everything else follows from

Even's client gives up at about **5 seconds**. This was measured on hardware,
not assumed: a 0.4s reply renders and a 5.6s reply does not, despite the bridge
returning `200 OK` both times. Streaming past the limit was tried and the
glasses reported a network error.

Omi's agent takes 6–18s, so it cannot be the answer path. Instead:

    retrieve facts (memories + conversations, in parallel)   1.4s
    compose them into prose with one LLM call                1.3-2.3s
    -----------------------------------------------------------
    total                                                    ~3s

Compose is a ladder — the local model on this Mac, then Omi's cloud endpoint,
then the raw retrieved lines. Every rung degrades; none can fail the answer.

**Local first, because the cloud rung is capped at 30/hour**
(`utils/rate_limit_config.py`, `"test:prompt": (30, 3600)`) — a day of actually
wearing the glasses spends that. Ollama already runs on the machine hosting the
bridge, so composing there is uncapped and free, and the retrieved facts never
leave the Mac to be turned into a sentence.

## Notable findings

* **Omi's chat stream is not spec SSE.** Frames are `\n\n`-*separated* with bare
  prefixes (`data:`/`think:`/`done:`), newlines escaped as `__CRLF__`, and the
  terminal `done:` is base64. The parser here is ported from
  `app/lib/backend/http/api/messages.dart`, the reference implementation.
* **Ollama sizes its KV cache from the model's declared maximum, not the
  prompt.** `ollama ps` showed 23GB of weights resident at **28 GB** for a
  262144-token context, serving prompts of ~500 tokens. Pinning `num_ctx` to
  4096 took it back to 23 GB.
* **Residency, not speed, is the local risk.** `qwen3.6:35b-a3b` answers in
  1.3–2.3s warm and 16.5s cold — three times the entire budget. A warmer runs
  at startup and every 10 minutes.
* **A server-side date window made search 4× slower** (13.5s vs 3.1s on the live
  account) for the same result, so date filtering happens client-side.
* **`CancelledError` is a `BaseException`**, so `suppress(Exception)` does not
  catch it — an ordinary shutdown printed a traceback until it was named.

## Verification

Live against the real Omi account, through the bridge:

    [2.9s] What did I do on 4 July?
           You shared burgers with friends on July fourth. You handled food prep
           details like burger assembly and serving.

    [3.1s] What do I need to do today?
           You need to complete the UMN Robotics workspace access form and
           investigate the invalid Omi Discord link on omi.me. Additionally,
           rebuild Sage's voice profile by setting xurl OAuth and review PR
           #10463 for the Google Calendar persistence fix.

    [2.0s] What am I behind on?
           You're behind on replying to Tharun, addressing requested changes for
           Omi desktop PR #10687, and submitting your Startup School weekly
           update.

Suites:

    integrations/omi-even/bridge      662 passed
    integrations/omi-even/app         40 unit tests, 82 simulator checks
    integrations/evenhub-g2           reference quickstart, unchanged

Capture was verified by making a real recording through the glasses and
confirming the conversation and transcript appeared in Omi.

## Known limits

* **Even's on-device intent classifier intercepts some utterances** before any
  network call — note-shaped phrasing gets saved to the quicklist and never
  reaches the bridge. Nothing server-side can change that.
* **A Cloudflare quick tunnel gets a new URL on restart**, so the agent has to
  be re-registered in the Even app. A named tunnel fixes this.
* The cloud compose rung appends a conversation transcript that the prompt
  instructs the model to disregard. That held on every question tried, but it
  is an instruction, not a guarantee.

## Not done yet

* `gather_facts` and `fast_answer` each retrieve independently, so the bullet
  fallback re-runs the search it just ran. Harmless today because compose
  almost always wins, but it should be one retrieval feeding both.
* One unexplained 76s stall of the bridge event loop was observed under memory
  pressure while a 28 GB model was resident. The `num_ctx` fix removes the
  suspected cause; catastrophic regex backtracking was measured and ruled out.

none

## Failure class (fixes)

Failure-Class: none

https://claude.ai/code/session_012aNauQfdfh98dZFWbE8PdP

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10748?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -247,3 +247,11 @@ builds-snapshot.md
 .hermes/
 backend/.venv
 backend/.harness/
+
+# omi-even bridge local state
+integrations/omi-even/bridge/.env
+integrations/omi-even/bridge/add-agent-capture.log
+integrations/omi-even/bridge/.venv/
+integrations/omi-even/bridge/__pycache__/
+integrations/omi-even/bridge/.pytest_cache/
+integrations/omi-even/bridge/tests/__pycache__/
--- a/integrations/evenhub-g2/.gitignore
+++ b/integrations/evenhub-g2/.gitignore
@@ -1,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
+
+node_modules
+dist
+dist-ssr
+*.local
+
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
--- a/integrations/evenhub-g2/README.md
+++ b/integrations/evenhub-g2/README.md
@@ -1,0 +1,86 @@
+# Omi G2 First App
+
+The Even Hub [first-app quickstart](https://hub.evenrealities.com/docs/get-started/quickstart/first-app)
+for Even Realities G2 smart glasses: one full-canvas text container that counts
+taps and exits on a double-tap.
+
+```
+Hello from G2!
+
+Tap to count: 0
+Double-tap to exit
+```
+
+## Run it
+
+```bash
+npm install
+npm run dev          # Vite on :5173, bound to the LAN
+```
+
+**In the simulator** (no hardware needed):
+
+```bash
+npm run simulator    # separate terminal
+```
+
+**On real glasses** — phone and laptop on the same Wi-Fi, developer mode enabled
+in the Even Realities companion app:
+
+```bash
+npm run qr           # auto-detects the LAN IP; scan from the phone app
+```
+
+Edits hot-reload on the glasses without rescanning.
+
+## Verify
+
+```bash
+npm run dev          # must be running first
+npm run verify
+```
+
+`scripts/verify-simulator.mjs` drives the simulator's automation HTTP API and
+asserts the app actually works: the bridge connects, the start-up container is
+created, the framebuffer is non-blank at 576x288, two taps advance the counter
+to 2, a double-tap requests the exit dialog, and nothing throws. It decodes the
+PNG framebuffer in pure Node, so there is no image-library dependency.
+
+It needs a GUI (the simulator opens a window), so it is a **local** check and is
+deliberately not wired into CI.
+
+## Notes for anyone extending this
+
+- **Touch input arrives on two different channels.** The simulator delivers tap
+  and double-tap as a `sysEvent`, and only scroll as a `textEvent`. The quickstart
+  code in the docs reads `event.textEvent` only, so taps never reach the handler.
+  `src/main.ts` reads whichever channel the event carries, which works in both
+  places. Verified against SDK 0.0.12:
+
+  | input | delivered as |
+  |---|---|
+  | `click` | `sysEvent: {eventSource: 1}` — `eventType` absent |
+  | `double_click` | `sysEvent: {eventType: 3}` |
+  | `up` / `down` | `textEvent: {containerID: 1, eventType: 1 / 2}` |
+
+  An absent `eventType` means a plain tap; every other event type is explicit.
+
+- **`server.host: true` in `vite.config.ts` is required.** Vite binds to
+  localhost by default and the phone would get connection refused. `strictPort`
+  is on so the dev server fails loudly rather than drifting to :5174 and leaving
+  a QR code pointing at a dead port.
+
+- **TypeScript is pinned to ^5.** `@evenrealities/evenhub-cli` declares a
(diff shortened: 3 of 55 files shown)


Git-on-my-level (changes requested):
Thanks for the very thorough hardware exploration here — the bridge/app split and the amount of regression coverage in the Python bridge are both useful. I’m going to request changes because this adds a new credentialed local bridge and dependency surface, and I found a few blockers before this should be considered mergeable.

Blocking issues:

1. `integrations/omi-even/bridge/server.py` exposes `WS /app` without any authentication or origin check, while that socket can read memories/action items/today, ask Omi using the user’s Firebase-backed session, and start/stop continuous capture. Because the bridge is bound on `0.0.0.0` for phone/glasses access, anything on the LAN — and potentially a browser-originated WebSocket if the user visits a hostile page — can attempt to connect and operate the bridge. Please require the same shared secret (or a dedicated app token) on the WebSocket handshake and reject unexpected origins/hosts before returning personal data or accepting microphone/capture commands.

2. `integrations/omi-even/bridge/server.py` logs the Even bearer token at startup (`bridge starting; Even token: ...`), and `run.sh` sends those logs to `/tmp/omi-even-server.log`. That token is the public-tunnel credential for the Add Agent endpoint; it should be printed only in the intentional setup summary, never emitted by the long-running server logger. Please remove or redact that startup log and add a regression test similar to the existing request-capture redaction tests.

3. The new Node apps are not installable in this Linux review/CI environment because `@evenrealities/sim-darwin-arm64` is a direct devDependency. `npm ci --ignore-scripts` fails in both `integrations/omi-even/app` and `integrations/evenhub-g2` with `EBADPLATFORM` on linux/x64, so the supposedly hermetic unit/typecheck path cannot run here and likely cannot be wired into CI as-is. Please move platform-specific simulator binaries behind optional dependencies or a platform-specific install path so `npm ci && npm test && npm run typecheck` works on non-Darwin hosts.

Agent-behavior documentation note: `integrations/omi-even/docs/add-agent-contract.md` is the kind of contract note future coding/review agents will use as source-of-truth for this integration. That is useful, but it currently says the client timeout is “not measured” and that ~10s answers are fine, while the PR body and bridge implementation are built around a measured ~5s limit. Please make that contract doc consistent with the current evidence (or clearly separate old observations from the newer hardware measurements), otherwise future agents may choose the wrong response path when modifying this bridge.

Validation I ran:
- `python -m pytest -q integrations/omi-even/bridge/tests` → 662 passed.
- `npm ci --ignore-scripts` in `integrations/omi-even/app` → fails with `EBADPLATFORM` for `@evenrealities/sim-darwin-arm64@0.8.0` on linux/x64.
- `npm ci --ignore-scripts` in `integrations/evenhub-g2` → same `EBADPLATFORM` failure.

Given the credential/user-data/capture surface, this should also get human maintainer security review before merge.

---

aryanorastar:
Not merging this.

Head is still 0565691 (Jul 28). The Jul 28 CHANGES_REQUESTED is unanswered, and the branch is now CONFLICTING on .gitignore. integrations/omi-even is not on main.

The three review blockers are still in the files:
- WS /app accepts with no bearer and no origin check, then serves memories/ask/capture. HTTP already has _authorized. The documented bind is 0.0.0.0:8788.
- server.py still logs the Even token at startup.
- @evenrealities/sim-darwin-arm64 is still a direct devDependency, so npm ci fails off Darwin.
- add-agent-contract.md still says the client timeout is "not measured" / ~10s is fine. The bridge is built around a measured ~5s cutoff.

The retrieval ladder and the Python suite are not the merge gate. I am not approving an unauthenticated LAN socket that holds a Firebase-backed session and a microphone. Do not expose this tip on a tunnel.

PR #10751, opened 2026-07-28
SCA-212: align dev public-build gateway contract

Git-on-my-level:
## Summary

- Route development public API build values to `https://api.omiapi.com`.
- Keep the gateway-only personas runtime contract explicit: retain only its gateway token and required public-service secrets, remove stale provider/LinkedIn host bindings, and reject empty gateway URLs before deploy.
- Give the app deployment the same direct-VPC network/subnet wiring; personas and admin retain their gateway URL wiring without min-instance flags.
- Allow `app` and `admin` to deploy an exact PR SHA to an explicitly selected development environment, with matching environment-scoped concurrency. This preserves automatic push behavior.

## Firebase decision

Development and prod intentionally use the same checked-in public Firebase client configuration (`based-hardware`) for persona authentication. This PR does not introduce Firebase secrets or change that auth tenancy; it only corrects the dev API bases.

## Stage 2 secrets

The personas contract requires Secret Manager bindings for `NEXT_PUBLIC_RAPIDAPI_KEY`, `NEXT_PUBLIC_RAPIDAPI_HOST`, `NEXT_PUBLIC_LINKEDIN_API_KEY`, and `OMI_LLM_GATEWAY_SERVICE_TOKEN`. Provider credentials (`ANTHROPIC_API_KEY`, `CLAUDE_API_KEY`, and `OPENROUTER_API_KEY`) are deliberately not required by the personas service.

## Verification

- `python3 .github/scripts/test_check_public_build_contract.py`
- `python3 .github/scripts/check_public_build_contract.py`
- `python3 .github/scripts/check-deployment-concurrency.py --self-test`
- `python3 .github/scripts/check-deployment-concurrency.py`
- `backend/.venv/bin/python -m pytest backend/tests/unit/test_llm_gateway_deploy_contract.py -q`
- `make preflight`

Closes SCA-212


Git-on-my-level:
Stage 3 is still blocked on exact-head development dispatch for `app` and `admin`.

- `gcp_app.yml` exposes an unused `branch` dispatch input, then selects `prod` for every ref except `refs/heads/development`. Dispatching PR 10751's head therefore uses the production environment/project variables.
- `gcp_admin.yml` has no environment input and skips the job unless the ref is exactly `main` or `development`, so its candidate cannot be dispatched against this PR head at all.

Please give both workflows the same explicit `workflow_dispatch.inputs.environment` choice and event-aware environment/concurrency expressions as personas (while preserving push behavior), and add contract tests that prove development dispatch on a non-development exact head resolves to the development environment rather than prod/skipped. This is required by SCA-211/SCA-214's exact-SHA development deployment acceptance.

PR #10758, opened 2026-07-28
fix: restore PTT voice tools + web-grounded escalations, notch polish

Git-on-my-level:
## Summary

Comprehensive fix for the PTT / voice-agent failure class observed live in Omi Beta 0.12.141, plus notch polish. Diagnosed from `/private/tmp/omi-beta.log` and the full git fix history of PTT.

**What was broken (from the running Beta's logs):**

1. `RealtimeHub[gemini]: kernel rejected tool ask_higher_model code=tool_not_allowed` on every PTT escalation. The kernel run-tool capability computes its allowlist from the *chat-adapter* projection of the tool manifest, but the realtime-voice surface advertises tools generated from the *surface* projection — so every voice-only tool (`ask_higher_model`, `point_click`) was structurally rejected and Gemini received a 112-byte error as the tool result. This is why voice tool calls fail.
2. Even when the escalation went through, its request body carries no client tools, and the gateway's agentic-turn heuristic (`!client_tools.is_empty()`) therefore never injected the server `web_search` tool — every voice escalation for current facts (weather, prices, scores; all visible in the log) reached Claude with zero web access. This is why PTT can't do web lookups.
3. (Historical, already fixed on main and confirmed shipped in 0.12.141: `tool_choice.name 'web_search'` 400s from the previous build's process were the old forced-caller bug, #10422 lineage.)

**Fixes:**

- `fix(agent)`: kernel run-tool capability registration now unions the run surface's manifest projection (`toolsForSurface("realtime_voice")`) with the adapter projection for `swift_realtime` runs — still filtered by execution role and spawn-time tool policy. Chat runs cannot inherit voice-only tools (asserted in test).
- `fix(desktop)`: the gateway accepts an explicit `omi_web_search: true` opt-in that injects the server `web_search` tool with no client tools present; the private/no-web prompt boundary still wins. The Swift escalation body opts in, the escalation prompt directs grounding in the lookup, and its timeout is raised to 60s for `pause_turn` continuations.

**Notch polish (as requested):**

- The legend now shows one **Open** row bound to the configured shortcut (matching Settings' "Open Omi Shortcut") instead of Ask + a hardcoded Summon row. The window-summon chord keeps working; it is just no longer listed.
- The Omi mark now pulses like a speaker while the assistant answers (two-harmonic breathing, bounded to the 21pt slot), and the listening waveform gains a center-weighted level-meter envelope with a travelling overtone. Reduce Motion suppresses both; the speaking state is announced to accessibility. Driven by the existing reducer-owned `isVoiceResponseGlowActive` projection — no new voice state.

**Follow-up commits (same PR):**

- `fix(desktop)`: **PTT stuck after a barge-in** (user-reported: PTT didn't stop after asking for the weather; chat appeared to show a duplicate). Root cause: `providerReplacementReady` (the barge-in session replacement) unconditionally rewrote the voice route from `.hubWarmWait` to `.hub`, stealing the captured-audio handoff that only `hubReady` can flush — so the turn had no armed deadline and hung ~2.5 minutes until the provider's idle teardown (log: turn `C6BFD805`, `dropped stale event ... deadline_fired`, terminal only at +150s). Both provider-ready admissions now converge on one `bindProviderReadyHubRoute` helper that never overwrites a manager-owned warm capture (`Failure-Class: FC-split-mutation-authority` — second competing writer for captured-input ownership). Three hermetic reducer tests reproduce the exact stale-drop and fail on the pre-fix reducer (8 failures), pass post-fix. The "duplicate message" was investigated and is NOT an INV-6 violation: kernel `conversation_turns` shows exactly one pair per continuity key; the appearance came from the `kernel_turn_tail` diagnostic returning both revisions of one turn plus two genuinely distinct weather asks.
- `feat(desktop)`: **notch animations now react to live audio.** The mic level was 5 Hz-throttled (waveform looked canned); `AudioLevelMonitor` now exposes un-throttled live mic + voice-playback levels polled per frame, `StreamingPCMPlayer` taps the main mixer for the actual audible output RMS, and the mark smooths both with an asymmetric attack/release envelope — the waveform follows the user's voice and the speaking pulse follows the assistant's speech (with a floor so quiet stretches still breathe).

**Internet-search findings (live-verified):**

- Typed main chat web search was never broken — verified live on the deployed dev backend (completed `web_search` tool call with current market data).
- PTT escalation search was broken **purely by deployment**: the `omi_web_search` opt-in exists only on this branch, and deployed gateways silently ignore the unknown field. Side-by-side proof with the identical escalation request: deployed dev backend → `web_search_requests: 0` plus a hallucinated tool-call and an "I don't have web access" answer; this branch's gateway → `web_search_requests: 1` with real source URLs. Three live PTT weather turns against the patched gateway completed with `web_search_exposed=true` and real current conditions.
- Real bug found while proving it: `should_record_web_search_fallback` still required client tools, so a voice escalation that asked for web and lost it (haiku route / kill switch) emitted no `chat_retrieval` fallback event — the exact regression the opt-in should surface was telemetry-blind. Fixed with the predicate mirroring `inject_web_search`; two regression tests; `cargo test --locked` 390 pass.
- **Rollout consequence:** dev bundles get the gateway fix automatically when this merges (`desktop_backend_auto_dev.yml` is merge-triggered; branch deploys are impossible by design). Omi Beta/stable users need a separate, explicitly-approved `desktop_backend_prod.yml` run after merge.

- INV-AGENT-* — run-tool capability allowlist widened for `swift_realtime` runs only; the guard test `agent/tests/run-tool-capability.test.ts` gains behavioral coverage that a swift_realtime run authorizes `ask_higher_model` end-to-end and a plain chat run still cannot.
- INV-VOICE-1 — the barge-in fix strengthens this invariant: the reducer remains the single lifecycle owner, and both provider-ready admissions now converge on one route-binding helper so no admission path can overwrite the manager-owned warm-capture claim. Guard: three new reducer tests in `VoiceTurnReducerTests`.
- INV-CHAT-1 — path-matched (gateway chat route, floating bar view); no transcript ownership or write-path behavior changes. The single shared `ChatProvider` timeline and kernel journal flow are untouched.

## Failure class (fixes)

Failure-Class: new

New class registered in this PR: an authority admitting tool invocations must compute its allowlist from the same manifest projection that advertised the tool to the invoking surface. Guard artifact: the behavioral capability test above.

The `omi_web_search` starvation fix declares no class of its own (single-instance heuristic gap; the release-boundary web-search probes already guard the deployed behavior).

## Verification

- `npx vitest run tests/run-tool-capability.test.ts` — 17/17 pass (new: swift_realtime surface-tool authorization + chat-run exclusion).
- `cargo test omi_web_search` — 2/2 pass (opt-in injects `web_search` with no client tools; explicit private/no-web turns stay tool-free). `cargo check` clean.
- `./scripts/agent-logic-harness.sh` — all four stages green (Swift lifecycle/state, agent runtime, pi-mono-extension, propagation self-check).
- `./scripts/agent-logic-harness.sh --cross-surface-smoke` — green (Swift, agent, Rust retrieval contract).
- `xcrun swift test --filter "NotchVoiceMorphMarkTests|OpenOmiShortcutSummonRoutingTests"` — 14/14 pass (new: speaker-pulse bounds/containment, envelope shape, amplitude ceiling, reduce-motion).
- `make preflight` — manifest checks pass with this PR body.
- Barge-in live check on the rebuilt bundle: a second PTT turn fired during the first turn's playback drain; both turns reached `terminal_success` with `stale_event_count=0` (previously the second turn wedged for 150s). Reducer regression tests cover the exact pre-fix wedge hermetically.
- Audio-reactivity tests: `VoiceLevelSmoother` attack/release/clamp/time-gap, speaking-level modulation floor/ceiling, `StreamingPCMPlayer.rmsLevel` on known buffers — 20/20 pass with the containment suite.
- QA bundle `omi-ptt-fix` built and launched via `OMI_APP_NAME=omi-ptt-fix ./run.sh --yolo`. **Live voice-path proof** — a natural authenticated PTT turn with real synthesized speech ("What's the weather in New York City right now?") through `ptt_test_turn`:
  ```
  22:59:28 RealtimeHub[gemini]: tool_call ask_higher_model({"query":"What's the weather in New York City right now?"})
  22:59:34 RealtimeHub: ask_higher_model ← claude-sonnet-4-6 OK in 6304ms (711 chars)
  22:59:34 RealtimeHub[gemini]: tool result ask_higher_model raw_bytes=1135 oversized=false
  22:59:57 RealtimeHub: reply received chars=423
  ```
  The identical query in Beta 0.12.141 logged `kernel rejected tool ask_higher_model code=tool_not_allowed` with a 112-byte error result. No rejection occurs on this build; the escalation completes and the reply is spoken.

Note: the `omi_web_search` gateway change takes effect in the deployed desktop backend via its normal Cloud Run release lane (whose candidate probes already require a real provider-reported web search); the client sends the opt-in immediately and older backends ignore the unknown field harmlessly.

## Latest UI iterations

The notch speaking/listening animation went through several passes after the initial audio-reactivity fix, each refining shape and responsiveness while staying on the same reducer-owned `isVoiceResponseGlowActive` projection (no new voice state added):

- `e237326c59` — circular sound-wave speaking ring (replacing the earlier pulse) plus a snappier input waveform response.
- `45d9de2a5f` — unidirectional wave propagation and a tail-of-phrase gate hangover so the ring doesn't cut off abruptly at the end of an utterance.
- `9773a8a43b` — equalizer-spike input wave and a brisker three-crest speaking ripple for a more percussive, less sluggish feel.
- `df2e003808` (current head) — final iteration: uniform fast-spiking ring expansion with a 12ms attack / 90ms release envelope, tuned to track live mic/playback RMS without overshoot.

**Verification of the latest state:**

- `xcrun swift test --filter NotchVoiceMorphMarkTests` — 16/16 pass on the current head (grew from 14/16 as the ring-expansion iterations added coverage).
- Live real-audio check: rebuilt the `com.omi.omi-ptt-fix` bundle and ran `ptt_test_turn` with real synthesized speech end-to-end — reached `terminal_success`, confirming the animation and escalation path both work together on the final build, not just in unit tests.


kodjima33 (approved):
Good diagnosis of the PTT tool-authority projection bug; holding the merge until Desktop Swift Build & Tests goes green.

PR #10763, opened 2026-07-28
fix(backend): surface denied hosted-MCP memory reads as errors (#10735)

### LINKED ISSUE #10735: Hosted MCP returns an empty 200 instead of an error when a memory read is denied

kodjima33:
Fixes #10735.

## Bug

A hosted-MCP memory read that the authorization check **denies** came back as a successful JSON-RPC result carrying `{"memories": []}`. The client reports "you have no memories" for an account whose memories are simply withheld — the failure mode most likely to make a user believe their data was deleted. Support cannot separate it from a genuinely empty account without a per-account Firestore trace, and it contradicts the memories screen, which lists the same user's memories fine.

## Root cause

`backend/routers/mcp_sse.py`, the two legacy-cohort memory reads (`get_memories` :866, `search_memories` :1094):

```python
if not mcp_legacy_read_authorized(memory_list_results):
    return {"memories": []}
```

Every reason that reaches this fail-closed branch is a server-side authorization state the caller cannot see or act on — `missing_mcp_default_memory_grant` (see #10734 for the grant-seeding gap), `malformed_rollout_state`, `rollout_read_failed`, `shadow_only`. Returning an empty list makes the denial indistinguishable from an empty account.

## Fix

Both sites now raise `ToolExecutionError(..., code=-32009)` carrying the denial reason, mirroring the app-key grant check a few lines above in the same handler. `handle_mcp_message` already converts that into a JSON-RPC error, so the client gets an explicit failure instead of a false empty.

Authorized paths are untouched: `USE_MEMORY` still returns memories, `USE_LEGACY_SAFE` and the un-enrolled cohort (`missing_rollout_state`, #9892) still fall through to the legacy read. Nothing that previously returned data returns less, and no data is exposed that was not exposed before.

The line-count ratchet baseline for `mcp_sse.py` is raised by the 15 lines of the shared helper, with its justification.

## What I tested

Drove `handle_mcp_message('tools/call')` — the exact seam the hosted SSE transport calls — with a denied decision, and read the envelope the MCP client receives.

Before:
```
get_memories    -> {"jsonrpc":"2.0","id":7,"result":{"content":[{"type":"text","text":"{\n  \"memories\": []\n}"}]}}
search_memories -> {"jsonrpc":"2.0","id":7,"result":{"content":[{"type":"text","text":"{\n  \"memories\": []\n}"}]}}
```

After:
```
get_memories    -> {"jsonrpc":"2.0","id":7,"error":{"code":-32009,"message":"Memory read is not authorized for this key (reason: missing_mcp_default_memory_grant)"}}
search_memories -> {"jsonrpc":"2.0","id":7,"error":{"code":-32009,"message":"Memory read is not authorized for this key (reason: missing_mcp_default_memory_grant)"}}
```

- New regression test `backend/tests/unit/test_mcp_sse_denied_memory_read_errors.py` (6 tests) drives the same client seam: 3 denial reasons on `get_memories`, one on `search_memories`, plus two that assert the authorized paths still serve memories. **4 of the 6 fail on the pre-fix code** (`git checkout routers/mcp_sse.py` → `4 failed, 2 passed`) and all 6 pass on the fix.
- `pytest tests/unit/test_mcp_memory_adapter.py tests/unit/test_mcp_data_endpoints.py tests/unit/test_mcp_sse_pagination_bounds.py tests/unit/test_mcp_auth_context_static.py tests/unit/test_mcp_action_item_writes.py tests/unit/test_mcp_conversations_pagination_clamp.py tests/unit/test_memory_api_egress_contract.py` → **118 passed**.
- `check_module_stub_pollution.py` / `scan_import_time_side_effects.py` → 0 violations. `make preflight` green.

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10763?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
@@ -3,7 +3,7 @@
     "backend/routers/apps.py": 2373,
     "backend/routers/chat.py": 1588,
     "backend/routers/developer.py": 2263,
-    "backend/routers/mcp_sse.py": 1858,
+    "backend/routers/mcp_sse.py": 1873,
     "backend/routers/sync.py": 2058,
     "backend/routers/users.py": 2076
   },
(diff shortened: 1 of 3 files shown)


kodjima33:
Closing as a duplicate — #10743 (mdmohsin7) already fixes #10735 and is a strict superset of this change: it covers the REST `routers/mcp.py` denial sites too, moves the guard into `utils/mcp_memories.py`, and registers FC-denial-rendered-as-empty-success. That PR is green and awaiting review. Opened by the hourly watchdog before it spotted the existing PR; no need to review this one.

PR #9597, opened 2026-07-13
Converge desktop agent, voice, and journal ownership

Git-on-my-level:
Closes #9515.

## What changed

This is the atomic convergence refactor from #9515. It gives each canonical model one mutation owner and removes the superseded authorities in the same PR:

- `VoiceTurnCoordinator`/`VoiceTurnReducer` exclusively own logical PTT/realtime lifecycle; capture, sockets, playback, and floating UI are drivers/projections.
- The Node kernel exclusively resolves route intent, owns session/run/attempt lifecycle, authorizes run-scoped tool capabilities, publishes versioned context snapshots, and pins immutable execution profiles.
- The kernel journal/outbox is the single durable main-chat-visible write path. Swift chat stores project journal events and backend reconciliation deduplicates by the stable canonical turn ID.
- Task chat takes amendment option (a): task-thread turns ride the journal with thread scope; legacy SQLite is a read-only, paginated compatibility import with owner, removal condition, and remove-by metadata.
- Effective-owner changes use a pre-visibility transition barrier, exact `(ownerID, generation)` authorization snapshots, correlated Node runtime revocation, commit leases for local mutations, and hard process teardown on account replacement.
- Foreground and direct-control child tools share one fail-closed run/attempt capability path. Request/client IDs remain correlation only.
- Superseded routing, authorization, persistence, session-identity, voice-lifecycle, provider-mutation, and dual-write paths were deleted rather than hidden behind flags.

The journal continues to sync through the existing backend message representation. Additive canonical-turn and revision fields remain backward-readable (a missing revision is revision 1), so a reverted client still sees ordinary history. Rollback is a revert of this complete PR, not a runtime dual-architecture switch.

## Root cause and durable guard

The system had centralized nouns but plural mutation verbs: Swift request lifetime could authorize kernel tools, physical voice drivers could terminalize logical turns, several stores could persist visible chat independently, and async work could recapture credentials after an owner transition. Recent fixes in these areas shared the same failure class: authority and identity were inferred at a call site instead of proven at the canonical boundary.

[...]

PR #9365, opened 2026-07-09
fix(desktop): make Conversations repository authoritative

Git-on-my-level:
Closes #8197

## Summary

- Add one account-scoped `ConversationRepository` as the authoritative desktop read/mutation boundary, with cache-first delivery, ID-based observation, and separate list/detail/transcript completeness and freshness.
- Admit server responses by version, generation, and revocation state so stale searches, details, deletes, locks, or previous-account requests cannot resurrect or overwrite newer state.
- Centralize durable title/star/folder mutations with coalescing, persisted pending intent, confirmation, retry, rollback, canonical-cache protection, and correct per-caller failure delivery.
- Preserve sub-millisecond Firestore revisions through SQLite, serialize projection merges in one transaction, and fail account-pool changes closed.
- Add lightweight versioned backend list/mutation contracts, real Firestore projections, minimal authorization reads, and an instrumented one-shot legacy endpoint fallback for rolling deploys.
- Separate capture/finalization persistence from conversation read authority and migrate desktop UI consumers to the repository.
- Add deterministic real-database, concurrency, rollback, account-isolation, and hermetic backend regression coverage.

## Architecture direction

This is the intended 80/20 move toward the platonic model: one authoritative repository per signed-in account; the backend owns version truth; views observe IDs instead of carrying divergent object snapshots; mutation intent is durable and reconciled centrally; and tests exercise deterministic state transitions without live services or timing sleeps.

The remaining complexity is concentrated at the boundary—rollout compatibility, persistence, and reconciliation—instead of being repeated across views.

## Product invariants affected

- `INV-AUTH-1` — account-scoped repository generations and cache guards preserve session isolation across sign-out/account changes.
- `INV-CHAT-1` — citation navigation now observes repository IDs while preserving the shared chat thread and citation-detail continuity.

## Verification

- Independent subagent review: **approved; no actionable findings remain after final re-review**
- Desktop Swift suites: **183 suites passed in isolated shards**
- `ConversationRepositoryTests`: **38 passed**
- Real SQLite `ConversationCacheStorageTests`: **7 passed**
- Targeted backend lock/projection suites: **69 passed**
- Backend hermetic E2E: **103 passed, 3 skipped**
- Backend runtime-env validator: **21 passed**
- Desktop strict e2e-flow coverage: **17/17 changed Swift files covered**
- Import-purity, async-blocker, module-stub-pollution, route-policy, formatting, typecheck (0 errors), OpenAPI, generated-client, and pre-push checks passed
[...]

PR #10772, opened 2026-07-28
feat(memory): establish canonical promotion pipeline

Git-on-my-level:
## What changed and why

Replace the remaining split memory lifecycle with one canonical pipeline: broad, bounded intake always lands in Short-term; canonical consolidation is the sole route into Long-term and atomically records its promotion receipt and graph assertion. Retrieval now collapses canonical lineage, while vector, keyword, compatibility, and graph projections converge from a durable outbox instead of acting as memory authority.

### High-level change map

- **Broad Short-term intake:** conversation extraction, explicit writes, observations, imports, APIs, plugins, and integrations now enter the same Short-term lifecycle. Capture is subject-aware, accepts useful short utterances, preserves source state when grounding is invalid, and retracts only genuinely empty extraction results.
- **One Long-term admission authority:** canonical consolidation gives every pending item exactly one route—promote, archive, review, or reject. Promotion is the only route into Long-term and atomically commits the promotion receipt, canonical item, ledger state, structured graph assertion, and derived-view outbox work. The old generic L2 promotion workers and helper paths are removed.
- **Cleaner, higher-context retrieval:** default access can use eligible Short-term and Long-term memory, but collapses aliases and superseded items by canonical lineage so one logical fact appears once. Vector hydration is fenced to the authoritative item revision/commit, restricted content stays provider-ineligible, and chat/developer/MCP compatibility adapters preserve the canonical policy.
- **Reliable derived projections:** vector, keyword, V3 compatibility, and graph cleanup are durable, retryable outbox effects rather than competing memory owners. Workers reload authoritative state, use leases and idempotent writes, reconcile after provider calls, and acknowledge only successful convergence.
- **Correct replacement and privacy closure:** conversation reprocessing replaces its complete source set atomically, restores independently supported facts when a merged source disappears, and keeps source-addressable lineage. Memory deletion, conversation cascades, exact 100-item batches, and account deletion now close over canonical lineage under durable deletion and processing fences.
- **Bounded cost and failure handling:** consolidation and required-processing work use stable cursors, bounded Firestore queries, capped batch sizes and LLM calls, retry leases/backoff, and terminal review/quarantine after repeated failure. Sparse legacy scans rotate durably so old rows cannot starve newer work.
- **Operational and contract updates:** the PR adds the canonical maintenance schedule/runtime contract, Firestore indexes and rules, malformed-document read-boundary coverage, deployment setting classification, architecture/runbook updates, the proposed `INV-MEM-4` lifecycle invariant, and a backend prevention artifact for `FC-split-mutation-authority`.

## Why these are improvements

- **Simpler ownership:** capture, admission, mutation, and projection responsibilities now have one named authority each. Removing parallel promotion paths makes contradictory tier/graph state structurally harder to create.
- **Better recall without simply returning more noise:** broad intake retains potentially useful observations in Short-term, while evidence/subject validation and the exact-one-route consolidation boundary decide what deserves durable Long-term status.
- **Better precision:** promotion requires structured grounding and a graph plan; retrieval rehydrates authoritative records and collapses lineage before returning results, rather than trusting stale provider rows or showing both an alias and its survivor.
- **Higher reliability and privacy:** canonical state and every derived effect are journaled together, external work is retried from authoritative state, stale generations are fenced, malformed Firestore documents use the shared sanitized boundary, and deletion wins over delayed projection work.
- **Controlled cost:** all new maintenance scans and LLM paths have explicit query, batch, attempt, and per-pass bounds. Failed or poisoned work is quarantined instead of being retried forever or starving the queue.
- **Evidence, not only structure:** behavioral tests execute production seams for capture, consolidation, atomic apply, replacement, deletion, retrieval, projection delivery, scheduling, and malformed data. Fault cases assert rollback, retry, redaction, and no-side-effect behavior.

The requested recall, precision, reliability, and cost percentages remain design targets. This PR establishes the mechanisms expected to improve those metrics but does not claim production benchmark results.

- INV-MEM-1
- INV-MEM-2
- INV-MEM-4 (proposed)

## How it was verified

- `BACKEND_PYTEST_WORKERS=1 bash test.sh` swept all 776 backend test files. It initially exposed nine integration-fixture regressions; after repairing those seams, `BACKEND_UNIT_TEST_FILE_LIST=.unit-failures.txt BACKEND_PYTEST_WORKERS=1 bash test.sh` passed every formerly failing file.
- Focused canonical capture, consolidation, apply/outbox, lineage/privacy, adapter, runtime-contract, and source-replacement suites passed, including the hermetic canonical pipeline path.
- `make preflight` passed the complete diff-scoped local manifest.
- The bounded pre-push gate passed all 60 selected backend test files with serial workers; the one prior parallel-only duration failure passed its required guarded rerun (`22 passed in 0.80s`).
- Pyright (`0 errors`), Black, Ruff, Python compilation, changed-workflow `actionlint`, Firestore index generation/coverage, runtime validators, and the app-client OpenAPI drift check passed.
- `npx tsc --noEmit --incremental false` passed for the admin web workspace.

No production deployment or live personal account data was used. A live-service/user-facing production exercise therefore remains intentionally unperformed.

## Tests

Core-path and error-path coverage was added or updated for broad Short-term capture, sole-route consolidation, atomic graph-backed promotion, required-processing retries, source replacement/reactivation, lineage collapse, projection retries/read fences, privacy deletion, account purge, bounded review queries, scheduler wiring, and compatibility adapters.

## Failure class (fixes)

Merged PRs #9365 and #9597 are the recorded instances this would have caught. `backend/tests/unit/test_canonical_maintenance_ordering.py` is intentionally a backend behavioral contract, not a new shared checker, because it must execute the production memory-maintenance owner and prove that no second promotion route can run when consolidation is blocked; the shared cross-component primitive remains the `FC-split-mutation-authority` registry contract.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10772?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Resolved Codex P2 thread r3665114173 — duplicate-supersede-target guard.

The batch-level guard added in `28db796a5c` rejects two decisions that supersede the same Long-term candidate before any apply phase runs, so the second decision can no longer reach `_validate_authoritative_targets` with an already-superseded (inactive) target after the first promotion commits. Verified locally: `test_batch_rejects_duplicate_supersede_across_decisions` and all 56 canonical_consolidation unit tests pass.

PR #10774, opened 2026-07-28
Track 3: feat(desktop): ask_user (elicitation) card

katipally:
Omi's agent can now stop mid-turn, put a question to you, and wait for the answer. The card takes over the chat composer, carries the whole set of questions at once, and hands the composer back when you are done.

## Demo

https://github.com/user-attachments/assets/aec1938b-e361-4cbb-a173-59761dee28f3

## What changed and why

**The contract, because two protocols needed one surface.** An ACP permission and an `ask_user` call normalize into one `ElicitationRequest` that one card renders and one resolver answers. `desktop_dispatches`, its `resolvedBy: "user"` column, the `grants` row, and the `waiting_approval` status were already built; what they never had was a producer a human answers.

**ACP permissions stopped answering themselves.** `resolveAcpPermission` selected `allow_always` without consulting anything and marked its audit event `automatic: true`, while `evaluateDesktopToolPolicy` returned `dispatch_required` for the same actions. Two contradicting policies in one file; this keeps the one that asks. A one-time grant on a read-shaped kind (`read`, `search`, `think`) still resolves without you. Anything that mutates, executes, fetches, or asks to be remembered goes to a person.

**The model had to be told to use it, not told to ask more.** MCP tool definitions carry `name`, `description` and `inputSchema` only, so trigger conditions moved into `description`. The kernel policy's only sentence about asking is a frequency restriction, so a channel rule went into `sharedSemanticGuidance`: when you do ask, ask with the tool. The frequency policy is untouched and a test asserts it survives.

**One call carries the whole decision.** `ask_user` takes `questions[]`, so a three-part decision is one interruption instead of three round trips. `allow_multiple` is a bright line rather than a judgment: default false, true only when the question says "any that apply".

**A waiting question had to survive four watchdogs.** The relay's 120s budget, the ACP idle watchdog, and the Swift stall guard each read "blocked on a person" as "hung agent"; separately, a bound MCP child was handed no `OMI_BRIDGE_PIPE` and failed every tool instantly. The three watchdogs now read one helper, `isUserInteractionTool`, which already owned "which tools ask a person".

**Untimed is not unbounded.** A question is released the moment its own run reaches a terminal state, and the card owns a visible two-minute idle budget that any interaction restarts. On expiry it sends what you answered and cancels only what you never reached.

**The rest came from using it.** The card was invisible for a while because Home uses `HomeAskBar`, not `ChatInputView`, and every automated check passed anyway; both composers now take a required, observed store so a new one cannot compile without handling a pending question. Calling `ask_user` used to raise a permission card asking whether it could ask. Answers came back empty because the card sent `optionIds` while the runtime read `optionId`. Pick-many rendered as pick-one because `allowsMultiple` was honoured on both ends and carried by neither. The schema rejected the option shapes its own normalizer was written to read. Each of those has its own commit and its own regression test.

## Commit map

48 commits, numbered oldest-first as `git log --reverse upstream/main..HEAD`. Grouped so a reviewer can take one slice rather than the whole thing; each commit carries its own verification evidence.

| Group | Commits | What it is |
|---|---|---|
| Contract and consent | 1, 2 | The shared `ElicitationRequest` and resolver; removing ACP auto-approval |
| Runtime to Swift | 3, 10 | The three protocol messages, the additive `elicitation` capability, owner-keyed answers, projection moved off the runtime actor |
| The card | 4, 5, 15, 17, 19, 21, 26, 36, 37, 40, 47 | Composer swap and Home; who is asking, without the duplicated path; option rows; choose-then-send with staging that survives navigation; Next until Send; mode-aware Cancel, Send naming what it cancels, scrolling lists, the recommendation shown but never staged; numbered options; the idle countdown; the typed answer as an option; the footer click |
| `ask_user` the tool | 7, 12, 14, 23, 24, 25, 27, 35, 44 | Registration; trigger conditions in `description`; no permission-to-ask; nested object items in the schema projector; `questions[]`; the channel rule; correcting the `realtimeHub` claim; `allow_multiple` and its bright line |
| Surviving the wait | 28, 29, 30, 31, 38 | The bridge pipe, the relay deadline, the ACP watchdog, the Swift stall guard, release on a terminal run |
| Reading what the model sent | 34, 39, 43, 48 | Labels printed as JSON, options buried in nested arrays, one recursive rule, schema tolerance |
| Boundary drift | 41, 45 | `optionIds` inbound, `allowsMultiple` outbound |
| Tests and QA | 8, 11, 13, 22, 42, 46 | The e2e flow and bridge actions, the pi-mono contract, the QA helper that minted "Always Allow", XCTUnwrap fixtures, cancel coverage, flow-lint registration |
| Bookkeeping | 6, 9, 16, 18, 20, 32, 33 | Changelog fragments and line-count baselines |

- INV-AGENT-*
- INV-AUTH-1
- INV-CHAT-1
- INV-UI-1

`INV-AGENT-*`: the model-context change is entirely in `agent/src/runtime/context-snapshot.ts`; no Swift prompt file is touched. `control-tools.test.ts`, a listed guard test, gained the `ask_user` branch it never had.

`INV-CHAT-1`: an answer is a reply to the agent, not a message the user sent, so it is not written to the transcript.

`INV-UI-1`: card colors come from `OmiColors` only.

## Tests

- `control-tools.test.ts`: the `ask_user` branch of `handleAgentControlToolCall`, which had no coverage at all, plus the tool driven through its real schema with object, JSON-encoded and nested-array options.
- `desktop-elicitation-resolver.test.ts`: both directions of the surface boundary round-trip through their projection function; `humanIsBeingAsked` cannot be stranded by a throw; a terminal run releases its own question and leaves another run's alone.
- `desktop-elicitation.test.ts`: normalization for both protocols, `allow_multiple` never reaching a permission, and `relayToolTimeoutMs` returning null for `ask_user` under every name the relay sees.
- `acp-permission-elicitation.test.ts`: 17 cases for the permission path, including every fail-closed branch and cancellation reaching a waiter exactly once.
- `run-attempt-lifecycle.test.ts`: a binding with a dead pipe path is handed the live one.
- `StallDetectorTests.swift`, `ElicitationStoreTests.swift`: progress accounting under an open card; owner refusal, free text refused on a permission, `allowsMultiple` refused where the protocol cannot carry it.
- `e2e/flows/chat-elicitation.yaml`: the composer swap, queueing, and both the answer and cancel paths from the automation bridge.

Mutation-checked: the wire-drift, schema-tolerance, flatten-cap and bridge-pipe fixes each fail their tests when reverted.

Not covered by a test: the footer click fix. SwiftUI hit regions are not reachable from XCTest, so it was verified by hand against a build where the same clicks did nothing.

## How it was verified

Named bundle `/Applications/omi-elicitation.app`, development backend, hermes adapter, no hint to use the tool in any prompt.

- One `ask_user` call carrying four questions; `tool no-progress guard fired` absent for the whole wait, where without the stall-guard change it appears at 90s and cancels the turn under the open card.
- Queue walked `0 → 1 → 2 → 1` with a staged answer surviving the switch; answered `4 → 3 → 2 → 1 → 0`; composer returns and the model resumes using the answers.
- Cancelling drains the queue and the model continues, `has_error` false.
- ACP permissions against real Claude Code: write → Allow → file created; read → no prompt; write → Deny → file absent.
- Pick-many read back from the dispatch rows: `allowsMultiple = 1` on the four questions that invited several answers and `0` on the two that did not. Rows from before that fix record `NULL`.

From `desktop/macos`:

- `agent`: `npx tsc --noEmit` clean; `npx vitest run` → **697 passed**, 58 files
- `pi-mono-extension`: `npm test` → **108 passed**
- `xcrun swift build -c debug --package-path Desktop` → build complete
- `xcrun swift test`: `StallDetectorTests` **21**, `ElicitationStoreTests` **21**, `ChatDiscoverabilityTests` **23**
- `./scripts/agent-logic-harness.sh` → 4/4
- `python3 scripts/desktop-flow-lint.py` → OK, 60 flows, 148 actions
- `scripts/pr-preflight --base upstream/main --pr-body-file <this file>` → **23 checks passed**
- full `git push --dry-run` pre-push gate → passed

Merged with `upstream/main` at `90cef82cd2`, no conflicts; upstream's 24 commits touch no file this branch touches.

Not verified: an ACP agent originating a *question* rather than a permission. `@zed-industries/claude-agent-acp@0.18.0` hard-disables `AskUserQuestion` (`dist/acp-agent.js:819`), so that path has hermetic coverage only.

## Failure class (fixes)

Failure-Class: none

`harden(desktop/agent): route ACP permission requests to a person` is classified `harden:` rather than `fix:` on purpose: it strengthens a consent boundary found by reading the policy module, not a defect reported from the field. If a real incident is known, it should be reclassified with a declaration.

There is a class worth registering that matches no existing one: a field one process sets and another honours must be carried by the message between them and required by its type, or both ends stay individually correct while the receiver silently defaults. This branch hit it twice in the same boundary, in opposite directions. It is declared `none` only because a definition file requires `evidence_prs` and the evidence is this PR, which has no number yet. The guard surface landed regardless: one exported projection function per direction, plus the field made required on `ElicitationPendingMessage`. Registering `FC-uncarried-boundary-field` is a follow-up.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (dismissed):
Thanks for the thoughtful work here. The direction is promising: replacing unattended ACP permission auto-selection with a visible user decision path is the right security posture, and the new store/resolver/card tests cover a lot of the important queue and answer-shaping behavior.

I do need to block this head for now because the required Desktop Swift checks are failing. In particular, `Desktop Swift Static & Test Contracts` reports `AgentRuntimeProcessTests.testRuntimeHandshakeRejectsStaleV2RuntimeWithoutRequiredCapability` failing after `elicitation` was added to the required runtime capabilities. Please update the Swift runtime test fixture/expectation so the stale-runtime rejection remains intentional and the suite is green again.

One maintainer note: this PR also changes AI-agent behavior, not just UI. `desktop/macos/agent/src/runtime/control-tool-manifest.ts` adds the `ask_user` tool and its behavioral guidance, and `desktop/macos/agent/src/runtime/context-snapshot.ts` changes shared semantic guidance so agents use `ask_user` rather than ending with a prose question when they do decide to ask. That seems directionally safe because the frequency restriction is explicitly preserved, but it affects how coding/review agents behave in Omi and should get human maintainer review before merge.

Labels added: `needs-tests`, `security-review`, `needs-maintainer-review`.

---

katipally:
@Git-on-my-level need human response .
@kodjima33 @undivisible 

I would like to know if the direction I am going is properly aligning with the vision of your team. For the next PR, I am thinking of working on making Omi generate artifacts similar to Claude artifacts. Please let me know what you think.

Git-on-my-level:
Thanks for the update here. I rechecked the current head (`9a1c119`) and the previous blocking feedback about the Desktop Swift runtime capability test looks resolved: GitHub now reports the required Desktop Swift / agent-runtime checks passing, and I also ran the focused desktop agent tests plus `npm run build` locally under a stripped environment. I dismissed the stale automation change-request review tied to the older failing head.

This remains a human-maintainer-review PR rather than an auto-approval. It changes a security-sensitive desktop agent permission path by replacing unattended ACP permission selection with a user-facing elicitation flow, and it also changes AI-agent behavior/instructions: `desktop/macos/agent/src/runtime/control-tool-manifest.ts` adds the `ask_user` tool/guidance and `desktop/macos/agent/src/runtime/context-snapshot.ts` tells agents to use that channel when they do decide to ask. Directionally, that seems safer and better aligned with explicit user consent, and the added resolver/store/card tests cover a lot of the important routing, cancellation, and answer-shaping cases. But because it can alter how coding/review agents behave in this repo and because the UI/runtime behavior is broad, David or another maintainer should still sign off before merge.

Keeping the existing `needs-maintainer-review`, `needs-tests`, and `security-review` labels for now so a human can verify the product/UX/security intent and decide when the added validation is sufficient.

---

aryanorastar:
Not merging this.

Head is still 9a1c119 (Jul 28). GitHub is CONFLICTING on ChatPage.swift, DashboardPage.swift, ChatInputView.swift, AgentRuntimeProcess.swift, and the agent elicitation files. ChatPage.swift is gone on main. Home is QueryShellHome; the one composer is QueryHeroBar (INV-NAV-1). This branch never touches QueryShellHome.swift. A conflict resolve that brings ChatPage back is not allowed.

The consent bug is still on main: resolveAcpPermission still selects allow_always and records automatic: true, while evaluateDesktopToolPolicy still returns dispatch_required for the same mutating tools. That half is still worth landing — on the current shell, not this July card.

Rebase without resurrecting ChatPage. The card has to take over the QueryShell composer the same way it had to take over HomeAskBar, or it will be invisible again and every composer test will still go green. I am not approving a DIRTY agent-permission and prompt change.

aryanorastar:
Keeping this open on `9a1c1192042e`.

ask_user elicitation 54-file keep with tests; DIRTY not shrink-only.

aryanorastar:
Keeping this open. Not approving `9a1c1192042e`. `desktop/macos/agent/tests/elicitation-store.test.ts` absent on origin/main. +5619.

Issue #10780, opened 2026-07-28
Action items filtered by conversation_id + a due-date range still 500: two more undeclared Firestore indexes


kodjima33:
`GET /v1/action-items` accepts `conversation_id` alongside `due_start_date`/`due_end_date`, and the chat `get_action_items` tool passes both through to `database.action_items.get_action_items`. `_apply_action_item_date_filters` orders that due range ascending, so those calls need composite indexes that are neither declared in `backend/database/firestore_index_registry.py` nor deployed:

- `conversation_id ASC, due_at ASC, __name__ ASC`
- `conversation_id ASC, completed ASC, due_at ASC, __name__ ASC`

## Evidence

Probed prod Firestore directly (read-only, non-existent uid so no user data is read):

```
conversation_id + due range       -> FailedPrecondition: 400 The query requires an index.
conversation_id + completed + due -> FailedPrecondition: 400 The query requires an index.
completed + created range         -> OK (that composite is deployed)
```

`gcloud firestore indexes composite list --project=based-hardware` shows six `action_items` composites; neither `conversation_id`+`due_at` pair is among them.

Same failure class as #10777 (`FC-undeclared-serving-query-index`), split out because it adds two more composites to a hot collection rather than declaring one that prod already had to hand-create. #10777's fix declares `(completed ASC, due_at ASC)` and adds the behavioral guard in `backend/tests/unit/test_firestore_query_contract.py`; extending that guard to the `conversation_id` chains is what should force these to be declared.

PR #10781, opened 2026-07-28
fix(backend): declare the (completed, due_at) action-items composite index

### LINKED ISSUE #10777: Action items with a due-date filter 500 in prod: missing (completed ASC, due_at ASC) Firestore index
## Symptom

Every due-date-filtered action-item read fails in prod with `google.api_core.exceptions.FailedPrecondition: 400 The query requires an index`.

- Chat's `get_action_items` tool returns `Error retrieving action items` — traceback: `utils/retrieval/tools/action_item_tools.py:256` → `database/action_items.py:_stream_action_items_bounded`.
- `GET /v1/action-items` with `due_start_date`/`due_end_date` takes the same path.
- Prod Cloud Run `backend`: 55 occurrences in the 7 days to 2026-07-28, 50 of them in the last 24h.

## Root cause

`_apply_action_item_date_filters` orders `due_at` ascending whenever a due range is requested, and `get_action_items` then adds the `completed ==` equality filter, so Firestore needs a composite index on `(completed ASC, due_at ASC)`. The only deployed `action_items` composite was `(completed ASC, due_at DESC)`, which serves the reverse ordering and not this query. Nothing declared the index in `backend/database/firestore_index_registry.py`, so `firestore.indexes.json` had no `action_items` entry at all and no gate caught it.

## Evidence

Reproduced directly against prod Firestore (query: `users/{uid}/action_items` where `completed == False`, `due_at` range, `order_by due_at ASC`) → `FailedPrecondition: 400 The query requires an index`, index hint `completed ASC, due_at ASC, __name__ ASC`.

Found by the hourly bug watchdog.

kodjima33:
## What

Declares the `action_items` composite index `(completed ASC, due_at ASC, __name__ ASC)` in `backend/database/firestore_index_registry.py`, regenerates `firestore.indexes.json` from it, and adds the regression test that would have caught its absence.

## Why

Every due-date-filtered action-item read was 500ing in prod (#10777): 55 `FailedPrecondition: 400 The query requires an index` in the 7 days to 2026-07-28, 50 of them in the last 24h. Both the chat `get_action_items` tool and `GET /v1/action-items?due_start_date=…` take the same path — `_apply_action_item_date_filters` orders a due range **ascending**, `get_action_items` adds the `completed ==` equality, and the only deployed composite was the descending twin `(completed ASC, due_at DESC)`, which does not serve that ordering.

Nothing declared any `action_items` index in the registry, so the generated manifest had no `action_items` entry at all and no gate could catch the gap.

## Failure class

`Failure-Class: new` — `FC-undeclared-serving-query-index`: a serving query's required composite index must be declared in the repository registry, not exist only as a hand-created live index. The guard artifact is the new behavioral test below, which derives the required signature from the production query builder instead of restating it.

## What I tested

- **Live reproduction against prod Firestore** (read-only, non-existent probe uid, so no user data touched) with the exact chain the issue reports — `due_at >=`/`<=`, `order_by due_at ASC`, `completed == False`:
  `400 The query requires an index. That index is currently building and cannot be used yet.` The index Firestore demands is exactly the one this PR declares; `gcloud firestore indexes composite list --project=based-hardware` shows it as `CREATING` (hand-created during the incident) and the other five `action_items` composites as `READY`.
- **New regression test drives the real read path.** `test_due_date_filtered_action_item_reads_have_a_declared_composite_index` monkeypatches `database.action_items.db` with a recorder, calls the production `get_action_items` for a due range (parametrized over `completed=None/False/True`), derives each executed equality+ordering chain's required index signature, and asserts the generated manifest declares it.
- **It fails without the fix.** Stashing only the registry + manifest change: 3 failed, with `('action_items', 'COLLECTION', (('completed','ASCENDING'), ('due_at','ASCENDING'), ('__name__','ASCENDING'))) not in declared` — the same signature Firestore's error hint names.
- `backend/tests/unit/test_firestore_query_contract.py` → **11 passed** (`.venv/bin/python -m pytest`, Python 3.11.15).
- `python3 backend/scripts/generate_firestore_indexes.py` → manifest matches the registry; `firestore_query_coverage.py --check-ratchet` and `check_firestore_model_read_boundary.py` → exit 0.
- `make preflight` with this diff → **23 checks passed** (including `firestore-index-manifest`, `firestore-query-coverage`, `failure-class-protocol`). `black --line-length 120 --skip-string-normalization` → unchanged.
- Touching the registry selects the **full backend unit suite** (774 files); ran it via `test.sh`. Everything passes except the local fast-unit CPU-time guard tripping under parallel load (CI uses the 1.0s ceiling) and `test_sync_v2.py::TestAsyncCoordinatorBehavioral::test_inline_renew_errors_cancel_before_capacity_admission_without_releasing_retry_material`, which **fails identically on `origin/main`** — pre-existing, unrelated to this change.

## Follow-up (not in this PR)

`GET /v1/action-items` also accepts `conversation_id` alongside a due range, and those combinations (`conversation_id ASC, due_at ASC`, plus the `completed` triple) are neither declared nor deployed — same failure class, but provisioning two more composites on a hot collection is a separate call. Filed as #10780 rather than widened here.

Fixes #10777

Failure-Class: new


undivisible (approved):
Looks good after resolving the main conflict.

The production fix for #10777 already landed via #10778 (and conversation-scoped indexes via #10786). This PR still adds real value: the behavioral query→manifest regression test and `FC-undeclared-serving-query-index`. Conflict resolution kept main’s index declarations (no duplicate registry entry) and retained those guardrails.

undivisible (commented):
Looks good after resolving the main conflict.

The production fix for #10777 already landed via #10778 (and conversation-scoped indexes via #10786). This PR still adds real value: the behavioral query→manifest regression test and `FC-undeclared-serving-query-index`. Conflict resolution kept main’s index declarations (no duplicate registry entry) and retained those guardrails.

undivisible (approved):
LGTM. Index already on main via #10778/#10786; this PR lands the regression test + failure class after conflict resolution.

PR #10786, opened 2026-07-28
fix(backend): declare the conversation-scoped action-item due-range Firestore indexes (#10780)

### LINKED ISSUE #10780: Action items filtered by conversation_id + a due-date range still 500: two more undeclared Firestore indexes

kodjima33:
## Bug

`GET /v1/action-items` and the chat `get_action_items` tool accept `conversation_id`
alongside `due_start_date`/`due_end_date`. Every such read 500'd with
`FailedPrecondition: 400 The query requires an index`. Issue #10780.

## Root cause

`_apply_action_item_date_filters` orders a due range ascending, and `get_action_items`
adds `conversation_id ==` (and, for the completed bucket, `completed ==`) on top. That
needs two composites — `(conversation_id ASC, due_at ASC)` and
`(completed ASC, conversation_id ASC, due_at ASC)` — neither declared in
`backend/database/firestore_index_registry.py` nor deployed. #10777/#10778 declared only
the `(completed ASC, due_at ASC)` chain, so the conversation-scoped reads kept failing.

## Fix

- Declare both composites in the index registry and regenerate `firestore.indexes.json`.
- Widen `test_action_items_due_range_index_contract.py` to every filter combination the
  endpoint accepts. The recorder now captures *every* query `get_action_items` builds (the
  completed bucket and the bounded legacy scan each start their own from
  `db.collection('users')`), and equality fields are compared as a set because Firestore
  orders the equality prefix itself.
- Both indexes are already created and READY in `based-hardware-dev` and `based-hardware`,
  so the deploy-time reconcile check-only gate stays green.

## Verified

- Prod Firestore probe (read-only, non-existent uid) **before** the indexes existed:
  `conversation_id + due range` and `conversation_id + completed + due range` both raised
  `FailedPrecondition`; `completed + due range` was OK (control).
- After creating the composites, re-probed `based-hardware-dev` once READY: both
  conversation_id chains return with no error.
- `backend/test.sh` over the index registry, manifest, reconcile and query-contract tests:
  46 pass.
- Removing the two registry entries fails the new parametrizations `[conversation]` and
  `[conversation+completed]`; restoring them passes.

Closes #10780

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10786?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/firestore_index_registry.py
+++ b/backend/database/firestore_index_registry.py
@@ -186,6 +186,18 @@ INDEX_ONLY_REQUIREMENTS = (
         (_asc('completed'), _asc('due_at'), _asc('__name__')),
     ),
     FirestoreIndexRequirement(
+        'action_items_conversation_due',
+        'action_items',
+        'COLLECTION',
+        (_asc('conversation_id'), _asc('due_at'), _asc('__name__')),
+    ),
+    FirestoreIndexRequirement(
+        'action_items_completed_conversation_due',
+        'action_items',
+        'COLLECTION',
+        (_asc('completed'), _asc('conversation_id'), _asc('due_at'), _asc('__name__')),
+    ),
+    FirestoreIndexRequirement(
         'candidate_integration_outbox_generation_status',
         'candidate_integration_outbox',
         'COLLECTION',
--- a/backend/tests/unit/test_action_items_due_range_index_contract.py
+++ b/backend/tests/unit/test_action_items_due_range_index_contract.py
@@ -1,17 +1,18 @@
-"""The due-range action-item read must be served by a declared Firestore index.
+"""Every due-range action-item read must be served by a declared Firestore index.
 
-``get_action_items`` filters ``completed`` for equality and orders ``due_at`` ascending
-whenever a due-date range is requested, so Firestore needs a composite index on
-(completed ASC, due_at ASC). Prod only had (completed ASC, due_at DESC), which serves the
-reverse ordering and not this one, so every due-range read -- the chat ``get_action_items``
-tool and GET /v1/action-items with due filters -- failed with
-``FailedPrecondition: 400 The query requires an index``.
+``get_action_items`` orders ``due_at`` ascending whenever a due-date range is requested and
+combines that order with whichever equality filters the caller asked for: ``completed``,
+``conversation_id``, or both. Each combination is a distinct Firestore composite. Prod only
+had (completed ASC, due_at DESC), which serves the reverse ordering and not this one, so every
+due-range read -- the chat ``get_action_items`` tool and GET /v1/action-items with due filters --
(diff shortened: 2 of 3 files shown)


kodjima33:
Watchdog verification before merge (2026-07-28):

- `bash test.sh` on this branch (file list runner) — `tests/unit/test_action_items_due_range_index_contract.py` 6/6 pass, plus `test_firestore_indexes.py`, `test_firestore_index_config.py`, `test_firestore_query_contract.py` all green.
- Mutation check: deleting the two new `FirestoreIndexRequirement` entries from the registry fails `test_due_range_composites_are_declared_in_the_firestore_index_manifest[conversation]` / `[conversation+completed]` **and** `test_generated_firestore_manifest_matches_the_checked_in_contract` — the guard is real, not vacuous.
- Prod (read-only, non-existent uid): both composites are now READY in `based-hardware`, and all three query shapes (`conversation_id`+due range, `completed`+`conversation_id`+due range, `completed`+due range) return OK instead of `FailedPrecondition: 400 The query requires an index`. This PR closes the declaration drift so a reconcile/redeploy can't drop the hand-created indexes.

PR #10823, opened 2026-07-29
test(desktop-windows): verify packaged Sentry dependencies

### LINKED ISSUE #10738: Bug: Omi Desktop for Windows crashes on launch after restart attempt with "Cannot find module 'ms'" error
## Summary
After closing Omi to check whether recording would resume, the Windows desktop app showed a startup error instead of reopening normally.

This happened after the earlier blank-screen capture problem reported in #10737.

## Related issue
- Related: #10737

## Problem
On launch, Omi shows a JavaScript main-process error and fails with:

`Error: Cannot find module 'ms'`

This appears during app startup, preventing normal use.

## Expected behavior
- Omi should restart normally after being closed
- Restarting the app should be a valid recovery step after capture issues
- The app should not fail on launch because of a missing module dependency

## Actual behavior
- Omi was closed to test whether recording would resume after the blank-screen issue
- On relaunch, the app showed a JavaScript error in the main process
- The visible error says `Cannot find module 'ms'`
- The stack trace points into the packaged Windows app resources

## Redacted stack trace detail
The screenshot shows local Windows installation paths. Redacting the personal username, the relevant path pattern is:

`C:\Users\<redacted>\AppData\Local\Programs\omi-windows\resources\app.asar\...`

## Impact
This turns a recording failure into a full app startup failure.

A user who tries to recover from broken capture by restarting the app can end up unable to reopen Omi at all.

## Suggested investigation
- Check why the packaged Windows app is missing the `ms` module at runtime
- Verify whether a partial update, corrupted install, or packaging issue can cause this dependency failure
- Check whether this can be triggered after restarting from a bad capture state
- Add a safer recovery path or clearer remediation message if a required module is missing
- Investigate whether the blank-screen capture issue and this startup crash share the same root cause

## Environment
- Omi Desktop
- Windows
- Observed on Monday, July 27, 2026

tianmind-studio:
## What changed and why

Adds a final-artifact guard for the Windows startup failure reported in #10738. The probe runs with the packaged Electron binary, reads the declared `main` entry from `app.asar/package.json`, verifies that entry exists, resolves the packaged `@sentry/electron/main` entry, and loads Sentry's `debug -> ms` dependency chain with Electron's ASAR-aware loader.

The original crash was produced by electron-builder 26.8.1 pruning `ms`; #10875 upgraded the builder and current `main` packages the transitive dependency correctly. This revision therefore removes the no-longer-needed direct `ms` dependency and keeps the durable guard that would have blocked the broken artifact.

Both unsigned and Azure-signed release paths run the probe before artifact publication. Windows CI now uses an explicit `build:win:unpack` command, while the platform-neutral `build:unpack` remains usable on macOS and Linux. Changes to the Windows release workflow also trigger Desktop Windows CI so the signed-path contract cannot drift untested.

none

## How it was verified

- Rebuilt `dist/win-unpacked` on Windows with electron-builder 26.15.3 after removing the direct `ms` dependency.
- Ran the probe against that real artifact; it resolved the packaged manifest main, `@sentry/electron/main`, `debug`, and `ms` successfully.
- `node node_modules/vitest/vitest.mjs run scripts/verify-packaged-main-runtime.test.mjs scripts/build-scripts-publish.test.mjs` - 14 passed.
- Node and renderer TypeScript checks passed under Node 22.23.1.
- `actionlint .github/workflows/desktop-windows-ci.yml .github/workflows/desktop_windows_release.yml` passed.
- Local PR preflight passed all 20 selected checks against current `main`.

## Tests

The runtime-probe tests cover the successful packaged chain, rejection of lookalike executables, the original missing-`ms` failure, absent/malformed/incomplete success output, and secret-environment filtering. The package-script tests prove the Windows unpacked build, unsigned installer, and signed release workflow all run the probe after electron-builder.

## Failure class (prevents recurrence)

Failure-Class: FC-runtime-image-boundary

## New guards

`desktop/windows/scripts/verify-packaged-main-runtime.mjs` is a final-artifact check rather than a source-only check: it requires the completed `win-unpacked` tree and uses the shipped Electron binary to traverse the packaged ASAR module graph.


Git-on-my-level:
Thanks for the focused Windows packaging fix — this is a good direction.

I reviewed the diff and ran the two focused desktop Windows tests in an isolated environment after installing dependencies with lifecycle scripts disabled:

- `corepack pnpm@10 exec vitest run scripts/verify-packaged-main-runtime.test.mjs scripts/build-scripts-publish.test.mjs` — 10 tests passed
- `git diff --check origin/main...HEAD` — clean

The direct `ms@2.1.3` runtime dependency matches the Sentry/debug failure mode, and the new packaged-main probe is a useful regression guard: it runs after both unsigned and Azure-signed Windows packaging, strips the child process environment down to a small allowlist, requires the expected `omi-windows.exe` artifact, and fails closed if the packaged runtime cannot resolve `ms`.

Because this touches dependency metadata and the Windows release workflow/signing path, I’m leaving this as a positive maintainer signal rather than a formal approval. A human maintainer should still review the release-path implications before merge, but I did not find a blocking code issue in the changes I inspected.

---

aryanorastar (commented):
Community review (COMMENT — Windows packaging):

Declaring `ms@2.1.3` as a direct Windows desktop runtime dep and probing the packaged Electron binary (`verify-packaged-main-runtime.mjs` resolving `ms` via `app.asar`) is exactly the right fix for #10738 (transitive prune → main-process crash). Wiring the probe into both unsigned `build:win` / `build:unpack` and the Azure-signed release path (plus FC-runtime-image-boundary) is strong fail-closed release hygiene.

Please keep explicit:
1. Signed release path cannot publish an installer if the probe fails (exit non-zero before publish).
2. Probe asserts the real packaged module graph (not just `node_modules` on the builder).
3. Lockfile/`pnpm-lock` stays in sync with the direct `ms` pin.

Not requesting changes — looks merge-shaped once required Windows checks stay green on tip.

formed2forge:
Checked this against current `main` (as of the v0.12.200 desktop changelog) while looking for stalled Windows fixes to help unstick:

- `ms` is still not a direct dependency in `desktop/windows/package.json` on `main`, so the underlying crash this fixes (#10738 — packaged app dies on launch with `Cannot find module 'ms'`) is still live and unaddressed elsewhere.
- The PR itself is now showing merge conflicts against current `main` (`mergeStateStatus: DIRTY`) — looks like it's just gone stale from the release cadence rather than anything wrong with the fix.

Happy to rebase this onto current `main` and re-verify the packaged-binary probe if that's useful — let me know, or feel free to push the rebase yourself if you'd rather keep driving it. Didn't want to force-push over your branch without asking first.

tianmind-studio:
Rebased this onto current main and narrowed it in light of #10875: the direct ms pin is gone, while the final-artifact probe now derives the packaged manifest main, resolves from @sentry/electron/main, and is wired into Windows-only package and release paths. I rebuilt win-unpacked without the direct dependency and the real probe passed; 14 focused tests, both typechecks, actionlint, and all 20 local preflight checks also pass.

Git-on-my-level:
Follow-up review after the rebase onto current main — this revision is materially stronger than the earlier round, and it resolves both the #10875 builder upgrade and the asks from the earlier community review.

What I verified on 718cd8b:

- desktop/windows/scripts/verify-packaged-main-runtime.mjs — the probe now derives everything from the packaged artifact itself: it reads main from app.asar/package.json, resolves that entry, then walks @sentry/electron/main -> debug -> ms via createRequire chains, exercising exactly the boundary that failed in #10738 without booting the Electron app. It runs the real omi-windows.exe under ELECTRON_RUN_AS_NODE, rejects symlinks/lookalike executables in findPackagedExecutable, and fails closed on non-zero exit, missing success marker, and malformed/incomplete details.
- Secret hygiene is right: PROBE_ENV_ALLOWLIST (PATH/TEMP/SYSTEMROOT and friends) keeps AZURE_CLIENT_SECRET and other signing secrets out of the child process, and verify-packaged-main-runtime.test.mjs asserts that directly.
- Dropping the direct ms pin is correct now that main carries electron-builder ~26.15 via #10875 — the diff vs main contains no dependency or lockfile changes at all, only the probe and its wiring. Since #10738 is still open on main, this guard has real protective value today.
- Release wiring: in desktop_windows_release.yml the signed path runs pnpm verify:packaged-main immediately after electron-builder, and the unsigned path inherits it via build:win, so a probe failure fails the job before artifact collection or publish. The three explicit asks from the earlier review — signed path cannot publish on probe failure, probe asserts the packaged module graph rather than the builder's node_modules, lockfile stays in sync — are all satisfied on this head.
- desktop/windows/scripts/build-scripts-publish.test.mjs extends the existing contract tests sensibly: build:win:unpack joins the --publish never and probe-ordering cases, and the signed-path test matches the literal 'Build signed installer (Azure Trusted Signing)' step (checked against the workflow text, so it is not vacuous).
- CI: the Windows build smoke job in desktop-windows-ci.yml now runs build:win:unpack, so the probe executed against a real packaged win-unpacked artifact on this head — and the check is green. Adding the release workflow to the CI path triggers also keeps the signed-path contract from drifting untested.

Minor nits, non-blocking:

- .github/failure-classes/FC-runtime-image-boundary.json now lists the desktop probe as a canonical prevention artifact, but scope_hints still only cover backend/**, plugins/**, and gcp_*.yml — consider adding desktop/windows/** (and evidence_prs could gain this PR) so future failures in this area map back to the entry.
- The signed-path contract test regex-parses workflow YAML as text, which is brittle to innocent step renames — acceptable given the file's existing string-assertion style, just worth knowing.

Label housekeeping: the current diff no longer touches dependencies or lockfiles, so I removed dependency-review; workflow-review stays because the CI and signed-release workflow changes still need maintainer sign-off.

Positive signal — not a formal approval. The remaining gate is the release-lane change itself (the probe now runs inside the Azure Trusted Signing build), which needs a human maintainer's sign-off before merge.

---

tianmind-studio:
Applied the scope-hint follow-up in 8e2cefc885: FC-runtime-image-boundary now includes desktop/windows/**, so future packaged-runtime changes surface this class during PR preparation. I left vidence_prs unchanged because the registry defines those as merged PR evidence; #10823 can be recorded after it lands. The branch is rebased onto current main, and all 20 selected local preflight checks pass.

Git-on-my-level:
Follow-up on 8e2cefc after the rebase onto current main — this resolves the scope-hint nit from the earlier automated review, and the remaining gates are unchanged.

What I verified on this head:

- `.github/failure-classes/FC-runtime-image-boundary.json` — `desktop/windows/scripts/verify-packaged-main-runtime.mjs` is now registered as a canonical prevention artifact and `desktop/windows/**` joined the scope hints, so future packaged-runtime failures in that tree map back to this entry. Leaving `evidence_prs` as merged-PR evidence (rather than adding this open PR) is the right read of the registry's semantics.
- The rebase onto v0.12.233 is clean for review purposes: the PR diff vs `main` still contains only the seven intended files, with no dependency or lockfile changes.
- `desktop/windows/scripts/build-scripts-publish.test.mjs` — the signed-path contract test's step-name regex still matches the current `desktop_windows_release.yml` after the rebase, and `pnpm verify:packaged-main` (line 313) runs after the electron-builder signing overrides and before the unsigned build, so a probe failure fails the job before artifact collection or publication.
- Checks on this head are green, including `Windows build smoke` — it runs `build:win:unpack`, so the probe executed against a real packaged `win-unpacked` artifact — plus `Typecheck · Lint · Test` and `Linux package · helper smoke`.

Label state stays as-is: `positive-signal` (good PR, not a formal approval), with `workflow-review` and `needs-maintainer-review` retained because the CI and signed-release workflow changes still need a maintainer's sign-off.

The human gate before merge is the same one as before: the packaged-runtime probe now runs inside the Azure Trusted Signing release build, and that release-lane change deserves a maintainer's direct sign-off rather than automation's.

---
*by AI on behalf of David. Before merge, a human maintainer should sign off on the release-lane change: the packaged-runtime probe running inside the Azure Trusted Signing build in `desktop_windows_release.yml`.*

kodjima33 (approved):
Useful regression guard + cleanup for the already-fixed (#10875) Windows Sentry/ms packaging crash. Doesn't clear the desktop bug-fix fast-lane bar (title isn't fix(windows), diff >200 lines) — approving, leaving merge to a maintainer.

aryanorastar (approved):
Verified on 8e2cefc88547.

`verify-packaged-main-runtime.mjs` loads the packaged main from `dist/win-unpacked` so a missing hoisted `ms` (Sentry/debug) fails CI before the installer hits the update feed. Tests pin the script and the workflow step order.

This touches `.github/workflows/**` (CODEOWNERS @Git-on-my-level). My approve is not a fake David CODEOWNERS review. I am not merging.

aryanorastar:
The packaged-runtime probe is a complementary guard, and `8e2cefc88547` also edits `desktop_windows_release.yml`. This stays open and unapproved.

tianmind-studio:
Friendly ping: this packaged Sentry runtime probe still has `positive-signal` / `needs-maintainer-review`, and Astar kept it open on 2026-09-22. When a human maintainer has a moment, could we get a sign-off / merge decision on the release-lane probe wiring? Happy to rebase or adjust anything needed.

PR #10825, opened 2026-07-29
fix(ci): make Windows preflight path-safe

tianmind-studio:
## What changed and why

Make the repository preflight executable from ordinary Windows PowerShell, including checkouts whose absolute path contains spaces and non-ASCII characters.

- Resolve Bash from the active Git for Windows installation instead of invoking the System32 WSL launcher.
- Make GNU Make, hooks, release guards, readiness checks, and dev-harness entrypoints share the same Bash/Python resolver.
- Reuse the repository Windows virtualenv (`Scripts/python.exe`) while preserving interpreter-specific `PYTHONPATH` separators.
- Normalize manifest `bash` and `python3` commands at the runner boundary, and decode Git/check output as UTF-8 instead of the Windows ANSI code page.
- Preserve Unicode Git paths and checkout roots through preflight selection, child environments, and single-flight fingerprints.
- Make Windows cancellation/liveness handling read-only and race-safe with isolated process groups, `CTRL_BREAK_EVENT`, and Win32 wait state.
- Run readiness behavior fixtures through Git Bash and the selected native Python.
- Keep Ruby/iOS and POSIX permission contracts on Linux/macOS, where their semantics are faithful, while retaining Linux CI coverage.
- Keep security-sensitive token mode checks fail-closed on Windows; only the unrelated candidate-provenance fixture mocks that boundary.
- Remove fixed test-output paths and nested Make banner leakage that made checks order- or host-dependent.

## Product invariants affected

none

## How it was verified

- After syncing `origin/main` and dropping overlap already merged through #10838, #10842, and #10843, the PR diff is down from 40 files to 34 files.
- The current PR head `ff80da22a` passed the official `make preflight`: **70 selected checks in 273.24s**.
- A larger pending Windows integration stack passed `make preflight` from `J:\codex-omi-work\路径 空格\omi-main-20260729`: **79 selected checks in 323.17s**.
- Dev-harness unit suite: 69 passed, 13 expected platform skips.
- Preflight contract suite: 36 passed, 2 expected platform skips.
- Manifest runner suite: 39 passed, 1 expected platform skip.
- Readiness behavior: 5 passed.
- Failure-class guard ratchet: 14 passed plus a real 90-day history scan.
- Desktop backend candidate probe: 14 passed.
- Desktop beta retry contract: 30 passed.
- Setup contract passed baseline provisioning, dependency-sync rerun, and linked-worktree fallback.
- Ruff, Bash syntax, and diff hygiene passed for touched modules.

[...]

Issue #10824, opened 2026-07-29
Bug: manifest preflight compares MSYS and Win32 temp paths literally


tianmind-studio:
## Summary

`make preflight` cannot complete on native Windows because `test_run_checks.py` assumes the path passed through Bash has the same string representation as Python's Win32 temporary path.

## Reproduction

1. Run the local manifest preflight on Windows.
2. Ensure the subprocess uses Git Bash (the bare `bash` executable otherwise resolves to `C:\Windows\System32\bash.exe`/WSL before PATH entries under Windows `CreateProcess`).
3. `RunnerBehaviorTests.test_release_process_guard_uses_locked_pyyaml_in_every_declared_lane` reaches its final assertion and fails:

```text
/tmp/tmpXXXX/.github/scripts/check-release-process-guards.py
!=
C:\Users\...\AppData\Local\Temp\tmpXXXX\.github\scripts\check-release-process-guards.py
```

The command itself succeeds; only the raw path-string comparison at `.github/scripts/test_run_checks.py:435` fails. This stops `check-manifest-contract`, the first always-selected preflight check, so later checks never run.

## Expected

The contract test should compare the same logical path across MSYS and Win32 representations (or capture a platform-neutral argument identity), and Windows subprocesses should resolve the intended Git Bash explicitly rather than relying on bare-name executable search.

## Evidence

Observed while validating #10823 on Windows 11 with Python 3.11.15 and Git Bash 5.2.15. The other 29 `test_run_checks.py` cases and all 13 remaining checks selected for that diff passed.

tianmind-studio:
#10825 now covers both Windows failure layers here: it compares MSYS/Win32 paths by filesystem identity and resolves manifest-leading `bash`/`python3` to Git for Windows Bash/current Python instead of the System32 WSL launcher or an absent executable. On current `main`, that head plus the related Windows fixture fixes completed the official `make preflight` with all 30 selected checks passing in 223.41s on native Windows. The PR is green but still awaiting maintainer/workflow review, so this issue should close when it merges.

aryanorastar:
VERIFY (2026-08-06): **FIXED_OPEN** — please close.

The failure described here (literal MSYS vs Win32 temp-path compare in `test_run_checks.py`, plus bare `bash` → System32/WSL) was fixed by [#10825](https://github.com/BasedHardware/omi/pull/10825) (`7dc9301ded`, merged 2026-07-30), with related Windows path helpers from the portable pre-push work.

[...]

Issue #10737, opened 2026-07-27
Bug: Omi Desktop for Windows stopped capturing screen and only recorded dark blank screens for hours without warning


hugoaap-code:
## Summary
Omi Desktop for Windows stopped saving actual screen content and only captured dark blank screens for about 3 hours.

I only noticed because Omi later sent a proactive/recommendation-style notification telling me to get back to work, even though the app had effectively been recording useless blank screens during that whole period.

## Problem
For some reason, screen capture silently failed and Omi kept running without warning the user.

Instead of capturing the actual desktop, it recorded only dark blank screenshots for hours.

## Expected behavior
- Omi should keep capturing real screen content while recording is active
- If screen capture fails and the app starts producing blank/dark frames, the app should detect that
- The app should warn the user immediately so they can restart Omi, restart screen capture, or take another action

## Actual behavior
- Omi kept running for around 3 hours
- Screen history for that period was only dark blank screens
- No warning or error was shown
- I only realized it after receiving an Omi notification during that failed capture period

## Impact
This causes silent data loss.

The user may assume Omi is working normally for hours, when in reality nothing useful is being captured. By the time it is noticed, the screen history is already lost.

## Suggested investigation
- Detect repeated blank/dark screenshots during active capture sessions
- Alert the user when capture appears to be failing for a sustained period
- Log whether the Windows desktop capture pipeline is returning empty/black frames
- Investigate whether the app can still generate recommendations/insights even when screen capture is broken
- Consider adding automatic recovery or a prompt to restart capture when too many invalid frames are detected

## Environment
- Omi Desktop
- Windows
- Observed on Monday, July 27, 2026

tianmind-studio:
I found a concrete post-#10549 blind spot: a desktop `MediaStreamTrack` can become `muted` while `readyState` remains `live`. The existing sampler treated that track as usable and could continue persisting blank frames. PR #10829 gates sampling on `live && !muted` and routes runtime mute, ended, and initially-muted states through the existing generation-fenced recovery path. This addresses the browser-declared unavailable-track state without adding a black-pixel heuristic; it does not assume every possible blank-frame source has the same cause.

undivisible:
Closing as completed by #10829 (merged in 99c838e44c).

The merged change treats `readyState === "live" && muted` as unavailable, prevents sampling in that state, and recovers through the generation-fenced restart path. It includes regressions for runtime mute and initially-muted streams.

[...]

PR #10829, opened 2026-07-29
fix(desktop-windows): recover muted Rewind capture tracks

tianmind-studio:
## What changed and why

Addresses #10737's recurrence after #10549.

#10549 made Rewind recover when a desktop-capture track reaches `readyState === "ended"`. A browser can also report that the source is unable to provide media by setting `MediaStreamTrack.muted` while leaving `readyState === "live"`. The existing `isLive()` check accepted that state, so the hidden capture video could keep producing and saving blank frames without taking the recovery path.

This change defines one usable-track predicate (`live && !muted`), listens for both `mute` and `ended`, and routes both states through the existing generation-fenced two-second restart. It also checks the initial track after `video.play()` and avoids duplicate recovery if a lifecycle event fired while `play()` was pending.

This deliberately does not add pixel-based black-frame heuristics, which would misclassify legitimate dark screens. It closes the remaining browser-declared unavailable-track state; a post-merge recurrence would need runtime health telemetry for a different source of blank pixels.

none

## How it was verified

- Before the implementation change, the new live-but-muted regression failed because `rewindSaveFrame` was called once after the `mute` event.
- `RewindCaptureHost.test.tsx`: 10/10 passed, including runtime mute and initially-muted stream recovery.
- All Rewind component tests: 20/20 passed.
- Full Windows desktop Vitest suite: 5,070 passed, 18 skipped across 532 test files.
- Node and renderer TypeScript checks passed.
- Focused ESLint and Prettier checks passed.
- Failure-class CLI tests: 17/17 passed.
- The Windows manifest preflight passed all 11 selected checks; the isolated verification worktree included #10825's pending path-portability fix because current `main` still contains #10824.

## Tests

- A live track becomes muted without ending: no frame is saved, the stream reopens, and the replacement resumes capture.
- `getUserMedia` resolves an already-muted stream: no sample is scheduled before the replacement stream is usable.
- Existing ended-track, mid-save generation, quality, and unmount coverage remains green.

## Failure class (fixes)

Failure-Class: new

Registers `FC-media-track-output-liveness`: a long-lived capture loop must treat every browser-declared media-output unavailable state as non-capturable and recover before persisting another frame. #10549 is the historical evidence PR, and `RewindCaptureHost.test.tsx` is the canonical prevention artifact.

The behavioral regression lives with `RewindCaptureHost` because that component is the sole owner of the Windows desktop `MediaStreamTrack`, sampling timer, and generation fence. There is no second media-track capture owner to justify a shared repository primitive; the single usable-track predicate is reused by event recovery, initial admission, and every sample.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-media-track-output-liveness.json
+++ b/.github/failure-classes/FC-media-track-output-liveness.json
@@ -1,0 +1,14 @@
+{
+  "schema_version": 1,
+  "id": "FC-media-track-output-liveness",
+  "violated_contract": "A long-lived capture loop must treat every browser-declared media-output unavailable state, including a muted live track and an ended track, as non-capturable and recover before persisting another frame.",
+  "canonical_prevention": "Centralize track usability as live and not muted, route ended, mute, and initially unusable tracks through one generation-fenced restart path, and behaviorally assert that no frame is saved until a replacement stream is usable.",
+  "canonical_prevention_artifact": [
+    "desktop/windows/src/renderer/src/components/rewind/RewindCaptureHost.test.tsx"
+  ],
+  "evidence_prs": [10549],
+  "scope_hints": [
+    "desktop/windows/src/renderer/src/components/rewind/**"
+  ],
+  "status": "open"
+}
--- a/desktop/windows/src/renderer/src/components/rewind/RewindCaptureHost.test.tsx
+++ b/desktop/windows/src/renderer/src/components/rewind/RewindCaptureHost.test.tsx
@@ -4,6 +4,8 @@
 // The host opens ONE persistent desktop stream; when that track dies (display
 // sleep, GPU/driver reset, session change) nothing noticed — the sampler kept
 // drawing a dead <video> and saving frames from it, and capture never recovered.
+// #10737 recurred after that fix because a source can become muted while its
+// track remains live; that state must use the same no-save + reopen path.
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 import { render, cleanup, act } from '@testing-library/react'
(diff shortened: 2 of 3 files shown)


Git-on-my-level (approved):
Thanks for this clean fix. The muted-but-live track state is a real gap — `MediaStreamTrack.muted` can be `true` while `readyState` stays `"live"`, and the old `isLive()` accepted that, persisting blank frames.

Your single-predicate approach (`live && !muted`) unified through one `onTrackUnavailable` handler is the right design. Two details stand out as well done:

- The post-`play()` usability re-check closes a subtle race where `mute` could fire during the pending `play()` promise.
- The `hasUsableVideoTrack()` guard inside `onTrackUnavailable` correctly prevents duplicate recovery when another usable track still exists.

Both new behavioral tests (runtime mute + initially-muted stream) directly assert the no-save-before-recovery contract. The failure-class registration follows the established schema.

Approving — small, well-scoped, low-risk fix with solid test coverage.

---

kodjima33 (approved):
Desktop bug fix for #10737's recurrence: isLive() accepted live-but-muted tracks so Rewind kept saving blank frames. Single usable-track predicate plus mute/ended listeners, with regressions for both runtime-mute and initially-muted streams. Green CI.

PR #10832, opened 2026-07-29
fix(desktop): keep headless pi-mono off user TUI extensions

undivisible:
## Summary
- Headless desktop pi-mono was auto-loading user ~/.pi TUI packages, which emit extension_ui_request and stall text chat until the 60s watchdog.
- Pass --no-extensions so only the explicit Omi -e extension loads; cancel blocking extension_ui_request so turns cannot hang.
- PTT/realtime path is unchanged.

Failure-Class: none

## Invariants
- INV-AGENT-1
- INV-CHAT-1

## Test plan
- [x] bun run test -- tests/pi-mono-adapter.test.ts
- [x] Sign in to desktop app and send a text chat; confirm it completes (no infinite spinner)
- [x] Confirm push-to-talk still works
- [x] Confirm /tmp/omi.log no longer floods with unknown extension_ui_request after a text send

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10832?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/agent/src/adapters/pi-mono.ts
+++ b/desktop/macos/agent/src/adapters/pi-mono.ts
@@ -492,19 +492,16 @@ export class PiMonoAdapter implements HarnessAdapter {
     const args = [
       "--mode",
       "rpc",
+      // Headless desktop RPC has no TUI host. Auto-discovered ~/.pi packages
+      // (ponytail, caveman, etc.) emit extension_ui_request and can stall the
+      // turn. Keep only the explicitly passed Omi extension (-e still works).
+      "--no-extensions",
       "-e",
       this.extensionPath,
       "--provider",
       "omi",
       "--model",
       "omi-sonnet",
-      // Auto-discover extensions and MCP servers from the user's machine
-      // to maximize pi-mono's capabilities (e.g. Playwright, filesystem tools).
-      // SECURITY NOTE: auto-discovered extensions run in the pi subprocess and
-      // can read process.env (including OMI_API_KEY). This is acceptable because:
-      // 1. OMI_API_KEY is a short-lived Firebase ID token (~1 hour expiry)
-      // 2. Extensions are user-installed — the trust boundary is the user's machine
-      // 3. ANTHROPIC_API_KEY is always scrubbed (never exposed to extensions)
     ];
     // Pi has no set_system_prompt RPC — system prompt must be baked at spawn
     // time via the --system-prompt CLI flag. To change it, restart the process.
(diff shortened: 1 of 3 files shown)


tianmind-studio (changes requested):
**Blocking: preserve headless-compatible user extensions.**

`--no-extensions` disables every auto-discovered user extension, not only packages that require a TUI. That reverses the explicit capability added in `04cab25d4` (`feat(desktop): enable pi-mono auto-discovery of extensions and MCPs`), where the flag was removed specifically to keep Playwright, filesystem, and other user automation available; the existing spawn-args test also documented that contract.

The new `extension_ui_request` handler already addresses the reported 60-second stall: it ignores fire-and-forget UI methods and returns `{ cancelled: true }` for blocking dialogs. Please keep that handler but drop `--no-extensions` and retain the previous spawn-args expectation. If retiring all user extension discovery is an intentional product decision, it needs to be called out as a capability removal with an explicit migration/setting and coverage for the lost headless-compatible integrations; the current PR presents this only as a hang fix.

Verified locally on this head with Node 22: `npm test -- tests/pi-mono-adapter.test.ts` (39/39 passed).

undivisible:
> **Blocking: preserve headless-compatible user extensions.**
> 
> `--no-extensions` disables every auto-discovered user extension, not only packages that require a TUI. That reverses the explicit capability added in `04cab25d4` (`feat(desktop): enable pi-mono auto-discovery of extensions and MCPs`), where the flag was removed specifically to keep Playwright, filesystem, and other user automation available; the existing spawn-args test also documented that contract.
> 
> The new `extension_ui_request` handler already addresses the reported 60-second stall: it ignores fire-and-forget UI methods and returns `{ cancelled: true }` for blocking dialogs. Please keep that handler but drop `--no-extensions` and retain the previous spawn-args expectation. If retiring all user extension discovery is an intentional product decision, it needs to be called out as a capability removal with an explicit migration/setting and coverage for the lost headless-compatible integrations; the current PR presents this only as a hang fix.
> 
> Verified locally on this head with Node 22: `npm test -- tests/pi-mono-adapter.test.ts` (39/39 passed).

hi tianmind im going to sleep but will review this tmr
like i use pi as one of my main agents and the extensions i have were frying omi i also have other things planned like replacing pi agent harness with something handrolled but for now this made text chat work for me and maybe other pi users again but maybe ill figure out something to have only certain extensions and all mcps enabled or sm but thanks for leaving this here. also pinging @Git-on-my-level maybe he has some other ideas

Git-on-my-level:
The hang fix on the current head (`ffea6b69`) is solid. The `extension_ui_request` handler is the right approach: fire-and-forget UI methods (notify/setStatus/setWidget/setTitle/set_editor_text) are dropped silently, and blocking dialogs (select/confirm/input/editor, plus anything unknown via the default case) get a `{ cancelled: true }` response so a turn cannot stall on an unavailable TUI. User extension auto-discovery is preserved, keeping the Playwright/filesystem/MCP capability from `04cab25d4` intact. The subprocess security controls (ANTHROPIC_API_KEY scrub, OMI_YOLO_MODE guard, Firebase-token-only auth) are untouched. The new regression test covers both the ignored and cancelled paths; Agent Runtime / Hygiene / Desktop Swift Build & Tests are green.

Two notes:
1. The PR body still reads "Pass `--no-extensions` so only the explicit Omi `-e` extension loads" — the current head does the opposite (user extensions are preserved). Please update the description to match the final approach so it does not mislead a final merge review.
2. The prior "drop `--no-extensions`, keep the handler" feedback is resolved on this head.

Once the description is reconciled, this looks good to land.

---

tianmind-studio (approved):
Re-reviewed current head ffea6b690: the blocker is resolved because spawn args retain user extension discovery while blocking extension_ui_request events echo their request id with cancelled: true, and the regression covers both paths. The PR description should still drop its stale --no-extensions sentence, but I found no remaining code blocker and Agent Runtime is green.

PR #10835, opened 2026-07-29
fix(desktop): restore onboarding permission and file discovery

Git-on-my-level:
## Summary

- Refresh macOS permission state before re-prompting and immediately reconcile an already granted system-audio permission.
- Restore the legacy Files-stage scan and aggregate local-file memory formation, with visible progress, retry, and safe continuation after a failure.
- Bind file scanning, profile-memory import, and knowledge-graph writes to the captured signed-in owner; cancellation stops the indexer before it persists a batch or starts profile import.
- Persist completed scanning at the next step to prevent re-import after relaunch, and let users request Full Disk Access when returning to a completed Files stage.

## Validation

- `xcrun swift test --package-path Desktop --filter OnboardingPermissionToolTests --filter OnboardingFlowTests --filter FileIndexerServiceTests --filter APIClientMemoryBulkSafetyTests` (45 tests)
- `./scripts/agent-logic-harness.sh --swift-only` (529 tests)
- `python3 scripts/check_desktop_test_quality.py`
- Pinned `swift-format` and SwiftLint on every changed Swift file.
- `make preflight` with this PR body.

## Manual verification limitation

The named bundle launched, but its Omi Dev auth seed was signed out and Terminal lacks Accessibility permission, so the live signed-in onboarding/Full Disk Access path could not be exercised on this machine. The Files-stage behavior is covered by the focused production-model tests above.

- INV-AUTH-1
- INV-CHAT-1

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-onboarding.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-onboarding.json
@@ -1,11 +1,11 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/Onboarding/OnboardingChatView.swift": 2180,
-    "desktop/macos/Desktop/Sources/Onboarding/OnboardingPagedIntroCoordinator.swift": 1750
+    "desktop/macos/Desktop/Sources/Onboarding/OnboardingPagedIntroCoordinator.swift": 1804
   },
   "raise_justifications": {
     "desktop/macos/Desktop/Sources/Onboarding/OnboardingChatView.swift": "Pinned swift-format normalizes line layout without changing this source's runtime behavior.",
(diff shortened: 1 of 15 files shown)


kodjima33 (approved):
Fix reads correct — permission state is refreshed before re-prompting, the Files-stage scan is restored with retry, and pre-granted FDA is held until scanning completes. Not merging: Desktop Swift Build & Tests and Desktop Swift Static & Test Contracts are failing on this head.

PR #10838, opened 2026-07-29
test(setup): make linked-worktree fixture Windows-safe

tianmind-studio:
## Summary

- Give the linked-worktree Python fixture a real shebang so Git for Windows treats the extensionless file as executable on NTFS.
- Compare the fixture's physical path so Git Bash's logical `/tmp` mount and the `BASH_SOURCE`-derived Windows path resolve to the same location.
- Let the existing `setup-pre-push-prerequisites` regression run through its linked-worktree fallback assertions on Windows instead of failing in fixture setup.
- Suppress nested Make directory banners so the strict stdout contract is stable when this script runs beneath `make preflight`.

## Verification

- Before this change, native Git for Windows failed with `Got: PYTHON=python3` before reaching the linked-worktree assertion.
- `PATH=/d/codex-omi-work/.tools/make-shim:$PATH env -u TMPDIR bash scripts/test-make-setup.sh` now passes all three sections.
- The same full script passes when launched as a recipe beneath a parent GNU Make process.
- `bash -n scripts/test-make-setup.sh`
- `git diff --check`

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10838?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/test-make-setup.sh
+++ b/scripts/test-make-setup.sh
@@ -136,7 +136,9 @@ print-resolved-python:
 	@printf 'PYTHON=%s\n' "$(PYTHON)"
 EOF
 git init -q --bare "$TMPDIR/fallback-bare.git"
-out="$(cd "$FB_ROOT" && env -u PYTHON GIT_DIR="$TMPDIR/fallback-bare.git" make print-resolved-python 2>/dev/null)"
+# This contract also runs beneath `make preflight`; suppress nested Make's
+# directory banner so stdout remains the resolver value under both entrypoints.
+out="$(cd "$FB_ROOT" && env -u PYTHON GIT_DIR="$TMPDIR/fallback-bare.git" make --no-print-directory print-resolved-python 2>/dev/null)"
 # Resolve the physical path because Git Bash exposes /tmp as a logical mount
 # while BASH_SOURCE resolves the same directory through its Windows path.
 expected="PYTHON=$(cd "$FB_ROOT" && pwd -P)/backend/.venv/bin/python"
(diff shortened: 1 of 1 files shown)


Git-on-my-level (commented):
Thanks for tightening this Windows fixture. I reviewed the shell diff statically because it touches setup/test workflow code.

This looks like a good, narrow fix: giving the extensionless Python fixture a shebang matches Git-for-Windows executable detection, `pwd -P` is a reasonable way to normalize Git Bash logical `/tmp` paths against the `BASH_SOURCE`-derived physical path, and `make --no-print-directory` keeps the stdout assertion stable when the test is invoked from a parent make recipe. `bash -n scripts/test-make-setup.sh` on the PR head also passes.

I’m not formally approving because this touches workflow/setup test machinery and should still get maintainer review before merge, but I don’t see a blocker in the current patch.

---

kodjima33 (approved):
Fixture-setup defect on Git for Windows (extensionless file needs a shebang; logical /tmp vs BASH_SOURCE path mismatch). Lets an existing regression actually run instead of erroring in setup. 15 lines, green.

PR #10841, opened 2026-07-29
fix(ci): make backend shell contracts Windows-safe

tianmind-studio:
## What changed and why

Resolve shell subprocesses through the active Git for Windows installation instead of invoking a bare `bash`, which Windows process lookup can resolve to the System32 WSL launcher. A shared `testing.shell` helper now discovers Git Bash from `git --exec-path`, converts native paths with `cygpath`, and keeps command arguments separate from shell source.

Migrate every direct Bash subprocess in the backend test tree to that helper. Tests that inject fake tools now prepend their POSIX path inside Git Bash, and shell-consumed temporary scripts and path lists use LF explicitly. This covers workflow scope rendering, VM startup and reaper contracts, backend test runners, changed-files, gateway deploy probes, and the LC3 replay harness.

Make `backend/scripts/needs-typecheck.sh` consume inherited stdin directly instead of redirecting `/dev/stdin`, which Git for Windows does not provide in this environment. This preserves file-list behavior while preventing the Windows pre-push hook from silently skipping backend type checking.

none

## How it was verified

- Ordinary PowerShell: `backend/.venv/Scripts/python.exe -m pytest` across the 9 affected test modules (`41 passed`)
- Python AST audit of `backend/tests/**/*.py`: zero subprocess commands whose executable is a literal `bash`, `/bin/bash`, or `/usr/bin/bash`
- `black --check` across the 10 affected Python files
- `ruff 0.4.4 check` across the 10 affected Python files
- Targeted Pyright with the backend venv (`0 errors`; existing warnings remain in older untyped tests)
- `C:\Program Files\Git\bin\bash.exe -n backend/scripts/needs-typecheck.sh`
- `git diff --check`
- The full local PR preflight still stops in `check-manifest-contract` at the existing main-branch Windows Bash/Python command-resolution boundary; #10825 fixes that independent release-guard test issue and is green.

## Tests

The focused suites were first run from ordinary PowerShell before the fixes and reproduced the WSL launcher failures, UTF-16/GBK decode warnings, fake-PATH misses, and `/dev/stdin` failure. The same combined suite now passes through Git Bash selected from the active Git installation. `test_windows_bash_resolution_uses_the_active_git_installation` also covers discovery without relying on the host installation layout.

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/backend/scripts/needs-typecheck.sh
+++ b/backend/scripts/needs-typecheck.sh
@@ -15,8 +15,8 @@ input="${1:--}"
 if [ "$input" != "-" ] && [ ! -f "$input" ]; then
   usage
 fi
-if [ "$input" = "-" ]; then
-  input=/dev/stdin
+if [ "$input" != "-" ]; then
+  exec < "$input"
 fi
 
 while IFS= read -r path || [ -n "$path" ]; do
@@ -25,6 +25,6 @@ while IFS= read -r path || [ -n "$path" ]; do
       exit 0
       ;;
   esac
-done < "$input"
+done
 
 exit 1
--- a/backend/testing/shell.py
+++ b/backend/testing/shell.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+import os
+from pathlib import Path
+import subprocess
+from typing import TypeAlias
+
+ShellArgument: TypeAlias = str | os.PathLike[str]
+GIT_LAYOUT_PARENT_LIMIT = 3
+
+
+def find_git_bash(git_exec_path: Path) -> Path:
+    """Find bash.exe in the Git for Windows installation that owns git."""
+    for parent in tuple(git_exec_path.parents)[:GIT_LAYOUT_PARENT_LIMIT]:
+        for relative_path in ('bin/bash.exe', 'usr/bin/bash.exe'):
+            candidate = parent / relative_path
+            if candidate.is_file():
+                return candidate
+    raise FileNotFoundError(f'Git Bash was not found above {git_exec_path}')
+
+
+def bash_executable(
+    *,
+    cwd: Path | None = None,
+    platform_name: str | None = None,
+    git_exec_path: Path | None = None,
+) -> str:
+    if (platform_name or os.name) != 'nt':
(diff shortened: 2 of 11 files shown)


Git-on-my-level:
Thanks for tightening the Windows shell contract here. I reviewed the diff and ran the affected backend shell-contract/unit tests locally in a clean env with no repo secrets:

```
cd backend
env -i HOME="$HOME" PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" PYTHONPATH="$PWD" .venv/bin/python -m pytest \
  tests/test_replay_harness_lc3_frame_timing.py \
  tests/unit/test_agent_vm_reaper.py \
  tests/unit/test_agent_vm_startup.py \
  tests/unit/test_auto_dev_backend_scope.py \
  tests/unit/test_backend_test_runner.py \
  tests/unit/test_backend_typecheck_predicate.py \
  tests/unit/test_backend_unit_ci_runner.py \
  tests/unit/test_changed_files_script.py \
  tests/unit/test_llm_gateway_deploy_contract.py
```

Result: `41 passed`. I also checked `python -m py_compile` on the touched Python files, `bash -n backend/scripts/needs-typecheck.sh`, and `git diff --check`; all passed.

The implementation looks directionally good: resolving Bash through Git for Windows avoids accidentally invoking the WSL launcher, `bash_path()` keeps Windows paths POSIX-normalized for Git Bash, and the `/dev/stdin` removal preserves inherited-stdin behavior while avoiding an unavailable Git-for-Windows path.

I’m not formally approving because this touches CI/shell-contract behavior and should still get maintainer review on the workflow impact, but I don’t see a blocker from this pass.

---

kodjima33 (approved):
Concrete defect: bare 'bash' resolves to the System32 WSL launcher on Windows, so backend shell contract tests failed with UTF-16/GBK garbage and needs-typecheck.sh silently skipped type checking via the missing /dev/stdin. Shared testing.shell helper is the right seam; 41 tests pass and an AST audit proves no literal bash invocations remain. Green CI.

PR #10844, opened 2026-07-29
fix(ci): decode failure-class history as UTF-8

tianmind-studio:
## What changed and why

Decode `git log` output as UTF-8 in the failure-class guard-artifact ratchet instead of relying on the host locale. Git emits repository history as UTF-8, while native Windows Python used GBK here; a non-ASCII first-parent commit caused the subprocess reader thread to fail and left `stdout` as `None`, followed by an `AttributeError` in the ratchet.

Make the checker's two disposable-repository tests use Python `TemporaryDirectory` cleanup instead of invoking the POSIX-only `rm` executable. This lets the complete checker module run under native Windows.

none

## How it was verified

- Before the fix, the real repository scan failed with `UnicodeDecodeError: 'gbk' codec can't decode byte 0x94`, then `AttributeError: 'NoneType' object has no attribute 'split'`.
- After the fix, the real 90-day first-parent scan completed successfully with 23 declared classes.
- `.github/scripts/test_check_failure_class_guard_ratchet.py`: 14 tests passed on native Windows; before the cleanup fix, 12 passed and 2 errored because `rm` was unavailable.
- Ruff 0.4.4 passed.
- Pyright 1.1.403 passed with 0 errors and 0 warnings.
- `py_compile` and `git diff --check` passed.

## Tests

The new regression supplies non-ASCII history text and verifies the production Git subprocess explicitly requests UTF-8. Existing end-to-end repositories continue to cover thresholds, artifacts, grandfathering, shallow history, short history, PR-body declarations, and the shipped registry.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10844?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/check_failure_class_guard_ratchet.py
+++ b/.github/scripts/check_failure_class_guard_ratchet.py
@@ -65,7 +65,7 @@ def run_git(root: Path, *args: str) -> str:
         check=False,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
-        text=True,
+        encoding="utf-8",
     )
     if result.returncode:
         detail = result.stderr.strip() or result.stdout.strip()
--- a/.github/scripts/test_check_failure_class_guard_ratchet.py
+++ b/.github/scripts/test_check_failure_class_guard_ratchet.py
@@ -8,6 +8,7 @@ allowlist behavior execute end to end. No source strings are asserted.
 
 from __future__ import annotations
 
+import importlib.util
 import json
 import os
 import subprocess
@@ -15,6 +16,8 @@ import sys
 import tempfile
 import unittest
 from pathlib import Path
+from types import ModuleType
+from unittest.mock import patch
 
 
 REPOSITORY_ROOT = Path(__file__).resolve().parents[2]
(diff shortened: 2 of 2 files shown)


tianmind-studio:
Integration follow-up: this head was stacked with the current Windows preflight fixes and exercised through the official `make preflight`. Both `failure-class-guard-artifact-ratchet` and its full 14-test fixture check passed, and the entire 30-check manifest run passed in 223.41s on native Windows.

Git-on-my-level (commented):
Thanks for the focused CI portability fix. I reviewed the diff and ran the guard-ratchet test module locally on the PR head (`fab8999e0`): 14 tests passed.

This looks like a good candidate from an implementation standpoint: `run_git()` now decodes Git output as UTF-8 instead of relying on the host locale, the regression covers non-ASCII history text, and the temporary repository cleanup is made portable by replacing the POSIX-only `rm` calls with `TemporaryDirectory` cleanup.

I’m not formally approving because this touches CI/guardrail scripts, so it should still get human maintainer review before merge, but I don’t see blocking changes needed from this automation review.

---

kodjima33 (approved):
Real crash with a named root cause: git log decoded via host locale (GBK on Windows) returned None stdout then AttributeError. Explicit UTF-8 decode plus a non-ASCII-history regression; also drops a POSIX-only rm from the tests. Green.

PR #10845, opened 2026-07-29
test(ci): keep candidate probe fixtures Windows-safe

tianmind-studio:
## What changed and why

Keep the desktop-backend candidate-probe fixtures faithful to the production security boundary on native Windows.

The token helper intentionally requires a regular POSIX mode-`0600` file. Windows cannot represent or prove that mode through `st_mode`, so the permission test now asserts the existing fail-closed result there instead of expecting `chmod(0o600)` to create POSIX semantics.

The full CLI provenance test now mocks the token-validation boundary and asserts that it receives the requested path. That test remains focused on forwarding candidate revision, image, tag, and workflow provenance instead of depending on host filesystem permission semantics.

none

## How it was verified

- Before the fix on native Windows: 12 tests passed, one failed, and one errored.
- After the fix: all 14 candidate-probe tests passed.
- Ruff 0.4.4 passed.
- Pyright 1.1.403 passed with 0 errors and 0 warnings.
- `py_compile` and `git diff --check` passed.

## Tests

The mode test still exercises successful mode-`0600` reads and rejection of mode `0644` on POSIX. On Windows it verifies that the production reader fails closed. The CLI test independently verifies the token path and all candidate provenance forwarded to the probe.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10845?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/test_desktop_backend_candidate_probe.py
+++ b/.github/scripts/test_desktop_backend_candidate_probe.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import importlib.util
 import json
+import os
 import sys
 import tempfile
 import unittest
@@ -183,6 +184,10 @@ class CandidateProbeTests(unittest.TestCase):
             path = Path(directory) / "token"
             path.write_text("abc.def.ghi", encoding="utf-8")
             path.chmod(0o600)
+            if os.name == "nt":
+                with self.assertRaisesRegex(PROBE.ProbeError, "mode-0600"):
+                    PROBE._valid_token(path)
+                return
             self.assertEqual(PROBE._valid_token(path), "abc.def.ghi")
             path.chmod(0o644)
             with self.assertRaisesRegex(PROBE.ProbeError, "mode-0600"):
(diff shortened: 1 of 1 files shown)


tianmind-studio:
Integration follow-up: this was the final blocker in the stacked native-Windows run. After applying this head, desktop-backend-candidate-probe-fixtures passed all 14 cases and the official make preflight completed all 30 selected checks successfully in 223.41s.

Git-on-my-level:
Thanks for the focused Windows-safe test fix. I reviewed the single `.github/scripts/test_desktop_backend_candidate_probe.py` change and also ran the candidate-probe unit test file from this head locally on Linux: all 14 tests passed.

This looks like the right direction to me: the token-permission assertion stays fail-closed on Windows, POSIX still exercises the `0600` success/reject path, and the full CLI provenance test now mocks the token-validation boundary while still asserting the token path is passed into it.

Because this touches CI/probe test infrastructure under `.github/scripts`, I’m not formally approving from automation. A human maintainer should still do the final workflow-sensitive review before merge, but I don’t see a blocker in this diff.

---

kodjima33 (approved):
Fixes candidate-probe fixtures that assumed POSIX mode-0600 semantics on Windows; the production reader's fail-closed behaviour is now asserted there instead of faking chmod. 14 tests pass, green CI.

Issue #10389, opened 2026-07-23
fix: create release-gate-failure label (break-glass audit job fails without it)


Git-on-my-level:
## Symptom

`record_break_glass` / the Release Eligibility break-glass audit job cannot create its intended GitHub issue when the label `release-gate-failure` is missing. That happened for the prod Cloud Run bypass recorded in #10301: the audit job tried to file the issue with `release-gate-failure`, the label did not exist, and the automated create failed (the audit record was preserved manually as #10301).

## Expected

- Label `release-gate-failure` exists in the BasedHardware/omi repo, **or**
- The audit job creates the issue without requiring a missing label / is resilient when the label is absent.

## Acceptance

- [ ] Label `release-gate-failure` is present in the repo.
- [ ] Next break-glass path can file its audit issue successfully.
- [ ] Optional: harden the audit job so a missing label does not fail the issue-create hard.

## Evidence

Closed #10301 (Release Eligibility bypass of `e262b2323b`, prod Cloud Run) — body documents the failed automated issue create due to the missing label.

PR #10852, opened 2026-07-29
fix(ci): repair and wire the release-process guard self-suite

### LINKED ISSUE #10351: Release-process guard self-tests: 9 failures on main and not wired into any CI lane

kodjima33:
## What

Two commits:

1. **Decouple the historical-parent probes from today's `codemagic.yaml`.** They now materialize the contract tree at the parent's own revision (`git show <rev>:codemagic.yaml` + its contract fixture) and assert the *current* global document lock separately against the current tree.
2. **Register the suite in `.github/checks-manifest.yaml`** (`desktop-release-process-guard-selftests`, both lanes) behind the same locked-PyYAML wrapper pattern the guard itself uses.

## Why

`.github/scripts/test_check_release_process_guards.py` had **9 of 90 cases failing on plain `main`** (#10351), and the suite ran in **no CI lane** — not in the manifest, not in any workflow — so the failures accumulated unnoticed.

Root cause of the 9: those probes assert *"the reviewed parent guard accepted this bypass, and the current global lock rejects it"* — a claim about the past — but fed the parent guard **today's** `codemagic.yaml`. That parent hardcodes `len(canonical_scripts) != 21`; the canonical workflow has since grown to **22** steps, so every probe reported `canonical workflow must retain exactly 21 approved script steps`. `_restore_parent_preview_credential_shape` was an earlier band-aid for exactly this coupling (a credential-group drift); it is deleted, not extended.

The guard itself (`check-release-process-guards.py`) was never wrong — it passes in `repo-checks.yml`. Only its self-suite had rotted.

## Tested

- **Reproduced first**, clean checkout of `main` (`df314540b0`): `python3 -m pytest .github/scripts/test_check_release_process_guards.py -q` → `9 failed, 81 passed`.
- **After the fix**: `90 passed in 36.11s`.
- **Via the new registered command**: `bash scripts/run-release-process-guard-tests.sh` → `90 passed`.
- **Confirmed the parent guard is green against its own era's files** (`ROOT` = tree at `4e391ee726`, unmutated) → `[]`, and that each of the 9 mutations still produces the expected parent verdict there (4 accept, anchor-spelling rejects, etc.).
- **Drift-immunity check**: temporarily added a 23rd script step to `codemagic.yaml` and re-ran the suite. The 9 repaired probes stayed green; only the 5 tests that intentionally pin the contract fixture to the current document went red — i.e. the coupling that caused this issue is gone, and the deliberate approval gate still bites. `codemagic.yaml` was restored (`git status` clean); it is not touched by this PR.
- **Manifest contract**: `python3 .github/scripts/test_run_checks.py` → `Ran 30 tests … OK`.
- **Selection wiring**: `resolve_checks` selects `desktop-release-process-guard-selftests` in **both** `local` and `ci` for `codemagic.yaml`, `check-release-process-guards.py`, `test_check_release_process_guards.py`, and the contract fixture. This PR's own diff triggers it, so CI here runs the suite it just registered.

## Notes

- `scripts/run-release-process-guard-tests.sh` mirrors the existing `scripts/run-release-process-guards.sh` (locked PyYAML from `backend/.venv`, self-healing via `sync-python-deps.sh`). `pytest` exits 5 on an empty collection, so a future emptied suite fails instead of passing silently.
- No workflow file is touched — CI reaches the new check through the existing manifest lane.
- Local `make preflight` ran 70/72 selected checks green. The two that fail — `dev-harness-unit-tests` and `setup-pre-push-prerequisites` — **fail identically on a pristine `origin/main` tree on this macOS host** (the latter is a `/private/var` vs `/var` symlink comparison), so they are pre-existing and unrelated to this diff.

fixes #10351

Failure-Class: FC-nonexecuting-verification-command

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10852?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -49,6 +49,11 @@ checks:
     triggers: [".github/workflows/desktop_qualify_beta.yml", ".github/scripts/check-release-process-guards.py", "scripts/run-release-process-guards.sh", "backend/tests/unit/test_desktop_release_scripts.py", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "SCA-174: qualification must check out the exact release tag inside its isolated source directory; the guard and its focused behavioral test run on every relevant diff"
+  - id: desktop-release-process-guard-selftests
+    command: ["bash", "scripts/run-release-process-guard-tests.sh"]
+    triggers: [".github/scripts/check-release-process-guards.py", ".github/scripts/test_check_release_process_guards.py", ".github/scripts/fixtures/codemagic_workflow_contract/**", "codemagic.yaml", "scripts/run-release-process-guard-tests.sh", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the guard's own self-suite ran in no lane, so 9 of its cases drifted red on main unnoticed; an unwired guard test suite is indistinguishable from a working one"
   - id: failure-class-retirement-author
     command: ["python3", ".github/scripts/test_failure_class_retirement.py"]
     triggers: [".github/scripts/failure_class_retirement.py", ".github/scripts/test_failure_class_retirement.py", ".github/workflows/repo-hygiene.yml", "scripts/failure-class", ".github/checks-manifest.yaml"]
--- a/.github/scripts/test_check_release_process_guards.py
+++ b/.github/scripts/test_check_release_process_guards.py
@@ -19,15 +19,36 @@ GUARDS = importlib.util.module_from_spec(SPEC)
 SPEC.loader.exec_module(GUARDS)
 REVIEWED_PARENT = "4e391ee726642d99abc2c61966dcd80a836e6c1c"
 RAW_SCANNER_PARENT = "11ac6d9e9d27677d06d513364f2e658f5ed99870"
+CONTRACT_RELATIVE_PATH = ".github/scripts/fixtures/codemagic_workflow_contract/v1.json"
+
+
+def _show(revision: str, relative: str) -> bytes:
+    return subprocess.check_output(["git", "show", f"{revision}:{relative}"], cwd=REPO_ROOT)
 
 
 def _copy_contract_tree(tmp_path: Path, monkeypatch) -> Path:
     shutil.copy2(REPO_ROOT / "codemagic.yaml", tmp_path / "codemagic.yaml")
-    fixture = tmp_path / ".github/scripts/fixtures/codemagic_workflow_contract/v1.json"
+    fixture = tmp_path / CONTRACT_RELATIVE_PATH
     fixture.parent.mkdir(parents=True, exist_ok=True)
-    shutil.copy2(REPO_ROOT / ".github/scripts/fixtures/codemagic_workflow_contract/v1.json", fixture)
+    shutil.copy2(REPO_ROOT / CONTRACT_RELATIVE_PATH, fixture)
     monkeypatch.setattr(GUARDS, "ROOT", tmp_path)
     return tmp_path / "codemagic.yaml"
+
+
+def _copy_parent_contract_tree(root: Path, revision: str = REVIEWED_PARENT) -> Path:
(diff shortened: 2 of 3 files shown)


.github/checks-manifest.yaml
```diff
@@ -49,6 +49,11 @@ checks:
     triggers: [".github/workflows/desktop_qualify_beta.yml", ".github/scripts/check-release-process-guards.py", "scripts/run-release-process-guards.sh", "backend/tests/unit/test_desktop_release_scripts.py", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "SCA-174: qualification must check out the exact release tag inside its isolated source directory; the guard and its focused behavioral test run on every relevant diff"
+  - id: desktop-release-process-guard-selftests
+    command: ["bash", "scripts/run-release-process-guard-tests.sh"]
+    triggers: [".github/scripts/check-release-process-guards.py", ".github/scripts/test_check_release_process_guards.py", ".github/scripts/fixtures/codemagic_workflow_contract/**", "codemagic.yaml", "scripts/run-release-process-guard-tests.sh", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
```
tianmind-studio, line 55: This entry has no `platforms`, so Windows preflight selects it, but the command is not Windows-executable. `run-release-process-guard-tests.sh` hardcodes `backend/.venv/bin/python`; Windows `make setup` creates `backend/.venv/Scripts/python.exe`, so sync completes and the wrapper still invokes a nonexistent path. Bypassing that wrapper with the canonical Windows interpreter and running `python -m pytest .github/scripts/test_check_release_process_guards.py -q` produced **7 failed, 82 passed, 1 skipped**: the raw-byte assertion sees CRLF working-tree bytes, the open-file replacement probes cannot replace an open file on Windows, and `os.O_NOFOLLOW` is absent. Because this check is now in the local lane, any Windows contributor touching one of these triggers gets a deterministic red preflight. Please add `platforms: ["linux", "macos"]` for the current POSIX contract, or make the wrapper and those tests genuinely Windows-portable before leaving it unrestricted.

tianmind-studio (changes requested):
The historical-parent repair and CI registration are sound on POSIX, but the new local-lane check currently has a blocking Windows portability gap. I reproduced this on native Windows with the repository's canonical venv; details are inline. Please either scope this self-suite to `linux`/`macos` for this PR or port the wrapper and POSIX-specific race tests before enabling it on Windows.

kodjima33:
Closing this rather than leaving it open: the watchdog that opened it may not merge CI wiring (`.github/checks-manifest.yaml`), and @tianmind-studio's changes-requested (scope the self-suite to linux/macos, or port the wrapper + POSIX race tests to Windows) is not something this run addressed. The branch `watchdog/release-process-guard-selftests` is untouched — reopen or re-cut from it once the Windows scoping decision is made. Tracking issues remain #10351 and #10389.

🤖 automated by hourly watchdog — closed, not abandoned

PR #7512, opened 2026-05-27
fix(desktop): enforce Rewind privacy exclusions on proactive assistant LLM uploads

beastoin:
## Summary

Fixes #7098 — Rewind privacy exclusions (password managers, keychains) did not protect the proactive assistant LLM upload path. Frames from excluded apps could reach Gemini via Memory, Insight, Focus, and Task assistants.

### Changes

**Primary fix — plugin-level privacy gate:**
- `ProactiveAssistantsPlugin.swift`: Block `trackFrame`, `distributeFrameIfChanged`, and `distributeFrameDuringDelay` for Rewind-excluded apps in both macOS 14+ and 13.x capture paths
- After `windowGone` retry and active-window capture fallbacks, re-resolve `appName` and `isRewindExcluded` to prevent stale-metadata races

**Defense-in-depth — assistant-level checks (all 4 assistants):**
- `MemoryAssistantSettings.isAppExcluded`: Now also checks `RewindSettings.shared.isAppExcluded`
- `InsightAssistantSettings.isAppExcluded`: Same
- `FocusAssistantSettings.isAppExcluded`: Same
- `TaskAssistant.analyze` and `onContextSwitch`: Now check `RewindSettings.shared.isAppExcluded` before processing

**Test runner audit (all 3 runners):**
- `InsightTestRunnerWindow`: Filter out Rewind-excluded screenshots
- `FocusTestRunnerWindow`: Filter out Rewind-excluded screenshots
- `TaskTestRunnerWindow`: Filter out Rewind-excluded screenshots before context switch detection

**Tests:**
- `ScreenPrivacyExclusionTests.swift`: 8 tests covering Rewind defaults, per-assistant exclusion, built-in exclusions, pass-through for normal apps, custom user exclusions, and RewindSettings direct coverage

### Review cycle (6 iterations)
- **R1:** Fixed `windowGone` retry race — re-resolve privacy gate after retry
- **R2:** Clarified architectural scope — capture/metadata separation is pre-existing
- **R3:** Fixed `trackFrame` context-switch leak — moved `trackFrame` inside privacy gate
- **R4:** Added TaskAssistant defense-in-depth + TaskTestRunnerWindow privacy filter
- **R5:** Added post-capture re-resolution for macOS 14 windowID==nil and macOS 13.x paths

### Risk assessment
- Low risk: privacy gate is additive, no existing behavior changes for non-excluded apps

## Confidence: HIGH

**Why high confidence:**
[...]

Issue #7098, opened 2026-04-30
Desktop: Rewind privacy exclusions do not protect proactive assistant LLM uploads


waffensam:
**Describe the bug**
A privacy-sensitive Rewind exclusion does not currently protect the desktop proactive assistant LLM upload path.

Rewind defaults exclude privacy-sensitive apps such as `Passwords`, `1Password`, `1Password 7`, `Bitwarden`, `LastPass`, `Dashlane`, `Keeper`, `Enpass`, `KeePassXC`, and `Keychain Access`. However, the desktop capture loop still distributes the captured frame to proactive assistants before the Rewind exclusion is applied to `RewindIndexer`.

`MemoryAssistant` then uses its own exclusion check, which only combines `TaskAssistantSettings.builtInExcludedApps` and user-configured memory exclusions. The shared built-in proactive assistant list does not include the password/keychain apps above. As a result, a privacy-sensitive app can be excluded from Rewind storage/OCR but still be queued by `MemoryAssistant` and sent as `jpegData` to Gemini for memory extraction.

This is a privacy boundary bug before memory creation. Even if no memory is ultimately saved, the sensitive frame may already have been uploaded to the LLM proxy.

**To Reproduce**
Code-path reproduction, no real password-manager screenshot required:

1. Enable desktop proactive memory extraction.
2. Keep the default Rewind excluded apps enabled.
3. Make a Rewind-excluded privacy app frontmost, for example `Passwords`, `1Password`, or `Keychain Access`.
4. In `ProactiveAssistantsPlugin`, observe that the frame is created and distributed via `AssistantCoordinator.shared.trackFrame(frame)` and `distributeFrameIfChanged(frame)` before the later `if !isRewindExcluded` gate for `RewindIndexer`.
5. In `MemoryAssistant.analyze(frame:)`, observe that only `MemoryAssistantSettings.shared.isAppExcluded(frame.appName)` is checked.
6. In `MemoryAssistantSettings.isAppExcluded`, observe that it checks only `TaskAssistantSettings.builtInExcludedApps` plus user memory exclusions, not `RewindSettings.shared.isAppExcluded`.
7. `MemoryAssistant.processFrame` can then call `extractMemories(from: frame.jpegData, appName: frame.appName)`, which sends the image to Gemini through `GeminiClient.sendRequest`.

Relevant code paths:

[...]

PR #10853, opened 2026-07-29
feat(desktop): keep password fields and private browsing out of screen capture

### LINKED ISSUE #6677: Exclude Safari private browsing windows from Rewind screen recording
Feature request: Add the ability to exclude Safari private browsing windows from Rewind screen capture, similar to how the original Rewind.ai app handled this.

Current behavior:
The Rewind screen recording feature supports excluding entire applications from capture (via rewindExcludedApps / RewindSettings.isAppExcluded()), but this is all-or-nothing at the app level. There is no way to record regular Safari windows while skipping private browsing windows.

Desired behavior:
When Safari is not in the excluded apps list, Rewind should detect and skip frames where the frontmost Safari window is a private browsing window — without needing to exclude Safari entirely.

Why this is important:
Users who want their regular browsing captured for context (Rewind's core value proposition) are currently forced to choose between recording private browsing sessions or excluding Safari entirely. A per-window privacy filter would match the behavior users expect from the original Rewind app.

Implementation hint:
The active window is already resolved in ScreenCaptureService.swift via the Accessibility API and CGWindowListCopyWindowInfo. Possible detection approaches include accessibility/private-window inspection or Safari scripting, depending on what is actually reliable on the target macOS version.

For example, this could be implemented by either:
- Checking Safari window/tab state through the Accessibility API if a reliable private-window signal is available
- Querying via AppleScript: tell application "Safari" to get private browsing of current tab of front window

The check could slot into the existing isRewindExcluded gate in ProactiveAssistantsPlugin.swift.

katipally:
## What changed and why

Closes #6677, and closes the same weakness on a second surface that no issue covers.

The Rewind privacy gate (#7098, built in #7512) is correct: an excluded frame reaches no consumer. Its only input is `RewindSettings.excludedApps`, a set of app display names the user maintains, which is why nine password managers ship as hardcoded defaults. An enumeration cannot express "this window, right now" or "the user is typing a credential".

- **Private browsing (#6677).** Exclusion is per app, so keeping private windows out meant excluding the whole browser. A private window is now recognized from its own title and skipped, while regular windows of the same browser still capture. Folded into `isRewindExcluded` at the three resolution points #7512 established: initial resolve, the `windowGone` retry, and the macOS 13.x fresh-capture path.
- **Secure keyboard entry.** `IsSecureEventInputEnabled()`, raised process-wide while any app holds focus on a password field, was not called anywhere in the repo. A credential typed into a web form inside a normally-captured browser was imaged, OCR'd, stored, and reachable by the agent through `get_work_context`. The gate now runs ahead of ScreenCaptureKit, so the frame is never imaged rather than imaged and discarded, with a 2s backoff after the flag clears.

A flag an app never clears would pause capture indefinitely. The gate keeps skipping and reports once per continuous run through `recordFallback` (`area=screen_capture`, `reason=policy`, `outcome=degraded`, both already allowlisted).

Private-window matching is scoped by bundle identifier and anchored to the title suffix per `FC-meeting-trigger-title-identity-drift`, whose canonical prevention names exactly that plus retained negative near-suffix cases.

None. `scripts/pr-preflight --suggest` reports none for the changed paths.

## How it was verified

Measured on macOS 25.5.

Private-window markers, by loading the same URL in a normal and a private window of each browser and diffing the accessibility title:

| browser | private window | normal window |
|---|---|---|
| Safari | `Example Domain, Private Browsing` | `Example Domain` |
| Chrome | `Example Domain - Google Chrome (Incognito)` | `Example Domain - Google Chrome` |
| Brave | `Example Domain - Brave (Private)` | `Example Domain - Brave` |

Markers were stable across repeated samples and survived navigation. `AXIdentifier`'s `IsSecure=` field is TLS state, not privacy state, and reports `true` for a private window on an HTTPS page.

Confirmed through `ScreenCaptureService.getActiveWindowInfo`, the resolver the capture loop calls, with each window frontmost:

```
Safari private     p6.html, Private Browsing                     isPrivate=true
Safari normal      Example Domain                                isPrivate=false
Chrome incognito   Example Domain - Google Chrome (Incognito)    isPrivate=true
Chrome normal      (empty, unloaded tab)                         isPrivate=false
Brave private      Example Domain - Brave (Private)              isPrivate=true
```

Secure input, A/B/A on a named `omi-privacy` bundle with Rewind capturing. One Safari window cycling full-screen background colours so frame dedupe could not mask the result; only the focused field changed between phases. Counts are rows in the live Rewind DB:

| phase | focused field | `IsSecureEventInputEnabled()` | frames stored in 34s |
|---|---|---|---|
| A | text | false (30/30 samples) | 5 |
| B | password | true (30/30 samples) | **0** |
| A2 | text | false (30/30 samples) | 1 |

- `xcrun swift build --package-path Desktop` — Build complete
- `xcrun swift test --filter 'SecureInputCaptureGateTests|PrivateBrowsingWindowTests'` — 20 passed
- `desktop/macos/test.sh` — exit 0, 434 Swift suites
- `python3 desktop/macos/scripts/check_desktop_test_quality.py` — passed, at or below baseline
- `bash desktop/macos/scripts/swiftlint-wrapper.sh lint` — 0 violations across 962 files
- `OMI_PR_BODY_FILE=... make preflight` — 20 selected local checks passed

Not verified: browsers other than Safari, Chrome and Brave, none of which are installed on the test machine.

## Tests

Both policies are pure and injectable, driven through their production API with no wall-clock waits.

- `SecureInputCaptureGateTests` (7) — skip while active, skip inside the backoff, resume after it, stuck-flag reported once per run and re-armed for a later run, reset clears state.
- `PrivateBrowsingWindowTests` (13) — the three measured markers, plus negative cases: regular windows of the same browsers, a marker embedded mid-title, a marker from the wrong browser, an unknown bundle, and empty or missing title or bundle.

Both were mutation-tested. Relaxing `hasSuffix` to `contains` and dropping the backoff each fail the suite. The first mutation initially passed, which exposed a missing negative case; the two mid-title cases close it.

## Failure class (fixes)

Failure-Class: none

Both commits are `feat:`. No stated contract said an unexcluded app must not be captured, so this adds protection that did not exist rather than repairing a violated boundary. No registry class covers "the privacy boundary's only input is a user-maintained enumeration", and every one of the 55 classes cites a real prior instance, which this has none of.

## Scoped cleanups

None.

## Known limitations

All resolve toward capture rather than toward a false sense of privacy.

- The private-window marker exists only in the accessibility title. `ScreenCaptureService` falls back to `kCGWindowName`, which carries the page title alone: a private Safari window arrives as `p6.html`, a Chrome incognito window as `New Incognito Tab`. Detection therefore holds while Omi is trusted for Accessibility, the permission it already requires. Recorded in the `PrivateBrowsingWindow` doc comment. Secure input needs no Accessibility and is unaffected.
- A private window that has not loaded a page reports an empty title and is captured; the content there is the browser's own incognito splash.
- Markers are the English localizations. A non-English browser keeps today's behaviour.
- Arc, Firefox, Edge, Vivaldi and Opera are absent rather than guessed. Each is a one-line entry once measured.

## Line-count ratchet

`ProactiveAssistantsPlugin.swift` moves 1739 to 1775. Both deciding policies live in their own files under `Core/`, leaving only call sites in the plugin; the baseline is raised with a justification rather than splitting a 1775-line file inside this change.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10853?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-proactiveassistants.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-proactiveassistants.json
@@ -1,11 +1,11 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/TaskExtraction/TaskAssistant.swift": 1742,
-    "desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift": 1739
+    "desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift": 1775
   },
   "raise_justifications": {
     "desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/TaskExtraction/TaskAssistant.swift": "Task candidate context now carries only backend action-item IDs, while local SQLite/staged evidence remains id-less so the model cannot target it for a backend mutation.",
-    "desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift": "Rewind capture policy adds per-app heartbeat and preview skip logic inline; also registers SuggestionAssistant (event-driven) with logic in Assistants/Suggestions/."
+    "desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift": "Adds the secure-input and private-browsing capture skips at the one place the privacy gate lives; the deciding logic sits in Core/SecureInputCaptureGate.swift and Core/PrivateBrowsingWindow.swift, leaving only the call sites here."
   },
   "threshold": 1500
 }
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/PrivateBrowsingWindow.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/PrivateBrowsingWindow.swift
@@ -1,0 +1,40 @@
+import Foundation
+
+/// Recognises a browser window in a private / incognito session so Rewind can skip that
+/// window alone, instead of forcing the user to exclude the whole browser and lose the
+/// regular browsing context Rewind exists to keep (#6677).
+///
+/// Matching is scoped by bundle identifier and anchored to the end of the window title, per
+/// `FC-meeting-trigger-title-identity-drift`: a title pattern must be bound to a known
+/// browser process and kept alongside negative near-suffix cases, or a generic title starts
+/// matching.
+///
+/// Suffixes were verified on macOS 25.5 by loading the same URL in a normal and a private
+/// window of each browser and diffing the accessibility title. A browser whose marker has
+/// not been measured is deliberately absent — a missing entry captures exactly as it does
+/// today, while a guessed one would claim protection nobody observed.
+///
+/// The marker exists only in the accessibility title. When `ScreenCaptureService` cannot use
+/// the Accessibility API it falls back to `kCGWindowName`, which was measured to carry the
+/// page title alone — a private Safari window arrives as `p6.html` and a Chrome incognito
+/// window as `New Incognito Tab`, with no marker to match. Detection therefore holds only
+/// while Omi is trusted for Accessibility, the permission it already requires.
+///
+/// Two further gaps, all three resolving toward capture rather than a false sense of privacy:
+/// a private window that has not loaded a page yet reports an empty title, and the markers
+/// are the English localizations.
+enum PrivateBrowsingWindow {
+  private static let privateTitleSuffixes: [String: String] = [
+    "com.apple.Safari": ", Private Browsing",
+    "com.google.Chrome": " (Incognito)",
+    "com.brave.Browser": " (Private)",
+  ]
+
+  static func isPrivate(windowTitle: String?, bundleIdentifier: String?) -> Bool {
+    guard let bundleIdentifier,
+      let suffix = privateTitleSuffixes[bundleIdentifier],
+      let windowTitle
+    else { return false }
+    return windowTitle.hasSuffix(suffix)
+  }
+}
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/SecureInputCaptureGate.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/SecureInputCaptureGate.swift
@@ -1,0 +1,82 @@
+import Foundation
+
+/// Decides, once per capture tick, whether Omi must skip screen capture because macOS
+/// reports **secure keyboard entry** active — the process-wide flag an app raises while a
+/// password field holds focus (browser logins, `sudo`, the login window, 2FA prompts).
+///
+/// `RewindSettings.defaultExcludedApps` can only name password managers someone thought to
+/// list; secure input is raised by every app that handles a credential correctly, including
+/// a web form inside a browser that is otherwise captured normally. Both feed the same
+/// privacy gate established by #7098 — an excluded frame reaches no consumer.
+///
+/// A short backoff after the flag clears covers the frames where the credential is still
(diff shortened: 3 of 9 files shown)


Git-on-my-level:
Thanks for the focused privacy work here. The direction looks useful for Rewind: checking secure keyboard entry before ScreenCaptureKit means credential pixels are not captured in the first place, and scoping private-window detection by browser bundle ID plus suffix keeps regular browser windows available. I also appreciate the unit coverage around the secure-input backoff/stuck-report behavior and the private-browsing near-miss cases; the relevant desktop Swift checks are passing on this head.

I’m not formally approving because this changes security/privacy-sensitive capture behavior and the PR also needs human product/maintainer sign-off on the exact privacy contract before merge. The main things I’d want a maintainer to explicitly confirm are:

- The stated fail-open private-browsing gaps are acceptable: missing Accessibility title, unloaded/blank private windows, unmeasured browsers, and non-English browser title markers currently resolve toward capture rather than protection.
- The secure-input global pause/backoff UX is acceptable if another app leaves secure keyboard entry stuck; the implementation does surface one degraded diagnostic while continuing to pause capture.
- The limited browser coverage (Safari, Chrome, Brave) is the desired initial scope rather than broadening to Edge/Chromium variants already known elsewhere in the desktop codebase.

No blocking implementation issue jumped out from the diff, so this is a positive signal from automation, with `security-review` / `needs-maintainer-review` left for human sign-off.

---

aryanorastar:
Not merging this.

Head is still 2d4cc8e (Jul 29). GitHub is CONFLICTING on ProactiveAssistantsPlugin.swift and the proactiveassistants line-count ratchet. #6677 is still OPEN: main's capture loop still sets isRewindExcluded from RewindSettings.isAppExcluded only. IsSecureEventInputEnabled is still unused.

The two Core/ helpers and the tests are the right shape — suffix+bundle, measured markers only, fail-open documented, mutation-tested near-misses. They cannot land until they are re-homed on current captureFrame.

That function grew after this tip. isAppExcluded now also gates an early excluded_app return and ContextDwellRefreshPolicy. A rebase that only patches the original three isRewindExcluded assignments will still capture a private window on the dwell-refresh path. Fold PrivateBrowsingWindow into every capture decision that has a window title; keep the secure-input skip ahead of ScreenCaptureKit.

Fail-open gaps (no AX title, blank private splash, English-only markers, Safari/Chrome/Brave only) stay a human privacy-review call. I am not approving a DIRTY capture-privacy change, and I am not treating unmeasured browsers as a merge blocker if they stay absent rather than guessed.

This does not gate floating-bar captureScreenData / captureScreenJPEG. Password pixels can still be attached on that path. Say so in the PR body if that is not already the contract.

aryanorastar (approved):
Verified on 2d4cc8e1fafe.

Capture returns before ScreenCaptureKit while `IsSecureEventInputEnabled()` is set (plus a 2s backoff). Private Safari/Chrome/Brave windows skip by bundle+title suffix. Tests cover the gate and the title markers. Fail-open when Accessibility is missing is documented in the source.

Honest gap: no on-device smoke (no kit). Branch is CONFLICTING. I am not merging.

aryanorastar:
Keeping this open on 2d4cc8e1fafe.

SecureInputCaptureGate and PrivateBrowsingWindow are absent on main. This head is dirty, so it is not an approval.

Not approving this SHA.

aryanorastar:
Keeping this open. Not approving `2d4cc8e1fafe`. `SecureInputCaptureGate.swift` absent on origin/main. mergeable CONFLICTING / DIRTY. Desktop capture privacy, not pendant BLE.

PR #10859, opened 2026-07-29
fix(macos): harden onboarding permissions and dev bundles

undivisible:
## Summary

- Make macOS onboarding language and permission handling complete at the point of use, including weather location requests and background connector safety.
- Package runtime resources correctly and use Apple’s strip tool so named development bundles launch reliably.
- Remove the transient authentication loading artifact and keep onboarding state resilient while users grant permissions.

- INV-AGENT-*
- INV-AUTH-1
- INV-CHAT-1
- INV-INT-1

## Validation

- focused macOS permission tests and tool-manifest tests
- upstream-based pre-push checks, including Swift format/lint, e2e coverage, tool-surface generation, and dev-bundle tests

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10859?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-chat.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-chat.json
@@ -1,11 +1,11 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": 2204,
-    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": 4369
+    "desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift": 4380
   },
   "raise_justifications": {
     "desktop/macos/Desktop/Sources/Chat/AgentBridge.swift": "deleteBackend flag threaded through both clearJournalTurns variants so a reset is local-only",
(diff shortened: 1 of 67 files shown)


Git-on-my-level (changes requested):
Thanks for hardening the onboarding and bundle paths here — the direction looks useful, especially avoiding background cookie/keychain access during onboarding import attempts and adding compact-layout coverage.

I found one blocker in the new location-permission path:

- `LocationPermissionService.requestCurrentLocation()` calls `requestWhenInUseAuthorization()`, but then only accepts `.authorizedAlways` before requesting the current location. A normal successful When In Use grant returns `.authorizedWhenInUse`, so the new weather/location tool path will report `denied` even after the user allows the macOS prompt; `check_permission_status` has the same issue through `isAuthorized`. Please accept `.authorizedWhenInUse` for this When In Use permission path, and add a focused test/adapter seam so the status/result mapping is covered without relying on the live CoreLocation prompt.

Also flagging for maintainer review: this PR changes agent/tool behavior by teaching the desktop agent to request Location for weather questions without an explicit place. That may be the right product behavior, but it is a privacy-sensitive agent-instruction change, so a human maintainer should confirm the wording and consent boundary before merge.

I did not run the macOS build/tests in this Linux review environment; this is a static maintainer review against the diff.

---

undivisible:
Superseded by #10874, which has the identical head commit on a BasedHardware-owned branch as part of the fork retirement.

PR #10861, opened 2026-07-29
fix(sync): replace previews from canonical capture

axAilotl:
## Summary

- Add an explicit `merge | replace` transcript mutation contract to the v2
  sync endpoint and carry it unchanged through Cloud Tasks.
- Admit canonical replacement only for one signed mobile capture bound to the
  authenticated device and existing conversation.
- Accumulate every recovered STT segment, then atomically replace the live
  preview timeline and reprocess the conversation once only after the entire
  capture succeeds.
- Preserve the deployed merge/idempotency behavior for existing clients and
  queued tasks while domain-separating replacement content claims.
- Regenerate the app-client OpenAPI contract and generated Swift/TypeScript
  clients; document replacement and retry semantics.
- Raise the exact frozen line-count baselines for the two existing sync owner
  files, with checked-in justifications: request authorization stays at the
  route boundary, and the all-or-nothing transcript commit stays inside the
  existing lease/ledger/retry coordinator rather than creating another owner.

## Root cause

The mobile recovery client already requested `transcript_mode=replace`, but the
backend endpoint did not define or propagate that operation. Recovery therefore
fell through the ordinary fragment-merge path, which could leave preview text
in place, publish partial mutations, and re-run conversation processing once
per fragment. The API boundary did not own a typed distinction between gap
repair and authoritative capture replacement.

## Safety and compatibility

- Replacement fails admission before staging unless the upload has exactly one
  file, an existing target conversation, a valid signed capture manifest, and
  matching server capture provenance.
- Any VAD/STT segment failure leaves the live preview unchanged and retries the
  complete canonical capture.
- Unknown explicit task modes fail closed; tasks admitted before this field
  existed continue as `merge`.
- A lease/lifecycle fence for the explicitly targeted replacement conversation now terminalizes the whole targeted job as superseded; ordinary multi-conversation merge work still isolates a fenced sibling and continues safely.
- A no-speech provider result retains the preview instead of deleting
  user-visible text.
- No production deployment is part of this PR.

## Verification

- `backend/.venv/bin/python -m pytest backend/tests/unit/test_sync_v2.py -k "merged_reprocess_fence_isolated or targeted_reprocess_fence_supersedes" -q` — 2 passed.
- `backend/.venv/bin/python -m pytest backend/tests/unit/test_sync_v2.py -q` — 169 passed.
- `npm run test:sync-cloud-tasks-stack:emulator` — passed the exact Sync Cloud Tasks stack gauntlet that had failed in CI.
- `make preflight` — 34/34 deterministic checks passed on the final rebased SHA.
- `backend/.venv/bin/pytest -q backend/testing/e2e/test_sync_lifecycle.py` — 3 passed, including signed-manifest upload → inline decode/VAD/STT seams → atomic Firestore replacement → one reprocess → completed job poll.
- `backend/.venv/bin/python backend/scripts/export_openapi.py --surface app-client --check` — passed.
- The normal bounded pre-push gate passed, including the full desktop Swift debug compile.
- The broad `backend/test.sh` run reached one unrelated static-guard mismatch already present on `origin/main`: `test_memory_import_static_guards.py` expects a single-line `post("v3/memory-imports/batch"` call while the current source formats the argument on the following line. PR-specific sync tests and the exact failing gauntlet pass.
- Android and iOS both use the same Dart canonical phone-side assembler. The Android physical run proved one canonical WAL but its acceptance record explicitly left terminal server-result chronology untested. The iPhone run followed that final artifact and exposed the shared backend boundary: the client sent `transcript_mode=replace`, while the deployed endpoint ignored it. Live replacement proof remains intentionally pending a non-production backend deployment.

None.

## Failure class

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json
@@ -4,7 +4,7 @@
     "backend/routers/chat.py": 1588,
     "backend/routers/developer.py": 2263,
     "backend/routers/mcp_sse.py": 1865,
-    "backend/routers/sync.py": 2058,
+    "backend/routers/sync.py": 2087,
     "backend/routers/users.py": 2076
   },
   "raise_justifications": {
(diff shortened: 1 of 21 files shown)


Git-on-my-level (commented):
Thanks for the detailed regression coverage here. The typed `merge`/`replace` contract, manifest-bound admission, replacement content-id domain separation, and the all-or-nothing commit shape look directionally right for canonical mobile recovery.

I’m leaving this as maintainer feedback rather than approval because this is still a draft, it changes the transcript mutation semantics for user conversation data, and required CI is currently red. Before marking ready, please clear the failing checks:

- Backend unit suite: `tests/unit/test_memory_import_static_guards.py::test_import_evidence_client_targets_import_endpoint`
- Desktop Swift Static & Test Contracts / Desktop Swift Build & Tests: Swift main-actor isolation failures in `ChatTimelineContinuityTests.swift` plus the launcher script test gate

Given the data-mutation surface, a human maintainer should also sanity-check the product/rollback semantics for replacing a live preview from a canonical capture before merge. No security/supply-chain concern found in this pass.

---

axAilotl:

PR #10869, opened 2026-07-29
SCA-251: skip dev startup Stripe price validation

Git-on-my-level:
## Summary

- Skip the diagnostic Stripe price-ID validation only during `OMI_ENV_STAGE=dev` startup.
- Keep the validation active in production and document that dev catalog/checkout calls still require mode-matched Stripe credentials.

## Verification

- `./backend/.venv/bin/pytest backend/tests/unit/test_subscription_plans.py -q` — 14 passed
- `black --check --line-length 120 --skip-string-normalization backend/utils/subscription.py backend/tests/unit/test_subscription_plans.py`
- `bash backend/test.sh` was stopped after unrelated fast-unit-duration guard failures in `tests/services/users/test_data_export.py` and `tests/unit/test_activate_task_intelligence_dogfood_user.py`; the touched subscription suite passed.

## Follow-up

The dev `STRIPE_API_KEY` must still be changed to the matching test-mode key (or the configured price IDs changed to prices for its current key) before the catalog and checkout flows can work in dev.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10869?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/AGENTS.md
+++ b/backend/AGENTS.md
@@ -13,7 +13,7 @@ source .venv/bin/activate
 uvicorn main:app --host 0.0.0.0 --port 8080
 ```
 
-**Env stages** (`OMI_ENV_STAGE`): `local` (emulator harness, `.env.local-dev`), `offline` (fake providers, `.env.offline`), `dev` (remote dev GCP, `.env.dev`), `prod` (reference only, `.env.prod`). `load_backend_env()` loads the stage file then `backend/.env` overrides. Templates: `backend/.env.*.template`. Harness: `PROVIDER_MODE=offline make dev-up` or `OMI_ENV_STAGE=offline`.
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
@cursor address review comment

Git-on-my-level:
Addressed Codex P1 feedback (discussion_r3678409063): the dev startup Stripe validation skip now calls `record_fallback` before the early return so `omi_fallback_total` tracks dev deployments that start without price validation. The existing regression test asserts the fallback call with the expected labels.

Verification: 14/14 subscription unit tests pass; `black --check` clean; CI green (Backend unit suite, Hermetic Backend E2E, Formatting, Public Developer API contract all pass).

kodjima33 (approved):
Dev-only skip of the startup Stripe price validation; prod path unchanged and covered by a new test.

PR #10831, opened 2026-07-29
fix(security): resolve Dependabot security vulnerabilities across workspace

undivisible:
## Summary
Resolves open Dependabot security vulnerabilities across the workspace by updating dependency manifests to **patch-level range specifiers** (`~=` for Python, `~` for Node package.json/overrides), allowing automatic inclusion of future security patch releases in each minor version group without major/minor version breaking changes.

## Changes by Subsystem & Commit Structure

1. **`fix(backend): update vulnerable python dependencies to patch ranges (~=)`** (`b970028fc6`)
   - Configured Python requirements (`backend/requirements.txt`, `backend/openapi-requirements.txt`, `backend/pusher/requirements.txt`, `backend/modal/requirements.txt`) using PEP 440 compatible release constraints (`~=` e.g., `aiohttp~=3.14.1`, `cryptography~=48.0.1`, `GitPython~=3.1.55`, `httplib2~=0.32.0`, `langchain~=1.3.14`, `langchain-core~=1.4.9`, `langgraph~=1.2.10`, `langgraph-checkpoint~=4.1.1`, `langgraph-sdk~=0.4.2`, `langgraph-prebuilt~=1.1.0`, `langsmith~=0.8.18`, `msgpack~=1.2.1`, `pillow~=12.3.0`, `pyarrow~=23.0.1`, `pyasn1~=0.6.4`, `PyJWT~=2.13.0`, `python-multipart~=0.0.31`, `starlette~=0.49.3`, `torch~=2.10.0`, `tornado~=6.5.7`, `websockets~=15.0.1`).
   - Regenerated Python lockfiles (`pylock.toml`, `pylock.macos.toml`, `pylock.windows.toml`, etc.) via `./backend/scripts/update-python-lock.sh`.

2. **`fix(desktop): update vulnerable dependencies in desktop apps to patch ranges (~)`** (`364570239b`)
   - Configured SemVer tilde patch ranges (`~`) in `package.json` and `overrides` for `desktop/macos/agent`, `desktop/macos/demo`, `desktop/macos/pi-mono-extension`, and `desktop/windows`.

3. **`fix(web): update vulnerable dependencies in web applications to patch ranges (~)`** (`c6e5bac9ee`)
[...]

PR #10875, opened 2026-07-30
fix(security): resolve Dependabot security vulnerabilities across workspace

undivisible:
Migrated from #10831 to move the work from the `undivisible/omi` fork into a BasedHardware-owned branch. The original PR remains temporarily for review-history reference until the migration is verified.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10875?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class

Failure-Class: none


Git-on-my-level (dismissed):
Thanks for migrating this security-update work onto a BasedHardware branch. I reviewed this as a dependency/security-sensitive PR.

Requesting changes before maintainer review/merge:

- Several backend runtime requirement files move from exact pins to compatible-release ranges, for example `backend/requirements.txt`, `backend/pusher/requirements.txt`, and `backend/openapi-requirements.txt`. For a broad security remediation touching backend, pusher, OpenAPI generation, desktop, web, SDK, docs, plugins, and MCP surfaces, please keep runtime/security-critical requirements reproducible unless there is a deliberate repo-wide policy change. The matching lockfiles still pin specific artifacts, but these requirement files can be installed directly in some environments, so `~=` ranges can pull unreviewed future patch/minor releases and make builds/security rollouts less deterministic.
- This also changes dependencies that influence AI/agent behavior in this repo: the backend requirement files update LangChain/LangGraph/LangSmith-related packages, and the desktop package overrides touch `@earendil-works/pi-coding-agent` transitive packages. That may change how Omi's AI runtime/coding-agent integrations behave, so a human maintainer should confirm these versions are safe and compatible rather than treating this as a purely mechanical Dependabot sweep.
- Please add a short validation note for the dependency-resolution process used for each workspace/package manager, especially where both manifests and lockfiles changed. A concise list of commands run (or why a workspace is lockfile-only) would make this much safer to review.

I did not see a supply-chain attack pattern in the diff itself; the concern is reproducibility/maintainer confidence for a very broad dependency/security update. Once the direct-install requirement files are re-pinned or the range change is explicitly justified and validated, this should be much easier to review.

---

Git-on-my-level (commented):
Thanks for the update. I re-reviewed the current head after the runtime requirements were re-pinned to exact versions and the validation notes were added in the latest commit.

The previous automation blocker about direct-install backend requirement files using broad compatible-release ranges looks resolved on this head: the backend/OpenAPI/pusher requirement files now use exact pins that match the lock-backed dependency update approach.

I still think this should stay with human maintainer/dependency review before merge rather than receive automatic approval, because it is a broad security/dependency sweep across backend, desktop, web, SDK, docs, plugins, and MCP surfaces. The CI signal is good, but maintainers should verify that the selected package upgrades are intended across all of those workspaces.

Agent/AI-runtime impact to be aware of: the backend requirement updates include LangChain/LangGraph/LangSmith-family packages in runtime/OpenAPI/pusher dependency sets, and the desktop agent/package overrides affect `@earendil-works/pi-coding-agent` transitive resolution. That can alter Omi AI runtime behavior and local coding-agent behavior, so this should be reviewed as an agent-behavior-sensitive dependency update, not just a mechanical vulnerability bump.

I do not see a supply-chain attack pattern in the diff from the cached context; this is a maintainer confidence/dependency-review call.

---

kodjima33 (approved):
Dependabot security bumps across the workspace — approving; too large a lockfile sweep to land unattended.

PR #10878, opened 2026-07-30
fix(app): suppress re-insertion of deleted memories/tasks during background refresh

Git-on-my-level:
## Summary

Fixes a race condition where deleted memories and tasks reappear in the UI until the app is restarted.

**Root cause:** Optimistic UI deletion removes items from the provider's in-memory list immediately, but the actual server DELETE fires after a delay (4-second undo timer for memories; async for action items). Any background data refresh (`loadMemories()`, `fetchActionItems()`) that fires during that window re-fetches the still-present Firestore document and overwrites the provider's list — the deleted item reappears.

**Fix:** Track pending-deletion IDs in both `MemoriesProvider` and `ActionItemsProvider`. Filter those IDs out of server response data during load/fetch cycles. Once the server delete completes (or fails), remove the ID from the pending set so subsequent refreshes behave normally.

## Changes

- **`app/lib/providers/memories_provider.dart`** (+9 lines): After `loadMemories()` assigns server response, filter out `_pendingDeletionId` if set
- **`app/lib/providers/action_items_provider.dart`** (+22 lines): New `_pendingDeletionIds` set; track ID before optimistic removal, filter in `fetchActionItems()`, clean up on success/failure/error
- **`app/test/providers/deletion_consistency_test.dart`** (new, +153 lines): 5 regression tests covering optimistic removal, restore, consecutive deletions, pending-state guard invariant, and undo-window race condition

## Test Results

- 5/5 new tests pass
- Full provider test suite: no regressions
- Pre-existing flaky test in `conversation_detail_provider_selection_test.dart` unchanged

## Failure-Class

Failure-Class: none

### DIFF
--- a/app/lib/providers/action_items_provider.dart
+++ b/app/lib/providers/action_items_provider.dart
@@ -232,9 +232,10 @@ class ActionItemsProvider extends ChangeNotifier {
         final serverIds = response.actionItems.map((e) => e.id).toSet();
         // Filter into a new list rather than mutating response.actionItems
         // in-place: the response list may be unmodifiable, and aliasing it
-        // would also retire tombstones prematurely.
+        // would cause removeWhere to throw in deleteActionItem/deleteSelectedItems
+        // and retire tombstones prematurely.
         _actionItems = _pendingDeletionIds.isEmpty
-            ? response.actionItems
+            ? List.of(response.actionItems)
             : response.actionItems.where((item) => !_pendingDeletionIds.contains(item.id)).toList();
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
Pending-deletion tracking stops deleted memories/tasks reappearing on background refresh; holding the merge until Dart Analyze is green.

kodjima33 (approved):
Bug fix 4/5: pending-deletion tombstones survive concurrent refresh in both providers; root cause stated, regression tests added, CI green.

PR #10883, opened 2026-07-30
SCA-256: gate Windows releases on live update-feed route

Git-on-my-level:
## Summary

- Root cause of Windows auto-update breakage after v1.0.30: production is missing `/v2/desktop/update-feed/windows` (merged in #10610) while Windows clients fail closed without it.
- Gate `desktop_windows_release.yml` on a production probe of that route **before** tagging, so a missing backend deploy cannot mint another broken cohort.
- Register failure class `FC-ship-before-required-route` and document the deploy-before-ship contract.
- Restore `--check-id` support in `run_checks.py` and the Windows portability manifest checks that #10825 dropped while `windows-preflight-portability.yml` still depends on them (fixes Native Windows process and launcher contracts).
- Keep the Windows python-path detect bash-3.2-safe so macOS pre-push can source the harness resolver.

Closes SCA-256

## Test plan

- [x] `python3 .github/scripts/test_probe_windows_update_feed.py` (8 passed)
- [x] `python3 .github/scripts/test_run_checks.py` (40 passed)
- [x] `python3 .github/scripts/run_checks.py --lane ci --base HEAD --head HEAD --platform windows --check-id check-manifest-contract --list`
- [ ] CI green on PR #10883

## Failure class

Failure-Class: new

## Product invariants

None matched by path globs for this change.

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -24,6 +24,11 @@ checks:
     triggers: [".github/scripts/verify-pre-tag-readiness.py", ".github/scripts/test_verify_pre_tag_readiness.py", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "the receipt verifier must reject stale, non-offline, failed, and promotion-bearing readiness evidence before a tag mutation"
+  - id: windows-update-feed-probe-tests
+    command: ["python3", ".github/scripts/test_probe_windows_update_feed.py"]
+    triggers: [".github/scripts/probe_windows_update_feed.py", ".github/scripts/test_probe_windows_update_feed.py", ".github/workflows/desktop_windows_release.yml", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "Windows releases must not tag while production is missing /v2/desktop/update-feed/windows; the probe's trust-boundary validation is the release-lane guard"
   - id: pre-push-final-pr-diff
     command: ["bash", "scripts/test-pre-push-diff-base.sh"]
     triggers: ["scripts/pre-push", "scripts/pre-push-diff-base", "scripts/test-pre-push-diff-base.sh", ".github/checks-manifest.yaml"]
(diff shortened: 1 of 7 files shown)


Git-on-my-level:
Monitor review-fix pass — addressed all 6 trusted Codex threads on the Windows update-feed deploy gate.

| # | Pri | Thread | Fix |
|---|-----|--------|-----|
| 1 | P2 | Break-glass path for probe | Added `bypass_update_feed_probe_reason` workflow_dispatch input |
| 2 | P1 | Probe default stable channel | Probe now runs `--channel beta --channel stable` (backend default is `stable`, never falls through to beta) |
| 3 | P2 | Require confirmed, tracked bypass | Bypass now requires the literal sentinel `I ACKNOWLEDGE THE ROUTE IS REQUIRED`; empty/whitespace/arbitrary text still runs the probe |
| 4 | P2 | Env-bind bypass reason | Reason passed via `BYPASS_REASON` env var + `printf`, not direct `${{ }}` interpolation (injection-safe in a `contents: write` checkout) |
| 5 | P2 | Re-probe on retried publish | Added re-probe step in `build-and-publish` (after checkout, before build) — covers the case where GitHub Actions reuses cached `plan-and-tag` outputs on retry |
| 6 | P1 | `python3` not found on Windows | Added `actions/setup-python@v6` + invoke `python` (not `python3`), matching `windows-preflight-portability.yml` |

**Tests:** `python3 .github/scripts/test_probe_windows_update_feed.py` — 10 passed (+2 new: multi-channel probe, stable-missing-no-fallthrough).

Git-on-my-level:
Resolved all 6 Codex review threads — the code fixes from the prior pass were already live at the PR head (`8cacbdd`) and CI is green, but the threads had been left unresolved.

Quick re-verification against the current workflow before resolving:

- Break-glass input + sentinel-confirmed bypass present
- Probe runs both `beta` and `stable` (no stable→beta fallthrough)
- Bypass reason bound via env var (`printf`, not direct interpolation)
- Re-probe step in `build-and-publish` + `actions/setup-python` before it (uses `python` on Windows)

All checks: pass.

Issue #10894, opened 2026-07-30
Feature Request: Add a clear way to return home from the chat window in the Omi desktop app


hugoaap-code:
## Summary
In the Omi desktop app, when opening a chat window, there is no clear way to return to the home screen.

There is no obvious home button, no clear `X` to close the chat view, and the navigation feels vague.

## Problem
Once inside a chat window, the app does not provide a strong visual cue for how to go back home.

This creates uncertainty in the interface and makes navigation less intuitive than it should be.

## Expected behavior
- The chat window should have a clear and visible way to return to the home screen
- Users should not have to guess how to leave the chat view
- The return action should be obvious even to first-time users

## Current behavior
- Open a chat window in the desktop app
- There is no obvious home button
- There is no clear `X` or close affordance for the chat view
- The user experience feels vague because it is not obvious what action returns the user to home

## Why this matters
This is a basic navigation affordance.

If users cannot immediately tell how to go back, the interface feels unclear and less polished. Even if there is technically a way back, it should be visually obvious.

## Suggested improvements
- Add a visible `Home` button in the chat view
- Or add a clear back button with explicit labeling
- Or add a visible close/exit affordance for the chat screen
- Consider adding a subtle visual hint or navigation indicator so users understand how to return to home

## Notes
This is primarily a UX/navigation clarity request for the Windows desktop app.

The goal is not to redesign the whole chat screen, only to make the path back to home obvious and discoverable.

PR #10901, opened 2026-07-30
feat(desktop-windows): give the Hub chat panel a visible way back home

kodjima33:
## What

The Hub chat panel now carries a close control in its header row, so there is a
visible way back to the home screen.

## Why

#10894 (user report, Windows desktop): *"when opening a chat window, there is no
clear way to return to the home screen — no obvious home button, no clear X, the
navigation feels vague."*

They were right: the panel could only be left with `Esc` or a mousedown on the
Hub paper around the stage. Both work, neither is visible, and neither is
discoverable to a first-time user.

The control lives in the panel's own header row rather than floating in the
corner, so when the multi-chat header is on it sits **beside** the `+` / history
buttons instead of on top of them. Scope is deliberately just the affordance —
no redesign of the chat screen, as the issue asked.

## Tested

- `pnpm typecheck` — clean; `pnpm lint` — 0 errors; `pnpm test` — **5117 passed**, 24 skipped, 0 failed.
- **Real app, not a mock:** drove the built app (`out/main/index.js`) through
  Playwright `_electron` — clicked the ask bar, the chat panel took the stage with
  the close control visible, clicked the control, the stage returned to the
  resting hub with the `omi.` wordmark back on screen. Screenshots captured at
  each step.
- **Regression proof:** the new `HomeHub` test fails on the parent commit
  (`Unable to find a label with the text of: Close chat`) and passes here.
- `make preflight` — 10 checks passed. No product invariants affected.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10901?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/changelog/unreleased/2026-07-hub-chat-close.json
+++ b/desktop/windows/changelog/unreleased/2026-07-hub-chat-close.json
@@ -1,0 +1,5 @@
+{
+  "changes": [
+    "The chat panel now carries a close button, so there is a visible way back to the home screen — Esc still works."
+  ]
+}
--- a/desktop/windows/src/renderer/src/components/home/hub/HomeHub.test.tsx
+++ b/desktop/windows/src/renderer/src/components/home/hub/HomeHub.test.tsx
@@ -168,6 +168,24 @@ describe('HomeHub — stage machine', () => {
     expect(mode()).toBe('hub')
   })
 
+  it('returns to the hub from the chat panel own close control', () => {
+    // Esc and a click on the paper both leave the panel, but neither is VISIBLE:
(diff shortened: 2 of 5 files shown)


undivisible (approved):
Reviewed the visible close action and its focused Hub regression coverage.

PR #6987, opened 2026-04-24
fix(ble): force reconnect on CDM appear (android) + kick stuck .connecting on foreground (ios)

mdmohsin7:
## Problem

In congested RF environments (events, venues with heavy BLE/WiFi traffic), when the device disconnects and the app retries, the pending reconnection attempt can sit indefinitely waiting for advertising packets. Users had to bring the phone within 1-2 inches of the device for it to finally reconnect.

Happens on both Android and iOS — different underlying mechanisms but same user experience.

## Android root cause

1. Device disconnects → \`handleRetryLogic\` schedules retry
2. Retry fires → \`connectGatt(autoConnect=true)\` → \`currentGattHash\` set, GATT sits on passive background scan
3. \`CompanionDeviceManager\` fires \`EVENT_BLE_APPEARED\` (the OS's more aggressive scan did find the device)
4. \`BleCompanionService.handleDeviceAppeared\` → \`startService(\"CompanionSvc.deviceAppeared\")\`
5. Service already running → \`manageDevice(addr)\`
6. \`manageDevice\` sees \`existing.currentGattHash != null\` → **returns without doing anything**

The most reliable \"device is here\" signal the OS gives us was being silently dropped.

### Android fix

Mirror Fieldy's pattern: when CDM triggers \`startService\`, bypass \`manageDevice\`'s guards and route to a new \`forceReconnect\` method that cancels the pending retry, closes the stuck GATT, and starts a fresh \`connectToDevice\`. Non-CompanionSvc callers (Dart \`manageDevice\`, BT toggle handler) still go through \`manageDevice\` as before. Rate limiting (15s) preserved.

## iOS root cause

\`OmiBleManager.reconnectStalePeripherals()\` is called on \`applicationWillEnterForeground\` and re-issues \`centralManager.connect(peripheral, options: nil)\` on previously-connected peripherals that aren't currently connected. But it skipped peripherals in \`.connecting\` state:

\`\`\`swift
switch peripheral.state {
case .connected, .connecting:
    continue   // <-- skip both
\`\`\`

A peripheral stuck in \`.connecting\` (iOS equivalent of Android's stuck \`autoConnect=true\`) got no kick on foreground.

### iOS fix

Match Fieldy's \`handleWillEnterForeground\`: re-issue \`connect()\` for any non-\`.connected\` peripheral. iOS's \`connect()\` is idempotent — safe for already-connecting peripherals, and it nudges CoreBluetooth to retry for truly stuck ones.

## Files

[...]

PR #10908, opened 2026-07-30
fix(android): re-emit device-ready when Background Mode's BLE service outlives the Flutter engine

### LINKED ISSUE #10847: Bug: Android background mode shows device as disconnected even though status says connected, and transcription stops until manual reconnect
## Summary
When using Omi on Android in background mode, the device can appear disconnected in the app even though the device light is blue and the device status page says it is connected.

In that state, transcription does not work. The only way to recover is to disconnect the device manually and reconnect it.

## Environment
- Omi Android app
- Device in Background mode
- Observed on Wednesday, July 29, 2026

## Problem
The app and the device status become inconsistent:
- The app shows the device as disconnected
- The device LED remains blue, indicating an active connection
- If I tap to view device status, it shows as connected
- Despite that, transcription does not happen

## Expected behavior
- Background mode should maintain a stable connection state
- The app should show the real connection status consistently
- If the device is actually connected, transcription should continue normally
- If the connection is broken, the app should detect it correctly and recover or prompt the user

## Actual behavior
- In background mode, the device appears disconnected in the main app state
- The device status view shows it as connected
- The blue LED also suggests it is connected
- Transcription stops in this state
- To fix it, I have to disconnect manually and reconnect

## Impact
This makes background mode unreliable.

The user can think the device is still active because the LED is blue and the status page says connected, but transcription is silently not working.

## Reproduction
1. Use the Omi Android app with the device in Background mode
2. Let it run until the app shows the device as disconnected
3. Open device status and observe that it still shows connected
4. Note that the device LED is blue
5. Try using transcription and observe that it does not transcribe
6. Manually disconnect and reconnect the device
7. Transcription resumes only after reconnecting

## Suggested investigation
- Check connection-state sync between the Android app UI, device status page, and actual device session
- Check whether background mode leaves the app in a stale or partially disconnected state
- Check whether transcription service/session is dead even while the BLE or device connection remains active
- Add recovery logic when the connection appears alive but transcription is no longer functioning
- Add a user-visible warning when transcription is not working despite an apparent connection

## Notes
The main issue is not just disconnecting. It is that the app reports conflicting connection states and fails silently until the user manually reconnects.

mdmohsin7:
Fixes #10847.

## Symptom

On Android with Background Mode on, the app shows the device as disconnected while the
LED is blue, the foreground-service notification says "Connected to Omi", and the
diagnostics page still counts connection uptime. Transcription stops silently. The only
recovery is a manual disconnect + reconnect.

## Root cause

Background Mode's whole point is that `OmiBleForegroundService` — and the GATT link it
owns — outlive the Flutter engine. When the user reopens the app, a fresh engine and a
fresh Dart isolate start at `isConnected = false` and call `manageDevice` to re-establish
their side of a link that is already up.

`startService()` routes that call to `forceReconnect()` when the service instance already
exists, and `forceReconnect()` returned immediately if the peripheral was connected. The
"already connected → re-emit ready" resync (`notifyReadyForConnectedGatt`) lives only in
`manageDevice()`, which that branch bypasses. `onDeviceReady` is the only signal Dart has
for "connected", so `NativeBleTransport.connect()` waited out its 60s timeout and settled
on `disconnected` — permanently, since nothing reconciles Dart state against native.

This is a regression from #6987, which replaced `inst.manageDevice(...)` with
`inst.forceReconnect(...)` to break stuck `autoConnect` passive scans. The already-connected
case was never the stuck case, and it lost its resync in the swap.

Two adjacent defects turn "wrong badge" into "silent loss of transcription":

- `ensureBackgroundAudioSubscription()` returned early whenever `isFlutterAlive` was true,
  deferring the audio CCCD subscribe to Dart. With Dart stuck disconnected, any subsequent
  BLE drop/reconnect left the audio characteristic subscribed by nobody — link up, zero
  packets.
- `isFlutterAlive` was cleared only when the activity was *finishing*. A system-initiated
  destroy (memory pressure while the foreground service pins the process) left it `true`
  pointing at a dead binary messenger, alongside a stale `nativeBleForegroundReady` that
  makes `OmiBackgroundAudioStreamer` bail with `foreground_ready`.

## Change

- `startService` → `onConnectRequest`, which adopts the live link and re-emits ready when
  the peripheral is connected, and otherwise forces a fresh attempt exactly as before.
  `manageDevice` shares the same `adoptConnectedDevice` helper. The decision is a pure
  policy function (`BleConnectRouter.kt`) so the contract is testable off-device.
- `adoptConnectedDevice` deliberately does not touch the managed-device or
  user-disconnected prefs, so a CompanionDeviceManager appear event can never resurrect a
  device the user explicitly disconnected.
- Native always ensures the audio subscription; enabling notifications is idempotent.
- `MainActivity.onDestroy` clears `isFlutterAlive` and `nativeBleForegroundReady` whether
  or not the activity is finishing. `configureFlutterEngine` re-arms both on attach.
- Dart: a guard test pinning that a device-ready never replaces stream controllers that
  still have subscribers. The native resync makes a repeated `onDeviceReady` routinely
  reachable, and replacing live controllers orphans every subscriber — the same silent
  audio stop in a new shape. `main` landed the controller-preservation branch itself while
  this PR was open (b0e2b2bd6f, d5b35a414b), so this commit is now the test plus a
  `hostApi` seam that lets the transport be driven without a platform channel.

## Verification

```
cd app/android && ./gradlew :app:testDevDebugUnitTest      # 15 tests, 0 failures (3 new)
cd app && flutter test test/unit/native_ble_transport_resync_test.dart   # 2 passed
cd app && flutter analyze <changed dart files>             # No issues found
cd app && flutter build apk --flavor dev --debug           # built
cd app && ./test.sh                                        # 956 passed, 1 failed
```

The new Dart test is a real guard: removing the controller-preservation branch makes it
fail with `Expected: [[1, 2, 3]] Actual: []`.

The single `./test.sh` failure is
`conversation_detail_provider_selection_test.dart: conversation getter resolves when
startedAt and createdAt fall on different days`, a date-boundary test unrelated to this
change. Confirmed pre-existing: it fails identically on `origin/main` with these commits
stashed.

Not verified by me: the on-device repro (Background Mode on, close the app, reopen, confirm
the app adopts the live link and resumes transcription) needs the pendant. `BleConnectRouterTest`
is a contract test on new code rather than a regression test, and Android unit tests are not
currently run by any CI workflow — a pre-existing gap, worth wiring into
`mobile-app-checks.yml` in its own change.

Failure-Class: new

---

Rebased on `main`, and re-verified there: `./gradlew :app:testDevDebugUnitTest`,
`flutter test test/unit/native_ble_transport_resync_test.dart`, `flutter analyze`, and
`flutter build apk --flavor dev --debug` all pass on the rebased branch.

The branch originally carried a fourth commit fixing `${resolved,,}` (a bash 4+ expansion)
in `scripts/dev-harness/_resolve_python.sh`, which crashed `pre-push-singleflight` on macOS
bash 3.2. `main` fixed the same defect in 5c8f8f62e4 / cd997c08c3, so that commit is dropped.

### DIFF
--- a/.github/failure-classes/FC-already-satisfied-request-as-noop.json
+++ b/.github/failure-classes/FC-already-satisfied-request-as-noop.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": 1,
+  "id": "FC-already-satisfied-request-as-noop",
+  "violated_contract": "A request to establish a session must still deliver the observation its requester depends on when the target is already in the desired state. A native owner that outlives its client — a foreground service holding a BLE link across a Flutter engine restart — makes \"already satisfied\" the ordinary case, and answering it with a bare early return strands the client projection with no path back to the truth.",
+  "canonical_prevention": "Route every establish request through one policy that separates not-established, already-established, and in-progress, and make the already-established branch re-emit the readiness event the client consumes instead of returning. Prove the policy through its own seam rather than by driving the platform service.",
+  "canonical_prevention_artifact": [
+    "app/android/app/src/main/kotlin/com/friend/ios/ble/BleConnectRouter.kt",
+    "app/android/app/src/test/kotlin/com/friend/ios/ble/BleConnectRouterTest.kt"
+  ],
+  "evidence_prs": [6987],
+  "scope_hints": [
+    "app/android/app/src/main/kotlin/com/friend/ios/ble/**",
+    "app/lib/services/devices/transports/**"
+  ],
+  "status": "open"
+}
--- a/app/android/app/src/main/kotlin/com/example/my_project/MainActivity.kt
+++ b/app/android/app/src/main/kotlin/com/example/my_project/MainActivity.kt
@@ -100,8 +100,16 @@ class MainActivity: FlutterActivity() {
     }
 
     override fun onDestroy() {
+        // The engine dies with the activity whether or not it is finishing, so these flags
+        // must clear outside the isFinishing guard — a system-initiated destroy otherwise
+        // leaves native deferring audio to an engine that is gone (issue #10847).
+        // configureFlutterEngine re-arms both on the next attach.
+        OmiBleManager.isFlutterAlive = false
+        getSharedPreferences("FlutterSharedPreferences", MODE_PRIVATE)
+            .edit()
+            .putBoolean("flutter.nativeBleForegroundReady", false)
+            .apply()
         if (isFinishing) {
-            OmiBleManager.isFlutterAlive = false
             // Engine + main isolate die with the activity; a live capture session must not outlive its consumer.
             if (PhoneMicController.isInitialized) PhoneMicController.instance.onFlutterEngineDestroyed()
             // Background Mode and Transcribe Later both need the foreground service to keep
--- a/app/android/app/src/main/kotlin/com/friend/ios/ble/BleConnectRouter.kt
+++ b/app/android/app/src/main/kotlin/com/friend/ios/ble/BleConnectRouter.kt
@@ -1,0 +1,30 @@
+package com.friend.ios.ble
+
+/**
+ * What a "please connect this device" request must do, given the current native link state.
+ *
+ * Requests arrive from Dart (`manageDevice` on app start / user reconnect), from
+ * CompanionDeviceManager appear events, and from the sticky-service restore after a
+ * process kill. Background Mode keeps OmiBleForegroundService — and the GATT link —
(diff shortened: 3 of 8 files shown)


kodjima33 (approved):
Android bug fix (5/5): re-emits device-ready when Background Mode's BLE service outlives the Flutter engine (#10847); clear root cause + unit tests. Not merged — branch has conflicts with main.

PR #10926, opened 2026-07-31
fix(desktop): restore byte-exact test-install contract fixture

kodjima33:
## Bug

`Desktop Swift CI` is red on `main` for every desktop-touching commit since b8a07d66b5 — both required jobs fail:

- `Desktop Swift Static & Test Contracts` → `desktop launcher script tests` → `tests/test-test-install-workflow-contract.sh` → `FAIL: admitted canonical test-install block must occur exactly once`
- `Desktop Swift Build & Tests` → `Require static contracts and Swift tests` (mirrors the verdict above)

Red runs: https://github.com/BasedHardware/omi/actions/runs/30601451952 (be04764efc) and https://github.com/BasedHardware/omi/actions/runs/30603540493 (90e06ae0a5, the v0.12.148 candidate).

## Root cause

b8a07d66b5 ("fix: update test-install workflow fixture + advance release path") rewrote `desktop/macos/tests/fixtures/test-install-job-contract.yml` through a YAML round-trip: it added a `---` document marker, dedented the block to 2-space indentation, stripped the blank lines between steps, and quoted `${{ github.token }}`.

That fixture backs a **byte-exact** contract. `tests/test-test-install-workflow-contract.sh` asserts `workflow_bytes.count(contract_bytes) == 1` and that the extracted `jobs.test-install` block byte-matches the fixture — the point of d2f60932b1 ("fix(release): lock installer test job bytes") is that no reserialization is admissible. After the round-trip the fixture occurs **zero** times in `desktop/macos/.github/workflows/test-install.yml`, so the check fails closed.

The workflow itself was never stale: the pre-b8a07d66b5 fixture still matches it exactly (`workflow_bytes.count(old_fixture) == 1`).

## Fix

Restore the pre-b8a07d66b5 fixture bytes verbatim. One file, fixture-only; the workflow is untouched.

Ignoring whitespace the diff is **+10/-2** (the `---` marker, the blank lines, the token quoting); the 388-line raw count is purely the reindentation artifact.

## Verification

```
# on origin/main (90e06ae0a5) — reproduces the exact CI message
$ bash desktop/macos/tests/test-test-install-workflow-contract.sh
FAIL: admitted canonical test-install block must occur exactly once

# on this branch — the same script CI discovers, full mutation suite
$ bash desktop/macos/tests/test-test-install-workflow-contract.sh; echo EXIT=$?
test-install workflow exact-DMG contract passed
EXIT=0
```

All 90 `run_mutation` cases still behave as specified — every hostile mutation is rejected, and the `canonical` / `unrelated-job-defaults-shell` benign cases are still accepted. `make preflight` passes.

none

## Failure class (fixes)

Failure-Class: none

Same declaration as d2f60932b1, the commit that established this byte-lock contract.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/tests/fixtures/test-install-job-contract.yml
+++ b/desktop/macos/tests/fixtures/test-install-job-contract.yml
@@ -1,190 +1,198 @@
----
-test-install:
-  runs-on: macos-15
-  timeout-minutes: 15
-  steps:
-  - name: System Info
-    run: |
-      echo "## System Information" >> $GITHUB_STEP_SUMMARY
-      echo "| Property | Value |" >> $GITHUB_STEP_SUMMARY
-      echo "|----------|-------|" >> $GITHUB_STEP_SUMMARY
-      echo "| macOS Version | $(sw_vers -productVersion) |" >> $GITHUB_STEP_SUMMARY
-      echo "| Build | $(sw_vers -buildVersion) |" >> $GITHUB_STEP_SUMMARY
-      echo "| Architecture | $(uname -m) |" >> $GITHUB_STEP_SUMMARY
-      echo "| Gatekeeper | $(spctl --status 2>&1 || echo 'unknown') |" >> $GITHUB_STEP_SUMMARY
-      echo "" >> $GITHUB_STEP_SUMMARY
-  - name: Download DMG from GitHub Release
-    id: download
-    env:
-      GH_TOKEN: "${{ github.token }}"
-    run: |
-      echo "=== Downloading DMG ==="
+  test-install:
+    runs-on: macos-15
+    timeout-minutes: 15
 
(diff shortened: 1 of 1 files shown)


undivisible (approved):
Verified the restored fixture is byte-identical to the prior canonical block and the full contract mutation suite passes.

PR #9412, opened 2026-07-10
Add bounded two-lane offline sync recovery

Git-on-my-level:
## Summary

- split offline sync into a trusted fresh lane and an isolated historical backfill lane
- cap backfill at one in-flight job per user, 4 processed speech hours per user per UTC day, 555 processed speech hours globally per UTC day, and 4 globally concurrent jobs
- preserve live-session economics by excluding backfill usage from realtime hard restrictions
- add signed fresh-capture manifests, durable idempotency and partial-result checkpoints, newest-first mobile scheduling, and lane-specific cooldowns
- provision the backfill Cloud Run service/queue, 70% and 90% routed alerts, Firestore TTL, runtime validation, generated API contracts, and an operator runbook

No shadow phase is included; the caps are enforced from launch.

## Verification

- `cd app && ./test.sh` — 732 passed
- relevant backend sync/fair-use suites — 405 passed
- `cd backend && uv run --python 3.11 bash scripts/pre-deploy-check.sh` — 40 passed
- full repository `./scripts/pre-push` — passed, including pyright, workflow/runtime contracts, OpenAPI/generated clients, backend tests, formatters, and macOS compile
- independent Sol review of the exact publishable commit — APPROVED

## Rollout requirement

Production deployment requires `SYNC_BACKFILL_ALERT_NOTIFICATION_CHANNELS` to contain comma-separated Cloud Monitoring notification-channel resource names.

Closes #9381

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10932, opened 2026-07-31
fix(app): offer sync when the lane the backlog needs is free

mdmohsin7:
## What changed and why

Reported: an account showing a backlog of recordings "waiting to sync", with no Sync button and no automatic drain. The backend shows nothing wrong for that account because the app never calls it — the block is entirely client-side.

Two layers disagree about what "rate limited" means:

- **Uploads** gate per lane — `SyncRateLimiter.isLimitedForLane(lane)`. A `fairUse` cooldown does **not** block the `backfill` lane.
- **The sync pages** gated on the account-wide `SyncRateLimiter.isLimited`, true when *any* lane is in cooldown.

So a fair-use cooldown hid the Sync control for exactly the files that upload through backfill — the app refusing a sync its own upload path would have accepted. It disappears completely rather than being disabled, because the card's rate-limited branch is evaluated before the two branches that render the Sync pill (`attention > 0`, `readyToBackUp > 0`).

The account-wide getter was correct when written — there was a single cooldown then, so global *was* per-lane. It became wrong when #9412 split the lanes underneath it without moving the callers.

### Changes

- **`pendingSyncUploadLanes(pending, now)`** — the lanes the pending recordings would actually use, from the existing rule (server capture proof plus the 6h fresh cutoff). No new policy.
- **`isRateLimitedForPendingUploads`** — true only when *every* lane the pending work needs is in cooldown. Both sync pages read this instead of the account-wide flag.
- **`prepareToUpload` reconciles on every lane**, not only `fresh`. This is the second half of the trap: a persisted fair-use cooldown does not gate backfill, but nothing else clears it, so a user whose whole backlog is backfill never reached the self-heal and a cooldown the server had already lifted could outlive it locally for up to 30 days.

Informative titles are unchanged — when the needed lanes really are limited, the card still explains why.

none

## How it was verified

- **`app/test.sh` — 972 tests pass.**
- **`test/unit/sync_lane_aware_gating_test.dart` (17 new).** Reverting just the reconciliation change turns the file red (`+13 -1`), so it fails against the pre-fix code rather than restating the new code. Covers lane classification (fresh vs backfill, mixed, empty), the all-lanes-limited truth table, and the UI-gate/upload-gate contract across all three cooldown kinds (`fairUse`, `backfill`, `backendBusy`).
- The page wiring is asserted by reading the two page sources. That is a static tripwire, not behavioural coverage, and is labelled as such in the file — the widget tree is not exercised, so it only pins which getter the pages call.

**Not verified against the reporting device.** This fixes a defect that produces exactly the reported symptoms, but two other branches render an actionless card as well — a stuck `isSyncing` (spinner + Cancel) and `uploaded > 0` ("Processing"). Distinguishing them needs a screenshot of the sync card or app logs. If that device turns out to be in one of those branches, this PR is not its fix.

## Tests

`app/test/unit/sync_lane_aware_gating_test.dart`.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10932?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/conversations/auto_sync_page.dart
+++ b/app/lib/pages/conversations/auto_sync_page.dart
@@ -143,7 +143,7 @@ class _AutoSyncPageState extends State<AutoSyncPage> {
     final hasAnyRecording = p.allWals.isNotEmpty;
 
     final isActive = s.isSyncing || s.isFetchingConversations;
-    final bool showSpinner = (isActive || uploaded > 0) && !p.isRateLimited;
+    final bool showSpinner = (isActive || uploaded > 0) && !p.isRateLimitedForPendingUploads;
 
     String title;
     String? progressText;
@@ -175,7 +175,7 @@ class _AutoSyncPageState extends State<AutoSyncPage> {
           title = s.isFetchingConversations ? l.syncCardProcessing : l.syncCardUploadingTitle;
       }
       action = _statusActionPill(l.cancel, Colors.redAccent, () => _confirmCancel(context, p));
-    } else if (p.isRateLimited) {
+    } else if (p.isRateLimitedForPendingUploads) {
       title = switch (p.rateLimitReason) {
         RateLimitReason.backendBusy => l.syncCardBackendBusy,
         RateLimitReason.backfillPaced => l.syncCardReadyCount(readyToBackUp),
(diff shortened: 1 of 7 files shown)


kodjima33 (approved):
Mobile bug fix (5/5): sync UI gated on the account-wide rate limit while uploads gate per lane, hiding Sync for a backlog that would have uploaded; lane-aware gate + unit test. CI green.

PR #10934, opened 2026-07-31
feat(llm-gateway): consolidate safe auto-router

undivisible:
## Summary

- Enforces that every prod-ready catalog entry matches its serving surface and active provider/model, with a real eval suite.
- Runs the capability smoke through the configured resolver and executor by default; synthetic mode is test-only.
- Removes inactive benchmark proposal tooling that could not emit a valid serving change.
- Hot-reloads the gateway config on YAML mtime change, including catalog demotions.
- Emits shared fallback telemetry when a failed reload falls back to the previously loaded config.

## Merge with main

The lane catalog and main's model_config-derived lanes conflicted. Resolved in favour of
main's shipped behaviour: `utils.llm.model_config` stays the source of truth for the feature
lanes it generates, and the catalog governs the hand-authored serving YAML only. Gating
generated lanes on the catalog would have cut the served set from 42 lanes to 2.
`GatewayConfig.generated_lane_ids` records which lanes are generated so
`validate_serving_config` skips them; catalog demotion still narrows the hand-authored
serving set, which is the safety property this PR adds.

## Validation

- `backend/.venv/bin/python -m pytest tests/unit/llm/test_lane_catalog.py tests/unit/llm_gateway/ tests/unit/test_llm_gateway_config.py tests/unit/test_llm_gateway_readiness.py tests/unit/test_llm_gateway_coverage_guardrails.py tests/unit/test_llm_gateway_anthropic_passthrough.py tests/unit/test_llm_gateway_output_budget.py -q` (164 passed)
- `backend/.venv/bin/python scripts/scan_async_blockers.py --dirs backend/llm_gateway`
- `make preflight`

- INV-INT-1

## Failure class (fixes)

Failure-Class: none


Git-on-my-level (commented):
Thanks for consolidating this behind the LLM gateway. The overall direction looks useful: the catalog/serving split, fake-default capability smoke, benchmark-to-proposed-artifact emitter, and mtime-based config reload all help keep the auto-router rollout controlled rather than widening production routing accidentally.

I’m leaving this as maintainer-review feedback rather than approval because the PR is still draft, backend checks are currently failing, and this is product/architecture-sensitive LLM routing work that should get human maintainer sign-off before merge.

A few review notes for the next pass:

- Please get the backend gate green before marking ready. GitHub currently reports failures in Backend unit suite, Hermetic Backend E2E, and the Backend Hermetic Merge Gate. The E2E failure is in the listen/STT language path, so it may be pre-existing or unrelated, but it still needs either a green rerun or a clear note explaining why it is not caused by this head.
- The serving/catalog boundary is the right shape for safe rollout. Before merge, I’d like a maintainer to confirm the exact prod-ready lane set and the behavior where adding/flipping a catalog entry can change the resolver allowlist after config reload.
- The new AA benchmark fetch path is appropriately optional/fallback-based from a quick static review. Please keep validating that no benchmark fetch/cache path can mutate the live serving artifacts directly; the proposed-artifact-only contract is important here.
- Since this changes model-routing infrastructure, please keep the PR validation focused on the real gateway path plus the emitter/smoke path. The added unit coverage is good, but human review should still confirm the rollout policy and operational ownership.

No urgent security/supply-chain issue found in this pass.

---

Git-on-my-level (dismissed):
Thanks for the continued work here. I found one concrete blocker in the benchmark refresh path that should be fixed before this is considered ready:

- `BenchmarksFetcher` defaults its fallback file to `backend/llm_gateway/gateway/_private/benchmarks.example.json`, but this PR does not add that file. With `AA_API_KEY` unset, `BenchmarksFetcher().fetch()` logs the intended fallback and then raises `FileNotFoundError` instead of returning example data. The same missing file also breaks the successful-AA path when it calls `_merge_with_uncovered_tasks(..., self._load_example())`, so STT/embedding preservation depends on a file that is not present. I reproduced this on the PR head by directly exercising the fetcher fallback path; it tried to open `/workspace/omi-pr-10934-review/backend/llm_gateway/gateway/_private/benchmarks.example.json`.

Please either commit the fallback fixture at the path the fetcher uses, point the fetcher at a real committed fallback payload, or make the scheduled/emitter path supply the fallback explicitly, and add a behavioral test that calls `fetch()` with no `AA_API_KEY` plus the AA-success merge path. The intended fallback behavior is important because this tooling is supposed to be safe/offline by default rather than failing the scheduled rotation path when the external benchmark key is absent or unavailable.

I am not approving this because the PR is still draft, checks are failing, and this model-routing/rollout path still needs human maintainer review before merge. No urgent security issue found in this pass.

---

Git-on-my-level (dismissed):
Thanks for the update here. The previous benchmark fallback blocker appears to be resolved on this head because the inactive benchmark proposal/fetcher tooling is no longer present in the final diff.

I found one new blocker in the config reload path that should be fixed before this is ready:

- `GatewayConfigReloader._load()` passes `required_lane_ids=get_supported_auto_lane_ids()` before `load_gateway_config()` reads the changed catalog file. Since the resolver catalog is process-cached, a running gateway can keep using the old prod-ready set when `lanes_catalog.yaml` changes. I reproduced this by priming the resolver catalog, demoting `omi:auto:public-shared-conversation-chat` in a copied catalog, and calling `GatewayConfigReloader.get()` again: the reload returned the stale cached config and kept `omi:auto:public-shared-conversation-chat` in `cfg.lanes` instead of narrowing the serving set. That means the intended “next request after merge sees the new artifact/catalog” behavior can fail specifically for catalog demotions/removals, leaving a lane servable until process restart.

Please make the reload path derive required lane IDs from the same freshly loaded catalog/config snapshot (or remove the stale resolver-derived requirement) and add a regression test that changes a prod-ready catalog entry to non-prod-ready and verifies the hot-reloaded config narrows the served lane set rather than returning stale.

I’m requesting changes rather than approving because the PR is still draft, the Hygiene check is failing, and this model-routing/config-reload behavior needs human maintainer sign-off before merge. No urgent security issue found.

---

Git-on-my-level (commented):
Thanks for the update. I re-reviewed the current head and the prior config hot-reload blocker looks resolved: `GatewayConfigReloader._load()` now loads the changed catalog first, passes that fresh catalog into `load_gateway_config()`, and only then updates the resolver catalog. The new regression `test_reloader_removes_demoted_catalog_lane` covers the demotion/removal case that previously kept `omi:auto:public-shared-conversation-chat` servable until restart.

Validation I ran locally on this head:

- `.venv/bin/python -m pytest tests/unit/llm/test_lane_catalog.py tests/unit/llm_gateway/test_capability_smoke.py tests/unit/llm_gateway/test_config_reload.py tests/unit/llm_gateway/test_daily_refresh.py -q` — 110 passed
- `.venv/bin/python scripts/scan_async_blockers.py --dirs llm_gateway` — 0 selected blocking findings

I don't see a remaining automation blocker in this diff, and the category remains backend. I’m not approving because this is still a draft and it changes LLM gateway routing/config reload/smoke behavior, so it still needs human maintainer sign-off before merge. No urgent security issue found.

---

Git-on-my-level (commented):
Thanks for the latest update. I re-checked this head against the earlier automation feedback:

- The previous benchmark fallback and catalog hot-reload blockers appear resolved in the current diff; the inactive benchmark proposal/fetcher path is gone, and the reloader now loads the changed catalog before refreshing the gateway config/resolver catalog.
- GitHub currently reports the relevant backend and hygiene checks passing on this head, which is good progress.
- I still would not merge this without human maintainer sign-off: it is still a draft, currently carries `needs-rebase`, and it changes LLM gateway routing/config reload/capability-smoke behavior. That rollout/operational boundary is a maintainer judgment call even though I do not see a new concrete automation blocker in this pass.

No urgent security/supply-chain issue found.

---

undivisible:
Rebased onto current `main` and addressed the open items.

**The `needs-rebase` conflict changed this PR's core policy — please look here first.**

`main` now generates a gateway lane for every `utils.llm.model_config` feature (42 served lanes, incl. the chat-agent Luna migration and the translation lane). This branch gated serving on `prod_ready` catalog entries, which on the merged tree would have cut the served set to **2 lanes** and silently stopped serving the other 40.

Resolved in favour of `main`'s shipped behaviour: `model_config` stays the source of truth for the lanes it generates, and the catalog governs the hand-authored serving YAML (`lanes.yaml` + `route_artifacts.yaml`) only. `GatewayConfig.generated_lane_ids` records which lanes are generated so `validate_serving_config` skips them. Catalog demotion still narrows the hand-authored serving set, so the hot-reload guarantee you verified is intact.

This means the "narrow the prod-ready lane set" outcome is **not** in this PR. That is the maintainer call you flagged; happy to do it as a follow-up once the intended prod-ready set is decided.

Also fixed:

- `get_gateway_config` became a coroutine on this branch but `main.lifespan` still called it synchronously, so startup raised `AttributeError: 'coroutine' object has no attribute 'lanes'` and aborted the lifespan. Now awaited, with the lifespan test updated.
- The reloader's stale-config fallback was silent. It now calls the shared `record_fallback` helper (`component=llm_gateway`, `to_mode=stale_config`, `outcome=degraded`) per `docs/agents/fallback-telemetry.md`, with a regression test.

Verification on this head:

- `pytest tests/unit/llm/ tests/unit/llm_gateway/ tests/unit/test_llm_gateway_*.py tests/unit/test_verify_llm_gateway_serving.py -q` → 320 passed
- `make preflight` → 23 checks passed
- `scripts/scan_async_blockers.py --dirs llm_gateway` → 0 blocking findings

Leaving as draft pending your call on the lane-set question above.

Git-on-my-level:
Thanks for the continued iteration here. I re-reviewed the current head and this is moving in a good direction, but I’m still leaving it for human maintainer review rather than approval because it is still a draft and it changes the LLM gateway’s routing/config reload behavior.

- `backend/llm_gateway/gateway/config_loader.py` now keeps `utils.llm.model_config` generated lanes as their own source of truth while applying the new catalog only to hand-authored serving YAML. The explicit `OMI_LLM_GATEWAY_SERVING_LANES` / `OMI_LLM_GATEWAY_SERVING_LANES_DROP_CONFIRM` guard is a good fail-closed shape for intentional lane narrowing.
- `backend/llm_gateway/gateway/lane_catalog.py` validates prod-ready catalog entries against serving config, provider/model, and route artifacts, and `backend/llm_gateway/config/lanes_catalog.yaml` keeps non-evaluated lanes catalog-only. That addresses the earlier concern about placeholder or dev-only lanes becoming silently servable.
- `backend/llm_gateway/gateway/config_reload.py` refreshes the catalog before rebuilding gateway config and then updates the resolver catalog, so catalog demotions should take effect without a restart; the new reload tests cover the demotion path.
- `backend/llm_gateway/scripts/capability_smoke.py` plus `backend/llm_gateway/scripts/deterministic_provider.py` keep the default CI/test smoke deterministic via `--provider fake`, while the real provider remains explicitly selectable for operator validation.
- `backend/llm_gateway/main.py`, `backend/llm_gateway/routers/dependencies.py`, and `backend/llm_gateway/routers/health.py` correctly await the hot-reloaded config path, and the OpenAI/Anthropic routers continue to use FastAPI’s async dependency injection for request handling.
- `docs/doc/developer/backend/llm_gateway.mdx` adds operator-facing guidance for the serving-lane allowlist. Because this documentation can influence future AI coding/review agents and operators working in this repo, the maintainer consequence is that agents may now treat lane narrowing as an explicit, confirmation-gated operational mode. The guidance looks safe and aligned with the implementation, but it still deserves maintainer sign-off because it documents production LLM-gateway behavior.

Validation I ran locally on this head:
- Focused backend gateway suite: 96 passed.
- Deterministic capability smoke with the fake provider: 2 lanes passed, 0 failed.

No blocking implementation issue found in this pass. Remaining merge blockers are qualitative/process: the PR is draft, currently carries rebase/workflow/human-review labels, and changes a security/operations-sensitive model-routing surface. Leaving this for human maintainer review of the LLM-gateway rollout direction and operational risk.

— Reviewed by glm-5.2 for Omi maintainer automation.

AI disclosure: this feedback was generated by Omi’s AI maintainer-review automation.

---

PR #10936, opened 2026-07-31
fix(stt): keep Velma off undecoded PCM codecs and log frame shape on rejection

mdmohsin7:
Velma is not served client codecs whose bytes reach the provider without a decoder in between, and a Velma rejection now logs the frame geometry it has to be explained by.

## Why

Velma-first was tried in prod on 07-31 and rolled back. Errors were confined to one code path:

| client / codec | reconnects per uid, Deepgram-first | reconnects per uid, Velma-first | change |
|---|---|---|---|
| `desktop` / `pcm16` | 2.00 | 9.17 | **4.58x** |
| `omi` / `opus_fs320` | 9.85 | 6.46 | 0.66x |
| `omi` / `opus` | 10.00 | 8.38 | 0.84x |

Two equal 10-minute windows. 80% of traffic is opus and did not move; only the raw-PCM path amplified. Velma was serving virtually everything in that window (1269 Modulate connections against 3 Parakeet), so this is not a routing artifact.

`pcm16`/`linear16` are the only codecs the listen receiver forwards verbatim — `decoded: bytes = data`. Every other codec passes through opus/aac/lc3 decoding or `audioop.lin2lin`, all of which emit whole 16-bit samples by construction.

## What this does NOT claim

**The mechanism is not proven.** The obvious candidate — frames that are not a whole number of samples — is ruled out as the production cause: `8fba206` shipped the alignment fix from #10558 to prod on 07-25, and pods running it produced 1,894 `Invalid input audio` errors while `omi_live_stt_misaligned_frames_total` never materialised at all. Those pods were scraped throughout; sibling metrics are present. Zero misaligned frames, ~1,900 rejections.

So the correlation above is solid and the mechanism behind it is open. This change is scoped to match: it removes the exposure and adds the measurement, rather than asserting a cause.

A probe against the live API (prod's exact connection parameters, 5 trials per case) does confirm what the error responds to in isolation — odd byte counts and nothing else:

```
ok    baseline even 300ms          ok    single 2-byte frame
FAIL  odd-length frame (961B)      ok    zero-filled silence 300ms
FAIL  odd tail after valid audio   ok    very large frame 640KB
ok    960B frames x10              ok    no language param / empty EOS
```

That is a property of the provider, not evidence about our traffic.

## Change

`modulate_supports_codec()` in the provider policy, consulted in `get_stt_service_for_language` alongside the existing language and surface checks. `preferred_service` cannot bypass it — the gate sits inside the same branch that selects Velma. Sessions on excluded codecs fall through to the next model in the list exactly as an unsupported language already does.

`SafeModulateSocket` tallies frame geometry and emits it with the rejection:

```
Modulate streaming error: Invalid input audio codec=pcm16 sample_rate=16000 language=en
  frames=412 bytes=395520 odd_frames=0 min_frame=640 max_frame=1280 last_frame=960
```

`odd_frames` is counted on the frame as received, before alignment, so it stays truthful even though the socket then corrects it. This is the field that distinguishes "we sent a bad frame" from "the provider rejected a well-formed stream" — the question three attempts have not been able to answer.

## Scope

Streaming only. `chat.py` (PTT) also calls `process_audio_modulate` and is left on its current behaviour; the codec parameter defaults to `None`, which is not excluded, so every other caller is unchanged.

The gate is deliberately a list of two codecs rather than an allowlist, so a new device codec is not silently denied Velma.

## Tests

`test_modulate_codec_gate.py`: decoded codecs stay eligible; `pcm16`/`linear16` are excluded including case and whitespace variants; absent codec is not excluded; selection routes a decoded codec to Velma and skips it for raw PCM; `preferred_service='modulate'` cannot bypass the gate; the shape log reports frame geometry and counts odd frames.

238 passed across the Modulate, live-STT-failure, Parakeet fallback, clean-close latch, Deepgram-serving and prerecorded-coverage suites. CI lane: 10/10 checks pass.

Note: a broad `pytest -k` collection segfaults on this machine inside `google.cloud.firestore` protobuf imports. It reproduces on unmodified `main` with this branch's test file excluded, so it is environmental; suites were run as an explicit file list.

## Rollout

This only takes effect where Velma is actually first. Prod `backend-listen` is currently `dg-nova-3,modulate-velma-2,parakeet`, so merging changes nothing until that ordering is moved back — at which point desktop and phone raw-PCM sessions stay on Deepgram and everything else goes to Velma.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/config/stt_provider_policy.py
+++ b/backend/config/stt_provider_policy.py
@@ -94,6 +94,12 @@ MODULATE_SUPPORTED_LANGUAGES: Final[frozenset[str]] = frozenset(
     }
 )
 
+# Client codecs whose bytes reach the provider without a decoder in between, so
+# nothing guarantees each frame is a whole number of 16-bit samples. Velma closes
+# the stream on any frame that is not, and live sessions have no in-session
+# failover, so it is not served these codecs until the cause is confirmed.
+MODULATE_EXCLUDED_CODECS: Final[frozenset[str]] = frozenset({'pcm16', 'linear16'})
+
 # This is the single source of truth for provider enablement. Cloud Deepgram is
 # intentionally absent from every serving surface. Self-hosted Deepgram is a
 # distinct product and remains available only to the streaming runtime that has
@@ -197,6 +203,11 @@ def normalized_stt_language(language: str | None) -> str:
 def modulate_supports_language(language: str | None) -> bool:
     """Return whether Velma-2 accepts a language code on a serving surface."""
     return normalized_stt_language(language) in MODULATE_SUPPORTED_LANGUAGES
+
+
+def modulate_supports_codec(codec: str | None) -> bool:
+    """Return whether a client codec may be served by Velma-2."""
+    return (codec or '').strip().lower() not in MODULATE_EXCLUDED_CODECS
 
 
 def supports_live_multilingual_mode(language: str | None) -> bool:
--- a/backend/routers/listen/receiver.py
+++ b/backend/routers/listen/receiver.py
@@ -165,6 +165,7 @@ class ListenReceiver:
                     modulate_callback or callback,
                     sample_rate,
                     self.host.stt_language,
+                    codec=self.host.request.codec,
                 ),
             )
             self.host.stt_service = actual_service
@@ -172,7 +173,12 @@ class ListenReceiver:
                 self.host.stt_model = 'velma-2'
             return socket
         if self.host.stt_service == STTService.modulate:
-            return await process_audio_modulate(modulate_callback or callback, sample_rate, self.host.stt_language)
+            return await process_audio_modulate(
+                modulate_callback or callback,
+                sample_rate,
+                self.host.stt_language,
+                codec=self.host.request.codec,
+            )
         if self.host.stt_service == STTService.deepgram:
             return await process_audio_dg(
                 callback,
--- a/backend/routers/listen/runtime.py
+++ b/backend/routers/listen/runtime.py
@@ -268,6 +268,7 @@ class ListenSessionRuntime:
             self.language,
             multi_lang_enabled=not single_language_mode,
             preferred_service=request.stt_service,
+            codec=request.codec,
         )
         self.parity_capture = ListenParityCapture.from_environ(
             principal_id=request.uid,
--- a/backend/tests/unit/test_modulate_codec_gate.py
+++ b/backend/tests/unit/test_modulate_codec_gate.py
@@ -1,0 +1,115 @@
+import asyncio
+
+import pytest
+
+from config.stt_provider_policy import STTServingSurface, modulate_supports_codec
+from utils.stt import streaming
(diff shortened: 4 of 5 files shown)


kodjima33:
Heads-up from the hourly bug watchdog: this looks like it takes desktop live transcription down under the **prod** provider config, and `Hermetic Backend E2E` has been red on every PR branched off main since it merged.

**Failing e2e (reproduced locally on `main`):**

```
FAILED testing/e2e/test_listen_stt.py::test_web_listen_custom_stt_multichannel_keeps_receiving_transcripts
AssertionError: websocket closed before expected message:
  {'type': 'websocket.close', 'code': 1008, 'reason': 'The language is not supported, en'}
```

**Why.** `modulate_supports_codec()` now removes `modulate-velma-2` from selection for `pcm16`/`linear16`. When it is the *only* provider that can serve the session, `get_stt_service_for_language()` returns `(None, None, None)` and `ListenSessionRuntime._bootstrap` (`backend/routers/listen/runtime.py:287`) closes the socket with 1008 instead of degrading.

Prod streaming config is exactly that shape — `STT_SERVICE_MODELS=dg-nova-3,dg-nova-2` with `DEEPGRAM_SELF_HOSTED_ENABLED=false`, so both `dg-*` entries are skipped and selection falls through to `DEFAULT_MODELS_BY_SURFACE[STREAMING] = ('modulate-velma-2', 'parakeet')`. Streaming Parakeet is the English-only `parakeet-rnnt-1.1b`, and live sessions request `multi` whenever multi-language mode is on (the default). Run against that config on current main:

```
codec=pcm16  lang=en multi=True  -> (None, None, None)      # socket closed, 1008
codec=pcm16  lang=en multi=False -> (parakeet, 'en', 'parakeet')
codec=pcm16  lang=es multi=True  -> (None, None, None)      # socket closed, 1008
codec=opus   lang=en multi=True  -> (modulate, 'multi', 'velma-2')
```

The macOS desktop connects with `codec=linear16` (`desktop/macos/Desktop/Sources/TranscriptionService.swift:112,375`), which is in `MODULATE_EXCLUDED_CODECS` — so on a prod deploy every desktop ambient session that is not English-with-single-language-mode gets 1008'd at bootstrap. Custom-STT sessions are closed too even though they never open a provider socket, which is what the e2e catches.

Suggested direction (not opened as a PR — the routing decision is yours): make the codec gate a *preference* rather than a hard exclusion, i.e. keep the Velma candidate as last resort when no other provider can serve the requested language, and/or skip the provider requirement entirely when `use_custom_stt` is set.

🤖 hourly watchdog — report only, no code change made here

PR #10956, opened 2026-08-01
fix(desktop): mute Floating Bar notification previews independently

### LINKED ISSUE #6765: Desktop: Add option to mute Floating Bar in-app notification previews (respect Focus/DND)
## Problem
On macOS Desktop, assistant notifications are shown twice:
1) as a Floating Bar in-app preview below the bar, and
2) as a native macOS notification.

The in-app Floating Bar preview is **not affected by macOS Focus/DND**, so users can't quickly silence it using system controls.

This also makes the Floating Bar appear to "randomly re-enable" itself when notifications arrive (it is shown temporarily to render the preview).

## Requested
Add a setting to control Floating Bar notification previews independently from native macOS notifications, e.g.
- Toggle: **Show notification previews in Floating Bar** (on/off)
- Optional: **Respect Focus/DND** (auto-mute in-bar previews when Focus is enabled)
- Optional: **Only show in-bar previews when Floating Bar is enabled**

## Expected behavior
Users should be able to keep native notifications (which respect Focus/DND and offer OS-level customization) while silencing Floating Bar previews without having to disable all notifications in-app each time.

## Notes
Implementation hint: `NotificationService.sendNotification(...)` currently calls `FloatingControlBarManager.shared.showNotification(...)` before sending the native notification.

Reported by: Discord user (alicja.szeptem) in #desktop-app

jnnabugwu:
## What changed and why

Floating Bar in-app notification previews had no independent off switch and
weren't affected by macOS Focus/DND, so proactive notifications showed
twice (in-bar preview + native banner) and the bar would temporarily
reappear just to render a preview users couldn't silence. Adds a
"Notification Previews" toggle (default on) so users can mute in-bar
previews while keeping native macOS notifications. Closes #6765.

## Out of scope

Real macOS Focus/DND auto-muting (the other optional item from #6765) is
deferred to a follow-up issue/PR, not included here. It requires
`INFocusStatusCenter`, which needs the `com.apple.developer.focus-status`
entitlement — an Apple-gated capability requiring a manual approval request
filed with Apple, not a self-service Xcode checkbox. This repo has no
existing usage of it and doesn't hold the entitlement today (`Omi.entitlements`
/ `Omi-Release.entitlements` only grant `applesignin`, audio-input,
screen-capture, and automation). Since that approval can't be obtained as
part of this PR, code built against it couldn't be exercised or verified, so
it's scoped out rather than shipped as unverifiable. The other optional item
from the issue — "only show in-bar previews when Floating Bar is enabled" —
is *not* deferred; the policy's `floatingBarEnabled` check already prevents
the in-bar preview (and the earlier forced bar reveal) whenever the bar is
off, with no separate toggle needed. That case intentionally stays silent
rather than falling back to a system banner, matching pre-existing behavior
and avoiding the contentless-banner confusion noted in `NotificationService`.

- INV-CHAT-1 (path-matched via `desktop/macos/Desktop/Sources/FloatingControlBar/ShortcutSettings.swift`; the toggle only mutes in-bar preview rendering and does not fork the shared transcript)

## How it was verified

- `./scripts/dev-feedback.py --once swift 'FloatingBarNotificationPreviewPolicyTests'`
- Manual QA on a named dev bundle (`OMI_APP_NAME=omi-fix-6765 ./run.sh --yolo`):
  - Previews ON → proactive notification still shows under the Floating Bar (unchanged default behavior)
  - Previews OFF (bar enabled) → no in-bar card; native macOS banner still appears instead
  - Floating Bar disabled → no in-bar card and no temporary bar reveal for notifications; no forced system banner either (matches pre-existing silent behavior for this case, since a contentless banner was previously reported as confusing)
  - Snooze still suppresses both paths
  - Master Notifications toggle OFF still suppresses proactive notifications
- Screen recording of the above QA checklist attached

## Tests

Added `FloatingBarNotificationPreviewPolicyTests` covering the policy matrix:
previews+bar enabled (preview shown, no forced banner), bar disabled with
previews on (no preview, no forced banner), previews muted with bar enabled
(no preview, banner fallback), bar disabled with previews also muted (still
no forced banner), and explicit `deliverSystemBanner: true` (always delivers
regardless of preview state).

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

None.

##Video proof

https://github.com/user-attachments/assets/5d365549-0875-4c88-b1b9-ea901a1e0790

### DIFF
--- a/desktop/macos/Desktop/Sources/DefaultsKey.swift
+++ b/desktop/macos/Desktop/Sources/DefaultsKey.swift
@@ -52,6 +52,7 @@ enum DefaultsKey: String {
   /// Test hook: forces TTS playback start to report failure (non-prod gauntlets).
   case forceTTSPlaybackStartFalse = "forceTTSPlaybackStartFalse"
   case shortcutPTTInputDeviceUID = "shortcut_pttInputDeviceUID"
+  case floatingBarNotificationPreviewsEnabled = "shortcut_floatingBarNotificationPreviewsEnabled"
   case desktopIsPaywalled = "desktop_isPaywalled"
   case rewindDisableContentCache = "rewindDisableContentCache"
 }
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/ShortcutSettings.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/ShortcutSettings.swift
@@ -322,7 +322,7 @@ class ShortcutSettings: ObservableObject {
   @Published var floatingBarNotificationPreviewsEnabled: Bool {
     didSet {
       UserDefaults.standard.set(
-        floatingBarNotificationPreviewsEnabled, forKey: "shortcut_floatingBarNotificationPreviewsEnabled")
+        floatingBarNotificationPreviewsEnabled, forKey: .floatingBarNotificationPreviewsEnabled)
     }
   }
 
@@ -596,7 +596,7 @@ class ShortcutSettings: ObservableObject {
 
     self.askOmiEnabled = UserDefaults.standard.object(forKey: "shortcut_askOmiEnabled") as? Bool ?? true
     self.floatingBarNotificationPreviewsEnabled =
-      UserDefaults.standard.object(forKey: "shortcut_floatingBarNotificationPreviewsEnabled") as? Bool ?? true
+      UserDefaults.standard.object(forKey: .floatingBarNotificationPreviewsEnabled) as? Bool ?? true
     self.pttEnabled = UserDefaults.standard.object(forKey: "shortcut_pttEnabled") as? Bool ?? true
     self.doubleTapForLock = UserDefaults.standard.object(forKey: "shortcut_doubleTapForLock") as? Bool ?? true
     self.solidBackground = UserDefaults.standard.object(forKey: "shortcut_solidBackground") as? Bool ?? false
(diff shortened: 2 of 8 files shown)


Git-on-my-level (commented):
Thanks for the focused fix here. The policy split is a nice direction: the new toggle is default-on, the notification delivery decision is isolated in `FloatingBarNotificationPreviewPolicy`, and the tests cover the important matrix including the “Floating Bar disabled should stay silent” case.

I’m not formally approving this one because it changes notification delivery/privacy behavior and there is no passing check summary visible on the PR. A maintainer should confirm the intended UX/privacy tradeoff before merge: when a user disables in-bar previews while the Floating Bar remains enabled, assistant notification content moves to a native macOS banner, which may be more visible at the system level than the in-app preview.

One small thing for maintainers to double-check: the settings copy says “When off, notifications use macOS banners instead.” That is accurate only while the Floating Bar itself is enabled; if the bar is disabled, floating-bar-only notifications still stay silent by design. If that nuance matters for user expectation, the copy could be narrowed before merge.

I could not run the macOS Swift build/tests in this Linux review environment (`swift`/`xcrun` are unavailable). Static review looks good, and the added policy tests are the right coverage once CI or a macOS maintainer runner confirms them.

---

jnnabugwu:
> need human response

As a maintainer, do you mean me?

undivisible (approved):
Maintainer sign-off (merge request from @undivisible): confirming the UX/privacy tradeoff the earlier review asked a maintainer to check. The notification-preview toggle is default-on, so existing users see no behavior change; it only mutes in-bar preview rendering, and the banner-fallback policy matrix is covered by FloatingBarNotificationPreviewPolicyTests. The in-bar preview remains a view over the shared kernel conversation and does not fork or store transcript state, so INV-CHAT-1 is unaffected.

PR #10963, opened 2026-08-01
feat(desktop): let users choose which Gmail account to read

undivisible:
## What

Users can now choose which Google account Omi reads email from (macOS desktop, cookie-based connector) — in Settings and during onboarding — instead of silently connecting the first readable browser profile, which is often the wrong ("junk") account.

## Why

The Mac Gmail connector replays browser cookies and picks the first readable profile (Default wins). There was no way to select a different inbox after onboarding connected the wrong one.

## How

- Probe each browser profile session for its account email via the cookie-authenticated AccountInfo endpoint, so picker rows show the real address, not an opaque browser name.
- Persist the chosen profile cookie path; the reader narrows to it and falls back to all profiles if the choice goes stale.
- Onboarding: when several readable accounts exist and none is chosen yet, the Gmail insight read waits for the pick instead of importing the first account silently.
- Shared `GmailAccountPickerView` drives both the Settings card and the onboarding data-sources step.

No OAuth — cookie-path only. OAuth multi-account grants are a follow-up.

Invariants: INV-INT-1 (functional probe added; no agent/heuristic automation).

## Verification

- `swift build` clean
- Tests: GmailAccountSelectionTests (6), GmailOutcomeParserTests (6), BrowserGoogleSessionTests (6), OnboardingGoogleInsightGateTests (1) — all pass
- swiftlint clean on new files
- Ratchet baseline raised with justification: OnboardingPagedIntroCoordinator 1804 → 1849

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10963?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Changes which browser cookies are read for email import and onboarding memories; mishandling could import the wrong inbox or skip Gmail on probe failure, though guards (snapshot reads, fail-closed probe, sign-out key clearing) target those cases.
> 
> **Overview**
> **macOS Gmail (cookie-based connector)** no longer always uses the first readable Chromium profile. Users pick **Automatic** or a specific browser profile (email resolved via AccountInfo probe) in **Settings → Gmail** and during **onboarding** when multiple accounts exist.
> 
> **`GmailSelectionStore`** persists the chosen cookie DB path and label; **`GmailReaderService`** filters browser configs to that profile and snapshots the selection for multi-fetch reads so mid-read picker changes cannot mix inboxes. Stale selections widen with a diagnostics fallback.
> 
> **Onboarding** blocks the Gmail insight task until selection when needed; probe failures **skip Gmail import** instead of silently using the wrong account; cancel/dismiss records **Automatic** and resumes the waiter so research cannot hang. Gmail keys clear on account-scoped onboarding reset so the next Omi user on the same Mac does not inherit the prior Gmail choice.
> 
> Shared **`GmailAccountPickerView`**, unit tests, changelog entries, and e2e coverage updates ship with the change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0395a0a5bc4aafca8b88ce70a913883a3b67a46f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Failure-Class: none

Product invariants: INV-AGENT-*, INV-AUTH-1, INV-CHAT-1, INV-DATA-1, INV-MEM-1, INV-BETA-1 (merge-brought paths)


Git-on-my-level (dismissed):
Thanks for working on this — choosing the Gmail account explicitly is a useful macOS desktop improvement, and the Settings/onboarding direction makes sense.

I’m requesting changes before this is merge-ready because the current onboarding path can still hang or silently read the wrong profile in several cases:

- `awaitGmailAccountSelectionIfNeeded()` sets `gmailAwaitingSelection`, but the view only presents the sheet from the manual “Choose account” action. If onboarding discovers multiple accounts on its own, the Gmail task can suspend without showing the picker.
- The picker cancel paths only hide the sheet. During onboarding that leaves `gmailSelectionWaiter` suspended indefinitely; cancel should either choose Automatic/fallback or otherwise resume and let the flow reach a terminal state.
- There is a race where the user can select an account while probing is still in progress, before the continuation is installed. Re-check `GmailSelectionStore.hasMadeChoice` after probing and immediately before waiting so a pre-wait selection cannot be lost.
- Probe errors/timeouts currently collapse to `[]`, which is treated the same as “zero or one account”; that can skip selection and fall back to the first readable browser profile, which is exactly the behavior this PR is trying to avoid. Please distinguish probe failure from a real single-account/no-account result, or surface a retry/explicit fallback.

A couple of repo-contract/test items should be fixed with the above:

- New desktop Swift files should not be added directly under `desktop/macos/Desktop/Sources/`; please move the new Gmail selection/picker code into an appropriate feature directory. Also route the new UserDefaults keys through the typed defaults-key convention rather than raw string keys.
- `google-connector-read.yaml` lists the picker/settings UI under `covers`, but that flow appears to exercise the reader bridge rather than rendering/selecting in the picker. Please either add coverage for the picker path or keep the covers list to what the flow actually exercises.

The desktop Swift checks are a good signal, but this still needs the onboarding fixes above plus human maintainer review for the Gmail account-selection UX/data-handling behavior before merge.

---

Git-on-my-level (dismissed):
Thanks for the update — this is much closer. The move into `Sources/Integrations`, typed defaults keys, picker auto-presentation/cancel resume, and the explicit selection snapshot in `GmailReaderService` address several of the earlier blockers.

I’m still requesting one focused fix before this is merge-ready: the onboarding race around a manual selection during the account probe is still present. In `awaitGmailAccountSelectionIfNeeded()`, the comment says to re-check after the probe, but the second `GmailSelectionStore.hasMadeChoice` guard currently runs before `GmailAccountProbe.availableAccounts()`. If the user opens the manual picker and selects an account while that probe is in flight, `selectGmailAccount()` calls `resumeGmailSelection()` before `gmailSelectionWaiter` exists, so it is a no-op. When the probe returns with multiple accounts, the code then installs a continuation and waits indefinitely even though a choice is already persisted.

Please move/add the `hasMadeChoice` re-check after `availableAccounts()` returns and immediately before installing the continuation / showing the required picker. That should close the main remaining hang case without needing to split this PR.

Also note that the current head still has failing repo checks (`Hygiene` and `PR Metadata Preflight`), so this needs those resolved plus human maintainer sign-off for the Gmail account-selection UX/data-handling behavior before merge.

---

Git-on-my-level (commented):
Thanks for the continued updates. I re-checked the current head, and the specific onboarding blockers from the earlier automation review look resolved now:

- The required onboarding picker is presented automatically when multiple Gmail accounts are discovered.
- Cancel resumes the onboarding Gmail path instead of leaving the continuation suspended.
- The manual-selection race is covered by the post-probe `GmailSelectionStore.hasMadeChoice` re-check immediately before the continuation is installed.
- The reader snapshots the selected cookie path for a whole Gmail import, and the new defaults keys/files are now routed through the expected conventions.
- Current desktop checks are passing.

I’m not formally approving because this is still a Gmail account-selection UX/data-handling change and should get human maintainer sign-off before merge, but from this automated maintainer pass I do not see a remaining code blocker. Keeping the existing `needs-maintainer-review` state is appropriate for that final product/data-handling decision.

---

Git-on-my-level (dismissed):
Thanks for the safety updates here — the selected-cookie snapshot and the onboarding cancel/re-check paths are moving in the right direction.

I’m still requesting one focused fix before this is merge-ready: the current probe can still fail open when a browser profile has Google auth cookies but the AccountInfo lookup does not return an email. In `GmailAccountProbe.availableAccounts()`, those profiles are silently skipped (`if email: accounts.append(...)`). That means onboarding can see `accounts.count <= 1`, skip the required picker, and then `GmailReaderService` may still read using the automatic first-readable Gmail profile. This is the same wrong-account data-handling risk the PR is trying to remove, just through a per-profile AccountInfo failure instead of a whole-process failure.

Please make that path explicit: either include the readable-but-unidentified profile as an “Unknown account” option, or treat any authenticated profile whose identity cannot be verified as an ambiguous probe failure and fail closed / require retry or explicit fallback. A small unit test around parse/probe output with an auth-capable row missing email would be useful too.

Keeping `needs-maintainer-review` is still appropriate for the final Gmail account-selection UX/data-handling sign-off.

---

undivisible:
The Gmail probe now preserves authenticated browser profiles whose account email is unavailable, and the regression test covers that unknown-account case. Swift formatting and desktop test-quality checks pass on head `a53ce8611f`; the full Swift package test remains toolchain-blocked by the existing AudioKit deployment-target error.

Git-on-my-level (commented):
Thanks for the update — I rechecked head `a53ce861`. The previous automation blocker about readable but unidentified Gmail profiles looks resolved now: authenticated browser profiles are still returned as picker options even when AccountInfo does not expose an email, the picker renders those as “Unknown account”, and the new unit coverage includes that case.

From this automated maintainer pass I do not see a remaining code blocker in the Gmail account-selection implementation. I removed the stale `needs-tests` label because the focused selection/probe tests are now present.

I’m still not formally approving because this changes Gmail account-selection UX/data-handling and the PR is currently not cleanly mergeable; it should keep human maintainer sign-off before merge, with normal desktop checks rerun after the branch is updated. Keeping `needs-maintainer-review` and `workflow-review` is appropriate for that final product/data-handling/tooling review.

---

Git-on-my-level (commented):
Thanks for the continued iteration here. I rechecked the current head and I don't see a remaining implementation blocker from this automated maintainer pass.

Specific notes from the diff:

- `desktop/macos/Desktop/Sources/Integrations/GmailAccountSelection.swift` now keeps authenticated profiles even when AccountInfo does not return an email, maps those to nullable `email`, and the selection filter snapshots a specific `db_path` for a read so query/feed passes don't mix accounts.
- `desktop/macos/Desktop/Sources/Integrations/GmailAccountPickerView.swift` renders those unknown-email profiles as “Unknown account” rather than dropping them, while only showing the Automatic checkmark after an explicit choice.
- `desktop/macos/Desktop/Sources/GmailReaderService.swift` captures `GmailSelectionStore.selectedCookiePath` once per `readRecentEmails` / `verifyConnection` call and threads it through both the direct Atom read and label-feed expansion, which addresses the mixed-account import race.
- `desktop/macos/Desktop/Sources/Onboarding/OnboardingPagedIntroCoordinator.swift` now defers passive Gmail/Calendar reads unless the flow is user-initiated, rechecks `hasMadeChoice` after probing before installing the continuation, and resumes the waiter on cancel so onboarding is not stranded.
- `desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Sections/SettingsContentView+Integrations.swift` adds the Settings picker entry point, surfaces probe errors instead of treating them as empty success, and uses `gmailReadGeneration` to prevent stale read/save results from updating the UI after an account switch.
- `desktop/macos/Desktop/Tests/GmailAccountSelectionTests.swift` covers selected-profile filtering, stale-selection fallback, snapshot behavior, explicit Automatic, and parser handling for unknown-email profiles.
- `desktop/macos/e2e/flows/ai-chat-settings.yaml` and `desktop/macos/e2e/flows/google-connector-read.yaml` now assign the picker UI and selection store to the relevant desktop flow contracts; the changelog entries and onboarding ratchet baseline are valid JSON.

CI is green in the provided context, including Desktop Swift Build & Tests and Desktop Swift Static & Test Contracts. I also sanity-checked the changed JSON artifacts locally.

I'm not formally approving because this is still a Gmail account-selection UX/data-handling change and the PR carries maintainer/workflow review labels. Marking this as a positive automated signal, with human maintainer sign-off still needed for the account-selection behavior before merge.

---
*Automated maintainer review on behalf of David. Human maintainer sign-off is still needed for the Gmail account-selection UX/data-handling decision before merge.*

---

Issue #10948, opened 2026-07-31
Bug: Omi offline sync gets stuck all day in rate-limited loop, sync button is not usable, and the app shows no actionable error


hugoaap-code:
## Summary
Omi offline sync gets stuck for hours, even for the whole day, without actually syncing.

The app just says the recordings are ready to be synced, but the sync button is not usable and there is no meaningful error message or guidance.

## Environment
- Omi app
- Offline sync flow
- Observed on Friday, July 31, 2026

## Problem
Offline recordings stay in a `ready to sync` state indefinitely.

From the user side:
- it stays like that for hours or the entire day
- there is no useful message explaining what is wrong
- the sync UI is effectively stuck
- the user cannot actually complete the sync

The attached log suggests the sync process is repeatedly entering a rate-limited upload path and finishing with no new conversations.

## Expected behavior
- Offline recordings should sync successfully after connection is restored
- If sync is rate-limited or blocked, the app should explain that clearly
- The sync button should remain usable or the app should provide a clear retry/wait state
- The user should get an actionable error or status message instead of a silent stuck state

## Actual behavior
- Recordings remain marked as ready to sync
- Sync does not complete for hours
- The app provides no meaningful error explanation
- The sync button is not usable/cannot be clicked to force a successful sync
- Logs show repeated rate limiting and no conversations being created

## Impact
This makes offline recording sync unreliable and opaque.

The user is left waiting all day with no idea whether sync is broken, delayed, or permanently blocked. That creates the practical effect of lost or inaccessible recordings.

## Reproduction
1. Use Omi in a way that creates offline recordings pending sync
2. Return to a state where sync should be possible
3. Observe the app showing recordings ready to sync
4. Wait for sync to run
5. Observe that sync does not complete for hours
6. Observe that there is no useful error message and the sync button is not usable

## Evidence
Relevant log excerpt:

```json
{"ts":"2026-07-31T22:06:08.027Z","level":"INFO","message":"Sync Phase -1: Uploading fresh phone files first"}
{"timestamp":"2026-07-31T22:06:08.066Z","level":"EVENT","type":"local_upload_started","walCount":4}
{"timestamp":"2026-07-31T22:06:08.305Z","level":"EVENT","type":"local_upload_rate_limited","until":"2026-07-31 19:06:38.304"}
[...]

PR #10967, opened 2026-08-01
fix(backend): never strand users in an expired fair-use restriction (#10948)

aryanorastar:
## What changed and why

`normalize_expired_restriction_state` skipped the `restrict → throttle` expiry transition whenever `restrict_until` was not a `datetime`. Firestore normally returns timestamp fields as `datetime`, but a **string** timestamp (older write, admin import, manual console edit) silently disabled expiry — the user stayed hard-restricted indefinitely, so every sync/upload returned 429 and the client looped on `Retry-After`. This is one server-side contributor to the "offline sync stuck all day" symptom in #10948.

The fix coerces ISO-8601 string timestamps (with or without `Z` suffix) and fails safe toward expiry on unparseable values: an inconsistent restriction clears rather than blocking a user forever.

None.

## How it was verified

- `ENCRYPTION_SECRET=... .venv/bin/python -m pytest tests/unit/test_fair_use_engine.py tests/unit/test_fair_use_async.py tests/unit/test_fair_use_flagged_limit_clamp.py tests/unit/test_fair_use_models.py` → **99 passed**.
- `black --line-length 120 --skip-string-normalization` clean.
- `backend/scripts/scan_import_time_side_effects.py` → 0 violations.

## Tests

Six new regression tests in `TestNormalizeExpiredRestrictionState`: active datetime kept, expired datetime clears, expired ISO string clears (the bug), active Z-suffixed string kept, unparseable clears fail-safe, non-restrict stage untouched.

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups

None.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10967?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_fair_use_engine.py
+++ b/backend/tests/unit/test_fair_use_engine.py
@@ -6,7 +6,7 @@ an autouse ``monkeypatch`` fixture instead of mutating ``sys.modules`` at module
 scope. See ``backend/docs/test_isolation.md`` (Tier-2 sanctioned seams).
 """
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @aryanorastar — this is a well-scoped fix for a genuine user-facing bug (one server-side contributor to #10948). The direction is correct: fail safe toward clearing an expired/inconsistent restriction rather than stranding a user in a permanent hard-restriction.

A few observations from review:

1. The `_coerce_restrict_until` helper cleanly consolidates the datetime/string/absent cases and is correctly shared by both `normalize_expired_restriction_state` and `_retry_after_seconds_from_restrict_until`, so the "Retry-After loop" symptom and the "stuck in restrict" symptom are both addressed from one root cause.

2. Fail-safe semantics are sound: unparseable `restrict_until` clears to `throttle` rather than being treated as "still active." This is the safe direction (under-restriction over indefinite lockout).

3. The six regression tests cover the right matrix (active datetime, expired datetime, expired ISO string, active Z-suffix string, unparseable fail-safe, non-restrict stage untouched). I confirmed they pass against the PR branch.

4. One non-blocking improvement: the malformed-timestamp fail-safe branch in `normalize_expired_restriction_state` (where `restrict_until` coerces to `None`) silently downgrades a restriction. This repo has an established `record_fallback` observability pattern (`backend/utils/observability/fallback.py`) for exactly this class of silent-healing path — `reason="'malformed_doc'`, `component="'other'`, `outcome="'recovered'` are all in the allowlist. Adding a `record_fallback` call there would make recurrence observable in ops telemetry. Not a merge blocker, but worth doing before or shortly after merge so corrupted fair-use state doesn't heal quietly.

This PR touches the fair-use enforcement surface (user-data persistence), so I'm leaving it for human maintainer review rather than approving. The code itself looks correct and safe.

---

kodjima33 (approved):
Bug fix 5/5: coerces ISO-8601 string restrict_until so expired fair-use restrictions actually clear; fail-safe on unparseable, 6 regression tests, CI green.

PR #10970, opened 2026-08-01
fix(app): transcript readiness is socket-driven, not connectivity-driven (#6311)

### LINKED ISSUE #6311: "Recording, reconnecting" shows falsely when transcription is working fine
## Bug

"Recording, reconnecting" status appears in the recording bar even when transcription is working perfectly — live transcript keeps flowing normally despite the UI showing a disconnected state.

## Root Cause

`transcriptServiceReady` in `capture_provider.dart:252` requires BOTH conditions:
```dart
bool get transcriptServiceReady => _transcriptServiceReady && _isConnected;
```

`_isConnected` comes from `ConnectivityService` which can briefly flicker `false` during network transitions (WiFi↔cellular, brief DNS hiccups) even when the WebSocket is still alive and streaming data. This causes the UI to show "Recording, reconnecting" with the red dot + cloud-off icon while transcription continues normally.

## Expected

The status should reflect actual transcript service state, not network connectivity checks. If the WebSocket is connected and segments are flowing, the UI should show "Listening" regardless of what `ConnectivityService` reports.

## Suggested Fix

Either:
1. Remove the `_isConnected` check from `transcriptServiceReady` — rely solely on `_transcriptServiceReady` which is set/cleared by actual WebSocket connection events
2. Add a debounce to the connectivity check (e.g., only show reconnecting if disconnected for >3 seconds)
3. Check if transcript segments are still arriving as a health signal

## Files
- `app/lib/providers/capture_provider.dart:252` — the combined check
- `app/lib/pages/conversations/widgets/processing_capture.dart:299` — UI rendering

Originally added in PR #5699 (Mar 20, 2026).

aryanorastar:
Fixes #6311

## What changed and why

`transcriptServiceReady` was `_transcriptServiceReady && _isConnected`, where `_isConnected` comes from `ConnectivityService`. That service can flicker `false` during a WiFi↔cellular handoff or a brief DNS hiccup even while the transcript WebSocket is alive and segments are flowing — so the recording header showed "Recording, reconnecting" (red dot + cloud-off icon) over perfectly healthy transcription.

`_transcriptServiceReady` is already the authoritative signal (set `true` on socket subscribe, `false` on close/error). This drops the `_isConnected` AND so the UI reflects the actual transcript pipeline rather than a network check.

`ConnectivityService` state remains in use where it belongs — `hasNetwork` for the recording manager and `onConnectionStateChanged` mirroring — only the readiness getter changes.

None (UI state fix on an existing path).

## How it was verified

- `flutter test test/providers/capture_provider_test.dart` → **67 passed** (66 existing + 1 new regression).
- `flutter analyze lib/services/capture/capture_controller.dart test/providers/capture_provider_test.dart` → no new issues (3 pre-existing infos in the test file, untouched by this change).
- `dart format --line-length 120` clean.

## Tests

New `transcriptServiceReady is socket-driven, not connectivity-driven (#6311)` group: asserts a connectivity flicker (`onConnectionStateChanged(false)` → `true`) cannot toggle transcript readiness — the exact regression that showed a false "reconnecting" state.

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups

None.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10970?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/services/capture/capture_controller.dart
+++ b/app/lib/services/capture/capture_controller.dart
@@ -411,7 +411,13 @@ class CaptureController extends ChangeNotifier
 
   bool _transcriptServiceReady = false;
 
-  bool get transcriptServiceReady => _transcriptServiceReady && _isConnected;
+  // The transcript service readiness is driven solely by the socket lifecycle
+  // (set true on subscribe, false on close). The `&& _isConnected` gate was
+  // removed (#6311): ConnectivityService can flicker false during a WiFi↔cellular
+  // handoff or a brief DNS hiccup even while the WebSocket is alive and segments
+  // are flowing, which made the UI show "Recording, reconnecting" over healthy
+  // transcription. The socket is the authoritative connectivity signal.
+  bool get transcriptServiceReady => _transcriptServiceReady;
 
   // having a connected device or using the phone's mic for recording.
   // Includes `interrupted` so the keep-alive/reconnect path keeps running
--- a/app/test/providers/capture_provider_test.dart
+++ b/app/test/providers/capture_provider_test.dart
@@ -1130,18 +1130,28 @@ void main() {
   // lifecycle is the authoritative signal; a connectivity change must never
   // change transcript-service readiness.
   group('transcriptServiceReady is socket-driven, not connectivity-driven (#6311)', () {
-    test('connectivity flicker does not toggle transcript readiness', () {
+    test('connectivity flicker does not toggle readiness of a ready socket', () {
       final provider = CaptureProvider();
-      // Fresh provider: socket not yet ready, so readiness must be false.
-      expect(provider.transcriptServiceReady, isFalse);
 
-      // A connectivity flip must not change readiness on its own — before the
-      // fix, onConnectionStateChanged(false) ANDed _isConnected into the getter
-      // and could manufacture a "reconnecting" state from a healthy socket.
+      // Drive the provider into the socket-subscribed state (the scenario the
+      // old getter got wrong): onConnected mirrors the transcript WebSocket
+      // subscribing, which sets _transcriptServiceReady = true.
+      provider.onConnected();
+      expect(provider.transcriptServiceReady, isTrue);
+
+      // A connectivity flicker must not toggle readiness off. Before the fix
+      // this ANDed _isConnected=false into the getter and manufactured a false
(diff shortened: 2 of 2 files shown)


Git-on-my-level (commented):
Thanks @aryanorastar — the core fix looks right. `_transcriptServiceReady` is the authoritative socket-lifecycle flag (`true` on `subscribe`/`onConnected`, `false` on `onClosed`/`onError`), so dropping the `&& _isConnected` AND correctly stops a transient `ConnectivityService` flicker (WiFi↔cellular handoff, brief DNS hiccup) from flashing a false "reconnecting" state while the WebSocket is healthy and segments are flowing. Minimal and well-scoped. CI is green.

One blocking concern on the regression test, which I agree with the cubic review on and verified against the diff:

The new `transcriptServiceReady is socket-driven...` test never drives `_transcriptServiceReady` to `true`. The provider is constructed fresh (so readiness is already `false`), and then `onConnectionStateChanged(false)` / `onConnectionStateChanged(true)` only flip `_isConnected`. All three `expect(..., isFalse)` assertions pass identically against both the old getter (`_transcriptServiceReady && _isConnected`) and the new one — so the test does not actually guard the #6311 behavior. The scenario the test's comment describes (a *ready* socket plus a connectivity flicker, readiness staying `true`) is never constructed.

For the test to actually protect this fix, it needs to put the provider into a socket-subscribed/connected state first (e.g. via a mock `TranscriptSocketService` that drives `subscribe`/`onConnected` so `_transcriptServiceReady == true`), then flicker `onConnectionStateChanged(false)` and assert `transcriptServiceReady` stays `true`. As written it adds no regression signal.

Not asking for a split — this is one cohesive fix. Just needs the test to cover the actual differing branch so a future regression is caught. Marking for human maintainer review alongside.

---

kodjima33 (approved):
Bug fix 5/5: transcriptServiceReady no longer ANDed with flaky connectivity state; root cause stated, regression test added, CI green.

PR #10972, opened 2026-08-01
feat(app): Omi HUD on display-capable Meta glasses, gated on the capability

undivisible:
## What this is

The Omi HUD on display-capable Meta glasses, gated on the capability rather than on the device being a Ray-Ban.

## The finding this rests on

`docs/rayban-meta-device.md` said the toolkit gives us camera only. That was true when written and is now stale: **DAT 0.7.0 (2026-05-14) added the Display capability**, and **0.8.0 ships it** plus `clearDisplay()`. The repo already pinned `meta-wearables-dat-ios` at exact 0.8.0 — it just never linked `MWDATDisplay`, so the display was sitting behind an unlinked module the whole time.

The no-microphone finding still holds. Audio remains the Bluetooth HFP route; nothing in the audio path changes here.

## Capability, not device type

`deviceSupportsDisplay()` returns the toolkit's own `Device.supportsDisplay()`. Any display-capable Meta hardware lights up with no new device case, and Ray-Ban Meta Gen 1/2 reports `false`. This is the same signal Meta's own sample uses to filter devices (`AutoDeviceSelector(wearables:filter:{ $0.supportsDisplay() })`).

`GlassesDisplayGate` keeps two conditions deliberately separate — the device can render, and the user opted in. Neither implies the other, so a display-less device can never be switched on by the flag alone.

## Content is device-neutral

`HudScreen`/`HudLine` describe *what* to show without naming an SDK; `RayBanMetaDisplayRenderer` translates into Meta's `FlexBox`/`Text`/`Button`. A future vendor implements `GlassesDisplay` and reuses the projector unchanged.

`HudProjector` owns the line budget, which is the actual design problem — a HUD shows roughly five lines and `TaskActionItem` alone carries ~30 fields. The projector decides what survives (due date, then priority, then text) and appends an explicit `+N more` rather than letting the renderer clip silently.

## Two things found on the way

**The graph test's exactness assertion was vacuous.** `test_separate_dat_target_owns_only_the_required_meta_products` asserts an exact module set, but `package_product_names` scanned for `MWDAT(?:Core|Camera)` — a newly linked product was invisible to it, so the assertion passed without checking anything. Broadened to `MWDAT\w+`, which is what makes the new `assert_equal` meaningful.

**Pigeon emits trailing whitespace** (`case 134: `) into generated Dart, which `diff-hygiene` rejects. Stripped in this PR; it will recur on the next `pigeon` run, so it likely wants a generator post-step rather than a manual fix each time. Not addressed here.

`MWDATDisplay` is linked into `RunnerRayBanDat` only, never default `Runner` — linking it there would re-trigger the `MWDATCore` SwiftProtobuf duplicate-class launch crash documented in `rayban-meta-troubleshooting.md`.

`RayBanMetaDisplayRenderer.swift` does not import SwiftUI: `MWDATDisplay` exports its own `Text`, `Button`, and `Image`, and importing both makes those names ambiguous (Meta documents this in the 0.7.0 notes).

## Verification

```
flutter test test/unit/hud_projector_test.dart                    12 passed
flutter test test/unit/rayban_meta_device_test.dart
             test/unit/rayban_meta_source_mapping_test.dart       22 passed (no regression)
ruby app/ios/test/rayban_dat_xcode_graph_test.rb                   9 runs, 402 assertions, 0 failures
ruby app/ios/test/rayban_dat_plugin_boundary_test.rb               9 runs,  99 assertions, 0 failures
ruby app/ios/test/rayban_dat_build_wrapper_test.rb                12 runs, 129 assertions, 0 failures
plutil -lint app/ios/Runner.xcodeproj/project.pbxproj              OK
scripts/pre-push (bounded preflight gate)                          passed
```

The graph-test count moving 393 → 402 is the evidence the pbxproj wiring actually took: the broadened scan now sees `MWDATDisplay` in `packageProductDependencies`.

## Compiled against the real SDK

The first pass of the Swift was written from Meta's changelog and one sample file without a compiler, and **three calls did not exist**. The package is now resolved locally and both files typecheck against the actual 0.8.0 xcframeworks:

```
swift package resolve                       meta-wearables-dat-ios 0.8.0
xcrun swiftc -typecheck -target arm64-apple-ios15.2-simulator \
  -F MWDATCore.xcframework/ios-arm64_x86_64-simulator \
  -F MWDATDisplay.xcframework/ios-arm64_x86_64-simulator \
  Renderer.swift HostProbe.swift            no diagnostics
```

What the interface corrected:

| Written from docs | Actual interface |
|---|---|
| `Display.stateStream()` | no such member — `statePublisher: any Announcer<DisplayState>`, whose `listen()` returns `AnyListenerToken` |
| `DeviceSession.removeDisplay(_)` | does not exist; `addDisplay()` has no inverse, so detach is `stop()` plus dropping the token |
| `DisplayState` falling through to `"error"` | `@frozen`, exactly `starting/started/stopping/stopped`, no error case — `.stopping` was being reported as an error |

The renderer's view API needed no change: `Text(_:style:color:)`, the `FlexBox` `ComponentBuilder` init, `padding`/`background`/`onTap`, and `Button(label:style:iconName:onClick:)` all matched.

## Review follow-up (this push)

Merged `origin/main` (594 commits) — no conflicts.

**Attach is now atomic.** `addDisplay()` was the only step inside the error-emitting `do/catch`; the state-publisher subscription and `capability.start()` ran outside it. A failure after `addDisplay()` succeeded left Dart holding `attaching` forever with no `error` transition *and* no way to retry, since `display` was already non-nil so `attachDisplay()` early-returns. The whole sequence now runs in one `do/catch` that rolls back the capability handle and its subscription before emitting `error`.

**Display state contract documented** in `docs/rayban-meta-device.md`: the five Dart-visible states, the atomic-attach guarantee, why `.stopping` and `.stopped` both normalize to `detached` (nothing may be sent to a display that is tearing down, so both are the same state to every Dart caller), and that best-effort `send`/`clear` failures surface via `onError` as `display_send_failed`/`display_clear_failed` without changing the display state — a dropped HUD refresh is not a detach.

No behavioral regression test accompanies the attach fix: `app/ios` has no Swift test target, and asserting on source-string order would be a static tripwire rather than behavioral coverage. The fix is structural (one `do/catch` spanning the whole sequence).

Verification:

```
ruby app/ios/test/rayban_dat_xcode_graph_test.rb     -> 9 runs, 402 assertions, 0 failures
ruby app/ios/test/rayban_dat_plugin_boundary_test.rb -> 9 runs,  99 assertions, 0 failures
flutter test test/unit/hud_projector_test.dart       -> 13 tests passed
scripts/pr-preflight --pr-body-file ...              -> 13 checks passed
```

## Still not done

**Not run on hardware, and not run in the simulator.** Typechecking proves the SDK calls are real; it does not prove anything renders. This wants one run against `MockDeviceKit.pairGlasses(model: .metaGlasses)` plus `MockCaptouchKit` for the tap path — simulator-only, no glasses needed.

**Nothing feeds the projector yet.** `HudProjector` is tested against `HudTask`, but `ActionItemsProvider`, the live transcript, and AI answers are not wired to it, and no preference persists `userEnabled`. The seam exists; the wiring is the next PR.

none

## Failure class

Failure-Class: none

### DIFF
--- a/app/android/app/src/main/kotlin/com/friend/ios/PigeonCommunicator.g.kt
+++ b/app/android/app/src/main/kotlin/com/friend/ios/PigeonCommunicator.g.kt
@@ -66,7 +66,7 @@ private object PigeonCommunicatorPigeonUtils {
     }
     return a == b
   }
-      
+
 }
 
 /**
@@ -898,7 +898,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onRecordingStopped(callback: (Result<Unit>) -> Unit)
@@ -915,7 +915,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onAudioData(audioDataArg: ByteArray, callback: (Result<Unit>) -> Unit)
@@ -932,7 +932,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onAudioChunk(audioChunkArg: ByteArray, chunkIndexArg: Long, isLastArg: Boolean, sampleRateArg: Double, callback: (Result<Unit>) -> Unit)
@@ -949,7 +949,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onRecordingError(errorArg: String, callback: (Result<Unit>) -> Unit)
@@ -966,7 +966,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onMicrophonePermissionResult(grantedArg: Boolean, callback: (Result<Unit>) -> Unit)
@@ -983,7 +983,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onMainAppMicrophonePermissionResult(grantedArg: Boolean, callback: (Result<Unit>) -> Unit)
@@ -1000,7 +1000,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onWatchBatteryUpdate(batteryLevelArg: Double, batteryStateArg: Long, callback: (Result<Unit>) -> Unit)
@@ -1017,7 +1017,7 @@ class WatchRecorderFlutterAPI(private val binaryMessenger: BinaryMessenger, priv
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
 }
@@ -1423,7 +1423,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onPeripheralDiscovered(peripheralArg: BlePeripheral, callback: (Result<Unit>) -> Unit)
@@ -1440,7 +1440,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onDeviceReady(peripheralUuidArg: String, servicesArg: List<BleService>, callback: (Result<Unit>) -> Unit)
@@ -1457,7 +1457,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onPeripheralDisconnected(peripheralUuidArg: String, errorArg: String?, callback: (Result<Unit>) -> Unit)
@@ -1474,7 +1474,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onCharacteristicValueUpdated(peripheralUuidArg: String, serviceUuidArg: String, characteristicUuidArg: String, valueArg: ByteArray, callback: (Result<Unit>) -> Unit)
@@ -1491,7 +1491,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onRssiUpdate(peripheralUuidArg: String, rssiArg: Long, callback: (Result<Unit>) -> Unit)
@@ -1508,7 +1508,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   fun onStateRestored(peripheralUuidsArg: List<String>, callback: (Result<Unit>) -> Unit)
@@ -1525,7 +1525,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
   /**
@@ -1546,7 +1546,7 @@ class BleFlutterApi(private val binaryMessenger: BinaryMessenger, private val me
         }
       } else {
         callback(Result.failure(PigeonCommunicatorPigeonUtils.createConnectionError(channelName)))
-      } 
+      }
     }
   }
 }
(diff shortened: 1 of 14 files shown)


Git-on-my-level:
Solid foundational work. The device-neutral content model, capability-gated display surface, and projector-owned line budget are exactly the right abstractions — keeping `HudScreen`/`HudLine` vendor-neutral while `RayBanMetaDisplayRenderer` does the Meta-specific translation means a future vendor just implements `GlassesDisplay` and reuses the projector unchanged.

The `GlassesDisplayGate` design (device capability **and** user opt-in, neither implying the other) is clean. The projector tests cover the important edge cases: budget overflow with explicit `+N more`, sorting precedence (due date → priority → text), whitespace collapse, truncation, and the three gate states.

A few observations for when this moves toward merge:

1. **This is plumbing, not yet wired** — no concrete `GlassesDisplay` implementation exists in the Dart layer, the projector isn't fed by any live app state, and there's no settings toggle for the user opt-in. That's fine for a draft; just flagging the remaining surface.
2. The `DisplayState` normalization maps `.stopping` and `.stopped` both to `"detached"` (`RayBanMetaHostApiImpl.swift:527`) — worth confirming that's the intended UX (a stopping display and a fully-stopped one are indistinguishable from Dart).
3. `sendDisplayScreen` fires `capability.send(view)` in an unstructured `Task` with error-only callback — if the view fails to send, the Dart side has no signal beyond `onError`. Consider whether a failed send should transition the display state to `error`, or whether that's acceptable for a best-effort HUD refresh.

Human maintainer review needed for product direction before this leaves draft. No security, privacy, or dependency concerns — the display renders only existing app content and `MWDATDisplay` was already pinned at 0.8.0.

---

Git-on-my-level:
Strong foundational work — the layering here is exactly right.

The device-neutral content model, capability gate, and projector-owned line budget are well separated: `HudScreen`/`HudLine` stay vendor-free, `GlassesDisplayGate` keeps device-capability and user-opt-in as independent conditions, and `HudProjector` makes the line-budget decision explicit (`+N more`) rather than letting the renderer clip silently. The generated Pigeon files are consistent across Dart/Swift/Kotlin, and broadening the Xcode graph test regex from `MWDAT(?:Core|Camera)` to `MWDAT\w+` is a genuine improvement over a vacuous assertion.

A few things I would flag for the next iteration:

1. **attachDisplay error-state gap** — `emitDisplayState("attaching")` fires before `session.addDisplay()`. If `addDisplay()` or `capability.start()` throws, the Dart side is left at `attaching` with no `error` transition. Consider wrapping the attach in a do/catch that emits an error state on failure.

2. **clearDisplay swallows errors** — `Task { try? await capability.clearDisplay() }` silently drops failures. If clear is intentionally best-effort, a one-line comment saying so would help; otherwise surface it through `emitError`.

3. **GlassesDisplay has no concrete implementation yet** — the abstract interface and gate are defined, the native Pigeon API exists (`deviceSupportsDisplay`, `attachDisplay`, `sendDisplayScreen`), and the bridge callbacks are wired (`onDisplayStateChangedCb`, `onDisplayActionTappedCb`), but no Dart class implements `GlassesDisplay` to connect them. Expected for a draft foundational PR, but worth tracking as the explicit next step.

No security, privacy, or data-handling concerns. Looking forward to seeing the integration layer land.

---

Git-on-my-level:
Followed up on the latest head (`ac0dfa91a`). The two concrete blockers from the earlier pass are now resolved:

- **attachDisplay** wraps `session.addDisplay()` in a `do/catch` that emits an `"error"` display state on failure (previously Dart could be left at `"attaching"`).
- **clearDisplay** surfaces failures through `emitError` instead of swallowing them with `try?`.

One small edge case remains, acceptable for a draft: `capability.start()` sits *outside* the error-emitting `do/catch` after `addDisplay()` succeeds — if `start()` throws synchronously before the state publisher emits, Dart could be left at `"attaching"`. Worth a one-line guard when you wire the integration layer. The `DisplayState` normalization mapping `.stopping`/`.stopped` both to `"detached"` reads as an intentional UX choice.

No security, privacy, or dependency concerns — the display renders only existing app content and `MWDATDisplay` was already pinned at 0.8.0. Human maintainer review is still needed for product direction before this leaves draft.

---

undivisible:
Pushed `e1634b6`.

`capability.start()` and the state-publisher subscription now run inside the same `do/catch` as `addDisplay()`, which rolls back the capability handle and the token before emitting `error`. The stuck-at-`attaching` case also had no retry path, since `display` was already non-nil and `attachDisplay()` early-returns.

`.stopping`/`.stopped` → `detached` is intentional, now recorded in `docs/rayban-meta-device.md` alongside the rest of the display state contract: nothing may be sent to a display that is tearing down, so both are the same state to every Dart caller.

Also merged `origin/main` (594 commits, no conflicts). Still draft — the projector has no live feed and no user-opt-in preference yet, and this needs a maintainer call on product direction before it leaves draft.

Git-on-my-level:
Solid work on this — the architecture is clean and the capability-vs-device-type gating is the right call. A few specific observations:

**Strengths**
- `GlassesDisplayGate.isOpen` (glasses_display.dart) requiring both `deviceSupportsDisplay` AND `userEnabled` means a display-less device can never be turned on by the flag alone. Good fail-safe.
- `HudProjector._taskOrder` (hud_projector.dart): deterministic tiebreaker chain (due date → priority → text), and the overflow path correctly reserves a slot for `+N more` instead of silently clipping. The line-budget ownership is the right design decision.
- `attachDisplay()` (RayBanMetaHostApiImpl.swift) is properly atomic — `addDisplay()` + state subscription + `start()` all live in one `do/catch` with rollback to nil on failure, so Dart can never get stuck at `attaching`. The follow-up push fixing this was important.
- `RayBanMetaDisplayRenderer` correctly avoids importing SwiftUI to sidestep the `Text`/`Button`/`Image` name collision with MWDATDisplay.
- The `rayban_dat_xcode_graph_test.rb` regex hardening (`MWDAT(?:Core|Camera)` → `MWDAT\w+`) was a good catch — the old assertion was vacuous.

**For maintainer attention**
- **Product direction needed for HUD content/scope:** what gets the limited line budget (tasks vs. capture state vs. AI answers) and in what priority is a product/UX decision that warrants maintainer sign-off, not just code review.
- **Not yet validated on hardware or simulator** (acknowledged in the PR body). The proposed MockDeviceKit test would be the natural next step.
- **Projector not wired to live data** yet (ActionItemsProvider, transcript, AI answers) and `userEnabled` is not persisted — worth confirming the seam design before that wiring lands in a follow-up.

The 13 projector unit tests and the passing checks (Dart Analyze & Tests, Generated Files, Desktop Swift Static & Test Contracts) look good for the unit-testable surface. Leaving for human maintainer review on product direction for the HUD feature.

---

Git-on-my-level:
Followed the branch to `4e6aa85`. The two newer pieces hold up, and both close threads from the earlier passes:

- **The bridge now matches the real `MWDATDisplay` 0.8.0 interface.** The corrections are all visible in `RayBanMetaHostApiImpl.swift` and consistent end to end: state via `statePublisher.listen` returning an `AnyListenerToken` (no `stateStream()`), detach as `stop()` + token drop (the SDK genuinely has no `removeDisplay`), and `normalizeDisplayState` covering exactly the frozen `starting/started/stopping/stopped` cases with no error case. The `.stopping`/`.stopped` -> `detached` rationale now recorded in `docs/rayban-meta-device.md` answers the question my earlier review raised.
- **Target isolation independently re-verified.** Parsing `project.pbxproj` directly: `MWDATDisplay` appears in exactly one frameworks phase, the one owned by `RunnerRayBanDat`, with its product dependency on that same target - default `Runner` stays clean, which is what the SwiftProtobuf duplicate-class history requires. The broadened `MWDAT\w+` scan in `rayban_dat_xcode_graph_test.rb` is what makes that assertion meaningful instead of vacuous.

Two notes, neither blocking for a draft:

1. `detachDisplay()` emits `"detached"` even when nothing was attached, and in non-toolkit builds `detachDisplay()`/`clearDisplay()` are silent no-ops while `sendDisplayScreen()` throws `display_unavailable`. Defensible best-effort-teardown asymmetry - worth one line in the state-contract doc when the wiring PR lands.
2. The external claims I can't check from the repo (DAT 0.7.0 added Display, 0.8.0 ships `clearDisplay()`) are consistent with everything verifiable here: the `meta-wearables-dat-ios` exact-0.8.0 pin is untouched and the code exercises only the interface the PR documents.

The remaining surface is unchanged and already owned in the PR body: no concrete `GlassesDisplay` implementation feeding the projector, no persisted `userEnabled` opt-in, and no simulator/hardware run yet (`MockDeviceKit.pairGlasses` is the right next step). The product call - what earns the ~5-line budget (tasks vs capture state vs answers) - is the maintainer decision this draft is waiting on.

---
*by AI on behalf of David*

PR #10973, opened 2026-08-01
perf(desktop): stop sending every tool's docs twice per chat turn

undivisible:
## What this changes

This branch withdraws the proposed prompt/tool-doc deduplication after review found that the active paths do not support a safe global projection. The removed Swift prompt is used by Chat Lab, while normal Home chat uses the kernel policy; globally expanding pi-mono and MCP tool descriptions would make ordinary turns larger and could drop or contradict adapter guidance.

The corrective commit restores the original prompt and tool-definition projections, removes the fake `chat-hermetic` coverage claim, and removes the changelog entry. The resulting tree matches `main`; this PR should be closed or replaced with a surface-specific implementation.

## Verification

- `xcrun swift test --package-path Desktop --filter 'ChatDiscoverabilityTests|ChatPromptsTests'`
- `./scripts/agent-logic-harness.sh --node-only --skip-install`
- `git diff origin/main...HEAD` is empty

Failure-Class: none

### DIFF
--- a/desktop/macos/Desktop/Sources/Chat/DesktopCapabilityRegistry.swift
+++ b/desktop/macos/Desktop/Sources/Chat/DesktopCapabilityRegistry.swift
@@ -19,18 +19,17 @@ enum DesktopCapabilityRegistry {
     let filteredCapabilities = capabilities(for: .desktopChat)
       .filter { !excludedToolNames.contains($0.toolName) }
     let availableToolNames = Set(filteredCapabilities.map(\.toolName))
+    let docs =
+      filteredCapabilities
+      .map { toolDoc($0, excluding: excludedToolNames) }
+      .joined(separator: "\n\n")
     let taskAgentAwareness = taskAgentAwarenessPrompt(availableToolNames: availableToolNames)
     let proactiveGuidance = proactiveGuidancePrompt(availableToolNames: availableToolNames)
     let usageGuidance = usageGuidancePrompt(availableToolNames: availableToolNames)
-    // Per-tool name/summary/bullets are deliberately absent. The adapter already
-    // registers every advertised tool with its description, promptGuidelines,
-    // and input schema, so restating them here sent the same content twice in
-    // every turn — roughly 3.5k tokens of it — and gave the two copies room to
-    // disagree as the manifest changed. What stays is the cross-tool guidance
-    // that has nowhere else to live: when to reach for a tool at all, and how
-    // the tools relate to each other.
     return """
-      Use your Omi tools before answering when the question depends on the user's personal data, tasks, conversations, memories, app/screen activity, or task-agent state. Do not guess when you can look it up. Do not call tools for simple chit-chat or general knowledge that does not depend on the user's data.
+      These Omi data/status tools are documented for desktop chat. Use them before answering when the question depends on the user's personal data, tasks, conversations, memories, app/screen activity, or task-agent state. Do not guess when you can look it up. Do not call tools for simple chit-chat or general knowledge that does not depend on the user's data.
+
+      \(docs)
 
       \(taskAgentAwareness)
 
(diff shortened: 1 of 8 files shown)


Git-on-my-level (dismissed):
Good change — the deduplication thesis is correct. I verified the bullet content ("Supports SELECT, INSERT, UPDATE, DELETE", "Vector similarity search on tasks", "find tasks about shopping", "More reliable than hand-writing MATCH queries", "Deduplication is handled automatically") all exists in `GeneratedToolCapabilities.swift` as `capability.bullets`, which feed the tool definitions' `promptGuidelines`. Removing the prompt-side copy is sound.

**Blocking: CI is red.** Six existing tests in `ChatDiscoverabilityTests.swift` assert content that only came from the removed `toolDoc` blocks and were not updated in this PR:

1. `testDesktopPromptMentionsEveryDesktopCapability` (line 122) — asserts `prompt.contains("**\(toolName)**")` for all 46 tools; the bold-marked names were produced only by `toolDoc`.
2. `testToolPromptIncludesSearchTasks` (lines 95–96) — asserts `"**search_tasks**"` and `"Vector similarity search on tasks"`.
3. `testDesktopPromptMentionsListAgentSessionsForSubagents` (lines 140, 143) — asserts `"**list_agent_sessions**"`, `"Call list_agent_sessions"`, `"floating_agent_pills"`.
4. `testDesktopPromptCanSpawnFloatingAgents` (line 148) — asserts `"**spawn_agent**"`.
5. `testDesktopPromptPreservesLegacyToolBehaviorGuidance` (lines 164–168) — asserts `"Supports SELECT, INSERT, UPDATE, DELETE"`, `"Supports FTS5 MATCH queries"`, `"More reliable than hand-writing MATCH queries for task search"`, `"**save_knowledge_graph**"`, `"Deduplication is handled automatically"`.
6. `testToolPromptListsSearchTasksInWhenToUse` (line 116) — asserts `"find tasks about shopping"`.

The PR description notes that existing `ChatPromptsTests` assertions on retained guidance still hold — that checks out for `ChatPromptsTests.swift`, but `ChatDiscoverabilityTests.swift` was missed.

These tests encode the old prompt contract (per-tool bold blocks + bullets in the system prompt). Since the new contract is "per-tool docs live only in tool definitions," these assertions should be migrated: either removed (the content is now verified by `testAgentControlCapabilitiesMatchCanonicalManifest`, which cross-checks bullets against the manifest) or rewritten to assert the new contract — e.g. that the prompt does NOT contain per-tool bold blocks, complementing the two tests you added in `ChatPromptsTests.swift`.

The two new tests in `ChatPromptsTests.swift` are well-paired (one proves dedup, the other proves the prompt is not emptied to satisfy it). Once the six `ChatDiscoverabilityTests` cases are reconciled, this should be ready.

---

Git-on-my-level:
Good change and well-scoped — deduplicating tool docs that were already carried by the tool definitions is the right call, and the new `modelVisibleDescription` correctly folds unique `capabilityDoc.bullets` into the adapter description so nothing is lost. The test migration is thoughtful: `testDesktopToolPromptDoesNotRestatePerToolDocs` guards the dedupe against regression, `testDesktopToolPromptKeepsCrossToolGuidance` prevents the first test from passing on an empty prompt, and the existing assertions were moved to assert against `capabilityDoc`/`tool definitions` rather than deleted. The changelog entry and `chat-hermetic.yaml` coverage update are appropriate.

One CI blocker to resolve before merge:

**PR Metadata Preflight is failing** because the diff touches two locked product-invariant path globs (`desktop/macos/Desktop/Sources/Chat/**` → `INV-CHAT-1`, and `desktop/macos/agent/src/runtime/**` → `INV-AGENT-*`) but the PR body does not name them. Run `scripts/pr-preflight --suggest` (or `python .github/scripts/check_product_invariants.py --changed-files <files> --pr-body-file <draft>`) and add the suggested block:

```

- INV-AGENT-*
- INV-CHAT-1
```

This looks purely administrative — the change does not alter session/run identity, authority, lifecycle, or chat continuity contracts; it only stops restating per-tool docs that the adapter already registers. Naming the invariants is the repo's required acknowledgement, not a sign of a semantic conflict. The `needs-tests` label also looks stale at this point — the PR adds two dedicated dedupe guards and migrates the existing assertions, so coverage is no longer the gap.

Minor observation (non-blocking): `modelVisibleDescription` uses substring containment (`covered.includes(bullet)`) for dedupe, so a bullet that is a substring of another covered string is dropped. That's fine for the current authored content and keeps the logic simple; just flagging in case future bullets are crafted where a partial match would over-suppress.

---

PR #10974, opened 2026-08-01
fix(windows): let Stop cancel a memories bulk delete during a rate-li…

toughc-web:
## What changed and why

`deleteMemoriesPaced` only observed `shouldStop()` between ids, so both of its waits were uninterruptible. A Stop pressed during a 429 `Retry-After` pause could not take effect until the whole pause elapsed, and when it ended the loop retried and deleted one more memory after the user had already asked to stop. The 1100ms pacing tail had the same property.

`waitInterruptible(ms, shouldStop)` now sleeps in 250ms slices and returns as soon as `shouldStop()` flips, and the 429 branch rechecks it after the pause and stops without deleting. An id cancelled mid-retry counts as neither deleted nor failed and fires no `onResult`, so its row stays: that memory still exists.

Callers that pass no `shouldStop` (`retentionSweep.ts`) take an early return, so each wait creates exactly one timer for exactly the original duration, with no polling and no repeated callback evaluation.

Deleting memories is irreversible, so an unresponsive Stop fails in the wrong direction. No API change, no new dependency, no UI change.

none

## How it was verified

Run from `desktop/windows` on Node 22.23.2 with pnpm 11.18.0 and `--frozen-lockfile`:

- `pnpm test src/renderer/src/lib/memoriesBulk.test.ts` -> 10 passed
- `pnpm test` (full suite) -> 533 files / 5131 tests passed, 6 files and 18 tests skipped, 0 failed
- `pnpm typecheck` -> clean (node and web)
- `pnpm exec eslint` and `pnpm exec prettier --check` on both changed files -> clean

Stated explicitly, since it matters: I did not exercise this through the signed-in desktop UI. A source build cannot complete Google sign-in without Omi's production OAuth client secret, which is not in the public repo, so the running app was not driven end to end. The behavior is covered at the module boundary instead, through the same `shouldStop` callback `Memories.tsx` passes, with the delete client mocked and fake timers driving the pauses.

## Tests

7 cases added to `src/renderer/src/lib/memoriesBulk.test.ts`.

Four of them fail on `main` and pass with this change, so they are the regression tests that would have caught the bug:

- a Stop during a rate-limit pause no longer lets one more delete through (asserts exactly one request reached the server, tally `{deleted: 0, failed: 0}`, and no `onResult`)
- the rate-limit pause ends within a poll interval instead of being served out
- the pacing tail is not served at all when Stop is already set (asserts zero scheduled timers)
- the pacing tail is cut short when Stop lands after the wait has begun

The other three pass on `main` by design, as non-regression guards: the no-`shouldStop` path still schedules a single unsliced 1100ms timer, normal completion still deletes every id and paces between them, and a rate limit still gets waited out and completes when the user lets it run.

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/desktop/windows/src/renderer/src/lib/memoriesBulk.test.ts
+++ b/desktop/windows/src/renderer/src/lib/memoriesBulk.test.ts
@@ -1,10 +1,16 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 import type { Memory } from '../hooks/useMemories'
 
 const omiApiGet = vi.fn()
-vi.mock('./apiClient', () => ({ omiApi: { get: (...args: unknown[]) => omiApiGet(...args) } }))
+const omiApiDelete = vi.fn()
+vi.mock('./apiClient', () => ({
+  omiApi: {
+    get: (...args: unknown[]) => omiApiGet(...args),
+    delete: (...args: unknown[]) => omiApiDelete(...args)
+  }
+}))
 
-import { fetchAllMemories } from './memoriesBulk'
+import { fetchAllMemories, deleteMemoriesPaced } from './memoriesBulk'
 
 function page(ids: string[]): { data: Partial<Memory>[] } {
   return { data: ids.map((id) => ({ id, uid: 'u', content: id, created_at: '', updated_at: '' })) }
@@ -12,6 +18,7 @@ function page(ids: string[]): { data: Partial<Memory>[] } {
 
 beforeEach(() => {
   omiApiGet.mockReset()
+  omiApiDelete.mockReset()
 })
 
 // Fakes GET /v3/memories: hard page cap 500, no first-page expansion.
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Nice fix — this addresses a genuine safety-of-direction bug. Deleting memories is irreversible, so a Stop button that cannot actually stop during a 429 Retry-After pause (tens of seconds) is failing in the wrong direction: the loop would wake up and delete one more memory after the user had already asked it to stop.

The approach is clean: `waitInterruptible` sleeps in 250ms slices and returns as soon as `shouldStop()` flips, the 429 branch rechecks it after the pause and breaks without retrying, and an id cancelled mid-retry correctly counts as neither deleted nor failed with no `onResult` (the memory still exists). The no-`shouldStop` path (`retentionSweep.ts`) takes the early `return sleep(ms)` so its single unsliced timer is preserved — verified against the actual call site. The seven new tests with fake timers and a mocked delete client are thorough, and the four that fail on `main` are exactly the regression guards this bug needs.

Two notes, neither blocking:

1. The repo's Hygiene and PR Metadata Preflight checks are currently failing on this head. Those are metadata/invariant gates rather than code-quality failures, but they'll need to be green before merge.

2. Separately (pre-existing, called out by cubic): the 1100ms pacing tail still runs after the last id, which delays `setDeleting(false)` / the completion toast by ~1.1s for single or last-item deletes. This PR doesn't introduce that and actually makes it cancellable, so it's a reasonable follow-up rather than something to fold in here.

Leaving a positive signal — a human maintainer should be good to merge once CI passes.

---

kodjima33 (approved):
Desktop bug fix: Stop now interrupts the 429/pacing waits so a cancelled bulk delete stops immediately; covered by new tests.

PR #10982, opened 2026-08-01
fix(mcp): anchor date-only search filters to UTC day bounds

aryanorastar:
## What

The MCP conversation-search surfaces parse `YYYY-MM-DD` `start_date`/`end_date` filters with a naive `datetime.strptime(...).timestamp()`. Naive datetimes are interpreted in the **server's local timezone**, so on any host east/west of UTC the filter window silently shifts by the UTC offset, and the end date only matched up to its local midnight (dropping the rest of the end day).

Vector `created_at` values are UTC epoch seconds (`int(datetime.now(timezone.utc).timestamp())` in `database/vector_db.py`), and Firestore conversation `created_at`/action-item `due_at` are timezone-aware UTC. The naive bounds compared against the wrong instants.

## Fix

Anchor both bounds to UTC and include the full end day, matching the established integration-router convention (`integration.py`: `YYYY-MM-DD` → `00:00:00 UTC` start, `23:59:59.999999 UTC` end):

- `routers/mcp.py`: new `_date_only_to_utc_epoch` helper used by the REST `/v1/mcp/conversations/search` endpoint.
- `routers/mcp_sse.py`: `_parse_mcp_date` now returns a UTC-anchored datetime, which also fixes the `get_conversations`, `get_action_items` due-date, and `get_screen_activity` tools that shared the same naive parse.

## What I deliberately did NOT change

- The `integration.py` date handling (already correct, used as the reference convention).
- The vector/firestore storage format — the bug is entirely on the filter-parsing side.
- Date-only string *validation* behavior (malformed dates still return 400 / `-32602`).

## Guard

Regression tests (`tests/unit/test_mcp_search_date_utc.py`) stub the data layer and assert the exact UTC bounds handed to `query_vectors` / `get_conversations` / `get_action_items`. The test uses the repo's snapshot/restore isolation (`memory_import_isolation.py`) so it does not pollute sibling MCP test modules.

## Verification

- `pytest tests/unit/test_mcp_search_date_utc.py` → **6 passed**
- `pytest tests/unit/test_mcp_sse_pagination_bounds.py tests/unit/test_mcp_search_date_utc.py` → **13 passed**
- `pytest ... test_mcp_search_date_utc.py test_desktop_rest_inventory.py` → **22 passed**
- `black` on both routers + test → clean (hook reported "3 files left unchanged")

None. This only changes which conversations/action-items/screen-activity rows an MCP date filter returns; it makes the filter match the user's stated UTC day instead of the server-local interpretation.

## Failure-Class

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (dismissed):
Thanks for tightening the UTC date-bound handling here — the shared helper direction and added coverage for REST, SSE search, action items, and screen activity look like the right shape.

I need to block this head for two concrete CI issues before it can move forward:

1. `backend/utils/mcp_data.py` imports `Optional` but does not use it, and the backend unit/type job is failing on pyright `reportUnusedImport`. Please remove that unused import.
2. The repo hygiene / PR metadata preflight is failing the failure-class declaration because the commit metadata uses a long prose value (`none — ...`). That guard only accepts an existing `FC-...` slug, `new`, or exactly `none`. Please update the declaration to the accepted form.

I also could not independently run the new pytest module in this review environment because the local checkout is missing backend test dependencies (`google.api_core`), but GitHub CI already shows the actionable failures above. Once those are fixed and checks are green, this looks like a focused backend bug fix rather than a product-direction concern.

---

aryanorastar:
@Git-on-my-level both requested changes are now addressed: unused `Optional` import removed (838caec) and failure-class declarations are valid (`none` / `new`). Hygiene, Hermetic Backend E2E, Sync Cloud Tasks, and Replay Harness all passing; Backend unit suite still running. The duplicate `fix(mcp): include the full end day...` commit is a harmless rebase artifact — happy to squash if preferred. Ready for re-review.

Git-on-my-level (commented):
Thanks for the follow-up. I re-reviewed the current head and the two blockers from my previous review look resolved: the unused `Optional` import is gone, the failure-class metadata is now in an accepted form, and the relevant backend CI/checks are green.

The UTC-bound handling now looks consistent across the MCP REST search and SSE tools: date-only starts anchor to UTC midnight, end dates include the full UTC day, and the shared helper keeps the convention centralized. The added regression coverage exercises the REST vector bounds plus SSE conversations, action items, and screen activity paths, including the screen-activity summary path.

I still cannot formally approve this automatically, but from this maintainer-automation pass this looks like a focused backend bug fix and I do not see remaining blockers.

---

kodjima33 (approved):
Correct fix: naive strptime made MCP date filters server-local; UTC day-bound anchoring matches the integration-router convention, and the regression tests assert the exact bounds. Confirmed still broken on main.

Issue #10984, opened 2026-08-01
Import history rows label the wrong day and time off-UTC (job.createdAt compared as UTC against a local today)


kodjima33:
Split out of #10980, which fixed the same derivation for conversation day keys. Different subsystem, so it is not in that PR.

`app/lib/pages/settings/import_history_page.dart:451-466`:

```dart
final now = DateTime.now();
final today = DateTime(now.year, now.month, now.day);
final jobDate = DateTime(job.createdAt!.year, job.createdAt!.month, job.createdAt!.day);
```

`now` is local, so `today` is the local calendar day. `job.createdAt` is a server timestamp and parses as UTC, so `jobDate` is its **UTC** day. The two disagree for every user whose local day differs from the UTC day, which flips the row between the `todayAtTime` / `yesterdayAtTime` / absolute-date branches at the wrong boundary.

The time each branch renders has the same problem independently: `job.createdAt!.hour` / `.minute` (and the `d/m/y` fallback) are raw UTC components, so an import started at 21:00 in New York renders as `01:00` — and, via the branch above, on the next day.

Both want `job.createdAt!.toLocal()` before any component is read.

## Why it is filed rather than fixed

Found by pattern-scanning for #10980's defect, in a page with no test seam and nothing in this repo's app suite that exercises it — the #10980 PR only lands changes it can verify. It wants either a widget test over `_buildJobCard`'s date formatting or pulling the formatting into a testable helper.

PR #10985, opened 2026-08-01
fix(app): derive every conversation day key from the local day (#10980)

### LINKED ISSUE #10980: Four conversation day-key call sites derive the key from raw UTC fields instead of conversationLocalDayKey
Split out of #10976 / #10979, which fixed the same defect in `ConversationDetailProvider.conversationOrNull`.

## The contract

Conversations are bucketed by `conversationLocalDayKey(startedAt ?? createdAt)` — the **local** calendar day (`app/lib/providers/conversation_provider.dart:33`, used by `_buildGroupedByDate`). Any code that looks a conversation up by day, or hands a day to `ConversationDetailProvider.updateConversation`, must derive its key the same way.

## Sites that don't

Each reads `.year/.month/.day` straight off the timestamp. Server timestamps parse as UTC, so those are UTC fields — a different day from the group key for every user not at UTC whose conversation sits near their local midnight. Two also use `createdAt` alone where the grouping uses `startedAt ?? createdAt`.

- `app/lib/pages/chat/widgets/ai_message.dart:810` — `DateTime(data.$2.createdAt.year, …)`; the miss makes `idx == -1`, so tapping a conversation link in chat silently does nothing.
- `app/lib/pages/onboarding/conversation_created_widget.dart:18` — same expression; the onboarding hand-off selects a day that has no group.
- `app/lib/pages/memories/widgets/memory_item.dart:199-200` (`_getConversationDate`, used by both `_ensureConversationInGroup` and the `updateConversation` call at :233) — inserts the conversation under one key and can select another.
- `app/lib/pages/conversation_detail/page.dart:210-211` — the fallback when `getConversationDateAndIndex` returns null sets `selectedDate` from raw components, so the detail page opens on a day key nothing is grouped under.

`getConversationDateAndIndex` (`conversation_provider.dart:247`) already routes through the helper and is the shape the rest should match.

## Why it is filed rather than fixed

These are widget paths with no seam I could drive from a unit test, and #10979 only lands changes it can actually verify. Fixing them wants either a widget test per call site or pulling the lookup behind a testable provider method.

Found while fixing #10976 (`conversationOrNull` compared raw UTC fields against the local day key); these are the remaining instances of the same derivation.

kodjima33:
Fixes #10980.

## Bug

Conversations are bucketed by `conversationLocalDayKey(startedAt ?? createdAt)` — the viewer's **local** calendar day (`app/lib/providers/conversation_provider.dart`, used by `_buildGroupedByDate`). Five call sites rebuilt that key from the raw UTC `year/month/day` of `createdAt` instead. Server timestamps parse as UTC, so for every viewer whose local day differs from their UTC day the rebuilt key names a different day than the group key, and `ConversationDetailProvider.updateConversation` silently no-ops because nothing is grouped under it.

| Site | User-visible effect |
|---|---|
| `chat/widgets/ai_message.dart` | tapping a conversation link in a chat message did nothing (`idx == -1`) |
| `onboarding/conversation_created_widget.dart` | the onboarding hand-off selected a day with no group |
| `memories/widgets/memory_item.dart` | inserted the conversation under one key and selected another |
| `conversation_detail/page.dart` | the not-found fallback set `selectedDate` to a key nothing is grouped under |
| `utils/conversation_sync_utils.dart` | **not in the issue** — the pointer `key` is handed straight to `updateConversation` when the user taps a just-synced conversation, so it is the same defect |

Two of them also used `createdAt` alone where the grouping uses `startedAt ?? createdAt`.

## Fix

The two sites that *cannot* derive the key correctly now go through the provider instead of deriving it at all:

- `ConversationProvider.findGroupedConversationById(id)` — a chat message's conversation reference carries only an id and a UTC `createdAt`, never `startedAt`, so it has nothing to derive the key from; it looks the conversation up by id.
- `ConversationProvider.ensureConversationGrouped(conversation)` — the memory path needs the conversation both filed under and selected by one key; the method returns the key it filed under.

The remaining three use the shared `conversationLocalDayKey(startedAt ?? createdAt)`.

This is the reusable guard surface FC-day-bucket-key-recomputed asks for on recurrence (third instance, after #10398 and #10981): the call sites no longer derive the key, so they can no longer derive it wrongly.

## Tested

- `app/test.sh` — **993 tests green** on the final tree.
- New `app/test/providers/conversation_day_key_call_sites_test.dart` green in UTC, `Asia/Kolkata`, `America/New_York`, `Europe/Berlin`, `Pacific/Auckland`, `Pacific/Kiritimati`, `Pacific/Midway`. Each test sweeps all 24 hours, because a single pinned hour only trips the bug at some UTC offsets — which is how these reached main green.
- **Red proof:** with the old raw-UTC derivation restored at the two call sites reachable from a test, both fail at `TZ=America/Los_Angeles` — `updateConvoDetailProvider` leaves `selectedDate` at its `DateTime.now()` default (because `updateConversation` no-ops on a key with no group, the exact reported symptom), and the sync pointer keys on `2026-07-18` where the group is `2026-07-17`.
- The three widget-only sites are the same substitution with no test seam; verified by `flutter analyze` (clean) and inspection. Called out rather than implied.

- INV-MEM-1

## Failure class (fixes)

Failure-Class: FC-day-bucket-key-recomputed

---

Found while sweeping for this pattern, **not fixed here** (different subsystem, no way to exercise it in this run): `app/lib/pages/settings/import_history_page.dart:452-453` compares a UTC `job.createdAt` day against a local `today` and renders `job.createdAt.hour` raw, so import-history rows show the wrong Today/Yesterday label and the wrong time off-UTC. Filed separately as #10984.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/10985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/chat/widgets/ai_message.dart
+++ b/app/lib/pages/chat/widgets/ai_message.dart
@@ -806,11 +806,10 @@ class _MemoriesMessageWidgetState extends State<MemoriesMessageWidget> {
                 final connectivityProvider = Provider.of<ConnectivityProvider>(context, listen: false);
                 if (connectivityProvider.isConnected) {
                   var memProvider = Provider.of<ConversationProvider>(context, listen: false);
-                  var idx = -1;
-                  var date = DateTime(data.$2.createdAt.year, data.$2.createdAt.month, data.$2.createdAt.day);
-                  idx = memProvider.groupedConversations[date]?.indexWhere((element) => element.id == data.$2.id) ?? -1;
+                  final located = memProvider.findGroupedConversationById(data.$2.id);
 
-                  if (idx != -1) {
+                  if (located != null) {
+                    final (date, idx) = located;
                     context.read<ConversationDetailProvider>().updateConversation(data.$2.id, date);
                     var m = memProvider.groupedConversations[date]![idx];
                     PlatformManager.instance.analytics.chatMessageConversationClicked(m);
@@ -822,7 +821,7 @@ class _MemoriesMessageWidgetState extends State<MemoriesMessageWidget> {
                     setState(() => conversationDetailLoading[data.$1] = true);
                     ServerConversation? m = await getConversationById(data.$2.id);
                     if (m == null) return;
-                    (idx, date) = memProvider.addConversationWithDateGrouped(m);
+                    final date = memProvider.addConversationWithDateGrouped(m).$2;
                     PlatformManager.instance.analytics.chatMessageConversationClicked(m);
                     setState(() => conversationDetailLoading[data.$1] = false);
                     if (context.mounted) {
--- a/app/lib/pages/conversation_detail/page.dart
+++ b/app/lib/pages/conversation_detail/page.dart
@@ -208,7 +208,7 @@ class _ConversationDetailPageState extends State<ConversationDetailPage> with Ti
         provider.updateConversation(widget.conversation.id, date);
       } else {
         final effectiveDate = widget.conversation.startedAt ?? widget.conversation.createdAt;
-        provider.selectedDate = DateTime(effectiveDate.year, effectiveDate.month, effectiveDate.day);
+        provider.selectedDate = conversationLocalDayKey(effectiveDate);
       }
 
       await provider.initConversation();
--- a/app/lib/pages/memories/widgets/memory_item.dart
+++ b/app/lib/pages/memories/widgets/memory_item.dart
@@ -196,20 +196,6 @@ class MemoryItem extends StatelessWidget {
     );
   }
 
-  DateTime _getConversationDate(DateTime createdAt) {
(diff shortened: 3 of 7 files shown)


kodjima33:
Superseded by #10987, which merged the same local-day-key fix for all four call sites (main: 131c295554). This branch is now conflicting against main and its remaining delta (`findGroupedConversationById`) is not needed by the landed approach. Closing rather than leaving it open — reopen if the helper is wanted separately.

PR #10995, opened 2026-08-02
fix(app): stop retrying a recording the server refuses for good (#10975)

### LINKED ISSUE #10975: Offline sync: a recording older than 30 days fails with no explanation and burns its retry budget
## Symptom

A WAL older than `SYNC_BACKFILL_MAX_AGE_SECONDS` (2592000 = 30 days, set in `backend/deploy/runtime_env.yaml`) can never sync, and the app gives the user no reason why. It shows up as a generic `failed` after three attempts.

## Evidence

Backend side — `classify_sync_lane` sets `automatic_recovery_allowed=False` past the window (`backend/utils/sync/lanes.py:103-112`), and both sync routes reject it:

```python
if not lane_decision.automatic_recovery_allowed:
    return JSONResponse(
        status_code=422,
        content={'code': 'backfill_lookback_exceeded', ...},
    )
```

App side — `uploadLocalFilesV2` in `app/lib/backend/http/api/conversations.dart` has no 422 branch. 422 is not 200/202/400/413/429/503-capacity/>=500, so it falls through to:

```dart
throw Exception('Upload failed unexpectedly');
```

That is caught by the generic batch handler, which increments `retryCount`. At `retryCount >= 3` the WAL renders as `WalSyncDisplayState.failed` (`app/lib/services/wals/wal.dart:152`) with no message distinguishing "we cannot process this" from "this is permanently outside the recovery window".

## Impact

Anyone draining a backlog older than a month — exactly the long-offline case sync exists for — spends three upload attempts per file on a rejection that can never succeed, and ends with an unexplained failure state.

## Two decisions here, worth separating

1. **Should the 30-day cliff exist at all?** A user who left a device offline for six weeks is a normal offline-sync user, not an abuse case.
2. **Regardless of (1), the app must handle 422.** Parse `backfill_lookback_exceeded`, treat it as terminal instead of retryable, and say so in the UI.

Found while collapsing the app-side two-lane sync model; deliberately left out of that change to keep it scoped.

kodjima33:
`/v2/sync-local-files` answers `422 backfill_lookback_exceeded` for a capture
older than `SYNC_BACKFILL_MAX_AGE_SECONDS` (30 days in
`backend/deploy/runtime_env.yaml`). `uploadLocalFilesV2` had no 422 branch, so
that response fell past every case to `Exception('Upload failed unexpectedly')`
— a generic, retryable failure. The batch handler clears `isSyncing` and leaves
the WAL `miss`, which is exactly the state `syncAll()` selects from, so the
same bytes are re-offered on every sync pass forever while the row reads
"Waiting to sync" and never explains anything. Anyone draining a backlog older
than a month — the long-offline case sync exists for — pays that upload
repeatedly for a rejection that can never succeed.

Classify the bounded `code` at the HTTP boundary into a typed
`SyncRecoveryWindowExceededException`, and give the recording a terminal status
(`WalStatus.outsideRecoveryWindow`) that leaves the retry pool and renders its
own reason instead of an unexplained failure.

Which recordings that terminal applies to is the load-bearing part. The backend
decides the lookback from the OLDEST capture in the upload
(`classify_sync_lane` measures `now - oldest_capture_at`), and a batch mixes
ages, so a rejection only proves that one recording is out of range — retiring
the whole batch would strand recordings the server would still accept. So
`_retireOutsideRecoveryWindow` retires exactly the proven capture and every
recording at least as old as it, which is sound (nothing older can be inside a
window that one fell outside of) and drains a long backlog in one pass instead
of costing another doomed upload per recording. The rest of the batch stays
`miss` and re-forms without the poison on the next drain.

Only the sync attempt is terminal: the local audio file is kept, the row stays
visible and individually deletable, and Clear All still removes it. No Retry
affordance is offered, because a manual retry cannot succeed either.

Deliberately not addressed here: whether the 30-day cliff should exist at all
(decision 1 in #10975) — that is a product/ops call, and widening the window
later would need its own pass to re-arm recordings already marked terminal.

Second instance of FC-permanent-write-rejection-retried-forever, and the first
outside `backend/database/`: a rejection no retry of the same payload can
satisfy must move the row to a terminal path rather than replaying it forever.

Tested:
- `flutter test test/unit/wal_upload_resilience_test.dart` — new group drives
  the production state machine through the injected uploader seam: the batch
  and single-recording paths both land on `outsideRecoveryWindow`, `retryCount`
  stays 0, `getMissingWals()` is empty, a second `syncAll()` is a no-op, the
  audio file still exists, and a generic failure still stays `miss`. Two cases
  pin the scoping: a mixed-age batch retires only the proven-oldest and leaves
  its newer members retryable, and a recording held out of the batch by its
  conversation grouping but older than the proven capture retires in the same
  pass. Mutating the bound to retire the whole batch fails both (+15 -2), so
  they are not tautological.
- `flutter test test/widgets/sync_row_recovery_window_test.dart` — pumps the
  real `WalListItem`: the row renders "Too old to sync — Omi can't accept it"
  and offers no Retry, while a retry-exhausted recording still shows both
  "Failed — tap Retry" and its Retry button.
- `flutter test test/unit/sync_rate_limit_response_test.dart` — the classifier
  matches both sync routes' real 422 bodies and does not widen to other 422s,
  other statuses, or unparseable bodies.
- `flutter test test/providers/sync_provider_terminal_wal_state_test.dart`,
  `test/unit/wal_sync_display_state_test.dart` — provider partitioning and the
  display-state mapping.
- Full suite `bash test.sh`: 1009 passed, 0 failed.
- `make preflight`: 11/11 manifest checks pass.

Not verified against a live 30-day-old recording — that needs a real month-old
WAL against prod, which this change could not reproduce.

Fixes #10975

Failure-Class: FC-permanent-write-rejection-retried-forever

### DIFF
--- a/app/lib/backend/http/api/conversations.dart
+++ b/app/lib/backend/http/api/conversations.dart
@@ -506,6 +506,33 @@ class SyncRateLimitedException implements Exception {
   String toString() => 'SyncRateLimitedException(kind=$kind, retryAfter=$retryAfterSeconds)';
 }
 
+/// Thrown when the server permanently refuses a recording because its capture
+/// time is older than the automatic-recovery window (HTTP 422
+/// `backfill_lookback_exceeded`). No amount of retrying can make this upload
+/// succeed, so callers must stop re-offering the file instead of spending the
+/// recording's retry budget on it.
+class SyncRecoveryWindowExceededException implements Exception {
+  const SyncRecoveryWindowExceededException();
+
+  @override
+  String toString() => 'SyncRecoveryWindowExceededException()';
+}
+
+/// Classifies a sync 422 as the backend's terminal lookback rejection.
+///
+/// Keyed on the bounded `code` the sync routes emit, never on human-readable
+/// detail text: any other 422 stays a generic retryable failure.
+@visibleForTesting
+bool isSyncRecoveryWindowExceededResponse(http.Response response) {
+  if (response.statusCode != 422) return false;
+  try {
+    final body = jsonDecode(response.body);
+    return body is Map && body['code'] == 'backfill_lookback_exceeded';
+  } catch (_) {
+    return false;
+  }
+}
+
 /// A synchronous upload response that durably reached the server but did not
 /// transcribe every segment. Callers keep the local WAL and retry it; the
 /// backend deduplicates any segments that already succeeded.
@@ -611,6 +638,8 @@ Future<UploadFilesResult> uploadLocalFilesV2(
       kind: syncRateLimitKindForResponse(response),
       retryAfterSeconds: _parseRetryAfterSeconds(response),
     );
+  } else if (isSyncRecoveryWindowExceededResponse(response)) {
+    throw const SyncRecoveryWindowExceededException();
   } else if (response.statusCode >= 500) {
     throw Exception('Server is temporarily unavailable');
   }
--- a/app/lib/l10n/app_ar.arb
+++ b/app/lib/l10n/app_ar.arb
@@ -2960,6 +2960,7 @@
     "syncStatusWaiting": "في انتظار المزامنة",
     "syncStatusRetrying": "تعذّرت المعالجة — جارٍ إعادة المحاولة",
     "syncStatusFailed": "فشل — اضغط إعادة المحاولة",
+    "syncStatusTooOld": "قديم جدًا للمزامنة — لا يمكن لـ Omi قبوله",
     "syncStatusFileUnavailable": "الملف غير متوفر",
     "syncStatusUploaded": "تم الرفع · قيد المعالجة على Omi",
     "noRecordingsYet": "لا توجد تسجيلات بعد",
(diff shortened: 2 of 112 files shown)


kodjima33:
Follow-up commit \`a7e573b5cb\`: the `backfill_lookback_exceeded` handler returned without clearing `isSyncing` on the batch members it deliberately leaves retryable, so a survivor rendered as `WalSyncDisplayState.syncing` ("Syncing…", never ending) for the rest of the session and dropped out of `pendingDeletableWals` — neither waitable nor deletable. Same hole on the single-recording path when the retired set came back empty. Both now clear the in-flight flags, mirroring the rate-limit handler beside them.

Tested: the existing mixed-age batch case now asserts the survivors are not left in flight and fails (+0 -1) with only the lib hunks reverted; `flutter test` full suite 1009 passed / 0 failed; `flutter analyze` on the changed files clean; `scripts/pr-preflight` 11/11.

PR #11000, opened 2026-08-02
feat(desktop): extract knowledge graph from screen OCR locally

undivisible:
- INV-AUTH-1
- INV-CHAT-1

## Summary
- `ScreenKnowledgeGraphExtractor` turns screenshot OCR text into `local_kg_nodes/edges` via on-device Foundation Models or Gemini proxy fallback (text only, never images).

## Test plan
- [ ] `swift test --filter ScreenKnowledgeGraphExtractorTests` in CI

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11000?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-rewind.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-rewind.json
@@ -1,12 +1,12 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift": 1688,
-    "desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift": 3499,
+    "desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift": 3554,
     "desktop/macos/Desktop/Sources/Rewind/UI/RewindPage.swift": 1856
   },
   "raise_justifications": {
     "desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift": "Add full local task retrieval for persistence-time category rebase, so clearing a search before the debounce never loses an off-page pending reorder id.",
-    "desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift": "Nullable screenshot capture provenance preserves the canonical-memory device identity and prevents multi-Mac screen activity from overwriting local row IDs.",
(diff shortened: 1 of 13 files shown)


Git-on-my-level (changes requested):
Thanks for pushing this forward — the owner fencing/backfill tests are useful, and keeping image bytes out of the extraction backend is the right direction.

I’m requesting changes on two maintainability/privacy gates before this should merge:

1. The on-device → Gemini proxy selection is a privacy-sensitive fallback but is not recorded through the desktop fallback telemetry contract. `ScreenKGExtractionBackendSelector.preferredBackend()` silently moves from the local Foundation Models path to `GeminiProxyScreenKGBackend` when on-device extraction is unavailable, which means OCR text can transit the Omi/Gemini path without the shared `DesktopDiagnosticsManager.recordFallback(area:from:to:reason:outcome:)` signal required by the desktop guide. Please add the fallback-health event (and keep the reason/outcome bounded, no raw OCR/window text), and make the local-vs-proxy behavior clear enough for maintainers to audit.

2. `desktop/macos/e2e/flows/capture-lifecycle.yaml` now lists the new screen-KG extractor/backend/builder and `RuntimeOwnerIdentity` under `covers:`, but this flow still only drives the transcript capture and conversation reconciliation bridge actions. It does not exercise screen OCR KG extraction, backend selection, KG record building, or owner retargeting. Please either add a real screen-KG E2E/bridge path for those files or remove these `covers:` additions so the coverage manifest stays truthful.

Because this feature distills screen OCR into a persistent local knowledge graph and may fall back to a remote text-processing path, it also needs human maintainer/product review for the privacy/UX policy before merge. I did not formally approve for that reason.

Validation I ran locally: `python3 scripts/check_desktop_test_quality.py` passed, and `git diff --check origin/fix/agent-sync-graph-only...HEAD` passed. I could not run the Swift test target in this Linux review environment because `swift`/`xcrun` are not installed here; CI’s desktop checks are green on the current head.

---

undivisible:
Superseded by #11012; the slice was rebuilt from current main under the security/transport-* branch split.

PR #11001, opened 2026-08-02
fix(desktop,backend): remove screen-activity cloud sync egress

undivisible:
- INV-INT-1
- INV-MEM-1

## Summary
- Deletes `ScreenActivitySyncService` and `POST /v1/screen-activity/sync` so verbatim OCR no longer ships to Firestore/Pinecone.
- Removes dead MCP screen-activity read surface; keeps deprecated empty `GET /v1/mcp/screen-activity` for OpenAPI compatibility.

## Test plan
- [x] Backend unit tests pass (91 in agent env)
- [ ] Verify `POST /v1/screen-activity/sync` returns 404

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11001?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks for tightening this privacy boundary. The main direction looks good: the macOS sync loop and backend write path are removed, the MCP screen-activity tool is no longer advertised, and the remaining cloud chat tools now fail closed instead of reading a stale Firestore/Pinecone mirror.

I do need to request a small follow-up before this lands:

1. `backend/route_policy_legacy_missing_routes.txt` has a scrambled header after the reorder. It currently says CI allows entries "added without a matching ..." and then "missing entries while failing...". Please restore the original sentence ordering so the route-policy ratchet remains clear.
2. Because the deprecated `GET /v1/mcp/screen-activity` endpoint and `McpScreenActivitySummaryResponse` remain in the OpenAPI contract for compatibility, please keep a minimal response-model/contract test for the retained empty-summary shape instead of dropping that model coverage entirely.

No supply-chain concern found. This still touches privacy/security-sensitive data handling and API/tool behavior, so I am leaving it for maintainer review rather than approving formally.

---

undivisible:
Superseded by #11018; the slice was rebuilt from current main under the security/transport-* branch split.

PR #11003, opened 2026-08-02
fix(desktop): sync only knowledge graph tables, not screen capture

undivisible:
- INV-AUTH-1

## Summary
- Removes `screenshots` and `observations` from AgentSyncService tableSpecs.
- Adds `local_kg_nodes` / `local_kg_edges` with explicit column allow-lists.

## Test plan
- [ ] `swift test --filter AgentSync` in CI

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11003?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

- INV-AUTH-1

Failure-Class: none

### DIFF
--- a/backend/agent_vm/main.py
+++ b/backend/agent_vm/main.py
@@ -23,15 +23,15 @@ from fastapi import FastAPI, HTTPException, Request, WebSocket, WebSocketDisconn
 
 SYNC_TABLES = frozenset(
     {
-        "screenshots",
         "action_items",
         "transcription_sessions",
         "transcription_segments",
         "memories",
         "staged_tasks",
         "focus_sessions",
-        "observations",
         "live_notes",
+        "local_kg_nodes",
+        "local_kg_edges",
         "ai_user_profiles",
         "task_dedup_log",
     }
--- a/backend/tests/unit/test_agent_vm_protocol.py
+++ b/backend/tests/unit/test_agent_vm_protocol.py
@@ -71,7 +71,7 @@ def test_http_protocol_requires_vm_token(tmp_path: Path) -> None:
         assert client.post("/upload", content=b"x").status_code == 401
         assert client.post("/auth", json={"firebaseToken": "firebase"}).status_code == 401
         assert client.post("/ping").status_code == 401
-        assert client.post("/sync", json={"table": "screenshots", "rows": [{"id": "1"}]}).status_code == 401
+        assert client.post("/sync", json={"table": "local_kg_nodes", "rows": [{"id": "1"}]}).status_code == 401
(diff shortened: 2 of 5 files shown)


Git-on-my-level (changes requested):
Thanks @undivisible — the direction here is exactly right: narrowing agent-VM sync to derived knowledge-graph facts and keeping raw screen capture local is a meaningful privacy improvement. The column allow-list for the KG tables and the egress integration test (`AgentSyncEgressTests`) are well done. Requesting changes for two concrete blockers before this can land.

1. **Backend CI is failing.** `tests/unit/test_agent_vm_protocol.py::test_sync_groups_rows_by_present_columns` POSTs `{"table": "screenshots"}` to `/sync` and now gets HTTP 400 (expected, since `screenshots` left `SYNC_TABLES`). That test needs to move to a table still in the whitelist (e.g. `local_kg_edges`), matching the Swift test you already updated.

2. **The changelog overstates the data boundary.** Removing `screenshots`/`observations` from `SYNC_TABLES` only stops *incremental* `/sync` egress. The full `omi.db` is still gzip-uploaded to the VM via `POST /upload` on every VM start, restart, and recovery re-upload (`AgentVMService.uploadDatabase` → `/upload`), and that copy contains the entire `screenshots` table (OCR text, image paths, embeddings). So as written, the claim "screenshots, OCR text, and screen embeddings no longer leave this Mac" is not yet true end-to-end. Either extend the sanitization to the bootstrap/recovery upload path (strip screen-capture tables/columns before the gzip upload), or narrow the changelog/title to describe incremental sync only and track the bulk-upload gap explicitly.

Two smaller notes:
- The agent VM still reads the `screenshots` table directly in `semantic_search`, `daily_recap`, and the system prompt (`backend/agent_vm/main.py`). After this change those surfaces depend solely on whatever the bulk `/upload` carries — worth a one-line decision on whether that is intended or whether the VM screen tools should also be retired.
- `forbiddenSyncFieldNames` is commented as a production guarantee but is only asserted in tests; the real production enforcement is the `includedColumns` allow-list. Fine as defense-in-depth, but a comment clarifying that the allow-list is the enforceable boundary would help future readers.

Needs human maintainer review on the data-boundary scope (bulk upload vs. incremental sync) before merge.

---

undivisible:
Superseded by #11013; the slice was rebuilt from current main under the security/transport-* branch split.

PR #11004, opened 2026-08-02
fix(firmware,app): stack BLE encryption with bond-clear unpair path

undivisible:
## Summary
- Requires L2 pairing on GATT characteristics, persists bonds via `BT_SETTINGS`, extends Android `needsBond` to Omi devices.

## Test plan
- [ ] NCS build + hardware QA (pair, unpair, re-pair on iOS and Android)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11004?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/services/devices/connectors/device_connection.dart
+++ b/app/lib/services/devices/connectors/device_connection.dart
@@ -100,7 +100,8 @@ class DeviceConnectionFactory {
 
     // Use name-based detection as fallback for OmiGlass devices (some advertise as DeviceType.omi).
     final deviceName = device.name.toLowerCase();
-    final isOmiGlass = device.type == DeviceType.openglass ||
+    final isOmiGlass =
+        device.type == DeviceType.openglass ||
         deviceName.contains('openglass') ||
         deviceName.contains('omiglass') ||
         deviceName.contains('glass');
@@ -109,7 +110,7 @@ class DeviceConnectionFactory {
       case TransportKind.bluetooth:
         final deviceId = locator.bluetoothId;
         if (deviceId == null || deviceId.trim().isEmpty) return null;
-        final needsBond = device.type == DeviceType.limitless || device.type == DeviceType.omi;
+        final needsBond = !isOmiGlass && (device.type == DeviceType.limitless || device.type == DeviceType.omi);
         transport = NativeBleTransport(deviceId, requiresBond: needsBond);
         break;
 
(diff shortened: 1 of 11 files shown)


Git-on-my-level:
Thanks @undivisible — the security direction here is the right one. L2 encryption on all user-content/control characteristics, bond persistence via `BT_SETTINGS`, no key-overwrite, RPA, and the `ble_perm.h` compile-time indirection are all solid and well-reasoned. A few things need addressing before this can move forward:

**1. OmiGlass gets caught in the bond path (high)**
`needsBond` now includes `DeviceType.omi`, but OmiGlass devices that advertise as `DeviceType.omi` (the `isOmiGlass` name-based fallback explicitly supports this) get `requiresBond: true` set before the `isOmiGlass` branch selects `OmiGlassConnection`. If OmiGlass runs firmware without this Kconfig, bonding can stall and device readiness can hang. Suggest excluding OmiGlass:
```dart
final needsBond = (device.type == DeviceType.limitless || device.type == DeviceType.omi) && !isOmiGlass;
```

**2. No teardown when encryption can't reach L2 (high)**
`_transport_security_changed` logs the failure but leaves the connection alive. With every characteristic now encrypted, a stale bond / rejected pairing / full bond slot means the link stays up while every read/write/subscribe returns "Insufficient Authentication" — the Android pipeline then reports a connected device whose access silently fails. Consider disconnecting (or signaling readiness failure) when the link can't reach L2 so normal recovery runs.

**3. Firmware revision still `3.0.20` (release blocker)**
The Kconfig text describes this as the 3.0.21 encryption behavior, but `CONFIG_BT_DIS_FW_REV_STR` is still `3.0.20`. Per `omi/firmware/AGENTS.md`, the backend only offers OTAs strictly newer than the current version, so devices already on 3.0.20 won't be offered this update. Please bump the revision before this ships.

**Noted for maintainer judgment (not a request):** the initial-claim window — a freshly flashed device advertises bondable and any nearby central can pair first. `MAX_PAIRED=1` + no-overwrite protects *after* the first bond, but the first-claim race is genuinely open. Worth a tracked decision on a time-limited pairing gesture vs. always-bondable.

This is a stacked PR (base → #11006) touching the BLE auth/security surface, so it needs human maintainer + security review and the hardware QA from the test plan before merge. Flagging `security-review` / `needs-maintainer-review`.

---

undivisible:
Superseded by #11016; the slice was rebuilt from current main under the security/transport-* branch split.

PR #11005, opened 2026-08-02
fix(backend): enforce runtime allowlist on fetch_url_tool

undivisible:
## Summary
- Rejects URLs not from the current-turn user message before HTTP fetch, matching prompt scoping.

## Test plan
- [x] `pytest tests/unit/test_fetch_url_allowlist.py` (14 passed)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11005?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none


Git-on-my-level (commented):
Thanks @undivisible — the per-turn URL allowlist here is a genuinely valuable security improvement, closing a prompt-injection/exfiltration vector where retrieved content could direct outbound fetches. The direction is right. Before this can land, a few things need attention:

**CI is red (blocking).** 12 tests in `test_prompt_cache_integration.py` fail with `ModuleNotFoundError: No module named 'langchain_core.tools'`. Root cause: `agentic.py` now directly imports `from utils.retrieval.tools.web_tools import (...)`, but that test loads `agentic.py` via `_load_module_from_file` with `utils.retrieval.tools` stubbed as a package and `web_tools` not stubbed. The pre-PR code imported `fetch_url_tool` only through `tools/__init__.py`, so this new direct dependency broke the test's stub surface. The test needs `utils.retrieval.tools.web_tools` stubbed (or the import restructured).

**URL normalization mismatch (security-relevant).** `normalize_user_url` strips trailing `)` (among `. , ; : ! ? ' "`), so a valid URL like `https://en.wikipedia.org/wiki/Function_(mathematics)` has its closing paren removed before allowlisting. The model then either fails to fetch or fetches a different route than the user intended. Markdown-delimiter balancing (or at least matched-paren handling) would be safer than unconditional trailing-punct stripping.

**Asymmetric host matching.** `is_url_allowlisted` does exact-string comparison plus hostname match, but the hostname match uses `_allowlist_hostname` only for the redirect path (`is_redirect_url_allowlisted`). The initial-URL path has the same hostname logic, but the asymmetry the bots flagged is that common variants (www vs bare host, trailing slash, scheme) are rejected on entry while accepted on redirect. Worth confirming the initial check is as permissive as the redirect check for the same host.

**Fragile ContextVar fallback.** `web_tools.py` does `try: from utils.retrieval.agentic import agent_config_context / except ImportError: agent_config_context = ContextVar('agent_config', default=None)`. In normal startup this resolves correctly, but if the fallback ever fires (partial import, test reload), `fetch_url_tool` reads from a *different* ContextVar than `agentic.py` sets — silently breaking the allowlist. Consider having `agentic.py` pass the context var explicitly, or importing it lazily inside `_user_provided_urls_from_config`.

Genuinely good security work here — these are implementation issues on a sensitive surface, not design concerns.

---

Git-on-my-level (commented):
Thanks @undivisible — good progress on this. The previously raised blockers are resolved on the current head: CI is green (langchain_core.tools stub added), the execute-tool route no longer self-allowlists params.url, redirect hops are re-checked against the allowlist, and host normalization is now symmetric. The 14-test suite covers the key paths well.

Two items remain before this lands:

1. Same-host fallback on the initial fetch (security). `is_url_allowlisted` accepts any same-host variant for the model's initial tool argument, not just redirects. Consequence: if a user chooses to fetch a page on `evil.example` and that page contains prompt injection, the model can be directed to fetch `evil.example/collect?data=<retrieved-data>` — a same-host exfiltration path that bypasses the per-turn scope. Tightening the initial check to an exact normalized match (keeping same-host flexibility for redirect hops only) would close that. This is a design call for a security maintainer, not clearly a defect.

2. URLs with closing parentheses are truncated (correctness). `USER_URL_PATTERN` excludes `)` and `normalize_user_url` strips trailing punctuation, so `https://en.wikipedia.org/wiki/Function_(mathematics)` loses its closing paren before allowlisting and fetch. Balanced-paren / Markdown-delimiter handling would fix common Wikipedia-style links.

Minor: the `try/except` ContextVar fallback in `web_tools.py` is still present; in normal startup it resolves correctly, but consider importing `agent_config_context` lazily inside `_user_provided_urls_from_config` to remove the silent-divergence risk.

The direction and implementation are solid; this needs a security maintainer's sign-off on point 1 before merge.

undivisible:
Superseded by #11015; the slice was rebuilt from current main under the security/transport-* branch split.

PR #11006, opened 2026-08-02
fix(firmware): add BLE bond-clear gesture before encryption ships

undivisible:
## Summary
- Double-tap then hold 3s calls `transport_clear_bonds()` via `bt_unpair` so users can pair a new phone when `CONFIG_BT_MAX_PAIRED=1`.

## Test plan
- [ ] NCS firmware build + hardware QA required before merge with encryption branch

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11006?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks @undivisible — solid, cohesive firmware change for the bond-clear gesture needed before BLE encryption ships with `CONFIG_BT_MAX_PAIRED=1`. The refactor (extracting `transport_reset_after_disconnect()`, `transport_stop_pusher()`/`transport_start_pusher()`, and `restart_advertising()`) is clean and the error/feedback paths look right. A few observations for hardware QA / maintainer sign-off:

Race safety
- Stopping the pusher before `bt_conn_disconnect`/`bt_unpair` and always calling `transport_reset_after_disconnect()` in the clear path is the right call — it prevents GATT traffic on a half-torn-down connection and keeps `is_connected`/`MTU`/TX-semaphore consistent. Good.
- One thing to confirm on hardware: the up-to-500 ms spin-wait on `current_connection` clearing (`for (i < 10) k_sleep(50)`) relies on the BT disconnected callback firing. After the wait you call `transport_reset_after_disconnect()` which force-unrefs `current_connection`, so even if the callback never fires the pointer is cleaned — but if the disconnect genuinely hasn't propagated to the controller yet, `bt_unpair` on `BT_ID_DEFAULT, NULL` will still remove the bond while the link may still be alive. In practice this is fine (bond removal disconnects the link), but worth a hardware QA check that a bond clear during an active connection fully tears down the link and re-advertises.

Advertising-failure return value
- Returning `adv_err` from `transport_clear_bonds` after a successful `bt_unpair` is correct: bonds are cleared (the primary intent) and the caller (`clear_ble_bonds`) shows the red-LED error path only if advertising genuinely failed to restart after 5 retries. This matches the Bugbot concern.

Gesture arming
- Arming `unpair_armed` on the second press of the double-tap (both in the press-transition and the double-tap event) and expiring via `k_uptime_get()` (not the reset-prone `current_time`) is correct. Single-tap and release paths clear the arm flag.

Before merge
- Needs NCS firmware build + hardware QA as noted in the PR's own test plan — CI here doesn't build firmware. Please confirm the double-tap → hold-3s gesture triggers bond clear (blue blink ×3 + haptic), the failure path shows red, and re-pairing to a new phone works, before this lands alongside the encryption branch.
- This is a security-sensitive BLE pairing surface and can't be fully verified without hardware, so leaving for human maintainer review rather than auto-approving.

---

undivisible:
Superseded by #11017; the slice was rebuilt from current main under the security/transport-* branch split.

PR #11010, opened 2026-08-02
Restore modern macOS Home chat and stable prompt scrolling

Git-on-my-level:
## What changed

- Restores the pre-chat-first modern Home shell: one top navigation bar, no default left sidebar, and no nested Home/assistant header.
- Keeps the chat-first in-transcript behavior, rich cards, prompt materialization, typed routing, and automation lifecycle.
- Keeps the native scrollbar hidden and preserves the right-hand notched prompt timeline.
- Starts restored chats at the live edge and prevents passive AppKit geometry samples from taking scroll ownership back from the user.
- Uses the same bounded message window for the rendered transcript and prompt rail.
- Retains the legacy sidebar only when the explicit legacy Home design is enabled.
- Splits presentation-only policy/tokens out of oversized Swift files and ratchets their line-count baselines downward.

## Root cause

The chat-first merge mounted `ChatPage` inside `ChatFirstShell`, even though the shell already owned the modern top bar. That composed a second chat header/navigation layer around the transcript.

The scroll regressions had two related causes:

1. Prompt-rail visibility previously mutated AppKit scrollbar chrome during SwiftUI layout, feeding width/wrapping changes back into transcript geometry and producing the beachball loop.
2. Passive bottom-position samples could re-arm auto-follow after programmatic jumps or layout changes, so a reader scrolling down could be snapped back to an older selected position. The rail geometry also observed more messages than the transcript rendered.

## User impact

The default cohort now presents the established modern Home/chat layout while preserving only the intended chat-first content behavior. The transcript and prompt rail track one another during physical scrolling, the native scrollbar remains hidden, and manual scroll position is stable.

## Validation

- 59 focused Swift tests passed.
- 19 shell-policy tests passed after the presentation extraction.
- Cross-surface Swift and agent contract smokes passed.
- Full `scripts/pr-preflight`: 22 checks passed.
- Bounded pre-push gate passed, including desktop build, launcher suite, and 652 agent tests.
- Named QA app: `/Applications/omi-qa-20260801.app` (`com.omi.omi-qa-20260801`).
- Runtime bridge confirmed `shellVariant=chat_first`, `homeMode=chat`, `showsPrimarySidebar=false`, and `usesLegacyHomeDesign=false`.
- Physical scroll-up/down captures showed the prompt rail moving with the transcript and no delayed snap-back.
- Five-second process sample contained none of the prior `GraphHost` / `ViewGraphRootValueUpdater` loop signatures.
- Final bundle passed strict deep code-signature verification.

## Invariants

- INV-CHAT-1
- INV-UI-1


Git-on-my-level:

PR #6822, opened 2026-04-19
feat(ble): improves BLE bonding persistence

TuEmb:
Our new connectivity flow relies on OS-level bonding keys (Android/iOS) for fast reconnect.
Previously, device-side keys were effectively lost after power-off/reboot, causing reconnect and re-pair issues.

- Enable `CONFIG_BT_SETTINGS` and `CONFIG_BT_KEYS_OVERWRITE_OLDEST`
- Load app settings separately from BT settings.
- Request `BT_SECURITY_L2` on connect to trigger bonding

## Expected behavior
- Bonding keys are retained across reboot/power cycle.
- If pairing limit is reached, pairing with a new phone replaces the oldest bond.
- Reconnect after reboot is more consistent and requires less active scanning/manual recovery.

PR #11017, opened 2026-08-02
security: add safe BLE bond-clear lifecycle

undivisible:
## Summary

- Add the physical long-press bond-clear gesture.
- Make bond teardown, advertising recovery, and TX cleanup race-safe.
- Surface bond-clear failures and preserve audio/workqueue health.

## Review status

Draft pending firmware build and hardware pairing/unpair QA on supported devices.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mdmohsin7:
Thanks for this — please hold before taking it further, and please don't merge this without that conversation happening first.

Bonding on Omi devices isn't a gap; it's a decision we've already been through. BLE bond persistence was added in April (#6822, shipped in firmware 3.0.18) and then rolled back, and the app stopped requesting bonding for Omi devices in `da9f8ff46a`. There were concrete reasons for that reversal that aren't captured in the code, so re-enabling it will reproduce the problems we backed out of.

Could you check with @mdmohsin7 or @TuEmb before continuing? Happy to have the change if we've solved the underlying issue — we just need that conversation first, since neither the firmware config nor the app call site records why it was removed.

Git-on-my-level (commented):
Thanks for this — a physical bond-clear gesture is a genuinely useful safety feature, and the implementation is clean. Some notes before this leaves draft:

**Gesture arming — a couple of correctness questions**
- In `check_button_level`, `unpair_armed` is set both on the press that becomes the double-tap (line ~190) and again when `BUTTON_EVENT_DOUBLE_TAP` fires (line ~244). The long-press branch (line ~253) consumes it. This is intentional and safe, but the two arming sites make the state machine a little harder to reason about — worth a one-line comment near the long-press branch noting that the double-tap is what arms the subsequent long-press to clear bonds (rather than power-off).
- The arm window uses `k_uptime_get()` while the rest of the tap logic uses the tick-counted `current_time`. They diverge only in the arm-timeout check (line ~210), which is fine, but a future reader may wonder why two clocks are mixed. Consider documenting that the arm timeout is wall-clock-bounded on purpose.

**`transport_clear_bonds` — race/teardown**
- Good sequencing: stop pusher → disconnect → reset → unpair → reset ring buffer/sem → restart adv → restart pusher, with bounded retries. The `bt_unpair(BT_ID_DEFAULT, NULL)` correctly clears all identities.
- The disconnect-wait loop (`for i < 10 && current_connection != NULL`, 50ms each) waits up to 500ms for the disconnected callback to null `current_connection`. If the callback is delayed beyond that, `transport_reset_after_disconnect()` will still null it directly — so the unpair is safe. Confirm this holds under load, since the pusher thread is already stopped (no `tx_queue_sem`/`audio_tx_sem` consumers) before the disconnect.
- One thing to verify on hardware: `restart_advertising` retries 5× with 100ms backoff. After a fresh `bt_unpair`, the first `bt_le_adv_start` may return `-EALREADY` if the stack has not yet fully torn down the old advertising set; the `-EALREADY → 0` mapping handles that, but the 100ms spacing should be enough in practice.

**Verification (this is the main gap before merge)**
- The PR description says "Draft pending firmware build and hardware pairing/unpair QA." That is the right call. The repository Definition of Done requires exercising the real user-facing path — for firmware that means a successful build (NCS 2.9.0 sysbuild) plus an actual pairing/unpair cycle on supported hardware. There are no unit tests for this gesture, which is expected for hardware button timing, so the hardware QA pass is what carries the regression confidence.
- Please document the commands run and results in the PR body once QA is complete.

**Risk surface**
This touches BLE bonding/encryption lifecycle — a security-sensitive area — so it warrants maintainer review even though the code looks correct. Not requesting changes; the above are verification/documentation items, not blockers. Labeling `security-review` + `needs-maintainer-review` accordingly.

Appreciate the careful refactoring of `transport_off`/pusher lifecycle into reusable helpers — that made the new flow easy to follow.

---

Git-on-my-level (changes requested):
Thanks for the update. I’m going to request changes on this head because there are now two distinct maintainer blockers before this can move toward review/merge:

1. The BLE bonding direction still needs explicit human sign-off. I verified the prior history referenced above: #6822 added BLE bonding persistence, and `da9f8ff46a` later stopped requesting BLE bonding for Omi devices. Since this PR re-enters that security/UX lifecycle, please keep it in draft and get the intended bonding direction confirmed with the firmware/app maintainers before continuing.

2. The mobile internal build cadence/guard changes are unrelated to the BLE bond-clear lifecycle. The `.github/scripts/*` and `.github/workflows/mobile_internal_build.yml` changes alter release/workflow guard behavior (`cancel-in-progress` plus the guard expectations). Even if those CI changes are correct, they are a separate workflow-risk surface and should be moved out of this firmware security PR so each can be reviewed and rolled back independently.

For the firmware portion itself, the transport teardown/restart shape still looks reasonable from static review, but the merge bar remains: firmware build plus real hardware pairing/unpair QA, with the commands/devices/results documented in the PR once the product/security direction is confirmed.

I’m keeping `security-review` and `needs-maintainer-review`, and adding `workflow-review` for the new CI/release-cadence changes.

---

Git-on-my-level:
Thanks for narrowing the scope — the mobile internal-build / CI-guard changes are gone from this head, so that side of the previous change request is resolved. The diff is now cleanly the firmware bond-clear lifecycle only.

What remains is the same product/security gate that's been open since @mdmohsin7 flagged it: the BLE bonding direction itself. Bonding persistence was added in #6822 (firmware 3.0.18) and then deliberately rolled back, and the app stopped requesting bonding for Omi devices in `da9f8ff46a`. Those reasons aren't recorded in the config or the call site, so re-entering this lifecycle (even just the unpair gesture) should get explicit confirmation from @mdmohsin7 or @TuEmb that the underlying issue that caused the reversal has actually been solved. I don't see that sign-off on the PR yet, so this should stay in draft until that conversation lands and is referenced here.

On the firmware itself, the static shape still looks sound: the double-tap-then-hold arming, the dedicated `bond_clear` workqueue (so `k_msleep`/LED/haptic never block the button workqueue), the stop-pusher -> disconnect -> wait -> reset -> `bt_unpair(BT_ID_DEFAULT, NULL)` -> ring-buffer/sem reset -> bounded adv/pusher restart sequence, and the `-EALREADY`/`-ENOENT` tolerance are all reasonable. The two correctness notes from the earlier pass (a comment on the double-tap-arms-long-press intent, and noting the mixed clock for the arm timeout) are still worth adding for future readers, but they're documentation, not blockers.

Keeping `security-review` and `needs-maintainer-review`. I'm dropping `workflow-review` since that surface is no longer in the PR. Once the bonding direction is confirmed by the firmware/app maintainers and the hardware pairing/unpair QA (NCS build + real device cycle, commands/results documented) is attached, this is in good shape to leave draft.

---

Git-on-my-level (commented):
Thanks for the update — the workflow/CI split is cleanly resolved on this head (the `.github/scripts/*` and `mobile_internal_build.yml` changes from the earlier head are gone; this is now a focused 4-file firmware change). Reviewing the current state.

**Per-file notes**

- `omi.conf` — the two added comment lines document `CONFIG_BT_MAX_PAIRED=1` and reference the clear gesture. Documentation-only, accurate.
- `button.c` — the gesture state machine is consistent: double-tap arms `unpair_armed` (5s `UNPAIR_ARM_WINDOW_MS`), a subsequent long-press consumes the arm and calls `schedule_clear_ble_bonds()`; ordinary taps clear it; the timeout branch expires it. Bond clear runs on a dedicated `bond_clear` workqueue (2048 B, `K_PRIO_PREEMPT(8)`) off the button-poll path, with a `k_work_busy_get` re-entry guard. LED + haptic feedback on both success and failure. Looks correct.
- `transport.c` — `_transport_disconnected` is cleanly split into `transport_reset_after_disconnect()` (shared teardown) + a thin callback wrapper; semantics preserved. `transport_off`'s pusher start/stop is factored into reusable helpers now also used by `transport_clear_bonds()`. `transport_stop_pusher` now also gives `audio_tx_sem` and re-joins after abort — a solid robustness improvement. `transport_clear_bonds()` itself: stop pusher → disconnect active central (bounded 500 ms busy-wait) → `transport_reset_after_disconnect()` → `bt_unpair(BT_ID_DEFAULT, NULL)` → re-init ring buffer/sem → retry advertising (5×) → restart pusher (3×), with granular error accounting. The bounded disconnect wait falls through to the unconditional `current_connection = NULL` in the reset path, so no connection leak if the callback is slow. No blocking defect found in the implementation.
- `transport.h` — Doxygen for `transport_clear_bonds()`, including the `CONFIG_OMI_REQUIRE_BLE_ENCRYPTION` + `CONFIG_BT_MAX_PAIRED=1` rationale. Accurate and source-backed.

**Still gating this before it can leave draft**

1. The BLE bonding/encryption direction still needs explicit human sign-off. This PR re-enters the bonding lifecycle that `da9f8ff46a` moved away from after #6822; that's a product/security decision, not an implementation one.
2. Firmware build on this head — the `Build` check is skipped, so CI hasn't verified compilation here.
3. Hardware pairing/unpair QA on supported devices, with results documented in the PR.
4. Rebase onto current `main` (`needs-rebase`).

Keeping `security-review` and `needs-maintainer-review`. Not requesting changes — the code is sound; the gate is the product/security direction plus build + QA. The earlier CHANGES_REQUESTED about splitting the CI changes is satisfied by this head.

---
*by AI (glm-5.2) on behalf of David — BLE bonding/encryption direction decision needs maintainer sign-off before this firmware security change leaves draft.*

---

undivisible:
Pushed 063d757576 addressing the code items. Still draft — the bonding-direction sign-off is unresolved and I can't resolve it.

**Arming was actually wrong, not just hard to read.** The button-down arming site fired on any press landing inside the double-tap window, so *single tap → press-and-hold* cleared the BLE bond instead of powering the device off. Removed that site; only a completed double-tap arms.

- Gesture FSM extracted to `button_gesture.h` as a pure step function with one arming site, plus the comment that the double-tap is what turns the following long press into a bond clear, and that the arm window is deliberately wall-clock (the tick counter resets on every release, so it can't measure a span that crosses releases).
- Regression test: `omi/firmware/omi/tests/test_button_gesture.c`, run by `bash omi/firmware/omi/tests/run_button_gesture_test.sh`, registered as `firmware-button-gesture-tests` in `.github/checks-manifest.yaml` (local + ci). Restoring the button-down arming fails `tap-then-hold does not clear bonds` / `tap-then-hold powers off`.
- `omi.conf`: dropped the commented-out `CONFIG_BT_EXT_ADV_MAX_ADV_SET=2` line (dead, no rationale); the note next to `CONFIG_BT_MAX_PAIRED=1` now states the gesture.

Outstanding and not mine to close: explicit maintainer sign-off from @mdmohsin7 or @TuEmb on re-entering the bonding lifecycle after #6822 / `da9f8ff46a`, plus NCS build and hardware pair/unpair QA.

Git-on-my-level (commented):
Thanks for the update on this head (063d757) — the three new commits since ead7533 substantively improve correctness and testability.

**Arming fix (021a78ad72) — resolves my earlier arming question**
My first review on 1827a09 flagged that the two `unpair_armed` set-sites made the gesture state machine hard to reason about. That concern was well-founded: the old inline code armed bond-clear on the button-down edge of *any* press inside the double-tap window, so a single tap followed by a press-and-hold would wipe the BLE bond instead of powering the device off. The new `button_gesture.h` extracts a pure, host-compilable `button_gesture_step()` with a single arming site — only a *completed* double-tap sets `unpair_armed`; a single tap explicitly disarms; the long-press consumes the arm (clears bonds) or falls back to power-off. The wall-clock `uptime_ms` arm window is the right call, and the doc comment explains why (the tick counter resets on every release). Clean fix.

**Host test harness (7c1e2fffbb) — verified green**
I compiled `button_gesture.h` + `test_button_gesture.c` with `-std=c11 -Wall -Wextra -Werror` and ran it: all 15 assertions pass. The regression `test_tap_then_hold_powers_off` directly encodes the bug above, and `test_double_tap_then_hold_clears_bonds`, `test_arm_expires_after_window`, `test_single_tap_disarms`, and `test_long_press_fires_once` cover the rest of the arming lifecycle. The `checks-manifest.yaml` entry wires it into the local+ci lanes with the right triggers. `run_button_gesture_test.sh` is a tidy CC-detected runner with mktemp cleanup.

**Docs cleanup (063d757576)** — drops the dead adv-set line from the `omi.conf` bond-slot note. Accurate, source-backed.

**Transport (unchanged since ead7533)** — no blocking defect; `transport_clear_bonds()` does a bounded 500 ms disconnect wait, then unconditional `transport_reset_after_disconnect()`, `bt_unpair(BT_ID_DEFAULT, NULL)`, ring/sem reinit, 5x advertising retry, 3x pusher restart, with granular error accounting. One minor non-blocking note: the unconditional reset-after-wait can re-run `transport_reset_after_disconnect()` if the BT disconnect callback already executed it — harmless here because the pusher is already stopped, but a `if (current_connection != NULL)` guard around that second call would make the no-double-teardown invariant explicit.

**Still gating before this leaves draft** (unchanged):
1. BLE bonding/encryption direction — this re-enters the lifecycle `da9f8ff46a` moved away from after #6822. That's a firmware/security product decision needing maintainer sign-off, not an implementation one.
2. Firmware build on this head (`Build` check is skipped).
3. Hardware pairing/unpair QA on supported devices, documented in the PR.
4. Rebase onto current `main` (branch is 38 commits behind; `needs-rebase`).

Keeping `security-review` and `needs-maintainer-review`. Not requesting changes — the code is sound and the new tests are a real improvement; the gate is the bonding/encryption direction plus build + QA.

---
*by AI (glm-5.2) on behalf of David — BLE bonding/encryption direction decision needs firmware/security maintainer sign-off before this draft leaves draft.*

---

undivisible:
Blocking question before this and #11016 can move: why was BLE bonding rolled back?

Bonding shipped in #6822 (firmware 3.0.18), was reverted, and the app stopped requesting it for Omi devices in `da9f8ff46a`. The reason isn't recorded in code, comments, or an issue.

@mdmohsin7 @TuEmb — what broke? Whatever it was is still latent here, and re-landing the same capability without knowing it risks repeating the field incident. If nobody recalls, say so and we'll gate this behind a staged rollout with a kill switch instead of a straight merge.

Git-on-my-level:
Thanks for the cleanup on c1dff44 — dropping the dead adv-set comment for an accurate bond-slot + gesture note is the right call.

**Stale review resolution:** the workflow/CI split concern from my earlier CHANGES_REQUESTED (the `.github/scripts/*` and `mobile_internal_build.yml` changes) is fully resolved — those files are gone from this head. The only remaining `.github/` change is the benign `firmware-button-gesture-tests` manifest registration, which is appropriate. Removing `workflow-review`.

**Host test verification:** I compiled and ran `test_button_gesture.c` against the current `button_gesture.h` (cc -std=c11 -Wall -Wextra -Werror) — all 16 checks pass, including the tap-then-hold regression, arm expiry, and single-tap disarm.

**Still gating:** the BLE bonding direction decision remains open. This PR re-enters the bond-persistence lifecycle that #6822 added and `da9f8ff46a` later stopped requesting — firmware/security maintainer sign-off on the intended bonding direction is needed before this leaves draft. I am not dismissing the stale CHANGES_REQUESTED for that reason.

*by glm-5.2 on behalf of David — BLE bonding direction decision needs firmware/security maintainer sign-off before merge.*

---

undivisible:
### Head update

- Merged `origin/main` (was 38 commits behind). 
- Applied the non-blocking note from review: `transport_clear_bonds` now only calls `transport_reset_after_disconnect()` when `current_connection` is still set after the bounded disconnect wait, so the no-double-teardown invariant is explicit.

Still gating (unchanged, human decisions not code): BLE bonding/encryption direction sign-off, firmware build on this head, hardware pairing/unpair QA. Remains a draft.

undivisible:
### CI note: desktop test failures are a main regression, not this branch

`Desktop Swift Static & Test Contracts` fails on `ChatCitationTests::testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID` and `AgentPillLifecycleTests` (signal 11). Introduced by main's `#11519` (clickable citations): `RewindCitationFocusState.consume()` now gates on the owner snapshot with no test setUp, making the singleton order-dependent in the test process. This branch's desktop changes are only a changelog entry (BLE/firmware PR); the failures reproduce on any branch merging that main.

undivisible:
Verification: the Desktop Swift failures are **main-owned, not PR-owned**. None of the failing test files (`ShellModalScrimTests`, `ChatCitationTests`, `AgentPillLifecycleTests`, `ChatTimelineContinuityTests`, `GlassSurfaceReviewRegressionTests`, `MarkdownNestedScrollTests`, `ProactiveLaneClientTests`) are touched by this PR's diff, and all were introduced on `main` by #11519 and adjacent desktop work. The geometry assertions fail identically on any branch merging that main (confirmed: identical test files on both sides). No PR fix applies — tracked as a main-side regression.

Git-on-my-level (commented):
Thanks — this head (`634be236`) lands the double-teardown guard and absorbs current `main`. I re-verified the full diff and re-ran the host gesture suite on this head.

**The new guard (`634be236`, transport.c)**

`transport_clear_bonds()` now calls `transport_reset_after_disconnect()` only when `current_connection != NULL` (transport.c:1337–1342). That's the right invariant: the BT disconnected callback (`_transport_disconnected`, transport.c:705) routes through the same reset and nulls `current_connection` (transport.c:691–694), so after a callback-driven disconnect the path out of the 500 ms wait loop no longer re-runs `k_sem_init(&audio_tx_sem)`, the NUS disable, and the storage power-down logic a second time.

One theoretical edge worth a log line in the hardware QA matrix (non-blocking, pre-existing in shape): `_transport_disconnected` ignores its `conn` argument, so if the callback fires later than the bounded wait — after the self-reset case — and a new central has already connected to the restarted advertising, the late callback would tear down the new connection's state. Not asking for a change; just worth watching during pairing QA.

**Checks — the two Desktop Swift failures are not from this diff**

Verified on this head: both failing jobs fail in surfaces untouched by and identical to `main` — `ShellModalScrimTests` in `desktop/macos/Desktop/Tests/PageGlassLaneTests.swift`, the `test-install` workflow-contract fixtures, and a desktop-core fault-suite port collision (the log shows it retrying). The only `.github` change here is the manifest entry below. This branch's Desktop Swift CI was green on `c1dff44` (Aug 10) and `063d757`, and `main`'s Desktop Swift CI is green again now, so the failures came in with the Aug-14 merge of an intermediate/flaky state. A rebase onto current `main` or a re-run should clear them; I don't think they should gate firmware review.

**Per-file notes (all 8 files)**

- `omi/firmware/omi/src/lib/core/button_gesture.h` — unchanged since `063d757`; arming still only on a completed double-tap (lines 117–122), single-tap disarm (line 113), 5 s arm window measured in wall-clock `int64_t uptime_ms` (lines 87–90) so tick-counter resets can't corrupt it.
- `omi/firmware/omi/src/lib/core/button.c` — unchanged since `063d757`; `check_button_level` stays a thin dispatcher over `button_gesture_step`, and `schedule_clear_ble_bonds` runs on a dedicated 2 KB workqueue with `k_work_busy_get` re-entry guarding, so the bounded waits inside `transport_clear_bonds` never block the system workqueue.
- `omi/firmware/omi/src/lib/core/transport.c` — guard covered above; the rest of `transport_clear_bonds` (disconnect → bounded wait → guarded reset → `bt_unpair(BT_ID_DEFAULT, NULL)` → ring-buffer/semaphore reset → advertising restart with retries → pusher restart with retries) re-verified unchanged on this head.
- `omi/firmware/omi/src/lib/core/transport.h` — the `transport_clear_bonds` doc comment is accurate, including the `CONFIG_OMI_REQUIRE_BLE_ENCRYPTION` / `CONFIG_BT_MAX_PAIRED=1` context it's required for.
- `omi/firmware/omi/omi.conf` — the three added comment lines document the single bond slot and the gesture, and correctly point at `button.c -> transport_clear_bonds()`.
- `omi/firmware/omi/tests/test_button_gesture.c` — ran the suite locally on this head: 16/16 checks pass, including the tap-then-hold regression and the arm-expiry-falls-back-to-power-off case.
- `omi/firmware/omi/tests/run_button_gesture_test.sh` — compiles the harness `-std=c11 -Wall -Wextra -Werror`, uses `mktemp -d` with a cleanup trap; clean.
- `.github/checks-manifest.yaml` — the +5 entry registers `firmware-button-gesture-tests` in both `local` and `ci` lanes with accurate reason text, matching the repo convention for new checks.

**Housekeeping**

- Removing `needs-rebase` — GitHub reports the PR mergeable again after the Aug-14 merge; the label predated it (Aug 8).
- On my earlier request-changes (`aba89766`): the workflow-split concern is long resolved; the one item still standing from it is the bonding-direction sign-off below, so I'm leaving that review in place.

**Still gating before this leaves draft (unchanged from prior heads)**

1. Human maintainer sign-off on the BLE bonding/encryption direction — this PR re-enters the #6822 / `da9f8ff46a` lifecycle, and that product/security decision belongs to the firmware maintainers, not to this review.
2. Firmware build plus real hardware pairing/unpair QA on supported devices, with the commands/devices/results documented here.

The firmware work itself looks ready for that sign-off: the gesture state machine is now the cleanest it has been across this PR's history, and the teardown lifecycle is guarded and testable. Leaving `security-review` and `needs-maintainer-review` in place — the remaining blocker is the bonding/encryption direction decision, which needs a firmware maintainer's call.

---

Git-on-my-level:
Re-verified on this head (`0234c58b`). The branch was rebased onto current `main`, and I confirmed via git diff against the previously reviewed `634be236` that the firmware tree is byte-identical — a pure rebase with no new firmware commits. I also re-ran the host gesture suite on this exact head: every check passes, including the tap-then-hold regression.

**About the red checks on this head** — `Backend Hermetic Merge Gate` is red only because `Detect Hermetic Backend Scope` was cancelled mid-run before scope was decided (`SCOPE_RESULT: cancelled`, `SCOPE_APPLIES` empty; the merge-gate script then exits 1 by design). Every lane that actually completed is green. This looks like a concurrency cancellation from the force-push rebase; a fresh run should clear it.

Per-file state, re-confirmed on this head:

- `button_gesture.h` — host-testable state machine; bond-clear arms only on a completed double-tap (`BUTTON_GESTURE_DOUBLE_TAP` sets `unpair_armed`), the arm window uses wall-clock uptime so it survives the tick-counter reset on release, and a following single tap disarms.
- `button.c` — `check_button_level` delegates to `button_gesture_step`; the long-press branch routes `clear_bonds` to `schedule_clear_ble_bonds()` on a dedicated `bond_clear` workqueue, so the LED/haptic feedback with its `k_msleep`s cannot block button polling or the system workqueue; `k_work_busy_get` guards re-entry.
- `transport.c` — the disconnect callback is refactored into `transport_reset_after_disconnect()`; `transport_clear_bonds()` stops the pusher (gives both semaphores, join with abort fallback), waits bounded for the disconnect callback, runs the reset only when `current_connection != NULL` (the double-teardown guard), tolerates `-ENOENT` from `bt_unpair`, resets the TX ring buffer, and retries advertising (5x) and pusher restart (3x) with error priority preserved.
- `transport.h` — documents the `transport_clear_bonds()` contract, including the `CONFIG_BT_MAX_PAIRED=1` / `CONFIG_OMI_REQUIRE_BLE_ENCRYPTION` precondition it unblocks.
- `omi.conf` — the three comment lines document the single bond slot and the gesture path (`button.c -> transport_clear_bonds()`); accurate against the code.
- `tests/test_button_gesture.c` + `tests/run_button_gesture_test.sh` — the host suite that captured the real arming bug (tap-then-hold wiping the bond); compiles the header with `-Wall -Wextra -Werror`; green on this head.
- `.github/checks-manifest.yaml` — registers `firmware-button-gesture-tests` in both lanes; the rest of the manifest delta versus the last head is churn absorbed from `main` by the rebase, not this PR's change.

The one open item is the same human gate as before: the bonding-direction decision. Bonding shipped in #6822 and was later stopped in `da9f8ff46a`, with no recorded reason; the question has been open since Aug 10 and now gates both this PR and the stacked #11016 (encrypted BLE bonding), which builds on this branch. Code-side, everything flagged across the last five heads is addressed here. Remaining before this leaves draft: that direction call, plus the firmware build and hardware pairing/unpair QA documented in the PR.

---

PR #11025, opened 2026-08-02
app: survive mic contention in voice memo and speech profile

mdmohsin7:
Fixes a Crashlytics fatal reported from production:

```
Bad state: Microphone is busy (held by conversation)
  at ArbitratedMic.start(mic_arbiter.dart:50)
  at VoiceRecorderProvider.startRecording(voice_recorder_provider.dart:145)
```

## What happened

The `MicArbiter` throw is the intended contract — it stops a chat voice memo from hijacking the recorder out from under a live phone-mic conversation. But neither caller guarded it: `VoiceRecorderProvider.startRecording` arms its session (state=recording, PCM sink open, waveform timer running) **before** awaiting `mic.start`, so contention escaped as an unhandled async crash from a fire-and-forget tap handler and wedged the memo UI. Latent on both platforms since the two-owner arbiter split (#9398); the Android native phone-mic rollout (#9882) made the contention path actually exercised.

## The fix

- **`VoiceRecorderProvider`**: `mic.start` wrapped; on failure a dedicated unwind reverses exactly what was armed (timer, sink, PCM file, counters) and returns to idle — the chat bar drops back to the text field. No rethrow. A minimal `mic` injection seam (same pattern as the existing `transcriber` seam) makes the production flow hermetically testable.
- **`SpeechProfileProvider`**: `_initiatePhoneMicStreaming` now returns `bool`; a mic-start failure surfaces the existing `SOCKET_INIT_FAILED` connection-error dialog instead of being swallowed by the outer catch, mislabeled as a socket failure, and leaving a recording UI that receives no audio.
- **Audit**: the only other `mic.start` / `startBatch` call sites (`capture_controller.dart`) were already guarded; `mic_arbiter.dart` unchanged — the throw stays the contract.

## Verification

- New hermetic regression tests (`test/unit/voice_recorder_mic_contention_test.dart`) run the **real** `startRecording()` flow (path_provider + permission channels mocked) against a fake mic throwing the exact production `StateError`: no escape, clean return to idle, PCM temp file deleted, provider not wedged for the next attempt.
- Regression proof: temporarily replacing the guard with a rethrow makes both tests fail with the production crash string.
- Full app suite **1019/1019**; `analyze_ratchet.sh` passes.
- Not exercised on-device: the live tap-memo-during-conversation path (needs a running phone-mic conversation + chat interaction); the hermetic tests drive the identical provider code path.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11025?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/providers/speech_profile_provider.dart
+++ b/app/lib/providers/speech_profile_provider.dart
@@ -125,7 +125,10 @@ class SpeechProfileProvider extends ChangeNotifier
         await _initiateWebsocket(codec: codec, sampleRate: 16000, force: true);
 
         // Start phone mic streaming
-        await _initiatePhoneMicStreaming();
+        if (!await _initiatePhoneMicStreaming()) {
+          notifyError('SOCKET_INIT_FAILED');
+          return false;
+        }
       } else {
         // Device mode - use device codec
         device = deviceProvider?.connectedDevice;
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
Correctly guards the MicArbiter contention contract at both callers — arming the session before awaiting mic.start was the actual defect behind the Crashlytics fatal, and the unit test pins it. Holding the merge only on the repo-wide Hygiene red (release-guard/codemagic digest drift, not this PR).

kodjima33 (approved):
Confirmed still broken on main: mic.start is awaited after the waveform timer and PCM sink are armed, so MicArbiter's intended throw escapes as an unhandled async crash. Scoped unwind + regression test.

PR #11026, opened 2026-08-02
fix: deduplicate local playback transcripts

Git-on-my-level:
## Summary

Fixes duplicate local transcripts when playback from a YouTube video is captured both by the macOS system-audio tap and by the microphone.

The implementation plan was independently reviewed by Sol. The review narrowed the fix to a conservative local-only policy and explicitly rejected fuzzy text matching, cloud-path changes, and a broad acoustic-echo-cancellation migration.

## Root cause

On Apple Silicon, local Parakeet transcription runs separate mic and system-audio services. Both recognize the same playback, but the local service assigns each result a different random segment ID. The existing AppState path therefore appended both segments; it had no source-aware duplicate boundary. The mic path also lacks acoustic echo cancellation, so YouTube speech can legitimately appear in both inputs.

## Changes

- Add an exact normalized-text plus close-timestamp duplicate policy for local STT only.
- Require opposite source labels and at least four words to avoid suppressing ordinary short acknowledgements or legitimate same-source speech.
- Keep one canonical row: suppress a later mic copy, or promote a later system-audio copy in place using the existing segment ID.
- Serialize transcript persistence and flush it before stop/rotation so the replacement cannot race the original SQLite upsert.
- Await the persistence tail in both max-duration local-session rotation handlers; keep the flush barrier limited to local STT rather than implying cloud callback quiescence.
- Make the non-production capture seam drain the same selected persistence queue as production, then exercise both suppression and promotion orderings in the capture lifecycle flow.
- Leave cloud transcription, the Always setting, and the audio capture graph unchanged.
- Add focused regression tests, capture-lifecycle flow coverage, and an unreleased desktop changelog entry.
- Update the existing AppState line-count ratchet with its required justification for the lifecycle guard additions.

## Verification

- `xcrun swift test --package-path desktop/macos/Desktop --filter LocalTranscriptionDuplicatePolicyTests` — 5 passed; full Swift package/test binary compiled successfully.
- `xcrun swift test --package-path desktop/macos/Desktop --filter CanonicalGoalsStoreTests/` — 7 passed locally; the initial PR run's 300-second timeout was isolated to this unrelated suite.
- `python3 desktop/macos/scripts/check_desktop_test_quality.py` — passed.
- `python3 desktop/macos/scripts/desktop-flow-lint.py` — passed for all 65 flows and 151 registered actions.
- `OMI_APP_NAME=omi-audio-echo OMI_AUTOMATION_PORT=50919 OMI_ALLOW_ADHOC_SIGN=1 OMI_SKIP_TUNNEL=1 ./run.sh --yolo` — named bundle compiled, ad-hoc signed, installed, launched, and passed the dependency audit.
- Named-bundle capture harness injected overlapping mic/system copies and reduced the live segment count from 2 to 1; the app log recorded `[DEDUP] Promoted system-audio copy over mic playback duplicate`.
- `OMI_PR_BODY_FILE=/tmp/omi-duplicate-transcript-pr-body.md make preflight` — pass after review fixes.

## Product invariants

None affected. This preserves the existing local transcript ownership path and improves capture trust without changing the cloud protocol.

Failure-Class: none

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-appstate.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-appstate.json
@@ -1,9 +1,9 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift": 1607
+    "desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift": 1610
   },
   "raise_justifications": {
-    "desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift": "Adds shared-capture silent-mic recovery telemetry so non-Bluetooth CoreAudio rebuild attempts are visible to release-health alongside the existing Bluetooth fallback. Transcription start also resolves the user's explicit microphone choice (UID -> device ID) before building the capture service so a picked Ray-Ban Meta mic survives reconnects (port of BasedHardware/omi#9181)."
+    "desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift": "Adds local transcript duplicate suppression and persistence flushes before stop or rotation so source-aware deduplication cannot leave stale rows."
   },
   "threshold": 1500
 }
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-appstate.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-appstate.json
@@ -1,6 +1,6 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift": 1610
+    "desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift": 1608
   },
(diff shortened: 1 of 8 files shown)


Git-on-my-level:
Monitor pass verified all three cubic review threads against current head `a3a26b7`.

- **Thread 1 (P1, max-duration rotation flush):** Verified both 4-hour rotation paths now call `await self.flushTranscriptPersistence()` after local STT `finish()` and before `finishSession` (AppState+Transcription.swift:285, :1109). Resolved as addressed.
- **Thread 2 (P2, cloud rotation barrier):** Already resolved — persistence flush correctly scoped to local-STT path only.
- **Thread 3 (P3, capture-lifecycle covers):** Already resolved — flow now exercises suppression/promotion branches.

CI note: prior run failure was `CanonicalGoalsStoreTests` timing out at 300s (pre-existing flake, unrelated to this PR's transcript-dedup changes). The PR's own tests (`LocalTranscriptionDuplicatePolicyTests`) pass.

PR #11038, opened 2026-08-03
fix(mcp): reject malformed action-item text arguments

peterolkhov:
## What

MCP advertises action-item descriptions, due dates, and search queries as strings, but the tool dispatcher does not validate JSON Schema inputs before calling the shared helper. Object- or array-valued arguments therefore reached `.strip()` and raised an uncaught `AttributeError` instead of returning MCP invalid params.

This adds direct type checks at the existing shared action-item boundary. The helper now raises `ValueError`, which both MCP transports already translate into their normal client error (`-32602` for tool calls). A parameterized dispatcher regression test covers malformed description, due date, and query values.

## Scope

No dependency or general-purpose schema validator was added. The existing `mcp_action_items` helper remains the single boundary shared by REST and MCP tool calls.

## Verification

- `backend/.venv/bin/pytest -q backend/tests/unit/test_mcp_action_item_writes.py` — 43 passed
- repository pre-push gate — full backend typecheck passed; 43/43 selected backend tests passed
- `BACKEND_FAST_UNIT_WARN_SECONDS=1.0 BACKEND_FAST_UNIT_FAIL_SECONDS=1.0 bash backend/test.sh` — all changed-path tests passed; one unrelated `test_sync_v2.py` process flaked in the parallel run, then passed 168/168 with the same runner in isolation
- `OMI_PR_BODY_FILE=/tmp/omi-mcp-invalid-params-pr.md make preflight`

The real production dispatcher was exercised through its controllable database seam; a live authenticated MCP server was not required for this input-boundary regression.

none

## Failure class

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11038?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_mcp_action_item_writes.py
+++ b/backend/tests/unit/test_mcp_action_item_writes.py
@@ -142,7 +142,7 @@ if not (isinstance(_existing_fp, type) and issubclass(_existing_fp, BaseExceptio
     _api_core_exc.FailedPrecondition = type('FailedPrecondition', (Exception,), {})
 
 if not isinstance(getattr(sys.modules['database._client'], '__file__', None), str):
-    sys.modules['database._client'].document_id_from_seed = lambda seed: 'id-' + str(abs(hash(seed)) % (10 ** 12))
+    sys.modules['database._client'].document_id_from_seed = lambda seed: 'id-' + str(abs(hash(seed)) % (10**12))
 sys.modules['dependencies'].get_uid_from_mcp_api_key = MagicMock(return_value='user-1')
 sys.modules['dependencies'].get_current_user_id = MagicMock(return_value='user-1')
 sys.modules['utils.other.endpoints'].with_rate_limit = MagicMock(side_effect=lambda dependency, _policy: dependency)
--- a/backend/tests/unit/test_mcp_action_item_writes.py
+++ b/backend/tests/unit/test_mcp_action_item_writes.py
@@ -142,7 +142,7 @@ if not (isinstance(_existing_fp, type) and issubclass(_existing_fp, BaseExceptio
     _api_core_exc.FailedPrecondition = type('FailedPrecondition', (Exception,), {})
 
 if not isinstance(getattr(sys.modules['database._client'], '__file__', None), str):
-    sys.modules['database._client'].document_id_from_seed = lambda seed: 'id-' + str(abs(hash(seed)) % (10**12))
+    sys.modules['database._client'].document_id_from_seed = lambda seed: 'id-' + str(abs(hash(seed)) % (10 ** 12))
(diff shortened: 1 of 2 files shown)


Git-on-my-level (commented):
Nice fix, @peterolkhov — thanks for the contribution!

The input-validation hardening here is correct and well-targeted. Adding explicit type checks at the shared `mcp_action_items` boundary (before `.strip()`) closes the gap where a malformed MCP argument (e.g. an object-valued `description` or array-valued `due_at`) reached string methods and surfaced as an uncaught `AttributeError` instead of a clean `-32602 invalid_params`. A few notes confirming this is solid:

- Both MCP transports already translate `ValueError` → `-32602`, so the new guards surface correctly on the SSE tool path with no dispatcher change needed.
- The REST route's Pydantic models (`McpCreateActionItem` / `McpUpdateActionItem`) already enforce `str` / `datetime` at the framework level, so those paths are defensively covered without behavioral change.
- The parameterized dispatcher test exercises all three surfaces (description, due_at, query) through `execute_tool`, which is the right integration boundary.
- Keeping the checks in the shared helper rather than adding a schema validator is the right minimal call — it preserves the single-boundary design.

The cosmetic signature reformatting is harmless. This looks good and ready for a maintainer merge.

---

kodjima33 (approved):
Correct input-boundary fix: MCP object/array args reached .strip() and raised AttributeError instead of -32602; regression test included. Not merging — the Formatting (black) check is red on both changed files.

Git-on-my-level:
Thanks for the formatting follow-up, @peterolkhov — confirming the Black 26.5.1 gate (`--line-length 120 --skip-string-normalization`) passes clean on both changed files at the current head, so the formatting note from the earlier review is resolved.

The input-validation guards remain correct and minimal: non-string `description` / `due_at` / `query` now raise `ValueError` at the shared `mcp_action_items` boundary, which both MCP transports already map to `-32602 invalid_params`. The parameterized dispatcher test covers all three surfaces, and the REST Pydantic models already enforce these types, so there is no behavioral change on that path.

One note for the maintainer doing the final merge: no CI run has published on the new head commit yet, so a green check on `55fa4ea` before merge would be good to see.

---

kodjima33 (approved):
Confirmed on main: _normalize_description calls .strip() unguarded, so an object/array argument raises AttributeError instead of MCP -32602. Type check at the shared boundary + parameterized dispatcher regression test. 5/5 - merging.

PR #11047, opened 2026-08-03
Desktop: add responsiveness performance gates

Git-on-my-level:
## Summary

- Add repeatable non-production responsiveness gates for Chat-first navigation and Ask Omi, with per-interaction 100 ms acknowledgement or open-and-focus budgets
- Preserve the existing mounted-visibility `/navigate` contract by default while allowing the benchmark to measure acknowledgement separately
- Bound transcript duplicate detection to the 500 rows that can render, avoiding unbounded history work on streamed-token body updates

## Why

Long chat histories were scanned again during SwiftUI transcript updates, and the desktop harness could not reliably distinguish navigation acknowledgement from destination mounting. This made input-latency evidence incomplete and let follow-up `/state` bookkeeping obscure the measured route trace.

## Validation

- `bash desktop/macos/tests/test-omi-harness.sh`
- `python3 desktop/macos/scripts/check_desktop_test_quality.py`
- `/Users/dazheng/workspace/omi/upstream-keep-clean/backend/.venv/bin/python desktop/macos/scripts/desktop-flow-lint.py`
- `xcrun swift test --package-path Desktop --filter 'ChatMessageDeduplicatorTests|DesktopAutomationBridgeRouteTests'`
- `OMI_APP_NAME=omi-ui-responsiveness ./run.sh --yolo --fast-only --no-wait`
- Live named-bundle runs: navigation 5/5 and Ask Omi 5/5 passed on `com.omi.omi-ui-responsiveness` (automation port 47831)
- `make preflight`

## UX review

Independent Sol review approved the final diff. It confirmed no interaction or accessibility regression from visible-window deduplication and verified the acknowledgement-vs-mounted semantics and Ask Omi end-to-end focus gate.

- INV-CHAT-1

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11047?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-root.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-root.json
@@ -2,7 +2,7 @@
   "files": {
     "desktop/macos/Desktop/Sources/AuthService.swift": 2812,
     "desktop/macos/Desktop/Sources/CloudConnectorFormAutomation.swift": 1678,
-    "desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift": 4469,
+    "desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift": 4461,
     "desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift": 1515,
     "desktop/macos/Desktop/Sources/MemoryExportService.swift": 1575,
     "desktop/macos/Desktop/Sources/OmiApp.swift": 1559
--- a/desktop/macos/Desktop/Sources/DesktopAutomationBridge+ChatFirst.swift
+++ b/desktop/macos/Desktop/Sources/DesktopAutomationBridge+ChatFirst.swift
@@ -27,12 +27,12 @@ enum DesktopAutomationNavigationResponseMode: Equatable {
 
   static func snapshot<T: Sendable>(
     waitForVisibility: Bool?,
-    cached: @Sendable () async -> T,
+    cached: @Sendable () async throws -> T,
     mounted: @Sendable () async throws -> T
   ) async throws -> T {
     switch resolve(waitForVisibility: waitForVisibility) {
     case .routeAcknowledged:
-      return await cached()
+      return try await cached()
     case .mountedVisible:
       return try await mounted()
     }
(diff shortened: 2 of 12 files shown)


Git-on-my-level:
Monitor: fixed two compile errors that blocked CI (Desktop Swift Static & Test Contracts). All checks now pass.

## What was fixed

The previous commit added unknown-target validation on the fast `/navigate` acknowledgement path but introduced two compile errors:

1. `cached` closure was declared non-throwing but called `validateKnownNavigationTarget` (a throwing function)
2. `cachedAutomationSnapshot()` was referenced as `self.` member when it is a file-scope function

Fixed by making the `cached` parameter throwing and calling the free function directly. CI is now green.

## Resolved threads

- **PRRT_kwDOLkKqys6V4pGc** (Cubic P2): unknown-target returns cached snapshot as success → addressed by `validateKnownNavigationTarget` guard + `testNavigationAcknowledgementRejectsUnknownTarget`.
- **PRRT_kwDOLkKqys6V4rOU** (Codex P2): reject unacknowledged nav requests → same guard.
- **PRRT_kwDOLkKqys6V4rOZ** (Codex P2): trace selector filtering by status → the `latest` selector uses only `trace.path` + `trace.method`; `statusCode` and `durationMs` remain assertion predicates, so a failed latest response cannot be hidden.

## Threads left open (out of scope for this PR)


- **PRRT_kwDOLkKqys6V4rOP** (Codex P1, main-actor queueing): the `elapsedMs` metric intentionally measures action-internal timing (open + focus). The bridge-level `/action` route is already traced end-to-end including the actor hop, and the benchmark's `state.askOmiFocused: true` wait covers the full path. Threading a pre-actor timestamp into the action result would require changing the action registry return type — separate change.


- **PRRT_kwDOLkKqys6V4rOW** (Codex P1, harness-only path): this PR explicitly adds `app: non-prod` responsiveness gates. `openAIInput()` is the harness-only entry point by design. Benchmarking the production shortcut path is a separate effort.

PR #11058, opened 2026-08-03
feat(web): desktop parity + redesign

undivisible:
"## What

Brings `web/app` to the shape of the desktop clients and makes it real: one
destination per idea, the macOS neutral theme throughout, and the stubs gone.

**Structure.** Chat folds into Home \u2014 desktop has no chat row either, and Home
rests on the hub for an empty account and opens in the transcript once there is
history (`restingMode`, ported with tests). Conversations and recaps become one
day-grouped `/timeline` gallery, since a recap is the summary of a day and
belongs in that day. My Apps and Settings \u2192 Integrations become `/connectors`.
Profile merges into Account. `/persona` is removed. None of these are
redirected \u2014 every in-tree caller migrated in the same change.

**Shell.** The content pane is inset and rounded to the macOS window geometry.
One shared `PageToolbar` gives every page its controls, right-aligned search and
actions, and renders no page title \u2014 the sidebar already names the page. The
rail's active row is the accent fill, and since the accent is white
(`OmiColors.accent`) that is a white pill with black text, shared across rows by
`layoutId` so it slides. The profile menu opens inside the profile surface
rather than floating over it.

**Chat.** The ask bar is one pill. Its two microphones are now distinguishable
by what they are: dictation keeps the mic, live conversation wears a waveform
next to send and, while running, is drawn by the ported Omi mark metering the
room. The transcript fades only at an end that has something behind it.

**Brand.** Purple is at zero and its tokens are deleted, so it cannot grow back.

**Backend.** A chat request can name the session it targets (unknown or foreign
id 404s), and a persona PATCH touches only the fields it was sent.

- INV-MEM-1
- INV-UI-1

## Failure class (fixes)

Failure-Class: none

## Verification

- `bun run check` \u2014 typecheck clean, 198 tests (was 0 gates on this component).
- `make preflight` green.
- Exercised signed-in against the real backend on a local build: walked Home,
  Timeline, Memories, Tasks, Connectors and Settings; confirmed the sliding rail
  pill, the profile menu and its Download link, the composer pill, and the
  confetti dismissal.
- Page transitions confirmed running on all five routes by sampling the wrapper
  mid-flight; they previously showed `opacity: 1; transform: none` on arrival.
- Content pane inset measured at 12.0/12.2/11.6px against a 2209\u00d71454 viewport.

## Not verified

- Live capture and the recording stage: web capture is behind
  `RECORDING_ENABLED` with no backend deployed.
- The hub resting mode, which needs an account with no chat history. Its rule is
  unit tested and shares the component tree.

## Requires

`@tschk/moonshine` >= 0.3.7. Below it, a navigation that changes only the query
re-renders nothing, so `?section=` tab bars change the URL and not the page.

---

> [!NOTE]
> **High Risk**
> Large web navigation and UX refactor plus backend chat-session resolution and persona PATCH touch core product flows (INV-MEM-1, INV-UI-1); mis-routing or session targeting bugs would affect all clients using chat.
> 
> **Overview**
> **`web/app`** is reshaped to match desktop: chat lives on **`/home`** (hub vs transcript **`restingMode`**), conversations and recaps share one day-grouped **`/conversations`** surface, apps/integrations move to **`/connectors`**, and legacy routes like **`/chat`**, **`/recaps`**, **`/my-apps`**, **`/persona`** are removed without redirects. The shell gets macOS-style inset content, shared **`PageToolbar`**, and a sliding white accent rail; purple brand tokens are stripped and the composer/chat UI is redesigned (pill ask bar, distinct mic vs live waveform).
> 
> **CI and repo hygiene** add a **`web-app-checks`** manifest lane (`web/app/test.sh` → `bun run check`), Bun **1.3.14** in repo and web workflows (replacing npm for **`web/app`** build), track **`web/app/bun.lock`**, and document **`web/app/AGENTS.md`**. The brand UI ratchet now counts **violet/indigo/fuchsia** Tailwind ramps and judges **hex colors by purple hue** so off-palette purples cannot slip past.
> 
> **Backend** lets chat requests **name a target session** (unknown/foreign ids **404**; quota replies attach to that session) and makes **persona PATCH** field-scoped partial updates. A new failure class **`FC-implicit-resource-selection`** documents that contract.
> 
> The visible diff also includes **Dart formatting-only** churn across **`app/lib`** (no behavioral changes called out there).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58cb9fe5f76fa8dc9798b48b8ced7923bfe09f23. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Line-Count-Exception: backend/routers/apps.py | 2425 -> 2439 | persona PATCH now fields-scoped and partial-update safe; router stays near threshold pending a split track
Line-Count-Exception: backend/routers/chat.py | 1797 -> 1846 | named chat-session resolution and quota-exceeded replies threaded into sessions; split tracked separately
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift | 1614 -> 1620 | first-run new-since seed fix; negligible 4-line growth
"


Git-on-my-level:
Maintainer automation pass (comment-only because this is a draft and it touches broad product/security/dependency surfaces):

Positive signal: the web/app quality gate is a useful addition, and I was able to run it locally with Bun 1.3.14: `bun run check` passed typecheck plus 4 Bun smoke tests and 137 Vitest tests. I also ran the focused backend tests for the new channel, chat-session targeting, and persona PATCH behavior: 35 passed.

A few areas need human maintainer review before this comes out of draft / is merged:
- The messaging-channel backend adds public webhooks, new provider secrets, SMS/iMessage/Telegram delivery, media download/vision/transcription handling, and a new core-chat tool. The local unit coverage is good, but this still needs a maintainer security/config pass and at least one real signed-in/provider smoke test with non-production secrets.
- The PR adds a new dependency/runtime path for `web/app` (`@tschk/moonshine` + Bun) and removes the npm/Next config path. The new gate passed locally, but the runtime/deploy switch should be reviewed as a dependency/deployment decision, not only as a test result.
- The new `AGENTS.md` entry and `web/app/AGENTS.md` change agent behavior for future coding/review agents in `web/app`: agents will now prefer Bun/Moonshine commands, treat lint as non-blocking for `check`, and follow the desktop-parity limits documented there. That guidance looks safe and concrete, but because it affects future automated work in this area it should be kept with the final implementation if maintainers accept the runtime direction.

I’m not requesting changes from this pass; this is a positive technical signal plus a maintainer-review marker for the security/dependency/product-deployment decisions.

---

Git-on-my-level:
Incremental maintainer automation pass on the latest head (`e7915cf`):

Positive signal on the new delta since my last pass:
- `bun run check` passed with the stripped/dummy web environment: typecheck, 4 Bun smoke tests, and 137 Vitest tests.
- `bun run build` completed, including the Moonshine build and asset copy step.
- I started the generated Bun server and smoke-checked the relevant routing fix: `/` returns `302` to `/login`, and `/login` renders `200 OK` with the expected frame-ancestor protections.

No new blocker found in this incremental pass, and I’m not requesting changes. The existing maintainer-review markers still apply before merge because this remains a draft with broad messaging-channel security/configuration, dependency/runtime, and product/deployment review surfaces. The `AGENTS.md` / `web/app/AGENTS.md` agent-instruction impact is also unchanged from the prior pass: it is concrete and appears safe, but maintainers should keep it aligned with the final Bun/Moonshine direction because it will guide future AI coding/review agents in `web/app`.

---

Git-on-my-level (dismissed):
Thanks for continuing to tighten this. I’m requesting one focused change on the latest head before this should move forward:

- `web/app/AGENTS.md` and `docs/agents/web-app-destinations.md` now tell future coding/review agents that `web/app` needs `@tschk/moonshine >= 0.3.7` because earlier versions keep query-only navigations stale. But the actual dependency manifest and lockfile still resolve the Moonshine packages at `0.3.5` (`web/app/package.json`, `web/app/bun.lock`). That means the repo now documents a required runtime fix while still installing the pre-fix runtime. Please either bump the Moonshine package family/lockfile to a version that satisfies the documented floor, or narrow/remove the `>= 0.3.7` instruction if that version is not actually required.

Agent-instruction impact: this PR adds `web/app/AGENTS.md`, links it from the root `AGENTS.md`, and adds `docs/agents/web-app-destinations.md` / `docs/agents/web-app-signals.md`. Those files will directly guide future AI coding/review agents in `web/app` toward Bun/Moonshine commands, route/destination rules, and signal-state patterns. The guidance is mostly concrete and useful, but the Moonshine version mismatch would cause future agents to rely on a runtime guarantee the checked-in dependency set does not provide.

I’m preserving the existing maintainer-review markers: this remains a draft with broad messaging-channel security/configuration, dependency/runtime, and product/deployment review surfaces. No security attack/supply-chain attempt found in this pass.

---

undivisible:
Fixed in b32ef8d7d6.

`moonshine-cli` was still `^0.3.5` and dragged a whole 0.3.5 tree into the lockfile alongside 0.3.7 — its 0.3.7 tarball had been registered without being uploaded, so the bump couldn't land at the time. It's installable now; the family moves together and no 0.3.5 resolution remains in `bun.lock`.

Verified on a clean reinstall: the router fix is in the installed package and `?section=` navigation re-renders.

undivisible:
Both valid, fixed in c7a76d6ab1.

**Quota path (Codex P2)** — the session is now resolved before `_build_quota_exceeded_reply`, and both the question and the canned reply join it.

**Session ordering (cubic)** — the `created_at` point is the more serious of the two. Firestore omits documents lacking the ordered field, and `add_chat_session` writes whatever dict it is handed, so an untimestamped session became invisible and the user's history was stranded behind a newly created one. Ordering now happens after the read; untimestamped sorts oldest, id breaks ties. That also drops the composite index the ordered query needed, so the registered spec and generated manifest lose it.

Six regression tests, including a static guard that fails if `order_by` returns to the query — verified by reinstating the ordering, which fails five of the six.

The remaining cubic findings are in the messaging-channel work rather than this session/web change; I have not touched those.

undivisible:
Moonshine floor is fixed: `web/app/package.json` and `bun.lock` now resolve the family at `^0.3.7`, matching the documented requirement.

Also fixed since your review: `test_chat_session_targeting.py` asserted the resolver's query called `.order_by('created_at', DESCENDING)`, which contradicted the fix that removed it — Firestore drops documents missing the ordered field, so an untimestamped session became invisible. The test now asserts the outcome (newest of three wins) instead of the mechanism.

Remaining red check is `Backend unit suite`: `test_byok_fingerprint_pepper_migration` and `test_ws_b_short_term_lifecycle`, neither touched by this branch's backend diff.

Ready for re-review.

undivisible:
All 38 review findings addressed across 13 commits. CI green (42/42).

Two were re-diagnosed rather than fixed as filed:

- **Home tasks / legacy items** — not the client filter. `database/action_items.py` takes the Firestore equality path when `completed` is passed explicitly, while the harvest for missing/null `completed` runs only in the unfiltered branch the Tasks page uses. Home now matches the Tasks page; **the backend asymmetry is unfixed** and needs its own change.
- **`/apps` CORS** — routing through `/api/proxy` would not have worked; it 401s without an `Authorization` header and public visitors have none. Added a same-origin GET passthrough restricted to three public read paths instead.

Two changes want explicit review attention:

1. **`database/channels.py` webhook lease** — a `processing` claim is now reclaimable after 5 minutes. Events already stuck carry no `lease_expires_at`, which reads as expired, so they become reclaimable on the next delivery. Recovers lost messages, but it is a live-data semantics change.
2. **`src/app/api/proxy/public/[...path]/route.ts`** — new unauthenticated route. GET only, anchored allowlist of three paths, forwards no `Authorization`, returns only the upstream content type. No rate limiting.

Not covered by tests, deliberately: the new Swift test compiles but cannot run locally (XCTest dies on a missing Sparkle dylib, reproduces on existing unrelated tests), and the SwiftUI/Flutter state fixes have no injectable seam.

Git-on-my-level (commented):
Maintainer automation pass on the latest head (`3577ad8`):

Positive signal: the focused blocker from my earlier change request is resolved. The web app manifest and lockfile now carry the Moonshine family at the documented `>= 0.3.7` floor, and the latest GitHub checks are green.

I’m not requesting more changes from automation on this pass, but this should still get human maintainer review before merge because it is a broad product/deployment/security surface: new Telegram/iMessage/SMS channel webhooks, provider secrets and runtime bindings, media download/vision/transcription handling, and first-party chat delivery paths. The existing `security-review`, `dependency-review`, `needs-maintainer-review`, and `web` labels still look appropriate.

Agent-instruction impact: this PR adds `web/app/AGENTS.md`, links it from the root `AGENTS.md`, updates `backend/AGENTS.md`, and adds web-agent documentation under `docs/agents/`. Those files will guide future coding/review agents toward the Bun/Moonshine web-app gates, the new route/destination model, and signal-state patterns. After the Moonshine floor fix, the guidance looks concrete and aligned with the checked-in dependency/runtime state; I still want a human maintainer to bless the product/runtime direction because these instructions will shape future agent work in this area.

No security attack or supply-chain attempt found in this pass.

---

Git-on-my-level (commented):
Maintainer automation pass on the current head (`c2675895`):

Positive signal on the implementation and validation I sampled:
- `backend/database/chat.py` replaces the unordered `limit(1)` session lookup with a bounded app-scoped scan and deterministic newest-session selection, while keeping untimestamped legacy sessions visible instead of hiding them behind a Firestore `order_by`.
- `backend/routers/chat.py` now threads `chat_session_id` through read/send/delete and the quota-exceeded reply path, so explicit sessions keep both the app identity and the persisted message membership the user selected.
- `web/app/src/hooks/useChat.ts` and `web/app/src/lib/api.ts` carry the selected session into history, streaming send, and clear calls, and the request-owner guard prevents late history/stream chunks from one thread landing in another after a user switches sessions.
- `web/app/package.json`, `web/app/bun.lock`, `web/app/Dockerfile`, `.github/workflows/web-checks.yml`, and `.github/checks-manifest.yaml` consistently move the web app to Bun/Moonshine and add a manifest-selected `web/app/test.sh` quality gate; the latest GitHub checks for build, frontend lint, generated files, and backend/web suites are green.
- `web/app/src/app/api/proxy/public/[...path]/route.ts` keeps the new unauthenticated same-origin proxy constrained to exact public read paths, with a timeout and success-only cache headers, rather than making the general authenticated proxy public.

Agent-instruction impact: this PR adds/changes `AGENTS.md`, `backend/AGENTS.md`, `web/app/AGENTS.md`, `docs/agents/web-app-destinations.md`, and `docs/runbooks/canonical-memory-rollout-flags.md`. Those files will steer future coding/review agents toward the Moonshine/Bun workflow, the `/home`-centric web destination model, the removed `/chat`/`/recaps` aliases, and the canonical-memory Luna lane ownership. The guidance looks intentional and mostly source-backed by the code changes, but it is still maintainer-policy material and should be signed off by a human maintainer before merge.

I am not requesting changes from automation on this pass. I am also not formally approving because this remains a broad web redesign with backend API/session behavior, deployment/dependency/workflow changes, agent-instruction changes, and current merge conflicts; leaving it for human maintainer review and conflict resolution before merge.

— Reviewed by gpt-5.5 maintainer automation.

---

Git-on-my-level (dismissed):
Maintainer automation pass on the current head (`14ea33662e`):

I’m requesting one focused change before this moves forward:

- `backend/routers/chat.py` still deletes the explicitly selected chat session in `DELETE /v2/messages` (`chat_db.delete_chat_session(uid, chat_session_id)`), then returns `initial_message_util(uid, compat_app_id)` without that session id. That creates/returns a different current session after the clear.
- `web/app/src/hooks/useChat.ts` calls `clearMessagesApi(appId, chatSessionId)` and then discards the response while keeping the same `chatSessionId` selected. After clearing a non-default thread, the next history load or send will keep targeting the now-deleted session and get `404 Chat session not found`.
- `web/app/src/components/chat/ChatPanel.tsx` wires the clear action directly to that hook path, so the regression is user reachable from the panel rather than only theoretical. Please either keep the selected session record and clear only its messages/files, or return/adopt the replacement session id on the web client; add a focused regression test for that clear-selected-session flow.

Positive signal on the rest of the sampled current diff:
- `backend/database/chat.py` makes the current-session fallback deterministic while keeping legacy untimestamped sessions visible.
- `web/app/src/hooks/useChat.ts` otherwise has the right owner guard for late history/stream chunks when switching sessions.
- `web/app/src/app/api/proxy/public/[...path]/route.ts` keeps the unauthenticated public proxy narrowed to explicit public read paths with an upstream timeout and success-only cache headers.
- `.github/checks-manifest.yaml` adds a real `web/app/test.sh` gate, and I verified the touched brand check locally with `python -m pytest .github/scripts/test_check_brand_ui.py -q` (10 passed).

Agent-instruction impact: this PR adds `web/app/AGENTS.md`, indexes it from `AGENTS.md`, and adds `docs/agents/web-app-destinations.md` / `docs/agents/web-app-signals.md`. Those files will change future coding/review-agent behavior for the web app by steering agents to Bun/Moonshine, the `/home` destination model, the Moonshine `>= 0.3.7` floor, and signal-backed state patterns. The guidance looks accurate against the sampled package/runtime files, but because it is agent-operational guidance plus a broad web runtime/dependency change, it still needs maintainer sign-off.

Leaving formal approval blocked for the concrete selected-session clear regression plus the existing broad web runtime/dependency/workflow review surfaces.

---

Git-on-my-level:
Maintainer automation pass on the current head (`0a22698c`):

Positive signal from this pass. I re-checked the previous selected-session clear blocker and it appears resolved on this head:

- `backend/routers/chat.py` now distinguishes an explicit `chat_session_id` in `DELETE /v2/messages`: explicit clears no longer call `delete_chat_session`, and the replacement greeting is created with `initial_message_util(..., chat_session_id=chat_session_id)` so the selected thread remains addressable.
- `backend/utils/chat.py` already supports that contract by resolving an explicit session id through `get_chat_session_by_id` and adding the initial message back to that same session.
- `web/app/src/hooks/useChat.ts` now resets `historyLoadedRef` after clear and continues loading/sending with the selected `chatSessionId`; the focused `useChat` tests also cover session ownership for late history/stream responses and realtime turns.

Other sampled areas look coherent with the desktop-parity/web redesign direction:

- `backend/database/chat.py` makes current-session fallback deterministic while keeping untimestamped legacy sessions visible, and `backend/tests/unit/test_chat_session_selection.py` covers newest selection, legacy visibility, and stable tie-breaking.
- `web/app/src/lib/api.ts` threads `chat_session_id` through get/send/clear, and `web/app/src/components/home/HomePage.tsx` uses the selected chat target for Home’s transcript rather than pinning Home to the shared thread.
- `web/app/src/app/api/proxy/public/[...path]/route.ts` keeps the unauthenticated proxy constrained to explicit public read paths, adds an upstream timeout, and only caches successful non-fair-use responses.
- `web/app/package.json`, `web/app/bun.lock`, `web/app/Dockerfile`, `.github/checks-manifest.yaml`, and `.github/workflows/web-checks.yml` consistently move the web app to Bun/Moonshine and add a manifest-backed `web/app/test.sh` quality gate. I could not run the Bun web tests locally because Bun is not installed in this review container, but the GitHub checks in the context are green.
- `.github/scripts/check_brand_ui.py` now detects purple-like Tailwind ramps and hex colors by hue; I ran `python -m pytest .github/scripts/test_check_brand_ui.py -q` locally and it passed (11 passed).

Agent-instruction impact: this PR adds `web/app/AGENTS.md`, indexes it from root `AGENTS.md`, and adds `docs/agents/web-app-destinations.md` / `docs/agents/web-app-signals.md` plus related runbook guidance. These will change future coding/review-agent behavior by steering web-app work toward Bun/Moonshine, the `/home` destination model, removed-route constraints, and signal-backed state patterns. The sampled guidance looks internally consistent with the code and package/runtime changes, but because it changes agent-operational guidance along with a broad web runtime/dependency/workflow surface, it should still get human maintainer sign-off before merge.

I’m not requesting further changes from automation on this head, and I dismissed my stale prior change request for the selected-session clear issue. Formal approval remains blocked by the broad web runtime/dependency/workflow and product-review surface, so leaving this for human maintainer review.

---
*by AI on behalf of David; leaving for human maintainer review because this changes the web runtime, dependency/workflow surface, and agent guidance.*

PR #11060, opened 2026-08-03
fix(desktop): await physical mic teardown before restart

### LINKED ISSUE #10921: Desktop mic picker: reapply preferred mic on reconnect; await physical stop when switching mics live
Review findings from #10917 (desktop microphone picker / Ray-Ban Meta support) deferred because they change shared capture-lifecycle behavior that can only be verified on macOS hardware:

1. **Preferred mic is not reapplied when it reconnects.** If the selected device (e.g. Ray-Ban Meta glasses) disconnects during transcription, `AudioCaptureService` falls back to the system default and keeps observing only that route. When the device comes back, nothing re-resolves the persisted UID — the session keeps recording from the fallback while Settings still shows the preferred mic selected. Fix direction: have the capture service (or `AppState`) observe `kAudioHardwarePropertyDevices` and restart through the existing `restartTranscriptionIfNeeded()` path when the saved UID resolves again.

2. **Live mic switch races the old capture's HAL teardown.** `restartTranscriptionIfNeeded()` stops the old `AudioCaptureService` asynchronously and starts the replacement on a fixed delay; `stopCapture()` documents that HAL teardown can block for seconds, so the new IOProc can open while the old one is still alive — on Bluetooth devices this recreates the A2DP↔HFP profile-flap contention the picker is meant to avoid. Fix direction: await the old service's `waitForPhysicalStop()` before constructing the replacement, instead of the timer. This is pre-existing behavior of the shared restart path (every settings-triggered restart has this window), so it needs its own verification pass on hardware.

3. **Active PTT turn vs. ambient capture arbitration.** #10917 arbitrates the *parked* warm PTT capture before ambient transcription opens (release + await physical stop), and PTT's routing yields to devices other captures hold — but an ambient start during a *live* PTT turn (e.g. Listening enabled mid-locked-turn) still opens a second IOProc if both resolve to the same device. This overlap predates #10917 (no arbitration existed at all); governing it means deciding which owner yields mid-turn — a product behavior decision plus a hardware-verified teardown handshake.

4. **Silent-mic watchdog state is unsynchronized across threads.** The watchdog counters (`consecutiveSilentWindows`, fire bookkeeping, window peaks) are mutated by `handleAudioInput` on the HAL IO thread and reset from the main actor — both by main's pre-existing `stopCapture()`-entry reset and by #10917's warm-adoption reset. Correct fix is a lock-guarded watchdog state container (e.g. `OSAllocatedUnfairLock`) on the hot audio path, which needs hardware verification that recovery behavior is unchanged. Worst case of the current race is one missed or spurious silent-mic recovery trigger, which self-corrects on the next watchdog window.

5. **Route-change reconfiguration leaves the new device unreserved for ~300 ms.** When an active default-input capture follows a system route change from device A to device B, `handleConfigurationChange()` stops A and waits ~300 ms before `reconfigureAfterChange()` resolves and registers B. Through that window the registry still carries the owner's entry for A — so `hasActiveCapture()` stays true and the Bluetooth-contention fallback still yields — but a PTT turn checking `isDeviceActivelyCaptured(B)` sees B free and can open it just before the ambient reconfiguration does. Fix direction: a reserving registry state (owner marked "reconfiguring", or B reserved at route-change time) so routing cannot admit a second owner mid-transition; needs hardware verification with real route flaps.

Evidence: review threads on #10917 (Codex P1/P2 across four rounds). All need physical-device verification (paired Bluetooth mic, disconnect/reconnect cycles, live PTT + ambient overlap, route flaps) before landing.

aryanorastar:
## Summary

Fixes #10921.

Changing the microphone or transcription language while recording stopped the
old capture and restarted it after a fixed one-second delay. CoreAudio teardown
is asynchronous and can take longer than that, so the replacement IOProc could
start while the old IOProc still owned the device. This is especially visible
when switching Bluetooth microphone routes.

The settings restart path now waits on the existing
`AudioCaptureService.waitForPhysicalStop()` boundary before starting the
replacement transcription session. A small coordinator makes the ordering
deterministic and adds a regression test proving that `start` cannot run before
physical teardown completes.

This PR intentionally does not change preferred-device reconnect handling,
active PTT arbitration, or route reservation; those are separate #10921
failure boundaries.

## Verification

- `xcrun swift test --package-path Desktop --filter TranscriptionRestartCoordinatorTests`
  - 1 test passed.
- `./scripts/swift-format-wrapper.sh lint Desktop/Sources/MainWindow/Pages/Settings/Sections/SettingsContentView+Transcription.swift Desktop/Tests/TranscriptionRestartCoordinatorTests.swift`
  - Passed with pinned swift-format 602.0.0.
- `python3 scripts/check_desktop_test_quality.py`
  - Passed; test-quality debt remains at or below baseline.
- `OMI_APP_NAME="omi-10921-teardown" ./run.sh --yolo --full --no-wait`
  - Named bundle built, signed, installed, dependency-audited, and launched.
  - Interactive microphone switching was not exercised because the seeded
    source dev bundle is signed out in this environment; no hardware behavior
    is being claimed from the launch alone.

## Invariants

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Sections/SettingsContentView+Transcription.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/Settings/Sections/SettingsContentView+Transcription.swift
@@ -318,17 +318,31 @@ extension SettingsContentView {
   func restartTranscriptionIfNeeded() {
     guard appState.isTranscribing else { return }
 
-    // Stop and restart to apply new language settings
-    appState.stopTranscription()
-
-    // Wait a moment for cleanup, then restart
-    DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
-      self.appState.startTranscription()
-    }
+    let capture = appState.audioCaptureService
+    _ = TranscriptionRestartCoordinator.restart(
+      stop: { self.appState.stopTranscription() },
+      waitForPhysicalStop: { await capture?.waitForPhysicalStop() },
+      start: { self.appState.startTranscription() })
   }
 
   // MARK: - Notifications Section
 
+}
+
+@MainActor
+enum TranscriptionRestartCoordinator {
+  @discardableResult
+  static func restart(
+    stop: @escaping @MainActor () -> Void,
+    waitForPhysicalStop: @escaping () async -> Void,
+    start: @escaping @MainActor () -> Void
+  ) -> Task<Void, Never> {
+    stop()
+    return Task { @MainActor in
+      await waitForPhysicalStop()
+      start()
+    }
+  }
 }
 
 /// Multi-select for the languages the user speaks to the VOICE ASSISTANT (push-to-talk).
--- a/desktop/macos/Desktop/Tests/TranscriptionRestartCoordinatorTests.swift
+++ b/desktop/macos/Desktop/Tests/TranscriptionRestartCoordinatorTests.swift
@@ -1,0 +1,24 @@
+import XCTest
+
+@testable import Omi_Computer
+
+@MainActor
+final class TranscriptionRestartCoordinatorTests: XCTestCase {
+  func testRestartWaitsForPhysicalStopBeforeStartingReplacement() async {
+    var events: [String] = []
+    let (stream, continuation) = AsyncStream<Void>.makeStream()
+
+    let task = TranscriptionRestartCoordinator.restart(
+      stop: { events.append("stop") },
+      waitForPhysicalStop: {
+        for await _ in stream { break }
+      },
+      start: { events.append("start") }
+    )
+
+    XCTAssertEqual(events, ["stop"])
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
Closing as superseded by #11031, which includes the same physical-mic teardown fix plus broader restart coordination and `TranscriptionSettingsRestartTests` coverage. Please continue review/merge on #11031.

Issue #7194, opened 2026-05-06
Bug: Offline/Bluetooth-disconnected recording is not working.


hugoaap-code:
Omi AI 1.0.534 (847) — samsung SM-S938B — Android 16

Firmware 3.0.18

Bug: Offline/Bluetooth-disconnected recording is not working.

Details: The app does not record when offline or when Bluetooth is disconnected. Because no recording is created, there is nothing to sync automatically once the connection or Bluetooth is restored.

Expected: The app should keep recording offline or during Bluetooth disconnection, save the recording locally, and sync it automatically when back online/reconnected.

Impact: Users can lose important conversations when there is no internet connection or Bluetooth disconnects.

PR #11066, opened 2026-08-03
fix(app): fall back to phone capture after BLE disconnect

aryanorastar:
## Summary

Relates to #11078.

**Retargeted 2026-08-04:** this PR was originally opened against #7194, but maintainers (@mdmohsin7, @Git-on-my-level) flagged that #7194 is specifically about on-device offline recording (a separate, existing-but-flaky path), and that a phone-capture fallback is a different behavior. This PR now tracks against #11078, filed to scope that phone-fallback behavior on its own. Using "Relates to" rather than "Fixes" since the approach for #11078 itself is still pending maintainer confirmation — see the also-still-open `deviceRecord` state-gating issue raised in review.

When an active supported wearable recording lost Bluetooth, the app stopped
receiving device audio but did not start its existing automatic phone-mic batch
fallback. The missing interval was therefore lost until the wearable reconnected.

The capture controller now starts the existing automatic phone batch writer on
an unexpected BLE disconnect during an active supported recording. The batch is
finalized before device streaming resumes on reconnect, and manual stops,
unsupported devices, inactive sessions, and already-active batch sessions do
not trigger the fallback.

## Verification

- `flutter test test/unit/batch_recording_phone_test.dart`
  - 20 tests passed, including the BLE-disconnect fallback policy matrix.
- `dart format --output=none --set-exit-if-changed test/unit/batch_recording_phone_test.dart`
  - Passed.
- `flutter analyze lib/providers/device_provider.dart lib/services/capture/capture_controller.dart lib/utils/batch_recording.dart test/unit/batch_recording_phone_test.dart`
  - No new diagnostics in changed logic; the command reports one existing
    `unnecessary_import` in `device_provider.dart`.

## Invariants

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11066?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/providers/device_provider.dart
+++ b/app/lib/providers/device_provider.dart
@@ -488,6 +488,7 @@ class DeviceProvider extends ChangeNotifier implements IDeviceServiceSubsciption
     setIsConnected(false);
     updateConnectingStatus(false);
 
+    await captureProvider?.onRecordingDeviceDisconnected();
     captureProvider?.updateRecordingDevice(null);
 
     // Batch mode: the native writer finalizes the in-progress recording on
--- a/app/lib/services/capture/capture_controller.dart
+++ b/app/lib/services/capture/capture_controller.dart
@@ -1575,11 +1575,20 @@ class CaptureController extends ChangeNotifier
   }
 
   Future<void> onRecordingDeviceDisconnected() async {
+    // #7194: if the wearable is already doing its own on-device offline/batch
+    // recording, do not stop it and do not start the phone mic fallback.
+    // Phone fallback is a complementary safety net for live streaming gaps (#11078),
+    // not a replacement for on-device offline capture.
+    final bool isOnDeviceOfflineBatchActive = SharedPreferencesUtil().batchModeEnabled &&
+        _recordingDevice != null &&
+        (_recordingDevice!.type == DeviceType.limitless || _offlineSessionStartSeconds != 0);
+
     if (!shouldFallbackToPhoneOnDeviceDisconnect(
       isRecordingDevice: _recordingDevice != null,
       isRecording: isRecordingDuringDeviceDisconnect(recordingState),
       supportsBatch: phoneMicSupportsTranscribeLater && deviceSupportsTranscribeLater,
       batchAlreadyActive: _phoneMicBatchActive,
+      isOnDeviceOfflineBatchActive: isOnDeviceOfflineBatchActive,
     )) {
       return;
     }
--- a/app/lib/services/capture/capture_controller.dart
+++ b/app/lib/services/capture/capture_controller.dart
@@ -452,7 +452,7 @@ class CaptureController extends ChangeNotifier
   void _updateRecordingDevice(BtDevice? device) {
     Logger.debug('connected device changed from ${_recordingDevice?.id} to ${device?.id}');
     _recordingDevice = device;
-    if (device == null) _endOfflineSession();
+    if (device == null && !_phoneMicBatchActive) _endOfflineSession();
     notifyListeners();
   }
 
@@ -1523,6 +1523,17 @@ class CaptureController extends ChangeNotifier
     }
     if (device != null) _updateRecordingDevice(device);
 
+    // A phone batch fallback covers the interval while the wearable was out of
+    // range. Finalize it before opening the device stream again so both capture
+    // paths never own the microphone concurrently.
+    if (_phoneMicBatchActive) {
+      _micInterrupted = true;
+      ServiceManager.instance().phoneMic.stop();
+      _phoneMicBatchActive = false;
+      _endOfflineSession();
+      _micInterrupted = false;
+    }
+
     bool wasPaused = _isPaused;
 
     // Ensure even very short device recordings have a location in Redis before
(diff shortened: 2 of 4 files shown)


Git-on-my-level (changes requested):
Thanks for this — the direction is right and the predicate/test structure is clean. There's a blocking issue before this can land.

**The fallback won't trigger for active BLE device streaming.**

`onRecordingDeviceDisconnected()` gates on `recordingState == RecordingState.record`, but active BLE device streaming uses `RecordingState.deviceRecord` (set in `_initiateDeviceAudioStreaming`). No path transitions a live device recording into `.record` — that state is only used by the phone-mic flows (`streamRecording`, `_startPhoneMicBatch`, `_resumeMicRecording`). So when a wearable drops mid-stream, `shouldFallbackToPhoneOnDeviceDisconnect` evaluates `isRecording: false` and returns early. The exact scenario this PR targets is missed.

The fix should check the device-record lane:
```dart
isRecording: recordingState == RecordingState.deviceRecord ||
    recordingState == RecordingState.record,
```
(and possibly `.interrupted` if a socket drop preceded the BLE disconnect).

**Two secondary points:**

1. The predicate checks `phoneMicSupportsTranscribeLater` (iOS/Android) but not `deviceSupportsTranscribeLater`. If the intent is to fall back only for supported wearable types, adding the device-side check (`deviceSupportsTranscribeLater`) would keep the fallback scoped.

2. In the reconnect hand-off, `ServiceManager.instance().phoneMic.stop()` is fire-and-forget. The comment says the phone batch is finalized before device streaming resumes, but since `stop()` isn't awaited the ordering isn't guaranteed. If `stop()` can be made awaitable (or the hand-off serialized), that would make the transition deterministic.

The unit tests cover the predicate matrix nicely but don't exercise the `RecordingState` integration, which is why they pass despite the mismatch. Adding a test that verifies the fallback triggers from `deviceRecord` would catch this.

Flagging for maintainer review since this touches the core capture state machine.

---

mdmohsin7 (changes requested):
The tagged issue actually means recording on the device offline when not connected to bluetooth. Right now it does record offline but it breaks sometimes. This PR is not the right fix for this

aryanorastar:
@mdmohsin7 thanks — understood. I’ll treat this PR as the wrong approach for #7194 if the intended behavior is **on-device offline recording when BLE is disconnected** (which already exists but is flaky), rather than falling back to phone capture.

@Git-on-my-level your `deviceRecord` gate finding is still correct for the fallback design this PR attempted, but I’m pausing further work on that path until we confirm product intent with @mdmohsin7.

Please confirm preferred next step:
1. **Close this PR** and open a new one that hardens on-device offline recording, or
2. Keep a phone-fallback path as a separate, explicitly scoped issue.

Happy to take whichever you prefer.

aryanorastar:
Retargeting this PR per @mdmohsin7's feedback above: #7194 is specifically about on-device offline recording continuing when BLE is disconnected (which already exists but is flaky) — this PR implements a different behavior, falling back to phone-mic capture while BLE is down, which isn't the right fix for #7194.

Filed [#11078](https://github.com/BasedHardware/omi/issues/11078) to scope and track the phone-capture-fallback behavior on its own, distinct from #7194, and updated this PR's description to reference it ("Relates to #11078" rather than an auto-closing keyword, since the approach itself is still pending maintainer confirmation).

@Git-on-my-level's `deviceRecord` state-gating finding from your review still stands and is called out in the new issue — that needs to be fixed here before this can land regardless of which issue it tracks against.

Not closing or merging anything — just re-pointing this at the correct issue for continued review.

aryanorastar:
Pushed an update for the remaining CHANGES_REQUESTED on this PR:
- Fix fallback predicate to treat  (and ) as the active recording lane.
- Gate fallback with wearable Transcribe Later support in addition to phone support.
- Serialize reconnect/device handoff by waiting for phone-batch teardown via an  completer (no more fire-and-forget stop ordering).
- Added unit coverage for the RecordingState->predicate mapping ().

Could you re-review and dismiss the Changes requested once CI is green?

aryanorastar:
Update pushed for CHANGES_REQUESTED:
- deviceRecord/Interrupted treated as active lane for disconnect fallback
- supportsBatch gated with wearable Transcribe Later support
- reconnect/device handoff serialized by waiting for phone-batch onStop completer
- added unit test for RecordingState -> predicate mapping

Please re-review / dismiss once CI is green.

aryanorastar:
Closing as superseded.

- **#7194 (on-device offline batch preserve)** is now tracked by https://github.com/BasedHardware/omi/pull/11084 — that matches @mdmohsin7 / @Git-on-my-level feedback that phone-capture fallback is the wrong fix for #7194.
- **Phone-side BLE-disconnect fallback** (the behavior this PR originally implemented) is tracked by https://github.com/BasedHardware/omi/issues/11078 and will land as a fresh, clean PR with the `deviceRecord` / Transcribe Later / serialized-stop review fixes — without sharing a tip with #11084.

Sorry for the tip/history confusion on this branch.

PR #11067, opened 2026-08-03
adding AI clone (track 2) making Omi respond to people on my behalf using my apps like telegram/whatsapp/imessage using beeper

abdolotfy:
## What changed and why
Adds Track 2's AI Clone feature: Omi can draft, or in trusted chats auto-send,
replies to inbound messages on the user's behalf. Connects through Beeper
Desktop's local API rather than integrating each messaging network directly,
so WhatsApp/Telegram/iMessage/etc. are all covered through one connection.

None. Checked against docs/product/invariants/ — no changed path matches a
locked invariant's path globs (chat-continuity.md's globs cover
agentKernel/**, ipc/mainChat.ts, and hooks/useChat.ts specifically; this
feature adds a separate draft/send path that doesn't touch any of those).

## How it was verified
- `npx vitest run src/main/aiClone src/main/ipc/aiClone.test.ts
  src/renderer/src/components/chat/AiCloneDraftHost.test.tsx
  src/renderer/src/components/settings/tabs/AiCloneTab.test.tsx` — 63 tests
  passing across 10 files
- `npx tsc --noEmit -p tsconfig.node.json` / `-p tsconfig.web.json` — no new
  errors introduced
- Manually exercised the real user-facing path end to end against a live
  Beeper Desktop instance: connected a real access token, listed real chats,
  set a chat to "Draft for review," sent a live test message from another
  device, confirmed a draft appeared in the review queue within the poll
  interval, and sent it from the UI.
- Not independently verified: `messages.list`'s exact ordering guarantee from
  Beeper's API is assumed newest-first per its documentation; this held in
  manual testing but wasn't stress-tested against a chat with high message
  volume or out-of-order delivery.

## Tests
Feature — core path and main error paths both covered:
- `beeperClient.test.ts` — API client mapping and error propagation on a bad
  token
- `chatMonitor.test.ts` — new-message detection against a cursor
- `autoReplyPolicy.test.ts` — send/queue/skip decision, including the
  sensitive-topic and needs-input gates that override auto-send
- `ipc/aiClone.test.ts` — end-to-end poll → detect → broadcast, and the
  submit-draft policy routing
- `AiCloneDraftHost.test.tsx` — draft generation and graceful handling of a
  failed model call
- `AiCloneTab.test.tsx` — connect flow (success and failure), mode changes,
  draft approve/dismiss

## Failure class (fixes)
Failure-Class: none

None added.

## Scoped cleanups (optional)
None.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11067?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks for putting this together — the direction is interesting, and I appreciate the tests around the new AI Clone path.

I’m going to request changes before this can move forward. This feature can send messages as the user through Beeper, so the safety/privacy bar is much higher than for a normal desktop feature. The main blockers I see on this head are:

- The Beeper token, per-chat settings, and queued drafts are stored at the install/userData level and are not tied to the signed-in Omi account or cleared as part of account switch/sign-out teardown. That can expose a previous user’s messaging account/drafts to the next Omi account on the same machine.
- `aiClone:setChatMode`/persisted settings are not runtime-validated, and `decideReplyAction` treats any value other than `off`/`draft` as auto-send. Unknown or corrupted state needs to fail closed to `off`, never to sending.
- The draft prompt is flattened into one string before calling the model, so untrusted chat/persona/history content can compete with the intended system instructions. For an auto-send path, please preserve instruction boundaries and treat remote messages as quoted data.
- Polling advances the cursor before draft generation/submission succeeds and has no in-flight/cancellation guard. Failures, overlapping polls, or disconnects can skip messages permanently or duplicate/continue sends after the user’s intent changed.
- Draft approval/auto-send paths are not idempotent against overlapping submits/clicks, so the same draft/message can be sent more than once.

Please address these before maintainer review: scope/clear all AI Clone state with the authenticated account lifecycle, validate all IPC and persisted modes, fail closed for ambiguous send decisions, make cursor advancement acknowledge successful processing, add in-flight/cancel guards, and make sends/draft approvals idempotent. I’d also like a human maintainer/product review before merge because auto-replying as the user is a sensitive product surface.

---

abdolotfy:
Thanks for the detailed review — addressed all five points below. Some required more digging than expected; noting where a fix took more than one attempt to get right.

1. Account/session teardown — Fixed. My first attempt wired this into the explicit data-wipe path (wipeUserData), which doesn't actually fire on an ordinary sign-out. Found this via a second review and re-wired it into onSessionReset, the hook that fires on every real sign-out (same one used for the app's other per-account cleanup). Verified with a test that drives the actual setBackendSession(null) sign-out path, not just the teardown function in isolation.

2. Fail-closed mode validation — Fixed. decideReplyAction is now an explicit switch with a skip default for anything unrecognized. Validated at both the IPC boundary and on every read of the persisted settings file (a corrupted/hand-edited file gets coerced to off, never trusted).

3. Prompt/instruction boundaries — Partially addressed, flagging honestly rather than claiming it's solved: untrusted content is now fenced (<untrusted_chat_content> tags with delimiter-escaping) and the system prompt hardened against embedded instructions. That said, the renderer's model-call path only accepts a single flattened string, so there's no true system/user role separation at the API level — this is a real mitigation, not a guarantee.

4. Cursor advancement / in-flight guards — Fixed for the core race: cursor now only advances after a message's outcome (sent/queued/skipped) is confirmed, never optimistically at poll time. Added a poll re-entrancy guard and an in-flight tracker with a staleness fallback so a crashed/failed draft doesn't lose a message forever. One narrower gap I want to be upfront about: a poll already mid-flight when a disconnect happens can still broadcast a message to the renderer before the session-generation check catches it at the submit step — sending itself is blocked, but that specific broadcast-after-disconnect window isn't fully closed yet.

5. Idempotent approval/send — Fixed. Draft approval does an atomic find-and-remove before sending (not after), and both the approval and submission paths have concurrency guards, so overlapping clicks/calls can't cause a duplicate send. Verified with tests that fire genuinely concurrent calls at the same draft/message.

Happy to keep closing out the remaining partial items, or take a checkpoint here if you'd rather review what's landed so far.

Git-on-my-level (changes requested):
Thanks for the follow-up fixes here — several of the earlier lifecycle and idempotency issues are moving in the right direction. I still need to keep this in changes-requested because the remaining gaps are on the path that can send messages as the user.

Blocking items on this head:

- The auto-send safety gate is still too narrow for this risk level. `decideReplyAction` only looks for a small set of sensitive words in the generated draft, so an incoming message can ask for a sensitive commitment and the model can produce a short agreement like “sounds good” or “I agree” that bypasses the gate. For auto-send, please fail closed when either the incoming request or the draft is uncertain/sensitive, or route through a stronger explicit policy check before `sendMessage`.
- The ordinary Disconnect path is not fail-safe. `aiClone:disconnect` calls `tokenStore().clear()` before bumping the session generation and stopping polling; if the file removal throws, the IPC call rejects before the in-memory client/poll loop is invalidated. The shutdown/cancellation state should be updated in a `finally`-style path so a failed token-file delete cannot leave AI Clone active.
- Multiple inbound messages from the same poll are still drafted/submitted concurrently in `AiCloneDraftHost`. In an auto-send chat, a newer reply can be sent before an older one if its model call returns first. Please serialize processing per chat or otherwise preserve the poll/message order before any send path.

This should remain human maintainer/product reviewed before merge: auto-replying through Beeper as the user is a sensitive product surface, and the dependency/lockfile addition should also get normal dependency review.

---

Git-on-my-level (changes requested):
Thanks for continuing to tighten this up. There are some good improvements here — the new `sessionGeneration` checks in `desktop/windows/src/main/ipc/aiClone.ts`, the fail-closed mode validation in `desktop/windows/src/main/aiClone/autoReplyPolicy.ts` / `chatSettingsStore.ts`, and the tests around the stores/IPC are all moving in the right direction.

I still need to request changes on this head because there are a few concrete blockers left on the path that can send messages as the user:

- `desktop/windows/package-lock.json` is not in sync with `desktop/windows/package.json`. A clean `npm ci --ignore-scripts --no-audit --no-fund` fails immediately because the lockfile resolves packages that do not satisfy the manifest (for example `axios@1.19.0` vs `package.json`'s `axios~1.18.0`, and `esbuild@0.27.7` vs `esbuild~0.28.1`). This blocks reproducible install/test/build for the Windows desktop package, and the new Beeper dependency needs a clean lockfile story before dependency review can complete.

- `desktop/windows/src/shared/types.ts` removes `attemptId` from the `meeting-capture-start` / `meeting-capture-stop` / `meeting-capture-status` message shapes and removes the `audio-source-ready` event, but existing code still uses those fields/events. For example `desktop/windows/src/main/meeting/meetingMonitor.ts` still sends `attemptId` with `meeting-capture-start` and `meeting-capture-stop`, and `desktop/windows/src/main/ipc/captureBridge.ts` / `captureBridge.test.ts` / `renderer/src/lib/omiListenClient.test.ts` still refer to `audio-source-ready`. This looks unrelated to AI Clone and should either be reverted or carried through all call sites with tests.

- The ordinary disconnect path in `desktop/windows/src/main/ipc/aiClone.ts` is still not fail-safe: `aiClone:disconnect` calls `tokenStore().clear()` before `bumpSessionGeneration()` and `stopAiClonePolling()`. If the file deletion throws, the IPC rejects before invalidating the cached client/session generation and before stopping the poller. Please update cancellation/invalidation in a `finally`-style path so a token-store failure cannot leave AI Clone active.

- The auto-send safety decision is still too narrow in `desktop/windows/src/main/aiClone/autoReplyPolicy.ts` and its use from `desktop/windows/src/main/ipc/aiClone.ts`. `decideReplyAction` only inspects the generated draft text (plus `[NEEDS_INPUT]`), not the incoming message/request or structured risk metadata. An incoming message can ask for a sensitive commitment and the model can produce a harmless-looking short agreement that bypasses the keyword gate. For auto-send, please fail closed when the incoming request is sensitive/uncertain too, or route through a stronger explicit policy check before `sendMessage`.

- `desktop/windows/src/renderer/src/components/chat/AiCloneDraftHost.tsx` still processes every incoming event with `void draftAndSubmit(event)`, so multiple messages from the same chat can draft and submit concurrently. In an `auto_send` chat, a later model call can finish first and send out of conversation order. Please serialize per chat (or otherwise preserve poll/message order before any send path).

I could not run the focused desktop tests because the clean install fails at `npm ci` before dependencies are available. Given the user-impersonation/security sensitivity, the dependency/lockfile failure, and the remaining ordering/disconnect/safety issues, this should stay changes-requested and get human maintainer/product sign-off before merge.

---
*by AI on behalf of David — leaving this for human maintainer review because auto-replying as the user through Beeper is a security- and product-sensitive surface.*

Git-on-my-level (changes requested):
Thanks for the continued iteration here. This head does resolve several of the earlier AI Clone safety/lifecycle concerns I had called out: `desktop/windows/src/main/ipc/aiClone.ts` now uses `sessionGeneration` on draft submit/approval, `aiClone:disconnect` invalidates in a finally-style path, `desktop/windows/src/main/aiClone/autoReplyPolicy.ts` checks the incoming message as well as the draft, and `desktop/windows/src/renderer/src/components/chat/AiCloneDraftHost.tsx` serializes drafting per chat.

I still need to keep this in changes-requested for two concrete blockers that remain:

- `desktop/windows/package-lock.json` is still not synchronized with `desktop/windows/package.json`. A clean `npm ci --ignore-scripts --no-audit --no-fund` fails before tests can run with `EUSAGE` because the lockfile resolves packages that do not satisfy the manifest (for example `axios@1.19.0` vs the manifest's `axios~1.18.0`, and `esbuild@0.27.7` / `@esbuild/*@0.27.7` vs the manifest's `esbuild~0.28.1`). The lockfile root also reports `version: 1.0.31` while the manifest is `1.0.33`. Please regenerate/commit the correct lockfile so the Windows desktop package can be installed and the new Beeper dependency can be reviewed reproducibly.

- `desktop/windows/src/shared/types.ts` still changes unrelated capture/meeting IPC contracts without updating the call sites. It removes `attemptId` from the `meeting-capture-start` / `meeting-capture-stop` / `meeting-capture-status` shapes, narrows meeting status values, and removes the `audio-source-ready` event; meanwhile `desktop/windows/src/main/meeting/meetingMonitor.ts` still sends/compares `attemptId`, `desktop/windows/src/renderer/src/capture/MeetingSessionHost.tsx` still reads `cmd.attemptId` on stop, and `desktop/windows/src/main/ipc/captureBridge.ts`, `desktop/windows/src/renderer/src/capture/AudioSessionHost.tsx`, and `desktop/windows/src/renderer/src/lib/omiListenClient.ts` / their tests still use `audio-source-ready`. This looks unrelated to AI Clone and will continue to break typechecking once install is fixed; please revert that shared-type churn or carry it through every call site with tests.

I could not run the focused desktop test suite because the clean install currently fails at `npm ci`. Given that this feature can send messages as the user through Beeper and introduces a new dependency, it still needs product/security/dependency maintainer sign-off before merge.

---
*by AI on behalf of David — leaving this for human maintainer review because auto-replying as the user through Beeper and the new desktop dependency need maintainer sign-off.*

Git-on-my-level:
Thanks for the PR, I think the engineering quality here is good, but it doesn't really meet the bar of a useful AI clone

Here's my definition: A twin that texts as you needs presence, per-relationship voice, current knowledge, and judgment about when to stay quiet. This has a poller, a daily bio, and a keyword gate.

- Voice is not your texting voice. Persona is getLatestProfileText(): a once-daily, <10k-char synthesis of Omi memories, tasks, goals, necklace conversation summaries, and Omi-app chat. It is not “how I text Mom vs my boss,” and it never trains on Beeper history. Thread context is the last six messages. In groups, every other person is labeled with the chat title, because senderName is setting.displayName.
- It is forbidden from doing the useful middle. The prompt says do not invent plans, availability, or commitments. Auto-send then only safely covers “yeah / haha / ok” — messages that do not need a twin. The messages where a twin would earn its keep (scheduling, favors, relationship maintenance) are exactly what the design refuses, or guesses.
- Judgment is a regex, not a twin. wire transfer and lawsuit get queued. “Venmo me $40,” “I’ll pick you up at 8,” “tell her I love her,” “yes I can host” do not. Reviewers already caught the “sounds good” hole for a short list of scary words; the same hole remains for everything not on that list.
- Presence is wrong for chat. 60s poll, Beeper Desktop must be running, Windows only. People do not experience a twin that answers a minute later from a second app they had to install. The internal consolidation PR (#10933) at least uses a listen loop and ships macOS; even that still requires approval in its summary, which is an admission that auto-send is the dangerous part, not the value.

So you get a paradox: auto-send is only safe when the reply does not matter, and only useful when it does. Draft-for-review escapes that paradox — a cross-network “reply later” queue is a real product for someone already in Beeper. That is an assistant with a send button, not a twin.

Git-on-my-level (changes requested):
Thanks for keeping the branch alive — I checked the three commits pushed since my last review (`6eeea93`, `6382379`, `a023cd43`): all three are the weekly guardrail-pulse bot chore touching only `.github/guardrail-pulse-history.jsonl`, so no AI Clone code has actually changed since `dadf99d`. That means the two concrete blockers from my earlier review still stand on this head, and there are two new facts on top. Keeping this in changes-requested; specifics below.

Still blocking, re-verified statically at `a023cd4`:

1. `desktop/windows/package-lock.json` (added by this PR) is still out of sync with `desktop/windows/package.json`: the lockfile root reports `1.0.31` vs the manifest's `1.0.33`, and it resolves `axios@1.19.0` (manifest: `~1.18.0`) and `esbuild@0.27.7` (manifest: `~0.28.1`), so a clean `npm ci` still fails before tests can run. One nuance I can now confirm: the PR's `pnpm-lock.yaml` **is** correctly synced (`@beeper/desktop-api` `^5.0.0` -> `5.0.0` in the root importer), and this pnpm workspace doesn't consume an npm lockfile at all — the cleanest fix is to drop the added `package-lock.json` entirely rather than regenerate it.

2. `desktop/windows/src/shared/types.ts` still narrows the unrelated capture/meeting IPC contracts without updating call sites. At this head it removes `attemptId` from `meeting-capture-start`/`meeting-capture-stop`, but `src/main/meeting/meetingMonitor.ts` still sends it at both call sites (start ~line 143, stop ~line 186); it removes the `audio-source-ready` event, but `src/main/ipc/captureBridge.ts` still lists it in `OWNED_EVENT_TYPES` (line 25) and `src/renderer/src/lib/omiListenClient.ts` still branches on it (line 110). These are type errors against live code, unrelated to AI Clone, and `main` still has the fields/events — please revert that churn or carry it through every call site with tests.

New since my last review:

3. The branch now conflicts with `main` (`mergeable: conflicting`), so it needs a rebase regardless — a good moment to resolve 1 and 2 in the same pass.

4. David's product feedback above (Aug 19) stands as the bar this needs to meet: per-relationship voice rather than the once-daily `getLatestProfileText()` bio, judgment beyond the `SENSITIVE_HINTS` keyword list, and a presence model beyond the 60s poll. Re-reading `autoReplyPolicy.ts` at this head: it does now check the incoming message alongside the draft and fails closed on unrecognized modes — real progress — but it is still exactly the keyword-gate judgment David described, so that gap won't close by itself. If the path forward is the draft-for-review queue without auto-send, that's a product decision worth making explicitly with him rather than leaving implied.

Smaller items for the eventual dependency/scope pass: `pnpm-workspace.yaml` adds `'@google/genai': true` to `allowBuilds` (a script-execution allowance for an existing dependency — fine if intentional, but it belongs in the dependency review), and the `codingAgent/agentSelection.ts` + `taskRunner.ts` task-fit ranking rides along as an unrelated, well-tested concern — fine to keep, just noting it's separable.

For the record, the engineering quality is not what's holding this up: `beeperTokenStore.ts` keeps the token in safeStorage ciphertext and never returns it to boolean-status callers, `sessionResetWiring.test.ts` drives the real `setBackendSession` sign-out path, and `AiCloneDraftHost.tsx`'s per-chat send serialization is exactly right.

---
*by AI on behalf of David — before this can move it needs his product-direction call on the auto-send bar from his Aug 19 comment, plus dependency sign-off on `@beeper/desktop-api`.*

aryanorastar:
Keeping this open. Not approving `28fa0447dcdc`. `desktop/windows/src/main/aiClone/beeperClient.ts` absent on origin/main. mergeable CONFLICTING / DIRTY. No check rollup.

PR #11076, opened 2026-08-04
fix(listen): make dev parity capture volume writable

Git-on-my-level:
## What changed and why

The dev listener image runs as the non-root `omi` user with GID `10001`, but
the parity-pack capture root is an `emptyDir` mounted without a pod `fsGroup`.
Kubernetes therefore presented a root-owned volume that the listener could not
write: allowlisted sessions reached capture, then `persist()` failed creating
`cassettes/` with `PermissionError`, leaving both the local root and GCS prefix
empty.

Set the development backend-listen pod `fsGroup` to the image's GID and guard
the complete rendered Helm boundary: image group, capture env root, volume,
mount, and pod security context. Listen remains fail-open and the change does
not add logging identifiers, Prometheus dependencies, IAM changes, or any
production configuration.

Closes fe728900-458f-4eb4-ad80-afe60423ed70

none

## Verification

- `PYTHONPATH=backend backend/.venv/bin/python -m pytest backend/tests/unit/test_backend_listen_helm_defaults.py backend/tests/unit/test_listen_parity_capture.py -q` — 19 passed.
- `npm run test:parity-pack-v0` — 16 passed.
- `BACKEND_UNIT_TEST_FILE_LIST=../backend-test-files.txt bash test.sh` (from `backend/`) — 23 passed with the local fast-unit timing guard.
- `OMI_PR_BODY_FILE=pr-body.md make preflight` — repository-required deterministic preflight.

The real dev WebSocket was not exercised from this workspace because no GCP
or Firebase credentials are available. The remaining live gate is a normal dev
backend-listen Helm rollout followed by one allowlisted transcription session
and confirmation of a new cassette under the configured GCS prefix.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11076?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/charts/backend-listen/dev_omi_backend_listen_values.yaml
+++ b/backend/charts/backend-listen/dev_omi_backend_listen_values.yaml
@@ -52,8 +52,12 @@ podLabels:
   # Keep those scrape targets in the same development-only evidence plane as
   # the Pusher targets.
 
-podSecurityContext: {}
-  # fsGroup: 2000
+podSecurityContext:
+  # The image runs as omi:omi (10001:10001). Kubernetes mounts emptyDir as
+  # root-owned, so the parity-pack cassette root needs this supplemental group
+  # before the non-root listener can create its cassettes/ directory.
+  fsGroup: 10001
+  fsGroupChangePolicy: OnRootMismatch
 
 securityContext: {}
   # capabilities:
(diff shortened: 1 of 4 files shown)


kodjima33 (approved):
Correct root cause: emptyDir without fsGroup gave the non-root listener a root-owned capture volume, so persist() failed. Dev-only values change, full rendered-boundary test coverage. Approving; leaving the actual merge to Nik since it touches deploy pod security context.

PR #11084, opened 2026-08-04
fix(app): preserve on-device offline batch across BLE disconnect (#7194)

### LINKED ISSUE #7194: Bug: Offline/Bluetooth-disconnected recording is not working.

aryanorastar:
## Summary
When a supported wearable is doing its own on-device offline/batch recording and BLE disconnects, the app should not interrupt that path by starting the phone-mic batch fallback.

This PR prevents the phone-mic fallback from triggering when an on-device offline batch recording is already active (fixes #7194).

## Implementation
- `app/lib/utils/batch_recording.dart`: extended `shouldFallbackToPhoneOnDeviceDisconnect` with `isOnDeviceOfflineBatchActive`.
- `app/lib/services/capture/capture_controller.dart`: computes `isOnDeviceOfflineBatchActive` from `batchModeEnabled` plus the current device/offline session state, and passes it into the predicate.
- `app/test/unit/batch_recording_phone_test.dart`: added unit coverage for the new gating.

## Tests
- `flutter test test/unit/batch_recording_phone_test.dart`

none

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/app/lib/providers/device_provider.dart
+++ b/app/lib/providers/device_provider.dart
@@ -488,6 +488,7 @@ class DeviceProvider extends ChangeNotifier implements IDeviceServiceSubsciption
     setIsConnected(false);
     updateConnectingStatus(false);
 
+    await captureProvider?.onRecordingDeviceDisconnected();
     captureProvider?.updateRecordingDevice(null);
 
     // Batch mode: the native writer finalizes the in-progress recording on
--- a/app/lib/services/capture/capture_controller.dart
+++ b/app/lib/services/capture/capture_controller.dart
@@ -1575,11 +1575,20 @@ class CaptureController extends ChangeNotifier
   }
 
   Future<void> onRecordingDeviceDisconnected() async {
+    // #7194: if the wearable is already doing its own on-device offline/batch
+    // recording, do not stop it and do not start the phone mic fallback.
+    // Phone fallback is a complementary safety net for live streaming gaps (#11078),
+    // not a replacement for on-device offline capture.
+    final bool isOnDeviceOfflineBatchActive = SharedPreferencesUtil().batchModeEnabled &&
+        _recordingDevice != null &&
+        (_recordingDevice!.type == DeviceType.limitless || _offlineSessionStartSeconds != 0);
+
     if (!shouldFallbackToPhoneOnDeviceDisconnect(
       isRecordingDevice: _recordingDevice != null,
       isRecording: isRecordingDuringDeviceDisconnect(recordingState),
       supportsBatch: phoneMicSupportsTranscribeLater && deviceSupportsTranscribeLater,
       batchAlreadyActive: _phoneMicBatchActive,
+      isOnDeviceOfflineBatchActive: isOnDeviceOfflineBatchActive,
     )) {
       return;
     }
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Thanks for this — the intent (don't let phone-mic fallback clobber an active on-device offline/batch recording on BLE drop) is clear and the pure predicate + `isRecordingDuringDeviceDisconnect` mapping are nicely unit-tested. A few observations for the maintainer and contributor:

1. `DeviceType.limitless` special-case looks over-broad. In `onRecordingDeviceDisconnected`:
   ```dart
   (_recordingDevice!.type == DeviceType.limitless || _offlineSessionStartSeconds != 0)
   ```
   When `batchModeEnabled` is on and the device is a `limitless`, the `isOnDeviceOfflineBatchActive` guard is always true regardless of whether an offline session is actually running. That disables phone fallback for *any* limitless+batchMode disconnect, even when no on-device offline batch is active. If limitless reliably runs on-device batch whenever batchMode is enabled that may be intentional, but worth confirming the invariant holds — otherwise prefer gating on a real "on-device batch active" signal rather than device type.

2. Disconnect cleanup ordering (`device_provider.dart`):
   ```dart
   await captureProvider?.onRecordingDeviceDisconnected();
   captureProvider?.updateRecordingDevice(null);
   ```
   If `onRecordingDeviceDisconnected()` throws, `updateRecordingDevice(null)` and the subsequent WAL/analytics teardown are skipped, leaving state inconsistent after a device drop. Consider wrapping the awaited call in try/catch so the rest of the disconnect path always runs. (cubic flagged the same; +1.)

3. Shared `_phoneMicBatchStopCompleter`. The reconnect teardown in `streamDeviceRecording` waits on a completer that the *shared* native `onStop` callback completes. `_onBatchStalled` and the manual `stopStreamRecording` batch branch also call `phoneMic.stop()`, so any of those stops can resolve the completer before this reconnect's own native finalization drains — the mic-ownership overlap this serialization is meant to prevent. Conversely, if `onStop` doesn't fire, reconnect blocks the full 5s. Tying the completer to the specific stop (e.g. a generation/token checked inside `onStop`) would make the serialization precise and remove the stall. (Also raised by cubic; the concurrency concern is real.)

The two failing Desktop Swift checks are unrelated to this Flutter/Dart change (no Swift files touched; the jobs were cancelled/skipped on the Swift-changes path and failed on a separate run).

Net: logic is reasonable and tested at the unit level, but the capture state machine has real concurrency/ordering edges here — marking for maintainer review before merge.

---

kodjima33 (approved):
Phone-mic batch fallback fired even while the wearable was doing its own on-device offline batch, interrupting that capture. Predicate now gates on isOnDeviceOfflineBatchActive, with unit coverage. Fixes #7194. 5/5 - merging.

PR #11085, opened 2026-08-04
fix(app): avoid synced UI when WALs remain pending (#7240)

### LINKED ISSUE #7240: Offline recordings: synced status OK but not synced some are pending
It says synced but there are several recordings pending.

Omi AI 1.0.535 (855) — samsung SM-S938B — Android 16

Firmware 3.0.19

aryanorastar:
## Summary
Fixes #7240 where the app can mark recordings/conversations as *synced* even though some associated WALs are still left in an uploaded/pending state.

## What changed
- `SyncProvider` now refreshes WAL state before updating `syncState.completed.syncedConversations`.
- It filters the “synced conversations” pointers down to only those conversations whose associated WALs are fully `WalStatus.synced`.

## Regression test
- `flutter test test/providers/sync_provider_fully_synced_conversations_test.dart`

none

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/app/lib/providers/sync_provider.dart
+++ b/app/lib/providers/sync_provider.dart
@@ -398,12 +398,16 @@ class SyncProvider extends ChangeNotifier implements IWalServiceListener, IWalSy
       updatedConversationIds: result.updatedConversationIds,
     );
     if (_isDisposed) return;
+    // Refresh WALs *before* updating the "synced" conversation UI so we don't
+    // label recordings as synced while the reconciler is still leaving some
+    // WALs in an uploaded/pending state (issue #7240).
+    await refreshWals();
     if (conversations.isNotEmpty && !_syncState.isProcessing) {
       // Append to whatever is already shown — jobs reconcile incrementally.
       final merged = [..._syncState.syncedConversations, ...conversations];
-      _updateSyncState(_syncState.toCompleted(conversations: merged));
+      final filtered = filterFullySyncedConversations(conversations: merged, wals: _allWals);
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @aryanorastar — this is a clean, well-scoped fix for #7240. Filtering synced-conversation pointers to only those whose WALs are fully `WalStatus.synced`, and refreshing WALs before updating the synced UI, directly addresses the reported mismatch (synced badge while recordings are still uploaded/pending). The regression test covers both the pending-WAL and no-local-WAL cases, and CI (Dart Analyze & Tests, Android Compile Smoke) is green on the current head.

One edge worth a look before merge: `filterFullySyncedConversations` treats a conversation as fully synced when it has no conversationId-linked WALs (`return true` on empty). In the normal flow WALs are stamped with a `conversationId` via `stampConversationId()` before upload, so this is rarely hit — but if any un-stamped WAL ever reaches `uploaded`/pending (e.g. legacy persisted WALs), it would not be counted against its conversation and the conversation could still surface as synced. Consider either conservatively withholding when unlinked uploaded/pending WALs exist, or documenting the invariant that uploaded WALs always carry a `conversationId`. Flagging for maintainer review on that point; otherwise this looks good to merge.

---

kodjima33 (approved):
Conversations were marked synced while their WALs were still pending; SyncProvider now refreshes WAL state and only counts fully-synced conversations. Regression test included. 5/5 - merging.

Issue #11078, opened 2026-08-04
Phone-side capture fallback when BLE device disconnects


aryanorastar:
## Summary

When an active recording from a supported wearable loses its Bluetooth connection mid-session, the app should automatically fall back to capturing audio via the phone's own microphone (in batch mode) for the duration of the disconnect, then hand off cleanly back to device streaming on reconnect. Today nothing fills that gap: the app stops receiving device audio on BLE disconnect and does not start the existing phone-mic batch fallback, so the interval is lost until the wearable reconnects.

This is **distinct from #7194**. #7194 is about the wearable/phone continuing **on-device offline recording** (recording locally without a live connection to sync later) — a separate, existing-but-flaky code path. This issue is specifically about **falling back to phone-mic capture** while BLE is down, as a complementary safety net, not a replacement for on-device offline recording.

## Expected behavior

- An active supported-wearable recording that loses BLE mid-session automatically starts the phone's own mic batch-recording fallback.
- The fallback correctly triggers for the actual device-streaming state (not just the phone-mic recording state).
- The phone batch is finalized before device streaming resumes on reconnect, with a deterministic (not fire-and-forget) hand-off.
- Manual stops, unsupported devices, inactive sessions, and already-active batch sessions do not trigger the fallback.

## Candidate implementation

[#11066](https://github.com/BasedHardware/omi/pull/11066) implements this behavior and is being retargeted to track against this issue instead of #7194. It has an open, unresolved review comment noting the fallback predicate currently gates on `RecordingState.record` and does not trigger for the actual live BLE device-streaming state (`RecordingState.deviceRecord`), so the exact disconnect scenario this issue describes is currently missed by that PR's implementation. That needs to be fixed before merge.

## Why a separate issue

A maintainer (@mdmohsin7) noted #7194 is specifically about on-device offline recording (which already exists but is flaky), and that a phone-capture fallback is not the right fix for that issue. Filing this separately so the phone-fallback behavior can be reviewed and scoped on its own merits.

undivisible:
Closing per automated triage (2026-08-20): superseded: phone-mic fallback deliberately reverted (#11134). See /tmp/omi-issue-triage.md.

PR #11086, opened 2026-08-04
fix(app): fall back to phone capture after BLE disconnect

aryanorastar:
## What changed and why

**Context:** the original phone-mic BLE-disconnect fallback for #11078 already
landed on `main` via #11084 (shared history). This PR rebases cleanly and closes
the disconnect/reconnect race @Git-on-my-level flagged on #11086.

When `onRecordingDeviceDisconnected()` awaits mic permission / docs path /
`startBatch()`, a fast wearable reconnect can install a new recording device
and then be cleared by the stale disconnect continuation — or the fallback can
still start `omibatchphoneauto` after the session is no longer current.

## Fix

- Capture-controller epoch cancels in-flight disconnect→phone-batch fallback on
  `streamDeviceRecording` / `stopStreamDeviceRecording`.
- Recheck epoch + recording-device id after each await before starting batch.
- `device_provider` only calls `updateRecordingDevice(null)` when reconnect has
  not already replaced the disconnected device.
- Fallback errors are caught so the rest of disconnect cleanup always runs.

none

## How it was verified

- `flutter test test/unit/batch_recording_phone_test.dart` (24 passed)

## Tests

- `app/test/unit/batch_recording_phone_test.dart` — revalidation helpers for
  epoch/device-id abort and conditional `updateRecordingDevice(null)`

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/app/lib/providers/device_provider.dart
+++ b/app/lib/providers/device_provider.dart
@@ -23,6 +23,7 @@ import 'package:omi/services/services.dart';
 import 'package:omi/services/battery_widget_service.dart';
 import 'package:omi/services/wals/wal_syncs.dart';
 import 'package:omi/services/wals/recording_transfer_coordinator.dart';
+import 'package:omi/utils/batch_recording.dart';
 import 'package:omi/utils/device.dart';
 import 'package:omi/utils/firmware_update_build_policy.dart';
 import 'package:omi/utils/firmware_update_check_session.dart';
@@ -488,8 +489,28 @@ class DeviceProvider extends ChangeNotifier implements IDeviceServiceSubsciption
     setIsConnected(false);
     updateConnectingStatus(false);
 
-    await captureProvider?.onRecordingDeviceDisconnected();
-    captureProvider?.updateRecordingDevice(null);
+    // Capture before the awaited fallback so a reconnect that installs a new
+    // recording device is not cleared by this stale disconnect continuation.
+    final disconnectedRecordingDeviceId = captureProvider?.recordingDevice?.id;
+    try {
+      await captureProvider?.onRecordingDeviceDisconnected();
+    } catch (e, st) {
+      // Fallback startup/teardown must not abort the rest of disconnect cleanup.
+      Logger.error('onRecordingDeviceDisconnected failed during device disconnect: $e\n$st');
+    }
+
+    final currentRecordingDeviceId = captureProvider?.recordingDevice?.id;
+    if (shouldClearRecordingDeviceAfterDisconnectFallback(
+      disconnectedDeviceId: disconnectedRecordingDeviceId,
+      currentRecordingDeviceId: currentRecordingDeviceId,
+    )) {
+      captureProvider?.updateRecordingDevice(null);
+    } else {
+      Logger.debug(
+        'Skipping updateRecordingDevice(null) after disconnect — reconnect already set '
+        '${currentRecordingDeviceId ?? "unknown"} (was $disconnectedRecordingDeviceId)',
+      );
+    }
 
     // Batch mode: the native writer finalizes the in-progress recording on
     // disconnect (.bin.part -> .bin). Rescan shortly after the rename completes
--- a/app/lib/services/capture/capture_controller.dart
+++ b/app/lib/services/capture/capture_controller.dart
@@ -131,6 +131,10 @@ class CaptureController extends ChangeNotifier
   // finalization is complete.
   Completer<void>? _phoneMicBatchStopCompleter;
 
+  /// Bumped to cancel an in-flight BLE-disconnect → phone-batch fallback when
+  /// reconnect / user-stop wins the race across the fallback's awaits.
+  int _disconnectPhoneFallbackEpoch = 0;
+
   bool get isPhoneMicBatchRecording => _phoneMicBatchActive;
 
   bool _isLoadingInProgressConversation = false;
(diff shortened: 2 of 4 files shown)


Git-on-my-level (commented):
Thanks for this — clean design. Extracting the fallback predicate into `batch_recording.dart` as a pure, well-tested function and serializing the reconnect teardown with a `Completer` + 5s timeout is the right shape, and the `deviceRecord`/`interrupted` lane in `isRecordingDuringDeviceDisconnect` is a good catch (a socket drop can precede the BLE disconnect).

One thing I'd want a maintainer to look at before merge: there's a revalidation gap in the disconnect path. `onRecordingDeviceDisconnected()` now awaits the mic permission request, `getApplicationDocumentsDirectory()`, and `startBatch()` before `device_provider.dart` runs its unconditional `captureProvider?.updateRecordingDevice(null)`. If the wearable reconnects during that window, `streamDeviceRecording` installs the new device first and the stale disconnect continuation then clears it — and `_startPhoneMicBatch` can also keep running after the user stopped or the device came back, starting an unexpected `omibatchphoneauto` recording.

A session/device generation token (or a cheap `recordingState`/device-id recheck after each await) would close that. The 5s timeout on the reconnect-side teardown is a fine safety net for the native-stop side. No security/privacy/dependency concerns from my read — the audio path reuses the existing permission-gated phone-mic batch service.

Flagging for maintainer review on the capture state-machine concurrency; otherwise this looks solid.

---

kodjima33 (approved):
Correct fix: BLE loss mid-recording left the gap uncaptured; starting the existing phone-mic batch fallback is the right lane. Approving — it now conflicts with #11084 (just merged, same predicate), so it needs a rebase from the author before it can land.

aryanorastar:
Ready for review — clean replacement for closed #11066.

Scope is phone-mic fallback on BLE disconnect for #11078 only (not #7194; that landed in #11084). Includes the earlier review fixes: `deviceRecord`/`Interrupted` as active lane, Transcribe Later `supportsBatch` gate, serialized phone-batch stop. Local: 20 unit tests + `pr-preflight` green; CI currently green.

aryanorastar:
@kodjima33 @Git-on-my-level rebased onto latest `main` (conflicts with merged #11084 resolved by keeping the landed phone-fallback + on-device offline gate).

This PR's unique delta is now the disconnect/reconnect revalidation David asked for:
- epoch cancel on reconnect/stop
- post-await device-id/epoch recheck before starting `omibatchphoneauto`
- conditional `updateRecordingDevice(null)` so a stale disconnect can't clear a reconnected device
- fallback errors no longer abort the rest of disconnect cleanup

`flutter test test/unit/batch_recording_phone_test.dart` → 24 passed.

Git-on-my-level:
This head closes the revalidation gap flagged earlier on this branch — the epoch-based cancellation plus the post-await device-id / recordingState recheck is exactly the shape that was asked for, and extracting the decision logic into batch_recording.dart as pure, tested predicates is clean.

Notes for the maintainer reviewing the capture concurrency:

- The epoch pattern is sound for Dart's single-threaded event loop; the reads/writes between awaits need no lock. Bumping on both streamDeviceRecording (reconnect) and stopStreamDeviceRecording (user stop) covers the two cancellation sources, and the recordingState == stop check in stillCurrent() is good belt-and-suspenders.
- shouldClearRecordingDeviceAfterDisconnectFallback correctly prevents a stale disconnect continuation from clearing a device that reconnect already installed.
- The try/catch around onRecordingDeviceDisconnected() means fallback startup/teardown failures no longer abort the rest of disconnect cleanup — solid robustness improvement.
- Unit tests cover the predicate logic well. Integration coverage of the epoch-bump points inside the live capture controller would strengthen confidence further, but is not blocking given the predicates are the actual decision logic.

Pre-existing (from the already-merged reconnect serialization in #11084, not introduced by this delta): the reconnect-side Completer can release before the native phone-mic stop() future resolves, so wearable capture may reopen while native teardown is still draining. Worth a separate look but orthogonal to this PR.

Blockers before merge:
- Needs a rebase — GitHub reports conflicts with the merged #11084 (mergeable state: dirty).
- Needs maintainer sign-off on the capture state-machine concurrency (label retained).

No security, privacy, dependency, or workflow concerns from my read — the audio path reuses the existing permission-gated phone-mic batch service.

---

aryanorastar:
@Git-on-my-level @kodjima33 @mdmohsin7 — stopping the rebase asked for above.

While this PR was waiting, @mdmohsin7 merged [#11134](https://github.com/BasedHardware/omi/pull/11134) which **reverts** the automatic phone-mic BLE-disconnect fallback that landed inside #11084. That revert cites the earlier product rejection on #11066 and notes that #11078 is still unapproved as a product direction.

Rebasing this tip onto current `main` would re-introduce the exact behavior #11134 just removed (plus the revalidation hardening). That would undo a maintainer product decision, so I’m **not** rebasing/pushing that.

I’ll close this PR as superseded by #11134’s product call. If phone-side fallback for #11078 is wanted after all, it should be a fresh proposal that addresses the Omi/ring-buffer double-capture concern raised in #11134 — not a revive of this tip.

aryanorastar:
Closing as superseded by maintainer product decision in #11134 (revert of automatic phone-mic BLE-disconnect fallback). Not rebasing this tip onto main.

PR #11090, opened 2026-08-04
fix(backend): make every chat agent terminal exit leave an answer and report failure

RatnamOjha:
## Summary

Makes every terminal exit from both chat agent loops leave the user an answer and report whether the turn failed, and adds a cost-attribution report over the gateway accounting ledger.

## Root cause

A terminal path that ends a turn without writing to `full_response` is indistinguishable from a model that legitimately said nothing: `routers/chat.py` builds both the persisted reply and the terminal `done:` frame from that list, so it renders its own canned error and records no failure.

Four exits had that shape. On the Anthropic path, a provider safety classifier declines a turn as an HTTP 200 with `stop_reason == "refusal"` and empty content, which the `stop_reason != "tool_use"` break read as the model choosing to answer nothing. On the managed OpenAI path — the runner gateway feature mode selects, so the one carrying production traffic — the client-construction and unrecoverable-provider exits each pushed their apology through `callback.put_data` alone, which reaches the live stream but not the persisted answer.

That path also has no refusal signal at all. A filtered turn simply arrives as a completion with no text and no tool calls, which the loop read as "done", fell out to `callback.end()`, and returned `None` — persisting a blank answer and recording the turn as a success.

## Impact

`_end_with_answer_guarantee` is the provider-independent form of the contract: it closes the stream, and if nothing was delivered it says so and returns `empty_answer` rather than `None`. Both loops exit through it, so the guarantee holds without either path having to detect a provider-specific refusal shape. A refusal additionally reports `provider_refusal` and logs `stop_details.category`, and its message does not invite a retry — the same prompt would be declined again.

Emptiness is judged on the joined, stripped text, not list truthiness: a stream can emit the inter-iteration separator or a whitespace-only delta and then fail, leaving `full_response` non-empty while the reply is still blank to the reader.

This class has now recurred once per exit path, so instead of another per-site fix, `TERMINAL_EXITS` and `OPENAI_TERMINAL_EXITS` enumerate every way each loop can end and assert each leaves a non-empty answer and reports failure honestly. They drive the real loops through the callback seam rather than asserting on source text.

`scripts/chat_agent_cost_report.py` reports spend split across cached reads, uncached input, cache writes and output. It groups the range by the model each attempt was billed as and prices each group with its own rate card — `chat_agent` has already moved provider once, and a single card cannot price a range spanning that. Output is aggregate token counts and costs only: no uid, no message content.

## Validation

- `BACKEND_UNIT_TEST_FILE_LIST=... BACKEND_FAST_UNIT_FAIL_SECONDS=1.0 bash test.sh` — 95 passed, 0 failed across the 3 touched test files (`test_chat_agent_provider_retry.py` 51, `test_prompt_cache_integration.py` 32, `test_chat_agent_cost_report.py` 12)
- Both guards verified to fail without their fix: reverting the two `put_data` calls and the shared guard fails 5 managed cases (`client_construction_failed`, `model_returned_nothing`, `provider_failed_unrecoverably`, `retry_budget_exhausted`, and the silent-completion test); reverting `_has_answer` alone fails `provider_refused_after_whitespace`
- `scripts/pr-preflight --base upstream/main --pr-body-file ...` — **passed: 23 checks**, including `product-invariants`, `failure-class-protocol`, and `backend-module-isolation`
- `scripts/check_module_stub_pollution.py` — 883 files, 0 violations
- Tool-schema sizing measured offline against the real 27 `CORE_TOOLS` schemas: tool definitions are 11,205 of the 11,759-token cached prefix (95.3%)

The `output_config.effort` A/B is written but not yet run — it needs a funded API key, and effort is only observable in output tokens because thinking is folded into `output_tokens`.

None — `scripts/pr-preflight` reports `OK: no locked invariants require naming for these changes`.

### DIFF
--- a/backend/scripts/chat_agent_cost_report.py
+++ b/backend/scripts/chat_agent_cost_report.py
@@ -1,0 +1,433 @@
+#!/usr/bin/env python3
+"""Attribute chat-agent LLM spend across the gateway accounting ledger.
+
+The ledger already records every billable component of an Anthropic Messages
+attempt (cached reads, uncached input, cache writes and their TTL, output), but
+nothing reads it back per feature. Cost work on the chat agent otherwise starts
+from guesses about which component dominates, and the two candidate levers pull
+in opposite directions depending on the answer: a low cache-read share means the
+request prefix is being rewritten rather than reused, while a high output share
+means generation settings matter more than prefix layout.
+
+Aggregation is pure and takes ledger rows as plain mappings, so the arithmetic is
+testable without Firestore. Only the fetch step touches the database.
+
+Reads one day per query using two equality filters, which Firestore serves from
+single-field indexes — no composite index and no schema change.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from collections import Counter
+from dataclasses import dataclass, field
+from datetime import date, datetime, timedelta, timezone
+from pathlib import Path
+from typing import Any, Callable, Iterable, Iterator, Mapping, Sequence
+
+BACKEND_ROOT = Path(__file__).resolve().parents[1]
+if str(BACKEND_ROOT) not in sys.path:
+    sys.path.insert(0, str(BACKEND_ROOT))
+
+ATTEMPTS_COLLECTION = 'llm_gateway_attempts'
+DEFAULT_FEATURE = 'chat_agent'
+DEFAULT_DAYS = 7
+MICRO_USD_PER_USD = 1_000_000
+TOKENS_PER_MILLION = 1_000_000
+
+# Cache writes are billed at a TTL-dependent multiple of the input rate, so the
+# 5-minute and 1-hour components are never collapsed into one number.
+TTL_FIVE_MINUTES = '5m'
+TTL_ONE_HOUR = '1h'
+
+
+@dataclass
+class Rates:
+    """Per-million-token micro-USD rates for one provider model."""
+
+    rate_card_id: str
+    input_micro_usd: int
+    cached_input_micro_usd: int
+    output_micro_usd: int
+    cache_write_micro_usd: int
+    cache_write_1h_micro_usd: int
+
+
+@dataclass
+class Totals:
+    """Accumulated ledger quantities for one feature over a date range."""
+
+    attempts: int = 0
+    priced_attempts: int = 0
+    cached_input_tokens: int = 0
+    uncached_input_tokens: int = 0
+    cache_write_5m_tokens: int = 0
+    cache_write_1h_tokens: int = 0
+    cache_write_untagged_tokens: int = 0
+    output_tokens: int = 0
+    ledger_cost_micro_usd: int = 0
+    attempts_with_cache_read: int = 0
+    attempts_with_cache_write: int = 0
+    cache_status: Counter[str] = field(default_factory=Counter)
+    outcome: Counter[str] = field(default_factory=Counter)
+    models: Counter[str] = field(default_factory=Counter)
+    days: set[str] = field(default_factory=set)
+
+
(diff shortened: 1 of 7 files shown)


Git-on-my-level:
Thanks for the careful follow-up here. I reviewed the current head and ran the focused backend unit coverage locally: `test_chat_agent_provider_retry.py` and `test_chat_agent_cost_report.py` both pass together (63 tests).\n\nThe chat-agent terminal-exit fix looks directionally solid: the failure apologies now go through the same persisted-answer path as streamed text, empty/refusal exits are reported as failures instead of silent successes, and the table-style tests cover both Anthropic and managed OpenAI loop exits. The new cost report is also structured as pure aggregation around ledger rows, with Firestore limited to the fetch path and no message/user content in the rendered output.\n\nI am not formally approving from automation because this touches backend chat-agent runtime behavior and operational cost-reporting tooling, so a human maintainer should still make the merge call. I added `needs-maintainer-review` plus the `backend` category label.

---

RatnamOjha:
Thanks for the review. All six findings are addressed. Note the inline comments are against the pre-rebase head — this branch was originally cut from a stale base and has since been rebased onto current `main`, so the line numbers won't match.

**1. Refusal can persist an empty answer after a whitespace-only stream** (`agentic.py`) — Fixed. Emptiness is now judged by `_has_answer()`, which tests the joined, stripped text rather than list truthiness. Applied at both sites: the refusal branch and the shared `_end_with_answer_guarantee`. Added `provider_refused_after_whitespace` to the terminal-exit table; verified it fails without the change.

**2. Pin `AGENT_STREAM_PROVIDER_MAX_ATTEMPTS` in the budget-exhaustion case** — Fixed. Both tables now monkeypatch it to `SCRIPTED_FAILURE_ATTEMPTS = 5`, matching the number of scripted streams, so raising the module default surfaces as a budget assertion rather than an `IndexError` off the end of the script.

**3. Module-scope `sys.modules` mutation** — Fixed, though not quite as suggested. The registration isn't optional: the script's dataclasses resolve their own module by name during `exec_module`, and dropping it fails with `AttributeError: 'NoneType' object has no attribute '__dict__'`. The real problem was the *scope*, so loading moved into a fixture that removes the entry in a `finally`. `scripts/check_module_stub_pollution.py` now reports 0 violations across 883 files.

**4. Component breakdown wrong when the range mixes models or rate-card versions** — Fixed, covering both dimensions. The range is grouped by `(model, rate_card_id)` via `build_totals_by_pricing_basis` and each group is priced on its own basis. `load_rates` takes an optional `rate_card_id` and prefers it over the model-name lookup, since a model's pricing can be revised while the name is unchanged; rows predating the field fall back to the name. An id absent from the card file raises rather than silently repricing history at current rates. One unpriceable group is reported to stderr without costing the rest of the report.

**5. Default date window not UTC** — Fixed. Derived from `datetime.now(timezone.utc).date()`. Covered by `test_the_default_window_ends_on_the_utc_day`, which asserts the queried day against an aware UTC clock.

**6. `yaml.YAMLError` escapes the clean failure path** — Fixed. Added to the handled tuple in both `main`'s config-parsing block and the per-group rate lookup, so a malformed `generated_route_overrides.yaml` or `cost_rate_cards.yaml` gives `error: ...` and exit 2 instead of a traceback.

Verification: 63 tests pass across the two files, preflight passes 23 checks against current `main`, and both new guards were confirmed to fail without their fix — reverting the two `put_data` calls and the shared guarantee fails 5 managed-loop cases, and reverting `_has_answer` alone fails the whitespace case.

kodjima33 (approved):
Making every chat-agent terminal exit leave an answer is a real fix; approving, 1k-line diff so leaving the merge to Nik.

PR #11093, opened 2026-08-04
fix(monitoring): disable invalid managed GKE alerts

### LINKED ISSUE #9138: prod monitoring has persistent false-positive control-plane alerts and severity noise
## Summary

Prod monitoring is firing persistent control-plane / scrape / rule alerts while application workloads appear healthy. This reduces trust in alerts during real incidents.

## Evidence

Observed via Prometheus/Grafana read-only checks on 2026-07-06 UTC.

Firing alerts included:

- `KubeProxyDown` critical
- `KubeSchedulerDown` critical
- `KubeControllerManagerDown` critical
- `PrometheusRuleFailures` critical
- `TargetDown` warnings for kubelet / coredns / gpu-metrics
- `Watchdog`

At the same time:

- Public API and agent healthchecks returned 200.
- `prod-omi-gke` pods in `prod-omi-backend` were `1/1 Running`.
- `kube_deployment_status_replicas_unavailable` was 0 for prod Omi backend services.
- Pending/Failed/Unknown pod counts from Prometheus were 0 for prod backend/monitoring/system namespaces in the sampled query.

Separately, pusher logs include many normal INFO request lines emitted at ERROR severity, which further increases monitoring noise.

## Expected behavior

Alerts should represent actionable service or infrastructure degradation. Managed-GKE control-plane scrape limitations and known unavailable targets should be adjusted, suppressed, or relabeled so they do not page as critical app-impacting issues.

## Acceptance criteria

- [ ] Determine whether the control-plane alerts are invalid for managed GKE scrape topology.
- [ ] Fix or suppress invalid `KubeProxyDown`, `KubeSchedulerDown`, and `KubeControllerManagerDown` alerts.
- [ ] Fix `PrometheusRuleFailures` in the affected rule groups.
- [ ] Triage `TargetDown` for kubelet/coredns/gpu-metrics and make the alert actionable or suppress expected gaps.
- [ ] Reduce INFO-as-ERROR application log noise, especially pusher request logs.
- [ ] Document which prod alerts should page vs remain dashboard-only.

## Priority

P2/P3: monitoring reliability issue; not itself a prod outage, but it weakens incident response.

aryanorastar:
## Summary

Fixes #9138.

The production kube-prometheus-stack values enabled scrapes for kube-proxy,
kube-scheduler, and kube-controller-manager. Those control-plane targets are
managed by GKE and are not exposed to this customer cluster, so the default
`*Down` rules became permanently false-positive after their 15-minute window.

The production values now set those three component scrapes to `enabled: false`.
In kube-prometheus-stack that alone gates the corresponding PrometheusRule
groups, so the invalid pages stop without a separate `defaultRules.disabled`
block (which would silently keep the alerts off if the components are later
re-enabled on a self-managed control plane). Application alerts, `TargetDown`,
and `PrometheusRuleFailures` are unchanged.

## Verification

- `python3 -m pytest backend/tests/unit/test_monitoring_alert_rule_contract.py backend/tests/unit/test_logging_config.py -q`
- `helm template prod backend/charts/deepgram-self-hosted/nova-3/charts/kube-prometheus-stack -f backend/charts/monitoring/kube-prometheus-stack/prod_omi_monitoring_values.yaml`
  - Rendered successfully; all three control-plane `*Down` alert rules rendered zero times.

## Risk

This intentionally stops collecting these managed control-plane metrics in the
production chart. If the monitoring deployment moves to a self-managed control
plane, flip the three `enabled` flags (and the alerts return with them).

## Invariants

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11093?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/charts/monitoring/kube-prometheus-stack/prod_omi_monitoring_values.yaml
+++ b/backend/charts/monitoring/kube-prometheus-stack/prod_omi_monitoring_values.yaml
@@ -1,3 +1,18 @@
+## GKE manages the Kubernetes control plane; these customer-cluster scrape
+## targets do not exist and their default "down" alerts would page forever.
+kubeProxy:
+  enabled: false
+kubeScheduler:
+  enabled: false
+kubeControllerManager:
+  enabled: false
+
+defaultRules:
+  disabled:
+    KubeProxyDown: true
+    KubeSchedulerDown: true
+    KubeControllerManagerDown: true
+
 grafana:
   enabled: true
   serviceAccount:
--- a/backend/charts/monitoring/kube-prometheus-stack/prod_omi_monitoring_values.yaml
+++ b/backend/charts/monitoring/kube-prometheus-stack/prod_omi_monitoring_values.yaml
@@ -1,17 +1,14 @@
 ## GKE manages the Kubernetes control plane; these customer-cluster scrape
 ## targets do not exist and their default "down" alerts would page forever.
+## Component `enabled: false` already gates the whole PrometheusRule group in
+## kube-prometheus-stack, so no separate `defaultRules.disabled` entries —
+## flipping enabled back on must restore the alert automatically.
 kubeProxy:
   enabled: false
 kubeScheduler:
   enabled: false
 kubeControllerManager:
   enabled: false
-
-defaultRules:
-  disabled:
-    KubeProxyDown: true
-    KubeSchedulerDown: true
-    KubeControllerManagerDown: true
 
 grafana:
   enabled: true
--- a/backend/tests/unit/test_monitoring_alert_rule_contract.py
+++ b/backend/tests/unit/test_monitoring_alert_rule_contract.py
@@ -102,16 +102,21 @@ def test_split_alert_exports_preserve_error_count_no_data_contract():
 
 
 def test_managed_gke_disables_unavailable_control_plane_scrapes_and_alerts():
-    """Managed GKE must not page on control-plane targets it cannot expose."""
+    """Managed GKE must not page on control-plane targets it cannot expose.
+
+    kube-prometheus-stack gates each control-plane PrometheusRule group on the
+    component ``enabled`` flag, so ``enabled: false`` alone suppresses the
+    ``*Down`` alerts. Do not also set ``defaultRules.disabled`` — that would
+    keep the alert off after a future self-managed control-plane re-enable.
+    """
     values = yaml.safe_load(PROD_STACK_VALUES.read_text(encoding="utf-8"))
 
     for component in ("kubeProxy", "kubeScheduler", "kubeControllerManager"):
         assert values[component]["enabled"] is False
 
-    disabled = values["defaultRules"]["disabled"]
-    assert disabled["KubeProxyDown"] is True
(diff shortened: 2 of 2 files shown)


Git-on-my-level (changes requested):
Thanks @aryanorastar — this is a genuinely useful fix. Suppressing the three managed-GKE control-plane `*Down` critical alerts is the right call, and the regression test in `test_monitoring_alert_rule_contract.py` is a nice touch that will prevent regressions. Issue #9138 is clearly addressed.

One small change request before merge.

`defaultRules.disabled` is redundant and slightly risky

I checked the vendored `kube-prometheus-stack` chart (`backend/charts/deepgram-self-hosted/nova-3/charts/kube-prometheus-stack`, v60.5.0). Each of the three PrometheusRule groups is gated at the **file level** by its component flag, e.g. in `kubernetes-system-kube-proxy.yaml`:

```
{{- if and ... .Values.defaultRules.create .Values.kubeProxy.enabled .Values.defaultRules.rules.kubeProxy }}
```

Setting `kubeProxy.enabled: false` (and likewise for `kubeScheduler` / `kubeControllerManager`) already prevents the entire `KubeProxyDown` / `KubeSchedulerDown` / `KubeControllerManagerDown` PrometheusRule from being rendered. So the `defaultRules.disabled` block doesn't suppress anything that `enabled: false` doesn't already suppress.

The mild risk: if the cluster ever moves to a self-managed control plane and someone flips `kubeControllerManager.enabled: true` (your own PR body notes these values "must be revisited" in that case), the leftover `KubeControllerManagerDown: true` would silently keep that alert off and could hide a real control-plane outage.

Suggested fix

Drop the `defaultRules.disabled` block and rely solely on the component `enabled: false` flags — then the alerts automatically come back when (and only when) the component is re-enabled. The test would then just assert the three `enabled: false` flags.

```yaml
## GKE manages the Kubernetes control plane; these customer-cluster scrape
## targets do not exist and their default "down" alerts would page forever.
kubeProxy:
  enabled: false
kubeScheduler:
  enabled: false
kubeControllerManager:
  enabled: false
```

No blocker on the core direction — flagging for human maintainer review given this touches prod alert routing.

---

aryanorastar:
@Git-on-my-level addressed — dropped the redundant `defaultRules.disabled` block and rely solely on `kubeProxy` / `kubeScheduler` / `kubeControllerManager` `enabled: false`.

Regression test now asserts those three flags are false and that the three `*Down` keys are **not** present under `defaultRules.disabled`, so re-enabling a component on a self-managed control plane restores the alert. Local contract test + `pr-preflight` green; helm template still renders zero of the three rules.

kodjima33 (approved):
Correct root cause (GKE-managed control plane targets never expose kube-proxy/scheduler/controller-manager), minimal chart fix, contract test added, CI green.

PR #11096, opened 2026-08-04
fix(app): stall watchdog for zombie Listening on BLE live capture (#6977)

### LINKED ISSUE #6977: BLE device audio: stale live socket state can drop gaps and leave app stuck 'Listening'
**Describe the bug**
During BLE device recording (Omi/CV1/DK2-class live streaming), the app can stay in an active "Listening" state while no real data is progressing. If phone internet flaps or the live socket gets into a stale half-connected state, the transcript can resume with a gap (for example `1,2,3,4,5,16,17,18,19,20`) and in worse cases the session appears stuck until the app is restarted.

This looks more serious than the current UX-only framing in #6171 for BLE device audio specifically: the app does **not** keep an always-on durable phone-side WAL for device audio while the socket merely *appears* connected.

**To Reproduce**
1. Connect an Omi/CV1/DK2-style BLE recording device.
2. Open the app and start a fresh transcription session.
3. Start counting aloud from `1` upward.
4. Around `5`, enable airplane mode on the phone.
5. Around `15`, disable airplane mode.
6. Keep speaking and watch transcript continuity and app state.

**Current behavior**
- The app can continue to look active / "Listening" even while no transcript progress is happening.
- When it resumes, the transcript can jump forward with the middle audio missing.
- In some cases the degraded state appears to persist until the app is force-closed/reopened.
- There is no strong user-facing signal for "device connected but no data is actually flowing".

**Expected behavior**
- Device audio should always be durably buffered somewhere during live streaming degradation.
- Reconnect should either resume the live path cleanly or upload/replay the backlog automatically.
- The app should surface an explicit warning if the device is connected but audio/transcript progress has stalled.
- The app should not stay in a zombie "Listening" state indefinitely.

**Code-backed root cause**
1. **BLE device audio does not have always-on phone WAL buffering during normal live streaming.**
   In `app/lib/providers/capture_provider.dart:622-625`, device WAL support is only enabled when the socket is not connected or `unlimitedLocalStorageEnabled` is on:
   ```dart
   (_socket?.state != SocketServiceState.connected || SharedPreferencesUtil().unlimitedLocalStorageEnabled)
   ```
   That means the normal "socket says connected" path does not keep an always-on durable phone-side copy for device audio.

2. **Frames are treated as synced immediately after local socket send, not after durable server receipt.**
   In `app/lib/providers/capture_provider.dart:639-649`, frames are marked synced right after `_socket?.send(...)`:
   ```dart
   _socket?.send(socketPayload);
   _wal.getSyncs().phone.markFrameSynced(frame.syncKey);
   ```
   There is no per-frame server ack boundary here.

3. **Recovery is tied to session end/startup/connectivity edges, not stalled live data flow.**
   WAL finalize/recovery paths are in `app/lib/providers/capture_provider.dart:1281-1392`, and connectivity-triggered recovery is only in `app/lib/providers/capture_provider.dart:1595-1600`.
   There does not appear to be a watchdog for:
   - BLE still connected
   - socket still reported connected
   - but no transcript progress / no real end-to-end data progress

4. **This differs from true offline device storage.**
   True offline storage exists when the device is actually disconnected (for example DK2 SD card flow documented in `docs/doc/developer/DevKit2Testing.mdx:19`), but this bug is the mid-session handoff / stale-live-path case.

**Why this likely causes the observed gap**
If network handoff or reconnect leaves the live socket in a stale "connected enough" state, the app may continue using the live-stream path without turning on durable WAL capture for device audio. That creates a gap window before recovery logic ever has a chance to help.

**Suggested fixes**
- Make BLE device audio use phone WAL as an always-on durability layer, not only when socket state is already disconnected.
- Treat websocket send as a fast path, not the durability boundary.
- Add a stall watchdog using some combination of:
  - BLE audio bytes received
  - websocket bytes successfully progressing
  - transcript segment activity
  - time since last good end-to-end progress
- Show a notification/banner such as: `Device connected, but no data is coming in` when the stall watchdog fires.
- Flush/finalize in-flight frames immediately on socket degradation / connectivity loss / background transitions instead of waiting for normal chunk/flush timing.
- Auto-sync backlog after reconnect instead of relying on manual recovery behavior.

**Related but separate follow-up**
There is also a WAL state-machine bug where `SyncProvider` shows `WalStatus.corrupted` as pending, but `LocalWalSyncImpl.syncAll()` only retries `WalStatus.miss` (`app/lib/services/wals/local_wal_sync.dart:504`). That can strand retryable recordings forever. I think that should be tracked/fixed too, but I am filing this issue specifically for the stale live-stream / missing-middle-audio path.

**user ID (can we access the user info to validate the bug?):**
Not included here.

**Smartphone + device (please complete the following information):**
- Device: reproduced by users across BLE Omi devices (not DK2-specific)
- OS: mobile repro uses airplane-mode network handoff; exact OS/device varies
- Browser: N/A
- App Version: not pinned here
- Device version: not pinned here

**Additional context**
Related issues:
- #5733 umbrella capture-loss issue
- #6171 currently frames a similar outage path as UX/manual-recovery only; this issue is specifically about BLE device audio not having guaranteed always-on durable WAL buffering during stale socket handoff

aryanorastar:
## Summary
- Fixes [#6977](https://github.com/BasedHardware/omi/issues/6977): BLE live capture can look permanently **Listening** while transcript progress stalls and audio mid-session is lost when the socket still appears connected.
- Keep phone WAL always-on for supported BLE device audio (no longer gated on socket disconnected / unlimited storage).
- Only `markFrameSynced` after a local websocket send while transcript progress is recent.
- Add a transcript-stall watchdog: after sustained silence, flip UI to reconnecting and stop the socket so reconnect can recover; clear stall on new segments / connect / close.
- Pure helpers + unit tests in `app/test/unit/transcript_stall_test.dart`.

## Test plan
- [x] `flutter test test/unit/transcript_stall_test.dart`
- [ ] Manual: start BLE live capture, confirm Listening while segments arrive
- [ ] Manual: force transcript silence with socket still connected → status becomes reconnecting (not indefinite Listening) and socket restarts
- [ ] Manual: after reconnect + new segments, status returns to Listening
- [ ] Manual: pause / mute does not falsely trip the stall watchdog
- [ ] Confirm mid-session audio remains in phone WAL when the live socket was stale

## Invariants

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/conversations/widgets/processing_capture.dart
+++ b/app/lib/pages/conversations/widgets/processing_capture.dart
@@ -91,8 +91,7 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
                 (SharedPreferencesUtil().batchModeEnabled && provider.havingRecordingDevice)) {
               return;
             }
-            final isCaptureActive =
-                provider.recordingState == RecordingState.record ||
+            final isCaptureActive = provider.recordingState == RecordingState.record ||
                 provider.recordingState == RecordingState.systemAudioRecord ||
                 provider.recordingState == RecordingState.deviceRecord ||
                 provider.recordingState == RecordingState.initialising ||
@@ -191,8 +190,7 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
     bool isHavingDesireDevice = SharedPreferencesUtil().btDevice.id.isNotEmpty;
     bool isHavingRecordingDevice = captureProvider.havingRecordingDevice;
 
-    bool isUsingPhoneMic =
-        captureProvider.recordingState == RecordingState.record ||
+    bool isUsingPhoneMic = captureProvider.recordingState == RecordingState.record ||
         captureProvider.recordingState == RecordingState.initialising ||
         captureProvider.recordingState == RecordingState.pause ||
         captureProvider.recordingState == RecordingState.interrupted;
@@ -200,8 +198,7 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
     // Check if any recording is active (phone mic, system audio, or device recording).
     // `interrupted` is included so the in-session UI stays visible while the
     // pipeline is transiently broken (e.g., iOS audio session interruption).
-    bool isAnyRecordingActive =
-        captureProvider.recordingState == RecordingState.record ||
+    bool isAnyRecordingActive = captureProvider.recordingState == RecordingState.record ||
         captureProvider.recordingState == RecordingState.systemAudioRecord ||
         captureProvider.recordingState == RecordingState.deviceRecord ||
         captureProvider.recordingState == RecordingState.initialising ||
@@ -342,8 +339,7 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
   }
 
   Widget _buildUnifiedRecordingUI(CaptureProvider provider, Widget? header) {
-    bool isDeviceRecording =
-        provider.havingRecordingDevice &&
+    bool isDeviceRecording = provider.havingRecordingDevice &&
         (provider.recordingState == RecordingState.deviceRecord || provider.recordingState == RecordingState.pause);
 
     // Offline/batch mode: device or phone-mic audio is saved locally with no live
@@ -355,8 +351,7 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
 
     // A phone-mic batch session reports RecordingState.record too; exclude it here so
     // the Live "Listening" card never renders for it (it is handled above).
-    bool isPhoneRecording =
-        !provider.isPhoneMicBatchRecording &&
+    bool isPhoneRecording = !provider.isPhoneMicBatchRecording &&
         (provider.recordingState == RecordingState.record ||
             provider.recordingState == RecordingState.systemAudioRecord ||
             provider.recordingState == RecordingState.initialising ||
@@ -384,14 +379,14 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
     String statusText = isAudioInterrupted
         ? context.l10n.paused
         : isPaused
-        ? (isDeviceRecording ? context.l10n.muted : context.l10n.paused)
-        : hasTerminalTranscriptionFailure
-        ? context.l10n.transcriptionUnavailable
-        : provider.isTranscriptStalled
-        ? context.l10n.transcriptionPausedReconnecting
-        : hasPhotos
-        ? 'Capturing'
-        : context.l10n.listening;
+            ? (isDeviceRecording ? context.l10n.muted : context.l10n.paused)
+            : hasTerminalTranscriptionFailure
+                ? context.l10n.transcriptionUnavailable
+                : provider.isTranscriptStalled
+                    ? context.l10n.transcriptionPausedReconnecting
+                    : hasPhotos
+                        ? 'Capturing'
+                        : context.l10n.listening;
 
     // When recording is active, show the unified UI design
     if (isDeviceRecording || isPhoneRecording) {
@@ -505,22 +500,22 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
                 decoration: BoxDecoration(
                   color: isPaused
                       ? isDeviceRecording
-                            ? const Color(0xFFFE5D50)
-                            : const Color(0xFF7C3AED)
+                          ? const Color(0xFFFE5D50)
+                          : const Color(0xFF7C3AED)
                       : isDeviceRecording
-                      ? const Color(0xFF35343B)
-                      : const Color(0xFFFF9500),
+                          ? const Color(0xFF35343B)
+                          : const Color(0xFFFF9500),
                   shape: BoxShape.circle,
                 ),
                 child: Center(
                   child: FaIcon(
                     isPaused
                         ? isDeviceRecording
-                              ? FontAwesomeIcons.microphoneSlash
-                              : FontAwesomeIcons.play
+                            ? FontAwesomeIcons.microphoneSlash
+                            : FontAwesomeIcons.play
                         : isDeviceRecording
-                        ? FontAwesomeIcons.microphone
-                        : FontAwesomeIcons.pause,
+                            ? FontAwesomeIcons.microphone
+                            : FontAwesomeIcons.pause,
                     color: Colors.white,
                     size: 12,
                   ),
@@ -608,8 +603,8 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
                       storageFull
                           ? context.l10n.paused
                           : muted
-                          ? context.l10n.muted
-                          : context.l10n.recording,
+                              ? context.l10n.muted
+                              : context.l10n.recording,
                       style: const TextStyle(color: Color(0xFFC9CBCF), fontSize: 14, fontWeight: FontWeight.w500),
                     ),
                   ],
@@ -633,8 +628,8 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
             isPendant
                 ? (prefs.pendantDraining ? context.l10n.pendantSyncingRecordings : context.l10n.pendantRecordingNote)
                 : storageFull
-                ? context.l10n.transcribeLaterStorageFull
-                : (muted ? context.l10n.transcribeLaterPaused : context.l10n.transcribeLaterNote),
+                    ? context.l10n.transcribeLaterStorageFull
+                    : (muted ? context.l10n.transcribeLaterPaused : context.l10n.transcribeLaterNote),
             style: TextStyle(color: Colors.grey.shade400, fontSize: 13, height: 1.35),
             maxLines: 2,
             overflow: TextOverflow.ellipsis,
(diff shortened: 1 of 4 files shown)


aryanorastar:
Addressed on \`f9eb4c77\`:

- **The important one**: 15s of *transcript silence* isn't a stall — a listener pausing, or just a quiet room, easily exceeds that with zero new segments. Raised the threshold to 45s so the watchdog stops restarting healthy sockets during ordinary conversation pauses.
- **Fake sync window**: `onConnected()` was seeding `_lastTranscriptProgressAt` to `now()` on every reconnect, even though a fresh connection isn't transcript progress. That defeated `shouldMarkFramesSyncedAfterSocketSend`'s whole contract ("only mark synced while progress is *recent*") for ~10s after every reconnect. Removed the seed.
- **UI precedence**: stall state was only cleared on the next 5s watchdog tick after an explicit pause, so "reconnecting" could outrank Paused/Muted for up to 5s. Now cleared synchronously in `pauseDeviceRecording()`.

Not fixed, flagging honestly rather than overclaiming: the "watchdog might miss a socket already connected before the listener attached" concern is a pre-existing characteristic of how this whole class starts everything off `onConnected()` (not something #6977 introduced), and fixing that properly needs a broader look at the connection-attach lifecycle that's out of scope here.

Verified: updated unit tests for the new threshold + an explicit "quiet room doesn't trip the watchdog" case, plus WAL/batch-recording tests for regressions (48/48), `dart analyze` ratchet passes, `dart format` clean.

Git-on-my-level (commented):
Thanks for this — it is a clean, well-tested approach to the zombie "Listening" problem (#6977). The pure-predicate extraction in `transcript_stall.dart` with the unit test suite is exactly the right shape for testing this contract without spinning up sockets, and the 45s threshold + the grace-window gating on `markFrameSynced` address the real failure mode.

A few observations for the maintainer:

**1. Watchdog startup gap (pre-existing, acknowledged):** `CaptureController.onConnected()` — which calls `_startTranscriptStallWatchdog()` — will not fire on the *initial* socket connection. In `_initiateWebsocket`, `await ...conversation()` resolves only after `PureSocket.connect()` has already synchronously called `onConnected()` (`pure_socket.dart` line 117), but `_socket?.subscribe(this, this)` runs afterward (line 692). So the `TranscriptSegmentSocketService._listeners` map is empty when `onConnected` iterates it on first connect. The watchdog will only start on a subsequent reconnect. You flagged this honestly as out of scope, which is fair — it is a pre-existing characteristic of the connection lifecycle, not something this PR introduced — but it does mean the watchdog will not catch the very first zombie stall of a fresh session until a reconnect cycle kicks in. Worth a maintainer decision on whether a follow-up to start the watchdog in `_initiateWebsocket` after `subscribe` would close that gap cleanly.

**2. WAL frame-sync semantics change:** Gating `markFrameSynced` on recent transcript progress (10s grace via `shouldMarkFramesSyncedAfterSocketSend`) is the correct fix for the zombie scenario — a socket send is a fast path, not durability. But it is a meaningful durability-semantics change: frames sent over a connected-but-not-yet-transcribing socket will now stay in the WAL recovery set rather than being marked synced. That is safer, but a maintainer should confirm the WAL recovery/sync flow handles the larger unsynced-frame set without unexpected UI or storage behavior.

**3. WAL always-on:** Removing the `_socket?.state != connected || unlimitedLocalStorageEnabled` gate so phone WAL stays on for supported BLE device audio is a good durability improvement — a stale "connected" socket can no longer drop the middle of a session. No concern there, just noting it is a behavior broadening.

The unit tests are solid (threshold boundary, conversational-pause cases, paused/not-ready/no-progress guards, grace-window boundary). The manual test plan checkboxes are appropriately unchecked. Labeling for maintainer review given the capture/durability sensitivity and the acknowledged lifecycle gap.

---

aryanorastar:
@Git-on-my-level follow-up on the lifecycle gap you flagged (and the remaining cubic P1s):

1. **Watchdog now arms on first connect** — after `_socket?.subscribe(this, this)` in `_initiateWebsocket`, we start the stall watchdog when `recordingState == deviceRecord`. That closes the race where `PureSocket.connect()` fired `onConnected()` before subscribe, so the initial BLE session never got a watchdog.
2. **Reconnecting UI survives the intentional restart** — when the watchdog calls `stop()`, `onClosed` no longer clears `isTranscriptStalled` if `_transcriptStallRestartInFlight` is set. Stall clears on the next `onConnected` / real transcript progress (not on the close itself), so the UI does not flash Listening mid-restart.
3. **Unified card status precedence** — pause / mute / call-interrupt now outrank stall in `_buildUnifiedRecordingUI` (header path already had pause first).

Still out of scope (as before): distinguishing “quiet room” vs “device frames still arriving but no STT” with a second liveness signal — the 45s transcript-silence threshold remains the intentional tradeoff.

Verified: `flutter test test/unit/transcript_stall_test.dart` (7/7) + local `pr-preflight` green.

kodjima33 (approved):
Stall watchdog for zombie Listening (#6977) is well-motivated and tested; approving, over the auto-merge size gate.

aryanorastar:
@kodjima33 Thanks for the approve on #6977.

All blockers look clear on `eef5efb1`:
- CI green
- Your APPROVE stands
- Watchdog first-subscribe arming + reconnecting-UI survival landed in that tip after David’s lifecycle note

Ready whenever you want to land it — no further author work pending on this head.

Git-on-my-level:
Confirmed on `eef5efb1` — the head addresses the open items from my earlier review:

- **Watchdog first-connect arming:** `_startTranscriptStallWatchdog()` now runs after `_socket?.subscribe(this, this)` in `_initiateWebsocket`, so a fresh BLE session gets a watchdog instead of only arming on reconnect. Good.
- **Threshold:** 45s is the right call — 15s of transcript silence is normal during conversational pauses / quiet rooms, not a dead socket.
- **Reconnecting-UI survival** across the intentional socket restart (`shouldPreserveTranscriptStallAcrossSocketClose`) and the synchronous stall clear on pause both look correct.

WAL frame-sync gating (10s grace via `shouldMarkFramesSyncedAfterSocketSend`) and WAL-always-on for supported BLE device audio are sound durability improvements — a stale connected socket can no longer drop mid-session audio. One thing worth a maintainer gut-check: with frames no longer marked synced on socket send alone, the WAL recovery set will be larger during early/transcribing sessions; confirm the recovery/sync flow handles that without unexpected UI or storage behavior.

Pure predicates + unit tests are exactly the right shape here. CI green; leaving for maintainer merge.

---

aryanorastar:
@Git-on-my-level Thanks for the re-confirm on `eef5efb1`.

Agreed on the WAL gut-check note — frames staying in the recovery set until recent transcript progress is intentional durability; happy for a maintainer to sanity-check recovery/sync UI under a larger early-session unsynced set when landing.

No further author work pending. @kodjima33 already APPROVED over the auto-merge size gate — ready whenever you want to merge.

kodjima33 (approved):
Transcript-stall watchdog stops the zombie Listening state and lets reconnect recover; pure helpers + 150 lines of tests.

aryanorastar:
@kodjima33 Rebased onto latest `main` to clear the merge conflict (`processing_capture.dart` + `capture_controller.dart`).

Resolution notes:
- Kept main’s `isAudioInterrupted` UI path and this PR’s stall-watchdog UI; pause / mute / interrupt still outrank `isTranscriptStalled`
- Preserved stall watchdog / WAL always-on / `markFrameSynced` grace / first-subscribe arming
- Dropped phone-mic batch-stop completer bits that main already reverted

Head: `9f2feff122` — `flutter test test/unit/transcript_stall_test.dart` 7/7. Ready to merge once CI re-greens on this tip.

Git-on-my-level:
Confirmed on the post-rebase head (`9f2feff122`) — all the open items from the earlier review landed cleanly:

- **First-connect arming:** the watchdog now starts after `_socket?.subscribe(this, this)` in `_initiateWebsocket`, so a fresh BLE session gets a watchdog instead of only arming on reconnect.
- **45s threshold** is the right call for transcript silence (a quiet room or a listener easily exceeds 15s).
- **Reconnecting-UI survival** across the intentional socket restart, plus the synchronous stall clear on pause, both look correct.

WAL frame-sync gating (10s grace) and WAL-always-on for supported BLE device audio are sound durability improvements. The one thing worth a maintainer gut-check at merge time: with frames no longer marked synced on socket send alone, the WAL recovery set will be larger during early/transcribing sessions — confirm the recovery/sync flow handles that without unexpected UI or storage behavior.

Pure predicates + the unit suite (7/7) are exactly the right shape. CI green; leaving for maintainer merge.

---

aryanorastar:
Closing for now.

This touches BLE live capture (stall watchdog / zombie Listening). Per our bar we will not ping for merge without on-device smoke. I do not have that hardware available right now, and unit tests alone are not enough.

Happy to reopen after an on-device smoke confirms a stalled BLE live session recovers (or fails closed cleanly) without leaving zombie Listening.

CI was green; approvals stand as historical signal only — reopen when device proof exists.

PR #11097, opened 2026-08-04
fix(app): auto-retry BLE connect after device-ready timeout (#6610)

### LINKED ISSUE #6610: [Limitless Pendant][iOS] BLE reconnect gets stuck ("connecting") after background disconnect, times out ~60s, no auto-retry until force-quit
## Summary

If the Limitless Pendant's BLE connection drops while the iOS app is backgrounded, reopening the app often shows it stuck "Connecting". Logs indicate the app attempts to reconnect, then times out and transitions to `disconnected`, and **does not automatically retry**. Users report it only recovers after force-quitting and reopening the app.

(Separate from conversation chunking issue #6551.)

## Evidence from user logs

- `device_connection_state: connecting` followed by `disconnected` after ~60 seconds (repeatable pattern)
- Example span (UTC):
  - `wal_initialized` 04:39:04Z
  - `device_connection_state: disconnected` 04:40:03.941Z
  - later reconnect succeeds quickly after app restart:
    - `device_connection_state: connecting` 04:53:19.966Z
    - `device_connection_state: connected` 04:53:21.064Z

User also notes multiple occurrences across the day with the same ~60s timeout behavior.

## Expected

If a reconnect attempt fails/times out, app should:
- reset BLE connection state (cancel connection, rescan, recreate central manager if needed), and
- automatically retry (with backoff) instead of stopping.

## Actual

Reconnect attempt times out and the app remains disconnected until force-quit.

## Suspected root cause (from user analysis)

Potential stale `CBCentralManager` / cached peripheral handle state after backgrounding. Force-quit resets BLE stack, making reconnect succeed in ~1-2 seconds.

## Environment

- App: Omi AI 1.0.531 (812)
- Device: iPhone
- OS: iOS 26.4
- Third-party device: Limitless Pendant
- Reporter: Discord @twilsonco

aryanorastar:
## Summary
- Fixes [#6610](https://github.com/BasedHardware/omi/issues/6610): Limitless/iOS BLE can sit in **Connecting**, hit the 60s Dart `deviceReady` timeout, then stay **disconnected with no auto-retry** until force-quit.
- After connect / device-ready failure, `DeviceService` schedules soft reconnects with backoff (`2s → 5s → 10s → 30s → 60s`) **without disposing** the existing `NativeBleTransport` (keeps BleBridge + native manage alive).
- Manual disconnect / forget suppresses retries; successful connect resets the counter.
- iOS: connecting watchdog cancels + re-issues `connect` for peripherals stuck in `.connecting` for ≥65s (background stall that never fires disconnect).
- Pure helpers + unit tests: `app/test/unit/ble_connect_retry_test.dart`.

## Test plan
- [x] `flutter test test/unit/ble_connect_retry_test.dart`
- [ ] Manual (Limitless iOS): background disconnect → reopen → confirm connecting timeout no longer requires force-quit; app retries and reconnects
- [ ] Manual: user disconnect / forget does **not** keep retrying
- [ ] Manual: successful reconnect clears stall and returns to connected without multiple overlapping hard reconnects

## Invariants

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11097?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/ios/Runner/Ble/OmiBleManager.swift
+++ b/app/ios/Runner/Ble/OmiBleManager.swift
@@ -47,6 +47,13 @@ final class OmiBleManager: NSObject {
     /// Periodic check for CoreBluetooth peripherals stuck in `.connecting`.
     private var connectingWatchdogTimer: Timer?
 
+    /// UUIDs the connecting watchdog just canceled and is about to re-issue.
+    /// `didFailToConnect`/`didDisconnectPeripheral` check this so a delegate
+    /// callback triggered by the watchdog's own cancel doesn't *also* schedule
+    /// its own retry — otherwise the same peripheral gets two overlapping
+    /// `connect()` calls for one stall.
+    private var watchdogPendingReissue: Set<String> = []
+
     /// Tracks peripherals that have connected at least once (for reconnection counting).
     private var everConnected: Set<String> = []
 
@@ -136,9 +143,7 @@ final class OmiBleManager: NSObject {
                 NSLog("[OmiBle] connectPeripheral: \(uuid) already connected, skipping")
                 return
             }
-            connectAttemptStartTimes[uuid] = Date()
-            startConnectingWatchdogIfNeeded()
-            centralManager.connect(peripheral, options: nil)
+            issueConnect(peripheral, uuid: uuid)
             return
         }
 
@@ -148,10 +153,18 @@ final class OmiBleManager: NSObject {
         if let peripheral = retrieved.first {
             peripheral.delegate = self
             peripherals[uuid] = peripheral
-            connectAttemptStartTimes[uuid] = Date()
-            startConnectingWatchdogIfNeeded()
-            centralManager.connect(peripheral, options: nil)
+            issueConnect(peripheral, uuid: uuid)
         }
+    }
+
+    /// Single funnel for every `centralManager.connect` issuance (#6610 review:
+    /// a connect issued outside this path is invisible to the watchdog and can
+    /// stall in `.connecting` forever with nothing left to recover it).
+    private func issueConnect(_ peripheral: CBPeripheral, uuid: String) {
+        connectAttemptStartTimes[uuid] = Date()
+        startConnectingWatchdogIfNeeded()
+        peripheral.delegate = self
+        centralManager.connect(peripheral, options: nil)
     }
 
     /// CoreBluetooth can leave a peripheral in `.connecting` after backgrounding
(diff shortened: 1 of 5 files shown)


aryanorastar:
Addressed the review on `5f30e536`:

- **P1** (retries outside the watchdog): funneled every `centralManager.connect` issuance through one `issueConnect()` helper (state restoration, foreground reconnect, and both delegate retry paths now all register with the watchdog, not just the original `connectPeripheral` path).
- **P2** (watchdog cancel → double connect): added `watchdogPendingReissue` so a `didFailToConnect`/`didDisconnectPeripheral` callback caused by the watchdog's own cancel defers to the watchdog's re-issue instead of scheduling a second, overlapping connect.
- **P2** (stale retry can override a newer device): this was the one I'd flag as the real bug — `_bleConnectRetryDeviceId` was set but never actually read anywhere, so a retry scheduled for a device the user has since moved away from could fire later, force-reconnect to the *old* device, and tear down + replace a connection just established to a different one. Added `shouldInvalidatePendingRetryForDifferentTarget` (unit tested) and check it at the top of `ensureConnection`.
- **P3** (attempt counter not reset): `stop()`/`disconnectDevice()`/`forgetDevice()` now call `_resetBleConnectRetryState()` instead of just cancelling the timer, so a later reconnect restarts backoff at 2s instead of resuming from wherever it left off.
- **P3** (watchdog timer never stops after manual disconnect with no callback): now clears the tracking entry inline for manually-disconnected UUIDs instead of relying on a delegate callback that isn't guaranteed to fire.
- **P3** (Dart constants only exercised by tests): left as-is with a cross-reference comment — it's a real but low-severity drift risk across the platform-channel boundary, and I'd rather keep the executable Dart spec than delete it.

Verified: `flutter test test/unit/ble_connect_retry_test.dart` (11/11, 3 new), plus `device_provider_test.dart` + `ble_bridge_test.dart` for regressions (26/26 combined), `dart analyze`/`dart format` clean. Swift changes syntax-parsed against the iOS SDK; I could not run a full Xcode build in this environment (no generated `.xcworkspace`, and this PR's CI doesn't gate on iOS compilation) — flagging that honestly so a maintainer with a full iOS toolchain gives it a real device/simulator pass before merge.

Git-on-my-level:
Thanks @aryanorastar — this is a well-scoped fix for #6610 and the review follow-up on `5f30e536` addresses the real gaps cleanly. Quick verification pass against the current head:

What checks out
- The retry chain is complete end-to-end: `NativeBleTransport`'s 60s device-ready `TimeoutException` → wrapped into `DeviceConnectionException` → caught in `ensureConnection` → `_scheduleBleConnectRetry`. Soft retries reuse the existing transport (no `dispose()`), so native auto-reconnect / BleBridge registration survives.
- Every `centralManager.connect` issuance now funnels through `issueConnect()` (connectPeripheral known/retrieved, state restoration, foreground reconnect, and both delegate retry paths) — so the watchdog sees all attempts.
- `watchdogPendingReissue` correctly prevents the double-connect: when the watchdog's `cancelPeripheralConnection` routes through `didFailToConnect`/`didDisconnectPeripheral`, those handlers defer to the watchdog's own 200ms re-issue instead of scheduling a second overlapping `connect()`.
- `shouldInvalidatePendingRetryForDifferentTarget` closes the stale-retry-replaces-newer-connection race, and `_resetBleConnectRetryState` (not just timer cancel) is wired into `stop`/`disconnectDevice`/`forgetDevice`/connect-success so backoff restarts at 2s.

Observations for a maintainer (non-blocking)
- The 65s watchdog / 60s device-ready constants are duplicated across `OmiBleManager.swift`, `ble_connect_retry.dart`, and `native_ble_transport.dart` with no shared source of truth. The unit tests document the contract but can't catch drift across the platform-channel boundary. Worth a note for a future pass; I agree with keeping the executable spec for now.
- The iOS watchdog and the Dart↔native retry interaction are only covered by pure-helper unit tests — the actual recovery path still relies on the manual Limitless/iOS steps in the test plan. Grep-verified the logic, but device-level validation is still needed before this reliably ships.

Leaving as a positive signal rather than an approval — BLE reliability behavior on a core device surface warrants a human maintainer's product/reliability judgment and the manual device validation above. CI is green.



---

mdmohsin7:
#6610 is a stale issue, it was already fixed, closing this PR for now.

PR #11101, opened 2026-08-04
fix(listen): guarantee parity capture teardown persistence

Git-on-my-level:
## What changed and why

The live listen runtime initialized parity capture during bootstrap, but called
`persist()` near the end of teardown after several fallible cleanup operations.
If usage persistence or another earlier teardown dependency raised, teardown
exited before the cassette write and GCS export seam. This exactly reproduces a
path where a real audio/transcript session can complete without exporting a
cassette.

Make parity capture persistence the unconditional final action of teardown.
The original cleanup exception remains visible, while capture persistence stays
fail-open and is attempted exactly once. The finalizer runs through
`run_blocking(storage_executor, ...)`, keeping cassette I/O and GCS upload off
the event loop. The normal supervisor-exit log no longer includes the task name
because that name contains the Firebase UID.

Add a hermetic runtime regression that constructs `ListenSessionRuntime`, runs
the real bootstrap and `ListenReceiver` STT callback attachment, sends client
audio through a fake WebSocket and fake STT socket, disconnects, and verifies
the cassette plus a fake exporter target. The adversarial case injects an
earlier teardown persistence failure and proves the offloaded cassette/export
finalizer still runs exactly once. The test asserts a production-valid
direction multiset, gives unexpected fake receives a deterministic diagnostic,
and guards UID, session ID, transcript, audio, credential, local-root, GCS URI,
and object-name absence from logs.

Closes SCA-292

## Reproduction

RED was observed on fresh `origin/main` (`c66679071a`):

- normal close wrote/exported the cassette but exposed the allowlisted UID in
  `Listen supervisor exited ... task=ws:<uid>:receive`;
- the injected teardown persistence failure raised before
  `ListenParityCapture.persist()`, so the required cassette/export assertions
  were never reached;
- on review correction, the strengthened regression failed both runtime cases
  because persistence had zero `storage_executor` offloads.

none

## Verification

- `PYTHONPATH=backend backend/.venv/bin/python -m pytest backend/tests/unit/test_listen_parity_runtime.py -q` — 2 passed.
- `PYTHONPATH=backend backend/.venv/bin/python -m pytest backend/tests/unit/test_listen_parity_runtime.py backend/tests/unit/test_listen_parity_capture.py backend/tests/unit/test_listen_runtime_regressions.py backend/tests/unit/test_listen_persistence.py backend/tests/unit/test_backend_listen_helm_defaults.py -q` — 41 passed.
- `PYTHON=.venv/bin/python bash test-preflight.sh` (from `backend/`) — 18 passed, 0 failed; optional integration/provider warnings only.
- `PYTHON=.venv/bin/python BACKEND_UNIT_TEST_FILE_LIST=../backend-test-files.txt bash test.sh` (from `backend/`, temporary list containing the five files above) — 41 passed under the strict timing guard.
- `scripts/pr-preflight --pr-body-file pr-body.md` — repository PR-body/failure-class checks.
- `OMI_PR_BODY_FILE=pr-body.md make preflight` — repository deterministic preflight.

A bare all-files `bash test.sh` was also attempted on the initial head; it
selected 821 files and was stopped after unchanged tests tripped the strict
0.12-second local CPU duration guard. The focused runner above uses the same
test harness and timing guard for every affected file.

No container/filesystem change was needed; the existing dev Helm `fsGroup:
10001` contract remains covered. No IAM, GCS permissions, production config,
Prometheus instrumentation, deployment, or merge is included.

## Failure class (fixes)

Failure-Class: new

## Remaining live gate

After review and an authorized dev rollout, run one allowlisted `/v4/listen`
session through audio, transcript, and close, then confirm exactly one new
cassette under the configured dev GCS prefix.


Git-on-my-level:
@cursor fix CI and review comments and merge conflict

PR #10969, opened 2026-08-01
feat(desktop): Google OAuth connector for Gmail and Calendar

undivisible:
## What

Add a Google OAuth connector to the macOS desktop app, replacing the browser-cookie path for Gmail and Calendar when a user connects an account. One grant per Google account; the settings UI lists them with per-account disconnect, and the existing Gmail/Calendar readers automatically prefer OAuth grants over cookies.

Depends on #10963 (both touch the settings integrations section; #10963 should merge first).

## Why

The cookie connector silently reads whichever browser session exists first (the "junk inbox" problem) and requires a signed-in Chromium browser. OAuth gives a real consent flow, works without a browser, and supports multiple accounts cleanly.

## How

- PKCE authorization-code flow with a loopback redirect server (`127.0.0.1:0`), `access_type=offline`, optional client secret for Google's newer Desktop-app client type.
- Grants stored in the login keychain via the existing `DesktopKeychainStore`, keyed per account (email from the OpenID userinfo endpoint). Reconnecting the same account replaces its grant.
- New reader paths over the Gmail and Calendar APIs (metadata only — subject/sender, no bodies); `GmailReaderService.readRecentEmails` and `CalendarReaderService.readEvents` route through OAuth when a grant exists and fall back to cookies otherwise, so onboarding, imports, and automation keep working unchanged.
- Settings → Advanced → Email & Calendar is now reachable (it existed in the enum but was never rendered) and hosts the OAuth accounts card plus the cookie account picker.

Invariants: INV-INT-1 (functional probe at connect; no agent/heuristic automation; tokens in keychain only).

## Product invariants affected

- INV-INT-1

## Review round: empty Gmail/Calendar list + merge with main

Addresses the one open blocker from the last review.

**Empty list is a successful read, not an error.** `GoogleOAuthGmailReader.readRecentEmails` treated a 200 response without a `messages` array as `.invalidResponse`. Gmail's `users.messages.list` omits `messages` entirely when the query matches nothing, so any account with no mail in the probe window failed `connect()` and `verifyConnection()` — both probe with `newer_than:1d` — despite working auth, scopes, and APIs.

`GoogleOAuthCalendarReader.readEvents` had the identical defect on `items`: `events.list` omits `items` for a window with no events, and `connect()` probes a one-day calendar window. Both list paths now read a missing or empty collection as an empty result. Non-JSON bodies still fail as `.invalidResponse`, and 401/403 still surface as reconnect-required, so this does not mask auth failures.

[...]

PR #11115, opened 2026-08-05
fix(desktop): refresh macOS connectors and read every Apple Note

Archit-lal:
## Problem

Omi's macOS import connectors do not stay up to date, and Apple Notes barely imports at all.

Measured on a real user's machine (prefs domain `com.omi.computer-macos`):

| Connector | Last synced | Result |
|---|---|---|
| Apple Notes | 19 days earlier | 75 sources — the library has **836 notes** |
| Email | 34 days earlier | 170 messages → 0 memories |
| Google Calendar | *no keys at all* | never synced once |

Two independent root causes.

**1. Nothing ever refreshes a connector.** Every call site of `AppleNotesReaderService` / `CalendarReaderService` is a human action — onboarding, Settings, "Sync now", or a chat tool. A search of the desktop codebase for a connector scheduler, timer, or refresh loop returned nothing. `ImportConnectorStatusStore` persisted only a UserDefaults timestamp, so "Connected" was a latch, not live state — which `INV-INT-1` explicitly forbids ("Gate 'Connected' on a timestamp latch or a one-time success").

**2. Apple Notes read the wrong thing, through a door most users cannot open.** `AppleNotesReaderService` opened `~/Library/Group Containers/group.com.apple.notes/NoteStore.sqlite` directly with GRDB. That path is TCC-protected and needs Full Disk Access. Worse, the query selected `WHERE ZNOTE IS NOT NULL`, which matches rows that *reference* a note (attachments) rather than notes — hence the `isLikelyAttachment()` heuristic that discarded ~90% of what it fetched, and hence 75 of 836. It also read only `ZTITLE`/`ZSUMMARY`; note **bodies were never read at all**. The remembered folder grant was stored as a plain path string rather than a security-scoped bookmark, so access died on relaunch — the code said so itself in a comment.

Also fixed: `OnboardingPagedIntroCoordinator` called `readRecentNotes(userInitiated: false)` while that method's first statement was `guard userInitiated else { throw }`. That auto-scan had never executed once.

## What changed

**Apple Notes now drives Notes.app over Apple Events (JXA), not the protected store.** No Full Disk Access. Measured against the live 836-note library: manifest read 0.26s, full export of every note with body text 1.69s. Incremental sync diffs a cheap manifest in Swift and only fetches bodies when something changed, so a steady-state background tick costs one ~0.2s call.

- Reads **all** notes with full body text (8000-char cap; 16 of 834 truncate), folder attribution, created/modified dates.
- Password-protected notes are skipped and *counted* — the import message says "(N locked notes skipped)" rather than silently under-importing.
- Note keys drop the CoreData store UUID (`x-coredata://<uuid>/ICNote/p1443` → `p1443`) so external IDs survive an iCloud re-sync or OS migration.
- The folder picker, the path-string persistence, and `isLikelyAttachment` are deleted outright — no compatibility shim.

**A background connector refresh scheduler.** Per-connector cadence with catch-up, exponential backoff with jitter, single-flight delegated to the existing `ConnectorImportRunner`, and at most one connector per tick.

The safety property that governs the whole design: **a background tick can never raise a macOS consent prompt.** Eligibility is a declared per-connector `supportsUnattendedRefresh`, and it is the highest-precedence skip in the policy. Today: Apple Notes `true` (the Apple Events grant persists), Local Files computed from a proven prior grant, **Calendar `false`** — its cookie read fails without a user-initiated window, which would park a healthy connector. That flips to `true` in one line once #10969 lands.

**Background refreshes do raw idempotent import only — never LLM synthesis.** The backend dedupes raw artifacts on `external_id` + `content_hash` (`backend/database/memory_imports.py` `_artifact_id` → `create()` → `AlreadyExists` → deduped), but synthesis emits items with no `external_id`, so varying model output would mint fresh duplicate artifacts on every refresh forever. This repo already took a mass-duplication incident that needed a server-side purge (see `import_write_guard.py`).

**"Connected" is now a live probe, not a latch** (`INV-INT-1`). The Apple Notes `requiresVerification` latch and the permanent "Reconnect to verify access" nag are gone; status comes from a prompt-free probe. A connector parked on `needsUserAction` renders that state instead of claiming a fresh sync time.

## Relationship to #10969

This PR deliberately does **not** migrate Calendar/Gmail to OAuth — that is #10969 (issue #10459), which is active. Zero files in this diff overlap that PR's 25. `DefaultsKey.swift`, `OmiApp.swift`, and `DesktopDiagnosticsManager`'s area list were all avoided by design (runtime-computed keys following the existing `ImportConnectorStatusStore` precedent; registration in `ViewModelContainer`; reuse of already-allowed fallback labels).

## Verification

```
xcrun swift build -c debug --package-path Desktop      -> Build complete!, 0 errors 0 warnings
./scripts/swift-test-suites.sh                          -> 483 suites, 3948 tests
xcrun swift test --filter ConnectorRefresh              -> Executed 43 tests, 0 failures
xcrun swift test --filter AppleNotes                    -> Executed 31 tests, 0 failures
./scripts/swift-format-wrapper.sh lint-scope            -> clean
./scripts/swiftlint-wrapper.sh lint                     -> 0 violations in 1064 files
check_desktop_test_quality.py                           -> at/below baseline (55/150, 19 waits)
check_product_file_line_count_ratchet.py                -> OK across 16 changed files
check_brand_ui.py                                       -> OK, no purple increase
check-e2e-flow-coverage.py --strict                     -> 0 uncovered
desktop-flow-lint.py                                    -> OK (68 flows, 152 actions)
desktop-changelog.py validate                           -> OK
```

Exercised the real path: extracted both JXA scripts and ran them against the live Notes library — 836 manifest ids, 834 exported notes, `lockedSkipped=2`, `folderMode=mapped`, max body 8000, 16 truncated. Manifest keys are a strict superset of export keys by exactly the 2 locked notes.

**Known pre-existing flake, not from this PR.** `AgentSyncBatchQueryTests.testDelayedOwnerATokenCannotResumeIntoOwnerBVM` is a concurrency race that sequences an actor handoff with `await Task.yield()`. Measured 10 runs on each tree: **clean `origin/main` 3/10 fail, this branch 4/10 fail** — indistinguishable. Not touched here, since quarantining an unrelated upstream test from a feature PR is scope creep.

**Not done:** verified by build + unit tests + direct script execution against the live library, not by a full named-bundle click-through of the Apps tab and onboarding. Worth a live pass before merge.

- INV-INT-1

The core change is what this invariant demands: "Connected" moves from a stored timestamp latch to a live functional probe, and the guard doc (`desktop/macos/docs/connector-checklist.md`) is updated in the same PR, since its Apple Notes section described the deleted folder-selection recovery.

Failure-Class: none

The violated contract here — a persisted one-time success rendered as live connector state, with no owner for periodic refresh — is already owned by `INV-INT-1` in the invariant registry rather than by a failure class, and the guard surfaces shipping with this change are the live probe plus the behavioral suites (`ConnectorRefreshPolicyTests`, `ConnectorRefreshSchedulerTests`, `AppleNotesReaderServiceTests`). Happy to declare a new class instead if a maintainer prefers that framing.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11115?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Archit-lal:
## CI status: this PR's own code is green

`Desktop Swift Static & Test Contracts` is red, but the failure is not from this diff.

```
Ran 483 Swift suites in isolation with 2 worker(s), 300s per-suite budget.
FAILED Swift suites: AgentSyncBatchQueryTests
```

**482 of 483 suites pass**, including all 74 tests added here (`ConnectorRefresh*` 43, `AppleNotes*` 31). `Desktop Swift Build & Tests` is red only because it gates on the suite job.

`AgentSyncBatchQueryTests` is a pre-existing flake on `main`, filed as #11119 with a measured rate: **3/10 failures on a clean `origin/main` checkout** versus 4/10 on this branch — indistinguishable. The file is not in this diff, and neither it nor `AgentSyncService` references any symbol this PR changes. It also reproduced after `rerun-failed-jobs`, so it is not a one-off.

Not quarantining or editing it from this PR — that would be scope creep into an unrelated subsystem. #11119 has the suggested fix (replace the bare `await Task.yield()` with a wait on the existing actor gate).

Everything that validates this change passed: PR Metadata Preflight, Hygiene, contracts, desktop-core-e2e-t0, Backend Hermetic Merge Gate, swift-format, SwiftLint, the line-count/brand/test-quality ratchets, and e2e flow coverage.

Git-on-my-level (changes requested):
Thanks for the thorough write-up and the test coverage here. The overall direction (moving Apple Notes off direct CoreData scraping and adding an unattended-refresh policy layer) looks promising, but I need to request changes before this can be considered mergeable.

Blocking items I verified on the current head:

1. Apple Notes incremental state can acknowledge notes before the raw import succeeds. `syncChangedNotes()` updates and persists entries for changed notes before `runAppleNotesImport()` calls `saveAsMemories()`, and `runAppleNotesImport()` does not turn `rawImport.failed` into a failed sync. A transient backend/write failure can therefore make changed notes look imported and cause later background ticks to skip them.

2. Passive Apple Notes refresh can be blocked when Notes.app is simply not running. `prepareAutomation(userInitiated: false)` rejects non-`noErr` permission status before attempting `launchNotesInBackground()`. If the preflight returns the stopped-target status, the already-granted background path never gets to launch Notes and refresh.

3. The onboarding Apple Notes path uses `sync.changed` as the displayed/imported note count. When the manifest is unchanged, `syncChangedNotes()` intentionally returns `changed: []` with a nonzero `totalNotes`, so onboarding can render an already-synced Notes library as zero notes / still connectable.

There are also CI blockers on this head (`Desktop Swift Build & Tests` and `Desktop Swift Static & Test Contracts`). I saw the note that the failing suite may be a main-branch flake, but this PR still needs a green or maintainer-accepted CI story before merge.

One maintainer-facing documentation note: `desktop/macos/Desktop/Sources/Integrations/BackgroundRefresh/README.md` is agent-readable subsystem guidance that will likely shape future AI coding/review agents working in this area. The safety intent is good (no background prompts, raw import only), but because some implementation details above diverge from the documented invariants, please keep the README synchronized with the fixed behavior and have a human maintainer review the background-refresh policy/permission model before landing.

---

aryanorastar:
Leaving this open. Main still reads Apple Notes through GRDB NoteStore.sqlite and has no unattended JXA refresh, so the gap is real.

This SHA is CONFLICTING and not rebaseable (thousands behind). CHANGES_REQUESTED on 631d4d3 still applies: sync can ack notes before saveAsMemories, refresh dies if Notes.app is closed, onboarding still calls readRecentNotes(userInitiated: false). Desktop Swift CI is red on this head.

Do not approve. Do not conflict-resolve this tree onto QueryShell. A follow-up needs those three blockers fixed on current connector code.

aryanorastar:
Keeping this open. Not approving `631d4d3dd855`. `Integrations/AppleNotes/AppleNotesReaderService.swift` absent on origin/main (root `AppleNotesReaderService.swift` is a different path). mergeable CONFLICTING / DIRTY. Checks FAILURE. +5027.

PR #11117, opened 2026-08-05
feat(desktop): rebuild the macOS app on a glass design system, and fix onboarding's dead ends

Archit-lal:
Rebuilds the macOS app on a glassmorphic design system modelled on the Context-for-Claude UI, and fixes three product defects found while doing it.

## Why this is not a reskin

`OmiColors` was a hardcoded **dark** palette (`textPrimary` `#FFFFFF`, grounds near-black). On a light-pinned glass panel every one of its ~1,900 call sites resolved to the panel's own colour — surfaces that compiled, drew, and were **blank**. So the conversion had to be total: a half-converted app is an invisible app, not an inconsistent one.

Sign-in and the chat onboarding were the sharpest case. Each painted a full-bleed dark photograph over the shell's ground, and their white type survived *only because* that art hid the panel. Removing either alone yields an empty screen; both had to go together.

## What changed

**The design system** (`Sources/Theme/{Ink,InkType,InkGlass,WindowGlass}.swift`) — behind-window `NSVisualEffectView(.hudWindow)` pinned to `.aqua`, 0.14 scrim, 22pt continuous corners, ambient shadow (r34 / 0.24 / y−10), and a semantic palette of `labelColor × alpha` so one value works on any ground. Two rules are enforced rather than documented: **on glass the type ladder has two rungs, not three** (tertiary fails AA on the shipped ground), and the accent pins `systemBlue` rather than `controlAccentColor`, which macOS lets users set to the off-brand hue — a defect no rendered-pixel check can catch.

Diverges from the reference in one place, deliberately: the panel sheen is clipped by the panel's own shape so it cannot square off the top corners.

**Typography** — Open Runde for display (≥22pt), SF Pro for reading, em-based tracking recomputed per size. Bundled **with its OFL-1.1 licence text**; the reference implementation ships these faces without one.

**Onboarding** — a six-beat cinematic on the eight-dot Omi mark (dots arrive one at a time on their own spring and click, then hand off to the comet pulse), an `AVAudioEngine` ambient bed that carries from the intro into the chat onboarding, and an exit that finally answers "what do I do now". Reduce Motion swaps a *different timing table* rather than branching. The intro plays once — never on a resume (granting Screen Recording restarts the process by design), never on a second run — and the flag is written before playback so a crash cannot re-arm it. **All audio is generated by a committed script; nothing is licensed.**

**Push-to-talk** is now clickable from the composer and the ask bar, driving the existing `PushToTalkManager` rather than a second audio path.

## Defects fixed (each was proven from code before it was fixed)

**Post-onboarding dead end.** `PostOnboardingPromptSuggestions.save(…)` was called only from the retired onboarding (`OnboardingView.swift:632`), so the live flow left the array empty, `DashboardPage.swift:489`'s guard could never fire, and `ChatProvider.presentOnboardingOpener()` built the Chat tab's starter chips from nothing. The feature existed and was orphaned. Guidance is now setup-aware and reads the user's **real** shortcut chords, so someone who rebound Ask-Omi is not told to press ⌘O, and the hold-to-talk cue is suppressed entirely if the mic was denied.

**Six onboarding permission defects.** The Automation prompt blocked the main thread behind a modal dialog (`NSAppleScript` on `@MainActor`); a grant landing after the 20s poll was never detected (no reactivation re-check, and Allow was `.disabled` while waiting); a Screen Recording grant never offered the relaunch it needs, so capture stayed dead while `.screenDemo` claimed "I can see it"; system audio looped with no escape; Automation state was read one cycle stale so an existing grant was re-asked; and AX / event-tap / filesystem probes ran on the main thread every 500ms. The pre-existing tests here asserted *source strings* against a view the app never renders.

**A latent voice-turn bug**, found while wiring the button: `enterLockedListening()` published `.lock` into any active turn, but the reducer only accepts `.lock` from `.recording`/`.pendingLockDecision`. From a response phase it was an invalid transition and the manager still opened a mic for a turn that never entered a recording phase.

Also fixed on the way: three filter rows had silently lost their active state when two dark tokens collapsed onto one wash; `OmiToggleStyle`'s "on" track was white, so every switch in Settings read as off; `opaqueGround` multiplied into existing alpha so Reduce Transparency never became opaque; the markdown renderer drew assistant text white on a light bubble; and `DesktopHomeView` re-stamped `.darkAqua` on every window, unpinning the glass it had just set.

## Home was rebuilt on top of the design system

The glass work above landed first; the surfaces below were built on it afterwards.

**Home is a query surface, not a hub.** Two glass objects with 12pt between them — a
query bar you type into, and a results panel that answers. `QueryShellLayout.panelGap`
is *derived* from `RewindSearchLayout.panelGap` rather than restated, so the product has
one opinion about how far apart its glass sits, and both panels take the top bar's own
lane so all three share a leading edge. `⌘⏎` swaps the panel body in place for an answer
thread through the one `ChatProvider`; the top bar's right side is icons only.

**The results panel is one chronological spine.** Conversations, memories, and screen
moments interleave by time under sticky day headers, with an hour rail that tracks the
topmost visible row. They are siblings rather than separate tabs because a conversation,
the memory it produced, and the screenshot from that minute belong together.

**Rewind's timeline is time-linear.** The old strip mapped a pixel to an *array index*,
so a three-hour gap occupied the same width as a three-second one. X now maps to the
capture's timestamp and the frame under the pointer comes from a binary search. Apps get
a deterministic colour identity (FNV-1a over the app name into seven bands, all outside
the forbidden hue range, every swatch ≥3:1 against white type) replacing a
`hashValue`-seeded full-wheel mapping that changed colours on every relaunch *and* handed
an `INV-UI-1`-forbidden hue to roughly a third of all apps.

**Interface sounds**, gated three ways: chrome cues honour macOS's own "Play user
interface sound effects" (read by hand — neither `AVAudioPlayer` nor `AVAudioEngine`
consults it), content cues ignore it, motion-tied cues respect Reduce Motion, and one
Settings switch silences everything including onboarding. No new assets.

## Defects the *build* surfaced that reading could not

These were all found by driving the running app, and are recorded because each one
compiles and reads correctly:

- **`⌘⏎` never fired.** The branch tested `press.modifiers.contains(.command)` inside
  `onKeyPress`, but AppKit routes Command-modified keys to key-equivalent handling before
  `onKeyPress` sees them. Unreachable code that looks right. Now a real `keyboardShortcut`.
- **The Ask button was unpressable** — a whole-bar `contentShape` + `onTapGesture` (tap
  anywhere to focus) is the outermost gesture and won every click landing on a button.
- **`ChatProvider.errorMessage` was surfaced nowhere in `MainWindow`.** The provider has
  two error shapes; only `currentError` had a renderer. A crashed agent runtime takes the
  other branch, so the panel showed the question and sat silent forever. Both shapes now
  render through the shared `AgentErrorClassifier`. This silence was app-wide, not new here.
- **Archived memories leaked onto Home.** The archive and device-scope guardrails lived
  inside one page's private recompute, so a second reader structurally could not obey them.
  Fixed by ownership — `MemoryVisibilityGuardrails` is now the single policy with two
  published projections — not by a call-site filter.
- **A day read before Rewind's database opened was cached as "no capture" forever**,
  which is why the moment count read low. Bounded readiness wait; the count now agrees
  with the database (569 screenshots, 135 today, 419 yesterday).
- **The Rewind search surface was dead code.** `RewindSearchBar`,
  `SearchResultsFilmstrip`, and `ScreenshotThumbnailView` had zero references outside a
  lint baseline; the live surface was inline in `RewindPage`. Rebuilt *and* rewired.

## Performance

Measured, not asserted. Glass compositing was profiled and found innocent
(+0.065 ms/frame — the blur is WindowServer work, not app work). The real costs were
elsewhere:

| | before | after |
|---|---|---|
| Rewind scrub, per chunk | 728.4 ms | **59.3 ms** (12.3×) |
| `AG::Graph::UpdateStack::update()` under typing | 5752 ms | **3083 ms** (−46%) |
| `DashboardPage.body` per keystroke | 309 ms | **0 ms** |
| Timeline drag, main-thread busy samples | 825 / 4931 | **174 / 5085** |

Causes: `extractFrameWithAVFoundation` reopened an `AVAssetReader` and walked from frame 0
per request (quadratic in chunk length), and handed frames to the UI as re-encoded JPEG;
the composer draft was `@Published` on the object the whole shell observes; and every
`mouseMoved` rebuilt an `NSWindow` + `NSHostingView` for the timeline tooltip.

## Invariants

- **INV-UI-1** — purple and `Color.accentColor` removed from the converted surfaces; the ratchet passes across 172 changed UI files. `FolderColors` dropped its Indigo/Purple swatches (stored user data keeps its hex — repainting it is the worse defect).
- **INV-NAV-1** — nav chrome was restyled, not reduced; parity guards green.
- **INV-6 / INV-VOICE-1** — no second `ChatProvider`, transcript, or turn handler; the mic button is a trigger, not a write path.

- INV-AUTH-1
- INV-CHAT-1
- INV-VOICE-1
- INV-INT-1
- INV-MEM-1

Merging `upstream/main` pulled the branch across four more locked path globs than the
pre-merge diff touched. Naming them is path-based, per `AGENTS.md`; none of the four has
a behavior change here:

- **INV-AUTH-1** — `SessionRecoveryView`, `OmiApp`, `ChatProvider`, `FileIndexerService`
  were restyled onto `Ink`/`InkGlass`. No change to session acquisition, storage, refresh,
  or the recovery path.
- **INV-CHAT-1** — the composer's draft moved off `ChatProvider`'s `@Published` surface
  into `ChatComposerDraft` for the keystroke-invalidation fix. `ChatProvider.draftText`
  is still the accessor and still the single owner; one transcript, one send path.
- **INV-INT-1** — connector chrome (`ConnectorBrandIcon`, the guidance overlay, the memory
  export sheet) repainted only. No connector added, removed, or re-scoped.
- **INV-MEM-1** — the Brain Map's atlas view takes upstream #11089's label truncation and
  this branch's palette swap. Tier assignment and graph boundary are untouched.

Failure-Class: none

## Verification

- `xcrun swift build -c debug --package-path Desktop` — clean, no warnings from changed files.
- `xcrun swift test --filter "Glass|Ink|Onboarding|PushToTalk|ChatFirst|MemoryGraph|ChatPromptTimeline|Markdown|Apps"` — **416 tests, 0 failures.**
- `make preflight` — 10 diff-scoped checks pass (`diff-hygiene`, `architecture-guardrails`, `product-invariants`, `failure-class-protocol`, `deferred-work-markers`, `lifecycle-headers`, …).
- `check_brand_ui.py` — OK, no purple increase across 172 UI files.
- `check_product_file_line_count_ratchet.py` — OK across 184 files. **Every baseline moved down, none up** (`TasksPage` 6633→6618, `OnboardingChatView` 2182→2156, `MemoriesPage` 3630→3605, `DesktopHomeView` 1885→1869, `SidebarView` 1616→1611, `CanonicalMemoryAtlasView` 4427→4425, `ChatBubble` 1998→1997).
- `check-e2e-flow-coverage.py --strict` — 181 covered, **0 uncovered**.
- The cinematic was **run and captured frame by frame** across all six beats.
- Zero new `TODO`/`FIXME`/`HACK` markers.

**Two tests were rewritten in this PR** (AGENTS.md treats that as suspect, so stating both explicitly):
1. `ChatPromptTimelineTests.testTheActiveMarkIsLitRatherThanResized` asserted `markColor == .white`. External source: `InkGlass`'s pinned-`.aqua` decision makes `labelColor` resolve near-black, so a white literal was only ever correct because the rail sat on a near-black page. It now asserts the two states are the *same* colour — the invariant the test's own name claims.
2. `MemoryGraphRevisitTests.testMemoryDropdownPillsUseAnOpaqueOmiSurface` pinned retired palette token names. It now asserts the contract its own comment states — occlusion plus neutral pill styling — against the shared `GlassPillBackground`.

## Two "inherited failures" resolved

An earlier revision of this description listed two failures as inherited from `main`.
Both are now settled and neither needs an exception:

- `check_agents_md_lean.py` — `backend/AGENTS.md` was 39089 bytes at the merge base and is
  38990 on `upstream/main`. Merging fixed it. This branch has no commit touching the file.
- `check-sources-root-layout.py` — **not wired into any CI lane** (no manifest entry, no
  workflow reference, not in the Makefile), so it gates nothing. Of the four files above
  its stale baseline, one was ours; it now lives in a feature directory.

## Verified live, and what still is not

Every surface here was driven in a named local bundle on the real account, not just built:
the query bar and its controls, the chip seam, push-to-talk, the spine against 973 real
conversations and 569 screen captures, Rewind's track and search grid, and the onboarding
cinematic frame by frame.

Two gaps stated plainly rather than implied:

- **Ask mode's success path has never rendered an answer on this branch.** Every chat turn
  in the local bundle dies with `pi-mono process exited (code 1)` — a bundled agent-runtime
  fault, not this change. Its *failure* path is verified live. Citation chips are code-live
  but unverified for that reason.
- Dark Mode, a real trackpad pinch (gesture `NSEvent`s cannot be synthesised), and the
  what's-new / usage-limit sounds (not triggerable on demand) were not exercised.

https://claude.ai/code/session_01Bro5DpYLqufeFKMYRbdY3Q

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks for the very cohesive macOS design/onboarding push — the direction is understandable, and I’m not asking for this to be split just because it is broad. I do need to hold it before merge because the current head is not buildable and a few user-facing regressions look real on static review.

Blocking items I verified:

- The Desktop Swift release compile currently fails in CI in `OmiOnboardingSound.swift`: `static let canonicalFormat = AVAudioFormat(...)` is rejected under Swift concurrency because `AVAudioFormat` is non-Sendable. That needs to be fixed and the desktop Swift build/test jobs need to go green before this can land.
- `ChatInputView` now defaults `showsPushToTalk` to true, and `TaskChatPanel` uses the shared composer without opting out. That makes the task-thread composer show a mic button that drives the global `PushToTalkManager`/floating-bar flow instead of `TaskChatState.sendMessage`, so voice input can land in the wrong conversation context until task routing is explicitly supported.
- Several selected/primary controls are rendering the same token for foreground and fill. For example, `RewindSearchBar` uses `Ink.primary` for both selected filter text and selected chip backgrounds, and the Rewind retry button uses `Ink.primary` on `Ink.primary`; those states become unreadable depending on appearance. Please audit the primary/selected states touched by the glass rewrite and use the inverted foreground token where the fill is primary.
- The new push-to-talk quota projection depends on `FloatingBarUsageLimiter.serverQuota` / `optimisticDelta`, but those are not published; `recordQuery()` and `applyQuota()` can change the blocked state without causing existing mic buttons to re-render. Please make quota changes observable or explicitly notify before relying on that state for UI.

Also: this PR updates `desktop/macos/AGENTS.md` with bundled-resource guidance for coding agents. The guidance looks plausible and useful, but because it changes instructions that future AI coding/review agents will load, a human maintainer should confirm the resource-location and manifest-cache guidance is accurate enough to become repo policy.

Once the compile failure and these concrete behavior/contrast issues are fixed, this should be much easier to review as a coherent macOS surface redesign.

---

Archit-lal:
## Update — the first version of this PR was broken, and here is what was wrong

The original PR was opened having visually verified exactly **one** screen (sign-in, a self-contained centred card — the least representative surface in the app). Running it revealed the app was unusable. Two distinct root causes, both found by looking rather than reasoning:

**1. Home rendered as a black slab with black text.** `HomeCanvasBackground` was still painting a literal dark gradient with `.ignoresSafeArea()` beneath Home, while every colour on the page had been remapped to resolve *dark* against the light panel. The token file's own comment already said "the page paints nothing: the glass panel owns the one ground" — the canvas contradicted it. Removing it exposed everything it had been hiding: a `white@0.14` Stop button (invisible exactly when it is the only control that matters), a `.white` "Show less" with no way back out of an expanded reply, ask-bar hover vs focus differing by ~0.4/255, and ~30 more.

**2. Every window had a 32pt bare strip at the top.** The window is `.hiddenTitleBar`, so its `NSHostingView` reports `safeAreaInsets.top == 32`, and a SwiftUI `.background{}` lays out *inside* that inset. Measured by pixel-scanning the window over a black backdrop: `y=0…31 → (0,0,0)`, then `y=32+ → (137,137,138)` continuous. The ground moved to AppKit — `InkGlassView` is now the window's `contentView` and the hosting view re-parents into it, which is how the reference implementation does it — so it is correct by construction rather than by modifier ordering. Its regression test asserts against **content-view bounds, not `contentLayoutRect`**; the latter already excludes the title bar and would have passed on the broken shape.

### Also fixed after looking
- "Granted" rendered **green-on-green** on permission rows (`systemGreen` is ~1.6:1 on this panel). A saturated system hue cannot be darkened, so the state moved to a 6pt disc and the word took `Ink.primary`.
- Filter chips on Conversations/Memories/Tasks had **silently lost their active state** when two dark tokens collapsed onto one wash. Its test now asserts the *ordering* (measured alpha gap) rather than a token identity — pinning identity is what let the defect through the first time.
- Four permission rows carried three different badge designs; white numeral on `Ink.hairline`; `OmiToggleStyle`'s "on" track was white so every switch read as off; the markdown renderer drew assistant text white on a light bubble.

Controls were exercised live, not read: a real send returning a reply, push-to-talk start→recording→finalized, nav pills, gear, show more/less, notification frequency 0→4, language en→es, shortcut ⌘O→⌘J.

### Still open — do not treat this as finished
- The **floating bar, notch pill, and full onboarding walkthrough have not been visually verified.** Three audit agents died mid-run (two stalled, one API error); their work was salvaged, validated and committed, but their visual passes did not complete.
- Disabled `OmiButtonStyle` is grey-on-grey (shared token, app-wide blast radius).
- 28 dead `Divider().background(...)` calls — `.background` cannot tint a `Divider`.
- Plan card nests a card inside a card; Settings still on the old 260px sidebar rather than the reference metrics.
- `ChatPromptTimeline` reads as a second scrollbar at high mark counts.
- Message column vs composer misaligned by ~54pt (pre-existing, not conversion-caused).

**Operational note worth capturing in the runbook:** the documented Sparkle re-sign fix (re-sign frameworks *and* app with `disable-library-validation`) breaks the bundle's keychain ACL and drops it to signed-out. Re-signing **frameworks only** avoids both the dyld crash and the auth loss.

Git-on-my-level (changes requested):
Thanks for the follow-up work here. The current head is in a better place than the prior review — the desktop Swift build/release checks are green now — but I still need to keep this in changes-requested because several user-facing regressions are still present on the current diff.

Blocking items I verified on this head:

- `TaskChatPanel` still constructs `ChatInputView` without opting out of the new default `showsPushToTalk: true`. That means the task-thread composer can show the mic button even though the click path drives the global `PushToTalkManager` / floating-bar flow, not `TaskChatState.sendMessage`, so voice can still go to the wrong conversation context.
- `FloatingBarUsageLimiter.serverQuota` and `optimisticDelta` are still plain `private(set)` properties. `PushToTalkButtonState.blocked` depends on `isLimitReached`, but `applyQuota()` and `recordQuery()` can change that value without publishing, so already-rendered mic buttons can stay visually enabled after the user hits the free-tier limit.
- There are still primary-on-primary contrast regressions in changed glass surfaces. Examples: `RewindPage`'s error-state Retry button and Rebuild Index button use `Ink.primary` for both text and fill, and the Apps Add Review sheet submit button/spinner also uses `Ink.primary` over an `Ink.primary` fill. Those controls can render blank in the state where the user most needs them.

Also noting the agent-instruction change: this PR updates `desktop/macos/AGENTS.md` with bundled-resource and fast-lane runtime-payload guidance. The guidance looks useful and not malicious, but it will affect future AI coding/review agents working in the macOS tree, so a human maintainer should confirm the SwiftPM resource-manifest and runtime-payload policy is accurate enough to become repo guidance.

I’m not asking for a split: the PR is broad, but it is cohesive around the macOS glass/onboarding rebuild. Once the remaining routing/observability/contrast blockers are fixed, this should be reviewable as one product-surface change.

---

PR #11121, opened 2026-08-05
fix(desktop): make delayed-owner-token sync guard deterministic (#11119)

### LINKED ISSUE #11119: AgentSyncBatchQueryTests.testDelayedOwnerATokenCannotResumeIntoOwnerBVM is flaky on main (~30% failure)
`AgentSyncBatchQueryTests.testDelayedOwnerATokenCannotResumeIntoOwnerBVM` fails intermittently on `main`, independent of any PR content. It blocks `Desktop Swift Static & Test Contracts` (and the dependent `Desktop Swift Build & Tests`) roughly a third of the time.

**Measured flake rate**, 10 consecutive isolated runs per tree via `xcrun swift test --package-path Desktop --filter AgentSyncBatchQueryTests` on macOS 26.5.2 / Xcode 26.2:

| Tree | Result |
|---|---|
| clean `origin/main` @ `b009279849` | **3/10 fail** |
| an unrelated connector branch | **4/10 fail** |

Statistically indistinguishable — the failure is present on a pristine base checkout with no local changes.

**Assertion:**
```
AgentSyncBatchQueryTests.swift:349: XCTAssertEqual failed:
("Optional("owner-a-vm")") is not equal to ("Optional("owner-b-vm")")
```

**Likely cause:** the test sequences an owner-A → owner-B `AgentSyncService` handoff and then relies on a single `await Task.yield()` to let owner-B's request start before asserting `requestURLs.first?.host`. A bare `Task.yield()` is not a guarantee that a specific continuation has resumed — under different cooperative-pool scheduling, owner-A's delayed request is still the first recorded URL. Runs that pass complete in ~0.005s; runs that fail take ~1.3s, consistent with the timing being scheduler-dependent rather than logical.

Suggested fix: replace the `Task.yield()` with a deterministic wait on the existing actor gate (the test already has `gate.waitForRequest()` / `waitUntilFirstFetchStarts()` primitives) so the assertion cannot run before owner-B's request is recorded.

**Observed on:** https://github.com/BasedHardware/omi/pull/11115 — 482 of 483 suites pass; this is the only failure, and it reproduced on a re-run of the failed jobs. Filing separately rather than touching the test from an unrelated connector PR.

kodjima33:
Fixes #11119.

## Bug

`AgentSyncBatchQueryTests.testDelayedOwnerATokenCannotResumeIntoOwnerBVM` fails intermittently on `main` (~30% in CI), blocking `Desktop Swift Static & Test Contracts` and the dependent `Desktop Swift Build & Tests` with no PR content involved. Measured on a clean `origin/main` checkout in isolation: **19 of 20 runs fail**, always at the same assertion:

```
AgentSyncBatchQueryTests.swift:361: XCTAssertEqual failed:
("Optional("owner-a-vm")") is not equal to ("Optional("owner-b-vm")")
```

## Root cause

The harness, not `AgentSyncService`. The test spawned the owner-A `stop()` into a detached `Task` and then released owner A's suspended token fetch immediately, with nothing ordering the two:

```swift
let stopOwnerA = Task { await service.stop(flushPendingChanges: false) }
await gate.releaseFirstFetch()   // may win the race against stop()
```

`stop()` bumps `syncGeneration` before it suspends on `await task?.value`, and `refreshFirebaseToken` re-checks the generation only *after* its `fetchIDToken()` await. When the release won the race, owner A resumed while its generation was still current, passed `isCurrent(...)`, and issued a real request to `owner-a-vm`. `waitForRequest()` then returned on owner A's request rather than owner B's — so the assertion read `owner-a-vm` as the first (and only) recorded URL. Failing runs took ~2.7s, passing runs ~0.005s: scheduler-dependent, exactly as the issue diagnosed.

## Fix (test-only — `AgentSyncService` is untouched)

The owner boundary already has an observable: `stop()` cancels the sync task *after* bumping the generation, so cancellation of the task that owns the first fetch proves the handoff happened. The gate records that cancellation via `withTaskCancellationHandler`, and the test waits on it before releasing the delayed token — which is what "delayed" was always supposed to mean. The trailing bare `Task.yield()` is dropped: `waitForRequest()` is now reached only with owner B's request pending, so it blocks until that request is recorded instead of returning on owner A's leak. No polling, no sleeps, no new production surface.

## What I tested

macOS 26.5.2 / Xcode 26.2, `xcrun swift test --package-path Desktop --filter AgentSyncBatchQueryTests/testDelayedOwnerATokenCannotResumeIntoOwnerBVM`:

| Tree | Result |
|---|---|
| clean `origin/main` @ `342290651f` | **19/20 FAIL** (failures ~2.7s, passes ~0.005s) |
| this branch | **30/30 PASS**, every run ~0.007s |

The guard still has teeth — this is a determinism fix, not a weakened assertion: deleting the post-await `guard isCurrent(generation:ownerID:vmIP:)` from `refreshFirebaseToken` makes the test fail **5/5** with the original `owner-a-vm != owner-b-vm` message, then restored. Also `xcrun swift test --package-path Desktop --filter AgentSync`: 16/16 pass.

Product invariants: none affected.

Failure-Class: none

Test-harness sequencing defect; no production contract changed and no production behavior fixed, so there is no semantic failure class to declare or open.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11121?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Tests/AgentSyncBatchQueryTests.swift
+++ b/desktop/macos/Desktop/Tests/AgentSyncBatchQueryTests.swift
@@ -9,6 +9,8 @@ private actor AgentSyncDelayedTokenGate {
   private var firstFetchStarted = false
   private var firstFetchWaiters: [CheckedContinuation<Void, Never>] = []
   private var firstFetchContinuation: CheckedContinuation<Void, Never>?
+  private var firstFetchCancelled = false
+  private var firstFetchCancelWaiters: [CheckedContinuation<Void, Never>] = []
   private var requests: [URLRequest] = []
   private var requestWaiters: [CheckedContinuation<Void, Never>] = []
 
@@ -37,6 +39,26 @@ private actor AgentSyncDelayedTokenGate {
   func releaseFirstFetch() {
     firstFetchContinuation?.resume()
     firstFetchContinuation = nil
+  }
+
+  /// `stop()` bumps the sync generation and only then cancels the sync task, so
+  /// the cancellation of the task that owns the first fetch is the observable
+  /// proof that the owner boundary has already moved. Waiting on it before
+  /// releasing the delayed token is what makes "the token resumes *after* the
+  /// handoff" a fact rather than a scheduling coincidence.
+  func noteFirstFetchCancelled() {
+    guard !firstFetchCancelled else { return }
+    firstFetchCancelled = true
+    let waiters = firstFetchCancelWaiters
+    firstFetchCancelWaiters.removeAll()
+    waiters.forEach { $0.resume() }
+  }
+
+  func waitUntilFirstFetchCancelled() async {
+    guard !firstFetchCancelled else { return }
+    await withCheckedContinuation { continuation in
+      firstFetchCancelWaiters.append(continuation)
+    }
   }
 
   func respond(to request: URLRequest) -> (Data, URLResponse) {
@@ -339,7 +361,13 @@ final class AgentSyncBatchQueryTests: XCTestCase {
     let gate = AgentSyncDelayedTokenGate()
     let service = AgentSyncService(
       networkHooks: AgentSyncService.NetworkHooks(
-        fetchIDToken: { await gate.fetchToken() },
+        fetchIDToken: {
+          await withTaskCancellationHandler {
+            await gate.fetchToken()
+          } onCancel: {
+            Task { await gate.noteFirstFetchCancelled() }
+          }
+        },
         dataForRequest: { request in await gate.respond(to: request) },
         reuploadDatabase: { _, _ in true },
         now: Date.init,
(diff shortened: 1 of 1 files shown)


Maslam-art (commented):
A Resource-Based Economy: A Pathway to Global Sustainability

Humanity faces an escalating, multi-dimensional crisis, including climate change, resource depletion, and systemic inequality. These issues are exacerbated by a monetary-based capitalist system that prioritizes infinite growth and profit over human welfare. Current political and economic frameworks are insufficient to address these technical challenges, as they rely on outdated ideologies that fail to account for the physical limits of the planet.

A Resource-Based Economy (RBE), as proposed by The Venus Project, offers a transition to a moneyless, sustainable system. In this model, Earth's resources are treated as the common heritage of all inhabitants, managed through advanced technology, automation, and scientific methodology. By replacing financial incentives with resource-based metrics and cybernetic governance, the system seeks to eliminate poverty, reduce environmental damage, and prioritize human development over capital accumulation.

The implementation of an RBE requires a staged transition: global cooperation, a comprehensive resource inventory, and the guaranteed fulfillment of basic human needs. Success depends on shifting societal values and overcoming psychological resistance to abandoning traditional monetary habits. By leveraging AI, clean energy, and circular production, this model aims to transform civilization into a restorative, efficient, and equitable society.

PR #11124, opened 2026-08-05
fix(app): hold capture on native BLE reconnect, not a grace timer (#6678)

### LINKED ISSUE #6678: BLE: constant device disconnects (dozens/day) visible in Device Diagnostics
## Problem
Multiple users are seeing **constant BLE disconnects** throughout the day (**dozens of disconnect events/day**).

These disconnects are visible in the **Device Diagnostics** screen (BLE diagnostics), and users can provide screenshots.

## Expected
Stable BLE connection during normal use, with occasional disconnects only when out of range / Bluetooth off.

## Actual
Repeated disconnects during the day, causing unreliable capture/sync and poor UX.

## Evidence
- Device Diagnostics screen shows frequent disconnect events
- Reporter will attach a screenshot

## Notes / What to attach
Please comment with:
- Phone model + OS version
- Omi app version/build
- Device firmware version
- Device Diagnostics export/screenshot (disconnect reasons, RSSI timeline)
- Whether battery optimizations are disabled for Omi

## Priority
High (impacts core reliability; affects multiple users).

aryanorastar:
## Summary
- Transient BLE drops were ending live capture (via `updateRecordingDevice(null)`) while Android/iOS native layers were still auto-reconnecting, spawning conversation shards (#6678).
- Native now reports `willAutoReconnect` on `onPeripheralDisconnected`. Dart clears connection UI / WAL / charging immediately, but **holds capture** until reconnect succeeds or a final disconnect arrives.
- Drops the previous guessed grace timer (3500ms / 1500ms) that measured the wrong endpoint (retry schedule ≠ link restore), raced multi-retry, and delayed every disconnect side effect.

## Native contract
- Android: `willAutoReconnect` mirrors `handleRetryLogic` gates (`managedDevices` present, not destroying, status not `-1`/`137`, BT on).
- iOS: `willAutoReconnect` matches the existing auto-reconnect predicates (`!manual && !pairingLost`, plus `everConnected` on fail-to-connect).

## Test plan
- [x] `flutter test test/providers/device_provider_test.dart test/services/bridges/ble_bridge_test.dart` → pass (reconnecting holds capture; final releases; reconnect clears hold without release; bridge forwards flag)
- [ ] CI Mobile App Checks green
- [ ] Not verified on-device: RF blip mid-session with Omi CV1 (no hardware here)

none

## Failure-Class
Failure-Class: none

Fixes #6678


aryanorastar:
Addressed cubic’s two points on `0d18d4e8`:

1. **Wire the helper into production** — the disconnect debouncer callback now gates through `shouldApplyDisconnectCaptureSideEffects(stillDisconnected: !isConnected)` so reconnect (which flips `isConnected` / cancels the timer) can’t still tear down capture if the callback races cancel.
2. **Reconnect-cancel coverage** — `ble_disconnect_grace_test.dart` now uses `fakeAsync` + `Debouncer` to prove disconnect→reconnect before grace does **not** invoke teardown, grace without reconnect does, and the residual stillDisconnected race no-ops.

Also fixed **Dart Analyze** (`dangling_library_doc_comments`) by adding an explicit `library;` so the file header doc is attached — that was the analyzer-ratchet failure.

`flutter test test/unit/ble_disconnect_grace_test.dart` → 6 passed.

Git-on-my-level (dismissed):
Found a real correctness bug in the production wiring of the guard — requesting changes.

The grace-delay concept (lengthening the debouncer past native auto-reconnect) is sound and well-documented. The cancel-on-reconnect path works correctly. But the residual-race guard introduced to address cubic's feedback has a circular dependency that breaks real disconnects.

**The bug**

`onDeviceConnectionStateChanged` gates teardown through:

```dart
_disconnectDebouncer.run(() {
  if (!shouldApplyDisconnectCaptureSideEffects(stillDisconnected: !isConnected)) {
    return;
  }
  onDeviceDisconnected();
});
```

`shouldApplyDisconnectCaptureSideEffects(stillDisconnected:)` is `stillDisconnected`. So the effective guard is "skip teardown if `isConnected` is still true." The problem: `isConnected` is **only** set to `false` inside `onDeviceDisconnected()` (via `setIsConnected(false)` at line ~492) — the function this guard gates.

On a real disconnect (device genuinely leaves, no reconnect within 3.5s):
1. Disconnect event schedules the debouncer; `isConnected` is still `true`.
2. 3.5s elapses with no reconnect → callback fires.
3. `isConnected` is still `true` (nothing else set it to `false`).
4. `!isConnected` → `false` → guard returns early.
5. `onDeviceDisconnected()` is **never called**.

That means capture teardown, BLE listener cleanup, charging-status cancel, `setIsConnected(false)`, analytics, and `onRecordingDeviceDisconnected()` are all skipped. The provider permanently believes the device is connected after a real drop — which is a worse failure mode than the conversation-shard problem this PR is fixing.

The reconnect-cancel path does work: a `connected` event calls `_disconnectDebouncer.cancel()` before the timer fires, so transient blips are correctly ignored. The bug is specifically the residual-race guard that was added to satisfy the "wire the helper into production" feedback.

**Why the tests don't catch it**

The unit tests pass because they control `stillDisconnected` as a local variable rather than modeling `DeviceProvider.isConnected`'s actual lifecycle (where nothing flips it to `false` until after teardown runs). A test that started from `isConnected = true` and asserted teardown runs after grace elapses would fail.

**Suggested fix**

Track the disconnect as a separate field so the guard checks an independent signal, e.g.:

```dart
DateTime? _disconnectStartedAt;
// on disconnected:
_disconnectStartedAt = DateTime.now();
_disconnectDebouncer.run(() { onDeviceDisconnected(); });
// on connected:
_disconnectStartedAt = null;
```

…and have the guard verify `_disconnectStartedAt != null` and (optionally) that the grace window has actually elapsed. Alternatively, since the debouncer is already cancelled on reconnect, the residual fire/cancel race is genuinely narrow, and the original timing-only debouncer may be sufficient — but if a guard is wanted it must not depend on `isConnected`, which can't change until after the guarded path runs.

Nice documentation on the native reconnect windows, and the helper extraction itself is clean. Would appreciate a follow-up that fixes the guard and adds a test modeling the real `isConnected` lifecycle so this regression can't hide again.

---

aryanorastar:
Addressing @Git-on-my-level  requested changes on the current head:

- Fixed the circular dependency: the disconnect debouncer guard no longer derives `stillDisconnected` from `!isConnected` (which stays true until `onDeviceDisconnected()` runs). Instead it tracks an independent `_disconnectStartedAt` marker and gates via that.
- Added a DeviceProvider fake-timer state-transition test that models the real lifecycle:
  - `disconnected` schedules teardown after grace even when `isConnected` remains true until teardown fires.
  - `connected` fired just before grace cancels the scheduled teardown.

Verified locally: `flutter test test/unit/ble_disconnect_grace_test.dart test/providers/device_provider_test.dart` (all pass).

kodjima33 (approved):
The 500ms teardown vs 3s native reconnect race is real and the helper is clean and unit-tested. Approving, but not merging: main just reverted the automatic phone-mic fallback (#11134), so the stated symptom path is stale, and the app maintainer owns this disconnect path.

aryanorastar:

Your circular-dependency CR was addressed on `592a367c` (independent `_disconnectStartedAt` marker; DeviceProvider fakeAsync lifecycle test proving grace teardown still runs while `isConnected` remains true until teardown). CI is green; @kodjima33 approved.

Please dismiss the stale Changes Requested / re-review this head.

@mdmohsin7 also needs code-owner APPROVE when you have a moment. Note on kodjima33’s #11134 comment: phone-mic fallback was reverted, but this PR’s core value still holds — waiting past the native reconnect window so RF blips don’t tear down live capture and spawn short conversation shards (#6678).

mdmohsin7 (changes requested):
Requesting changes. The problem this targets is real and the two native constants it cites are accurate, but the fix as written doesn't close the race and changes far more than the description says.

Verified before writing this:
- `OmiBleForegroundService.kt:47` → `RECONNECT_DELAY_MS = 3_000L` ✅
- `OmiBleManager.swift:673` → `asyncAfter(.now() + .milliseconds(200))` ✅
- the previous 500ms debouncer ✅

The circular-dependency bug from the earlier review is genuinely fixed by `_disconnectStartedAt`. Four things still block this.

**1. The 3500ms is measured from the wrong endpoint, so the race stays open.**

`RECONNECT_DELAY_MS` is when Android *schedules* the retry, not when the link is back:

```kotlin
handler.postDelayed(runnable, RECONNECT_DELAY_MS)   // → connectToDevice(addr, "retry_N")
```

After that fires you still need GATT connect + service discovery, and then Dart's own `_connectDebouncer` (100ms) before `setIsConnected(true)` cancels anything. That is routinely 1–3s more. A 3500ms grace therefore leaves the reconnect roughly 500ms to complete end to end — on the common Android path the timer expires while the reconnect is still in flight, teardown runs anyway, and the shards persist. What we've bought is 3.5s of extra latency on every genuine disconnect.

`handleRetryLogic` also retries repeatedly (`managed.retryCount++`), so attempt #2 lands around 6s and #3 around 9s. The grace covers neither.

**2. This is not a capture grace — it delays every disconnect side effect.**

The helper is named `shouldApplyDisconnectCaptureSideEffects`, but it gates `onDeviceDisconnected()` in full:

```dart
_bleChargingStatusListener?.cancel();
setConnectedDevice(null);
setIsConnected(false);
ServiceManager.instance().wal.getSyncs().sdcard.setDevice(null);
ServiceManager.instance().wal.getSyncs().flashPage.setDevice(null);
```

For the whole grace window the app believes a dead device is live: connected UI, charging status still polling, and the sdcard/flashPage WAL syncs still holding a device handle. That is a much wider blast radius than "capture teardown", and this is the area where non-obvious breakage costs us recordings. If a timer stays, it has to be scoped to capture alone — `isConnected`, charging, and the WAL device handles should still clear immediately.

**3. The extracted helper asserts nothing.**

```dart
bool shouldApplyDisconnectCaptureSideEffects({required bool stillDisconnected}) => stillDisconnected;
```

It returns its argument, and the call site already does the real check inline (`_disconnectStartedAt != null`). The 84 lines of tests around it read as thorough coverage while verifying that `f(true) == true`. Please drop the helper and put the tests on the actual behaviour: a real disconnect with no reconnect must tear down after the grace, and a reconnect inside the grace must not.

**4. The stated harm is partly gone, and the branch needs a rebase.**

The headline justification — blips "turned RF blips into phone-mic fallback" and started `omibatchphoneauto` — was `onRecordingDeviceDisconnected()`, which was reverted in #11134. This diff is based on the post-#11084 blob (`31be4d92d6`), so it needs a rebase onto current `main` and a rewritten rationale.

What survives is conversation shards from session teardown. That is a real problem, but the PR doesn't evidence it: "dozens of short conversation shards per day" needs logs, a query, or a diagnostics export attached, not an assertion.

**What I'd like instead**

Don't guess a timeout. The native layer already knows the difference between "reconnecting" and "gone" — Android schedules the retry and keeps `managedDevices[addr]`; iOS holds an indefinite pending `centralManager.connect`. Surface that as an explicit state (`disconnected_retrying` vs `disconnected_final`) and let Dart hold capture open until the reconnect actually resolves. That removes the magic number, which is currently wrong on both platforms, and it degrades correctly when a reconnect takes longer than we guessed.

If the timer approach is kept for now, then at minimum: derive the window from reconnect *completion* rather than the retry delay, scope it to capture only, and cover the multi-retry case.

Happy to look again once it's rebased — please loop me in before this merges, it's in the BLE/offline-sync paths.

aryanorastar:
@mdmohsin7 Addressed your CHANGES_REQUESTED on head `0d0e5274` (rebased onto current main post-#11134).

**What changed vs the grace-timer design you blocked:**

1. **No guessed timeout.** Pigeon `onPeripheralDisconnected` now takes `willAutoReconnect`. Android sets it from the same gates as `handleRetryLogic`; iOS from the same predicates as the existing auto-reconnect `connect` call. Multi-retry is covered because Dart stays in hold until native reports final (`false`) or `connected`.
2. **Capture-only hold.** `reconnecting` → `_applyImmediateDisconnectSideEffects()` (UI / WAL / charging / `isConnected=false`) without `updateRecordingDevice(null)`. Final `disconnected` → `_releaseHeldCapture()`. Successful reconnect clears the hold with no capture teardown.
3. **Dropped the tautological helper + grace file.** Tests assert real DeviceProvider behaviour (hold / release / reconnect), plus BleBridge flag forwarding.
4. **Rationale rewritten** for conversation shards from early `updateRecordingDevice(null)` — phone-mic fallback justification removed (#11134).

Please re-review when you can. Looping you before merge as requested.

@Git-on-my-level the timer path (and the old `isConnected` circular guard) is gone; this is the native-state redesign Mohsin asked for — happy to have another pass when useful.

aryanorastar:
**Preflight / Hygiene:** `diff-hygiene` failed on trailing whitespace in regenerated `pigeon_communicator.g.dart` (and matching Swift/Kotlin stubs). Stripped trailing whitespace and pushed.

aryanorastar:
Added the missing fakeAsync reconnect-cancellation coverage requested by cubic P2 on the current head.

Test added/covered:
- app/test/providers/device_provider_test.dart
- test name: reconnecting then connected cancels any disconnect-capture teardown race (fakeAsync)

Verified locally:
- flutter test test/providers/device_provider_test.dart (pass)

Could you please re-review / clear the needs-tests request?

aryanorastar:
heads-up: the current PR head is 064f85a8ef3a (after the native-state redesign).

In app/lib/providers/device_provider.dart, the logic is now driven by native willAutoReconnect state (reconnecting/connected/disconnected + _captureHeldForReconnect). The earlier _disconnectStartedAt / grace-timer residual guard isn’t present on this head.

To address cubic's test request explicitly, I also added a fakeAsync DeviceProvider reconnect-cancellation regression test on this head:
- app/test/providers/device_provider_test.dart
- test name: reconnecting then connected cancels any disconnect-capture teardown race (fakeAsync)
- verified locally: flutter test test/providers/device_provider_test.dart

Could you please re-review this latest head and dismiss CHANGES_REQUESTED if the race is fully closed?

aryanorastar:
@mdmohsin7 need human response — re-review request.

Head `@064f85a8ef` replaces the grace-timer design you blocked with native `willAutoReconnect` state (Pigeon iOS/Android → `reconnecting` holds capture, final `disconnected` releases). Cubic P2 fakeAsync coverage included. CI fully green (incl. Android Compile Smoke). kodjima33 already APPROVED. Could you re-pass?

Git-on-my-level:
Re-reviewing head `064f85a8` (the native-state redesign). The grace-timer design that carried the circular-dependency bug and the `shouldApplyDisconnectCaptureSideEffects` no-op helper is entirely gone on this head, so I'm dismissing my own stale Changes Requested (review `4862669022`, commit `0d18d4e8`) — its blocker no longer exists in the codebase.

**What I verified on this head**

The redesign directly implements what @mdmohsin7 asked for on `592a367c`:
1. No guessed timeout — Pigeon `onPeripheralDisconnected` now carries `willAutoReconnect` from the same gates Android/iOS already use to schedule reconnect, so Dart holds capture until native reports a resolution rather than racing a magic number.
2. Immediate side-effects — `_applyImmediateDisconnectSideEffects()` clears connection UI, charging listener, WAL device handles, and `isConnected` immediately on `reconnecting`, with capture held separately. mdmohsin7's "blast radius" concern is addressed.
3. The extracted helper is removed; the five new `device_provider_test.dart` cases assert real state transitions (`reconnecting` holds capture, `connected` clears hold without release, `disconnected` releases once), including a `fakeAsync` reconnect-cancel regression test.

**The one thing I checked carefully — can capture be held forever?**

The hold is bounded by native's retry lifetime, which itself is bounded by real escape hatches I confirmed in the native code:
- Android `willScheduleAutoReconnect` returns `false` (→ Dart releases) once the device is unmanaged, `isDestroying`, BT off, or status is `-1`/`137`. Each failed retry re-enters `handleDisconnection` and re-emits the flag, so a device that never comes back eventually hits one of these (BT toggle, app/service teardown, pairing loss) and capture releases.
- iOS emits `willAutoReconnect: false` on manual disconnect or `pairingLost`; otherwise the chipset-level pending `connect` either succeeds (`connected` clears the hold) or the peripheral is removed.

This matches the contract documented in the Pigeon doc comments. The unbounded Android retry count is pre-existing behavior this PR surfaces rather than introduces, so it's not a regression.

**Checks:** Dart Analyze & Tests, Android Compile Smoke, Generated Files, Hygiene, Formatting, and the backend merge gate are all green.

**Status:** Removing `needs-tests` (the core state machine is now covered). Leaving this for @mdmohsin7's code-owner re-review on the current head — happy to defer to your call on whether the native `willAutoReconnect` predicates match your intent for the BLE/offline-sync paths.

---

aryanorastar:
@mdmohsin7 need human response — code-owner re-review

Head remains `064f85a8ef` (native `willAutoReconnect` redesign you asked for).

@Git-on-my-level just dismissed their stale CR and left a **positive-signal** after verifying this head implements your ask:
- no guessed timeout
- immediate UI/WAL/charging clear; capture held separately
- helper removed; DeviceProvider state-transition + fakeAsync coverage
- CI green (Dart Analyze, Android Compile Smoke, Generated Files, Hygiene)

Merge is only blocked by your earlier CHANGES_REQUESTED on the old grace-timer design. Could you re-pass / APPROVE on this head when you have a moment?

aryanorastar:
Closing for now.

This touches BLE / live capture. Per our bar we will not escalate @mdmohsin7 or push for merge without on-device smoke (RF blip mid-session on real hardware). I do not have that hardware available right now, and unit tests alone are not enough.

Happy to reopen after an on-device smoke confirms:
- reconnecting holds capture (no conversation shards)
- final disconnect still releases capture + immediate UI/WAL/charging clear

CI was green on head `064f85a8`; native `willAutoReconnect` redesign remains the intended approach when we return.

PR #11133, opened 2026-08-05
feat(memory): add bounded canonical legacy backfill

Git-on-my-level:
## What changed

Add a whitelist-only, checkpointed, bounded legacy-to-canonical staging page.
It uses the existing migration state machine, resumes safely after interruption,
and deliberately leaves enrollment, promotion, and graph assertions to their
respective canonical owners.

- INV-MEM-1

Failure-Class: none

## Verification

- `backend/.venv/bin/python -m pytest -q backend/tests/unit/test_canonical_legacy_backfill.py`
- `make preflight` with this PR body

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11133?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/testing/workflow_contracts.json
+++ b/backend/testing/workflow_contracts.json
@@ -124,13 +124,15 @@
         "backend/utils/memory/legacy_backfill_bulk_support.py",
         "backend/utils/memory/legacy_backfill_inventory.py",
         "backend/utils/memory/bulk_legacy_backfill.py",
+        "backend/utils/memory/canonical_legacy_backfill.py",
         "backend/scripts/backfill_legacy_memories.py",
         "backend/scripts/bulk_backfill_legacy_memories.py"
       ],
       "tests": [
         "tests/unit/test_ws_c_backfill.py",
         "tests/unit/test_bulk_legacy_backfill.py",
-        "tests/unit/test_bulk_backfill_legacy_memories_cli.py"
+        "tests/unit/test_bulk_backfill_legacy_memories_cli.py",
+        "tests/unit/test_canonical_legacy_backfill.py"
       ],
       "checks": ["no_large_tuple_results"],
       "invariants": [
--- a/backend/tests/unit/test_canonical_legacy_backfill.py
+++ b/backend/tests/unit/test_canonical_legacy_backfill.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import inspect
 from dataclasses import dataclass
+from types import SimpleNamespace
 
 import pytest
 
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Addressed both trusted review threads on this orchestration seam:

**dry_run fail-safe default (Cubic P2, thread PRRT_kwDOLkKqys6WjsyH)**
`CanonicalLegacyBackfillConfig.dry_run` now defaults to `True`, matching the `BulkMigrationConfig` convention it wraps. An unparameterized `run_canonical_legacy_backfill_page()` call inventories only and writes no checkpoints or staged data.

**Enrollment guard before checkpoint advancement (Codex P1, thread PRRT_kwDOLkKqys6WjvNN)**
The no-op enrollment hook is replaced with a cohort-membership assertion. The bulk state machine treats a returning `enroll_fn` as write-readiness and advances the checkpoint toward terminal `read_ready`; this seam is not the enrollment owner, so it now raises `ValueError` for any uid not in the code-defined canonical whitelist. This prevents a non-enrolled uid from reaching terminal `read_ready` and permanently blocking the real enrollment owner.

Added 4 regression tests: dry_run default assertion, default-config no-write path, and enrollment guard accept/reject. Backend unit suite: 10 passed.

Git-on-my-level:
Not implementing: this seam's contract (documented in the module docstring and PR body) is that it "does not change rollout documents or read grants" and "leaves enrollment, promotion, and graph assertions to their respective canonical owners." Writing the write-stage control document (`build_rollout_documents(stage="write")`) is the bulk CLI's enrollment responsibility — not this orchestration seam's. The `_cohort_enrollment_hook` is deliberately a membership gate only: it raises if the uid is outside the code-defined whitelist, which prevents a non-cohort uid from advancing to terminal `read_ready`. A cohort member whose `memory_control/state` is blocked has a separate readiness owner that must clear independently.

Git-on-my-level:
Addressed the two follow-up trusted review threads:

**Register in workflow contracts (Codex P2, thread PRRT_kwDOLkKqys6WmCfc) — resolved**
Added `backend/utils/memory/canonical_legacy_backfill.py` to the `legacy_memory_backfill` workflow contract sources and `tests/unit/test_canonical_legacy_backfill.py` to its tests list. Future edits to the helper now select its checkpoint/enrollment/page-bound regression coverage instead of falling back to the generic INV-MEM suite. Added a selection assertion to `test_workflow_contracts.py`. Contract check passes; 34 tests green.

**Write-stage control doc in enrollment hook (Codex P1, thread PRRT_kwDOLkKqys6WmCfS) — not implemented**
This seam's stated contract is that it does not change rollout documents and leaves enrollment to the canonical owner. Writing the write-stage control doc is the bulk CLI's responsibility, not this orchestration seam's. The `_cohort_enrollment_hook` is deliberately a cohort-membership gate only. See the inline acknowledgement for full rationale.

kodjima33 (approved):
Bounded canonical legacy backfill reads cleanly and is well covered — approving; backend feature so leaving the merge call to Nik.

PR #11140, opened 2026-08-05
fix(memories): restore default-scope deletion

Git-on-my-level:
## Summary

- Restore the desktop Memories menu action for deleting default memories.
- Add an explicit `scope=default` backend contract that tombstones canonical Short-term and Long-term memories while leaving Archive untouched.
- Keep legacy accounts and the dual-read rollout mirror behavior safe, and leave the unscoped visibility actions disabled.

## Root cause

The macOS bulk mutation capability was hardcoded to `false` during the tiered-memory rollout (`fcc20de16e4`). That disabled the Delete Default Memories button before it could issue a request. The existing unscoped DELETE endpoint also was not safe to re-enable because canonical delete-all includes Archive, so the fix adds an explicit scoped server contract instead of reusing the unscoped operation.

## Invariants and failure class

- `INV-MEM-1`: default access remains Short-term + Long-term; Archive is excluded unless explicitly targeted.

## Verification

- `backend/.venv/bin/pytest -q tests/unit/test_memories_batch_delete.py tests/unit/test_ws_j_delete_privacy.py` — 63 passed
- `xcrun swift test -c debug --package-path Desktop --filter APIClientMemoryBulkSafetyTests` — 12 passed
- app-client OpenAPI export/check and compatibility against `origin/main` — passed
- `make preflight` — passed
- Oversized-file ratchets were updated with one-line ownership justifications for the canonical adapter and Memories page, as required by the repository guard.
- Built and launched isolated `omi-memory-delete` named bundle against the hermetic offline stack. The app reached the local-auth path, but this machine’s ad-hoc Keychain returned `-25308` before the signed-in Memories UI could be driven; no production app or data was touched.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11140?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Resolved the P1 version-skew thread (Codex). The capability-gating fix was already on the PR head — verified end-to-end:

- Backend advertises `X-Omi-Memory-Default-Delete-Supported` on GET `/v3/memories` (`true` for canonical users only; `false` for legacy/error paths).
- Desktop defaults `defaultMemoryDeleteSupported` to `false`, sets it only when the header is present and `true`, and gates `bulkDeletionServerMutationAvailable` on it — so an updated desktop against an older backend (no `scope` param) keeps the button disabled rather than falling through to unscoped delete-all.

Verification: `test_v3_real_router_default_off_f4.py` (16 passed), `test_ws_i_write_convergence.py` (26 passed), `test_memories_batch_delete.py` + `test_ws_j_delete_privacy.py` (27 passed), Swift parse clean, all CI checks green.

kodjima33 (approved):
Restores the desktop default-memory delete regression with an explicit scope=default contract; Archive exclusion covered by INV-MEM-1 tests.

PR #11149, opened 2026-08-05
fix(desktop): route hermes/openclaw harness modes through AgentRuntimeRouting

Christiankbr:
## Problem

AgentControlService.currentHarnessMode() hardcoded the return to either "piMono" or "acp", silently discarding "hermes" and "openclaw" even when the user selected them in Settings > Advanced > AI Provider.

When chatBridgeMode was set to "hermes" (via the Settings UI), currentHarnessMode() returned "acp", causing the Node agent to register the generic ACP (Claude Code) adapter instead of the Hermes adapter, binding chat sessions to adapterId "acp" and failing with "Claude sign-in is required" errors.

This made the "Local Hermes Agent via ACP" option in Settings appear non-functional — the preference was written to the database correctly, but the chat engine ignored it.

## Root Cause

```swift
// AgentControlService.swift:74-76 (before)
static func currentHarnessMode() -> String {
    let mode = UserDefaults.standard.string(forKey: "chatBridgeMode") ?? "piMono"
    return mode == "piMono" ? "piMono" : "acp"  // hermes/openclaw discarded
}
```

The function only distinguished between piMono and acp, returning "acp" for everything else — including "hermes" and "openclaw", which are valid BridgeMode values (see ChatProvider.BridgeMode and AgentRuntimeRouting).

## Fix

Delegate to AgentRuntimeRouting.harnessMode(from:) which already knows all production harness modes (piMono, acp, hermes, openclaw) and maps them correctly. The routing helper now also accepts the persisted Claude Code spelling and maps it to ACP.

## Verification

1. Set chatBridgeMode = "hermes" in UserDefaults
2. Built and launched the desktop app
3. Confirmed HARNESS_MODE=hermes in the Node agent environment
4. Sent a chat message — got a response from Hermes (gpt-5.4-mini via ACP)
5. DB shows adapterId "hermes" in adapter_bindings, run.succeeded events
6. Added a regression assertion for the persisted "claudeCode" preference mapping to ACP

## Failure-Class

Failure-Class: none

- INV-CHAT-1

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11149?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Chat/AgentControlService.swift
+++ b/desktop/macos/Desktop/Sources/Chat/AgentControlService.swift
@@ -73,7 +73,11 @@ final class AgentControlService {
 
   static func currentHarnessMode() -> String {
     let mode = UserDefaults.standard.string(forKey: "chatBridgeMode") ?? "piMono"
-    return mode == "piMono" ? "piMono" : "acp"
+    // Delegate to AgentRuntimeRouting which knows all harness modes
+    // (piMono, acp, hermes, openclaw). Previously this hardcoded "acp" for
+    // anything that wasn't "piMono", which silently broke the Hermes and
+    // OpenClaw adapters even when selected in Settings.
+    return AgentRuntimeRouting.harnessMode(from: mode)?.rawValue ?? "piMono"
   }
 
   func logDetail(name: String, arguments: [String: Any]) -> String {


Git-on-my-level (changes requested):
Thanks for the investigation and clear write-up — the root cause you identified (hardcoding `"acp"` for any non-`piMono` mode, silently discarding `hermes`/`openclaw`) is real, and delegating to `AgentRuntimeRouting` is the right direction.

Unfortunately the current implementation introduces a regression for Claude/ACP users.

**The problem**

`chatBridgeMode` in `UserDefaults` stores the `ChatProvider.BridgeMode` **raw value**, not the adapter string. For Claude that raw value is `"claudeCode"`, not `"acp"`:

```swift
enum BridgeMode: String {
    case omiAI = "agentSDK"   // Legacy, auto-migrated to piMono
    case userClaude = "claudeCode"
    case piMono = "piMono"
    case hermes = "hermes"
    case openClaw = "openclaw"
}
```

But `AgentRuntimeRouting.harnessMode(from:)` only recognizes the adapter spellings (`"piMono"`, `"acp"`, `"hermes"`, `"openclaw"`, `"openClaw"`). Passing `"claudeCode"` returns `nil`, so the `?? "piMono"` fallback kicks in:

| chatBridgeMode stored | Before PR | After PR |
|---|---|---|
| `"piMono"` | `"piMono"` ✓ | `"piMono"` ✓ |
| `"claudeCode"` (Claude) | `"acp"` ✓ | **`"piMono"` ✗ regression** |
| `"hermes"` | `"acp"` ✗ (the bug you fixed) | `"hermes"` ✓ |
| `"openclaw"` | `"acp"` ✗ | `"openclaw"` ✓ |

So selecting Claude in Settings → Advanced → AI Provider would now silently route agent control tools to piMono instead of the ACP/Claude adapter — the same class of silent-misrouting bug this PR is fixing, just for a different provider.

The legacy `"agentSDK"` value is also affected until migration runs.

**Suggested fix**

Resolve through `BridgeMode` first (which maps `userClaude → acp`), the same way `ChatProvider.resolvedHarnessMode()` already does:

```swift
static func currentHarnessMode() -> String {
    let mode = UserDefaults.standard.string(forKey: "chatBridgeMode") ?? "piMono"
    return ChatProvider.harnessMode(for: ChatProvider.BridgeMode(rawValue: mode) ?? .piMono)
}
```

`ChatProvider.harnessMode(for:)` delegates to `AgentRuntimeRouting.harnessMode(for: BridgeMode)`, which correctly maps all four cases. This preserves your hermes/openclaw fix without breaking Claude.

**Test request**

A small regression test covering all four `BridgeMode` values through `currentHarnessMode()` (e.g. alongside `ChatBridgeProfilePreferenceTests`) would catch this class of bug going forward. `BridgeMode` cases and their expected harness outputs:

- `.piMono` → `"piMono"`
- `.userClaude` → `"acp"`
- `.hermes` → `"hermes"`
- `.openClaw` → `"openclaw"`

Happy to re-review once updated — the underlying fix is valuable.

---

Git-on-my-level:
Thanks @Christiankbr — clean, well-scoped fix.

Verified against `main`: `AgentControlService.currentHarnessMode()` did hardcode the return to `piMono` or `acp`, so `hermes`/`openClaw` values stored in `chatBridgeMode` were silently collapsed to `acp`. Delegating to `AgentRuntimeRouting.harnessMode(from:)` (which already maps all four production modes plus the `pi-mono`/`openClaw` aliases) is the correct fix and avoids re-encoding the mode table here.

One subtle note for awareness, not a blocker: unrecognized `chatBridgeMode` values now fall back to `piMono` instead of the old `acp`. That's the safer default (matches the `UserDefaults` fallback), but it is a behavior change for any non-canonical string.

Marking this as a positive signal from automation review. A human maintainer still needs to sign off before merge (this touches agent/provider routing), and it would be great to add a small unit test covering the four canonical modes plus the unknown-value fallback so this regression can't quietly return.

---

aryanorastar (commented):
Community review — agreeing with the existing CHANGES_REQUESTED on the Claude regression:

Root cause write-up is right (`piMono` vs everything-else → `acp` silently killed Hermes/OpenClaw). Delegating to `AgentRuntimeRouting` is the right *direction*, but as already called out, `chatBridgeMode` stores `BridgeMode` raw values (`claudeCode`, etc.), while `harnessMode(from:)` expects adapter spellings (`acp`). That turns Claude → `piMono` via `?? "piMono"` — same silent-misroute class.

Please route through `BridgeMode` / `ChatProvider.harnessMode(for:)` (as suggested), and add explicit tests for:
- `claudeCode` → `acp`
- `hermes` / `openclaw`
- unknown / legacy `agentSDK` fallback

Happy to re-review once that mapping is in — this is exactly the kind of small desktop-non-BLE correctness fix worth landing.

aryanorastar (changes requested):
Requesting changes (upgrade from earlier COMMENT — reinforce @Git-on-my-level):

Routing `hermes` / `openclaw` through `AgentRuntimeRouting` is the right fix for the hardcoded `"acp"` discard. **Do not merge until the Claude regression is fixed.**

**Blocker:** `chatBridgeMode` stores `ChatProvider.BridgeMode` **raw values**. Claude is `"claudeCode"`, not `"acp"`. Mapping only on adapter strings (or treating unknown as `acp`) breaks Claude/ACP users.

Please:
1. Translate `BridgeMode` raw → runtime adapter via the same path `AgentRuntimeRouting` already owns (include `claudeCode` → `acp`).
2. Add a focused unit/fixture that proves `claudeCode` still routes to ACP after the hermes/openclaw fix.
3. Keep hermes/openclaw coverage.

Happy to re-review once that regression test is green on tip.

PR #11127, opened 2026-08-05
fix(app): run bounded cloud sync grace when screen locks (#7221)

aryanorastar:
## Summary
When the screen locks, Flutter hits `paused` and `#9763`'s foreground-only `RecordingTransferCoordinator` stops all upload/reconcile work. That matches [#7221](https://github.com/BasedHardware/omi/issues/7221): overnight with the app "open" on fiber Wi-Fi still only drains ~5% of a 1000+ file backlog, and "processing on server" freezes at 0%.

## Implementation
- Pure `decideScreenLockedSyncContinuation` policy (`app/lib/utils/sync_continuation_policy.dart`).
- New `WakeTrigger.screenLockedGrace`: while backgrounded, runs **reconcile + bounded phone-disk uploads only** (max 3 batches / ~15 WALs). Skips BLE device discovery/drain.
- `SyncReconciler.onBackground` starts that grace when miss/uploaded work exists; iOS `beginBackgroundTask` via `com.omi/sync_background_task`.
- `WalSyncs.syncLocalUploadsOnly` + `LocalWalSyncImpl.maxBatches` for the bounded pass.

Out of scope: overnight BGProcessingTask for full 1000+ drain, BLE flash download in background, server job pipeline stalls.

## Tests
- `flutter test test/unit/sync_continuation_policy_test.dart`
- `flutter test test/services/wals/recording_transfer_coordinator_test.dart`

## Product invariants affected

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11127?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #7221, opened 2026-05-10
bug: Background sync stalls on iOS when screen is off — 1000+ files queue


galamarco:
**Describe the bug**
Background sync is extremely slow — almost completely stalled when the screen is off. After 5 days of recording, 1294 files accumulated in the sync queue. Leaving the app open overnight on fiber Wi-Fi resulted in less than 5% synced by morning.

**To Reproduce**
1. Record audio for several days without syncing (e.g. offline/delayed sync) or with iOS app in background
2. Open the app with a large queue of files (1000+)
3. Leave the app open overnight connected to Wi-Fi with screen off
4. Check sync progress in the morning

**Current behavior**
After a full night with the app open on Wi-Fi, only ~5% of 1294 queued files were synced. The process appears nearly stalled when the screen is off.

**Expected behavior**
The app should be able to process and upload a backlog of 1000+ files overnight without requiring the screen to stay on.

**Screenshots**
N/A

**user ID:**
marco.galassi.cri@gmail.com

**Smartphone + device:**
- Device: iPhone 15
- OS: 26.4.2
- App Version: 10535
- Device version: iphone 15

**Additional context**
The bottleneck is not the network — fiber Wi-Fi connection confirmed. Accumulation period: 5–10 May (5 days, 1294 files). iOS background processing limits are likely the root cause.

beastoin:
@marco-galassi Thanks for the detailed report with device info. This is a significant capture-layer issue — 1294 files queued with only 5% synced overnight on fiber Wi-Fi indicates iOS background processing limits are throttling sync aggressively.

I scored it as Layer capture / P1 / maintainer with score 25 (w=5, sev 4, trust 3, freq 3, lev 2, cost 3). [Triage guide](https://github.com/BasedHardware/omi/blob/main/docs/developer/triage.md).

---
_by AI for @beastoin_

hugoaap-code:
same on android.

galamarco:
## Urgent follow-up — service effectively unavailable for 2+ weeks

Hi team, posting an update because this issue has now crossed a threshold where the product is effectively non-functional for me.

### Current state (as of today)

- **Device storage full**: the Limitless Pendant is now blinking red (8x), indicating it can no longer buffer new recordings. From this point on, any unsynced audio is at risk of being permanently lost.
[...]

PR #11166, opened 2026-08-06
fix(app): soft-retry leave/background offline upload aborts (#4587)

### LINKED ISSUE #4587: Offline sync errors if I leave offline sync page
I start a offline sync, it syncs, starts processing. If I leave that screen it fails. Both navigating in app and another app. 

this issue happens since offline sync was released, it's not a new issue.

Omi AI 1.0.522 (686) — samsung SM-S938B — Android 16

latest firmware

aryanorastar:
## Summary
- Leaving the Offline Sync page (or backgrounding the app) mid-upload no longer paints a hard **Processing Failed** / `ClientSoftware caused connection abort` red error (#4587).
- Root cause: UI promises leave-safe sync (`syncProcessingBackgroundHint`), but architecture is foreground-only. OS aborts in-flight multipart; batches increment `localUploadFailures` and correctly leave WALs as `miss`, then `SyncProvider._performSync` fail-closed with `toError`.
- Shrink-only: on `localUploadFailures > 0` → `toIdle` + cooldown wake (matches schema: retryable, re-arm recovery — not completed, not terminal error). Transient network catch (incl. connection abort / ClientSoftware) same soft path. No FGS/Workmanager.
- Orthogonal to [#11127](https://github.com/BasedHardware/omi/pull/11127) (#7221 screen-lock grace).

## Test plan
- [x] `flutter test test/providers/sync_provider_sync_wal_wake_test.dart test/unit/transient_network_error_test.dart` → **7 passed**
- [ ] CI Dart Analyze / hermetic green
- [ ] Not verified on-device: Android leave sync page mid-upload with pendant backlog (needs device). Hermetic tests cover the fail-closed → soft-retry contract.

Failure-Class: none

Fixes #4587

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11166?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/backend/http/shared.dart
+++ b/app/lib/backend/http/shared.dart
@@ -33,25 +33,28 @@ class AuthTokenUnavailableException implements Exception {
 }
 
 // Normal-mode connectivity failures on mobile (no network, DNS failure,
-// connection reset, TLS handshake during reconnect, request timeout). Reporting
-// these to Crashlytics drowns out real signal — caller logs them locally and
-// either returns null or rethrows for the upstream sync state machine.
-bool _isTransientNetworkError(Object e) {
+// connection reset, TLS handshake during reconnect, request timeout, OS abort
+// when the app backgrounds mid-upload). Reporting these to Crashlytics drowns
+// out real signal — caller logs them locally and either returns null or
+// rethrows for the upstream sync state machine.
+bool isTransientNetworkError(Object e) {
   if (e is SocketException) return true;
   if (e is HandshakeException) return true;
   if (e is TimeoutException) return true;
-  if (e is http.ClientException) {
-    final m = e.message;
-    return m.contains('SocketException') ||
-        m.contains('HandshakeException') ||
-        m.contains('TimeoutException') ||
-        m.contains('Connection closed') ||
-        m.contains('Connection reset') ||
-        m.contains('Failed host lookup') ||
-        m.contains('Network is unreachable') ||
-        m.contains('Bad file descriptor');
-  }
-  return false;
+  final text = e is http.ClientException ? e.message : e.toString();
+  return text.contains('SocketException') ||
+      text.contains('HandshakeException') ||
+      text.contains('TimeoutException') ||
+      text.contains('Connection closed') ||
+      text.contains('Connection reset') ||
+      text.contains('Failed host lookup') ||
+      text.contains('Network is unreachable') ||
+      text.contains('Bad file descriptor') ||
+      // Android leaves/backgrounds mid-multipart (#4587 screenshot):
+      // "ClientSoftware caused connection abort"
+      text.contains('connection abort') ||
+      text.contains('ClientSoftware') ||
+      text.contains('Software caused connection abort');
 }
 
 Future<String> getAuthHeader({bool expireTerminalSession = true}) async {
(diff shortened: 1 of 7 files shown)


aryanorastar:
Addressed all four cubic findings on the new tip:

**P1 — no double cooldown wake:** local-failure / transient soft paths wake only when `!rethrowOnError` (direct/manual). Coordinator drains keep `rethrowOnError: true` and rely on `_runPass` → `_scheduleRetry`. `syncWal` post-success wake now requires zero failures so soft-retry does not emit a second wake.

**P2 — ClientSoftware token:** `isTransientNetworkError` matches only the full phrase `software caused connection abort` (case-insensitive). Bare `ClientSoftware` no longer counts as transient.

**P2 — permanent refusals:** batch catch now tracks `localUploadPermanentFailures` / `localUploadPermanentError`. Soft `toIdle` only when failures are transient; permanent 400/413 still `toError`.

**P3 — wake assertion:** regression test expects exactly `[cooldownElapsed]` and `WalStatus.miss`; added permanent-failure case expecting error + zero wakes.

`flutter test` on the touched suites → **9 passed**.

Git-on-my-level:
Thanks @aryanorastar — solid fix for #4587. The approach is well-scoped and the reasoning is sound:

- Correctly distinguishes transient leave/background multipart aborts from permanent server refusals (400/413) via the new `localUploadPermanentFailures`/`localUploadPermanentError` fields. Soft-retry (`toIdle` + cooldown wake) applies only to transient failures; permanent refusals still surface `SyncStatus.error`.
- The `software caused connection abort` matching (case-insensitive, full-phrase only) avoids over-classifying bare `ClientSoftware` tokens, which was the right tightening after cubic's review.
- No-double-wake guard (`!rethrowOnError`) correctly prevents the coordinator drain path from double-poking cooldown.
- Tests cover both the transient soft-retry contract (`miss` + exactly one cooldown wake, no error) and the permanent-failure error path, plus a dedicated unit suite for the network-error classifier.

A couple of things worth a human maintainer's eye before merge:

1. The `lower` variable in `isTransientNetworkError` is computed but only the abort-phrase branch uses it; the classic connectivity phrases still match against the original-cased `text`. That's intentional for the existing token list but worth confirming that none of those substrings depend on case (e.g. `http.ClientException` messages from the `http` package are typically already mixed-case). Not blocking.
2. This shifts `localUploadFailures` from fail-closed (`toError`) to fail-open-soft-retry. That's the intended fix, but it's a behavior change in the user-visible sync status surface — worth a quick on-device sanity check that a genuine persistent network failure still eventually surfaces (the PR notes on-device verification is still pending).

No security, privacy, or dependency concerns. Checks are green. Leaving a positive signal — requesting a maintainer's final product/UX judgment before merge since the on-device confirmation is still outstanding.

---

aryanorastar:
@mdmohsin7 all required checks are green on the current head (fa7750c58e), and @Git-on-my-level left a positive signal after the cubic follow-up hardening.\n\nThis is now only waiting on code-owner approval.\n\nCould you please take a quick review pass and approve if it looks good?

kodjima33 (approved):
Leave/background mid-upload now soft-retries instead of painting a terminal Processing Failed; 7 passing tests.

PR #11168, opened 2026-08-06
fix(app): tear down Android BLE GATT on disconnect tap even if unmanaged (#5361)

### LINKED ISSUE #5361: Device stays connected (blue LED) after tapping to disconnect — Pixel 6A / GrapheneOS (Android 16)
## Bug Report

**Device:** Omi
**Phone:** Pixel 6A
**OS:** GrapheneOS (Android 16)

### Steps to reproduce
1. Open Omi app — device auto-connects (blue LED)
2. Tap device button in app to disconnect
3. Device vibrates briefly but stays blue and keeps listening
4. App shows disconnected (red) but device remains connected

### Expected behavior
Device should disconnect and LED should turn red/off after tapping disconnect.

### Additional context
- Force-closing the app does disconnect, but reopening immediately reconnects without user action
- Audio capture works even while app shows disconnected (red status)
- May be related to GrapheneOS Bluetooth permission handling

_Reported via Discord by Mr. M_

aryanorastar:
## Summary
- Disconnect tap no longer leaves the pendant connected (blue LED / still listening) while the app shows red/disconnected (#5361).
- Root cause: `OmiBleForegroundService.unmanageDevice` early-returned when the address was missing from `managedDevices`, so a FGS/map desync skipped `disconnectGatt`/`closeGatt`. Dart still cleared prefs and flipped UI. Service-not-running fallback only called `closeGatt` (no `disconnect()`, no `user_disconnected` / `managed_device` clear).
- Shrink-only: always tear down GATT; set `user_disconnected`; remove `managed_device`; notify Flutter; stop service. `BleUnmanagePolicy` encodes the contract. Fallback path mirrors the same teardown.
- Orthogonal to [#10908](https://github.com/BasedHardware/omi/pull/10908) (#10847 — inverse desync / forceReconnect).

## Test plan
- [x] `BleUnmanagePolicyTest` (JVM hermetic) — managed-entry absent still plans full teardown
- [ ] CI Android Compile Smoke / Dart Analyze green
- [ ] Not verified on-device: Pixel / GrapheneOS disconnect tap → LED red + capture stops (needs device). Hermetic policy covers the fail-closed contract.

none

Failure-Class: none

Fixes #5361

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11168?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/android/app/src/main/kotlin/com/friend/ios/ble/BleHostApiImpl.kt
+++ b/app/android/app/src/main/kotlin/com/friend/ios/ble/BleHostApiImpl.kt
@@ -7,6 +7,7 @@ import com.friend.ios.BleHostApi
 
 import android.app.Activity
 import android.bluetooth.BluetoothAdapter
+import android.content.Context
 import android.content.Intent
 import android.util.Log
 import androidx.core.content.ContextCompat
@@ -53,8 +54,21 @@ class BleHostApiImpl(private val getActivity: () -> Activity?) : BleHostApi {
 
     override fun unmanageDevice(uuid: String) {
         Log.i(TAG, "unmanageDevice: $uuid")
-        OmiBleForegroundService.instance?.unmanageDevice(uuid)
-            ?: bleManager.closeGatt(uuid) // Fallback if service not running
+        val service = OmiBleForegroundService.instance
+        if (service != null) {
+            service.unmanageDevice(uuid)
+            return
+        }
+        // Service not running — still tear down any live GATT and record user
+        // disconnect intent so companion / sticky restart cannot revive (#5361).
+        bleManager.disconnectGatt(uuid)
+        bleManager.closeGatt(uuid)
+        val context = getActivity()?.applicationContext ?: return
+        context.getSharedPreferences(OmiBleForegroundService.PREFS_NAME, Context.MODE_PRIVATE)
+            .edit()
+            .putBoolean(OmiBleForegroundService.PREFS_USER_DISCONNECTED, true)
+            .remove(OmiBleForegroundService.PREFS_KEY)
+            .apply()
     }
 
     // ── Bonding ──
--- a/app/android/app/src/main/kotlin/com/friend/ios/ble/BleUnmanagePolicy.kt
+++ b/app/android/app/src/main/kotlin/com/friend/ios/ble/BleUnmanagePolicy.kt
@@ -1,0 +1,40 @@
+package com.friend.ios.ble
+
+/**
+ * Pure contract for user-initiated BLE disconnect (#5361).
+ *
+ * The foreground service's `managedDevices` map can desync from a live GATT
+ * (strict BG/BLE policy, GrapheneOS, FGS restart races). Unmanage must still
+ * tear down GATT and record user intent — never no-op just because the map
+ * entry is missing.
+ */
+object BleUnmanagePolicy {
+    data class Plan(
+        /** Cancel scheduled reconnect / stability timers if a managed entry existed. */
+        val cancelManagedTimers: Boolean,
+        /** Always call disconnectGatt then closeGatt for the address. */
+        val tearDownGatt: Boolean,
+        /** Persist user_disconnected=true so companion / sticky restart do not reconnect. */
+        val markUserDisconnected: Boolean,
+        /** Remove managed_device so sticky restore cannot revive the session. */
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Thanks @aryanorastar — solid, well-scoped fix for #5361. The fail-closed reasoning is right: the old `managedDevices.remove(addr) ?: return` could skip GATT teardown entirely on a map desync, and the service-absent fallback only did `closeGatt` with no `disconnect()` and no `user_disconnected`/`managed_device` clear. Having both paths always tear down GATT + persist user intent is the correct direction, and extracting `BleUnmanagePolicy` as a pure contract is a nice testable seam.

A few notes for a human maintainer / follow-up:

1. Reconnect/restore gating looks consistent. `manageDevice()` clears `user_disconnected` on an explicit re-manage, and `handleDisconnection` honors the flag before scheduling a retry — so a queued FGS start or `START_STICKY` restore won't silently revive a session the user just disconnected. Worth confirming on-device that no companion (`EVENT_BLE_APPEARED`) / `forceReconnect` path can race the teardown within the same window.

2. The JVM `BleUnmanagePolicyTest` covers the pure `plan()` contract for both managed-entry present/absent. It doesn't exercise the actual `unmanageDevice()` fail-closed path in `OmiBleForegroundService` / `BleHostApiImpl` (e.g. that `disconnectGatt`+`closeGatt` are actually invoked when the map entry is missing). A Robolectric/instrumentation test asserting the no-map-entry teardown would lock in the regression this fixes.

3. Minor doc nit: in `BleUnmanagePolicy.kt`, the `plan()` KDoc says "whether [address] was present in `managedDevices`", but `address` isn't a parameter of `plan()` (only `managedEntryPresent` is) — the `[address]` renders as a dangling link in generated docs.

Needs a human maintainer's eyes before merge (Android BLE/connectivity surface + on-device verification gap). Not blocking on anything structural — the change itself is coherent and the test plan is honest about what is and isn't covered.

---

aryanorastar:
Thanks @Git-on-my-level — agreed on the framing.

**Doc nit:** fixed the dangling `[address]` KDoc on `BleUnmanagePolicy.plan()` (cubic P3 / your note). Head: latest push on this branch.

**Cubic P1/P2 (queued FGS / sticky restore):** as you noted, the fail-closed contract already clears `managed_device` and sets `user_disconnected`; companion paths honor the flag, and `handleDisconnection` skips retry when it is set. Sticky restore only calls `manageDevice` when `managed_device` is still present, which unmanage removes. Leaving those as-is rather than expanding scope — on-device companion/`forceReconnect` race confirmation still needs a device (honest gap in the test plan).

**Service-absent diagnostics (cubic P2):** service path already emits `persistDisconnectEvent` + `onPeripheralDisconnected`; fallback remains GATT teardown + prefs only. Happy to mirror the emissions in a follow-up if you want parity there before merge; not expanding this PR unless asked.

**Robolectric no-map-entry teardown:** noted as a nice lock-in; JVM policy test stays the hermetic coverage for this PR.

@mdmohsin7 ready for code-owner review when you have a window — Android BLE disconnect-tap fail-closed (#5361).

kodjima33 (approved):
Always tears down GATT on disconnect tap so the pendant can't stay connected behind a red UI; policy encoded + JVM test.

mdmohsin7:
Were you able to recreate the issue locally @aryanorastar?

aryanorastar:
@mdmohsin7 Sorry — I missed your question before the merge/revert.

Honest answer: I did **not** recreate #5361 on a physical Android device. The change was driven from the code path (`managedDevices.remove(addr) ?: return` skipping teardown on map desync) + the pure `BleUnmanagePolicy` contract tests. David’s pass also flagged the on-device companion/`forceReconnect` race as an open gap.

The revert is fair given that. If you can point me at the **new bug** you saw from this tip (or the scenario where the partial fix is wrong), I’ll fix that specifically before any re-land — and I won’t re-open without device verification (or an explicit maintainer smoke) of disconnect-tap → GATT teardown + `user_disconnected` persistence with no sticky revive.

PR #11170, opened 2026-08-06
fix(app): strict segment seek + stop-at-end for transcript playback (#4471)

### LINKED ISSUE #4471: Segment playback starts at wrong
timestamp and keeps playing
Bug: Segment playback starts at wrong
timestamp and keeps playing forward
ios

◦ Omi (CV1

App/ Version: Omi Al 1.0.521 (659
Omi CV1 Firmware: 3.0.15

Device/OS: iPhone. iOS 26.2.1

Steps:
1. Open a conversation with timestamped segments.
2. Tap play on a specific segment (example: 00:11:41 to 00:11:43).

Expected: Plays only 00:11:41 to 00:11:43

Actual: Starts midway through a later segment
(example: 00:12:03 to 00:12:17) and continues playing forward to the end of the conversation.

Frequency: Always.

aryanorastar:
## Summary
- Tapping a transcript segment no longer seeks into a *later* span and plays to the end of the conversation (#4471).
- Root cause: segment taps used scrubber-oriented `wallToArtifact` **gap-snap**. After dense MP3 playback collapses gaps >90s, a segment whose wall start sits in that gap (e.g. 11:41) snaps to the next span (e.g. 12:03). macOS already uses strict in-span mapping (`CapturePlayback.artifactOffset`).
- Shrink-only: `wallToArtifactStrict` / `wallToArtifactStrictInclusive` for segment start/stop; require dense artifact + spans (no per-part fallback for taps); pause via `positionStream` at segment end. Scrubber keeps gap-snap.

## Test plan
- [x] `flutter test test/unit/audio_timeline_mapper_test.dart` → **14 passed** (incl. reporter-shaped 701→null vs scrubber snap)
- [ ] CI Dart Analyze / Android Compile green
- [ ] Not verified on-device: multi-part CV1 conversation with >90s gap; tap segment before/after gap (needs device). Hermetic tests cover the mapping contract.

none

Failure-Class: none

Fixes #4471

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11170?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/conversation_detail/page.dart
+++ b/app/lib/pages/conversation_detail/page.dart
@@ -333,8 +333,8 @@ class _ConversationDetailPageState extends State<ConversationDetailPage> with Ti
         final conversation = provider.conversation;
         final summaryContent =
             conversation.appResults.isNotEmpty && conversation.appResults[0].content.trim().isNotEmpty
-            ? conversation.appResults[0].content.trim()
-            : conversation.structured.toString();
+                ? conversation.appResults[0].content.trim()
+                : conversation.structured.toString();
         _copyContent(context, summaryContent);
         break;
       case 'download_audio':
@@ -764,8 +764,8 @@ class _ConversationDetailPageState extends State<ConversationDetailPage> with Ti
                                         provider.conversation.starred = newStarredState;
                                         // Update in conversation provider
                                         context.read<ConversationProvider>().updateConversationInSortedList(
-                                          provider.conversation,
-                                        );
+                                              provider.conversation,
+                                            );
                                         // Track star/unstar action
                                         PlatformManager.instance.analytics.conversationStarToggled(
                                           conversation: provider.conversation,
@@ -1104,15 +1104,13 @@ class _ConversationDetailPageState extends State<ConversationDetailPage> with Ti
                   child: Consumer<ConversationDetailProvider>(
                     builder: (context, provider, child) {
                       final conversation = provider.conversation;
-                      final hasActionItems = conversation.structured.actionItems
-                          .where((item) => !item.deleted)
-                          .isNotEmpty;
+                      final hasActionItems =
+                          conversation.structured.actionItems.where((item) => !item.deleted).isNotEmpty;
                       return ConversationBottomBar(
                         mode: ConversationBottomBarMode.detail,
                         selectedTab: selectedTab,
                         conversation: conversation,
-                        hasSegments:
-                            conversation.transcriptSegments.isNotEmpty ||
+                        hasSegments: conversation.transcriptSegments.isNotEmpty ||
                             conversation.photos.isNotEmpty ||
                             conversation.externalIntegration != null,
                         hasActionItems: hasActionItems,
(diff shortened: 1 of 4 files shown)


aryanorastar:
Addressed both cubic P2s:

**Formatting:** `dart format` restored the six-space initializer indent on `AudioTimelineMapper`.

**Rapid second tap:** `_segmentSeekGeneration` owns the segment end-handler. Scrubber seeks bump it (invalidate). Segment taps bump after seek and only pause/clear when the callback still owns the generation — a late pause from the previous tap cannot stop the new playback.

`flutter test test/unit/audio_timeline_mapper_test.dart` → **14 passed**.

aryanorastar:
**Formatting** was red on `app/lib/pages/conversation_detail/page.dart` (not the mapper/bottom_bar files). Ran `dart format --line-length 120` on that file and pushed.

aryanorastar:
**Android Compile Smoke** — infra flake, not this diff.

`Could not resolve androidx.test:runner:1.2+` → GET `storage.googleapis.com/download.flutter.io/.../maven-metadata.xml` returned **403 Forbidden**. No Android/Gradle changes here (Dart mapper + bottom_bar + page format only). Neighboring Mobile App Checks on other branches are green.

Couldn’t `gh run rerun` (needs admin). Pushed an empty commit to retrigger.

Git-on-my-level:
Nice work on this — the strict-mapping approach is clean and the generation-guarded stop handler is the right pattern for preventing stale end-handlers from pausing newer taps. The reporter-shaped test case is a good touch.

Two behavior changes I'd want a maintainer to confirm before merge:

1. **Stop-at-end is new UX.** Tapping a transcript segment now auto-pauses at the segment boundary instead of continuing through the rest of the conversation. That's a deliberate product direction (matches the PR title), but it changes what users expect from segment playback, so it's worth a product judgment call.

2. **Per-part fallback removed.** Segment taps now require `_singleArtifact && _timelineMapper != null`; conversations without the dense artifact + spans manifest get an `audioPlaybackUnavailable` error instead of the old imprecise `_calculateFilePositionForTimestamp` seek. Reasonable for correctness, but it's a regression for those edge-case conversations — worth confirming that's acceptable.

Minor edge case: if `wallToArtifactStrictInclusive(segmentEnd)` returns null (unmappable end), `stopSeconds` falls back to `filePosition`, so the positionStream listener could fire on the first position event and pause immediately. Probably only affects malformed segments, but worth a guard or a comment.

CI is green and the mapper contract is well-tested. Leaving for maintainer review on the two UX/behavior questions above.

---

kodjima33 (approved):
Strict in-span segment mapping matches the macOS behaviour and stops seeking past the gap; 14 unit tests.

PR #11172, opened 2026-08-06
fix(chat): add authoritative current-time context

Git-on-my-level:
## Summary

Fix cross-platform chat date context so Windows and the mobile Agent VM path use an authoritative current time, matching the behavior already present on the normal Mac/mobile backend path.

## Root cause

- Windows kernel-routed typed chat built a stable system prompt but never added a per-turn current-time block.
- Mobile Claude-agent queries bypass the normal chat backend's datetime injection; the Agent VM proxy forwarded raw prompts and did not receive the device timezone.
- Typed and multipart streaming requests handled error responses as unmaterialized streams, so clock-skew responses could not reach the existing detector.

## Changes

- Add a per-turn `# Current Time` block to Windows main-chat dispatch, using the host clock, local IANA timezone, and an explicit ISO offset while keeping the system prompt byte-stable.
- Send the Android/iOS device timezone on the Agent VM WebSocket and per query, so long-lived sessions follow travel and explicitly fall back to UTC when a lookup is unavailable. The proxy prefixes every valid VM query with server-clock time rendered in that timezone.
- Preserve the VM's protocol-level `Invalid query` behavior for malformed prompts and keep current-time prefixing active when history seeding fails.
- Materialize non-200 typed/multipart streaming responses before invoking clock-skew detection, preserving the existing 402 error payload behavior even when a response body resets mid-stream.
- Add deterministic Windows, Flutter, proxy, and loopback streaming regression coverage.

- INV-CHAT-1

## Verification

- Windows focused tests: 39 passed.
- Windows full suite: 536 files passed, 6 skipped; 5,157 tests passed, 24 skipped.
- Windows ESLint and Node typecheck passed.
- Flutter focused tests: 11 passed.
- Flutter full suite: 1,033 tests passed.
- Flutter analyzer ratchet and Dart formatting passed.
- Backend focused proxy tests: 26 passed.
- Backend prescribed reproduction of the five files that tripped the parallel CPU-time guard passed all five files, with warnings only.
- Full backend suite executed; all reported functional tests passed, but five unrelated fast-unit CPU-time guard checks exceeded the tight 0.12s local threshold under parallel load. The exact failed-file reproduction passed with the same `test.sh` timing guard.
- `make setup`, `make preflight`, and `make runtime-image-source-closure` passed.

## Failure class (fixes)

Failure-Class: none

## Scope notes

This PR does not reintroduce the removed clock-skew snackbar, which had produced false-positive user warnings from stale request timestamps. The detector remains available for logging/telemetry. Physical Windows, Android, and iOS devices were not available in this macOS validation environment.

### DIFF
--- a/app/lib/backend/http/shared.dart
+++ b/app/lib/backend/http/shared.dart
@@ -573,10 +573,14 @@ Stream<String> makeStreamingApiCall({
     }
 
     if (streamedResponse.statusCode != 200) {
-      Logger.error('Streaming request failed: ${streamedResponse.statusCode}');
-      if (streamedResponse.statusCode == 402) {
+      // Materialize error responses so clock-skew detection sees the JSON body;
+      // streamed responses previously bypassed _checkClockSkewResponse().
+      final errorResponse = await http.Response.fromStream(streamedResponse);
+      _checkClockSkewResponse(errorResponse);
+      Logger.error('Streaming request failed: ${errorResponse.statusCode}');
+      if (errorResponse.statusCode == 402) {
         try {
-          var body = await streamedResponse.stream.bytesToString();
+          final body = errorResponse.body;
           yield 'error:402:$body';
         } catch (_) {
           yield 'error:402:{}';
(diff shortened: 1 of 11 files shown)


Git-on-my-level:
Reviewed both remaining P3 threads from the Cubic review run.

**Thread on `test_agent_proxy_async_boundaries.py` (invalid timezone fallback)** — already covered. The fallback-to-UTC path Cubic suggests testing exists as a dedicated unit test: `test_current_time_prompt_falls_back_to_utc_for_invalid_mobile_timezone` in `test_agent_proxy_history_seeding.py:54`, which passes `"Not/AZone"` and asserts UTC output. The integration test in `test_agent_proxy_async_boundaries.py` exercises the end-to-end forwarding path; the unit test covers the formatter's fail-closed branch directly.

**Thread on `desktopChatPrompt.ts` (duplicated timezone formatting)** — out of scope for this PR. The two `localIsoParts` implementations live in different Electron process contexts (main `src/main/agentKernel/` vs renderer `src/renderer/src/lib/voice/`) and use different strategies (`Intl.DateTimeFormat` with `timeZoneName: 'longOffset'` vs manual `offsetMinutes` wall-clock reconstruction). Consolidating into a shared utility is a reasonable follow-up refactor, but it belongs in its own PR rather than this bug-fix change.

kodjima33 (approved):
Correct per-turn current-time injection across Windows and the Agent VM path, with deterministic tests. Approving; holding the merge since it adds behaviour across three clients rather than repairing a single defect.

PR #11173, opened 2026-08-06
fix(listen): keep web + device recordings on separate conversations (#5388)

### LINKED ISSUE #5388: Bug: Simultaneous device + web app recording fails — both are lost
### Problem

When the Omi pendant is actively recording and the user starts a second recording via the web app (e.g. for an online meeting using headphones), the web app shows it's recording but the meeting audio is never captured. The web app recording silently fails while the device recording continues.

### Steps to Reproduce

1. Wear the Omi pendant — it's connected and recording normally
2. Join an online meeting using headphones (audio goes through headphones, not picked up by pendant)
3. Start recording on the **web app** (h.omi.me) to capture the meeting audio
4. Web app shows recording indicator — appears to be working
5. After the meeting, the web app recording is missing — no transcription, no conversation

### Expected Behavior

Device recording and web app recording should work **independently and simultaneously**. They are capturing different audio sources (pendant mic vs. browser audio) and should create separate conversations.

### Actual Behavior

The web app recording silently fails when the device is also recording. The user sees a recording indicator on the web app but nothing is actually captured. Only the pendant recording works.

### Impact

High — users who rely on the pendant for in-person audio and the web app for online meetings lose their meeting recordings. There's no warning that the web app recording isn't working, so users only discover the loss after the meeting is over.

### Use Case

This is a common scenario: wearing the pendant all day for in-person conversations, but switching to the web app for online meetings where audio comes through headphones. Both sources should be captured independently.

---

Reported by community member on Discord.

aryanorastar:
## Summary
- Simultaneous pendant + web (`h.omi.me`) recording no longer silently loses the web meeting (#5388 / related #5389).
- Root cause: `prepare()` attached new listen sockets to the single Redis `in_progress` conversation. Web audio could write into a device session, and stop/finalize could complete the shared pointer.
- Shrink-only changes:
  - Backend: refuse cross-source attach (`should_attach_to_existing_in_progress`) so `omi` and `unknown` sources do not share one in-progress conversation.
  - Web: pass `client_conversation_id`, keep `conversation_session` only when status is `in_progress`, and finalize by explicit ID via `/v1/conversations/{id}/finalize` instead of the global pointer.
  - Compatibility: `ConversationSource.web` is not present in the public app-client contract; `source=web` still maps through `_missing_` to `unknown`.

## Test plan
- [x] `pytest tests/unit/test_transcribe_decisions.py` (cross-source attach refusal coverage)
- [x] CI lanes: backend/web/OpenAPI green
- [ ] Manual: pendant recording + start web meeting record → two conversations after stop

## Follow-up (separate from this shrink fix)
- Finalize-vs-drain race on immediate web socket close (tail-end audio timing)

Failure-Class: none

Fixes #5388

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11173?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/models/conversation_enums.py
+++ b/backend/models/conversation_enums.py
@@ -50,6 +50,7 @@ class ConversationSource(str, Enum):
     phone = 'phone'
     phone_call = 'phone_call'
     desktop = 'desktop'
+    web = 'web'
     openglass = 'openglass'
     screenpipe = 'screenpipe'
     workflow = 'workflow'
--- a/backend/models/conversation_enums.py
+++ b/backend/models/conversation_enums.py
@@ -50,7 +50,6 @@ class ConversationSource(str, Enum):
     phone = 'phone'
     phone_call = 'phone_call'
     desktop = 'desktop'
-    web = 'web'
     openglass = 'openglass'
     screenpipe = 'screenpipe'
     workflow = 'workflow'
--- a/backend/routers/listen/conversations.py
+++ b/backend/routers/listen/conversations.py
@@ -24,6 +24,7 @@ from utils.transcribe_decisions import (
     decide_recording_session_reconnect_action,
     recording_session_id_for_lifecycle_event,
     select_recording_session_id,
+    should_attach_to_existing_in_progress,
 )
 from utils.transcribe_store import calendar_db, conversations_db, redis_db
 
@@ -260,6 +261,16 @@ class LiveConversationController:
             return None
         existing = await self.host.persistence.call(retrieve_in_progress_conversation, self.host.request.uid)
         if not existing:
+            await self.create_new_in_progress_conversation()
+            return None
+        existing_source = existing.get('source')
+        if hasattr(existing_source, 'value'):
+            existing_source = existing_source.value
+        # Cross-source sockets (pendant + web meeting) must not share one conversation (#5388).
+        if not should_attach_to_existing_in_progress(
+            existing_source=existing_source if isinstance(existing_source, str) else None,
+            request_source=self.host.request.source,
+        ):
             await self.create_new_in_progress_conversation()
             return None
         finished_at = datetime.fromisoformat(existing['finished_at'].isoformat())
--- a/backend/tests/unit/test_transcribe_decisions.py
+++ b/backend/tests/unit/test_transcribe_decisions.py
@@ -224,8 +224,10 @@ def test_cross_source_in_progress_must_not_attach():
     assert should_attach_to_existing_in_progress(existing_source='omi', request_source='web') is False
     assert should_attach_to_existing_in_progress(existing_source='web', request_source='desktop') is False
     assert should_attach_to_existing_in_progress(existing_source=None, request_source='') is True
-    assert normalize_listen_source('Web') == 'web'
+    assert normalize_listen_source('Web') == 'unknown'
     assert normalize_listen_source(None) == 'omi'
+    assert normalize_listen_source('web') == 'web'
+    assert should_attach_to_existing_in_progress(existing_source='unknown', request_source='Web') is True
 
 
 def test_recording_session_identity_retries_and_rollovers_are_distinct():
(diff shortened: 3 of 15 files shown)


aryanorastar:
OpenAPI contract failure was from adding `ConversationSource.web` without refreshing the committed public schema. Regenerated `docs/api-reference/openapi.json` (`web` enum member) — should clear **Public Developer API contract**.

aryanorastar:
Follow-up pushed for Cubic P2s + OpenAPI lane:

1. **OpenAPI** — refreshed `public` / `app-client` / `integration-public` + generated TS/Swift clients (`web` enum).
2. **Source canonicalize** — `normalize_listen_source` now uses `ConversationSource` (same as create persistence), so unknown/mixed-case tokens attach consistently instead of spawning a new convo every reconnect.
3. **conversation_session** — only accept `status === 'in_progress'` before updating the finalize id.
4. **Refs** — single `conversationIdRef`.

Deferred (pre-existing / out of shrink): finalize-vs-drain race on socket close; ultra-short stop before prepare completes (same class as prior `processInProgressConversation` path).

aryanorastar:
**OpenAPI still red** was not stale schemas — it was `check_app_client_openapi_compatibility.py` rejecting `ConversationSource.web` as **14 breaking response-union changes** (released clients cannot decode a new enum member).

**Fix:** removed `ConversationSource.web`. Query `source=web` still maps to **`unknown`** via `_missing_` (same as before this PR). Cross-source attach continues to refuse `omi` ↔ `unknown`, so pendant + web stay separate without expanding the released app-client contract.

Local: `check_app_client_openapi_compatibility.py --base-ref origin/main` → **passed**.

Git-on-my-level (commented):
Thanks for this well-targeted fix for #5388 — the cross-source attach refusal in `prepare()` and the `client_conversation_id` + finalize-by-ID pattern are the right approach for keeping pendant and web recordings on separate conversations.

A few observations for the maintainer:

1. **PR description is now stale.** It still says "add `ConversationSource.web`," but that member was removed for OpenAPI compatibility and `source=web` now maps to `unknown` via `_missing_`. The behavior is correct (`unknown != omi`, so cross-source attach still refuses), but updating the description would help reviewers.

2. **Finalize-vs-drain race (pre-existing, acknowledged).** The backend only auto-finalizes `desktop`-source conversations on clean disconnect (runtime.py `close_code == 1000` path). Web recordings rely on the frontend `/finalize` call, which fires immediately after `socket.disconnect()` and may race with trailing buffered-audio segments. This is strictly better than before (correct conversation vs wrong conversation), but worth tracking as a follow-up so short tail-end audio isn't dropped.

3. **Unknown-source bucketing.** Since `source=web` maps to `unknown`, any other unrecognized source would share that bucket with web. Not a blocker for #5388 (pendant is `omi`), just noting the coarseness is intentional to avoid an OpenAPI breaking change.

Unit tests for the cross-source decision logic are solid. Requesting human maintainer review since this touches conversation routing and finalization (user-data persistence).

---

aryanorastar:
@Git-on-my-level PR description is now updated to match the landed diff (no ConversationSource.web contract expansion; source=web stays mapped to unknown).\n\nAll scoped checks are green, and this is ready for human maintainer approval/merge when you’re comfortable.\n\nFollow-up race (finalize vs tail-drain) is tracked as separate out-of-scope hardening.

aryanorastar:
@Git-on-my-level All CI checks are green on the current head of #11173 (only the human approval gate remains).\n\nAlso, I updated the PR description to remove the stale wording about adding ConversationSource.web (the merged behavior keeps source=web mapping via _missing_ to unknown, and cross-source attach still refuses omi vs unknown).\n\nWhen you’re comfortable, could you please proceed with approval to clear the remaining needs-maintainer-review gate? The finalize-vs-drain timing item is still tracked as an explicit follow-up/out-of-scope, per the PR.

kodjima33 (approved):
Refuses cross-source in-progress attach so web and device recordings no longer collide; unit test covers the refusal.

PR #11176, opened 2026-08-06
feat: reconcile dev Agent VM boot-image migrations

Git-on-my-level:
## Summary

- add an explicit development-only Agent VM boot-image migration plan, rendered and activated through a canonical, generation-guarded manifest path
- reconcile stopped allowlisted VMs through a journaled candidate, private authenticated readiness check, fenced pointer cutover, retained soak, and ID/label-fenced predecessor retirement
- preserve safe retry and cleanup behavior: durable candidate tokens, legacy instance-ID binding, candidate cleanup on terminal quarantine, and Firestore retirement claims before provider deletion
- extend the reconciler IAM role only with the GCE create/image/label/delete permissions needed for the fenced dev workflow

## Safety and rollout

Ordinary release manifests cannot enable replacement. The runtime additionally rejects migration outside `based-hardware-dev`, requires an explicit UID allowlist and one-at-a-time concurrency, and does not stop an active VM. Production replacement remains hard-disabled.

## Validation

- `PYTHONPATH=. .venv/bin/python -m pytest tests/unit/test_agent_vm_reconciler_job.py tests/unit/test_agent_vm_reconciler.py tests/unit/test_agent_vm_reconciler_deployment_scripts.py` (64 passed)
- `bash backend/scripts/typecheck.sh` (0 errors)
- `make preflight` (24 selected checks passed)
- Luna swarm review, including follow-up review of retry, retirement, scheduling, activation, and cleanup fences

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11176?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/backend-other.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/backend-other.json
@@ -1,8 +1,10 @@
 {
   "files": {
+    "backend/services/agent_vm_lifecycle.py": 1682,
     "backend/testing/listen_pusher_stack/run.py": 1790
   },
   "raise_justifications": {
+    "backend/services/agent_vm_lifecycle.py": "#11164's lifecycle owner now keeps the migration journal's account-deletion, lease, pointer, GCE identity, and missing-candidate retry fences alongside the existing Agent VM transitions; splitting those transaction primitives would duplicate the safety boundary.",
     "backend/testing/listen_pusher_stack/run.py": "#10577 keeps the operation-scoped persistence-fault controls beside the real listener/Pusher/Firestore assertions they qualify. +79 for #9809: the cloud-orphan-recovery scenario stays in the existing end-to-end listener/Pusher gauntlet rather than forking a second harness."
   },
   "threshold": 1500
--- a/backend/jobs/agent_vm_reconciler.py
+++ b/backend/jobs/agent_vm_reconciler.py
@@ -17,6 +17,7 @@ import json
 import logging
 import os
 import re
+import secrets
 import time
 import uuid
 from dataclasses import dataclass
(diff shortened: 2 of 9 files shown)


Git-on-my-level:
**Thread `W4ahL` — shared IAM role carries migration permissions in prod (Cubic, conf 6)**

Acknowledged, intentionally not split. The PR's security model is runtime hard-disable, not IAM least-privilege per environment:

- The migration path is triple-gated at runtime: `release.environment != "development"` AND `AGENT_VM_ENVIRONMENT != "development"` AND `GCE_PROJECT_ID != "based-hardware-dev"`. Any one failing aborts before any create/delete call.
- The installer (`apply-agent-vm-reconciler-iam.sh`) is a shared, refuse-by-default script run once per environment with explicit env vars. Splitting into a dev-only migration role would require conditional installer branches, a second role definition, runbook changes, and test coverage for two IAM shapes — all outside this PR's scope.
- The pre-existing prod reconciler already held `compute.instances.start/stop` (ability to halt production VMs). The new permissions are additive to a role that was never least-privilege in the strict sense; the defense boundary is the code-level hard-disable, documented in the runbook and enforced by unit tests.

If defense-in-depth via environment-scoped IAM is desired, it should be a dedicated hardening follow-up (separate role + conditional installer), not bolted onto this feature PR.

---

**This run:** pushed `094a314891` — black-formatted `test_agent_vm_reconciler_job.py` and added `Failure-Class: none` to the PR body. Both `Hygiene` and `Formatting` checks now pass (all 26 CI checks green). Resolved threads `W4ahH` (UID whitespace) and `W4ahS` (deployment-script test) as code-addressed.

PR #11177, opened 2026-08-06
fix(app): localize the plans sheet

mdmohsin7:
## Summary

The plans sheet rendered hardcoded English on top of a localized billing toggle. On a Japanese device the user saw 年額 / 約25%お得 / 月額 and プロモコード wrapped around "Choose your plan to unlock unlimited Omi.", "Doesn't work on Desktop" and "3 Months Free".

This fixes everything on that screen that the app owns. The plan-card interiors stay English for now — they arrive pre-rendered from the backend (see **Not in scope**).

## What was actually wrong

Three highlight bullets **already had ARB keys, translated into all 49 locales**, and the widget passed English literals straight past them:

```dart
_buildFeatureItem(faIcon: FontAwesomeIcons.infinity, text: 'Unlimited conversations'),
```

`unlimitedConversations`, `askOmiAnything` and `unlockOmiInfiniteMemory` were sitting right there — `ja` has 無制限の会話 — two lines below a working `context.l10n.youAreOnAPaidPlan`. Same story for `selectedPlanNotAvailable`, `upgradeAlreadyScheduled`, `annualPlanStartsAutomatically`, `planRenewsOn`, `youreOnAnnualPlan` and `alreadyBestValuePlan`. Those nine are wired up here at zero translation cost.

## Changes

**Commit 1 — plans sheet.** Wires the nine existing keys and adds 16 new ones, translated across all 48 non-template locales: the header, the platform bullet, `POPULAR`, both desktop-access rows, the annual billing summary, the months-free badge, the "Ends on" badge, the resubscribe/cancel notices, the freemium limits intro and delay row, the downgrade action, and the three training-data card strings.

`annualSaveTag()` built `"N Months Free"` in Dart, so it could never be translated. It becomes `annualMonthsFree()` returning the count, and the widget renders it through an ICU plural — which also gets the singular right per locale rather than by string concatenation.

**Commit 2 — unrelated, same files.** `pinAsBaseline`, `unpinAsBaseline`, `baselineMemory` and `alwaysInContext` were added to the template only, so all 47 other locales fell back to English and `gen-l10n` reported **188 untranslated messages**. That standing noise is part of why the plans-sheet gap went unnoticed. Now translated; `gen-l10n` reports zero.

## Deliberately unchanged

- **`eyebrow == 'Most popular'`** at the POPULAR badge call site (`plans_sheet.dart:1675`). It string-matches a backend value, not display copy — localizing it would silently drop the badge. It should become a boolean on the API before the backend work starts.
- **Card interiors** — plan titles, subtitles, feature rows, `$13/mo`. These come from `/v1/payments/available-plans` as pre-rendered English (`get_paid_plan_definitions()` / `get_plan_features()` f-strings) and are painted verbatim. The backend has no i18n layer; that is the follow-up.
- **The hardcoded `$` on prices.** Stripe is already billing some of these customers in JPY/KRW/GBP (`presentment_details.currency` on live subscriptions), so the displayed price can differ from the amount charged. That is a currency bug, not a translation bug, and wants `NumberFormat.currency` plus the presentment currency on the wire.

## Note on the generated files

`lib/l10n/app_localizations*.dart` are committed and stored at 120 columns; `flutter gen-l10n` emits 80. They are re-run through `dart format --line-length 120` so the diff stays additive instead of rewriting every line of all 49 files.

## Verification

```
flutter gen-l10n --untranslated-messages-file
  -> {}   (zero untranslated across all 49 locales; was 188)

flutter analyze lib/l10n/ lib/pages/settings/widgets/plans_sheet.dart \
  lib/utils/plan_pricing.dart test/unit/plans_sheet_l10n_test.dart
  -> 1 info (prefer_final_fields, plans_sheet.dart:52) — confirmed pre-existing on main, 0 new

./test.sh
  -> 1082 tests, all passed

make preflight
  -> all manifest checks passed
```

New coverage in `test/unit/plans_sheet_l10n_test.dart` (58 tests): every locale carries all 25 plans-sheet keys and none is blank; `{date}` / `{months}` / `{price}` / `{count}` placeholders survive every translation; the generated lookups return locale-specific copy (`ja.popularBadge` is asserted *not* to equal `POPULAR`, so a silent English fallback fails); ICU plural and interpolation resolve. `test/utils/plan_pricing_test.dart` moves from asserting the English label to asserting the count — the label assertion did not disappear, it moved to the localized-rendering test.

**Not verified on a device.** This is string-level and covered by the locale tests; I did not build and run the app to eyeball the Japanese paywall. Worth a look on a real device before it ships if you want the layout checked — some of these translations are longer than the English and the badges are narrow.

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11177?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mdmohsin7:
Both cubic findings were valid. Fixed in 99b0cd0ade.

**`endsOnDate` duplicate (P1) — my defect.** The key already existed in all 49 ARBs as `"Ends {date}"`, fully translated. I checked for collisions by comparing English *values* before adding keys, so a same-name/different-value key slipped straight through, and appending a second definition made the key duplicate in every file. `jsonDecode` keeps the last, so the pre-existing translation went dead in 49 locales and every check I had still passed — the value lookups only assert the key is *present*, and it was, twice.

Correction to the report: no Dart code called `context.l10n.endsOnDate` on main, so nothing was rendering the old string; the pre-existing key was orphaned. The duplication and the silently-shadowed translations are real regardless. Rather than renaming to `planEndsOnDate`, the sheet now uses the key that was already there — which is the same thing this PR is about. Visible effect in English: "Ends on Aug 6" → "Ends Aug 6", shorter and better for a 9pt badge. Most locales were already identical (`ja` 「{date}に終了」 in both).

**Incomplete plurals (P2/P3) — valid.** `=1/other` rendered the 2–4 noun form for every larger count. The count is derived from live Stripe prices, so 5+ is reachable if a tier's annual discount deepens. Full CLDR categories now declared for the 14 languages that have them: ar (zero/one/two/few/many/other), be, bs, cs, hr, lt, lv, pl, ro, ru, sk, sl (one/two/few/other), sr, uk. Slovenian's `two` and Arabic's dual are included.

One correction there too: Croatian has one/few/other, not one/few/many.

**Added the guard.** `no ARB defines the same key twice` reads the raw file text rather than the parsed map, because parsing is exactly what hid this. Verified by planting a duplicate `popularBadge` in `app_ja.arb` and confirming it fails before keeping it.

```
flutter gen-l10n --untranslated-messages-file -> {}
./test.sh -> 1084 tests, all passed
make preflight -> 11 checks passed
```

mdmohsin7:
Second pass — three of these were valid and are fixed in 8df06bddef. The `endsOnDate` items are stale; they were fixed in 99b0cd0ade before this review ran (one `endsOnDate` per ARB now, one getter per generated file).

**`planCancelsResubscribe` vs `planSetToCancelOn` — valid, and worse than reported.** Both that pair *and* `planEndedResubscribe` vs `planEndedOn` were character-identical duplicates. Same defect as `endsOnDate`, but invisible to the duplicate-key guard because the names differ. Both new keys dropped; the sheet uses the pre-existing ones, already translated in all 49 locales.

**Portuguese variant — valid.** `app_pt.arb` is Brazilian ("Você está em um plano pago", "Seu plano é renovado em") — 315 strings carry pt-BR markers against 93 pt-PT. I wrote the new copy in European Portuguese: telemóvel, O seu plano, Subscreva, faturação, Partilhar. Rewritten to pt-BR.

**`trainingDataProgram` missing from the test — valid.** Added, along with `upgradeScheduled`, `planEndedOn` and `planSetToCancelOn`, which the sheet also reuses without asserting.

**New guard:** `no plans-sheet key repeats another key's text`, comparing values rather than names — the thing the duplicate-key check structurally cannot see. Scoped to the plans-sheet keys, because the template already carries 220 duplicate values elsewhere and a repo-wide version would be noise.

It earned its keep immediately: it flagged a pre-existing pair, `renewsOn` and `planRenewsOn`, both "Your plan renews on {date}." with no caller on either. Removed the dead one.

Net effect of this commit is 42 insertions against 1,079 deletions — three keys' worth of duplicate copy across 49 locales, gone.

```
flutter gen-l10n --untranslated-messages-file -> {}
flutter test (plans_sheet_l10n + plan_pricing) -> 69 passed
./test.sh -> 1085 tests, all passed
make preflight -> 11 checks passed
```

PR #11194, opened 2026-08-07
fix(app): restore Omi double-tap mute after Android reconnect

undivisible:
## What changed

- Keep Omi BLE notification streams alive across transient Android reconnects and re-subscribe the native characteristics without orphaning the existing Dart listeners.
- Label the double-tap action as `Mute / Unmute` in device settings, matching the onboarding behavior.
- Add hermetic coverage for double-tap characteristic delivery before and after reconnect, plus the settings copy.

## Root cause

`NativeBleTransport` closed the Dart notification stream during a reconnect and created a replacement stream. The existing `OmiDeviceConnection` button subscription remained attached to the closed stream, so firmware double-tap notifications stopped reaching `CaptureController` after the Android BLE lifecycle changed.

## Validation

- `cd app && bash test.sh` — 1,112 tests passed.
- `app/scripts/analyze_ratchet.sh` — passed.
- Focused transport/settings tests — passed.
- `make preflight` — passed.

## Scope boundary

Firmware 3.0.20 has no dedicated BLE mute command. Its mic-gain level `0` only attenuates samples while the capture thread continues, so this PR does not claim to add true device-side microphone mute. The existing app-side recording pause remains unchanged; a true hardware mute needs a coordinated firmware protocol and firmware rollout.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11194?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches core BLE transport reconnect/subscription logic used for device button events; behavior change is localized but affects capture after GATT drops.
> 
> **Overview**
> Fixes **Omi double-tap (mute/unmute) stopping after transient Android BLE reconnects** by changing how `NativeBleTransport` handles disconnect and re-subscription.
> 
> On disconnect, the transport **no longer tears down Dart notification streams** that upper layers still listen to; it tracks active vs. native-subscribed characteristics, avoids wiping subscription state on duplicate disconnect callbacks, and **re-subscribes on reconnect** (including after first `deviceReady`) while reusing or recreating controllers only when needed. `disconnect`/`dispose` now unsubscribe via the active-key set.
> 
> **Device settings** copy for double-tap action 1 switches from pause/resume strings to **`Mute / Unmute`** (`deviceOnboardingMuteUnmute`), aligned with onboarding.
> 
> Adds **widget and transport tests** for the new labels and for characteristic delivery before/after reconnect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87dab07596b2c9da722cb298cf7c10a9820db1c9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/lib/pages/settings/device_settings.dart
+++ b/app/lib/pages/settings/device_settings.dart
@@ -375,7 +375,7 @@ class _DeviceSettingsState extends State<DeviceSettings> {
       case 0:
         return context.l10n.endConversation;
       case 1:
-        return context.l10n.pauseResume;
+        return context.l10n.deviceOnboardingMuteUnmute;
       case 2:
         return context.l10n.starConversation;
       default:
@@ -421,7 +421,7 @@ class _DeviceSettingsState extends State<DeviceSettings> {
                   ),
                   ListTile(
                     title: Text(
-                      context.l10n.pauseResumeRecording,
+                      context.l10n.deviceOnboardingMuteUnmute,
                       style: const TextStyle(color: Colors.white, fontWeight: FontWeight.w400),
                     ),
                     trailing: currentAction == 1 ? const Icon(Icons.check, color: Colors.white, size: 20) : null,
--- a/app/lib/services/devices/transports/native_ble_transport.dart
+++ b/app/lib/services/devices/transports/native_ble_transport.dart
@@ -83,12 +83,12 @@ class NativeBleTransport extends DeviceTransport {
 
   @override
   Future<void> disconnect() async {
-    if (_state == DeviceTransportState.disconnected) return;
+    if (_state == DeviceTransportState.disconnected && _streamControllers.isEmpty) return;
 
     _updateState(DeviceTransportState.disconnecting);
 
     // Unsubscribe all active streams
-    for (final key in _streamControllers.keys.toList()) {
+    for (final key in _activeSubscriptionKeys.toList()) {
       final parts = key.split(':');
       if (parts.length == 2) {
         try {
@@ -97,6 +97,8 @@ class NativeBleTransport extends DeviceTransport {
       }
     }
 
+    _activeSubscriptionKeys.clear();
+    _subscribedSubscriptionKeys.clear();
     _closeAllStreams();
     _services = [];
 
@@ -145,18 +147,23 @@ class NativeBleTransport extends DeviceTransport {
 
     if (!_streamControllers.containsKey(key)) {
       _streamControllers[key] = StreamController<List<int>>.broadcast();
+      _activeSubscriptionKeys.add(key);
       if (_hasCharacteristic(serviceUuid, characteristicUuid)) {
-        _subscribeCharacteristic(serviceUuid, characteristicUuid);
+        unawaited(_subscribeCharacteristic(serviceUuid, characteristicUuid));
       }
     }
 
     return _streamControllers[key]!.stream;
   }
 
-  void _subscribeCharacteristic(String serviceUuid, String characteristicUuid) {
+  Future<void> _subscribeCharacteristic(String serviceUuid, String characteristicUuid) async {
+    final key = '${serviceUuid.toLowerCase()}:${characteristicUuid.toLowerCase()}';
+    if (_subscribedSubscriptionKeys.contains(key)) return;
+    _subscribedSubscriptionKeys.add(key);
     try {
-      _hostApi.subscribeCharacteristic(_peripheralUuid, serviceUuid, characteristicUuid);
+      await _hostApi.subscribeCharacteristic(_peripheralUuid, serviceUuid, characteristicUuid);
     } catch (e) {
+      _subscribedSubscriptionKeys.remove(key);
       Logger.debug('[NativeBleTransport] Failed to subscribe $serviceUuid:$characteristicUuid: $e');
     }
   }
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Reviewing this draft — strong work on the root cause. The fix is exactly right: keeping the Dart broadcast stream controllers alive across transient Android BLE reconnects (instead of closing/recreating them) means OmiDeviceConnection's button subscription stays attached and keeps receiving double-tap notifications after re-subscription. The activeSubscriptionKeys/subscribedSubscriptionKeys split cleanly tracks intent-to-subscribe vs. already-subscribed, and the guard against the double-fire disconnect case (checking _streamControllers.isNotEmpty before clearing) is well-reasoned.

A few observations for when this moves out of draft:

1. _subscribeActiveCharacteristics on initial connect. getCharacteristicStream now adds to _activeSubscriptionKeys and fires _subscribeCharacteristic itself, so calling _subscribeActiveCharacteristics at device-ready is only relevant for subscriptions requested before discovery completed. That is correct for the button-listener case (requested before services are known), just worth confirming the re-subscribe path is idempotent — it is, via the _subscribedSubscriptionKeys.contains(key) guard. Good.

2. unawaited on subscribe calls. Moving _subscribeCharacteristic to async + unawaited(...) is fine since failures are caught and logged with rollback of the subscribed-key set. If subscribe fails, the key is removed from _subscribedSubscriptionKeys but stays in _activeSubscriptionKeys, so a later reconnect will retry. That is the desired behavior.

3. Settings label change. Switching double-tap action 1 from pauseResume/pauseResumeRecording to deviceOnboardingMuteUnmute (resolves to "Mute / Unmute") aligns the settings UI with onboarding. The scope boundary note in the PR body — that firmware 3.0.20 has no true hardware mute command — is accurate and appropriately scoped out.

CI passes (Dart Analyze and Tests, Android Compile Smoke, Hygiene, Formatting). Tests are hermetic and exercise the reconnect + resubscribe path through the BleBridge seam. No concerns blocking this once it is ready.

---

Git-on-my-level:
Reviewed this as glm-5.2. Nice fix — the root-cause analysis and the repair both look sound.

**`native_ble_transport.dart`** — The core fix is correct. Previously `_resubscribeAfterReconnect` created fresh stream controllers, orphaning the existing `OmiDeviceConnection` button subscription (closed stream → no more double-tap events to `CaptureController`). The new path reuses the existing controller when it's still open and only creates one if it's null/closed. I verified the ordering is safe: `_services` is populated from the native callback before `_hasCharacteristic` is consulted in both `_subscribeActiveCharacteristics` and `_resubscribeAfterReconnect`, so the post-reconnect service discovery drives the guard correctly. The `_subscribedSubscriptionKeys` dedup is handled well — it blocks duplicate native subscribe calls while in-flight, and the catch path removes the key so a failed subscription remains retryable. `disconnect()`'s early-return now also requires `_streamControllers.isEmpty`, which correctly lets a dangling-stream cleanup proceed even when the transport already reports disconnected.

**`device_settings.dart`** — The copy change from `pauseResume`/`pauseResumeRecording` to `deviceOnboardingMuteUnmute` ("Mute / Unmute") is the right call for consistency with onboarding, and the key already exists across locales.

**Tests** — Both new files are solid. The transport test exercises the real reconnect lifecycle (disconnect → device-ready → resubscribe → value delivery) through mock Pigeon handlers and asserts both that subscribe is called twice and that values reach the same listener before and after reconnect — exactly the regression this fix targets. The widget test covers the rebranded label.

The scope boundary note in the PR body (no true hardware mute on firmware 3.0.20, mic-gain `0` only attenuates) is honest and accurate; app-side recording pause is unchanged here.

Leaving for human maintainer review for final merge sign-off.

---

Git-on-my-level:
Thanks for this — the root-cause analysis and regression coverage are solid. The fix correctly preserves Dart notification streams across transient BLE reconnects instead of orphaning existing listeners.

Per-file notes:

**`native_ble_transport.dart`**
- Core fix in `_handleConnectionState`: no longer calling `_closeAllStreams()` on a transient native disconnect is the right call. Explicit `disconnect()` and `dispose()` still close streams, so the cleanup paths are preserved while the button listener survives the reconnect.
- The `_subscribedSubscriptionKeys` dedup is sound — the key is added synchronously before the `await _hostApi.subscribeCharacteristic(...)`, which prevents concurrent double-subscription races, and it is removed on failure so a retry is possible.
- `_handleDeviceReady` now storing `_services = services` and calling `_subscribeActiveCharacteristics()` correctly covers the case where `getCharacteristicStream` was called before service discovery completed.
- `_resubscribeAfterReconnect` reusing the existing open `StreamController` (instead of unconditionally creating a replacement) is exactly what prevents the orphaned `OmiDeviceConnection` button subscription.

**`device_settings.dart`**
- Straightforward consistency fix — double-tap action 1 now uses the existing `deviceOnboardingMuteUnmute` string to match onboarding.

**`native_ble_transport_test.dart`**
- Good hermetic regression test: verifies the listener survives a disconnect→reconnect cycle, receives both characteristic values, and that `subscribeCharacteristic` is invoked twice (once per connection). The pigeon mock note for `getBluetoothState` returning `['on']` is accurate.

**`device_settings_double_tap_test.dart`**
- Clean widget test confirming the label swap and absence of the old 'Pause/Resume' strings.

One note for a human maintainer: the PR scope boundary honestly states FW 3.0.20 has no true device-side mic mute and the app-side recording pause is unchanged. The 'Mute / Unmute' label is consistent with onboarding terminology, but it describes app-side recording pause rather than hardware mic muting — worth a quick confirm that this is the desired user-facing framing.

_Review by glm-5.2 (Omi maintainer monitor)._

---

PR #11195, opened 2026-08-07
fix(billing): recover failed payments and defer canceled plan changes

undivisible:
## What changed and why

Make billing recovery discoverable for inactive or past-due Stripe subscriptions, return plan descriptions in the catalog, and prevent changing plans until a cancel-at-period-end subscription actually ends. Same-plan reactivation remains available.

none

## How it was verified

- `make preflight` passed all 10 selected checks.
- Focused backend tests passed: payment promotions (18), subscription plans (16), and available plans resilience (5).
- Python compilation and `git diff --check` passed.
- Live Stripe account recovery and portal invoice configuration were not exercised locally.

## Tests

- Added regression coverage for pending-cancellation plan changes and cancellation timing.
- Added catalog description assertions and preserved the existing payment/subscription test paths.

## Failure class (fixes)

Failure-Class: none

Line-Count-Exception: backend/routers/payment.py | 1483 -> 1534 | Stripe-backed stale-subscription recovery and deferred plan-change boundary with regression coverage.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11195?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches payment eligibility, upgrades, and Stripe/Firestore reconciliation—incorrect matching could block legitimate upgrades or allow wrong checkouts; UI/backend must stay aligned on pending-cancellation rules.
> 
> **Overview**
> Tightens **cancel-at-period-end** behavior end-to-end: plan switches to a different price are blocked until the current period ends, while **same plan + billing interval** reactivation stays allowed (including legacy price IDs via `price_ids_match_plan_and_interval`).
> 
> **Backend** enforces this in `can_user_make_payment`, upgrade (409 when pending cancellation), and `_try_reactivate_subscription` with a **Firestore → Stripe** fallback when local subscription rows are stale, plus observability via `record_fallback`. The available-plans API now surfaces plan **descriptions** on monthly/annual options.
> 
> **Web and Flutter** mirror the rules (disable other plans, client-side guards, treat `current_price_id` as “current”), expose **billing portal** for anyone with a Stripe subscription (not only “unlimited”), refresh subscription after returning from the portal, and plumb `stripe_subscription_id` through the web client types/API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c9a2d58933cf66fd648751c5141e0efe842dcd1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/lib/pages/settings/widgets/plans_sheet.dart
+++ b/app/lib/pages/settings/widgets/plans_sheet.dart
@@ -358,13 +358,13 @@ class _PlansSheetState extends State<PlansSheet> {
     Map<String, dynamic>? selectedPlanData;
     if (tierId != null) {
       selectedPlanData = plans.cast<Map<String, dynamic>>().firstWhereOrNull(
-        (plan) => plan['plan_id'] == tierId && plan['interval'] == (isYearly ? 'year' : 'month'),
-      );
+            (plan) => plan['plan_id'] == tierId && plan['interval'] == (isYearly ? 'year' : 'month'),
+          );
     }
     // Fallback to old behavior (first plan matching interval) for backwards compat
     selectedPlanData ??= plans.cast<Map<String, dynamic>>().firstWhereOrNull(
-      (plan) => plan['interval'] == (isYearly ? 'year' : 'month'),
-    );
+          (plan) => plan['interval'] == (isYearly ? 'year' : 'month'),
+        );
 
     if (selectedPlanData == null) {
       AppSnackbar.showSnackbarError(context.l10n.selectedPlanNotAvailable);
@@ -939,8 +939,7 @@ class _PlansSheetState extends State<PlansSheet> {
                             builder: (context) {
                               // Check if subscription period has ended
                               final sub = provider.subscription?.subscription;
-                              final periodEnded =
-                                  sub?.currentPeriodEnd != null &&
+                              final periodEnded = sub?.currentPeriodEnd != null &&
                                   DateTime.fromMillisecondsSinceEpoch(
                                     sub!.currentPeriodEnd! * 1000,
                                   ).isBefore(DateTime.now());
@@ -1015,8 +1014,7 @@ class _PlansSheetState extends State<PlansSheet> {
                         // Training Data Opt-in Option - only show after plans are loaded
                         Consumer2<UsageProvider, UserProvider>(
                           builder: (context, usageProvider, userProvider, child) {
-                            final shouldShowTrainingOption =
-                                _showTrainingDataOptIn &&
+                            final shouldShowTrainingOption = _showTrainingDataOptIn &&
                                 !usageProvider.isLoadingPlans &&
                                 usageProvider.availablePlans != null;
 
@@ -1178,8 +1176,7 @@ class _PlansSheetState extends State<PlansSheet> {
                             final isOnAnnualPlan = currentPlan?['interval'] == 'year';
                             final hasScheduledUpgrade = _hasScheduledUpgrade();
                             final usageProvider = context.read<UsageProvider>();
-                            final shouldShowContinueButton =
-                                !isOnAnnualPlan &&
+                            final shouldShowContinueButton = !isOnAnnualPlan &&
                                 !hasScheduledUpgrade &&
                                 !isCancelled &&
                                 !usageProvider.isLoadingPlans &&
--- a/app/lib/pages/settings/widgets/plans_sheet.dart
+++ b/app/lib/pages/settings/widgets/plans_sheet.dart
@@ -376,6 +376,9 @@ class _PlansSheetState extends State<PlansSheet> {
     // Check if user is upgrading from monthly to annual
     final provider = context.read<UsageProvider>();
     final currentSub = provider.subscription?.subscription;
+    if (currentSub?.cancelAtPeriodEnd == true && priceId != currentSub?.currentPriceId) {
+      return;
+    }
     // Only show "no charge until renewal" dialog for same-tier monthly→annual switch.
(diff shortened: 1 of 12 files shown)


Git-on-my-level (dismissed):
Thanks for tackling this — the overall direction is cohesive and the added focused backend coverage is helpful. I found one billing-state blocker before this should merge:

- `can_user_make_payment()` still skips active Stripe subscriptions that are `cancel_at_period_end` when the local Firestore subscription is missing/stale/basic. In that path `_has_active_stripe_subscription()` returns false for canceling subscriptions, `_try_reactivate_subscription()` only handles same-price reactivation, and then checkout can create a new subscription for a different target price before the existing subscription period ends. Since Stripe is already used as the recovery source of truth in this branch, the pending-cancellation/defer-plan-change rule needs to be enforced there as well: allow same-price reactivation, but reject different target prices until the current period ends.

Also, the Formatting check is currently failing and needs to go green.

Given this touches billing/subscription state transitions, I’m leaving this as changes requested and adding maintainer/security review labels for the billing-risk surface. No agent-behavior changes were found.

---

kodjima33 (approved):
Payment recovery + deferred plan changes look correct and are unit-tested. Approving only — billing paths are Nik's call to land.

Git-on-my-level (dismissed):
Thanks for the follow-up — the earlier stale/basic Firestore branch blocker is largely addressed, and the checks are now green. I still think this needs one more billing-path fix before merge:

- In the stale/basic Firestore recovery path, `can_user_make_payment()` now returns success for a same-price pending-cancellation subscription found from Stripe, but `create_checkout_session_endpoint()` then calls `_try_reactivate_subscription()`, which reloads only `users_db.get_user_subscription(uid)` and returns `None` when that local row has no/currently-stale `stripe_subscription_id`. That drops into the normal checkout-session flow instead of clearing `cancel_at_period_end` on the Stripe subscription that was just found. Please carry the discovered Stripe subscription through to reactivation, or otherwise make `_try_reactivate_subscription()` use the same Stripe source-of-truth lookup for this branch, so same-plan recovery actually reactivates instead of creating/attempting a new checkout.

One additional recovery UX point worth covering while you are there: the newly exposed mobile billing-portal path opens the portal, but after the user returns/closes it the app does not refresh subscription state, so a paid invoice/recovered subscription can continue to display as Free until a manual reload. Mirroring the checkout refresh behavior would make the recovery path visible immediately.

Given this still touches billing/subscription state transitions, keeping this as changes requested and preserving the existing maintainer/security review labels. No agent-behavior changes were found.

---

kodjima33 (approved):
Billing recovery + deferred plan changes on cancel-at-period-end — approve only; Stripe money path stays with Nik.

Git-on-my-level (commented):
Thanks for the update — I rechecked the current head and the stale/basic Firestore reactivation blocker from the earlier automation review looks addressed now.

- `backend/routers/payment.py`: `_try_reactivate_subscription()` now falls back to `find_active_paid_subscription_for_user()` when the local subscription row is missing/stale, so the same-plan pending-cancellation recovery path can clear `cancel_at_period_end` instead of falling through to a fresh checkout. `upgrade_subscription_endpoint()` also blocks plan changes while Stripe still has a pending cancellation, which matches the intended billing invariant.
- `backend/utils/subscription.py`: `is_pending_cancellation()` centralizes the period-end check, and `can_user_make_payment()` now treats a Stripe-discovered pending-cancellation subscription as active for different-price checkout while still allowing same-price reactivation. That covers the stale/basic branch the previous review called out.
- `backend/tests/unit/test_payment_promotion_codes.py` and `backend/tests/unit/test_subscription_plans.py`: the new regression tests cover both the Stripe fallback reactivation and the stale/basic defer-different-price behavior, including same-price reactivation staying allowed.
- `backend/tests/unit/test_available_plans_resilience.py`: the added assertions verify plan descriptions are returned in the available-plans catalog, matching the backend response change in `get_available_plans_endpoint()`.
- `app/lib/pages/settings/widgets/plans_sheet.dart`: the mobile portal path now refreshes subscription state after returning from Stripe billing, which addresses the earlier mobile recovery-visibility concern.
- `web/app/src/components/settings/PlansSheet.tsx`, `web/app/src/components/settings/SettingsPage.tsx`, `web/app/src/lib/api.ts`, and `web/app/src/types/user.ts`: the web UI now carries `stripe_subscription_id`/`current_price_id`, disables non-current plan selection during pending cancellation, and exposes billing/invoice access for users who still have a Stripe subscription.

One non-blocking follow-up I’d still consider: the newly exposed web “Billing & Invoices” portal path opens Stripe but does not refresh subscription state on window focus/return the way the checkout path does. If a user pays an overdue invoice from that portal, the web UI may stay stale until another refresh. Not a reason for me to keep a blocking change request, but it is worth tightening before/after merge if this is part of the recovery UX.

I could not run the focused backend pytest suite in this sandbox because the checkout is missing backend test dependencies (`google.api_core` import fails from `tests/unit/conftest.py`). Static review of the changed billing paths and the added regression coverage looks consistent.

Because this touches Stripe billing/subscription state, I’m leaving final sign-off to the human maintainer/billing owner rather than formally approving. No agent-behavior changes were found.

---
*by AI on behalf of David — final billing/Stripe behavior should be confirmed by a human maintainer.*

Issue #10950, opened 2026-07-31
[Alert] Firestore Document Reads exceeded 10k/s (10,214/s) — 2026-07-31 20:40 UTC


beastoin:
## Alert

- **Source:** Google Cloud Alerting (automated)
- **Policy:** Firestore Instance - Document Reads > 10k/s
- **Project:** based-hardware
- **Metric:** `firestore.googleapis.com/document/read_count`
- **Threshold:** 10,000 reads/s
- **Observed value:** 10,214.47 reads/s
- **Start time:** 2026-07-31 20:40 UTC
- **Type:** QUERY
- **Prior occurrence:** #10817 (2026-07-29 03:08 UTC, 10,145/s)

beastoin:
## Third occurrence — 2026-08-09 01:36 UTC

- **Observed value:** 10,277/s (escalating: 10,145 → 10,214 → 10,277 across 3 alerts in 11 days)
- **Source:** Google Cloud Alerting, same policy `Firestore Instance - Document Reads > 10k/s`

### Escalation pattern

| Date | Value | Issue |
|------|-------|-------|
| 2026-07-29 03:08 UTC | 10,145/s | #10817 |
| 2026-07-31 20:40 UTC | 10,214/s | #10950 |
| 2026-08-09 01:36 UTC | 10,277/s | this comment |

### Fixes merged since first alert

| PR | Title | Merged |
|----|-------|--------|
| #10750 | bound Firestore list reads (504 fix) | 2026-07-28 |
| #10778 | action-items due-range index | 2026-07-28 |
| #10781 | action-items composite index | 2026-07-29 |
| #10902 | tighten knowledge graph read budget | 2026-07-30 |
| #11062 | deduplicate all-time usage reads | 2026-08-04 |
| #11197 | bound chat-quota month read | 2026-08-07 |

Despite 6 merged read-reduction PRs, the alert continues to fire at increasing values. The fixes addressed individual hot paths but the aggregate read volume has not decreased below threshold.

### Root cause investigation

#9889 (Reduce Firestore query and listener read amplification) is still open — the broader investigation into read amplification patterns across backend, mobile, desktop, and web. No assignee, no timeline.

**Recommendation:** This needs a systematic audit of top read-contributing paths (via Firestore usage metrics breakdown by collection/operation), not more individual hot-path fixes. The escalating trend suggests either traffic growth is outpacing optimizations, or major read sources haven't been identified yet.

beastoin:
[...]

Issue #10817, opened 2026-07-29
[Alert] Firestore Document Reads exceeded 10k/s (10,145/s) — 2026-07-29 03:08 UTC


beastoin:
## Alert

- **Source:** Google Cloud Alerting (automated)
- **Policy:** Firestore Instance - Document Reads > 10k/s
- **Project:** based-hardware
- **Metric:** `firestore.googleapis.com/document/read_count`
- **Threshold:** 10,000 reads/s
- **Observed value:** 10,145.17 reads/s
- **Start time:** 2026-07-29 03:08 UTC
- **Type:** QUERY

beastoin:
Consolidated tracking issue filed: **#11282** — includes all 3 alert occurrences, the full 6-PR fix history, and acceptance criteria requiring production metric verification. This issue is tracked there.

undivisible:
Closing per automated triage (2026-08-20): superseded: folded into live #11282 Firestore-reads tracking. See /tmp/omi-issue-triage.md.

Issue #9889, opened 2026-07-16
Reduce Firestore query and listener read amplification


Git-on-my-level:
## Goal

Identify and reduce avoidable Firestore read amplification in Omi's high-volume query and realtime-listener paths without changing product semantics, consistency expectations, or user-data retention.

## Current shape of the problem

The application has both:

- backend paths that execute Firestore queries and collection scans; and
- client/realtime paths that attach snapshot listeners and can reconnect or refetch.

The initial investigation suggests likely opportunity areas are duplicate listener attachment, reconnect loops, polling/repeated fetches, broad or unbounded list queries, fan-out reads, and cache misses. This issue intentionally excludes billing figures, production record contents, user identifiers, and exports of production data.

## Requested investigation

1. Build a source-level map of the main Firestore read paths in backend, mobile, desktop, and web surfaces.
2. Identify query/listener patterns that can amplify reads, prioritizing shared/high-traffic code paths rather than one-off admin scripts.
3. Propose privacy-safe aggregated telemetry that can attribute read behavior without storing user content or raw identifiers. Useful dimensions may include route/feature, collection category, query shape/boundedness, listener attach/re-attach/dispose events, returned-vs-read document counts, retry reason, and cache source/hit status.
4. Recommend a small, testable first fix with explicit expected behavior and regression coverage.
5. Document which proposals require product/consistency validation before rollout.

## Constraints

- Do not inspect, export, or include production document contents, user identifiers, credentials, or billing data in issue discussion, tests, or logs.
- Prefer source inspection, local emulator tests, and aggregated privacy-safe instrumentation.
- Do not change listener semantics, query ordering, pagination, cache policy, or data retention without focused behavioral tests and a separately reviewed implementation.
- Treat Firestore reads as a cross-surface concern: a local optimization must not introduce stale UI, missed realtime updates, duplicate processing, or retry storms.

## Acceptance criteria

- [ ] Source-level inventory of top shared read/query/listener paths and their suspected amplification mechanism.
- [ ] Proposal for minimal privacy-safe aggregated telemetry, including cardinality controls and no user-content logging.
- [ ] Ranked shortlist of candidate changes, each with risk and validation plan.
- [ ] One narrowly scoped first implementation proposal with regression-test outline.
- [ ] Clear statement of any production evidence still needed before selecting the first fix.

aryanorastar:
**VERIFY (contributor triage):** Still OPEN as an investigation epic, but substantial amplification work already landed:

[...]

PR #11197, opened 2026-08-07
perf(backend): bound the chat-quota month read to the current month

kodjima33:
## What

`get_monthly_chat_usage` — the chat-quota counter read — now issues one bounded
document-id range query over the current month instead of listing the account's
entire `llm_usage` collection and fetching the in-month days one at a time.

## Why

This is not a reporting path. It runs inside `enforce_chat_quota`, so **every
chat request pays for it** (`backend/routers/desktop_chat.py:1217`,
`backend/routers/chat.py:299,586`), plus the overage check
(`backend/utils/overage.py:112`) and `/v1/users/me/usage-quota`.

The old shape (`backend/database/user_usage.py`):

```python
for doc in llm_usage_ref.list_documents():      # every day the account ever wrote
    if not doc.id.startswith(month_prefix):     # filtered client-side
        continue
    snap = doc.get()                            # one more RPC per in-month day
```

`llm_usage` holds one `YYYY-MM-DD` doc per day of activity, so both the reads and
the round-trips grew with **account age**: a two-year daily-active account listed
~730 document ids and then made ~30 sequential `get()` calls to answer "how many
questions this month". New accounts are cheap; the longest-tenured, heaviest users
— the ones behind the read spikes in #10950 / #10817 — are charged the most per
message, and they pay it in latency on every send.

This is the "narrowly scoped first implementation" #9889 asks for, with the source
and measurement its last triage comment asked for before another change landed.

## What changed

- `_current_month_llm_usage_docs()` streams `[YYYY-MM-01, <next month>-01)` by
  document id — a single-field `__name__` range, so no composite index (the
  `firestore-query-coverage` preflight check agrees).
- The read is reported through the existing `record_firestore_read` telemetry as
  a new `chat_quota_monthly_usage` **bounded** family, matching how
  `get_monthly_usage_stats_since` and `_read_all_time_usage` already report — the
  per-family, no-user-content attribution #9889 wants.
- `_next_month()` extracted so the query boundary and the `reset_at` boundary
  cannot drift apart.

No change to which documents are summed or how their fields are interpreted.

## Verification

**Differential run against a real Firestore emulator** (`gcloud emulators
firestore`, 56 seeded `llm_usage` docs spanning 2025–2027 for one uid), executing
the old reader and the new one side by side:

| `now` | old | new | docs read |
|---|---|---|---|
| 2026-08 | 28 questions / $2.75 | 28 / $2.75 | **61 → 5** |
| 2026-12 (Dec→Jan rollover) | 38 questions | 38 | **60 → 4** |
| 2024-03 (month with no data) | 0 | 0 | **56 → 0** |

`RESULT: PASS — identical totals, bounded reads`. Boundary documents
`2026-07-31`, `2026-09-01` and `2027-01-01` stay excluded; `2026-08-01` and
`2026-08-31` stay included.

**Regression tests** (`backend/tests/unit/test_user_usage.py`): the collection
fake now applies the real `__name__` range filters instead of returning
everything, and raises if anything calls `list_documents()`, so an all-time scan
cannot come back unnoticed. Both new tests fail on the parent commit
(`AssertionError: chat-quota reads must stay bounded to the current month`) and
pass here.

The full suite also caught a first attempt that imported
`google.cloud.firestore_v1.field_path`: several backend tests stub
`google.cloud.firestore_v1` as a plain module, so that submodule import broke
`test_users_add_sample_transaction.py` at import time. The field path is now
spelled literally (`'__name__'`, which is exactly what `FieldPath.document_id()`
returns) and that file passes again.

Commands run:

- `pytest tests/unit/test_user_usage.py tests/unit/test_chat_quota.py` → 27 passed
- `backend/test.sh` (full, 841 files) → the 14 failing files are **identical on
  this branch and on `origin/main`** when re-run with the same runner
  (`test_agent_vm_startup`, `test_sync_v2`, `test_http_timeout_defaults`,
  `test_action_item_canonical_contract`, … — local CPU-time guard/environment
  failures, none of them touched here)
- `backend/scripts/typecheck.sh` → 0 errors
- `make preflight` → 24 checks passed
- `black --line-length 120 --skip-string-normalization` → clean

Refs #9889, #10950, #10817.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11197?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/firestore_read_metrics.py
+++ b/backend/database/firestore_read_metrics.py
@@ -13,6 +13,7 @@ class FirestoreReadFamily(StrEnum):
     ACTION_ITEMS_LIST = 'action_items_list'
     LISTEN_MONTHLY_USAGE = 'listen_monthly_usage'
     ALL_TIME_USAGE = 'all_time_usage'
+    CHAT_QUOTA_MONTHLY_USAGE = 'chat_quota_monthly_usage'
 
 
 class FirestoreReadMode(StrEnum):
--- a/backend/database/user_usage.py
+++ b/backend/database/user_usage.py
@@ -12,11 +12,42 @@ from models.user_usage import UsageStats
 
 logger = logging.getLogger(__name__)
 
+# Firestore's document-id sentinel field path (`FieldPath.document_id()`), spelled
+# literally: tests that stub `google.cloud.firestore_v1` as a plain module cannot
+# import its `field_path` submodule.
+_DOCUMENT_ID_FIELD = '__name__'
+
 
 def _typed_doc(doc: Any) -> Dict[str, Any]:
     """Typed adapter for a Firestore snapshot's `to_dict()` (SDK stub gap)."""
     raw: object = doc.to_dict()
     return cast(Dict[str, Any], raw) if isinstance(raw, dict) else {}
+
+
+def _next_month(now: datetime) -> Tuple[int, int]:
+    """(year, month) of the UTC month after ``now`` — the quota bucket boundary."""
+    if now.month == 12:
+        return now.year + 1, 1
+    return now.year, now.month + 1
+
+
+def _current_month_llm_usage_docs(llm_usage_ref: Any, now: datetime) -> Iterable[Any]:
+    """Stream only the current month's `llm_usage/{YYYY-MM-DD}` docs.
+
+    Bounded by document id, so the read stays at "days elapsed this month"
+    regardless of how long the account has existed. Listing the whole
+    collection and fetching the in-month days one at a time grew both the read
+    count and the round-trips with account age, on the path every chat request
+    takes through ``enforce_chat_quota``.
+    """
+    next_year, next_month = _next_month(now)
+    start = llm_usage_ref.document(f'{now.year}-{now.month:02d}-01')
+    end = llm_usage_ref.document(f'{next_year}-{next_month:02d}-01')
+    return (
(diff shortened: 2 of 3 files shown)


kodjima33:
CI note: the one red check (`Backend unit suite` → `tests/unit/test_kg_user_type_mismatch.py`) is pre-existing on `main`, not from this PR.

- This PR's run: [31196375116](https://github.com/BasedHardware/omi/actions/runs/31196375116) — only `test_kg_user_type_mismatch.py` fails, with `ModuleNotFoundError: No module named 'utils.conversations.transcript_for_llm'` raised from `utils/conversations/process_conversation.py:15`.
- `main`'s own Backend Unit Tests run from before this branch existed, [31188359943](https://github.com/BasedHardware/omi/actions/runs/31188359943), fails on the same single file with the same error.

Nothing in this diff touches `utils/conversations/`. Every other check on this PR is green (26 pass), including the `firestore-query-coverage` and `backend-module-isolation` manifest checks and the cubic review (no issues, 5/5).

Locally I also ran the full `backend/test.sh` on this branch and on a `origin/main` worktree with the same file list — **identical failure sets**, all environment/CPU-time-guard noise.

PR #11212, opened 2026-08-07
feat(app): replace single-date conversation filter with a date range

jnnabugwu:
## Summary
- Replaces the single-day conversation filter with a start/end date range (`selectedStartDate`/`selectedEndDate` on `ConversationProvider`, inclusive filtering).
- Adds a range-mode calendar picker (`calendar_date_picker_sheet.dart`) and wires it into the conversations page, home page, and search widget.

## Test plan
- [x] `flutter test test/providers/conversation_date_range_filter_test.dart` — 6/6 passing (range inclusivity, no-range-selected, filter set/clear, clearUserData reset).
- [ ] Manual verification on iOS Simulator — **blocked**, not completed.

## Known blocker: could not run the app locally to verify manually
Attempted `flutter run` on iOS Simulator to exercise the real user-facing path per the Definition of Done, and hit an environment blocker unrelated to this change:
- The `dev` scheme's Watch app companion (`omiWatchApp`) is signed to Omi's Apple Developer team (`9536L8KLMP`), which this machine has no account/certificate for.
- Working around it with `CODE_SIGNING_ALLOWED=NO` unblocks compilation but then fails at Watch-app embed validation (`Embedded binary's bundle identifier is not prefixed with the parent app's bundle identifier` / platform mismatch) — signing-disabled builds can't satisfy Xcode's embed checks for the companion app.
- This is a local signing/team-access issue, not a defect in this diff — none of the changed files touch the Watch app or native iOS code.

Manual verification is deferred until iOS Simulator access is unblocked (team credentials) or someone with a working local build can confirm the UI path.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/conversations/conversations_page.dart
+++ b/app/lib/pages/conversations/conversations_page.dart
@@ -144,7 +144,7 @@ class _ConversationsPageState extends State<ConversationsPage> with AutomaticKee
   // folder-tab chips would hide on empty filtered results, leaving no way to
   // clear filters short of restarting the app.
   bool _hasActiveFilter(ConversationProvider provider) {
-    return provider.showStarredOnly || provider.selectedFolderId != null || provider.selectedDate != null;
+    return provider.showStarredOnly || provider.selectedFolderId != null || provider.selectedStartDate != null;
   }
 
   Widget _buildNoConversationsHero(BuildContext context) {
@@ -308,7 +308,7 @@ class _ConversationsPageState extends State<ConversationsPage> with AutomaticKee
               Consumer<HomeProvider>(
                 builder: (context, homeProvider, _) {
                   final isSearchActive = homeProvider.showConvoSearchBar || convoProvider.previousQuery.isNotEmpty;
-                  final hasCalendarFilter = convoProvider.selectedDate != null;
+                  final hasCalendarFilter = convoProvider.selectedStartDate != null;
                   final prefs = SharedPreferencesUtil();
                   if (convoProvider.showDailySummaries || isSearchActive || hasCalendarFilter) {
                     return const SliverToBoxAdapter(child: SizedBox.shrink());
--- a/app/lib/pages/conversations/conversations_page.dart
+++ b/app/lib/pages/conversations/conversations_page.dart
@@ -52,7 +52,8 @@ typedef _ConversationPageSnapshot = ({
   String previousQuery,
   String? selectedFolderId,
   String? selectedSpeakerId,
-  DateTime? selectedDate,
+  DateTime? selectedStartDate,
+  DateTime? selectedEndDate,
   bool showStarredOnly,
   bool showDailySummaries,
   bool hasDailySummaries,
@@ -81,7 +82,8 @@ String conversationLoadMoreFilterKey({
   required String query,
   required String? folderId,
   required String? speakerId,
-  required DateTime? date,
+  required DateTime? startDate,
+  required DateTime? endDate,
   required bool starredOnly,
   required bool discarded,
   required bool shortOnly,
@@ -91,7 +93,8 @@ String conversationLoadMoreFilterKey({
       query,
       folderId ?? '',
       speakerId ?? '',
-      date?.toIso8601String() ?? '',
+      startDate?.toIso8601String() ?? '',
+      endDate?.toIso8601String() ?? '',
       starredOnly,
       discarded,
       shortOnly,
@@ -110,7 +113,8 @@ _ConversationPageSnapshot _conversationPageSnapshot(
     previousQuery: conversations.previousQuery,
     selectedFolderId: conversations.selectedFolderId,
     selectedSpeakerId: conversations.selectedSpeakerId,
-    selectedDate: conversations.selectedDate,
+    selectedStartDate: conversations.selectedStartDate,
+    selectedEndDate: conversations.selectedEndDate,
     showStarredOnly: conversations.showStarredOnly,
     showDailySummaries: conversations.showDailySummaries,
     hasDailySummaries: conversations.hasDailySummaries,
@@ -283,7 +287,8 @@ class _ConversationsPageState extends State<ConversationsPage> with AutomaticKee
       query: provider.previousQuery,
       folderId: provider.selectedFolderId,
       speakerId: provider.selectedSpeakerId,
-      date: provider.selectedDate,
+      startDate: provider.selectedStartDate,
+      endDate: provider.selectedEndDate,
       starredOnly: provider.showStarredOnly,
       discarded: provider.showDiscardedConversations,
       shortOnly: provider.showShortConversations,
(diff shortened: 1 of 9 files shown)


Git-on-my-level (dismissed):
Thanks for the focused cleanup here — the duplicated picker code is a good thing to consolidate, and the new provider tests cover the in-memory inclusive range predicate.

I’m requesting changes because the fetch path can still return incomplete results for a selected range. I verified that the app’s local grouping/filtering uses `startedAt ?? createdAt`, but `getConversationsResult()` sends `start_date`/`end_date` to `/v1/conversations`, and that backend list path filters Firestore on `created_at`. If a conversation’s recording `startedAt` falls inside the selected range but its `created_at` is outside it, the server prefilter drops it before the client-side range predicate can include it.

Please align the server request/query with the same effective conversation date used by the provider (or avoid server-side narrowing before the local predicate), and add coverage for the server-fetch boundary in addition to the current in-memory provider tests. The PR description also calls out that the real UI path could not be manually verified yet, so this should still get human maintainer/product review before merge.

---

kodjima33 (approved):
Date-range conversation filter — sound implementation; holding the merge since it is a feature, not a bug fix.

Git-on-my-level:
Follow-up after the merge from main — the feature code is unchanged, so my prior technical concern persists but I want to put it in context.

The prior review flagged that `getConversationsResult()` sends `start_date`/`end_date` to `/v1/conversations`, which filters Firestore on `created_at`, while the local predicate in `_filterOutConvos` uses `startedAt ?? createdAt`. This means a conversation whose `startedAt` and `createdAt` fall on different days could be returned by the backend but dropped locally (or vice versa) at range boundaries.

To be clear: this mismatch is **pre-existing** — the old single-date filter had the exact same `startedAt ?? createdAt` vs `created_at` discrepancy. This PR does not introduce it, and in fact improves the situation by switching from exact-day equality to an inclusive range. The practical impact is limited to conversations captured very close to midnight where `startedAt` and `createdAt` straddle a day boundary.

What I checked file-by-file:

- **conversation_provider.dart**: `_filterOutConvos` range predicate (`convoDate.isBefore(startDay) || convoDate.isAfter(endDay)`) is correctly inclusive on both ends. `_getDateFilterRange()` emits `00:00:00.000`–`23:59:59.999` day boundaries. `filterConversationsByDateRange()`, `clearDateFilter()`, and `clearUserData()` all set/null both fields atomically.
- **calendar_date_picker_sheet.dart**: New `showConversationDateRangePicker()` correctly handles the single-date-selection edge case (`end` defaults to `start`). The `calendarType: CalendarDatePicker2Type.range` config is appropriate.
- **search_widget.dart** / **home/page.dart**: Both now delegate to the shared picker — ~187 lines of duplicated inline picker code removed and replaced with a single call. All `selectedDate` references correctly migrated to `selectedStartDate`.
- **conversations_page.dart**: `_hasActiveFilter` and `hasCalendarFilter` correctly check `selectedStartDate != null`.
- **analytics_manager.dart**: `calendarFilterApplied` now tracks `start_date`, `end_date`, and `range_days` — reasonable for range analytics.
- **conversation_date_range_filter_test.dart**: 5 tests cover range inclusivity (first/last day included, before/after excluded), no-range passthrough, filter set/clear, and `clearUserData` reset. The test docstring honestly notes it does not exercise `_getDateFilterRange()` or the server fetch path.

@kodjima33 has approved the feature commit. I'm leaving the prior CHANGES_REQUESTED in place since the `startedAt`/`created_at` concern is not resolved on this head, but I'm not re-requesting changes given the pre-existing nature and human approval. If you'd like to address it, the fix would be to align the local predicate field with the backend filter field (or vice versa) — but that's a separate cleanup, not a blocker for this PR.

Removed `needs-tests` (tests added), `dependency-review` (no dependency changes), and `privacy-review` (no persistence/permission/privacy-surface changes).

---

Git-on-my-level (commented):
Thanks @jnnabugwu — this is a solid, well-scoped change. Consolidating the two near-identical inline pickers into one shared `showConversationDateRangePicker` is the right call, and the provider tests give good coverage of the inclusive range predicate.

Per-file notes:

- **conversation_provider.dart**: The range predicate (`isBefore(startDay) || isAfter(endDay)`) is correct — day-truncated `DateTime`s give clean inclusive boundaries on both ends. `_getDateFilterRange` now returns `23:59:59.999` for the end boundary, the right inclusive full-day span. `filterConversationsByDateRange` / `clearDateFilter` / `clearUserData` all keep start+end in sync.
- **calendar_date_picker_sheet.dart**: Clean extraction. The Done callback correctly normalizes a single-pick (`end ?? start`) so a one-tap selection applies `start == end` rather than dropping the filter. The Done-button color change from `deepPurple` to white aligns with INV-UI-1 (verified in `docs/product/invariants/brand-ui.md`).
- **search_widget.dart / home/page.dart**: Good dedup — both now delegate to the shared picker.
- **analytics_manager.dart**: Sensible additions (`range_days` + both dates).
- **conversation_date_range_filter_test.dart**: Good boundary coverage (first-day, last-day, before, after, no-filter, set, clear, `clearUserData`). The scoping note about not exercising `_getDateFilterRange` is honest and appropriate.
- **pubspec.lock**: Transitive-only bumps (`characters`, `matcher`, `material_color_utilities`, `test_api`) — benign.

One pre-existing note (not a regression from this PR): the backend `get_conversations` path filters Firestore on `created_at` while the client groups/re-filters on `startedAt ?? createdAt`. For conversations where `startedAt` and `createdAt` fall on different local days this can cause edge-case mismatches. This predates this PR (same divergence exists in `main`'s single-date filter), so it's worth a separate follow-up rather than blocking here.

Minor nit: `filterConversationsByDateRange` does not enforce `start <= end` on the client side (the backend validates via HTTP 400 and the picker constrains selection, so this only affects a programmatic caller).

I also dismissed the prior changes-requested review on this PR — the `created_at`/`startedAt` concern it raised is reassessed as pre-existing rather than introduced here.

Leaving for human maintainer product review of the date-range filter UX (kodjima33 has already approved this head).

---

aryanorastar (changes requested):
Requesting changes (merge realism):

Shared `showConversationDateRangePicker` + inclusive day-truncated range on `ConversationProvider` is a clean UX win, and the 6 provider tests (inclusivity / clear / clearUserData) are the right lock. David’s latest notes on the tip look consistent with that.

**Blocker:** tip is `CONFLICTING` / `DIRTY` vs `main`. Size-gate APPROVE does not survive a conflicted tip — please rebase and keep `conversation_date_range_filter_test.dart` green.

Non-blocking after rebase:
- Author notes manual iOS smoke was blocked — fine to keep honest in the PR body (no kit here either).
- Confirm start>end is impossible in the sheet (or swapped) so the predicate cannot silently empty the list.

Happy to re-review a clean tip.

PR #11214, opened 2026-08-07
Preserve Agent VM state across image replacement

Git-on-my-level:
## Summary

- give every newly provisioned Agent VM a dedicated owner-labelled state disk for SQLite and the agent workspace
- migrate a stopped legacy VM through a read-only clone, exact manifest comparison, SQLite integrity check, and authenticated guest receipt
- move the same verified state disk across future image replacements and compensate every partial pre-cutover failure before reopening admission
- keep the browser profile explicitly ephemeral and keep production boot-image replacement hard-disabled

## Safety contract

- ordinary releases cannot replace boot images; migration is a separate development-only, one-owner allowlist
- candidate and disk mutations require deterministic names, numeric IDs, migration labels, owner labels, and expected attachment users
- proxy admission remains closed through the soak; candidate release identity and state receipt are rechecked before predecessor retirement
- interrupted first-copy attempts reconstruct only the unreceipted destination from the still-read-only source clone
- malformed or concurrent SQLite uploads cannot replace the prior open database
- existing mutable state is not snapshot-locked to the first-copy receipt; boot revalidates the durable database without rehashing or rejecting normal workspace changes
- state attachments remain `autoDelete: false` through clone cleanup, predecessor deletion, and journal completion; a failed final repair leaves the only state disk protected
- every owner records the exact durable migration journal ID atomically with journal creation, so crash recovery performs one bounded document read without a collection-group index or an unbounded history scan
- omitted GCE disk `users` metadata is treated as the provider's detached representation during rollback; malformed present metadata still fails closed
- pre-instance-ID deferred cleanup records are upgraded through a deletion-marker CAS and an exact provider-ID reread before deleting only a historically UID-derived VM name
- account deletion validates a reused state disk against the union of identity-fenced VM generations, while migration soak uses its immutable journaled release even after the active pointer advances
- migration metadata outages now fail startup without rewriting a valid receipt; legacy VM-name fallback is allowed only after a confirmed metadata 404
- deferred post-retirement state-disk lifecycle repair emits shared fallback telemetry while leaving the sole state disk protected
- provider-terminated legacy VMs may self-heal a stale Firestore `ready` status, while running providers, active sessions, and queued start/drain demand remain blocked
- legacy account deletion recovers identity from the UID-owned auth token plus a stable provider numeric ID, without requiring new IAM
- deferred post-deletion VM cleanup persists that same numeric provider ID before the user document can disappear
- both desktop request services use the same private-VPC readiness channel as the reconciler

Production migration remains disabled. This PR is intentionally inert until the checked-in development activation flow names one UID. The real GCP replacement path has not been exercised on this commit yet; after merge and dev auto-deploy I will run success, forced-failure compensation, soak, and identity-fenced cleanup against the authorized development owner only.

## Verification

- `BACKEND_PYTEST_WORKERS=8 ./backend/test.sh` — all 843 backend unit test files passed
- changed Agent VM, reconciler, account-cleanup, deployment-policy, fallback-observability, and route-isolation suite — 288 passed
- desktop deployment policy suite — 29 passed
- committed-diff PR preflight — all 28 selected checks passed
- `black --check --line-length 120 --skip-string-normalization ...` — passed
- `bash -n backend/agent_vm/startup.sh backend/scripts/apply-agent-vm-reconciler-iam.sh` — passed
- post-review Luna fix slices closed the fresh crash-recovery, retirement-ordering, late-cleanup, legacy-label, and cancelled-upload findings; direct integration checks passed on current `main`

none

Failure-Class: none


Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

PR #11218, opened 2026-08-08
fix(desktop): make Rewind span retained history

Archit-lal:
## Summary

- make Rewind one continuous retained-history timeline: it opens on a readable recent window, then progressively reveals older history as the user zooms or pans
- query an overscanned, bounded sample for the visible time window instead of fixing one 500-frame sample across the entire database
- route wheel/trackpad input through the timeline viewport and make horizontal swipes pan time in the same direction as the gesture
- apply app filters before bounded sampling and fence in-flight viewport reads when search takes ownership of the results
- keep dates only as labels and jump targets; day navigation centers a window on the chosen day without replacing the timeline source

## Root cause

The captured-history survey introduced the list of retained days, but `viewModel.screenshots` remained owned by `loadScreenshotsForDate(selectedDate)`. The first revision repaired that ownership by loading one fixed all-time sample, but that was still the wrong interaction boundary: zoom changed only the scale of the same sparse rows and could not progressively reveal history. Rewind now owns exact global bounds separately from a replaceable viewport sample, and every debounced zoom or pan queries that continuous visible time range with overscan.

Swipe handling lived in a page-wide local event monitor. It added the horizontal and vertical axes together and applied the signed AppKit delta directly to an older-to-newer frame index, so a horizontal gesture moved opposite the finger, slightly diagonal gestures could cancel themselves, and scrolling stepped frames instead of moving continuously through time. The track now pans the time window by the dominant gesture axis. Explicit frame navigation reveals the playhead, but a stationary playhead cannot pull a freely panned viewport back.

## Verification

- `xcrun swift test --package-path Desktop --filter 'RewindTrackTests|RewindAllTimeSearchTests'` — 32 passed
- `xcrun swift test --package-path Desktop --filter Rewind` — 192 passed
- `agent-logic-harness --cross-surface-smoke` — passed
- built and launched `/Applications/omi-rewind-history.app`; the isolated profile still contains 336 captures on three non-adjacent days (Aug 4, Aug 5, Aug 7). A fresh UI pass was blocked before the shell by macOS asking for the login-keychain password after the named app was re-signed, so no post-revision visual claim is made.

The swipe expectations follow AppKit's signed horizontal delta contract: [NSEvent.deltaX](https://developer.apple.com/documentation/appkit/nsevent/deltax). The updated E2E expectation is also grounded in the live three-day measurement above.

## Product invariants

- INV-AUTH-1 — cited because the desktop view-model path is covered by the invariant's path rule; this PR does not change authentication or session ownership.

## Failure class

Failure-Class: none

This is a local Rewind source-ownership and gesture-sign defect; no registered cross-system failure class applies. Regression coverage now exercises the production database/view-model viewport boundary, progressive zoom bounds, and continuous gesture navigation policy.

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-rewind.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-rewind.json
@@ -1,7 +1,7 @@
 {
   "files": {
     "desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift": 1734,
-    "desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift": 3428
+    "desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift": 3423
   },
   "raise_justifications": {
     "desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift": "The bounded incomplete-task storage surface reads all dated rows and pages only No Deadline rows, so TasksStore can preserve bucket boundaries without loading the entire undated local universe; it remains beside the existing task query and persistence owner rather than duplicating SQLite semantics. Single-snapshot dated/no-deadline reads prevent due-date races from omitting tasks between bucket queries.",
--- a/desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift
@@ -2940,47 +2940,42 @@ actor RewindDatabase {
   }
 
   /// Get screenshots sampled evenly across a date range, ordered ASC (oldest first).
-  /// If the total count for the range is <= targetCount, returns all rows ASC.
-  /// Otherwise picks every Nth screenshot to fit ~targetCount frames.
-  func getScreenshotsSampled(from startDate: Date, to endDate: Date, targetCount: Int) throws -> [Screenshot] {
+  /// Returns all rows within `targetCount`; otherwise samples evenly while retaining both endpoints.
+  func getScreenshotsSampled(
+    from startDate: Date, to endDate: Date, targetCount: Int, appFilter: String? = nil
+  ) throws -> [Screenshot] {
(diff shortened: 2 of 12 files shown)


Git-on-my-level (changes requested):
Thanks for pushing this forward — making Rewind's timeline continuous across retained history is the right direction, and the added viewport/window tests cover the core panning and endpoint behavior well.

I found one blocker before this is safe to merge:

- `loadTimelineWindow(...)` only checks `activeSearchQuery == nil` before awaiting the database query. If the user starts a search while a debounced pan/zoom load is still in flight, the stale timeline load can still pass the `timelineLoadID` check and assign `screenshots` after the search has started. That can replace filtered search results with the unfiltered timeline, and if the search fails/returns slowly the UI can remain on the wrong source. Please re-check that search mode is still inactive after the await and before assigning `visibleTimelineRange`/`screenshots` (or invalidate timeline loads when search starts), and add a regression test for the race.

I also recommend adding coverage for the app-filtered timeline path: `loadTimelineWindow` samples the whole visible window and only then applies `selectedApp`, so sparse apps can be underrepresented or disappear from a large mixed-app window. If that behavior is intentional for the initial version, please document the limitation; otherwise the app predicate needs to be applied before sampling.

Because this changes how retained screen history is queried and displayed, I’m leaving it for human maintainer review after the race is fixed.

---

kodjima33 (approved):
Rewind now owns global bounds separately from a replaceable viewport sample so history is reachable; needs a rebase on top of #11223 (both touch RewindTrack) before it can land.

Archit-lal:
Addressed the requested blockers in 6dad89d22c: the viewport loader now re-checks search ownership after its awaited query, with a deterministic in-flight race regression test; app filtering now occurs in the database before bounded sampling, with sparse-app coverage. The branch is also rebased onto main/#11223, and horizontal swipe plus vertical-wheel directions are separately tested. Verification: 32 focused tests, 192 Rewind tests, and all 20 push-gate contracts pass. @Git-on-my-level need human response for re-review so this can merge.

PR #11220, opened 2026-08-08
fix(backend): stop a group binding resolving to a member's personal account

undivisible:
Ports the channel fixes from #11058 onto this branch, where the feature actually lands.

Both PRs descend from `3fb99bc77b`, and they drifted into complementary halves: this branch has the Twilio signature validation, vision-context framing and caller-id work; #11058 has the fixes below. This branch is the one that ships the feature, so the fixes belong here.

### Group binding could resolve to a member's personal account

`get_binding` looped over `[channel_chat_id, channel_user_id]`, and the document id hashes only the identity. An unlinked **group** therefore fell through to the second probe and resolved **the sender's personal DM binding** — the router then answered from that person's private Omi account, with their private context, into the group chat. `revoke_binding` routes through the same function, so a group `/unlink` could revoke the sender's personal binding too.

A lookup now resolves exactly one identity and rejects a document whose stored chat id is not the one asked for. DMs are unaffected: there chat id and user id are equal, so the resolved document is unchanged.

### Also

- **Webhook claims were permanent tombstones.** Any failure before the handler wrote `ready` left `status: processing` forever, and every provider retry got `duplicate` — message lost, no way back. The claim is now a 5-minute lease a stale attempt can reclaim. **This changes durable semantics**: events already stuck in `processing` carry no `lease_expires_at`, which reads as expired, so they become reclaimable on the next delivery. That recovers lost messages but is a live-data change, not additive.
- **`revoke_channel` exceeded Firestore's 500-write batch cap**, so revoking failed outright past 500 bindings. Now chunked.
- **Media ran before the quota gate**, so over-quota users still triggered attachment downloads and vision/STT spend. Quota is the first gate now.
- **Twilio SMS ran on `critical_executor`**, which `backend/AGENTS.md` reserves for auth gates at 8 workers. Moved to `postprocess_executor`.

### Verification

`BACKEND_UNIT_TEST_FILE_LIST=... bash test.sh` on `tests/unit/test_channels.py` — **27 passed** against this branch's code. Each fix was proven to fail without its change on the source branch: the group lookup returns the personal binding instead of `None`, the batch raises `maximum 500 writes allowed per request`, and the router returns `{'status': 'duplicate'}` where it should return `{'status': 'delivered'}`.

- INV-MEM-1

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **High Risk**
> Group binding could leak private Omi context into group chats (memory isolation), and webhook reclaim changes behavior for in-flight `processing` events in production Firestore.
> 
> **Overview**
> Fixes a **privacy bug** where unlinked group chats could resolve to a member’s personal DM binding (and group `/unlink` could revoke that personal link). `get_binding` now looks up a single chat identity and rejects documents whose stored `channel_chat_id` does not match.
> 
> **Webhook idempotency** no longer permanently drops messages after a failed handler: claims get a **5-minute processing lease**, stale `processing` / `failed` events can be reclaimed transactionally, and `_handle_payload` calls `release_webhook_claim` on failure so provider retries can run again. Events already stuck in `processing` without a lease are treated as reclaimable on the next delivery.
> 
> Also chunks `revoke_channel` past Firestore’s **500-write batch limit**, runs **chat quota before** attachment/vision work, and sends Twilio SMS via **`postprocess_executor`** instead of `critical_executor`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d76a3fa20dc8a70e3dbc457eb084d2fadfa1aef. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/database/channels.py
+++ b/backend/database/channels.py
@@ -297,13 +297,29 @@ def revoke_binding(
 
 
 def _webhook_claim_is_stale(event: Dict[str, Any], now: datetime) -> bool:
+    """Whether a claim can be taken from whoever holds it.
+
+    A claim written before leases existed has no `lease_expires_at`. Reading
+    that as "expired" would make every such event reclaimable the instant this
+    deploys, including one a handler picked up seconds earlier — two handlers
+    on the same event means the user gets the reply twice. Its `received_at`
+    dates it just as well, so the lease is derived from that instead, and only
+    a genuinely old claim is reclaimed.
+    """
     status = event.get('status')
     if status not in WEBHOOK_RECLAIMABLE_STATUSES:
         return False
     if status == 'failed':
         return True
     lease_expires_at = event.get('lease_expires_at')
-    return not isinstance(lease_expires_at, datetime) or lease_expires_at <= now
+    if not isinstance(lease_expires_at, datetime):
+        received_at = event.get('received_at')
+        if not isinstance(received_at, datetime):
+            # Neither field is usable, so nothing can date this claim. Leaving
+            # it held forever is the failure this lease exists to end.
+            return True
+        lease_expires_at = received_at + WEBHOOK_PROCESSING_LEASE
+    return lease_expires_at <= now
 
 
 def claim_webhook_event(
--- a/backend/database/channels.py
+++ b/backend/database/channels.py
@@ -14,6 +14,9 @@ from database.firestore_index_registry import CHANNEL_BINDINGS_BY_UID_CHANNEL_QU
 CHANNELS = frozenset({'telegram', 'imessage', 'sms'})
 LINK_TOKEN_TTL = timedelta(minutes=15)
 LINK_TOKEN_PATTERN = re.compile(r'^[0-9a-f]{48}$')
+FIRESTORE_MAX_BATCH_WRITES = 500
+WEBHOOK_PROCESSING_LEASE = timedelta(minutes=5)
+WEBHOOK_RECLAIMABLE_STATUSES = frozenset({'processing', 'failed'})
 
 
 class ChannelLinkError(ValueError):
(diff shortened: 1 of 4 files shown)


Git-on-my-level (changes requested):
Thanks for porting this over — the group-binding fix itself is important, and the new tests cover the original privacy failure well.

I think the webhook-lease part still needs one concurrency guard before this lands. `claim_webhook_event()` increments `attempts`, but `_handle_payload()` never gets an attempt/claim token back, and `release_webhook_claim()` / `update_webhook_event()` only check the event id and current status. That means an older handler can still settle a claim after its lease has expired and a retry has reclaimed the same event:

1. attempt 1 claims the event and runs longer than the 5-minute lease;
2. provider retry reclaims it as attempt 2 and starts processing;
3. attempt 1 later fails and calls `release_webhook_claim()`, which sees `status == processing` and marks attempt 2 as `failed`; or attempt 1 later succeeds and writes/sends a reply without proving it still owns the claim.

For this idempotency path, please fence release and ready/delivered settlement on the active claim owner — for example, return the attempt number or a random claim token from `claim_webhook_event()` and require `release_webhook_claim()` / `update_webhook_event()` to compare it transactionally before changing the document. A stale handler should be unable to fail, ready, deliver, or overwrite a newer reclaimed attempt.

Given this touches channel auth/privacy behavior and webhook retry semantics, I’m leaving this as changes requested for the claim-fencing issue rather than approving automatically.

---

kodjima33 (approved):
Group bindings no longer fall through to the sender's personal DM binding (private context leaking into group chats); lookup now resolves exactly one identity, with unit coverage. Lands on its feature base branch.

PR #11222, opened 2026-08-08
Show task provenance in the Home timeline

Archit-lal:
## Summary

- add Tasks to the existing macOS Home timeline and filter chips
- attach conversation-derived tasks beneath their source conversation, while keeping manual tasks as standalone timeline entries
- show conversation provenance on task and memory rows and route source clicks through the existing Conversations detail view and full transcript drawer
- preserve exact transcript segment IDs and time bounds from structured task extraction, validate them against the source transcript, and focus/highlight those segments in the macOS transcript drawer
- keep legacy provenance compatible by opening the full source transcript at the top when segment-level evidence is unavailable

## User-facing behavior

Home now makes the chain `conversation -> memory/task -> transcript evidence` visible in the normal timeline. Users can filter to Tasks, complete a task inline, open the source conversation, and inspect the complete transcript. Newly extracted conversation tasks jump to and highlight their supporting transcript segments.

## Verification

- built and launched `/Applications/omi-task-provenance.app` (`com.omi.omi-task-provenance`)
- exercised the real Home UI with authenticated data: All/Tasks filtering, attached and standalone task rows, source labels, source navigation, Conversation Details, and the full transcript drawer
- `./scripts/dev-feedback.py --once swift 'SpineCompositionTests|QueryShellTests|ConversationDetailAutomationStateTests'` (75 passed)
- focused backend provenance/capture/schema tests (21 passed)
- app-client OpenAPI export and generated Swift API checks
- `./scripts/agent-logic-harness.sh --cross-surface-smoke` (passed)
- `make preflight` (31 checks passed)

## Product invariants

- `INV-NAV-1`: source navigation reuses the established Conversations destination and its feature-complete detail/transcript owner.
- `INV-UI-1`: the Home and transcript additions use existing neutral design tokens; no purple increase.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11222?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks @Archit-lal — this is a genuinely nice piece of work. Making conversation → task → transcript evidence visible inline in Home, with segment-level focus/highlight, is a cohesive and well-motivated feature, and the desktop tests (SpineCompositionTests, QueryShellTests, ConversationDetailAutomationStateTests, new provenance tests) are solid. A few things to address before this lands:

**1. Backend unit suite is failing on this head.** `tests/unit/test_candidate_lifecycle.py::test_task_mutation_preserves_origin_and_merges_provenance` now fails because `EvidenceRef` always serializes `transcript_segment_ids`, `start_seconds`, and `end_seconds` (even when empty/null), so the test's bare-dict provenance equality no longer holds. The schema change is fine; the pre-existing test stub needs updating to the new serialized shape.

**2. Non-finite transcript bounds.** In `backend/models/action_item.py`, `start_seconds`/`end_seconds` use `Field(ge=0)`, which accepts `float('inf')`. `inf` is not valid JSON and cannot focus a finite transcript position, so it can be persisted/emitted as broken provenance. Please reject non-finite values (e.g. `allow_inf_nan=False` or an explicit finite check) before save/emit.

**3. Desktop provenance correctness (the cubic review flags look valid on inspection):**
- Only the first matching evidence span is highlighted when a task has merged provenance — collect segment IDs from every matching conversation ref before opening the transcript.
- Prior `focusedTranscriptSegmentIds` can survive a conversation switch in `syncPresentedDetail` when there's no matching pending evidence — clear focus on switch so stale IDs can't affect the new drawer.
- The new SpineStore task digest omits the very provenance/source fields this PR adds, so an in-place server re-extraction (task stays incomplete, provenance changes) would show stale provenance/label until some other task field changes. Consider hashing `sourceSegmentIDs`/`provenance` and `source` into the digest.
- Tasks linked to an unloaded source conversation show a provenance button labeled "Task"; a "From conversation" fallback when only `conversationId` is known would preserve the source relationship.

Labeling `needs-maintainer-review` for the desktop UX calls (timeline settled/undercount semantics, label/fallback wording) and `needs-tests` for the failing backend suite. No security, privacy, auth, workflow, or dependency concerns on my review. Looking forward to the next revision.



---

PR #11238, opened 2026-08-08
fix(app): validate phone numbers with libphonenumber, not a fixed digit count

### LINKED ISSUE #11209: Signup blocked for valid Estonian phone numbers (+372)
## Summary
Valid Estonian phone numbers are rejected on the phone signup screen, leaving the submit/continue button disabled and blocking signup.

## Environment
- Product: Omi app
- Screen: phone signup with country code dropdown
- Country code: `+372` (Estonia)

## Reproduction
1. Open the phone signup screen.
2. Select Estonia in the country code dropdown (`+372`).
3. Enter `5142537`.
4. Observe that the Continue/Submit button stays disabled.

## Also observed
- Entering `05142537` also keeps the button disabled.
- Manually prefixing the number with `+372` in the input field produces a duplicated/garbled number because the dropdown already prepends the country code.

## Expected
The field should accept valid Estonian mobile numbers and enable the Continue/Submit button accordingly.

## Actual
Validation appears to enforce an incorrect fixed length / US-centric format, which blocks valid Estonia signup entirely.

## Impact
Users with valid Estonian numbers cannot complete signup.

mdmohsin7:
Fixes #11209.

## Why Estonian signup was impossible

The Continue button was gated on a digit count:

```dart
digits.length >= country.telephoneMinLength &&
digits.length <= country.telephoneMaxLength
```

`intl_country_data` declared Estonia `10/10`. Real Estonian numbers are 7–8 digits, so **no input could ever enable the button** — the reported `5142537` is 7 and `05142537` is 8.

## Why not just fix Estonia

**219 of the package's 241 countries have `min == max`** — one fixed length each, which cannot express a country whose numbers vary. The fork's own history already carries two one-off fixes of exactly this shape (Luxembourg/Finland, NZ landlines). Country-by-country patching does not converge; the next report is already latent.

Validation moves to [`phone_numbers_parser`](https://pub.dev/packages/phone_numbers_parser) — a pure-Dart port of Google's libphonenumber metadata, single dependency (`meta`). Pure Dart is the deciding factor: `flutter_libphonenumber` and `libphonenumber_plugin` are platform channels and cannot run under `flutter test`, which this repo requires. `intl_country_data` stays for what it's good at — the picker's names, flags and dial codes.

## The second reported bug

`_fullNumber` concatenated `'+${dialCode}$digits'` over every digit in the field, so typing `+372 5142537` produced **`+3723725142537`**. An explicit `+` or `00` is now parsed as international and wins over the picker, so pasting a foreign number keeps that number rather than being silently re-homed.

## Upstream

Estonia is corrected at source too — [mdmohsin7/intl_country_data#2](https://github.com/mdmohsin7/intl_country_data/pull/2), re-pinned here at `bb5355c` — since the picker still reads that package and `10/10` was plainly wrong.

**Only Estonia.** I attempted a blanket regeneration of all 241 entries from libphonenumber and reverted it: for NANP entries the two sources count different things. Jamaica's `telephoneCode` is `"1876"`, so its length field means digits *after* the area code (7), while libphonenumber's NSN includes it (10). The regeneration set 26 such countries to 10/10 and broke the fork's tests. Widening also broke its length-based country disambiguation. Left alone — the app no longer gates on those fields.

## One correction to the report

`05142537` stays invalid, and that's right. Estonia abolished trunk prefixes, so the leading `0` isn't a prefix to strip and that string isn't a real number. libphonenumber strips a leading `0` only where a country actually has a national prefix — the GB case in the tests covers exactly that (`02071838750` → `+442071838750`). The reporter's main complaint was correct; this sub-case was a wrong expectation.

## Verification

```
flutter test test/unit/phone_number_input_test.dart -> 19 tests, all passed
  covers 5142537, the +372-typed duplicate, 00-prefixed, an explicit + winning
  over the picker, GB with/without trunk 0, US, DE, and the invalid cases

flutter analyze (helper, page, test) -> No issues found
./test.sh                            -> 1153 tests, all passed
upstream: dart test -> 20 passed | dart analyze -> no issues
```

**Not verified on a device.** The parser is covered by unit tests, but I did not build the app and walk the phone-setup screen. Worth doing before release.

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/phone_calls/phone_setup_number_page.dart
+++ b/app/lib/pages/phone_calls/phone_setup_number_page.dart
@@ -168,7 +168,7 @@ class _PhoneSetupNumberPageState extends State<PhoneSetupNumberPage> {
                           hintText: context.l10n.phoneNumberHint,
                           hintStyle: TextStyle(color: Colors.grey[600]),
                         ),
-                        inputFormatters: [FilteringTextInputFormatter.allow(RegExp(r'[0-9\s\-\(\)]'))],
+                        inputFormatters: phoneFieldInputFormatters,
                         onChanged: (_) => setState(() => _errorMessage = null),
                       ),
                     ),
--- a/app/lib/pages/phone_calls/phone_setup_number_page.dart
+++ b/app/lib/pages/phone_calls/phone_setup_number_page.dart
@@ -6,6 +6,7 @@ import 'package:provider/provider.dart';
 import 'package:omi/pages/phone_calls/phone_setup_verify_page.dart';
 import 'package:omi/providers/phone_call_provider.dart';
 import 'package:omi/utils/l10n_extensions.dart';
+import 'package:omi/utils/phone_number_input.dart';
 
 class PhoneSetupNumberPage extends StatefulWidget {
   const PhoneSetupNumberPage({super.key});
@@ -36,15 +37,14 @@ class _PhoneSetupNumberPageState extends State<PhoneSetupNumberPage> {
     super.dispose();
   }
 
-  bool get _isValid {
-    var digits = _phoneController.text.replaceAll(RegExp(r'\D'), '');
-    return digits.length >= _selectedCountry.telephoneMinLength && digits.length <= _selectedCountry.telephoneMaxLength;
-  }
+  PhoneNumberInput get _parsed => parsePhoneNumberInput(
+        raw: _phoneController.text,
+        isoCode: _selectedCountry.codeAlpha2,
+      );
 
-  String get _fullNumber {
-    var digits = _phoneController.text.replaceAll(RegExp(r'\D'), '');
-    return '+${_selectedCountry.telephoneCode}$digits';
-  }
+  bool get _isValid => _parsed.isValid;
+
+  String get _fullNumber => _parsed.e164;
 
   void _showCountryPicker() {
     showModalBottomSheet(
--- a/app/lib/utils/phone_number_input.dart
+++ b/app/lib/utils/phone_number_input.dart
@@ -1,4 +1,15 @@
+import 'package:flutter/services.dart';
 import 'package:phone_numbers_parser/phone_numbers_parser.dart';
+
+/// Characters the phone field accepts.
+///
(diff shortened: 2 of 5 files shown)


mdmohsin7:
Valid, and a good catch — fixed in a0c300f0a1.

The field's formatter was `FilteringTextInputFormatter.allow(RegExp(r'[0-9\s\-\(\)]'))`, so `+` was stripped before the controller ever saw it. That made the explicit-international branch I added **dead code in the UI**: a UK number pasted with Estonia selected arrived as `442071838750` and got parsed as an Estonian national number. My tests passed because they call `parsePhoneNumberInput` directly and never went through the widget — the gap was exactly where the review pointed.

Fixes:

- The formatter allows `+` and now lives beside the parser as `phoneFieldInputFormatters`, so the accepted character set and the code that depends on it can't drift apart again.
- The parser keeps only a *leading* `+` and drops any other, so a stray `+` mid-string is ignored rather than making the whole input unparseable.
- Coverage runs through the same formatter object the widget uses, plus a widget test that types into a real `TextField` and asserts a pasted `+442071838750` survives and still resolves to `+442071838750` with Estonia selected.

```
flutter test test/unit/phone_number_input_test.dart -> 26 passed (was 19)
flutter analyze (helper, page, test)                -> No issues found
./test.sh                                           -> 1160 tests, all passed
```

kodjima33 (approved):
Replaces the fixed digit-count phone gate with libphonenumber metadata (pure-Dart phone_numbers_parser) so variable-length countries like Estonia can sign up; fixes #11209 with unit tests.

PR #11251, opened 2026-08-08
fix(macos): isolate signed release canary startup

Git-on-my-level:
## Summary

- keep signed release canaries ahead of font/resource startup and the single-instance guard
- make the UserNotifications smoke-request predicate explicit and testable
- add regression coverage for the launch-order contract

## Failure classification

Failure-Class: FC-usernotifications-callback-isolation

The v0.12.152 signed stable artifact passed build, signing, dependency audit,
notarization, DMG creation, and Beta packaging, then failed the isolated stable
smoke step. The redesign introduced font registration before the canary early
return. This change restores callback-only canary startup without weakening any
artifact, signing, Keychain, or UserNotifications gate.

- INV-AUTH-1

## Verification

- `xcrun swift-format lint --strict` on all changed Swift files
- `xcrun swift test -c debug --package-path desktop/macos/Desktop --filter AuthStorageCanaryTests`
- `xcrun swift test -c debug --package-path desktop/macos/Desktop --filter UserNotificationCallbackBridgeTests`

## Risk

Low. Normal launches retain font registration and single-instance enforcement.
Only explicit distribution-canary launches return before those product startup
side effects.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11251?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/OmiApp.swift
+++ b/desktop/macos/Desktop/Sources/OmiApp.swift
@@ -274,8 +274,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSMenuDelegate, @unchecked S
     // Publish the live delegate instance for callers that can't rely on
     // `NSApp.delegate as? AppDelegate` (nil under SwiftUI's delegate adaptor).
     AppDelegate.shared = self
+    if AuthStorageCanary.isRequested || UserNotificationCallbackBridge.isSignedSmokeRequested() { return }
     OmiFontRegistration.registerAll()
-    if AuthStorageCanary.isRequested { return }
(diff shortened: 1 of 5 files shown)


kodjima33 (approved):
Restores callback-only signed-canary startup ordering ahead of font registration; regression coverage added.

PR #11265, opened 2026-08-08
feat(app): UI redesign — Convos-first nav, Chat tab, Brain tab, consolidated Todos

jacobtrebil:
## What changed and why

A UI redesign pass over the Flutter app. The central change is that **chat
becomes a first-class destination** rather than a route pushed from Home's "Ask
Omi" bar — that push was residue from when Chat was its own nav tab (the tap
still fired `bottomNavigationTabClicked('Chat')`).

- **Nav is now Convos / Chat / Brain / Todos / Apps.** Conversations lead; the
  old Home tab is renamed **Chat** and holds the conversation with Omi.
- **Brain tab** — memories and the knowledge graph move out of the settings
  drawer into a nav tab, and the Mind Map card comes off Home where it rendered
  unlabeled dots across a third of the screen.
- **Todos** collapse from five headings (Today, Tomorrow, Later, No deadline,
  Overdue) to one list plus Completed. Due dates already appear on the rows.
- **Typography** moves to Space Grotesk with a shared ramp in
  `lib/utils/app_typography.dart`.
- **Neutral accents** — the completed-task check and several controls drop their
  colour, continuing the `INV-UI-1` direction.
- Daily recaps carousel, starter prompts under the composer, task-completion
  feedback, and bottom-bar spacing work.

- **INV-MEM-1** — memories move to a dedicated Brain tab. Storage, promotion and
  tiering are untouched; this is presentation only.

`INV-UI-1` is not cited: per its own PR rule, routine UI work relies on the brand
ratchet. This branch *reduces* purple (Record button, task FAB, completed check).

## How it was verified

Exercised on the iOS simulator (iPhone 17 Pro, iOS 26.5) — every tab, the chat
empty state and composer focus behaviour, task completion, and the recaps
carousel. Spacing changes were measured against reference screenshots rather
than eyeballed.

**Three limits worth stating plainly:**

1. **Not verified against real data.** A local build cannot authenticate:
   `app/setup.sh` copies `based-hardware-dev` Firebase config while setting
   `API_BASE_URL=https://api.omiapi.com/`, so `signInWithCustomToken` fails with
   `custom-token-mismatch`. Every screen here was exercised against fabricated
   local data. That looks like a bug in the documented setup path rather than
   anything specific to this branch, and deserves its own issue.
2. **Not verified on a physical device.** Device signing is blocked on an Apple
   account permission, not on anything in the project. BLE, the widget and the
   watch app are therefore unexercised — none are touched here, but the
   app-group entitlement path is unproven.
3. **Tab indices are not compiler-checked.** Tab position lives as bare integer
   literals across several files (page list, scroll-to-top, merge action bar,
   sync/calendar buttons, app-bar chrome gate, `BatteryInfoWidget`). The reorder
   updated all known sites and each was checked on screen, but a future reorder
   has no guard. Converting these to a named enum is the obvious follow-up.

## Tests

`flutter test test/widgets/` — **148 pass, 0 fail**, 27s.

Run in a fresh checkout of this branch with no local index state, after
`flutter pub get`, `dart run build_runner build` and the `setup.sh`-generated
Firebase config — i.e. what a maintainer gets, not my working tree. Same
checkout: `flutter analyze lib/ test/` reports 0 errors.

New coverage lands with the behaviour it asserts: `action_item_completion_test`,
`action_item_completed_view_test`, `action_items_page_scroll_test`, and
`home_hero_entrance_test`.

Two problems the first push had, both fixed here and both invisible from my
working tree, since the files carrying them were marked `skip-worktree`:

- `users.dart` imported `package:omi/dev_demo_data.dart`, a local-only file
  excluded from git, so no clean clone could build. Reported in review.
- Three action-item tests called `ActionItemsProvider.seedDemoData`, which also
  existed only locally, so the suite did not compile either. They now use the
  constructor injection this provider already had, plus a fake update request in
  `test/widgets/support/`.

An earlier `home_hero_pixel_test` has been removed. It awaited `toByteData()`
outside `tester.runAsync`, which never completes under fake async — the runner
wedged and only failed at the 10-minute default timeout. It was a debugging
probe, not a guard.

## Failure class (fixes)

Failure-Class: none

The one `fix:` commit (`995cd71`) corrects a build-configuration typo: the dev
BatteryWidget derived `$(APP_BUNDLE_IDENTIFIER).development.widget` while
`APP_BUNDLE_IDENTIFIER` already ended in `.development`, producing
`…ios12.development.development.widget`. That identifier does not exist on the
Apple team, so any dev-flavour device build failed at signing. The prod widget
and the watch app already used the correct single-suffix pattern; this brings the
three dev configs in line. No existing class covers a per-flavour identifier
derivation typo, and one instance does not warrant a new one.

## Scoped cleanups

- Removed `ChatPage`'s second `BottomNavBar`, which rendered inside the page and
  appeared/disappeared on focus — the cause of the composer shifting ~100pt.
- The reply-reserve spacer now sizes off the message list's viewport rather than
  the whole screen; as a tab body the list is far shorter, so the reserve pushed
  the newest message off the top.
- `.gitignore` exception for `app/assets/sounds/*.wav`. The repo-wide `*.wav`
  rule hid the completion chime while `pubspec.yaml` declared it, so every other
  clone would have built with the asset missing.

## Known gaps

Embedding `ChatPage` removed its app bar, which was the only caller of
`openEndDrawer()` — **Clear chat and the chat-app selector are unreachable from
the Chat tab.** They need a new entry point (the composer's `+` menu is the
natural place) before this is user-ready.


Git-on-my-level (dismissed):
Thanks for pushing this redesign work — the direction is cohesive around the Flutter app shell/Home/chat/Todos experience, and I like that you included widget coverage for the new home hero and task-completion behavior.

I need to request changes before this can be reviewable/mergeable because the current head appears to leave local demo scaffolding wired into production code:

- `app/lib/backend/http/api/users.dart` imports `package:omi/dev_demo_data.dart` and calls `kDemoDataEnabled` / `demoDailySummaries()` inside `getDailySummaries()`, but `app/lib/dev_demo_data.dart` is not present in the PR or repository at this head. A fresh checkout therefore cannot resolve this import and the Flutter app will fail analysis/build before runtime.
- `app/lib/providers/goals_provider.dart` and `app/lib/providers/message_provider.dart` add `seedDemoData(...)` APIs and comments that also point at `lib/dev_demo_data.dart`. If these hooks are only for local prototype data, they should be removed before merging; if they are intended as a real fixture/demo mode, please add the missing implementation behind an explicit non-production/test boundary.
- `app/lib/pages/conversations/widgets/daily_recaps_carousel.dart` now calls `getDailySummaries()` directly, so the unresolved demo-data branch in `users.dart` affects the new carousel path too, not just an unused helper.
- `app/pubspec.yaml` and `.gitignore` correctly add `assets/sounds/` / allow committed WAV assets for `app/lib/utils/audio/ui_sounds.dart`, but I could not complete a local Flutter analyzer run in this environment because Flutter/Dart are not installed here; the unresolved import above is from static inspection of the checked-out PR head.
- `app/lib/pages/home/page.dart`, `app/lib/pages/chat/page.dart`, `app/lib/widgets/bottom_nav_bar.dart`, and `app/lib/pages/action_items/action_items_page.dart` make broad navigation/product-surface changes. Once the build issue is fixed, this still needs maintainer product/UX sign-off because Home becoming the embedded chat, adding Brain, and moving goals/Todos are substantial app-direction choices.

Suggested next step: remove the `dev_demo_data` import/branch and any local-only seed hooks that are not meant to ship, or include a properly scoped test/demo fixture that is available in clean clones and cannot affect production users. After that, please rerun Flutter analysis/tests for the app/widget suite so maintainers can review the UX changes on a buildable head.

Leaving this for human maintainer review as a product-direction decision for the app navigation/Home redesign after the build blocker is fixed.

— Reviewed by gpt-5.5 for the Omi maintainer automation.

---

jacobtrebil:
Fixed in `f9c5c39` and `b099e01`.

The `dev_demo_data` import is gone from `users.dart`. You were right that it broke clean clones — the file is local-only and excluded from git, and the `skip-worktree` flag on the files around it meant this was invisible from my working tree.

There was a second instance you couldn't have seen without running Flutter: three action-item tests called `ActionItemsProvider.seedDemoData`, which also existed only locally, so the test suite didn't compile either. They now use the constructor injection this provider already had for `getActionItems`/`deleteActionItem`, with a fake update request in `test/widgets/support/`. The `_demoMode` flag those hooks relied on is removed — it suppressed the server write in production code, which was the more serious half of the problem.

Also removed `home_hero_pixel_test`, which awaited `toByteData()` outside `tester.runAsync` and wedged the runner until the 10-minute timeout.

Verified on a fresh checkout with no local index state: `flutter analyze lib/ test/` 0 errors, `flutter test test/widgets/` 148 pass / 0 fail. Note that no CI has run on this branch yet — the checks are still pending workflow approval — so those numbers are from my machine, not from Omi's CI.

The nav and Home/Chat changes are still product-direction calls and need a maintainer's view.

undivisible:
yo can we get screenshots pls :))

Git-on-my-level (changes requested):
Thanks for the follow-up. I re-reviewed the current head and confirmed the previous build blocker around local demo scaffolding has been addressed: `app/lib/backend/http/api/users.dart` no longer imports `dev_demo_data`, and the prior `seedDemoData` references are gone from the providers.

I still need to request changes for a remaining navigation regression introduced by the tab reorder:

- `app/lib/pages/home/page.dart` now defines the tab order as Conversations(0), embedded Chat/Home(1), Brain(2), Tasks(3), Apps(4), and `app/lib/widgets/bottom_nav_bar.dart` renders the same five-tab order. However the deep-link initialization still maps `/action-items` to `homePageIdx = 2` at lines 307-310, and the later switch still says the action-items tab has already been selected. With the new order, index 2 is `BrainPage`, so notification/deep-link entry points for action items open the Brain tab instead of Todos. Please update that mapping to the new Tasks index and add/adjust a regression test for the route/initial-index mapping.

Other reviewed areas:

- `app/lib/pages/chat/page.dart` correctly avoids disposing the shared `HomeProvider.chatFieldFocusNode` when embedded, and the empty-state composer/prompt structure is coherent with Home-as-chat.
- `app/lib/pages/action_items/action_items_page.dart` now consolidates active tasks into one Todos bucket and delays provider completion until after the tap feedback; the provider seam in `app/lib/providers/action_items_provider.dart` is a reasonable way to keep widget tests hermetic.
- `app/lib/pages/conversations/widgets/daily_recaps_carousel.dart` uses the existing `getDailySummaries(limit: 7)` path directly and stays out of the page when there are no summaries; I did not see the prior missing local fixture issue on this head.
- `app/lib/pages/memories/brain_page.dart` embeds `MemoryGraphPage` with bottom-nav clearance, which matches the new tab-shell architecture.
- `app/pubspec.yaml`, `.gitignore`, `app/lib/gen/assets.gen.dart`, and `app/lib/gen/fonts.gen.dart` are consistent with shipping Space Grotesk and `assets/sounds/`; this is product/brand-sensitive but not a supply-chain concern from the diff I inspected.
- The new widget tests in `app/test/widgets/` cover the action-item completion and home-hero behavior, but I could not run them in this review environment because Flutter is not installed here, and GitHub still has no passing check summary for this draft.

This PR is still a substantial app navigation/Home redesign, so after the routing bug is fixed it should remain with a human maintainer for product/UX sign-off before merge.

---
*by AI on behalf of David; human maintainer review is needed for the app navigation and Home/Chat product direction.*

Reviewed by glm-5.2 for the Omi maintainer automation.

aryanorastar:
Closing as a maintainer.

Git-on-my-level requested changes on b099e01 (Aug 10): /action-items still sets homePageIdx = 2 after the tab reorder, so that deep link opens Brain instead of Todos. That is still the head. There was no follow-up.

The mapping is still wrong in app/lib/pages/home/page.dart (Conversations 0, Chat 1, Brain 2, Todos 3, Apps 4; case "action-items": homePageIdx = 2, and the later switch still says ActionItemsPage). memories/facts still land on index 0. The author already noted this is not user-ready: embedding ChatPage dropped openEndDrawer(), so Clear chat and the chat-app selector are unreachable.

This is still a draft, CONFLICTING against current main (gitignore, generated assets/fonts, l10n), with no repository checks. Main still uses Home / Conversations / Apps. Landing a 151-file Flutter shell rewrite from August would fight that, not compose with it.

A new PR on current main can take another run after a product yes, named tab indices with a route test, and a buildable clean clone — not this branch.

PR #11283, opened 2026-08-09
feat(llm): backend SSOT — managed text via OpenRouter Luna

undivisible:
## Summary
- Make `model_config` + gateway overrides the product SSOT for managed text routing, with `chat_agent` and all former OpenAI/Gemini-lite/`wrapped_analysis` features resolving to `gpt-5.6-luna` on OpenRouter (`openai/gpt-5.6-luna`).
- Add shared OpenRouter vendor-prefix helper (`openai/` / `google/`), gate `prompt_cache_retention` to direct OpenAI, and remap OpenAI-family BYOK on OpenRouter routes to the user's OpenAI key.
- Keep Perplexity for `web_search`; update cost cards, inventory, readiness, and QoS/gateway unit tests.

## Test plan
- [x] `BACKEND_UNIT_TEST_FILE_LIST=… bash test.sh` for QoS + gateway unit files (including fast-unit CPU guard)
- [ ] Smoke chat-agent / knowledge_graph / memory_l2 against OpenRouter with `OPENROUTER_API_KEY`
- [ ] Confirm OpenRouter Luna rate card vs live OpenRouter pricing before relying on COGS alerts
- [ ] Manual: mobile omit-model chat + desktop managed rewrite still hit `omi:auto:chat-agent`

## Failure class (fixes)

Failure-Class: none

The only `fix:` commit here restores main's Dart formatting on files this merge merely
carried across; it repairs no product contract.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11283?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks for the careful consolidation here — the SSOT direction is coherent, and I like that the OpenRouter model-name prefixing is being centralized instead of duplicated.

I found one blocking issue before this can merge:

- `backend/utils/llm/clients.py` now intentionally remaps OpenRouter-hosted OpenAI-family BYOK traffic to the user's OpenAI key in gateway mode (`provider=_effective_byok_provider(model, provider)`, so `openrouter` + `gpt-5.6-luna` becomes `openai`).
- `backend/utils/llm/gateway_byok.py` then forwards that key as `X-Omi-Byok-openai-Key`, which matches the stated product behavior for OpenAI-family BYOK remapping.
- But the gateway executor still checks the route provider literally: `backend/llm_gateway/gateway/executor.py` calls `credential_context.has_provider_key(provider_ref.provider)`, and the generated route provider is now `openrouter` for these lanes (`backend/llm_gateway/config/generated_route_overrides.yaml` / `backend/llm_gateway/config/route_artifacts.yaml`). With only the forwarded OpenAI BYOK key present, the gateway will see no `openrouter` key and fail closed as `missing_byok_key` before the provider call. `backend/llm_gateway/gateway/credentials.py` also has no effective-provider/remap layer for this check.
- The tests do not cover that production path: `backend/tests/unit/test_llm_gateway_client_config.py` asserts that the backend sends provider `openai` into the gateway BYOK helper, while `backend/tests/unit/test_llm_gateway_openai_compatible.py` switched the direct gateway API tests to `X-Omi-Byok-OpenRouter-Key`. Neither test exercises an OpenRouter route artifact receiving only an OpenAI forwarded BYOK key, so this regression can stay green.

Please add/fix the gateway-side BYOK effective-provider mapping (or otherwise make the route credential check accept the intended OpenAI key for OpenRouter OpenAI-family models) and add a focused gateway executor/openai-compatible test for an `openrouter` + `openai/gpt-5.6-luna` route with only `X-Omi-Byok-OpenAI-Key` present.

Other notes from the sampled files:

- `backend/utils/llm/openrouter_model_names.py` and `backend/utils/llm/providers.py` look like the right shared boundary for vendor-prefixing OpenRouter model IDs.
- `backend/llm_gateway/routers/health.py` correctly moves readiness to `OPENROUTER_API_KEY` when managed chat lanes are OpenRouter-backed.
- `backend/llm_gateway/gateway/config_loader.py` now uses the shared prefix helper, keeping generated route artifacts and runtime provider requests aligned.
- `backend/llm_gateway/gateway/executor.py` correctly preserves GPT-5.6 cache fields for vendor-prefixed `openai/gpt-5.6-luna`, but this is also where the BYOK provider-key check needs to stay consistent with the remap.
- `backend/tests/unit/test_openrouter_model_names.py` covers the new helper, but the missing test is the cross-boundary BYOK/gateway credential path above.

Because this is model/provider routing and BYOK behavior, this still needs maintainer sign-off after the blocking credential-path fix.

---
Automated maintainer review. I may be wrong; please verify the concern against the code and tests.

---

undivisible:
Confirmed and fixed in 79fca60286.

`_byok_vendor_provider_ref` in the executor now follows the forwarded key: a BYOK request on an `openrouter` route whose model carries a vendor prefix is served by that vendor's provider with the prefix dropped, when the caller supplied that vendor's key. Managed traffic is untouched. The vendor mapping lives in `openrouter_model_names.py` next to the existing prefixing helper, so both sides of the boundary read from one place.

Two tests in `test_llm_gateway_openai_compatible.py` cover the path you named — an `openrouter` + `openai/gpt-5.6-luna` route with only `X-Omi-Byok-OpenAI-Key` (calls `openai` with `gpt-5.6-luna`), and the fail-closed counterpart with an unrelated vendor key.

Also in this head: the merge with `main` keeps `reasoning_effort=none` from #11285 on the now-OpenRouter `chat_agent` route, and extends the GPT-5.6 sanitizer to the `openai/` namespace so both live-400 guards survive the repoint.

PR #11284, opened 2026-08-09
feat(llm): dynamic OpenRouter model catalog for context/max tokens

undivisible:
## Summary
- Add `OpenRouterModelCatalog` that fetches/caches OpenRouter `GET /api/v1/models` for `context_length`, `max_completion_tokens`, and `supported_parameters`.
- Clamp gateway provider requests for OpenRouter routes against live completion ceilings (rs_ai-inspired thin registry; OpenRouter is the numeric metadata source because rs_ai does not expose context windows).
- Stacked on #11283 (`backend-ssot-openrouter`).

## Test plan
- [x] `tests/unit/test_openrouter_model_catalog.py` via `test.sh`
- [x] `tests/unit/test_llm_gateway_executor.py` still green
- [ ] Optional live check: with `OPENROUTER_API_KEY`, confirm `openai/gpt-5.6-luna` resolves with non-null context/max completion

## Failure class (fixes)

Failure-Class: none

The `fix:` commits here come from the merged parent branch; this PR adds the dynamic
OpenRouter model catalog and its clamp.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11284?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for this — the dynamic catalog is a clean approach to keeping provider limits accurate without hardcoding them in route config.

Reviewed all 3 changed files:

**backend/utils/llm/openrouter_model_catalog.py** — Solid implementation. Thread-safe TTL cache with double-checked locking (lines 74-89), fail-open design that retains stale entries when refresh fails (line 85), and defensive parsing that correctly rejects bools from `_positive_int` (line 139). The `apply_openrouter_completion_clamp` helper correctly gates on `provider == 'openrouter'` (line 184) so non-OpenRouter routes pass through untouched.

**backend/llm_gateway/gateway/executor.py** — The clamp is applied as the final transform in `_provider_request` (line 438), after output budget. This is the right position — budget shaping happens first, then the live ceiling is enforced last. One note: the singleton reads `OPENROUTER_API_KEY` at fetch time (line 93) rather than construction, which is fine for a long-lived server process.

**backend/tests/unit/test_openrouter_model_catalog.py** — Good coverage across the core paths: limit loading, clamping to provider ceiling, non-OpenRouter passthrough, stale retention on refresh failure, and non-JSON rejection. The `autouse` reset fixture (line 17) correctly isolates each test from the module-level singleton.

The 5 new tests pass (verified locally). The Backend unit suite failure is present on the parent branch (#11283 `backend-ssot-openrouter`) as well — the same 3 test files fail there independently of this commit.

Leaving for human maintainer review: this is a stacked PR making an architectural decision about whether Omi should source LLM model limits dynamically from OpenRouter rather than hardcoding them, and it touches provider routing in the gateway executor.

---

Issue #11292, opened 2026-08-09
App search fetches and enriches the whole catalog on every query, and matches name only


mikejsmith1985:
Searching the Apps tab is slow no matter what I type — typing `s` costs the same as
typing `sports` — and some searches return nothing when they clearly shouldn't. Searching
`adhd` gives "No apps found".

Reading `/v2/apps/search` against `d60ab7a1c`, both look like the same root cause: **the
query never reaches the database.**

- `backend/database/apps.py:83` — `search_apps_db()` takes `category`, `capability`,
  `my_apps`, `installed_apps`. There is no query parameter.
- `backend/routers/apps.py:672` — so it is called without the text query, and returns the
  whole catalog for the given filters.
- `backend/routers/apps.py:693-694` — `get_apps_installs_count(app_ids)` and
  `get_apps_reviews(app_ids)` then run over **every app in that catalog**, followed by an
  `App(**app_dict)` construction per record.
- `backend/routers/apps.py:725` — only then is the text applied, in Python:
  `search_query in app.name.lower()`.

So the per-request cost is independent of the query, which matches the symptom: every
keystroke pays for the full catalog plus its install counts and reviews, then discards
almost all of it.

The second half is the empty results. That final filter matches **the name only**, while
the parameter is documented as `'Search query for app name or description'`
(`backend/routers/apps.py:650`). An app whose description is about ADHD but whose name
does not contain the string is unreachable by search. I do not know whether the docstring
or the behaviour is the intended one, which is part of why I am asking rather than
sending a patch.

Happy to work on this if you want it — I would rather agree the direction first, since
pushing the query down to Firestore is a real change to a hot path and the substring
match has no obvious index-friendly equivalent.

*Investigated with Claude Code; every file:line above was verified against `d60ab7a1c`.*

kodjima33:
Both halves confirmed against `d60ab7a1c` — thanks for the file:line trace, it was exact.

I've taken the **name-only match** half in #11294, since that one is unambiguous: the `q`
parameter documents "app name or description", and every client's offline fallback ranks over
name/description/category/author (`desktop/.../appRanking.ts` → `matchesQuery`). The remote
endpoint was strictly narrower than the fallback it replaces — an app findable with the network
down vanished once the endpoint answered. Fixed to name OR description, with a relevance tier
(exact name → name prefix → description-only) ahead of the installs sort, because results are
paginated and widening the predicate alone would let a 10k-install description match push the
app you typed the name of off page 1.

[...]

PR #11293, opened 2026-08-09
fix(app): keep a progress indicator running for the whole app search

mikejsmith1985:
## Problem

Searching the Apps tab shows five shimmer placeholders. `ShimmerWithTimeout` stops
animating after five seconds and falls back to static blocks
(`app/lib/widgets/shimmer_with_timeout.dart:57`), so any search slower than that leaves
five motionless grey rectangles and nothing else moving on screen.

At that point a working search and a wedged one look identical. This is the state users
report as the app having hung: on a slow search the screen settles into five unmoving grey
rectangles with no spinner, no bar, and no text.

Related: #11292 covers why those searches are slow in the first place. This PR is only
about the screen not looking dead while they run, and stands on its own.

## Change

Adds an indeterminate `LinearProgressIndicator` above the placeholders that does not time
out, so there is always one moving thing for as long as the request is in flight.

Indeterminate on purpose: a search has no known duration, and a determinate bar would have
to invent a percentage that then appears to stall — the same doubt in a different shape.

The loading state moves into `SearchLoadingSliver` so it can be pumped on its own in a
widget test. Reaching it through `ExploreInstallPage` would need a live backend, which
would put the regression test outside CI.

`ShimmerWithTimeout` itself is untouched. It has 45 call sites across 22 files, and its
timeout is reasonable where something else already signals liveness; the defect here is
that on this screen nothing else did.

No new user-facing strings, so no ARB changes.

## Verification

- `flutter test test/widgets/search_loading_sliver_test.dart` — 3/3 pass. The two
  progress-indicator assertions fail on the parent commit; that is the regression this fixes.
- `flutter test test/widgets` — 146 pass.
- `flutter test` — 1119 pass, 2 fail: `plans_sheet_l10n` and `voice_recorder_mic_contention`.
  Both reproduce with this work stashed on `d60ab7a1c`, so neither is from this change; the
  latter is a Windows temp-directory teardown lock.
- `dart analyze` on all three changed files — no issues.
- `dart format --line-length 120` — no changes.

**I could not run the app.** No macOS and no Android SDK on this machine, so this is not
verified on a device. The widget test asserts the behaviour, but that is not the same as
having seen it, and I would rather say so than imply otherwise.

## Product invariants

`scripts/pr-preflight --suggest` reports none affected. `brand-ui` (INV-UI-1) is selected
because the diff touches UI sources: the indicator is `AppStyles.textPrimary` on
`AppStyles.backgroundSecondary` — white on neutral, no purple.

## Failure class

Failure-Class: none

No existing class covers it — the registry's closest entries are scoped to specific
subsystems (`FC-media-track-output-liveness` is desktop Windows rewind capture). I have not
declared `new`, because a new class asks for a canonical prevention artifact and a reusable
guard surface, and I do not think one UI liveness affordance earns that yet. Happy to
classify it differently if you would rather it were tracked.

---

*Written with Claude Code. Every file:line above was verified against `d60ab7a1c`.*

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/apps/explore_install_page.dart
+++ b/app/lib/pages/apps/explore_install_page.dart
@@ -15,6 +15,7 @@ import 'package:omi/pages/apps/widgets/category_apps_page.dart';
 import 'package:omi/pages/apps/widgets/category_section.dart';
 import 'package:omi/pages/apps/widgets/filter_sheet.dart';
 import 'package:omi/pages/apps/widgets/popular_apps_section.dart';
+import 'package:omi/pages/apps/widgets/search_loading_sliver.dart';
 import 'package:omi/providers/app_provider.dart';
 import 'package:omi/providers/home_provider.dart';
 import 'package:omi/utils/app_localizations_helper.dart';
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Thanks for this — it's a genuinely well-scoped fix. I verified the premise against the source: `ShimmerWithTimeout` defaults to `timeoutSeconds: 5` (`app/lib/widgets/shimmer_with_timeout.dart:35`) and then renders its child statically, so a slow app search does leave five motionless grey rectangles with nothing else moving. The bug is real and your framing of it is accurate.

What I like:

- `search_loading_sliver.dart`: the indeterminate `LinearProgressIndicator` (no `value` set) is the right call — a search has no known duration, so a determinate bar would have to invent a percentage that then appears to stall, recreating the same doubt. Keeping it outside `ShimmerWithTimeout` so it never times out is exactly the point.
- The extraction is clean: `SliverMainAxisGroup` lets the bar and the shimmer list share one padded sliver, and the old inline `_buildShimmerListItem` is preserved 1:1 in `_ShimmerListItem`.
- `explore_install_page.dart`: the call site swap (`const SearchLoadingSliver()` replacing `_buildSearchLoadingSliver()`) is correct — it still only shows when `state.isSearching` is true.
- `test/widgets/search_loading_sliver_test.dart`: the three tests pin the precise regression — 5 placeholders, the `LinearProgressIndicator` still present after pumping past 10s, and `indicator.value` being null. Pumping the widget in isolation (rather than through `ExploreInstallPage`) is the right way to keep this regression test backend-free and in CI.

All `AppStyles` tokens used (`spacingL`, `radiusSmall`, `textPrimary`, `backgroundSecondary`, `backgroundTertiary`) resolve in `ui_guidelines.dart`. No new dependencies, no native/auth/data surfaces touched.

One optional, non-blocking nit: `placeholderCount` defaults to 5 and is honored, but it's only exercised at the default in the tests. Not worth blocking on for a loading placeholder.

Leaving a positive signal rather than a formal approval.

---

mikejsmith1985:
Same note as on #11297, since it applies here too: **no CI has been triggered on this PR.** Zero workflow runs — the only check reporting is cubic's review.

Not path filtering. `.github/workflows/backend-hermetic-e2e.yml` fires on every `pull_request` to `main` with no `paths:` filter, doing its scope detection inside the job, which is what makes `Backend Hermetic Merge Gate` report on every PR per AGENTS.md. It has not run on either of my PRs.

Most likely the first-time-contributor approval gate on fork pull requests. If so, a maintainer clicking **Approve and run workflows** unblocks it; I have no way to trigger it myself.

Local results on `d60ab7a1c` in the meantime: `flutter test test/widgets` 146 pass, full `flutter test` 1119 pass / 2 fail (`plans_sheet_l10n` and `voice_recorder_mic_contention`, both failing identically with this work stashed), `pr_preflight --lane local` 11/11. Still not run on a device — no macOS or Android SDK here — which is the one Definition of Done item I cannot satisfy.

kodjima33 (approved):
Adds a non-timing-out progress indicator so a slow apps search no longer looks hung; widget test included, clear symptom write-up. Approve-only — no CI has run on this branch yet.

kodjima33 (approved):
Search shimmer stops animating after 5s, so a slow search looks wedged; adds a non-timing-out indeterminate bar and extracts the sliver for a widget test. 5/5.

PR #11297, opened 2026-08-09
fix(app): hold the apps search in its searching state until the newest query lands

mikejsmith1985:
## Problem

Typing in the Apps tab intermittently shows "No apps found" for a query that does match.
Searching `adhd` returns nothing; the same search a moment later returns ADHD Assistant.
Nothing about the account or the app changed in between.

`performServerSearch` used `isSearching` for two jobs at once:

- the flag the apps screen renders its loading state from, and
- the re-entrancy guard that stops two requests running at the same time.

When a query was superseded mid-flight, the `finally` block cleared `isSearching` and
notified listeners **before** starting the queued follow-up. For that frame the screen saw
"not searching" alongside the previous query's results — which
`_buildFilteredAppsSlivers` renders as "No apps found" — for a search that was still
outstanding.

The superseded request was already discarding its results correctly. What it must also stop
doing is ending the searching state, which by then belongs to its replacement.

## Change

Splits the two jobs. `_isDrainingSearchQueue` guards re-entrancy, and one run drains the
queue in a loop rather than ending and recursing per query. `isSearching` now stays true for
the whole chain, so the screen cannot fall into the empty state between two queries.

The filter-building block moves into small named helpers so the request path stays readable.
The filters it produces are unchanged — same keys, same types, same null-vs-true semantics
for the mutually exclusive `my_apps` / `installed_apps` flags.

This also removes an unawaited recursive call to `performServerSearch()` from the `finally`
block.

## Test seam

Adds `searchAppsOverride`, following the `enableAppOverride` / `disableAppOverride` pattern
already in this provider. Without it the queueing behaviour is only reachable through a live
backend, which would put the regression test outside CI. The test drives two overlapping
queries through completers and records what the screen would render on every notification.

## Verification

- `flutter test test/providers/app_provider_search_flash_test.dart` — 3/3 pass. The first
  assertion fails on the parent commit with `(isSearching: false, resultCount: 0)` recorded
  while a newer query was still in flight. That is the flash.
- `flutter test test/providers/app_provider_toggle_test.dart` — 3/3 pass, unchanged.
- `flutter test` — 1119 pass, 2 fail: `plans_sheet_l10n` and
  `voice_recorder_mic_contention`. Both fail identically on `d60ab7a1c` with this work
  stashed. An earlier run also showed `voice_recorder_test` failing; it passes in isolation
  twice and did not recur on a second full run, so it is order-dependent and unrelated to
  this change.
- `dart analyze` — no new issues; two pre-existing `avoid_print` infos remain.
- `dart format --line-length 120` — applied.

**I could not run the app.** No macOS and no Android SDK on this machine, so this is proven
by test rather than on a device. I would rather say so than imply otherwise.

## Product invariants

`scripts/pr-preflight --suggest` reports none affected.

## Failure class

Failure-Class: none

**This is declared `none` reluctantly, and the reason is probably worth your attention.**

`FC-superseded-connection-mutates-live-session` describes this defect exactly. Its violated
contract — *a callback registered on an operation that has since been replaced must never
mutate the state its replacement now owns* — is precisely what went wrong; the retired
operation is an in-flight search rather than a socket, and the state it wrongly wrote is
`isSearching` rather than session connectedness. Its canonical prevention matches the fix
too: return early when the operation is no longer current, and prove it by delivering the
retired one's completion after the replacement is live. That is what the new test does.

I declared it, and `failure-class-protocol` passed. But `failure-class-guard-artifact-ratchet`
then failed:

```
FC-superseded-connection-mutates-live-session: 3 declarations in the window
(threshold 3) with no 'canonical_prevention_artifact'.
```

So this would be the third recurrence in 90 days of a class that still has no reusable guard
surface — which is exactly the signal that ratchet exists to raise. Clearing it means either
adding a `canonical_prevention_artifact` to the definition or grandfathering it in the
allowlist, and both are registry edits that `AGENTS.md` says an instance fix must never make.

I have declared `none` so this PR does not smuggle a registry change through a bug fix. If
you would rather it were classified, the honest sequence is a registry-only PR naming a guard
artifact for that class, then re-declaring here — happy to do either, but that is your call,
not mine.

---

*Written with Claude Code. Every file:line above was verified against `d60ab7a1c`.*

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11297?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/providers/app_provider.dart
+++ b/app/lib/providers/app_provider.dart
@@ -231,6 +231,10 @@ class AppProvider extends BaseProvider {
     searchQuery = query.toLowerCase();
 
     if (query.trim().isEmpty && !_hasServerSideFilters()) {
+      // Emptying the box is a search change like any other. Without its own
+      // revision, a request already in flight would still count as current and
+      // would land its results on top of the cleared list.
+      _latestSearchRevision++;
       searchResults = [];
       isSearching = false;
       filterApps();
@@ -248,15 +252,23 @@ class AppProvider extends BaseProvider {
         filters.containsKey('Apps');
   }
 
-  String _pendingSearchQuery = '';
+  /// Identifies the newest search the user has asked for.
+  ///
+  /// A revision rather than the query text, because the text is not what makes a
+  /// search unique: narrowing by category while the same word is still loading
+  /// changes the results completely and changes the text not at all. Anything
+  /// keyed on the text alone cannot tell those two requests apart, and answers
+  /// the second with the first one's unfiltered results.
+  int _latestSearchRevision = 0;
 
   bool _isDrainingSearchQueue = false;
 
   Future<void> performServerSearch() async {
-    // Always update pending query to the latest
-    _pendingSearchQuery = searchQuery;
+    // Every call is a fresh statement of what the user wants, whether the text or
+    // the filters moved.
+    final revision = ++_latestSearchRevision;
 
-    // A run already under way will pick this query up when its current request
+    // A run already under way will pick this revision up when its current request
     // returns. Starting a second one would race it for the results.
     if (_isDrainingSearchQueue) {
       return;
@@ -268,16 +280,16 @@ class AppProvider extends BaseProvider {
 
     try {
       // Drain the queue inside one run rather than ending and restarting per
-      // query. Ending in between would publish "not searching" with the results
-      // of a query the user has already left behind — which the apps screen
+      // request. Ending in between would publish "not searching" with the results
+      // of a search the user has already moved past — which the apps screen
       // renders as "No apps found" for a search that is still outstanding.
-      var queryBeingSearched = _pendingSearchQuery;
+      var revisionBeingSearched = revision;
       while (true) {
-        await _runOneSearch(queryBeingSearched);
-        if (_pendingSearchQuery == queryBeingSearched) {
+        await _runOneSearch(revisionBeingSearched);
+        if (_latestSearchRevision == revisionBeingSearched) {
           break;
         }
-        queryBeingSearched = _pendingSearchQuery;
+        revisionBeingSearched = _latestSearchRevision;
       }
     } finally {
       _isDrainingSearchQueue = false;
@@ -287,7 +299,18 @@ class AppProvider extends BaseProvider {
   }
 
   /// Runs one search request and publishes it only if it is still the current one.
-  Future<void> _runOneSearch(String queryBeingSearched) async {
+  Future<void> _runOneSearch(int revisionBeingSearched) async {
+    // Read the query and filters this request is for before awaiting, so a change
+    // made while it is in flight cannot be mistaken for what it asked the server.
+    final queryBeingSearched = searchQuery;
+
+    // Nothing to ask the server: an empty box with no filter narrowing it. Reached
+    // when clearing the box supersedes a request that was already in flight;
+    // searchApps has cleared the results for that case already.
+    if (queryBeingSearched.trim().isEmpty && !_hasServerSideFilters()) {
+      return;
+    }
+
     try {
       final result = await (searchAppsOverride ?? retrieveAppsSearch)(
         query: queryBeingSearched.isEmpty ? null : queryBeingSearched,
@@ -300,8 +323,8 @@ class AppProvider extends BaseProvider {
         limit: 100,
       );
 
-      // The user typed on while this was in flight; these results are stale.
-      if (queryBeingSearched != _pendingSearchQuery) {
+      // The user changed the search while this was in flight; these results are stale.
+      if (revisionBeingSearched != _latestSearchRevision) {
         return;
       }
 
--- a/app/lib/providers/app_provider.dart
+++ b/app/lib/providers/app_provider.dart
@@ -14,7 +14,25 @@ import 'package:omi/utils/alerts/app_snackbar.dart';
 import 'package:omi/utils/l10n_extensions.dart';
 import 'package:omi/utils/logger.dart';
 
+/// The shape of the apps-search request, named once so the test seam can refer to it.
+typedef AppsSearchRequest = Future<({List<App> apps, Map<String, dynamic> pagination, Map<String, dynamic>? filters})>
+    Function({
+  String? query,
+  String? category,
+  double? minRating,
+  String? capability,
+  String? sort,
+  bool? myApps,
+  bool? installedApps,
+  int offset,
+  int limit,
+});
+
 class AppProvider extends BaseProvider {
+  /// Test seam — overrides [retrieveAppsSearch] in [performServerSearch].
+  @visibleForTesting
+  AppsSearchRequest? searchAppsOverride;
+
   /// Test seam — overrides [enableAppServer] in [toggleApp].
   @visibleForTesting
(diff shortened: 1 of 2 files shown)


mikejsmith1985:
@cubic-dev-ai this is a real bug and it is now fixed in c23b235.

You were right that queue identity was keyed only on query text. Worth adding that it is not new to this branch — the code being replaced used the same `_pendingSearchQuery != queryBeingSearched` check, so `applyFilters()` during an in-flight request was already being dropped. My refactor carried the blind spot forward rather than introducing it. Since it is in the path this PR is rewriting anyway, it is fixed here in its own commit rather than left for later.

**What was wrong:** narrowing by Category, Rating, Capability or Apps while the same word is still loading changes the results completely and changes the query text not at all. The drain loop saw no new text, concluded there was nothing further to run, and published the unfiltered results as the answer to the filtered search.

**The fix,** as you suggested — a monotonically increasing revision bumped on every statement of search intent, rather than a text comparison. `_runOneSearch` also now reads the query and filters immediately before issuing its request, so what a request asked for cannot drift from what it is later judged against.

One case your comment did not name, which the same root cause produces: **clearing the search box.** That sets no new pending text either, so a request already in flight still counted as current and landed its results on top of the emptied list. It takes a revision now too, and `_runOneSearch` returns early when the box is empty with no filter narrowing it, so superseding a search that way costs no request.

**Regression tests** (`app_provider_search_flash_test.dart`), both failing on the parent commit:
- filter change mid-flight — 1 request made instead of 2, and the unfiltered results published
- clear mid-flight — `Late Result` lands on an emptied list

`flutter test test/providers/` — 167 pass. Full suite 1121 pass, 2 fail (`plans_sheet_l10n`, `voice_recorder_mic_contention`), both identical on `d60ab7a1c` with this work stashed. `pr_preflight --lane local` — 11/11.

Still not verified on a device: no macOS and no Android SDK on this machine, so this is proven by test rather than by use.

mikejsmith1985:
Thanks for the re-review. On the one thing you couldn't verify from the slice — it does notify, and immediately.

`searchApps` has always called `notifyListeners()` on the cleared path; it just isn't in the diff because it is an unchanged line. `app/lib/providers/app_provider.dart:230-243`:

```dart
if (query.trim().isEmpty && !_hasServerSideFilters()) {
  _latestSearchRevision++;   // <- the only line this PR adds here
  searchResults = [];
  isSearching = false;
  filterApps();
  notifyListeners();         // <- pre-existing
  return;
}
```

So the empty state repaints synchronously, before the drain loop's `finally` is reached. The `finally` is a second notification for the same settled state rather than the first one.

For completeness on the other half of your question: `filterApps()` (`:374`) does not notify on its own — it is a pure recompute and every caller notifies afterwards. That is why the pairing above is `filterApps(); notifyListeners();` rather than one or the other, and the convention is unchanged by this PR.

Git-on-my-level:
Thanks for this — the diagnosis is sharp and the fix targets the actual root cause.

What I verified:

- **app/lib/providers/app_provider.dart** — The dual-purpose `isSearching` (UI flag + re-entrancy guard) is correctly split into `isSearching` (UI only) and `_isDrainingSearchQueue` (re-entrancy). The drain loop holds `isSearching` true across the whole chain, so the apps screen can no longer render "No apps found" between a superseded request and its replacement. The `finally` block always resets `_isDrainingSearchQueue`, so there's no stuck-searching risk on exception paths.
- The revision-based identity (`_latestSearchRevision`, commit 2) is a real improvement over the text comparison it replaces: narrowing by Category/Rating/Capability/Apps while the same query text is loading changes results completely but changes the text not at all. The original `_pendingSearchQuery != queryBeingSearched` check had that same blind spot, so fixing it here rather than deferring is the right call. Clearing the empty box now takes a revision bump too, and `_runOneSearch` early-returns for empty+no-filters, so a superseding clear costs no request.
- The `AppsSearchRequest` typedef matches `retrieveAppsSearch` (`app/lib/backend/http/api/apps.dart`) exactly, so `searchAppsOverride ?? retrieveAppsSearch` is type-correct and the `@visibleForTesting` seam mirrors the existing `enableAppOverride` pattern.
- The extracted filter helpers preserve the original null-vs-true semantics (e.g. `_isAppsFilter` returns `true` or `null`, never `false`), so the mutually-exclusive `my_apps` / `installed_apps` flags are unchanged.
- The unawaited recursive `performServerSearch()` in the old `finally` block is gone — good; the drain loop makes it unnecessary.

**app/test/providers/app_provider_search_flash_test.dart** — 5 cases covering the flash, newest-query-wins-when-late, filter-change-queues-replacement, clear-discards-in-flight, and single-search-settles. The filter-change case is the one that would have silently broken under the text-only identity, so it's the right thing to pin down.

One thing I can't verify from here: there's no CI check data on the PR yet, and the queueing is only reachable through the test seam. A maintainer with the app running should confirm the behavior on-device (particularly the `applyFilters` path during an in-flight search). The logic is sound on static review.

Positive signal — not formally approving because there's no CI signal to confirm the build yet. Appreciate the honesty in the description about the two pre-existing test failures and not being able to run on a device.

---

mikejsmith1985:
@Git-on-my-level thanks for the review — and for reading the filter helpers closely enough to check the null-vs-true semantics. That was the part I was least sure would survive review.

On the one thing holding approval: **no CI has been triggered on this PR at all.** Not failing, not pending — zero workflow runs. The only check reporting is cubic.

That is not path filtering. `.github/workflows/backend-hermetic-e2e.yml` triggers on every `pull_request` to `main` with no `paths:` filter and does its scope detection inside the job, which is what makes `Backend Hermetic Merge Gate` report on every PR per AGENTS.md. It has not run here, and the same is true of #11293.

Most likely the first-time-contributor approval gate on fork pull requests — Actions will not run until someone with write access approves the workflow run. If that is it, a maintainer clicking **Approve and run workflows** should be all it takes, and I cannot do it from my side.

Not escalating this as urgent; it just seemed worth naming, since you said the CI signal was the only thing between this and a formal approval.

For what it is worth locally, on `d60ab7a1c`: `flutter test` 1121 pass / 2 fail (`plans_sheet_l10n`, `voice_recorder_mic_contention`, both failing identically with this work stashed), and `pr_preflight --lane local` 11/11. That is not a substitute for their CI and I would rather it ran.

On your other point — a maintainer confirming `applyFilters` during an in-flight search on a real device is exactly the check I could not do, and I would rather someone did it than take the tests as proof.

kodjima33 (approved):
Good root cause: isSearching was doing double duty as loading flag and re-entrancy guard, so a superseded query flashed "No apps found". Regression test included. Approve-only — no CI has run on this branch yet.

kodjima33 (approved):
isSearching was doing double duty as loading flag and re-entrancy guard, so a superseded query flashed 'No apps found'; splits the guard and drains the queue in one run, with a regression test that fails on the parent commit. 4/5.

PR #11364, opened 2026-08-10
feat: serve the primary-language picker list from the backend

mdmohsin7:
## Summary

The primary-language picker's list lived in `home_provider.dart`, so adding a language needed an app release even when the backend could already transcribe it. It now comes from `GET /v1/users/available-languages`.

**This is a move, not a product change.** The served list is the app's current 69 options, verbatim, in the same order. Nothing added, nothing removed.

## What it unblocks

The backend already accepts **12 base languages the app cannot offer**:

```
af Afrikaans   az Azerbaijani   cy Welsh      eu Basque
gl Galician    gu Gujarati      kk Kazakh     ml Malayalam
mt Maltese     pa Punjabi       sq Albanian   sw Swahili
```

Adding any of them is now a one-line edit in `PRIMARY_LANGUAGE_OPTIONS`. I deliberately did **not** add them here — that's a product call, and this PR is the plumbing.

## Backend

The list lives in `utils/user_language.py`, next to the normalization it has to agree with, rather than in a new module. The endpoint sits in the users router; the picker renders after `kAuthPage` in onboarding, so it can stay authenticated like its neighbours.

`tests/unit/test_language_catalog.py` is what makes the move safe: every offered code must survive `normalize_user_language` **unchanged**, must be serviceable by some STT provider, and must not collide with another entry. Without that, a future edit could add an option that `PATCH /v1/users/language` then rejects — a picker that offers something you can't save.

Generated app-client artifacts (Dart, TS, Swift, OpenAPI spec) regenerated for the two new schemas.

Two repo gates needed feeding, both intentionally: `routers/users.py` is line-count frozen, so the response models live in `models/users.py` and the router grows only by the route itself (2147 → 2159, baseline raised with a justification); and the new route carries a **reviewed** entry in `route_policy_manifest.yaml` rather than joining the 477-strong legacy-unreviewed baseline.

## App

`availableLanguages` prefers, in order: the served list → the list cached from the last successful fetch → the bundled copy.

**The hardcoded map is kept**, renamed `_bundledLanguages`. Deleting it would mean a first run with no network shows an empty picker. It stops being the source of truth and becomes the floor. The accessor keeps its `Map<String, String>` shape, so the four UI call sites are untouched.

A failed *or throwing* fetch leaves whatever the picker already had. `Env` isn't always initialised when this runs, and a stale list beats an empty one.

## Two latent bugs the move exposed

Both were safe only while the list was frozen into the build, and both are reachable once it can change under the app:

- `getLanguageName` was `firstWhere` with no `orElse` — threw on a stored code the list no longer carries. Now falls back to the bundled name, then the raw code.
- `language_selection_dialog.dart` had the same unguarded lookup and would throw a `StateError` on open. Its two sibling call sites were already guarded (`orElse` / `try-catch`); this was the last one.

## Not touched

`SttLanguages` in `models/stt_provider.dart` looks like a duplicate list but isn't — it's per-provider capability for BYOK custom STT (Whisper/Deepgram/Gemini), a different concept. Collapsing it onto this source would be wrong.

## Verification

```
pytest tests/unit/test_language_catalog.py                    -> 7 passed
pytest (catalog + user_language_normalization + trial_metadata) -> 138 passed
pytest (app_client_schema_inventory + dart_generator)         -> 46 passed
generate_dart_models.py --all --check                         -> up to date
generate_ts_openapi_types.py --check                          -> exit 0
generate_swift_openapi_types.py --check                       -> exit 0
export_openapi.py --app-client --check                        -> up to date
check_app_client_openapi_compatibility.py --base-ref origin/main -> passed

flutter test test/providers/home_provider_languages_test.dart -> 7 passed
app ./test.sh                                                 -> 1177 tests, all passed
flutter analyze -> 3 infos in users.dart, all confirmed pre-existing on main; 0 new
```

The full backend `pytest tests/unit` run dumps core on this machine — I confirmed it does the same on pristine `origin/main`, so it's a local environment problem, not this change. Targeted runs above plus CI cover it.

**Not verified on a device.** The fallback ordering is unit-tested, but I did not build the app and open the picker with the network off.

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mdmohsin7:
All three were valid — fixed in e236421951.

**Auth mismatch (P2).** Confirmed: the handler took no auth dependency and the users router has no global one, so the route was served anonymously while `route_policy_manifest.yaml` claimed `firebase_id_token` and the generated spec advertised `firebaseBearer` + a 401 it could never return. Added `Depends(auth.get_current_user_uid)`. The picker renders after `kAuthPage`, so nothing is lost by requiring auth — the declared policy was the correct intent, the implementation just didn't match it. Spec and clients regenerated.

**Startup ordering (P2).** Right. `loadAvailableLanguages()` and `setupUserPrimaryLanguage()` were fired independently, so a fresh user could get the picker on the bundled list while the served one was in flight. Chained with `whenComplete` rather than awaited, so the ordering is guaranteed without the rest of startup waiting on the request.

**Untested success path (P2).** The sharpest of the three. Every test that reached the fetch depended on it *failing* in the hermetic environment, so the branch that installs the served list — the headline behaviour — never executed. `loadAvailableLanguages` now takes an optional `fetch` seam, and the suite pins served-over-bundled, served-over-cached with the cache rewritten, a null fetch preserving the cache, and a throwing fetch.

That coverage immediately earned itself: **an empty served list replaced the picker with an empty map.** `getAvailableLanguages` already returns null for an empty payload, so it wasn't reachable through the real API, but the provider trusted whatever it was handed. Guarded at the assignment.

Also stripped the explanatory comments this branch had accumulated.

```
pytest (catalog + schema inventory + dart generator)  -> 46 passed
generate_{dart,ts,swift} --check                      -> all exit 0
check_app_client_openapi_compatibility --base-ref main -> passed
flutter test home_provider_languages_test             -> 12 passed (was 7)
app ./test.sh                                         -> 1182 passed
make preflight                                        -> 29 checks passed
```

mdmohsin7:
All three valid again — b2347ff8b6.

**Sign-out race (P1).** Mine, and introduced by the previous round's fix. `setupUserPrimaryLanguage` captures `_sessionGeneration` when it *starts*; chaining it onto the catalog request meant it started after a network round trip, so a `clearUserData()` during the request produced a generation it then happily agreed with — and the required picker could open over a signed-out app. `loadLanguagesThenSetupPrimary` captures the generation once, before the load, and spans both halves.

**Tests hitting the real network (P2).** Correct, and the failure scenario is exactly right: four fallback tests called `loadAvailableLanguages()` with no seam, ran the real authenticated `makeApiCall`, and passed only because it fails under the flutter_test HTTP block. A dev with an API base URL wired would have watched 'a cached list wins' fail against the real 69-language response. They inject a fetch now, matching the served-list tests.

**Overwritten justification (P3).** Also mine. I replaced the cumulative history for `users.py` — account-deletion telemetry, the +56 AI-profile route, the +5 executor move — rather than appending, which is what every other entry in that file does. Restored verbatim from `origin/main` with this route appended as `+8`.

One note on the new coverage: my first attempt at a sign-out test called `loadLanguagesThenSetupPrimary()` and *then* `clearUserData()`, asserting a flag that would have been false regardless — it exercised nothing. Replaced with a stub that signs out from inside the fetch, asserting the in-flight list is neither installed nor cached.

```
flutter test home_provider_languages_test -> 14 passed (was 12)
flutter analyze (provider, app_shell, test) -> No issues found
app ./test.sh -> 1184 passed
make preflight -> 29 checks passed
```

Issue #11365, opened 2026-08-10
Developer webhooks never fire for Limitless Pendant sessions (but do fire for phone recordings)


simonsbookclub:
# Developer webhooks (Real-Time Transcript, Conversation Events, Audio Bytes) never fire for Limitless Pendant sessions — but do fire for phone recordings

**App:** Omi iOS 1.0.531 (812) · **Device:** Limitless Pendant · **STT:** custom Cloud Provider (Deepgram nova-3, own API key)

## Summary

With all three developer webhooks enabled and saved, conversations captured **via the Limitless Pendant** never produce a webhook call. The **same app, same webhook URLs, same custom STT provider, same account** does deliver webhooks for conversations recorded with **Record with Phone**. The pendant conversations themselves are fine — they appear in the app and are retrievable via the Developer API (`GET /v1/dev/user/conversations?include_transcript=true`) with full transcript segments. Only the push path is silent.

## Evidence

Receiving endpoint is a Cloudflare Worker that logs **every** inbound request (method, headers, raw body) before parsing, so a malformed or unparseable delivery would still be recorded.

| Time (UTC+1) | Source | Webhook deliveries received |
|---|---|---|
| 17:00–17:03 | Record with Phone | `POST /ingest` ×4, `POST /memory` ×4 — HTTP 200 |
| 17:30 → next day | Limitless Pendant (many conversations) | **none** — zero requests logged |

Audio Bytes was also enabled with a valid URL for the pendant period: zero deliveries.

## Expected

Per https://docs.omi.me/doc/developer/apps/Integrations, the Real-Time Transcript webhook fires as segments are transcribed, and Conversation Events fires on conversation creation. No device-source limitation is documented.

## Actual

No webhook of any type fires for pendant-sourced conversations. Workaround in use: polling `GET /v1/dev/user/conversations?include_transcript=true` on a 60-second cron.

## Minor, related

1. The Real-Time Transcript webhook appends the uid with `?` instead of `&` when the configured URL already has a query string: configured `https://host/ingest?token=abc` is called as `https://host/ingest?token=abc?uid=xyz`, so the last configured query parameter's value is corrupted. (Conversation Events correctly uses `&`.)
2. The Audio Bytes toggle re-enables itself after being switched off and saved. While it is enabled with an empty URL, saving Developer Settings appears to silently fail — other webhook settings on the screen are not persisted either. Workaround: give Audio Bytes a valid URL.

kodjima33:
[...]

PR #11366, opened 2026-08-10
fix(app): stop a Developer Settings save from re-enabling toggled-off webhooks (#11365)

kodjima33:
## Bug

Reported in #11365: "The Audio Bytes toggle re-enables itself after being switched off and saved. While it is enabled with an empty URL, saving Developer Settings appears to silently fail — other webhook settings on the screen are not persisted either."

## Root cause

Two independent defects in `DeveloperModeProvider.saveSettings()`.

1. **The enabled state has two writers.** The toggle calls `enable/disableWebhook`, but `POST /v1/users/developer/webhook/{wtype}` also enables the webhook whenever the stored URL is non-empty (`backend/routers/users.py:486`). `saveSettings` POSTed all four URLs unconditionally, so any webhook the user had just switched off was turned back on by the next save. Audio Bytes hit this even with an empty URL, because the client always sends `"<url>,<delay>"` and the `",5"` payload is neither `''` nor `','`.

   The URL field is only rendered while its toggle is on (`developer.dart:222`, `if (isEnabled)`), so the toggle is the sole authority for the enabled state and the stale controller text behind an off toggle should never re-enable anything.

2. **A failed save reported success.** `Future.wait([w1, w2, w3, w4]);` was never awaited, and `setUserWebhookUrl` returns `false` on failure rather than throwing — so the `try/catch` was dead code. Every save wrote the local prefs and showed "Settings saved!" regardless of what the backend did.

## Fix

`saveSettings` awaits the four writes, treats a `false` as a failure (log + error snackbar, prefs untouched, no success toast), and then re-asserts `disableWebhook` for every toggle the user left off. Two `@visibleForTesting` seams follow the existing `AppProvider.enableAppOverride` pattern.

## Also in #11365

- The realtime-transcript webhook appending `?uid=` to a URL that already has a query string is **already fixed on main** — `_append_query_params` in `backend/utils/webhooks.py` rebuilds the query with `urlencode`.
- The headline complaint (no webhooks at all for Limitless Pendant sessions) is **not** addressed here; it needs pendant hardware to reproduce and is left open on the issue.

## Tested

- `flutter test test/providers/developer_mode_provider_save_settings_test.dart` → 2 passed. Both cases were confirmed to **fail** against the pre-fix method body with the test seams kept in place (`disabledTypes` empty; the rejected save still persisted the URL).
- `flutter test test/providers/` → 178 passed.
- `flutter analyze` on both changed files → no issues.
- **Not exercised on a device:** no iOS/macOS run of the Developer Settings screen. The regression test drives the real `saveSettings()` through a real `MaterialApp`/`ScaffoldMessenger`, and asserts the snackbar text, but a human has not tapped Save in the running app.

## Product invariants

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11366?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/providers/developer_mode_provider.dart
+++ b/app/lib/providers/developer_mode_provider.dart
@@ -13,6 +13,14 @@ import 'package:omi/utils/logger.dart';
 import 'package:omi/utils/other/validators.dart';
 
 class DeveloperModeProvider extends BaseProvider {
+  /// Test seam — overrides [setUserWebhookUrl] in [saveSettings].
+  @visibleForTesting
+  Future<bool> Function({required String type, required String url})? setUserWebhookUrlOverride;
+
+  /// Test seam — overrides [disableWebhook] in [saveSettings].
+  @visibleForTesting
+  Future<void> Function({required String type})? disableWebhookOverride;
+
   final TextEditingController webhookOnConversationCreated = TextEditingController();
   final TextEditingController webhookOnTranscriptReceived = TextEditingController();
   final TextEditingController webhookAudioBytes = TextEditingController();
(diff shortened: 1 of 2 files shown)


kodjima33:
Closing this rather than leaving it open.

**Why its CI was red — not this diff.** The `Dart Analyze & Tests` failure here
(`override_on_non_overriding_member: 5 found, baseline 0`) is pre-existing on `main`:
6908c4f5f6 removed `openAIAPIKey` from `EnvFields` but left the `@override` stub on five
`_TestEnvFields` classes. I reproduced the identical ratchet failure on a clean `origin/main`
checkout (f51681dc9e). The job is path-filtered, so post-merge push runs on `main` skip it and
the breakage only fires on PRs touching `app/`. Fixed separately in #11368.

**Why this PR is still being closed.** Two reasons:

1. The user-reported symptom in #11365 ("the Audio Bytes toggle re-enables itself after being
   switched off and saved") is an empty-URL case, and #11367 fixes it at the source in
   `set_user_webhook_endpoint` — for every client, not just the Flutter app. Landing both would
   be duplicate coverage of the same defect.
2. This diff adds `@visibleForTesting` override-function fields (`setUserWebhookUrlOverride`,
   `disableWebhookOverride`) to the production provider purely as test seams. That is the wrong
   shape for the repo.

**What is genuinely not covered by #11367** (real, worth a follow-up, filed here so it is not lost):

- `saveSettings()` does not `await` its `Future.wait([w1..w4])`, so prefs are persisted and
  "Settings saved!" is shown even when the backend calls fail. This is the "saving Developer
  Settings appears to silently fail" half of #11365.
- A webhook toggled **off** whose URL field still holds a valid URL is re-enabled on save,
  because `set_user_webhook_endpoint` enables on any non-empty URL and the app posts every
  field on every save. #11367 only handles the empty/`,5` case.

#11365 stays open for its headline defect (no webhooks at all for Limitless Pendant sessions),
which neither PR addresses.

🤖 hourly watchdog

PR #11387, opened 2026-08-10
Improve mobile conversation responsiveness

Git-on-my-level:
## Summary

- Make the production-data `mobile_beta` profile explicit while keeping production Firebase/Auth and beta OAuth routing.
- Prewarm and preserve mobile tabs without blocking the tap frame, and show skeletons instead of transient empty states.
- Flatten conversation and detail content into lazy slivers, isolate repaint-heavy rows, and defer analytics/provider work until after visible navigation.
- Restore the Home conversation preview and keep Daily Recaps, Mind Map, and compact journey locations intact.
- Bound goal slider divisions so large numeric goals cannot block scrolling or route transitions.

## Root cause

Physical-device Time Profiler captured six severe main-thread hangs lasting 3.22-3.74 seconds. The dominant sampled frame was `RenderSlider.paint`. A numeric goal with a target near 1,000,000,000 created one slider division per target unit, so Flutter walked roughly one billion divisions during every repaint even though tick marks were hidden. Scrolling Conversations and opening either a goal or conversation repainted that source page and hit the same pathological loop.

Large and fractional goals now use a continuous slider. Small integral goals up to 100 retain exact one-unit divisions. The Goals section is also isolated behind a repaint boundary.

## User impact

- Conversations scrolling and row taps remain responsive with very large numeric goals.
- First-load navigation displays immediate skeleton feedback instead of suggesting data is gone.
- Tab state, restored Home content, and existing goal editing behavior are preserved.

David verified the final Profile-prod `mobile_beta` build on a physical iPhone 15 Pro Max and confirmed the lag is fixed.

## Validation

- `flutter test --no-pub test/widgets/bottom_nav_bar_test.dart test/widgets/conversation_list_item_performance_test.dart test/widgets/conversation_markdown_widget_test.dart test/widgets/daily_summary_detail_page_test.dart test/widgets/goals_widget_performance_test.dart test/widgets/home_conversations_preview_test.dart test/providers/conversation_provider_tap_test.dart test/pages/conversations_group_widget_test.dart` (14 passed)
- `bash scripts/analyze_ratchet.sh` (passed)
- Repository pre-push gate (passed), including PR preflight, production routing invariants, Flutter generated output, localization output, and Dart formatting.
- Physical-device build: `flutter build ios --profile --flavor prod --dart-define=OMI_APP_PROFILE=mobile_beta --dart-define=OMI_API_BASE_URL=https://api.omiapi.com/`
- Installed bundle verified as `com.friend-app-with-wearable.ios12`, `omi-beta`, and `https://api.omiapi.com/`.

Failure-Class: none


Git-on-my-level:
Monitor review-fix pass for commit `c16b9367`.

**Code fix (1 commit):**
- `app/lib/pages/apps/page.dart`: Wrapped `_loadApps()` body in try/catch with `Logger.handle` so a future provider error cannot surface as an unhandled async exception from the post-frame callback. (Cubic P2, thread `…Dpb2`)

**Threads resolved:**
- `…Dpb2` (`page.dart`): Fixed — error containment added.
- `…DpcN` (`firebase_options_local.dart`): Already addressed by prior commit `cfa527f576` — the duplicate `setup/prebuilt/firebase_options_local.dart` was deleted and all setup/test consumers now source from `lib/firebase_options_local.dart`, eliminating the dual source of truth.

**Threads left open (reviewed at stale commit `9b779ca`, not actionable against current diff):**
- `…Dpbu` (dev `AndroidManifest.xml`): No current diff on this file (unchanged from `main`).
- `…Dpb6` (`generate_ios_dev_info_plist.sh` as Xcode build phase): Scope creep for a performance PR; `Info-Dev.plist` is committed and available without `setup.sh`.
- `…DpcA` (`prodRelease.xcconfig`): `GOOGLE_REVERSE_CLIENT_ID` was already re-added by commit `cfa527f576` with a deterministic override after `Custom.xcconfig`.

**Verification:**
- `flutter analyze lib/pages/apps/page.dart` — no issues.
- `flutter test test/providers/app_provider_load_test.dart` — 3/3 passed.

PR #11403, opened 2026-08-11
fix(desktop): probe fallback must honour app privacy exclusions

kodjima33:
## The bug

`probeSuggestionNudge` falls back to capturing the active window when no frame is pending. That fallback called `ScreenCaptureService` directly with **no exclusion check**.

The sharp edge: `latestCapturedFrame` is nil *precisely* when the frontmost app is privacy-excluded, because the capture gate refuses those apps. So the fallback fired exactly in the apps the user told Omi never to look at — and sent the screenshot to a model.

Introduced in #11392 (merged), which is why this is a follow-up rather than an amend.

## The fix

- Exclusion is resolved **before** anything is captured, not filtered afterwards.
- An excluded app returns `excluded_app`, not `no_frame_captured` — a privacy decision and a capture hiccup are different answers and shouldn't read the same in automation output.
- `SuggestionProbePrivacy.isExcluded` consults the same `RewindSettings` list the normal capture path uses, factored out so it is testable without a running app or a real screen.
- Removes a duplicated doc line a previous comment split left above `refreshGoalsIfStale`.

## Verification

- `SuggestionProbePrivacyTests`: excluded app refused, non-excluded allowed, exact-name matching (`Messages` excluded ≠ `Messages Beta`).
- **122 focused tests pass**, swift-format clean.
- Exercised on the running named bundle — delivery still works with the gate in place: *"Discord onboarding is fine - but you said you'd recruit people for Omi by Saturday"*.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11403?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/Suggestions/SuggestionAssistant.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/Suggestions/SuggestionAssistant.swift
@@ -267,8 +267,6 @@ actor SuggestionAssistant: ProactiveAssistant {
     return grounding
   }
 
-  /// Describe a commitment the way a person would say it out loud.
-  ///
   /// Fire-and-forget goal refresh. Never awaited by grounding: the next evaluation gets the
   /// fresher list, this one is not delayed for it.
   ///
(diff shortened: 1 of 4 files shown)


kodjima33:
Superseded: this fix landed on main verbatim via the combined proactivity merges — main's ProactiveAssistantsPlugin+SuggestionProbe.swift contains the window-ID-bound capture, allowsCapture both-ends check, and cached-frame exclusion re-check from this branch. The only branch-side differences are cosmetic (outcome string 'excluded_app' vs main's 'skipped_excluded_app') or redundant (an explicit RewindSettings check that SuggestionAssistantSettings.isAppExcluded already delegates to). Nothing protective is lost by closing.

Issue #11412, opened 2026-08-11
Anthropic server-side web_search is an unguarded egress path for prompt-injected exfiltration


undivisible:
PR #11015 bounds `fetch_url_tool` to URLs the user typed in the current turn, and enforces the same policy on redirect hops. It does not cover Anthropic's server-side `web_search`.

`WEB_SEARCH_TOOL` (`backend/utils/retrieval/agentic.py`) is attached to the request whenever `include_server_web_search=True`. Anthropic executes those `server_tool_use` blocks itself; the backend only observes them at `content_block_start` to emit a status thought. They never reach `_execute_tool`, so the current-turn allowlist, the SSRF address guard, and the redirect check do not apply.

Consequence: after a memory, Gmail, or screen tool returns sensitive data, an injected instruction can get that data placed in a `web_search` query string, which leaves the boundary as an outbound request. The only control today is the prose in `AGENT_SAFETY_INSTRUCTIONS`.

Reported by Codex review on #11015 (comment 3714902708). Left out of that PR deliberately — it is a product-behaviour change to the search surface, not part of the bounded `fetch_url` slice.

Options worth evaluating: drop the server tool from loops that have already consumed untrusted tool output, or route search through a function tool that can apply a runtime provenance check.

PR #11414, opened 2026-08-11
security(backend): gate Anthropic server-side web search on private-context taint

undivisible:
Fixes the exfiltration path in #11412.

## Problem

Anthropic's `web_search_20250305` runs **at Anthropic**. Its `server_tool_use` blocks never reach `_execute_tool`, so the `fetch_url` allowlist, SSRF guard and redirect checks in `backend/utils/retrieval/tools/web_tools.py` never apply to the query string. The only control was prose in `AGENT_SAFETY_INSTRUCTIONS`, which prompt injection defeats.

Worse, `web_search_requested` was true whenever `bool(client_tools)` was truthy: merely sending any client tool implicitly enabled server-side search. So a turn that had just returned memory / screen / connector output could have that data carried out in a search query.

The route is already authenticated; authentication was never the gap.

## What this changes

1. **Taint gate (primary control).** `_request` refuses to inject `web_search` when the request already carries private tool output. No client change is required, so ordinary turns behave exactly as before for every shipped client.
2. **Implicit trigger removed.** `bool(client_tools)` no longer requests web search. Only explicit `omi_web_search: true` or the existing public-web routing instruction do.
3. **Per-user authorization (defence in depth).** Evaluated server-side on the authenticated `uid` before injection, against `assistant_settings.web_search.enabled` — the same per-user settings map `focus`/`task`/`memory` use. **Absence means allowed**, so principals that predate the gate keep today's behavior; denial has to be a stored decision. Lookup failure fails closed to model knowledge.
4. **Telemetry.** Every newly withheld path calls the shared `record_fallback` with the existing `from_mode='anthropic_web_search'` shape: `private_tool_output_in_context`, `not_authorized`, `authorization_unavailable`. No new counters.

`server_tool_use` usage accounting and the content-block passthrough are untouched.

## What counts as private, and how it is identified

Two detection surfaces, both keyed on literal contract shapes rather than content heuristics:

**A. OpenAI `role: "tool"` messages.** Each `tool_call_id` is resolved back to the tool name declared in the preceding assistant `tool_calls` entry. Private unless the name is on a small allowlist of tools that return no user data:

`cancel_agent_run`, `check_permission_status`, `create_action_item`, `create_calendar_event`, `point_click`, `request_permission`, `set_desktop_attention_override`, `update_action_item`, `update_agent_artifact_lifecycle`

Everything else in the shipped hub surface (`desktop/macos/Desktop/Sources/Generated/GeneratedRealtimeTools.swift`) is treated as private — `search_screen_history`, `get_daily_recap`, `get_conversations`, `search_conversations`, `get_memories`, `search_memories`, `get_action_items`, `get_tasks`, `screenshot`, `report_screen_observation`, `inspect_agent_artifacts`, `get_agent_run`, `list_agent_sessions`, `ask_higher_model`, `spawn_agent`. **Unknown or unresolvable tool names are private**, so a future connector (Gmail, files) is covered on the day it ships, before anyone remembers to update a list.

**B. Tool output inlined into the user turn.** `RealtimeHubTools.swift:290` appends tool context to the query behind the literal marker `\n\nTool-provided context (untrusted):\n`. That path sends no `tool` message at all, so surface A cannot see it. Any user message containing that marker is treated as private.

### Named gap (deliberately not gated)

The desktop kernel also prefixes user turns with a **context snapshot** (`[Kernel Context Snapshot `, `# Omi Context Snapshot`). That marker is present on essentially every desktop turn, so tainting on it would disable web search for all desktop traffic. Whatever private material the snapshot carries is therefore still reachable by an Anthropic-side query. This is a real residual gap, called out rather than half-gated; closing it needs the snapshot's contents classified, or the follow-up below.

Follow-up direction (not built here): a first-party search tool routed through the same allowlist/SSRF path as `fetch_url`, which removes the provider-executed egress entirely. Nothing here makes that harder — `_web_search_requested` and `_carries_private_tool_output` are pure predicates over the request body.

## Tests

New, all executing production behavior through `_request` / `chat_completions` / `_web_search_authorized`:

- implicit `client_tools` no longer injects web search
- private tool output in the transcript blocks injection (unit + full route through `chat_completions` asserting the Anthropic payload)
- tool output inlined behind the untrusted marker blocks injection
- unrecognized tool name is treated as private
- only-public-safe tool output still gets web search
- unauthorized principal is refused, with the `not_authorized` fallback
- **legacy principal**: no user doc / no `web_search` section → allowed; only a stored `enabled: false` denies; lookup failure fails closed and records `authorization_unavailable`

Two existing assertions were changed deliberately, both because they encoded the vulnerability:

- `test_request_adds_web_search_alongside_client_tools_but_not_for_haiku_or_none` asserted `['web_search', 'weather']` for a turn whose only signal was the presence of client tools — that is the implicit trigger this PR removes. Renamed and now asserts `['weather']`; its haiku/`tool_choice` halves are unchanged.
- `test_request_classifies_only_trusted_query_before_tool_context` asserted that a public-looking query **with tool output inlined behind the untrusted marker** still got `web_search`. That is exactly the exfiltration shape in #11412. It now asserts no injection, and a new clean-turn case preserves the positive assertion it used to carry.

The remaining `_request` call sites go through a test-local `_authorized_request` wrapper that defaults `web_search_allowed=True`, so those cases keep asserting the non-authorization conditions with no change to their bodies or assertions.

## Verification

```
BACKEND_UNIT_TEST_FILE_LIST=<test_desktop_chat.py + test_assistant_settings_response_models.py> bash backend/test.sh
  69 passed / 3 passed
```

Mutation verification (break, observe failure, revert):

| mutation | result |
|---|---|
| drop `and not private_context_present` | 3 failed — private-output unit, unrecognized-tool, route-level |
| make `_carries_private_tool_output` ignore the inlined marker | 2 failed — inlined-marker test, trusted-query test |
| drop `and web_search_allowed` | 1 failed — unauthorized-principal test |
| restore `or bool(client_tools)` | 14 failed |
| make authorization require a stored `enabled: true` | 3 failed — incl. the legacy-principal test |

`make preflight` run before opening.

Not exercised against live Anthropic: this is a payload-construction change, and the route-level test asserts the exact tool list handed to the Anthropic client.

Failure-Class: new

Line-Count-Exception: backend/routers/desktop_chat.py | 1397 -> 1505 | server-side Anthropic web-search gate (taint check, per-user web_search setting, typed auth status) crosses the 1500-line threshold pending a router split
Line-Count-Exception: backend/routers/users.py | 2174 -> 2179 | assistant_settings.web_search.enabled joined to the strict PATCH model

New class: a capability the provider executes on its own infrastructure bypasses the egress guard that the in-process tool path enforces, so the guard's coverage silently ends at the trust boundary. The gate must live where the request is composed, not in the tool executor the provider never calls.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **High Risk**
> Changes security-sensitive desktop chat egress and can silently degrade web search for many turns; misclassification of “safe” tool output or settings lookup behavior would have data-leak or availability impact.
> 
> **Overview**
> **Closes an exfiltration path** where Anthropic’s provider-executed `web_search` could run with private transcript data because it never hits the in-process `fetch_url` / SSRF controls.
> 
> Desktop chat now **only injects** Anthropic `web_search` when the request is explicitly opted in (`omi_web_search` or the public-web routing instruction)—**not** merely because client tools are present—and when **`_carries_private_tool_output`** is false. That check treats OpenAI `tool` results as private unless the tool name is on a small allowlist, flags unknown tools as private, and treats user turns containing the untrusted inlined-tool marker as private.
> 
> **Per-user defense in depth:** before injection, `chat_completions` loads `assistant_settings.web_search.enabled` (missing section stays **allowed** for legacy users; only stored `enabled: false` denies). Settings lookup failures fail closed and record `authorization_unavailable`, distinct from `not_authorized`.
> 
> Withheld paths use **`record_fallback`** with new reason labels (also allowlisted in observability). Assistant settings API/OpenAPI/generated clients gain optional `web_search.enabled`. A failure-class doc captures the provider-executed-tool egress pattern.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f5fb5af9e6dbbc77342d9a5096ae33448816725. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Git-on-my-level (dismissed):
Thanks for tightening this. The core backend taint gate looks directionally right, but I think this needs a small telemetry-contract fix before merge.

Blocking:
- `backend/routers/desktop_chat.py`: the new `record_fallback(...)` calls emit `private_tool_output_in_context`, `not_authorized`, and `authorization_unavailable`, but `backend/utils/observability/fallback.py` does not allowlist those reason labels. `record_fallback` will bucket them all to `other`, so the newly introduced security withhold paths are not distinguishable in metrics/logs despite the PR saying telemetry preserves those exact shapes. Please add the new reasons to the shared fallback contract and include a regression test at that boundary.
- `backend/routers/desktop_chat.py`: `_web_search_authorized()` records `authorization_unavailable` and returns `False`; then `_request(..., web_search_allowed=False)` records `not_authorized` for the same request when there is no private context. That makes a Firestore/settings outage look like both an infra failure and an explicit per-user denial. Please make these mutually exclusive so lookup failures do not also emit `not_authorized`.

File-by-file notes:
- `.github/failure-classes/FC-provider-executed-tool-bypasses-egress-guard.json`: useful failure-class capture for provider-executed tools crossing the in-process executor trust boundary.
- `.github/scripts/product_file_line_count_ratchet_baseline/backend-routers.json`: the ratchet bump matches the small `users.py` schema addition and rationale.
- `backend/routers/desktop_chat.py`: removing implicit `bool(client_tools)` web-search injection and checking `_carries_private_tool_output()` before offering Anthropic server-side `web_search` is the right security direction; the fallback telemetry issues above are the blockers.
- `backend/routers/users.py`: adding `assistant_settings.web_search.enabled` to the strict PATCH model gives the route a writable per-user deny state instead of relying on an undeclared Firestore field.
- `backend/tests/unit/test_desktop_chat.py`: good coverage for explicit opt-in, unknown/private tool output, inline untrusted tool context, legacy allowed settings, stored deny, and auth lookup failure; please extend it (or fallback unit tests) to verify the shared fallback reason bucketing and no duplicate auth-failure reason.
- `desktop/windows/src/renderer/src/lib/omiApi.generated.ts`: generated API type now exposes `web_search?: WebSearchAssistantSettings | null` consistently for desktop clients.
- `docs/api-reference/app-client-openapi.json`: OpenAPI now includes `WebSearchAssistantSettings` on assistant settings response/update request.
- `web/admin/lib/services/omi-api/omiApi.generated.ts`: generated web admin client matches the new assistant-settings shape.
- `web/app/src/lib/omiApi.generated.ts`: generated main web client matches the new assistant-settings shape.
- `web/personas-open-source/src/lib/omiApi.generated.ts`: generated personas client matches the new assistant-settings shape.

Because this is a security-sensitive backend gate, I’m leaving this blocked on the telemetry-contract fix rather than formal approval.

---

kodjima33 (approved):
Gating provider-executed web search on private-context taint is the right boundary; approve-only pending a rebase (conflicts with main).

Git-on-my-level:
Verified on current head `1f5fb5af9e` — following up on my earlier change request.

**Resolved — dismissing my Aug 11 CHANGES_REQUESTED.** `backend/utils/observability/fallback.py` now allowlists `private_tool_output_in_context`, `not_authorized`, and `authorization_unavailable`, and `test_web_search_security_reasons_keep_their_exact_labels` pins them so they can't silently collapse to `other`. The tri-state refactor in `desktop_chat.py` additionally keeps `authorization_unavailable` (lookup failure, fail-closed) mutually exclusive with `not_authorized` (stored denial) — one step beyond what I asked for.

File-by-file pass:

- `backend/routers/desktop_chat.py` — injection is a conjunctive gate (`_request`: supported model ∧ `tool_choice != 'none'` ∧ not prohibited ∧ requested ∧ authorized ∧ untainted). Taint detection classifies `role='tool'` results against `_PUBLIC_SAFE_CLIENT_TOOLS` with unknown tool names defaulting private, and also catches the desktop hub's inlined `Tool-provided context (untrusted)` user-turn variant. `_request` has a single production call site in this module (`chat_completions`), `_web_search_authorized` fails closed on Firestore errors, and the implicit `bool(client_tools)` trigger is gone (`test_request_does_not_let_client_tools_alone_enable_server_side_web_search`).
- `backend/routers/users.py` — `WebSearchAssistantSettings.enabled` rides the existing `assistant_settings` deep-merge; absent = allowed keeps legacy principals on today's behavior.
- `backend/tests/unit/test_desktop_chat.py` (+283) — 13 new gate tests, including an end-to-end one asserting at the real Anthropic client seam that `_WEB_SEARCH_TOOL` is absent from the outgoing payload when private tool output is in context. Both touched suites pass locally under my run (81 tests).
- `backend/tests/unit/test_fallback_observability.py` / `backend/utils/observability/fallback.py` — exact-label pinning discussed above.
- `.github/failure-classes/FC-provider-executed-tool-bypasses-egress-guard.json` — schema matches sibling entries and describes the bypass class accurately; note it feeds `scripts/pr-preflight --suggest`, so future `fix:` PRs in this area will cite it.
- `desktop/windows/…/omiApi.generated.ts`, `web/admin/…`, `web/app/…`, `web/personas-open-source/…`, and `docs/api-reference/app-client-openapi.json` — pure regenerated `web_search` settings surface, consistent across all four clients.

**One thing needing maintainer eyes:** @kodjima33's Aug 13 approval covers `7a312b26`, which is not an ancestor of the current head. The delta that matters: `_PUBLIC_SAFE_CLIENT_TOOLS` was widened after that approval — `create_action_item`, `create_calendar_event`, `update_action_item`, and `update_agent_artifact_lifecycle` were added, and the earlier "write tools are excluded because their arguments carry user data" caution was replaced by a returns-no-user-data rationale. That rationale holds for tool *results* (what the taint check classifies), but prior `tool_use` *arguments*, which can embed private screen/memory content, do remain in the transcript when search is offered. It's a defensible trade-off — the model can echo user-turn text into a query regardless — but it's a security-boundary judgment made after the approval, so it warrants your re-sign-off rather than mine.

Label changes: dropped `needs-tests` (coverage is now thorough and green), keeping `security-review`, adding `needs-maintainer-review` for the boundary re-approval above.

— glm-5.3 maintainer review

---

PR #11281, opened 2026-08-09
fix(backend): pin chat agent route + prod gateway off after chat outage

Git-on-my-level:
## Summary
Fix the 2026-08 agentic chat outage class and re-enable LLM gateway + Luna chat safely.

1. **`OMI_LLM_CHAT_AGENT_ROUTE`** (`direct` | `gateway`, aliases `luna`/`off`): independent kill switch for managed agentic/desktop chat.
2. Prod defaults: **`FEATURE_MODE=gateway`** + **`CHAT_AGENT_ROUTE=gateway`** (Luna chat on via `omi:auto:chat-agent` OpenAI surface / gpt-5.6-luna).
3. Wire agentic + desktop + Anthropic client selection so chat can stay `direct` while other gateway features stay on.
4. Runbook: emergency kill switch = set `OMI_LLM_CHAT_AGENT_ROUTE=direct` only.

## Failure-Class
Failure-Class: none

## Test plan
- [x] unit: gateway client config + deploy contract + runtime_env render (61 passed)
- [x] surface: `chat_agent`+openai → `openai.chat_completions`
- [ ] after merge: prod cloud-run-only deploy
- [ ] smoke agentic chat; if bad, env-only `CHAT_AGENT_ROUTE=direct` (no full gateway off)

## Related
- Multica SCA-304 / SCA-305
- Live emergency rollback was FEATURE_MODE=off; this PR re-enables with chat kill switch

PR #11419, opened 2026-08-11
fix(chat): accept file attachments under gateway feature mode

mdmohsin7:
## Summary

Since prod flipped `OMI_LLM_GATEWAY_FEATURE_MODE=gateway` (#11281), attaching a file in chat has been broken for all users: every `POST /v2/files` 500'd because `FileChatTool.upload()` raised `GatewayDirectModelSurfaceBlocked` — file chat runs directly on OpenAI Files/Assistants/vision and has no gateway lane. #11191 gave the chat *stream* a typed user-safe failure but the upload endpoint kept throwing. The app compounded it: the thrown upload error skipped spinner/selection cleanup, leaving a forever-loading thumbnail whose remove button crashed with `RangeError`.

- **Backend:** file chat is now an *acknowledged* direct surface — the gate records `record_direct_exception_surface(surface='file_chat.openai_files_assistants_vision')` under gateway feature mode and never raises, so uploads and file-chat answers work in every environment without env overrides. The typed `GatewayDirectModelSurfaceBlocked` handling in `graph.py` stays for the generic gateway surfaces.
- **App:** `uploadFiles` routes a thrown `uploadFilesServer` error through the existing failure path (spinner reset, selection cleared, failure snackbar), and `clearSelectedFile` is bounds-safe so removing a failed attachment can no longer throw `RangeError`.

Failure-Class: none

The regression is an instance of a mode-flip gating an ungated surface; no existing registry class matches, and the reusable guard here is the new behavioral test pinning upload's non-blocking contract under gateway mode.

## Test plan

- [x] `pytest tests/unit/test_chat_file_gateway_surface.py tests/unit/test_chat_async_offload.py tests/unit/test_llm_gateway_client_config.py tests/unit/test_chat_file_skip_malformed.py` → 58 passed
  - New `test_chat_file_gateway_surface.py`: upload succeeds and records the acknowledged surface under `OMI_LLM_GATEWAY_FEATURE_MODE=gateway`; records nothing outside gateway mode.
- [x] `flutter test test/providers/message_provider_attachment_test.dart` → 2 passed (regression for the `RangeError` misalignment)
- [x] `dart analyze lib/providers/message_provider.dart` clean (one pre-existing unrelated `unnecessary_import` info)
- [x] `make preflight`

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11419?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/providers/message_provider.dart
+++ b/app/lib/providers/message_provider.dart
@@ -360,7 +360,13 @@ class MessageProvider extends ChangeNotifier {
       if (res != null) {
         uploadedFiles.addAll(res);
       } else {
-        clearSelectedFiles();
+        final failedPaths = files.map((e) => e.path).toSet();
+        for (var i = selectedFiles.length - 1; i >= 0; i--) {
+          if (failedPaths.contains(selectedFiles[i].path)) {
+            selectedFiles.removeAt(i);
+            selectedFileTypes.removeAt(i);
+          }
+        }
         final l10n = globalNavigatorKey.currentContext?.l10n;
         AppSnackbar.showSnackbarError(l10n?.msgUploadFileFailed ?? 'Failed to upload file, please try again later');
       }
--- a/app/lib/providers/message_provider.dart
+++ b/app/lib/providers/message_provider.dart
@@ -320,9 +320,9 @@ class MessageProvider extends ChangeNotifier {
   }
 
   void clearSelectedFile(int index) {
-    selectedFiles.removeAt(index);
-    selectedFileTypes.removeAt(index);
-    uploadedFiles.removeAt(index);
+    if (index < selectedFiles.length) selectedFiles.removeAt(index);
+    if (index < selectedFileTypes.length) selectedFileTypes.removeAt(index);
+    if (index < uploadedFiles.length) uploadedFiles.removeAt(index);
     notifyListeners();
   }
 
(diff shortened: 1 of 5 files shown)


mdmohsin7:
Addressed all three cubic findings in 29900e46dc:

- **P1 (gateway misconfig still raises):** valid — `should_route_features_through_gateway()` raises `RuntimeError` in a prod runtime when the allow-prod flag or gateway URL is missing. The gate now treats that as gateway mode and records the surface instead of propagating, so upload can never be blocked by rollout config. New regression test `test_upload_proceeds_when_gateway_mode_is_misconfigured_in_prod`.
- **P2 (failure branch wipes tray / leaves stale `uploadedFiles`):** valid — the failure path now removes only the failed batch from `selectedFiles`/`selectedFileTypes`, preserving earlier successfully uploaded selections and their `uploadedFiles` entries, so a later message can't silently attach invisible stale files. New regression test drives the real `uploadFilesServer` throw and asserts only the failed batch is removed.
- **P3 (test env not self-contained):** valid — `_gateway_mode` now pins `OMI_ENV_STAGE=dev`.

`pytest tests/unit/test_chat_file_gateway_surface.py` → 3 passed; `flutter test test/providers/message_provider_attachment_test.dart` → 3 passed.

mdmohsin7:
Addressed both round-2 findings in c405d89284:

- **P1 (duplicate-path removal):** valid — cleanup now removes the exact `File` instances of the failed batch instead of matching by path; every `uploadFiles` caller passes the instances stored in `selectedFiles`, so identity matching is sound. Regression test now uses a duplicate path and asserts the earlier successful selection survives.
- **P2 (`clearSelectedFile` edge indices):** valid — the index is validated against `selectedFiles` once up front (negative or out-of-range → no-op), so no list is touched and a stale `uploadedFiles` entry can no longer be removed by an unrelated index. Test pins both cases.

`flutter test test/providers/message_provider_attachment_test.dart` → 3 passed.

PR #11425, opened 2026-08-11
fix(backend): enforce desktop Gemini quotas

### LINKED ISSUE #11423: Desktop Gemini proxy inverts rate-limit check: Pro silently downgraded to Flash from the first call; daily hard limit never enforces
## What

`backend/routers/desktop_proxy.py` misreads the rate limiter's return contract, with two production consequences (verified read-only on `main` @ `454b153ab7`):

1. **Every server-paid `gemini-2.5-pro` request is silently rewritten to `gemini-2.5-flash` from the user's first call of the day.** The Pro "soft quota" is inverted: it downgrades while the user still has quota and only serves Pro once ≤30 daily calls remain.
2. **The daily hard limit (1,500/day) never enforces.** Only the per-minute burst limit works.

InsightAssistant is Pro-primary on this path, so desktop Insights have effectively been served by Flash in production.

## Root cause

`redis_db.check_rate_limit` returns `(allowed, remaining, retry_after)` — the second element is **remaining**, not the used count (`backend/database/redis_db.py` ~L860–877, `remaining = max(0, max_requests - current)`).

`_meter_server_request` (`backend/routers/desktop_proxy.py` ~L383–416) unpacks `_, current, _` and then treats `current` as the used count:

- `if int(current) > _DAILY_HARD_LIMIT:` → compares **remaining** (≤1500) against 1500 → can never fire.
- `if int(current) > soft_limit and ... model == 'gemini-2.5-pro':` → true while **remaining** > 30 → fires for roughly the first 1,470 calls of the day → Pro→Flash rewrite (`record_fallback(component='gemini_model', from_mode='pro', to_mode='flash', reason='quota', outcome='degraded')`).

The unit test encodes the wrong contract: `backend/tests/unit/test_desktop_proxy.py` ~L153–169 mocks the tuple's second element as if it were a count, so the suite passes while production inverts.

Note the soft limit was also always a **shared** counter (`desktop_gemini_daily` — all desktop Gemini traffic including flash/flash-lite/embeddings), not a Pro-only meter; whether that is intended is a policy question for the fix reviewer, but the inversion is a straight bug either way.

## Fix (single PR)

1. In `_meter_server_request`: use `allowed` from the daily call for the hard limit; derive used count as `_DAILY_HARD_LIMIT - remaining` (or extend `check_rate_limit` to also return the raw count) for the soft-limit comparison.
2. Fix `test_desktop_proxy.py` to mock the real `(allowed, remaining, retry_after)` contract.
3. Regression tests (automatic-or-dead):
   - first Pro call of the day is **not** downgraded;
   - Pro **is** downgraded once >30 (or >300 for max tier) daily calls are consumed;
   - request 1,501 in a day is rejected with 429.
4. In the PR description: check fallback telemetry (`component='gemini_model'`, `reason='quota'`) in Sentry/PostHog to size how long and how widely this fired in production.

## Cut list (out of scope)

- Making the soft quota Pro-only vs shared (file a follow-up if the reviewer wants the policy changed).
- Any gateway/lane migration work (tracked in the context-aware-proactivity project).
- Stale `GeminiClient.swift` header comment still referencing the deleted Rust backend (`desktop/macos/Desktop/Sources/ProactiveAssistants/Core/GeminiClient.swift` ~L181) — optional one-line drive-by only if touching adjacent code.

Found during adversarial design review of the desktop proactivity pipelines (independent verification of reviewer claims against `main`).

Git-on-my-level:
## Summary

- honor Redis's daily rate-limit `allowed` result so request 1,501 returns 429
- derive consumed daily requests from the documented `remaining` value before applying the Pro soft quota
- preserve the existing shared daily counter, BYOK bypass, burst limit, embedding exemption, and fallback telemetry
- replace the contract-inverted unit mock with regression coverage for calls 1, 30/31, 300/301, 1,500/1,501, both embedding actions, and the exact fallback fields

Fixes https://github.com/BasedHardware/omi/issues/11423.

## Root cause and impact

`database.redis_db.check_rate_limit` documents and returns `(allowed, remaining, retry_after)`, but `_meter_server_request` treated `remaining` as a consumed count. That made the hard limit unreachable and degraded server-paid `gemini-2.5-pro` requests while quota was still available. The issue's independently verified production report is the external source for the corrected test expectation: https://github.com/BasedHardware/omi/issues/11423.

The fix uses `allowed` as the hard-limit authority and computes `daily_used = 1_500 - remaining` for the unchanged soft-limit policy. Standard users now keep Pro through request 30 and max-tier users through request 300; the next request degrades to Flash. Request 1,500 remains allowed and request 1,501 is rejected.

## Failure class

Failure-Class: none

No registered failure class matches this isolated caller-side misread of an already documented return contract. The behavioral regression tests exercise the production metering seam and pin both sides of every affected boundary.

## Telemetry impact check

Issue #11423 requested a Sentry/PostHog query for `component='gemini_model'` / `reason='quota'`. The live implementation does not send this backend signal to either system: `record_fallback` increments Prometheus `omi_fallback_total` and emits `omi_fallback_event`; production backend PostHog is intentionally disabled. I attempted the production Grafana dashboard on 2026-08-11, but `https://monitor.omi.me/` required credentials unavailable in the current contributor sessions, so I could not responsibly claim a live affected-request count.

The follow-up production query is:

```promql
sum(increase(omi_fallback_total{component="gemini_model",from_mode="pro",to_mode="flash",reason="quota",outcome="degraded"}[30d]))
```

Cloud logs can be cross-checked for `omi_fallback_event component=gemini_model from=pro to=flash reason=quota outcome=degraded`. This PR changes when that existing signal fires; it does not add or rename telemetry.

## Validation

- `cd backend && .venv/bin/pytest -q tests/unit/test_desktop_proxy.py tests/unit/test_rate_limiting.py` — 100 passed
- `black --line-length 120 --skip-string-normalization backend/routers/desktop_proxy.py backend/tests/unit/test_desktop_proxy.py` — clean
- `git diff --check` — clean
- three independent Luna reviews against the original issue, followed by final-diff re-review — no findings

The full `backend/test.sh` runner was also attempted. Under its 864-file parallel load, unrelated fast-duration guards failed and two unrelated files (`test_action_item_vector_best_effort.py`, `test_workstream_router_contract.py`) stalled. The duration-guard files passed when rerun without that load; the two stalled files reproduce independently. The focused affected suites above are green.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11425?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Acknowledged. This is a P3 style preference on test setup deduplication. The boundary tests are already parametrized where the boundary values map naturally (`test_..._keeps_pro_within_the_soft_limit` and `test_..._downgrades_pro_after_the_soft_limit`). Factoring the remaining mock setup into a shared fixture would broaden a focused quota-enforcement bug fix beyond its scope. Happy to revisit if the suite grows more boundary variants.

Git-on-my-level:
Monitor pass for context `714f75f44c`.

- Resolved Codex P2 (non-retryable daily quota): commit `43abd049de` already classifies daily exhaustion via `_GeminiRateLimitExceeded(retryable=False)` and propagates `X-Omi-Retryable: false` + `Retry-After` through `_proxy`. Regression tests pin both headers. Thread resolved.
- Acknowledged Cubic P3 (test setup deduplication): out of scope for this focused quota-enforcement fix; no code change.
- CI: all checks green at `d89b1f7`.

PR #11437, opened 2026-08-12
Fix processing recovery and stale conversation state

Git-on-my-level:
## Summary

- bound hosted VAD connect/read deadlines so an unreachable service falls back to local ONNX before sync jobs go stale
- reconcile conversation Processing cards against authoritative per-ID lifecycle state without reviving cards cleared by websocket completion
- replace misleading upload progress and connectivity copy with localized background-processing guidance and truthful retry semantics

## Root cause

The hosted VAD request inherited a 300-second timeout. When the configured endpoint was unreachable, queued sync jobs spent most of their 600-second stale budget waiting before local VAD could run, and eventually failed without processing a segment. The app then collapsed those failures into a connectivity banner and displayed uploaded/pending WAL counts as if they were server progress.

Conversation Processing cards were held in a separate in-memory list. A missed websocket completion plus the first-page `processing,completed` list filter left no authoritative transition for cards whose server state became `in_progress`, `failed`, deleted, or off-page.

## User impact

- hosted VAD faults degrade to local ONNX after bounded deadlines instead of timing out whole sync jobs
- recordings remain pending with an accurate retry message when upload or processing fails
- sync screens no longer render uploaded/pending queue counts as completion progress
- completed, failed, rolled-back, deleted, and off-page conversations no longer leave permanent Processing cards

## Validation

- `backend/.venv/bin/python -m pytest tests/unit/test_vad_onnx.py tests/unit/test_clean_sweep_migrations.py tests/unit/test_fallback_observability.py -q` — 103 passed
- `(cd app && bash test.sh test/providers/conversation_provider_processing_reconcile_test.dart test/providers/sync_provider_sync_wal_wake_test.dart)` — 16 passed
- `(cd app && scripts/analyze_ratchet.sh)` — passed
- `git diff --check` — passed
- Cursor Codex review — all three findings fixed (localization, 404 coverage, invalid timeout coverage)
- Cursor Grok 4.5 review — both P2 races fixed and regression-tested (websocket removal revival and mixed-status publication)

The original failure was reproduced on a connected physical iPhone and correlated with dev backend job/log evidence. The repaired backend path has not been deployed or re-exercised on-device in this PR.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11437?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
Bounded hosted-VAD deadlines and authoritative Processing-card reconciliation; root cause + regression tests, CI green. 4/5.

PR #11439, opened 2026-08-12
feat(analytics): add Android reliability telemetry

axAilotl:
## Summary

- attach trusted platform, app version/build, and release-channel context to every Flutter analytics event
- make the authoritative connected-to-disconnected transition emit one bounded, content-free `Device Disconnected` event with useful BLE diagnostics
- add deduplicated live transcription errors plus Transcribe Later capture and processing outcomes
- restore the analytics reachability ratchet to the local/CI check manifest so orphaned emitters are caught before charts silently lose their denominator

## Why

The analytics migration left Android pendant reliability, transcription failure, and offline-capture behavior incomplete in PostHog. The earlier reactive reconnect refactor also demonstrated that call-site-owned telemetry can disappear while the product continues working, producing misleading usage charts.

The root cause was split ownership: analytics lived on individual callbacks and UI paths instead of authoritative state transitions, while shared event context and the existing reachability guard were not consistently applied by the Flutter runtime and CI manifest.

## Impact

Android/iOS Flutter events can now be compared by release cohort, Android pendant disconnects have stable operational dimensions, and live/batch transcription failures have bounded outcomes without transcript text, raw error messages, device names, IDs, or file paths. This change only emits telemetry; it does not create or modify PostHog dashboards or alerts.

## Verification

- `bash app/test.sh` — 1,219 tests passed
- `app/scripts/analyze_ratchet.sh` — passed
- `python3 .github/scripts/check_analytics_reachability.py` — passed
- `backend/.venv/bin/python .github/scripts/test_check_analytics_reachability.py` — 7 tests passed
- `backend/.venv/bin/python .github/scripts/test_run_checks.py` — 44 tests passed
- `make preflight` — passed with the repository Python environment on `PATH`

No physical Android/BLE device was available on this macOS host, so the production device path was not exercised locally.

None.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11439?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for the careful telemetry work here. I reviewed the diff as a maintainer automation pass and this looks directionally useful: it moves reliability telemetry to authoritative state transitions, keeps analytics values bounded, and avoids raw provider/device/file details in the new events.

Specific review notes:
- `app/lib/providers/device_provider.dart` moves `Device Disconnected` emission from the callback into `setConnectedDevice(null)`, so duplicate native callbacks should not double count. The helper bounding for reason/app state/RSSI/HCI values is the right shape for privacy-preserving reliability metrics.
- `app/lib/utils/analytics/analytics_manager.dart` adds global event context and lets bootstrap-owned platform/version/release fields override call-site values; that should prevent split mobile dashboards. The new `Transcription Error` and transcribe-later methods also keep property vocabularies closed.
- `app/lib/utils/platform/platform_manager.dart` wires platform/app version/build/release channel before `AnalyticsManager.init()`, which is the right lifecycle point for shared event context.
- `app/lib/services/capture/capture_controller.dart` records terminal live transcription failures once per socket lifecycle and intentionally bounds provider/outcome values; I like that raw `statusText`/exception content is not forwarded.
- `app/lib/providers/local_recordings_provider.dart` derives source from known batch filename families and records user-triggered upload outcomes without exposing file names or backend errors.
- `.github/checks-manifest.yaml` and `.github/scripts/analytics_reachability_baseline.json` re-enable the analytics reachability ratchet in local/CI lanes; because this changes CI/check coverage, I’m flagging it for maintainer awareness rather than treating it as purely app code.
- The added tests in `app/test/providers/capture_provider_test.dart`, `app/test/providers/device_provider_test.dart`, `app/test/unit/analytics_manager_fail_open_test.dart`, and `app/test/unit/local_recordings_analytics_policy_test.dart` cover the key privacy and one-shot emission contracts. `docs/analytics/events.md` matches the implementation claims I checked, including no raw device ID/name/file path/error text.

I did not find a concrete blocker in the sampled code. I’m leaving this for human maintainer review because the PR is still draft and it changes product analytics semantics plus the workflow/check manifest; once it is ready for review, the main maintainer question is whether these reliability/transcription event definitions are the analytics surface Omi wants to standardize on.

Reviewed by Omi maintainer automation (glm-5.2).

---

PR #11448, opened 2026-08-12
fix(macos): route memory export fallbacks to main actor

undivisible:
## Summary

Route memory-export fallback opens to the main actor when NSWorkspace invokes its completion off-thread.

## Sentry issues

- [OMI-DESKTOP-2Z1](https://omi-nk3.sentry.io/issues/7630388038/?project=4511086024851456)
- [OMI-DESKTOP-2YF](https://omi-nk3.sentry.io/issues/7630358206/?project=4511086024851456)

- INV-INT-1

## Failure class

Failure-Class: none

## Verification

- Swift format, test-quality, and diff checks passed.
- Focused XCTest is still compiling in the local SwiftPM cache.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11448?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Low Risk**
> Narrow concurrency fix in export open fallbacks, with a unit test and no auth, data, or architectural changes.
> 
> **Overview**
> Ensures memory-export **fallback opens** hop back to the main actor when `NSWorkspace` completion handlers fire off-thread.
> 
> Adds a `nonisolated` `performOnMainActor` helper on `MemoryExportDestinationSheetModel` and uses it for both branded-app and default-handler fallbacks (with a weak self capture). Includes a focused XCTest that verifies a background callback is rerouted onto the main thread.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 587751dff74eb0c6eda67e2937a66fff10e0ef7b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/MemoryExportDestinationSheet.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/MemoryExportDestinationSheet.swift
@@ -202,7 +202,26 @@ private struct MemoryExportRow: View {
 }
 
 @MainActor
+struct MemoryExportWorkspace {
+  var installedApplicationURL: (ConnectorBrand) -> URL?
+  var defaultApplicationURL: (URL) -> URL?
+  var openWithApplication: ([URL], URL, NSWorkspace.OpenConfiguration, @escaping @Sendable (Error?) -> Void) -> Void
+  var open: (URL) -> Void
+
+  static let live = Self(
+    installedApplicationURL: { $0.installedApplicationURL },
+    defaultApplicationURL: { NSWorkspace.shared.urlForApplication(toOpen: $0) },
+    openWithApplication: { urls, applicationURL, configuration, completion in
+      NSWorkspace.shared.open(urls, withApplicationAt: applicationURL, configuration: configuration) { _, error in
+        completion(error)
+      }
+    },
+    open: { NSWorkspace.shared.open($0) })
+}
+
+@MainActor
 final class MemoryExportDestinationSheetModel: ObservableObject {
+  private let workspace: MemoryExportWorkspace
   @Published var isRunning = false
   @Published var statusMessage: String?
   @Published var errorMessage: String?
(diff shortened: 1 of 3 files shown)


Git-on-my-level (approved):
Reviewed the full diff across all three commits. This is a clean, well-targeted concurrency fix.

MemoryExportDestinationSheet.swift — the core fix is sound. `performOnMainActor` is correctly declared `nonisolated static` so it can be called from the NSWorkspace completion (which may fire off the main actor), and it dispatches via `Task { @MainActor in }`. Both fallback paths now route through it: the `openInDefaultHandler(url)` retry in the branded-app path and the `workspace.open(url)` final fallback in the default-handler path.

The `[self]` capture in the branded-app completion is the right call over `[weak self]` — it ensures the model stays alive long enough for the fallback retry to fire rather than silently dropping it if the sheet is dismissed first. No retain cycle: `MemoryExportWorkspace` is a value-type struct so the `[workspace]` capture in `openInDefaultHandler` copies by value, and the NSWorkspace completion is one-shot.

The `MemoryExportWorkspace` seam is idiomatic — closure-backed struct with a `.live` default, injectable for tests. Good design choice that keeps the production path unchanged while making the async fallback chain fully testable.

MemoryExportSetupTests.swift — the rewritten test exercises the real `openDestination -> openInDefaultHandler -> fallback open` path end-to-end, simulates an off-thread NSWorkspace completion via `DispatchQueue.global().async`, asserts the fallback `open` runs on the main thread, verifies both app-launch attempts occurred, and checks the model doesn't leak. This is genuine behavioral coverage of the bug, not a static tripwire.

Changelog entry follows the documented format and the user-visible message is accurate.

INV-INT-1 citation is appropriate (path glob matches `MemoryExport*`); the change is a pure concurrency fix and doesn't violate any invariant MUST NOT rules.

Review by glm-5.2 (Omi maintainer automation).

---

PR #11451, opened 2026-08-12
fix(desktop): make admission waiter handoff cancellation-safe

undivisible:
## Summary

Prevent a cancelled context-admission waiter from resuming twice after its handoff.

## Sentry issue

- [OMI-DESKTOP-3ZA](https://omi-nk3.sentry.io/issues/7662129686/?project=4511086024851456)

- INV-CHAT-1

## Failure class

Failure-Class: none

## Verification

- Swift format, parse, test-quality, and diff checks passed.
- Focused XCTest is still compiling in the local SwiftPM cache.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches concurrency control for context admission used by chat/voice refresh paths. The change is narrowly scoped and covered by a race-focused test, but incorrect waiter handoff could still stall or crash admission.
> 
> **Overview**
> Prevents a **double-resume crash** in `AgentContextAdmissionGate` when a waiter is cancelled during grant handoff (e.g. while refreshing live voice context).
> 
> Waiter state is now lock-protected on `WaiterHandle`, with cancel/register/take as atomic operations so grant and cancel cannot both resume the same continuation. Cancelled waiters also flush a pending release so the queue keeps moving.
> 
> Adds a focused race test that pauses mid-handoff and cancels before resume.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c8e39342b51e60c9e510026ab9765f5f6c22ad4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/desktop/macos/Desktop/Sources/Chat/AgentContextAdmissionGate.swift
+++ b/desktop/macos/Desktop/Sources/Chat/AgentContextAdmissionGate.swift
@@ -110,7 +110,7 @@ actor AgentContextAdmissionGate {
     case .granted:
       return true
     case .cancelled:
-      finalizeCancelledWaiter(at: slotIndex)
+      finalizeCancelledWaiter(handle: handle)
       return false
     }
   }
--- a/desktop/macos/Desktop/Sources/Chat/AgentContextAdmissionGate.swift
+++ b/desktop/macos/Desktop/Sources/Chat/AgentContextAdmissionGate.swift
@@ -17,8 +17,40 @@ import Foundation
 ///   serialized; the query stream itself runs outside the gate.
 actor AgentContextAdmissionGate {
   private final class WaiterHandle: @unchecked Sendable {
-    var cancelled = false
-    var continuation: CheckedContinuation<WakeReason, Never>?
+    private let lock = NSLock()
+    private var cancelled = false
+    private var continuation: CheckedContinuation<WakeReason, Never>?
+
+    func cancel() -> CheckedContinuation<WakeReason, Never>? {
+      lock.lock()
+      defer { lock.unlock() }
+      cancelled = true
+      let continuation = continuation
+      self.continuation = nil
+      return continuation
+    }
+
+    func register(_ continuation: CheckedContinuation<WakeReason, Never>) -> Bool {
+      lock.lock()
+      defer { lock.unlock() }
+      guard !cancelled else { return false }
+      self.continuation = continuation
+      return true
+    }
+
+    func isCancelled() -> Bool {
+      lock.lock()
+      defer { lock.unlock() }
+      return cancelled
+    }
+
+    func takeContinuation() -> CheckedContinuation<WakeReason, Never>? {
+      lock.lock()
+      defer { lock.unlock() }
+      let continuation = continuation
+      self.continuation = nil
+      return continuation
+    }
   }
 
   private enum WakeReason {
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Thanks for this fix — the analysis and implementation are solid.

**Root cause verified.** The original `WaiterHandle` stored `cancelled` and `continuation` as non-atomic `var`s accessed from both the actor thread (`release()`, `registerWaiter()`) and the `onCancel` handler (arbitrary thread). More critically, the continuation was duplicated across `WaiterHandle.continuation` and `WaiterSlot.continuation`, so `release()` could resume via the stale `WaiterSlot` copy while `onCancel` simultaneously resumed via `WaiterHandle` — a classic double-resume race that matches the Sentry crash (OMI-DESKTOP-3ZA).

**The fix is correct and well-structured.** Moving to an `NSLock`-protected `WaiterHandle` with atomic claim-and-nil operations makes the mutual exclusion airtight:

- `cancel()` and `takeContinuation()` each atomically extract-and-nil the continuation, so exactly one path can resume it — no double-resume possible regardless of interleaving.
- `register()` guards against the cancel-before-register window: if cancel ran first, register returns false and the caller immediately resumes with `.cancelled`.
- Eliminating the duplicate `WaiterSlot.continuation` removes the stale-copy bug at its source.
- The `isCancelled()` re-check after a nil `takeContinuation()` in `release()` correctly distinguishes "cancel already claimed it" (skip) from "continuation not registered yet" (set `releasePending` and wait).

**One detail worth noting in `finalizeCancelledWaiter`.** The new `releasePending` flush (lines 150-154) is a good addition — without it, a cancelled head waiter holding a pending release could stall the queue. The lookup-by-identity (`firstIndex(where: { $0.handle === handle })`) is robust against compaction shifting indices, which is safer than the old index-based approach.

**Regression test quality.** `testContextAdmissionGateCancellationAfterGrantResumesWaiterOnce` deterministically reproduces the exact race using the `onWaiterRegistered`/`beforeResumingGrantedWaiter` hooks to pause the handoff at the vulnerable point, then cancels and verifies no crash plus continued gate usability. The synchronous semaphore block in `pauseGrantedHandoff()` is appropriate for test-only synchronization hooks (production uses nil callbacks). The `GateGrantCancellationRaceProbe` probe class is cleanly isolated with its own lock and properly handles the async coordination.

**Changelog.** The fragment accurately describes the user-facing impact.

Not formally approving because `Desktop Swift Static & Test Contracts` is still running. The release compile already passed, and the implementation review is clean.

---

Git-on-my-level (approved):
Approving. Well-scoped fix for a double-resume crash in the admission gate's waiter handoff.

AgentContextAdmissionGate.swift — WaiterHandle state is now NSLock-protected with atomic single-owner handoff. cancel() and takeContinuation() both nil the continuation under the lock, so exactly one wins — the continuation can only be resumed once. register() rejects an already-cancelled handle, closing the cancel-then-register window. The continuation was removed from WaiterSlot and lives only in the handle, eliminating the dual-reference that made the old race possible. The re-check in release() (isCancelled() after takeContinuation() returns nil) correctly distinguishes "not yet registered" from "cancelled meanwhile," preventing a queue stall. The releasePending flush in finalizeCancelledWaiter keeps the queue moving when a head waiter is cancelled.

AgentRuntimeProcessTests.swift — testContextAdmissionGateCancellationAfterGrantResumesWaiterOnce uses the new init hooks to pause release() between takeContinuation() and resume(), cancels the waiter in that window, then unblocks. This exercises the exact race the fix addresses and verifies the continuation resumes exactly once. GateGrantCancellationRaceProbe is a clean lock-protected coordination harness.

20260812-fix-admission-handoff.json — accurate, user-facing changelog fragment for the crash fix.

The test-injection hooks default to nil and the production call site (AgentClient.Session) uses the default init, so no test coupling reaches production.

Review by glm-5.2 (automated maintainer review).

---

PR #11459, opened 2026-08-12
fix(desktop): index the turn-revision lookup so chat context snapshots finish

formed2forge:
## Problem

Desktop Ask AI answered **nothing**. Every turn returned "Omi couldn't answer this one. Try sending it again." and the model was never reached.

```
kernel contract timeout operation=get_context_snapshot elapsed_ms=15013
ChatProvider: kernel context preparation timed out before an agent query started
chat_agent_error duration_ms=15142 error_class=timeout partial_response=false watchdog_fired=false
```

Field counts from the affected machine: 8 context-preparation timeouts, **0** agent-query timeouts, **0** completed turns. 100% of turns died in setup.

## Cause

Each turn waits on the kernel's `get_context_snapshot`, budgeted at 15s by `AgentRuntimeKernelContractTimeoutPolicy`. `buildContextSnapshot` builds its recent-turn window with:

```sql
FROM conversation_turns ct
LEFT JOIN conversation_turn_revisions revision
  ON revision.conversation_id = ct.conversation_id
 AND revision.turn_id = ct.turn_id       -- no index carries turn_id
WHERE ct.conversation_id = ?
GROUP BY ct.conversation_id, ct.turn_id
ORDER BY ct.created_at_ms DESC, insertion_seq DESC
LIMIT ?
```

`conversation_turn_revisions` is keyed `PRIMARY KEY(conversation_id, turn_seq)` and carried **no index on `turn_id`**, so the join could only seek on `conversation_id` and then walked every revision row of the conversation for every turn in it — quadratic in conversation length. `GROUP BY` plus `ORDER BY insertion_seq` also prevents `LIMIT` from short-circuiting (`USE TEMP B-TREE FOR ORDER BY`).

That query runs synchronously through `node:sqlite` on the kernel's event loop, so the entire runtime stalls while it scans. The trivial `get_context_snapshot` handler cannot reply until it finishes, the Swift side abandons the turn at 15s, and the runtime's eventual answer is discarded as a late contract response.

This is why "fresh machine, long-lived account" was the shape that tripped it: cost scales with turns × revisions.

## Fix

One additive index. Migration 33, following the existing `hasMigration` / `runTransaction` pattern.

```sql
CREATE INDEX IF NOT EXISTS conversation_turn_revisions_turn_id_idx
  ON conversation_turn_revisions(conversation_id, turn_id);
```

## Verification

Measured on a real 35 MB database with 7,263 turn revisions in one conversation:

| | plan | time |
|---|---|---|
| before | `SEARCH revision USING INDEX sqlite_autoindex_conversation_turn_revisions_1 (conversation_id=?)` | **27.65 s** |
| after | `SEARCH revision USING INDEX conversation_turn_revisions_turn_id_idx (conversation_id=? AND turn_id=?)` | **0.03 s** |

Index builds in 0.03 s.

- `tsc --noEmit` clean; `npm run build` succeeds; `vitest run` → **58 files, 671 tests, all passing**
- **Negative check:** with the migration call removed, all 3 new tests fail, including `expected 'SEARCH revision USING INDEX sqlite_autoindex_...' to contain 'conversation_turn_revisions_turn_id_idx'`
- **Upgrade path:** applied to a pre-existing 7,291-revision database (what a shipped user has), plan became the two-key seek, query timed 0.043 s on the live file
- **Real path:** rebuilt dev bundle, one query → `chat_agent_query_completed duration_ms=3695`, reaching `APIClient: Quota plan=Operator unit=questions allowed=true`, which no previous turn ever got to
- **Sustained:** a combined build ran 11 turns — 9 completed, **0** `kernel contract timeout`, **0** `dropping late kernel`

Root cause was located with a Node inspector CPU profile of the live kernel (`SIGUSR1` + CDP `Profiler`): 20.9% of samples under `next(native) ← allRows ← buildContextSnapshot ← contextSnapshot ← get_context_snapshot`, with journal replay at 0.1%.

## Tests

Assert the **plan and the migration**, not a wall-clock threshold — a timing assertion on shared CI hardware is the flaky kind this suite is told to delete. The plan is the real contract: an equality seek on both join keys.

1. The revision step names the index and `turn_id=?`
2. Snapshot output is identical with and without the index (protects the chronology contract documented at the query)
3. The migration creates the index on a database that predates it, then reopens idempotently — including the already-migrated principal, which must not re-run or fail

## Known limit

The query still aggregates every turn in the conversation, so cost stays linear in conversation length (0.04 s at ~7,300 turns). Restructuring it to apply `LIMIT` before aggregating is riskier than this index because of the chronology semantics documented at the query, so that is tracked separately rather than bundled here.

- INV-AGENT-* — *Omi owns session/run/attempt identity, authority, and lifecycle.* Matched via `desktop/macos/agent/src/runtime/sqlite-store.ts`. No guard-test update: this adds an index and changes no ownership, identity, authority or lifecycle behaviour.
- INV-CHAT-1 — *Every canonical conversation has one kernel-owned transcript.* Matched via the same file. No guard-test update: the transcript query's results and their order are unchanged — one of the new tests asserts snapshot output is identical with and without the index, precisely so the transcript contract is provably untouched.

Both are cited on the path-based rule. The change is additive to the schema and behaviour-preserving by construction, which is what the identical-output test exists to demonstrate.

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11459?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-deadline-bounded-query-without-covering-index.json
+++ b/.github/failure-classes/FC-deadline-bounded-query-without-covering-index.json
@@ -1,0 +1,18 @@
+{
+  "schema_version": 1,
+  "id": "FC-deadline-bounded-query-without-covering-index",
+  "violated_contract": "A query that serves a deadline-bounded interactive contract must be covered by an index on every predicate it joins and filters on. An uncovered join degrades with accumulated user data, so a contract that passes on a small store becomes a guaranteed timeout on a large one \u2014 and when the query runs synchronously on a single-threaded runtime, it stalls every other request behind it rather than only itself.",
+  "canonical_prevention": "Declare the index in the same change as the query, and assert the plan in a behavioural test that runs the real engine: the query plan must show an equality seek on each join key, never a partial seek that leaves a predicate to a row-by-row scan. Assert the plan rather than a wall-clock threshold, which is unstable on shared hardware and passes on small fixtures for the wrong reason.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/agent/tests/context-snapshot-turn-revision-index.test.ts"
+  ],
+  "evidence_prs": [
+    9597,
+    9660
+  ],
+  "scope_hints": [
+    "desktop/macos/agent/src/runtime/**",
+    "desktop/macos/agent/tests/**"
+  ],
+  "status": "open"
+}
--- a/desktop/macos/agent/src/runtime/sqlite-store.ts
+++ b/desktop/macos/agent/src/runtime/sqlite-store.ts
@@ -80,6 +80,7 @@ const CHAT_FIRST_MATERIALIZATION_RECEIPTS_MIGRATION_VERSION = 29;
 const CHAT_FIRST_COLD_START_SEQUENCE_RECEIPTS_MIGRATION_VERSION = 30;
 const LOCAL_ONLY_JOURNAL_DELIVERY_MIGRATION_VERSION = 31;
 const CHAT_FIRST_COLD_START_SEQUENCE_RECEIPTS_OWNER_SCOPE_MIGRATION_VERSION = 32;
+const CONVERSATION_TURN_REVISION_TURN_ID_INDEX_MIGRATION_VERSION = 33;
 
 const ACTIVE_ATTEMPT_STATUSES = ["queued", "starting", "running", "waiting_input", "waiting_approval", "cancelling"] as const;
 const TERMINAL_ATTEMPT_STATUSES = ["succeeded", "failed", "cancelled", "timed_out", "orphaned"] as const;
@@ -548,6 +549,9 @@ export class SqliteAgentStore implements AgentStore {
     }
     if (!this.hasMigration(CHAT_FIRST_COLD_START_SEQUENCE_RECEIPTS_OWNER_SCOPE_MIGRATION_VERSION)) {
       runChatFirstColdStartSequenceReceiptsOwnerScopeMigration(this.db, this.nowMs());
+    }
+    if (!this.hasMigration(CONVERSATION_TURN_REVISION_TURN_ID_INDEX_MIGRATION_VERSION)) {
+      runConversationTurnRevisionTurnIdIndexMigration(this.db, this.nowMs());
     }
   }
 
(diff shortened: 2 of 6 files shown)


Git-on-my-level (commented):
Solid fix. Verified each changed file against main:

- `sqlite-store.ts` — Migration #33 adds `CREATE INDEX conversation_turn_revisions_turn_id_idx ON conversation_turn_revisions(conversation_id, turn_id)`. This directly covers the join in `buildContextSnapshot` (`context-snapshot.ts:226-228`: `ON revision.conversation_id = ct.conversation_id AND revision.turn_id = ct.turn_id`) and the `EXISTS` subquery in the downgrade repair (`sqlite-store.ts:1840`). The table PK is `(conversation_id, turn_seq)`, which doesn't carry `turn_id` as a secondary seek key, so without this index the join walked every revision row per turn — the quadratic diagnosis is correct. `IF NOT EXISTS` + `schema_migrations` version-33 gating makes it idempotent, and the `runTransaction` wrapper matches the established migration pattern.
- `context-snapshot-turn-revision-index.test.ts` — Asserting `EXPLAIN QUERY PLAN` shows an equality seek on both keys (`conversation_turn_revisions_turn_id_idx` + `turn_id=?`) rather than a wall-clock threshold is exactly the right contract for a shared-CI suite. The drop-and-reproject result/order-equivalence check and the pre-migration-reopen + idempotency case are both valuable.
- `sqlite-store.test.ts` / `workstream-continuity.test.ts` — Migration count bumped 32->33, correct.
- `FC-deadline-bounded-query-without-covering-index.json` — Accurate failure-class record; the prevention-artifact pointer and scope hints match the actual fix.
- `changelog/...json` — Matches the user-visible symptom (Ask AI answering nothing on long-history accounts).

No query result or ordering changes, no data rewrite — purely additive index. Relevant desktop checks pass. Great write-up.

Leaving formal approval to a maintainer; the change itself looks correct and low-risk.

---

kodjima33 (approved):
Missing composite index on conversation_turn_revisions(conversation_id, turn_id) made the context snapshot quadratic and blew the 15s kernel budget. Additive index + migration follows the existing pattern, with a regression test.

kodjima33 (approved):
Missing composite index on conversation_turn_revisions(conversation_id, turn_id) made the context snapshot quadratic and blew the 15s kernel budget; additive index + migration follows the existing pattern with a regression test.

PR #11460, opened 2026-08-12
fix(desktop): keep deleted-lane tasks retired so they stop resurfacing as live

formed2forge:
## Problem

Tasks deleted on another device came back as **live** tasks on a fresh machine, and the Removed list stayed empty:

```
TasksStore: Fetched 100 deleted tasks from API
TasksStore: Showing 0 deleted tasks from merged local cache
```

## Cause

`TaskActionItem.isRetired` re-derives retirement from the response body (`deleted == true`, or a canonical status of `cancelled`/`superseded`). Its own doc comment records that list responses may omit legacy `deleted` and project retirement through canonical lifecycle status — and that status vocabulary is **open**, so any other value reads as live. `ActionItemRecord.from(_:)` persists `deleted: item.isRetired`, so a row arriving from the deleted lane without either marker was written to SQLite as live: absent from the `.deleted` scope query, present as a live task.

## Fix

The lane is the authority on retirement, not a re-derivation from whichever fields a response happened to carry. `fetchDeletedPage` stamps retirement after **both** transports — live API and injected operations — so neither path can skip it, and both callers of that page (first load and auto-refresh) sync a correctly retired row.

## Verification

macOS 27.0, Xcode 26.6 (stable, not the beta):

- `xcrun swift test --filter TasksStoreDeletedLaneRetirementTests` → **2 passed**
- Reverting both production files to the parent commit → **2 failed**, asserting the reported symptom: `XCTAssertEqual failed: ("[]") is not equal to ("["deleted-on-phone"]")`
- All **87** tests matching `TasksStore` pass together, so nothing leaks through the `TasksStore.shared` singleton
- `swift-format lint --strict` clean on all three files (linter confirmed live by checking it flags a deliberately malformed file)
- `desktop/macos/scripts/check-main-actor-xctest-hooks.py` passes
- Product-file line-count ratchet passes with the baseline raised 3688 → 3697

**App-level:** a combined build carrying this fix was run as the named bundle `omi-combined-verify`. The deleted lane now reports `Fetched 100 deleted tasks / Showing 20 deleted tasks` — `Showing 0` against a non-zero fetch was the entire symptom, so this is the first run with a populated Removed list. The live list still sheds rows as the lane pages through (581 → 547 → 513 → 483, `added=0`), which is now the correct outcome: those rows land in the `.deleted` scope instead of vanishing.

## Scope

Fixes retirement **state**, not refresh cadence — the deleted refresh still re-fetches and re-syncs a full page per tick. Tracked separately. Also corrects the write path only: rows already persisted as live by an affected build are reconciled by later refreshes, not rewritten.

Explicitly **not** a contributor to the desktop chat outage — that was measured to a missing index on `conversation_turn_revisions(conversation_id, turn_id)` and is fixed on its own branch.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11460?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-stores.json
+++ b/.github/scripts/product_file_line_count_ratchet_baseline/desktop-swift-stores.json
@@ -1,9 +1,9 @@
 {
   "files": {
-    "desktop/macos/Desktop/Sources/Stores/TasksStore.swift": 3688
+    "desktop/macos/Desktop/Sources/Stores/TasksStore.swift": 3697
   },
   "raise_justifications": {
-    "desktop/macos/Desktop/Sources/Stores/TasksStore.swift": "Suggestion triage: AI captures excluded from dashboard lanes, acceptSuggestedTask"
+    "desktop/macos/Desktop/Sources/Stores/TasksStore.swift": "Deleted-lane retirement authority: fetchDeletedPage stamps retirement after both transports so neither can write a retired row as live"
   },
   "threshold": 1500
 }
--- a/desktop/macos/Desktop/Sources/Stores/TasksStore.swift
+++ b/desktop/macos/Desktop/Sources/Stores/TasksStore.swift
@@ -1405,17 +1405,26 @@ class TasksStore: ObservableObject {
     operations: OwnerBoundOperations
   ) async throws -> OwnerBoundOperations.ActionItemsPage {
     guard isCurrent(lease) else { throw LocalMutationAuthorizationError.revoked }
+    let page: OwnerBoundOperations.ActionItemsPage
     if let fetchDeletedPage = operations.fetchDeletedPage {
-      return try await fetchDeletedPage(offset, limit, lease.ownerID)
+      page = try await fetchDeletedPage(offset, limit, lease.ownerID)
+    } else {
+      let response = try await APIClient.shared.getActionItems(
+        limit: limit,
+        offset: offset,
+        deleted: true,
+        expectedOwnerId: lease.ownerID,
+        authorizationSnapshot: lease.authorizationSnapshot
+      )
+      page = .init(items: response.items, hasMore: response.hasMore)
     }
-    let response = try await APIClient.shared.getActionItems(
-      limit: limit,
-      offset: offset,
-      deleted: true,
-      expectedOwnerId: lease.ownerID,
-      authorizationSnapshot: lease.authorizationSnapshot
-    )
-    return .init(items: response.items, hasMore: response.hasMore)
+    // The lane is the authority on retirement, not `isRetired`'s re-derivation
+    // from whatever fields this response happened to carry. Stamping it here —
+    // after both transports, so neither can skip it — is what stops a retired
+    // row being written to the local cache as live and resurfacing as a live
+    // task. Every caller of this page (first load and auto-refresh) syncs it
+    // into SQLite, so normalizing anywhere later would leave one path wrong.
+    return .init(items: page.items.map { $0.retired() }, hasMore: page.hasMore)
   }
 
   private func syncPage(
--- a/desktop/macos/Desktop/Sources/TaskActionItem.swift
+++ b/desktop/macos/Desktop/Sources/TaskActionItem.swift
@@ -22,7 +22,7 @@ struct TaskActionItem: Codable, Identifiable, Equatable {
   /// Classification category: personal, work, feature, bug, code, research, communication, finance, health, other
   let category: String?
   /// Soft-delete: true if this task has been deleted by AI dedup
-  let deleted: Bool?
+  var deleted: Bool?
   /// Who deleted: "user", "ai_dedup"
   let deletedBy: String?
   /// When the task was soft-deleted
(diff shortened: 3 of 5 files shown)


kodjima33 (approved):
Deleted-lane rows persisted as live because retirement was re-derived from an open status vocabulary; stamping retirement at the lane is the right authority. Root cause + regression test + revert-proof verification.

formed2forge:
Both review findings were investigated against current code. **Neither requires a change to this PR**, and no code was changed. Evidence below, plus two genuine defects found in passing that are filed separately.

## Finding 1 — `retired()` leaves `deleted` nil for status-only rows: **harmless**

True that the `guard !isRetired else { return self }` returns early, leaving `deleted` nil when a lane row is retired only via `taskStatus`. It cannot produce the stated consequence, for two independent reasons.

**Every writer that persists a `TaskActionItem` derives the stored column from `isRetired`, never from `item.deleted`:**

```
Rewind/Core/ActionItemModels.swift:323    deleted: item.isRetired,
Rewind/Core/ActionItemModels.swift:376    self.deleted = item.isRetired
Rewind/Core/ActionItemModels.swift:750    deleted: item.isRetired,
Rewind/Core/ActionItemStorage.swift:744   let incomingDeleted = item.isRetired
```

The early return fires *only* when `isRetired` is already true — i.e. precisely the cases those writers already resolve to `true`. `ActionItemRecord.deleted` is non-optional `Bool`, so such rows land in SQLite as `deleted = 1`. `taskStatus` also survives the round-trip (`RewindDatabase.swift:2535` column; written `ActionItemModels.swift:335`/`:387`, read `:489`), so a cached row satisfies both projections.

**The one un-stamped in-memory object is contained.** It reaches published state only via the sync-failure fallback `deletedTasks = response.items` (`Stores/TasksStore.swift:2093`), and `deletedTasks` is read only inside `TasksStore.swift`. `TasksPage`'s in-memory sources are `completedTasks`/`incompleteTasks` (`TasksPage.swift:2223-2225`); its Removed lanes are SQLite-backed.

I did try the suggested normalisation — dropping the guard so the lane always stamps — and reverted it: `testAlreadyRetiredRowIsPassedThroughUnchanged` asserts *"an already-retired row must round-trip untouched"*, and the custom `==` compares `.deleted` (`TaskActionItem.swift:113`). Making the change would mean rewriting an existing guard with no external source for the new expected value, to fix nothing observable.

## Finding 2 — lane stamps only `deleted`, provenance left nil: **mechanism correct, attributed harm does not exist**

The mechanism is verified exactly as described: `TasksPage.swift:200-201` partitions on `deletedBy == "user"` (nil ⇒ `removedByAI`); `ActionItemStorage.swift:1420-1432` filters `AND deletedBy = ?` (never matches NULL); `TaskAssistant.swift:1588` queries `deletedBy: "user"` — note that file is under `TaskExtraction/`, not `TaskAgent/`.

But the harm cannot occur, because **nothing in the macOS target ever writes `deletedBy = "user"`**. `ActionItemStorage.deleteActionItemByBackendId` accepts a `deletedBy` argument, never reads it, and hard-deletes the row (`ActionItemStorage.swift:1186-1207`). So `removedByMe` is structurally empty and that query structurally returns `[]` — with or without this PR. No task can lose provenance it never had.

Three further corrections:

- `retired()` does not drop provenance. `var copy = self` carries `deletedBy` through unchanged; nil-ness originates at decode, upstream of the lane.
- `deletedAt` and `deletedReason` are never persisted for action items at all — hardcoded nil at `ActionItemModels.swift:481-482`. Two of the three named fields are structurally unpersistable, independent of this PR.
- Nil-provenance rows in Removed are **pre-existing**: rows carrying `cancelled`/`superseded` were already `isRetired` before this fix, already landed in Removed, already had NULL provenance. The only newly-reachable population is rows with neither marker — and the sole in-tree server retirement path (`retire_action_items_for_conversation`) is genuinely system-side, so "Removed by AI" is the correct bucket for them.

Stamping provenance in the lane would fabricate a value the server never sent, and would require changing three `let` declarations to `var` to do it. Not done.

One limit stated plainly: whether the deployed server omits `deleted_by` could not be proven from this repo — the in-tree `ActionItemResponse` (`backend/models/action_item.py:188-219`, `extra='ignore'`) cannot carry it and the backend never writes it, but the deployed backend is ahead of this tree (it honours a `deleted` query param this tree lacks). The conclusion does not depend on that, since it rests on the missing client-side writer.

## Filed separately

- #11517 — "Removed by me" is structurally empty: `deleteActionItemByBackendId` ignores its `deletedBy` argument and hard-deletes, while its caller's comment claims a soft delete. Also starves `TaskAssistant`'s user-deletion list, so extraction can keep re-suggesting tasks the user removed.
- #11518 — ~32 raw `TaskActionItem.deleted` reads disagree with the ~20 `isRetired` reads, so a status-only-retired task reads as live on objects that have not round-tripped through SQLite (e.g. `ChatFirstShell.swift:569` returns a cancelled task as live). Same failure class as this PR, on paths the deleted lane never touches.

PR #11476, opened 2026-08-13
fix(desktop): fail dev-feedback when a test filter matches zero tests (#11441)

### LINKED ISSUE #11441: swift test --filter matching zero tests exits 0, so dev-feedback.py reports PASS on a run that executed nothing
`swift test --filter` exits **0** when the filter matches zero tests, so a renamed, mistyped, or deleted filter reports a green run instead of an error. `desktop/macos/scripts/dev-feedback.py` decides pass/fail from the exit code (`scripts/dev-feedback.py:190-194`, `if result.returncode == 0: emit("... PASS")`), so the documented focused-feedback workflow prints `PASS` for a run that executed nothing.

Measured on macOS 27.0 / Xcode 26.6 / Swift 6.3.3, in `desktop/macos/Desktop`:

```
$ xcrun swift test --filter ThisClassDoesNotExistAnywhere_XYZ ; echo $?
	 Executed 0 tests, with 0 failures (0 unexpected) in 0.000 seconds
✔ Test run with 0 tests in 0 suites passed after 0.001 seconds.
0

$ xcrun swift test --filter TasksStoreDeletedLaneRetirementTests ; echo $?
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.046 seconds
0
```

Both exit 0 — a real run and a run that matched nothing are indistinguishable by exit code.

The trailing `✔ Test run with 0 tests in 0 suites passed` line makes it worse to eyeball: that summary comes from swift-testing, which has nothing to run in this XCTest-only package, and it prints **after** XCTest's real result. Reading the tail of the output — which is what a human does — shows a green "0 tests passed" regardless of whether the actual tests passed, failed, or never matched.

Encountered live: while verifying a filtered regression class on this package, the tail alone read as a pass with zero tests, and only the earlier XCTest section showed the 2 tests had genuinely run.

No existing issue found for this (searched `swift test filter 0 tests`).

kodjima33:
## Summary

Fixes #11441. `swift test --filter` exits **0** when the filter matches zero tests, and `desktop/macos/scripts/dev-feedback.py` decided pass/fail purely from that exit code — so a renamed, mistyped, or deleted filter printed `PASS` for a run that executed nothing. The trailing swift-testing line (`✔ Test run with 0 tests in 0 suites passed`) makes it read green to a human too.

## Root cause

`emit_iteration_result` ran the command with `subprocess.run` (no output captured) and reported `PASS` whenever `returncode == 0`. A zero-match filtered run is indistinguishable from a real one by exit code alone.

## Fix

- `run_streaming` replaces `subprocess.run` as the default runner: it echoes the test output live (the watch loop still streams) while retaining it.
- A swift iteration must now prove coverage: a green exit with a reported count of zero fails as `FAIL (filter '...' matched 0 tests)`, and a green exit with **no** reported count at all fails as `FAIL (no executed-test count reported; coverage unproven)` — an unrecognized summary can never fall through to PASS (review follow-up). Any non-zero reported count still passes, so an XCTest-only package whose swift-testing section reports `0 tests` is unaffected. pytest keeps deciding on its exit code (it exits 5 when nothing is collected).
- Regression tests in `desktop/macos/tests/test-dev-feedback.py` plus the matching doc lines in `desktop/macos/AGENTS.md` and `desktop/macos/README.md`.

## Failure-Class

Failure-Class: FC-nonexecuting-verification-command

The violated contract is exactly this class: a registered verification command must execute the cases it claims to verify, and an invocation that can exit 0 without evaluating a single assertion must never report success.

none

## Test Plan

- [x] `desktop/macos/tests/test-dev-feedback.sh` — 10 tests OK. Mutation-verified: the three new tests error against the pre-fix `dev-feedback.py` from `origin/main`.
- [x] Real path, real SwiftPM (`--once swift` against a real XCTest package):
  - non-matching filter → `Iteration 1: FAIL (filter 'ThisClassDoesNotExistAnywhere_XYZ' matched 0 tests)`, **exit 1** — the same invocation on `main` prints `Iteration 1: PASS`, **exit 0**.
  - matching filter → `Executed 1 test, with 0 failures` then `Iteration 1: PASS`, **exit 0**; the swift output streamed live as before.
- [x] `make preflight`


kodjima33:
Valid catch — fixed in 060dfac147. An unmatched summary no longer falls through: a swift iteration that exits 0 without a reported executed-test count now fails with `no executed-test count reported; coverage unproven`, so only a positive count passes. pytest still decides on its exit code (5 = nothing collected), covered by a new test.

PR #11524, opened 2026-08-13
harden(ci): match deepPurple and the Dart hex form in the INV-UI-1 ratchet, and clear the apps pages

mikejsmith1985:
Two spellings of purple that the Flutter app uses fall just outside the INV-UI-1 ratchet's
current patterns. This adds them, then clears the purple those patterns bring into view on
the apps pages. Offered as a contribution — take either commit, both, or neither.

## The two patterns

- **`deepPurple`.** `\bpurple\b` cannot match inside `deepPurple`, because camelCase leaves
  no word boundary before it. `Colors.deepPurple`, `Colors.deepPurpleAccent` and
  `Colors.deepPurple.shade300` therefore count as zero. **183 references** in `app/lib`.
- **The Dart hex form.** The hex pattern is written for `#RRGGBB`, which is the right shape
  for the web and Swift surfaces. Flutter writes `Color(0xFF8B5CF6)`, so the Dart literals
  sit outside it — **60 references**, including the star on every app row
  (`app/lib/pages/apps/list_item.dart:105`).

Counting `app/lib` with the patterns as they stand: **118 hits in 29 files**. With the two
added: **357 in 94 files**.

This is not a new guard, just the existing one taught two more spellings. Both patterns cite
instances in the tree today rather than a hypothesis.

**It cannot fail an open PR.** The ratchet counts base and head with the same script, so both
sides move together and every existing comparison is unchanged. What it adds is coverage for
the next one.

## Then the apps pages

49 references by the ratchet's new count, now 0. Replaced by role, because contrast decides
what "neutral" means in each place:

| Role | Was | Now |
|---|---|---|
| Fill behind white text (selected pills, chips, badge) | purple | white @ alpha 0.22 |
| Solid primary action (submit review) | purple bg, white label | white bg, black label |
| Icon or border on dark (stars, ticks, outlines) | purple | white |
| Pull-to-refresh arc | purple | **black** — it is drawn on an explicitly white `backgroundColor` |
| Radio tick | white on purple | black on white |

Two are judgement rather than rule, and easy to change if you would prefer something else:

- The **AI generator banner** gradient ended on violet; it now ends on a deeper indigo,
  keeping the banner's identity. Indigo is not purple and the project already uses `#4f46e5`
  elsewhere.
- The **per-category hue palette** had purple for `entertainment` and deepPurple for `music`.
  Those become `pinkAccent` and `tealAccent`, following the palette's existing habit of
  pairing a base hue with an accent variant. Both are arbitrary; only "not purple" is
  required.

Replacements were applied by an explicit mapping asserting how many times each must match, so
a miss or an over-eager match failed loudly. It did: a bare `color: Colors.deepPurple,` also
matched inside a `TextStyle`, caught before it could recolour something unintended.

## Verification

Against `main` at `8ddcff958`.

- `check_brand_ui.py` vs `origin/main` — *"no purple increase across 15 changed UI file(s)"*
- ratchet count under `app/lib/pages/apps`: **49 → 0**; a raw case-insensitive grep for
  `purple` over the same tree also returns 0
- `python -m unittest test_check_brand_ui` — 9 tests pass, including three new for
  `deepPurple`, two for the Dart hex form, and one asserting a non-purple hue sharing hex
  digits is *not* counted
- `flutter test` — 1221 pass, 2 fail (`plans_sheet_l10n`, `voice_recorder_mic_contention`).
  Both fail identically with this work reverted on the same base, so neither is ours. The
  mic-contention one is a Windows tempdir teardown that cannot delete a still-open PCM
  file — this host, not the test.
- `dart analyze lib/pages/apps` — 10 issues, and analyzing `origin/main`'s version of the
  same tree reports the identical 10. Nothing new.
- `dart format --line-length 120` — 40 files, 0 changed
- `scripts/pr-preflight` — 12 checks pass on both the `local` and `ci` lanes

**Not seen running.** No macOS and no Android SDK on this machine, so every contrast decision
above is reasoned from the surrounding code rather than from looking at it.

That caveat earned its keep. Review found **six places where a fill went white while its
foreground stayed white** — invisible, not merely off-brand: the "Enable" pill's label in two
files, `AnimatedLoadingButton`'s white-by-default label and spinner behind Subscribe and Enable,
two submit spinners, and a `RefreshIndicator` arc drawn on its own white background. All fixed in
`fix(app): give the new white surfaces a foreground that contrasts`.

The shape is worth naming, because the tooling could not catch it. The replacement mapping asserts
how many times each edit matches, so it cannot change something unintended — but a count assertion
says nothing about a foreground the mapping never listed, and in all six the text or indicator
lived in a sibling property, a different widget, or a widget default. I have since swept the whole
diff for the same shape: everything else is either alpha-0.22 behind white text on a dark sheet or
an icon on a dark surface, and the filter-sheet radio already flips its tick to black.

Eyes on a running build would still be worth more than my reading of it.

## Product invariants

**INV-UI-1.** `pr-preflight --suggest` reports no required citation, and
`docs/product/invariants/brand-ui.md` asks that routine UI PRs not name it. Naming it anyway
only because the first commit changes what the guard detects, which is adjacent to the
"changing brand color policy or the allowlist" case the doc reserves the ID for. Happy to
drop the mention if you would rather keep that line bright.

## Failure class

Failure-Class: none

No existing class covers "a pattern did not reach a spelling its subject uses in practice".
`new` felt like overreach — a new class wants a canonical prevention artifact and a reusable
guard surface, and the honest prevention here is the widened pattern this PR already
contains. Happy to classify it differently if you would rather it were tracked.

---

*Written with Claude Code. Every count above is reproducible from
`.github/scripts/check_brand_ui.py` at this commit.*

### DIFF
--- a/.github/scripts/check_brand_ui.py
+++ b/.github/scripts/check_brand_ui.py
@@ -31,6 +31,10 @@ ALLOWLIST_FILES: set[str] = {
     "desktop/macos/Desktop/Sources/Theme/OmiColors.swift",
 }
 
+# The purple hues, named once. Both the `#RRGGBB` and the Dart `0xAARRGGBB` pattern below
+# interpolate this, so adding or retiring a hue is a single edit and the two cannot drift.
+PURPLE_HEX_HUES = "7C3AED|8B5CF6|A855F7|9333EA|6D28D9|AF52DE|D946EF|A78BFA|C4B5FD"
+
 PURPLE_PATTERNS = [
     re.compile(r"Color\.purple\b"),
     re.compile(r"\.purple\b"),  # SwiftUI shorthand: .foregroundStyle(.purple)
@@ -39,11 +43,15 @@ PURPLE_PATTERNS = [
     # cannot match inside `deepPurple` because camelCase puts no word boundary
     # before it, and none of the other patterns spell it either.
     re.compile(r"\bdeepPurple(?:Accent)?\b"),
-    re.compile(r"#(?:7C3AED|8B5CF6|A855F7|9333EA|6D28D9|AF52DE|D946EF|A78BFA|C4B5FD)\b", re.I),
+    re.compile(rf"#(?:{PURPLE_HEX_HUES})\b", re.I),
     # The same hues as written in Dart. Flutter spells colours `Color(0xFF8B5CF6)`,
     # so the `#RRGGBB` pattern above never matched a Flutter literal — the alpha
     # channel is optional because both forms appear in the app.
-    re.compile(r"0x(?:[0-9A-F]{2})?(?:7C3AED|8B5CF6|A855F7|9333EA|6D28D9|AF52DE|D946EF|A78BFA|C4B5FD)\b", re.I),
+    #
+    # The lookbehind requires `0x` to start a token. Without it the pattern matches inside a
+    # longer literal or identifier, so an unrelated constant that merely ends in a purple hue
+    # would be counted and could fail the ratchet on a file that contains no purple at all.
+    re.compile(rf"(?<![0-9A-Za-z_])0x(?:[0-9A-F]{{2}})?(?:{PURPLE_HEX_HUES})\b", re.I),
     re.compile(r"purple(?:Primary|Secondary|Accent|Light|Gradient|LightGradient)\b"),
     re.compile(r"purple-(?:primary|secondary|accent|\d{2,4})\b"),  # Tailwind: bg-purple-500 etc.
     re.compile(r"--purple-"),
--- a/.github/scripts/check_brand_ui.py
+++ b/.github/scripts/check_brand_ui.py
@@ -35,7 +35,15 @@ PURPLE_PATTERNS = [
     re.compile(r"Color\.purple\b"),
     re.compile(r"\.purple\b"),  # SwiftUI shorthand: .foregroundStyle(.purple)
     re.compile(r"Colors\.purple\b"),  # Flutter
+    # Flutter's most common purple, and previously invisible here: `\bpurple\b`
+    # cannot match inside `deepPurple` because camelCase puts no word boundary
+    # before it, and none of the other patterns spell it either.
+    re.compile(r"\bdeepPurple(?:Accent)?\b"),
     re.compile(r"#(?:7C3AED|8B5CF6|A855F7|9333EA|6D28D9|AF52DE|D946EF|A78BFA|C4B5FD)\b", re.I),
+    # The same hues as written in Dart. Flutter spells colours `Color(0xFF8B5CF6)`,
+    # so the `#RRGGBB` pattern above never matched a Flutter literal — the alpha
+    # channel is optional because both forms appear in the app.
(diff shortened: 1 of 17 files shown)


mikejsmith1985:
All eight valid, all fixed. Thank you — the six contrast ones were real invisibility bugs, not
style nits, and I would not have found them by reading.

**The six white-on-white cases.** Each is a fill that went white while its foreground was defined
somewhere I did not look:

| Where | What was invisible |
|---|---|
| `popular_apps_section.dart`, `list_item.dart` | "Enable" label — a sibling `const TextStyle(color: Colors.white)` |
| `app_detail.dart` Subscribe + Enable | `AnimatedLoadingButton` defaults `textStyle` **and** `loaderColor` to white |
| `app_detail.dart` review submit, `reviews_list_page.dart` reply send | the `CircularProgressIndicator` during submission |
| `capability_apps_page.dart` | `RefreshIndicator` arc on its own white `backgroundColor` |

Labels now follow their fill (black on white "Enable", white on grey "Open"), both loading buttons
pass black `textStyle`/`loaderColor` explicitly, both spinners are black, and the refresh arc is
black — matching what `explore_install_page.dart` already does for the same reason.

**Worth naming, because it says something about the method.** The replacements were applied by a
mapping that asserts how many times each must match, which is why nothing unintended was edited.
But a count assertion cannot see a foreground the mapping never listed — and in all six the text or
indicator lived in a sibling property, a different widget, or a widget default. I have since swept
the whole diff for that shape; the remainder is alpha-0.22 behind white text on a dark sheet, or an
icon on a dark surface, and the filter-sheet radio already flips its tick to black.

**The two regex points, both taken.** The `0x` pattern had no leading boundary, so
`count_purple("foo0x8B5CF6")` returned 1 and any longer literal ending in a purple hue counted — on
a no-increase ratchet that is worse than a miss, because it fails a changed file containing no
purple at all. Now behind a negative lookbehind, with a test pinning `foo0x8B5CF6` and
`SOME_CONST0x8B5CF6` at zero while `Color(0xFF8B5CF6)` and a bare `value = 0x8B5CF6;` still count.
The nine hues are now one `PURPLE_HEX_HUES` constant both patterns interpolate, so they cannot
drift.

Verification after the fixes: `check_brand_ui.py` vs `origin/main` — no purple increase across 15
changed UI files; `python -m unittest test_check_brand_ui` — 10 pass; `dart analyze lib/pages/apps`
— 10 issues, identical to `origin/main`'s version of the same tree; `dart format --line-length 120`
— 0 changed; `scripts/pr-preflight` — 12 checks pass. Rebased onto current main.

Still not seen running — that limitation is unchanged, and it is exactly what this round
demonstrated.

kodjima33 (approved):
Ratchet hardening plus the INV-UI-1 cleanup it now catches; CI green. Approving — not merging, since this is a hardening/UI sweep rather than a bug fix.

kodjima33 (approved):
the INV-UI-1 ratchet could not see Colors.deepPurple (camelCase, no word boundary) or Dart 0xAARRGGBB literals — patterns added with token-boundary guards and unit tests, plus the apps pages cleared

Issue #11282, opened 2026-08-09
Firestore Document Reads persistently above alert threshold — day 25, 10 read-reduction PRs merged but still firing, no assignee


beastoin:
## Problem

GCP alert **"Firestore Instance - Document Reads > 10k/s"** (project `based-hardware`, metric `firestore.googleapis.com/document/read_count`, type QUERY) has been breaching since **Aug 7, 2026**. The trend is **worsening** — daily average reads climbing from ~10k/s (Aug 8) to ~14k/s (Aug 21).

**This issue has no assignee (day 28).**

### Breach data (GCP Monitoring API, updated Aug 22)

**233 breach hours out of 609 total** since Aug 7 (38% of all hours above threshold).

| Day | Breach hours | Peak | Floor | Avg |
|-----|-------------|------|-------|-----|
| Aug 7 | 3/24 | — | — | — |
| Aug 8 | 24/24 | — | — | ~10,000/s |
| … | | | | |
| Aug 17 | 23/24 (96%) | 15,753/s | 9,930/s | 12,607/s |
| Aug 18 | 20/24 (83%) | 18,438/s | 8,750/s | 12,972/s |
| Aug 19 | 22/24 (92%) | 13,768/s | 9,546/s | 11,847/s |
| Aug 20 | 19/24 (79%) | 18,599/s | 4,545/s | 13,438/s |
| Aug 21 | 23/24 (96%) | 19,061/s | 8,353/s | 14,431/s |
| Aug 22 | 8/9 so far (89%) | 14,374/s | 4,515/s | 11,878/s |

**Read type breakdown:** QUERY dominates (max 18.3k/s). NOT_FOUND up to 1.1k/s, LOOKUP up to 400/s.

### Trend

- **Daily average is climbing**: ~10k/s (Aug 8) → ~13k/s (Aug 17) → ~14.4k/s (Aug 21)
- **Breach coverage is persistent**: 79–96% of hours per day above threshold (Aug 17–22)
- **New worst peak**: 19,061/s on Aug 21 (~91% above threshold)

### Investigation findings (@Git-on-my-level, Aug 13)

Pulled prod Cloud Monitoring data for Aug 7 18:00–20:00 window:

- **QUERY type = 97% of all document reads** (~9,458/s of 9,762/s)
- **~27 docs per RunQuery** (68M reads / 2.5M RunQuery calls)
- GCP metrics have **no collection-level attribution** — `module`/`version` labels are `__unknown__`
[...]

PR #11526, opened 2026-08-13
perf(backend): serve app search from the catalog the browse endpoints already cache

mikejsmith1985:
Follows #11292 and #11294. This is the *cheap* option from that thread, not the pushdown — the
`search_terms` / Typesense question is still open and still waiting on your read of catalog size.

## What it does

`/v2/apps/search` called `search_apps_db` on every request, which streams the whole approved public
catalog out of Firestore before any text filter runs. Typing `s` cost the same as typing `sports`,
and each keystroke was a collection scan.

`get_approved_available_apps()` already holds that exact set. `search_apps_db`'s default branch
selects `approved == True AND private == False` (`backend/database/apps.py:131-132`); so does
`get_public_approved_apps_db()` (`:67`). The difference is that the second one is already behind a
10-minute Redis copy with a singleflight memory cache in front of it, and is already invalidated by
`invalidate_approved_apps_cache()` on approve, publish, unpublish and visibility change — eight
call sites in `routers/apps.py`. The browse endpoints above search already read it.

So search now reads the same catalog instead of scanning Firestore itself.

**I did not add a cache.** My comment on #11292 proposed a new snapshot behind `set_generic_cache`;
reading the code, reusing the existing one is strictly better. A second cached copy of the public
catalog would need its own invalidation, and a public list that can go stale on its own is exactly
how an unpublished app stays visible after its owner hides it. There is no new cache key here and
no new thing to keep true.

The handler splits into two paths:

| Path | Source | Why |
|---|---|---|
| default (public catalog) | `get_approved_available_apps()` — shared, cached | identical for every user |
| `my_apps` / `installed_apps` | `search_apps_db()` — live, per-user | not shareable, unchanged |

## The per-user field, which is the part worth reviewing

`enabled` is the one fact in the response that differs per caller. The cache hands the *same* `App`
instances to every request that asks for the catalog, so stamping `enabled` onto them in place
would publish one user's installed apps to everyone served from that list until it expired.

The catalog is therefore copied before it is stamped (`_public_catalog_for_search`, last line), and
a test pins it: two users search the same catalog object, and the catalog must still read
`enabled=False` afterwards. This was a real bug in my first draft, not a hypothetical.

## Capability filtering — caught in review

The first version filtered capability with `filter_apps_by_capability`, already imported for the
browse sections in this file. It answers a different question: `_get_app_capability` resolves each
app to one section by priority order and notification apps are dropped outside their own section.
The predicate it replaced was `array_contains`, i.e. membership — so an app declaring
`{external_integration, chat}` would have stopped being findable by `chat`.

Now `app.has_capability(capability)`. Caught by cubic on this PR; the regression test fails on the
commit before the fix.

## A latent 500 on the shared path, fixed first

Switching search onto the shared builder surfaced that `get_approved_available_apps` reads
`app['id']` unguarded — once to build the ids for the installs and reviews lookups, and again per
app in the loop. A legacy document without that field raises `KeyError` out of the builder.

That is the same failure `search_apps` was already hardened against
(`test_apps_search_poison_guard.py`), on a worse path: this list is Redis- and process-cached and
serves the marketplace to everyone, so one bad document takes out browse for every user rather than
one person's search page. Fixed in its own commit ahead of the perf change, with its own test.

I have not seen this fire in production — I found it by reading the path I was about to depend on.

## Why six existing tests moved

`test_apps_search_matches_description.py` and `test_apps_search_poison_guard.py` both patched
`search_apps_db` and drove the default public path. That path no longer calls it, so the fixtures
had to move. **No assertion changed** — only where the apps come from:

- the description-matching tests now feed `get_approved_available_apps`, so the name-or-description
  predicate and the #11294 relevance tiers are asserted exactly as before;
- the poison guard now runs against `my_apps=True`, which is the scope that still builds apps from
  raw documents. The public path's equivalent guard is `_safe_build_app` upstream, plus the new
  missing-id test above.

Flagging this explicitly because a rewritten test in the same PR as its subject is worth a second
look, and I would rather point at it than have it found.

## One behaviour change, declared

The shared catalog skips apps flagged `disabled` (`backend/utils/apps.py:517`). The old search path
did not, so a platform-disabled app — one whose endpoints failed health validation, per
`validate_app_endpoints_for_reenable` — was reachable from search while being absent from browse.

After this change search agrees with browse and those apps are excluded. I believe that is the
correct behaviour and it is why I did not work around it, but it is a change in what the endpoint
returns and it is yours to overrule. Say the word and I will preserve the old inclusion explicitly.

## What this does not do

It does not push the query down, so cost is still proportional to catalog size rather than to the
query — it is just no longer paid in Firestore reads on every keystroke. The A-vs-B design question
in #11292 is untouched and still needs your call on how large the catalog is and how fast it grows.

Worth noting for #11282: the removed call was an unbounded Firestore collection scan on an
interactive path, which is the shape that issue is chasing. I am not claiming it is *the* cause —
just that it is one instance, measured rather than speculative.

## Verification

- `backend/tests/unit/test_apps_search_shared_catalog.py` — 6 new tests: the public path never
  reaches `search_apps_db`; `enabled` is per-user and does not persist into the shared catalog;
  unapproved and private apps are excluded even if present in the source; the category filter
  applies over the shared list; and both `my_apps` and `installed_apps` still read live
- `backend/tests/unit/test_apps_catalog_missing_id_guard.py` — 2 new tests for the second commit;
  both fail with `KeyError` before it
- 14 passed across the four apps-search files; 57 more across
  `test_apps_marketplace_poison_guard.py` (4), `test_app_visibility_cache_invalidation.py` (3),
  `test_firestore_read_ops_cache.py` (48 passed, 3 skipped) and `test_apps_installs_count_clamp.py` (2)
- `black --line-length 120 --skip-string-normalization` — clean
- `scripts/pr-preflight` — 21 of 23 selected checks pass. Two could not run here and neither is
  reachable from this diff:
  - `backend-route-policy-baseline` provisions its own venv from `backend/.python-version`
    (3.11). This machine has no 3.11, so the runner fell back to 3.13 and `onnxruntime==1.19.0`
    has no cp313 wheel. It fails identically with an empty diff (`--base-ref HEAD`), so the
    failure is venv provisioning, not the change.
  - `desktop-backend-candidate-probe-fixtures` reports *"total response deadline is unsupported
    on this runner"*. Its test file is not in this diff.

  The remaining checks were run individually after the first failure stopped the runner:
  `firestore-query-coverage`, `lifecycle-headers`, `version-prefixed-filenames`,
  `backend-import-purity`, `backend-runtime-image-source-closure`, `backend-module-isolation`,
  `backend-workflow-contracts`, `conversation-lifecycle-write-guard`,
  `desktop-backend-release-boundary` and its fixtures — all pass.

**Not exercised against a live backend.** No deployment or Firestore/Redis instance here, so the
cache-sharing behaviour is proven by unit test against a controllable seam rather than by watching
real requests. The claims about which fields the shared source already populates are read from
`get_approved_available_apps` rather than measured. If you want a live smoke before this lands, that
is a reasonable ask and I cannot supply it.

## Line-count exceptions

Both files are already over the 1,500-line threshold. Splitting a 2,400-line router is not in scope
for a perf change, so both declare their growth.

Line-Count-Exception: backend/routers/apps.py | 2425 -> 2487 | splits /v2/apps/search by whether its result is shareable so the public catalog reads the list the browse endpoints already cache while my_apps and installed_apps stay live; the seam and the per-user `enabled` copy belong in the handler that owns the response contract, and the extracted helpers carry the reasoning that makes the aliasing hazard visible

Line-Count-Exception: backend/utils/apps.py | 1652 -> 1661 | drops marketplace records with no id before the shared catalog builder reads `app['id']` unguarded, keeping the guard at the builder that serves every browse and search caller rather than at each call site

## Failure class

Failure-Class: none

This is a `perf:` change rather than a `fix:`, so no declaration is strictly required; recording
`none` because the aliasing bug it would otherwise have shipped was caught in review here rather
than in production, and no existing class covers "per-user field stamped onto a shared cached
object". Happy to file one if you would rather it were tracked.

---

*Written with Claude Code. Every file:line above verified against `8ddcff958`.*

### DIFF
--- a/backend/routers/apps.py
+++ b/backend/routers/apps.py
@@ -667,6 +667,97 @@ def _name_match_tier(app: App, query: str) -> int:
     return 2
 
 
+def _serves_public_catalog(my_apps: bool | None, installed_apps: bool | None) -> bool:
+    """Whether this search reads the shared public catalog rather than one user's own apps.
+
+    `my_apps` and `installed_apps` both scope the result to the caller, so they cannot be served
+    from a catalog shared between users. Everything else is the same approved public list for
+    everybody.
+    """
+    return not my_apps and not installed_apps
+
+
+def _public_catalog_for_search(category: str | None, capability: str | None) -> List[App]:
+    """The approved public apps for a search, taken from the catalog the browse endpoints share.
+
+    `search_apps_db`'s default branch selects `approved == True AND private == False`, which is
+    exactly what `get_approved_available_apps()` already holds: a 10-minute Redis copy behind a
+    singleflight memory cache, invalidated by `invalidate_approved_apps_cache()` on every approve,
+    publish, unpublish and visibility change.
+
+    Reading it here is what removes the per-keystroke Firestore scan. It deliberately reuses the
+    existing catalog rather than adding a second cached copy, because a second copy would need its
+    own invalidation — and a public list that goes stale independently is how an unpublished app
+    stays visible after its owner hides it.
+    """
+    apps = get_approved_available_apps(include_reviews=True)
+    # Re-assert visibility on the way out, exactly as the browse endpoints do. The source is
+    # already approved-and-public, so this changes nothing today; it is here so a future widening
+    # of the shared catalog cannot quietly make search the endpoint that leaks a private app.
+    apps = [app for app in apps if app.approved and not app.private]
+    if category:
+        apps = [app for app in apps if app.category == category]
+    if capability:
+        apps = filter_apps_by_capability(apps, capability)
+
+    # Copy before returning. These App objects are the cache's own, handed to every request that
+    # asks for the catalog, so the caller stamping a per-user `enabled` onto them would publish one
+    # user's installs to everyone served from the same cached list until it expired.
+    return [app.model_copy() for app in apps]
+
+
+def _personal_apps_for_search(
+    uid: str,
+    category: str | None,
+    capability: str | None,
+    my_apps: bool,
+    installed_apps: bool,
+) -> List[App]:
+    """The caller's own or installed apps, read live because they are not shareable between users."""
+    enabled_app_ids = list(get_enabled_apps(uid)) if installed_apps else None
+    apps_data = search_apps_db(
+        uid=uid,
+        category=category,
+        capability=capability,
+        my_apps=my_apps,
+        installed_apps=installed_apps,
+        enabled_app_ids=enabled_app_ids,
+    )
+
+    # Drop any malformed record missing an id before enrichment: id drives the installs/reviews
+    # lookups below and the pre-loop app_ids list, so a missing id would KeyError before the
+    # per-record ValidationError guard can catch it.
+    valid_apps_data = [a for a in apps_data if a.get('id')]
+    skipped_no_id = len(apps_data) - len(valid_apps_data)
+    if skipped_no_id:
+        logger.warning("Skipping %d malformed app record(s) without an id in search results", skipped_no_id)
+
+    app_ids = [app['id'] for app in valid_apps_data]
+    apps_installs = get_apps_installs_count(app_ids)
+    apps_reviews = get_apps_reviews(app_ids)
+
+    apps: List[App] = []
+    for app_dict in valid_apps_data:
+        app_dict['installs'] = apps_installs.get(app_dict['id'], 0)
+
+        # Calculate average from reviews
+        reviews = apps_reviews.get(app_dict['id'], {})
+        scores = [_clamp_review_score(x['score']) for x in reviews.values()]
+        app_dict['rating_avg'] = sum(scores) / len(scores) if scores else None
+        app_dict['rating_count'] = len(scores)
+
+        # Skip a malformed/legacy app document rather than 500 the whole search page.
+        try:
+            apps.append(App(**app_dict))
+        except ValidationError as e:
+            logger.warning(
+                "Skipping malformed app %s in search results: %s",
+                app_dict.get('id'),
+                [err['loc'][0] for err in e.errors() if err.get('loc')],
+            )
+    return apps
+
+
 @router.get('/v2/apps/search', tags=['v2'], response_model=AppSearchResponse)
 def search_apps(
     q: str | None = Query(default=None, description='Search query for app name or description'),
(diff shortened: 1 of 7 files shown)


mikejsmith1985:
Valid, and thank you — fixed in the latest push.

`search_apps_db` filtered capability with `FieldFilter('capabilities', 'array_contains', capability)`,
which is membership. I reached for `filter_apps_by_capability` because it was already imported and
already used by the browse endpoints in the same file, without checking that it answers a different
question: it compares against a single primary capability from `_get_app_capability`, drops
notification apps outside the notification section, and has a `popular` branch keyed on `is_popular`
rather than on the capability list at all.

So a multi-capability app would have stopped being findable by any capability other than its
primary one, and notification apps would have vanished from most capability filters entirely. Now
`app.has_capability(capability)`, which is the same membership test the Firestore predicate was.

Added `test_capability_filter_matches_any_capability_not_just_the_first` — an app with
`{chat, memories}` searched by `memories` must come back, which fails on the previous commit.

Two other things came out of this round, both now fixed:

- CI's backend unit suite caught `test_app_review_score_clamp.py`, a fourth file driving the public
  search path that my local subset had missed. Repointed at `my_apps`, which is the scope that still
  computes `rating_avg` in the handler; the static sweep at the bottom of that file still covers both
  aggregation sites.
- I have since enumerated every test importing `routers.apps` or `utils.apps` rather than picking
  files by name, which is how the first one got missed.

Git-on-my-level:
Solid performance work. Routing the public search path to the already-cached `get_approved_available_apps()` catalog instead of a per-keystroke Firestore scan is the right call, and reusing the existing cache rather than adding a second cached copy avoids a new staleness/invalidation surface.

A few specifics worth highlighting:

- **Per-user `enabled` isolation is correct.** The cache hands the same `App` instances to every request, and you copy via `model_copy()` in `_public_catalog_for_search` before stamping `enabled` — the test pinning two users on the same catalog object (`test_enabled_is_per_user_and_does_not_persist_into_the_shared_catalog`) is exactly the right regression guard.
- **Missing-id guard in `utils/apps.py`** is a good catch. A single legacy document without `id` raising `KeyError` on the shared/cached path would have taken out browse for everyone.
- **Capability filter fix** (`has_capability` membership instead of `filter_apps_by_capability`'s primary-capability resolution) preserves findability for multi-capability apps. Good that the cubic review caught it and you added a regression test.
- **Declared behaviour change — disabled apps now excluded from search.** The shared builder skips `disabled` apps, so search now agrees with browse. This is likely the correct behaviour, but it's a change in what the endpoint returns, so leaving it for a maintainer to confirm.

Tests are thorough: the `_forbidden` sentinel asserting `search_apps_db` is never reached on the public path, the per-user `enabled` non-persistence test, and the missing-id guard tests all directly pin the real regression risks. The three migrated existing test files keep their assertions unchanged and only move the seam.

Positive signal — the implementation is correct and well-tested. The one declared behaviour change (disabled-app inclusion) is yours to confirm before merge.

---

*Automated review by glm-5.2 (omi-pr-monitor). Leaving for human maintainer review on the disabled-app search-inclusion behaviour change.*

---

kodjima33 (approved):
Reuses the already-cached browse catalog for app search; CI green. Approve-only: perf rewrites stay out of the auto-merge lane.

mikejsmith1985:
Closing this — every commit here is now either landed upstream by another route or superseded.

| Commit | Status |
|---|---|
| `ab2e33b384` perf: serve search from the cached catalog | **Landed via #11851.** `search_apps_db` on `main` now reads `get_public_approved_apps_cached_db()` on the public-set path, which is the same change with a cleaner seam than mine. |
| `cac6424b94` fix: match any capability | **Not a real bug.** It patched `_public_catalog_for_search`, a helper that only ever existed on this branch — introduced by my own perf commit. `main`'s search has always used membership (`array_contains`, and `capability in (app.get('capabilities') or [])`). My mistake; thanks to the reviewer for catching the original regression. |
| `691a7618c6` fix: drop records with no id | **Reopened standalone as #12790**, rebased on current `main` with the regression test and a full green preflight. |

Thanks for the review, @Git-on-my-level — the perf reasoning you signed off on is what shipped, just through #11851 rather than here.

The remaining live work is #12790 for the `KeyError`, and #11292 where the A-vs-B design question is still open.

PR #11532, opened 2026-08-14
docs(app): point mobile setup at the local backend harness, not api.omiapi.com

formed2forge:
## Problem

`docs/doc/developer/AppSetup.mdx` and `app/README.md` point a new mobile developer at `https://api.omiapi.com/`, but `app/setup.sh ios` configures the app for **local development**: it writes `API_BASE_URL=http://127.0.0.1:8000/` into `app/.dev.env` and passes `--dart-define=OMI_FIREBASE_AUTH_EMULATOR_HOST` (`setup.sh:417-420`). The documented target and the actual one disagree.

Following the docs as written, a developer gets an app that builds, installs, launches — and then sits on the splash screen, because it is waiting on a backend nobody told them to start. There is no error and no hint of the cause.

Production Firebase makes the documented path worse than useless: the app authenticates against Omi's production Firebase project, and those tokens are rejected on every call, so sign-in appears to succeed and nothing else works.

## Change

Documents what `setup.sh` actually does, and what it does **not** do:

- The build targets the **local backend harness** — the Python API on port `8000` and the Firestore emulator — not production.
- **`setup.sh` does not start either of them.** This prerequisite was previously undocumented and is the single thing most likely to strand a new contributor.
- Its real prerequisites: **Python 3.11** (not 3.12+ — the backend pins 3.11) and a **Java runtime** for the Firestore emulator.
- Addressing differs by platform: iOS reaches the harness at `127.0.0.1`, Android defaults differently. Both are overridable via `OMI_DEV_HOST` / `OMI_LOCAL_API_BASE_URL`.
- A one-command path (`make dev-init`) for creating `backend/.venv` and the env template.

## Verification

Rebased onto current `main`; both files have zero upstream commits since the change was written, so it applies cleanly. `main` still contains the stale `api.omiapi.com` reference in each file, so the correction is still needed.

The splash-screen symptom was reproduced today on an iOS device build: `setup.sh ios` completed, the app installed and launched, and hung on the logo with nothing listening on port 8000.

## Scope

Documentation only. It does not change `setup.sh`, and deliberately does not try to make the script detect or start the harness — that is a separate change, filed as its own issue.

Note that the prod-simulating path (`setup.sh ios beta`, which targets `https://api.omiapi.com/`) is gated behind `FIREBASE_SERVICE_ACCOUNT_KEY` and so is unavailable to community contributors by design. These docs therefore describe the local harness as the supported route rather than implying production is reachable.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -8,12 +8,30 @@ The Omi App is a Flutter-based mobile application that serves as the companion a
 
 Before getting started, make sure your device is connected and unlocked. If you're using an iPhone, ensure that Developer Mode is enabled — you can toggle this in the iPhone settings. For Android devices, make sure the device is connected and USB debugging is enabled in Developer Options
 
-1. Navigate to the app directory:
+1. Start the local backend harness from the **repo root**. `setup.sh` builds
+   against it but does not start it, so do this first or the app will launch with
+   nothing to talk to:
+   ```bash
+   make dev-init   # once: creates backend/.venv and copies the env template
+   make dev-up     # Firestore + Auth emulators and the Python API on :8000
+
+   # No provider API keys? Use fake providers:
+   PROVIDER_MODE=offline make dev-up
+   ```
+
+   `make dev-init` builds `backend/.venv` from whatever `python3` resolves to,
+   and the backend requires **Python 3.11** (not 3.12+). The harness also needs a
+   Java runtime and `firebase-tools`/`npx`; `make dev-up` names what's missing.
+
+   `make dev-status` shows what came up; `make dev-down` stops it. Ports, seeded
+   users, and troubleshooting: [`docs/runbooks/local-emulator-manual-qa.md`](../docs/runbooks/local-emulator-manual-qa.md).
+
+2. Navigate to the app directory:
    ```bash
    cd app
    ```
 
-2. Run the setup script for your platform:
+3. Run the setup script for your platform:
    ```bash
    # macOS/Linux: iOS
    bash setup.sh ios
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified this against the codebase and the technical claims all check out — `setup.sh` does write `API_BASE_URL=http://127.0.0.1:8000/` and sets `OMI_FIREBASE_AUTH_EMULATOR_HOST`, the `make dev-init`/`dev-up`/`dev-status`/`dev-down` targets all exist, `PROVIDER_MODE=offline` is valid (`providers.py`), the Auth emulator port `9099` matches `config.py` and `env.dart`, the `demo-omi-local` project matches `firebase_options_local.dart`, and the Python 3.11 pin matches `backend/.python-version`. The linked runbook `docs/runbooks/local-emulator-manual-qa.md` exists.

The core fix — documenting that `setup.sh` does not start the backend/emulator and that `make dev-up` must be run first — addresses a real onboarding gap (the splash-screen hang the PR describes). The warnings about not pointing at `api.omiapi.com` with emulator-only Firebase configs are technically accurate given the project-scoping behavior.

Two minor nits (non-blocking, already surfaced by cubic):
- **`AppSetup.mdx` manual-setup table**: the `API_BASE_URL` row now recommends `http://127.0.0.1:8000/` but doesn't distinguish iOS (`127.0.0.1`) from Android emulator (`10.0.2.2`), while the automatic-setup section above does. An Android developer following manual setup would hit connection-refused. A one-line platform note would match the automatic section's precision.
- **Troubleshooting accordions** largely restate the inline `<Warning>` blocks in the same file. Consolidating or cross-linking would reduce drift risk if ports/projects change.

Good documentation contribution. Leaving for human maintainer review for a docs accuracy decision.

---
*Automated review by glm-5.2 · [Omi PR Maintainer Monitor](https://github.com/BasedHardware/omi)*

---

kodjima33 (approved):
Verified every claim against main: dev-init/dev-up/dev-status/dev-down exist, setup.sh defaults the dev flavor to 127.0.0.1:8000 (10.0.2.2 on Android) and honours OMI_DEV_HOST, api.omiapi.com is the beta profile only, and the demo-omi-local configs are emulator fakes. Docs were stale; this matches the code.

PR #11536, opened 2026-08-14
fix(desktop): do not turn other people's commitments into the user's tasks

Git-on-my-level:
## What changed and why

The context director used `task_candidate` whenever a validated fact recorded an explicit commitment that was absent from the task list. It never required that commitment to be the user's own, so other people's obligations became interactive interruptions and durable entries on the user's task surface.

Tightened `ContextProactivityPromptBuilder.directorStablePrompt` so `task_candidate` requires a first-person commitment the user made or accepted. A third-party commitment that bears on the user's tracked work may at most be `insight`; a commitment between other parties is `silence`. Decision semantics, schema, grounding rules, and delivery gates are unchanged.

## Prompt sentence being tightened

Before:

> Use task_candidate only when a validated fact explicitly records a new commitment, promise, or request with an accountable action, and that commitment is absent from the supplied task list. ... never infer an owner or due date ...

"Never infer an owner" was read as "don't invent an owner", not as "only act on the user's commitments". An explicit third-party commitment satisfied every stated condition.

After (ownership precondition plus third-party rule, existing constraints kept):

> Use task_candidate only when a validated fact explicitly records a new commitment, promise, or request with an accountable action that the user personally made or accepted (first person), and that commitment is absent from the supplied task list. A commitment made by another person is never a task candidate, however explicit or well-dated it is; if it genuinely bears on the user's tracked work it may at most be insight, and a commitment between other parties that does not involve the user is silence.

## Reproduced against gpt-5.6-luna (`probe_context_bucket_director`, faithful snapshot wire format)

**Third party commits to a third party.** Fact: *"Dmitri Vance committed to publish the adapter rollback runbook by Friday."* Director returned `task_candidate`, title *"Track rollback runbook publication"*, message *"Dmitri committed to publish the adapter rollback runbook by Friday. Add this as an open task..."*, reasoning *"The validated thread records an explicit commitment with a deadline, and no corresponding task appears in the open task list."*

**Third party owes the user.** Fact: *"Rina Okonkwo committed to send the updated pricing table tomorrow."* Director returned `task_candidate`, *"Track pricing table delivery"*, message *"Rina Okonkwo committed to send the updated pricing table tomorrow... Consider tracking this as an open follow-up."*

**Third party, different domain, user absent from the thread.** Fact: *"Priya Raman committed to return the signed DPA to the Helio team by Wednesday."* Director returned `task_candidate`, title *"Follow up on overdue signed DPA"*, message *"Priya committed to return the signed DPA by Wednesday, which has passed as of Aug 14. Follow up with Priya or the Helio team on the outstanding DPA."*

**Control cases in the same run behaved correctly**, so this is ownership-specific, not general over-firing: a first-person commitment ("You replied that you will post the cutover plan by tomorrow morning") correctly produced `task_candidate`; the same commitment already present in the task list produced `resurface`; an already-fulfilled first-person commitment produced `silence`; a >48h "reference only" task produced `silence`; injected instruction-shaped screen text produced `silence`.

## Stale prompt-contract test (already red on main)

`testDirectorPromptContainsSafetyPreambleBucketTasksAndFrameMetadata` is a full-equality contract. It has been failing since `8fda810a1e` ("fix: link meeting notes from always-on capture (#11468)") added the two meeting-notes lines to `directorStablePrompt` and did not update the expected string. This PR brings the expected string back in line with the real prompt (those two lines plus the new ownership sentences). It does not weaken `XCTAssertEqual` into a `contains` check.

The desktop Swift test lane **does run this test**. `swift-test-suites.sh` auto-discovers every `XCTestCase` under `Desktop/Tests/`; `ContextBucketPromptAssemblerTests` is not in the skip ratchet. Changing `ContextBucketRollup.swift` selects `desktop-swift-tests`. After #11468 merged, the main-push `Desktop Swift Build & Tests` run executed this suite (14 tests) and failed on exactly `testDirectorPromptContainsSafetyPreambleBucketTasksAndFrameMetadata` (`https://github.com/BasedHardware/omi/actions/runs/31734789303`). The red equality test reached `main` because #11468 merged while that required check was red / completed after merge, not because CI omits the suite.

none

## How it was verified

- Source and test expected instruction lines match 1:1 (Python line diff of `directorStablePrompt` vs the equality fixture): MATCH, 15/15 lines.
- `python3 scripts/check_desktop_test_quality.py`: OK, debt at or below baseline.
- `./scripts/dev-feedback.py --once swift 'ContextBucketPromptAssemblerTests'`: compile blocked only by the known pre-existing `RewindPage.swift:153` type-check timeout on this machine's Xcode 26.2 / Swift 6.2.3. CI pins Xcode 16.4 and is unaffected. Pushed with `PRE_PUSH_SKIP_DESKTOP_SWIFT=1` for that unrelated file only.
- Could not re-run the live `probe_context_bucket_director` dogfood on this toolchain for the same `RewindPage.swift` compile block. The production prompt string is what the probe sends; the equality test is the hermetic guard. Original dogfood against gpt-5.6-luna is quoted above.

## Tests

- Restored and extended `testDirectorPromptContainsSafetyPreambleBucketTasksAndFrameMetadata` (full prompt equality, including the #11468 meeting-notes lines and the new ownership sentences).
- Added `testDirectorStablePromptRequiresFirstPersonOwnershipForTaskCandidate`, which calls `directorStablePrompt` and asserts the first-person precondition and third-party insight/silence rule.

## Failure class (fixes)

Failure-Class: new

**Violated contract:** a durable user-surface action (`task_candidate` graduation into the user's task list, plus an interactive interruption) was gated on the existence of a commitment but not on who owns it.

**Canonical prevention:** state first-person ownership as an explicit `task_candidate` precondition in the director prompt, keep the third-party insight-or-silence rule beside it, and retain a full-equality prompt-contract test that includes both sentences.

**Evidence:** the three gpt-5.6-luna reproductions above. Registry file: `.github/failure-classes/FC-task-candidate-unowned-commitment.json`. Guard artifact: `desktop/macos/Desktop/Tests/ContextBucketPromptAssemblerTests.swift`.

## New guards

The prompt-contract test is the existing production-API equality guard for this prompt, not a new shared primitive. It is the check that would have caught both the #11468 expected-string drift and a future drop of the ownership sentences. A model-eval / shadow-scoring gate is deliberately not added.

🤖 Generated with Cursor

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-task-candidate-unowned-commitment.json
+++ b/.github/failure-classes/FC-task-candidate-unowned-commitment.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-task-candidate-unowned-commitment",
+  "violated_contract": "A durable user-surface action may be gated on the existence of a commitment only when that commitment is the user's own. task_candidate graduates into the user's task list and presents an interactive interruption; a third-party commitment, however explicit or dated, must not satisfy that gate.",
+  "canonical_prevention": "State first-person ownership as an explicit precondition of task_candidate in the director prompt, keep the third-party insight-or-silence rule beside it, and retain a full-equality prompt-contract test that includes both sentences.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/Desktop/Tests/ContextBucketPromptAssemblerTests.swift"
+  ],
+  "evidence_prs": [
+    11536
+  ],
+  "scope_hints": [
+    "desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketRollup.swift",
+    "desktop/macos/Desktop/Tests/ContextBucketPromptAssemblerTests.swift"
+  ],
+  "status": "open"
+}
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketRollup.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketRollup.swift
@@ -181,10 +181,14 @@ enum ContextProactivityPromptBuilder {
       Use resurface or suggest for an actionable open task supplied below. Entries marked
       reference-only are identity context: do not notify about or recreate them yet. Use
       task_candidate only when a validated fact explicitly records a new commitment, promise,
-      or request with an accountable action, and that commitment is absent from the supplied
-      task list. A material change, status update, recommendation, or useful follow-up without
-      an explicit commitment, promise, or request is insight or suggest; never infer an owner or
-      due date and never create a task candidate from actionability alone.
+      or request with an accountable action that the user personally made or accepted (first
+      person), and that commitment is absent from the supplied task list. A commitment made by
+      another person is never a task candidate, however explicit or well-dated it is; if it
+      genuinely bears on the user's tracked work it may at most be insight, and a commitment
+      between other parties that does not involve the user is silence. A material change, status
+      update, recommendation, or useful follow-up without an explicit commitment, promise, or
+      request is insight or suggest; never infer an owner or due date and never create a task
+      candidate from actionability alone.
 
       \(stableBucket)
       """
--- a/desktop/macos/Desktop/Tests/ContextBucketPromptAssemblerTests.swift
+++ b/desktop/macos/Desktop/Tests/ContextBucketPromptAssemblerTests.swift
@@ -64,13 +64,19 @@ final class ContextBucketPromptAssemblerTests: XCTestCase {
       ScreenDerivedContent.untrustedPreamble,
       "Decide whether interrupting now adds concrete value. Return silence unless the validated",
       "facts support a specific, timely action. Use only supplied bucket-entry refs.",
+      "Never announce that meeting notes, a transcript, or a call summary are ready. The",
+      "conversation-finalization lane owns that claim and attaches the exact conversation link.",
       "Use resurface or suggest for an actionable open task supplied below. Entries marked",
       "reference-only are identity context: do not notify about or recreate them yet. Use",
       "task_candidate only when a validated fact explicitly records a new commitment, promise,",
-      "or request with an accountable action, and that commitment is absent from the supplied",
-      "task list. A material change, status update, recommendation, or useful follow-up without",
-      "an explicit commitment, promise, or request is insight or suggest; never infer an owner or",
-      "due date and never create a task candidate from actionability alone.",
+      "or request with an accountable action that the user personally made or accepted (first",
+      "person), and that commitment is absent from the supplied task list. A commitment made by",
+      "another person is never a task candidate, however explicit or well-dated it is; if it",
+      "genuinely bears on the user's tracked work it may at most be insight, and a commitment",
+      "between other parties that does not involve the user is silence. A material change, status",
+      "update, recommendation, or useful follow-up without an explicit commitment, promise, or",
+      "request is insight or suggest; never infer an owner or due date and never create a task",
+      "candidate from actionability alone.",
       "",
       bucket,
       "",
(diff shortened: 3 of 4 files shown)


Git-on-my-level:
## Empirical before/after against the real model boundary

Verified by driving `probe_context_bucket_director` (real director prompt, real
`proactive_reasoning` lane, `gpt-5.6-luna`) from a local named QA bundle, using snapshot fixtures in
the faithful wire format (`entry:<id> …` tail lines, `fact:<id> … [evidence: …; refs: …]` facts).

"Before" = `/Applications/omi-qa-main.app`, built from `main` earlier today (pre-fix).
"After" = a bundle built from this branch's prompt change applied on top of `main` + the
`RewindPage` type-check fix (#11537), which is required to build locally on Xcode 26.2 at all.

| case | before | after |
| --- | --- | --- |
| **A** control — the user's own commitment, absent from the task list | `task_candidate` | `task_candidate` (unchanged) |
| **K** third party commits to another third party | `task_candidate` | `silence` |
| **L** third party commits to the user | `task_candidate` | `silence` |
| **N** third party, user absent from the thread entirely | `task_candidate` | `insight` |

Reasoning strings after the change show the ownership test being applied directly:

- **K**: "It is not the user's commitment, and there is no open task or explicit user action to resurface."
- **L**: "Rina's commitment is an update from another person, not a commitment owned or accepted by the user."
- **N**: "the commitment belongs to Priya rather than the user, so it is an insight—not a task candidate."
- **A** (control, still fires): "This is an explicit first-person commitment, absent from the supplied open task list, with a near-term deadline."

Before the change, **N** additionally asserted a missed deadline on another person's behalf
("which has passed as of Aug 14. Follow up with Priya or the Helio team"), which is the shape that
made this worth fixing: `task_candidate` graduates durable candidates via
`CandidateSink.graduateValidatedFacts` and then presents an interactive notification, so each of
these wrote another person's obligation into the user's task surface.

All fixture content is invented (fictional people, company, and issue IDs). No real user content was
used. 20 real model calls total across the run, all HTTP 200.

Git-on-my-level:
The failing **Desktop Swift Static & Test Contracts** lane here is not caused by this PR.

The job log is dominated by `RewindPage.swift:153: error: the compiler is unable to type-check this expression in reasonable time` — a pre-existing regression on `main` from #11519, fixed in #11537. The two earlier-looking failures in the same job (`FAIL: TERM case never launched detached app`, `FAIL: sentinel target build failed`) are downstream of that same compile error, since both steps have to build and launch the app.

I re-ran the failed jobs once to test whether they were flaky; they reproduced identically on attempt 2, which rules out flakiness and matches the compile-blocker explanation.

#11537 passes Desktop Swift Release Compile. Merging it first and rebasing this branch should clear these lanes. The prompt/test changes in this PR were verified separately against the live model — see the before/after comment above.

Git-on-my-level:
**Correction to my previous comment.** I claimed the `TERM case never launched detached app` / `sentinel target build failed` steps were downstream of the RewindPage compile error. I tested that and it is **wrong**.

#11537 carries the RewindPage fix, and its `Desktop Swift Static & Test Contracts` lane **still fails**, on:

```
FAIL: TERM case never launched detached app
ERROR: timed out after 4s waiting for ./run.sh build lock (.../omi-run-sh-lock-test.../wt-a/.dev/run-sh-build.lock.d)
FAIL: workflow test-install job must semantically match the complete-job fixture   (x4)
```

So there are two independent problems on the desktop lanes right now:

1. **RewindPage type-check** (#11519 regression, fixed by #11537) — this genuinely explains the `Desktop Swift Release Compile` failures and any compile of the app.
2. **A separately broken `Desktop Swift Static & Test Contracts` lane** — launcher-script tests: a detached-launch TERM case, a `run.sh` build-lock timeout, and a workflow-fixture mismatch. These reproduce **with** the RewindPage fix applied, so they are not caused by it and are not caused by this PR either. None of these three PRs touch `.github/workflows/*.yml` or `run.sh`.

Net: #11537 still unblocks the release-compile lane and should still land first, but it will **not** by itself turn Static & Test Contracts green. That lane looks broken for every desktop PR at the moment and needs a separate owner.

Git-on-my-level:
**Correcting my earlier comment on this PR.** I attributed the `Desktop Swift Static & Test Contracts` failure partly to `TERM case never launched detached app` / `sentinel target build failed` / a workflow-fixture mismatch. That was a misread on my part: I grepped the job log for `FAIL:` and those lines are *expected* output from mutation sub-cases where the guard is meant to reject. `desktop/macos/tests/test-test-install-workflow-contract.sh` exits **0** when run directly.

The lane's actual failing step is `Test desktop Swift app`. On the runs I inspected the failing suites were `ContextBucketPromptAssemblerTests` (the equality test red on `main` since #11468, fixed by #11536), plus `AgentPillLifecycleTests` / `ChatTimelineContinuityTests` / `MarkdownNestedScrollTests` failing with `error: Exited with unexpected signal code 11` (test-runner segfaults), and `ChatCitationTests.testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID`, which asserts on the global singleton `RewindCitationFocusState.shared` and is order-dependent on shared state.

Also for the record: the RewindPage type-check regression from #11519 was fixed directly on `main` in `37aa667262` (prompted by Codemagic v0.12.170 failing `swift build -c release`), so my #11537 was superseded and is now closed.

PR #11537, opened 2026-08-14
fix(desktop): split RewindPage.body so Swift 6.2 can type-check it

Git-on-my-level:
## What

`RewindPage.body` was a single ~15-modifier chain. On the current maintainer-local toolchain it does not compile:

```
desktop/macos/Desktop/Sources/Rewind/UI/RewindPage.swift:153:23: error: the compiler is unable to
type-check this expression in reasonable time; try breaking up the expression into distinct
sub-expressions
```

Toolchain: Xcode 26.2 (build 17C52), Apple Swift 6.2.3 (swiftlang-6.2.3.3.21), target `arm64-apple-macosx26.0`.

This reproduces on **unmodified `origin/main`** (verified at `6083f7de58` in a clean detached worktree before any edit) — it is not caused by recent feature work. CI pins Xcode 16.4, so the release lane is green and structurally cannot observe this.

**Impact:** `./run.sh` cannot build any local named QA bundle from `main` on this toolchain, so every local desktop dogfooding and QA path is blocked. This was found while trying to build a bundle to dogfood the context-bucket proactive pipeline.

## Change

`body` is now `keyboardHandledContent`, and the original chain is split into named `some View` helpers applied in the same order: `chromeContent` → `lifecycleAttachedContent` → notification-observer groups → view-model-observer groups → keyboard-handler groups. The track-window Combine chain is extracted to `trackWindowRangePublisher` so `combineLatest`/`debounce` overloads are no longer solved inside the SwiftUI modifier chain.

Behavior is unchanged:

- 43 view modifiers before, 43 after, in the same applied order. (`.debounce` moves earlier in *file text* only, because the publisher is now a computed property declared above its use; `.onReceive(trackWindowRangePublisher)` keeps its exact position in the chain, still followed by `.onChange(of: viewModel.activeSearchQuery)`.)
- No closure body was edited, no observer added or removed.
- Comments moved with the modifiers they explain, including the focus-ring rationale, the owner-change frame-invalidation note, and the preserve-position note.

## Verification

Local debug build with the toolchain that previously failed:

```
cd desktop/macos && xcrun swift build -c debug --package-path Desktop
```

succeeds, producing `.build/debug/Omi Computer` (238 MB). No `RewindPage.body` type-check warning remains.

Not claimed: this PR was not exercised at runtime. It is a compile-time restructuring with no behavior change, and the app was not launched from this build.

## Line-Count-Exception

Line-Count-Exception: desktop/macos/Desktop/Sources/Rewind/UI/RewindPage.swift | 1493 -> 1543 | Splitting one oversized type-check expression into named sub-expressions necessarily adds declaration lines. Moving the helpers to a separate file would force the `private` state they read (`isPageFocused`, `currentImage`, `searchViewMode`, the view model) to widen to internal across the module, which trades a local line count for a wider API surface.

Failure-Class: new

🤖 Generated with Cursor

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11537?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-swiftui-body-typecheck-timeout.json
+++ b/.github/failure-classes/FC-swiftui-body-typecheck-timeout.json
@@ -1,0 +1,13 @@
+{
+  "schema_version": 1,
+  "id": "FC-swiftui-body-typecheck-timeout",
+  "violated_contract": "A SwiftUI View.body that compiles on CI's pinned Xcode must also compile on the current maintainer local toolchain. A single modifier chain the Swift 6.2 type checker cannot solve in reasonable time blocks every local named-bundle QA path while CI stays green.",
+  "canonical_prevention": "Break the failing body into named some View sub-expressions and extract Combine publishers from the SwiftUI modifier chain so each expression type-checks independently, preserving original modifier order.",
+  "evidence_prs": [
+    11536
+  ],
+  "scope_hints": [
+    "desktop/macos/Desktop/Sources/**/*.swift"
+  ],
+  "status": "open"
+}
--- a/desktop/macos/Desktop/Sources/Rewind/UI/RewindPage.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/UI/RewindPage.swift
@@ -151,6 +151,16 @@ struct RewindPage: View {
   }
 
   var body: some View {
+    keyboardHandledContent
+  }
+
+  // MARK: - Page modifier groups
+  //
+  // `body` used to be one ~15-modifier chain. Swift 6.2 cannot type-check that expression in
+  // reasonable time (Xcode 26.2). Each helper below is a distinct sub-expression; modifier order
+  // matches the original chain and is semantically significant.
+
+  private var chromeContent: some View {
     pageContent
       .glassContent()
       // The page answers arrow keys wherever the pointer is, so it holds keyboard focus itself. The
@@ -159,6 +169,10 @@ struct RewindPage: View {
       // window with no visible extent is a blue rectangle on the wallpaper. See
       // `shellPageKeyboardTarget`.
       .shellPageKeyboardTarget($isPageFocused)
+  }
+
+  private var lifecycleAttachedContent: some View {
+    chromeContent
       .task {
         await viewModel.loadInitialData()
         await resolveCitationFocusIfNeeded()
@@ -168,6 +182,10 @@ struct RewindPage: View {
         screenCaptureHealth = ProactiveAssistantsPlugin.shared.screenCaptureHealth
         isPageFocused = true
       }
+  }
+
+  private var monitoringNotificationContent: some View {
+    lifecycleAttachedContent
       .onReceive(NotificationCenter.default.publisher(for: .assistantMonitoringStateDidChange)) { _ in
         let pluginState = ProactiveAssistantsPlugin.shared.isMonitoring
         let state = RewindCaptureState.afterMonitoringChange(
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
Follow-up nit for whoever reviews/merges this: `.github/failure-classes/FC-swiftui-body-typecheck-timeout.json` currently lists `"evidence_prs": [11536]`. 11536 is the sibling director-ownership PR; this one should cite **11537**. One-line change, not worth another CI cycle on its own — please fold it in on merge, or I can push it separately.

(I attempted the correction locally but the pre-push desktop build could not re-resolve dependencies: SwiftPM served `FluidInference/FluidAudio` "from cache" and that cached mirror lacks pinned commit `19600a485`, even though the commit is fetchable from the upstream remote. Clearing `~/Library/Caches/org.swift.swiftpm` should fix it for anyone who hits the same wall after wiping `Desktop/.build`.)

Git-on-my-level:
**This is blocking the desktop CI release lane, not just local builds — please prioritize.**

I originally wrote this up as a local-toolchain problem (Xcode 26.2) that CI could not observe. That was wrong, and here is the evidence:

- PR #11538 (a desktop change branched from current `main`, **without** this fix) just failed **Desktop Swift Release Compile** on the CI runner with the identical error: `/Users/runner/work/omi/omi/desktop/macos/Desktop/Sources/Rewind/UI/RewindPage.swift:153:23: error: the compiler is unable to type-check this expression in reasonable time`.
- This PR, which carries the fix, **passed** Desktop Swift Release Compile in 18m47s.

So any desktop PR cut from current `main` fails the release lane until this lands.

Why `main` looks green: the recent commits on `main` are backend-only, so `Detect Desktop Swift Changes` skips the desktop jobs entirely (e.g. run 31759118584 on `eb90b610` shows both desktop lanes `skipped`). Nothing has actually release-compiled the desktop app on `main` since the regression landed.

Where it came from: `RewindPage.body` was restructured and grew several observers in #11519 (`00cc9ed9d6`, "feat(desktop): add clickable source citations"), merged earlier today — that is the commit that pushed the chain past the type checker's budget.

Merging this first should unblock #11536 and #11538, both of which branch from `main` without it.

Git-on-my-level:
Closing this as **superseded** — the fix already landed on `main` independently, and rebasing would just force a duplicate.

**The conflict is the fix itself.** `main` now has commit `37aa667262` ("fix(desktop): split RewindPage.body so release compile type-checks", *"Codemagic v0.12.170 failed swift build -c release on a 168-line modifier chain"*). That splits the same `body` into `pageChrome` / `rewindPageLifecycle` / `rewindPageKeyHandlers`, and ships its own changelog fragment. Two independent splits of one expression cannot merge cleanly. Same defect, same root cause (#11519), already resolved — so this PR has nothing left to contribute.

**On the failing checks — none of them are caused by this PR.** The failing step is `Test desktop Swift app`, and the suites are:

- `ContextBucketPromptAssemblerTests` — the prompt-equality test that had been red on `main` since #11468. Fixed by #11536, which this branch predates.
- `AgentPillLifecycleTests`, `ChatTimelineContinuityTests`, `MarkdownNestedScrollTests` — `error: Exited with unexpected signal code 11`. Test-runner segfaults, unrelated to this diff.
- `ChatCitationTests.testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID` — asserts on the global singleton `RewindCitationFocusState.shared` (`request(42)` then `consume()`). It touches no view code and is order-dependent on shared state; worth a look on its own, but not attributable here.

**Correcting my own earlier comment on this PR:** I previously reported the static lane as failing on `TERM case never launched detached app`, a `run.sh` build-lock timeout, and `workflow test-install job must semantically match the complete-job fixture`. That was a misread — I grepped the job log for `FAIL:` and those lines are *expected* output from mutation sub-cases where the guard is supposed to reject. `bash desktop/macos/tests/test-test-install-workflow-contract.sh` exits **0** locally. Sorry for the noise.

The one thing here not on `main` is the `FC-swiftui-body-typecheck-timeout` registry entry; `37aa667262` declared `Failure-Class: none`. Happy to open a one-file PR for it if you want the class recorded — otherwise nothing else is lost.

PR #11538, opened 2026-08-14
fix(desktop): scope proactive-lane quota cooldown per operation

Git-on-my-level:
## What changed and why

PR #11527 added a quota cooldown to `ProactiveLaneClient` so a 429 stops the client from hammering the desktop backend. That was the right idea, but the cooldown was scoped wrong and unbounded. This PR keys it per operation, clamps `Retry-After` to `[60s, 3600s]`, and records self-suppressed attempts as `quota_cooldown` instead of a fake server 429.

Decision semantics are unchanged: terminal state stays `silence`/`failed`, no user-visible surface, no retry loop.

### 1. One cooldown covered two independent server quotas

Server side, `backend/routers/desktop_proactivity.py` reserves quota under a **per-operation** Redis key: `f"desktop_{operation_value}"`, with separate ceilings `_OPERATION_DAILY_LIMITS = {"proactive_extraction": 200, "proactive_reasoning": 100}`. Extraction and reasoning have completely independent budgets.

Client side, `ProactiveLaneClient.complete(...)` called `checkQuotaCooldown()` at the very top — before the `operation` argument was considered at all. Both callers share `ProactiveLaneClient.shared`:
- `ContextBucketRollupWriter` (extraction) at `ContextBucketRollup.swift:609`, `ModelQoS.Proactivity.extractionOperation`
- `ContextProactivityEngine` (director / reasoning) at `ContextProactivityEngine.swift:97`, `ModelQoS.Proactivity.reasoningOperation`

Extraction runs on every qualified visit departure and has the larger ceiling (200), so it exhausts first. One 429 then suppressed the director's reasoning lane too, even though reasoning has its own untouched budget. The director is the only path that can produce a user-visible proactive notification, so a partial quota problem became a total notification outage.

### 2. The cooldown was unbounded

`backend/database/redis_db.py` `reserve_rate_limit` returns `retry_after = ttl`, the remaining TTL of a **fixed 24-hour window key** (`_QUOTA_WINDOW_SECONDS = 24 * 60 * 60`) — not a short per-slot delay. So `Retry-After` can legitimately be many hours. `armQuotaCooldown` accepted it verbatim with no upper bound, so a single late-in-window 429 could disable the proactive lane for the rest of the day in that process. Nothing re-checked sooner if the operator raised the limit or the window rolled.

### 3. A self-suppressed attempt was indistinguishable from a real server 429

`checkQuotaCooldown()` threw `.http(status: 429, ...)`, which `ProactiveLaneFailureClassification` recorded as `{"failure":"http_error","status":429}` — byte-identical to an actual server rejection. In the delivery ledger you could not tell "we asked the server and it refused" from "we never sent anything".

none

## How it was verified

- `cd desktop/macos && ./scripts/dev-feedback.py --once swift 'ProactiveLaneClientTests'` — **10/10 passed**. Headline: a 429 on `proactive_extraction` does **not** suppress a subsequent `proactive_reasoning` call (the reasoning request reaches the URLProtocol stub; a later extraction attempt in the same window does not). Same-operation suppress-then-resume, `Retry-After: 86400` clamped to 3600s, and tiny/zero `Retry-After` armed at least the 60s floor all passed. Injected clock; no wall-clock sleeps.
- `cd desktop/macos && ./scripts/dev-feedback.py --once swift 'ContextProactivityEngineTests|ContextProactivityDirectorFailureTests'` — **19/19 passed**. Stored delivery row is `{"failure":"quota_cooldown","status":429}` for a cooldown-skipped director attempt and `{"failure":"http_error","status":429}` for a real server 429.
- Local test compile required a temporary copy of the `RewindPage.swift` split from open PR #11537. That type-check timeout (`unable to type-check this expression in reasonable time`) is pre-existing on unmodified `origin/main` with local Xcode 26.2 / Swift 6.2.3; CI pins Xcode 16.4. The split was reverted after the tests ran and is **not** in this PR.
- I did not re-run the live quota-exhaustion path against a signed bundle; the regression is the hermetic client/engine tests above.

## Tests

- `ProactiveLaneClientTests.testExtractionQuotaCooldownDoesNotSuppressReasoning` — the headline test that fails on the pre-change shared cooldown.
- `ProactiveLaneClientTests.testExtractionSkipsNetworkDuringQuotaCooldownThenResumesAfterDeadline` — same operation is suppressed before the deadline and resumes after the injected clock passes it.
- `ProactiveLaneClientTests.testRetryAfterAboveADayIsClampedToOneHour` / `testTinyOrZeroRetryAfterArmsAtLeastTheFloor` — ceiling 3600s, floor 60s.
- `ContextProactivityDirectorFailureTests.testEngineRecordsQuotaCooldownProvenanceOnSelfSuppressedLaneError` / `testEngineRecordsHttp429ProvenanceOnTypedLaneError` — assert on the stored delivery row, not a source string.

## Failure class (fixes)

Failure-Class: new

**Violated contract:** a client-side cooldown that backs off from a server quota must be keyed to the same identity the server uses to reserve that quota. Sharing one process-wide deadline across independently governed operations converts a partial exhaustion into a total outage of unrelated lanes.

**Canonical guard:** `ProactiveLaneClient` now keys `quotaCooldownUntil` by the `operation` string, clamps `Retry-After` into `[60s, 3600s]`, and throws `.quotaCooldown` so provenance records `{"failure":"quota_cooldown","status":429}` instead of `http_error`. Behavioral tests drive production `complete()` through a URLProtocol stub with an injected clock, and assert the director delivery row.

**Evidence:** #11527 shipped the actor-wide cooldown. Server policy is per-operation (`desktop_{operation}` Redis key; extraction 200 / reasoning 100) with a fixed 24h window TTL feeding `Retry-After`. This PR is recorded as `evidence_prs: [11538]`.

## New guards

Not a new CI check. The reusable surface is the per-operation cooldown inside `ProactiveLaneClient` (not a new subsystem), proven by `testExtractionQuotaCooldownDoesNotSuppressReasoning`. #11527 is the merged PR that would have caught this if the headline test had existed.

## Local build hatch

Local pre-push `xcrun swift build -c debug --package-path Desktop` fails on unmodified `RewindPage.swift` under Xcode 26.2 / Swift 6.2.3 (same failure on `origin/main`; being fixed in #11537). This PR does not touch that file. Push uses the documented hatch `PRE_PUSH_SKIP_DESKTOP_SWIFT=1`. CI pins Xcode 16.4 and is unaffected. The focused test targets above still ran and passed.

🤖 Generated with Cursor

### DIFF
--- a/.github/failure-classes/FC-shared-backoff-conflated-independent-budgets.json
+++ b/.github/failure-classes/FC-shared-backoff-conflated-independent-budgets.json
@@ -8,7 +8,7 @@
     "desktop/macos/Desktop/Tests/ContextProactivityEngineTests.swift"
   ],
   "evidence_prs": [
-    11527
+    11538
   ],
   "scope_hints": [
     "desktop/macos/Desktop/Sources/ProactiveAssistants/**"
--- a/.github/failure-classes/FC-shared-backoff-conflated-independent-budgets.json
+++ b/.github/failure-classes/FC-shared-backoff-conflated-independent-budgets.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-shared-backoff-conflated-independent-budgets",
+  "violated_contract": "A client-side cooldown that backs off from a server quota must be keyed to the same identity the server uses to reserve that quota. Sharing one process-wide deadline across independently governed operations converts a partial exhaustion into a total outage of unrelated lanes.",
+  "canonical_prevention": "Key the cooldown by the operation string the server reserves, clamp Retry-After into a closed [60s, 3600s] range so the lane re-checks within an hour, and record self-suppressed attempts as quota_cooldown rather than http_error.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/Desktop/Tests/ProactiveLaneClientTests.swift",
+    "desktop/macos/Desktop/Tests/ContextProactivityEngineTests.swift"
+  ],
+  "evidence_prs": [
+    11527
+  ],
+  "scope_hints": [
+    "desktop/macos/Desktop/Sources/ProactiveAssistants/**"
+  ],
+  "status": "open"
+}
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketRollup.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketRollup.swift
@@ -641,10 +641,14 @@ actor ContextBucketRollupWriter {
         normalizedContextKey: ContextTitleNormalizer.identityKey(
           appName: frame.appName, windowTitle: frame.windowTitle) ?? "")
     } catch {
-      if case .http(let status, _) = error as? ProactiveLaneClientError, status == 429 {
+      switch error as? ProactiveLaneClientError {
+      case .http(let status, _) where status == 429:
         return
+      case .quotaCooldown(_):
+        return
+      default:
+        log("Context bucket extraction failed silently: \(error.localizedDescription)")
       }
-      log("Context bucket extraction failed silently: \(error.localizedDescription)")
     }
   }
 
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ProactiveLaneClient.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ProactiveLaneClient.swift
@@ -127,6 +127,7 @@ actor ProactiveLaneClient {
   private var quotaCooldownUntil: [String: Date] = [:]
   private var loggedQuotaSkip: Set<String> = []
   private var loggedQuotaClamp: Set<String> = []
+  private var cooldownOwner: String?
 
   init(
     session: URLSession = .shared,
@@ -154,6 +155,8 @@ actor ProactiveLaneClient {
     maxCompletionTokens: Int = 1024,
     authorizationSnapshot: RuntimeOwnerAuthorizationSnapshot? = nil
   ) async throws -> ProactiveLaneResult {
+    let currentOwner = authorizationSnapshot?.ownerID
+    clearCooldownsIfOwnerChanged(currentOwner)
     try checkQuotaCooldown(operation: operation)
     if let authorizationSnapshot {
       guard RuntimeOwnerIdentity.isAuthorizationCurrent(authorizationSnapshot) else {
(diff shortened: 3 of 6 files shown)


Git-on-my-level:
Heads-up: the failing **Desktop Swift Release Compile** here is not caused by this PR. It fails on `RewindPage.swift:153` (`unable to type-check this expression in reasonable time`), which is a pre-existing regression on `main` introduced by #11519 and fixed in #11537. #11537 passes the same lane. Rebase on top of #11537 once it merges and this lane should go green.

Git-on-my-level:
**Correction to my previous comment.** I claimed the `TERM case never launched detached app` / `sentinel target build failed` steps were downstream of the RewindPage compile error. I tested that and it is **wrong**.

#11537 carries the RewindPage fix, and its `Desktop Swift Static & Test Contracts` lane **still fails**, on:

```
FAIL: TERM case never launched detached app
ERROR: timed out after 4s waiting for ./run.sh build lock (.../omi-run-sh-lock-test.../wt-a/.dev/run-sh-build.lock.d)
FAIL: workflow test-install job must semantically match the complete-job fixture   (x4)
```

So there are two independent problems on the desktop lanes right now:

1. **RewindPage type-check** (#11519 regression, fixed by #11537) — this genuinely explains the `Desktop Swift Release Compile` failures and any compile of the app.
2. **A separately broken `Desktop Swift Static & Test Contracts` lane** — launcher-script tests: a detached-launch TERM case, a `run.sh` build-lock timeout, and a workflow-fixture mismatch. These reproduce **with** the RewindPage fix applied, so they are not caused by it and are not caused by this PR either. None of these three PRs touch `.github/workflows/*.yml` or `run.sh`.

Net: #11537 still unblocks the release-compile lane and should still land first, but it will **not** by itself turn Static & Test Contracts green. That lane looks broken for every desktop PR at the moment and needs a separate owner.

Git-on-my-level:
**Correcting my earlier comment on this PR.** I attributed the `Desktop Swift Static & Test Contracts` failure partly to `TERM case never launched detached app` / `sentinel target build failed` / a workflow-fixture mismatch. That was a misread on my part: I grepped the job log for `FAIL:` and those lines are *expected* output from mutation sub-cases where the guard is meant to reject. `desktop/macos/tests/test-test-install-workflow-contract.sh` exits **0** when run directly.

The lane's actual failing step is `Test desktop Swift app`. On the runs I inspected the failing suites were `ContextBucketPromptAssemblerTests` (the equality test red on `main` since #11468, fixed by #11536), plus `AgentPillLifecycleTests` / `ChatTimelineContinuityTests` / `MarkdownNestedScrollTests` failing with `error: Exited with unexpected signal code 11` (test-runner segfaults), and `ChatCitationTests.testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID`, which asserts on the global singleton `RewindCitationFocusState.shared` and is order-dependent on shared state.

Also for the record: the RewindPage type-check regression from #11519 was fixed directly on `main` in `37aa667262` (prompted by Codemagic v0.12.170 failing `swift build -c release`), so my #11537 was superseded and is now closed.

PR #11554, opened 2026-08-14
fix(desktop): context-buckets mode must not silence the focus suggestion assistant

kodjima33:
## What

The `context_buckets` rollout silently killed focus (suggestion) nudges for everyone in the flag cohort — and the flag was at **100% of all users**. Nik's last delivered focus nudge before the fix: Aug 12 23:06; nothing after, with zero errors logged anywhere. Insights/memories/tasks kept working, which is exactly the signature: focus is the only assistant whose pipeline starts from a context switch.

Three defects, one per commit hunk:

1. **Coordinator starved the assistants.** The buckets path of `AssistantCoordinator.checkContextSwitch` fired `onContextSwitch` only on `assistants["task-extraction"]`. `SuggestionAssistant.analyze` returns nil before any logging when no switch ever armed its dwell anchor. Now a shared `fireContextSwitchOnAllAssistants` fires on every registered assistant in buckets mode and out of it — awaited, and outside the bucket-persistence do/catch.
2. **`SuggestionAssistant.isEnabled` was hard-gated on `!ContextBucketsFeature.isEnabled`** — the context director was supposed to replace live suggestions and delivered almost nothing. The gate is removed; replacing the assistant must be an explicit evidenced change, not a rollout side effect.
3. **Context tracking died with the bucket DB.** `lastTrackedApp` was only assigned inside the `do` blocks, so any bucket-visit write failure left tracking unprimed and every assistant starved. Tracking now updates regardless of persistence health. (Found by the new regression test failing in the hermetic test host, where the bucket store throws.)

Live mitigation already shipped separately: the PostHog flag (id 817239) was disabled at 02:00; this PR makes re-enabling it safe.

## Product invariants

None matched (`scripts/pr-preflight --suggest`: no invariant IDs emitted for these paths).

Failure-Class: none

## Verification

- New deterministic regression test `testBucketsModeDispatchesContextSwitchToEveryAssistant` (spy assistant hears switches with buckets forced on; no wall-clock sleeps — dispatch is awaited, registration visibility via bounded `Task.yield()`); `AssistantCoordinatorTests` 4/4 green; `check_desktop_test_quality.py` at/below baseline (16 waits, unchanged).
- Live on `omi-ankara` built from this branch with `OMI_FORCE_CONTEXT_BUCKETS=1` (real signed-in account):
  - `probe_suggestion_nudge` pre-fix returned `{"outcome": "disabled"}`; post-fix it ran the full pipeline — `{commitments: 3, goals: 5, confidence: 80, outcome: filtered_low_confidence}` (honest below-bar refusal).
  - Organic proof: the running bundle logged real `Suggestion: skipped (skippedDwell) app=Warp` / `below bar [80% < 85%]` lines from the user's own app switches with buckets ON — the exact signature absent all day on broken builds.
- Fleet evidence of the outage: PostHog fleet gate outcomes continued (other users mostly weren't on builds with the buckets code yet); the affected machine logged zero `Suggestion:` lines across every app instance from Aug 13 02:59 (first build carrying the rollout) until the flag was disabled.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/Suggestions/SuggestionAssistant.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/Suggestions/SuggestionAssistant.swift
@@ -20,8 +20,14 @@ actor SuggestionAssistant: ProactiveAssistant {
 
   var isEnabled: Bool {
     get async {
+      // Deliberately independent of ContextBucketsFeature. The buckets rollout gated this
+      // on `!ContextBucketsFeature.isEnabled`, betting the context director would replace
+      // live suggestions — it delivered almost nothing, and with the flag at 100% of all
+      // users focus nudges went silent fleet-wide with no error logged (Aug 13–14 2026).
+      // If the director is ever meant to replace this assistant again, that must be an
+      // explicit, evidenced change — never a side effect of a rollout flag.
       await MainActor.run {
-        !ContextBucketsFeature.isEnabled && SuggestionAssistantSettings.shared.isEnabled
+        SuggestionAssistantSettings.shared.isEnabled
       }
     }
   }
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/AssistantCoordinator.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/AssistantCoordinator.swift
@@ -114,8 +114,13 @@ class AssistantCoordinator {
   /// and window title changes — one unified path with one delay mechanism.
   /// - Returns: `true` if a context switch was detected and fired.
   @discardableResult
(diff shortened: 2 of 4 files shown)


kodjima33:
Superseded by #11562, which contains these commits rebased plus the Maximum pacing work verified in the 10-trial live marathon.

PR #11528, opened 2026-08-13
feat(desktop): speak proactive notifications aloud (off-by-default Advanced setting) + unified voice speed

kodjima33:
## What

Demo feature: **Speak Notifications Aloud** — an off-by-default toggle in Advanced settings → Preferences. When enabled, proactive notifications (suggestions, insights, tasks, context-director) are read out loud through the existing floating-bar voice pipeline (user's chat voice, system-voice fallback) at the moment a surface actually presents them. Plus a voice-speed parity fix the demo surfaced: the system-voice path ignored the Voice Speed setting.

- `NotificationSpeech`: pure policy — speaks only when the user opted in, only for proactive notifications (`respectFrequency` marker; functional notices like the onboarding test and screen-repair prompts never speak), only non-empty messages.
- `NotificationSpeechOnDelivery`: once-per-notification speaker hooked into the shared presentation callbacks in `NotificationService`, so bar-immediate, bar-queued, and system-banner deliveries each speak exactly once and a muted-preview dual delivery cannot speak twice.
- `FloatingBarVoicePlaybackService.systemSpeechRate`: system voice scales its utterance rate by `ShortcutSettings.voicePlaybackSpeed` (default 1.4×) exactly like the OpenAI audio path applies it via `AVAudioPlayer.rate`, clamped to `AVSpeechUtterance`'s legal range — one Voice Speed setting drives both output paths.

## Product invariants

- **INV-CHAT-1** — path-matched (`NotificationService.swift`). No transcript behavior change: delivery surfaces, chat journaling, and notification persistence are untouched; the change only adds a speech side-effect at the existing presentation callbacks.
- **INV-VOICE-1** — path-matched (`FloatingBarVoicePlaybackService.swift`). Voice-turn lifecycle ownership unchanged: speech goes through the existing `speakOneShot`, which already acquires/validates PTT leases via `VoiceTurnCoordinator`; the rate fix only changes `AVSpeechUtterance.rate` arithmetic.

Failure-Class: none

## Verification

- Focused suites green: `xcrun swift test --filter "NotificationSpeechTests|FloatingBarVoiceResponseSettingsTests"` — NotificationSpeechTests 7/7 (default-off contract, proactive gate, once-guard dedup, functional-notice silence) + `testSystemSpeechRateScalesWithVoiceSpeed` (scaling + clamping).
- Live on `omi-ankara` (this branch on top of main, real signed-in account):
[...]

PR #11556, opened 2026-08-14
fix(desktop): media playback exempts the capture idle gate

kodjima33:
## What

Movie-watching no longer counts as being away: the capture loop's idle gate skipped every tick after 60 s without HID input, so an hour of passive video (Nik's report: Maximum notifications, one nudge in a movie hour, 14 `skipping capture ticks (idle)` cycles logged) produced almost no screen analyses and therefore almost no proactive notifications.

- `MediaPlaybackIdlePolicy.effectiveIdleSeconds` pins the gate's idle value to 0 while another process holds a display-sleep-prevention assertion (`PreventUserIdleDisplaySleep` / legacy `NoDisplaySleepAssertion` — what browsers and video players hold during playback). No media → HID idle passes through untouched.
- `MediaPlaybackDetector`: 10 s-cached `IOPMCopyAssertionsByProcess` walk, ignoring our own pid (our capture stack's own assertions must not count as "the user is watching").
- One log line per idle-but-watching episode (`CaptureGate: HID-idle but media playback active — capture continues`) for observability.

## Product invariants

None matched (path-based check: none).

Failure-Class: none

## Verification

- `MediaPlaybackIdleExemptionTests` 3/3 (policy pass-through / pin-to-zero, detector TTL cache re-probe).
- IOKit probe live-tested both ways on the target machine: standalone run of the identical assertion-walk code printed `PREVENTED=false` at rest and `PREVENTED=true` under a held `PreventUserIdleDisplaySleep` assertion.
- **End-to-end on the running app** (omi-notch built from this branch, real user idleness, real assertion): `02:44:46 CaptureGate: HID-idle but media playback active — capture continues` → `02:45:09 RewindIndexer: Last 60s — 4 frames…`, zero idle skips during the hold.
- **Negative path live**: with the assertion lapsed and the user idle, `02:27:09 CaptureGate: skipping capture ticks (idle)` — no media keeps the old power-saving behavior.

Line-Count-Exception: desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift | 1761 -> 1779 | The idle-gate exemption must live at the gate check itself (effective-idle computation + once-per-episode log flag); the policy and detector logic already live in the new MediaPlaybackIdleExemption.swift, and splitting the plugin for 18 gate-site lines would fragment the capture-gate sequence.

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/MediaPlaybackIdleExemption.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/MediaPlaybackIdleExemption.swift
@@ -1,0 +1,76 @@
+import Foundation
+import IOKit.pwr_mgt
+
+/// Watching a movie is not being away from the machine.
+///
+/// The capture loop's idle gate reads HID idle time, and a person watching video
+/// produces no HID input — so after 60 quiet seconds every capture tick was skipped
+/// and the proactive assistants went blind exactly when the user asked "why no nudges
+/// while I watch TikTok for an hour?" (measured live: 14 idle-skip cycles in one
+/// movie hour, near-zero analyses at Maximum frequency).
+///
+/// The system already knows the difference: video players and browsers hold a
+/// power-management assertion that prevents display sleep while media plays
+/// (`PreventUserIdleDisplaySleep`, or the legacy `NoDisplaySleepAssertion`). While any
+/// other process holds one, the display would stay lit for a human who is presumed to
+/// be watching — the same presumption the idle gate should make.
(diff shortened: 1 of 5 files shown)


kodjima33:
Merging past two known-flaky checks, same pair and precedent as #11528 tonight: 'TERM case never launched detached app' and 'workflow test-install fixture' fail identically on unrelated diffs (previously on a 2-line color-token change). This PR touches only the capture idle gate (Swift policy + tests + e2e cover + changelog); Backend Hermetic Merge Gate and all other checks are green. Fix verified live 4× (transitions at 02:44/02:56/03:03/03:25, frames flowing, negative path preserved).

Issue #11517, opened 2026-08-13
Desktop: 'Removed by me' is structurally empty — deleteActionItemByBackendId ignores its deletedBy argument and hard-deletes


formed2forge:
The desktop "Removed by me" filter can never contain anything, and the task-extraction model is never told which tasks the user deleted, because the only code that would record user provenance passes it to a function that ignores it.

`ActionItemStorage.deleteActionItemByBackendId` accepts a `deletedBy` argument and never reads it — it hard-deletes the row (`desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift:1186-1207`):

```swift
/// Hard-delete an action item by backend ID
func deleteActionItemByBackendId(
  _ backendId: String,
  deletedBy: String? = nil,           // never referenced in the body
  authorization: LocalMutationAuthorization
) async throws {
  ...
      if let record = try Self.fetchRecord(database, surfacedId: backendId) {
        try record.delete(database)   // hard delete, no soft-delete write
      }
```

Its caller believes the opposite (`desktop/macos/Desktop/Sources/Stores/TasksStore.swift:3179-3186`):

```swift
// Local-first: soft-delete in SQLite immediately for instant UI update
try await ActionItemStorage.shared.deleteActionItemByBackendId(
  task.id,
  deletedBy: "user",
```

Both halves of that comment are wrong: the row is hard-deleted, not soft-deleted, and the provenance is discarded.

Two user-visible consequences:

1. **"Removed by me" is structurally empty.** `TasksPage` partitions the Removed lane on that column (`desktop/macos/Desktop/Sources/MainWindow/Pages/TasksPage.swift:200-201`):
   ```swift
   case .removedByAI: return task.deleted == true && task.deletedBy != "user"
   case .removedByMe: return task.deleted == true && task.deletedBy == "user"
   ```
   Since nothing ever writes `"user"`, every retired row has `deletedBy` NULL, and `nil != "user"` is true — so everything lands in "Removed by AI".

2. **Task extraction never learns user deletions.** `TaskAssistant` feeds the extraction model a "don't re-extract what the user deleted" list (`desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/TaskExtraction/TaskAssistant.swift:1588`):
   ```swift
   deletedTasks = try await ActionItemStorage.shared.getRecentDeletedTasks(limit: 10, deletedBy: "user")
   ```
[...]

PR #11559, opened 2026-08-14
fix(desktop): keep user-deletion provenance so "Removed by me" is not empty (#11517)

kodjima33:
## Bug (#11517)

Desktop **"Removed by me" is structurally empty**, and task extraction never learns which tasks the user deleted.

`TasksStore.deleteTask` tombstones a synced task with `deletedBy = "user"` — and then the acknowledgement path **hard-deleted** the row (`deleteActionItemByBackendId`, whose `deletedBy:` argument was accepted and never read). The Removed lane re-fetches deleted tasks from the backend, and the backend has no `deleted_by` field, so the row came back with `deletedBy: nil`. `TasksPage` partitions the lane on exactly that column:

```swift
case .removedByAI: return task.deleted == true && task.deletedBy != "user"
case .removedByMe: return task.deleted == true && task.deletedBy == "user"
```

so every user deletion landed under "Removed by AI". The same NULL empties `getRecentDeletedTasks(deletedBy: "user")` (SQL `= ?` never matches NULL) — the list `TaskAssistant` feeds the extraction model as "don't re-extract these" — so the model kept re-suggesting tasks the user had explicitly removed.

Second clobber on the same column: `ActionItemRecord.updateFrom(_ item: TaskActionItem)` copied the server's (always nil) `deleted_by` over the local value, so hydration erased the attribution even inside the pending window.

## Fix

- **`ActionItemStorage.markActionItemDeletionAcknowledged(backendId:)`** — the ack now clears the pending flag and keeps the tombstone (`deleted = 1`, `deletedBy` intact) instead of purging the row. The retirement is already durable server-side; the local row is the only place the provenance exists. Used by both ack callers (`deleteTask` after a 2xx, and the pending-deletion flush).
- **`updateFrom(_:)`** keeps local provenance when the server omits `deleted_by`, and clears it only when the server reports the task live again (nothing left to attribute).
- Dropped the never-read `deletedBy:` parameter from `deleteActionItemByBackendId` and documented its two genuine hard-delete callers (a local-only task the server never saw; the undo purge before a restore re-inserts).

Every live-lane query in `ActionItemStorage` already pairs `completed` with `deleted == false` (audited all 17), so a retained tombstone stays invisible — asserted in the test.

## Tested

`swift test --filter ActionItemDeletionSyncTests` — **5 passed, 0 failures** (3 new/updated: the ack keeps `deleted = 1` + `deletedBy = "user"` and stays out of the live lane; the deleted-page hydration keeps that attribution and the row shows up in `getRecentDeletedTasks(deletedBy: "user")`; a server row that comes back live clears the stale provenance).

**The regression test provably catches the bug.** Reverting only the `updateFrom` hunk and re-running:

```
ActionItemDeletionSyncTests.swift:126: XCTAssertEqual failed: ("nil") is not equal to ("Optional("user")")
  - hydration must not re-attribute a user deletion to the AI
ActionItemDeletionSyncTests.swift:131: XCTAssertTrue failed
  - task extraction reads this list to stop re-suggesting what the user removed
Executed 5 tests, with 2 failures
```

Adjacent suites green after the change: `ActionItemLocalIdentityMutationTests` (6), `TasksStoreOwnerBoundaryTests` (29, including the static mutation-authority guard the new function is registered in), `LocalMutationAuthorizationTests` (5), `ActionItemStorageVisibilityReconciliationTests` (2). `swift build --build-tests` clean.

Not verified through the running app: exercising the lane end-to-end needs a signed-in desktop build against the real backend. The lane predicate is a direct read of the column this PR fixes (`task.deleted == true && task.deletedBy == "user"`), and the storage-level assertions cover the write and both hydration directions.

Line-Count-Exception: desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift | 1795 -> 1830 | adds the acknowledgement path next to the tombstone write it pairs with; splitting this file is out of scope for a bug fix

## Known gap (tracked, not this PR)

Bulk delete (`Select All → Delete` → `ActionItemStorage.deleteActionItemsByBackendIds`) still hard-deletes its rows, so bulk deletions remain unattributed. Same failure class, different function; noted on #11517 rather than widened into this fix.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11559?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Rewind/Core/ActionItemModels.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Core/ActionItemModels.swift
@@ -374,7 +374,15 @@ extension ActionItemRecord {
     self.description = item.description
     self.completed = item.completed
     self.deleted = item.isRetired
-    self.deletedBy = item.deletedBy
+    // The backend has no `deleted_by` field, so a server row always reports nil here.
+    // Retirement provenance ("user" / "ai_dedup" / "staged") is local-only state: taking
+    // the server's nil re-attributed every user deletion to the AI on the next hydration.
+    // A task the server reports as live has no retirement left to attribute.
+    if let itemDeletedBy = item.deletedBy {
+      self.deletedBy = itemDeletedBy
+    } else if !self.deleted {
+      self.deletedBy = nil
+    }
     self.source = item.source
     self.conversationId = item.conversationId
     self.priority = item.priority
--- a/desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift
@@ -1197,7 +1197,6 @@ actor ActionItemStorage {
     return count
   }
 
-  /// Hard-delete an action item by backend ID
   /// Mark a synced task deleted locally while the backend delete is still in flight.
   ///
   /// The previous shape was a local hard delete plus a fire-and-forget backend call: one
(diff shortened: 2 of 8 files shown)


kodjima33:
Closing this rather than leaving it open: the merge gate I run under requires green CI, and \`Desktop Swift Static & Test Contracts\` cannot go green on any branch right now — it is red on unmodified \`main\` (77 failures across 3 suites) and equally red on #11553. Filed as #11563 with the root cause of the largest suite (15a51fb985 added `.background(InkGlassHitRegionReporter())` to the modal scrim's host-sized barrier; reverting that one line takes ShellModalScrimTests from 74 failures to 0).

The fix itself is finished and independently verified — `make preflight` 20/20, `ActionItemDeletionSyncTests` 5/5, and reverting only the `updateFrom` hunk makes the new regression test fail, so it does catch the bug. Nothing about it is in question; only the lane is.

The branch is kept, unrebased, at `watchdog/issue-11517-deleted-by-user-provenance` (fa3aa517e6). Once #11563 is fixed, reopen this PR — no rebase should be needed. Issue #11517 stays open.

PR #11566, opened 2026-08-14
fix(desktop): unblock the Swift test lane — 77 failures on every desktop PR (#11563)

### LINKED ISSUE #11563: Desktop Swift test lane red on main: 77 failures across 3 suites block every desktop PR (scrim regression root-caused to 15a51fb985)
`Desktop Swift Static & Test Contracts` fails on every desktop PR right now, so no desktop change can reach `main` and the release train has no qualification signal. It is red on PR #11553 (58m25s, Nik's own branch) and on #11559, and I reproduced it locally on unmodified `main` at `d20f89153d` with the identical failure counts, so it is not any one branch's diff.

Three independent suites are red — 77 assertion failures total:

| Suite | Failures |
|---|---|
| `ShellModalScrimTests` (`Tests/PageGlassLaneTests.swift`) | 74 |
| `GlassSurfaceReviewRegressionTests` | 1 |
| `ProactiveLaneClientTests` | 2 |

`main` shows green because the lane is path-filtered: `Detect Desktop Swift Changes` skips the test job on every commit that does not touch `desktop/macos/**`, so runs 31779563942 / 31778317070 / 31778252500 all report `success` having skipped it. The red only appears on desktop PRs.

## 1. `ShellModalScrimTests` — 74 failures, root-caused to 15a51fb985

`15a51fb985` ("fix(desktop): pass clicks through transparent dead zones of the shell and notch windows", Nik Shevchenko, 2026-08-14 02:10 -0400) added a reporter to the modal scrim's host-sized transparent barrier — `desktop/macos/Desktop/Sources/MainWindow/ShellModalScrim.swift`:

```swift
Color.clear
  .contentShape(Rectangle())
  .onTapGesture { onTap?() }
  .background(InkGlassHitRegionReporter())   // <- added by 15a51fb985
```

`PaintedExtent.of` measures what the view actually paints, and the reporter makes that host-sized barrier paint, so the measured extent becomes the whole window instead of the shell's lane:

```
XCTAssertEqualWithAccuracy failed: ("3440.0") is not equal to ("900.0") +/- ("1.5")
  - the confirmation's card is drawn but its dim is not: at (3440.0, 1440.0) the paint is narrower than the shell's lane
XCTAssertGreaterThan failed: ("0.0") is not greater than ("0.0")
  - wholeShell at (3440.0, 1440.0): the confirmation painted to the top edge, which on this window is the user's wallpaper
```

Verified by reverting only that one line locally: `swift test --filter ShellModalScrimTests` goes from **7 tests / 74 failures** to **7 tests / 0 failures**. Nothing else changed.

The reporter is there for a real reason (it keeps the barrier inside the window's pointer ownership so a modal stays modal under the new click-through sync), so the fix is not a plain revert — the barrier needs to register its hit region without contributing paint, or the test's measurement has to exclude a zero-alpha reporter. That is a call for the author of 15a51fb985 / `InkGlassHitRegions`, which is why I am filing rather than patching it.

## 2. `GlassSurfaceReviewRegressionTests` — 1 failure

```
Tests/GlassSurfaceReviewRegressionTests.swift:61: testInkGlassPanelClipsReturnedContentButNotAnOuterOverlay
  XCTAssertLessThan failed: ("0.8") is not less than ("0.5")
  - the returned red content must not paint through the panel's rounded corner
```

Still fails with the 15a51fb985 hunk reverted, so it is a separate defect.

## 3. `ProactiveLaneClientTests` — 2 failures

```
Tests/ProactiveLaneClientTests.swift:303,312: testLaterShorterRetryWindowDoesNotShortenActiveCooldown
  XCTAssertEqual failed: ("1") is not equal to ("2")
  - the later shorter Retry-After must not have shortened the active cooldown
```

Also independent of 15a51fb985.

## Repro

```
git checkout d20f89153d          # unmodified main
cd desktop/macos/Desktop
swift test --filter ShellModalScrimTests              # 7 tests, 74 failures
swift test --filter GlassSurfaceReviewRegressionTests # 3 tests, 1 failure
swift test --filter ProactiveLaneClientTests          # 11 tests, 2 failures
```

Found by the hourly desktop bug watchdog while trying to land #11559; that PR is otherwise green (20/20 `make preflight`, its own suites pass) and is blocked purely by this.

kodjima33:
## Bug

`Desktop Swift Static & Test Contracts` has been red on **every** desktop PR since `15a51fb985` (2026-08-14 02:10 EDT), so no desktop change could reach `main` and the release train had no qualification signal. `main` reads green only because the lane is path-filtered: `Detect Desktop Swift Changes` skips the test job on every commit that does not touch `desktop/macos/**`, so the red only appears on desktop PRs. Filed as #11563.

The failing set, identical across CI runs 31778872737 and 31779149249:

| Suite | Failures |
|---|---|
| `ShellModalScrimTests` | 74 |
| `GlassSurfaceReviewRegressionTests` | 1 |
| `ProactiveLaneClientTests` | 2 |
| `ChatCitationTests` | 1 |

## Root causes

**1. 75 of the 77 — the hit-region marker paints.** `InkGlassHitRegionReporter` mounts a zero-drawing AppKit marker so a transparent window can tell visible content from air. `ImageRenderer` cannot rasterise an `NSViewRepresentable` and substitutes an opaque placeholder for it, so every pixel-truth test of a surface that mounts one measured the placeholder covering the marker's whole host: the modal dim read as painting to the window's edge — which on this window is the user's wallpaper — and the glass panel's rounded corner read as filled. Measured directly: `Color.clear.background(InkGlassHitRegionReporter())` rendered at 40×40 came back fully opaque (centre RGBA `255,56,60,255`), where `Color.clear` alone is `0,0,0,0`.

**2. `testLaterShorterRetryWindowDoesNotShortenActiveCooldown` (`ff974901b6`) was never reachable.** It issued its two 429s sequentially, but an armed cooldown short-circuits the network before the second request is built (as `testExtractionSkipsNetworkDuringQuotaCooldownThenResumesAfterDeadline` asserts), so the shorter `Retry-After` was never parsed and `requestCount` could not reach the 2 it asserted.

**3. `testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID` depended on a leaked global.** `RewindCitationFocusState.request` refuses an id when `RewindCaptureOwnerSnapshot.capture()` is nil, and the only thing that makes it nil is `RewindCaptureOwnerGeneration`'s suspended flag — which is process-global and outlives whichever suite suspended it. Class execution order is not stable across machines, which is the whole of the local/CI divergence: this one passes locally and has failed on CI since #11519.

## Fix

- Mount the marker at zero opacity. The placeholder then contributes no pixels, while the AppKit view stays mounted, unhidden and registered: SwiftUI applies opacity to the host layer, not to the view's own `alphaValue`, which is what `InkGlassHitRegions.containsPoint` reads. New `InkGlassHitRegionReporterTests` pins both halves — dropping the reporter would satisfy the paint half and silently return the shell to passing clicks through its own modals.
- Overlap the two 429s on a rendezvous, which is the only way a second `Retry-After` for one operation is ever observed, so the max-deadline preservation the test is named for is actually exercised.
- Start the citation-focus test from a clean fence and assert an owner exists, so it fails on the handoff it is named for or not at all.

No production behaviour changes: the only non-test edit is where the marker's opacity is applied.

## What I tested

- The six affected suites: **44/44 pass** (`ShellModalScrimTests`, `GlassSurfaceReviewRegressionTests`, `ProactiveLaneClientTests`, `ChatCitationTests`, `InkGlassHitRegionReporterTests`, `ShellClickThroughPolicyTests`) — was 77 failures.
- Full package suite on this branch: 5085 tests, 0 failures attributable to this change. Two unrelated failures reproduce identically on unmodified `main` sources (`KnowledgeGraphPaginationTests.testCatalogOnlyCanonicalAtlasIsRenderableAndActivationCached`, verified by reverting only this diff, and a `MemoryDetailPanelTests` SQLite disk-I/O flake that passes in isolation). Neither fails on CI.
- Root cause 3 reproduced under CI's precondition locally: with a suite that leaks the fence ahead of it, `consume()` returns nil exactly as CI reports without this change, and 42 with it.
- Click-through unaffected: `ShellClickThroughPolicyTests` plus the new registration test confirm a surface mounting the marker still owns the pointer.
- `make preflight` and the macOS-scoped manifest checks (swift-format lint, SwiftLint 1258 files / 0 violations) pass.

Labelled `no-changelog-needed`: nothing here changes what a user sees.

Fixes #11563.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11566?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Theme/InkGlassHitRegions.swift
+++ b/desktop/macos/Desktop/Sources/Theme/InkGlassHitRegions.swift
@@ -70,12 +70,27 @@ package final class InkGlassHitRegionView: NSView {
 
 /// SwiftUI mount for the marker. Attach as a `.background` so it spans exactly the surface that
 /// should own the pointer.
-package struct InkGlassHitRegionReporter: NSViewRepresentable {
+///
+/// Mounted at zero opacity, which is load-bearing rather than decorative. The marker paints nothing
+/// on screen, but `ImageRenderer` cannot rasterise an `NSViewRepresentable` and substitutes an
+/// opaque placeholder for it — so a marker at full opacity reads as *painted content covering its
+/// whole host* to every pixel-truth test of a surface that mounts one, which is a claim about the
+/// app that is not true. At zero opacity the placeholder contributes no pixels while the AppKit view
+/// stays mounted, unhidden and registered — SwiftUI applies the opacity to the host layer, not to the
+/// view's own `alphaValue`, which is what `InkGlassHitRegions.containsPoint` reads.
+/// `InkGlassHitRegionReporterTests` holds both halves.
+package struct InkGlassHitRegionReporter: View {
   package init() {}
 
-  package func makeNSView(context: Context) -> InkGlassHitRegionView {
+  package var body: some View {
+    InkGlassHitRegionMarker().opacity(0)
+  }
+}
+
+private struct InkGlassHitRegionMarker: NSViewRepresentable {
+  func makeNSView(context: Context) -> InkGlassHitRegionView {
     InkGlassHitRegionView(frame: .zero)
   }
 
-  package func updateNSView(_ view: InkGlassHitRegionView, context: Context) {}
+  func updateNSView(_ view: InkGlassHitRegionView, context: Context) {}
 }
--- a/desktop/macos/Desktop/Tests/ChatCitationTests.swift
+++ b/desktop/macos/Desktop/Tests/ChatCitationTests.swift
@@ -291,6 +291,16 @@ final class ChatCitationTests: XCTestCase {
 
   @MainActor
   func testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID() {
+    // `request` refuses every id while the Rewind capture fence is suspended, and that fence is
+    // process-global: it outlives whichever suite suspended it. That is why this has failed on CI
+    // since it landed while passing locally — the id never arrived, and the failure was reported
+    // against this contract rather than against the suite that left the fence down. Start from a
+    // known-clean fence so the assertions below are about the one-shot handoff and nothing else.
+    RewindCaptureOwnerGeneration.endTransition()
+    XCTAssertNotNil(
+      RewindCaptureOwnerSnapshot.capture(),
+      "no Rewind owner for the citation focus to attach to")
+
     RewindCitationFocusState.shared.request(42)
     XCTAssertEqual(RewindCitationFocusState.shared.consume(), 42)
     XCTAssertNil(RewindCitationFocusState.shared.consume())
--- a/desktop/macos/Desktop/Tests/ChatCitationTests.swift
+++ b/desktop/macos/Desktop/Tests/ChatCitationTests.swift
@@ -291,16 +291,6 @@ final class ChatCitationTests: XCTestCase {
 
   @MainActor
   func testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID() {
-    // `request` refuses every id while the Rewind capture fence is suspended, and that fence is
-    // process-global: it outlives whichever suite suspended it. That is why this has failed on CI
-    // since it landed while passing locally — the id never arrived, and the failure was reported
-    // against this contract rather than against the suite that left the fence down. Start from a
-    // known-clean fence so the assertions below are about the one-shot handoff and nothing else.
-    RewindCaptureOwnerGeneration.endTransition()
-    XCTAssertNotNil(
-      RewindCaptureOwnerSnapshot.capture(),
-      "no Rewind owner for the citation focus to attach to")
-
     RewindCitationFocusState.shared.request(42)
     XCTAssertEqual(RewindCitationFocusState.shared.consume(), 42)
     XCTAssertNil(RewindCitationFocusState.shared.consume())
--- a/desktop/macos/Desktop/Tests/InkGlassHitRegionReporterTests.swift
+++ b/desktop/macos/Desktop/Tests/InkGlassHitRegionReporterTests.swift
@@ -1,0 +1,73 @@
+import AppKit
+import OmiTheme
+import SwiftUI
+import XCTest
+
+@testable import Omi_Computer
+
+/// Regression: `InkGlassHitRegionReporter` is a marker for the pointer, and it has to be *both*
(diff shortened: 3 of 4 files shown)


kodjima33:
Closing this rather than leaving it open: it is verified but not mergeable, and the branch `watchdog/issue-11563-desktop-swift-lane-red` is ready to re-open the moment the last blocker below is resolved.

**What CI proved.** Run 31792619068 on this branch took `Desktop Swift Static & Test Contracts` from **77 failures to 1**. `ShellModalScrimTests` (74), `GlassSurfaceReviewRegressionTests` (1) and `ProactiveLaneClientTests` (2) are all green on CI with these commits — the two fixes are confirmed, not just locally tested.

**Why it cannot merge.** `ChatCitationTests.testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID` (from #11519) still fails, exactly as it has on every CI run of this lane since it landed. It is unrelated to this diff and it keeps the lane red, so nothing here unblocks desktop PRs on its own.

I tried one fix for it and CI disproved the hypothesis, so I reverted it (8e73f6a9fe) rather than leave a wrong explanation in the tree. What is now known:
- `RewindCaptureOwnerSnapshot.capture()` is **non-nil** on CI at that point (my added `XCTAssertNotNil` passed), so the suspended-capture-fence theory is dead.
- The failure is therefore inside `consume()`'s remaining guards: either the owner scope captured by `request` no longer equals the one `consume` reads, or `currentOwner.isCurrent()` is false. Both are process-global owner identity/generation state that another suite mutates; class execution order is not stable across machines, which is the whole of the CI-red/local-green split.
- I could not reproduce it locally, including by forcing an owner-id split between `RewindDatabase.currentUserId` and the persisted auth owner.

Its owner should either make it establish the owner identity it needs or delete it — a test that has never passed in CI is not guarding anything.

🤖 hourly desktop watchdog

Issue #11534, opened 2026-08-14
setup.sh ios builds an app that hangs on the splash screen when the local backend harness isn't running


formed2forge:
`bash setup.sh ios` builds, installs and launches an app that then hangs on the splash screen, with no error and no indication of the cause, whenever the local backend harness is not running.

Reproduced today on an iOS device build from current `main`. `setup.sh ios` (the non-beta branch, `setup.sh:417-420`) is the *local-dev* path: it writes

```
API_BASE_URL=http://127.0.0.1:8000/
USE_WEB_AUTH=true
USE_AUTH_CUSTOM_TOKEN=true
```

into `app/.dev.env` and passes `--dart-define=OMI_FIREBASE_AUTH_EMULATOR_HOST=127.0.0.1`. With nothing listening on port 8000 the app waits on a backend that does not exist. The build itself succeeds, so every signal a contributor has says the setup worked.

Two things make this hard to self-diagnose:

1. **The prerequisite is invisible.** `setup.sh` neither starts the harness nor checks for it, and never mentions it. (Docs for this are proposed separately in #11532.)
2. **`127.0.0.1` is the phone's loopback, not the Mac's.** Even with the harness running, a physical iOS device cannot reach it. `setup.sh` honours `OMI_DEV_HOST` (`setup.sh:53`) for exactly this, but nothing tells a contributor they need it or what to set it to.

Suggested shape, deliberately small: before building, check whether the configured `API_BASE_URL` is reachable, and if not, fail with a message naming the harness and the `OMI_DEV_HOST` override — rather than producing an app that looks fine and hangs. Failing fast is the whole point; the current behaviour spends a full iOS build before stranding the developer.

For context on why the local harness is the only option here: the prod-simulating path (`setup.sh ios beta`, targeting `https://api.omiapi.com/`) is gated behind `FIREBASE_SERVICE_ACCOUNT_KEY` and so is unavailable to community contributors by design. Production Firebase tokens are also rejected against that backend, so pointing a dev build at production yields a sign-in that appears to succeed while every call fails.

Found while verifying the iOS community-build signing work (#7641), which is unrelated to this and does not introduce it — this reproduces on `main` today.

formed2forge:
**Correcting my own framing.** I filed this as "hangs when the local backend harness *isn't* running." That is not the real problem, and I don't want the title to send anyone down the wrong path.

[...]

Issue #11533, opened 2026-08-14
make dev-init fails in a git worktree: dev-init.sh assumes .git is a directory


formed2forge:
`make dev-init` fails when run from a git worktree, because `scripts/dev-harness/dev-init.sh:33` assumes `.git` is a directory:

```bash
hook=".git/hooks/pre-commit"
if [ ! -f "$hook" ]; then
  ln -s -f ../../scripts/pre-commit "$hook"
```

In a linked worktree `.git` is a **file** containing a `gitdir:` pointer, not a directory, so `.git/hooks/` does not exist and the symlink cannot be created. Observed on this machine:

```
main checkout   /Users/tim/DevProjects/omi                 .git is a directory
worktree        /Users/tim/DevProjects/omi-worktrees/…     .git is a FILE
                                                            (gitdir: …/.git/worktrees/…)

in the worktree:  .git/hooks           DOES NOT EXIST
                  git rev-parse --git-path hooks   → /Users/tim/DevProjects/omi/.git/hooks
```

This bites the workflow the repo itself recommends: AGENTS.md says *"Always work in a git worktree for code changes (`git worktree add`)"*, so a contributor following that guidance and then running `make dev-init` hits it immediately. It also lands late in the script — after `pip install -r backend/requirements.txt` has already run — so the failure costs a full dependency install first.

Git provides the worktree-safe lookup:

```bash
hook="$(git rev-parse --git-path hooks)/pre-commit"
```

`git rev-parse --git-path` resolves correctly in both a main checkout and a linked worktree. (`make setup` already advertises "installs repo Git hooks … with linked-worktree-safe paths", so the correct approach exists in the repo — `dev-init.sh` just doesn't use it.)

Worth checking the rest of `scripts/dev-harness/` for the same assumption while fixing; this was the only `.git` reference in `dev-init.sh` itself.

formed2forge:
The downstream consequence is worth recording, because the worktree failure is not where a contributor feels it.

`dev-init.sh` is the **only** thing that installs backend requirements (`scripts/dev-harness/dev-init.sh:30`, `pip install -q -r backend/requirements.txt`), and the hook line that fails in a worktree is at `:33` — *after* that install in the happy path, but a contributor who hits the failure and re-runs, or who runs it in a worktree from the start, can easily end up with a venv that is not fully provisioned.

[...]

PR #11569, opened 2026-08-14
fix(dev-harness): detect prerequisites that are present but unusable

formed2forge:
## Problem

Getting the local dev harness up on a clean Mac took about an hour, and both blockers presented as the *same* misleading failure — a 135-second wait ending in:

```
Health checks failed:
  - firestore: not healthy after 45s at http://127.0.0.1:8085/
  - auth: not healthy after 90s at http://127.0.0.1:9099/
```

Neither emulator was the problem. The message names the wrong components, so the natural next step is to go debugging Firestore.

## Two fixes

**1. `npx firebase-tools` blocked forever on an install prompt.** `_firebase_command()` built `["npx", "firebase-tools"]`. On a machine without firebase-tools, npx asks `Need to install the following packages: firebase-tools@15.27.0 / Ok to proceed? (y)` — but the emulator runs detached with stdout redirected to a log file, so nothing can answer and it blocks indefinitely. The entire `firebase-emulators.log` was that prompt, printed twice, with no other output. Fixed by passing `--yes`.

**2. The java preflight passed on every Mac.** The check was `_which("java")`. macOS ships a stub at `/usr/bin/java` that is always present and exits 1 with "Unable to locate a Java Runtime" when no JDK is installed:

```
$ ls -la /usr/bin/java     -> exists, 135376 bytes
$ java -version            -> "Unable to locate a Java Runtime", exit 1
$ shutil.which("java")     -> /usr/bin/java          <- false pass
```

So the preflight always passed on the platform the harness primarily targets, and the missing JVM surfaced 135 seconds later as the health-check timeout above. Replaced with `_java_runtime_present()`, which runs `java -version` and checks the exit code.

## Verification

```
$ bash scripts/dev-harness/run-tests.sh
85 passed, 9 skipped
```

Four new tests cover both fixes: the macOS stub exiting non-zero is not a runtime, a real runtime is, the missing runtime is reported as a prerequisite, and the npx invocation carries `--yes`.

Exercised the real path, not just the tests:

- **Before:** on a Mac with no JDK, `make dev-up` waited 135s and blamed firestore/auth.
- **After:** it names the real blocker in ~2s — `java runtime (required by the Firestore and Auth emulators; install one with 'brew install --cask temurin' ...)`. Installed Temurin, re-ran, and the harness came up.
- The `--yes` path is the one actually taken on this machine (`which firebase` -> not found), and `firebase-emulators.log` went from containing *only* `Ok to proceed? (y)` to showing a started Emulator Hub on port 4400.

## Notes

`--yes` converts the hang into a silent first-run download of `firebase-tools@15.27.0`. That is a deliberate trade — a slow first run beats an infinite hang — but worth stating, since `prerequisite_report` accepts `_which("npx")` without checking firebase-tools is actually installed.

Related contributor-onboarding issues found on the same path: #11533 (`make dev-init` fails in a git worktree) and #11534.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/dev-harness/dev_harness/cli.py
+++ b/scripts/dev-harness/dev_harness/cli.py
@@ -283,6 +283,27 @@ def _python_importable(module: str) -> bool:
     )
 
 
+def _java_runtime_present() -> bool:
+    """Report whether a usable JVM exists, not merely whether `java` is on PATH.
+
+    macOS ships a stub at /usr/bin/java that is always present and exits 1 with
+    "Unable to locate a Java Runtime" when no JDK is installed, so a PATH lookup
+    passes on every Mac. Running the binary is the only check that distinguishes
+    the stub from a real runtime.
+    """
+    if not _which("java"):
+        return False
+    try:
+        return (
+            subprocess.run(
+                ["java", "-version"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, timeout=30
+            ).returncode
+            == 0
+        )
+    except (OSError, subprocess.SubprocessError):
+        return False
+
+
 def prerequisite_report(cfg: config.HarnessConfig) -> tuple[list[str], list[str]]:
     missing: list[str] = []
     warnings: list[str] = []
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Correctly diagnosed: /usr/bin/java stub always passes a PATH check, and detached npx blocks on its install prompt. Both fixes are minimal and covered by tests.

PR #11571, opened 2026-08-14
fix(desktop): unblock the red Swift test lane — 77 failures across 4 suites (#11563)

kodjima33:
## Bug

`Desktop Swift Static & Test Contracts` has been red on **every** desktop PR since `15a51fb985` (2026-08-14 02:10 EDT), so no desktop change could reach `main` and the release train had no qualification signal. `main` reads green only because the lane is path-filtered: `Detect Desktop Swift Changes` skips the test job on every commit that does not touch `desktop/macos/**`, so the red only appears on desktop PRs. Filed as #11563.

| Suite | Failures |
|---|---|
| `ShellModalScrimTests` | 74 |
| `GlassSurfaceReviewRegressionTests` | 1 |
| `ProactiveLaneClientTests` | 2 |
| `ChatCitationTests` | 1 |

## Root causes

**1. 75 of the 77 — the hit-region marker paints.** `InkGlassHitRegionReporter` mounts a zero-drawing AppKit marker so a transparent window can tell visible content from air. `ImageRenderer` cannot rasterise an `NSViewRepresentable` and substitutes an **opaque placeholder** for it, so every pixel-truth test of a surface that mounts one measured the placeholder covering the marker's whole host: the modal dim read as painting to the window's edge — which on this window is the user's wallpaper — and the glass panel's rounded corner read as filled. Measured directly: `Color.clear.background(InkGlassHitRegionReporter())` rendered at 40×40 came back fully opaque (centre RGBA `255,56,60,255`), where `Color.clear` alone is `0,0,0,0`.

Fix: mount the marker at **zero opacity**. The placeholder contributes no pixels while the AppKit view stays mounted, unhidden and registered — SwiftUI applies opacity to the host layer, not to the view's own `alphaValue`, which is what `InkGlassHitRegions.containsPoint` reads. `InkGlassHitRegionReporterTests` pins **both** halves, because simply dropping the reporter would satisfy the paint half and silently return the shell to passing clicks through its own modals.

**2. The 2 in `ProactiveLaneClientTests`.** `testLaterShorterRetryWindowDoesNotShortenActiveCooldown` (`ff974901b6`) asserted a request count an armed cooldown makes unreachable: the two 429s were issued sequentially, but the cooldown short-circuits the network before the second request is built, so the shorter `Retry-After` was never parsed. The two attempts now overlap on a rendezvous — the only way a second window is ever observed — so the max-deadline preservation the test is named for is actually exercised.

**3. The 1 in `ChatCitationTests` — a fourth failure #11563 did not record.** `testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID` has failed on every CI run of this lane since #11519 landed, and passes locally.

`RewindCaptureOwnerSnapshot` resolves an owner id two different ways: `capture()` prefers the authorization snapshot's owner (from the persistent `auth_userId` default), while `isCurrent()` compares against `RewindDatabase.currentUserId ?? "anonymous"`, which is in-memory and starts nil in a fresh process. `swift-test-suites.sh` runs each suite in its own process, but UserDefaults is on disk and outlives all of them — so on a runner where an earlier suite left `auth_userId` set, the focus request captured owner `"X"` and `consume()` compared it against `"anonymous"` and refused.

Fix: pin the two owner sources together for the duration of the test. The assertion is about the one-shot handoff contract, not owner plumbing.

## What I tested

- **Reproduced the CI failure locally**, which is what previous attempts were missing: setting `auth_userId` in the test process before the unmodified test turns it red with CI's exact message — `XCTAssertEqual failed: ("nil") is not equal to ("Optional(42)")`. With the fix and that default still set: 18/18 green. With it cleared: 18/18 green.
- `xcrun swift test --filter` on all four suites the lane reports, **77 failures → 0**:
  - `ChatCitationTests` 18/18
  - `ShellModalScrimTests` 7/7
  - `GlassSurfaceReviewRegressionTests` 3/3
  - `ProactiveLaneClientTests` 11/11
  - `InkGlassHitRegionReporterTests` 2/2 (new)
- `scripts/pre-push` bounded gate passed.

## Follow-up, deliberately not in this PR

The divergence between the two owner sources in `RewindCaptureOwnerSnapshot` is a real product defect — any snapshot captured before `RewindDatabase.currentUserId` is resolved at launch can never validate, so frames and citation handoffs are silently dropped in that window. Narrowing a fail-closed ownership fence is not a change to make alongside a CI unblock; filed as #11572.

Failure-Class: none

### DIFF
--- a/desktop/macos/Desktop/Sources/Theme/InkGlassHitRegions.swift
+++ b/desktop/macos/Desktop/Sources/Theme/InkGlassHitRegions.swift
@@ -70,12 +70,27 @@ package final class InkGlassHitRegionView: NSView {
 
 /// SwiftUI mount for the marker. Attach as a `.background` so it spans exactly the surface that
 /// should own the pointer.
-package struct InkGlassHitRegionReporter: NSViewRepresentable {
+///
+/// Mounted at zero opacity, which is load-bearing rather than decorative. The marker paints nothing
+/// on screen, but `ImageRenderer` cannot rasterise an `NSViewRepresentable` and substitutes an
+/// opaque placeholder for it — so a marker at full opacity reads as *painted content covering its
+/// whole host* to every pixel-truth test of a surface that mounts one, which is a claim about the
+/// app that is not true. At zero opacity the placeholder contributes no pixels while the AppKit view
+/// stays mounted, unhidden and registered — SwiftUI applies the opacity to the host layer, not to the
+/// view's own `alphaValue`, which is what `InkGlassHitRegions.containsPoint` reads.
+/// `InkGlassHitRegionReporterTests` holds both halves.
+package struct InkGlassHitRegionReporter: View {
   package init() {}
 
-  package func makeNSView(context: Context) -> InkGlassHitRegionView {
+  package var body: some View {
+    InkGlassHitRegionMarker().opacity(0)
+  }
+}
+
+private struct InkGlassHitRegionMarker: NSViewRepresentable {
+  func makeNSView(context: Context) -> InkGlassHitRegionView {
     InkGlassHitRegionView(frame: .zero)
   }
 
-  package func updateNSView(_ view: InkGlassHitRegionView, context: Context) {}
+  func updateNSView(_ view: InkGlassHitRegionView, context: Context) {}
 }
--- a/desktop/macos/Desktop/Tests/ChatCitationTests.swift
+++ b/desktop/macos/Desktop/Tests/ChatCitationTests.swift
@@ -291,6 +291,14 @@ final class ChatCitationTests: XCTestCase {
 
   @MainActor
   func testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID() {
+    // The handoff is owner-scoped, and it reads two owner sources that are ambient here:
+    // `auth_userId` is a persistent UserDefaults value that outlives every suite process on a
+    // CI runner, while `RewindDatabase.currentUserId` starts nil in each one. Pin them to the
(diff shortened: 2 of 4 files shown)


kodjima33:
Closing without merging — CI cannot go green for reasons that have nothing to do with this diff, and an open PR nobody will merge is worse than a closed one with the evidence written down. The branch `watchdog/issue-11563-desktop-swift-lane-red` is kept at 44b39a7a5a, rebased and ready to reopen.

**What this PR proved.** All four assertion suites go to zero, confirmed by CI run [31809160446](https://github.com/BasedHardware/omi/actions/runs/31809160446): `ShellModalScrimTests`, `GlassSurfaceReviewRegressionTests`, `ProactiveLaneClientTests` and `ChatCitationTests` are all absent from its failure list, where every earlier run of this lane had them. 77 assertion failures → 0.

**Why it is still red.** Three suites crash with `signal code 11` — `AgentPillLifecycleTests`, `ChatTimelineContinuityTests`, `MarkdownNestedScrollTests`. They are not mine: they crash identically on [31795097198](https://github.com/BasedHardware/omi/actions/runs/31795097198) (`security/transport-egress`), [31778872737](https://github.com/BasedHardware/omi/actions/runs/31778872737) (#11553) and [31779149249](https://github.com/BasedHardware/omi/actions/runs/31779149249) (#11559), none of which contain a commit from this branch. So #11563 undercounted the lane at 3 suites; it is 7. Filed as #11573, with the evidence that the crash point moves between runs and that all three pass when the built bundle is driven directly on a local machine.

Reopen this the moment #11573 lands — the diff is rebased, the manifest checks pass, and the four suites are proven.

Issue #7206, opened 2026-05-07
Bug Report: Voice Detection Delay Causes Missed Conversation Start


hugoaap-code:
## Bug Report: Voice Detection Delay Causes Missed Conversation Start

**Voice detection battery-saving mode misses the first seconds of conversations after silence**

There seems to be a bug with the new voice detection feature designed to save battery.

When the device has been in silence for a while and a conversation starts, the device does not begin recording immediately. Instead, it only starts recording after a delay, which causes the beginning of the conversation to be missed.

In my testing, I have noticed that up to **10 seconds** of conversation can be missed before recording starts.

### Steps To Reproduce

1. Enable the new voice detection / battery-saving mode.
2. Leave the device in a silent environment for a while.
3. Start a conversation normally.
4. Check the recording/transcription afterward.

### Expected Behavior

The device should detect the beginning of the conversation quickly and start recording immediately, or at least preserve the first few seconds using a buffer.

### Actual Behavior

The device starts recording only after a delay, causing the first phrase or first few seconds of the conversation to be missing.

### Impact

This makes the feature unreliable for real conversations because important context can be lost at the beginning. Missing the first sentence often makes the rest of the conversation harder to understand or less useful.

### Additional Notes

A short pre-roll buffer could help prevent this issue by capturing a few seconds before voice detection fully triggers recording.

aryanorastar:
### VERIFY (2026-08-06) — DEVICE / firmware (no app-only residual)

Checked against current `main` before pickup:

- “Battery-saving voice detection” is **pendant hardware AAD**, not the app/backend Silero toggle (`vadGateEnabled`). After `CONFIG_OMI_VAD_HOLD_MS` (10s) of silence, `enter_hw_aad()` in `omi/firmware/omi/src/mic.c` pauses PDM and puts the T5838 into hardware sleep — **mic off ⇒ zero PCM**, so conversation onset is lost until WAKE + resume (+ SD remount offline).
- Software VAD + ~800ms pre-roll lived in `aad.c` (#6950) and was **removed** by [#9048](https://github.com/BasedHardware/omi/pull/9048) in favor of hardware-only AAD. `CONFIG_OMI_ENABLE_VAD_GATE=y` is still set but does not gate PCM on current main.
- App/backend 300ms streaming pre-roll cannot restore audio that never left the device.
[...]

PR #11578, opened 2026-08-14
fix(firmware): keep CV1 capture reliable across silence and SD faults

axAilotl:
## Summary

- replace the unreliable CV1 hardware-AAD sleep path with continuous T5838 capture at a compliant 640 kHz PDM clock and software VAD
- preserve 500 ms of pre-roll, debounce activity for 300 ms, and hold capture for 10 seconds after voice so quiet audio is gated without losing opening speech
- recover from the observed SD CMD18/CMD12 multi-sector read failure by switching to checked CMD17 single-sector reads for the rest of the boot
- preserve unread ring data on every failed read and keep SD-worker buffers alive until the worker completes

This supersedes the earlier narrow AAD-entry patch on this PR. Physical testing proved that patch fixed the lost-edge race but exposed a real-hardware 66-67 ms sleep/re-wake loop. The qualified fix therefore does not claim hardware-AAD low-power operation: it chooses reliable capture and durable recovery over the unqualified CV1 v0.5 WAKE path.

## Failure boundary

Firmware 3.0.21 could fail in two independent ways:

1. During hardware-AAD entry, masking and later clearing WAKE could strand the microphone asleep. Arming sooner retained the event, but the T5838 entry acknowledgement/transient then woke physical hardware almost immediately, producing a high-churn loop instead of useful low-power dwell.
2. During ring download, the card/controller combination could fail a CMD18/CMD12 multi-sector read. The old path surfaced the failure without a bounded recovery mode, leaving valid unread audio stuck on the device. A related worker-lifetime bug allowed a caller-owned buffer to expire before the SD worker finished using it.

The new boundary keeps the microphone/PDM active, performs silence gating after capture, and treats multi-sector read failure as a one-way boot-local transition to individually verified sectors. Ring cursors advance only after successful delivery; failed reads never consume the unread tail.

## Exact scope relative to firmware 3.0.21

The qualified image is exactly two commits on tag `Omi_CV1_v3.0.21`:

- `3.0.21+56`: continuous capture, 640 kHz PDM, software VAD, bounded pre-roll, and SWD-readable soak counters
- `3.0.21+57`: CMD17 SD-read fallback plus SD-worker buffer-lifetime repair

For review, those same changed files are transplanted onto current `main`; every firmware file changed by the qualified +57 image is byte-identical on this PR head.

No app change is required. This patch does not change BLE services, ring opcodes, record layout, Opus framing, or cursor semantics. The existing app ring-sync implementation for firmware 3.0.20+ was used throughout physical qualification and drained disconnected backlogs successfully.

## Physical qualification

Signed `3.0.21+57` ran on CV1 hardware attached to J-Link serial `69650697`. All accepted observations were identity-locked to the production pendant (`Omi`, BLE `D0:25:80:82:17:70`, model `Omi CV 1`); the briefly selected DevKit was excluded before the Android clock started.

### iOS away/return qualification

- 32h42 from the first 100% indication to the final intact 7% snapshot, with four deliberate phone-away/return cycles.
- The iPhone was away for an estimated 5-5.5 hours (at least 3.5 hours directly bracketed by consecutive disconnected snapshots). Admitted audio accumulated on SD while `is_connected=0`; returned-phone drains ultimately consumed every eligible backlog without firmware drops or read/write errors.
- 36 newly recovered files, 300,869 Opus frames, and 1h40m17.38s of away/return audio validated with zero malformed frames and zero decode errors; all files contained non-silent samples.
- Including the original recovered backlog: 49 files, 411,796 frames, and 2h17m15.92s validated.

### Android continuous-live qualification

- 23h01 unassisted from 100% to the observed 1% charging boundary, only 59 minutes short of the planned 24-hour endpoint.
- Android exercised a materially harder transport workload: one roughly six-second reconnect and otherwise essentially continuous live BLE at a negotiated 15 ms interval, 2M PHY, and MTU 498. This contrasts with the iOS away periods and makes the raw runtimes unsuitable as a controlled platform battery A/B.
- The final intact SWD snapshot at 22h46 / 7% showed ring read/write `408925/408940`, dropped 0, exact software-VAD accounting, zero ring/write-queue/VAD/fallback-sector failures, mounted writable SD, and healthy mic/PDM. The core was resumed.
- All 101 correct-device catalog entries were locally present: 24 SD-recovered and 77 live files totaling 13,697,765 bytes, 190,098 Opus packets, and 3,783 catalog seconds. Every file had exact frame structure/counts, decoded successfully, and contained non-silent PCM.
- The production Android app process remained alive. Server transcription websocket 1011/flapping and upload `miss` states were downstream; neither represented missing or corrupt local audio.

### Cross-platform interpretation

- Firmware VAD-admitted duty was 12.88% on iOS and 13.34% on Android. Android was only 3.6% more speech-active proportionally, but it was connected almost continuously, so continuous BLE—not speech alone—is the leading known workload difference.
- SD recovery remained exactly one multi-sector failure, one fallback attempt, and one recovered batch; CMD17-only mode latched for the boot. Across both qualifications there were zero fallback-sector failures, ring drops, write-queue drops, VAD failures, or SD write errors.
- The Android 15-record tail was durable and below the app minimum-drain threshold. It is an adjacent app-policy behavior, not firmware data loss; no app change is required for this firmware reliability boundary.

The qualified result supports review of the firmware change: continuous capture remained alive through silence, software VAD accounting stayed exact, disconnected audio reached SD, returned-phone recovery produced structurally valid non-silent Opus, and the forced SD-read fallback preserved cursor/data integrity.

The earlier hardware-AAD permanent-sleep failure and 66-67 ms entry-transient loop cannot occur in this build because hardware AAD is disabled.

## Verification

- `cc -std=c11 -Wall -Wextra -Werror -Isrc src/audio_frontend.c src/software_vad.c tests/plus56/test_plus56.c -o /tmp/omi-plus56-host-test && /tmp/omi-plus56-host-test`
- `cc -std=c11 -Wall -Wextra -Werror -Isrc src/sd_read_recovery.c tests/plus57/test_plus57.c -o /tmp/omi-plus57-host-test && /tmp/omi-plus57-host-test`
- clean NCS 2.9.0 sysbuild and signed image generation
- J-Link flash and verify on the physical CV1
- non-resetting SWD snapshots throughout the multi-day soak and away/return test

## Known adjacent behavior

- This is a reliability fix, not a hardware-AAD power fix. Software VAD reduces stored/transmitted silence, but the mic/PDM remain active.
- The app intentionally defers a ring tail estimated below 10 seconds. The observed 64-record tail was durable, survived reboot, and drained once enough audio accumulated; changing that policy belongs in a separate app PR.
- App scheduling can delay automatic backlog drain while connected, but the tested backlog remained durable and subsequently drained without firmware errors.
- Transcription-unavailable flapping is downstream app/backend behavior; direct firmware counters and recovered Opus audio showed capture continuing.
- Recovered-file wall-clock timestamps can be stale until the device RTC is re-established. That does not affect frame integrity and is separate from this firmware repair.

Product invariants: none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11578?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


axAilotl:
Physical validation (2026-08-14, CV1 hardware):

- Flashed the signed `3.0.21+2` MCUboot OTA (`dfu_application.zip` SHA-256 `dda74b00cbb359d86229775d918875784b9d836284d0c4a7c56a4ef4181d872a`) through the iOS MCU DFU path with app-settings erase disabled.
- Transfer advanced from 0% to 100% and the app reported `Firmware flashed successfully`.
- The first app instance stayed in `Disconnected / Waiting for device` for more than 60 seconds after reboot; a cold restart of only the app reconnected the same saved pairing immediately, showing 100% battery and live listening. No pendant reset, unpair, or re-pair was required.
- After a quiet interval, speech resumed through the live path; a subsequent BLE cold reconnect also returned to listening.
- The Device Information UI still displays `3.0.21`; it does not expose the MCUboot `+2` build metadata.

The post-DFU stale iOS BLE-manager instance is existing app behavior and is recorded separately from this firmware wake-boundary fix.

Git-on-my-level:
Verified this independently — the fix is sound and the harness genuinely pins the bug. Detail per file:

- `omi/firmware/omi/src/mic.c` — `enter_hw_aad()` now arms the WAKE IRQ immediately after `t5838_aad_enter()` (before the 800 ms `CONFIG_OMI_AAD_SETTLE_MS` sleep) and drops the old `atomic_clear(&aad_wake_pending)` that ran after the settle window. That clear was what discarded a real edge landing during entry; latching instead of clearing is the right call, and the already-HIGH `gpio_pin_get_dt()` re-check covers the missed-edge case. The new `mic_request_aad_wake()` keeps the wake path under the AAD worker's single ownership, so a racing reconnect can't drive pins concurrently.
- `omi/firmware/omi/src/t5838_aad.c` (not changed, checked for the timing claim) — `t5838_aad_enter()` ends with `clock_bitbang(ENTER_SLEEP_CLOCKING_US / ...)` = 2.5 ms of activation clocking before it returns, so the "arm immediately" premise in the mic.c comment is accurate.
- `omi/firmware/omi/src/lib/core/transport.c` — the `_transport_connected` hook is guarded by `CONFIG_OMI_ENABLE_T5838_AAD`, and `mic_request_aad_wake()` is a no-op when the mic is already active, so the reconnect path can't misfire on non-AAD builds.
- `omi/firmware/omi/tests/aad_wake/` — I ran `bash omi/firmware/omi/tests/aad_wake/run.sh` on Linux GCC 13.3: 3/3 passed (settle / reconnect / gpio). As a mutation control I moved the WAKE arming back after the settle interval (the 3.0.21 ordering): `aad_wake_during_settle` failed as expected while the other two still passed, so the regression test really does cover this fix, not just adjacent behavior. Compiling the production `mic.c` against the host shim is a genuinely behavioral seam rather than a source-order tripwire.
- `.github/checks-manifest.yaml` — the new `firmware-aad-wake-state-machine` check (local + ci lanes) followed the documented registration process, and Hygiene passed on this PR, so the manifest-selected check ran in CI.
- `omi/firmware/omi/Kconfig` + `tests/aad_wake/README.md` — help text and README now describe the armed-during-settle behavior accurately; nothing aspirational.

The two failing Desktop Swift checks are unrelated to this change: the log shows macOS fault-suite infrastructure errors (`desktop-core-harness fault-suite failed`, "Retrying failure detached-app cleanup case after port collision", "TERM case never launched detached app") and this PR touches no Swift or desktop files.

Nice work on the failure-boundary writeup and the mutation-control evidence in the PR description — that's exactly the kind of verification record that makes firmware review tractable.

Leaving for human maintainer review before it moves out of draft — mainly to sanity-check the reconnect one-shot policy (one bounded wake request per BLE connection) as the intended product behavior for #7206.

— glm-5.3, Omi maintainer automation

---

axAilotl:
Physical-validation status update for @Git-on-my-level:

The host harness still validates the narrow bug fixed by this PR, but further testing of the physical `3.0.21+2` build reproduced another failure in the same user-visible path and under the same quiet/AAD → reconnect conditions. The pendant could be blue and fully reconnected—with service discovery, notification-configuration requests, and continuing RSSI traffic—while producing neither live audio nor new SD/WAL audio. A soft power reset completed another BLE reconnect but did not restore audio. After several additional power cycles, high-rate GATT indications eventually returned, but recovery was non-deterministic: observed BLE-ready delays were approximately 20 seconds, 44 seconds, and 5 minutes 54 seconds, with the producer stream appearing later. The durable WAL remained frozen at its earlier boundary during that recovery capture.

So this PR repairs a real AAD-entry race, but it is **not yet the final fix** for the connected-but-not-recording/transcribing failure class or for #7206. The remaining adjacent fault is below the live-preview/backend layer and may involve the physical AAD/PDM lifecycle, actual audio-subscription state, or the firmware's live-vs-durable routing boundary. The current host shim cannot distinguish those states.

We are keeping this PR in draft while building a physical test jig that will leave the CV1 attached to a Segger and expose the board safely enough to trace AAD/PDM state, encoded-frame progress, GATT subscription/notification progress, discard routing, and SD write advancement. The jig needs to be completed before additional firmware flashes and physical validation can continue. No broader firmware workaround should be inferred from the current patch until that hardware trace is available.

Git-on-my-level:
Thanks @axAilotl — this is a well-scoped repair for the persistent no-wake part of #7206, and the production-linked harness is exactly the kind of regression this subsystem needs. Notes from glm-5.3 (automated maintainer review):

**Core fix is sound**

- `omi/firmware/omi/src/mic.c` — `enter_hw_aad()` now arms WAKE immediately after `t5838_aad_enter()` returns. That matches the driver contract: `t5838_aad.c` clock-bangs `ENTER_SLEEP_CLOCKING_US` (2500 µs) internally, so the mic has latched AAD sleep before the call returns. The old ordering kept WAKE masked through the additional settle window and then did `atomic_clear(&aad_wake_pending)`, discarding any edge latched during entry. The already-HIGH check at arm time closes the missed-edge case.
- `omi/firmware/omi/src/lib/core/transport.c` — the one-shot `mic_request_aad_wake()` in `_transport_connected()` (on the success path, after `is_connected = true`) is bounded: `aad_thread_fn` consumes `aad_wake_pending` as a no-op when `aad_in_sleep == 0`, so a reconnect can only resume capture, never force or extend sleep.
- `omi/firmware/omi/src/lib/core/mic.h` — the new API doc ("safe to call when the microphone is already active") matches the implementation.
- `omi/firmware/omi/Kconfig` — help text correctly re-describes `OMI_AAD_SETTLE_MS` as settle-only with WAKE armed throughout; consistent with the code.

**Harness verified**

I built and ran `omi/firmware/omi/tests/aad_wake/run.sh` on Linux GCC 13.3: 3/3 pass. I also re-ran your mutation control — restoring the 3.0.21 arm-after-settle ordering in `enter_hw_aad()` fails exactly `aad_wake_during_settle` (the other two cases still pass), so the regression genuinely pins this boundary. The shim kernel (`host/shim.c`) injecting a GPIO edge from inside `k_sleep(50)` is a clean controllable seam for the production `mic.c`, and the check is registered in `.github/checks-manifest.yaml` with both `local` and `ci` lanes per repo convention.

**Notes**

- The two failing Desktop Swift checks also fail on this PR's base commit — pre-existing, not caused by this firmware-only diff.
- One thing on-device validation should confirm before this leaves draft: the old masking-during-settle existed to swallow the AAD entry transient, and WAKE is now armed through that window. If the entry transient asserts WAKE on real hardware, a pendant could oscillate wake → VAD hold → re-sleep and burn power. The host harness can't observe that (its GPIO is test-driven), so a CV1 soak after AAD entry is the missing evidence. The trade-off direction (prefer a spurious wake over stranded audio) is right for #7206; it just deserves a device-level confirmation.

Marking ready-for-review plus a firmware maintainer's on-device sign-off on the settle-window arming is what's left here; leaving for human maintainer review with that specific ask.

---

axAilotl:
Extended physical validation update (2026-08-15/16, CV1 hardware):

The instrumented signed `3.0.21+3` build contains this narrow AAD repair plus black-box diagnostics. OTA ZIP SHA-256: `f0c620cd237411dca68cb8652f152ae4c2eb09948ea9a15f3fe3db49d941ecae`.

The good result is that the original permanent-sleep failure did not recur. On one uninterrupted application boot:

- the approximately 12-hour checkpoint had 2,814 AAD sleeps and 2,814 wakes, 2,067,980 encoded frames, and 2,066,683 one-for-one accepted/completed BLE notifications;
- five hours later the counters reached 4,345 sleeps and 4,345 wakes, 2,932,420 encoded frames, and 2,931,123 one-for-one accepted/completed notifications;
- BLE link-setup errors, microphone read errors, queue errors, notification errors, storage errors, rejected frames, and dropped frames all remained zero;
- the later checkpoint was still the same boot and connection, at 37% battery / 3,788 mV.

However, this run also confirms the power concern raised above. The retained trace covers the final 4.45 hours of the first checkpoint and contains 225 rate-limited sleep events with 225 matching wakes; every sampled wake follows its sleep by only about 66–67 ms. The unthrottled counters show 2,814 cycles in roughly 12 hours and another 1,531 cycles in 5 h 16 min. With the configured 10-second quiet hold, that cadence is strongly consistent with the newly armed WAKE path immediately undoing nearly every AAD entry rather than the microphone remaining asleep until later acoustic activity.

The 09:15→14:31 interval was consequently capture-active for about 91.2% of wall time and consumed 14 battery-percentage points (about 2.7 points/hour; gauge-derived, not calibrated current measurement). So the patch fixes the catastrophic stranded-asleep behavior, but the present implementation does not yet pass the intended low-power AAD boundary: it appears to trade permanent sleep for near-continuous sleep/wake cycling.

One user-visible no-transcription incident during the soak was not a recurrence of the pendant failure: direct SWD counters and iPhone GATT traffic continued advancing while text was absent, and restarting only the app restored transcription without resetting the pendant.

Conclusion: keep this PR in draft. The production-linked harness remains valid for the lost-edge regression, but the physical acceptance test now also needs to require a real quiet dwell in AAD sleep, bounded wake latency on actual sound, and no entry-transient wake loop. The pendant is still running toward a near-depletion checkpoint; final RAM trace and battery/state evidence will be captured before it is allowed to shut down.

Git-on-my-level:
Thanks @axAilotl — this is a focused repair of the #7206 no-wake boundary with an unusually strong verification story. Independent results from my side:

**Ran the new harness** (`bash omi/firmware/omi/tests/aad_wake/run.sh`) on Linux GCC 13.3: 3/3 passed, matching your report. I also re-ran the mutation control: moving `aad_wake_irq(true)` back to after `k_msleep(CONFIG_OMI_AAD_SETTLE_MS)` in `enter_hw_aad()` makes `aad_wake_during_settle` fail while the reconnect and gpio cases still pass — so the settle regression genuinely pins the arming order this fix introduces.

File notes:
- `src/mic.c` — arming WAKE right after `t5838_aad_enter()` (before the settle sleep) is the correct boundary: the shim's `k_sleep` seam injects an edge from inside the 50 ms settle window, and the latched `aad_wake_pending` is consumed by the worker loop after `enter_hw_aad()` returns. `mic_request_aad_wake()` is atomic + semaphore based and safe to call from the BLE thread; when the mic is already active it degrades to a no-op as documented in `mic.h`.
- Dropping the old `atomic_clear(&aad_wake_pending)` before arming means a stale request set during entry now produces one bounded wake-and-resume instead of being silently discarded — this matches the trade-off in the PR body (worst case is a return to active capture, never a disabled AAD).
- `src/lib/core/transport.c` — the `_transport_connected()` recovery call is guarded by `CONFIG_OMI_ENABLE_T5838_AAD` and runs after `is_connected = true`, so SD power stays up along the reconnect path. No issues found.
- `Kconfig` — the rewritten `OMI_AAD_SETTLE_MS` help text now describes the actual behavior (WAKE armed during settle). Good docs hygiene.
- `.github/checks-manifest.yaml` — `firmware-aad-wake-state-machine` follows the existing schema (both lanes, diff-scoped triggers), and the `ci` lane executed it green on this head via `run_checks.py` in Repo Checks.
- The harness compiles the production `mic.c` against test-owned GPIO/kernel/DMIC seams rather than asserting on source string order — exactly the behavioral coverage style the repo's bug-fix bar asks for.

CI note: the two failing checks (Desktop Swift Build & Tests, Desktop Swift Static & Test Contracts) are in the desktop Swift lane, which this diff does not touch; their logs show a macOS harness port collision / detached-app TERM case and workflow fixture mismatches, and the same lane has intermittent failures on `main`. I don't see anything in this PR that could affect them.

Minor, non-blocking: the shim's settle injection matches `k_sleep` by exact timeout value (`wake_during_delay_ms` = 50), which is distinct from `AAD_PDM_SETTLE_MS` (20) today; if those constants ever collide the seam would need an explicit marker. Fine to leave as-is.

Since this changes on-device power-state behavior for the listening pendant and ships through the manual firmware release flow, keeping `needs-maintainer-review` for a human sign-off on the hardware behavior — from the automation side this looks ready when you are.

---

Git-on-my-level:
Thanks @axAilotl — supersession understood: the narrow AAD-arming patch verified earlier on hardware was overtaken by your finding of the 66-67 ms entry-transient loop, and this rework chooses reliable continuous capture over the unqualified CV1 v0.5 WAKE path. The failure-boundary writeup is unusually precise. Notes per file on the new head (both host regressions re-run independently from this branch: they compile clean under -Wall -Wextra -Werror and pass):

- omi/firmware/omi/src/mic.c — capture moves to 8 kHz stereo with the PDM clock window narrowed to 630-650 kHz (ratio 80 -> 640 kHz, inside the T5838 low-power band), and process_audio_buffer() now requires exactly OUTPUT_FRAMES_PER_BLOCK (1600) frames from the interpolator before forwarding — a good fail-closed check. The AAD sleep machinery is correctly re-gated behind CONFIG_OMI_ENABLE_T5838_HW_AAD_SLEEP while t5838_aad_power() rail ownership stays under OMI_ENABLE_T5838_AAD.
- omi/firmware/omi/src/audio_frontend.c + .h — stateful stereo-to-mono mix plus 2:1 linear interpolation that preserves previous_mono across DMA block boundaries; the capacity guard (output_capacity < input_frames * 2) returns 0 before touching memory, and the int32 mix cannot overflow, so dropping the old clip guards is correct.
- omi/firmware/omi/src/software_vad.c + .h — preroll ring (5 x 1600 samples = 16 KB static), 300 ms debounce, 10 s hold, replayed oldest-first via the modular first index, and fail-open on emit error exactly as the header documents. The exact-1600-sample requirement matches MIC_BUFFER_SAMPLES in lib/core/config.h. The 16 KB preroll in .bss is a deliberate, bounded cost.
- omi/firmware/omi/src/sd_read_recovery.c + .h — the CMD18/CMD12 -> checked CMD17 fallback is boot-local and one-way, never resets the controller or mutates ring metadata, and never trusts the failed multi-block buffer (single-sector reads re-fetch every sector at exact offsets). The magic-tagged metrics struct for SWD snapshots is a nice diagnostic touch.
- omi/firmware/omi/src/sd_card.c — load_batch_for_seq() routes through the fallback; restore_tail_batch() now preserves the tail slot and metadata on a read failure instead of truncating away valid unread audio; and sd_ring_read() no longer returns at the 15 s warning threshold — since the queued request owns the caller's buffer pointer, waiting for the in-flight worker is the correct use-after-scope fix. Worth being aware: a wedged SD worker now blocks the caller indefinitely where it previously errored out at 15 s; with the watchdog in place that is a defensible trade, but it is a behavior change.
- omi/firmware/omi/Kconfig — clean split of rail/driver ownership (OMI_ENABLE_T5838_AAD) from clock-off sleep (OMI_ENABLE_T5838_HW_AAD_SLEEP, default n) with honest 'laboratory-only until physical qualification' help text; OMI_AAD_SETTLE_MS correctly re-parented onto the sleep symbol.
- omi/firmware/omi/omi.conf — 3.0.21+57 with matching CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION and CONFIG_OMI_ENABLE_T5838_HW_AAD_SLEEP=n, consistent with the qualification claims.
- omi/firmware/boards/omi/omi_nrf5340_cpuapp.dts and omi/firmware/omi/CMakeLists.txt — comment-only dts change documenting the 640 kHz choice; build wiring for the three new translation units is straightforward.
- omi/firmware/omi/tests/plus56/ and tests/plus57/ — both READMEs accurately describe what the tests cover (verified by running them); plus57's scripted-IO harness asserting exact buffer pointers and sector offsets for the fallback is exactly the right shape.

Two items before this leaves draft:

1. Formatting CI is red on the new files: clang-format --dry-run --Werror flags src/sd_read_recovery.c:42 and tests/plus57/test_plus57.c:31 (brace placement on the compound-literal casts). clang-format -i on those two files should clear it.
2. The previously-verified immediate-WAKE-arm ordering did not carry into the dormant path. enter_hw_aad() on this head matches current main — atomic_clear(&aad_wake_pending) after the settle sleep, then arm — i.e. the lost-edge fix verified earlier in this PR's history is gone. It is unreachable in this build (CONFIG_OMI_ENABLE_T5838_HW_AAD_SLEEP=n), so it is not a blocker, but if the lab-only path is ever re-qualified the race comes back. Please either carry the arming-order fix into the dormant block or note it in the Kconfig help text so nobody re-enables blind.

One judgment call for maintainers (label already present): this ships continuous PDM capture with software gating, so the mic rail never sleeps. Given the WAKE-path hardware findings that is the right reliability call, but it is a CV1 power-budget/product direction decision that needs maintainer sign-off before merge. Leaving for human maintainer review on that trade-off; the mechanical work here looks sound.

---

Issue #11481, opened 2026-08-13
Shipped default task prompt (13.7k chars) exceeds the 10k assistant-prompt sync bound, so it never syncs


kodjima33:
Commit 38d3b8eb2f ("fix: harden combined desktop QA workflows", merged 2026-08-12) added `APIClientAssistantSettingsTests.testShippedTaskPromptFitsBackendSyncContract` and `testEachAssistantPromptUsesItsBackendBound`, which assert that the task assistant's shipped default prompt fits its sync bound and that task has a bound above the legacy 10,000. Neither is true today, so the desktop Swift suite is red on main (first seen on PR #11479's CI; reproduced at main `8442e89f03`), which also blocks the hourly desktop release train's qualification lane.

Facts:
- `TaskAssistantSettings.defaultAnalysisPrompt` is ~13.7k chars, and has been over 10k since at least 2026-06 (14.3k on Jul 1). It has therefore never synced: `SettingsSyncManager.promptForSync` deliberately omits oversized prompts to keep partial PATCH semantics.
- Backend `ASSISTANT_ANALYSIS_PROMPT_MAX_LENGTH = 10000` (`backend/routers/users.py`, task model), same 10k as insight/memory; the client's `TaskAssistantSettings.maximumSyncedAnalysisPromptLength` is also 10_000.

Raising the bound is NOT client-safe on its own: if the client constant is raised before the backend limit is deployed to prod, every user on the default task prompt starts including a ~13.7k `analysis_prompt` in the settings PATCH, Pydantic 422s, and the entire assistant-settings sync breaks for that build. Sequence must be: raise backend limit → deploy prod backend → then raise the client constant in a desktop release.

PR #11479 restores the two tests to assert the shipped contract (oversized default is omitted; all three assistants share the 10k bound) so main and the release train are unblocked; this issue tracks the intended bound raise with correct sequencing (or, alternatively, trimming the default prompt under 10k).

Sun-HakunaMatata:
Hi @kodjima33 — following the contribution guide's bounty-suggestion process, is there an available paid scope here for the **prompt-reduction** option?

I reviewed #11580 and the merged client-side fix in #11678. I understand that raising the existing API limit needs a contract decision, and that the existing PR author has offered to continue. I am asking only about work that is still unassigned.

If you choose to keep the 10,000-character contract, I propose **US$75** for a focused contribution:

- Shorten the shipped task prompt while preserving the behaviors you identify as required.
[...]

PR #11580, opened 2026-08-14
fix(backend): admit the shipped task-extraction prompt into settings sync (#11481)

RatnamOjha:
## What

`TaskAssistantSettings.analysis_prompt` was bounded at 10,000 code points while the desktop client's shipped default task prompt is **13,120**. That prompt has therefore never synced — and never will on any deployed build — because the mismatch is silent by construction.

This is the backend half of #11481, and backend-only on purpose.

## Why it is silent

An oversized prompt does not 422. `SettingsSyncManager.promptForSync` omits it from the settings PATCH to preserve partial-update semantics, so the field simply never arrives:

```swift
// desktop/macos/Desktop/Sources/ProactiveAssistants/Services/SettingsSyncManager.swift:212
let length = prompt.unicodeScalars.count
guard length <= maximumLength else { ... return nil }
```

For every user still on the default task prompt, the server has held no `analysis_prompt` at all since at least 2026-06.

## Why task gets its own bound

Task is the only shipped default that outgrew 10k. Measured at `66fe4673ce` by extracting each `defaultAnalysisPrompt` literal and counting code points:

| assistant | shipped default | bound |
|---|---:|---:|
| task | 13,120 | **20,000** |
| memory | 6,262 | 10,000 |
| insight (wire key `advice`) | 3,576 | 10,000 |
| focus | ships no default | 10,000 |

This mirrors the client, which already carries a task-specific `TaskAssistantSettings.maximumSyncedAnalysisPromptLength`. The headroom above 13,120 is for user edits, which write the same field.

## Sequencing

Order matters and this PR is deliberately step 1 of 3:

1. **this change** — backend accepts up to 20k
2. prod backend deploy
3. client constant raised in a desktop release

Raising the client constant first would make every user on the default start sending a 13.1k `analysis_prompt` that the deployed backend still 422s, breaking assistant-settings sync wholesale for that build. The desktop tests that pin the client at 10k (`testShippedTaskPromptExceedsBackendBoundAndIsOmittedFromSync`, `testEachAssistantPromptUsesItsBackendBound`) are untouched and stay green.

Raising a `max_length` is backward compatible: every payload valid before is still valid.

## The regression test

The existing backend test was named `..._accepts_shipped_default_size_...` but only compared the bound against itself, so it passed at any value — which is why the defect survived. Each side asserted against its own constant and nothing asserted the contract between them.

`test_backend_bounds_admit_every_shipped_desktop_prompt` now parses each assistant's real `defaultAnalysisPrompt` literal out of the Swift sources and validates it through the real `UpdateAssistantSettingsRequest` model. It is behavioral — real Pydantic validation on the real shipped text — not a source scrape.

## Verification

```
$ cd backend && python -m pytest tests/routers/test_users.py -q
25 passed

$ python -m pytest tests/unit/test_app_client_openapi_compatibility.py tests/routers/test_users.py -q
45 passed
```

Proof the new test catches the reported defect — with the bound reverted to 10000, it fails on task **only**:

```
E  String should have at most 10000 characters [type=string_too_long,
   input_value='You are a task commitmen...l_system / project_tool']
FAILED tests/routers/test_users.py::test_backend_bounds_admit_every_shipped_desktop_prompt[task-...]
1 failed, 2 passed
```

Effective bounds read off `model_fields` after the change, confirming the other three are untouched:

```
FocusAssistantSettings     max_length=10000
TaskAssistantSettings      max_length=20000
AdviceAssistantSettings    max_length=10000
MemoryAssistantSettings    max_length=10000
```

## Note on the OpenAPI spec

`docs/api-reference/app-client-openapi.json` carries the single exported bound that changed. It was hand-patched rather than regenerated: `scripts/export_openapi.py` on this machine also rewrites `format: binary` to `contentMediaType` and adds `ctx`/`input` to `ValidationError`, which is local pydantic/fastapi version drift unrelated to this change. Worth a maintainer regenerating on the pinned toolchain if you'd rather have the generated artifact.

Line-Count-Exception: backend/routers/users.py | 2174 -> 2180 | two named bounds replacing a duplicated literal, plus the contract comment explaining why a bound may not sit under its assistant's shipped default

Refs: #11481
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11580?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/users.py
+++ b/backend/routers/users.py
@@ -1975,12 +1975,24 @@ class FocusAssistantSettings(BaseModel):
     excluded_apps: list[str] | None = None
 
 
+# Analysis prompts are user-editable, and the desktop client ships a default for each
+# assistant. A bound below an assistant's own shipped default silently disables settings
+# sync for every user still on that default: `SettingsSyncManager.promptForSync` omits an
+# oversized prompt from the PATCH rather than truncating it, so the prompt never reaches
+# the server (issue #11481). Each bound must therefore clear the prompt its assistant
+# actually ships; `test_backend_bounds_admit_every_shipped_desktop_prompt` enforces that
+# against the real Swift sources.
 ASSISTANT_ANALYSIS_PROMPT_MAX_LENGTH = 10000
+
+# Task extraction is the one shipped default that outgrew the shared bound: 13,120 code
+# points at 66fe4673ce, and between 13k and 15k since 2026-06. The headroom above it is
+# for user edits, which are the same field.
+TASK_ANALYSIS_PROMPT_MAX_LENGTH = 20000
 
 
 class TaskAssistantSettings(BaseModel):
     enabled: bool | None = None
-    analysis_prompt: str | None = Field(None, max_length=ASSISTANT_ANALYSIS_PROMPT_MAX_LENGTH)
+    analysis_prompt: str | None = Field(None, max_length=TASK_ANALYSIS_PROMPT_MAX_LENGTH)
     extraction_interval: float | None = None
     min_confidence: float | None = Field(None, ge=0.0, le=1.0)
     notifications_enabled: bool | None = None
--- a/backend/routers/users.py
+++ b/backend/routers/users.py
@@ -1967,27 +1967,21 @@ class SharedAssistantSettings(BaseModel):
     screen_analysis_enabled: bool | None = None
 
 
+# Each bound must clear its own assistant's shipped desktop default. An oversized prompt is
+# omitted from the settings PATCH by `SettingsSyncManager.promptForSync` rather than
+# rejected, so a bound under the shipped default unsyncs that prompt silently (#11481);
+# `test_backend_bounds_admit_every_shipped_desktop_prompt` guards this against the Swift source.
+ASSISTANT_ANALYSIS_PROMPT_MAX_LENGTH = 10000
+# Task ships the only default over 10k: 13,120 code points, and 13k-15k since 2026-06.
+TASK_ANALYSIS_PROMPT_MAX_LENGTH = 20000
+
+
 class FocusAssistantSettings(BaseModel):
     enabled: bool | None = None
-    analysis_prompt: str | None = Field(None, max_length=10000)
+    analysis_prompt: str | None = Field(None, max_length=ASSISTANT_ANALYSIS_PROMPT_MAX_LENGTH)
     cooldown_interval: int | None = None
     notifications_enabled: bool | None = None
     excluded_apps: list[str] | None = None
-
-
-# Analysis prompts are user-editable, and the desktop client ships a default for each
-# assistant. A bound below an assistant's own shipped default silently disables settings
-# sync for every user still on that default: `SettingsSyncManager.promptForSync` omits an
-# oversized prompt from the PATCH rather than truncating it, so the prompt never reaches
-# the server (issue #11481). Each bound must therefore clear the prompt its assistant
-# actually ships; `test_backend_bounds_admit_every_shipped_desktop_prompt` enforces that
-# against the real Swift sources.
-ASSISTANT_ANALYSIS_PROMPT_MAX_LENGTH = 10000
-
-# Task extraction is the one shipped default that outgrew the shared bound: 13,120 code
-# points at 66fe4673ce, and between 13k and 15k since 2026-06. The headroom above it is
-# for user edits, which are the same field.
-TASK_ANALYSIS_PROMPT_MAX_LENGTH = 20000
 
 
 class TaskAssistantSettings(BaseModel):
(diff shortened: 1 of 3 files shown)


RatnamOjha:
Marking this draft: **`Public Developer API contract` fails, and I do not think it should be forced green.** The gate is right and it exposes a constraint the issue did not account for.

`check_app_client_openapi_compatibility.py` rejects the change on the response side:

```
BREAKING paths./v1/users/assistant-settings.get.responses.200...properties.task.anyOf[0]
BREAKING paths./v1/users/assistant-settings.patch.responses.200...properties.task.anyOf[0]
Version the endpoint instead of breaking a released app-client contract.
```

That is `_compare_bound` doing exactly what it should. For `direction == 'response'` an upper bound breaks when `new_value > old_value`: released clients were told `analysis_prompt` is at most 10,000 characters, and widening it means the server can now return something they were promised it would not.

The reason it cannot be worked around by splitting request and response models:

- `AssistantSettingsResponse` extends `UpdateAssistantSettingsRequest`, so both surfaces share one `TaskAssistantSettings` schema. Splitting them so only the request widens leaves the response validating at 10,000 — and since the endpoint declares `response_model=AssistantSettingsResponse`, a stored 13.1k prompt then fails response validation and **500s the GET**, which is worse than today's silent non-sync.
- Dropping `maxLength` from the response instead also trips the same rule (`new_value is None` breaks a released response bound).

So there is no non-breaking way to let the server return more than it promised. That leaves the two paths the gate itself names, and both are maintainer calls rather than something I should pick:

1. **Trim `TaskAssistantSettings.defaultAnalysisPrompt` under 10,000** — the alternative #11481 already mentions. No contract change at all. It is 13,120 code points today, so it needs ~3.2k trimmed, and that is a prompt-behavior change I should not make unilaterally.
2. **Version the endpoint** (`/v2/users/assistant-settings`) and widen the bound there, leaving v1 on 10,000.

Everything else in the PR stands and is independent of which path you choose:

- `test_backend_bounds_admit_every_shipped_desktop_prompt` is the missing regression test. The old `..._accepts_shipped_default_size_...` compared the bound against itself and passed at any value, which is why this went unnoticed since 2026-06. The new one parses each assistant's real `defaultAnalysisPrompt` literal out of the Swift sources and validates it through the real request model.
- Measurements at `66fe4673ce`: task **13,120** code points, memory 6,262, insight 3,576, focus ships no default. Task is the only one over.

Happy to redo this as either path — say which and I will push it. Leaving it draft rather than green-by-force.

Git-on-my-level:
Verified this independently before it waits on a maintainer call — the analysis holds up end to end.

**Measurements reproduced exactly.** I re-extracted each `defaultAnalysisPrompt` literal from the head tree and measured Unicode code points: task **13,120** (`TaskExtraction/TaskAssistantSettings.swift`), memory **6,262** (`MemoryExtraction/MemoryAssistantSettings.swift`), insight **3,576** (`Insight/InsightAssistantSettings.swift`), and confirmed focus ships no default (no `Assistants/Focus` directory exists). The silent-failure mechanics also check out: `SettingsSyncManager.promptForSync` returns `nil` for anything over the bound, and the client pins task at `10_000` (`TaskAssistantSettings.swift:119`) — so the shipped default has indeed never synced.

**The structural blocker is real.** `AssistantSettingsResponse(UpdateAssistantSettingsRequest)` in `backend/routers/users.py` means request and response share one `TaskAssistantSettings` schema, so widening only the request side would leave response validation at 10k and 500 the GET for a stored 13.1k prompt. Splitting the models does not avoid the contract break.

**The failing contract gate is correct, not a false positive.** Widening a released response bound means the server can now return something released clients were promised it would not — that is a genuine breaking change under any sane compatibility rule, and the hand-edited `docs/api-reference/app-client-openapi.json` line (single `maxLength` occurrence in the shared `TaskAssistantSettings` component) is generator-consistent, so the failure is purely the contract decision, not a spec-editing mistake.

**The new test is the real regression guard.** `test_backend_bounds_admit_every_shipped_desktop_prompt` parsing the actual Swift literals closes the gap where the old bound test compared the bound against itself and passed at any value.

What is needed now is a product/release decision between the two options already laid out: (1) trim `TaskAssistantSettings.defaultAnalysisPrompt` under 10,000, or (2) version the endpoint (`/v2/users/assistant-settings`) and widen there. Neither is something automation should pick — (1) changes assistant prompt behavior and (2) adds a public API surface — so flagging for human maintainer review on that choice. Thanks for the unusually rigorous draft note; the sequencing rationale (backend bound → deploy → client constant) is exactly right.

---

aryanorastar (changes requested):
Verified on af61b0cf43fe (draft).

Raising the task `analysis_prompt` bound so the shipped TaskAssistant default actually syncs is the right production hole. Do not do it by bumping the unversioned OpenAPI `maxLength` 10000 → 20000 on GET/PATCH `analysis_prompt`. Public Developer API contract is red on this SHA for that break.

Please keep the unversioned contract at 10k or version the schema; admit the Swift default without a silent unversioned maxLength change. I am not merging this draft.

PR #11581, opened 2026-08-14
feat(desktop): index recent work by URL and file handles

Git-on-my-level:

- INV-CHAT-1

## Summary
- Persist durable source handles (browser URL, file path, app+window fallback) on context visits at the switch edge.
- Prefer those handles for context-bucket identity so the same Google Doc in Chrome and Arc is one workstream.
- `get_work_context` now returns visit handles and workstream briefs first; the screenshot timeline stays as fallback.

## Test Plan
- [x] `xcrun swift test -c debug --package-path Desktop --filter 'WorkHistoryHandleTests|WorkHistoryIndexTests|ContextBucketVisitResolverTests|ContextVisitCoordinatorTests'` (19 passed)
- [ ] Dogfood: switch between two browsers on the same URL twice and ask chat where that doc is.

## Verification
Focused Swift suites above passed after a real compile. Live AX URL capture was not exercised in a named bundle in this change.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11581?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Chat/ScreenContextTelemetry.swift
+++ b/desktop/macos/Desktop/Sources/Chat/ScreenContextTelemetry.swift
@@ -663,13 +663,15 @@ enum ScreenContextWorkContextBuilder {
         "Prefer visits[].handles to identify the document, URL, or file the user means, then open or read that source. The screenshot timeline is fallback evidence only. This is recent historical activity, not proof of the current visible screen.",
     ]
     if let queue = await RewindDatabase.shared.getDatabaseQueue() {
+      let excluded = RewindSettings.shared.excludedApps
       let visits =
         (try? await queue.read { db in
-          try WorkHistoryIndex.fetchRecentVisits(in: db, from: start, to: now, limit: 20)
+          try WorkHistoryIndex.fetchRecentVisits(
+            in: db, from: start, to: now, limit: 20, excludedApps: excluded)
         }) ?? []
       let briefs =
         (try? await queue.read { db in
-          try WorkHistoryIndex.fetchBriefs(in: db, limit: 5)
+          try WorkHistoryIndex.fetchBriefs(in: db, limit: 5, excludedApps: excluded)
         }) ?? []
       payload["visits"] = visits.map { $0.jsonObject(clock: clock) }
       payload["briefs"] = briefs.map { $0.jsonObject() }
--- a/desktop/macos/Desktop/Sources/Chat/ScreenContextTelemetry.swift
+++ b/desktop/macos/Desktop/Sources/Chat/ScreenContextTelemetry.swift
@@ -660,8 +660,20 @@ enum ScreenContextWorkContextBuilder {
       "timeline": timeline,
       "memories_hint": "For the user's operating principles/preferences, also call search_memories (omi-memory).",
       "guidance":
-        "This is recent historical on-screen activity, not proof of the current visible screen. Use it for history; for a direct current-screen question, obtain a live capture instead of answering from this payload.",
+        "Prefer visits[].handles to identify the document, URL, or file the user means, then open or read that source. The screenshot timeline is fallback evidence only. This is recent historical activity, not proof of the current visible screen.",
     ]
+    if let queue = await RewindDatabase.shared.getDatabaseQueue() {
+      let visits =
+        (try? await queue.read { db in
+          try WorkHistoryIndex.fetchRecentVisits(in: db, from: start, to: now, limit: 20)
+        }) ?? []
+      let briefs =
+        (try? await queue.read { db in
+          try WorkHistoryIndex.fetchBriefs(in: db, limit: 5)
+        }) ?? []
+      payload["visits"] = visits.map { $0.jsonObject(clock: clock) }
+      payload["briefs"] = briefs.map { $0.jsonObject() }
+    }
     if let failureCode {
       payload["failure_code"] = failureCode.rawValue
     }
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
@@ -178,21 +178,6 @@ enum ContextBucketVisitResolver {
       arguments: [
         lookupHash, resolvedBucketID, subjectKind, subjectID, confidence, source, startedAt, startedAt,
       ])
-    if handleIdentity != nil, referenceHash != lookupHash, !referenceHash.hasPrefix("ephemeral:") {
-      try db.execute(
-        sql: """
-          INSERT INTO subject_bindings
-            (referenceHash, bucketID, subjectKind, subjectID, confidence, source,
-             occurrenceCount, createdAt, updatedAt)
-          VALUES (?, ?, ?, ?, 0.6, 'title_cooccurrence', 1, ?, ?)
-          ON CONFLICT(referenceHash) DO UPDATE SET
-            bucketID = excluded.bucketID,
-            updatedAt = excluded.updatedAt
-          """,
-        arguments: [
-          referenceHash, resolvedBucketID, subjectKind, subjectID, startedAt, startedAt,
-        ])
-    }
     return resolvedBucketID
   }
 }
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
@@ -83,17 +83,20 @@ enum ContextBucketVisitResolver {
     in db: Database,
     referenceHash: String,
     normalizedTitle: String?,
-    startedAt: Date
+    startedAt: Date,
+    primaryHandle: WorkHistoryHandle? = nil
   ) throws -> String? {
+    let handleIdentity = primaryHandle?.isDurable == true ? primaryHandle : nil
+    let lookupHash = handleIdentity?.identityKey ?? referenceHash
     let binding = try Row.fetchOne(
       db,
       sql: "SELECT * FROM subject_bindings WHERE referenceHash = ?",
-      arguments: [referenceHash])
-    let subjectKind: String = binding?["subjectKind"] ?? "context"
-    let subjectID: String = binding?["subjectID"] ?? referenceHash
+      arguments: [lookupHash])
+    let subjectKind: String = binding?["subjectKind"] ?? handleIdentity?.kind.rawValue ?? "context"
+    let subjectID: String = binding?["subjectID"] ?? handleIdentity?.value ?? referenceHash
     let workstreamID: String? = binding?["workstreamID"]
-    let confidence: Double = binding?["confidence"] ?? 0.5
-    let source: String = binding?["source"] ?? "repeat_cooccurrence"
+    let confidence: Double = binding?["confidence"] ?? (handleIdentity == nil ? 0.5 : 0.8)
+    let source: String = binding?["source"] ?? (handleIdentity == nil ? "repeat_cooccurrence" : "source_handle")
 
     if let existing: String = binding?["bucketID"] {
       let consistent =
@@ -111,17 +114,30 @@ enum ContextBucketVisitResolver {
       // subject. Fall through and re-resolve for the binding's current subject.
     }
 
-    guard normalizedTitle != nil else { return nil }
+    guard normalizedTitle != nil || handleIdentity != nil else { return nil }
     // A subject binding (including explicit_open) already owns this identity, so
     // do not require a prior completed visit before attaching the subject bucket.
     if binding == nil {
       let recentCutoff = startedAt.addingTimeInterval(-7 * 24 * 60 * 60)
-      let previousVisits =
-        try Int.fetchOne(
-          db,
-          sql:
-            "SELECT COUNT(*) FROM context_visits WHERE referenceHash = ? AND outcome = 'completed' AND startedAt >= ?",
-          arguments: [referenceHash, recentCutoff]) ?? 0
+      let previousVisits: Int
+      if let handleIdentity, try db.columns(in: "context_visits").map(\.name).contains("primaryHandleValue") {
+        previousVisits =
+          try Int.fetchOne(
+            db,
+            sql: """
+              SELECT COUNT(*) FROM context_visits
+              WHERE primaryHandleType = ? AND primaryHandleValue = ?
+                AND outcome = 'completed' AND startedAt >= ?
+              """,
+            arguments: [handleIdentity.kind.rawValue, handleIdentity.value, recentCutoff]) ?? 0
+      } else {
+        previousVisits =
+          try Int.fetchOne(
+            db,
+            sql:
+              "SELECT COUNT(*) FROM context_visits WHERE referenceHash = ? AND outcome = 'completed' AND startedAt >= ?",
+            arguments: [referenceHash, recentCutoff]) ?? 0
+      }
       guard previousVisits >= 1 else { return nil }
     }
 
@@ -141,9 +157,12 @@ enum ContextBucketVisitResolver {
           INSERT INTO context_buckets
             (id, subjectKind, subjectID, workstreamID, displayLabel, visitCount,
              lastVisitedAt, createdAt, updatedAt)
-          VALUES (?, ?, ?, ?, NULL, 0, ?, ?, ?)
+          VALUES (?, ?, ?, ?, ?, 0, ?, ?, ?)
           """,
-        arguments: [resolvedBucketID, subjectKind, subjectID, workstreamID, startedAt, startedAt, startedAt])
+        arguments: [
+          resolvedBucketID, subjectKind, subjectID, workstreamID,
+          handleIdentity?.value ?? normalizedTitle, startedAt, startedAt, startedAt,
+        ])
     }
     try db.execute(
       sql: """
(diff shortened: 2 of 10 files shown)


desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandle.swift
```diff
@@ -0,0 +1,94 @@
...
+    components.host = host
+    if let port = url.port, !((scheme == "http" && port == 80) || (scheme == "https" && port == 443)) {
+      components.port = port
+    }
+    var path = url.path
+    if path.count > 1, path.hasSuffix("/") {
+      path.removeLast()
+    }
+    components.percentEncodedPath = path.isEmpty ? "/" : path
+    if let query = url.query, !query.isEmpty {
```
Git-on-my-level: ⚠️ **Warning:** Keeping the raw query string is both an identity and a privacy leak.

Chrome vs Arc on the same Google Doc often differ only by `usp` / `tab` / `authuser` / `sid`. Those will hash to different `url::…` keys, so the core “same URL in two browsers is one workstream” case will split. Query values also land in `primaryHandleValue`, `handlesJson`, `context_buckets.subjectID`/`displayLabel`, and then in `get_work_context` for the model.

Drop userinfo/fragment as you already do, then strip known tracking/auth params (or drop query by default for http(s) document hosts). Add a test that `…/document/d/abc?usp=sharing` and `…/document/d/abc` share one primary handle.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandleExtractor.swift
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  static func liveSnapshot(appName: String, windowTitle: String?) -> WorkHistoryFrontmostSnapshot {
+    var snapshot = WorkHistoryFrontmostSnapshot(
+      appName: appName,
+      windowTitle: windowTitle,
+      bundleID: NSWorkspace.shared.frontmostApplication?.bundleIdentifier
+    )
+    guard AXIsProcessTrusted() else { return snapshot }
+    guard let pid = NSWorkspace.shared.frontmostApplication?.processIdentifier else { return snapshot }
```
Git-on-my-level: ⚠️ **Warning:** This snapshots whichever process is frontmost *now*, not the arriving visit’s app.

`ContextVisitCoordinator.transition` calls this after `waitForTransitionTurn` + `ensureReconciled`. If the user has already moved on, `handles(from:)` still uses the *arriving* `appName` for `isBrowser`, so a Chrome visit can inherit Slack’s first `AXURL` (or the reverse). That corrupts durable identity.

Resolve the PID from the arriving app (or require `frontmost.localizedName`/`bundleID` to match) and fall back to title-only handles when it does not.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandleExtractor.swift
```diff
@@ -0,0 +1,128 @@
...
+    "Chromium",
+    "Vivaldi",
+    "Orion",
+    "Dia",
+    "Comet",
+  ]
+
+  static func handles(from snapshot: WorkHistoryFrontmostSnapshot) -> [WorkHistoryHandle] {
+    var result: [WorkHistoryHandle] = []
+    if isBrowser(snapshot.appName), let url = snapshot.browserURL ?? httpURL(from: snapshot.documentURL),
```
Git-on-my-level: ⚠️ **Warning:** `browserURL` (first DFS `AXURL`) is preferred over `kAXDocument`.

Safari/Chrome commonly put the *page* URL on `AXDocument`. The child walk can return a bookmark, tab favicon target, or in-page link first (`prefix(8)` / `remaining: 24` is not address-bar-aware). For browsers, take http(s) `documentURL` first, then a conservative `AXURL`.

Also: `isBrowser` is an English localized-name set even though `bundleID` is now persisted — localized “Chrome” / Arc / Brave variants will silently degrade to `app_window`.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
```diff
@@ -157,8 +176,23 @@ enum ContextBucketVisitResolver {
           updatedAt = excluded.updatedAt
         """,
       arguments: [
-        referenceHash, resolvedBucketID, subjectKind, subjectID, confidence, source, startedAt, startedAt,
+        lookupHash, resolvedBucketID, subjectKind, subjectID, confidence, source, startedAt, startedAt,
       ])
+    if handleIdentity != nil, referenceHash != lookupHash, !referenceHash.hasPrefix("ephemeral:") {
```
Git-on-my-level: ⚠️ **Warning:** This title-hash side-binding is extra scope and an identity collision.

Handle lookup already uses `url::` / `file::`. Writing `subject_bindings` for the app+title `referenceHash` means two different URLs that share a normalized title (`Inbox`, `Untitled document`, `Dashboard`) will `ON CONFLICT` flip that title onto whichever URL was bucketed last. A later AX miss then attaches the wrong workstream.

Prefer leaving title identity alone unless you can prove the title is unique to this handle. If you keep it, do not overwrite an existing binding that points at a different `subjectID`.

desktop/macos/Desktop/Sources/Chat/ScreenContextTelemetry.swift
```diff
@@ -660,8 +660,20 @@ enum ScreenContextWorkContextBuilder {
...
+    if let queue = await RewindDatabase.shared.getDatabaseQueue() {
+      let visits =
+        (try? await queue.read { db in
+          try WorkHistoryIndex.fetchRecentVisits(in: db, from: start, to: now, limit: 20)
+        }) ?? []
+      let briefs =
+        (try? await queue.read { db in
+          try WorkHistoryIndex.fetchBriefs(in: db, limit: 5)
+        }) ?? []
+      payload["visits"] = visits.map { $0.jsonObject(clock: clock) }
```
Git-on-my-level, line 676: 💡 **Suggestion:** `visits` / `briefs` ship raw `handle.value` (full URLs and `/Users/…` paths) to the model, and `fetchRecentVisits` / `fetchBriefs` do not apply the excluded-app predicate.

Visit-row purge is not enough: `fetchBriefs` rebuilds a URL handle from `context_buckets.subjectKind/subjectID`, so a purged Chrome workstream can still appear. Filter by remaining non-excluded visits, and consider redacting query strings even if you persist them for identity.

Git-on-my-level (commented):
## Code Review Summary

**Verdict: Reviewed 💬** (0 blockers, 4 warnings, 4 follow-ups)

**PR:** #11581 — feat(desktop): index recent work by URL and file handles
**Author:** @Git-on-my-level
**Files changed:** 10 (+761 / -26)

Reviewed the implemented diff only. Scope stays inside the four product constraints: persist handles at the switch edge, prefer them for bucket identity (first visit unbucketed), surface visits+briefs from `get_work_context`, no consent/personality/automation expansion.

### ⚠️ Warnings
- **WorkHistoryHandle.swift:72** — Canonical URLs keep the full query string. Chrome `?usp=sharing` and Arc's bare `/document/d/abc` will not share a workstream, which breaks the stated Chrome/Arc identity goal. Query strings also persist tokens, emails, and session IDs into SQLite and then into the model via `get_work_context`.
- **WorkHistoryHandleExtractor.swift:50** — `liveSnapshot` always AX-walks `NSWorkspace.frontmostApplication`, after `waitForTransitionTurn`. The arriving visit is keyed by `appName`, so a queued/stale transition can stamp Slack's first `AXURL` onto a Chrome visit (or the reverse). Resolve AX against the arriving app's PID, or skip durable handles when frontmost ≠ arriving app.
- **WorkHistoryHandleExtractor.swift:23** — For browsers, the first DFS `AXURL` wins over `kAXDocument`. Chromium trees often expose bookmark/toolbar/link URLs before the page URL; Safari's page URL is usually on `AXDocument`. Prefer http(s) `AXDocument` for browsers, then address-bar/`AXURL`.
- **ContextBucketStore.swift:181** — `title_cooccurrence` binds the app+title `referenceHash` onto the handle bucket. Shared titles (`Inbox`, `Dashboard`, two Google Docs both titled `Untitled document`) will flip or merge title-only visits onto the wrong URL bucket when AX later fails.

### 💡 Suggestions
- **WorkHistoryHandleExtractor.swift:6** — Browser detection is an English localized-name allowlist. `bundleID` is now stored and would be the stable `isBrowser` signal (`com.google.Chrome`, `company.thebrowser.Browser`, etc.).
- **WorkHistoryHandleExtractor.swift:107** — `firstURL` has no AX timeout, runs on the coordinator actor (not main), and the remaining-budget decrement does not account for nodes consumed by the recursive call. A slow Chrome tree will stall the switch-edge visit write.
- **ScreenContextTelemetry.swift:665 / WorkHistoryIndex.swift:78** — `visits`/`briefs` emit raw handle values with no excluded-app filter. Visit-row purge still leaves `context_buckets.subjectID` as the URL, so `fetchBriefs` can resurface a purged browser workstream.
- **WorkHistoryIndex.recapSection** is unused. `get_daily_recap` is unchanged. The PR body claims daily recap wiring that is not in this diff.

### Tests / missing follow-ups
- Happy-path tests are good: canonicalize userinfo/fragment, first visit unbucketed, second URL visit creates a `url` subject, Chrome/Arc primary-handle equality when the URL already matches.
- Missing: same document with different tracking queries stays one bucket; `app_window` is not durable identity; title-binding collision; `get_work_context` payload includes `visits`/`briefs`; excluded-app/purge does not leak stored URLs.
- E2E yaml only added `covers:` paths — no dogfood step for “same URL in two browsers, ask chat.” Live AX capture is still unexercised, as the PR notes.

### ✅ Looks Good
- Durable vs fallback split is correct (`url`/`file` only for identity; `app_window` is extra evidence).
- Userinfo and fragments are dropped; `file:` rejected as a URL handle.
- Bucket creation still requires one prior **completed** visit unless a subject binding already exists.
- `get_work_context` guidance correctly demotes the screenshot timeline to fallback.
- Migration is additive/nullable; existing visit-insert tests keep compiling.

Do not merge on this review. I did not implement fixes — these need a product call on query stripping before a one-line change is safe.

---
*Reviewed by Hermes Agent*

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandle.swift
Git-on-my-level: Addressed in 34c6accade: tracking/auth params (`usp`, `utm_*`, `sid`, `authuser`, …) are stripped so Chrome/Arc share one URL. Follow-up cae7cebe04 also drops credential query names (`access_token`, `code`, names containing token/secret/session) from both identity and outbound `jsonObject()`.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandleExtractor.swift
Git-on-my-level: Addressed in 34c6accade (refuse AX unless frontmost still matches arriving app) and cae7cebe04 (snapshot handles *before* `waitForTransitionTurn` / persistence awaits on both `transition` and `rearmAfterSystemResume`).

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandleExtractor.swift
Git-on-my-level: Addressed in 34c6accade: browsers prefer http(s) `AXDocument` over the first child `AXURL`, and `isBrowser` now also matches known bundle IDs.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
Git-on-my-level: Addressed in 34c6accade: removed the `title_cooccurrence` overwrite. Title-only identity is no longer flipped onto a URL/file bucket.

desktop/macos/Desktop/Sources/Chat/ScreenContextTelemetry.swift
Git-on-my-level: Addressed in 34c6accade: `fetchRecentVisits` / `fetchBriefs` take `excludedApps` and `get_work_context` applies `RewindSettings.shared.excludedApps`. Briefs whose remaining visits are all excluded are omitted.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextVisitCoordinator.swift
```diff
@@ -198,6 +210,19 @@ actor ContextVisitCoordinator {
...
+  private static func resolvedVisitIdentity(
+    appName: String,
+    windowTitle: String?,
+    handles: [WorkHistoryHandle],
+    bundleID: String?
+  ) -> (handles: [WorkHistoryHandle], bundleID: String?) {
+    if !handles.isEmpty {
+      return (handles, bundleID)
+    }
+    let snapshot = WorkHistoryHandleExtractor.liveSnapshot(appName: appName, windowTitle: windowTitle)
```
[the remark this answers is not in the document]
Git-on-my-level: Fixed in cae7cebe04: `resolvedVisitIdentity` now runs at the start of `transition` / `rearmAfterSystemResume`, before any awaits. The frontmost-match guard from 34c6accade remains as defense in depth.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandle.swift
```diff
@@ -0,0 +1,104 @@
...
+struct WorkHistoryHandle: Equatable, Hashable, Codable, Sendable {
+  var kind: WorkHistoryHandleKind
+  var value: String
+
+  var identityKey: String { "\(kind.rawValue)::\(value)" }
+
+  var isDurable: Bool { kind == .url || kind == .file }
+
+  func jsonObject() -> [String: String] {
+    ["kind": kind.rawValue, "value": value]
```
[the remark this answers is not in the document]
Git-on-my-level: Fixed in cae7cebe04: credential-bearing query names are dropped from the persisted canonical URL, and `jsonObject()` re-canonicalizes so outbound `get_work_context` handles cannot leak leftover tokens from older stored values.

desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandle.swift
```diff
@@ -0,0 +1,104 @@
...
+    guard let url = URL(string: trimmed), let scheme = url.scheme?.lowercased() else { return nil }
+    guard scheme == "http" || scheme == "https" else { return nil }
+    guard let host = url.host?.lowercased(), !host.isEmpty else { return nil }
+    var components = URLComponents()
+    components.scheme = scheme
+    components.host = host
+    if let port = url.port, !((scheme == "http" && port == 80) || (scheme == "https" && port == 443)) {
+      components.port = port
+    }
+    var path = url.path
```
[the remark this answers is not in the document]
Git-on-my-level: Fixed in cae7cebe04: canonicalize now copies `percentEncodedPath` / `percentEncodedQuery` instead of decoded `url.path` / `queryItems`. Added `testPreservesPercentEncodedPathIdentity`.

desktop/macos/e2e/flows/context-buckets-dogfood.yaml
```diff
@@ -24,6 +24,9 @@ covers:
   - desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextSubjectBindingService.swift
   - desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextTitleNormalizer.swift
   - desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextVisitCoordinator.swift
+  - desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandle.swift
+  - desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryHandleExtractor.swift
+  - desktop/macos/Desktop/Sources/ProactiveAssistants/Core/WorkHistoryIndex.swift
```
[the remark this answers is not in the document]
Git-on-my-level: Fixed in cae7cebe04: added dogfood step S6 that asks chat where the S1 document is and requires a visit/brief handle (URL or file), not only a Rewind screenshot.

Git-on-my-level:
Review comments are addressed on `cae7cebe04` (this push) plus the earlier harden commit `34c6accade`.

**Original review warnings**
- Tracking query params stripped so Chrome/Arc share one URL
- AX refused unless frontmost still matches the arriving app
- Browsers prefer `AXDocument` over the first child `AXURL`
- Removed `title_cooccurrence` overwrite
- `get_work_context` hides excluded-app visits/briefs

**Cubic follow-ups**
- Handle snapshot now happens *before* visit persistence awaits
- Credential query params dropped from identity and outbound JSON
- Percent-encoded path/query preserved
- Dogfood S6 now asks chat for the document by handle

Left out on purpose: daily recap wiring (frozen `ChatToolExecutor` ratchet) and live AX dogfood in a named bundle. Not merging from here.

PR #11582, opened 2026-08-14
fix(desktop): new beta installs run context buckets without waiting on targeting

Git-on-my-level:
Follow-up to #11577. That change made the beta/stable split exact by keying it on `AppBuild.isBetaProductionBundle`, which fixed *stable* leakage. This one fixes the other half: which beta users actually get the feature.

## Problem

After #11577, a beta client still asks PostHog whether `context_buckets` is on. That flag's audience can only be described two ways, and both fail for a **newly installed** beta client:

- **The person property `update_channel`.** Registered as a super property at init, so it rides events — but the person record is only written by `identify`, which returns early when nobody is signed in. Across 14 days of macOS clients, 215 of ~260 beta installs had a null person-side channel and 11 reported `stable`. It resolved for roughly a sixth of the channel.
- **A static cohort.** Describes only the users who existed when it was built. A new install is not in it.

Behavioral cohorts would solve this, since they recalculate — but PostHog refuses them in flag targeting:

```
Cohort 'macOS beta channel (auto-updating)' has an event-based condition on
'App Launched' (performed_event) and cannot be used in feature flags.
```

So there is no server-side expression of "is running the beta build" that stays correct as users arrive. A new beta user matches nothing and silently gets no notifications — the same class of silent-off failure #11577 was opened to remove, one layer up.

## Change

Bundle identity already answers "is this the beta build" exactly, offline, before any network call. So beta enablement no longer consults `context_buckets` at all.

What is still worth having remotely is the ability to **stop** the pipeline, so beta reads a dedicated kill switch instead:

```swift
guard AppBuild.isBetaProductionBundle else { return false }
return !PostHogManager.shared.isFeatureEnabled(killSwitchFlagName)
```

`context_buckets_kill` is inverted deliberately. `isFeatureEnabled` returns false when PostHog is unavailable or uninitialized, so absent, unresolved, and false all mean "run the pipeline"; only an explicit true stops it. An unreachable PostHog cannot switch the dogfood channel off by accident.

The flag exists and is currently **off**, so this ships enabled: https://us.posthog.com/project/302298/feature_flags/821229

## Blast radius

Stable is gated out one line above and never reaches this branch, so the fail-open default cannot touch it. The audience is the beta channel — roughly 337 macOS clients.

`context_buckets` is left in place and still governs clients built before #11577, which continue to read it.

## Trade-off worth naming

Fail-open is the right default for a dogfood channel that is currently failing *closed* by accident, but it does mean a PostHog outage leaves notifications running rather than pausing them. Given stable is untouchable here and the kill switch is one toggle, that seems the correct side to err on. Worth a second opinion.

## Testing

Not covered by a unit test: `isEnabled` reads `AppBuild` globals and `PostHogManager.shared`, neither of which has an injection seam today. Verified behaviorally — a bundle carrying #11577 and no environment overrides reports `context_buckets_enabled: true` and accumulated 75 visits, 9 buckets, 143 facts and 20 deliveries.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.pr-body.md
+++ b/.pr-body.md
@@ -1,36 +1,48 @@
+Follow-up to #11577. That change made the beta/stable split exact by keying it on `AppBuild.isBetaProductionBundle`, which fixed *stable* leakage. This one fixes the other half: which beta users actually get the feature.
+
 ## Problem
 
-`ContextBucketsFeature.isEnabled` gated non-production bundles on `OMI_FORCE_CONTEXT_BUCKETS=1`, read from `ProcessInfo.processInfo.environment` at runtime. That only survives a launch that goes through `run.sh`.
+After #11577, a beta client still asks PostHog whether `context_buckets` is on. That flag's audience can only be described two ways, and both fail for a **newly installed** beta client:
 
-Any other launch path drops it — opening the bundle from Finder, or macOS restoring the app after the user grants Screen Recording. `isEnabled` then falls through to the PostHog `context_buckets` flag, which does not resolve true for dev bundles. The result is the worst possible shape: screen capture keeps running normally (`is_monitoring: true`, `capture_gate: flowing`) while the bucket pipeline is off and records nothing.
+- **The person property `update_channel`.** Registered as a super property at init, so it rides events — but the person record is only written by `identify`, which returns early when nobody is signed in. Across 14 days of macOS clients, 215 of ~260 beta installs had a null person-side channel and 11 reported `stable`. It resolved for roughly a sixth of the channel.
+- **A static cohort.** Describes only the users who existed when it was built. A new install is not in it.
 
-Observed today on a QA bundle during dogfood, immediately after granting Screen Recording:
+Behavioral cohorts would solve this, since they recalculate — but PostHog refuses them in flag targeting:
 
 ```
-has_screen_recording_permission: true
-is_monitoring: true
-capture_gate: flowing
-context_buckets_enabled: false   <-- silently off
+Cohort 'macOS beta channel (auto-updating)' has an event-based condition on
+'App Launched' (performed_event) and cannot be used in feature flags.
 ```
 
-Zero `context_visits` rows accumulated. This is indistinguishable from "the feature never fires," and cost real debugging time before the environment was inspected with `ps eww`.
+So there is no server-side expression of "is running the beta build" that stays correct as users arrive. A new beta user matches nothing and silently gets no notifications — the same class of silent-off failure #11577 was opened to remove, one layer up.
 
 ## Change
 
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
## Heads-up for whoever reviews this: Desktop Swift CI is disabled repo-wide

This PR shows `mergeStateStatus: CLEAN`. That is **not** evidence the desktop tests passed — they did not run, and currently cannot.

```
$ gh api repos/BasedHardware/omi/actions/workflows
308485337   disabled_manually   Desktop Swift CI   .github/workflows/desktop-swift-ci.yml
```

The workflow was manually disabled around 17:10–17:15 UTC on 2026-08-14. Consequences worth knowing before reading any desktop PR's check list today:

- No `Desktop Swift Static & Test Contracts` or `Desktop Swift Release Compile` check runs are produced at all, so they are absent rather than failing. A short check list looks identical to a green one.
- Re-triggering does not help. A force-push (`synchronize`) and a close/reopen (`reopened`) both produced no run; the workflow has no `workflow_dispatch` trigger.
- **#11577 merged to `main` with zero Desktop Swift check runs.** Its merge commit `0aac049fd5` has none. That is the change now gating context buckets for the whole beta channel.

I have not re-enabled it. It was switched off deliberately and recently, plausibly to unblock the beta train, and re-enabling is a repo-wide change that belongs to whoever turned it off.

## Local test evidence instead

Ran the affected suites on this branch, rebased onto current `main`:

```
cd desktop/macos && ./scripts/dev-feedback.py --once swift \
  'CandidateSinkDeliveryGateTests|CandidateSinkGraduationTests|ContextProactivityDirectorFailureTests|ContextProactivityEngineTests|ContextBucketPromptAssemblerTests|ContextBucketRecentDeliveryTests'
```

```
Executed 46 tests, with 0 failures (0 unexpected) in 0.381 seconds
Iteration 1: PASS in 49.27s
```

Notably included: `testSnapshotExcludesExpiredValidatedFactsBeforeGarbageCollection`, `testSnapshotWorthinessIgnoresExpiredValidatedFacts`, `testValidatedFactIDsFailClosedWhenDirectorCitesNothing`, and `testQuotaCooldownIsADistinctProvenanceClassFromHttp429`.

This is one machine, one toolchain — it is weaker than the CI matrix and is not a substitute for the release-compile lane, which is the one that caught the `RewindPage` type-check regression earlier this week. Treat it as evidence the logic is sound, not as a green build.

Issue #11204, opened 2026-08-07
GRDB records with 'mutating didInsert' + PersistableRecord silently never capture their rowid (10 types)


Archit-lal:
A record type that conforms to GRDB's `PersistableRecord` and declares `mutating func didInsert(_:)` to capture its rowid **never runs that method**, so `id` is `nil` after every insert — silently, because the field is optional by design.

`PersistableRecord`'s `didInsert` requirement is **non-mutating** and ships an empty default implementation (`PersistableRecord+Insert.swift:10`). A `mutating` method cannot witness a non-mutating requirement, so the declaration compiles, is never called, and the default no-op runs instead. The fix is to conform to `MutablePersistableRecord`, where the mutating `didInsert` is the witness. `PersistableRecord` refines `MutablePersistableRecord`, so call sites are unaffected.

**Confirmed instance, already fixed on the desktop glass-UI branch:** `Screenshot` in `desktop/macos/Desktop/Sources/Rewind/Core/RewindModels.swift`. The impact there was not cosmetic. Everything keyed to a freshly captured frame is written `if let id = inserted.id`, so it was all dead code — `RewindIndexer` never queued a single captured frame for embedding (3 call sites). Semantic search still *looked* complete only because `OCREmbeddingService.backfillIfNeeded()` derives its own ids from `SELECT id … WHERE embedding IS NULL`; once that backfill marks itself complete, nothing newly captured is ever embedded again. It surfaced only because an unrelated test tripped over an empty result.

**Same pattern, not yet migrated** (`grep -l "mutating func didInsert"` cross-referenced with `PersistableRecord` conformance, verified at `desktop/macos/Desktop/Sources/`):

```
FileIndexing/IndexedFileRecord.swift
FileIndexing/KnowledgeGraphRecord.swift
LiveNotes/LiveNoteModels.swift
Rewind/Core/ProactiveModels.swift
Rewind/Core/ActionItemModels.swift
Rewind/Core/MemoryModels.swift
Rewind/Core/GoalRecord.swift
Rewind/Core/TranscriptionModels.swift
Rewind/Core/ObservationRecord.swift
ProactiveAssistants/Services/AIUserProfileService.swift
```

Each returns `id == nil` after insert by the same mechanism. Whether that is harmful depends per type on whether the id is consumed downstream, which is why this is filed rather than migrated wholesale — it wants per-type evidence, not a blanket sweep.

[...]

PR #11586, opened 2026-08-14
fix(desktop): put every GRDB insert on the rowid-keeping idiom, and guard it (#11204)

RatnamOjha:
## What

#11204 asked for the class-level guard after two rowid-drop instances were found and fixed one at a time. This lands it, plus the one remaining call site in the hazardous shape.

## The mechanism

A record that captures its rowid in a `mutating func didInsert` never runs that method when inserted through `PersistableRecord`. Measured against the pinned GRDB in `Desktop/Package.resolved`:

- `PersistableRecord+Insert.swift:34` — `extension PersistableRecord { func insert(...) }` is **non-mutating**, and its `didInsert` resolves to the empty default at `:10`. A mutating `didInsert` cannot witness a non-mutating requirement, so it compiles, never runs, and the rowid is dropped.
- `MutablePersistableRecord+Insert.swift:79` — `inserted(_:)` does `var result = self; try result.insert(db)` inside `extension MutablePersistableRecord`, so it binds the **mutating** insert and the record's own `didInsert` is the witness.

The field is optional by design, so the failure is silent: no error, no warning, no crash — the value is just always nil and every dependent path quietly does nothing.

## The call sites

| site | type | status before |
|---|---|---|
| `ActionItemStorage.swift:699` | `ActionItemRecord` (`PersistableRecord`) | **the trap shape** — rowid dropped; nothing reads it, so a dead store today, not a live defect |
| `AIUserProfileService.swift:78` | `AIUserProfileRecord` | safe since #11216 |
| `RewindAbandonedVideoChunkRecovery.swift:119` | `Screenshot` | safe since #11208 |
| `RewindDatabase.swift:2781` | `Screenshot` | safe since #11208 |

All four now use `inserted(...)`, which is what the other 19 persistence seams already use. `insertProfile` also drops a redundant `var inserted = record` copy.

## Why this guard shape

The precise hazard is *"a direct `record.insert(db)` on a concrete type whose `didInsert` is mutating"*, which needs receiver-type resolution a source scrape cannot do soundly. Flagging the declaration shape instead (`PersistableRecord` + `mutating didInsert`) would flag 14 types that are **not** affected, because they all insert via `inserted(db)` — exactly the blanket sweep the issue set out to avoid.

So the checker enforces the single idiom: the direct form buys nothing here, every seam now uses `inserted(db)`, and any new direct `.insert(db)` is rejected regardless of what the receiver conforms to. No allowlist, no type resolution.

It keys on the **database-handle argument**, not `.insert(` alone. Set and Array inserts in `Desktop/Sources` pass ids, keys and elements (`.insert(task.id)`, `.insert(task, at: 0)`, `.insert(key)`), never a bare `db`/`database`, so there is no collision — covered by `test_does_not_flag_set_and_array_inserts`.

**Not a shared primitive** because the hazard is specific to GRDB's split between its mutating and non-mutating insert overloads; there is nothing for another component to reuse. It scans production sources only — tests may construct the broken shape deliberately to prove it is broken.

This is a **static checker**, not behavioral coverage, and is labelled as such in its own docstring. Its behavioral coverage is `test_check_grdb_insert_idiom.py`, which runs the real `main()` against disposable source trees.

## Evidence it would have caught both real instances

Run against the parent commit of each merged fix:

```
parent of 18d4974339 (#11208, Screenshot)    -> exit=1, 5 sites, including
  Rewind/Core/RewindDatabase.swift:2754: try record.insert(db)
parent of 634f093d4e (#11216, AIUserProfile) -> exit=1, 5 sites, including
  ProactiveAssistants/Services/AIUserProfileService.swift:162: try mutableRecord.insert(database)
```

Both are the exact lines those PRs fixed.

## Verification

```
$ xcrun swift build -c debug --package-path Desktop
Build complete! (20.41s)

$ xcrun swift test --package-path Desktop --filter "RowIDCapture"
Executed 2 tests, with 0 failures

$ xcrun swift test --package-path Desktop --filter "ActionItemDeletionSyncTests|ActionItemStorageVisibilityReconciliationTests|StagedTaskSyncIntegrityTests|LocalMutationAuthorizationTests"
Executed 11 tests, with 0 failures

$ python3 desktop/macos/tests/test_check_grdb_insert_idiom.py
Ran 7 tests - OK

$ python3 desktop/macos/scripts/check-grdb-insert-idiom.py
ok: desktop GRDB inserts all use the inserted(db) idiom   (exit 0)

$ git stash && python3 desktop/macos/scripts/check-grdb-insert-idiom.py
FAIL - 4 sites (exit 1)

$ python3 .github/scripts/test_run_checks.py
Ran 45 tests - OK
```

The two `RowIDCapture` suites are the behavioral proof for the idiom conversion: both assert a non-nil rowid back from the real production seams (`insertProfile`, `insertScreenshot`) and both stay green across the change.

## Invariants

**INV-AUTH-1** — matched on path only. `RewindDatabase.swift` is in this invariant's globs, but the change there is the screenshot-rebuild loop's insert idiom; session invalidation policy, 401 handling, owner-transition pool teardown and mutation-lease behavior are all untouched. The `ActionItemStorage` change sits inside the existing `authorization.require()` block and does not alter lease acquisition, commit, or fencing. `LocalMutationAuthorizationTests` (5 tests) passes unchanged.

Refs: #11204
Failure-Class: FC-nonmutating-witness-drops-rowid

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -224,6 +224,16 @@ checks:
     lanes: ["local", "ci"]
     platforms: ["macos"]
     reason: "desktop Swift sources changed; compare against the pinned macOS formatter before requiring a flow cover"
+  - id: desktop-grdb-insert-idiom-tests
+    command: ["python3", "desktop/macos/tests/test_check_grdb_insert_idiom.py"]
+    triggers: ["desktop/macos/scripts/check-grdb-insert-idiom.py", "desktop/macos/tests/test_check_grdb_insert_idiom.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "the insert-idiom checker must keep flagging the two merged rowid-drop instances (#11208, #11216) without flagging the Set and Array insert shapes that are live in Desktop/Sources"
+  - id: desktop-grdb-insert-idiom
+    command: ["python3", "desktop/macos/scripts/check-grdb-insert-idiom.py"]
+    triggers: ["desktop/macos/Desktop/Sources/**/*.swift", "desktop/macos/scripts/check-grdb-insert-idiom.py", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "#11204: a mutating didInsert never witnesses PersistableRecord's non-mutating insert, so a direct record.insert(db) drops the rowid silently onto an optional field; two shipped instances went undetected until unrelated work tripped over them"
   - id: brand-ui-ratchet-tests
     command: ["python3", ".github/scripts/test_check_brand_ui.py"]
     triggers: [".github/scripts/check_brand_ui.py", ".github/scripts/test_check_brand_ui.py", ".github/checks-manifest.yaml"]
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Services/AIUserProfileService.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Services/AIUserProfileService.swift
@@ -74,9 +74,7 @@ actor AIUserProfileService {
   func insertProfile(_ record: AIUserProfileRecord) async throws -> AIUserProfileRecord {
     let db = try await ensureDB()
     return try await db.write { database in
-      var inserted = record
-      try inserted.insert(database)
-      return inserted
+      try record.inserted(database)
     }
   }
 
--- a/desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Core/ActionItemStorage.swift
@@ -696,7 +696,7 @@ actor ActionItemStorage {
             if !newRecord.completed && !newRecord.deleted {
               visibilityChanged = true
             }
-            try newRecord.insert(database)
+            _ = try newRecord.inserted(database)
           }
         }
         try authorization.require()
--- a/desktop/macos/Desktop/Sources/Rewind/Core/RewindAbandonedVideoChunkRecovery.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Core/RewindAbandonedVideoChunkRecovery.swift
@@ -116,8 +116,7 @@ extension RewindDatabase {
       // `imagePath` is NOT NULL in SQLite, whereas video-backed screenshots
       // carry nil in the model.
       if record.imagePath == nil { record.imagePath = "" }
-      try record.insert(db)
-      return record
+      return try record.inserted(db)
     }
   }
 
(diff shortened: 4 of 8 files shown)


kodjima33 (approved):
Putting every GRDB insert on the rowid-keeping idiom with a guard is the right shape for #11204; holding the merge on the red Desktop Swift build.

aryanorastar (changes requested):
Requesting changes (community review — own lane is red):

Putting every desktop GRDB write on `inserted(db)` and guarding with `check-grdb-insert-idiom.py` (+ unit fixtures) is exactly the right class-level fix for #11204. The PersistableRecord non-mutating `insert` / dropped-`didInsert` explanation in the PR body is clear and correct.

**Blocker:** required **Desktop Swift Static & Test Contracts** is failing on this tip (`TEST_OUTCOME: failure`; suites include `ShellModalScrimDismissTests`, `AgentPillLifecycleTests`, `ChatTimelineContinuityTests`, …). Release compile is green, so this is test-lane red, not “cannot compile.”

A size-gate APPROVE does not replace a green required desktop suite on a desktop Swift PR. Please either:
1. Fix any failure caused by this tip, or
2. If these are known flakes unrelated to the GRDB idiom change, re-run until green and note that in the PR — do not merge on a red required lane.

Happy to re-review once Desktop Swift Static & Test Contracts is green on the tip.

RatnamOjha:
Addressed all four Cubic review comments and pushed the fixes.

**Checker**
- Fixed false positives from Swift comments/string literals by masking those regions before matching.
- Added coverage for comments, strings, multiline/raw strings, and collection inserts.
- Strengthened positive tests to assert the exact expected violations rather than substring matches.
- Historical detection for #11208 and #11216 was re-verified unchanged.

**Changelog**
- Updated the unreleased changelog entry to describe the user-visible impact.

**Desktop Swift Static & Test Contracts**
- Investigated the previous red lane and confirmed the failures were deterministic and pre-existing at this branch's merge base, rather than caused by this PR.
- Upstream had already fixed those suites in subsequent commits, so this branch is now rebased onto current `main`.
- The required lane is now **green on the current tip**.

**Verification**
- GRDB checker tests: 14/14 passed
- Full check suite: 46/46 passed
- `RowIDCapture`: 2/2 passed
- Targeted Desktop Swift suites: 13/13 passed
- Desktop Swift Static & Test Contracts: **green**
- `pr-preflight`: 96/96 checks passed

The production GRDB changes themselves are unchanged from the reviewed version.

@aryanorastar re-requesting your review now that the required lane is green.

PR #11592, opened 2026-08-15
feat(desktop): group browser tabs of one site into a shared context bucket

Git-on-my-level:
## Problem

Bucket identity is `sha256(appName::normalized window title)` — `subjectID` defaults to the reference hash, so **one distinct title is always exactly one bucket**. Measured on a real beta profile over ~38h:

- 742 visits → **83 buckets**, 83 distinct normalized keys → 83 buckets, **1:1, zero collapsing ever occurred**
- 50 of the 83 are Google Chrome, from only 135 entries
- X alone is 10 buckets (`(4) home / x`, `(2) notifications / x`, `home / x`, individual posts…)
- 4 separate buckets for different PRs in one repository
- 26 buckets (31%) hold exactly one entry and never accumulate
- **74 of 83 buckets never produced a single notification**

The cost is not notification noise — the dwell gate already keeps this debris out of the reasoning path. It is **missed joins**: nothing can connect four views of one repo, or a repo and the channel where it is discussed.

`workstreamID` is *not* the fix. It sits inside `UNIQUE(subjectKind, subjectID, COALESCE(workstreamID,''))`, so it can only ever partition — two different titles have different `subjectID`s and stay two buckets whatever `workstreamID` holds. The only merge lever in the existing design is `subjectID` convergence via `subject_bindings`, which today only ever writes self-referential rows (verified: 83 rows, all `subjectID == referenceHash`, all `workstreamID` NULL).

## Change

**Leading unread-badge strip.** The existing count regexes are `$`-anchored, so `(4) Home / X` and `Home / X` hash apart. Now stripped for browsers and messaging apps. The trailing form stays messaging-only — `Issue (123)` outside messaging is genuine identity, and `ContextTitleNormalizerTests` still guards that. Also fixes a shipped pattern that matched `– 4 new messages` but never Slack's web title `- 1 new item`.

**Browser destination keys.** Browser contexts get a key like `x.com/feed` or `github.com/acme/repo`, proposed by the extraction call that already runs and cached on the reference hash in `subject_bindings`. Several hashes then resolve to one bucket. The model is consulted **once per novel title and never again**, so identity stays stable and replayable and nothing is added to the hot path.

Riding the existing extraction call is deliberate: the backend's `ProactiveOperation` enum is closed, so a new lane would need a provisioned gateway lane plus its own quota. Extraction already runs at exactly the moment a bucket exists and content is known.

## Why this shape

Ten configurations over 134 real contexts, scored against a label rule fixed before any model output was inspected:

| approach | precision | recall |
| --- | --- | --- |
| route against candidate buckets, all apps | 38% | 18% |
| + richer candidate context (accumulated facts) | 40% | 19% |
| canonical key, all apps | **6%** | 35% |
| canonical key, browser-scoped | 61% | 33% |
| + deterministic site hint | 76% | 34% |
| route against candidates, browser-scoped | 53% | 19% |
| canonical key + hint, reasoning-lane model | 85% | 54% |
| + grounding guard | **89%** | 54% |

The shipped configuration is the extraction lane's own model (`gpt-5-nano`, `reasoning_effort: minimal` → this call reads `low`), which lands at **74% precision / 31% recall** after the guard was tightened in review. The reasoning-lane model reaches 89%/52% on the identical prompt; that is one backend routing line if dogfooding shows the error rate is too high.

Three results drove the design:

1. **Emitting a key beats matching candidates** (76% vs 53%), and is order-independent — the outcome cannot depend on which buckets happened to be recent, which is the failure mode a candidate-set design carries.
2. **Browser scoping is load-bearing.** Unscoped, the model answers `telegram` for every thread and collapses 29 distinct private conversations into one bucket: 6% precision, the worst available outcome. 26 of the 34 available merges are inside Chrome anyway, and native-app title identity is already correct.
3. **Identity signal beat context.** A mechanically-extracted trailing site token was worth +15 points; feeding richer accumulated facts was worth +2.

Self-consistency was tested and rejected — 2-of-3 agreement moved precision 80%→80%, and unanimity cost 16 points of recall. The errors are systematic, not sampling noise.

## Safety

Over-merging is strictly worse than under-merging: a wrong merge pollutes an accumulated fact set that the director later cites in a real notification. Every gate fails toward the existing per-title identity.

- Unrecognized, abstained (`unknown/…`), browser-named, or ungrounded keys are all rejected → context keeps its own bucket
- **Grounding guard**: the claimed domain must be evidenced by the title, rejecting hallucinations like `scalingforever.com/inbox` for a title reading `…ever.com`. Free, deterministic, worth +4 points
- `explicit_open` bindings are never overwritten; an applied destination is never re-adjudicated
- Entries already written are never re-parented — retroactive fact mixing is the one operation that could poison an existing bucket
- Routing rules sit **below** `ScreenDerivedContent.untrustedPreamble`: window titles are attacker-controlled, since any page sets `document.title`
- Native apps are untouched, so no code path can merge two DM threads
- Forward-only. No migration, no history rewritten
- Separate kill switch (`context_buckets_destination_kill`, fail-open) so this can be stopped without taking the pipeline down

**Known residual risk:** at 74% precision on the shipped lane, roughly one in four merges is wrong. The blast radius is bounded to browser buckets, and `DELETE FROM subject_bindings WHERE source LIKE 'derived_destination:%'` re-fragments and lazily re-derives.

## Review findings, and what changed

An adversarial review (cursor, grok-4.6) found four real defects, all fixed here:

- **The grounding guard was exploitable.** Domain labels were substring-matched with no length floor, so `x.com` reduced to `"x"` and grounded against any title containing that letter — `fix: thing #11 · acme/omi` accepted `x.com/feed`, permanently parking a GitHub tab in the X bucket. Short labels now must match a whole title token; only labels of 4+ characters may substring-match. Cost: 2 points of recall on the measured set.
- **Forbidden domains were matched against the whole string**, so `chrome.com/tabs` passed while `chrome/tabs` was rejected. Now checked per label.
- **Web messengers were reachable through the browser path.** A Chrome title like `general | acme | Slack` grounded `slack.com/acme`, merging distinct private threads — precisely the failure browser-scoping exists to prevent. Messenger hosts and messenger-looking titles are now excluded outright.
- **`isBrowser` used substring matching**, so `Ledger Live` (contains "edge") and `Archive Utility` (contains "arc") were treated as browsers. Now an exact application-name set.

Two further issues it raised, also fixed: the join path never touched the destination bucket's `lastVisitedAt`/`visitCount`, so both stale GC and overflow GC would prefer deleting the destination over the orphan buckets feeding it — worst for exactly the visit-once titles this feature targets. And `promptFragment` interpolated the raw title into the rule text, so a newline in an attacker-controlled title could forge rule structure while still sitting below the preamble.

## Verification

126 proactive tests pass, 23 new. Beyond the happy paths they pin: short-label grounding rejection, per-label forbidden domains, messenger exclusion, browser-substring false positives (`Ledger Live`, `Archive Utility`), newline titles not forging prompt structure, decoding a response with no `destination` field, a zero-row claim leaving the binding untouched, and the join path keeping the destination bucket fresh.

One bug was caught by these tests during development: a minimum-length filter on domain parts made `x.com` reduce to nothing after the generic TLD was dropped, silently making the single most fragmented site ineligible to merge. The fix for that is what the review then showed had gone too far the other way.

Not covered: the measurement is one dev-centric profile. The 34 available merges are almost entirely developer tools, so generalization to non-dev usage is unverified.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.pr-body.md
+++ b/.pr-body.md
@@ -1,29 +1,83 @@
 ## Problem
 
-`plan-desktop-release.py` gates every desktop release on this workflow's checks existing for the **exact** source SHA:
+Bucket identity is `sha256(appName::normalized window title)` — `subjectID` defaults to the reference hash, so **one distinct title is always exactly one bucket**. Measured on a real beta profile over ~38h:
 
-```python
-REQUIRED_SOURCE_CHECK_NAMES = (
-    RELEASE_ELIGIBILITY_CHECK_NAME,
-    "Desktop Swift Build & Tests",
-    "Desktop Swift Release Compile",
-)
-```
+- 742 visits → **83 buckets**, 83 distinct normalized keys → 83 buckets, **1:1, zero collapsing ever occurred**
+- 50 of the 83 are Google Chrome, from only 135 entries
+- X alone is 10 buckets (`(4) home / x`, `(2) notifications / x`, `home / x`, individual posts…)
+- 4 separate buckets for different PRs in one repository
+- 26 buckets (31%) hold exactly one entry and never accumulate
+- **74 of 83 buckets never produced a single notification**
 
-Until today this workflow had no manual trigger, only `push` and `pull_request`. That makes the gate unrecoverable whenever a run does not happen: push events cannot be replayed, and for a commit already on `main` neither a force-push nor a PR close/reopen produces one. The SHA stays permanently unreleasable, and the only remedy is to land an unrelated commit purely to move `main` forward.
+The cost is not notification noise — the dwell gate already keeps this debris out of the reasoning path. It is **missed joins**: nothing can connect four views of one repo, or a repo and the channel where it is discussed.
 
-That is not hypothetical. This workflow was disabled manually for part of today. During that window several changes merged to `main` with no checks, so the planner could not resolve a releasable SHA. It polls every 30s for up to 5400s and then fails — two release runs burned the full window and ended `cancelled` with `tag-release` skipped. Re-enabling the workflow did not fix it, because enabling does not retroactively create runs for commits that already merged.
-
-Meanwhile the release train looked healthy: builds were dispatched, PRs showed `mergeStateStatus: CLEAN`, and no check reported failure. A required check that is *absent* is materially different from one that fails, and only the second is visible.
+`workstreamID` is *not* the fix. It sits inside `UNIQUE(subjectKind, subjectID, COALESCE(workstreamID,''))`, so it can only ever partition — two different titles have different `subjectID`s and stay two buckets whatever `workstreamID` holds. The only merge lever in the existing design is `subjectID` convergence via `subject_bindings`, which today only ever writes self-referential rows (verified: 83 rows, all `subjectID == referenceHash`, all `workstreamID` NULL).
 
 ## Change
 
-Adds `workflow_dispatch` so the evidence can be re-minted for any SHA without inventing a commit.
+**Leading unread-badge strip.** The existing count regexes are `$`-anchored, so `(4) Home / X` and `Home / X` hash apart. Now stripped for browsers and messaging apps. The trailing form stays messaging-only — `Issue (123)` outside messaging is genuine identity, and `ContextTitleNormalizerTests` still guards that. Also fixes a shipped pattern that matched `– 4 new messages` but never Slack's web title `- 1 new item`.
 
-No job logic changes. The `changes` job already tolerates the dispatch event: `github.event.pull_request.base.sha` is empty, so `DIFF_BASE` falls through to the existing `git rev-parse HEAD^` fallback, and `github.event.action` is empty so the `!= 'closed'` guard passes.
+**Browser destination keys.** Browser contexts get a key like `x.com/feed` or `github.com/acme/repo`, proposed by the extraction call that already runs and cached on the reference hash in `subject_bindings`. Several hashes then resolve to one bucket. The model is consulted **once per novel title and never again**, so identity stays stable and replayable and nothing is added to the hot path.
+
+Riding the existing extraction call is deliberate: the backend's `ProactiveOperation` enum is closed, so a new lane would need a provisioned gateway lane plus its own quota. Extraction already runs at exactly the moment a bucket exists and content is known.
+
+## Why this shape
+
+Ten configurations over 134 real contexts, scored against a label rule fixed before any model output was inspected:
+
+| approach | precision | recall |
+| --- | --- | --- |
+| route against candidate buckets, all apps | 38% | 18% |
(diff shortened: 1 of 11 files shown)


Git-on-my-level:
## Review comments addressed in `21fb810488`

Five of six were valid and are fixed. All four structural ones measured **free** — precision and recall identical on the same 134 contexts before and after (nano 74%/31%, reasoning lane 89%/52%).

- **Bare domain accepted** — `github.com` with no section satisfied every other guard and would have collapsed every repository into one bucket. The prompt always asks for `<domain>/<section>`; that contract is now enforced.
- **Forbidden labels checked only in first position** — `google-chrome/tabs` hides the browser name in its *second* label. Good catch that my own test read as covering this but only passed because the title `"anything"` grounded nothing; a title like `"chrome new tab"` would have produced `dest:google-chrome/tabs`. All labels are now checked.
- **Generic-only domain** — `com/feed` carried no site identity yet could ground via its section. Rejected outright.
- **`legacy_explicit_open` unprotected** — correct. The one-time UserDefaults import records a user's own prior binding under that source, and the guard only covered `explicit_open`.
- **Raw `Window:` prompt line** — true, and it predates this change, but routing turns extraction output into an *identity* decision, which raises what the vector is worth. Now flattened.

## One taken in substance but not literally

The section-fallback finding identifies a real hole — `x.com/feed` could ground on any title containing "feed". But requiring domain evidence, as suggested, regresses the feature's motivating case. Browser titles truncate from the left, so `fix: thing #11 · acme/omi` never contains "github" — only the repo path. Measured:

| variant | nano | reasoning lane |
| --- | --- | --- |
| current | 74% / 31% | 89% / 52% |
| require domain evidence (as suggested) | 71% / **26%** | 88% / **47%** |
| block generic section words (implemented) | 74% / 31% | 89% / 52% |

Requiring domain evidence costs 5 points of recall and removes **zero** false merges on the measured set. Blocking only generic section words (`feed`, `inbox`, `home`, `builds`, …) closes the hole at no measured cost:

- `x.com/feed` on "My Feed - Random Site" → rejected
- `evil.com/inbox` on "Inbox - Gmail" → rejected
- `github.com/acme/omi` on the truncated GitHub title → still accepted

132 tests pass, 29 in `ContextDestinationKeyTests`. All 20 preflight gates green.

PR #11594, opened 2026-08-15
fix(app): local-dev Firebase placeholders must be format-valid, not just fake

formed2forge:
## Why

Firebase's native SDK format-validates `GOOGLE_APP_ID` and `API_KEY` at `Firebase.initializeApp()` time — before any network call, and before Auth is routed to the local emulator. The local-harness placeholder Firebase configs (`setup/prebuilt/GoogleService-Info-Local.plist`, `google-services-local.json`, `lib/firebase_options_local.dart`) used syntactically invalid values (`GOOGLE_APP_ID = "1:000000000000:ios:omi-dev-local"`, `API_KEY = "local-firebase-auth-emulator-api-key"`), so a `bash setup.sh ios` dev build crashes on launch — never gets anywhere near the local emulator it's supposed to talk to.

## What changed

Measured directly on iPhone 17 Pro / iOS 27.0 while verifying an unrelated iOS UIScene fix:

1. `'Configuration fails. It may be caused by an invalid GOOGLE_APP_ID in GoogleService-Info.plist...'` — `GOOGLE_APP_ID` must look like `<project_number>:<number>:<platform>:<hex>`; `omi-dev-local` isn't hex.
2. After fixing that: `'API Key length must be 39 characters, API Key must start with `A`.'` — the placeholder API key was 37 chars and didn't start with `A`.

Fixed both in all three sources that describe this same fake identity (iOS plist, Android json, Dart `FirebaseOptions` consts) — `main.dart` selects the Dart consts explicitly for `OMI_APP_PROFILE=local_dev`, but the crash traced to native Firebase plugin registration reading the bundled plist even earlier, before Dart's `main()` runs, so both had to change to actually fix it end to end.

Also patches `BUNDLE_ID` in the copied plist (`generate_ios_custom_config` in `setup.sh`) to track the per-hostname-suffixed `APP_BUNDLE_IDENTIFIER` that dev builds already get, since the placeholder's static `BUNDLE_ID` never matched it. This turned out not to be the actual crash trigger for the two errors above — Firebase's error message named the wrong field both times — but it's a real mismatch, cheap to fix alongside this, and could plausibly bite some other Firebase code path that does check bundle identity.

Values remain non-secret placeholders that reach no real Firebase project — `IS_ANALYTICS_ENABLED`/`IS_GCM_ENABLED`/`IS_APPINVITE_ENABLED`/`IS_ADS_ENABLED` stay `false`, matching the existing posture of this config.

## Tests

- `app/test/unit/firebase_local_options_test.dart` (new) — asserts the Dart `FirebaseOptions` consts and the raw plist/json source files match the format Firebase's SDK actually enforces (39-char API key starting with `A`; `\d+:\d+:(ios|android|web):hex` app ID), so a future edit can't reintroduce a human-readable placeholder in either field without a red test.
- `app/scripts/generate_ios_custom_config_bundle_id_test.sh` (new) — hermetic shell test for the `BUNDLE_ID` patch, following the existing `setup_app_env_test.sh` pattern (sources `setup.sh`, stubs the machine-dependent `generate_device_suffix`/`detect_apple_team_id` helpers, asserts the patched plist's `BUNDLE_ID` matches the generated `APP_BUNDLE_IDENTIFIER`).

Both pass locally:
```
flutter test test/unit/firebase_local_options_test.dart   # 6/6 passed
bash scripts/generate_ios_custom_config_bundle_id_test.sh # passed
```

Manually re-verified the actual crash-to-fix progression on-device (iPhone 17 Pro / iOS 27.0) by patching the built app's config file at each step and relaunching via `devicectl`, before committing the same fix to the source files here.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/firebase_options_local.dart
+++ b/app/lib/firebase_options_local.dart
@@ -26,16 +26,16 @@ class DefaultFirebaseOptions {
   }
 
   static const android = FirebaseOptions(
-    apiKey: 'local-firebase-auth-emulator-api-key',
-    appId: '1:000000000000:android:omi-dev-local',
+    apiKey: 'AIzaSyDEMOOMILOCALFAKEKEY00000000000000',
+    appId: '1:000000000000:android:0000000000000000',
     messagingSenderId: '000000000000',
     projectId: 'demo-omi-local',
     storageBucket: 'demo-omi-local.localhost',
   );
 
   static const ios = FirebaseOptions(
-    apiKey: 'local-firebase-auth-emulator-api-key',
-    appId: '1:000000000000:ios:omi-dev-local',
+    apiKey: 'AIzaSyDEMOOMILOCALFAKEKEY00000000000000',
+    appId: '1:000000000000:ios:0000000000000000',
     messagingSenderId: '000000000000',
     projectId: 'demo-omi-local',
     storageBucket: 'demo-omi-local.localhost',
@@ -43,8 +43,8 @@ class DefaultFirebaseOptions {
   );
 
   static const macos = FirebaseOptions(
-    apiKey: 'local-firebase-auth-emulator-api-key',
-    appId: '1:000000000000:ios:omi-dev-local',
+    apiKey: 'AIzaSyDEMOOMILOCALFAKEKEY00000000000000',
+    appId: '1:000000000000:ios:0000000000000000',
     messagingSenderId: '000000000000',
     projectId: 'demo-omi-local',
     storageBucket: 'demo-omi-local.localhost',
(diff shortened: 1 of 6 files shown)


kodjima33 (approved):
Local-dev-only Firebase placeholders were format-invalid so the native SDK crashed at initializeApp before reaching the emulator; values stay fake, and the new format tests lock it in.

PR #11595, opened 2026-08-15
fix(app): migrate flutter_contacts to 2.x — 1.x crashes under UIScene

formed2forge:
## Why

`flutter_contacts` 1.1.9+2's iOS registration force-unwraps `UIApplication.shared.delegate!.window!!.rootViewController!`, which is nil under the UIScene lifecycle (#11568) and crashes with `EXC_BREAKPOINT` on launch. Plugin registration is alphabetical and synchronous (`GeneratedPluginRegistrant.m`), so this crash takes down every alphabetically later plugin too — `SharedPreferences`, `Sqflite`, `PermissionHandler`, `Geolocator`, `WebView`, ~27 in total. Confirmed via full crash backtrace on iPhone 17 Pro / iOS 27.0 while verifying the UIScene migration.

## What changed

No 1.x patch exists past 1.1.9+2 — checked the full pub.dev version list. The fix only landed in 2.x, a complete Dart-API rewrite. Migrated the 3 call sites:

- `requestPermission({readonly})` → `bool` becomes `permissions.request(PermissionType.read|readWrite)` → `PermissionStatus` (`granted`/`limited` stand in for the old `true`).
- `getContacts(withProperties: true, withPhoto: false)` becomes `getAll(properties: {ContactProperty.phone})` — narrower and cheaper, since these call sites only ever read phones/displayName, and those two are always fetched regardless of the requested property set.
- `Contact.displayName`: non-null `String` → nullable `String?` in 2.x.
- `Phone.label`: bare `PhoneLabel` enum → `Label<PhoneLabel>` wrapper; `phone.label.name` → `phone.label.label.name`.

`phone_calls_page.dart` and `phone_call_provider.dart` both already import `permission_handler`, which also exports a type named `PermissionStatus` — writing `flutter_contacts`' bare `PermissionStatus` in either file is an ambiguous-import error. Fixed with `hide PermissionStatus` on the `permission_handler` import in both files; neither file spells that package's own `PermissionStatus` by name anywhere, so nothing else changes.

Bonus: 2.3.1 ships a `Package.swift`, so `flutter_contacts` drops off the "doesn't support Swift Package Manager" warning Flutter prints on every iOS build.

## Verification

`flutter analyze`: 0 errors on the 3 changed files (1 pre-existing unrelated lint), 0 new errors project-wide. Combined with the UIScene migration (#11570) and a Firebase local-config fix (separate PR) on a throwaway community-build test, the app reached the sign-in screen on iPhone 17 Pro / iOS 27.0, confirmed visually — the first time this plugin registration chain has completed under UIScene on this project.

No dedicated regression test: there's no way to unit-test a native UIScene-launch crash without device access. The version constraint itself (commented in `pubspec.yaml`) is the guard — a downgrade to 1.x would silently reintroduce the crash.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11595?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/conversation_detail/widgets/share_to_contacts_sheet.dart
+++ b/app/lib/pages/conversation_detail/widgets/share_to_contacts_sheet.dart
@@ -75,7 +75,8 @@ class _ShareToContactsBottomSheetState extends State<ShareToContactsBottomSheet>
     });
 
     // Request contacts permission using flutter_contacts' own method
-    final permissionGranted = await FlutterContacts.requestPermission();
+    final status = await FlutterContacts.permissions.request(PermissionType.readWrite);
+    final permissionGranted = status == PermissionStatus.granted || status == PermissionStatus.limited;
 
     if (!permissionGranted) {
       if (!mounted) return;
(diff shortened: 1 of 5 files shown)


kodjima33 (approved):
flutter_contacts 1.x force-unwraps rootViewController and takes down the whole plugin registration chain under UIScene; 2.x migration is the only fix, and all 3 call sites are migrated correctly.

PR #11597, opened 2026-08-15
feat(desktop): give the context director a single bounded retrieval hop

Git-on-my-level:
# feat(desktop): give the context director a single bounded retrieval hop

The director can only reason over the current bucket's facts plus the task
list, so it cannot answer a question that is on screen when the answer lives
elsewhere in the user's history. Live example: an email draft asking "when was
the last omi release?" while the answer ("The macOS beta is now live on
0.12.171") sat in a different bucket's Telegram-derived insight.

## What changed

The director's structured output (flag-on only) gains a `lookup_query` string.
When the first response sets one, the engine runs one retrieval — the backend
tool-search endpoints the chat surface already uses,
`/v1/tools/conversations/search` and `/v1/tools/memories/search` via
`APIClient` — quotes up to 3+3 result previews into the prompt, and calls the
director exactly once more. The second response is final; `plan` refuses any
further lookup and the engine has exactly one second call site, so there is no
loop: at most two director calls per visit, and the second happens only when
the director asked for it.

That REST seam was chosen over the agent tool-execution layer (which drags in
chat-session scaffolding) and over a new proactive lane (the backend
`ProactiveOperation` enum is closed). No backend change.

## Invariants held

- **Grounding does not weaken.** `store.validatedEntryRefs` /
  `validatedFactIDs` semantics are untouched, and the non-silence gate still
  requires a validated bucket entry ref and bucket fact, exactly as before.
  Retrieved items are citable only under their own namespaces
  (`conversation:<id>`, `memory:<id>`) against an allowlist containing exactly
  the items quoted to that one call — empty whenever no hop completed, so a
  hallucinated retrieved ref fails closed. The delivery's provenance row
  records the query, every ref offered, and bounded title/preview per item, so
  a delivered citation stays verifiable from the database alone.
- **Untrusted framing.** Retrieved content rides only in the uncached suffix,
  below the volatile prompt, which sits below
  `ScreenDerivedContent.untrustedPreamble`; titles and previews are flattened
  to single lines so they cannot forge `== SECTION ==` structure, and the
  section's static framing lines precede every retrieved byte.
- **Prompt caching.** Both calls share the identical stable prompt and
  `cacheKey`; the lookup instruction is static text covering both call states,
  so the second call re-reads the first call's cached prefix.
- **Never lose a decision.** Retrieval failure, empty results, a stale visit,
  a flipped free gate, or a failed second call all keep the first decision in
  force (`finalDecision(first:second:)`); each source also fails independently.
- **Flag off is byte-identical to today**: same schema (no `lookup_query`
  property), same prompt, one call. `isRetrievalHopEnabled` follows the
  destination-routing pattern — gated under `isEnabled`, non-production
  defaults on with `OMI_FORCE_BUCKET_RETRIEVAL=0` as the only off switch,
  production beta reads the inverted fail-open kill switch
  `context_buckets_retrieval_kill`.

## Testing

148 tests pass under the proactive filter (132 on main, 16 new in
`ContextProactivityRetrievalHopTests`). Logic tests only — no eval harness.
Pinned: flag-off byte-identity of schema and prompt, the two-call ceiling
(`plan` with `priorHops: 1` refuses), retrieval failure preserving the first
decision, per-call allowlist validation with hallucinated refs rejected and
bucket refs keeping their existing path, untrusted ordering and cache-prefix
exclusion of retrieved content, query/preview flattening and bounds, and
provenance auditability.

## Tuning knobs

- Trigger wording: `ContextProactivityPromptBuilder.directorLookupInstruction`.
- Result count: `ContextDirectorRetrievalHop.perSourceResultLimit` (3 per
  source, conversations + memories).
- Preview sizes: `maximumTitleLength` / `maximumPreviewLength` (120/400).
- Query bounds: `maximumQueryLength` / `minimumQueryLength` (200/3).
- Remote stop: `context_buckets_retrieval_kill`; local off:
  `OMI_FORCE_BUCKET_RETRIEVAL=0`.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.pr-body.md
+++ b/.pr-body.md
@@ -1,83 +1,76 @@
-## Problem
+# feat(desktop): give the context director a single bounded retrieval hop
 
-Bucket identity is `sha256(appName::normalized window title)` — `subjectID` defaults to the reference hash, so **one distinct title is always exactly one bucket**. Measured on a real beta profile over ~38h:
+The director can only reason over the current bucket's facts plus the task
+list, so it cannot answer a question that is on screen when the answer lives
+elsewhere in the user's history. Live example: an email draft asking "when was
+the last omi release?" while the answer ("The macOS beta is now live on
+0.12.171") sat in a different bucket's Telegram-derived insight.
 
-- 742 visits → **83 buckets**, 83 distinct normalized keys → 83 buckets, **1:1, zero collapsing ever occurred**
-- 50 of the 83 are Google Chrome, from only 135 entries
-- X alone is 10 buckets (`(4) home / x`, `(2) notifications / x`, `home / x`, individual posts…)
-- 4 separate buckets for different PRs in one repository
-- 26 buckets (31%) hold exactly one entry and never accumulate
-- **74 of 83 buckets never produced a single notification**
+## What changed
 
-The cost is not notification noise — the dwell gate already keeps this debris out of the reasoning path. It is **missed joins**: nothing can connect four views of one repo, or a repo and the channel where it is discussed.
+The director's structured output (flag-on only) gains a `lookup_query` string.
+When the first response sets one, the engine runs one retrieval — the backend
+tool-search endpoints the chat surface already uses,
+`/v1/tools/conversations/search` and `/v1/tools/memories/search` via
+`APIClient` — quotes up to 3+3 result previews into the prompt, and calls the
+director exactly once more. The second response is final; `plan` refuses any
+further lookup and the engine has exactly one second call site, so there is no
+loop: at most two director calls per visit, and the second happens only when
+the director asked for it.
 
-`workstreamID` is *not* the fix. It sits inside `UNIQUE(subjectKind, subjectID, COALESCE(workstreamID,''))`, so it can only ever partition — two different titles have different `subjectID`s and stay two buckets whatever `workstreamID` holds. The only merge lever in the existing design is `subjectID` convergence via `subject_bindings`, which today only ever writes self-referential rows (verified: 83 rows, all `subjectID == referenceHash`, all `workstreamID` NULL).
+That REST seam was chosen over the agent tool-execution layer (which drags in
+chat-session scaffolding) and over a new proactive lane (the backend
+`ProactiveOperation` enum is closed). No backend change.
 
-## Change
+## Invariants held
 
-**Leading unread-badge strip.** The existing count regexes are `$`-anchored, so `(4) Home / X` and `Home / X` hash apart. Now stripped for browsers and messaging apps. The trailing form stays messaging-only — `Issue (123)` outside messaging is genuine identity, and `ContextTitleNormalizerTests` still guards that. Also fixes a shipped pattern that matched `– 4 new messages` but never Slack's web title `- 1 new item`.
+- **Grounding does not weaken.** `store.validatedEntryRefs` /
+  `validatedFactIDs` semantics are untouched, and the non-silence gate still
+  requires a validated bucket entry ref and bucket fact, exactly as before.
+  Retrieved items are citable only under their own namespaces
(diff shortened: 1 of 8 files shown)


Git-on-my-level:
## Tuned against live data — the hop was missing its own motivating case

The implementation was structurally right, but the instruction wording made the hop fire only a third of the time on the scenario it exists for. Measured against the reasoning-lane model on the case from the live profile (drafting a message asking when the last release shipped, with the answer present elsewhere in history but absent from the bucket):

| instruction | fires on target (n=8) | fires on 60 ordinary real snapshots | control: answer already supplied | control: nothing asked |
| original | **3/8** | 0/60 | 0/3 | 0/3 |
| **shipped in `9e9831f819`** | **8/8** | **0/60** | 0/3 | 0/3 |

Strictly better on both axes: reliable when it matters, and it still buys **zero** additional director calls in ordinary use. A third variant carrying a worked example scored identically to the shipped one, so it was dropped rather than carried on the cached prefix.

Emitted query on the target case: `most recent omi macOS release date` — which is the phrase that retrieves the relevant history.

## Two findings worth carrying forward

**No naturally occurring instance of the target shape in 90 real snapshots.** The situation this hop serves — a question being written whose answer lives elsewhere — did not occur once across the sampled recent activity. A near-zero live trigger rate is therefore *expected* and is not evidence the hop is broken.

**The director has a ~23% intrinsic decision flip rate.** Identical prompts, two runs, 7 of 30 decisions differed. Any future A/B on director behaviour has to clear that floor before a difference means anything — an earlier 38% raw difference in this same exercise turned out to be noise.

Both are recorded in the commit message.

## What is still unproven

This measures the **trigger** end only. End-to-end answer quality — the second call with real retrieved results — is not verified, because the full `contextEntered` path is not harnessable outside the app. Closing that requires dogfooding a bundle with the flag on.

148 tests pass, 20/20 preflight checks green.

Git-on-my-level:
## Review comments addressed in `ec8b106c7d`

**P2 — stale outcome after the hop. Valid, and a real regression of an existing invariant.** The pre-hop engine never persisted a delivery once ownership was revoked or the visit ended: the guard after the director call terminalized as `stale_visit` and returned. The hop reopened that window — its guard runs *before* a retrieval round trip plus a second model call, and a stale hop returned a nil decision that the caller treated as an ordinary fallback, keeping the first decision and writing a delivery against context the previous code would have refused.

The same guard now runs again after the hop, re-checked directly rather than by trusting the hop to self-report, so a future failure path added inside the hop cannot quietly bypass it.

**P3 — provenance preview truncation. Valid.** Provenance recorded 200 characters while the prompt quoted up to 400, so a citation grounded between those cuts was not verifiable from the delivery row — exactly the auditability this hop claims. Provenance now records what was actually quoted.

**P3 — changelog trailing period. Valid.** Confirmed no other unreleased fragment ends with one.

## Not changed: the `covers:` entry

The comment reads that list as a per-file test claim. The flow carries **25 `covers:` entries against 7 steps**, so it denotes area ownership rather than per-file coverage, and the desktop e2e coverage gate *requires* an entry for every new desktop Swift source — removing it fails preflight. If the intent is that `covers:` should imply an exercising step, that is a change to the gate itself and to the 24 pre-existing entries, not something to fix in this PR.

148 tests pass, 20/20 preflight checks green.

PR #11600, opened 2026-08-15
feat(desktop): open the task composer inside Today, aligned with its rows

kodjima33:
## What

Adding a task needed the composer to be summoned before there was anywhere to
type, and the composer that appeared sat at the top of the whole list rather
than in the section it would file into.

The composer now opens **as the first row of Today** — from the **+** button or
Cmd+N — and closes on Esc. Because it lives under Today, what you type there is
filed under Today: Enter creates the task with an end-of-day due date, and the
row stays open so tasks can be entered back to back.

It is deliberately not a standing row: the page shows no input until you ask
for one.

### Alignment

The composer sat a checkbox-width left of every task, because task rows open
with a drag-handle column it did not reserve. It now uses the same
`RigidLeadingGutter` that already solved this for Suggested rows (promoted from
private to shared rather than duplicated), plus TaskRow's checkbox geometry — a
20pt circle in a 24pt frame — so one checkbox column runs straight down the
list.

none

## Verification

Built and ran the named bundle `omi-today-btn` (dev backend) and drove the real
UI: AX press on the **+** button, then synthetic key events posted through the
app's event queue — actual typing, not a view-model call.

- Page loads with no composer; pressing **+** opens it as Today's first row with
  the caret already in it.
- Typed "align", pressed Return: task created (`ub7yuJFHguyWsBzWs3ow`), filed
  into Today (13 → 14) directly under the composer, which stayed open for the
  next entry. Screenshot: `.cross-review-verify.png`.
- Alignment confirmed against the same screenshot: composer circle and text sit
  on the same columns as the task rows beneath it.
- Test task deleted afterwards.

Hermetic coverage for the states that are awkward to reach in the UI: the
composer is suppressed while a row-anchored insert is open (they share text and
focus state), Cmd+N leaves the completed view, and an empty To-Do list renders
the list rather than the empty state so the composer has a home.

Commands run: `xcrun swift build -c debug`, `xcrun swift test --filter
"InlineTaskTodayButtonTests|TaskDetailPanelTests|TasksViewModel"` (30 tests, 0
failures), `make preflight` (20 checks).


aryanorastar (changes requested):
Requesting changes (own desktop lane red):

Opening the task composer as the first Today row (+ / Cmd+N), filing end-of-day due dates, staying open for back-to-back entry, and Esc-to-close is strong Tasks UX. Alignment with Today rows is the right product shape.

**Blockers on this tip:**
1. **Desktop Swift Static & Test Contracts / Build** are failing — required lanes on a desktop Tasks PR. Green them or prove unrelated flake with a clean re-run.
2. Confirm Preflight/Failure-Class (and any Line-Count exceptions) are honest on the tip if Hygiene/Preflight wake.

Soft asks once green: bulk-edit / completed views correctly hide the composer; empty Today still shows while composer is open.

Re-review after Desktop Swift required checks are green.

PR #11603, opened 2026-08-15
fix(desktop): give notification cards one size per display mode

kodjima33:
## Problem

Proactive suggestion cards arrive at different sizes, and sometimes render collapsed — the card squeezed down to just its icon and dismiss button, with the text gone.

The notification surface was computed independently at four call sites: `showNotification`, both `currentSurfaceSize` overloads, and `frameForCurrentState`. They disagreed:

- two added the **hover-expanded** bar height (`isHoveringBar ? expandedBarSize.height : minBarSize.height`), so the same suggestion got a different height depending on where the pointer happened to be;
- one always added the **notch chrome** height, even on a non-notch display.

Whichever path last resized the window decided the card's size, and the shortest of them squeezed the card when the bar expanded underneath it.

## Fix

One authority — `notificationSurfaceSize(usesNotchIsland:chromeHeight:)` — with no transient pointer state among its inputs. All four call sites route through it. The pill card stacks under the hover-expanded bar height so an expanding bar can never squeeze it; the only thing that may move the height is the display's own chrome.

Also adds `debug_suggestion_card` to the non-production bridge. The real delivery path needs a live model verdict (`probe_suggestion_nudge` returns `assistant_unavailable` offline), which left the most-seen proactive card unverifiable outside production; it presents the same card through the same presenter.

## Two adjacent fixes in the same surface

- **A dropped animation completion could freeze the bar's size.** When an implicit
  `NSWindow` animation was superseded, its completion never ran, so the pending target stayed set and
  every later resize was compared against a frame the window would never reach. Extracted as
  `FloatingControlBarGeometry.shouldSkipResize(...)` with behavioral coverage.
- **The spoken name was read as "oh-my".** `NotificationSpeech.spokenSpelling(of:)` respells the
  standalone word for the synthesizer only; the displayed text is untouched. `debug_suggestion_card`
  now goes through `NotificationService.shared.sendNotification(...)`, so the debug seam exercises the
  same speech and telemetry the real path does instead of bypassing them.

## Product invariants

- INV-CHAT-1 — the diff touches floating-bar chat paths. Chat behaviour is unchanged: only the notification card's window geometry moved, and the card's own content and interactions are untouched.

Failure-Class: none

## Verification

- Built and ran the `omi-notifsize` bundle. Four consecutive deliveries — long copy and short copy — every one presented at exactly **556×203** (508 surface + 24pt glow each side; 43pt chrome + 8 + 128).
- Captured card matches the intended layout: icon, "Suggested by Omi", two-line message, dismiss button (screenshot in review).
- `swift test --filter FloatingBarGeometryTests` → **30 passed**, including the new `testNotificationSurfaceIsOneSizePerDisplayMode` guard.

Line-Count-Exception: desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift | 4602 -> 4632 | debug_suggestion_card makes the most-seen proactive card verifiable offline, alongside the existing debug_reach_error seam
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift | 5255 -> 5285 | the single notification-size authority plus the comment recording why hover state is excluded; it replaces four divergent computations

### DIFF
--- a/desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift
+++ b/desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift
@@ -1699,7 +1699,13 @@ final class DesktopAutomationActionRegistry {
           assistantId: "suggestion"
         )
       )
-      return ["shown": "true", "frame": NSStringFromRect(window.frame)]
+      return [
+        "shown": "true",
+        "frame": NSStringFromRect(window.frame),
+        "hasNotification": window.state.currentNotification == nil ? "false" : "true",
+        "showingAIConversation": window.state.showingAIConversation ? "true" : "false",
+        "isVisible": window.isVisible ? "true" : "false",
+      ]
     }
 
     // Cursor-free click diagnosis: report which window (any app's) is topmost at a screen point,
--- a/desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift
+++ b/desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift
@@ -1676,6 +1676,32 @@ final class DesktopAutomationActionRegistry {
       return ["shown": "true"]
     }
 
+    // The suggestion card is the most-seen proactive surface and the one whose geometry regressed
+    // (four call sites sized it differently). The real delivery path needs a live model verdict, so
+    // rendering it was unverifiable offline; this presents the same card through the same
+    // presenter. Non-prod bridge only.
+    register(
+      name: "debug_suggestion_card",
+      summary: "Show a sample 'Suggested by Omi' card on the bar for visual/geometry verification",
+      params: ["message"]
+    ) { params in
+      let mgr = FloatingControlBarManager.shared
+      guard let window = mgr.window else { return ["error": "no bar window"] }
+      if !mgr.isVisible { mgr.show() }
+      let message =
+        params["message"].flatMap { $0.isEmpty ? nil : $0 }
+        ?? "Film and post 'what did we ship' video with Chase From mark today"
+      window.showNotification(
+        FloatingBarNotification(
+          ownerID: RuntimeOwnerIdentity.currentOwnerId() ?? "",
+          title: "Suggested by Omi",
+          message: message,
+          assistantId: "suggestion"
+        )
+      )
+      return ["shown": "true", "frame": NSStringFromRect(window.frame)]
+    }
+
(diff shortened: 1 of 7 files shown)


aryanorastar (changes requested):
Requesting changes (CI / ratchet):

One `notificationSurfaceSize` authority (fixed hover-expanded bar height, no pointer-dependent resize) is exactly the right fix for collapsed suggestion cards. `FloatingBarGeometryTests` asserting notch vs pill parity looks like the invariant lock. Desktop Swift lanes are green on tip — good.

**Blocker:** **Hygiene / `product-file-line-count-ratchet` is red.** PR-body `Line-Count-Exception` lines must match the **exact tip** line counts for:
- `DesktopAutomationBridge.swift`
- `FloatingControlBarWindow.swift`

Update the exception numbers to the tip (or shrink), then re-run Hygiene.

Soft: `Failure-Class: none` on a user-visible geometry fix is fine if Preflight accepts it; otherwise use `new` / an FC id.

Happy to re-review once Hygiene is green.

undivisible (changes requested):
## Request changes

The current head is blocked: Desktop Swift Static & Test Contracts fails `AgentPillLifecycleTests.testFloatingBarResizeCoalescesNoopFrames` after this refactor removes the source-contract guard it asserts, and Desktop Swift Build & Tests is also red. Update that behavioral/static contract for the new resize authority (or preserve the required contract), then rerun the required Swift lanes. The PR body’s Line-Count-Exception entries are also stale for the current base/head (4577→4607 and 5207→5237), so update them before rerunning preflight.

aryanorastar:
Closing as superseded by #12630, #12882 and #13337.

Current `main` now owns notification geometry through shared `notificationSurfaceSize` / `closedSurfaceSize` authorities, routes the remaining closed-surface transitions through that composition, and enforces a grow-only surface floor at every resize, animation completion and state-change boundary. A mounted card therefore cannot settle below its required width or height, including during hover, PTT, thinking, hints, replacement and conversation transitions. Those paths have behavioral tests and live E2E geometry evidence.

The current `send_test_notification` automation action also delivers through the real `NotificationService` path, covering the debug-seam goal.

This branch remains at `8a0ea7a`, conflicts with the replaced geometry/automation implementation and fails the older resize contract. The standalone product-name pronunciation change did not land; if still wanted, it should be proposed separately rather than keeping this superseded geometry aggregate open.

PR #11621, opened 2026-08-15
Break the database.users import cycle with utils.subscription and honor the dev goals limit

ZachL111:
# Problem

Two reliability defects, rebuilt from #9966 and #9977, which had gone stale against current main. Both bugs are still present on `origin/main` today; I re-verified each against the current code before rebuilding.

**1. `database/users.py` and `utils/subscription.py` form an import cycle.**

`backend/database/users.py` imported back into `utils.subscription` at module scope:

```python
from models.other import Person
from utils.subscription import get_default_basic_subscription
import logging
```

while `backend/utils/subscription.py` does `import database.users as users_db` at module scope. Whichever module is imported first decides whether the pair works. Importing `utils.subscription` first raises:

```
ImportError: cannot import name 'get_default_basic_subscription' from partially initialized
module 'utils.subscription' (most likely due to a circular import)
```

Importing `database.users` first happens to succeed, which is why this surfaces intermittently as unrelated files change what gets imported first. The database -> utils edge is also the reverse of the documented backend layering (database/ -> utils/ -> routers/).

**2. `GET /v1/dev/user/goals` ignores its documented `limit` when `include_inactive=true`.**

The route clamps `limit` and then drops it on one branch:

```python
# oversized limit cannot stream the whole collection. Mirrors the GET /v3/memories hardening.
if include_inactive:
    goals = goals_db.get_all_goals(uid, include_inactive=True)
else:
    goals = goals_db.get_user_goals(uid, limit=limit)
```

`get_all_goals` had no limit parameter and streams the whole goals collection, so the clamp is dead code on that branch and the endpoint returns every goal the user has ever created, ignoring its own documented "**limit**: Maximum number of goals to return".

# Reachability / failure scenario

- Import cycle: any entry point that imports `utils.subscription` before `database.users` fails at import time with the error above. Reproduced with a fresh interpreter running `import utils.subscription` from `backend/`.
- Goals limit: `GET /v1/dev/user/goals?include_inactive=true&limit=10` from any developer API consumer returns the full collection and reads every goal document in Firestore. The path that lets it through is `routers/developer.py::get_goals` -> `database/goals.py::get_all_goals`.

# Fix

**Import cycle.** The `get_default_basic_subscription` import in `database/users.py` moves from module scope to its call sites: inside `get_user_subscription`, and at the top of `get_user_valid_subscription`. Since #9966 was written, main added a `provision=False` branch to `get_user_valid_subscription` that also calls `get_default_basic_subscription`, earlier in the function than the old fallback site, so the function-level import sits above the first call site and covers both. Each deferred import carries a comment recording why it must not be folded back to the top.

**Goals limit.** `get_all_goals` accepts an opt-in `limit` applied after the in-Python newest-first sort. The bound is deliberately not pushed into the Firestore query: `order_by('created_at')` excludes documents that lack the field entirely, and legacy or manually created goals can lack `created_at`, so a query-level order+limit would silently drop them. Goals without `created_at` coerce to `datetime.min` and sort last, so the bounded page is the newest goals with dateless legacy goals appearing only once dated goals run out, and the response honours the documented limit.

Deliberately not changed: every other `get_all_goals` caller (the goal-by-id lookup, `routers/goals.py`, and the MCP goal reads) omits `limit` and keeps its fetch-everything behavior; bounding those is a behavior change with no bug attached.

# Tests

- `backend/tests/unit/test_subscription_import_cycle.py`: runs `import utils.subscription` (and separately `import database.users`) first in a fresh interpreter and asserts both succeed standalone. A fresh interpreter is the only honest seam here: once pytest has imported either module, the order is decided for the whole session, so an in-process assertion would pass either way.
- `backend/tests/unit/test_dev_goals_limit_include_inactive.py`: router-level tests assert the clamp is delegated (captured kwargs for limit 5, ceiling clamp to 1000, and the active-only branch unchanged); database-level tests assert the bounded page is the newest goals, that a legacy goal without `created_at` is retained and sorts last instead of being dropped, that dated goals fill the page first, and that callers omitting `limit` keep the full unordered fetch with nothing pushed into the query.

# Verification

All commands run from `backend/` on Windows (Git Bash), `PYTHONUTF8=1`.

Prove-fail. Reverted the three product files to `origin/main` (`git diff origin/main -- <files>` printed nothing, byte-identical), then ran both test files:

```
FAILED tests/unit/test_subscription_import_cycle.py::test_utils_subscription_imports_standalone
  AssertionError: utils.subscription is not importable on its own:
  ImportError: cannot import name 'get_default_basic_subscription' from partially initialized
  module 'utils.subscription' (most likely due to a circular import)
FAILED tests/unit/test_dev_goals_limit_include_inactive.py::test_get_all_goals_bounds_the_query_when_limit_is_given
  TypeError: get_all_goals() got an unexpected keyword argument 'limit'
5 failed, 3 passed
```

The three passes on the unfixed tree are the behavior-preservation guards (importing `database.users` first, the active-only branch, and the unbounded default), which must pass on both sides. Restored the fix: `8 passed`.

Static and contract gate:

```
black --line-length 120 --skip-string-normalization --check <5 files>   -> 5 files would be left unchanged
python -m pyright -p pyrightconfig.json database/goals.py               -> 0 errors (users.py and routers/developer.py are in the pyright exclude list)
python scripts/check_module_stub_pollution.py                           -> 898 test files, 0 violations
python scripts/scan_async_blockers.py --dirs routers utils              -> 0 findings in fail scope
python scripts/scan_import_time_side_effects.py                         -> 804 files, 0 violations
python scripts/check_conversation_lifecycle_writes.py                   -> passed
```

Workflow contracts: the touched files match three high-risk workflows (goals, users, developer). Ran all 13 unit and service test files listed by those workflows: `391 passed, 2 failed`, and the sync workflow file separately: `77 passed, 10 failed`. All 12 failures reproduce identically on a clean unmodified `origin/main` checkout in the same batch order (`tests/services/users/test_account_deletion.py` pair passes standalone on both trees), so they are pre-existing local-environment and test-ordering issues, not caused by this change. `testing/e2e/test_account_deletion_cloud_tasks.py` needs live services and errors locally on both trees.

`scripts/pr-preflight --suggest`: Product invariants affected: none. No `fix:` commits, so no failure-class declaration is required.

Full `scripts/pr-preflight --pr-body-file` run: every selected manifest check passes (including backend-route-policy-baseline, backend-async-blockers, backend-import-purity, backend-workflow-contracts, the line-count ratchet with the declarations below) except `desktop-backend-candidate-probe-fixtures`, which fails on this Windows machine with "gemini_proxy: total response deadline is unsupported on this runner". The files that check exercises are byte-identical to `origin/main` in this branch (the diff is five backend files), so it is a local runner-capability issue of the same class as the known Windows-only `detect_platform` false failure; CI runs on Linux.

# Impact

- The import cycle breaks any standalone consumer of `utils.subscription` and makes backend import order fragile: whether it bites depends on which module a given entry point or test session happens to import first. Developer-facing reliability, low direct user impact.
- The goals endpoint bug affects developer API consumers who pass `include_inactive=true`: response size and Firestore read cost grow with the user's total historical goal count instead of respecting the documented bound. Narrow trigger, real cost on goal-heavy accounts.

Line-Count-Exception: backend/database/users.py | 2131 -> 2140 | Breaking the utils.subscription import cycle moves one module-level import to deferred call sites with comments recording why each must stay deferred so the import is not folded back to the top and silently re-broken.
Line-Count-Exception: backend/routers/developer.py | 2107 -> 2111 | The four added lines are the comment recording why the goals bound is applied after the in-Python sort rather than pushed into a Firestore order_by that would drop legacy goals lacking created_at.

### DIFF
--- a/backend/database/goals.py
+++ b/backend/database/goals.py
@@ -342,27 +342,18 @@ def get_all_goals(
 ) -> List[Dict[str, Any]]:
     """Fetch a user's goals, newest first.
 
-    ``limit`` bounds the read at the query instead of in Python, so a caller that only needs a
-    page cannot stream the whole collection. It is opt-in: every existing caller omits it and
-    keeps the fetch-everything behaviour they rely on.
+    ``limit`` bounds the returned page after the in-Python newest-first sort. It is opt-in:
+    every existing caller omits it and keeps the full-list behaviour they rely on.
 
-    The query orders by ``created_at`` descending so the bounded page is the newest ``limit``
-    goals rather than an arbitrary slice that only looks sorted after the in-Python sort below.
-
-    ``limit`` is only supported together with ``include_inactive=True``. That shape carries no
-    equality filter, so ordering by ``created_at`` is served by Firestore's automatic
-    single-field index. Combining a limit with the ``is_active`` filter would need a composite
-    index that this project does not declare, and Firestore answers a missing composite index
-    with an opaque 500, so the unsupported combination is rejected here rather than in
-    production.
+    The bound is deliberately NOT pushed into the Firestore query: ``order_by('created_at')``
+    excludes documents that lack the field entirely, and legacy or manually created goals can
+    lack ``created_at`` — a query-level order+limit would silently drop them. Instead the full
+    stream is sorted here (goals without ``created_at`` coerce to ``datetime.min`` and sort
+    last) and the page is sliced, so the bound caps the response payload while dateless legacy
+    goals still appear once dated goals run out.
     """
-    if limit is not None and not include_inactive:
-        raise ValueError('get_all_goals(limit=...) is only supported with include_inactive=True')
-
     collection = _get_db(firestore_client).collection(users_collection).document(uid).collection(goals_collection)
     query = collection if include_inactive else collection.where(filter=FieldFilter('is_active', '==', True))
-    if limit is not None:
-        query = query.order_by('created_at', direction=firestore.Query.DESCENDING).limit(limit)
     goals = [normalize_goal_storage(_goal_dict(doc), goal_id=doc.id) for doc in query.stream()]
     if not include_inactive:
         goals = [goal for goal in goals if goal['is_active']]
--- a/backend/database/goals.py
+++ b/backend/database/goals.py
@@ -337,15 +337,37 @@ def get_all_goals(
     uid: str,
     include_inactive: bool = False,
     *,
+    limit: Optional[int] = None,
     firestore_client: Any = None,
 ) -> List[Dict[str, Any]]:
+    """Fetch a user's goals, newest first.
+
+    ``limit`` bounds the read at the query instead of in Python, so a caller that only needs a
+    page cannot stream the whole collection. It is opt-in: every existing caller omits it and
+    keeps the fetch-everything behaviour they rely on.
+
+    The query orders by ``created_at`` descending so the bounded page is the newest ``limit``
+    goals rather than an arbitrary slice that only looks sorted after the in-Python sort below.
+
+    ``limit`` is only supported together with ``include_inactive=True``. That shape carries no
+    equality filter, so ordering by ``created_at`` is served by Firestore's automatic
+    single-field index. Combining a limit with the ``is_active`` filter would need a composite
+    index that this project does not declare, and Firestore answers a missing composite index
+    with an opaque 500, so the unsupported combination is rejected here rather than in
+    production.
+    """
+    if limit is not None and not include_inactive:
+        raise ValueError('get_all_goals(limit=...) is only supported with include_inactive=True')
+
     collection = _get_db(firestore_client).collection(users_collection).document(uid).collection(goals_collection)
     query = collection if include_inactive else collection.where(filter=FieldFilter('is_active', '==', True))
+    if limit is not None:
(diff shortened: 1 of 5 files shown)


ZachL111:
Addressed all three findings in ae3fcd2a0a:

- Legacy goals without created_at: correct catch. Firestore's order_by excludes documents missing the field, so the query-level bound could silently drop legacy or manually created goals. The limit now applies after the in-Python newest-first sort (missing created_at coerces to datetime.min and sorts last), so every goal stays reachable and the response still honours the documented limit. The PR body is updated to describe the payload-level bound.
- The import-cycle tests now carry pytest.mark.slow with a comment citing the tests README rule; they deselect cleanly under the PR lane's 'not integration and not slow' expression. The fresh-interpreter seam stays, since in-process assertions cannot detect import-order bugs once either module is cached.
- The database tests now assert newest-first directly: the fake streams docs oldest-first and the test requires the returned page to be the last N ids in reverse creation order, plus a dateless-legacy-goal retention case and a dated-goals-fill-the-page-first case.

kodjima33 (approved):
Breaking the database.users -> utils.subscription cycle and honoring the dev goals limit both look right; holding on the red required checks.

PR #9055, opened 2026-07-05
Add GET /v3/speech-profile/status: consolidated speech-profile status

ZachL111:
## What
Adds `GET /v3/speech-profile/status`, a single consolidated speech-profile status object.

## Why
The speech-profile duration is cached on upload (`set_speech_profile_duration` writes `users:{uid}:speech_profile_duration`) but no endpoint ever reads it back, and existence vs url are split across `GET /v3/speech-profile` and `GET /v4/speech-profile`. A settings screen that wants to show "profile recorded, N seconds" has to make two calls and still cannot see the duration. This folds them into one.

## Details
- New Redis getter `get_speech_profile_duration(uid)` mirrors the existing setter (same key, same `@try_catch_decorator` fail-open pattern, returns 0.0 when unset).
- The endpoint returns `{has_profile, duration_seconds, url}`, reusing `get_user_has_speech_profile` and `get_profile_audio_if_exists` (already imported). Plain `def` like its `/v3` and `/v4` siblings.
- No route collision: the existing `/v3/speech-profile*` routes are all literal.

## Test
`backend/tests/unit/test_speech_profile_status.py`: the Redis getter parses bytes and str values, returns 0.0 when unset, and fails open to None on a Redis error; the endpoint mapping (including the None-duration coercion to 0.0) runs in CI, where `routers.speech_profile` imports PyAV (absent locally).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/9055?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #8940, opened 2026-07-03
Add GET /v1/screen-activity/summary for a user's aggregated screen activity

ZachL111:
## What

The desktop app captures screen-activity rows into `users/{uid}/screen_activity`, read back only internally by the MCP server. There was no first-party REST endpoint for the aggregated view a client would show a user (per-app screenshot counts, first/last seen). The `get_screen_activity_summary(uid, start_date, end_date)` database helper already computes it, but nothing exposed it.

## Change

Add `GET /v1/screen-activity/summary`, authenticated as the current user, returning the aggregated per-app summary with an optional `date` (YYYY-MM-DD, validated -> 422 on a bad date) narrowed to a single-day range. Single new endpoint in `backend/routers/focus_sessions.py`, reusing the existing helper. No change to the database layer.

## Test

New `backend/tests/unit/test_screen_activity_summary_endpoint.py`: no-date passes `None`; a date becomes a single-day range; a bad date returns 422. Imports the router normally, patches the db helper with `monkeypatch.setattr`, and calls the handler directly. Passes the import-purity and stub-pollution scanners.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/8940?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

PR #8939, opened 2026-07-03
Add GET /v1/screen-activity to list a user's screen-activity rows

ZachL111:
## What

The desktop app captures screen-activity rows into `users/{uid}/screen_activity` (app name, window title, OCR text, timestamp), but they were only ever read back internally by the MCP server. A first-party mobile/desktop client had no REST endpoint to fetch a user's own screen activity. The `get_screen_activity(uid, start_date, end_date, app_filter, limit)` database helper already reads them, but nothing exposed it.

## Change

Add `GET /v1/screen-activity`, authenticated as the current user, returning the rows newest-first with:
- an optional `date` (YYYY-MM-DD, validated with `validate_calendar_date` -> 422 on a bad date) narrowed to a single-day range,
- an optional `app_filter`,
- a bounded `limit` (1..5000, default 500).

Single new endpoint in `backend/routers/focus_sessions.py` (its existing focus-stats endpoints already read a uid + date the same way), reusing the existing helper. No change to the database layer.

## Test

New `backend/tests/unit/test_screen_activity_list_endpoint.py`: no-date passes `None` bounds; a date becomes a single-day range; a bad date returns 422. Imports the router normally, patches the db helper with `monkeypatch.setattr`, and calls the handler directly. Passes the import-purity and stub-pollution scanners.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/8939?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

PR #8431, opened 2026-06-27
Fix data loss and a layout crash in the Windows local knowledge graph

ZachL111:
## Summary

Correctness fixes for the local knowledge graph and the memory-extraction dedupe that feeds it. Each one silently drops or corrupts data that the user expects to see in their graph or memories. Found while reading the code; no filed issue, no happy-path behavior change.

## What this fixes

- C, C++, and C# all collapsed into a single technology node. `slugify` strips `+` and `#`, so `slugify('C')`, `slugify('C++')`, and `slugify('C#')` all return `c`, giving every one the id `c:technology`. `mergeGraph` keys nodes by id, so the last one silently overwrites the others and two of the three languages disappear. `slugify` now maps `+` to `p` and `#` to `sharp` so the ids stay distinct. This is consistent with `EXT_TO_TECH`, which already maps `.c`, `.cpp`, and `.cs` to distinct labels.
- The force-directed layout threw on a dangling edge. `mergeGraphs` intentionally keeps edges whose endpoints are not both present, and onboarding can emit a user-to-app edge before the user node exists, but `computeLayout` mapped every edge into the simulation without checking, so `forceLink` threw `missing: <id>` and took down the whole layout. It now filters edges to those whose endpoints both exist, matching what `useGraphSimulation` already does.
- Onboarding emitted duplicate app nodes and edges. `buildApps` had no dedupe (unlike its `deriveAppNodes` sibling), so a repeated app from an installed-apps scan, or two names that slug to the same id, produced two nodes and two edges with identical ids. It now dedupes by id.
- One oversized screen window lost all of its extracted facts forever. `budgetSegments` returns `[]` the moment the first segment alone exceeds the budget, and the caller treats an empty result as handled and advances its watermark past those frames. A single window with more than the budget of OCR text (a long doc or log) therefore had all of its facts skipped permanently. It now keeps a truncated first segment when nothing else fits.
- The memory dedupe key collapsed distinct facts. `normalize` stripped every non-letter/number/space char, so "Proficient in C++" and "Proficient in C#" both normalized to `proficient in c` and the second was dropped as a duplicate. It now keeps `+` and `#` in the key.

## Testing

- Extended the existing vitest suites for `kgTech`, `graphLayout`, `onboardingGraphModel`, and `screenGrouping`, and added a `memoryExtract.normalize` test. New assertions verified red to green; the prior assertions still pass.
- `npm run typecheck` passes for both projects.

## PR status check

[...]

PR #11622, opened 2026-08-15
Consolidate five approved open PRs: screen-activity reads, speech-profile status, and Windows hardening

ZachL111:
# Summary

This PR consolidates my five remaining older open PRs into one reviewable unit, so the open list stays short and everything in it can merge as it stands. Each change is squashed unchanged from the head of its source PR, one commit per original PR, and the source PRs are closed in favor of this one:

- #8939: GET /v1/screen-activity, an authenticated list read over the rows the desktop app syncs via POST /v1/screen-activity/sync
- #8940: GET /v1/screen-activity/summary, the aggregated per-app view (screenshots per app, first and last seen)
- #9055: GET /v3/speech-profile/status, one consolidated read for has_profile, sample count, and duration
- #8430: hardening for the Windows OCR and automation helper processes
- #8431: data-loss and crash fixes for the Windows local knowledge graph

All five were MERGEABLE against main at consolidation time. Three (#8939, #8940, #9055) were approved by kodjima33, and all five carry Git-on-my-level review passes on their final heads (the earlier changes-requested reviews were resolved and dismissed on the source PRs; the review history lives there).

# What each change does

**GET /v1/screen-activity (#8939).** The desktop app captures screen activity, but the rows were only readable internally (MCP). This exposes them to the user's own client: date filter validated with validate_calendar_date (422 on an impossible date), app_filter, and a bounded limit Query(ge=1, le=5000). A single-day query uses an inclusive same-day upper bound (23:59:59) because get_screen_activity applies end_date as timestamp <= 'YYYY-MM-DD HH:MM:SS.999', so a next-day-midnight end would leak the following day's first second into a privacy-sensitive single-day result.

**GET /v1/screen-activity/summary (#8940).** Same exposure logic and the same day-boundary handling for the aggregated per-app summary, reusing the existing get_screen_activity_summary helper with no DB-layer change.

**GET /v3/speech-profile/status (#9055).** The app needed several round trips to answer "does this user have a usable speech profile and how long is it". This returns has_profile, sample count, and duration in one read. duration_seconds is gated on has_profile so a stale write-ahead cache entry that outlives a deleted profile reports 0.0 instead of a phantom duration; the endpoint stays a pure read and does not clear state on a GET.

**Windows OCR and automation helper hardening (#8430).** A 64 MiB MAX_FRAME_BYTES ceiling on the helper protocol decoder plus a poisoned-buffer reset before the throw, so an oversized or corrupt frame cannot wedge the stream; a spawn cooldown and an identity guard (this.child === child) around recycle and exit handling in both supervisors; no-op stream error listeners so an EPIPE from a dying helper no longer crashes the Electron main process; and try/finally disposal of SoftwareBitmap and process handles in the C# helpers (a real native-memory leak on the exception path).

**Windows local knowledge graph fixes (#8431).** Membership checks use Object.hasOwn, so a node id matching an inherited property name (such as __proto__) is no longer treated as present, which previously crashed the force layout; edges with missing endpoints are filtered before layout; C, C++, and C# slugify to distinct ids matching what EXT_TO_TECH expects; normalize keeps + and # so C++ and C# stop deduplicating to the same memory key; the onboarding graph model dedupes apps; and screen grouping keeps a truncated first segment.

# Consolidation mechanics

Each source branch was squash-merged onto current origin/main as its own commit. The only overlap between the five is routers/focus_sessions.py, where #8939 and #8940 both append an endpoint at the same anchor; the resolved file contains both functions, each byte-identical to its source branch (verified with a block-level diff against the branch head). Everything else merged without conflicts, and the new test files are byte-identical to their source branches.

# Tests

- backend/tests/unit/test_screen_activity_list_endpoint.py: no-date passes None bounds, a date becomes an exact single-day range, a bad date is 422, and a real-boundary test records the actual >= and <= timestamp strings through the Firestore-shaped fake.
- backend/tests/unit/test_screen_activity_summary_endpoint.py: the same coverage for the summary endpoint, including the exact-day boundary strings.
- backend/tests/unit/test_speech_profile_status.py: status shape, and the stale-cache regression (has_profile false with a positive cached duration maps to 0.0).
- desktop/windows: helperProtocol oversized-frame rejection and poisoned-buffer recovery; graphLayout dangling-edge and __proto__ regression with finite coordinates after layout; kgTech slug distinctness; memoryExtract normalize behavior behind an apiClient mock; onboardingGraphModel dedupe; screenGrouping truncated-first-segment.

# Verification

Backend, from backend/ on Windows (Git Bash):

```
python -m pytest tests/unit/test_screen_activity_list_endpoint.py \
  tests/unit/test_screen_activity_summary_endpoint.py \
  tests/unit/test_speech_profile_status.py -q
-> 15 passed
```

Prove-fail: reverted routers/focus_sessions.py, routers/speech_profile.py, and database/redis_db.py to origin/main (git diff against origin/main printed nothing, byte-identical) and reran; every endpoint test fails with AttributeError: module 'routers.focus_sessions' has no attribute 'screen_activity_db' (and the speech-profile equivalents), which is the expected failure shape for feature tests against a tree without the feature. Restored: 15 passed.

```
black --line-length 120 --skip-string-normalization --check <6 backend files>  -> unchanged
python -m pyright -p pyrightconfig.json routers/focus_sessions.py database/redis_db.py -> 0 errors
  (routers/speech_profile.py is in the pyright exclude list)
python scripts/check_module_stub_pollution.py     -> 899 test files, 0 violations
python scripts/scan_async_blockers.py --dirs routers utils -> 0 findings in fail scope
python scripts/scan_import_time_side_effects.py   -> 804 files, 0 violations
```

Line-count ratchet: all touched backend files are under the 1500-line threshold (redis_db.py is at 1177 after the change), so no exception declarations are needed. testing/workflow_contracts.json lists none of the touched files as workflow sources.

Desktop, from desktop/windows (pnpm install, then vitest on the six touched test files):

```
pnpm exec vitest run src/main/ocr/helperProtocol.test.ts src/renderer/src/lib/graphLayout.test.ts \
  src/renderer/src/lib/kgTech.test.ts src/renderer/src/lib/memoryExtract.test.ts \
  src/renderer/src/lib/onboardingGraphModel.test.ts src/renderer/src/lib/screenGrouping.test.ts
-> Test Files 6 passed (6), Tests 38 passed (38)
```

scripts/pr-preflight: Product invariants affected: none. No fix: commits, so no failure-class declaration is required. The full local preflight passes apart from the known Windows-only runner limitations already documented for this machine (CI on Linux is the authority).

# Impact

Nothing here changes existing behavior: the three backend endpoints are additive reads scoped to the authenticated caller's own uid, and the Windows changes only harden paths that previously crashed, leaked, or lost data. Consolidating trades five approved-but-idle PRs for one unit that can be reviewed and merged in a single pass.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11622?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/redis_db.py
+++ b/backend/database/redis_db.py
@@ -1057,6 +1057,15 @@ def set_speech_profile_duration(uid: str, duration: float) -> None:
     r.set(f'users:{uid}:speech_profile_duration', str(duration))
 
 
+@try_catch_decorator
+def get_speech_profile_duration(uid: str) -> float:
+    """Read the cached speech profile duration in seconds (0.0 if unset)."""
+    val = r.get(f'users:{uid}:speech_profile_duration')
+    if not val:
+        return 0.0
+    return float(val.decode() if isinstance(val, bytes) else val)
+
+
 # ******************************************************
 # ************ DAILY SUMMARY NOTIFICATIONS *************
 # ******************************************************
--- a/backend/database/redis_db.py
+++ b/backend/database/redis_db.py
@@ -1058,8 +1058,9 @@ def set_speech_profile_duration(uid: str, duration: float) -> None:
 
 
 @try_catch_decorator
-def get_speech_profile_duration(uid: str) -> float:
-    """Read the cached speech profile duration in seconds (0.0 if unset)."""
+def get_speech_profile_duration(uid: str) -> Optional[float]:
+    """Read the cached speech profile duration in seconds (0.0 if unset; None if
+    the read itself fails, per try_catch_decorator's fail-open contract)."""
     val = r.get(f'users:{uid}:speech_profile_duration')
     if not val:
         return 0.0
--- a/backend/route_policy_manifest.yaml
+++ b/backend/route_policy_manifest.yaml
@@ -74,6 +74,75 @@ routes:
     path: /v1/screen-activity/sync
     policy: *desktop_migration_policy
   - route_type: http
+    method: GET
+    path: /v1/screen-activity
+    policy:
+      review_status: reviewed
+      auth:
+        mechanisms:
+          - firebase_id_token
+        placement: dependency
+        scopes: []
+      byok: not_applicable
+      rate_limit:
+        policy_name: none
+        key_subject: none
+        enforcement: none
+        placement: none
+      timeout_class: default_method
+      surface: first_party_app
+      visibility: first_party
+      data_domain: user_profile
+      deprecation:
+        state: active
+      owner: backend
+  - route_type: http
+    method: GET
+    path: /v1/screen-activity/summary
+    policy:
+      review_status: reviewed
+      auth:
+        mechanisms:
+          - firebase_id_token
+        placement: dependency
+        scopes: []
+      byok: not_applicable
+      rate_limit:
+        policy_name: none
+        key_subject: none
+        enforcement: none
+        placement: none
+      timeout_class: default_method
+      surface: first_party_app
+      visibility: first_party
+      data_domain: user_profile
+      deprecation:
+        state: active
+      owner: backend
+  - route_type: http
+    method: GET
+    path: /v3/speech-profile/status
+    policy:
+      review_status: reviewed
+      auth:
+        mechanisms:
+          - firebase_id_token
+        placement: dependency
+        scopes: []
+      byok: not_applicable
+      rate_limit:
+        policy_name: none
+        key_subject: none
+        enforcement: none
+        placement: none
+      timeout_class: default_method
+      surface: first_party_app
+      visibility: first_party
+      data_domain: user_profile
+      deprecation:
+        state: active
+      owner: backend
+  - route_type: http
     method: POST
     path: /v1/tts/synthesize
     policy: *desktop_migration_policy
--- a/backend/route_policy_manifest.yaml
+++ b/backend/route_policy_manifest.yaml
@@ -81,9 +81,10 @@ routes:
       auth:
         mechanisms:
           - firebase_id_token
+          - admin_key_uid_prefix
         placement: dependency
         scopes: []
-      byok: not_applicable
+      byok: validated_when_headers_present
       rate_limit:
         policy_name: none
         key_subject: none
@@ -104,9 +105,10 @@ routes:
       auth:
         mechanisms:
           - firebase_id_token
+          - admin_key_uid_prefix
         placement: dependency
         scopes: []
-      byok: not_applicable
+      byok: validated_when_headers_present
       rate_limit:
         policy_name: none
         key_subject: none
@@ -127,9 +129,10 @@ routes:
       auth:
         mechanisms:
           - firebase_id_token
+          - admin_key_uid_prefix
         placement: dependency
         scopes: []
-      byok: not_applicable
+      byok: validated_when_headers_present
       rate_limit:
         policy_name: none
         key_subject: none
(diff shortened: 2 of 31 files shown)


ZachL111:
Addressed the review in 64b87c6984:

- helperProcess and bridge: explicit dispose() now clears the crash cooldown so the next request spawns immediately, recycle() with no live child is a no-op, and windowInfo() keeps its rejection contract instead of returning an empty result. The three dispose regression tests pass again.
- systemAudioMute: added the same catch-and-recycle guard around decoder.push the OCR helper and automation bridge use, so the decoder's fail-fast throw has no unguarded consumer.
- graphLayout: the degree map is now a null-prototype object, with a regression test for a node actually named __proto__ keeping its edges. Good catch on the setter path.
- focus_sessions: both screen-activity endpoints now declare typed response models and the /v1/screen-activity prefix joined the app-client contract surface; specs and generated clients regenerated.
- speech_profile: the status response now includes sample_count (from the expand recordings) with a response model, and the upload caches duration only after upload_profile_audio succeeds.
- redis_db: get_speech_profile_duration annotated Optional[float].

Two findings I am not taking:

- The two route_policy_manifest suggestions to mark the new routes byok: validated_when_headers_present with admin_key_uid_prefix. The manifest currently records 59 reviewed get_current_user_uid routes as byok: not_applicable and 5 as validated_when_headers_present, so not_applicable is the dominant convention my entries follow. If the inventory owners decide the 5-route form is the correct semantics for every get_current_user_uid route, that flip should happen across all 59 in one dedicated change, not piecemeal in this consolidation.

The supervisor-duplication note is fair; extracting a shared helper for the three process supervisors is follow-up-sized and does not belong in a consolidation PR, so I left the two mirrored implementations as they were shipped.

undivisible (changes requested):
The route-policy entries for `/v1/screen-activity`, `/v1/screen-activity/summary`, and `/v3/speech-profile/status` must reflect `get_current_user_uid`: BYOK is validated when headers are present, and the applicable admin-key auth mechanism must be declared.

ZachL111:
You are right on both counts, thanks for the catch. I verified against the dependency itself: get_current_user_uid routes through verify_token, which accepts admin-key impersonation alongside Firebase ID tokens, and it calls validate_byok_request unconditionally after auth, which is a no-op only when no BYOK headers arrive. So not_applicable was wrong for all three routes.

Fixed in ddf573cc5e: /v1/screen-activity, /v1/screen-activity/summary and /v3/speech-profile/status now declare mechanisms [firebase_id_token, admin_key_uid_prefix] and byok validated_when_headers_present, matching the existing entries for this dependency (for example /v1/chat/materialize-prompts). test_route_policy_inventory passes 26/26 locally and CI is rerunning.

For context on how the wrong values got in: I seeded the new entries from the majority convention in the manifest (most entries are not_applicable) instead of deriving them from the route's actual dependency. The materialize-prompts entries were the right precedent to copy.

undivisible (approved):
Verified the auth-policy correction for all three routes; it resolves my requested changes.

Issue #11518, opened 2026-08-13
Desktop: raw TaskActionItem.deleted reads disagree with isRetired, so status-only-retired tasks read as live


formed2forge:
`TaskActionItem` has two ways to ask "is this task retired?", and they disagree. Roughly 32 call sites read the raw `deleted` field while ~20 read the `isRetired` projection, so a task retired *only* through canonical lifecycle status reads as **live** in the former.

The type documents why the projection exists (`desktop/macos/Desktop/Sources/TaskActionItem.swift:82-87`):

```swift
/// Server list/detail responses project soft retirement through canonical
/// lifecycle status and may omit the legacy `deleted` field. Keep that
/// projection here so every local cache and surface applies the same rule.
var isRetired: Bool {
  deleted == true || taskStatus == "cancelled" || taskStatus == "superseded"
}
```

So a server row with `status: "cancelled"` and no `deleted` field is retired — but `task.deleted != true` evaluates to "live" for it. Verified examples that read the raw field on objects that have not been through the local cache (where the column is written from `isRetired`, masking the difference):

- `desktop/macos/Desktop/Sources/MainWindow/ChatFirst/ChatFirstShell.swift:569` — a fresh detail fetch:
  ```swift
  if let task = try? await APIClient.shared.getActionItem(id: id), task.deleted != true {
    return task
  }
  ```
  A cancelled/superseded task is returned here as though live.
- `desktop/macos/Desktop/Sources/Stores/TasksStore.swift:2761`, `:2766` — `remoteTask.deleted != true`, `hydratedTask.deleted != true` on freshly fetched rows.
- `desktop/macos/Desktop/Sources/MainWindow/Pages/TasksPage.swift:200-201`, `:2546-2549`, `:5525` — filter predicates and `isDeleted`.
- Live-lane filters over `tasksStore.tasks`: `MainWindow/DesktopTopBar.swift:58`, `MainWindow/ChatFirst/ChatFirstTasksPage.swift:153`, `:413`, `:417`, `:547`, `MainWindow/ChatFirst/ChatFirstShell.swift:544`, `:566`, `:573`, `:604`, `:610`, `MainWindow/Pages/DashboardPage.swift:1006`.

[...]

PR #11626, opened 2026-08-15
fix(desktop): stop dashboard navigation from opening a retired task (#11518)

kodjima33:
## Bug

`DashboardIntelligenceStore.taskForNavigation(taskID:)` fetched the task detail and returned it with **no retirement check at all**, so a "What matters now" recommendation minted before its task was cancelled, superseded, or deleted still routed the user to that task.

The user-visible symptom is a silent dead end rather than a stale card: every live Tasks lane filters retired rows out, so the `TaskNavigationRequestStore` request `openRecommendation` enqueues can never be consumed — the dashboard switches to the Tasks tab and selects nothing, with no explanation.

Its sibling in the same file, `candidateForNavigation`, already gates on `SuggestedTasksStore.canPresentForNavigation(candidate)` and reports "This Suggested item is no longer available." The task path had no equivalent.

## Root cause

Retirement is re-derived by each reader instead of being asked of the one authority. `TaskActionItem.isRetired` exists precisely for this, and its doc comment records why:

```swift
/// Server list/detail responses project soft retirement through canonical
/// lifecycle status and may omit the legacy `deleted` field. Keep that
/// projection here so every local cache and surface applies the same rule.
var isRetired: Bool {
  deleted == true || taskStatus == "cancelled" || taskStatus == "superseded"
}
```

So a raw `deleted` read is not sufficient on anything straight off the wire, and `taskForNavigation` did not even do that much.

This is mostly invisible on cache paths because every writer normalizes: `ActionItemRecord.from` persists `deleted: item.isRetired`. It is observable only on objects read directly from an API response.

## Fix

- `DashboardIntelligenceStore.taskForNavigation` now requires `!task.isRetired` alongside the existing id match, and surfaces the same "This task is no longer available." error the Suggested path uses.
- `ChatFirstShell.taskForFocus` — the only other site that tests an API response directly rather than a cache row — moves from `task.deleted != true` to `!task.isRetired` (#11518).

Cache reads elsewhere are left alone: they are already normalized, and #11518 explicitly warns that `deleted` is a field on unrelated types, so a blind sweep is not the fix.

## Tested

```
cd desktop/macos
./scripts/dev-feedback.py --once swift 'DashboardIntelligenceStoreTests'
  → Executed 30 tests, with 0 failures  (PASS in 593.66s)
```

The two new tests execute production behavior through the injected `DashboardIntelligenceClient` seam, not a source scrape. To prove they are load-bearing, the `!task.isRetired` guard alone was reverted and they were re-run:

```
testLegacyDeletedTaskIsNotOpenedByNavigation  failed — deleted: true opened as live
testRetiredTaskIsNotOpenedByNavigation        failed — cancelled and superseded both opened as live
  → Executed 2 tests, with 6 failures
```

The guard was then restored and the suite re-run green (above).

Not exercised live: reproducing it in the running app needs a server-side task retired between projection mint and click. The seam the fix sits behind is the same one `taskForNavigation`'s existing coverage uses.

## Product invariants

None matched — `scripts/pr-preflight --suggest` reports no invariant citations required for these paths.

Fixes part of #11518.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/ChatFirst/ChatFirstShell.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/ChatFirst/ChatFirstShell.swift
@@ -580,7 +580,10 @@ private struct ChatFirstRestoredTasksHost: View {
     if let task = tasksStore.tasks.first(where: { $0.id == id && $0.deleted != true }) {
       return task
     }
-    if let task = try? await APIClient.shared.getActionItem(id: id), task.deleted != true {
+    // Straight off the wire, so legacy `deleted` may be absent and retirement
+    // only stated through canonical lifecycle status. Cache reads above and
+    // below are already normalized (`ActionItemRecord.from` stores `isRetired`).
+    if let task = try? await APIClient.shared.getActionItem(id: id), !task.isRetired {
       return task
     }
     await tasksStore.loadCompletedTasks()
--- a/desktop/macos/Desktop/Sources/MainWindow/Dashboard/DashboardIntelligenceStore.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Dashboard/DashboardIntelligenceStore.swift
@@ -400,7 +400,10 @@ final class DashboardIntelligenceStore: ObservableObject {
     do {
       let task = try await client.getActionItem(id: taskID)
       guard requireCurrentOwner(ownerScope) else { return nil }
-      guard task.id == taskID else {
+      // The detail response is the freshest word on retirement, and it projects
+      // it through canonical lifecycle status — a recommendation minted before
+      // the task was cancelled/superseded/deleted must not open it as live.
+      guard task.id == taskID, !task.isRetired else {
         error = "This task is no longer available."
         return nil
       }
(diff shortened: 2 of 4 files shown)


kodjima33:
Independently re-verified before merge (not trusting the opening run's claims):

- `dev-feedback.py --once swift 'DashboardIntelligenceStoreTests'` on this branch: **30 tests, 0 failures**.
- Proved the new tests are real guards, not tautologies: reverted only the production line (`guard task.id == taskID, !task.isRetired` → `guard task.id == taskID`) and re-ran the two new tests — **6 failures**; both `taskStatus: cancelled`/`superseded` and legacy `deleted: true` handed the retired task back as a live navigation target with `error == nil`. Restored the line, suite green again.
- `isRetired` is the repo-wide retirement projection (`TaskActionItem.swift:92`), already used by SpineStore, TasksStore, ChatFirstContentBlockViews and ActionItemStorage — this fix mirrors the existing pattern rather than inventing a check.

CI green (Desktop Swift Static & Test Contracts, Release Compile, Build & Tests). Merging.

PR #11627, opened 2026-08-15
fix(backend): unbreak conversation merge and cascade delete behind the canonical fence

mdmohsin7:
## Summary

Conversation merge **and cascade delete** have both been broken since **13 Aug**, from the same cause. Reported symptom: the app sits on "merging" until it is restarted, and afterwards the merged conversation **and both originals** are all listed.

## Root cause

`perform_merge_async` builds the merged conversation, then deletes each source. `_delete_conversation_and_related_data` retracts the source's canonical memories first and **re-raises** if that fails — deliberately, so a source is never deleted while active memories still point at it.

Retraction runs through the canonical replace boundary, which `_require_canonical_intake_enabled()` fences whenever `MEMORY_MODE` is not `write`/`read`. **Production runs `off`.**

So every merge created the merged conversation, raised on the first source, unwound into `_handle_merge_failure`, and reset the sources to `completed`. **Step 10 — the FCM that tells the app the merge finished — was never reached**, which is why the UI never left "merging" without a restart, and why both originals came back.

Production logs, every merge, 1 ms apart:

```
Error processing merged conversation: 503: Memory writes are globally paused
Error deleting memories for <id>: canonical memory intake is globally paused
Merge failed with exception: canonical memory intake is globally paused
Merge failed for conversations
```

| Day | Merge failures |
|---|---:|
| 10 Aug | 0 |
| 12 Aug | 0 |
| **13 Aug** | **27** |
| 14 Aug | 18 |

Same origin as #11623: `5724a10084` introduced `_require_canonical_intake_enabled`, and the first revision carrying it deployed 13 Aug 05:09.

## Fix

The fence closes *intake*, so a conversation ingested while it was closed has **no canonical memories to retract in the first place**. Cleanup now reads the source cohort — an unfenced read — and skips retraction only when the source has no active canonical memories. Nothing can dangle because there is nothing there.

Unchanged:
- A source that **does** have memories still aborts exactly as before.
- A cohort read that **fails** also aborts, rather than assuming the source is empty.
- The fence callback still fires on the skip path, since everything after retraction destroys source state and failure handling must not treat the source as untouched.

## On the existing tests

`test_merge_conversations_canonical_delete.py` describes a source that *has* memories. Rather than rewrite its assertions — which would delete the guard it exists to provide — the precondition is stated in a fixture. Every assertion in that file is byte-identical; all 8 still pass.

## A third breakage, found by self-review and unreported

`DELETE /v1/conversations/{id}?cascade=true` is the app's **only** delete path (`app/lib/backend/http/api/conversations.dart:136`). It called `retract_conversation_memories` with no error handling, so the same fence turned every delete into a 500:

| Day | cascade-delete 500s |
|---|---:|
| 11 Aug | 0 |
| 12 Aug | 0 |
| **13 Aug** | **300** (query cap; 400 of 400 sampled were 500) |
| 14 Aug | 300 |

Users have been unable to delete conversations for two days and nobody reported it. Same 13 Aug onset as merge and as the knowledge-graph 503s in #11623.

The skip logic now lives in `utils/memory/retraction_scope.py` and both call sites share it, so merge and delete cannot drift apart.

## Scope

This does **not** turn canonical intake back on. Merges work again; memory extraction stays fenced while `MEMORY_MODE=off`. That flag is still the underlying condition behind this, #11623, and the silently-skipped memory extraction on every processed conversation — worth a separate decision.

## Verification

```
pytest test_merge_skips_retraction_without_canonical_memories.py  -> 4 passed
  reverted the fix, re-ran                                        -> 3 of 4 FAIL (tests catch it)
pytest (5 merge test files)                                       -> 59 passed
black --line-length 120 --skip-string-normalization               -> unchanged
```

Not exercised against production — the fenced path is covered by unit tests, but I have not run a real merge on the deployed backend.

## Line-Count-Exception

Line-Count-Exception: backend/routers/conversations.py | 1508 -> 1515 | guards the cascade-delete retraction behind the shared skip helper; the fence otherwise makes every delete a 500

- **INV-MEM-1** — three product memory tiers. Matched on path globs; the tier model is untouched. This change only decides *when* retraction runs, never what a memory is or which tier it belongs to. No tier vocabulary, promotion rule, or storage shape is altered, and the guard test for the invariant is unchanged.

Worth noting **INV-MEM-3** (no legacy fallback after canonical selection) is *not* violated: nothing here falls back to a legacy store. When retraction is required it still runs through the canonical path and still aborts on failure; the skip applies only where the canonical scope is provably empty.

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/backend/routers/conversations.py
+++ b/backend/routers/conversations.py
@@ -48,6 +48,7 @@ from utils.conversations.process_conversation import process_conversation, retri
 from utils.conversations import lifecycle as lifecycle_service
 from utils.executors import db_executor, llm_executor, postprocess_executor, run_blocking, submit_with_context
 from utils.memory.memory_service import MemoryService
+from utils.memory.retraction_scope import retraction_can_be_skipped
 from utils import byok
 from utils.conversations.search import (
     ConversationSearchUnavailableError,
@@ -785,7 +786,13 @@ def delete_conversation(
         # Delete associated memories and action items before removing the conversation doc
         # so a partial failure cannot orphan derived data.
         db_client = getattr(db_client_module, 'db', None)
-        MemoryService(db_client=db_client).retract_conversation_memories(uid, conversation_id)
+        memory_service = MemoryService(db_client=db_client)
+        # Retraction is fenced with canonical intake (MEMORY_MODE). Skipping it
+        # when there is provably nothing to retract keeps delete working while
+        # the fence is closed; anything real still raises rather than orphaning
+        # live memories against a deleted conversation.
+        if not retraction_can_be_skipped(uid, conversation_id, memory_service=memory_service, db_client=db_client):
+            memory_service.retract_conversation_memories(uid, conversation_id)
 
         action_items_db.delete_action_items_for_conversation(uid, conversation_id)
         background_tasks.add_task(delete_conversation_audio_files, uid, conversation_id)
--- a/backend/tests/unit/test_delete_conversation_cascade_retraction_fence.py
+++ b/backend/tests/unit/test_delete_conversation_cascade_retraction_fence.py
@@ -1,0 +1,60 @@
+"""Cascade delete must consult the retraction-skip guard.
+
+`DELETE /v1/conversations/{id}?cascade=true` is the app's only delete path
+(`app/lib/backend/http/api/conversations.dart:136`). It called
+`MemoryService.retract_conversation_memories` with no error handling, and that
+runs through the canonical replace boundary fenced by
+`_require_canonical_intake_enabled()` whenever `MEMORY_MODE` is not write/read.
+
+Production runs `off`, so every cascade delete raised and returned 500 — 100% of
+sampled requests from 13 Aug onward. It was never reported; it was found while
+fixing the merge failure that shares the same cause.
+
+This is a **static checker, not behavioural coverage**: it reads the endpoint's
+source rather than exercising the route, because importing `routers.conversations`
+into a unit-test session pulls in enough global state to break neighbouring
+import-isolated suites. The behaviour it protects — when the skip is allowed and
+when retraction must still run — is covered against the real helper in
+test_merge_skips_retraction_without_canonical_memories.py.
+"""
+
+from __future__ import annotations
+
+import pathlib
+import re
+
+ROUTER = pathlib.Path(__file__).resolve().parents[2] / "routers" / "conversations.py"
+
+
+def _delete_conversation_source() -> str:
+    source = ROUTER.read_text()
+    start = source.index("def delete_conversation(")
+    end = source.index("\n@router.", start)
+    return source[start:end]
+
+
+def test_the_cascade_branch_guards_retraction_behind_the_skip_helper():
+    body = _delete_conversation_source()
+    assert "retract_conversation_memories" in body, "expected the cascade branch to still retract"
+    assert "retraction_can_be_skipped" in body, (
+        "cascade delete must consult retraction_can_be_skipped; without it the canonical "
+        "fence turns every delete into a 500"
+    )
+
+
+def test_the_guard_precedes_the_retraction_call():
+    body = _delete_conversation_source()
+    assert body.index("retraction_can_be_skipped") < body.index(
+        "retract_conversation_memories"
+    ), "the guard must run before the retraction it protects"
+
+
+def test_the_retraction_call_is_inside_the_guard():
+    # `if not retraction_can_be_skipped(...)` — a guard that does not actually
+    # gate the call would satisfy the ordering check above but fix nothing.
+    body = _delete_conversation_source()
+    guarded = re.search(
+        r"if not retraction_can_be_skipped\([^)]*\):\s*\n\s+memory_service\.retract_conversation_memories\(",
+        body,
+    )
+    assert guarded, "retraction must be inside the `if not retraction_can_be_skipped(...)` branch"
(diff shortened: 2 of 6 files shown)


mdmohsin7:
Both valid — fixed in the latest commit. The second one pointed at a better design than what I'd shipped.

**Historical records outside the predicate's scope (P2).** Correct, and it's the same dangling-evidence bug I was fixing. `retract_conversation_memories` retracts the canonical cohort *and* tombstones historical live rows from `self.history.all_live(uid)` whose `conversation_id` or evidence names the source. My predicate only read the canonical cohort, so a source with historical rows and no canonical items would have read as empty and been deleted with those rows still pointing at it. `_source_retraction_is_a_noop` now covers both stores.

**Race with intake enabled (P1).** Also correct, and your suggested remedy is the right one. The skip now applies **only while the fence is closed** — which is the only case it was ever meant to address. With intake enabled, retraction works and the original always-retract path is correct, so there is no window to race.

New coverage exercises the real predicate rather than mocking it: intake-enabled always retracts and never consults the probe; historical records naming the source (by `conversation_id` or by conversation evidence) are not a no-op, while unrelated ids and non-conversation evidence are; an active canonical item is never a no-op.

```
test_merge_skips_retraction_without_canonical_memories.py -> 12 passed (was 4)
5 merge test files                                        -> 67 passed
make preflight                                            -> 24 checks passed
```

Worth noting the shape of what you caught: my first version widened a safety invariant further than the bug required. The fix is narrower now — it changes behaviour only in the state that is actually broken in production, and leaves every other path exactly as it was.

mdmohsin7:
All three valid. Fixed in `e01f5a2`-range latest commit.

**Full history scan on the hot path (P2).** Right, and I measured it before changing anything rather than guessing. Random 120-user sample: median **0** memories, p90 24, p99 280 — a single page for most accounts. The heavy cohort is where it bites: median 54, **max 6,352**, and 9 of 41 sampled need more than one page. Merge scanned *per source*, so that account merging five conversations paid it five times.

Merge now takes **one** pass, collects the conversation ids referenced by live history, and reuses the set across sources. Single deletes still scan once, which the measurements say is one page. I did not add a targeted Firestore query — there is no index on `memories.conversation_id` and adding one is a prod schema change I would not fold into an incident fix.

**Fence transition race (P1).** Valid. The fence is now re-read *after* the scope reads; if it changed during them, the skip is refused and the caller retracts as before. To be precise about what this does and does not do: it removes the transition case you described, it does **not** make check-and-delete atomic. That needs a transaction spanning the memory store and the conversation document, which is a larger change than this fix warrants — flagging rather than quietly claiming it is closed.

**Tests never drove the real decision (P2).** The sharpest of the three, and correct. Every `_delete_conversation_and_related_data` test stubbed `retraction_can_be_skipped` to a fixed boolean, so a helper with the fence check inverted would have shipped green. There are now end-to-end cases that mock only the two data reads and let the real `canonical_intake_is_fenced` → `source_retraction_is_a_noop` → `retraction_can_be_skipped` chain decide, in both fence states. Verified by inverting the fence comparison: exactly those 2 fail.

**One thing you did not flag, found while fixing yours.** These merge suites replace *attributes* on real modules, which `restore_sys_modules` cannot undo, so they break any later suite calling the real `database.conversations` helpers. Mine added a second instance; the pre-existing `test_merge_conversations_canonical_delete.py` has had it all along. CI has never seen it because collection is alphabetical and the cascade suite sorts first. Both files now restore what they stubbed.

```
8 merge + delete + privacy files, merge ordered first -> 113 passed
same set, same order, before this commit             ->   3 failed
fence comparison inverted                            ->   2 end-to-end fail
make preflight                                       ->  24 checks passed
```

mdmohsin7:
One of the two holds and is fixed. The other I don't think can occur — reasoning below, happy to be shown wrong.

**Commit epoch across the scope reads (re-detected P1) — valid, fixed.** You're right that the fence re-read is insufficient: a write already in flight against the *pre-fence* mode commits without the fence ever moving, so both reads see it closed and the snapshot is stale. `retraction_can_be_skipped` now captures a canonical commit epoch — account generation, commit sequence and head commit id from `memory_state/head`, which the apply boundary writes on every committed canonical write — before the scope reads and compares it after. Any advance refuses the skip. A missing head is treated as a stable value rather than a mismatch, which matters because most accounts have no head at all while intake is fenced.

Still not atomic, and I'm not claiming otherwise. Closing it fully needs a transaction spanning the memory store and the conversation document. The epoch narrows the window to "nothing canonical committed while we looked".

**Stale precomputed set with a row referencing two sources (new P1) — I can't reproduce it.** The set is built *only* when the fence is closed (`merge_conversations.py`, guarded by `canonical_intake_is_fenced()`), and `retraction_can_be_skipped` short-circuits on the fence before consulting it.

So with the fence closed: source A is in the set → not a no-op → retraction runs → **the fence makes it raise** → the merge aborts. The loop never reaches source B. For A's retraction to succeed and remove the shared row, the fence must be open — and then the set is never read, because the short-circuit returns False first.

The reachable worst case is a stale entry causing a *conservative abort*, which is exactly the behaviour that predates this PR, not a missed retraction. If there's a path I've missed where the set is consulted with the fence open, I'd genuinely like to see it — that would change the answer.

```
test_merge_skips_retraction_without_canonical_memories.py -> 21 passed
8 merge + delete + privacy files, merge ordered first     -> 116 passed
make preflight                                            -> 24 checks passed
```

New tests cover the epoch directly: an epoch that advances during the reads refuses the skip, a stable one allows it, and a missing state head reads as stable rather than as a change.

kodjima33 (approved):
Conversation merge + cascade delete behind the canonical fence reads correct; holding the merge on the red Backend unit suite and the 787-line diff.

PR #11630, opened 2026-08-15
fix(backend): fail-closed chat quota accounting before billable turns

aryanorastar:
## Summary
- `/v2/messages` used to swallow `record_chat_quota_question` failures and still run the billable chat stream, so Free-plan question counters could fail open under Firestore outages.
- Recording now happens **before** persisting the human turn / session link. On failure: emit an SSE `done:` canned reply mobile can render (not a bare HTTP 503), and **do not** persist the human message (avoids orphan copies on retry).
- Voice `/v2/voice-messages` keeps best-effort recording after a visible `message:` frame (billable work already started there).
- Allowlists main-wide `#11613` desktop inventory gap (`/v1/tools/conversations/search-chunks`) so tip CI is not poisoned.

## Test plan
- [x] `pytest backend/tests/unit/test_chat_quota_counting_router.py` + `test_desktop_rest_inventory.py` (18 passed)
- [ ] CI backend unit suite on the PR
- Honest gaps: no induced Firestore outage in prod; desktop `/v2/desktop/messages` still best-effort (out of this shrink).

Failure-Class: none

Line-Count-Exception: backend/routers/chat.py | 1734 -> 1791 | Adds fail-closed quota accounting before billable `/v2/messages` turns, with SSE-visible retry copy and no orphan human persistence (#11630).

### DIFF
--- a/backend/routers/chat.py
+++ b/backend/routers/chat.py
@@ -257,7 +257,7 @@ def _build_quota_exceeded_reply(
     return ResponseMessage(**ai_msg.model_dump(), ask_for_nps=False)
 
 
-def _record_chat_quota_question_safe(
+def _record_chat_quota_question(
     uid: str,
     *,
     idempotency_key: str,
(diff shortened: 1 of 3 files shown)


aryanorastar:
Added the required PR metadata: `Failure-Class: none` and the line-count exception for `backend/routers/chat.py` (`1734 -> 1769`). The preflight failure was metadata-only; backend and hermetic checks are still running.

aryanorastar:
Addressed both cubic P2s + Backend Unit red:

1. **SSE vs 503** — accounting failure now returns HTTP 200 + `done:` canned reply (same mobile contract as quota-exceeded), not a bare 503.
2. **Orphan human messages** — counter write runs **before** `add_message` / `add_message_to_chat_session`; on failure nothing is persisted, so retries cannot multiply unanswered human rows.
3. **CI** — allowlisted `/v1/tools/conversations/search-chunks` (#11613 main-wide inventory gap).

Tip pushed with updated tests. Local: 18 passed on the touched suites.

kodjima33 (approved):
Records the chat-quota question before any billable work and returns an SSE-visible retry instead of failing open, without persisting an orphan human turn; covered by new router tests.

Issue #11511, opened 2026-08-13
ChatToolExecutorPolicyTests flakes with authorized_execution_owner_changed and blocked a desktop release cut


kodjima33:
`ChatToolExecutorPolicyTests.testTaskDeleteAndCompleteReachTaskStorage` failed on `main` at bdc5bc2233414b2b92be38028c282cc7a59febad (Desktop Swift CI run 31710277815, job 94481412653), which failed `Desktop Swift Build & Tests` and blocked the release train: the 14:33Z `Build Desktop Release Candidate` run (31710846941) refused to cut with `Desktop candidate source gate blocked: required check Desktop Swift Build & Tests for exact source SHA bdc5bc2233... completed with failure`. The same suite passes on the neighbouring SHAs.

Failure signature — both tool calls in the test returned the owner-authority revocation payload:

```
ChatToolExecutorPolicyTests.swift:26: XCTAssertTrue failed - complete_task returned:
{"ok":false,"error":{"code":"authorized_execution_owner_changed",
 "message":"The signed-in account changed while the authorized tool was executing."}}
... same for delete_task
```

That payload is emitted when `RuntimeOwnerAuthorizationAuthority` sees the persisted owner change out-of-band (`revokeUnexpectedOwnerMismatch`) — i.e. the process-global auth owner moved underneath the test after `setUp` established it.

Suspected cause: the suite drives `RuntimeOwnerAuthorityTestFixture`, which writes `auth_userId` into the **production-standard** `UserDefaults` domain, but it is not part of the sequential cluster in `desktop/macos/scripts/swift-test-suites.sh` (`SERIAL_SUITES`). That cluster exists precisely because, per the comment there, hosted runners can still route the standard domain through the shared `cfprefsd` even when workers have distinct `CFFIXED_USER_HOME`. Of the six suites that use the fixture — `CaptureScreenToolTests`, `ChatToolExecutorCreateMemoryTests`, `ChatToolExecutorPolicyTests`, `CreateCalendarEventToolTests`, `HomeSuggestionsStoreTests`, `StartupWarmupPolicyTests` — only `ChatToolExecutorCreateMemoryTests` is listed, so the other five run in parallel and can overwrite each other's owner.

The membership is a hand-maintained string, so every new fixture adopter silently re-opens this race; it should be derived from the source instead.

PR #11635, opened 2026-08-15
test(desktop): pin the signed-out owner in RealtimeHubSessionHandoffPolicyTests

kodjima33:
## Bug

The desktop release train is blocked. `Build Desktop Release Candidate` (run 31894961914, 16:14Z) failed its source gate:

```
required check Desktop Swift Build & Tests for exact source SHA 197e6997cbe3ee3301b4c0609ac065c90efecfc3 completed with failure
```

That SHA's `Desktop Swift CI` run (31890381385) failed one suite — `RealtimeHubSessionHandoffPolicyTests`, in `testBargeInFailoverWaitsForTransportAcknowledgementBeforeSpecializedStart`, at lines 200/203/209/211. The same code passed pre-merge; the diff on that SHA touched `DashboardIntelligenceStore`, not the realtime hub. No desktop candidate can be cut while the newest releasable desktop SHA carries a red required check.

## Root cause

Every policy in the suite is asserted against `RealtimeHubOwnerScope.signedOut`, and `.signedOut` is only *current* while the standard defaults domain holds no `auth_userId` (`RealtimeHubOwnerScope.isCurrent` → `RuntimeOwnerIdentity.currentOwnerId()`). The suite never pinned that owner, so it inherited whatever the process's standard domain held.

`scripts/swift-test-suites.sh` gives each worker its own `CFFIXED_USER_HOME`, but that does not isolate this domain: the xctest process's `UserDefaults.standard` resolves to the shared `com.apple.dt.xctest.tool` domain regardless of the worker home (verified locally — the run's writes land in the real user's `~/Library/Preferences/com.apple.dt.xctest.tool.plist`, never in the worker home). A concurrently running suite that establishes a signed-in owner therefore makes `isOwnerScopeCurrent(.signedOut)` false, `failoverBargeInReplacement` returns `false` at its owner guard, and the four assertions fail exactly as CI reported.

This is the failure class #11511 already documented for `ChatToolExecutorPolicyTests` ("a concurrent suite moving `auth_userId` underneath it"), which the runner mitigates by auto-serializing every suite that drives `RuntimeOwnerAuthorityTestFixture`.

## Fix

Adopt that same fixture in `setUp`/`tearDown` and establish a signed-out owner (`authOwnerID: nil`), restoring the prior owner afterwards. Test-only, +21/-1 in one file. Two effects:

1. The suite's owner is pinned instead of inherited.
2. It joins `swift-test-suites.sh`'s derived sequential cluster (membership is derived by grepping for `RuntimeOwnerAuthorityTestFixture`), so no concurrent suite can move the owner — or the BYOK key the barge-in test seeds — underneath it.

## Verification

Reproduced the CI-only failure deterministically by seeding the leak, then proving the fix removes it:

```
defaults write com.apple.dt.xctest.tool auth_userId "leaked-concurrent-owner"
env CFFIXED_USER_HOME=... xcrun swift test --package-path Desktop --filter 'RealtimeHubSessionHandoffPolicyTests/'
```

| state | result |
|---|---|
| before this commit, leaked owner seeded | 21 tests, **7 failures** — including the exact CI signature at lines 200/203/209/211 |
| after this commit, same leaked owner seeded | 21 tests, 0 failures |
| after this commit, clean domain | 21 tests, 0 failures |

`make preflight` passes (18 checks).

Separately, the failed `Desktop Swift CI` run on `197e6997cb` was re-run to clear the release-train source gate for that SHA; this PR removes the cause so it does not recur.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11635?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Tests/RealtimeHubSessionHandoffPolicyTests.swift
+++ b/desktop/macos/Desktop/Tests/RealtimeHubSessionHandoffPolicyTests.swift
@@ -10,6 +10,27 @@ import XCTest
   // notification regression step must compile the bundle without them.
 
   final class RealtimeHubSessionHandoffPolicyTests: XCTestCase {
+    // Every policy here is evaluated against `RealtimeHubOwnerScope.signedOut`,
+    // which is only current while the standard domain holds no `auth_userId`.
+    // Inheriting that domain let a concurrent suite's signed-in owner make
+    // `isOwnerScopeCurrent(.signedOut)` false, so `failoverBargeInReplacement`
+    // returned false and the barge-in test failed on CI only. Pin the owner
+    // (which also moves this suite into the sequential cluster).
+    private var ownerFixture: RuntimeOwnerAuthorityTestFixture?
+
+    override func setUp() async throws {
+      try await super.setUp()
+      let fixture = await RuntimeOwnerAuthorityTestFixture()
+      ownerFixture = fixture
+      await fixture.establish(authOwnerID: nil)
+    }
+
+    override func tearDown() async throws {
+      await ownerFixture?.restore()
+      ownerFixture = nil
+      try await super.tearDown()
+    }
+
     @MainActor
     func testPhysicalReplacementGateDrainsBeforeStartingAndCoalescesDuplicates() async {
       let gate = RealtimeHubTransportReplacementGate()


aryanorastar (commented):
Community review (test-only):

Pinning `RealtimeHubOwnerScope.signedOut` via `RuntimeOwnerAuthorityTestFixture` in setUp/tearDown is a plausible fix for the CI-only flake where a concurrent suite’s signed-in owner made `isOwnerScopeCurrent(.signedOut)` false and broke barge-in assertions. Moving the suite into the sequential cluster is the right accompaniment.

Please confirm before merge:
1. The barge-in / failover tests still exercise the **policy branch**, not only “fixture established.”
2. Add or keep a **signed-in owner** assertion path so pinning signed-out doesn’t become the only covered mode.
3. Wait for Desktop Swift Release Compile / related checks on this tip to finish green.

Tiny release-train unblock — thanks for the clear comment on the flake mechanism.

PR #11640, opened 2026-08-15
fix(desktop): cycle listening through all three audio recording modes

Archit-lal:
## Problem

`AssistantSettings.AudioRecordingMode` has three cases — `off`, `always`, `onlyMeetings` — but the listening control could only reach two of them. `toggleListening` read:

```swift
let nextMode = currentMode == .off ? .onlyMeetings : .off
```

so `.always` was **unreachable from the top bar and from Home**. Turning the microphone "on" always armed the meetings gate, which deliberately holds the mic shut until a call is detected; there was no way to ask for continuous recording from either shell. Reported from the top-bar button: it "only toggles between off and only meetings".

Only Meetings was also indistinguishable from Always On at a glance — same `mic`, same dot — even though the two behave completely differently.

## Change

- **A cycle, not a flip.** `CaptureListeningLogic.cycleListening` advances `Off → Always On → Only Meetings → Off`. Both shells already share this one function, so both gain the third mode.
- **Only Meetings is marked.** It carries `person.2.fill` in the corner. The mark is *additive corner ink* like the state dot — the base `mic` silhouette stays byte-identical in every state, which is the rule the cluster's header sets and `ShellStatusIconLegibilityTests` measures in pixels.
- **Only Meetings explains itself once.** Selecting it shows a popover — "the microphone stays closed until Omi detects a call, then records until it ends" — which auto-dismisses. It fires on the transition only, never on hover and never at rest, so the resting cluster stays wordless.
- **The tooltip names the destination.** "Click to stop" was true of a two-state switch and is a false promise from Always On, where a click selects Only Meetings and stops nothing.

## Product invariants

**INV-CHAT-1** (one shared transcript across surfaces) — cited because the diff touches
`desktop/macos/Desktop/Sources/MainWindow/Pages/DashboardPage.swift`, which is inside its path globs.
The change there is a single call-site rename (`toggleListening` → `cycleListening`); it does not
touch conversation ownership, transcript routing, or any chat surface, so the invariant's guard tests
are unchanged.

## Failure class

The control is the authority for the listening mode's transitions, but it only ever authored two of the three states the mode declares, leaving the third reachable exclusively from Settings. The fix is the class's canonical prevention: one transition function owns the whole state machine, and `testTheCycleCoversEveryDeclaredMode` fails if a future mode is added that the control cannot select.

## Review follow-ups (second commit)

Both findings from the automated review were valid and are fixed:

**P1 — the microphone stayed open when Only Meetings could not yet prove a call.** `reconcileCapture`'s
`guard meetingStateReady else { return }` returned *before* either pause branch. Selecting Only Meetings
from a live Always session builds a fresh detector, so that first pass runs with
`hasObservedState == false` and the microphone the previous mode opened kept running until the detector's
first asynchronous probe landed. A gate the user selects in order to close the mic has to fail closed, so
"not known yet" now means "not in a call". `MeetingGateReadinessPolicy` names the rule and
`pauseCaptureWhileMeetingGateUnknown` takes the same stop the normal gating takes, earlier.

This path only became reachable from the button in this PR: before it, Always On could not be selected
outside Settings, so `Always → Only Meetings` was not a transition the control could make.

**P2 — the tooltip promised a mode the click could not reach.** Without the microphone grant, a click
spends itself on the permission prompt and the mode does not move. The tooltip now says what is actually
missing until the grant exists.

Line-Count-Exception: desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift | 1736 -> 1738 | Fail-closed privacy fix for P1. The pause helper and its policy live in new files (`AppState+MeetingGatePause.swift`, `MeetingGateReadinessPolicy`); the two lines added here are the guard's call and its comment, which have to sit at the point reconcileCapture returns early.

## Verification

- `xcrun swift build -c debug --package-path Desktop --build-tests` — clean.
- `ShellListeningCycleTests` 14/14, `ShellStatusIconLegibilityTests` 13/13, `DashboardCaptureStateTests` 10/10, `MeetingGatedSystemAudioTests` — **43 tests, 0 failures**.
- **The regression test fails against the shipped behaviour.** Reverting `nextAudioRecordingMode` to the two-state flip fails 4 tests and reproduces the report verbatim:

  > the listening cycle ran off → onlyMeetings → off → onlyMeetings. It has to offer all three modes…

- `make preflight` — all checks pass.

### What is *not* verified, explicitly

The final code has **automated coverage only — it was not click-verified in a running build.** The full three-state cycle, the `person.2.fill` badge and the popover *were* exercised live (clicking the real control, reading `defaults` and screenshotting each state) on a build of the pre-rebase implementation, which carried the same badge, popover and cycle order against the older two-setting model. That evidence does not transfer to this diff.

The blocker was environmental, not the change: the summoned shell places itself on this machine's secondary display, synthesized clicks are not delivered there (a control-tab click failed too), and the panel auto-dismisses on any focus change, so it could not be relocated to the primary display. Worth a reviewer clicking the button once before merge.

## Notes for review

Three `ShellStatusTooltip.audio` call sites in `ShellStatusIconLegibilityTests` gained a `next:` argument. Two are mechanical — their assertions (`hasPrefix("Audio")`, `contains("Meetings only")`) are byte-identical. The third, `testTheAwaitingMeetingAudioTooltipDoesNotClaimOffOrStart`, changed one expected literal from `"Click to turn off"` to `"Click for Off"`: from Only Meetings a click still lands on Off, so the guard's intent is unchanged and its real assertion — `XCTAssertFalse(contains("Click to start"))` — is untouched. Only the copy it looks for moved.

### DIFF
--- a/desktop/macos/Desktop/Sources/AppState/AppState+MeetingGatePause.swift
+++ b/desktop/macos/Desktop/Sources/AppState/AppState+MeetingGatePause.swift
@@ -1,0 +1,37 @@
+import Foundation
+
+extension AppState {
+  /// Pause both capture paths without ending the session, for the window in which the meeting gate
+  /// has not answered yet. Mirrors the pause branches `reconcileCapture` uses once it does answer;
+  /// this is the same stop, taken earlier, when the honest answer is "not in a call as far as we
+  /// know".
+  /// Apply `MeetingGateReadinessPolicy` and pause when it says the gate cannot yet justify capture.
+  @MainActor
+  func pauseCaptureIfMeetingGateUnknown(
+    mode: AssistantSettings.AudioRecordingMode, meetingStateReady: Bool
+  ) {
+    guard
+      MeetingGateReadinessPolicy.shouldPauseCapture(
+        mode: mode, meetingStateReady: meetingStateReady)
+    else { return }
+    pauseCaptureWhileMeetingGateUnknown()
+  }
+
+  @MainActor
+  func pauseCaptureWhileMeetingGateUnknown() {
+    if let mic = audioCaptureService, mic.capturing {
+      mic.stopCapture()
+      AudioLevelMonitor.shared.updateMicrophoneLevel(0)
+      log("Transcription: Microphone capture paused (meeting state not yet known)")
+    }
+    if #available(macOS 14.4, *) {
+      if let systemService = systemAudioCaptureService as? SystemAudioCaptureService,
+        systemService.capturing
+      {
+        systemService.stopCapture()
+        AudioLevelMonitor.shared.updateSystemLevel(0)
+        log("Transcription: System audio capture paused (meeting state not yet known)")
+      }
+    }
+  }
+}
--- a/desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift
+++ b/desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift
@@ -616,6 +616,8 @@ extension AppState {
     isAwaitingMeeting = mode == .onlyMeetings && !meetingActive
 
     guard meetingStateReady else {
+      // Fail closed while the gate has not answered — see `pauseCaptureWhileMeetingGateUnknown`.
+      pauseCaptureIfMeetingGateUnknown(mode: mode, meetingStateReady: meetingStateReady)
       log("Transcription: waiting for meeting detector before changing capture state")
       return
     }
--- a/desktop/macos/Desktop/Sources/AppState/MeetingConversationBoundaryPolicy.swift
+++ b/desktop/macos/Desktop/Sources/AppState/MeetingConversationBoundaryPolicy.swift
@@ -1,5 +1,20 @@
 import Foundation
 
+/// Whether capture may continue while the meeting gate has not yet answered.
+///
+/// Only Meetings is a *closed* gate that a detected call opens, so "we do not know yet" has to be
+/// treated as "not in a call". Selecting Only Meetings from a live Always session builds a fresh
+/// detector, so its first reconcile pass runs with `hasObservedState == false`; leaving capture
+/// alone until the first probe lands would keep the microphone the previous mode opened running
+/// after the user asked for it to be closed.
+enum MeetingGateReadinessPolicy {
+  static func shouldPauseCapture(
+    mode: AssistantSettings.AudioRecordingMode, meetingStateReady: Bool
+  ) -> Bool {
+    mode == .onlyMeetings && !meetingStateReady
+  }
+}
+
 enum MeetingConversationBoundaryPolicy {
   typealias Role = TranscriptionConversationRole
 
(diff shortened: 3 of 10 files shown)


Git-on-my-level:
Maintainer-automation review (glm-5.3). I verified the full path in the diff — this is a clean fix. Notes per file:

**The fix — `CaptureListeningLogic.swift`:** the old `toggleListening` could only ever reach `.off` and `.onlyMeetings`, so `.always` (continuous recording) was unreachable from both the top bar and Home. `nextAudioRecordingMode` cycles Off → Always On → Only Meetings → Off, and both shells drive the one `cycleListening`, so the third mode lands everywhere at once. `ShellListeningCycleTests` pins the cycle as pure functions rather than view scraping — a good regression surface, including the "cycle must cover every declared mode" guard for future fourth modes.

**The fail-closed follow-up — `AppState+MeetingGatePause.swift` plus the not-ready branch of `reconcileCapture()` in `AppState+Transcription.swift`:** verified the wiring end-to-end. Switching Always → Only Meetings mid-session posts `audioRecordingModeDidChange`; the observer in `AppState.swift` calls `reconcileCapture()`; `ensureMeetingDetector` rebuilds the detector for the new mode (fresh `hasObservedState == false`); and the new `MeetingGateReadinessPolicy` branch now stops both mic and system audio before returning early — instead of leaving the previously-open mic running until the detector's first async probe lands. Correct semantics for a gate the user selected in order to close the mic: "not known yet" means "not in a call". This closes the state-transition gap flagged earlier on the first commit.

**Status cluster — `ShellStatusIcons.swift`:** the `person.2.fill` badge rides outside the `mic` silhouette (additive corner ink like the state dot, not a glyph swap), so the pixel-level silhouette guard in `ShellStatusIconLegibilityTests` keeps measuring what it was written to measure; `modeBadge(for:)` states the exactly-one-badged-mode rule as a testable function. The tooltip now names the destination ("Click for X") instead of the two-state-era "Click to stop", and correctly declines to promise a next mode while the microphone grant is missing. The Only-Meetings popover fires on the transition only, keeping the resting cluster wordless.

**Smaller pieces:** `DashboardPage.swift` is the one-line rebind to `cycleListening`; `MeetingConversationBoundaryPolicy.swift` hosts `MeetingGateReadinessPolicy` beside the boundary policy it complements; the `changelog/unreleased/20260815-listening-mode-cycle.json` entry matches the sibling schema and is accurate; `e2e/flows/audio-recording.yaml` adds the new file to the coverage map.

Two non-blocking observations for a maintainer:
1. Product: the cycle order means the first click from Off now selects **Always On** (continuous capture), where the old button armed Only Meetings. I think that's the right default — turning the mic on should record — and the tests state that intent, but it is a user-facing behavior change worth a quick product gut-check.
2. `AppState.toggleTranscription()` (floating control bar play/pause and the RewindPage pill) still uses the two-state off↔onlyMeetings flip. Pre-existing and out of scope here, but those two surfaces can now disagree with the cycle until they're migrated — worth a small follow-up issue.

Checks are green (Desktop Swift Build and Tests, Release Compile, Static and Test Contracts, desktop-core-e2e-t0). Leaving for human maintainer review before merge.

---

PR #11642, opened 2026-08-15
fix(admin): source macOS activation from conversation records, not telemetry

Archit-lal:
## What

The macOS Activation KPI on `admin.omi.me` read **0.2%**. The real figure, measured from production conversation records, is **36.5%** — and rising from 20.2% (week of Jun 29) to 57.1% (week of Aug 3).

## Why it was wrong

The metric counted the desktop `Memory Created` event. Builds before `98f1ee7c7f` (2026-08-11) never emit it on the normal durable-session path, and stable `0.12.144` — ~66% of the macOS fleet — produced **0** events across 454 recording users in three days. The metric was reporting build age, not behaviour.

Three independent query defects compounded it, each only ever pushing the number down:

1. **Cohort** — the signup subquery filtered by `timestamp` *before* `min()`, so "first sign-in" meant first sign-in *inside the window*. 237 of 1,810 were returning users re-authenticating, counted as new signups.
2. **Numerator** — the join keyed off `min(timestamp)` of `Memory Created`, so a user whose first-ever memory predated their window was marked unactivated even when they created one inside it.
3. **Maturation** — every signup was pooled, including 167 that had not yet had seven days to activate. The weekly chart already excluded these; the headline did not.

## What changed

No amount of query-fixing gets a real number out of a source two thirds of the fleet cannot write to, so activation moves off telemetry. A conversation document exists whether or not the client managed to report it, so the new route reads `users/{uid}/conversations` directly: a macOS signup is activated if a conversation was created inside its first 7 days.

Two traps in the cohort query, both covered by tests:

- `created_at` is **null** on user documents written by the current signup path; `signup_platform_at` is the field that is actually populated. Cohorting on `created_at` returns 35 users instead of 10,375.
- Filtering on `signup_os` alongside a `signup_platform_at` range needs a composite index that does not exist on this project, so the range is the query and the platform is filtered in memory.

Fanning out one aggregation query per signup is too slow for a request, so it runs in the existing cron precompute endpoint (`maxDuration` 3600) and the GET is a payload-cache read — the same shape as `profitability` and `infra-costs`.

The PostHog-derived series is kept as a fallback for the window before the activation route has a cached payload, with its three defects fixed and its telemetry coverage now explicit, so it reports "not measurable" rather than a confident near-zero.

## Verification

- `npx tsc --noEmit` — clean
- `npm test` (web/admin) — 15 files, **101 tests passing**, 10 new
- `npx eslint` on every changed file — 0 errors
- `make preflight` — 14/14 checks pass
- New HogQL validated directly against PostHog project 302298
- Ran the viral-metrics route locally (`next dev`, dev-auth bypass) against live PostHog: `GET /api/omi/stats/viral-metrics?days=60` → 200
- Loaded `/dashboard/classic` in a browser and confirmed the tiles render
- The activation algorithm was validated end-to-end against **production Firestore** before being written in TypeScript: 1,456 matured macOS signups over 60 days, 532 activated, 0 read errors

New behavioural tests for `computeActivation` run against a faked Firestore and cover the 7-day boundary, the macOS filter, maturity exclusion, 500-doc pagination, and per-user read failures being reported rather than silently scored as unactivated.

### Not verified

The TypeScript route has **not** been executed against Firestore — this machine has no `FIREBASE_CLIENT_EMAIL` / `FIREBASE_PRIVATE_KEY`, which `lib/firebase/admin.ts` requires. Its first real exercise will be the cron run.

## Not addressed here

Both need a decision rather than a patch:

- Stable is still on a build that cannot report activation via telemetry.
- Cloud Run service `backend` has `POSTHOG_HOST` but no `POSTHOG_PROJECT_API_KEY`, so every backend-side product event is silently dropped — this is also why `MCP Tool Call` does not exist in PostHog.

Failure-Class: FC-unestablished-capability-default

### DIFF
--- a/.github/failure-classes/FC-unestablished-capability-default.json
+++ b/.github/failure-classes/FC-unestablished-capability-default.json
@@ -3,6 +3,10 @@
   "id": "FC-unestablished-capability-default",
   "violated_contract": "A surface that selects its presentation from a rollout or account capability must treat \"the capability is not established yet\" as its own state. A capability carried as a plain Bool cannot express it, so the default value is silently rendered as an authoritative answer and the wrong surface mounts until the real answer arrives.",
   "canonical_prevention": "Publish establishment alongside the capability and resolve presentation to an explicit undetermined state that mounts neither branch. Independently, any branch that performs destructive or shared-state work on first appearance must require a load that authoritatively succeeded, never an initial-value sentinel such as isEmpty.",
+  "canonical_prevention_artifact": [
+    "web/admin/lib/growth-metrics.ts",
+    "web/admin/lib/__tests__/growth-metrics.test.ts"
+  ],
   "evidence_prs": [9499],
   "scope_hints": [
     "desktop/macos/Desktop/Sources/MainWindow/Pages/MemoryGraph/MemoryGraphPage.swift",
--- a/web/admin/app/(protected)/dashboard/classic/page.tsx
+++ b/web/admin/app/(protected)/dashboard/classic/page.tsx
@@ -1713,7 +1713,7 @@ export default function AnalyticsPage() {
         id: "viral-activation",
         title: "New Activated Users / Week",
         periodChange: latestPeriodChange(weeklyActivationData, (point) => point.activated, "vs previous mature cohort week"),
-        subtitle: "Memory created within 7 days of signup · fully matured cohorts only",
+        subtitle: "Conversation created within 7 days of signup · fully matured cohorts only",
         icon: <Target className="h-4 w-4" />,
         initialLayout: { cols: 12, rows: 4 },
         render: () => (
--- a/web/admin/app/(protected)/dashboard/classic/page.tsx
+++ b/web/admin/app/(protected)/dashboard/classic/page.tsx
@@ -228,6 +228,9 @@ interface ViralMetrics {
   summary: {
     quickRatio: number | null;
     activationRate: number | null;
+    activationTelemetryCoverage: number | null;
+    activationSignups: number;
+    activationPooledRate: number | null;
     dauMau: number;
     dauWau: number;
     dau: number;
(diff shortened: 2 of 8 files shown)


Git-on-my-level:
Verified this end-to-end in a worktree at d5e502f: `npx tsc --noEmit` clean and `npx vitest run` 103/103 passing (web/admin), matching the commit's stated evidence. The three query defects the PR description claims to fix are all visible in the diff, and each of the earlier bot findings checks out as resolved on this head. Per-file notes:

- `web/admin/app/api/omi/stats/activation/route.ts` (new) — the core of the PR. The cohort paginates `users` on a `signup_platform_at` range with `signup_os === "macos"` filtered client-side (the missing-composite-index constraint is documented inline); signups inside the 7-day maturity window are excluded rather than scored as guaranteed zeros; per-user conversation `count()` runs 16-wide; and read failures stay `null` so unreadable users surface via `erroredUsers` instead of depressing the rate. The comment explaining why `null` must not collapse to `false` is exactly right. `days` is clamped to 1..180.
- `web/admin/lib/__tests__/activation-route.test.ts` (new) — the fake applies the route's own `where()` predicates and rejects filters not on `created_at`, so the 7-day window is a real contract rather than a mirror of the seed data; the boundary test pins the window closed at both ends, and the pagination test drives past the 500-doc page.
- `web/admin/lib/growth-metrics.ts` — `summarizeActivation` pools only matured days; `capableRate`/`telemetryCoverage` make "nobody could report" read as unmeasurable rather than zero; legacy series without capable counts default to fully-capable rather than zero-coverage. Sound.
- `web/admin/lib/__tests__/growth-metrics.test.ts` — good new cases for the unmeasurable week, the maturity exclusion, and the day-the-window-closes boundary.
- `web/admin/app/api/omi/stats/viral-metrics/route.ts` — the rewritten HogQL moves the window filter outside the `min(timestamp)` subquery (true first-ever sign-in instead of first-inside-window), keys activation off any memory inside the window instead of the user's earliest, and tags capability via `$app_version` at first sign-in vs 0.12.167. Each of the three old defects is now structurally hard to reintroduce.
- `web/admin/app/(protected)/dashboard/classic/page.tsx` — the headline tile keys on `activationStats.rate` with `n=` and an amber unreadable-users note; the weekly chart prefers Firestore weeks and falls back to telemetry only before the payload loads (`activationStats ?` — payload presence, not `weeks.length`, which was the earlier empty-cohort bug); `macosKpis` now lists `activationStats` in its deps; the chart subtitles now say "Conversation created".
- `web/admin/app/api/internal/precompute/route.ts` — warms `activation:v1:macos:60` behind the existing `run()` helper; `ACTIVATION_DAYS = 60` matches the dashboard request, so the request path should normally hit the cache.
- `.github/failure-classes/FC-unestablished-capability-default.json` — `canonical_prevention_artifact` is a valid registry field (10 other FC files carry it) and points at real files in this PR.

Cross-checked the `created_at`-is-null claim against `backend/database/users.py:184-189` — `signup_platform_at` is indeed the field the signup path populates. Declining the `app_version` fallback is reasonable given the 2,294/2,294 `$app_version` coverage cited in aa3bcae, and the telemetry series is only the pre-cache fallback now.

Two non-blocking notes: (1) a cold-cache GET can still trigger the full per-user fan-out on the request path (`maxDuration = 3600`) — fine while precompute warms it, just worth watching first-request latency; (2) the signups subquery now scans all-time macOS `Sign In Completed` events to get the true first sign-in — correct, just a heavier HogQL query behind the 30-minute cache.

This redefines a headline admin KPI (the dashboard will show ~36.5% where it showed 0.2%), so leaving formal merge sign-off to a human maintainer who can confirm the new methodology is the number they want on the dashboard.

---

PR #11651, opened 2026-08-16
Stop desktop tests deciding a day in one zone and asserting it in another

Git-on-my-level:
## Summary

Three desktop tests decided a day in one time zone or at one instant and asserted it in another, so their result depended on when and where they ran. One of them blocked a desktop release tonight.

`RewindAllTimeSearchTests.testCapturedDayStartsReportsEveryDayThatHoldsCaptureNewestFirst` seeded a frame at the live time-of-day shifted back N days, plus a duplicate at that stamp `+120s` — same-day only if the run misses the last two minutes of a day. The CI job started 23:40Z on 2026-08-15 and hit that window: each duplicate crossed midnight and reported a day one newer than the one it was meant to duplicate, so three seeded days came back as five.

Replaying the old seeding arithmetic over every minute of a day in four zones fails at 23:58 and 23:59 (and 22:58 on a spring-forward day). The new arithmetic fails at none of 17,280 sampled minutes.

## Which layer

The walk was already correct: `capturedDayStarts(calendar:)` takes its calendar and buckets with `Calendar` rather than in SQL deliberately. The tests never exercised that injection point, so they bucketed in whatever zone the machine had. They now derive every instant from one reading of the clock in one pinned zone, and seed a day's first and last instants rather than a stamp and a stamp `+120s` — which also drives the day-edge case the old fixture could not reach.

Two production paths did have a real defect of the same shape — they took a `calendar` and honoured it for *deciding* a day but not for *naming* one:

- `RewindHistoryReach.spanLabel` set `formatter.calendar` but not `formatter.timeZone`, which does not follow it.
- `ChatMessageTimestampFormat` decided day-ness with the injected calendar while `Date.FormatStyle` rendered against `.autoupdatingCurrent`.

Every shipping caller passes `.current`, so no user sees this today — but `SpineStore` already passes its own calendar to the day walk, and the label would then name a day either side of the one it points at. Reverting the one-line fix now fails three assertions with "Jul 1 – Aug 4" where the days say "Jul 2 – Aug 5".

- INV-CHAT-1

Cited on paths: `ChatBubbleSupport.swift` is matched by the invariant's globs. The change is confined to how an already-rendered message's timestamp is worded — no transcript store, session identity, continuity ring, or ordering behavior is touched, so the invariant's guard behavior is unchanged.

## Failure class (fixes)

Failure-Class: FC-day-bucket-key-recomputed

## Verification

- `xcrun swift test --package-path Desktop` — 5329 tests, 1 skipped, 0 failures, exit 0.
- The three suites also pass under `TZ=UTC`, `Asia/Tokyo`, `Pacific/Kiritimati`, `America/Los_Angeles`.
- Negative control: reverting `formatter.timeZone = calendar.timeZone` fails 3 assertions in `RewindHistoryReachTests`.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11651?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubbleSupport.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubbleSupport.swift
@@ -44,13 +44,30 @@ struct ChatMessageTimestamp: View {
 /// the message and started reading as page furniture. Today says the time; this
 /// year adds the day; only another year is worth naming.
 enum ChatMessageTimestampFormat {
-  static func text(for date: Date, now: Date = Date(), calendar: Calendar = .current) -> String {
-    let time = date.formatted(.dateTime.hour().minute())
+  /// - Parameters:
+  ///   - calendar: decides *and* renders. Which day a message belongs to and which day the stamp
+  ///     says have to be the same question: `Date.FormatStyle` otherwise resolves against
+  ///     `.autoupdatingCurrent`, so a caller passing any other calendar would get "today" decided in
+  ///     one zone and the clock time printed in another — a 1:28 PM stamp on a row the same call
+  ///     just decided was yesterday.
+  ///   - locale: how the stamp is worded; the user's, so the month reads in their language.
+  static func text(
+    for date: Date, now: Date = Date(), calendar: Calendar = .current, locale: Locale = .current
+  ) -> String {
+    func render(_ style: Date.FormatStyle) -> String {
+      var style = style
+      style.calendar = calendar
+      style.timeZone = calendar.timeZone
+      style.locale = locale
+      return date.formatted(style)
+    }
+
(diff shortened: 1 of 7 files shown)


Git-on-my-level:
Addressed the changelog wording feedback: the unreleased fragment now states the user-visible outcome ("always show the date of the day they point at, even around midnight and across time zones") without describing the internal bucketing mechanism. Pushed as an amend to the PR head; the desktop changelog gate passes locally.

PR #11652, opened 2026-08-16
fix(app): four iOS startup fixes — CGNAT dev endpoints, and hangs that presented as a blank splash

formed2forge:
Four fixes to iOS startup, found while getting a community build to run against the local dev harness on real hardware. They are independent of each other and of the sign-in work that surfaced them.

## 1. `local_dev` rejected Tailscale's address range

`Env._isLocalDevelopmentApi` allowed loopback and RFC 1918 only. Tailscale assigns from **100.64.0.0/10** (RFC 6598 shared address space), which is neither, so `validateApplicationStartupRouting()` threw before any network call:

```
StateError: Profile local_dev requires a loopback or private-network API endpoint;
use mobile_beta for https://api.omiapi.com/.
```

This is not an exotic setup. The dev harness binds loopback only *by design* (`scripts/dev-harness/dev_harness/safety.py:194` is a fail-closed guard holding real provider keys), so a physical device cannot reach it on `127.x` and cannot reach it over the LAN either. A tailnet address is in practice the only route from real hardware to a developer's harness.

Measured on iPhone 17 Pro / iOS 27.0, one variable at a time:

| `OMI_API_BASE_URL` | Reachable? | Result |
|---|---|---|
| loopback default | no | reaches sign-in |
| `http://100.105.2.5:8000/` | yes, responds | splash hang |
| `http://100.105.2.5:9999/` | nothing listening | splash hang |

The dead-port run is the discriminator: no responding server is required, which is what pointed at a validation guard rather than a connection. Bounded to the real /10 — tests assert `100.63.x` and `100.128.x` stay rejected so this cannot decay into "any 100.x host".

## 2. A startup throw showed a blank splash screen

`_init()` runs inside `runZonedGuarded`, whose handler logs with `debugPrint` and nothing else — and `debugPrint` is invisible in profile and release builds. Any throw before `runApp()` left the launch storyboard up indefinitely: no UI, no crash, no message.

The failure mode was worse than the failure it hid. Fix 1 above threw a precise, actionable `StateError` naming the exact misconfiguration; presenting that as a frozen splash is what made it expensive — five connectivity hypotheses were investigated on device before anyone read the guard.

`StartupFailureApp` is deliberately dependency-free: no providers, services, theme lookups or localisation, since all of those are precisely what may have just failed to initialise. Message is selectable because the situation it exists for is a device with no debugger attached.

## 3. Startup awaited the auth refresh with no timeout

`_init()` awaited `getIdToken()` before `runApp()`. Measured on the same device against a Firebase Auth emulator on a non-loopback host, FirebaseAuth's forced refresh **never returned at all** — not an error, just silence. File-backed probes recorded entry to `_refreshIdTokenWithRetries` with every guard passing, then no further line. The app could not be launched again until it was deleted, because the cached session made every start hang there.

Note fix 2 does not help here: it catches throws, and this is a hang.

## 4. The forced refresh could freeze every request app-wide

`backend/http/shared.dart:73` refreshes on the way into **every** authenticated request, so the same stall took the whole app offline. Observed end to end: the app signed in, ran onboarding normally on its cached token — a clean sequence of 200s — then went permanently silent the moment a refresh was first required. Tapping a button did nothing at all, because neither the success nor the failure path ran.

Each attempt is now bounded, and a stall is reported as a transient failure — which the surrounding retry loop and every existing caller already handle, since `getIdToken()` returns null on four failure branches today. So this only makes a hang behave like the failure it effectively is; it never fails a refresh the old code would have completed. `refresh_timeout` is a distinct failure class so a stall stays distinguishable in telemetry from a genuine failure.

**This does not fix the underlying stall.** Why FirebaseAuth's forced refresh never returns against the Auth emulator on a non-loopback host is still unexplained — the SDK does route it through the emulator (`SecureTokenRequest.swift:129-130` builds exactly the URL that answers instantly to `curl`), the device reaches the public internet, and ATS permits the plain-http call that sign-in already makes to the same host and port. This makes that failure survivable rather than fatal.

## Verification

```
flutter test test/unit  ->  772 passed
```

11 new tests. The CGNAT case fails without fix 1 with exactly the `StateError` quoted above. The hang guards supply futures that never complete and assert startup and `refreshIdToken` both resolve. Slow-but-successful refreshes inside the budget are asserted unaffected, so neither bound can sign a user out on a slow network.

Also exercised on device throughout: iPhone 17 Pro, iOS 27.0, profile build, against a local harness reached over Tailscale.

One implementation note worth flagging for review: the bounds are written as races rather than `Future.timeout(onTimeout:)`. `onTimeout` must return the *concrete* future's type, so a callback producing `Future<String>` rather than `Future<String?>` makes `() => null` throw a `TypeError` at runtime — turning a hang-guard into a new startup crash. A test caught exactly that during development and now pins it.

## INV-DATA-1

`app/lib/env/env.dart`, `app/lib/main.dart` and `app/lib/startup_routing.dart` are path globs of `INV-DATA-1`. Production-family routing is unchanged: production keeps its own exact-match branch, fix 1 widens only the `local_dev` branch, and fixes 2-4 change no routing at all. The invariant's MUST NOT clauses all concern production-family endpoints.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11652?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/ios/Runner/Info-Dev.plist
+++ b/app/ios/Runner/Info-Dev.plist
@@ -131,7 +131,15 @@
 	</dict>
 	<key>NSAppTransportSecurity</key>
 	<dict>
-		<key>NSAllowsLocalNetworking</key>
+		<!-- Dev-only plist. NSAllowsArbitraryLoads rather than NSAllowsLocalNetworking,
+		     because the latter covers only .local, unqualified hostnames and link-local
+		     addresses — never a Tailscale CGNAT address (100.64.0.0/10), which is what
+		     the local dev harness is reached on.
+
+		     These two keys must not be combined: declaring NSAllowsArbitraryLoads
+		     alongside NSAllowsLocalNetworking makes the former a documented no-op on any
+		     OS that understands the newer key, silently leaving CGNAT hosts blocked. -->
+		<key>NSAllowsArbitraryLoads</key>
 		<true/>
 	</dict>
 	<key>NSBluetoothAlwaysUsageDescription</key>
--- a/app/lib/env/env.dart
+++ b/app/lib/env/env.dart
@@ -184,6 +184,12 @@ abstract class Env {
     return first == 10 ||
         (first == 172 && second >= 16 && second <= 31) ||
         (first == 192 && second == 168) ||
+        // 100.64.0.0/10 — RFC 6598 shared address space, the range Tailscale
+        // assigns. Included because a physical device has no other route to a
+        // developer's local harness: the harness binds loopback only by design,
+        // so the device cannot use 127.x, and a plain LAN address does not reach
+        // it either. Bounded to the real /10 — 100.63.x and 100.128.x are public.
+        (first == 100 && second >= 64 && second <= 127) ||
         (first == 127);
   }
 
--- a/app/lib/main.dart
+++ b/app/lib/main.dart
@@ -36,6 +36,7 @@ import 'package:omi/env/prod_env.dart';
 import 'package:omi/firebase_options_local.dart' as local;
 import 'package:omi/firebase_options_prod.dart' as prod;
 import 'package:omi/flavors.dart';
+import 'package:omi/startup_failure_app.dart';
 import 'package:omi/startup_routing.dart';
 import 'package:omi/l10n/app_localizations.dart';
 import 'package:omi/pages/apps/providers/add_app_provider.dart';
@@ -228,7 +229,21 @@ void main() {
       } else {
         WidgetsFlutterBinding.ensureInitialized();
       }
-      await _init();
+      try {
+        await _init();
+      } catch (error, stack) {
+        // Startup failed before the first frame. Without this the launch
+        // storyboard stays on screen forever: runApp() is never reached, and the
+        // zone handler below only calls debugPrint, which goes nowhere in
+        // profile/release builds. A misconfigured OMI_API_BASE_URL cost about a
+        // day of investigation for exactly this reason — the app looked hung
+        // when it had in fact thrown a precise, actionable StateError.
+        if (Firebase.apps.isNotEmpty) {
+          FirebaseCrashlytics.instance.recordError(error, stack, fatal: true);
+        }
+        runApp(StartupFailureApp(error: error, stack: stack));
+        return;
+      }
       runApp(const MyApp());
     },
     (error, stack) {
--- a/app/lib/main.dart
+++ b/app/lib/main.dart
@@ -36,6 +36,7 @@ import 'package:omi/env/prod_env.dart';
 import 'package:omi/firebase_options_local.dart' as local;
 import 'package:omi/firebase_options_prod.dart' as prod;
 import 'package:omi/flavors.dart';
+import 'package:omi/startup_auth.dart';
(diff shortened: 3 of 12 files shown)


kodjima33 (approved):
All four startup fixes are real: CGNAT 100.64/10 is legitimately private, and the blank-splash hangs are genuine. Holding the merge — Dart Analyze fails the ratchet with a new invalid_use_of_visible_for_testing_member (baseline 0).

formed2forge:
Pushed a follow-up commit to this PR: the CGNAT/ATS fix a couple commits back in this same PR (`db65122f0a`, "let the dev build reach a CGNAT dev host over plain HTTP") turned out to be silently undone on every `setup.sh ios` run.

`app/scripts/generate_ios_dev_info_plist.sh` regenerates `Info-Dev.plist` from `Info.plist` from scratch on each run and unconditionally added `NSAllowsLocalNetworking` — discarding the `NSAllowsArbitraryLoads` fix committed earlier in this same PR. `NSAllowsLocalNetworking` only covers `.local`/unqualified/link-local hosts, never a Tailscale CGNAT address (100.64.0.0/10), which is exactly the address this PR's own fix is for.

Caught live, not just by reading the code: built and ran on a physical iPhone over a real Tailscale path after merging this branch with a couple others — every `NSURLSession` call failed with `CFNetwork Code=-1022` (the ATS-block signature from #11730) despite this PR's fix being present in the source tree. Restoring the committed `Info-Dev.plist` with no other change, then rebuilding, gave zero `-1022` errors and a real `200` from the harness — confirming both that the fix is correct and that the generator script was the thing undoing it.

Filed as #11782, fixed here: `generate_ios_dev_info_plist.sh` now writes `NSAllowsArbitraryLoads`, with a comment tying it to #11730/#11652/#11782 so it can't silently regress again, plus a new hermetic shell test (`app/test/shell/ios_dev_ats_config_test.sh`) that fails against the old script and passes against the fix.

formed2forge:
All CI checks are green and this is ready for maintainer review/merge — I don't have write access to merge it myself.

Summary of what's in this PR as of the latest push: the CGNAT/ATS fix (`NSAllowsArbitraryLoads` for reaching a local-dev harness over Tailscale's CGNAT range), plus two follow-up fixes found while verifying it end-to-end on a real device: #11782 (the iOS dev Info.plist generator was silently reverting the ATS fix on every `setup.sh ios` run) and an unrelated pre-existing `@visibleForTesting` misuse that was failing the Dart analyzer ratchet.

kodjima33 (approved):
Four independently root-caused iOS startup defects (CGNAT dev endpoint rejected, silent blank splash on startup throw, unbounded auth refresh) with unit tests for each; dev-flavor plist only, prod signing/ATS untouched.

PR #11664, opened 2026-08-16
fix(gateway): price GPT-5.6 long-context requests at the provider's long-context rate

Git-on-my-level:
## Summary

`cost_rate_cards.yaml` carried only the short-context price band for `gpt-5.6-luna`, and no card at
all for `gpt-5.6-sol` / `gpt-5.6-terra`. OpenAI publishes a second, more expensive band for the
GPT-5.6 family once a request's input context crosses 272,000 tokens. Every request over that
threshold was therefore priced at roughly half its real cost, and the two unlisted models were
recorded as `rate_card_missing` — unpriced rather than mispriced.

This adds an optional long-context tier to `RateCard` and the cards for the missing models.

## Which layer

The tier is selected in `_estimate_cost`, from **total request context** — cached + uncached +
cache-write input tokens — never from output tokens, which do not affect which band applies.

The sum is built from the split usage fields rather than `usage.prompt_tokens` because that field's
relationship to cache-write tokens is provider-specific: OpenAI's raw `prompt_tokens` already
includes cache-write tokens, while Anthropic's `input_tokens` excludes
`cache_creation_input_tokens`. Reading the aggregate would have under-counted Anthropic requests
near the boundary and double-counted nothing on OpenAI — a silent, provider-dependent error.

A card with no `long_context_threshold_tokens` has a single tier, and `effective_rates` returns it
for any token count, so every existing card is unaffected.

None. This changes recorded cost attribution only — no routing, admission, quota, or user-visible
behavior. Pricing is read after a request completes.

## Verification

- `pytest backend/tests/unit/test_llm_gateway_accounting.py backend/tests/unit/test_llm_gateway_deploy_contract.py` — 30 passed, exit 0.
- Boundary cases covered in tests: exactly at the threshold (short band), one token over (long
  band), and a card without a long-context tier (unchanged).
- Rebased onto `origin/main` — the deploy-contract test asserts against `.github/workflows/*` and
  failed only because this branch's `.github` had diverged. It passes on the rebased tip.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11664?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/llm_gateway/config/cost_rate_cards.yaml
+++ b/backend/llm_gateway/config/cost_rate_cards.yaml
@@ -4,6 +4,16 @@
 # The ledger deliberately marks models without a card as `unpriced`, rather
 # than pretending their COGS are zero. Update cards from the provider's public
 # pricing page and create a new ID when a rate changes.
+#
+# Long-context tier: some providers publish a second, pricier band once a
+# request's total input context (cached + uncached + cache-write tokens)
+# crosses a threshold; the higher rates then apply to the whole request, not
+# just the tokens past the boundary. A card represents this with the
+# `long_context_*` fields alongside its normal (short-context) fields; a card
+# with none of those fields is single-tier and unaffected. This is additive
+# schema, not a rate change to the short-context fields, so it does not by
+# itself require a new rate_card_id — see the OpenAI GPT-5.6 cards below for
+# the reasoning on why their existing ID was kept.
 rate_cards:
   - rate_card_id: openai.gpt-5-nano.2026-07-17
     provider: openai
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Addressed the two pricing-feedback threads; kept the rate-card ID decision as designed.

**Incomplete long-context tiers now fail at load** (`accounting.py`): a card declaring a partial long-context tier (threshold or any long input/cached-input/output rate missing) is rejected at load instead of letting `effective_rates` bill omitted components at $0 past the threshold. Long cache-write rates remain optional — a long-context attempt that writes cache with no long write rate is recorded as unpriced via the existing `cache_write_rate_missing_or_mixed_ttl` guard, consistent with how the ledger treats every other missing write rate.

**Cost report prices the long band** (`chat_agent_cost_report.py`): the report now reads the `long_context_*` fields and buckets attempts whose recorded total input context crossed the threshold into their own components, priced at the long rates. Tier selection mirrors `_estimate_cost` (cached + uncached + cache-write input tokens, never output), so the modeled breakdown no longer understates long-context attempts relative to the ledger total, and the report surfaces how many attempts were billed at the long band.

**On publishing a new rate-card ID for Luna** — keeping the existing `openai.gpt-5.6-luna.2026-07-30` ID is the deliberate choice documented in the PR body and the card file. The card-file rule is "new ID when a rate changes," and the provider's rates on 2026-07-30 did not change: the published price list already carried both bands while the card encoded only the short one. Every >272K request since that date has been recorded at roughly half its real cost; minting a new ID would preserve that under-recording as the old ID's "correct" interpretation and leave a cutover window pricing at the wrong band. Short-tier pricing is byte-identical, so the only behavioral delta is that >272K requests now record what the provider actually charged on that date.

Verification: `test_llm_gateway_accounting.py`, `test_chat_agent_cost_report.py`, `test_llm_gateway_deploy_contract.py` through `backend/test.sh` — 46 passed.

kodjima33 (approved):
Correct layer for the long-context tier (total request context, not output tokens) and the missing sol/terra cards are a real gap. Not merging: branch has conflicts with main and the diff is past the scoped-fix size.

PR #11666, opened 2026-08-16
fix(desktop): attest a workstream label against the bucket it labels

Git-on-my-level:
## Summary

Two scope defects in the workstream reconciler, both found by reading live dogfood rows rather
than by a failing test.

**1. A label was attested against the wrong text.** `acceptedAssignments` validated a proposed
label by searching the concatenated observations of *every group in the batch*. Any term that
appeared anywhere in the batch therefore satisfied the check for every bucket in it. Measured on
live data: bucket `1bfc86de-c6e5-4aaf-91ad-3b40fce94e2d` carries 193 facts, none of which contain
the string `stably`, and it was tagged `stably` — a label earned by a different bucket in the same
batch.

The check now runs against the tagged bucket's own facts, threaded through as
`observationsByBucket`. A bucket that cannot support a label keeps none, so `null` remains a
correct and expected outcome rather than the bucket borrowing a neighbour's label.

**2. An expired candidate could be reported as armed.** `state` stays `'armed'` in the column
until a sweep runs, so a row whose `expiresAt` had passed was still handed to the gate as live.
`lookupArmed` now runs `expireStale` in the same write and filters on `effectiveState(at:)`.

## Which layer

The attestation fix is in the acceptance step, not the prompt: the model may propose whatever it
likes, and the client decides what is supportable. Doing it here keeps `null` cheap and means a
weaker model cannot widen label scope.

The expiry fix deliberately changes `lookupArmed` from a read to a write so the sweep and the
lookup cannot interleave. The alternative — filtering in the caller — leaves the stale row in the
table to be re-read by the next caller.

## What is deliberately NOT in this PR

The branch originally also defined `notify_worthiness` in the extraction prompt. That change is
**dropped**, because measurement did not support it. Scoring 54 real statements shown with their
on-screen evidence, 7 runs per arm through the `proactive_extraction` lane:

| arm | AUC | overall mean |
|---|---|---|
| bare field (ships today) | 0.546 (sd 0.059) | 0.257 |
| with the proposed definition | 0.580 (sd 0.074) | 0.383 |

The ranking gain is inside run-to-run noise, while the mean rises 49%. Four gates key on absolute
thresholds of this field (0.3 and 0.6), so shipping it would arm materially more candidates
without ranking them any better. It needs a different approach, not a prose definition.

None. Labelling and candidate liveness only; no delivery, transcript, or session behavior changes.

## Verification

- `xcrun swift test --package-path Desktop --filter 'Context'` — 376 tests, 0 failures, exit 0.
- New reconciler tests cover a label supported by one bucket and not another in the same batch,
  and a candidate whose `expiresAt` has passed.

## Failure class (fixes)

Failure-Class: FC-attestation-scoped-to-batch-not-item

Defined in the registry-only PR #11665, which must merge first.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11666?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextBucketStore.swift
@@ -557,8 +557,9 @@ actor ContextBucketStore {
     let (pool, _) = await RewindDatabase.shared.getDatabaseQueueWithGeneration()
     guard let pool else { return [] }
     return
-      (try? await pool.read { db in
-        try ContextProactiveCandidateLookup.lookupArmed(
+      (try? await pool.write { db in
+        try ContextProactiveCandidateLookup.expireStale(now: now, in: db)
+        return try ContextProactiveCandidateLookup.lookupArmed(
           bucketID: bucketID, tags: tags, now: now, in: db)
       }) ?? []
   }
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextWorkstreamReconciler.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextWorkstreamReconciler.swift
@@ -316,7 +316,8 @@ enum ContextWorkstreamTagging {
     }
     var counts: [String: Int] = [:]
     for bucketID in bucketOrder {
-      counts[proposalsByBucket[bucketID]!.first!, default: 0] += 1
+      guard let firstProposal = proposalsByBucket[bucketID]?.first else { continue }
+      counts[firstProposal, default: 0] += 1
     }
 
     func isAllowed(_ tag: String, bucketID: String) -> Bool {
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Monitor follow-up for the failing desktop checks on `da70fac8`:

- **`desktop-swiftlint` manifest check** — the proposal-collection rewrite introduced three `force_unwrapping` violations in `ContextWorkstreamReconciler.swift` (lines 319×2, 336). Fixed in 76c516b by binding the same structurally-guaranteed values with `guard let` / optional chaining; behaviour unchanged (every `bucketOrder` entry is inserted alongside its first proposal, so the guards never fire). The `Run macOS-scoped manifest checks` step is green on the new head.
- **`TasksStoreEmptyCloudReconcileTests.testAutoRefreshConvergesStaleListWhenCloudEmptiedElsewhere`** — did not reproduce locally: the full suite passes 14/14 on this branch, and the PR touches no TasksStore code. Classified as a CI-only flake, not addressed here.
- Branch rebased onto current `main` (`14e7eed`) and force-pushed to `fix/workstream-label-attestation`.
- Local verification: `swift test --filter ContextWorkstreamReconcilerTests` — 23 passed, 0 failures.

kodjima33 (approved):
Workstream labels are now attested against their own bucket's facts, and expired candidates no longer report as armed. Scoped fix with tests.

PR #11673, opened 2026-08-16
Refresh the Codemagic fixture digests broken by the Flutter 3.44.5 bump (unblocks Hygiene on every PR)

ZachL111:
## Summary

Main's Hygiene lane is red for every PR right now: #11672 (92a2ff1172) bumped the eight Codemagic Flutter pins to 3.44.5 without refreshing the reviewed digest fixture, and the release-process guard is built to fail exactly then. This PR is the two-line fixture refresh that re-approves the current document.

This is the same unblock class as #11629 (a desktop route landed without its spec entry and broke an inventory guard for every open PR).

## What changed and how it was computed

Only `.github/scripts/fixtures/codemagic_workflow_contract/v1.json`, two values. Both digests were recomputed with the check script's own functions, not by hand: `_read_security_bound_file` + `_sha256_bytes` for the raw digest, and `_sha256_text(_canonical_json(document))` over the safe-loaded document for the semantic digest, against `codemagic.yaml` exactly as merged at 94487a41fb.

- codemagic_raw_sha256: 941fd758... to 826c7aba...
- codemagic_semantic_sha256: a2d952c0... to a72c79d0...

## Review surface

The fixture is a reviewed security boundary, so what matters is what these digests newly approve: precisely the 16-line version-pin diff of 92a2ff1172 (eight `flutter:` pins to 3.44.5), which was itself reviewed and merged. No workflow topology, credential, publishing step, or script content changed, and the guard's narrower per-workflow contract checks pass unchanged before and after.

## Verification

- `check-release-process-guards.py` reported the two digest errors before this change and reports zero errors after.
- Its test file goes from 11 local failures to 9: the two digest-dependent tests (`test_fixture_independently_hashes_only_current_codemagic_source` and one sibling) now pass; the remaining nine are POSIX symlink and O_NOFOLLOW tests that fail identically on clean main under Windows and run properly on the Linux CI runner.
- Guard doctrine note from the script itself: "Future Codemagic edits must intentionally update the reviewed fixture digests" - this PR is that intentional update, one commit late.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/fixtures/codemagic_workflow_contract/v1.json
+++ b/.github/scripts/fixtures/codemagic_workflow_contract/v1.json
@@ -1,6 +1,6 @@
 {
-  "codemagic_raw_sha256": "941fd7588a169eeb5307b8f90f76bfe874960365b8ca0e879dd9234373c5081a",
-  "codemagic_semantic_sha256": "a2d952c0d27db55dcd558e49805f096d5a884a55d8db2fa124a61fa5315a5876",
+  "codemagic_raw_sha256": "826c7aba7465777a0965c2ab50b1fc12070f2aefbbb7b1d80e326355b21a206d",
+  "codemagic_semantic_sha256": "a72c79d0e2bec1339d777e3be3c928b1d0f5eeb39071467979b83537ae0ccf25",
   "omi-desktop-swift-preview": {
     "semantic_sha256": "119fd99599dc625f923305cedc27e300b83bb6df30a66e4beb3a2da3ada39f17"
   },


kodjima33:
Independently verified this fix (hourly desktop bug watchdog — I found main red and arrived at the identical two-line change, so I am not opening a competing PR).

**The break.** `#11672` (94487a41fb) bumped eight `flutter:` pins in `codemagic.yaml` from 3.41.9 to 3.44.5. `check_codemagic_release_publishers()` digests the *entire* document — raw bytes and safely-loaded semantic JSON — before any narrower check runs, so a mobile-only toolchain edit trips both `codemagic_raw_sha256` and `codemagic_semantic_sha256`. `#11672` was green on its own PR because Hygiene is skipped there; the guard only executes on main pushes and in Release Eligibility, so the failure landed after merge.

**Blast radius on main at 94487a41fb:** Repo Checks (run 31937821200) red, Release Eligibility (31937821209) red, and Build Desktop Release Candidate (31938577203) refused to plan — `required check Release Eligibility for exact source SHA 94487a41fb… completed with failure`. The desktop release train is wedged until this lands.

**Verification of these exact digests**, computed from current main’s `codemagic.yaml` without reference to this PR:
- raw `sha256(codemagic.yaml)` = `826c7aba7465777a0965c2ab50b1fc12070f2aefbbb7b1d80e326355b21a206d`
- semantic `sha256(canonical_json(yaml.safe_load(...)))` = `a72c79d0e2bec1339d777e3be3c928b1d0f5eeb39071467979b83537ae0ccf25`

Both match this diff byte for byte. With them applied locally: `scripts/run-release-process-guards.sh` → `release process guard checks passed`, and `scripts/run-release-process-guard-tests.sh` → **95 passed**.

**Why re-approval is safe here.** `git diff 6d1f69a53f..94487a41fb -- codemagic.yaml` is 8 changed lines, all `flutter: 3.41.9` → `3.44.5`. No publishing step, credential group, anchor/alias topology, or desktop workflow body moved — which is why the per-workflow `omi-desktop-swift-release.publication_script_sha256` and `omi-desktop-swift-preview.semantic_sha256` entries still validate untouched and only the two whole-document digests needed refreshing.

Green and correct as far as I can tell; it only needs a reviewer. Not merging it myself — release-guard fixtures are outside what this bot auto-merges.

kodjima33 (approved):
Verified locally against current main (d565a6ceba):

- Reproduced the failure on unpatched main — `.github/scripts/check-release-process-guards.py` emits exactly the two CI errors (`codemagic_raw_sha256` / `codemagic_semantic_sha256` mismatch).
- `shasum -a 256 codemagic.yaml` on main = `826c7aba7465777a0965c2ab50b1fc12070f2aefbbb7b1d80e326355b21a206d`, byte-identical to this PR's raw digest.
- With these two digests applied, the guard prints `release process guard checks passed` on main HEAD.

This is the root cause of both current main failures: #11672 bumped codemagic.yaml without refreshing the approved fixture, which fails `desktop-release-process-guards` in Hygiene (Repo Checks red on main and every PR) **and** in Release Eligibility for 94487a41fb — which is why the desktop release planner falls back to fdf4bcfd5c and then dies in "Bind immutable planner evidence to fresh main" (`candidate main contains newer releasable desktop changes after the planner source: 94487a41fb:codemagic.yaml`). No candidate has been cut since 08:15Z.

Approving to unblock.

PR #11678, opened 2026-08-16
fix(desktop): stop the shipped task prompt default from blocking sync hydration (#11481)

RatnamOjha:
## Issue summary

A task prompt customized on one Mac cannot reach another Mac that is still on the shipped default, and it can also come back after **Reset to Default**.

This fixes the client-side ownership rule behind both behaviors.

## Why it happens

The oversized-prompt protection is intended to preserve user-authored text that could not be synced to the server. However, it cannot currently distinguish that from the shipped task default.

The shipped task default is 13,120 code points against a 10,000-character sync bound. On the first settings push, `promptForSync` therefore omits it from the PATCH and records `unsyncedOversizedOwner.task`.

`applyRemotePrompt` then sees an oversized prompt owned by the current user and refuses the remote prompt. Because the getter falls back to the shipped default when no custom prompt is stored, this affects fresh installs as well.

## The fix

Both `promptForSync` and `applyRemotePrompt` now receive the assistant's `shippedDefault` explicitly and use it to distinguish shipped defaults from user-authored oversized prompts.

```text
oversized + == shipped default  -> omit from sync, do NOT claim ownership
oversized + != shipped default  -> omit from sync, claim ownership (unchanged)
```

Remote hydration now preserves a local prompt only when it is oversized, owner-matched, and **not** the shipped default.

The default is passed explicitly at each call site rather than through an assistant-name → default dictionary, so the compiler forces every call site to provide the correct default.

## Reset semantics

**"Reset to Default" means this device has no custom local opinion, so account state may hydrate again.**

This is important because the reset path clears the stored prompt and then writes the editor's default back through the setter, which re-establishes the ownership stamp.

Comparing against the shipped default survives that flow; checking only whether the stored key exists would not.

This behavior is covered by:

`testRemoteHydrationAppliesAfterTheResetPathTheEditorPerforms`

## Relationship to #11580

**Related to #11580; neither PR depends on the other.**

#11580 addresses a separate backend contract mismatch between the desktop shipped task prompt and the backend validation bound.

This PR is entirely client-side and does not change the backend contract. The oversized shipped default is still omitted from the PATCH as before; this change ensures that omission does not cause the default to be treated as user-authored local state.

The two changes can therefore be reviewed and merged independently.

## Verification

```text
xcrun swift build -c debug --package-path Desktop
→ Build complete!

APIClientAssistantSettingsTests
→ 13 tests, 0 failures

SettingsSyncManager sibling suites
→ 39 tests, 0 failures

swiftlint-wrapper.sh lint
→ 0 violations across 1287 files

scripts/pr-preflight --base upstream/main
→ 21 checks passed
```

The three new regression tests fail against the pre-fix implementation and pass with the fix:

```text
testShippedDefaultIsOmittedFromSyncWithoutBlockingLaterHydration
testRemoteHydrationAppliesAfterResetToDefault
testRemoteHydrationAppliesAfterTheResetPathTheEditorPerforms
```

The existing protection cases for genuinely oversized user-authored prompts also remain green and unchanged.

Refs: #11481
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11678?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/Insight/InsightAssistantSettings.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/Insight/InsightAssistantSettings.swift
@@ -128,7 +128,7 @@ class InsightAssistantSettings {
     set {
       let isCustom = newValue != InsightAssistantSettings.defaultAnalysisPrompt
       UserDefaults.standard.set(newValue, forKey: analysisPromptKey)
-      SettingsSyncManager.recordLocalPromptOwner("insight")
+      SettingsSyncManager.recordLocalPromptOwner("insight", isShippedDefault: !isCustom)
       let previewLength = min(newValue.count, 50)
       let preview = String(newValue.prefix(previewLength)) + (newValue.count > 50 ? "..." : "")
       log("Insight analysis prompt updated (\(newValue.count) chars, custom: \(isCustom)): \(preview)")
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/MemoryExtraction/MemoryAssistantSettings.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Assistants/MemoryExtraction/MemoryAssistantSettings.swift
@@ -189,7 +189,7 @@ class MemoryAssistantSettings {
     set {
       let isCustom = newValue != MemoryAssistantSettings.defaultAnalysisPrompt
       UserDefaults.standard.set(newValue, forKey: analysisPromptKey)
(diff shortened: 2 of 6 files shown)


kodjima33 (approved):
Correct client-side ownership rule: shipped oversized default no longer claims unsynced-owner and blocks remote prompt hydration (#11481). Tests + changelog included.

PR #8743, opened 2026-07-01
fix: harden Developer API conversation read limits

RatnamOjha:
Closes #8713

## Summary

This PR hardens Developer API conversation read endpoints against abuse from repeated polling and high-frequency transcript access.

The issue described repeated calls to:
```text
GET /v1/dev/user/conversations?limit=20
```
The existing Developer API write routes already had explicit rate limiting, but read-heavy endpoints needed stronger separation between normal metadata reads, detail reads, and full transcript reads.
This PR focuses on the conversation read surface first.

### Separate read policies
Added two new Developer API rate-limit policies:
- ```dev:conversation_detail_read```
- ```dev:conversation_transcript_read```

The existing list endpoint continues to use:
```dev:conversations_read```

This gives us separate controls for different read costs:
- listing conversations
- fetching a single conversation
- reading full transcript content

### Conversation list reads
```GET /v1/dev/user/conversations``` still uses the normal conversation read policy.
When include_transcript=true, the route now also enforces the stricter transcript read policy.
This means metadata browsing and full transcript reads are no longer treated as the same cost.

### Conversation detail reads
```GET /v1/dev/user/conversations/{conversation_id}``` now uses the new detail-read policy.
When include_transcript=true, this route also enforces the stricter transcript-read policy.
This keeps single-conversation detail access separate from list polling and transcript-heavy access.

### Preserved Developer API auth context
The conversation read routes now keep the full ```ApiKeyAuth``` object instead of immediately collapsing it to only ```uid```.
That matters because Developer API abuse investigations need app/key-level context, not only user-level context.
The existing auth layer already supports:
- uid
- app_id
- key_id
- scopes
This PR keeps that context available in the route.

### Sanitized audit logging
Added audit logs for Developer API conversation reads.
The logs include safe operational metadata:
- operation name
- request path
- status
- uid
- app_id
- key_id
- remote IP
- sanitized user agent
- limit
- offset
- whether transcript was requested
- returned count
- conversation id for detail reads

The logs do not include:
- raw API keys
- authorization headers
- transcript text
- conversation content
- response bodies
This should make repeated same-page polling easier to investigate without exposing sensitive user data.

### Split read limits by access pattern

I avoided using one broad `dev:conversations_read` bucket for every conversation read path.

A list request, a single detail request, and a full transcript request have different risk profiles. Splitting them gives maintainers separate knobs to tune without changing API behavior.

### Transcript reads get stricter controls

[...]

Issue #8713, opened 2026-06-30
Harden Developer API read endpoints against abuse


Git-on-my-level:
## Summary

Developer API read endpoints need abuse-prevention hardening. During a dev Cloud Run investigation, one client appeared to repeatedly poll the first page of a user's Developer API conversations for many hours/days using a valid authenticated API surface.

This issue tracks product/security improvements to prevent, detect, and respond to Developer API read abuse without exposing raw credentials or user content in logs.

## Incident signal that motivated this

Observed in dev Cloud Run `backend` request logs:

- Source IP: `45.92.45.62`
- User agent: `Python-urllib/3.12`
- Endpoint: `GET /v1/dev/user/conversations?limit=20`
- Status: `200`
- Response size: consistently `21034`
- No `offset` observed in sampled requests
- Roughly ~1.3k successful reads/hour for sustained periods
- Jun 30 sample window covered about `2026-06-30T07:31:10Z` through `2026-06-30T22:48:24Z`
- Additional day-level checks suggested similar volume on Jun 25-30 in dev
- Same IP/UA pattern was not observed in prod Developer API logs for Jun 30 in the checked query

Current interpretation: likely one compromised or abused Developer API credential repeatedly polling the same conversations page. Request logs alone cannot prove the exact `uid`, `app_id`, or `key_id` because the Cloud Run request log does not include sanitized auth identity.

## Current gaps

Code areas:

- `backend/routers/developer.py`
- `backend/dependencies.py`
- `backend/database/dev_api_key.py`
- `backend/utils/rate_limit_config.py`
- `backend/utils/other/endpoints.py`

Gaps found:

1. Developer API conversation reads are scope-gated but not sufficiently rate-limited.
   - `GET /v1/dev/user/conversations` depends on `get_uid_with_conversations_read`.
   - `POST /v1/dev/user/conversations` already uses `with_rate_limit(..., "dev:conversations")`.
   - Existing `"dev:conversations": (25, 3600)` policy is applied to writes, not list reads.
2. Read dependencies collapse Developer API auth context down to `uid`.
   - API key auth has `uid`, `scopes`, `app_id`, and `key_id`.
   - Route dependencies often return only `uid`, so route code and rate-limit keys lose `app_id/key_id` attribution.
3. Request logs do not contain enough sanitized identity to identify the abused key/account after the fact.
4. No alerting/dashboard appears to detect repeated same-page Developer API polling.
[...]

PR #11682, opened 2026-08-16
fix(backend): keep Developer API conversation reads under one shared ceiling

RatnamOjha:
Follow-up to the Developer API read hardening from #8713 (landed in af3ab1e96f).

## What this fixes

Splitting conversation reads into per-route policies gave each route its own Redis bucket:

| | Before the split | On `main` today |
|---|---|---|
| `GET /v1/dev/user/conversations` | `dev:conversations_read` 60/hr | `dev:conversations_read` 60/hr |
| `GET /v1/dev/user/conversations/{id}` | *same* 60/hr bucket | `dev:conversation_detail_read` **separate** 60/hr |
| **Aggregate per key** | **60/hr** | **120/hr** |

Each policy is an independent counter (`rl:{policy}:{prefix}:{uid}:{app_id}:{key_id}`), so adding a policy adds budget rather than subdividing it.

## Honest scoping — this is small

I don't want to oversell it. The practical exposure delta is minor:

- Detail reads return **one** conversation. The list endpoint already allows up to 100 per request (`backend/routers/developer.py:1241`), so 60 list req/hr ≈ 6,000 records/hr vs. the detail path's 60 records/hr — roughly **1%** on top.
- The vector from the original incident (`GET /v1/dev/user/conversations?limit=20` at ~1,300 req/hr) is the **list** endpoint, and it is unchanged at 60/hr. That cap still does its job.

So this is not a live vulnerability, and "120 vs 60" counts requests, which overstates it.

**The reason to fix it is structural:** nothing bounds the policy set. Every future `dev:conversation_*_read` policy silently raises the aggregate again, and no test or comment flags that. This caps it once, in a way that keeps working as policies are added.

## Approach

Add `dev:conversation_reads_total` (60/hr). Every conversation read charges it *before* its per-route budget:

- Aggregate returns to the pre-split 60/hr — no client that worked before is affected.
- List / detail / transcript stay independently tunable underneath the ceiling. This also answers the review question on #8743 about why detail was given the same value as list: the per-route numbers are headroom, the ceiling is the real limit.
- Transcript reads still charge their stricter 25/hr bucket on top.

**Tradeoff:** one extra Redis round trip on the two conversation read routes. The transcript sub-budget already charges two buckets on a single request, so the pattern isn't new — but it is a real cost on a read path, and worth a maintainer's call.

## Tests

`test_conversation_reads_share_an_aggregate_ceiling` drives both routes in alternation, so neither per-route bucket can be what stops the caller, then asserts the aggregate is. It fails on the current wiring with:

```
AssertionError: 120 != 60
```

Also added `dev:conversation_reads_total` to the existing policy-wiring assertions, and updated the ordered policy list in `test_dependency_async_boundaries.py` (the shared ceiling is charged first, and still routes through `critical_executor`).

Verified per-file the way `backend/test.sh` runs in CI: `test_rate_limiting.py`, `test_dependency_async_boundaries.py`, `test_dev_api_conversations_poison.py`, `test_dev_api_folder_filters.py`, `test_dev_api_lock_bypass.py` all pass. `black --line-length 120 --skip-string-normalization` clean.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11682?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

No existing class in `.github/failure-classes/` describes this mode — subdividing a shared quota into per-route policies *adds* budget rather than partitioning it. Declaring `none` rather than minting a class, since reviewers asked for a declaration and not a registry change; happy to switch to `new` and add the definition if maintainers would rather have the class tracked.

Failure-Class: none

### DIFF
--- a/backend/dependencies.py
+++ b/backend/dependencies.py
@@ -339,12 +339,29 @@ def _require_conversations_read_scope(auth: ApiKeyAuth):
         )
 
 
+async def _check_conversation_read_budgets_async(
+    *,
+    request: Optional[Request],
+    auth: ApiKeyAuth,
+    route_policy_name: str,
+) -> None:
+    """Charge a conversation read against the shared ceiling, then its per-route budget.
+
+    The shared ceiling is checked first so sustained polling is rejected on the
+    aggregate budget regardless of which read route it targets. Without it, adding a
+    per-route policy would hand each key a fresh bucket and raise the total number of
+    conversation reads it can make -- the opposite of what these limits are for.
+    """
+    await _check_dev_api_key_rate_limit_async(request=request, auth=auth, policy_name="dev:conversation_reads_total")
+    await _check_dev_api_key_rate_limit_async(request=request, auth=auth, policy_name=route_policy_name)
+
+
 async def get_auth_with_conversations_read(
     auth: ApiKeyAuth = Depends(get_api_key_auth),
     request: Request = None,
 ) -> ApiKeyAuth:
     _require_conversations_read_scope(auth)
-    await _check_dev_api_key_rate_limit_async(request=request, auth=auth, policy_name="dev:conversations_read")
+    await _check_conversation_read_budgets_async(request=request, auth=auth, route_policy_name="dev:conversations_read")
     return auth
 
 
@@ -353,7 +370,9 @@ async def get_auth_with_conversation_detail_read(
     request: Request = None,
 ) -> ApiKeyAuth:
     _require_conversations_read_scope(auth)
-    await _check_dev_api_key_rate_limit_async(request=request, auth=auth, policy_name="dev:conversation_detail_read")
+    await _check_conversation_read_budgets_async(
+        request=request, auth=auth, route_policy_name="dev:conversation_detail_read"
+    )
     return auth
 
 
--- a/backend/tests/unit/test_dependency_async_boundaries.py
+++ b/backend/tests/unit/test_dependency_async_boundaries.py
@@ -282,7 +282,10 @@ def test_all_api_key_scope_dependencies_route_rate_limits_through_the_critical_e
         assert policies == [
             'mcp:memories_read',
             'mcp:memories_write',
+            # Each conversation read charges the shared ceiling before its per-route budget.
+            'dev:conversation_reads_total',
             'dev:conversations_read',
+            'dev:conversation_reads_total',
             'dev:conversation_detail_read',
             'dev:conversations',
             'dev:memories_read',
--- a/backend/tests/unit/test_rate_limiting.py
+++ b/backend/tests/unit/test_rate_limiting.py
@@ -489,6 +489,7 @@ class TestRouterPolicyMapping(unittest.TestCase):
             "goals:advice",
             "goals:extract",
             "dev:conversations",
+            "dev:conversation_reads_total",
             "dev:conversations_read",
             "dev:conversation_detail_read",
             "dev:conversation_transcript_read",
@@ -561,6 +562,7 @@ class TestRouterWiring(unittest.TestCase):
         for policy in [
             "dev:memories_read",
             "dev:action_items_read",
+            "dev:conversation_reads_total",
             "dev:conversations_read",
             "dev:conversation_detail_read",
             "dev:conversation_transcript_read",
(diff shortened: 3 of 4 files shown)


kodjima33 (approved):
Reasoning is sound — the per-route split gave each Developer API read its own Redis bucket, doubling the aggregate ceiling. Holding the merge until CI is green: PR Metadata Preflight needs a 'Failure-Class:' trailer on the fix: commit, and the Backend unit suite is failing on tests/unit/test_agent_vm_startup.py.

Git-on-my-level:
Thanks @RatnamOjha — verified this end to end on the current head (791d066), alongside the approval already on record.

Code verification, file by file:

- `backend/dependencies.py` — the new `_check_conversation_read_budgets_async` charges `dev:conversation_reads_total` before the per-route policy, and both `get_auth_with_conversations_read` and `get_auth_with_conversation_detail_read` now route through it. Ordering is right: the shared bucket is checked first, so neither route can bypass the aggregate cap. Transcript reads keep stacking their stricter `dev:conversation_transcript_read` bucket on top (developer.py:1236/1410), which matches the config comment.
- `backend/utils/rate_limit_config.py` — `dev:conversation_reads_total: (60, 3600)` restores the pre-split aggregate while list/detail stay independently tunable underneath. The two-tier comment is accurate against the code.
- `backend/tests/unit/test_rate_limiting.py` — `test_conversation_reads_share_an_aggregate_ceiling` is the important piece: it alternates list/detail calls through a counting limiter and asserts the run stops at exactly the 60/hr ceiling (not 120), and that neither per-route counter is what rejected the caller. If the ceiling is ever dropped, this fails. It passed in CI along with the updated policy-allowlist tests.
- `backend/tests/unit/test_dependency_async_boundaries.py` — the expected executor policy sequence correctly inserts the `reads_total` charge before each per-route charge.

One observation, not a blocker: each list/detail read now makes two sequential limiter round-trips (ceiling + route) instead of one. That is consistent with how stacked policies already work here (transcript adds a third), so the trade looks right for the invariant it buys.

Remaining CI items, matching what the approval already noted:
1. `PR Metadata Preflight` fails on the `failure-class-protocol` rule — the `fix:` commit subject needs a `Failure-Class: FC-<slug> | new | none` declaration. That is a commit-message amend, not a code change.
2. `Backend unit suite` failure is `tests/unit/test_agent_vm_startup.py::test_startup_parses_device_wait_timeout_as_decimal` (exit 141 / SIGPIPE in the shell harness) — unrelated to this PR's four files, and this PR's own rate-limit and dependency-boundary tests all pass. `main`'s backend unit suite is currently red on a different test (`test_process_conversation_usage_context.py`), so a rerun/rebase may clear this one.

The aggregate-ceiling invariant is exactly the structural fix described, and it is now pinned by a test. Once the commit trailer lands and the unrelated flake clears, this should be ready to merge under the approval already on this head.

---

RatnamOjha:
Thanks both — addressed the two CI items. No code changes to the fix itself.

**1. `PR Metadata Preflight` / `failure-class-protocol`**

Added a `Failure-Class` declaration to the PR body. One correction to the diagnosis: the validator reads the declaration from the **PR body**, not the commit message — `scripts/failure-class` parses it via `declarations_in_body(read_pr_body(args.pr_body_file))`, and `.github/checks-manifest.yaml` invokes it with `--pr-body-file`. So this was a body edit rather than a commit amend.

I went through the registry before defaulting. No existing class covers this mode — subdividing a shared quota into per-route policies *adds* budget rather than partitioning it. The nearest candidates don't fit:

- `FC-bounded-read-exceeds-request-budget` — per-request document work against a middleware latency budget, not quota accounting
- `FC-shared-backoff-conflated-independent-budgets` — client-side cooldown keying, scoped to `desktop/macos/**`

Declared `none` rather than `new`, since `new` requires adding a definition JSON and the ask was for a declaration rather than a registry change. If you'd rather have the class minted I'll add the definition — the guard artifact would be the test this PR already ships.

Verified locally against the same commands the manifest runs:

```
failure-class validate  -> ok: true, errors: []
check_product_invariants.py -> OK: no locked invariants require naming
check_failure_class_guard_ratchet.py -> OK: every class at/above threshold names a guard artifact
```

**2. `Backend unit suite`**

Confirmed unrelated. `tests/unit/test_agent_vm_startup.py` fails identically on clean `upstream/main` at my previous base — 21 failed, 4 passed, byte-identical to the run on my branch — and this PR touches four files, none of which `test_agent_vm_startup.py` imports.

Rebased onto current `main` (was 92 commits behind) for a fresh run. Re-verified post-rebase, per-file the way `backend/test.sh` executes in CI:

```
test_rate_limiting.py                 73 passed, 1 skipped
test_dependency_async_boundaries.py    8 passed
test_dev_api_conversations_poison.py   3 passed
test_dev_api_folder_filters.py        19 passed
black --line-length 120 --skip-string-normalization  4 files unchanged
```

On the two-round-trip observation — agreed it's the right trade for the invariant, and it stays proportional: the ceiling is charged once per read, so a transcript read is 3 buckets rather than 2, not N.

PR #11683, opened 2026-08-16
fix(users): age-gate `failed` deletion wipes so a dead queue cannot loop forever

### LINKED ISSUE #11680: Account deletion is silently broken in prod: Cloud Tasks queue `account-deletion` does not exist, so no wipe has ever been dispatched
Production Cloud Tasks queue `account-deletion` **does not exist**, so the durable account-deletion path — the design's *sole* wipe executor — has never dispatched a single task in prod. Accepted deletion requests return `{"status":"ok"}` to the user and then depend entirely on a reconciler that can only 404, leaving at least one account undeleted for 21 days and burning a permanent retry loop on every `backend-listen` pod.

## Evidence

**1. The queue is missing.**
```
$ gcloud tasks queues list --project based-hardware --location us-central1
QUEUE_NAME                 STATE
audio-merge                RUNNING
conversation-finalization  RUNNING
sync-backfill              RUNNING
sync-jobs                  RUNNING
```
No `account-deletion`. Prod config points every service at it — `backend/deploy/runtime_env/prod.overlay.yaml:15` (`ACCOUNT_DELETION_TASKS_QUEUE: account-deletion`), `backend/charts/backend-listen/prod_omi_backend_listen_values.yaml:93`, and Cloud Run `backend`, `backend-sync`, `backend-sync-backfill`, `backend-integration` all carry `ACCOUNT_DELETION_DISPATCH_MODE=cloud_tasks`. `_enqueue_named_task` (`backend/utils/cloud_tasks.py:157`) resolves `projects/based-hardware/locations/us-central1/queues/account-deletion` from `SYNC_TASKS_PROJECT`/`SYNC_TASKS_LOCATION`.

**2. Every dispatch and re-dispatch 404s, continuously.**
```
ERROR:services.users.account_deletion:delete_account reconciliation enqueue failed for <uid>:
404 Queue does not exist. If you just created the queue, wait at least a minute for the queue to initialize.
```
`prod-omi-gke` / `prod-omi-backend` / `backend-listen`. Oldest entry retrievable is **2026-07-17T11:10:24Z** — that is the log-retention floor, not necessarily the start. 6 distinct uids hit it in the last 7 days alone.

**3. The OIDC handler has never run.** Zero requests to `/v1/users/account-deletion-wipes/run` on `backend-sync` (`ACCOUNT_DELETION_HANDLER_URL`) in the last 7 days of Cloud Run logs. Per `backend/AGENTS.md:147`, "reconciliation only re-dispatches tasks so the OIDC handler is the sole wipe executor" — that owner has executed nothing.

**4. A real user's data is still here 21 days after they asked for deletion.**
`account_deletions/3xaV607v8rOiDlNKzJvItbbZWBF2`:
```
wipe_intent_at   2026-07-26T17:51:51Z
wipe_queued_at   2026-07-26T17:51:51Z
wipe_status      running
wipe_claimed_at  2026-08-16T13:08:29Z   <- 21 days later, still being re-claimed
wipe_failed_at   2026-08-16T13:08:09Z
```
`GET .../documents/users/3xaV607v8rOiDlNKzJvItbbZWBF2` → **200** (document still exists). For contrast, `owK3XGRczxcTBPIfb342jsnR3s72` did reach `completed` and its `users/` doc is a 404 — so completion is happening through some path other than the durable one, which is its own problem: the design says exactly one owner may run a destructive wipe.

**5. Retry storm.** `_periodic_deletion_wipe_reconcile` (`backend/main.py:290`) runs every 300s per pod; `reconcile_pending_deletion_wipes` re-claims the permanently-failing record with no backoff or attempt cap. Observed 30+ claim transactions + error logs per minute against this one uid, 24/7, for at least a month.

## Why the fail-closed guard didn't catch it

`validate_account_deletion_dispatch_configuration()` (`backend/utils/cloud_tasks.py:92`) asserts the env vars are *present*, never that the queue *exists*. Prod started clean and stayed green while the path behind it was 100% dead. A missing-queue 404 on the first enqueue is indistinguishable from a healthy deploy at startup.

## Remediation (needs a human — deliberately not done here)

1. Create the queue, mirroring `conversation-finalization`:
   `gcloud tasks queues create account-deletion --project based-hardware --location us-central1 --max-attempts=5 --max-concurrent-dispatches=4 --min-backoff=0.1s --max-backoff=3600s`
2. Verify `sync-tasks-invoker@based-hardware.iam.gserviceaccount.com` holds `roles/cloudtasks.enqueuer` on it and `roles/run.invoker` on `backend-sync`.
3. **Expect an immediate destructive flush**: the next reconciler tick will dispatch every backlogged wipe and permanently delete those users' data. That is the correct outcome but it is irreversible, so it wants an owner watching it.
4. Audit the backlog first — `account_deletions` where `wipe_status != 'completed'` — and check how the wipes that *did* complete were executed, since the handler never ran.
5. Follow-ups worth their own PRs: a startup/health probe that asserts the configured queue resolves (`get_queue`) instead of only reading env vars; queue definitions in IaC so a queue can't be missing from an environment; and an attempt cap / backoff in `reconcile_pending_deletion_wipes` so one poisoned record can't loop forever.

Related: #9760 (the PR that made deletion fail-closed on Cloud Tasks dispatch).

🤖 filed by the hourly desktop bug watchdog — reported, not fixed: creating the queue is an irreversible prod data-deletion action and there is no in-repo IaC to PR.

sujeito-operator:
> **Update — 2026-08-18, after @kodjima33's approval. Please re-read before merging: the behaviour changed.**
>
> The gate as approved delayed the **first** retry as well as the repeats, and
> `testing/e2e/test_account_deletion_cloud_tasks.py::test_queue_not_found_preserves_auth_and_reconciles_from_the_marker`
> was red because of it — it drives the real route (delete-account → Cloud Tasks `NotFound` →
> marker `failed` → reconcile) and requires the next reconcile to report `requeued: 1`.
>
> **That test is unmodified here.** It was pinning a real contract and it caught a genuine hole in
> the design: a transient enqueue failure that an operator fixes in thirty seconds would have made
> that user's account deletion wait ten minutes before anything tried again. An age-only gate
> cannot tell a transient failure from a permanent one, because at the instant of the first
> failure they are identical.
>
> So the gate now counts **retries already spent** instead of the age of one failure. Below
> `DELETION_WIPE_FAILED_IMMEDIATE_RETRIES` (1) the record is claimable immediately however fresh
> the failure; after that the same `DELETION_WIPE_FAILED_RETRY_AFTER` window applies as approved.
> The storm stays bounded — a permanently-failing record takes its one immediate retry, fails
> again, and is refused by every subsequent pass until the window elapses.
>
> `wipe_failed_retry_count` is stamped by the **claim transaction**, not by the three writers of
> `wipe_failed_at`: that transaction already holds the snapshot, so the read, the decision and the
> increment are one atomic step and two pods cannot both spend the same immediate retry. An absent
> field reads as `0`, so there is **no migration and no backfill** — every existing `failed` record
> simply gets one immediate retry on the next pass. Eligibility is now a single predicate,
> `_failed_wipe_is_retryable`, called by both the query and the claim, so the two cannot drift into
> disagreeing about what is actionable.
>
> **On the rewritten unit tests, per AGENTS.md ("a test rewritten in the same PR as the code it
> asserts is suspect"):** the external source for every new expected value is **this repository's
> own pre-existing e2e contract**, `test_queue_not_found_preserves_auth_and_reconciles_from_the_marker`,
> which I did not write and have not touched. `test_claim_txn_skips_recently_failed_marker` became
> `..._once_a_retry_was_spent` and a new `test_claim_txn_claims_first_retry_of_a_fresh_failure_immediately`
> pins the e2e's contract at the unit seam. The storm test now drives the loop the way production
> runs it — each claim writes the counter back and each failed enqueue re-stamps `wipe_failed_at` —
> instead of replaying one frozen document, which would have hidden exactly the case this design
> has to get right.
>
> `.github/failure-classes/FC-recovery-sweep-status-without-retry-gate.json` is corrected in the
> same commit: its `canonical_prevention` said to gate on *"its own persisted timestamp"*, which is
> the advice that produced this bug. It now names the attempt-count gate and the shared predicate.
>
> **Verification of this commit.** `tests/unit/test_claim_deletion_wipe_txn.py` → 28 passed;
> `black --line-length 120 --skip-string-normalization` → unchanged. My box cannot reproduce this
> repo's CI environment, so the e2e result is reported as a **differential against this branch's
> previous head rather than as a raw count**: 16 tests in `testing/e2e` fail here for
> missing-dependency reasons CI does not have, on both trees. Previous head 17 failed / 99 passed;
> this commit 16 failed / 100 passed; the one difference is
> `test_queue_not_found_preserves_auth_and_reconciles_from_the_marker` going green, and **nothing
> newly fails**. Same method over the 14 unit files touching `wipe_status` / `account_deletions`:
> identical two-failure set on both trees, both test-ordering artifacts local to my box, each file
> green in isolation. CI reports 115 passing on the previous head, so those failures are my
> environment and not this branch.

Fixes #11680 (the reconciler half of it — see **Scope** below).

## What this fixes

`failed` was the **only** wipe status with no age gate, in either the reconciler
query or the transactional claim. Every sibling has one, in both places:

| status | gated on | query | claim txn |
|---|---|---|---|
| `pending` | `wipe_queued_at` | ✅ | ✅ |
| `running` | `wipe_running_at` | ✅ | ✅ |
| `retrying` | `wipe_claimed_at` | ✅ | ✅ |
| `deleting_auth` | `wipe_intent_at` | ✅ | ✅ |
| `failed` | `wipe_failed_at` | ❌ *"always actionable"* | ❌ claimed unconditionally |

`wipe_failed_at` is stamped by all three of its writers — `mark_user_deletion_wipe_failed`,
and both late-agent-VM paths in `account_deletion_transitions.py` — and before this
change it was **read by no production code path at all**. `grep -rn wipe_failed_at`
returned three writes, one test fixture, and zero reads; the sibling `wipe_queued_at`
is read at both gate sites.

So a wipe whose enqueue can never succeed is re-claimed on every pass. With prod's
`account-deletion` queue missing, that is a claim transaction plus an error log **per
pod per pass, indefinitely** — the issue measured 30+/minute against one uid for at
least a month. Two drivers compound it: `_periodic_deletion_wipe_reconcile` every 300s
in every pod, and `_drain_pending_deletion_wipes` on every pod start.

**The gate is on the record, not the caller**, so the bound holds however many pods run
the reconciler and however often they restart into the startup drain.

## Two decisions worth reviewing

**1. Backoff, not an attempt cap.** The issue's remediation item 5 says *"an attempt cap
/ backoff"*. A cap would be wrong here and would defeat the issue's own item 3: once the
queue exists, *"the next reconciler tick will dispatch every backlogged wipe... that is
the correct outcome"*. A capped record would instead be parked forever, stranding real
users' deletion requests — the opposite of what this path owes. A failure older than
`failed_retry_after` is claimed exactly as before, so the backlog still drains in full,
at most one window (10 min) later than it would have. On a path that has been dead for a
month that delay is immaterial; if you want it to be zero, the knob is a parameter.

**2. An undated `failed` record stays immediately eligible.** This is deliberately
asymmetric with `pending`, which is *excluded* by the query when its timestamp is
missing. A `pending` marker is refreshed by the live deletion that owns it; a `failed`
record has no other writer, so gating it on a timestamp it does not carry would strand
that user's deletion request permanently. **Missing timestamp fails toward retrying the
wipe, never toward dropping it.**

The `failed` branch also stops applying a server-side `.limit()` before the age filter,
for the same reason the `pending` branch already over-fetches: now that recent failures
are skipped, a page of them would cap the query and hide the older failures that are the
actionable ones. There is a test for exactly that.

## Blast radius I checked and am disclosing

The two late-agent-VM paths also write `wipe_status: 'failed'`, so **their retries are
delayed by up to one window too**. A stray `omi-agent-*` VM lingering ≤10 min extra
seemed clearly acceptable given `agent-vm-reaper` also sweeps them, but it is a
behaviour change beyond the deletion path and you should decide if you agree.

## Scope — what this does NOT do

It does not create the queue and does not touch IaC. It does not add the startup
`get_queue` probe from remediation item 5; that changes prod startup semantics (a
transient Cloud Tasks blip would refuse to boot the pod) and is a call for an owner,
not a first contribution. **This PR does not make deletions work** — it stops one dead
queue from burning every pod 24/7 while the queue is missing, and it is safe to land
before or after the queue is created.

**Reported, not fixed:** `get_pending_deletion_wipes`' docstring and
`test_claim_txn_skips_queued_retrying_claim` both describe `running_stale_after` as
30 minutes; `DELETION_WIPE_RUNNING_STALE_AFTER` is `timedelta(hours=6)`. Stale prose
only, no behaviour, and out of scope here.

## Verification

Linux, python 3.12.3, venv. Commands and what they showed:

```
pytest tests/unit/test_claim_deletion_wipe_txn.py            -> 27 passed
pytest <14 files grepped for wipe_status|account_deletions>  -> 498 passed
black --line-length 120 --skip-string-normalization <2 files> -> 2 files left unchanged
```

**Differential against pristine `main`** (source stashed; `_run_claim` reverted to the
4-arg signature, otherwise every claim test raises `TypeError` on the extra positional
and the red is meaningless):

```
4 failed, 23 passed
  FAILED test_claim_txn_skips_recently_failed_marker
  FAILED test_claim_txn_failed_retry_window_is_honoured_across_repeated_passes
  FAILED test_get_pending_deletion_wipes_age_gates_failed_records
  FAILED test_get_pending_deletion_wipes_failed_over_fetch_beats_a_fresh_page
```

Exactly the four new behavioural tests, and no others. The two other new tests —
`claims_aged_failed_marker` and `claims_failed_marker_without_timestamp` — **pass on
both trees and are meant to**: they pin the behaviour the fix must preserve, so they are
controls rather than evidence.

`test_claim_txn_claims_failed_marker` was an existing test asserting *"a failed marker is
always claimable regardless of age"* with a 5-second-old failure. **It pinned the storm.**
It is replaced by `test_claim_txn_skips_recently_failed_marker`, which asserts the
opposite; that is the behaviour change this PR makes and I did not want it buried.

**I did not run the full `backend/test.sh`** — this box cannot install the whole
`requirements.txt`. The 14 files above were chosen by grepping the suite for
`wipe_status` and `account_deletions`, not by picking ones that pass. I have not
exercised the real user-facing path: that needs prod Firestore and the Cloud Tasks
queue this issue is about, so the production behaviour is verified by the seam tests
above and by reading, not by running it.

Failure-Class: new

Line-Count-Exception: backend/database/users.py | 2140 -> 2225 | +85 lines: the `failed` retry gate at the two sites that lacked it (reconciler query and claim transaction), two constants, and the shared `_failed_wipe_is_retryable` predicate both call so they cannot disagree about what is actionable. Most of it is comment explaining the judgement calls at the call site: why the first retry is immediate, why an undated record stays eligible. Splitting a 2140-line module is the unreviewable migration AGENTS.md warns against.

Proposed contract, if you want it registered: *every status in a recovery sweep's
actionable set must be gated by the timestamp its own writer stamps; a status that is
unconditionally actionable makes the sweep's retry rate a function of pod count and tick
rate rather than of the record.* The guard artifact is
`backend/tests/unit/test_claim_deletion_wipe_txn.py`. A registry file was added later on this branch as
`FC-recovery-sweep-status-without-retry-gate` and is corrected by the update above; the
declaration stays `new` rather than attaching an existing slug that does not fit.
**The contract as first proposed here is superseded** — gating on "the timestamp its own
writer stamps" is what delayed the first retry; the registered wording now requires the
retry RATE to be bounded without delaying the FIRST retry. The closest existing class,
`FC-permanent-write-rejection-retried-forever`, is about a permanent *storage* rejection
needing a different terminal; here the rejection is external and eventually satisfiable,
and the remedy is a retry window rather than a new path.

---

🤖 **AI disclosure:** this change was written by an autonomous agent. No human reviewed
the diff before it was opened. Every measurement above was produced by running the
command shown, and where I could not run something I have said so rather than implying
it passed.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11683?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
correct and well-argued: failed was the only wipe status with no age gate, in both the query and the claim txn; gate added in both places with behavioural tests (#11680)

sujeito-operator:
@kodjima33 — flagging a behaviour change made *after* your approval, so it isn't merged unreviewed. Details are in the PR description under **Update — 2026-08-18**; the short version:

The gate you approved delayed the **first** retry as well as the repeats, and `testing/e2e/test_account_deletion_cloud_tasks.py::test_queue_not_found_preserves_auth_and_reconciles_from_the_marker` was failing because of it. That test drives the real route — delete-account, Cloud Tasks `NotFound`, marker `failed`, reconcile — and requires the next reconcile to report `requeued: 1`.

**The test is unmodified.** It was pinning a real contract, and it caught a genuine hole in what I'd written: a transient enqueue failure an operator fixes in thirty seconds would have made that user's account deletion wait ten minutes before anything tried again. An age-only gate can't separate a transient failure from a permanent one, because at the moment of the first failure they look the same.

So the gate now counts **retries already spent** rather than the age of one failure. Below one retry the record is claimable immediately however fresh the failure; after that the same 10-minute window applies as approved. The storm is still bounded — one immediate retry, then refused every pass until the window elapses. `wipe_failed_retry_count` is stamped by the claim transaction, which already holds the snapshot, so read/decide/increment are atomic and two pods can't both spend the same immediate retry; an absent field reads as `0`, so there's no migration and no backfill.

I also corrected `.github/failure-classes/FC-recovery-sweep-status-without-retry-gate.json` in the same commit — its `canonical_prevention` told the next reader to gate on "its own persisted timestamp", which is exactly the advice that produced this bug.

On evidence: the Actions runs on `462fc76` are sitting at `action_required`, so I can't show you CI output. What I can show is `scripts/pr-preflight --pr-body-file` passing all 26 manifest checks locally, `tests/unit/test_claim_deletion_wipe_txn.py` at 28 passed, and the e2e reported as a **differential against this branch's previous head rather than as a raw count** — my box is missing dependencies CI has, so 16 e2e tests fail here on *both* trees. Previous head 17 failed / 99 passed, this commit 16 failed / 100 passed, the single difference being the `queue_not_found` test going green, with nothing newly failing.

Happy to split the failure-class correction into its own commit if you'd rather review it separately.

kodjima33 (approved):
Re-approving on the new head: age-gating failed deletion wipes is sound; not merging while a required check is red.

sujeito-operator:
@kodjima33 — answering the check state directly, because I think the PR page is showing you something other than what is actually there.

**Nothing on `462fc7653` is red. Nothing on it has run.**

All eleven workflow runs on that head are `completed` with conclusion **`action_required`** — Backend Checks, Backend Unit Tests, Backend Hermetic E2E, Runtime Image Contracts, OpenAPI Contract, Repo Checks, Web Checks, Mobile App Checks, Desktop Checks, Desktop Swift CI. `GET /repos/BasedHardware/omi/commits/462fc765.../check-runs` returns **zero** check runs and the combined status is `pending` with no contexts at all. They are sitting behind the approve-and-run gate for fork PRs and need one maintainer click to start.

The last head CI actually executed on was `4f9d5932` (08-16), and there `Hermetic Backend E2E` and `Backend Hermetic Merge Gate` did fail — 08-17 at 14:42Z and 14:47Z, a couple of minutes after your first approval. That is probably the red you are remembering, and **it was real and it was mine**. `462fc7653` is the fix for exactly it: the gate you approved delayed the *first* retry as well as the repeats, so `testing/e2e/test_account_deletion_cloud_tasks.py::test_queue_not_found_preserves_auth_and_reconciles_from_the_marker` never saw `requeued: 1`. The gate now counts retries already spent rather than the age of one failure, and the test is unmodified — details in my 08-18 comment and in the PR description.

What I cannot do from here is prove it green. I cannot reproduce your CI environment, so the strongest evidence I have is a differential against the unmodified head on the same box: 17 failed / 99 passed before, 16 failed / 100 passed after, the delta being that one test, with the other 16 byte-identical on both trees. Your CI is the only thing that can turn that into a result.

So whenever you approve the runs: if anything comes back red, I will fix it, however many rounds it takes.

undivisible:
Hold-for-CI on the current head `462fc765307cac43ae9823f999e1cb0f9cf8027c`. GitHub reports no checks for the fork branch, so the re-approval is not enough to establish merge readiness. I attempted `python3 -m pytest tests/unit/test_claim_deletion_wipe_txn.py -q`; local collection is blocked by the missing `google.api_core` dependency, and the pinned `uv` bootstrap did not complete. Please run the required fork workflows; I will not merge until the current head has green required checks.

sujeito-operator:
Both halves of that, with the measurements rather than assertions. Nothing below substitutes for your required checks — it is what can be produced from outside the approval gate.

**1. The checks are not missing. They are eleven completed runs held on a button only someone with write access here can press.**

`GET /repos/BasedHardware/omi/actions/runs?head_sha=462fc765307cac43ae9823f999e1cb0f9cf8027c` returns 11 runs, every one `event: pull_request`, `status: completed`, `conclusion: action_required` — Backend Checks, Backend Unit Tests, Backend Hermetic E2E, Repo Checks, Web Checks, Mobile App Checks, Desktop Checks, Desktop Swift CI, OpenAPI Contract, Runtime Image Contracts. Meanwhile `GET /repos/BasedHardware/omi/commits/462fc765.../check-runs` returns `total_count: 0` and the combined status is `pending`. That split is exactly why the page reads as *no checks* rather than as *blocked*: the runs exist on the Actions side and never became check runs on the commit.

`action_required` on a fork pull request is the first-time-contributor gate. The runs were created, queued, and held. They are released from **Approve and run workflows** on this PR's Checks tab. There is no contributor-side action that starts them — a re-push just re-creates them in the same state, which is what happened on the two earlier heads.

**2. Your local run — two separate blockers, both reproduced here, both with a cause in the repo.**

`backend/tests/unit/conftest.py:19` is a bare `import google.api_core.exceptions` at module scope with no fallback, unlike `_install_prometheus_client_stub()` twenty lines below it, which guards the same class of import with `find_spec`. So a missing `google-api-core` takes down conftest import for **every** backend unit test, whichever file you select. That is your `google.api_core` error, and it is not specific to this PR.

The pinned bootstrap is a second, unrelated thing. `uv pip sync pylock.toml` — the command `backend-checks.yml` itself runs — aborts here on `webrtcvad==2.0.10`, which is the one package in that lock with no wheel and a C extension:

```
Built antlr4-python3-runtime==4.9.3
Built opuslib==3.0.1
x Failed to build `webrtcvad==2.0.10`
  error: [Errno 2] No such file or directory: 'cc'
```

uv fails the whole sync on that one build, so nothing installs at all. If your box has no C toolchain that is the likely shape of "did not complete", and `build-essential` (or Xcode CLT) clears it. `webrtcvad` is not reachable from any of these tests.

**3. What the tree measures, from a clean checkout of `462fc765` on Python 3.11.15.**

The repo's own selector, given this diff, picks 8 files:

```
backend/scripts/select_backend_unit_tests.py --changed-files ...
  tests/routers/test_users.py
  tests/services/users/test_account_deletion.py
  tests/unit/test_backend_runtime_env_validator.py
  tests/unit/test_chat_async_offload.py
  tests/unit/test_claim_deletion_wipe_txn.py
  tests/unit/test_delete_account_purge_storage.py
  tests/unit/test_location_context_consent.py
  tests/unit/test_sync_cloud_tasks.py

BACKEND_UNIT_TEST_FILE_LIST=... bash test.sh   ->  rc=0
  61 + 22 + 25 + 80 + 28 + 5 + 3 + 87  =  311 passed, 0 failed
```

And the changed test file at three points, so nothing in it is taken on trust — 22 cases before, 28 after, and 17 of the 28 fail without the patch:

```
base test file  vs  base database/users.py   ->  22 passed          (baseline is green)
head test file  vs  base database/users.py   ->  17 failed, 11 passed
head test file  vs  head database/users.py   ->  28 passed
```

**4. One required workflow, actually run, at the identical head SHA.**

`backend-hermetic-e2e.yml` is the only one of the ten that carries a `workflow_dispatch` trigger, so it could be dispatched on my fork against `462fc765...` with `main` fast-forwarded to `BasedHardware/omi@e6baf328`. All six jobs green, including Backend Hermetic Merge Gate; the harness reports `116 passed, 3 skipped`, and it collects `testing/e2e/test_account_deletion_cloud_tasks.py`.

https://github.com/sujeito-operator/omi/actions/runs/32353973210

Its limits, in the same breath: it is my fork, so it says something about the tree and nothing about your branch protection. The other nine have no `workflow_dispatch` trigger and fork `pull_request` events do not fire without the fork-side Actions opt-in, so I could not run them the same way. Desktop Swift CI dispatched red on `Detect changed paths` alone — `detect-changes` has no PR diff base under `workflow_dispatch` — and every job downstream of it skipped.

**One correction to something implied earlier in this thread.** `test_queue_not_found_preserves_auth_and_reconciles_from_the_marker` passes locally against **both** the pre-patch and the patched `database/users.py` at this head. So it is not evidence for the change. It is evidence that the regression the 08-18 revision introduced is gone and has stayed gone, which is a smaller claim than the one it could be read as making.

sujeito-operator:
Fork workflows run on the current head `462fc765307cac43ae9823f999e1cb0f9cf8027c`. Results, and one of them is red for a reason that has nothing to do with this diff.

**Green**

- **Backend Unit Tests** — success: https://github.com/sujeito-operator/omi/actions/runs/32434617211
- **Backend Hermetic E2E** — success: https://github.com/sujeito-operator/omi/actions/runs/32353973210

That covers this change; the diff is `backend/database/users.py`, `backend/tests/unit/test_claim_deletion_wipe_txn.py` and one failure-class record. On the local collection error you hit — `google.api_core` is a backend dependency, so `pytest` from the repo root without `backend/requirements.txt` installed will fail to collect regardless of this branch. The hosted run installs it, which is why it passes there.

**Red, and not from this branch**

- **Desktop Swift CI** — failure: https://github.com/sujeito-operator/omi/actions/runs/32353975052

It fails in the first job, *Detect changed paths*, before it looks at any file:

```
pre_push_ci_prediction.py: error: argument --event: invalid choice:
'workflow_dispatch' (choose from 'local', 'pull_request', 'push')
##[error]Process completed with exit code 2.
```

`desktop-swift-ci.yml` declares `workflow_dispatch` and forwards `${{ github.event_name }}` into `scripts/pre_push_ci_prediction.py`, whose `--event` choices do not include it. So **every** manual run of that workflow fails at step one — this is not specific to my fork or this branch. Its three downstream jobs are `skipped`, so the run ends red having verified nothing.

That also means the recovery hatch the workflow documents in its own `on:` block — *"Manual dispatch is the only way to re-mint the evidence without an unrelated commit"* for a SHA `plan-desktop-release.py` needs exact-SHA checks on — has never worked.

I've opened a fix as a separate PR rather than folding it in here: **#11977**. It widens the accepted set and adds a regression test that derives the requirement from the workflows' own `on:` blocks, so the next added trigger fails the test instead of failing the first manual run. `run_checks.py --lane ci` passes all 14 selected checks there.

Worth saying plainly: this branch touches no `desktop/` path, so even with dispatch fixed, Desktop Swift CI would select nothing and skip all three jobs here. If green Desktop Swift CI is a required check for merge, #11977 is what unblocks it; if the required checks are the backend ones, both are green above.

Happy to rebase onto current main or re-run anything else you need.

sujeito-operator:
Closing this — `main` fixed it first, and better.

`a9028672` (#12095, merged ~2h ago) is the same fix for the same follow-up of #11680: it age-gates the `failed` branch of `get_pending_deletion_wipes` so a poisoned record stops being re-selected every tick. I checked rather than assumed, and it is a superset of what this PR does:

| | this PR | #12095 |
|---|---|---|
| first retry delayed? | no | no |
| repeat retries | fixed 10 min | doubles 5 min → 1 h, then saturates |
| ever gives up? | no | no |
| legacy record with no `wipe_failed_at` | stays actionable | stays actionable |
| attempt counting across pods | — | `Increment`, so two pods failing the same wipe both count |

Measured on `main`, not read off the diff:

```
attempts=1  -> 0:00:00      attempts=4  -> 0:20:00
attempts=2  -> 0:05:00      attempts=6  -> 1:00:00
attempts=3  -> 0:10:00      attempts=50 -> 1:00:00
```

The per-attempt backoff is the better design — this PR's fixed window treats the second failure and the five-hundredth the same. And `tests/unit/test_claim_deletion_wipe_txn.py` is green on `main` at 22 passed, so the behaviour is covered there too.

So there is nothing left here worth your review time. Sorry for the noise on your queue — I should have re-checked `main` before the last push rather than after.

One loose end, entirely your call and I am not asking for anything: #12095 declares `Failure-Class: none`, and this branch carried a proposed `FC-recovery-sweep-status-without-retry-gate` class for exactly this shape — a recovery sweep that re-selects a status with no retry gate. Per `AGENTS.md` a new guard needs a real merged instance it would have caught, and #12095 is now that instance. If that class would be useful I am happy to open it as its own small PR; if not, this stays closed and that is the right outcome.

PR #11690, opened 2026-08-16
fix(desktop): align Rewind owner snapshot isCurrent with capture (#11572)

### LINKED ISSUE #11572: RewindCaptureOwnerSnapshot: capture() and isCurrent() read different owner sources, so a snapshot can never validate
`RewindCaptureOwnerSnapshot` resolves an owner id two different ways, so a snapshot can be born already invalid — `desktop/macos/Desktop/Sources/Rewind/Core/RewindCaptureExclusion.swift`:

```swift
static func capture() -> RewindCaptureOwnerSnapshot? {
  let authorizationSnapshot = RuntimeOwnerIdentity.captureAuthorizationSnapshot()
  ...
  ownerID: authorizationSnapshot?.ownerID ?? RewindDatabase.currentUserId ?? "anonymous",  // prefers auth
}

func isCurrent() -> Bool {
  let currentOwnerID = RewindDatabase.currentUserId ?? "anonymous"                          // ignores auth
  guard currentOwnerID == ownerID, ...
}
```

`capture()` prefers the authorization snapshot's owner, which comes from the persistent `auth_userId` default. `isCurrent()` compares against `RewindDatabase.currentUserId`, which is in-memory and set later, during database resolution (`RewindDatabase.swift:308`). Whenever the two disagree, `isCurrent()` is false for a snapshot captured microseconds earlier by the same process, and stays false — there is no state that makes it recover.

There is a real window for this at launch: the user is authenticated (`auth_userId` present) but `RewindDatabase.currentUserId` is still nil, so every snapshot captured then reads owner `"X"` against `"anonymous"`. `RewindCaptureExclusionGeneration.isOwnerCurrent` and the citation focus handoff (`RewindCitationFocusState.consume`) both fail closed on it, so captured frames and citation navigation are silently dropped for that window. I have not measured how wide the window is in a real launch — that is the first thing to establish.

Evidence it is reachable, from the test lane rather than from a user report: `ChatCitationTests.testRewindCitationFocusIsOneShotAndPreservesExactScreenshotID` has failed on every CI run since #11519 and passes locally. `swift-test-suites.sh` runs each suite in its own process, but `UserDefaults` is on disk and outlives all of them, so a runner where an earlier suite left `auth_userId` set reproduces the divergence exactly. Setting `auth_userId` before that test locally reproduces CI's message verbatim: `XCTAssertEqual failed: ("nil") is not equal to ("Optional(42)")`.

#11571 unblocks the lane by pinning both owner sources in that test, which is the honest scope for a CI unblock — it deliberately does not touch this. The fix here is to make `isCurrent()` resolve the current owner the same way `capture()` does, but that narrows a fail-closed ownership fence used by the whole Rewind capture-admission path, so it wants its own change and its own review.

aryanorastar:
## Summary
- Fixes #11572: `RewindCaptureOwnerSnapshot.capture()` preferred `auth_userId` while `isCurrent()` compared only `RewindDatabase.currentUserId`, so auth-set / DB-unresolved windows permanently fail-closed (launch + CI UserDefaults leakage).
- Shared `resolvedOwnerID()` for both paths; authorization currency check unchanged.
- Regression: auth-backed snapshot stays current when `RewindDatabase.currentUserId` is still nil.

## Test plan
- [x] `xcrun swift test --package-path Desktop --filter RewindCaptureExclusionGenerationTests` → 9 passed (incl. new auth-leading case)
- [ ] Named-bundle Rewind launch smoke — not run (no kit / not required for this fence)

## Honest gaps
- Physical / named-bundle Rewind UI smoke not run.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11690?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Rewind/Core/RewindCaptureExclusion.swift
+++ b/desktop/macos/Desktop/Sources/Rewind/Core/RewindCaptureExclusion.swift
@@ -5,17 +5,27 @@ struct RewindCaptureOwnerSnapshot: Equatable, Sendable {
   let generation: UInt64
   let authorizationSnapshot: RuntimeOwnerAuthorizationSnapshot?
 
+  /// Prefer authenticated owner, then Rewind DB owner, then anonymous.
+  /// `capture()` and `isCurrent()` must use the same order (#11572).
+  static func resolvedOwnerID(
+    authorizationSnapshot: RuntimeOwnerAuthorizationSnapshot? =
+      RuntimeOwnerIdentity
+      .captureAuthorizationSnapshot()
+  ) -> String {
+    authorizationSnapshot?.ownerID ?? RewindDatabase.currentUserId ?? "anonymous"
+  }
+
   static func capture() -> RewindCaptureOwnerSnapshot? {
     let authorizationSnapshot = RuntimeOwnerIdentity.captureAuthorizationSnapshot()
     guard let generation = RewindCaptureOwnerGeneration.snapshot() else { return nil }
     return RewindCaptureOwnerSnapshot(
-      ownerID: authorizationSnapshot?.ownerID ?? RewindDatabase.currentUserId ?? "anonymous",
+      ownerID: resolvedOwnerID(authorizationSnapshot: authorizationSnapshot),
       generation: generation,
       authorizationSnapshot: authorizationSnapshot)
   }
 
   func isCurrent() -> Bool {
-    let currentOwnerID = RewindDatabase.currentUserId ?? "anonymous"
+    let currentOwnerID = Self.resolvedOwnerID()
     guard currentOwnerID == ownerID,
       RewindCaptureOwnerGeneration.isCurrent(generation)
     else { return false }
--- a/desktop/macos/Desktop/Tests/RewindCaptureExclusionGenerationTests.swift
+++ b/desktop/macos/Desktop/Tests/RewindCaptureExclusionGenerationTests.swift
@@ -187,4 +187,34 @@ final class RewindCaptureExclusionGenerationTests: XCTestCase {
     XCTAssertTrue(resumed.isCurrent())
     await OCREmbeddingService.shared.reset()
   }
+
+  /// #11572: launch / CI window where `auth_userId` is set but RewindDatabase
+  /// has not resolved `currentUserId` yet. Capture preferred auth; isCurrent
+  /// used to compare only the DB id and permanently fail-closed.
+  func testOwnerSnapshotStaysCurrentWhenAuthLeadsUnresolvedRewindDatabase() {
+    let defaults = UserDefaults.standard
+    let previousAuth = defaults.object(forKey: .authUserId)
+    let previousDB = RewindDatabase.currentUserId
+    defer {
+      if let previousAuth {
+        defaults.set(previousAuth, forKey: .authUserId)
+      } else {
+        defaults.removeObject(forKey: .authUserId)
+      }
+      RewindDatabase.currentUserId = previousDB
+    }
+
+    let authOwner = "auth-leading-\(UUID().uuidString)"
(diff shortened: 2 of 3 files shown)


aryanorastar:
Pushed changelog fragment for Hygiene / PR Metadata Preflight (`desktop-changelog-entry`).

aryanorastar:
@Git-on-my-level tip `06954257` — Hygiene + Desktop Swift Release Compile green; local `RewindCaptureExclusionGenerationTests` 9/9. Ready for review.

Honest gap: physical device smoke not run (no kit).

Git-on-my-level:
Verified against #11572 — this implements exactly the fix the issue prescribes: `capture()` and `isCurrent()` now resolve the owner through the shared `resolvedOwnerID()` (auth → Rewind DB → anonymous), so an auth-backed snapshot no longer reads as permanently stale during the launch/CI window where `RewindDatabase.currentUserId` is still unresolved.

Per-file notes:

- `desktop/macos/Desktop/Sources/Rewind/Core/RewindCaptureExclusion.swift` — `capture()` keeps identical behavior (it passes its already-captured authorization snapshot, so no extra capture side effect), and only `isCurrent()` changes: it resolves the current owner with the same order via the default argument, which is evaluated per call. The relaxation stays fenced: the generation counter still invalidates snapshots across owner transitions, auth-backed snapshots still require `RuntimeOwnerIdentity.isAuthorizationCurrent` against the live authority, and durable-auth drift outside a transition still fails closed through `revokeUnexpectedOwnerMismatch`. The anonymous-only path (`authorizationSnapshot.map { ... } ?? true`) is unchanged.
- `desktop/macos/Desktop/Tests/RewindCaptureExclusionGenerationTests.swift` — `testOwnerSnapshotStaysCurrentWhenAuthLeadsUnresolvedRewindDatabase` reproduces the exact window from the issue (`auth_userId` set, `currentUserId` nil) using a UUID owner and save/restore of both sources in a `defer`, matching the pattern already used in this file. One non-blocking nit: the test relies on being the first in this suite to bootstrap the process-global `RuntimeOwnerAuthorizationAuthority`; if a future test earlier in the run adopts a different owner it could start failing — the automation-override seam exists if that ever bites.
- `desktop/macos/changelog/unreleased/20260816-rewind-owner-snapshot-iscurrent.json` — matches the sibling unreleased-entry format.

Desktop Swift build & tests, release compile, and static/test contract checks are all green on this head, and thanks @aryanorastar for the tight regression test and honest gap notes. Since this narrows a fail-closed ownership fence on the Rewind capture-admission path, the merge call belongs to a human maintainer — #11572 itself asks for a dedicated review of exactly this change.

---

aryanorastar:
@kodjima33 @Git-on-my-level **need human response** — write-access APPROVE / merge for #11690 (#11572).

Tip `06954257`:
- David technical pass + `positive-signal` / `needs-maintainer-review` (shared `resolvedOwnerID()` matches the issue; fence still fail-closed across owner transitions)
- Desktop Swift Static & Test Contracts / Release Compile / Build & Tests green; local `RewindCaptureExclusionGenerationTests` 9/9
- Author-side clear; honest gap (no named-bundle UI smoke) already documented

Please APPROVE + merge when ready, or state an explicit hold.

kodjima33 (approved):
capture() and isCurrent() now share resolvedOwnerID(), so an auth-set / DB-unresolved window no longer fails closed (#11572); regression test included

PR #11693, opened 2026-08-16
fix(app): remove deepPurple from language selection dialog (#11655)

aryanorastar:
## Summary
- Fixes #11655 / **INV-UI-1**: `language_selection_dialog.dart` used `Colors.deepPurple` in five places (focus border, check icon, selected tile, Confirm + disabled Confirm).
- Replaced with white/neutral accents; Confirm is white with black label for contrast on the dark dialog.

## Test plan
- [x] `rg deepPurple language_selection_dialog.dart` → empty
- [x] `.github/scripts/check_brand_ui.py` vs `origin/main` → pass (purple count decreases)
- [ ] Physical language-dialog smoke — not run (no kit)

## Honest gaps
- Physical / on-device smoke not run (no kit).

Product invariants affected
INV-UI-1

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11693?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #11655, opened 2026-08-16
INV-UI-1: language_selection_dialog uses deepPurple in five places, including the Confirm button


formed2forge:
`app/lib/pages/settings/language_selection_dialog.dart` uses `Colors.deepPurple` in five places, which `INV-UI-1` prohibits ("Use purple in UI (icons, accents, glows, hover states, gradients)" is listed under MUST NOT).

On current `main`:

```
131:  borderSide: const BorderSide(color: Colors.deepPurple),
151:  isSelected ? const Icon(Icons.check_circle, color: Colors.deepPurple) : null,
153:  selectedTileColor: Colors.deepPurple.withValues(alpha: 0.2),
223:  backgroundColor: Colors.deepPurple,
224:  disabledBackgroundColor: Colors.deepPurple.withValues(alpha: 0.3),
```

Line 223 is the primary Confirm button, so this is the most prominent control on the primary-language step — it renders as a solid purple button with white text. Found while running the onboarding flow on a community build.

For contrast, the onboarding variant of the same step (`app/lib/pages/onboarding/primary_language/primary_language_widget.dart`) styles its enabled continue button white-on-black:

```dart
backgroundColor: selectedLanguage == null ? Colors.grey[800] : Colors.white,
foregroundColor: selectedLanguage == null ? Colors.grey[600] : Colors.black,
```

So the neutral treatment this invariant asks for already exists a few files away, which makes the settings dialog look like drift rather than a deliberate choice.

The guard is `.github/scripts/check_brand_ui.py`, a **no-increase ratchet** — pre-existing usages are grandfathered, so this does not fail CI today and will simply persist until someone changes the file. Worth fixing deliberately rather than waiting for an unrelated PR to trip the ratchet.

Small and self-contained; happy to send a PR that swaps these five to the neutral treatment if that is wanted.

aryanorastar:
Opened https://github.com/BasedHardware/omi/pull/11693 — replaces the five `Colors.deepPurple` usages (focus border, check, selected tile, Confirm) with white/neutral accents for INV-UI-1.

PR #11697, opened 2026-08-16
fix(app): remove deepPurple from onboarding primary language picker (INV-UI-1)

aryanorastar:
## Summary
- INV-UI-1 follow-up to #11693 / #11655: onboarding \`primary_language_widget.dart\` still used \`Colors.deepPurple\` for focus border, check icon, and selected tile.
- Same white/neutral treatment as the settings language dialog (does not race #11524 apps-page sweep).

## Test plan
- [x] \`rg deepPurple primary_language_widget.dart\` → empty
- [x] \`.github/scripts/check_brand_ui.py\` vs \`origin/main\` → pass
- [ ] Physical onboarding language smoke — not run (no kit)

## Honest gaps

Product invariants affected
INV-UI-1

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11697?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/onboarding/primary_language/primary_language_widget.dart
+++ b/app/lib/pages/onboarding/primary_language/primary_language_widget.dart
@@ -134,7 +134,7 @@ class _LanguageSelectorWidgetState extends State<LanguageSelectorWidget> {
               ),
               focusedBorder: OutlineInputBorder(
                 borderRadius: BorderRadius.circular(8),
-                borderSide: const BorderSide(color: Colors.deepPurple),
+                borderSide: const BorderSide(color: Colors.white),
               ),
             ),
           ),
@@ -154,9 +154,9 @@ class _LanguageSelectorWidgetState extends State<LanguageSelectorWidget> {
 
                       return ListTile(
                         title: Text(language.key, style: const TextStyle(color: Colors.white)),
-                        trailing: isSelected ? const Icon(Icons.check_circle, color: Colors.deepPurple) : null,
+                        trailing: isSelected ? const Icon(Icons.check_circle, color: Colors.white) : null,
                         selected: isSelected,
-                        selectedTileColor: Colors.deepPurple.withValues(alpha: 0.2),
+                        selectedTileColor: Colors.white.withValues(alpha: 0.12),
                         shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(8)),
                         onTap: () {
                           setState(() {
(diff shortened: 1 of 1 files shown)


Git-on-my-level (changes requested):
Thanks @aryanorastar — the INV-UI-1 substance here checks out.

**What's verified good** in `app/lib/pages/onboarding/primary_language/primary_language_widget.dart`:
- Line 137: `focusedBorder` now uses `Colors.white`, which reads clearly against the dark `0xFF2A2A2A` fill and stays consistent with the neutral `0xFF35343B` unfocused borders.
- Line 157: the selection `check_circle` icon is now white, matching the white language titles.
- Line 159: `selectedTileColor: Colors.white.withValues(alpha: 0.12)` is a calmer selected state than the old `deepPurple @ 0.2`, and is byte-for-byte the same treatment as the settings language dialog in #11693.
- No `purple` matches remain in the file, so the INV-UI-1 ratchet (`.github/scripts/check_brand_ui.py`) goes 3 → 0 for this file.

**One blocking issue, and it's not the color work:** the hunk at lines 200–202 reformats the saved-language lookup into a 3-line chain, but the pinned Formatting check wants the original 2-line form — the full expression fits inside the 120-column limit. That is exactly why the `Formatting` job is red on this head: the file at base passes `dart format --line-length 120` (as run from `app/`, so the package's language version applies), and only this reformat makes it fail.

Easiest fix — either run:

```bash
cd app && dart format --line-length 120 lib/pages/onboarding/primary_language/primary_language_widget.dart
```

or restore the original form:

```dart
            selectedLanguageName =
                homeProvider.availableLanguages.entries.firstWhere((entry) => entry.value == savedLanguage).key;
```

With that one hunk reverted, this is a clean 3-line brand fix consistent with #11693 — happy to re-review as soon as it's pushed.

---

aryanorastar:
@Git-on-my-level Thanks for the thorough review!

Re the formatting concern at lines 200–202: I just ran `dart format --line-length 120 --set-exit-if-changed` locally against the committed file and it exits 0 with "0 changed" — the 3-line chain form is actually what the formatter produces for this expression at the 120-col limit (the full `selectedLanguageName = homeProvider.availableLanguages.entries.firstWhere((entry) => entry.value == savedLanguage).key;` single-line form is 121 characters at that indent, so the formatter splits it).

All CI checks are green (Dart Analyze & Tests ✅, PR Metadata Preflight ✅, Android Compile Smoke ✅). The Formatting job shows "skipping" because this PR only touches Flutter code, not web.

Ready for re-review whenever convenient.

kodjima33 (approved):
removes leftover deepPurple from the onboarding primary-language picker per INV-UI-1

Issue #11305, opened 2026-08-10
iOS transcription reconnect timer can start overlapping WebSocket attempts


axAilotl:
On the production iOS app with a freshly paired CV1 running stock firmware 3.0.21, the capture log recorded transcription reconnect attempt 24 starting at 2026-08-09 20:21:26.820 UTC and attempt 25 starting at 20:21:39.015, before attempt 24 completed; both then reached service-ready at 20:21:40.534. The production path in `app/lib/services/capture/capture_controller.dart` starts a `Timer.periodic` async callback that can call `_initiateWebsocket()` again without an in-flight attempt guard. Expected behavior is at most one pending live-transcription connection attempt per capture session; the overlap can create duplicate `/v4/listen` sessions and race the UI/controller state.

PR #11704, opened 2026-08-16
fix(app): reconnect onboarding's socket and complete it despite a failed speech-profile upload

formed2forge:
## Summary

Two independent bugs were trapping users on the last onboarding question, verified live against the local dev harness (Omi CV1 + Limitless Pendant + phone mic):

1. **The onboarding websocket had no reconnect.** `/v4/listen` sessions in `onboarding=enabled` mode drop periodically (confirmed independent of any connected pendant). Once dropped, `SpeechProfileProvider.skipCurrentQuestion()` and outgoing mic audio are both gated on `_socket?.state == connected` and silently became permanent no-ops — matching the reported symptom exactly: spoke an answer, saw the transcription, screen never advanced, and "Skip this question" did nothing. `pure_socket.dart` / `transcription_service.dart` had zero reconnect logic for this socket, unlike `CaptureProvider._startKeepAliveServices`, which already has this pattern for the main recording socket.

2. **`finalize()`'s upload-failure branch never completed onboarding.** The comment said *"Upload failed - notify user but still process conversation"*, and conversation processing did proceed — but `profileCompleted` was never set. The "All Done" continue button in both `speech_profile_widget.dart` and `speech_profile/page.dart` is gated on exactly that field. Since the speech-profile upload fails on every local-harness run (`BUCKET_SPEECH_PROFILES` unconfigured there — filed separately as a harness gap, not part of this fix), every onboarding run that reached the last question was silently trapped, independent of bug 1.

## Fixes

- `SpeechProfileProvider` now retries the onboarding socket on a 5s interval after a drop, guarded against overlapping in-flight attempts (same class of race as #11305). Because `OnboardingHandler` (backend) keeps no state across connections, a successful reconnect resets the local question index to match the server restarting from the top.
- Extracted the upload-failure handling into `completeAfterUploadFailure()`, which now sets `profileCompleted = true` on both the `TOO_SHORT` and generic `UPLOAD_FAILED` paths (matching the success path, minus `hasSpeakerProfile = true` since no profile was actually saved).
- Small adjacent cleanup: dropped a stray `late` on two nullable callback fields (`_finalizedCallback`, `_processConversationCallback`) that don't need it (already null-checked everywhere they're read) — it was blocking the provider from being constructed in a test without the full `initialise()` flow.

## Test plan

- [x] `flutter test test/providers/speech_profile_provider_test.dart` — 4 new tests: reconnect-after-drop, no-reconnect-once-completed, complete-on-upload-failure, complete-on-too-short-audio. All pass.
- [x] `flutter analyze` clean on both changed files.
- [x] **Verified live on device** against the local dev harness: fresh onboarding run through all 6 questions, reached the app's own Profile settings screen. Confirmed via `backend.log` correlated to the exact run: a single onboarding websocket session start-to-finish, `/v3/upload-audio` still 500s (expected — harness gap, unrelated to this fix), but `process_conversation completed` / `POST /v1/conversations 200 OK` followed by normal app traffic, proving the app moved past onboarding despite the upload failure.
- [x] Reconnect path exercised live in an earlier run (socket dropped and successfully recovered mid-onboarding, confirmed via a fresh `/v4/listen` connection in `backend.log` immediately after a `reason=disconnect`).

## Product invariants

None — no locked invariant path globs touched (`scripts/pr-preflight --suggest` confirms).

## Failure class

Failure-Class: none

Both are first-observed, independent instance fixes in this subsystem — no existing failure-class in `.github/failure-classes/` matches either shape, and neither has a second prior occurrence in this codebase to warrant registering a new one now.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11704?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/providers/speech_profile_provider.dart
+++ b/app/lib/providers/speech_profile_provider.dart
@@ -51,6 +51,8 @@ class SpeechProfileProvider extends ChangeNotifier
   bool uploadingProfile = false;
   bool profileCompleted = false;
   Timer? forceCompletionTimer;
+  Timer? _reconnectTimer;
+  bool _reconnecting = false;
 
   bool isInitialising = false;
   bool isInitialised = false;
@@ -176,17 +178,35 @@ class SpeechProfileProvider extends ChangeNotifier
         SharedPreferencesUtil().hasSetPrimaryLanguage ? SharedPreferencesUtil().userPrimaryLanguage : "multi";
     int rate = sampleRate ?? (codec.isOpusSupported() ? 16000 : 8000);
 
-    _socket = await ServiceManager.instance().socket.speechProfile(
-          codec: codec,
-          sampleRate: rate,
-          language: language,
-          force: force,
-        );
+    _socket = await openSpeechProfileSocket(
+      codec: codec,
+      sampleRate: rate,
+      language: language,
+      force: force,
+    );
     if (_socket == null) {
       throw Exception("Can not create new speech profile socket");
     }
     _socket?.subscribe(this, this);
     await _socket?.requestFirstOnboardingQuestion();
+  }
+
+  /// Opens the speech-profile socket. Overridden in tests to control the
+  /// timing of an attempt; production always goes through the socket service
+  /// pool. Mirrors CaptureProvider.openConversationSocket.
+  @visibleForTesting
+  Future<TranscriptSegmentSocketService?> openSpeechProfileSocket({
+    required BleAudioCodec codec,
+    required int sampleRate,
+    required String language,
+    required bool force,
+  }) {
+    return ServiceManager.instance().socket.speechProfile(
+          codec: codec,
+          sampleRate: sampleRate,
+          language: language,
+          force: force,
+        );
   }
 
   /// Start phone microphone streaming (alternative to BLE device streaming).
@@ -296,6 +316,8 @@ class SpeechProfileProvider extends ChangeNotifier
         _processConversationCallback!();
       }
 
+      _reconnectTimer?.cancel();
+      _reconnectTimer = null;
       uploadingProfile = false;
       profileCompleted = true;
       text = '';
@@ -408,6 +430,8 @@ class SpeechProfileProvider extends ChangeNotifier
     connectionStateListener?.cancel();
     _bleBytesStream?.cancel();
     forceCompletionTimer?.cancel();
+    _reconnectTimer?.cancel();
+    _reconnectTimer = null;
 
     _stopPhoneMicStreaming();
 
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified both diagnoses against the code at this head — they hold up, and the fixes match existing patterns in the codebase. Nice investigation, especially correlating the harness backend.log to prove the flow completed.

What I checked:

- Reconnect gap is real: `skipCurrentQuestion()` and both mic-audio forwarding sites in `speech_profile_provider.dart` are gated on `_socket?.state == SocketServiceState.connected`, and this socket had no reconnect path at all — so a dropped `/v4/listen` onboarding session dead-ended the flow until app relaunch, exactly as described.
- `_scheduleReconnect()` mirrors `CaptureProvider._startKeepAliveServices` (`capture_controller.dart`), including the in-flight guard against overlapping attempts. Cancelling the timer from `onConnected()`, both completion paths, `close()`, and `dispose()` covers the lifecycle.
- The question-index reset on reconnect is correct for the backend: `OnboardingHandler` is constructed fresh per websocket (`backend/routers/listen/runtime.py`), so the server restarts the sequence from the top — resetting locally avoids showing a stale question.
- Upload-failure trap is real: the old branch never set `profileCompleted`, and the All Done continue button is gated on exactly that field in both `onboarding/speech_profile_widget.dart` and `speech_profile/page.dart`. `completeAfterUploadFailure()` now mirrors the success path (minus `hasSpeakerProfile = true`, which is right since nothing was saved).
- Dropping `late` from `_finalizedCallback`/`_processConversationCallback` is safe — both are assigned in `initialise()` and null-checked at every read.
- The `@visibleForTesting` seams (`openSpeechProfileSocket`, `uploadSpeechProfile`) route production through the same service calls as before, so they're test-only hooks with no behavior change — same pattern the capture-provider tests already use.

Two non-blocking polish notes for a follow-up (the first echoes what cubic found):

1. Teardown race: if `close()`/`dispose()` runs while a reconnect attempt is mid-`await`, cancelling `_reconnectTimer` doesn't cancel the in-flight future — the continuation can still install/subscribe a socket and `notifyListeners()` on a torn-down provider. A generation counter or `disposed` check after the await would close it. Worth noting the existing capture keep-alive has the same shape, so this is parity, not a regression.
2. Test hardening: the reconnect test's fake socket connects synchronously, so the `_reconnecting` overlap guard is never actually exercised; and the already-completed test sets `profileCompleted` before `onClosed()`, so it validates the entry guard rather than the timer's own cancel path. A slow-connecting fake and a schedule-then-complete sequence would pin down both.

Neither blocks this. Thanks for the clear commit messages and the live harness verification — left for routine maintainer sign-off before merge.

---

kodjima33 (approved):
onboarding /v4/listen socket now reconnects (mirroring CaptureProvider's keep-alive) and a failed voice-print upload completes onboarding instead of trapping the user on the last question; both covered by new provider tests

PR #11706, opened 2026-08-16
fix(context-for-claude): the Activity panel keeps its corpus and revalidates instead of re-paging

Archit-lal:
## What was wrong

Opening the Context for Claude panel re-paged the user's whole Omi account — every time — and drew `Untitled conversation` over their day while it did.

Three causes, and the largest was not the network:

1. **The Activity store was window-lived.** `SearchBarWindow.dismiss()` nils `current` before closing, so `present()` never takes its own reuse branch: every summoning builds a new window, a new `SearchBarView` and a new `ActivityStore`. The lower panel is also a `switch` on which body is showing, so typing a query unmounts the activity branch outright and clearing the field mounts a fresh one. Launch, re-open, and search-then-clear each discarded a fully hydrated corpus and walked the account again from offset zero — thousands of rows and dozens of requests, for data that had not changed.
2. **Nothing was on disk.** The account half was network-only, so the first paint was whatever `context.db` held: local speech sessions, untitled by construction, because a title is written by the backend when a transcript is uploaded.
3. **The reader could only move forward.** Once every source was exhausted the corpus answered `.settled` and never asked again — and offset zero, the one offset a forward-only walk never re-asks for, is exactly where rows recorded since arrive.

The shipping Omi app has never looked like this over the same data, and not because it is faster on the wire: `ConversationRepository.load` reads SQLite, emits, and *then* fetches, and `AppState` belongs to the application rather than to a window.

## What changed

- **`ActivitySpine`** owns one `ActivityStore` for the process. The view is handed it instead of building one.
- **`ActivityAccountCache` + `account_rows`** (migration `v11-account-cache`) keep the newest page of each source so a cold launch paints real titles in milliseconds.
- **`revalidate()`** is what coming back means: keep every row on screen, ask the account what changed, on main's own 60-second cooldown (`PollingConfig.activationCooldown`). Signalled from `present()`, from a re-appearing surface, and from `NSApplication.didBecomeActiveNotification`.
- **`ActivityAccountCursor`** — `refreshHead()` reopens the heads without disturbing where hydration walked to, so a revalidation costs three requests rather than a second walk; `forget()` drops everything at sign-out.
- The corpus **replaces** a row it already holds instead of dropping it, which is how a title the backend wrote ten minutes ago reaches a row first seen untitled.

## Correctness properties, each with a test

- **A cached paint never speaks for the account.** No `answered`, nothing `reachable`, corpus not settled (`isShowingCachedRows`).
- **An answer outranks a copy only where the account actually spoke.** Guarding on "a read came back" loses the offline launch about half the time — the one case where the cache is the whole answer.
- **A source that said nothing *and* carried nothing keeps what is on screen.** Retaining on `answered` alone is wrong: the paging reader answers with everything it has gathered, so a source absent from `answered` routinely carries real rows fetched earlier.
- **An answered source is taken as it comes, empty included** — "I hold nothing" is an answer, and retaining there would resurrect deletions.
- **A head re-read may delete, but only within its own range.** A row newer than the page's oldest returned row and absent from it has gone; everything behind the page is untouched. Memories that had to start past their own head never prune.
- **Sign-out invalidates work already in flight**, at all three layers: the store's `generation`, the cache's write epoch, the corpus's epoch. An account read is a network round trip and signing out is a keystroke.

## Review history worth reading

Five findings in the first review round and four in the second were real; all are fixed. Two of my own fixes were wrong and the tests caught them:

- My first corpus regression test **passed against its bug** — the corpus replaces a row it already holds and both reads returned the same id, so it could not tell a survivor from a re-fetch. Distinct ids make the difference a row count.
- I reported the cache save/clear race fixed one round before it was. The epoch check sat before `await store()`, and **actors are reentrant**: that await is a suspension point inside the actor, so a `clear()` arriving while a save is parked there runs to completion and the save resumes past a check it had already passed. The handle is now acquired first and the epoch checked after, so check-to-write is synchronous.

Every fence was verified by removing it and confirming its test fails — including **both** directions of the prune bound, since a test that only proves deletion happens would pass while it deleted the whole account.

**Not changed, deliberately:** `locallySourced` excluding the whole memories source from caching. After `ActivityLocalMemories` merges there is no per-row provenance, so caching "only the account's rows" needs the decorator to keep them separate — and the users it would help already read memories from `omi.db` locally on every launch. Caching them would mean a second copy of another app's table that a later launch reads back as *the account's* answer.

## Verification

```
swift test   # 1581 tests, 7 skipped, 0 failures — three consecutive clean runs
```

New coverage: `ActivityRevalidationTests` (14), `ActivityAccountCacheTests` (13), `AccountCacheQueriesTests` (9), `OmiActivityFeedTests` (+5).

**Test rewrite disclosure.** `StoreTests`' two migration-ledger enumerations gained `v11-account-cache`. They assert the exact set of identifiers `ContextStore.makeMigrator()` registers; the new value is read off the migration in this PR rather than chosen, and a missing entry means the migration re-runs on every launch and throws on the second — which is what those assertions exist to catch.

**Live UI verification is not available for this app from a local build, and I am saying so rather than implying a run happened.** A dev-signed build under the shipped bundle id rewrites both of the user's Claude MCP configs at startup and writes TCC records shared with the installed app. Coverage here is behavioural; the notarized build is what exercises it end to end.

None — `scripts/pr-preflight --suggest` reports no invariant globs matched by this diff.

## Failure class (fixes)

Failure-Class: new

`FC-durable-state-scoped-to-view-lifetime`. State whose cost to rebuild exceeds the lifetime of the view that displays it must be owned by something that outlives the view. The guard surface is `ActivitySpine` plus `ActivityRevalidationTests`, which asserts the three properties a re-entry must have: the list does not blink, a second appearance revalidates rather than re-pages, and the account half is dropped on sign-out.

`evidence_prs` cites #11641, the PR that shipped the app carrying the defective contract.

### DIFF
--- a/.github/failure-classes/FC-durable-state-scoped-to-view-lifetime.json
+++ b/.github/failure-classes/FC-durable-state-scoped-to-view-lifetime.json
@@ -1,0 +1,18 @@
+{
+  "schema_version": 1,
+  "id": "FC-durable-state-scoped-to-view-lifetime",
+  "violated_contract": "State whose cost to rebuild exceeds the lifetime of the view that displays it must be owned by something that outlives the view. A corpus paged in over dozens of network requests, held as a SwiftUI @StateObject inside the surface that draws it, is discarded and re-derived on every ordinary re-entry — a window rebuilt on each present(), a branch of a switch the user leaves by typing — and the surface reads as slow for work it has already done. The owning scope is a property of what the state costs, not of who renders it.",
+  "canonical_prevention": "Own the store at the process (ActivitySpine.shared), hand it to the view, and make re-entry a revalidation that keeps every composed row and asks only what changed — on a cooldown, never clearing, never re-opening the window. State that now outlives a sign-out must be told to forget it, and the paint before the first answer comes from a bounded on-disk copy that is explicitly not an authority. Regression coverage in ActivityRevalidationTests asserts the list does not blink mid-request, that a second appearance revalidates rather than re-pages, and that the sign-out edge clears the account half.",
+  "canonical_prevention_artifact": [
+    "desktop/context-for-claude/Sources/ContextApp/Activity/ActivitySpine.swift",
+    "desktop/context-for-claude/Tests/ContextAppTests/ActivityRevalidationTests.swift"
+  ],
+  "evidence_prs": [
+    11641
+  ],
+  "scope_hints": [
+    "desktop/context-for-claude/**",
+    "desktop/macos/**"
+  ],
+  "status": "open"
+}
--- a/desktop/context-for-claude/ARCHITECTURE.md
+++ b/desktop/context-for-claude/ARCHITECTURE.md
@@ -88,7 +88,7 @@ live, a dedicated transcription gap reason is published the same way.
 `~/Library/Application Support/ContextForClaude/`
 
 ```
-context.db              sessions, segments, frames + two FTS5 indexes
+context.db              sessions, segments, frames + two FTS5 indexes, and the account cache
 Frames/YYYY-MM-DD/      screen JPEGs, pruned after 30 days
 capture-state.json      heartbeat: capturing / paused reason / capabilities
 last-query.json         the MCP server's proof it served a tool call: tool name + time, nothing else
@@ -104,6 +104,13 @@ staged. Neither ever holds anything the user said, saw, or asked — see `QueryS
 A session is a contiguous run of speech; a gap over five minutes starts a new one
 (`SessionPolicy`). Transcripts are kept forever; frames are the expensive part and are the only
 thing pruned.
+
+`account_rows` is the one table in here that is **a copy and not an authority**: the last answer the
+user's Omi account gave, bounded to one page per source, so the Activity panel paints real titles on
+a cold launch instead of the untitled local sessions it used to show while three HTTP round trips
+were in flight. Every row came off the wire and is rewritten by the wire; nothing reads it once a
+real answer exists, and signing out empties it. See `AccountCacheQueries` and
+`ActivityAccountCache`.
 
 ## The MCP surface
 
@@ -142,6 +149,9 @@ Sources/ContextApp/      ContextApp  Engine  Permissions
                          Onboarding/ Ink  Backdrop  RandomizedText  OnboardingWindow  OnboardingView
                          Integration/ClaudeRegistrar  LoginItem
                          Shortcuts/  GlobalShortcuts  ShortcutConflicts
+                         Activity/   ActivitySpine (the one process-lived store)  ActivityStore
+                                     ActivityComposer  ActivityAccount  ActivityAccountCache
+                                     ActivityLocalMemories  ActivitySurface  ActivityStream
                          Search/     SearchBarWindow  SearchBarView  SearchSurface
                                      SearchResultsModel  SearchResultsView  SearchRanking
                                      ClaudeRouter (tutorial + settings only, not the bar)
--- a/desktop/context-for-claude/Sources/ContextApp/Activity/ActivityAccount.swift
+++ b/desktop/context-for-claude/Sources/ContextApp/Activity/ActivityAccount.swift
@@ -197,6 +197,31 @@ protocol ActivityAccountReading: Sendable {
     func read(since: Double?, until: Double?, limit: Int) async -> ActivityAccountFeed
 }
 
+/// An account reader whose walk through the account can be moved by the caller.
+///
+/// **Separate from `ActivityAccountReading` for the reason `ActivityAccountDiagnosing` is**: a
+/// preview's account, a test's account and `ActivityAccountAbsent` have no cursor and nothing to
+/// move, and making every one of them implement two no-ops would be a protocol demanding a promise
+/// only one implementation can keep.
+///
+/// It exists because a reader that pages is a reader that only ever moves *away* from the newest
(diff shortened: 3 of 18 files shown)


Archit-lal:
Thanks — the three P1s were one real defect I introduced, and both P2s were valid. All five addressed in 7a65839.

**The three P1s (sign-out races).** Correct diagnosis: I wrote sign-out as a series of resets with no epoch, so everything already in flight still believed it belonged. An account read is a network round trip and signing out is a keystroke — the window is not narrow.

- *Store* — `forgetTheAccount()` now advances `generation`. Every landing path was already fenced on it, so one increment invalidates all of them. It advances **without** re-opening the window: frames and speech on this Mac are not account-scoped.
- *Cache* — now an `actor` with a write epoch. The actor makes the save and the clear orderable; the epoch decides *which account the write describes*, which is the right question — a save can arrive second, be perfectly serialised, and still be wrong. The epoch advances on intent, so a delete that cannot reach the database still stops the writes behind it.
- *Corpus* — the epoch is captured before a request leaves and quoted by every `commit`.

**P2 (tie ordering)** — valid. `at` is not unique here: a date-only task is stamped 11:59 PM like every other one, and a backfilled batch of memories shares an instant. `recent` and the trim subquery now both break ties on `id DESC`, so the page a launch paints can't disagree with the page the trim kept.

**P2 (fixed 60 ms sleep)** — valid, and it violated this repo's own hermetic-test rule. Replaced with `waitUntil`, which returns the instant the work lands and spends its deadline only when something is genuinely broken.

**On the regression tests.** Each of the three fences has one, and each was verified by removing its fence and confirming the test fails. That check earned its keep: my first corpus test *passed against the bug*, because the corpus replaces a row it already holds and both reads returned the same id, so it could not tell a survivor from a re-fetch. Distinct ids make the difference a row count.

`swift test` — 1574 tests, 0 failures, three consecutive clean runs.

Archit-lal:
Second round addressed in c305ba5. Four more real ones — thank you, two of these were worse than the originals.

- **Offline launch was a coin flip.** The seed paints from disk, the failing read lands carrying nothing, and taking it wholesale wiped what the user had just been shown. Whichever landed first decided it, so the panel was empty about half the time — on the one launch where the cache *is* the answer. A source that said nothing **and** handed over nothing now keeps what is on screen.

  Worth recording: my first attempt retained on `answered` alone and was wrong. The paging reader answers with everything it has gathered, so a source absent from `answered` routinely carries hundreds of real rows fetched earlier — the documented shape of a page that did not arrive, not a stale leftover. The existing spine tests caught it on the first run. Both halves of the condition are now required, and an *answered* source is taken as it comes, empty included.

- **A revalidation could only add.** Correct, and it's a regression the process-lived store introduced: a panel rebuilt per window used to lose a deleted row simply by being rebuilt. A head re-read is now authoritative **for its own range** — a row newer than the page's oldest returned row and absent from it has gone; everything behind the page is untouched, because a head page is evidence about nothing older. Memories that had to start past their own head never prune, since that page is incomplete at the top by construction and deleting there would read "not fetched" as "deleted".

- **`readFailure` outlived the failure** — `openWindow` was the only thing clearing it, which was enough until a revalidation started running a second opening read.

- **Single-flight slot didn't cover the reopen** — the intent is now a parameter of `readAccount`, so the sequence sits behind the guard that already existed.

Not changing one thing, and saying so rather than quietly leaving it: the `locallySourced` exclusion dropping the whole memories source from caching is deliberate. Once `ActivityLocalMemories` has merged, the feed carries no per-row provenance, so caching "only the account's rows" would need the decorator to keep them separate — and the users it would help are exactly the users who already have Omi installed and get their memories from `omi.db` locally on every launch. Caching them here would mean a second copy of another app's table that a later launch reads back as *the account's* answer. Happy to revisit if you disagree.

Each fix has a regression test verified by removing the fix — including **both** directions of the prune bound, since a test that only proves deletion happens would pass while it deleted the whole account.

`swift test` — 1580 tests, 0 failures, three consecutive clean runs.

Archit-lal:
**Correction: I reported the cache save/clear P1 as fixed in 7a65839 and it was not.** You were right to leave the thread open. Genuinely fixed in c8575c7.

The epoch check sat *before* `await store()`, and **actors are reentrant** — that await is a suspension point inside the actor, so a `clear()` arriving while a save is parked there runs to completion and the save then resumes past a check it had already passed. Serialising through an actor makes calls orderable; it does not make a single call atomic across its own awaits. My previous reply reasoned about ordering and missed reentrancy entirely.

The store handle is now acquired first and the epoch checked after it, so everything from the check to the last write is synchronous. `clear()` bumps before its own await for the same reason. The regression test parks a save inside the provider and reenters with a clear, and **fails against the previous order** — which is how I confirmed the hole rather than arguing about it.

**Fixed sleeps** — also right, and the conversion surfaced something. Waiting on first paint wasn't enough: the opening read is followed by hydration's own reads, so a baseline captured mid-walk had the revalidation refused by the single-flight guard. Those tests were green while measuring the wrong thing. Baselines now wait on `corpusSettled`. Five `settle()` calls remain, all on assertions about something *not* happening, each behind a positive wait. Suite time went 0.81s → 0.23s.

**`locallySourced`** — still declining, reasoning in the previous comment. Short version: after the merge the feed has no per-row provenance, and the users it would help already read their memories from `omi.db` locally on every launch.

`swift test` — 1581 tests, 0 failures, three consecutive clean runs.

PR #11721, opened 2026-08-17
fix(desktop): repair the shell monitor retain cycle and unbreak swift test

undivisible:
## Why

Both of these made local desktop development impossible on Xcode 27, and one of them is a real memory leak that has been shipping.

## 1. Shell mouse-monitor retain cycle

```swift
let scheduleReconciliation: @Sendable () -> Void = {
  DispatchQueue.main.async { [weak self] in self?.sync() }   // weak on the wrong closure
}
```

`ShellMouseInterceptionSync` stores its event monitors in `self.monitors`, and those handlers call `scheduleReconciliation` — which captured `self` **strongly**. The `[weak self]` applied to the inner dispatch block, not to the stored capture, so the cycle was never broken:

`self → monitors → handler → closure → self`

Every shell window leaked its interception sync for the process lifetime. Moving the capture list to the outer closure makes the weak capture the one that actually governs the stored reference.

Swift 6.4 reports this as `ImplicitStrongCapture`. Because the package builds with `-warnings-as-errors`, that one line fails the entire debug build of `main` on Xcode 27. Earlier toolchains accepted it silently — which is exactly how a real leak sat behind a diagnostic nobody could see.

## 2. Test bundles missing an rpath

`swift test` could not load **any** xctest bundle:

```
Library not loaded: @rpath/Sparkle.framework/Versions/B/Sparkle
```

Generated rpaths cover `Products/<config>/PackageFrameworks/`, but Sparkle and friends are built into `Products/<config>/` itself — one level up. One unloadable bundle aborts the whole run, so a single target's missing rpath hid every test in the package, and it surfaces as a test failure rather than a link error.

Every test target now carries `@loader_path/../../..`. Applied uniformly rather than working out which targets link Sparkle transitively — a search path costs nothing on the ones that don't need it.

## Verification

```
xcrun swift build -c debug --package-path Desktop     Build complete
  (-warnings-as-errors intact; fails on main without fix 1)

xcrun swift test --package-path Desktop               5272 tests, 1 skipped, 0 failures
  (on main this cannot load a single bundle)

swift-format lint                                     clean
```

The full desktop suite runs locally, which it did not before this PR.

## Note

No test is added for the retain cycle. A leak check would need a teardown harness this package doesn't have, and the compiler now rejects the pattern outright under `-warnings-as-errors` — which is a stronger guard than a test, as long as the flag stays. Found while working on #11657, where both problems forced me to verify desktop changes through workarounds.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11721?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Low Risk**
> Targeted linker and capture-list fixes for tests and shell click-through; no auth, data, or broad behavioral refactors.
> 
> **Overview**
> Fixes two desktop/macOS issues that blocked **Xcode 27** local builds and **`swift test`**.
> 
> **`ShellMouseInterceptionSync`** now applies **`[weak self]`** on the stored `scheduleReconciliation` closure instead of only on the inner `DispatchQueue.main.async` block, breaking **`self → monitors → handler → closure → self`** so each shell window’s mouse helper can be released instead of leaking for the process lifetime (also clears Swift 6.4 **`ImplicitStrongCapture`** under **`-warnings-as-errors`**).
> 
> **`Package.swift`** adds a shared linker **rpath** (`@loader_path/../../..`) on every test target so xctest bundles can **dlopen** frameworks like **Sparkle** from `Products/<config>/`, not only `PackageFrameworks/`.
> 
> Changelog notes the shell mouse-tracking leak fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb028f9549fc822814f9900558b86f22cd70dafd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/desktop/macos/Desktop/Package.swift
+++ b/desktop/macos/Desktop/Package.swift
@@ -1,6 +1,15 @@
 // swift-tools-version: 6.0
 import PackageDescription
 
+// Frameworks such as Sparkle are built into Products/<config>/, but a test bundle's
+// generated rpaths only cover PackageFrameworks/. Without this the bundle builds and
+// then fails to dlopen, which reads as an unrelated test failure.
+// ponytail: one rpath on every test target, rather than working out which ones link
+// Sparkle transitively.
+let testBundleFrameworkSearchPath = LinkerSetting.unsafeFlags([
+  "-Xlinker", "-rpath", "-Xlinker", "@loader_path/../../..",
+])
+
 let package = Package(
   name: "Omi Computer",
   platforms: [
@@ -129,7 +138,8 @@ let package = Package(
       ],
       swiftSettings: [
         .unsafeFlags(["-strict-concurrency=complete", "-warnings-as-errors"])
-      ]
+      ],
+      linkerSettings: [testBundleFrameworkSearchPath]
     ),
     .testTarget(
       name: "OmiSupportTests",
@@ -137,7 +147,8 @@ let package = Package(
       path: "Tests/OmiSupportTests",
       swiftSettings: [
         .unsafeFlags(["-strict-concurrency=complete", "-warnings-as-errors"])
-      ]
+      ],
+      linkerSettings: [testBundleFrameworkSearchPath]
     ),
     .testTarget(
       name: "OmiWALTests",
@@ -145,7 +156,8 @@ let package = Package(
       path: "Tests/OmiWALTests",
       swiftSettings: [
         .unsafeFlags(["-strict-concurrency=complete", "-warnings-as-errors"])
-      ]
+      ],
+      linkerSettings: [testBundleFrameworkSearchPath]
     ),
     .testTarget(
       name: "VoiceTurnDomainTests",
@@ -153,7 +165,8 @@ let package = Package(
         .target(name: "Omi Computer"),
         "VoiceTurnDomain",
       ],
-      path: "Tests/VoiceTurnDomainTests"
+      path: "Tests/VoiceTurnDomainTests",
+      linkerSettings: [testBundleFrameworkSearchPath]
     ),
     .testTarget(
       name: "SemanticFeatureSentinels",
@@ -161,7 +174,8 @@ let package = Package(
       path: "Tests/SemanticFeatureSentinels",
       swiftSettings: [
         .unsafeFlags(["-strict-concurrency=complete"])
-      ]
+      ],
+      linkerSettings: [testBundleFrameworkSearchPath]
     ),
   ],
   swiftLanguageModes: [.v6]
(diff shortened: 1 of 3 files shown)


Git-on-my-level (changes requested):
Verified both fixes on this head — genuinely useful PR: the shell-monitor leak has been shipping, and `swift test` was fully broken.

**desktop/macos/Desktop/Sources/MainWindow/ShellClickThrough.swift** — the capture-list move is correct. Previously `[weak self]` sat on the inner `DispatchQueue.main.async` block while the outer `scheduleReconciliation` closure — the reference actually stored through the NSEvent monitors in `self.monitors` — captured `self` strongly, so `self → monitors → handler → closure → self` never broke and every shell window leaked its `ShellMouseInterceptionSync` for the process lifetime. With `[weak self]` on the outer `@Sendable` closure the stored reference is now weak; a pending `sync()` after dealloc is a no-op, which is the intended semantics. Behavior is otherwise unchanged.

**desktop/macos/Desktop/Package.swift** — all five test targets (`Omi ComputerTests`, `OmiSupportTests`, `OmiWALTests`, `VoiceTurnDomainTests`, `SemanticFeatureSentinels`) now carry the shared `testBundleFrameworkSearchPath` (`@loader_path/../../..`). Applying it uniformly rather than tracing which targets transitively link Sparkle is the right call — an unused search path costs nothing. The green *Desktop Swift Build & Tests* run on this head corroborates the local `swift test` result, and the `ponytail:` note matches the repo's existing comment convention.

One blocking item before merge — both red checks (*Hygiene*, *PR Metadata Preflight*) fail on the same deterministic gate, `desktop-changelog-entry`:

- Desktop file changes require a one-line user-facing JSON fragment under `desktop/macos/changelog/unreleased/` (the directory currently holds only `.gitkeep`). Since the retain-cycle fix repairs shipped app behavior, a line like "Fixed a memory leak in the desktop shell window's mouse handling" fits the existing release-fragment style. The gate's alternative is the `no-changelog-needed` label, but that is meant for internal-only changes, and the leak fix doesn't read as internal-only.

Skipping a dedicated leak regression test is reasonable here: with `-warnings-as-errors` the toolchain now rejects the strong-capture pattern outright, which is the stronger guard as long as the flag stays on.

Once the fragment lands and both preflight checks re-run green, this is a straightforward maintainer merge — the retain-cycle fix is real and worth shipping.

---

kodjima33 (approved):
real retain-cycle fix (capture list moved to the outer closure) plus the test-bundle rpath that unbreaks swift test

PR #11727, opened 2026-08-17
fix: restore conversation deletion

undivisible:
## Summary

- Let conversation deletion retract existing canonical memory while canonical intake is paused.
- Keep failed mobile and macOS deletions visible and recoverable instead of presenting them as success.

## Root cause

The deletion path used the same intake fence as canonical memory creation, so conversations with existing canonical memory could not be deleted while intake was paused.

## Validation

- `backend/.venv/bin/pytest backend/tests/unit/test_ws_j_delete_privacy.py backend/tests/unit/test_memory_apply_store.py -q`
- `backend/.venv/bin/python backend/scripts/scan_async_blockers.py`
- `backend/scripts/typecheck.sh`
- `bash app/test.sh`
- Focused macOS deletion XCTest, Swift formatting, and test-quality checks
- `make preflight`

## Product invariant

- `INV-MEM-1`: preserves the canonical memory collection and existing tier policy while allowing only zero-write retractions.

Failure-Class: none

Line-Count-Exception: backend/database/memory_apply_store.py | 1932 -> 1961 | A dedicated transaction preserves the paused-intake boundary for ordinary writes while allowing only zero-write retraction.
Line-Count-Exception: backend/utils/memory/canonical_memory_adapter.py | 2388 -> 2406 | The adapter selects the dedicated zero-write retraction path without changing normal canonical replacement behavior.
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift | 3468 -> 3482 | The existing page owns optimistic deletion state, restoration, and its user-visible failure alert.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11727?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #11687, opened 2026-08-16
fix(backend): declare opposite-direction Firestore list-scan indexes (#11684)

aryanorastar:
## Summary
- Fixes the root-cause registry bug behind prod `GET /v3/memories` → **503 Canonical memory unavailable** (#11684): `updated_at DESC, __name__ ASC` is not an automatic Firestore index, but `INDEX_REQUIREMENTS` dropped every single-field+`__name__` spec.
- Narrow the manifest predicate: opposite-direction `__name__` tie-breaks are real composites; same-direction singles stay out (Firestore rejects them as redundant).
- Regenerates `firestore.indexes.json` with the four missing indexes (`memory_items` updated_at, `memories` updated_at/created_at, `memory_review_queue` status).

## Merge sequencing (required)
**Do not merge this PR until the four indexes exist in prod** (and preferably auto-dev).

`firestore_readiness` runs `reconcile_firestore_indexes.py --check-only` on backend deploy. Merging the manifest *before* APPLY blocks deploys until the indexes are live.

Correct order (from #11684):
1. Human: `Reconcile Firestore Indexes` → `prod` → `APPLY_FIRESTORE_INDEXES` (stops the 503s even before this PR).
2. Then merge this PR so the registry matches the deployed schema and regressions stay pinned.

## Test plan
- [x] `pytest tests/unit/test_firestore_index_config.py test_firestore_indexes.py test_reconcile_firestore_indexes.py test_firestore_query_contract.py` → 61 passed
- [x] `python3 backend/scripts/generate_firestore_indexes.py` verify clean after `--write`
- [ ] Maintainer APPLY of the four indexes in prod / auto-dev before merge

## Honest gaps
- Author cannot run `APPLY_FIRESTORE_INDEXES` (prod environment reviewers). Live 503 stop depends on that apply; this PR locks the declaration + tests.

Failure-Class: none

Product invariants affected
INV-MEM-1

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #11685, opened 2026-08-16
One ungrounded memory candidate aborts the whole conversation: ~1 in 3 lazy enrichments fails on 'evidence without a unique source binding'


kodjima33:
Canonical memory extraction aborts the whole conversation when **one** L1 candidate fails quote grounding, so conversation finalization dies with it. In prod this is the single largest cause of failed enrichment right now.

**Symptom (prod, `based-hardware`, 2026-08-16 03:00Z → 15:30Z):**

```
ERROR:routers.conversations:lazy enrich failed uid=<redacted> conv=<redacted>:
  canonical memory extraction returned evidence without a unique source binding
ERROR:utils.sync.pipeline:sync: failed to reprocess merged conversation <redacted>:
  canonical memory extraction returned evidence without a unique source binding
```

- `lazy enrich` outcomes in that window: **42 complete / 39 failed**, and **27 of the 39 failures** are this one error. So ~1 in 3 of all lazy enrichments dies here.
- Same error also fires continuously on `backend-sync` (`_reprocess_merged_conversations`), where it leaves the merged conversation's summary/structured/speakers stale.
- Meanwhile extraction is *not* systematically broken: in the same window 332 conversations saved ≥1 canonical memory (209 of them saved ≥2) and 468 extracted none. Grounding succeeds routinely — it fails per candidate.

**Root cause** — `backend/utils/conversations/process_conversation.py:912-918`:

```python
for candidate in extracted_candidates:
    evidence_quotes = _grounded_l1_evidence_quotes(candidate.evidence_quotes, conversation.transcript_segments)
    if not evidence_quotes:
        raise ValueError("canonical memory extraction returned evidence without a unique source binding")
```

`_grounded_l1_evidence_quotes` returns `[]` when any one quote of that candidate does not match **exactly one** transcript segment — 0 matches (the model paraphrased, or quoted across a segment boundary) or 2+ matches (a short phrase repeated in two segments). That is a per-candidate trust verdict, but it is raised as a whole-run failure.

**Blast radius of the raise.** `_extract_memories` is called synchronously near the end of `process_conversation` (`:1532`), so the exception takes down everything after it as well:

- the run's other, correctly-grounded candidates are discarded;
- `_save_action_items` and `_update_goal_progress` never run;
- the lazy-enrich handler (`routers/conversations.py:129-135`) marks the conversation `deferred: True` again — no summary for the user until they reopen it;
[...]

PR #11731, opened 2026-08-17
fix(backend): declare opposite-direction Firestore list-scan indexes + converge cascade-retract CAS (#11684, #11726)

### LINKED ISSUE #11684: GET /v3/memories 503s in prod: the canonical scan orders updated_at DESC + __name__ ASC, an index Firestore never builds automatically and the registry never declares
`GET /v3/memories` has been answering **503 "Canonical memory unavailable"** in prod since the canonical scan shipped on **2026-08-12** (efba1a05b5, "fix: finish combined proactive QA hardening"). The canonical list scan asks Firestore for an index that has never been declared, so every page-1 read fails at query planning.

## Evidence (prod, `based-hardware`)

`gcloud logging read severity>=ERROR` over the last 24h: **915 of a 1000-record sample** are this one traceback.

```
File "/app/utils/memory/memory_service.py", line 1105, in _ensure_slots
  raw_slots, stream_exhausted = read_canonical_scan_page(
File "/app/utils/memory/canonical_memory_adapter.py", line 623, in read_canonical_scan_page
  snapshots = list(query.limit(bounded_limit).stream())
google.api_core.exceptions.FailedPrecondition: 400 The query requires an index.
```

The remediation link in that error decodes to exactly:
`memory_items` COLLECTION — `updated_at DESCENDING, __name__ ASCENDING`.

`memory_service.py:1116` maps that exception to `HTTPException(503, "Canonical memory unavailable")`, so users of `GET /v3/memories` (`routers/memories.py:513`) get a 503, not a degraded list.

## Root cause

`read_canonical_scan_page` orders `updated_at DESC, __name__ ASC` (`canonical_memory_adapter.py:612`). That tie-break direction is deliberate and correct — `MemoryService.keyset_sort_key` is `(-updated_at_us, memory_id)` (`memory_service.py:1597`), so the whole service's public order is updated_at DESC with **ascending** document id. The emitted-keyset monotonicity filter (`memory_service.py:1153`) and the canonical/historical merge (`:1725`) both depend on the Firestore scan matching that order. Flipping `__name__` to DESC would be the wrong fix.

The index is missing because the registry believes Firestore provides it automatically:

```python
# backend/database/firestore_index_registry.py:332
# Firestore manages the single-field updated_at index (plus automatic __name__
# tie-break) itself — this spec records the serving query contract only.
```

Firestore's automatic single-field indexes serve `field ASC, __name__ ASC` and `field DESC, __name__ DESC` only. `field DESC, __name__ ASC` is a *different* index that is never built automatically. `INDEX_REQUIREMENTS` (`:600`) drops any spec with one non-`__name__` field from the manifest, so four specs silently never reach `firestore.indexes.json`:

| spec | index | live in prod? |
|---|---|---|
| `memory_items_universal_list_scan` | `memory_items` updated_at DESC, `__name__` ASC | **missing — this outage** |
| `memories_universal_list_scan_updated_at` | `memories` updated_at DESC, `__name__` ASC | missing |
| `memories_universal_list_scan_created_at` | `memories` created_at DESC, `__name__` ASC | missing |
| `memory_review_queue_by_status_id` | `memory_review_queue` status ASC, `__name__` DESC | missing |

Verified against `gcloud firestore indexes composite list --project based-hardware` (89 indexes): none of the four exist.

`test_firestore_index_config.py:30` encodes the same wrong assumption ("rejected at deploy time as redundant"), which is why no test caught it.

## The assumption is wrong — measured, not argued

I created the exact index in **`based-hardware-dev`** and Firestore accepted it (it is not rejected as redundant), then A/B'd the real serving query:

```
# index CREATING
FAILED_PRECONDITION based-hardware-dev: 400 ... That index is currently building and cannot be used yet
# index READY
OK based-hardware-dev: query planned and ran
# prod control, index still absent
FAILED_PRECONDITION based-hardware: 400 The query requires an index.
```

Firestore itself matched my created index to the query — confirming both the shape and that it is not automatic.

## Fix

Narrow the manifest predicate to what Firestore actually rejects: a single ordered field is automatic only when `__name__` tie-breaks in the *same* direction. Patch (registry + regenerated `firestore.indexes.json` + the two tests), 4 files / +122 / -13, is ready and green:

- `backend/tests/unit/test_firestore_indexes.py`, `test_firestore_index_config.py`, `test_reconcile_firestore_indexes.py` — 34 passed
- `backend/tests/unit/test_firestore_query_contract.py` — 26 passed
- `backend/scripts/generate_firestore_indexes.py` verify — clean; adds exactly the 4 indexes above, removes none

## Why this is not a drive-by PR — sequencing needs a human

`firestore_readiness` runs `reconcile_firestore_indexes.py --check-only` in both `gcp_backend.yml` and `gcp_backend_auto_dev.yml`. **Merging the manifest change before the indexes exist blocks backend deploys** (prod and auto-dev) until the create-only migration runs. The correct order is:

1. Apply the four indexes (`Reconcile Firestore Indexes` → `prod` → `APPLY_FIRESTORE_INDEXES`). This alone stops the 503s — the serving code needs no change.
2. Then merge the registry/manifest patch so the declaration matches the deployed schema and the regression tests hold.

That workflow is `workflow_dispatch` with an explicit confirmation string and the `prod` environment requires reviewers — deliberately a human-approved writer — and `AGENTS.md` puts schema migrations in the explicit-sign-off bucket. So this is filed rather than pushed.

## Follow-ups

- `validate_account_deletion_dispatch_configuration`-style gap: nothing asserts a declared serving query is actually servable. A startup or readiness probe that plans each `QUERY_SPECS` entry once would have caught this on the deploy that shipped it.
- Separate live error, same subsystem, 29 occurrences/24h: `ValueError: canonical memory extraction returned evidence without a unique source binding` (`process_conversation.py:918`) escaping `_run_enrichment`.

🤖 filed by the hourly desktop bug watchdog

### LINKED ISSUE #11726: Concurrent cascade conversation deletes 500: same-uid retract races the canonical memory-control CAS
## Bug Description

Concurrent Flutter `DELETE /v1/conversations/{id}?cascade=true` calls for the **same account** race the account-global canonical memory-control CAS inside `replace_conversation_sourced_memories`. After three `ConversationSourceReplacementConflict` retries the handler raises `RuntimeError: canonical conversation replacement conflicted repeatedly` and the client sees **HTTP 500**. Client retries (up to 12× on the same conversation id in this incident) amplify the burst.

This is **not** the missing Firestore list-scan index. That 503 class is #11684 / PR #11687 (APPLY indexes first). Do not put index declaration or `APPLY_FIRESTORE_INDEXES` sequencing in this PR.

## Incident evidence (prod, read-only)

- Source: Omi prod anomaly watchdog `75afe342f6f8` at **2026-08-17T03:27:27Z** (recovered 03:43–03:59Z).
- Serving: Cloud Run service `backend`, revision `backend-memon0815`, commit `6316415534fffc4636efdab224a49a3638cf046f` (merged #11627), 100% traffic. Project `based-hardware`.
- Alert-window HTTP 5xx (03:12–03:28Z, 345 requests): **39** `DELETE ...?cascade=true` → **500** (mostly 1–5s). Same window also had 268 DELETE **503**s from the #11684 index hole — out of scope here.
- Earlier tick 02:56–03:12Z: 38 DELETE 500 + 303 DELETE 503; 151 unique conversation ids.
- Dart/Flutter `userAgent` family only. Same conversation ids retried (max 12).
- App traces in 03:16–03:19Z traceback sample:
  - `routers/conversations.py` `delete_conversation`
  - `utils/memory/memory_service.py` `retract_conversation_memories`
  - `utils/memory/canonical_memory_adapter.py` `retract_conversation_sourced_memories` / `replace_conversation_sourced_memories`
  - `database/memory_apply_store.py` `replace_conversation_source_firestore`
  - `RuntimeError: canonical conversation replacement conflicted repeatedly`
  - `ConversationSourceReplacementConflict: memory control changed during conversation replacement`

No user ids, conversation ids, or payloads in this issue.

## Steps to Reproduce

1. Use an account with several conversations that still have live canonical (or historical) memories so `retraction_can_be_skipped` is false.
2. Issue **parallel** `DELETE /v1/conversations/{id}?cascade=true` for many of those conversations (same uid). Optionally retry 5xxs immediately, as the mobile client did.
3. Observe 500s and the RuntimeError / ReplacementConflict traces above.

Serial deletes alone will under-reproduce.

## Expected Behavior

- Parallel cascade deletes for one uid do not return 5xx.
- Retracting an already-retracted / empty source is idempotent (2xx).
- Memory-control CAS conflicts retry or serialize until the retract converges; they must not escape as an unhandled 500.
- Required retraction still happens. Do not skip retract just to avoid conflicts, and do not swallow conflicts that would leave live memories pointing at a deleted conversation.
- Repeating the same DELETE is safe.

## Actual Behavior

Concurrent same-uid cascade deletes enter `replace_conversation_sourced_memories`, which plans against account-global `source_generation` / head commit and retries only 3 times. Overlapping retracts invalidate each other's control snapshot → `ConversationSourceReplacementConflict` → `RuntimeError` → HTTP 500. Flutter retries make it a 5xx storm.

## Environment

- OS / runtime: Cloud Run `backend` prod (`us-central1`)
- Version/Commit: `6316415534fffc4636efdab224a49a3638cf046f` (`backend-memon0815`)
- Later `main` may differ; verify the retract/CAS path on the serving SHA **and** current `origin/main` before coding. Do not assume #11685 (`5af7e53fd7`, unique source binding) changed this path — it did not.

## Error Output

```
File "/app/routers/conversations.py", line 795, in delete_conversation
File "/app/utils/memory/memory_service.py", line 2347, in retract_conversation_memories
File "/app/utils/memory/canonical_memory_adapter.py", line 2161, in retract_conversation_sourced_memories
File "/app/utils/memory/canonical_memory_adapter.py", line 1424, in replace_conversation_sourced_memories
RuntimeError: canonical conversation replacement conflicted repeatedly
```

```
ConversationSourceReplacementConflict: memory control changed during conversation replacement
```

## Root cause (for the fix PR)

`replace_conversation_sourced_memories` is the account-global replace/retract boundary. Cascade delete calls it once per conversation. Concurrent deletes for one uid race the control CAS. The HTTP layer does not serialize per uid, does not treat “already retracted” as success, and maps exhausted conflicts to 500.

#11627 (on the serving SHA) made fenced no-op retract skippable. It did not make required concurrent retracts converge.

## Proposed fix boundary (keep narrow)

Fix concurrency/idempotency on the cascade-retract path only. Do **not** redesign the memory-control model.

Likely shapes (pick the smallest that passes the tests):

- Per-uid serialization around conversation source replacement / cascade retract; and/or
- Treat an already-empty source set as a successful retract after a control-change conflict; and/or
- Map remaining retryable conflicts to a bounded 409/503 with idempotent retry, never an unhandled 500 — only if retract is still guaranteed to complete.

Independently re-read `replace_conversation_sourced_memories`, `replace_conversation_source_firestore`, and `delete_conversation` on current `origin/main` before editing.

## Acceptance tests (required)

Behavioral, same-uid, **parallel**:

1. N concurrent `DELETE ?cascade=true` for distinct conversations on one uid → no 5xx; conversations gone; no live memories left pointing at those sources.
2. Repeat DELETE of an already-deleted / already-retracted conversation → success, not 500.
3. Injected/forced `ConversationSourceReplacementConflict` during retract either converges or fails closed **without** deleting the conversation while live memories remain, and without an unhandled 500.
4. Serial cascade delete still retracts required memories (do not regress #11627’s “never orphan live memories”).

## Explicit non-goals

- #11684 / PR #11687 Firestore index APPLY / registry predicate (`updated_at DESC, __name__ ASC`). That owns the DELETE/GET **503**s.
- #11685 / PR #11688 `evidence without a unique source binding` on conversation processing.
- Changing `MEMORY_MODE`, cohort enrollment, or the canonical control schema.
- Client-only retry backoff as the sole fix.

## Suggested labels

`bug`, `backend`, `stability`

## Related

- #11684 — same incident window; 503/index class
- #11627 — serving SHA; unbroke fenced no-op cascade delete
- #11685 — separate extract abort; already merged, not this 500

Git-on-my-level:
## What changed and why

Two production 5xx classes from the 2026-08-17 incident window (watchdog `75afe342f6f8`, serving SHA `6316415` = merged #11627), fixed in two independent commits:

1. **Missing opposite-direction Firestore list-scan indexes → 503 (#11684).** Serving queries order `updated_at DESC, __name__ ASC` (`read_canonical_scan_page`, plus the historical dual-stream scans and the review-queue status scan). Firestore automatic single-field indexes only cover `field+__name__` in the *same* direction, so these need real composites — but `_query_spec_index_requirements()` dropped every spec with ≤1 non-`__name__` field. Prod answered `FailedPrecondition: The query requires an index`, mapped to 503 `Canonical memory unavailable` / `Historical memory unavailable` on `GET /v3/memories` and the cascade-delete skip-guard scan. Fix: the predicate now declares exactly the opposite-direction pairs and regenerates the manifest with the four missing indexes (`memory_items` `updated_at DESC`+`__name__ ASC`; `memories` `updated_at DESC`+`__name__ ASC`; `memories` `created_at DESC`+`__name__ ASC`; `memory_review_queue` `status ASC`+`__name__ DESC`). Same-direction singles and array-contains pairs stay out (deploy rejects those as redundant).

2. **Cascade-retract CAS exhaustion → 500 (#11726).** Concurrent same-uid canonical writes all bump the one account-global `memory_state/apply_control` document. `replace_conversation_sourced_memories`' three *immediate* CAS attempts exhausted under that contention and raised an unhandled `RuntimeError` → `DELETE /v1/conversations/{id}?cascade=true` returned 500 while the Flutter client retried one conversation up to 12× and kept the storm fed. A repeat delete whose committed receipt went stale (another conversation's replacement advanced `source_generation`) livelocked on `source replacement operation already exists` the same way. Fix on the retract entry only (cascade delete + merge; the extraction replace path keeps its existing contract): bounded backoff retry (5 rounds, 0.05→0.5s), and after each conflict an epoch-guarded empty-cohort check — the same double-read control fence the replacement loop uses — returns the committed-empty result when a peer already retracted the source. Exhaustion raises typed `ConversationReplacementConflictError` (a `RuntimeError` subclass, so existing callers are unchanged), which `delete_conversation` maps to a retryable **503** *before anything is deleted*: the #11627 never-orphan-live-memories fence holds, and a retried delete is idempotent (receipt replay / already-retracted short-circuit).

Closes #11684
Closes #11726

## Relationship to open work

- **Supersedes #11687** (aryanorastar's index fix). That branch conflicts with current main — main refactored the predicate into `_query_spec_index_requirements()` — so this PR re-expresses the same semantic change there, ports its test fix verbatim (`test_firestore_config_does_not_declare_same_direction_single_field_composites` + the four-tuple declaration test), and regenerates the manifest through the generator rather than hand-patched JSON hunks. Credit for the measured predicate direction belongs to #11684/#11687.
- **Does not replace #11727** (draft: zero-write retraction while intake is paused). That PR targets the fenced-intake path; this PR converges the open-intake CAS path. Whichever lands second will need a small rebase; the touched regions barely overlap.
- Out of scope, per #11726: #11685/#11688 extract abort, MEMORY_MODE/cohort/control-schema redesign, client-only backoff.

## Merge sequencing (required)

`firestore_readiness` runs `reconcile_firestore_indexes.py --check-only` on backend deploys and **fails closed until every manifest index is READY**. Merge only after a human runs `Reconcile Firestore Indexes` → `prod` → `APPLY_FIRESTORE_INDEXES` (dev first is fine) — that APPLY alone stops the 503s even before this PR merges. Split-merging (declaration after APPLY lands separately) is a later human decision; nobody should run APPLY from this PR.

## How it was verified

- `pytest tests/unit/test_firestore_index_config.py tests/unit/test_firestore_indexes.py tests/unit/test_reconcile_firestore_indexes.py tests/unit/test_firestore_query_contract.py` → **62 passed** (was 61 on the wrong predicate; the renamed same-direction test no longer codifies the bug and the new opposite-direction declaration test pins all four indexes).
- `python3 backend/scripts/generate_firestore_indexes.py --write` then read-mode → `firestore.indexes.json matches the Firestore index registry`; diff adds exactly the 4 entries (+56 lines, same shape as #11687).
- `python3 backend/scripts/firestore_query_coverage.py --check-ratchet` → exit 0 (coverage keys on query signatures; QUERY_SPECS unchanged → no baseline drift).
- `pytest tests/unit/test_cascade_retract_convergence.py tests/unit/test_delete_conversation_cascade_retraction_fence.py` → **7 passed**: injected `ConversationSourceReplacementConflict` rounds converge with bounded backoff; repeat retract with a stale receipt returns already-retracted instead of livelocking; exhausted conflicts with live items still raise (conversation + memories intact, #11627 fence); route maps the typed conflict to 503 before any delete.
- Affected memory suites: `pytest tests/unit/test_ws_i_hardening.py tests/unit/test_ws_j_delete_privacy.py tests/unit/test_memory_apply_store.py tests/unit/test_memory_replace_policy.py tests/unit/test_merge_conversations_canonical_delete.py tests/unit/test_merge_skips_retraction_without_canonical_memories.py tests/unit/test_universal_memory_service.py tests/unit/test_atomicity_lifecycle_regressions.py` → **173 passed**.
- Scan-read/fake-contract: `pytest tests/unit/test_bounded_firestore_list_reads.py tests/unit/test_firestore_fake_contract.py` → **12 passed**.
- `scripts/pr-preflight --suggest` → INV-MEM-1 required; failure-class protocol honored below.

Not verified here: live Firestore behavior (no APPLY, no prod/dev writes — the four indexes must be provisioned by a human via the workflow above).

## Tests

- Index half regression: `backend/tests/unit/test_firestore_index_config.py::test_firestore_config_declares_opposite_direction_universal_list_scans` (fails on main's manifest) and the renamed same-direction test (fails if the predicate over-declares).
- CAS half regression: `backend/tests/unit/test_cascade_retract_convergence.py` (three behavioral tests: convergence under injected conflicts, stale-receipt idempotency, fail-closed exhaustion) + `test_delete_conversation_cascade_retraction_fence.py::test_exhausted_replacement_conflict_maps_to_retryable_503_before_any_delete`.

INV-MEM-1

## Failure class (fixes)

(That class covers the index half exactly — a serving query's composite index missing from the registry manifest. The CAS half has no existing class in the registry; it is a candidate for a new class — "account-global CAS exhaustion surfaced as an unhandled 500 on a retryable, idempotent operation" — to be registered separately per the transition-narrative protocol, not inline in this fix PR.)

Line-Count-Exception: backend/routers/conversations.py | 1515 -> 1527 | retryable-503 mapping for the cascade-retract conflict (#11726) is a few lines inside the existing delete route; splitting the router is out of scope for an incident fix
Line-Count-Exception: backend/utils/memory/canonical_memory_adapter.py | 2388 -> 2491 | retract convergence loop + epoch-guarded completion check + typed conflict error live with the CAS loop they wrap (#11726); extraction into a new module would split the replace/retract boundary mid-incident

## Residual risk

- The 503-after-exhaustion path is still a 5xx by design — but only after 5 backoff rounds *and* a proven-non-empty source, with nothing deleted and a safe client retry; the incident's unbounded 12×-retry 500 loop is gone (converges or fails closed).
- The already-retracted short-circuit trusts the same double-read fence the replacement loop uses; it cannot skip a retract that still has live items (live items → keep retrying → raise).
- Multi-instance Cloud Run: the backoff converges cross-instance via the control CAS itself; there is no cross-instance lock (deliberately — the smallest shape #11726 sanctioned).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11731?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/firestore_index_registry.py
+++ b/backend/database/firestore_index_registry.py
@@ -351,8 +351,9 @@ CANONICAL_MEMORY_ATLAS_READ_QUERY = FirestoreQuerySpec(
 # Collection-scoped newest-first scan for universal mixed list cursor paging.
 # Equality filters are intentionally empty: access/device/pending/archive are
 # applied after each bounded raw page so filtered rows still advance the keyset.
-# Firestore manages the single-field updated_at index (plus automatic __name__
-# tie-break) itself — this spec records the serving query contract only.
+# Firestore auto single-field indexes only cover field+__name__ in the *same*
+# direction (DESC+DESC / ASC+ASC). updated_at DESC + __name__ ASC is a real
+# composite and must be declared (#11684).
 UNIVERSAL_CANONICAL_LIST_SCAN_QUERY = FirestoreQuerySpec(
     identifier='memory_items_universal_list_scan',
     collection_group='memory_items',
@@ -363,8 +364,8 @@ UNIVERSAL_CANONICAL_LIST_SCAN_QUERY = FirestoreQuerySpec(
 
 # Historical dual-stream keysets for effective updated_at-or-created_at order.
 # Docs with updated_at ride the updated stream; created stream skips those
-# duplicates in Python so each document is emitted once. Single-field+__name__
-# indexes stay out of the composite manifest.
+# duplicates in Python so each document is emitted once. Opposite-direction
+# __name__ tie-breaks need composite indexes (same class as #11684).
 UNIVERSAL_HISTORICAL_UPDATED_LIST_SCAN_QUERY = FirestoreQuerySpec(
     identifier='memories_universal_list_scan_updated_at',
     collection_group='memories',
@@ -614,12 +615,36 @@ QUERY_SPECS = (
 _INDEX_ONLY_REQUIREMENT_SIGNATURES = frozenset(requirement.signature for requirement in INDEX_ONLY_REQUIREMENTS)
 
 
+def _index_fields_need_composite_manifest(index_fields: tuple[FirestoreIndexField, ...]) -> bool:
+    """Return True when Firestore will not serve this order from automatic indexes.
+
+    Automatic single-field indexes cover ``field ASC, __name__ ASC`` and
+    ``field DESC, __name__ DESC`` only. A lone ordered field with an opposite
+    ``__name__`` direction is a composite Firestore must be given explicitly
+    (#11684). Multi-field orders always need the composite manifest.
+    Array-contains (+ ``__name__``) stays out of the composite manifest — the
+    existing unified-memory index contract keeps those automatic.
+    """
+
+    non_name = [field for field in index_fields if field.field_path != '__name__']
+    name_fields = [field for field in index_fields if field.field_path == '__name__']
+    if len(non_name) > 1:
+        return True
+    if len(non_name) != 1 or len(name_fields) != 1:
+        return False
+    ordered = non_name[0]
+    name = name_fields[0]
+    if ordered.order is None or name.order is None:
+        return False
+    return ordered.order != name.order
+
+
 def _query_spec_index_requirements() -> tuple[FirestoreIndexRequirement, ...]:
     """One composite index per signature, even when two serving queries share it."""
     seen = set(_INDEX_ONLY_REQUIREMENT_SIGNATURES)
     requirements: list[FirestoreIndexRequirement] = []
     for spec in QUERY_SPECS:
-        if len([field for field in spec.index_fields if field.field_path != '__name__']) <= 1:
+        if not _index_fields_need_composite_manifest(spec.index_fields):
             continue
         signature = spec.index_requirement.signature
         if signature in seen:
--- a/backend/routers/conversations.py
+++ b/backend/routers/conversations.py
@@ -48,6 +48,7 @@ from utils.conversations.process_conversation import process_conversation, retri
 from utils.conversations import lifecycle as lifecycle_service
 from utils.executors import db_executor, llm_executor, postprocess_executor, run_blocking, submit_with_context
 from utils.memory.memory_service import MemoryService
+from utils.memory.retraction_scope import retraction_can_be_skipped
 from utils.memory.canonical_memory_adapter import ConversationReplacementConflictError
 from utils import byok
 from utils.conversations.search import (
--- a/backend/routers/conversations.py
+++ b/backend/routers/conversations.py
@@ -48,7 +48,7 @@ from utils.conversations.process_conversation import process_conversation, retri
 from utils.conversations import lifecycle as lifecycle_service
 from utils.executors import db_executor, llm_executor, postprocess_executor, run_blocking, submit_with_context
 from utils.memory.memory_service import MemoryService
-from utils.memory.retraction_scope import retraction_can_be_skipped
+from utils.memory.canonical_memory_adapter import ConversationReplacementConflictError
 from utils import byok
 from utils.conversations.search import (
     ConversationSearchUnavailableError,
@@ -792,7 +792,18 @@ def delete_conversation(
         # the fence is closed; anything real still raises rather than orphaning
         # live memories against a deleted conversation.
         if not retraction_can_be_skipped(uid, conversation_id, memory_service=memory_service, db_client=db_client):
-            memory_service.retract_conversation_memories(uid, conversation_id)
+            try:
+                memory_service.retract_conversation_memories(uid, conversation_id)
+            except ConversationReplacementConflictError as error:
+                logger.exception('cascade retraction conflicted uid=%s conversation_id=%s', uid, conversation_id)
+                # Concurrent same-account memory writes kept winning the
+                # account-global control CAS. Nothing has been deleted yet, so
+                # fail closed with a retryable answer instead of an opaque 500;
+                # the retraction is idempotent, a retried delete is safe (#11726).
+                raise HTTPException(
+                    status_code=503,
+                    detail='Conversation memory retraction is busy, please retry',
+                ) from error
 
         action_items_db.delete_action_items_for_conversation(uid, conversation_id)
         background_tasks.add_task(delete_conversation_audio_files, uid, conversation_id)
--- a/backend/tests/unit/test_cascade_retract_convergence.py
+++ b/backend/tests/unit/test_cascade_retract_convergence.py
@@ -1,0 +1,209 @@
+"""Cascade retraction converges under account-global control CAS races (#11726).
+
+Concurrent same-uid canonical writes — parallel ``DELETE
+/v1/conversations/{id}?cascade=true``, extractions, merges — all bump the one
+``memory_state/apply_control`` document. ``replace_conversation_sourced_memories``
+answers those races with ``ConversationSourceReplacementConflict`` and, after
+three immediate attempts, ``RuntimeError`` → unhandled HTTP 500, while the
+Flutter client retried the same conversation up to 12× and kept the storm fed.
+
+These tests pin the convergence contract on the retraction entry
+(``retract_conversation_sourced_memories``), the only path cascade delete and
+merge use:
+
+* injected control-change conflicts are retried with bounded backoff and
+  converge instead of raising;
+* a source whose retraction was already committed elsewhere — including the
+  stale-receipt shape after another conversation's replacement advanced the
+  generation — returns the committed-empty result instead of livelocking on
+  ``source replacement operation already exists``;
+* a source that still has live items after every bounded round keeps raising,
+  so the caller fails closed with the conversation and its memories intact
+  (the #11627 fence must not regress).
+"""
+
+from __future__ import annotations
+
+import importlib
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from tests.unit.memory_import_isolation import (
+    ensure_utils_memory_packages_importable,
+    install_database_client_stub,
+    install_ws_j_heavy_import_stubs,
+    restore_sys_modules,
+    snapshot_sys_modules,
+)
+
+BACKEND_DIR = Path(__file__).resolve().parents[2]
+
+
+@pytest.fixture(scope="module", autouse=True)
+def _cascade_retract_import_isolation():
+    saved = snapshot_sys_modules(["database._client", "firebase_admin", "utils.subscription", "database.users"])
+    install_database_client_stub()
+    install_ws_j_heavy_import_stubs()
+    yield
+    restore_sys_modules(saved)
+
+
+ensure_utils_memory_packages_importable(str(BACKEND_DIR))
+from database.memory_apply_store import ConversationSourceReplacementConflict  # noqa: E402
+from models.memory_apply import MemoryControlState  # noqa: E402
+from models.product_memory import MemoryItemStatus  # noqa: E402
+from utils.memory.canonical_memory_adapter import ConversationReplacementConflictError  # noqa: E402
+from tests.unit.fixtures.canonical_memory_fakes import (  # noqa: E402
+    _FakeDb,
+    _sample_memory_payload,
+    _trusted_account_generation,
+)
+
+UID = "uid-cascade-retract"
+
+
+@pytest.fixture(autouse=True)
+def _reset_universal_memory_env(monkeypatch):
+    from tests.unit.universal_memory_test_helpers import reset_universal_memory_fixture
+
+    canonical_adapter = importlib.import_module("utils.memory.canonical_memory_adapter")
+    globals()["retract_conversation_sourced_memories"] = canonical_adapter.retract_conversation_sourced_memories
+    globals()["write_canonical_extraction_memory"] = canonical_adapter.write_canonical_extraction_memory
+    monkeypatch.setattr(
+        canonical_adapter,
+        "read_memory_v3_trusted_account_generation",
+        lambda **_: _trusted_account_generation(),
+    )
+    reset_universal_memory_fixture(monkeypatch)
+
+
+def _cascade_db() -> _FakeDb:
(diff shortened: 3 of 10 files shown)


Git-on-my-level:
Monitor run summary (2026-08-17):

- **Hygiene / `product-file-line-count-ratchet` failure resolved.** The PR body declared `backend/routers/conversations.py | 1515 -> 1526`, but the third commit ("restore dropped retraction-scope import") took the file to 1527, leaving the declaration stale. Updated the `Line-Count-Exception` to the actual `1515 -> 1527` (reason unchanged). Verified locally against the ratchet script with the synthetic merge before editing: old body reproduced the exact CI failure; corrected body passes. The `canonical_memory_adapter.py | 2388 -> 2491` declaration was already accurate.
- **Rebased branch pushed.** The monitor-rebased branch (3 PR commits re-applied onto current `origin/main` @ `81bcd25c`, patch-ids verified identical) was force-pushed to `sca-316-scan-index-cas-fix` (`4685b0c` → `5d11789`). No code changes.
- **CI on the new head: 26 pass / 13 skip / 0 fail**, including the previously failing Hygiene check. No review threads open.

kodjima33 (approved):
incident fix for both #11684 (opposite-direction list-scan indexes) and #11726 (cascade-retract CAS convergence), well covered by tests; merge is gated on a human running Reconcile Firestore Indexes -> prod APPLY first

PR #11743, opened 2026-08-17
Windows search: find anything across conversations, memories, tasks and screen

ZachL111:
You could not search your own data on Windows. This adds a search that covers all of it.

## The gap

The only two search inputs in the whole Windows app were `components/rewind/RewindSearchBar.tsx` (28 lines, frames only) and `components/settings/SettingsSearchProvider.tsx` (37 lines, settings only). A conversation, a memory or a task could be found by scrolling and no other way.

The capability was partly there and unreachable. `POST /v1/conversations/search` exists, and the generated client already has `search_conversations_endpoint_v1_conversations_search_post` in `omiApi.generated.ts` - nothing in the UI ever called it. Memories had no index at all, while rewind frames, action items and staged tasks each had one.

macOS solves this with QueryShell (`MainWindow/QueryShell/`, ~2,900 lines): Home itself becomes a search over rows it has already hydrated into memory. That works there because macOS mirrors everything locally in GRDB. Windows has no local mirror of cloud conversations, so the same design would have searched a fraction of someone's history and presented it as all of it. This uses full-text indexes and the server endpoint instead, which is a deliberate divergence and is why the shape differs.

## What is searched, and from where

| Corpus | Source | Coverage |
|---|---|---|
| Conversations | `POST /v1/conversations/search` | Full history |
| Memories | new `memories_fts` index | Everything on this machine |
| Tasks | existing `action_items_fts` + `staged_tasks_fts` | Everything on this machine |
| Screen | existing `rewind_frames_fts` | Everything on this machine |

One matcher runs all of it. `buildRewindFtsMatch` was already the app's tuned user-facing FTS5 matcher: quoted phrase terms so punctuation can never become query syntax, a trailing star for prefix search, camelCase and digit sub-parts expanded, tokens AND-joined so typing more narrows. Using it everywhere means one query behaves the same way against every corpus, rather than adding a fourth matcher next to the two the codebase already has for other jobs (exact-AND for agent tool calls, prefix-OR for dedup recall).

## Three things the design refuses to do

**Never interleave BM25 scores across corpora.** Scores from two indexes are computed over different document populations, so comparing them is meaningless and would silently order results by which index happens to hold shorter documents. Each corpus is ranked inside itself and shown in its own group.

**Never let a capped list read as a complete one.** Every slice carries the true total next to the hits it is showing, so the page says "Showing 20 of 300" instead of "20". A search that silently truncates makes someone stop looking, believing they have seen everything. Conversation totals come from a page count rather than a row count, so they are shown as a lower bound ("41+") rather than a number that would invent up to nineteen conversations.

**Never show an unreachable corpus as an empty one.** The two sources fail independently: a conversation request that times out still shows local results, and an unusable local index still shows conversations. Each corpus carries its own failure state, because "could not be searched" and "you have nothing about that" are different claims and only one of them is about the user's data.

## Privacy

The memories index covers content, context summary and source app, and deliberately not `window_title`. `assistants/memory/persist.ts` keeps raw window titles local precisely because they are unfiltered PII ("Chase - Log in", "MyChart - Lab Results"); indexing them would make that PII reachable by typing a fragment of it into a search box. A test asserts it stays unsearchable.

## Races

Typing fires a request per keystroke and a two-letter query can easily resolve after a five-letter one. Every run takes a sequence number, aborts its predecessor, and delivers only while it is still the latest, so the list on screen always belongs to the text in the box. Results also carry the text they were computed from, so "nothing matches X" can never name a different word than the one that was searched.

## Divergence: Ctrl+7

`manifest.test.ts` pins the Windows shortcut table to macOS's `OmiApp.swift:163-214` (Cmd+1..6 and Cmd+,) and asserted that no seventh key was bound. That test is updated, and the reason belongs in the record: macOS has no Search destination to bind, because its search is the QueryShell that Home becomes when you type. Windows has no QueryShell, so search is its own destination and needs its own key. Nothing else in the table moved, and Search is appended to the rail rather than placed near Home so the existing order is untouched.

## Migration

`memories_fts` is external-content and trigger-maintained, following `rewind_frames_fts` exactly, so it stays out of `USER_DATA_TABLES`: the account-switch `DELETE FROM memories` empties it through the delete trigger. Migration v3 runs the one-time rebuild for installs whose memories predate the index, and the index is added to the post-recovery rebuild list alongside the other three.

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,293 passed, 18 skipped, 547 files. 85 of those are this feature, run against a real SQLite database via node:sqlite and a real jsdom render of the page.
- `npx electron-vite build` - built in 17.14s.
- `check_lifecycle_headers.py`, `check_brand_ui.py`, `check_arch_guardrails.py` - all pass.

## Mutation audit

24 regressions seeded into the source, each confirmed to turn the suite red:

- memories index also covers the raw window title
- memory delete trigger dropped, leaving deleted content in the index
- memories ranked oldest-first on a tie
- existing memories never backfilled into the index
- deleted tasks become findable again
- completed tasks dropped from search
- suggested tasks not searched at all
- total reports only what is shown, hiding truncation
- a malformed query passed straight to FTS5
- IPC accepts a non-string query
- IPC lets a database failure reach the renderer
- IPC does not bound a pasted query
- a local failure blanks the conversation results too
- a conversation failure blanks the local results too
- a stale answer overwrites a newer one
- a newer search does not abort the one in flight
- a multi-page conversation total presented as exact
- a one-letter query is searched
- results echo the live input instead of what was searched
- an unreachable corpus renders as no matches
- the count line hides that the list is capped
- a failed corpus counts as an empty result
- empty is shown while a search is still running
- no results yet is reported as nothing found

Two of these were caught only after a test was strengthened. The delete-trigger test could not fail as first written: `searchMemoriesFts` joins `memories`, so a deleted row is filtered out by the join whether or not the index was updated. Asserting on the count, which reads the index directly, is what actually exercises the trigger.

## What is not here

The Activity spine (macOS `MainWindow/Spine/`, ~3,000 lines) is the other half of the macOS surface and is not in this PR. Deep links to a single memory, task or frame are also absent: those pages have no per-item route to land on yet, so a result opens the page the item lives on. Memories are searched from the local index, which covers memories extracted on this machine; there is no server-side memory search endpoint to complement it the way conversations have one.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11743?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/index.ts
+++ b/desktop/windows/src/main/index.ts
@@ -29,6 +29,7 @@ import { registerCaptureBridge } from './ipc/captureBridge'
 import { registerSoak } from './soak'
 import { createCaptureWindow, getCaptureWindow, getCaptureWc } from './captureWindow'
 import { registerFileIndexHandlers } from './ipc/fileIndex'
+import { registerSearchHandlers } from './ipc/search'
 import { cancelStartupRescan } from './fileIndex/indexer'
 import { registerMemoryImportHandlers } from './ipc/memoryImport'
 import { registerMemoryExportHandlers } from './ipc/memoryExport'
@@ -855,6 +856,7 @@ app.whenReady().then(async () => {
   // byte counters to userData/soak.jsonl for the 8h idle-soak verification.
   registerSoak()
   registerFileIndexHandlers()
+  registerSearchHandlers()
   registerLocalGraphHandlers()
   registerMemoryImportHandlers()
   registerMemoryExportHandlers()
--- a/desktop/windows/src/main/ipc/db.ts
+++ b/desktop/windows/src/main/ipc/db.ts
@@ -8,6 +8,12 @@ import { addColumnIfMissing as ensureColumn, runMigrations } from './dbMigration
 import { applyRewindEmbeddingSchema } from './rewindEmbeddingSchema'
 import { LOCAL_CONVERSATION_SCHEMA } from './localConversationSchema'
 import {
+  MEMORY_SEARCH_SCHEMA,
+  searchMemoriesFts,
+  type MemorySearchDb,
+  type MemorySearchRow
+} from '../search/memorySearchStore'
+import {
   clearCorruptionFlags,
   isCorruptionError,
   isCorruptionSuspected,
@@ -656,6 +662,12 @@ function get(): Database.Database {
     );
     CREATE INDEX IF NOT EXISTS idx_memories_created_at ON memories(created_at);
   `)
+  // Full-text index over the memories just created. DDL lives in
+  // search/memorySearchStore.ts so prod and the node:sqlite tests run the same
+  // SQL; the index is external-content + trigger-maintained, so it is
+  // deliberately absent from USER_DATA_TABLES (see dbWipe.ts) and is backfilled
+  // for existing installs by migration v3.
+  db.exec(MEMORY_SEARCH_SCHEMA)
   // Track 3 local task storage (action_items + staged_tasks + their FTS indexes).
   // DDL lives in taskStore.ts so prod and the node:sqlite CRUD tests run the same
   // SQL; both tables are user-scoped (see USER_DATA_TABLES in dbWipe.ts).
@@ -702,7 +714,12 @@ function get(): Database.Database {
     // Rebuild every external-content FTS index from its recovered base rows (salvage
     // skips virtual tables, leaving the shadow tables empty). Same 'rebuild' idiom as
     // migration v2. Each is independent — one failing must not skip the others.
-    for (const fts of ['rewind_frames_fts', 'action_items_fts', 'staged_tasks_fts']) {
+    for (const fts of [
+      'rewind_frames_fts',
+      'action_items_fts',
+      'staged_tasks_fts',
+      'memories_fts'
+    ]) {
       try {
         db.exec(`INSERT INTO ${fts}(${fts}) VALUES('rebuild')`)
       } catch (e) {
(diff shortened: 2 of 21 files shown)


ZachL111:
Merge-order note for whoever lands this, found while sweeping my open PRs for mergeability.

This PR and #11802 (Rewind video chunks) both add a database migration numbered `version: 3`:

- #11743 adds `memories_fts_backfill`
- #11802 adds `rewind_video_chunks`

Each is clean against main on its own, so GitHub reports both as mergeable and will keep doing so right up until the second one lands. The collision is in `desktop/windows/src/main/ipc/dbMigrations.ts`.

This one is worth resolving carefully rather than by hand-merging the hunk, because the failure is not cosmetic. `runMigrations` enforces contiguity:

```
if (m.version !== i + 1) {
  throw new Error(`migrations must be contiguous from 1; found version ${m.version} at index ${i}`)
}
```

So a resolution that keeps both entries at 3, or that leaves a gap, throws on every launch for every user rather than failing visibly in CI.

Whichever merges second needs its migration renumbered to 4 and its own test updated. Nothing else about either change interacts; the two migrations touch different tables.

Happy to do the rebase on whichever of the two you want to land second, once the order is decided.

ZachL111:
Merge-order note, found while sweeping my open PRs for mergeability.

This PR and #11746 (Activity) are each clean against main but conflict with each other once either one lands. Simulating the two merges in sequence, in both orders, the overlap is:

- `desktop/windows/src/renderer/src/routes/manifest.ts` and its test
- `desktop/windows/src/main/index.ts`
- `desktop/windows/src/main/ipc/db.ts`
- `desktop/windows/src/preload/index.ts`

All four are append-at-the-end registration points, so the resolution is keep-both rather than pick-one, and no behaviour is in question. I flagged the manifest half of this on #11746 already; this is the complete list.

I will rebase whichever of the two lands second. No action needed before then.

aryanorastar:
Still the right gap on current main: there is no Search page, and conversation search is still an in-memory filter of the loaded list.

Not approving this head. It is ~3311 commits behind, CI is from 17 Aug, and this PR and #11802 both add dbMigrations version 3 (`memories_fts_backfill` vs `rewind_video_chunks`). `runMigrations` requires contiguous versions; a keep-both-at-3 resolution will throw on every launch. Rebase onto current main, pick a free version, and update the migration test.

#13702 is not a substitute (Conversations list search only). #11746 overlaps manifest/index/db/preload — keep-both if that one lands first.

Privacy note from the branch is the one to keep: do not index raw window titles.

aryanorastar:
Keeping this open. Not approving `46e1ba998fae`. `desktop/windows/src/main/search/desktopSearch.ts` absent on origin/main. +2350.

PR #11746, opened 2026-08-17
Windows Activity: one day-grouped stream of conversations, memories, tasks and screen

ZachL111:
Windows had four separate lists and no way to see a day. This composes them into one.

## The gap

Answering "what happened on Tuesday" meant opening Conversations, Memories, Tasks and Rewind and cross-referencing them by hand, because nothing on Windows put a day's records next to each other. macOS has had this for a while as the Activity tab of the Memory hub (`MainWindow/Spine/`, ~3,000 lines). The survey of remaining macOS-only surfaces ranked it the largest gap with no platform blocker, and it is the other half of the surface #11743 started.

This is a port of that model: conversations, memories, tasks and captured screen moments composed into one reverse-chronological stream, grouped by day, with a 24-hour density rail beside each day and chips to solo a kind.

## The rules that carry the weight

Almost every rule in `spineModel.ts` exists to stop a specific wrong-looking timeline rather than to produce a right-looking one. The four worth reading the diff for:

**Frames attach to conversations through a two-pointer merge whose cursor advances only past conversations that START after the frame.** Advancing on the finish time instead lets one frame newer than everything consume the whole list and orphan every conversation behind it. macOS regression-tests this by name (`testAFrameNewerThanEveryConversationDoesNotOrphanTheOnesBehindIt`); the test here is the same case.

**Attached rows are re-seated directly under their conversation**, deliberately defeating the pure-anchor ordering. A memory recorded mid-conversation would otherwise interleave with unrelated rows and read as though it came from somewhere else. They sit in a fixed order (memories, tasks, screen) rather than by anchor, so the same card does not reorder itself depending on which attachment happened to be recorded last.

**A memory pointing at a conversation that is not loaded becomes its own row.** Hiding the user's data behind a paging boundary is worse than showing it unattached.

**A day's moment count is null when its capture has not been read and 0 when it was read and held nothing.** Collapsing the two makes an unread day claim the user captured nothing. The rail renders the first as an em dash and "counting screen moments".

## Local time, deliberately

Days and hours are bucketed in JS against the local calendar and never in SQL. A UTC-keyed `GROUP BY` shifts every row near midnight into the wrong day, and SQLite's date functions would need an explicit `localtime` modifier and still bucket against a UTC day, so the histogram would disagree with the day grouping beside it.

For the same reason the day boundary is computed by adding one calendar day rather than 24 hours, and the day walk steps by calendar day: a DST transition must not drop, duplicate or misplace an hour.

## Reading a day without shipping it

A busy day is thousands of frames and the stream shows eight per strip, so `screenIndex.ts` reads a day at three resolutions: an exact count, an exact 24-bucket histogram built from the day's timestamps alone, and a bounded evenly-spaced sample of whole frames. The histogram is built from every timestamp rather than from the sample, so a busy hour still reads as busy.

The sample is taken on a row number over the day's own ordering rather than on rowid, so retention deleting older frames does not reshuffle which frames a day shows.

## Degradation

Any of the four sources may fail. The stream shows whatever loaded and says the day may be incomplete, because a timeline missing one source is far more useful than no timeline and much better than one that quietly implies the missing records never existed.

Fold state is keyed by the day's own identity, so paging older days in cannot move which day is folded, and is session-only, so a day folded last week is not still hidden today.

## Divergences from macOS, stated

- **No Ctrl+N.** macOS reaches this as a tab inside the Memory hub rather than as a destination, so it binds no key there either. Windows has no Memory hub with destination tabs, so Activity is its own panel route; taking a number would also collide with the Search page in #11743. The `manifest.test.ts` shortcut pin is untouched.
- **No video frames.** macOS moments carry a video chunk path and frame offset; Windows Rewind is screenshot-only (`grep -riE "videoChunk|mp4|h264|ffmpeg"` across `desktop/windows/src` returns 0), so the moment model carries the five fields a strip actually renders.
- **The row comparator's kind tie-break cannot be observed through the composer.** Rows are emitted conversation-first and `Array.prototype.sort` is stable, so the order is already right without it; macOS needs it because Swift's `sort` is not stable. It is kept as an explicit invariant against a future change to the emit order, and exported so it is tested directly rather than inferred.

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,301 passed, 18 skipped, 546 files. 105 of those are this feature, including a real SQLite database via node:sqlite for the screen index and a jsdom render of the page.
- `npx electron-vite build` - built in 21.47s.
- `check_lifecycle_headers.py`, `check_brand_ui.py` - pass.

## Mutation audit

28 regressions seeded into the source, each confirmed to turn the suite red:

- frame cursor advances on finish instead of start, orphaning conversations
- a frame exactly at a conversation's end no longer attaches
- a memory whose conversation is not loaded is dropped
- attachments are not re-seated under their conversation
- an attachment whose owner is missing is dropped
- attachments order by anchor instead of a fixed type order
- rows order oldest first
- days order oldest first
- a conversation no longer wins a tie against its own attachment
- overlapping pages produce duplicate rows
- cluster splits at exactly 45 minutes instead of beyond it
- a strip reports only what it shows
- the brain map card counts as a conversation row
- a brain map card appears on a day with no memories
- an attached record is filed under its own day, not its conversation's
- a day that has not been read reports zero moments
- filtering keeps days that matched nothing
- the query is matched case-sensitively
- an untimed conversation is shown at epoch zero
- an unfinished conversation is treated as zero length
- an inverted conversation range is kept
- the day ends 24 hours later instead of one calendar day
- the histogram is built from the sample instead of every timestamp
- the sample is taken from the start of the day only
- a day with no capture is indistinguishable from one never read
- the rail runs oldest hour first
- density normalises against a fixed ceiling rather than the day
- the rail headline loses its not-read-yet branch

Four of these escaped on the first pass and produced the last commit. Two were tests that could not fail: the ordering tie-break was masked by a stable sort plus a conversation-first emit order, and the conversation time window had two overlapping guards so neither could be shown to matter alone. The other two were simply missing cases.

One coverage limit worth stating rather than hiding: the "24 hours instead of one calendar day" mutation is only observable in a timezone that observes DST. The test asserts the property across every day of a year, so it catches the mutation on a developer machine in such a zone, but CI runs in UTC where no day transitions.

## What is not here

Paging further back than the first load (200 conversations, 300 memories, 14 screen days), and thumbnails on the screen strips - the strip shows window titles, and rendering frame images needs the same decode path Rewind uses. macOS's `SpineHydrator` background paging is the natural follow-up.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11746?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/index.ts
+++ b/desktop/windows/src/main/index.ts
@@ -29,6 +29,7 @@ import { registerCaptureBridge } from './ipc/captureBridge'
 import { registerSoak } from './soak'
 import { createCaptureWindow, getCaptureWindow, getCaptureWc } from './captureWindow'
 import { registerFileIndexHandlers } from './ipc/fileIndex'
+import { registerSpineHandlers } from './ipc/spine'
 import { cancelStartupRescan } from './fileIndex/indexer'
 import { registerMemoryImportHandlers } from './ipc/memoryImport'
 import { registerMemoryExportHandlers } from './ipc/memoryExport'
@@ -855,6 +856,7 @@ app.whenReady().then(async () => {
   // byte counters to userData/soak.jsonl for the 8h idle-soak verification.
   registerSoak()
   registerFileIndexHandlers()
+  registerSpineHandlers()
   registerLocalGraphHandlers()
   registerMemoryImportHandlers()
   registerMemoryExportHandlers()
--- a/desktop/windows/src/main/ipc/db.ts
+++ b/desktop/windows/src/main/ipc/db.ts
@@ -8,6 +8,12 @@ import { addColumnIfMissing as ensureColumn, runMigrations } from './dbMigration
 import { applyRewindEmbeddingSchema } from './rewindEmbeddingSchema'
 import { LOCAL_CONVERSATION_SCHEMA } from './localConversationSchema'
 import {
+  projectScreenDay,
+  screenDaysWithCapture,
+  type ScreenDayProjection,
+  type ScreenIndexDb
+} from '../spine/screenIndex'
+import {
   clearCorruptionFlags,
   isCorruptionError,
   isCorruptionSuspected,
@@ -1475,6 +1481,19 @@ export function listRewindFramesSampled(
     const step = rewindSampleStep(n, target)
     return d.prepare(buildRewindSampledSql(REWIND_COLUMNS)).all(from, to, step) as RewindFrame[]
   })
+}
+
+/** Activity spine screen projections. Thin get()-bound wrappers over
+ *  spine/screenIndex.ts, which owns the SQL so production and the node:sqlite
+ *  tests run the same statements. */
+export function spineScreenDays(limit: number): number[] {
+  return timed('spineScreenDays', () =>
+    screenDaysWithCapture(get() as unknown as ScreenIndexDb, limit)
+  )
+}
+
+export function spineScreenDay(dayId: number): ScreenDayProjection {
+  return timed('spineScreenDay', () => projectScreenDay(get() as unknown as ScreenIndexDb, dayId))
 }
 
 // --- Track 4: Rewind FTS5 search ---
--- a/desktop/windows/src/main/ipc/spine.ts
+++ b/desktop/windows/src/main/ipc/spine.ts
@@ -1,0 +1,56 @@
+// Renderer-facing surface for the Activity spine's screen data.
+//
+// Reads only, so it is not gated to the main window: this returns the same
+// capture the Rewind page already shows. Two channels rather than one, because
+// the two answers have very different sizes - the day list is a handful of
+// numbers and is needed immediately, while a day projection carries up to a few
+// hundred frames and is only worth fetching for days actually on screen.
+
+import { ipcMain } from 'electron'
+import { spineScreenDay, spineScreenDays } from './db'
+import type { ScreenDayProjection } from '../spine/screenIndex'
+
+/** Most days offered to the spine at once. The stream pages further back by
+ *  conversation and memory, which are cheap; screen days are not. */
+export const MAX_SCREEN_DAYS = 30
+
+const emptyDay = (dayId: number): ScreenDayProjection => ({
+  dayId,
+  total: 0,
+  hourCounts: new Array<number>(24).fill(0),
+  sampled: []
+})
+
+export function listScreenDays(limit?: unknown): number[] {
+  const bounded =
+    typeof limit === 'number' && Number.isFinite(limit)
+      ? Math.max(1, Math.min(Math.floor(limit), MAX_SCREEN_DAYS))
+      : MAX_SCREEN_DAYS
+  try {
+    return spineScreenDays(bounded)
+  } catch (e) {
+    // A timeline that cannot read screen capture still has conversations,
+    // memories and tasks to show; failing the whole page over this would be a
+    // worse answer than a timeline with no screen rows in it.
+    console.error('[spine] screen day list failed', e)
+    return []
+  }
+}
+
+export function readScreenDay(dayId: unknown): ScreenDayProjection | null {
+  if (typeof dayId !== 'number' || !Number.isFinite(dayId)) return null
+  const day = Math.floor(dayId)
+  try {
+    return spineScreenDay(day)
+  } catch (e) {
+    console.error('[spine] screen day projection failed', e)
+    // A zeroed day, not null: the caller asked and got an answer, so the rail
+    // shows 0 rather than sitting on "counting" forever.
+    return emptyDay(day)
+  }
+}
+
+export function registerSpineHandlers(): void {
+  ipcMain.handle('omi-spine:screen-days', (_e, limit?: unknown) => listScreenDays(limit))
+  ipcMain.handle('omi-spine:screen-day', (_e, dayId: unknown) => readScreenDay(dayId))
+}
--- a/desktop/windows/src/main/spine/screenIndex.test.ts
+++ b/desktop/windows/src/main/spine/screenIndex.test.ts
@@ -71,12 +71,23 @@ describe('bucketByLocalHour', () => {
 
 describe('endOfLocalDay', () => {
   it('ends one calendar day later, not 24 hours later', () => {
-    // Adding 86_400_000 would land an hour early or late across a DST boundary
-    // and drop or double-count the frames in that hour.
     const end = endOfLocalDay(DAY_15)
     const d = new Date(end)
     expect([d.getHours(), d.getMinutes()]).toEqual([0, 0])
     expect(d.getDate()).toBe(16)
+  })
+
+  it('lands on local midnight for every day of a year', () => {
+    // Adding a fixed 86_400_000 lands an hour early or late on the two DST
+    // transition days, dropping or double-counting the frames in that hour. In a
+    // timezone that observes DST this loop fails for the fixed-constant version;
+    // in UTC (which CI uses) no day transitions, so this asserts the property
+    // holds everywhere it can be observed rather than proving the DST case.
+    for (let i = 0; i < 365; i += 1) {
+      const day = startOfLocalDay(new Date(2026, 0, 1 + i, 12).getTime())
+      const end = new Date(endOfLocalDay(day))
+      expect([end.getHours(), end.getMinutes(), end.getSeconds()]).toEqual([0, 0, 0])
+    }
   })
 })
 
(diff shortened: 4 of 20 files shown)


Git-on-my-level (dismissed):
Thanks @ZachL111 — this is a genuinely well-built port. The composition rules in `lib/spine/spineModel.ts` carry over the load-bearing macOS behaviors, and the two-pointer `attachMoments` keeps the exact regression macOS pins by name (`testAFrameNewerThanEveryConversationDoesNotOrphanTheOnesBehindIt`), with the same case covered in `spineModel.test.ts`. The three-resolution day projection in `main/spine/screenIndex.ts` (exact total, exact histogram, ROW_NUMBER-sampled frames) and the DST-safe day walk in `screenDaysWithCapture` are the right shape, and `main/ipc/spine.ts` degrading to "no screen rows" instead of failing the whole page matches how the app treats optional sources elsewhere. Tests are unusually thorough.

One blocking bug, verified against the backend contract:

**`pages/Activity.tsx` — the star toggle sends the wrong query param.** `toggleStar` calls `omiApi.patch(\`/v1/conversations/${id}/starred\`, null, { params: { value: !starred } })`, but the endpoint requires the query param to be named `starred` (see `backend/routers/conversations.py`, `set_conversation_starred`, which takes `starred: bool`). With `value`, the required param is missing, so the request fails validation and every star tap lands in the "Could not update that conversation." toast. The rest of the app already uses the right name — see `lib/conversations/mutations.ts`: `{ params: { starred } }`. The test "stars a conversation without opening it" in `pages/Activity.test.tsx` mocks `patch` and asserts `{ params: { value: true } }`, so the suite currently enshrines the wrong contract; please fix both together.

Two non-blocking notes:

- `components/spine/SpineRow.tsx`: the star control is a `span[role=button]` nested inside the row's `<button>`. The `stopPropagation` handling is tested and correct; flagging only that nested interactive elements are an accessibility smell — worth revisiting if you touch this file again.
- `shared/types.ts`: most of this diff is formatting churn on existing unions (`ConversationSyncState`, `IndexedFileType`, `LocalKGNodeType`) alongside the real new `SpineScreenDay`/`SpineScreenMoment` types. Harmless, just noting it slightly blurs the actual type additions.

Feature fit: composing the four existing lists into one day view is straight macOS parity for Windows; no concerns with the direction. One-line fix plus the matching test assertion, and this looks ready for a maintainer's merge pass.

— automated maintainer feedback on behalf of the Omi maintainers

ZachL111:
Thanks for the review. The blocking bug is real and is fixed in 19a0a1bfc4, along with both non-blocking notes.

## The star param

You are right about the cause and the fix, and you are also right that the test was the worse half of the problem: it asserted the same wrong name, so the suite was actively holding the bug in place rather than catching it.

I fixed it one level up from the one-line change. `Activity.tsx` no longer builds the request at all. It calls the existing helper you pointed at:

```ts
import { setConversationStarred } from '../lib/conversations/mutations'
...
await setConversationStarred(id, !starred)
```

The hand-rolled `omiApi.patch` was the actual defect. Any new page that spells the request out itself can get the param name wrong again, and a page that calls the shared mutation cannot. That is why the fix went there instead of into the string.

The test now names the contract and cites where it comes from:

```
it('stars a conversation with the query param the endpoint requires', ...)
    expect(patch).toHaveBeenCalledWith('/v1/conversations/c1/starred', null, {
      params: { starred: true }
    })
```

I checked that the corrected test can actually fail, since a test rewritten in the same change as the code it asserts is worth nothing until it is shown to be load-bearing. Changing `params: { starred }` to `params: { value: starred }` inside `lib/conversations/mutations.ts` turns it red:

```
 FAIL  src/renderer/src/pages/Activity.test.tsx > Activity page >
       stars a conversation with the query param the endpoint requires
 Tests  1 failed | 9 passed (10)
```

The mutation is in the shared helper rather than in the page, which is the point: the assertion now guards the helper for every caller, not just this page.

## SpineRow nesting

Fixed rather than deferred. The star is no longer a `span[role=button]` inside the row's `<button>`. It is a sibling `<button>`, so there is no nested interactive element and the row is a single control. The existing `stopPropagation` tests and the "does not open the conversation when the star is pressed" test still pass unchanged, which is what I wanted to confirm before changing the markup.

## shared/types.ts

Also fixed. The diff against main is now 24 insertions and 0 deletions, so the file contains only the new `SpineScreenDay` and `SpineScreenMoment` types with the existing unions untouched. You were reading it correctly, the reformatting was real, and it is gone.

## Verification

Run in `desktop/windows`:

- `npx vitest run` on the full suite, green.
- `npx tsc --noEmit` on `tsconfig.web.json` and `tsconfig.node.json`, clean.
- `npx eslint` on every touched file, 0 problems.

One thing worth flagging that is outside this review: this branch and #11743 both append to `routes/manifest.ts`, so whichever merges second will conflict there. It is a two-line textual conflict, not a behavioral one, and I will rebase whichever is second.

Git-on-my-level:
Thanks @ZachL111 — re-reviewed on the current head, and everything from my earlier review is resolved:

- **Star toggle fixed.** `pages/Activity.tsx` now goes through `setConversationStarred` from `lib/conversations/mutations.ts`, which sends the query param as `starred` — matching `set_conversation_starred` in `backend/routers/conversations.py`. And `pages/Activity.test.tsx` now asserts `{ params: { starred: true } }` instead of enshrining the wrong `value` param, so the suite pins the real contract.
- **Star control is now a sibling `<button>`** next to the open button in `components/spine/SpineRow.tsx` rather than a nested interactive element — kinder to keyboard users, and the `stopPropagation` gymnastics are gone.
- **`shared/types.ts` churn is gone** — the additions there are now just the `SpineScreenDay`/`SpineScreenMoment` bridge types and the two `spineScreen*` methods, matching the handlers in `main/ipc/spine.ts`.

Also re-checked the surfaces I hadn't blocked on: the new IPC channel pair stays read-only with `limit` clamped to [1, 30] (`MAX_SCREEN_DAYS`) and a `Number.isFinite`-checked `dayId`; `main/spine/screenIndex.ts` uses positional `?` parameters only, and its ROW_NUMBER-based sampling keeps the sample stable when retention deletes frames mid-day; `hooks/useSpine.ts` degrades per-source so a failed screen read still renders conversations/memories/tasks under the "may be incomplete" notice. Both `Windows build smoke` and `Typecheck · Lint · Test` pass on this head.

No blocking concerns remain from my side. The one thing left is a maintainer's product call rather than a code fix: this adds a new top-level rail destination to the Windows app, where macOS reaches the same spine as a tab inside the Memory hub — worth a deliberate sign-off on that placement before merge.

— automated maintainer feedback on behalf of the Omi maintainers

---

Git-on-my-level:
Two small non-blocking notes on top of my earlier re-review of this head — neither changes its conclusion, and the pending maintainer call on navigation placement stands.

1. `main/spine/screenIndex.test.ts` — "keeps sampling stable when retention has deleted earlier frames" is currently a no-op. It runs `DELETE FROM rewind_frames WHERE id <= 0`, but `id` is `INTEGER PRIMARY KEY AUTOINCREMENT`, so ids start at 1 and the DELETE removes nothing; `after` then equals `before` for any sampling implementation, including a rowid-based one. The ROW_NUMBER approach in `projectScreenDay` is the right one — this test just doesn't pin it. Seeding frames on an earlier day and deleting those (leaving a real rowid gap below the day under test) would make it load-bearing.

2. `screenDaysWithCapture` returns every calendar day between the oldest and newest capture, not "days that hold any capture at all" as the header comment (and the test name) say — its own test asserts day 14 is listed for frames placed only on days 13 and 15. Downstream, that means the unfiltered stream renders empty day sections ("0 entries" header plus a zeroed rail) for capture-less days inside the range. If gap days are meant to show as honest empty days, that's a defensible choice — but the name and comment undersell it; if not, the walk should skip zero-count days, which would also save projecting them in `useSpine`.

Both are quality nits, not merge blockers: the composition rules, IPC input validation, and per-source degradation all held up under a full-file pass, and both Windows checks pass on this head.

---

aryanorastar:
Keeping this open. Not approving `19a0a1bfc44b`. `desktop/windows/src/main/spine/screenIndex.ts` absent on origin/main. +2712.

PR #11750, opened 2026-08-17
Windows memory atlas: the brain map as named regions instead of hidden nodes

ZachL111:
The brain map keeps a large graph legible by hiding nodes. This adds a view that hides none.

## The problem, as the codebase already states it

`KnowledgeGraphViewer.tsx` records the measurement in its own header:

> a real account's graph is large (measured: ~188 nodes / ~474 edges, one 226-degree hub, ~40% degree-1 leaves) - drawn whole it is an unreadable, laggy hairball

Its answer is a default node cap plus label declutter, with a "Show all" escape hatch. That is a reasonable answer, and it is the only one Windows has.

macOS answers the same problem the other way, in `MainWindow/Pages/MemoryGraph/` (9,463 lines against Windows' ~1,500): group the entities into regions, give each region a name, and let detail arrive with magnification instead of throwing it away up front. This ports that.

Both views ship. The Brain Map page gains a Map/Graph toggle; the existing scene is untouched, which matters because onboarding mounts it too.

## Pipeline

```
relatedness  ->  communities  ->  layout  ->  coastlines  ->  territories
```

Communities are detected **before** positions are computed, which is the reverse of the obvious order, and the reason is the most interesting thing in the diff. See "what building it taught me" below.

## The rules that carry the weight

**Relatedness is two signals.** Edges the graph states outright, weighted logarithmically in the memories behind them, plus entities that keep turning up in the same memory. Each memory divides one unit of relatedness among its participants, each weighted by how rare it is, so two entities that only ever co-occur inside one ubiquitous entity's memories are not treated as related. A memory naming more than twenty entities is skipped entirely: it relates everything to everything, which carries the same information as relating nothing.

**Coastlines are a scalar field, not a hull.** Each community's members are splatted as Gaussian bumps into a shared 128x128 field; a cell belongs to whichever community peaks highest, provided the peak is decisive; the mask is blurred and traced with marching squares. A hull would swallow every other community standing between a community's members. A field lets two interleaved communities own nothing in the contested middle, and lets a community that lives in two places come out as two islands rather than one region spanning the gap.

**Three constants shape every coastline.** The sea level sits above 1 because a lone entity's bump peaks at exactly 1, so one entity can never mint a territory. The decisive margin means a cell only belongs to the leader if it beats the runner-up by a quarter. Reach is not an absolute distance but the map's own median nearest-neighbour spacing times 2.5, because the same spacing means "tight" on a dense map and "scattered" on a sparse one.

**Marching squares stitches by integer edge identity**, not by coordinate. Comparing interpolated coordinates leaves hairline gaps wherever two cells disagree in the last bit, and a ring with a gap cannot be filled.

**A region the map cannot name is not drawn.** An unnamed shape is noise; the entities inside it are still drawn as entities.

**Determinism is a contract.** Every traversal that can affect an output walks a sorted array, a tied Louvain candidate cannot pull a node, and the layout seeds from a phyllotaxis spiral by index with a fixed tick count. The graph is merged from an onboarding floor plus a server fetch whose interleaving is not fixed, so without this the map would rearrange itself between runs for no reason a user could see.

## What building it taught me

Wired to the app's existing `computeLayout`, the ported island algorithm produced almost nothing: two of three obvious clusters had no land at all, and the third produced a ridge running BETWEEN its members with none of them standing inside it.

That is why macOS carries its own force layout, and it is not incidental. The existing layout spreads entities evenly, which is right for a node-link graph. A coastline is drawn where a community's own field beats the sea level, and a community spread as thinly as the rest of the map never clears it.

Rather than port a second 1,600-line relaxation, this adds one force to the engine already in the tree: every entity is pulled toward its community's centre of mass, recomputed each tick. That is the minimum needed, and it is why communities have to be detected first.

## Deliberate divergences from macOS

| Divergence | Why |
|---|---|
| One added force on the existing d3 engine, not a bespoke relaxation | The bespoke layout exists to pack communities; one community-attraction force achieves what the island algorithm needs, and leaves the existing brain map untouched. |
| A purity floor is **enforced**, not just asserted | macOS asserts in its tests that every drawn region holds its own members, because its relaxation packs tightly enough that it is always true. Here it can fail, and a region drawn over ground none of its entities stand on is worse than no region. |
| No playback | Playback replays the map growing over time, and `graphDisplay.ts` records that the server graph carries no per-node timestamp. Deriving one from each node's memories is possible but is its own change. |
| No render-plan or snapshot cache | Those exist for a 4,000-line SwiftUI canvas under continuous camera motion; this redraws a few hundred shapes. |
| `ForceDirectedSimulation.swift` not ported | It is the legacy 3D engine for the old page, has no Atlas references, and uses a random source, which would break the determinism contract. |

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,339 passed, 18 skipped, 548 files. 127 of those are this feature; the branch point was 5,212.
- `npx electron-vite build` - built in 22.58s.
- `check_lifecycle_headers.py`, `check_brand_ui.py`, `check_arch_guardrails.py` - pass.

The component test records what reaches the canvas and asserts regions are drawn under entities, captions over both, the fill uses the even-odd rule matching the containment test, and every one of the thirty entities is drawn rather than capped.

The four fully-labelled zoom values are the ones macOS pins in its own performance harness tests (1 -> 16, 1946 -> 160, 2400 -> 180, 10000 -> 360), so they are a cited contract rather than a guess.

## Mutation audit

33 regressions seeded, **32 caught**. They cover every constant and guard in the package: the sea level, the decisive margin, relative reach, the area floor, the ring-vertex floor, even-odd containment, the caption tiers and truncation, the size floor, the purity floor, every zoom band and ceiling, the anchor exclusion, the community pull, and the deterministic seeds.

Two things worth reporting rather than burying.

**One escape is real, and it means an earlier commit in this branch was wrong.** That commit claimed the eager edge-crossing computation was a bug and that a NaN from a non-crossing edge was reachable from the neighbouring cell. The audit showed otherwise: making all four crossings eager again fails no test, and it cannot, because the two cells sharing an edge read the same two corner values and always agree on whether the contour crosses it. The misplaced coastlines that change was credited with fixing were caused entirely by the layout, which the same commit also changed. There is a commit correcting the record, and the lazy version stays as hygiene rather than as a fix.

**Several guards genuinely cannot be observed in the output**, and the code says so where they sit rather than pretending a test covers them: island ownership ties (a tied cell fails the decisive margin either way), the weightless-graph guard in Louvain (the gains become NaN and no node moves regardless), and the final dense relabelling (membership is already dense by then). They are kept so a later change to the surrounding rule cannot silently start depending on the wrong behaviour.

One redundancy the audit did find and fix: the self-edge guard existed both where the adjacency is built and again inside the modularity loop, so neither could be shown to matter. It is now in one place.

## What is not here

Playback, the caches, and macOS's label-dodging placement pass, which re-places region labels twice so they avoid each other. Labels currently sit at their region's centre.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11750?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/renderer/src/components/graph/KnowledgeGraphViewer.tsx
+++ b/desktop/windows/src/renderer/src/components/graph/KnowledgeGraphViewer.tsx
@@ -2,6 +2,7 @@ import { useMemo, useState } from 'react'
 import { ArrowLeft, Brain, Loader2, RefreshCw } from 'lucide-react'
 import type { KnowledgeGraph } from '../../../../shared/types'
 import { BrainGraph } from './LazyBrainGraph'
+import { MemoryAtlas } from './MemoryAtlas'
 import { EmptyState } from '../ui/EmptyState'
 import { capGraph, isCapped, DEFAULT_NODE_CAP, DEFAULT_LABEL_TOPK } from '../../lib/graphDisplay'
 
@@ -38,6 +39,11 @@ export function KnowledgeGraphViewer(props: {
   // showAll (4 states) — all labels of the key nodes, or all labels of the full
   // graph. Component state (not durable), matching showAll above.
   const [showAllLabels, setShowAllLabels] = useState(false)
+  // Two ways of reading the same graph. The graph hides nodes to stay legible;
+  // the map hides none and groups them into named regions instead. Component
+  // state, like the two toggles above, so neither view is imposed on a return
+  // visit.
+  const [view, setView] = useState<'graph' | 'map'>('graph')
 
   // The full graph stays reachable: showAll lifts the cap entirely (capGraph
   // returns the same graph ref, so the sim re-lays out the complete set). When
@@ -57,7 +63,8 @@ export function KnowledgeGraphViewer(props: {
     // blurred surface pins the GPU re-blending on every unrelated repaint (same
     // reason the Memories card uses a solid tint).
     <div className="relative h-full w-full overflow-hidden bg-black/40">
-      {hasGraph ? (
+      {hasGraph && view === 'map' && <MemoryAtlas graph={graph} centerNodeId={centerNodeId} />}
+      {hasGraph && view === 'graph' ? (
         <BrainGraph
           interactive
           graph={visibleGraph}
@@ -69,7 +76,7 @@ export function KnowledgeGraphViewer(props: {
           frameLoop="demand"
           labelMode={showAllLabels ? 'all' : 'declutter'}
         />
-      ) : (
+      ) : hasGraph ? null : (
         <div className="flex h-full w-full items-center justify-center">
           <EmptyState
             icon={Brain}
@@ -113,7 +120,21 @@ export function KnowledgeGraphViewer(props: {
           </span>
         </div>
         <div className="pointer-events-auto flex items-center gap-2">
-          {hasGraph && capsBite && (
+          {hasGraph && (
+            <button
+              onClick={() => setView((v) => (v === 'graph' ? 'map' : 'graph'))}
+              className="btn-ghost px-3 py-2"
+              title={
+                view === 'graph'
+                  ? 'Group your memories into named regions instead of hiding nodes'
+                  : 'Back to the connection graph'
+              }
+              aria-pressed={view === 'map'}
+            >
+              {view === 'graph' ? 'Map' : 'Graph'}
+            </button>
+          )}
+          {hasGraph && view === 'graph' && capsBite && (
             <button
               onClick={() => setShowAll((v) => !v)}
               className="btn-ghost px-3 py-2"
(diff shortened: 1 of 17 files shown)


aryanorastar:
Still the right gap on current main: the brain map still hides nodes behind DEFAULT_NODE_CAP; there is no named-region atlas.

Not approving this head — ~3306 behind, CI from 17 Aug. Rebase onto current main and re-run the Windows suite.

This is additive UI (Map/Graph toggle; existing viewer stays for onboarding). No merge from me until the rebased SHA is green.

aryanorastar:
Keeping this open. Not approving `58576aefaa1d`. `desktop/windows/src/renderer/src/lib/atlas/buildAtlas.ts` absent on origin/main. +3242.

Issue #11296, opened 2026-08-09
"My Apps" filters by authorship, so it is permanently empty for anyone who has not built an app


mikejsmith1985:
I have ADHD Assistant installed and enabled. It shows under the **Installed Apps** filter and
its detail page offers "Disable", so the install is real. Under **My Apps** I get "No apps found".

That is working as written. `backend/database/apps.py:112`:

```python
if my_apps:
    filters.append(FieldFilter('uid', '==', uid))
```

`my_apps` filters on authorship — apps whose `uid` is mine — not on apps I have. ADHD
Assistant is by "ace", so it is correctly excluded.

The problem is that nothing in the UI says so. The two filters sit next to each other as a
person icon ("My Apps") and a download icon ("Installed Apps"), and "my apps" is a natural
way to describe the apps you have. For anyone who has never authored an app — which I would
guess is the large majority of users — **"My Apps" is permanently empty and reads as broken**,
especially right next to an "Installed Apps" filter that works.

I hit this while trying to confirm an app was actually installed, and it cost me a while to
work out that both filters were behaving correctly.

Renaming it to something explicit — "Created by me", "My Creations", "Developer" — would make
the distinction obvious without touching behaviour. I did not send a patch because it is a
user-facing string across every ARB and the wording should be yours, not mine. Happy to do the
work once you pick the term.

Reading against `d60ab7a1c`. Investigated with Claude Code; the file:line above was verified
against that commit.

aryanorastar:
Closing as already fixed on current main.

#12432 Installed Apps primary + #11779 'Created by me' label. Authorship filter remains as the named created-by filter, not the empty default My Apps trap.

PR #11779, opened 2026-08-18
fix(app): disambiguate My Apps filter label from Installed Apps (#11296)

aryanorastar:
## Summary

The **My Apps** filter in the apps explore/install page and the manage-create page reads as "apps I have", but it actually filters by **authorship** — `backend/database/apps.py` applies `FieldFilter('uid', '==', uid)` when `my_apps=True`. For anyone who has never authored an app (the large majority), the filter is permanently empty and looks broken right next to a working **Installed Apps** tab.

**Fix:** rename the "My Apps" filter label to **"Created by me"** across all 49 locales. Internal filter identifier and backend behavior unchanged.

## Verification

- `flutter gen-l10n` content reproduced for all 49 locale getters (`myApps`), verified no untranslated messages.
- ARB files updated via `jq --indent 4` (1-line change per file); generated dart getters updated to the exact translated strings.
- `flutter analyze` clean on `lib/l10n/` + new test.
- `bash scripts/analyze_ratchet.sh` — passed.
- Not manually exercised in the running app UI (hosted/simulator); string-only change.

## Tests

New `app/test/unit/apps_filter_labels_test.dart` (2 tests):
- authorship and possession filter labels stay visually distinct (`myApps != installedApps`)
- authorship label no longer reads as a generic "My Apps"

```text
00:00 +2: All tests passed!
```

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11779?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks @aryanorastar — verified the diagnosis, and the fix holds together well.

**What I verified**

- The confusion is real: `backend/database/apps.py:112-113` applies `FieldFilter('uid', '==', uid)` when `my_apps=True`, so the filter is authorship, not possession — exactly as described in #11296. Renaming the label while keeping the identifier and backend behavior untouched is the right minimal fix.
- Both Flutter consumers pick the new string with no code change: `app/lib/pages/apps/explore_install_page.dart:590` and `app/lib/pages/apps/manage_create_page.dart:39` render `context.l10n.myApps` as chip labels, so the copy flows straight through.
- ARB ↔ generated Dart are consistent: `app_en.arb` now has `"myApps": "Created by me"`, matching `app_localizations_en.dart` (`String get myApps => 'Created by me';`) and the updated doc comment in `app_localizations.dart`; the `Generated Files` check passes, so the gen-l10n output matches all 49 ARB edits.
- Spot-checked several translations (de, es, fr, ru, zh, ja) — faithful renderings of the authorship meaning.

**One blocking fix**

- `app/test/unit/apps_filter_labels_test.dart` is missing the trailing newline at EOF, so `dart format` rewrites it and the Formatting check fails ("Formatted 51 files (1 changed)", exit 123). Running `dart format app/test/unit/apps_filter_labels_test.dart` and pushing should turn it green. The tests themselves are a nice touch — a cheap pin that the two filter labels stay distinct.

**Non-blocking nits**

- `app_he.arb`: `"נוצר על ידי"` reads as "created by" with the agent dropped; `"נוצר על ידיי"` would carry the "me". Worth a native-speaker pass over any locales you're fluent in.
- Web still shows a hardcoded "My Apps" (`web/app/src/components/apps/AppsExplorer.tsx`, ~line 560), so the same ambiguity remains on that surface. A follow-up issue/PR would keep the two from diverging — no need to expand this PR.

Once the newline lands this is in good shape. Since this renames user-facing copy in every locale, the final wording sign-off and merge belong to a human maintainer.

---

aryanorastar:
@Git-on-my-level — the blocking fix is in: `dart format app/test/unit/apps_filter_labels_test.dart` run locally and pushed (7478761cad). The only diff is the trailing newline at EOF, so the Formatting check should turn green.

On the non-blocking nits:
- I'll leave the he translation and the Web `My Apps` hardcode (web/app/src/components/apps/AppsExplorer.tsx) for follow-ups rather than expand this PR, as you suggested.
- Agreed the final wording sign-off belongs to a human maintainer; happy to adjust any strings once that lands.

Could you re-review / dismiss the blocking change so this can proceed to maintainer sign-off?

kodjima33 (approved):
Renames the authorship filter to 'Created by me' so it stops reading as 'apps I have' — string-only, backend filter unchanged, test pins the distinction. Fixes #11296.

PR #11787, opened 2026-08-18
fix(dev-harness): let OMI_DEV_HOST make the harness actually reachable

### LINKED ISSUE #11774: OMI_DEV_HOST cannot make the dev harness reachable from a physical device — the harness binds 127.0.0.1 unconditionally
The documented physical-device setup cannot work: `OMI_DEV_HOST` only changes what the app is compiled against, and the harness always binds loopback.

`docs/doc/developer/AppSetup.mdx:152` tells a contributor: *"On a physical iPhone, set `OMI_DEV_HOST` to your Mac's LAN address so the device can reach the harness."* `AppSetup.mdx:328` and `app/README.md:50` say the same. Following that literally cannot succeed, because nothing ever makes the harness listen on that address.

Verified on a clean clone of `main` at `015172400b`, harness started with `PROVIDER_MODE=offline make dev-up` (all seven services healthy).

**The harness binds loopback with a hardcoded literal.** `scripts/dev-harness/dev_harness/cli.py:767` starts the backend as:

```python
[sys.executable, "-m", "uvicorn", "main:app", "--host", "127.0.0.1", "--port", str(cfg.backend_port)],
```

The host is a string literal, not read from config or environment. The same is true of `llm-gateway` (`cli.py:753-758`) and `desktop-backend` (`cli.py:778-787`), and `cli.py:609` rewrites the Firebase emulator hosts to `127.0.0.1` when generating the harness's `firebase.json`.

**`OMI_DEV_HOST` is build-side only.** Across the whole repo it occurs in exactly two code locations, both in `app/setup.sh` (`:53`, `:55`), plus three doc locations (`app/README.md:50`, `AppSetup.mdx:152`, `:328`). It appears nowhere under `scripts/dev-harness/`. **The harness never reads it.** It changes the URL the app is compiled against and nothing else — so the app is pointed at an address where no service is listening.

**Measured, from the Mac itself:**

```
lsof -nP -iTCP:8000 -sTCP:LISTEN   ->  Python  TCP 127.0.0.1:8000 (LISTEN)
lsof -nP -iTCP:9099 -sTCP:LISTEN   ->  node    TCP 127.0.0.1:9099 (LISTEN)

curl http://192.168.2.81:8000/     ->  000 in 0.0039s   (refused; 192.168.2.81 is this Mac's LAN address)
curl http://127.0.0.1:8000/        ->  200 in 0.0021s
curl http://192.168.2.81:9099/     ->  000 in 0.0013s
curl http://127.0.0.1:9099/        ->  200 in 0.0009s
```

A phone pointed at the LAN address reaches nothing — the connection is refused before it leaves the Mac.

**This is not a one-line bind change, which is why I'm filing rather than PR-ing.** The loopback posture looks deliberate and fail-closed: `scripts/dev-harness/dev_harness/safety.py:181` defines `is_loopback_host`, `:194` defines `validate_loopback_emulator_host` ("host must point to loopback"), and it is enforced at `safety.py:210`, `safety.py:635` (Redis) and `desktop_profile.py:259-260` ("Firebase Auth emulator host must be loopback"). Binding the LAN interface would relax a guard someone added on purpose, so the call belongs to whoever owns that guard.

Two shapes, both reasonable:

1. **Docs-only** — state that a physical device needs a proxy/tunnel from the Mac's reachable address to `127.0.0.1:8000` and `:9099`, and drop the claim that `OMI_DEV_HOST` alone is sufficient. Cheapest, and honest about the current design.
2. **Opt-in bind** — an explicit `OMI_DEV_BIND_HOST` (or `OMI_DEV_HOST` also driving the bind) that binds the app-facing services to the LAN interface, with the safety guard widened deliberately to permit RFC 1918 while still refusing public addresses.

Whichever is chosen, `OMI_DEV_HOST` should not silently produce a build that cannot reach anything.

**Context / relation to other issues.** This is the complement to #11534, and it corrects an assumption in it: #11534 says "`setup.sh` honours `OMI_DEV_HOST` (`setup.sh:53`) for exactly this, but nothing tells a contributor they need it" — which is true of the *build* side, but setting it still does not yield a reachable harness. So the reachability advice in #11534 and in the docs is incomplete in the same way.

Worth noting for anyone who has had device builds working: a proxy in front of the harness (in my case `tailscale serve` mapping `:8000`/`:9099` to `127.0.0.1`) makes all of this work and completely hides the gap. Every device-side success I had recorded turned out to depend on that undocumented extra step, not on `OMI_DEV_HOST`.

Found while running the documented iOS pathway end-to-end from a fresh clone as a new contributor would.

formed2forge:
## Summary

`OMI_DEV_HOST` only ever changed what the app was compiled to expect (`app/setup.sh:53`) — the harness bound `backend`, `llm-gateway`, `desktop-backend`, and the Firebase emulators to a hardcoded `127.0.0.1` regardless, so a physical device pointed at the documented LAN address reached nothing.

The harness now resolves a `dev_bind_host` from `OMI_DEV_BIND_HOST` (falls back to `OMI_DEV_HOST`, defaulting to loopback), fails closed via a new `validate_dev_bind_host` guard unless the requested address is loopback, RFC 1918 (LAN), or `100.64.0.0/10` (Tailscale CGNAT) — never public — and binds `0.0.0.0` rather than the literal address so loopback callers (health checks, a booted simulator) keep working alongside the newly reachable LAN/tailnet address.

This is deliberately a **new, separate** guard: `validate_loopback_emulator_host` protects a different boundary (the backend's own connection to its co-located emulators, which must always stay loopback) and is untouched.

## Testing

- New unit tests in `test_safety.py` (loopback/LAN/CGNAT accepted, public rejected) and `test_cli.py` (env resolution priority/fallback, public address rejected at `load_config`, `_firebase_command` and `_start_app_services` both actually bind to the resolved host) — 94 passed, 6 skipped (platform-gated).
- **Verified live end to end**, not just in unit tests: with `OMI_DEV_HOST` set to this Mac's Tailscale address, `make dev-up` brings up all seven services healthy, and `curl http://<tailscale-ip>:8000/` and `:9099/` both return `200` — mirroring the issue's own repro, which measured a refused connection on the LAN address before this fix.
- Along the way, found and worked around two testing artifacts that are **not** defects in this change: a leftover `tailscale serve` config from earlier work occupying the same tailnet address/port, and lingering Firebase emulator processes from rapid manual iteration that `lsof -sTCP:LISTEN` didn't surface but `netstat` did.

Failure-Class: none

Fixes #11774.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11787?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -47,8 +47,9 @@ Before getting started, make sure your device is connected and unlocked. If you'
 
    `bash setup.sh ios` is the safe local-development path: it uses the local
    API/emulator harness and the `demo-omi-local` Firebase project. For a real
-   iPhone, set `OMI_DEV_HOST` to the Mac's LAN address when the local harness is
-   reachable from the device.
+   iPhone, set `OMI_DEV_HOST` to the Mac's LAN or Tailscale address before
+   running both `setup.sh ios` and `make dev-up` (export it so both commands
+   see it) — the harness now binds there too, not just the app build.
 
    iOS setup requires macOS/Xcode, so Windows developers should use the Android setup path.
 
--- a/docs/doc/developer/AppSetup.mdx
+++ b/docs/doc/developer/AppSetup.mdx
@@ -149,8 +149,13 @@ data requires the explicit `beta` profile (see [Mobile beta](https://github.com/
 <Tip>
 The automatic setup runs entirely on your machine — local API, local emulators,
 no production data — making it the fastest way to start building apps and making
-changes. On a physical iPhone, set `OMI_DEV_HOST` to your Mac's LAN address so
-the device can reach the harness.
+changes. On a physical iPhone, set `OMI_DEV_HOST` to your Mac's LAN or Tailscale
+address before running both `setup.sh ios` and `make dev-up` (in the same shell,
+or exported so both commands see it) — this now makes the harness itself listen
+there too, not just the app you build. Only a private address works (LAN, e.g.
+`192.168.x.x`, or Tailscale/CGNAT, e.g. `100.x.x.x`); a public address is
+rejected. Use `OMI_DEV_BIND_HOST` instead if you want the harness to bind a
+different address than the one the app is compiled to reach.
 </Tip>
 
 ---
@@ -326,8 +331,11 @@ ln -s -f ../../scripts/pre-commit .git/hooks/pre-commit
     Run `make dev-up` from the repo root first, and confirm with `make dev-status`.
 
     On a physical device, `127.0.0.1` is the phone — set `OMI_DEV_HOST` to your
-    Mac's LAN address before running `setup.sh` so the build points at your
-    machine. The Android emulator uses `10.0.2.2` by default.
+    Mac's LAN or Tailscale address before running **both** `setup.sh` (so the
+    build points at your machine) and `make dev-up` (so the harness actually
+    listens there — it defaults to loopback-only otherwise, which is why a
+    device build alone used to reach nothing). Export it in the same shell so
+    both commands see it. The Android emulator uses `10.0.2.2` by default.
   </Accordion>
   <Accordion title="Sign-in works but every request returns 401" icon="triangle-exclamation">
     The app and the backend are on **different Firebase projects**. Firebase ID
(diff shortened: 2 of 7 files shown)


kodjima33 (approved):
Makes OMI_DEV_HOST actually bind the harness services, fail-closed against public addresses via a new dedicated guard; tests + live end-to-end verification. Fixes #11774.

PR #11788, opened 2026-08-18
fix(backend): x-connector provider 5xx must not starve the remaining chunks in a cycle

aryanorastar:
## Summary

Follow-up hardening for the same failure class fixed in #11780 (strict memory
extraction provider 5xx): in `x_connector._extract_and_index`, one chunk's
`MemoryExtractionError` (strict flex lane) previously aborted the whole
per-user sync, so the batches behind it had to wait for the next hourly cycle
even though the provider fault only hit the first chunk.

Now the typed extraction failure is caught per chunk: the failed raw posts are
**not acknowledged** (they stay pending and the next sync replays them — the
existing raw-owns-progress invariant is untouched), the remaining chunks still
extract in the same cycle, and the skip is recorded via `record_fallback`
(`provider_5xx` → `chunk_deferred` → `degraded`). `PromotionFlexDeferred` and
untyped exceptions still propagate unchanged (the deferral and fail-closed
invariants pinned by the existing tests are preserved).

## Behavior change → test

- `test_provider_5xx_skips_only_the_failed_chunk_and_leaves_it_pending`: two
  forced chunks; the first raises `MemoryExtractionError(query)`, the second
  still runs; asserts total == 1, only chunk 2 is acknowledged, chunk 1 is left
  pending for the next sync, and the fallback event fires with
  `provider_5xx`/`chunk_deferred`/`degraded`.

## How I ran it

- `pytest tests/unit/test_x_memory_extraction_retry.py` → 5 passed (incl. the
  new regression; the `PromotionFlexDeferred` propagation and strict+fenced ack
  tests are unchanged)
- `pytest tests/unit/test_x_posts_kind_filter_limit_before_sort.py
  tests/unit/test_x_posts_list_endpoint.py
  tests/unit/test_integration_sync_telemetry.py` → 16 passed
- `pytest tests/unit/test_universal_memory_task_authority.py
  tests/unit/test_universal_memory_route_surfaces.py` → 9 passed
- `scripts/check_module_stub_pollution.py` / `scan_import_time_side_effects.py`
  / `check_isinstance_return_ratchet.py` / `scan_async_blockers.py` → clean
- `scripts/check_fallback_instrumentation.py` → no warnings
- `black --line-length 120 --skip-string-normalization` clean

Same local constraint as before: the locked `onnxruntime 1.19.0` has no macOS
arm64 wheel, so the full `uv pip sync` (and the pyright lane) run on the Linux
CI runner.

## Failure class (fixes)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11788?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_x_memory_extraction_retry.py
+++ b/backend/tests/unit/test_x_memory_extraction_retry.py
@@ -10,6 +10,7 @@ from fastapi import HTTPException
 os.environ.setdefault('ENCRYPTION_SECRET', 'omi_ZwB2ZNqB2HHpMK6wStk7sTpavJiPTFg7gXUHnc4tFABPU6pZ2c2DKgehtfgi4RZv')
 
 from models.memories import Memory, MemoryCategory
+from models.memory_contracts import MemoryExtractionError
 from utils import x_connector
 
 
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified the fix against both changed files — this is a clean, tightly-scoped follow-up to #11780's strict-lane hardening.

**`backend/utils/x_connector.py`** — in `_extract_and_index`, the new `try/except MemoryExtractionError` wraps only the `extract_memories_from_text` call, and the `continue` skips both `result_guard()` and the tail-of-loop `x_posts_db.mark_memory_extraction_completed` ack for the failed chunk. That's the right shape: the failed posts stay pending (raw-owns-progress is untouched, so the next hourly sync replays them) while the remaining chunks still extract in the same cycle. Cross-checked `utils/llm/memories.py`: `PromotionFlexDeferred` re-raises before the strict wrap and untyped exceptions still propagate out of `_extract_and_index`, so the deferral and fail-closed contracts pinned by the existing tests are preserved. The `record_fallback(...)` fields all land in the allowlists in `utils/observability/fallback.py`.

**`backend/tests/unit/test_x_memory_extraction_retry.py`** — the new regression forces two chunks via `MEMORY_BATCH_CHARS=10`, fails the first with `MemoryExtractionError`, and asserts the right outcomes: `total == 1`, both chunks attempted, only `post-2` acknowledged, one batch of one memory, and the `provider_5xx`/`chunk_deferred`/`degraded` fallback event. The pre-existing strict/fenced-ack and deferral tests are unchanged.

Two minor non-blocking notes:
1. Strict mode wraps *any* extraction failure as `MemoryExtractionError`, so e.g. an output-parse failure would also be counted under `reason='provider_5xx'` in `omi_fallback_total`. Behavior (skip chunk, stay pending, replay next cycle) is identical and safe; just be aware when reading that metric later.
2. On the legacy lane (`llm is None`) the except is unreachable since that path returns `[]` rather than raising — harmless, and it keeps the diff minimal.

Backend hermetic merge gate and E2E are green. Thanks for the hardening!

---

aryanorastar:
@Git-on-my-level — thanks for the positive-signal review. The PR is green (15 successful checks, hermetic gate + E2E passed) and mergeable, with no changes requested. Could you submit the formal approving review so this can proceed to maintainer merge? If you'd rather batch it, no rush — just wanted to surface that it's waiting only on the approval click.

kodjima33 (approved):
Per-chunk containment of a provider 5xx in the x-connector sync — root-caused, scoped to 2 files, regression test pins the new behaviour, CI green.

PR #7641, opened 2026-06-04
fix(ios): enable community builds — dynamic DEVELOPMENT_TEAM, APP_GROUP_IDENTIFIER, and stripped dev entitlements

formed2forge:
## Problem

Community developers with a paid Apple Developer account cannot build the dev flavor of the Omi iOS app out of the box. There are three blockers:

1. **Hardcoded team ID** — `project.pbxproj` has the BasedHardware team ID (`9536L8KLMP`) in all 9 dev-flavor build configurations. Xcode rejects signing for any other team.
2. **Bundle ID override in `devDebug.xcconfig`** — `APP_BUNDLE_IDENTIFIER=com.friend-app-with-wearable.ios12.development` is set here, but `Custom.xcconfig` (written by `setup.sh`) sets a machine-specific bundle at higher priority. The stale duplicate in `devDebug.xcconfig` causes confusion and overrides the dynamic value in some tool flows.
3. **Unprovisionable capabilities in dev entitlements** — `RunnerDebug/Profile/Release-dev.entitlements` request push notifications (`aps-environment`), associated domains (`h.omi.me`, `try.omi.me`), HotspotConfiguration, and Wi-Fi info. Community developers cannot provision these against their own team without modifying files, causing code-signing to fail with "entitlements do not match."

### Commit 1 — Dynamic DEVELOPMENT_TEAM and APP_GROUP_IDENTIFIER

- **`setup.sh`**: adds `detect_apple_team_id()` which auto-discovers the developer's team ID from their local provisioning profiles (with `APPLE_DEVELOPMENT_TEAM` env var override and interactive fallback). Writes both `DEVELOPMENT_TEAM` and `APP_GROUP_IDENTIFIER` (machine-specific, hostname-derived) to `Custom.xcconfig`.
- **`project.pbxproj`**: all 9 dev-flavor `DEVELOPMENT_TEAM` entries changed from literal `9536L8KLMP` to `$(DEVELOPMENT_TEAM)`, resolved at build time from `Custom.xcconfig`.
- **`devDebug.xcconfig`**: removes `APP_BUNDLE_IDENTIFIER` and `GOOGLE_REVERSE_CLIENT_ID` duplicates — both are already set correctly via `Custom.xcconfig` at higher xcconfig priority.
- **`BatteryWidget.entitlements`**, **`Runner.entitlements`**: use `$(APP_GROUP_IDENTIFIER)` instead of the hardcoded `group.com.friend-app-with-wearable.ios12`.
[...]

PR #11789, opened 2026-08-18
chore(dev): add personal build config overlay for local dev builds

formed2forge:
## Summary

Split out of #7641 at maintainer request — that PR mixed two unrelated concerns (dynamic `DEVELOPMENT_TEAM`/entitlements signing fix, and this personal-configs overlay), and splitting makes both easier to review and land independently.

Introduces a `.personal_configs/` convention at the repo root for contributors to store machine-local Firebase credentials and dev env config without committing them:

- `.personal_configs.example/` — templates for `.dev.env`, `GoogleService-Info.plist`, and `firebase_options_dev.dart`, with instructions for where to get real values
- `app/setup-personal.sh` — copies `.personal_configs/` into the app tree after `setup.sh`, with an optional ngrok-URL override for `API_BASE_URL`
- `.personal_configs/` added to `.gitignore` so real credentials never get committed

Carries the same content as #7641's commits `1295dc33a2`/`b246990fa9`, rebased onto current `main` — no functional changes from what was already reviewed there.

## Known nit (carried forward, not fixed here)

`setup-personal.sh` uses `sed -i ''` (BSD/macOS syntax) — fine since this is iOS-only tooling, but flagged in the original PR review as worth a comment if Linux contributors might use it. Left as-is to keep this a pure split with no new changes; happy to address in review if wanted.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11789?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for the clean split out of #7641 — I verified file-by-file that this PR stands on its own against `main`, with no dependency on the still-open #7641.

**`.gitignore`** — the single added line `.personal_configs/` sits below the `!.env*.template` negation block, so it doesn't disturb the env-template un-ignore rules, and nothing in that directory matches the earlier `*.env` / `.env*` patterns in a way that could re-include private files. Only the private copy is ignored; the committed example directory is unaffected. Correct placement.

**`.personal_configs.example/.dev.env.template`** — keys-only template (`OPENAI_API_KEY=`, `API_BASE_URL=`, `GOOGLE_MAPS_API_KEY=`, …) with no real values or real URLs. Matches the env keys the app's existing setup flow consumes via `.dev.env` + `build_runner`.

**`.personal_configs.example/GoogleService-Info.plist`** — structurally valid but *empty* plist (`<dict></dict>`) with comments pointing to the Firebase console for the real download. Good: it can't be mistaken for working credentials.

**`.personal_configs.example/firebase_options_dev.dart`** — comments only, including a genuinely useful warning against running `flutterfire configure` inside `app/` (which would overwrite the committed prod credentials). This matches a real hazard: `/app/lib/firebase_options_dev.dart` is gitignored on `main` precisely because it's generated per-machine.

**`app/setup-personal.sh`** — all operations are local `cp`/`sed`/`mkdir`/`grep`; no network access, no `curl`/`wget`, no eval of file contents, no writes outside the app tree. Destinations verified to exist as conventions on `main` independent of #7641: `ios/Config/Dev/GoogleService-Info.plist` and `ios/Runner/` (the Xcode pre-build script in `project.pbxproj` copies `Config/Dev` → `Runner` at build time), `lib/firebase_options_dev.dart` (already created/expected by `app/setup.sh` and `app/test.sh`), and `.dev.env` (consumed via `dart run build_runner build`, which the script's final hint correctly references). The optional ngrok-URL argument only rewrites `API_BASE_URL=` in the generated local `.dev.env` — machine-local, never committed.

Two small non-blocking notes:
- `sed -i ''` is BSD/macOS-only and will fail on GNU sed (Linux). You already flagged this as a carried nit; fine for iOS-focused tooling, just noting Linux contributors will hit it.
- Writing the same plist to both `ios/Config/Dev/` and `ios/Runner/` is harmless (Runner is refreshed from Config at build time anyway) and keeps on-disk copies consistent with what `app/setup.sh` does on `main`.

Checks that ran (Dart Analyze & Tests, Formatting, Hygiene, Backend Hermetic Merge Gate) all pass; the skipped heavier jobs are expected for non-app-code changes.

Nice developer-experience contribution — the warning about `flutterfire configure` overwriting prod credentials alone will save someone a bad afternoon. Leaving for human maintainer review before merge.

---

formed2forge:
All CI checks are green and the review above raised no blocking issues — this is ready for maintainer merge whenever convenient. I don't have write access to merge it myself.

kodjima33 (approved):
Pure split of #7641 at maintainer request — gitignored .personal_configs overlay plus empty templates; no real credentials, no product code.

PR #11793, opened 2026-08-18
fix(app): pin an iOS build destination and validate prerequisites

### LINKED ISSUE #11775: setup.sh ios never pins a device and validates none of the prerequisites it prints
Two related gaps in `app/setup.sh`, both verified by reading `main` at `015172400b`. Neither is a niche edge case: they are what a new contributor hits before any of the app code matters.

## 1. `run_build_ios()` never pins a target device

`app/setup.sh:241-249`:

```bash
function run_build_ios() {
  local flavor="${1:-dev}"
  shift || true
  flutter pub get \
    && pushd ios && pod install --repo-update && popd \
    && dart run build_runner build \
    && flutter run --flavor "$flavor" "$@"
}
```

There is no `-d`, so Flutter picks a device itself. When I ran `bash setup.sh ios` on a Mac with **no iOS simulator runtime installed** and only a *wirelessly* paired iPhone visible, Flutter selected the **macOS desktop** target and the run died:

```
Launching lib/main.dart on macOS in debug mode...
Error: No macOS desktop project configured. See https://flutter.dev/to/add-desktop-support
```

Exit code 1 — so it does fail loudly, which is good; the problem is that a command explicitly told `ios` spent a full `pod install --repo-update` and build_runner pass and then reported an error about macOS, a platform the contributor never mentioned. The diagnosis ("you have no simulator runtime") is nowhere in the output.

**Attribution note, so this is not misread:** I was on Xcode 27.0 / Flutter 3.44.9, ahead of the versions the script prints (Xcode 16.4). *Which* fallback target Flutter chooses is version- and environment-dependent. The gap being reported is the version-independent one: `run_build_ios()` pins no device and the script never checks that an iOS destination exists, so the failure mode is "silently build something else" rather than "tell me what's missing."

Suggested shape: pass `-d` explicitly, or fail up front with something like *"no iOS device or simulator found — boot a simulator or connect a device."*

## 2. `setup.sh` prints a prerequisite list and validates none of it

`setup.sh:29-51` prints:

```
For iOS Developers:
- Xcode (v16.4)
- CocoaPods (v1.16.2)
```

…and the entire script contains exactly one `command -v` check — for `fastlane`, at `setup.sh:145`, inside `setup_provisioning_profile()`. Nothing verifies Xcode, CocoaPods, Flutter, or their versions. A missing or wrong-version prerequisite therefore surfaces as a confusing downstream failure instead of a named error.

**The repo already has the pattern worth copying, in the harness.** `make dev-up` refuses to start and names every missing prerequisite *with both remedies*:

```
Cannot start; missing prerequisites:
  - OPENAI_API_KEY (openai; set in backend/.env.local-dev or use PROVIDER_MODE=offline)
  ...
  - docker daemon (docker CLI present but `docker info` failed; start Docker/colima
    or install typesense-server for OMI_TYPESENSE_RUNTIME=native)
```

That is exemplary — it names the thing, where to set it, and the escape hatch. (Both of those are real outputs I hit today; the second is what a fresh reboot with colima stopped produces.) `setup.sh` never got the same treatment.

Both gaps push the same cost onto newcomers: the script knows what it needs, prints what it needs, and then declines to check, so the first signal of trouble is an unrelated error several minutes into a build.

Found while running the documented iOS pathway end-to-end from a fresh clone as a new contributor would. Related but distinct: #11534 (build succeeds, then hangs on splash when the harness isn't running) and #11774 (`OMI_DEV_HOST` cannot make the harness reachable from a device).

formed2forge:
## Summary

Two related gaps in `app/setup.sh`, both hit before any app code matters:

1. **`run_build_ios()` never pinned a target device.** With no `-d`, Flutter picked a destination itself. On a machine with no iOS simulator runtime installed and only a wirelessly-paired phone visible, it silently built for **macOS desktop** instead — after a full `pod install --repo-update` and `build_runner` pass — and failed with an unrelated "No macOS desktop project configured" error.
2. **`setup.sh` prints a prerequisite list but validates almost none of it.** Xcode v16.4, CocoaPods v1.16.2, Flutter v3.44.5 are documented in the script's own header, but the whole script has exactly one `command -v` check (for `fastlane`). A missing or outdated tool surfaces as a confusing downstream failure instead of a named error — unlike the dev harness's own `Cannot start; missing prerequisites:` pattern, which names each gap with a remedy.

## What changed

- `select_ios_device()` enumerates iOS-platform destinations from `flutter devices --machine` (via `jq`): returns the one candidate directly, prompts when there are several (failing fast without a TTY — same reasoning as `detect_apple_team_id`'s existing prompt), and fails with a named error instead of a silent fallback when there are none.
- `check_ios_prerequisites()` validates Flutter/Xcode/CocoaPods/`jq` against the versions the script already documents, listing every gap at once with a remedy, before any of the slow setup steps run.
- `run_build_ios()` now calls both before doing any real work, and passes the resolved `-d <device_id>` through to `flutter run`.

## Testing

- New shell tests: `app/test/shell/ios_device_selection_test.sh` (single device returns directly; zero devices fails named; multiple devices with no TTY fails fast, not hangs; empty device list fails named) and `app/test/shell/ios_prerequisites_test.sh` (`_version_at_least` comparisons; each of the four prerequisites individually missing/outdated with the correct remedy; all four missing at once).
- **Verified live, not just in the new tests** — ran the real `bash setup.sh ios` three ways:
  - Non-interactively with two real devices connected (a physical iPhone + a booted simulator) → correctly failed fast with the multi-device prompt rather than hang.
  - Through a real pty, feeding the interactive prompt "2" → correctly proceeded through `pod install`/`build_runner` and reached `Launching lib/main.dart on iPhone 17 Pro in debug mode...` — the exact device chosen.
  - With a stubbed toolchain reporting only a macOS destination → correctly failed with the named "No iOS device or simulator found" error instead of the original silent-fallback bug.

Failure-Class: none

Fixes #11775.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/setup.sh
+++ b/app/setup.sh
@@ -235,16 +235,132 @@ function run_build_android() {
     && flutter run "${flutter_args[@]}"
 }
 
+# #####################################
+# iOS prerequisite and device selection
+# #####################################
+
+# True if $1 (a dotted version, e.g. "16.4") is >= $2. Relies on `sort -V`,
+# which Apple's /usr/bin/sort on macOS supports (verified; this is iOS-only
+# tooling, so a non-macOS sort is not a concern).
+function _version_at_least() {
+  local have="$1" want="$2"
+  [ "$(printf '%s\n%s\n' "$want" "$have" | sort -V | head -n1)" = "$want" ]
+}
+
+# Named checks with remedies, matching the harness's own pattern
+# (scripts/dev-harness's `Cannot start; missing prerequisites:` block) instead
+# of letting a missing/outdated tool surface as a confusing downstream failure
+# several minutes into a build.
+function check_ios_prerequisites() {
+  local missing=()
+
+  local flutter_version
+  flutter_version=$(flutter --version 2>/dev/null | grep -oE '^Flutter [0-9]+\.[0-9]+\.[0-9]+' | awk '{print $2}')
+  if [[ -z "$flutter_version" ]]; then
+    missing+=("Flutter SDK (v3.44.5 or later) — install from https://docs.flutter.dev/get-started/install")
+  elif ! _version_at_least "$flutter_version" "3.44.5"; then
+    missing+=("Flutter ${flutter_version} found, but v3.44.5 or later is required — run: flutter upgrade")
+  fi
+
+  if ! command -v xcodebuild &>/dev/null; then
+    missing+=("Xcode (v16.4 or later) — install from the App Store, then run: sudo xcode-select --switch /Applications/Xcode.app")
+  else
+    local xcode_version
+    xcode_version=$(xcodebuild -version 2>/dev/null | awk '/^Xcode/ {print $2; exit}')
+    if [[ -n "$xcode_version" ]] && ! _version_at_least "$xcode_version" "16.4"; then
+      missing+=("Xcode ${xcode_version} found, but v16.4 or later is required — update via the App Store")
+    fi
+  fi
+
+  if ! command -v pod &>/dev/null; then
+    missing+=("CocoaPods (v1.16.2 or later) — install with: sudo gem install cocoapods")
+  else
+    local pod_version
+    pod_version=$(pod --version 2>/dev/null)
+    if [[ -n "$pod_version" ]] && ! _version_at_least "$pod_version" "1.16.2"; then
+      missing+=("CocoaPods ${pod_version} found, but v1.16.2 or later is required — update with: sudo gem install cocoapods")
+    fi
+  fi
+
+  if ! command -v jq &>/dev/null; then
+    missing+=("jq (used to select an iOS build destination) — install with: brew install jq")
+  fi
+
+  if [[ "${#missing[@]}" -gt 0 ]]; then
+    echo "❌ Cannot build for iOS; missing prerequisites:" >&2
+    local item
+    for item in "${missing[@]}"; do
+      echo "   - ${item}" >&2
+    done
+    return 1
+  fi
+}
+
+# Picks the iOS destination `flutter run -d` should target, instead of leaving
+# Flutter to fall back to whatever else it finds (macOS desktop, on a machine
+# with no simulator runtime and only a wirelessly-paired phone visible) and
+# reporting an error about a platform the developer never asked for.
+function select_ios_device() {
+  local devices_json
+  devices_json=$(flutter devices --machine 2>/dev/null) || {
+    echo "❌ Could not query connected devices (flutter devices --machine failed)." >&2
+    return 1
+  }
+
+  local ios_devices
+  ios_devices=$(echo "$devices_json" | jq -c '[.[] | select(.targetPlatform == "ios")]')
+  local count
+  count=$(echo "$ios_devices" | jq 'length')
+
+  if [[ "$count" -eq 0 ]]; then
+    echo "❌ No iOS device or simulator found." >&2
+    echo "   Boot a simulator (open -a Simulator) or connect a physical device, then retry." >&2
+    return 1
+  fi
+
+  if [[ "$count" -eq 1 ]]; then
+    echo "$ios_devices" | jq -r '.[0].id'
+    return 0
+  fi
+
+  echo "⚠️  Multiple iOS destinations found. Choose one:" >&2
+  local i=0 line
+  while IFS= read -r line; do
+    i=$((i + 1))
+    echo "   $i) $line" >&2
+  done < <(echo "$ios_devices" | jq -r '.[] | "\(.name) (\(.id))"')
+
+  # Same reasoning as detect_apple_team_id's prompt: never block on read
+  # without a TTY, or a non-interactive run (CI, nested automation) hangs.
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Thanks @formed2forge — reviewed head `2847da1`. Well-scoped fix with real regression tests; I verified the key claims and ran the new suites.

**Verified against main and the head:**
- `run_build_ios()` on main really passes no `-d` to `flutter run`, so Flutter's silent destination pick (macOS desktop when no simulator runtime is installed) is a real failure mode, not a hypothetical.
- The enforced floors in `check_ios_prerequisites()` (Flutter 3.44.5, Xcode 16.4, CocoaPods 1.16.2) match `setup.sh`'s own header exactly — the check enforces documented policy rather than inventing new minimums.
- `_version_at_least()`'s `sort -V` reliance: macOS's BSD-lineage `/usr/bin/sort` documents `-V, --version-sort`, and any Mac that can run the required Xcode 16.4 has it — the comment's portability claim holds.
- The `Cannot start; missing prerequisites:` precedent cited in the comment exists in `scripts/dev-harness/dev_harness/cli.py`.
- Ran both new suites hermetically against the head's `setup.sh`: `app/test/shell/ios_device_selection_test.sh` 4/4 (single device, zero devices, multi-device no-TTY fail-fast, empty machine list) and `app/test/shell/ios_prerequisites_test.sh` 10/10 (version comparisons, all-met, each named failure, all-four-missing-at-once). The function-extraction + stubbed-PATH harness design is genuinely behavioral — nice.

**Non-blocking notes:**
- `check_ios_prerequisites()`: if `xcodebuild` exists but is broken (license unaccepted, missing components), `xcodebuild -version` returns nothing and the `[[ -n "$xcode_version" ]]` guard skips the check — a broken install still passes the gate and fails downstream. A named "xcodebuild present but not usable" message would close that gap.
- `select_ios_device()`'s comment cites `detect_apple_team_id`'s prompt as precedent, but that function doesn't exist in `setup.sh` on main (it only appears as a stub in `app/scripts/generate_ios_custom_config_bundle_id_test.sh`). The no-TTY fail-fast itself is right; the reference is just stale.
- jq is now a hard prerequisite for every iOS build. Reasonable given the named remedy, but it's one new install step for contributors with previously-working builds; and for brew-installed CocoaPods the remedy `sudo gem install cocoapods` should be `brew upgrade cocoapods`.
- Neither suite is wired into CI (`mobile-app-checks.yml` runs `flutter test` only). That matches the existing `app/scripts/*_test.sh` convention, so not blocking — a small ubuntu runner (bash+jq) would make these stick.

What I can't verify from here is the macOS behavior itself. Before this becomes the default `setup.sh ios` path, one on-Mac pass of the prereq-failure messages and the multi-device prompt would confirm the UX end-to-end.

---
*by AI on behalf of David — needs one on-Mac validation run of the new `setup.sh ios` flow and a maintainer call on making jq a hard iOS prerequisite; please @Git-on-my-level if David is needed.*

formed2forge:
Thanks for the thorough review, @Git-on-my-level — addressed all three notes in a follow-up commit (`500e9762`):

1. **Broken/unlicensed Xcode silently passing** — fixed. `check_ios_prerequisites()` now explicitly names "xcodebuild is on PATH but not usable" when the version line can't be parsed, instead of treating that as "nothing to check." New test confirms this was a real gap: it fails (rc=0, silent pass) against the pre-fix check, passes against the fix.
2. **Stale `detect_apple_team_id` reference** — removed. You're right that it doesn't exist on `main`; the comment now just states the reasoning directly.
3. **CocoaPods update remedy** — now names both `brew upgrade cocoapods` and `sudo gem install cocoapods`, since gem-installing over a Homebrew-managed CocoaPods doesn't actually update what's on `PATH`.

On the two open questions:
- **jq as a hard prerequisite**: agreed this is a real, if small, cost — happy to defer to a maintainer call on whether that's acceptable, per your note.
- **On-Mac validation of the actual UX**: this was tested live, not just in the shell test suites — three real runs of `bash setup.sh ios` / `flutter run`, covering the multi-device-no-TTY fail-fast path (hit unplanned, with a real wirelessly-reconnected phone), the interactive prompt through an actual pty (selected a device by number, watched it build and reach `Launching lib/main.dart on <chosen device>`), and the original bug's exact repro (stubbed toolchain reporting only macOS, confirmed the new named error fires instead of the old silent fallback). Details are in the PR description.

kodjima33 (approved):
setup.sh only: pins an iOS build destination instead of silently falling back to macOS, and validates the prerequisites the script already documents. Shell tests + live verification. Fixes #11775.

PR #11802, opened 2026-08-18
Windows Rewind video chunks: stop storing a JPEG per second of screen time

ZachL111:
Windows Rewind stores one JPEG per captured frame. Capture runs at one frame per second and retention defaults to 14 days, so the folder grows without any bound the user can see. This packs older frames into inter-frame-compressed video chunks, which is what macOS has done since the start.

## The measurement

Taken on this machine's own capture path at the shipped `standard` quality tier (1280x720, JPEG quality 0.7), over a real 60-second run of ordinary desktop work:

| Storage | Per frame | Per 60s chunk |
|---|---|---|
| JPEG (what ships today) | 115.8 KB | 6.95 MB |
| H.264 chunk @ 400 kbps | **2.26 KB** | **136 KB** |
| VP9 chunk @ 400 kbps | 3.35 KB | 201 KB |

51x, with all 60 frames decoding back at 37 to 41 dB PSNR against their sources. The saving is that large because consecutive frames of a screen are nearly identical, which is exactly the redundancy a JPEG-per-frame store cannot use and an inter-frame codec exists for. Scaled to the shipped defaults that is roughly 25 GB against roughly 0.5 GB for a fortnight of active use.

## Compaction, not live encoding

This is the deliberate divergence and everything else follows from it.

macOS encodes live: `VideoChunkEncoder.addFrame` takes each frame as it is captured and appends it to the chunk being written. Windows compacts after the fact. Capture keeps writing JPEGs exactly as it always has, and a background pass every 30 minutes packs runs older than 30 minutes.

Live encoding puts the codec in the capture path, and a bug there costs frames that no longer exist anywhere. Compacting instead means capture is untouched, a failed compaction leaves the JPEGs where they were, turning it off changes nothing about how frames are stored, and the backlog already on disk gets compacted too.

It also removes most of the machinery macOS needs. Because macOS encodes live, its rows necessarily exist while the chunk is still being written, so it carries a sidecar journal of abandoned writers, a quarantine table, and a generation/reservation ownership model to stop a stale finaliser clearing a newer writer's state. Here nothing references a chunk until it is finished, so none of that is load-bearing. The tombstone table is kept, because a chunk can still be found corrupt at read time long after it was written.

## The ordering that makes it safe

```
write chunk  ->  read it back and verify  ->  claim frame  ->  delete its JPEG
```

Every prefix is a safe place to crash.

| Crash after | State | Who cleans up |
|---|---|---|
| write | an unreferenced chunk file | the chunk sweep, after its grace period |
| verify | same | same |
| a claim | claimed frames read from a chunk proven to hold them; unclaimed frames still have their JPEGs | nothing to clean, the chunk holds a superset |
| a claim, before the delete | an orphaned JPEG | the existing orphan sweep |

Verify re-reads from disk rather than checking the buffer just written, so it also catches a write that never landed or landed short, and it checks the capture timestamp at every offset. That last check is the one that matters: a chunk with the right frame count but a shifted order would make every read return a neighbouring frame, silently.

## Rules worth knowing

**Only OCR'd frames are compacted.** The OCR backlog sweep reads `image_path` for `indexed = 0` frames and, when the file is missing, marks the frame indexed with empty text rather than failing (`ocrService.ts`). Compacting an un-OCR'd frame would therefore not error, it would quietly cost that frame its searchable text forever. This is the reason the compactor cannot simply take the oldest frames.

**Runs shorter than 8 frames are left alone.** Measured, a 5-frame run reached only 2.4x against a 60-frame run's 51x, because the opening key frame is most of a short chunk.

**Presentation timestamps come from the frame index, not the capture clock.** macOS derives them from a live capture rate and documents what that cost: a mid-chunk rate change made a later frame's timestamp fall below an already-appended one, which `AVAssetWriter` rejects as non-monotonic, dropping frames and sometimes the whole chunk. An index is monotonic by construction, and the real capture time travels per record in the container.

**The read cursor is one-way.** A chunk has no random access, so reopening per request makes a scrub quadratic. macOS measured 728 ms to walk an 18-frame chunk reopening each time against 59 ms keeping the reader alive, 12.3x and pixel-identical. The decision of when the open decoder can be reused is a pure state machine; the decoding that executes it needs a codec and lives in the renderer.

**A chunk is self-describing.** `rebuildIndex.ts` exists because Windows can re-create `rewind_frames` rows from `<day>/<ts>.jpg` after a database wipe. Compaction would have quietly taken that away from every frame it packed, leaving intact pixels that nothing references and that the sweep would then delete, so each record carries the timestamp its filename used to.

## Why a custom container

`.omichunk` rather than MP4. AVFoundation is a muxer-first API, so macOS gets a container whether it wants one or not. WebCodecs is codec-first, emitting bare `EncodedVideoChunk`s and accepting them back, so an MP4 here would mean writing a muxer and a demuxer to put bytes into a box structure and immediately take them out again. Frame N is record N, which is the same number stored in `chunk_offset`, and parsing is defensive: a truncated file raises rather than returning the frames it managed to read, because the compactor treats "reads back exactly" as the precondition for deleting anything.

## What running the real codec found

Three defects that reasoning did not, and all three would have shipped broken:

- `createImageBitmap` taken as a bare property reference throws "Illegal invocation". It is a method on the global object and loses its receiver.
- The container used Node's `Buffer`, which does not exist in a renderer.
- Flushing the decoder between reads breaks the cursor on the very next step, because WebCodecs requires a key frame after a flush. The tape now decodes incrementally and only flushes at the end of a chunk.

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,414 passed, 18 skipped, 552 files. 179 of those are this feature; the branch point was 5,235.
- `npx eslint .` - 0 errors.
- `npx electron-vite build` - clean.
- `check_lifecycle_headers.py`, `check_brand_ui.py`, `check_arch_guardrails.py` - pass.
- `node --test e2e/rewind-chunks.spec.mjs` - passes against a real H.264 encoder.

The E2E is the only place the central claim can be made honestly, so it drives the production encoder and reader inside a real Electron renderer rather than a copy of them: 60 frames in, 60 out, 11,687 KB of JPEGs into a 1,813 KB chunk, worst mean per-channel pixel difference 1.69 of 255. Its frames are deliberately adversarial, dense high-entropy text with a full line changing every frame, which is why it measures 6.4x where real capture output measured 51x.

## Mutation audit

67 regressions seeded, 62 caught. Three findings are worth reporting rather than burying.

**One test could not fail.** It sized its fixtures from the very constant it was asserting, so lowering the minimum chunk length from 8 to 1 shrank the fixtures with it and the assertion still passed. It pins the literal now.

**Three guards were masking each other.** The compaction candidate query had two clauses that both excluded the same row, so either alone looked sufficient. The offset ordering was asserted with offsets that happened to ascend with rowid, so SQLite returned the right answer with the `ORDER BY` deleted. A chunk-listing extension filter was already subsumed by the path validator two lines below it. The first two now have cases that separate them; the third is deleted, because a guard that cannot be shown to matter is a guard that will drift.

**Four corruption cases in the container had no coverage at all** - a zero frame count, an empty declared geometry, an out-of-range header length, and the header scan's own frame ceiling. All four do now.

Five mutations still escape, and all five are defence in depth whose effect is genuinely unobservable today: the path containment check and the whitespace check, both behind a regex that already rejects what they catch; the atomic write's rename, which only shows up if the process dies mid-write; the day-directory filter, behind the same path validator; and the candidate query's id tiebreak, which SQLite happens to satisfy anyway. Each is documented where it sits with why it is kept, rather than covered by a test that would only be asserting the mutation.

## What is not here

Live encoding, and with it macOS' aspect-ratio debounce, frame-rate freeze, buffer-overflow emergency reset and writer-not-ready retry ladder, all of which exist to keep a live encoder healthy in the capture path. The reservation and generation ownership model, for the reason above. macOS' 3000px downscale cap, which would never bind because Windows already caps captured frames by quality tier at 2560x1440.

OCR cannot read a chunk-backed frame, which only matters after a database wipe: rebuilt chunk rows come back with their pixels but without their text, because the text lived only in the wiped database and was never written into a chunk. Teaching OCR to decode chunks needs a renderer and is its own change.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11802?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/e2e/rewind-chunks.spec.mjs
+++ b/desktop/windows/e2e/rewind-chunks.spec.mjs
@@ -1,0 +1,116 @@
+// The real WebCodecs round trip, in a real Electron renderer.
+//
+// Everything else about compaction is unit-tested in plain node: the container's
+// bytes, the grouping, the SQL, the write/verify/claim ordering, the cursor's
+// state machine. What none of that can cover is whether an actual H.264 encoder
+// and decoder, driven by the production modules, turn frames into a small chunk
+// and give the same frames back. That claim is the entire reason this feature
+// exists, and this is the only place it can be made honestly.
+//
+// The hooks it drives (`window.__omiRewindChunkE2E`) call the PRODUCTION
+// `encodeFramesToChunk` and `ChunkFrameReader`, not a copy — see
+// src/renderer/src/rewind/chunkE2EHooks.ts.
+import test from 'node:test'
+import assert from 'node:assert/strict'
+import path from 'node:path'
+import { mkdtempSync, rmSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { fileURLToPath } from 'node:url'
+import { _electron as electron } from 'playwright'
+
+const here = path.dirname(fileURLToPath(import.meta.url))
+const mainEntry = path.join(here, '..', 'out', 'main', 'index.js')
+
+const WIDTH = 1280
+const HEIGHT = 720
+const FRAMES = 60
+
+/**
+ * The first window with the chunk hooks installed.
+ *
+ * Any window will do — the hooks are installed by both the main renderer and
+ * the capture window — so this takes whichever is ready first rather than
+ * waiting on the capture window, which only opens while capture is running.
+ */
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type -- plain .mjs; the repo's other e2e specs are the same shape
+async function captureWindow(app) {
+  const deadline = Date.now() + 60_000
+  for (;;) {
+    for (const win of app.windows()) {
+      const installed = await win
+        .evaluate(() => Boolean(window.__omiRewindChunkE2E))
+        .catch(() => false)
+      if (installed) return win
+    }
+    if (Date.now() > deadline) throw new Error('timed out waiting for the capture window')
+    await new Promise((r) => setTimeout(r, 250))
+  }
+}
+
+test('encodes real frames into a chunk and reads every one of them back', async (t) => {
+  const dir = mkdtempSync(path.join(tmpdir(), 'omi-e2e-chunks-'))
+  const app = await electron.launch({
+    args: [mainEntry, `--user-data-dir=${dir}`],
+    env: { ...process.env, OMI_E2E: '1', OMI_AUTOMATION: '0', OMI_SKIP_TUNNEL: '1' }
+  })
+  t.after(async () => {
+    try {
+      await app.close()
+    } catch {
+      /* already closed */
+    }
+    rmSync(dir, { recursive: true, force: true })
+  })
+
+  const win = await captureWindow(app)
+
+  const codec = await win.evaluate(
+    ([w, h]) => window.__omiRewindChunkE2E.codecFor(w, h),
+    [WIDTH, HEIGHT]
+  )
+  // A machine with no usable encoder is a supported state — it just never
+  // compacts — but the harness machine is not allowed to be one silently.
+  assert.ok(codec, 'expected this machine to support at least one chunk codec')
+
+  const result = await win.evaluate(
+    ([n, w, h]) => window.__omiRewindChunkE2E.roundTrip(n, w, h),
+    [FRAMES, WIDTH, HEIGHT]
+  )
+
+  // --- Every frame came back. ---
+  // The property the compactor bets the JPEGs on. A chunk that returned 59 of
+  // 60 frames would have cost the user the sixtieth permanently.
+  assert.equal(result.framesBack, FRAMES, 'every frame must decode back out of the chunk')
+
+  // --- It is dramatically smaller. ---
+  // The generated frames are deliberately adversarial: dense high-entropy text
+  // with a full line changing every frame, which is far worse for both JPEG and
+  // inter-frame prediction than a real screen. This run measures about 6.4x
+  // against the 51x measured on real capture output (see
+  // main/rewind/chunks/ARCHITECTURE.md). The floor is set well below what the
+  // adversarial case achieves, so it asserts "inter-frame compression is
+  // working" without pinning a number that varies with the encoder build.
+  const ratio = result.jpegBytes / result.chunkBytes
+  assert.ok(
+    ratio > 3,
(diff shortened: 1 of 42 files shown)


ZachL111:
Merge-order note for whoever lands this, found while sweeping my open PRs for mergeability.

This PR and #11743 (Windows search) both add a database migration numbered `version: 3`:

- #11743 adds `memories_fts_backfill`
- #11802 adds `rewind_video_chunks`

Each is clean against main on its own, so GitHub reports both as mergeable and will keep doing so right up until the second one lands. The collision is in `desktop/windows/src/main/ipc/dbMigrations.ts`.

This one is worth resolving carefully rather than by hand-merging the hunk, because the failure is not cosmetic. `runMigrations` enforces contiguity:

```
if (m.version !== i + 1) {
  throw new Error(`migrations must be contiguous from 1; found version ${m.version} at index ${i}`)
}
```

So a resolution that keeps both entries at 3, or that leaves a gap, throws on every launch for every user rather than failing visibly in CI.

Whichever merges second needs its migration renumbered to 4 and its own test updated. Nothing else about either change interacts; the two migrations touch different tables.

Happy to do the rebase on whichever of the two you want to land second, once the order is decided.

aryanorastar:
Still the right gap on current main: Rewind ingest still writes one JPEG per frame; migrations stop at version 2.

Not approving. Head is ~3221 commits behind, and this PR’s migration is also version 3 — same slot as #11743 `memories_fts_backfill`. Rebase onto current main, take the next free version, and re-run Windows CI plus the rewind-chunks e2e on the new SHA.

Nothing from me to merge. Compaction is local Rewind storage, but it still rewrites the screen-capture corpus; rebase before anyone treats MERGEABLE as landable.

aryanorastar:
Keeping this open. Not approving `a7f9b7f94c25`. `desktop/windows/e2e/rewind-chunks.spec.mjs` absent on origin/main. +4940. Desktop screen storage, not BLE.

PR #11801, opened 2026-08-18
feat(desktop): hands-free wake word to command the assistant during ambient listening

aryanorastar:
## Summary

> **Status note.** #12181 (pause endpointing) has merged, so the first of the two always-on
> surfaces is off this PR. Echo suppression stays, and is load-bearing rather than riding along —
> `c5f95d9` removes the wake word's own "assistant is speaking" guard precisely because
> `VoicePlaybackEchoPolicy.classify` runs ahead of `WakeWordService.observe`.

Adds an opt-in **Wake Word** feature for macOS desktop: during ambient listening, say `"Omi"` followed by a command and the assistant runs it hands-free — no clicks, no Push-to-Talk.

## What changed

- **`Sources/WakeWord/WakeWordService.swift`** (new) — trigger engine: strips the wake phrase, extracts the command, and submits it to the assistant (`openAIInputWithQuery`). Guards: 30s cooldown against rapid repeats, per-segment-ID deduplication, user-speech only, and suppression while the assistant is already busy.
- **`Sources/WakeWord/WakeWordSegmentParser.swift`** (new) — splits "Omi, do X" into wake phrase + command; requires a 2+ word command so a bare "Omi" never misfires.
- **`AppState+ListenEvents.swift`** — feeds every incoming transcript segment into `WakeWordService.observe(...)`, the single funnel for ambient speech.
- **`AssistantSettings.swift`** — persisted `wakeWordEnabled` / `wakeWordPhrase` / `wakeWordCooldown` (default **off** so current users' behavior is unchanged).
- **`SettingsContentView+General.swift`** — new **Wake Word** settings card with on/off toggle and a dynamic subtitle tied to the audio-recording mode ("Listens in the background during meetings and calls").
- **Tests** — `WakeWordServiceTests` + `WakeWordSegmentParserTests` (15 cases).

## How it works

```
ambient transcript segment ("Omi, order asian food")
  → WakeWordService.observe
  → WakeWordSegmentParser extracts wake phrase + command
  → gated: enabled? busy? user speech? cooldown? dedup?
  → onTrigger("order asian food") → submitted to assistant
```

## Verification

- `xcrun swift test --package-path Desktop --filter WakeWord` → **15 tests, 0 failures**.
- Exercised end-to-end in a running named bundle via the hermetic capture seam (drives the real `handleBackendSegments` → `WakeWordService` path):

```
WakeWord: submitting 'order asian food' to the assistant   ← wake phrase stripped
[...]

PR #11452, opened 2026-08-12
feat(dev-api): add /v1/dev/user/ask and omi ask CLI

aryanorastar:
## Addressed review (David CR 2026-08-12)

- [x] IANA timezone validated via `ZoneInfo` at the API boundary
- [x] `ConversationSearchUnavailableError` → 503 (parity with `/v1/conversations/search`)
- [x] Transcript-chunk retrieval merged with summary search before RAG

## Summary

Clean current-main replacement for #10314. Adds the read-only `/v1/dev/user/ask` Developer API endpoint and `omi ask` CLI command, grounded in cited conversation RAG.

## Security and cost controls

- Dedicated fail-closed `dev:ask` per-key budget (25/hour).
- Requires `conversations:read`.
- Excludes discarded conversations and re-checks locked records after authoritative hydration.
- Developer-API-key-only route; explicitly excluded from the Firebase app-client OpenAPI contract.

## Verification

- `backend/tests/unit/test_dev_ask_endpoint.py`: 9 passed locally on tip.
- Product line-count ratchet passed.
- CLI / OpenAPI export smoke may still rely on CI where local extras are missing.

## Honest gaps

- Full CLI `respx` suite and OpenAPI export not re-run locally in this environment; CI is the gate.

Failure-Class: none

Line-Count-Exception: backend/routers/developer.py | 2126 -> 2244 | Adds the read-only developer Ask endpoint with scoped retrieval, rate limiting, and locked/discarded conversation safeguards.

Supersedes stale conflicted #10314 with the same intended implementation on current `main`.

PR #11400, opened 2026-08-11
feat(app): find spoken moments in search + play on open

aryanorastar:
## Addressed review (David CR 2026-08-12)

- [x] Preserve server search rank in UI grouping (`groupSearchResultsPreservingRank` + tests)
- [x] Analyzer `avoid_print` baseline aligned (`30`)
- [x] Prior round: first timed snippet + sanitized transcript-index logs

## Summary

In-app conversation search now finds transcript-only spoken phrases when the optional transcript-chunk index is available, attaches bounded timed match snippets, and opens the transcript at the matching moment for playback.

The API keeps the search-only `ConversationSearchItem` contract, preserves locked-conversation redaction, and fails open when indexing or playback is unavailable.

## Verification

- Focused backend unit tests for transcript search, date/speaker filtering, generated-client contracts, and desktop REST inventory.
- Flutter search-seek + search-rank grouping tests passed locally (8/8 on tip).
- Dart formatting and Black checks passed.
- Product file line-count ratchet: declared exception below.
- Physical device/audio seek smoke was not run (no kit).

## Honest gaps

- Physical / on-device seek smoke not run (no kit).

Failure-Class: none

Line-Count-Exception: backend/routers/conversations.py | 1819 -> 1865 | Owns bounded transcript-match hydration and privacy redaction for spoken-moment search; page/date filters stay at the route so transcript-only hits cannot starve summary results (#11400).

Supersedes stale conflicted PR #11205; same verified final tree, rebased onto current `main` without merge.

PR #11804, opened 2026-08-18
feat(desktop): transcript-driven proactivity during ambient listening

aryanorastar:
## Summary

The desktop context-proactivity ("director") engine historically only evaluated a screen context on **dwell** — after you settle into a context through `contextEntered`, or right as you leave via `evaluateAfterDeparture`. Ambient speech was never a trigger: if you *asked something aloud* while a context visit was active, nothing happened until the next dwell.

This PR makes speech a first-class evaluation trigger. While a context visit is active, a fresh user utterance arriving through the ambient transcript funnel now evaluates **that** context immediately, grounded on the live screen **plus** a `== LIVE SPEECH ==` section quoted from what you just said — so hands-free requests ("where does the deploy script live?", "what's open on this page?") get answered during ambient listening, not on the next dwell.

This stacks on top of #11801 (wake word); together the pair closes the loop: the wake word opens the mouthpiece, speech triggers the evaluation, and the director delivers a grounded answer.

### Before
- Triggers were dwell-only: `contextEntered` and `evaluateAfterDeparture`.
- `evaluateAndDeliver` produced a prompt composed of the tracked frame + task list + recent-delivery dedup only.
- Ambient transcript segments built chat rows and nothing else.
- No permission/eligibility round-trip was reachable without a dwell.

### After
- New entry point `ContextProactivityEngine.evaluateFromSpeech(speech:)` — reuses the exact same authorized-delivery tail as dwell evaluations (approve → gate → fence → admission-queue → eligibility → frame grounding), so behavior is consistent with dwell, scaled to one-shot speech slices.
- `AppState+ListenEvents` feeds each new backend segment to `SpeechProactivityCoordinator.shared.observe(...)` (no-op when the feature flag is off).
- The engine prompt now appends a bounded `== LIVE SPEECH ==` section (up to 6 slices) behind the candidates section, so the director answers against both the screen and what you said.
- Speech is treated as **evidence, not instructions**: slices are single-line flattened under an untrusted-speech boundary so nothing you say can forge prompt structure elsewhere.

## Evaluation policy (admission)

A pure, unit-tested decision loop (`SpeechProactivityAdmission`) decides whether speech warrants an evaluation:

| Check | Value |
|---|---|
| Feature flag | `isTranscriptProactivityEnabled` (off in prod/beta; on in dev bundles; `OMI_FORCE_SPEECH_PROACTIVITY=0` force-disables) |
| Conversation state | Must **not** be mid assistant voice-turn (uses `VoiceTurnCoordinator.activeTurnID`) |
| Speaker | Latest slice must be **user** speech |
| Minimum length | ≥ 4 words (`minimumUserWordCount`) |
| Cooldown | ≥ 90s since last speech evaluation (`evaluationCooldownSeconds`) |

Skip reasons (`admission.outcome`) are observable: `.flagDisabled`, `.conversationActive`, `.noUserSpeech`, `.utteranceTooShort`, `.coolingDown`.

The speech window itself:
- `TranscriptSpeechSlice` is a small Sendable value (segment ID, speaker, text, `isUser`, start/end).
- `SpeechProactivityWindow` keeps max 12 slices, prunes by **seen time** (180s) — not backend timestamps (which can be in the future or missing) — and dedups by segment ID (transcripts are delivered repeatedly).

### Added
- `desktop/macos/Desktop/Sources/ProactiveAssistants/Core/TranscriptSpeechWindow.swift` — slice + window types; `SpeakerSegment.speechProactivitySlice` extraction.
- `desktop/macos/Desktop/Sources/ProactiveAssistants/Core/SpeechProactivityAdmission.swift` — pure admission policy.
- `desktop/macos/Desktop/Sources/ProactiveAssistants/Core/SpeechProactivityCoordinator.swift` — `@MainActor` observer that hops to the engine actor.
- `desktop/macos/Desktop/Tests/TranscriptDrivenProactivityTests.swift` — 14 hermetic tests.

### Modified
- `Core/ContextProactivityEngine.swift` — `evaluateFromSpeech(speech:)` entry; `evaluateAndDeliver` gained defaulted `speechSection:` appended after the candidates section. All existing dwell call sites compile unchanged.
- `Core/ContextBucketRollup.swift` — `ContextProactivityPromptBuilder.liveSpeechSection(_:maximumSliceCount:timeZone:)`; `[You]`/`[Other speaker N]` tags, `ContextDestinationKey.singleLine(limit:)` flattening, `== LIVE SPEECH ==` header.
- `Core/ContextBucketsFeature.swift` — `isTranscriptProactivityEnabled` + `OMI_FORCE_SPEECH_PROACTIVITY` override.
- `Core/ContextVisitCoordinator.swift` — production `activeFence()` accessor beside the existing `activeFenceForTesting()`.
- `AppState/AppState+ListenEvents.swift` — funnel wiring after the new segment is built.

### Notes / design choices
- Serialization: speech evaluations ride the same `dwellAdmission.begin(visitID:)` queue as dwell evaluations, so concurrent speech + dwell evaluations can't interleave on the same visit.
- Freshness: the same fence-freshness check applies; a speech evaluation won't fire on a stale fence.
- Conversation-active uses `VoiceTurnCoordinator.shared.activeTurnID != nil` (WakeWordService isn't on this base yet; #11801 introduces it and this PR is stacked on it, so no behavioral conflict).

## Testing / Verification
- `xcrun swift build -c debug --package-path Desktop` — clean.
- `TranscriptDrivenProactivityTests` — 14/14 pass (window dedup/retention/bounds, every admission outcome, prompt-section tags/flattening/slice cap).
- `ContextProactivity*` engine suites — all pass (dwell paths unaffected by the defaulted param).
- swift-format applied; swiftlint reports 0 violations.

Stack: #11801 → **this PR**.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11804?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

Line-Count-Exception: desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextProactivityEngine.swift | 1437 -> 1536 | evaluateFromSpeech is the third entry point into evaluateAndDeliver and shares the other two's guarded tail exactly - flag, authorization snapshot and recheck, free-gate preflight, active fence, dwell serialization, fence freshness, eligibility, frame grounding, then the injected jitHandle admission. A sibling file would duplicate that ordering rather than share it, and the ordering is the contract.


aryanorastar:
Both failing checks fixed and pushed (0cf23fc024):

1. **Repo Checks / PR Metadata Preflight + Hygiene — `desktop-changelog-entry`**: added the unreleased changelog fragment `20260818-speech-driven-proactivity.json`. Verified locally with `check-desktop-changelog.py` — passes.

2. **Desktop Swift Static & Test Contracts — `desktop-e2e-flow-coverage`** (same class as #11801): the three new speech files had no `covers:` entry. Added `SpeechProactivityAdmission.swift`, `SpeechProactivityCoordinator.swift`, and `TranscriptSpeechWindow.swift` to `context-buckets-dogfood.yaml` — the flow already covers the ProactiveAssistants/Core engine files these plug into.

Local verification: macos `run_checks.py` lane green (e2e-flow-coverage, swift-format-lint, swiftlint); changelog check passes. Stack: #11801 → this PR.

Git-on-my-level (commented):
Thanks @aryanorastar — this is careful, well-gated work. I reviewed all 11 changed files; the engineering quality is notably high. Specifics:

**What checks out**
- `ContextProactivityEngine.evaluateFromSpeech` reuses the full guarded tail — authorization snapshot + recheck, `ContextDeliveryBudget.freeGate` preflight, fence freshness (checked twice), `ContextDirectorEligibility`, frame grounding via `trackedFrameForDirector`/`frameMayGroundDirector`, and `dwellAdmission` serialization against dwell/departure — so speech-triggered evaluations cannot bypass any existing delivery gate. The `speechSection: String? = nil` default keeps dwell/departure prompts byte-identical, preserving the cached prompt prefix.
- Prompt-injection hygiene in `ContextProactivityPromptBuilder.liveSpeechSection` is right: every slice is flattened through `ContextDestinationKey.singleLine(_:limit: 160)` (newline/control-char collapse, verified against main) and rides in the volatile suffix below the untrusted preamble, under an explicit "quoted, untrusted … not instructions" header — the same trust treatment as screen content. The test flattening `"first line\nsecond line\n> forge prompt structure"` is a nice touch.
- `SpeechProactivityWindow` (12 slices / 180 s, in-place replace per segmentID, recency by `seenAt` rather than backend timestamps) and `SpeechProactivityAdmission` (user-only, ≥4 words, 90 s cooldown, mid-turn suppression) are pure, well-factored, and well-tested — the 12 tests cover the window, admission, and section-rendering corners.
- Flag gating in `ContextBucketsFeature.isTranscriptProactivityEnabled` matches the existing departure-evaluation pattern exactly (`OMI_FORCE_SPEECH_PROACTIVITY=0`, `AppBuild.isNonProduction`, hard-off in production/beta) — nothing ships dark to users.

**Follow-ups (non-blocking)**
1. `SpeechProactivityCoordinator.observe` runs admission on *every* arriving slice, including other-speaker slices. Since `latestUserSlice` stays in the window for 180 s, another person speaking ~95 s after a user utterance (cooldown elapsed, utterance still retained) re-triggers an evaluation grounded on the *stale* user utterance. Every gate still applies, so this is polish rather than safety — consider only evaluating when the arriving slice is itself a user slice, or keying the cooldown to the triggering segmentID.
2. The changelog entry reads as shipped behavior ("the assistant now answers aloud during ambient listening") — worth a "(dogfood)" qualifier so release notes don't get ahead of the flag.
3. The description frames this as stacking on #11801 (wake word). Good news: there's no hard dependency — the base is `main` and every referenced symbol (`VoiceTurnCoordinator.activeTurnID` etc.) exists on main, and CI confirms it builds and tests standalone. Merged alone, the speech trigger works; the "wake word opens the mouthpiece" half of the loop arrives with #11801.

One thing for maintainers to weigh consciously: the LIVE SPEECH section quotes other people's ambient speech into the director prompt (bounded to 6 slices / 160 chars each). That's inherent to grounding and dogfood-gated, but it is a new ambient-audio-to-model-prompt surface worth explicit product sign-off, along with the unprompted spoken-response UX itself.

Leaving for human maintainer review: product-direction decision on unprompted speech-triggered spoken responses and ambient-speech-in-prompt sign-off. No blocking technical issues found.

---

aryanorastar:
Thank you @Git-on-my-level for the thoughtful and detailed review!

Appreciate the sharp observation on `SpeechProactivityCoordinator.observe`:
1. **Arriving slice gating**: Gating evaluation specifically on `arrivingSlice.isUser` (or tracking the last-evaluated `segmentID`) is a great refinement so non-user turns don't re-trigger an evaluation on a retained user slice.
2. **Changelog qualifier**: Happy to add the `(dogfood)` qualifier to keep the release notes strictly aligned with the feature flag.
3. **Ambient Speech in Prompt**: Completely agree that keeping ambient speech bounded (6 slices, 160 chars, single-line flattened) under the untrusted section gives the director accurate grounding while protecting user safety behind the dogfood flag.

Ready for final maintainer sign-off and merge!

Git-on-my-level:
Status check on the updated head (`e1e8068`): that commit is a merge of `main` only — none of the merged files touch `desktop/`, so the speech-proactivity code is unchanged from the previously reviewed revision. Verified on this pass:

- Both refinements from the earlier review haven't landed yet: `SpeechProactivityCoordinator.observe` still runs admission on every arriving slice (an other-speaker slice arriving after the 90 s cooldown can re-trigger evaluation on a retained user utterance — every delivery gate still applies, so this stays non-blocking polish), and the changelog fragment still reads as shipped behavior without a `(dogfood)` qualifier. Worth folding both in before this leaves dogfood.
- All desktop checks pass on the merged head (build & tests, release compile, static & test contracts, e2e t0), and the merge didn't disturb anything this PR depends on.

The open item remains the product call rather than the engineering: unprompted speech-triggered spoken responses, and quoting bounded ambient speech into the director prompt, need maintainer sign-off.

---
*by AI on behalf of David — product-direction decision needed for unprompted speech-triggered spoken responses and ambient-speech-in-prompt; tag @Git-on-my-level to weigh in.*

aryanorastar:
@Git-on-my-level both refinements from your review are now on tip `b4ea2b6`, and the one red check is not this diff.

**Refinements landed**

`SpeechProactivityAdmission` now decides about the slice that just *arrived* rather than about whatever user slice the window still retains, so a non-user slice can never trigger. Other-speaker slices are still appended, so they keep grounding the LIVE SPEECH section — they just stop being triggers. I also keyed the decision to the backend `segmentID`: a segment is re-delivered as it grows, so the same utterance would otherwise evaluate again on every re-delivery once the cooldown lapsed. That adds a `.alreadyEvaluated` skip reason alongside the existing ones.

Two regression tests pin exactly the scenario you described — an other-speaker slice arriving 95s later with the user's utterance still retained, and the same segment re-arriving after the cooldown. Suite is 16 tests, up from 14.

I checked the new guard actually does the work rather than assuming it: dropping the `isUser` gate makes `testOtherSpeakerAfterCooldownDoesNotRetriggerOnAStaleUserUtterance` fail with `evaluate` instead of `skip(.noUserSpeech)`; restoring it passes. `swift build` clean, `check-desktop-changelog.py` passes.

Changelog fragment now reads "(dogfood builds only)" so release notes cannot get ahead of `isTranscriptProactivityEnabled`.

**The red `desktop-core-e2e-t0` is a bug on main, not here**

```
desktop-flow-lint: 2 error(s)
```

`d5596a6` relocated those two actions into `DesktopAutomationBridge+Notifications.swift` without adding that file to `ACTION_SOURCE_RELATIVE_PATHS`, which is the only list the lint reads. Reproducible against a pristine checkout of main with no branch involved:

```bash
git archive origin/main | tar -x -C /tmp/mainlint
cd /tmp/mainlint && python3 desktop/macos/scripts/desktop-flow-lint.py
```

This PR only surfaced it by being merged with current main. Filed as #11985. It reddens every desktop PR that merges main; it went unnoticed because main's own Desktop Backend Contracts runs are sitting at `action_required` rather than executing.

Resolved: main fixed this itself in `4d7c209e` (#11962), which added the missing entry alongside unrelated memory-TTL work. I have merged current main, so this branch now carries the fix and `desktop-flow-lint` passes locally — `OK (72 flows, 162 registered actions)`. The one-line PR I had opened for it (#11984) is closed as a no-op against current main.

<sub>Edited to correct this paragraph: it originally said the fix was pending in #11984 and that I was holding the merge back. Main landed its own fix in the meantime, so the merge is pushed and #11984 is closed.</sub>

Product call is unchanged and still yours: unprompted speech-triggered spoken responses, and quoting bounded ambient speech into the director prompt.

Git-on-my-level:
Follow-up on the current head (`4d337f4`): both refinements from the earlier review landed on `b4ea2b6`, verified — and the three merges from current main didn't disturb the speech path.

- `SpeechProactivityAdmission.decides` now decides about the slice that just *arrived* (`arrivingSlice.isUser`) rather than the window's retained `latestUserSlice`, so another speaker after the cooldown can no longer re-trigger an evaluation grounded on a stale user utterance — `testOtherSpeakerAfterCooldownDoesNotRetriggerOnAStaleUserUtterance` pins that exact scenario. Keying admission to the backend `segmentID` (`.alreadyEvaluated`) also stops a growing segment from re-evaluating on every re-delivery once the cooldown lapses — good catch on that re-delivery behavior; it's pinned by `testSameSegmentDoesNotEvaluateTwiceAfterTheCooldownLapses` as well.
- The changelog fragment now carries "(dogfood builds only)", aligned with `ContextBucketsFeature.isTranscriptProactivityEnabled`'s hard-off in production/beta.
- Merge check on `4d337f4`: `evaluateFromSpeech` still rides the full guarded tail (authorization snapshot + recheck, `ContextDeliveryBudget.freeGate` preflight, fence freshness, `ContextDirectorEligibility`, frame grounding, `dwellAdmission` serialization), `ContextProactivityPromptBuilder.liveSpeechSection` still lands additively behind the `speechSection: String? = nil` default so dwell/departure prompts are untouched, and the `AppState+ListenEvents` → `SpeechProactivityCoordinator.observe` hook still no-ops while the flag is off. Desktop checks are green on this head; the suite is 16 tests now.

No open technical concerns from my side. What remains is the product decision for a maintainer: unprompted speech-triggered spoken responses, and quoting bounded ambient speech into the director prompt, are a new UX/privacy surface even dogfood-gated — ready to merge once someone who owns that UX signs off.

---
*by AI on behalf of David — remaining before merge: product sign-off on unprompted speech-triggered spoken responses and ambient speech in the director prompt; tag @Git-on-my-level to weigh in.*

aryanorastar:
@Git-on-my-level thanks — one reframing of what the remaining sign-off actually commits to, because I think the current framing asks for more than the merge does.

Each of your passes has landed on some version of "a new ambient-audio-to-model-prompt surface needs product sign-off before merge". That reads as though merging turns the surface on. It cannot. The gate is bundle-derived, not remote:

```swift
@MainActor static var isTranscriptProactivityEnabled: Bool {
  guard isEnabled else { return false }
  if AppBuild.isNonProduction {
    return ProcessInfo.processInfo.environment["OMI_FORCE_SPEECH_PROACTIVITY"] != "0"
  }
  return false
}
```

Following that through `AppBuild`:

```swift
var isNonProduction: Bool {
  bundleIdentifier.hasPrefix("com.omi.")
    && !AppBuild.productionFamilyBundleIdentifiers.contains(bundleIdentifier)
}
```

`productionFamilyBundleIdentifiers` is exactly `{productionBundleIdentifier, betaProductionBundleIdentifier}` — the shipped stable app and the shipped Beta. Both are members, so both take the unconditional `return false`. There is no remote enable and no Info.plist key that reaches this; the only path to `true` is running a non-production bundle id, and even then the env var can force it off. `bundleIdentifier` also fails closed — `Bundle.main.bundleIdentifier ?? productionBundleIdentifier`.

So merging this ships no behavior change to any user. Nothing quotes ambient speech into a prompt on a shipped build, because the funnel no-ops before the window is ever appended to.

That splits the decision you are being asked to make into two, with very different weights:

1. **Merge this now** — commits to nothing user-visible, and stops the branch from re-merging `main` indefinitely. This is the one that is currently blocked.
2. **Sign off on the UX/privacy surface** — unprompted speech-triggered responses, and bounded ambient speech in the director prompt. That genuinely needs a human, and it is the gate on *removing the flag*, not on merging the code behind it.

Right now (2) is blocking (1), and I do not think it needs to. Deciding (2) well wants dogfood evidence — how often it fires, what it quotes, whether the 6-slice / 160-char bound holds up against real multi-speaker rooms — and that evidence cannot exist until the code is on `main` and running in dogfood builds. The current ordering asks for the judgement before the data that would inform it.

I am not asking anyone to skip (2). I am asking whether it should gate the merge or gate the flag flip. If the answer is the flag flip, this is mergeable today. If you would rather hold the whole thing until the UX call is made, that is a legitimate answer too and I will stop merging `main` into it and mark it draft — I would just rather that be a decision than a default.

One honest note on the asymmetry with my other stalled PR: I am **not** making this argument for #11801. That one genuinely does ship to users on merge, and its measured ~15–35 s ambient latency makes it weak as command-and-control, so the product call there really is a merge blocker. The difference is exactly this flag.

Nothing else is outstanding here — both refinements from your earlier passes landed on `b4ea2b6`, checks are green on `4d337f4`, and the suite is 16 tests.

aryanorastar:
A/B, identical script: 1 delivery transcripts-off, 1 transcripts-on. No increase.

Video: 10.5 min real Hotstar + room talk, 213 segments → 1 notification, and it was correct — flagged a real open task while I was on leisure content. 17 cooldown suppressions behind it.

Budget: a transcript evaluation only consumes allocation if it delivers; the counter excludes `suppressed`/`failed`.

Caveat: single session, and the injected half tests gating not STT.

Git-on-my-level:
Follow-up on the current head (`774ae9d`), covering the two commits since `4d337f4` plus the merge.

**Lane-dimension telemetry (`2bf4aa2f`) — verified, one small gap**

`ContextProactivityTelemetry.GateTrigger` now records which lane started a rejected evaluation, derived once in `evaluateAndDeliver` via `forSpeechSection(speechSection)` — the speech lane is the only entry point that supplies a section, so the argument carries the distinction instead of each gate site repeating it. I checked all nine `recordGateRejection` call sites in `ContextProactivityEngine.swift`: the speech preflight passes `trigger: .speech`, and the shared-tail sites (attempt / reservation / preModel / presentation / handoff) all thread `gateTrigger` through. The three new tests in `ContextProactivityEngineTests.swift` pin the lane split, the empty-section-still-speech case, and the raw values — good, since renaming a value would silently break the analytics split.

One nit: the `.retrievalHop` site (`ContextProactivityEngine.swift:1208`) still uses the default `trigger:`. `performRetrievalHop` is called from inside `evaluateAndDeliver` (line 647), which the speech lane reaches, so a speech-grounded evaluation rejected during the retrieval hop would be counted as `screen_visit` — exactly the crowding-out miscount this dimension exists to prevent. A one-word fix whenever the file is next touched; not blocking.

**Merge check (`774ae9d`)**

The final merge brought roughly 150 desktop files in from main. Nothing PR-owned changed: `liveSpeechSection` in `ContextBucketRollup.swift` is byte-identical (main's edits there are director prompt-policy refinements elsewhere in the file), `ContextBucketsFeature.isTranscriptProactivityEnabled` is unchanged, and `evaluateFromSpeech` still runs the full guarded tail — flag, authorization snapshot + recheck, free-gate preflight, active fence, `dwellAdmission` serialization, fence freshness twice, eligibility, frame grounding — before `evaluateAndDeliver(speechSection:)`. Desktop checks are green on the head.

**On the merge-vs-flag-flip framing**

I verified the gate chain you quoted: `AppBuild.productionFamilyBundleIdentifiers` covers both the stable and Beta bundle ids, `bundleIdentifier` fails closed to production, and there is no remote or Info.plist path that reaches `isTranscriptProactivityEnabled` — merging this ships no behavior change on any shipped build. The argument is technically sound, and the dogfood evidence (213 segments → 1 correct delivery, 17 cooldown suppressions, budget consumed only on delivery) is the right kind of data, single session notwithstanding.

Whether that makes the UX/privacy call a gate on the merge or a gate on removing the flag is a product-ordering decision I'm leaving explicitly with David rather than deciding here. No blocking technical issues on this head.

---
*by AI on behalf of David — decision needed on merge-vs-hold ordering for this dogfood-hard-off feature, alongside the standing UX/privacy sign-off for unprompted speech-triggered responses and ambient speech in the director prompt; @Git-on-my-level.*

aryanorastar:
@Git-on-my-level the `.retrievalHop` nit is fixed on `b401c1c2`.

`performRetrievalHop` now takes `gateTrigger` and passes it to the `.retrievalHop` rejection, so a speech-grounded evaluation gated while the hop rebuilds the free gate is counted as `speech` rather than `screen_visit`. I made the parameter non-defaulted rather than defaulting it to `.screenVisit` — there is exactly one call site today, and without a default the compiler is what enforces that a future second call site declares its lane, instead of it silently inheriting the wrong one. That is the failure mode this dimension exists to catch, so it seemed worth spending the argument.

No test added: this is an argument threaded through a private method with a single caller, and the engine has no end-to-end retrieval-hop harness — building one to pin a pass-through would be more machinery than the change. The compiler covers it, and the three existing lane tests still pin `forSpeechSection` and the raw values.

`swift build` clean, `swift-format lint --strict` clean, 39/39 in `ContextProactivityEngineTests` + `TranscriptDrivenProactivityTests`. All checks green on the head.

Nothing else outstanding from any of your passes. The open item is still the product one, and I take your point that the merge-vs-hold ordering is David's call rather than yours to make here.

Git-on-my-level:
Follow-up on the current head (`b401c1c2`) — the `.retrievalHop` attribution nit from my last pass is fixed, verified on the head blob.

`performRetrievalHop` now takes `gateTrigger` as a non-defaulted parameter and passes it to the `.retrievalHop` rejection (`ContextProactivityEngine.swift:1210`), so a speech-grounded evaluation gated while the hop rebuilds the free gate is counted as `speech` instead of `screen_visit`. Making the parameter non-defaulted is the right call for exactly the reason you gave: with one call site, the compiler now forces any future second caller to declare its lane rather than silently inheriting `.screenVisit`. I re-audited all 11 `recordGateRejection` sites on the head — the speech preflight passes `trigger: .speech`, every shared-tail site (attempt / reservation / preModel / presentation / handoff, on both the main and candidate paths) threads `gateTrigger` through, and the single remaining defaulted call is the `contextEntered` dwell preflight, where `.screenVisit` is the correct lane. Skipping a dedicated test for a compiler-enforced argument pass-through with one caller is reasonable; the three lane tests still pin `forSpeechSection` and the raw analytics values.

With that, every technical item raised across the passes on this PR is resolved: admission decides on the arriving slice and dedupes by segment ID, the changelog fragment carries the dogfood qualifier, the speech path rides the full guarded delivery tail, and the `== LIVE SPEECH ==` section stays bounded, single-line flattened, and quoted as untrusted data under the preamble. Desktop checks are green on the head. Thanks for the persistence through the `main` re-merges — the lane-dimension telemetry and the dogfood A/B evidence materially raised the quality of the product conversation on this thread.

The one open item is the product decision the thread has converged on: whether UX/privacy sign-off for unprompted speech-triggered responses (and ambient speech quoted into the director prompt) gates this merge or gates removing the dogfood flag. That ordering call stays with a human maintainer.

---
*by AI on behalf of David — remaining before merge: product decision on whether UX/privacy sign-off gates the merge or the flag removal, for the speech-triggered response surface; @Git-on-my-level.*

aryanorastar:
@Git-on-my-level thanks for the re-audit of all 11 `recordGateRejection` sites — confirming the one remaining defaulted call is the `contextEntered` dwell preflight is a stronger check than the one I did, and it is the right lane there.

With that pass, this PR is in an unusual state that I think is worth naming plainly, because it looks like a review gap rather than a decision:

- Every technical item raised across four passes is resolved, in your words.
- All checks are green on `b401c1c2` — 18 successful, 9 skipped, 0 failing, 0 pending.
- There is no `CHANGES_REQUESTED` on file.
- The merge is still blocked on "at least 1 approving review by reviewers with write access."

The gap is mechanical: your passes on this thread have been issue comments rather than submitted reviews. The only formal review on this PR is a single `COMMENTED` from Aug 18 on `0cf23fc0`, so none of the four subsequent passes — including the one where you wrote "no blocking technical issues found" — count toward that gate.

@kodjima33 — this one has never reached you. You have given approve-only passes on #11206, #11400, #11452 and #11807 under the same feature policy; #11804 is the sibling of #11807 and is the only one of the set with nothing outstanding against it. I cannot add you as a reviewer myself (the API returns 404 for a fork contributor), so this comment is the only route I have.

On the product question, I am not re-arguing it — I made that case above and @Git-on-my-level has already verified the gate chain independently. I would just note the two are separable: the approving review is about whether the code is correct, which the passes above have answered; the UX/privacy sign-off is about whether the flag ever flips, which stays open either way and does not need to be settled to approve correct code that is hard-off on every shipped bundle.

If the intent is to hold this until the UX call is made, that is a fine answer and I will mark it draft and stop merging main into it. I would rather have that than have it sit on a gate nobody realised was unsatisfied.

aryanorastar:
@Git-on-my-level ## Current-main refresh + verification proof

Refreshed this branch onto `origin/main` at `cd6765ba9c` and pushed head `017b2f6292`.

### Scope integrity

- Merge completed without conflicts.
- The PR diff remains exactly **13 macOS files**: 9 production Swift files, 2 Swift test files, 1 changelog fragment, and 1 macOS e2e-flow update.
- Diff remains **607 additions / 18 deletions**.
- No iOS, Android, backend, or Windows changes were pulled into this PR.
- `git diff --check origin/main...HEAD` passes.

### Behavior verification on current main

```text
xcrun swift test -c debug --package-path Desktop \
  --filter 'ContextProactivityEngineTests|TranscriptDrivenProactivityTests'

ContextProactivityEngineTests:      23 passed
TranscriptDrivenProactivityTests:   16 passed
Total:                              39 passed, 0 failed
```

This covers the existing dwell path plus transcript-window dedup/bounds, admission decisions, arriving-slice semantics, cooldown, prompt grounding, stale-fence rejection, lane attribution, and delivery behavior.

### Static and repository gates

- PR-scoped pinned `swift-format --strict`: **pass**
- Pinned SwiftLint: **0 violations across 1,429 files**
- Desktop test-quality ratchet: **pass**
- Desktop flow contract: **74 flows / 181 actions, pass**
- Desktop changelog requirement: **pass**
- Diff hygiene, author identity, line-count ratchet, failure-class protocol, e2e coverage (9/9 sources covered), GRDB idiom, brand invariant, and deferred-work checks: **pass**

### Known base-branch preflight issue

The aggregate preflight currently stops on a pre-existing `origin/main` legacy-memory ratchet mismatch in:

```text
desktop/windows/src/main/assistants/insight/prompt.ts: 4 -> 5
```

That file is not in this PR. The isolated main-wide repair is [#12339](https://github.com/BasedHardware/omi/pull/12339); its hygiene and backend unit gates are green. I deliberately did not mix that unrelated repair into this feature branch.

### Rollout safety

The feature remains hard-off in production and Beta. Only non-production/dogfood bundles enable it, with `OMI_FORCE_SPEECH_PROACTIVITY=0` available as an explicit kill switch. This refresh does not change that rollout boundary.

Git-on-my-level:
Refresh check on `017b2f6292` (merge of main at `cd6765ba9c`) — verified from the head blobs, not the description.

**PR-owned code is unchanged.** All 13 files re-checked against the last-reviewed head (`b401c1c2`): `SpeechProactivityAdmission.swift` (arriving-slice gating + `segmentID` dedup), `SpeechProactivityCoordinator.swift` (MainActor observe loop), `TranscriptSpeechWindow.swift` (12-slice / 180 s bounded window), `ContextBucketRollup.swift`'s `liveSpeechSection` (≤6 slices, single-line flattened, quoted as untrusted under the preamble), `ContextBucketsFeature.isTranscriptProactivityEnabled` (hard-off in production/beta), `ContextVisitCoordinator.activeFence()`, the changelog fragment's dogfood qualifier, the `covers:` entries, and both test files (39 tests) are blob-identical. The four files that did change (`AppState+ListenEvents.swift`, `ContextProactivityEngine.swift`, `ProactiveLaneClient.swift`, `context-buckets-dogfood.yaml`) changed only through main's side of the merge — `ProactiveListenAdmission`/`proactive_message` routing and the `hasTranscriptionBYOK` rename in the first, the JIT hooks and JIT client methods in the others. `evaluateFromSpeech` still rides the full guarded tail (flag, authorization snapshot + recheck, free-gate preflight with `trigger: .speech`, active fence, `dwellAdmission` serialization, fence freshness twice, eligibility, frame grounding).

**The red Hygiene check is not this diff.** From the job log: `legacy-memory-surface-ratchet` fails on `desktop/windows/src/main/assistants/insight/prompt.ts: 4 -> 5`, and that file is byte-identical (blob `51782366bf5`) between main `cd6765ba9c` and this head. The growth landed on main in `61dcb9473c` (Aug 27); the isolated repair is #12339. Nothing for this PR to do beyond waiting on that fix.

**One new integration point worth a look before dogfood widens.** Main's JIT work now fronts the dwell and departure lanes: `contextEntered` and `evaluateAfterDeparture` call `JITProactivityCoordinator.handle(...)` first (engine lines 220 / 278), and only a `.legacyContextBucketFallback` decision falls through to `evaluateAndDeliver`. `evaluateFromSpeech` calls `evaluateAndDeliver` directly, so the speech lane never consults the JIT admission stack. With an owner moved onto the JIT rollout (rollout enabled, kill switch disabled), dwell/departure deliveries come from the JIT lane — or are JIT-suppressed — while a spoken mid-visit question still evaluates through the legacy director tail. Practically: the JIT kill switch does not stop speech-lane deliveries in dogfood builds, and the `GateTrigger` telemetry split (`screen_visit` vs `speech`) predates this third admission path. Blast radius today is dogfood-only (the transcript flag is hard-off in production/beta, and JIT default-unknown fails closed to legacy), so this reads as a sequencing question to settle before the dogfood flag flips or JIT rollout widens — either route speech through the same admission or note that the lanes are intentionally independent.

Desktop checks are green on this head (build & tests, release compile, static & test contracts, e2e t0). The open item is unchanged from the previous passes: whether UX/privacy sign-off for unprompted speech-triggered responses (and ambient speech in the director prompt) gates this merge or gates removing the dogfood flag — that stays with a human maintainer.

---

aryanorastar:
@Git-on-my-level thanks for the blob-level refresh check, and for catching the JIT integration point — the speech lane genuinely does not consult `JITProactivityCoordinator`, and that is worth settling before either flag widens.

One correction to the framing, because I think it points at the wrong fix. I traced the kill-switch path rather than reasoning from the call sites:

```swift
// JITProactivityPolicy.swift:29
rollout == .enabled && killSwitch == .disabled   // permitsNewLane

// :89
guard flags.permitsNewLane else {
  case (_, .enabled): reason = "kill_switch"
  return .legacyContextBucketFallback(reason: reason)
}
```

`.legacyContextBucketFallback` makes `handle()` return `false`, and `contextEntered` then falls through to `evaluateAndDeliver`. So with the kill switch engaged, **dwell delivers through the legacy director too** — it is not suppressed. Both lanes end up in the same place, and the speech lane isn't escaping a control that dwell obeys.

The asymmetry is the other case, and it is real: when JIT is *enabled* and returns `.suppressed`, `handle()` returns `true` and dwell returns without delivering, while a spoken mid-visit question still goes down `evaluateAndDeliver`. So it is JIT's **suppression decisions** the speech lane bypasses — dedup, continuity keys, planned-trigger precedence — not the kill switch.

That difference matters for what to do about it. "The kill switch is leaky" would argue for an urgent fix; "the two lanes have independent admission" is a design question, and the answer depends on whether a spoken question should be subject to JIT's planned-trigger precedence at all. My instinct is that it should not — a user asking out loud is a direct request, where dwell is the engine volunteering — but that is exactly the kind of call I should not make unilaterally on a PR already waiting on a product decision.

So I have not changed behaviour. If you want speech routed through the same admission, say so and I will do it with a test; if the lanes are intentionally independent, that is worth a line in the code so the next reader does not have to derive it. Either way it wants deciding before the dogfood flag flips, as you said.

Everything else is unchanged: 13 files, green on `9ee607f4` after the main refresh that cleared the `legacy-memory-surface-ratchet` red (that was #12339's territory and has since landed). @undivisible

Git-on-my-level:
Follow-up on `9ee607f4` — and first, @aryanorastar: your correction is right. I verified it against the code rather than taking it on faith.

**The kill-switch framing in my last pass was wrong.** Tracing the same path you did: `JITProactivityPolicy.decide` returns `.legacyContextBucketFallback` whenever `permitsNewLane` is false — including the kill-switch case (`JITProactivityPolicy.swift:87–95`) — and `JITProactivityCoordinator.handle` returns `false` for fallback, so `contextEntered` (engine line 220) falls through to `evaluateAndDeliver`. With the kill switch engaged, dwell delivers through the legacy director exactly like speech does. The kill switch is not a control the speech lane escapes, and my earlier sentence read as though it were.

**What actually differs** is the case you isolated: when JIT is *enabled* and returns `.suppressed` — dedup, continuity keys, planned-trigger precedence — `handle()` returns `true` and dwell/departure end without delivering, while a spoken mid-visit question still evaluates through the legacy tail (`evaluateFromSpeech` calls `evaluateAndDeliver` directly). So the open question is precisely yours: should a spoken question be subject to JIT's planned-trigger precedence, or are the lanes intentionally independent because a direct spoken request outranks an engine-volunteered nudge? That is a design decision, and I agree with your instinct not to make it unilaterally — it needs a maintainer ruling (route speech through the same admission, or document the independence where the next reader will find it) before the transcript flag widens beyond dogfood or the JIT rollout grows.

**On this head:** all 13 PR-owned files are blob-identical to the last-reviewed `017b2f6292` — the only change is the merge of main, which also cleared the `legacy-memory-surface-ratchet` hygiene red (#12339's territory, landed on main). Desktop checks are green on `9ee607f4` (build & tests, release compile, static & test contracts, e2e t0).

Nothing blocking in the code as it ships behind the dogfood flag. What remains are decisions, not defects, and they stay with a human maintainer: the JIT/speech admission question above, and UX/privacy sign-off for unprompted speech-triggered responses (ambient speech quoted into the director prompt — bounded to 6 slices at 160 chars, single-line flattened, framed as untrusted data, hard-off in production/beta).

---

aryanorastar:
Filed the JIT/speech admission question as #12407 so it outlives this thread — that's the "document the independence where the next reader will find it" half of your two options, without picking the answer.

It records the verified asymmetry (kill switch sends all three lanes to legacy alike; `.suppressed` is the only case where dwell stops and speech doesn't), both readings of whether that's right, and the fact that nothing is user-visible while the transcript flag is hard-off and JIT fails closed.

I deliberately didn't add a code comment. The PR's 13 files are blob-stable across your last two passes, and if the ruling is "route speech through the same admission" the comment would just come straight back out. Once someone decides, whichever half is needed is a small push — the code comment pointing at #12407, or the routing with a test.

Nothing else outstanding here. @Git-on-my-level @undivisible

Git-on-my-level:
Merge check on `4c1ba5ee` (merge of main at `4fe6b5c2`). This is the second merge on this branch that needed conflict resolution in `ContextProactivityEngine.swift`, so I verified the resolution against both sides rather than assuming it.

**What the resolution preserved, from the head blobs:**
- Main's JIT integration is intact: `JITProactivityCoordinator.shared.handle` still gates both dwell entry points (`contextEntered` at engine line 222, `evaluateAfterDeparture` at line 280).
- The speech path is equally intact: `evaluateFromSpeech` (line 301) still rides the full guarded tail - flag, authorization snapshot + recheck, free-gate preflight with `trigger: .speech`, active-fence check, `dwellAdmission` serialization, fence freshness (twice), eligibility, frame grounding - and the effective delta vs current main is exactly the PR-owned hunks: the `speechSection: String? = nil` default (dwell/departure prompts stay byte-identical), `gateTrigger` threaded through the ten speech-reachable rejection sites on both the main and candidate paths (the one remaining default, the `contextEntered` dwell preflight, is correctly `.screenVisit`), and the `== LIVE SPEECH ==` section appended after the candidates section in the volatile prompt.
- The other three merge-touched files (`AppState+ListenEvents.swift`, `ContextBucketRollup.swift`, `ProactiveLaneClient.swift`) match main plus only their PR hunks; the remaining nine PR files are blob-identical to the last-reviewed revision.

One thing the merge makes more concrete: the asymmetry filed as #12407 is now visible on this head - the engine consults `JITProactivityCoordinator` from both dwell entry points, while `evaluateFromSpeech` still calls `evaluateAndDeliver` directly. Nothing is user-visible while the transcript flag is hard-off in production and beta, but the ruling (route speech through the same admission, or document the lanes as intentionally independent) is now the only open design question besides the product sign-off itself.

Nothing blocking in the code. What remains is unchanged in substance from the last pass: a product decision on unprompted speech-triggered spoken responses with bounded ambient speech quoted into the director prompt, and the #12407 admission ruling.

---
*by AI on behalf of David - the open items are product rulings rather than code fixes: speech-triggered-response sign-off and the #12407 JIT-admission decision; no urgency while the feature stays dogfood-gated.*

aryanorastar:
Speech now routes through JIT admission, on `b81fa9805c`. That closes the #12407 ruling, which was the last open item that wasn't the product call.

`evaluateFromSpeech` calls `JITProactivityCoordinator.shared.handle(...)` in the same position `contextEntered` and `evaluateAfterDeparture` use it, returning early when JIT handles the visit. Routed rather than documented as independent, per the ruling recorded on #12407.

**One correction worth stating plainly.** The first draft of this change carried a rationale — in the code comment and the test doc — saying the kill switch suppressed dwell and departure while speech delivered anyway. That is the claim raised in your earlier pass, traced on this thread, retracted, and then verified by you independently. It should not have come back in the commit that fixes the thing. Each version even contradicted itself, stating the `false` return one clause after asserting the switch suppressed the other two.

Both now say what is actually true: `permitsNewLane` false — kill switch included — yields `.legacyContextBucketFallback`, `handle()` answers `false`, and all three lanes fall through to the legacy director alike. What speech escaped was `.suppressed`: JIT enabled and declining on dedup, continuity keys, or planned-trigger precedence, where dwell and departure stop and speech did not. That asymmetry is the reason routing was the right answer.

**On the test.** `testEveryProactivityEntryPointPassesJITAdmission` counts `JITProactivityCoordinator.shared.handle(` occurrences against the three entry points, so a fourth entry point must either route through the coordinator or change the number deliberately. It is a source scrape, not behavioural coverage — the engine has no seam to drive all three entry points against a fake coordinator — and the doc comment now says so rather than implying otherwise.

`desktop-test-quality` caught it as expected, taking production-source inspection from 147 to 148 sites. I used the gate's own remedy — a reasoned `omi-test-quality: source-inspection` annotation on the counted line — rather than raising the baseline, since this is a genuine static contract. Debt back at baseline: 54 files / 147 sites.

Verification: `swift build` clean; `ContextProactivityEngineTests` + `TranscriptDrivenProactivityTests` 40/40. Mutation-checked — deleting the speech routing fails the admission count with `("2") is not equal to ("3")`.

Remaining is the product decision alone: unprompted speech-triggered responses and bounded ambient speech in the director prompt, and whether that sign-off gates this merge or gates removing the dogfood flag.

Git-on-my-level:
Follow-up on the new head (`b81fa980`) — the three commits since `4c1ba5e` are the #12407 resolution, so this pass focuses on that delta plus a head-level re-verification of the speech path. #12407 is confirmed closed with the routing ruling recorded there.

**Speech now routes through JIT admission — verified.** `evaluateFromSpeech` calls `JITProactivityCoordinator.shared.handle(...)` at the same position `contextEntered` (engine L222) and `evaluateAfterDeparture` (L280) use it — after eligibility and frame grounding, immediately before `evaluateAndDeliver` (speech at L359) — returning early when the coordinator handles the visit. Tracing `handle()` in `JITProactivityCoordinator.swift`: it returns `false` only for `.legacyContextBucketFallback` (so kill-switch/JIT-off falls through to the legacy director on all three lanes alike) and `true` for `.suppressed` and `.deliver` (where JIT delivers itself). The `.suppressed` asymmetry #12407 isolated — speech escaping JIT's dedup, continuity-key, and planned-trigger-precedence decisions — is closed, and the code comment and test rationale now match the corrected kill-switch semantics rather than the retracted framing.

**The static-contract test** (`testEveryProactivityEntryPointPassesJITAdmission` in `ContextProactivityEngineTests.swift`) counts `JITProactivityCoordinator.shared.handle(` occurrences against the three entry-point declarations via a source scrape. Given the engine has no seam to drive all three entry points against a fake coordinator, that's an honest substitute, and counting rather than pinning signatures is robust to reformatting. Its known blind spot: it verifies presence, not position — a `handle()` call placed after `evaluateAndDeliver` would still pass. The ordering is correct on the head, so this is fine as a ratchet.

One design nuance the routing surfaces, worth a line on #12407 or a follow-up issue rather than a change here: `handle()` takes no speech argument, so when JIT *delivers* (`.deliver`), the live-speech evidence never reaches the JIT prompt — only the legacy-director path sees the `== LIVE SPEECH ==` section. With the transcript flag dogfood-only and JIT failing closed to legacy by default there's no shipped impact, but if both lanes ever widen, a spoken question answered by the JIT path would be answered without the speech that prompted it.

**Unchanged and re-verified on the head:** the guarded tail ordering in `evaluateFromSpeech` (flag → authorization snapshot → free-gate preflight with `trigger: .speech` → active fence → `dwellAdmission` serialization → fence freshness → eligibility → frame grounding → JIT admission), `liveSpeechSection`'s untrusted-data framing with `ContextDestinationKey.singleLine(_:limit: 160)` flattening and the ≤6-slice cap, `SpeechProactivityWindow` bounds (12 slices / 180 s, in-place per-segmentID replace), admission on the arriving user slice with segmentID dedup and the 90 s cooldown, `isTranscriptProactivityEnabled` hard-off outside `AppBuild.isNonProduction`, lane-attributed gate telemetry including the `.retrievalHop` site, the changelog's dogfood qualifier, and the e2e `covers:` entries. Checks green on the head.

That closes every technical item raised across the passes on this PR. What remains is the product call, unchanged: unprompted speech-triggered spoken responses, with bounded ambient speech quoted into the director prompt, need maintainer sign-off before this leaves dogfood — and this PR still sits on #11801, which is open, so merge order is part of that decision.

---
*by AI on behalf of David — product sign-off needed for speech-triggered spoken responses and ambient-speech-in-prompt, plus merge order relative to open #11801; tag @Git-on-my-level to weigh in.*

aryanorastar:
Filed the follow-up you asked for: #12852.

I verified the nuance against `b81fa980` rather than taking it on the description, and it holds exactly as stated. In `evaluateFromSpeech`, `handle(fence:snapshot:frame:authorizationSnapshot:)` at L359 takes no speech argument; `liveSpeechSection` is built at L365, *after* the early return at L363; and in `JITProactivityCoordinator.handle`, `.deliver` returns `true`. So on the delivering JIT path the engine returns before the speech has been passed anywhere. The same call shape at L222 and L280 is correct — those lanes have no speech to give — so the speech lane is the only one that loses evidence.

I recorded why it is not urgent as well as why it matters: the transcript flag is hard-off outside `AppBuild.isNonProduction` and JIT fails closed to the legacy director, which *does* get the speech section, so the gap is only reachable if both lanes widen. The issue also notes that a fix should reuse `liveSpeechSection`'s existing bounded framing — untrusted-data framing, `singleLine(_:limit: 160)` flattening, ≤6-slice cap — rather than hand the JIT path raw transcript text on a looser contract than the legacy path holds itself to.

No code change here, so nothing on this branch moved. Checks remain green on the head (14 pass, 13 skipped), and the PR is `MERGEABLE`.

Agreed on what is left, and it is not mine to settle: the product call on unprompted speech-triggered spoken responses with bounded ambient speech quoted into the director prompt, and the merge order relative to #11801, which is still open. This PR sits on that one.

aryanorastar:
Pushed 71baa2ec8e — merged current main (1247 commits), and that merge turned one of this PR's own decisions stale, so it is fixed here rather than left to rot.

**Main added the seam this PR's test comment said did not exist.** JIT admission now sits behind an injected `jitHandle` closure on the engine's `init`, and both dwell entry points share a single call through it — the only remaining mention of `JITProactivityCoordinator.shared.handle(` is that closure's default argument.

`evaluateFromSpeech` still called the singleton directly. Not wrong at runtime, since the default forwards to the same coordinator — but it made the one lane this PR adds the one lane a test cannot point at a fake, which is exactly the limitation `testEveryProactivityEntryPointPassesJITAdmission` cites as its reason for being a source scrape. Speech now goes through `jitHandle` like the other two.

**The contract test's premise died with the same refactor**, and it would have failed on this merge: it asserted one admission call per entry point, and the dwell lanes now share one. Counting was never the invariant — the route is. It now asserts:

- the singleton appears **exactly once**, the seam's default, so no lane reaches the coordinator around it;
- `evaluateFromSpeech` consults `jitHandle` **before** `evaluateAndDeliver`.

That second assertion also closes the blind spot the earlier pass named — *"it verifies presence, not position — a `handle()` call placed after `evaluateAndDeliver` would still pass."* For the speech lane it no longer would.

**One thing worth flagging about the merge itself.** Counting admission calls across the three trees reads alarming at first: the branch had 3, main has 1, the merged result has 2. That is correct, not a dropped hunk — main consolidated its two dwell calls into one shared call, so 1 + this PR's 1 is the right answer. I checked it against all three trees rather than trusting the arithmetic, because a silently dropped call here is exactly the failure this PR's test exists to catch.

**Verification:** 27 tests pass across `ContextProactivityEngineTests` and the speech suites, including the rewritten contract test. Scope unchanged at **13 files, +664/−16**, `desktop/macos` only.

**One new Line-Count-Exception**, added because main crossed the threshold under this branch rather than because this PR grew: `ContextProactivityEngine.swift` is 1437 on main and 1536 here. The reason is recorded in the PR body — the speech entry point shares the other two's guarded tail exactly, so a sibling file would duplicate that ordering rather than share it.

The open items are unchanged and are not code: the product call on unprompted speech-triggered responses with bounded ambient speech in the director prompt, and merge order relative to #11801. #12852 still tracks the JIT-delivers-without-speech-evidence nuance.

Git-on-my-level:
Merge check on the new head (`71baa2ec`) — the two commits since `b81fa980` are the main refresh (1247 commits) plus one PR-owned change, so this pass focuses on that delta.

**The substantive part of this head: main moved the engine onto an injected JIT seam, and the merge turned this PR's own test stale — the fix here is correct.** Verified against the head blobs, not the description:

- `ContextProactivityEngine` now stores `jitHandle: JITHandle` as an injected property whose *default argument* routes to `JITProactivityCoordinator.shared.handle` — the single surviving singleton reference in the file (the reshaped test pins exactly that count, which is now the right invariant: a lane calling the singleton directly is one a test can't drive against a fake). Both the dwell lane and the speech lane reach the coordinator through the seam, and in `evaluateFromSpeech` the `await jitHandle(...)` call still precedes `await evaluateAndDeliver(...)`, so the #12407 fix survived the merge — now through the injected route rather than the singleton.
- The reworked `testEveryProactivityEntryPointPassesJITAdmission` in `ContextProactivityEngineTests.swift` is an honest reshape, not a weakening: the old per-entry-point call count was the wrong shape once main consolidated the dwell lanes onto the shared seam, and the replacement asserts presence of all three entry points, exactly one singleton reference (the seam's default), and admission-before-delivery ordering pinned within the speech body. The accompanying comment on why `.suppressed` — not the kill switch — was the case speech used to escape remains accurate against `JITProactivityPolicy`.
- `evaluateAndDeliver` now derives the telemetry trigger in one place via `GateTrigger.forSpeechSection(speechSection)`, and the speech section rides last in the volatile prompt after the candidates section — the cached stable prefix and main's volatile ordering are both preserved. All `recordGateRejection` sites, including `performRetrievalHop`, carry the trigger; the payload stays bounded enums.

**Re-verified unchanged on the head:** the guarded tail in `evaluateFromSpeech` (flag → authorization snapshot → free-gate preflight with `trigger: .speech` → active fence → `dwellAdmission` serialization → fence freshness → eligibility → frame grounding → JIT admission), `liveSpeechSection`'s quoted-untrusted-data framing with `ContextDestinationKey.singleLine` flattening and the ≤6-slice cap, `SpeechProactivityWindow` bounds (12 slices / 180 s, per-segmentID in-place replace), admission on the *arriving* user slice with segment dedup and the 90 s cooldown, `isTranscriptProactivityEnabled` hard-off outside `AppBuild.isNonProduction` with the `OMI_FORCE_SPEECH_PROACTIVITY=0` kill, the dogfood-only changelog wording, and the e2e `covers:` entries for the three new core files. The follow-up nuance from the last pass remains tracked in #12852. Checks green on the head, including Desktop Swift Build & Tests and the static/test contracts.

No new technical items on this head. What remains is the product call, unchanged: unprompted speech-triggered spoken responses, with bounded ambient speech quoted into the director prompt, need maintainer sign-off before this leaves dogfood — and this PR still sits on #11801, which is open, so merge order is part of that decision.

---
*by AI on behalf of David — product sign-off needed for speech-triggered spoken responses and ambient-speech-in-prompt, plus merge order relative to open #11801; tag @Git-on-my-level to weigh in.*

PR #11806, opened 2026-08-18
Windows chat markdown: render tables, and stop destroying links inside bold

ZachL111:
Windows renders chat markdown with a 173-line file that splits the text on one alternation regex. macOS uses 1,635 lines (`OmiMarkdown.swift`). The gap is not only missing features: two of the differences are wrong output.

## What the shipped renderer does today

Each of these was run against the current build and the result recorded before anything was changed.

| Input | Renders as |
|---|---|
| a four-line GFM table | one paragraph of literal `\| a \| b \|` |
| `> quoted line` | literal `&gt; quoted line` |
| `- a` / `  - b` / `- c` | one FLAT list; the nesting is discarded |
| `- [ ] todo` | literal `[ ] todo` |
| `~~gone~~` | literal tildes |
| `Before` / `---` / `After` | one paragraph with the dashes inside it |
| `#` and `######` | identical bold paragraphs |
| `**see [docs](https://x.com) now**` | **the link is destroyed**: bold text containing the raw markdown characters |
| `a \*not italic\* b` | **mangled**: the backslashes show AND it italicises |

The last two are the ones that matter most. They are not missing cases, they are what a flat split does: it cannot nest, so anything inside a matched pair is treated as literal text, and escapes are invisible to it.

Tables are the reason this is worth doing at all. The model emits them constantly, because a table is the natural shape for the comparisons users actually ask for, and every one of them has been arriving as unreadable pipe soup.

## Shape

```
lib/markdown/inline.ts   emphasis, code spans, links, escapes  -> tree
lib/markdown/table.ts    GFM tables                            -> table
lib/markdown/blocks.ts   the block grammar                     -> blocks
components/Markdown.tsx  blocks + inline                       -> elements
```

Parsing is separated from rendering, so the entire grammar is a pure function of a string and is tested without a DOM. The component only turns the tree into elements.

## Rules worth knowing

**Scanner order is load-bearing** and is macOS': fenced code first, then tables, then everything else. Code first is what keeps a `---` or a `| a | b |` inside a code sample literal instead of turning it into a rule or a table.

**Code spans outrank everything inline.** That is what makes `` `**not bold**` `` work, and it means a delimiter inside a code span can no longer close an emphasis run that opened outside it.

**A single asterisk steps over a double run** when looking for its closer. Without that, `*italic with **bold** inside*` closed the italic on the opening `**` and produced garbage for the rest of the line.

**Underscores may not open or close inside a word,** so `my_var_name` stays an identifier. Asterisks still work intraword, which is what CommonMark says.

**Link destinations are read with balanced parentheses.** The first `)` is not the end: real URLs contain parens, and so does the very href this app refuses, so `[x](javascript:alert(1))` has to be read whole before it can be judged at all.

**Table cells split on pipes that are neither escaped nor inside a code span,** ported from macOS rather than guessed, so a cell can hold `\|` or `` `a | b` ``. And a table is only a table if the separator parses as alignments for exactly the header's columns, which is what stops a paragraph containing a pipe being read as one.

**Unclosed markers stay literal.** A reply is streamed, so a half-typed `**` is the normal state, not an error. One test renders every prefix of a representative document, because the renderer runs on every token and any prefix that looped would hang the app.

## Security

Chat is untrusted input in the ordinary case here, not only under attack: the prompt includes OCR of whatever is on the user's screen, so a reply can be steered by indirect prompt injection into emitting any markup it likes. Three rules follow, and they live in the parser with their own tests rather than inside a component.

**The href allowlist is carried over and extended.** http, https and mailto become links; anything else renders its label as plain text with no anchor at all. New: an href containing a control character is refused outright, because URL parsers strip those before reading the scheme. That blocks `java\tscript:alert(1)`, and it also blocks `https://omi.me\t@evil.example`, where the scheme really is https so the allowlist alone passes it while the host actually visited is `evil.example`.

**Images are reduced to alt text and nothing is fetched.** Deliberate, not unimplemented. Rendering `` would make an injected address a tracking pixel that reports the user's IP and that the chat was opened, on load, with nothing clicked.

**No raw HTML, ever.** React elements only, which the tests assert directly: raw tags render as text, HTML inside a code block renders as text, and a markdown image produces no `img` element and no request.

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,340 passed, 18 skipped, 546 files. 105 of those are this change; the branch point was 5,235. The three pre-existing `Markdown` tests pass unmodified.
- `npx eslint .` - 0 errors.
- `npx electron-vite build` - clean.
- `check_brand_ui.py`, `check_lifecycle_headers.py` - pass.

## Mutation audit

49 regressions seeded, 48 caught.

It found four real gaps, all now closed, and three of them are the same shape: a guard that looked covered because every input tested against it also failed a different guard.

**The control-character rule in the href check** could be deleted with no test failing, because every case tested also failed the scheme allowlist. The case that gives it its own weight is `https://omi.me\t@evil.example`, above.

**The href check on the autolink path** was the same story until tested with a URL carrying a control character, which the autolink pattern admits because it excludes only whitespace.

**The intraword-underscore rule has two halves** and only the closing one was covered. `my_var_ end` separates them: the closer there is legal, so only the opening rule stops `var` italicising.

**The CRLF test asserted block kinds,** which pass either way because a lone carriage return still trims to empty and still splits the paragraphs. The carriage returns were surviving inside the text, where they reach the DOM.

One escape is left and is documented where it sits: a table ends at a blank line, which the cell-count rule below it already subsumes, since a blank line splits into a single empty cell. It stays because it states the rule a reader is looking for, which the cell count only implies, and because macOS carries the same pair.

## Deliberate divergences from macOS

| Divergence | Why |
|---|---|
| An unclosed fence keeps the rest of the reply as code; macOS keeps it as text | Mid-stream the closing fence has not arrived yet. Watching the block form reads better than watching raw markdown reflow the moment it closes. This is also the shipped Windows behaviour, so it is a divergence being kept rather than introduced. |
| No citation tokens | macOS' inline citation masking belongs to its chat citation feature, which Windows does not have. |
| No inline-code copy button | macOS makes every inline code span its own copy target. Windows keeps the copy button on fenced blocks only, which is the existing behaviour and the one the current tests pin. |
| Images render as alt text | See above; this is a security position, not a gap. |

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11806?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/renderer/src/components/Markdown.render.test.tsx
+++ b/desktop/windows/src/renderer/src/components/Markdown.render.test.tsx
@@ -1,0 +1,148 @@
+// @vitest-environment jsdom
+// What actually reaches the DOM. The grammar is covered in lib/markdown; this
+// asserts the things only a render can show: that a table is a real table, that
+// a refused link is not an anchor, and that nothing here can inject HTML.
+import { describe, expect, it, afterEach } from 'vitest'
+import { render, cleanup, screen } from '@testing-library/react'
+import { Markdown } from './Markdown'
+
+afterEach(cleanup)
+
+function html(text: string): string {
+  return render(<Markdown text={text} />).container.innerHTML
+}
+
+describe('tables', () => {
+  const TABLE = '| Name | Size |\n| :--- | ---: |\n| alpha | 1 |\n| beta | 2 |'
+
+  it('renders a real table, not a paragraph of pipes', () => {
+    render(<Markdown text={TABLE} />)
+    expect(screen.getByRole('table')).toBeTruthy()
+    expect(screen.getAllByRole('columnheader').map((c) => c.textContent)).toEqual(['Name', 'Size'])
+    expect(screen.getAllByRole('row')).toHaveLength(3)
+  })
+
+  it('applies each column’s alignment', () => {
+    render(<Markdown text={TABLE} />)
+    const [left, right] = screen.getAllByRole('columnheader')
+    expect(left.className).toContain('text-left')
+    expect(right.className).toContain('text-right')
+  })
+
+  it('scrolls the table rather than the page', () => {
+    // A wide table must not widen the chat column.
+    render(<Markdown text={TABLE} />)
+    const region = screen.getByRole('region', { name: 'Scrollable table' })
+    expect(region.className).toContain('overflow-x-auto')
+  })
+
+  it('renders inline markdown inside cells', () => {
+    render(<Markdown text={'| a | b |\n| --- | --- |\n| **bold** | `code` |'} />)
+    expect(screen.getByText('bold').tagName).toBe('STRONG')
+    expect(screen.getByText('code').tagName).toBe('CODE')
+  })
+})
+
+describe('blocks', () => {
+  it('renders headings at their level', () => {
+    render(<Markdown text={'# One\n\n### Three'} />)
+    expect(screen.getByRole('heading', { level: 1 }).textContent).toBe('One')
+    expect(screen.getByRole('heading', { level: 3 }).textContent).toBe('Three')
+  })
+
+  it('renders a blockquote', () => {
+    expect(html('> quoted')).toContain('<blockquote')
+  })
+
+  it('renders a thematic break', () => {
+    expect(html('a\n\n---\n\nb')).toContain('<hr')
+  })
+
+  it('renders nested lists as nested lists', () => {
+    const { container } = render(<Markdown text={'- outer\n  - inner'} />)
+    expect(container.querySelectorAll('ul ul')).toHaveLength(1)
+  })
+
+  it('renders a task item as a read-only checkbox', () => {
+    render(<Markdown text={'- [ ] todo\n- [x] done'} />)
+    const boxes = screen.getAllByRole('checkbox') as HTMLInputElement[]
+    expect(boxes.map((b) => b.checked)).toEqual([false, true])
+    // A rendered reply is a transcript; the box reports, it does not edit.
+    expect(boxes.every((b) => b.readOnly)).toBe(true)
+  })
+
+  it('starts an ordered list at the number given', () => {
+    const { container } = render(<Markdown text={'3. a\n4. b'} />)
+    expect(container.querySelector('ol')?.getAttribute('start')).toBe('3')
+  })
+})
+
+describe('links', () => {
+  it('renders an allowed link as an anchor that opens externally', () => {
+    render(<Markdown text={'[docs](https://omi.me)'} />)
+    const link = screen.getByRole('link', { name: 'docs' })
+    expect(link.getAttribute('href')).toBe('https://omi.me')
+    expect(link.getAttribute('rel')).toBe('noreferrer')
+  })
+
+  it('renders a refused link as text with no anchor at all', () => {
+    // The label still shows; only the navigation is withheld.
+    render(<Markdown text={'[click me](javascript:alert(1))'} />)
+    expect(screen.queryByRole('link')).toBeNull()
+    expect(screen.getByText('click me')).toBeTruthy()
+  })
+
+  it('never emits a javascript href even inside other formatting', () => {
+    const rendered = html('**bold [x](javascript:alert(1)) tail**')
+    expect(rendered).not.toContain('javascript:')
+    expect(rendered).toContain('<strong>')
+  })
+
+  it('keeps a bolded link clickable', () => {
+    // The measured defect: this used to render the raw markdown characters.
+    render(<Markdown text={'**see [docs](https://omi.me) now**'} />)
+    expect(screen.getByRole('link', { name: 'docs' })).toBeTruthy()
+    expect(html('**see [docs](https://omi.me) now**')).not.toContain('](')
+  })
+})
+
+describe('nothing here can inject HTML', () => {
+  it('renders raw tags as text', () => {
+    const rendered = html('<script>alert(1)</script> and <img src=x onerror=alert(1)>')
+    expect(rendered).not.toContain('<script')
+    expect(rendered).not.toContain('<img')
+    expect(rendered).toContain('&lt;script&gt;')
+  })
+
+  it('renders a markdown image as its alt text and requests nothing', () => {
+    const rendered = html('![a diagram](https://tracker.example/p.png)')
+    expect(rendered).not.toContain('<img')
+    expect(rendered).not.toContain('tracker.example')
+    expect(rendered).toContain('a diagram')
+  })
+
+  it('renders HTML inside a code block as text', () => {
+    const rendered = html('```\n<script>alert(1)</script>\n```')
+    expect(rendered).not.toContain('<script')
+    expect(rendered).toContain('&lt;script&gt;')
+  })
+})
+
+describe('streaming', () => {
+  it('renders every prefix of a reply without throwing', () => {
+    // The renderer runs on every token, so each prefix must render.
+    const doc =
+      '# Title\n\nSome *text*.\n\n| a | b |\n| --- | --: |\n| 1 | 2 |\n\n' +
+      '> quoted\n\n- [ ] task\n  - deep\n\n---\n\n```ts\nconst x = 1\n```\nEnd.'
+    for (let n = 0; n <= doc.length; n++) {
+      expect(() => {
+        render(<Markdown text={doc.slice(0, n)} />)
+        cleanup()
+      }).not.toThrow()
+    }
+  })
+
+  it('keeps a half-typed marker literal', () => {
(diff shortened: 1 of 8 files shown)


Git-on-my-level:
Thanks @ZachL111 — this is a genuinely well-executed rewrite, and the PR description (defects measured against the shipped build before touching code) set a standard most PRs don't meet.

I reviewed every changed file at `a6cb0513` and ran the parser test suites in an isolated sandbox (86 tests, all green). What stood out:

- `lib/markdown/table.ts` — faithful GFM port: escaped pipes and pipes inside code spans stay cell content, the separator must parse as alignments for exactly the header's width, and body rows stop at a blank line or a <2-cell line so prose can follow. The two-column minimum correctly keeps ordinary prose containing a pipe out of table parsing.
- `lib/markdown/inline.ts` — the two headline defects are structural fixes, not patches: `**see [docs](https://x.com) now**` keeps a live link inside strong, and `\*not italic\*` shows neither backslashes nor emphasis. The scanner handles the subtle cases too (single `*` stepping over a `**` run instead of closing on its first char; intraword `_` rules).
- Security posture is **stronger** than what it replaces, which matters since this renders untrusted model output: the `https?/mailto` allowlist is kept and `isSafeHref` additionally rejects any href containing a control character — closing both the `java\tscript:` scheme-split and the tab-strip userinfo trick where the host a user reaches differs from the one they read. Images deliberately render as alt text only (no fetch-on-load tracking pixel), refused links keep their label but no anchor, and React-only rendering means `<script>` stays text — all pinned by tests in `Markdown.render.test.tsx` and `inline.test.ts`.
- `lib/markdown/blocks.ts` — fence-first scanner order keeps `---` and pipes inside code samples literal; the paragraph loop guarantees forward progress; and the every-prefix termination tests address the real risk of a parser that runs on every streamed token.
- `components/Markdown.tsx` — the component is now just a tree→element mapper; the `group/codeblock` scoping fixes the copy-button hover leak, and the table lives in its own `overflow-x-auto` region so wide tables scroll instead of widening the chat column.

One minor, non-blocking nit I verified while probing edges: a deeper-indented plain-text line after a list item (e.g. `- item` followed by a 4-space `wrapped text`) becomes a separate paragraph rather than a CommonMark lazy continuation. Rare in chat replies; fine to ignore or note for later.

No split requested — this is cohesive work on one surface. Flagging for human maintainer review before merge: this replaces the entire rendering stack for untrusted chat output with ~1,000 new lines, and while the security design here is better than the base, adopting a new parser implementation for that surface deserves a maintainer's sign-off.

---

ZachL111:
Thanks for going through it file by file, and for running the parser suites yourself rather than taking the description's word for it.

I reproduced the lazy-continuation nit and it behaves slightly differently from how it reads, so here is what actually happens, in case it matters to whoever picks it up later.

Input:

```
- item one
    wrapped text
- item two
```

The indented line does become its own paragraph, and the list is also **split in two**: one list holding `item one`, the paragraph, then a second list holding `item two`. So the structural damage is a little wider than a stray paragraph.

The visible result is milder than that sounds, which is why I agree it is not worth holding the PR for. The wrapped text is still rendered, just outside its bullet, and ordered numbering survives the split because each list carries the number it starts at:

```
1. first step
    continued line
2. second step
3. third step
```

parses to a list starting at 1, a paragraph, then a list starting at 2, so the reader still sees 1, 2, 3. Nothing is dropped and nothing is misnumbered.

I would rather not fix it inside this PR. Lazy continuation means treating an indented non-marker line as part of the item above, and indented lines are also how a nested code block or quote arrives, so the rule needs its own cases to avoid swallowing those. That is a contained follow-up, and doing it here would churn a diff you have already read end to end.

On the maintainer question you flagged: agreed that adopting a new parser for untrusted model output deserves a sign-off rather than my say-so. The one thing I would point at for that decision is that the security properties are the part I made hardest to regress: the href allowlist, the control-character rule, no-image-fetch, and no-raw-HTML each have their own cases, and a mutation audit confirmed each of them fails the suite when removed. That is the surface where being wrong is expensive, and it is the surface with the most tests per line.

Git-on-my-level (changes requested):
Follow-up to my earlier review of this same head (`a6cb0513`) — I found one real gap it missed, this time by executing the parser modules rather than only reading them. Everything in that earlier pass still holds, and the lazy-continuation analysis in @ZachL111's reply matches what I measure (the list splits in two, numbering survives via each list's `start`); I agree it is a fine follow-up.

## The finding: `parseBlocks` recurses without a depth cap, so reply content can crash the chat panel

`inline.ts` guards its own recursion the right way (`if (depth < 6)` in `parseInline`, content past the cap emitted as plain text). The two block-level recursions have no such guard:

- `lib/markdown/blocks.ts`, quote branch — `parseBlocks(quoted.join('\n'))` recurses once per `>` marker.
- `lib/markdown/blocks.ts`, `readList()` — a deeper-indented line recurses into `readList(lines, cursor)`, once per indent step.

Measured by running this PR's own modules in an isolated sandbox:

- A single line of 4,750+ consecutive `>` — exactly the shape of a model repetition loop, on the surface this PR itself documents as steerable by indirect prompt injection — throws `RangeError: Maximum call stack size exceeded`.
- ~4,500 spaced quote levels and ~5,961 indent-nested list levels throw the same way.

Consequences, checked against `main`:

- The throw is contained: every route panel is wrapped in an `ErrorBoundary` (`components/layout/MainViews.tsx`, `panel:${entry.id}` + `PanelErrorFallback`), so this is a dead chat panel, not a dead app.
- The triggering message is persisted with the conversation, so the panel crash-loops on every mount until that message is gone. One poisoned reply kills the chat surface for that conversation.
- It also contradicts `blocks.ts`'s own header contract — "Never throws and never drops input" — on precisely the untrusted-input surface this rewrite exists to harden. The renderer being replaced had no quote rule, so the same input was just a paragraph; this is a new failure mode, introduced with the feature.

The fix should be small and mirrors the precedent already in this PR: thread a depth through `parseBlocks`/`readList` (or cap quote-marker stripping) and emit literal text past the cap, the way `parseInline` already does. A test in the spirit of the existing robustness suite would pin it: a line of 5,000 `>` renders as text and does not throw.

## File-by-file, verified by execution

- `lib/markdown/blocks.ts` — the finding above. Everything else I ran behaves as documented: fence-first scanning keeps `---` and pipes inside code samples literal, and the paragraph loop's unconditional take guarantees forward progress.
- `lib/markdown/blocks.test.ts` — the every-prefix termination tests are the right idea but cover one small benign document; the adversarial-depth cases above are the gap.
- `lib/markdown/inline.ts` — reproduced exactly: `**see [docs](https://x.com) now**` keeps a live link inside strong, `\*not italic\*` shows neither backslash nor emphasis, `[x](java\tscript:alert(1))` and `https://omi.me\t@evil.example` are refused, `JAVA\tSCRIPT:` is refused case-insensitively. The `depth < 6` guard here is the pattern the block level is missing.
- `lib/markdown/inline.test.ts` — the control-character href tests (scheme-split, NUL prefix, userinfo `@` trick) all reproduce; consistent with the mutation-audit claim.
- `lib/markdown/table.ts` — verified: escaped pipes and pipes inside code spans stay cell content, the separator must parse as alignments for exactly the header's width, and the two-column floor keeps ordinary prose with a stray pipe out of table parsing. A 5,000-row table parses in ~7 ms.
- `lib/markdown/table.test.ts` — the seven "what is not a table" cases all reproduce; that negative suite is the valuable part.
- `components/Markdown.tsx` — the component is now just a tree→element mapper; table cells share the paragraph inline path (so the href rules apply there too), and the `overflow-x-auto` region keeps wide tables from widening the chat column.
- `components/Markdown.render.test.tsx` — the right level for the DOM contract (real table roles, refused link is not an anchor, no HTML injection); adding the deep-nesting input to this suite would cover the rendering path of the finding too.

This is the only defect I could find in an unusually well-built change, and it is a contained fix on top of work that deserves to land.

Still leaving the merge decision for human maintainer review, as before: adopting a new parser implementation for untrusted chat output is a maintainer sign-off call, and the depth cap above should ride along with it.

---
*by AI on behalf of David*

aryanorastar:
Keeping this open. Not approving `a6cb0513c424`. `desktop/windows/src/renderer/src/lib/markdown/table.ts` absent on origin/main. +1668. Hygiene SUCCESS.

PR #11807, opened 2026-08-18
feat(listen): deliver cloud-generated proactive messages to desktop over listen websocket

aryanorastar:
## What changed and why
Delivers cloud-generated proactive messages (mentor prompts, realtime third-party app interjections) directly to the desktop macOS app in real time over the active `/v4/listen` WebSocket stream. Previously, mid-session proactive interjections from `utils/app_integrations.py` only reached mobile FCM notifications/webhooks, leaving the desktop ambient listener disconnected from realtime suggestions.

none

## How it was verified
1. Ran backend unit tests covering `ProactiveMessageEvent` serialization, Redis pub/sub dispatching, and process-local session registry:
   `python3 -m pytest tests/unit/test_proactive_listen_delivery.py -v` -> 10 passed
2. Ran integration regression suite:
   `python3 -m pytest tests/unit/test_async_app_integrations.py -v` -> 18 passed
3. Built and executed desktop Swift unit tests covering event decoding, empty body gating, type recognition, and TTS delivery policy:
   `swift test --filter ProactiveListenEventTests` -> 9 passed
4. Formatted modified Python files with `black` and verified zero regressions.

## Tests
- Backend: `backend/tests/unit/test_proactive_listen_delivery.py` (10 unit tests)
- Integration: `backend/tests/unit/test_async_app_integrations.py` (18 integration tests)
- Desktop: `desktop/macos/Desktop/Tests/ProactiveListenEventTests.swift` (9 Swift unit tests)

## Failure class (fixes)
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aryanorastar:
@Git-on-my-level @kodjima33 review request — this one looks like it got missed. Its siblings in the same proactivity/wake-word batch (#11801, #11804, #11808, #11809) all picked up review on 08-18; this one hasn't had a look yet.

Tip `4635b74e`: delivers cloud-generated proactive messages to desktop over the listen websocket. 22/22 checks green — 10 backend + 18 integration + 9 Swift tests pass.

aryanorastar:
@Git-on-my-level @undivisible @kodjima33 review ping — this one has never had a reviewer assigned.

MERGEABLE, no failing checks. Open 2 days. Its three siblings from the same proactivity batch (#11801, #11804, #11864) all picked up review passes; this one was skipped rather than held.

Delivers cloud-generated proactive messages to desktop over the listen websocket — 10 backend + 18 integration + 9 Swift tests. Happy to answer anything inline.

backend/routers/listen/registry.py
```diff
@@ -0,0 +1,96 @@
...
+        import os
+        import redis.asyncio as aioredis
+
+        redis_host = os.getenv("REDIS_HOST", "localhost")
+        client = aioredis.Redis(
+            host=redis_host,
+            port=6379,
+            db=0,
+            decode_responses=True,
+        )
```
Git-on-my-level: 🔴 **Critical:** This subscriber will not see what `publish_proactive_message` publishes in prod.

Publisher uses the shared client in `database/redis_db.py` (`REDIS_DB_HOST`, `REDIS_DB_PORT`, `REDIS_DB_PASSWORD`, username `default`). This loop invents a second client on `REDIS_HOST` (wrong env), port 6379, no password.

Subscribe failure then `return`s forever, so the feature can pass unit tests and do nothing live.

`get_async_redis_client()` was added on this branch and is unused — use that (or the existing client) instead of a one-off constructor.

desktop/macos/Desktop/Sources/AppState/AppState+ListenEvents.swift
```diff
@@ -520,6 +520,42 @@ extension AppState {
...
+      FloatingControlBarManager.shared.showNotification(
+        ownerID: authorizationSnapshot.ownerID,
+        title: title,
+        message: message,
+        assistantId: appId.isEmpty ? "proactive-listen" : appId,
+        sound: .default,
+        kind: .general,
+        authorizationSnapshot: authorizationSnapshot,
+        onPresented: { speech.notificationWasPresented() }
+      )
```
Git-on-my-level, line 610: ⚠️ **Warning:** This bypasses `NotificationService`.

Owner gating here is good, but the bar manager does not enforce the master notifications toggle, the off-by-default migration, or frequency throttle. `isProactive: true` then speaks the card if TTS is on.

Please deliver through `NotificationService` (same door as other proactive cards) so Off / frequency / future presence-withhold (#11864) still apply. Also drop the `message.prefix(80)` log — that is user-conversation content.

The `conversation_id` block below only logs; it does not refresh conversations.

desktop/macos/Desktop/Tests/ProactiveListenEventTests.swift
```diff
@@ -0,0 +1,158 @@
...
+  // MARK: - handleListenEvent gate: empty message body is dropped
+
+  func testHandlerDropsEmptyMessage() {
+    let state = AppState()
+
+    // An empty message body should be silently dropped (no crash, no
+    // notification). Without a logged-in runtime owner there's no
+    // FloatingControlBarManager delivery to observe, but the guard on empty
+    // message precedes the owner check, so we verify it doesn't crash.
+    state.handleListenEvent(
```
Git-on-my-level: ⚠️ **Warning:** These tests do not prove delivery.

Calling `handleListenEvent` with no runtime owner and asserting “didn’t crash” does not cover show / suppress / TTS. The `NotificationSpeech.utterance` cases below re-test existing policy, not this handler.

Please add a behavioral test that a non-empty `proactive_message` is presented when notifications are allowed, and suppressed when the master toggle / frequency / owner gate would drop it.

Git-on-my-level (commented):
## Review

**Verdict: changes requested** — not an obvious bugfix. This is a new always-on interrupt path.

### What this is adding
Today cloud mentor / third-party realtime interjections are saved to chat and pushed to **phone**. Desktop can be the listener and still never see them live.

This PR publishes the same text to Redis, a new backend subscriber forwards it over the live `/v4/listen` socket, and macOS pops the floating bar and can speak it.

The *gap* is real. The *implementation* is a new bus, not a one-liner.

### UX
Desktop already has a real notification policy in `NotificationService` (master toggle, off-by-default migration, frequency throttle, snooze, owner checks). This handler skips that service and calls `FloatingControlBarManager.showNotification` directly, then queues TTS with `isProactive: true`.

That means:
- Notifications Off on Mac can still get a spoken floating-bar card
- Desktop frequency limits do not apply
- Meetings / other people present are not considered — this would punch a hole in #11864
- #11804 (speech-driven director) is flag-gated and off in prod/beta; this path has no flag
- Phone already got the FCM; desktop would now interrupt too

If this lands, the first time someone is on a call and Omi talks over them from a mentor prompt, that is the bug.

### Please fix before merge
1. Use the existing Redis client (`REDIS_DB_HOST` / port / password). The subscriber as written is likely dead in prod.
2. Deliver through `NotificationService` so Off / frequency / future presence-withhold stay one door.
3. Add a test that actually asserts show vs suppress, not just “didn’t crash.”

I did not run the suites or exercise a live listen session; this is from the diff and surrounding policy code.

aryanorastar:
@Git-on-my-level thanks — all four are fixed, and I took the feature out of unit tests and ran it.

## The four findings

| Finding | Fix |
|---|---|
| 🔴 Subscriber won't see prod publishes | `cb5dfc6f` — now `await get_async_redis_client()`, the helper you spotted I'd added and left unused |
| ⚠️ Bypasses `NotificationService` | `bb11c5ee` — `sendNotification(..., respectFrequency: true)` |
| ⚠️ Tests prove nothing | `04eb951d` — `ProactiveListenAdmission` + 14 assertions |
| ⚠️ `message.prefix(80)` logs user content | logs provenance only |

I also removed the `conversation_id` block (`47db3ee8`). Its comment said "Refresh conversations if the message is tied to a specific conversation" and the body only logged the id — a comment describing behavior the code does not have. The id stays on the wire, so adding a real refresh later needs no protocol change.

## Live run — backend leg

Real Redis, the real `publish_proactive_message`, the real dispatcher, a registered session:

| | Result |
|---|---|
| Fixed — publisher and subscriber on one server | `DELIVERED` — `mentor` / `Omi` / `"You said you'd call the bank."` / `conv-e2e-1` |
| Old — subscriber on a different reachable server | `NOTHING DELIVERED`, no error raised |

The second row is your point reproduced rather than reasoned about: subscribe succeeds against the wrong server and the feature is silently dead.

## Live run — desktop leg

Named dev bundle, signed in, real handler in-process. I added a temporary `inject_listen_event` bridge action to feed one raw event through `handleListenEvent`, drove it with `omi-ctl`, and toggled the setting through the real `set_notification_settings` action. The seam was removed afterwards and is not in this diff.

```
notifications ON   09:21:50  Transcription: Proactive message from mentor
                   09:21:50  FloatingControlBarManager: dropping notification because window is not set up

notifications OFF  09:22:23  Transcription: Proactive message from mentor
                             (no FloatingControlBarManager line)
```

With the toggle off the message is swallowed inside `NotificationService` before the presentation layer — which is only true because it now routes through it. The old code would have presented either way. That log line is also the PII fix: provenance only, no message body.

Bar-window drop in the first case is the onboarding state of a fresh dev profile, not a gate.

## The one red check

`Hygiene` fails on `backend-async-blockers`, and it is not this diff. The scanner marks every name imported from `database.*` as a blocking DB call without checking whether it is awaited, so the correct `await get_async_redis_client()` trips it:

```
async_helpers_with_blocking: backend/routers/listen/registry.py:49 | proactive_message_dispatcher | get_async_redis_client:58
```

That gate is why the original code hand-rolled a client — commit `11e7de9b` is literally titled "use direct async redis client to satisfy async blocker check". Working around the linter is what pointed it at the wrong server in the first place. #12060 fixes the scanner; this goes green once it lands.

Suites: 14 Swift, 10 backend, 18 integration.

kodjima33 (approved):
Feature (new desktop delivery path for cloud-generated proactive messages over listen websocket), not a bug fix — mixed area (backend registry/redis + desktop AppState), riskiest area is backend. Per policy, backend features are approve-only, not merged. Well-tested (10 backend + 18 integration + 9 Swift unit tests), CI green.

PR #11809, opened 2026-08-18
feat(desktop): halt voice playback on user speech barge-in

aryanorastar:
## What changed and why
Adds voice barge-in interruption during ambient listening. When the assistant is speaking (delivering proactive voice advice, answering queries, or synthesizing TTS) and the user begins speaking, playback is halted immediately so Omi never talks over the user.

- INV-CHAT-1

## How it was verified
1. Built and executed unit tests:
   `swift test --filter VoiceBargeInPolicyTests` -> 5 passed in 0.006s
   - User speech interrupts active playback
   - User speech is a no-op when not playing
   - Other speaker turns on multi-party calls do not interrupt
   - Empty/whitespace utterances are ignored
   - Speaker 0 is treated as primary user
2. Added e2e flow coverage entry in `capture-lifecycle.yaml`
3. Added changelog fragment

## Tests
- `desktop/macos/Desktop/Tests/VoiceBargeInPolicyTests.swift` (5 unit tests)

## Failure class (fixes)
Failure-Class: none

### DIFF
--- a/desktop/macos/Desktop/Sources/AppState/AppState+ListenEvents.swift
+++ b/desktop/macos/Desktop/Sources/AppState/AppState+ListenEvents.swift
@@ -23,7 +23,7 @@ extension AppState {
         isSpeaking: FloatingBarVoicePlaybackService.shared.isSpeaking
       ) {
         log("Transcription [BARGE-IN]: User spoke mid-playback; interrupting voice output")
-        FloatingBarVoicePlaybackService.shared.stop()
+        FloatingBarVoicePlaybackService.shared.interruptCurrentResponse()
       }
 
       // Convert backend segment to local SpeakerSegment
--- a/desktop/macos/Desktop/Sources/AppState/AppState+ListenEvents.swift
+++ b/desktop/macos/Desktop/Sources/AppState/AppState+ListenEvents.swift
@@ -14,6 +14,18 @@ extension AppState {
       // Extract speaker_id from backend (e.g. "SPEAKER_00" → 0)
       let speakerId = segment.speaker_id ?? 0
 
+      // Barge-in interruption: if the user speaks while voice playback is active,
+      // halt playback immediately so Omi never talks over the user.
+      if VoiceBargeInPolicy.shouldInterrupt(
+        isUser: segment.is_user,
+        speaker: speakerId,
+        text: segment.text,
+        isSpeaking: FloatingBarVoicePlaybackService.shared.isSpeaking
+      ) {
+        log("Transcription [BARGE-IN]: User spoke mid-playback; interrupting voice output")
+        FloatingBarVoicePlaybackService.shared.stop()
+      }
+
       // Convert backend segment to local SpeakerSegment
       let translations = (segment.translations ?? []).map {
         SegmentTranslation(lang: $0.lang, text: $0.text)
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/VoiceBargeInPolicy.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/VoiceBargeInPolicy.swift
@@ -5,7 +5,7 @@ import Foundation
 /// When the assistant is speaking (proactive advice, voice response, or TTS)
 /// and the user begins speaking, barge-in halts playback immediately so Omi
 /// never talks over the user.
-enum VoiceBargeInPolicy {
+enum VoiceBargeInPolicy: Sendable {
   /// Evaluates whether an incoming transcript segment warrants interrupting voice playback.
   ///
   /// - Parameters:
--- a/desktop/macos/Desktop/Sources/FloatingControlBar/VoiceBargeInPolicy.swift
+++ b/desktop/macos/Desktop/Sources/FloatingControlBar/VoiceBargeInPolicy.swift
@@ -1,0 +1,29 @@
+import Foundation
+
+/// Pure policy deciding when incoming ambient speech interrupts active voice output.
+///
+/// When the assistant is speaking (proactive advice, voice response, or TTS)
+/// and the user begins speaking, barge-in halts playback immediately so Omi
+/// never talks over the user.
+enum VoiceBargeInPolicy {
+  /// Evaluates whether an incoming transcript segment warrants interrupting voice playback.
+  ///
+  /// - Parameters:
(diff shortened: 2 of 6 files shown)


Git-on-my-level:
Thanks @aryanorastar - clean, well-scoped v1 for barge-in, with the decision logic isolated in a testable policy type and CI green across the desktop suites. Reviewed each changed file:

- `VoiceBargeInPolicy.swift` - good extraction of the interrupt decision into a pure `Sendable` policy; the guard ordering (isSpeaking -> user identity -> non-empty text) reads well. One semantic worth a conscious product call: `speaker == 0` interrupts even when `is_user` is false, and `speaker_id` defaults to 0 when the backend omits it, so in a multi-party conversation where diarization collapses speakers, a non-user's utterance can also halt playback. "Never talk over anyone" is a defensible default - just want it to be deliberate (it is well covered by `testSpeakerZeroTreatedAsUserEvenIfIsUserFlagFalse`).
- `AppState+ListenEvents.swift` - `isSpeaking` covers synthesis and queued audio, not just audible output, so barge-in also prevents a queued TTS chunk from starting after the user begins speaking: nice. Two questions here:
  1. The call site uses the full-teardown `stop()` rather than `interruptCurrentResponse()`. `stop()` clears mode/tracer and, with the default `notifyPTTDrain: false`, drops `activePTTLease` without `VoiceTurnCoordinator.releaseOutput(...)`. In the pure ambient case (proactive advice with no active turn) there is no lease, so this is fine; but if a PTT output lease can be held while a backend ambient segment lands (race at PTT engage, or a second capture surface streaming into the same conversation), the lease would be dropped without a reducer event - the desktop voice-turn invariants keep user interrupt as a typed reducer event. Can you confirm ambient `handleBackendSegments` cannot run while a PTT lease is held, or route through `interruptCurrentResponse()`/the coordinator for that window?
  2. Interruption triggers on transcript-segment arrival, so "immediately" in practice means "when STT returns the segment" (cloud roundtrip vs local STT latency). Fine for v1 - just flagging that perceived responsiveness scales with STT latency; a VAD/energy-based early trigger could be a nice follow-up.
- `ChatProvider+JournalProjection.swift` - behavior-identical if-let refactor of `createdAt.map { formatter.string(from: $0) }`. Unrelated to barge-in but harmless.
- `VoiceBargeInPolicyTests.swift` - the 5 tests cover the matrix nicely (interrupt active playback, no-op when idle, other-speaker no-interrupt, whitespace-only no-interrupt, speaker-0 fallback), and `Desktop Swift Build & Tests` is green on this head.
- `changelog/unreleased/20260818-voice-barge-in-interrupt.json` - accurate and matches the changelog-fragment convention.
- `e2e/flows/capture-lifecycle.yaml` - registering `VoiceBargeInPolicy.swift` under capture-lifecycle coverage is the right slot given the call site lives in the transcription ingestion path; `desktop-core-e2e-t0` passes.

The direction (halting output when the user starts speaking) fits the floating-bar voice answers surface well. Leaving for human maintainer review before merge: product sign-off on the barge-in trigger semantics (diarization-fallback interrupts + transcript-arrival latency) and the PTT lease interaction question above.

---

aryanorastar:
Thank you @Git-on-my-level for the thorough review and great feedback!

Addressed the architectural points:
1. **PTT Lease & Teardown**: Updated the barge-in call site from full-teardown `stop()` to `FloatingBarVoicePlaybackService.shared.interruptCurrentResponse()`. This preserves the mode/tracer and properly resets the playback pipeline without dropping active PTT leases untracked. (Pushed in `eb3bd918`).
2. **Speaker 0 / Diarization Default**: Agree that "never talk over anyone" is the best safety default for v1 during live audio playback, while keeping multi-party non-user turns gated when distinct non-0 speaker IDs are resolved.
3. **Responsiveness / VAD Follow-up**: Excited to explore energy/VAD early triggering as a low-latency follow-up on top of the transcript-driven base.

Ready for final maintainer sign-off and merge!

Git-on-my-level:
Follow-up after `eb3bd918` — thanks @aryanorastar, the architectural points from the earlier review are addressed and verified on this head:

- `AppState+ListenEvents.swift` — the barge-in call site now routes through `interruptCurrentResponse()` instead of full-teardown `stop()`. Verified in `FloatingBarVoicePlaybackService`: this preserves `currentMode`/`tracer` (the tracer-drop comment explicitly reserves clearing for full teardown), cancels in-flight synthesis/playback via the `playbackGeneration` bump, drains both queues, and records `interruptedResponseID` so later chunks of the interrupted response are swallowed in the enqueue path rather than restarting playback after the user has barged in. That is exactly the routing the earlier review asked for, and it now matches the established ambient-interrupt idiom used at PTT engage in `PushToTalkManager`.
- `VoiceBargeInPolicy.swift` — unchanged on this head and still reads well; the 5-test matrix in `VoiceBargeInPolicyTests.swift` covers the cases that matter (idle no-op, other-speaker no-interrupt, whitespace guard, speaker-0 fallback).
- `ChatProvider+JournalProjection.swift` — the if-let form is behavior-identical to the previous `.map`; fine as a strict-concurrency cleanup, though unrelated to barge-in.
- Changelog fragment and the `capture-lifecycle.yaml` coverage entry are accurate, and the desktop suites are green on this head.

One residual note for the record (non-blocking, pre-existing idiom rather than something this PR introduces): a bare `interruptCurrentResponse()` goes through `resetPlaybackPipeline(clearMode: false)` whose `notifyPTTDrain` defaults to false, so in the narrow window where a PTT output lease is somehow held while an ambient backend segment lands, that lease would still be dropped without a `VoiceTurnCoordinator.releaseOutput` event — identical to the existing PTT-engage call sites. If that window ever proves reachable, threading a `leaseID:` through (or defaulting the drain there) would be the natural follow-up.

Adding `positive-signal` — clean policy extraction, responsive iteration, tests and CI green. Keeping `needs-maintainer-review` for product sign-off on the barge-in trigger semantics: the speaker-0/diarization-collapsed fallback (a non-user utterance halting playback when diarization collapses speakers) is deliberate for v1, but it deserves a human product yes before merge.

---

undivisible (approved):
Cline review agrees: VoiceBargeInPolicy is a small pure interrupt rule (user/speaker 0, non-empty, while speaking) with XCTests. Desktop Swift contracts/compile green. Merge.

PR #11808, opened 2026-08-18
feat(desktop): extract environmental speaker context to detect multi-party calls and meetings

aryanorastar:
## What changed and why
Extracts environmental multi-party speaker context from live diarization segments during ambient listening. Previously, diarization tags (`speaker_id`, `is_user`) were stored in raw segments but never summarized into an environmental context signal.

With this change:
- `EnvironmentalSpeakerAnalyzer` extracts active speaker count, non-user participant count, and identified participant names (e.g. Maya / Participant 1) over a bounded 180s sliding window.
- Produces a structured `EnvironmentalSpeakerSignal` and renders a safe, injection-sanitized prompt section (`== ENVIRONMENTAL / CALL CONTEXT ==`) for the director and assistant.
- Enables the assistant to recognize when the user is on a call or meeting with others (*"I noticed another person in your call"*).

## Product invariants affected
none

## How it was verified
1. Built and ran unit test suite covering all cases:
   `swift test --filter EnvironmentalSpeakerContextTests` -> 7 passed in 0.007s
   - Solo/monologue (no false multi-party trigger)
   - Two speakers (unnamed participant)
   - Two speakers (named participant from person map)
   - Multi-party group meeting (3+ speakers)
   - Active window pruning (stale turns from earlier calls excluded)
   - Label sanitization (prompt injection safe)
   - Director volatile prompt integration
2. Added e2e flow coverage entry in `context-buckets-dogfood.yaml`
3. Added changelog fragment

## Tests
- `desktop/macos/Desktop/Tests/EnvironmentalSpeakerContextTests.swift` (7 unit tests)

## Failure class (fixes)
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11808?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #11810, opened 2026-08-18
feat(desktop): multi-step action chaining pipeline with reversible receipts

aryanorastar:
## What changed and why
Adds multi-step action chaining pipeline to plan context-grounded agent actions with reversible receipts:
1. **Inspect Context**: Extracts active multi-party call participants and resolves individual preferences from long-term memory facts.
2. **Plan Proposal**: Assembles structured proposals (e.g. food orders, meeting setups) with items, costs, and participant assignments.
3. **Execute & Undo**: Emits reversible `ActionReceipt` with an undo/cancellation deadline window (e.g. 30 seconds).
4. **Cancel / Rollback**: Handles immediate cancellation requests before deadline expiration.

- INV-CHAT-1

## How it was verified
1. Ran unit test suite:
   `swift test --filter ActionChainingPipelineTests` -> 6 passed in 0.004s
   - Single-user order planning with memory facts
   - Multi-party participant order planning (e.g. Maya + User)
   - Execution producing reversible receipt
   - Cancellation within deadline window
   - Expired deadline rejection
   - Non-matching intent filter
2. Added e2e flow coverage entry in `capture-lifecycle.yaml`
3. Added changelog fragment

## Tests
- `desktop/macos/Desktop/Tests/ActionChainingPipelineTests.swift` (6 unit tests)

## Failure class (fixes)
Failure-Class: none


RatnamOjha:
Thanks for putting this together. I don't think it's ready to merge yet - the main issue is that nothing in the app actually calls this. `ActionChainingPipeline`, `ActionProposal`, and `ActionContextEnvironment` are only referenced by the tests, so this currently ships no user-facing behavior despite the changelog describing it as a feature.

Main things to address:

1. **`execute()` doesn't execute anything.** It doesn't call a service or persist state; it only returns a receipt. `cancel()` similarly mutates a local struct, so the reversible-receipt behavior isn't implemented yet.

2. **Conflicts with `INV-CHAT-1`.** This moves intent parsing, planning, execution, and undo ownership into Swift, while the existing `AuthorizedToolExecution` path says the kernel owns invocation truth and authorization state.

3. **`covers:` isn't real coverage.** `capture-lifecycle` doesn't exercise this code, and `desktop-flow-lint.py` only checks that the path exists. Please add a flow that drives this code or remove the entry.

4. **Intent matching is too broad.** `lower.contains("order")` also matches words like `"border"` and `"reorder"`. Use word-boundary matching at minimum.

5. **Remove demo logic from production code.** The `"Chef's Special / Pad Thai"` fallback and `items.count * 16` pricing shouldn't be in the shipped implementation.

Also, #11808 already adds `EnvironmentalSpeakerSignal`, which looks very similar to `ActionContextEnvironment`. I'd reuse that rather than introducing another type.

Happy to take another look once this is wired to a real call site through the kernel path.

aryanorastar:
Thanks for the thoughtful review and feedback @RatnamOjha! Completely agree with your architectural points — client-side hardcoded demo heuristics don't belong in the production tree, and tool execution/undo semantics belong squarely under the kernel-backed `AuthorizedToolExecution` / `RealtimeToolAuthority` boundary per `INV-CHAT-1`.

Closing this PR in favor of keeping the core tool-authority architecture clean. We'll integrate multi-step action proposals directly through the server-side tool execution pipeline in a dedicated follow-up.

PR #11811, opened 2026-08-18
Mutation-audit the Windows suite: close 24 proven holes, and one real Rewind defect

ZachL111:
This adds no features and changes no production source. Every file it touches is a test.

`AGENTS.md` is explicit that coverage grows by ratchet and not by mandate, and that a suite people distrust is worse than a smaller suite. So rather than adding tests to raise a number, I ran a mutation audit against the existing Windows suite and only wrote cases where a specific regression provably shipped silently. Every case below cites the mutation it closes, and each mutation was re-run afterwards to confirm the case actually turns the suite red.

## How this was found

For a target module, find the test files that reach it transitively by import, seed a mutation, and run only those tests. A mutation that leaves them green is a proven hole: that exact regression ships today with no failure anywhere.

The mutations are generated rather than chosen, from two operators that almost always compile and almost always mean a real behaviour change: a comparison moved one step (`>=` to `>`, `<=` to `<`), and a numeric literal changed. Everything the generator flagged was then adjudicated by hand, because plenty of mutations are equivalent and prove nothing. The equivalent ones are called out below rather than quietly dropped.

## The pattern this kept finding

Almost every hole had the same shape: **the tests sit in the comfortable middle of a range and never on its edge.**

`captureDecision` has three comparisons that decide what of the user's screen is recorded. The existing cases test idle at 120 seconds against a threshold of 60, "very different" against identical, and 29 and 31 seconds against a 30 second window. All correct, all well clear of the boundary, and so all three comparisons could be moved a step with the suite staying green.

The boundaries are not academic. They are whether a frame is recorded at the instant the user walks away from the machine, whether a screen that changed by a hair is stored again, and whether a static screen still gets its periodic anchor in the timeline.

`orphanSelection` is the same shape with more at stake, because its return value is passed to `unlink`. The grace window is the only thing standing between the sweep and a file whose database row is still being written, and every existing case cleared it by a second or more.

## What else the audit turned up

**A frame hash that cannot see colour.** `frameHash` builds every fixture from a helper that sets blue, green and red to the same value. A greyscale fixture cannot see which channels the luminance reads, so swapping red for a second copy of blue left all four cases green. This hash is the only thing deciding whether two frames are "the same screen", so a hash blind to a channel means a screen that changed only in that channel is treated as unchanged and never stored. The user finds nothing in Rewind for that stretch and nothing anywhere says why.

**Three unreached rules in the timeline geometry, behind 47 tests.** `REWIND_LINEAR_MIN_PX` is the floor that gives a short burst of activity a clickable width: measured, a two second block between two idle gaps has a natural width of 0.056px, so without the floor it is invisible, and nothing noticed the floor going to zero. The tick-budget comparison is inclusive because the tick count across a span is `floor(span/interval) + 1`, and one step to exclusive silently coarsens every axis. And the six-hour tick interval was never selected by any test, so it could be deleted from the table without a failure; reaching it takes a twenty-hour span with a five-tick budget.

**A privacy list with no test at all.** `BUILT_IN_EXCLUDED_APPS` and `SENSITIVE_WINDOW_MARKERS` are the lists that stop Rewind storing password manager windows and login screens. Nothing tested either. They were reached only indirectly, and three of the twelve title markers were exercised.

## A real defect, found by writing that last one

The exclusion list is matched as a case-insensitive **substring** against the app and process name together, and two entries are short enough to appear inside unrelated names.

**`OBS` is a substring of Obsidian, so Rewind never records an Obsidian window.** Omi ships an Obsidian export integration (`main/memoryExport/obsidian.ts`), so this is a product the app expects its users to have. `Loom` is a substring of Bloomberg. Also caught: Observer, and any executable with "obs" in the name.

I have documented this rather than fixed it, deliberately. The two failure directions are not symmetric. An over-exclusion means an app is not recorded. A botched narrowing means a password manager IS recorded. Narrowing these entries safely needs the real process names of OBS Studio and Loom on Windows, which is a maintainer's call rather than something a test change should decide. The cases are written so that fixing the list turns them red, which tells whoever does it exactly which behaviour they changed.

A second, smaller finding from the same file: four entries contain a space, and a Windows process name never does, so `Snipping Tool`, `Snip & Sketch`, `Proton Pass` and `Keeper Password` cannot match an executable and depend entirely on the friendly app name being reported.

## The trap I walked into myself

The first version of the `captureDecision` boundary block could not fail. It built its fixtures with `hashDifferingBy(DUP_HAMMING_THRESHOLD)` and then asserted behaviour at that distance, so zeroing the constant moved the fixture with it and the assertion still passed. The audit caught it in a test written minutes earlier.

Every fixture here is now a literal rather than an expression over the constant it brackets. A boundary test has to state the boundary.

Two constants are pinned directly, and both cite a source outside this repo rather than the code under test. `KEYFRAME_ANCHOR_MS` matches macOS' `frameDedupeMaxInterval = 30.0` (`RewindIndexer.swift:36`, which also uses `<=` at line 218). The dedup threshold is pinned because macOS uses 5 where Windows uses 4, so the two platforms already disagree about a screen that changed by exactly five bits, and that disagreement should not move silently.

The exclusion tests also failed me twice before this was committed: I asserted twelve markers as fifteen, and asserted that an app called "Password Generator" was excluded when the app list holds product names and not the bare word. That is the argument for having them.

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,330 passed, 18 skipped, 543 files.
- `npx eslint .` - 0 errors.
- No production source file is modified; `git diff --stat` is five test files.

**24 mutations seeded across the five modules, every one of them re-run after the cases were written. All 24 turn the suite red now. All 24 left it green before.**

| Module | Mutations | What they represent |
|---|---|---|
| `captureDecision.ts` | 5 | idle, dedup and anchor boundaries; both constants |
| `orphanSelection.ts` | 3 | grace boundary, grace window, newest-signal rule |
| `frameHash.ts` | 5 | each colour channel, the average comparison, the length guard |
| `timelineGeometry.ts` | 4 | the min-width floor, the tick budget, the six-hour interval |
| `rewindExclusions.ts` | 7 | a dropped password manager, a dropped marker, case-sensitivity, each check disabled |

## What is deliberately not here

Bulk coverage. I measured first: of 650 Windows source files over 40 lines, only 75 are never reached by any test, and most of those are Electron wiring or React pages where a unit test would assert the mock. This is a well-tested codebase, and adding cases to raise a number would be the noise `AGENTS.md` warns about.

Two escapes are recorded as equivalent mutations rather than papered over with a case that would only be asserting the mutation. In `xToTs`, changing `px <= pc.xEnd` to `<` changes nothing observable, because adjacent pieces share the timestamp at their common edge and both readings return the same value. In `retentionCutoff`, changing `retentionDays <= 0` to `< 0` is equivalent because the arithmetic on the other branch produces the same answer at zero.

The harness itself is not included. It is a developer tool with no blocking audience, and `AGENTS.md` is clear that a check nothing runs is a dead check. I am happy to contribute it wired into a CI lane as a separate change if that is wanted.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11811?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/rewind/captureDecision.test.ts
+++ b/desktop/windows/src/main/rewind/captureDecision.test.ts
@@ -19,10 +19,16 @@ describe('shouldCaptureFrame', () => {
     expect(shouldCaptureFrame(base)).toEqual({ capture: true })
   })
   it('skips when the screen is locked', () => {
-    expect(shouldCaptureFrame({ ...base, locked: true })).toEqual({ capture: false, reason: 'locked' })
+    expect(shouldCaptureFrame({ ...base, locked: true })).toEqual({
+      capture: false,
+      reason: 'locked'
+    })
   })
   it('skips when the user is idle past the threshold', () => {
-    expect(shouldCaptureFrame({ ...base, idleSeconds: 120 })).toEqual({ capture: false, reason: 'idle' })
+    expect(shouldCaptureFrame({ ...base, idleSeconds: 120 })).toEqual({
+      capture: false,
+      reason: 'idle'
+    })
   })
   it('skips when a previous frame is still processing', () => {
     expect(shouldCaptureFrame({ ...base, busy: true })).toEqual({ capture: false, reason: 'busy' })
@@ -52,18 +58,26 @@ describe('shouldCaptureFrame', () => {
     expect(shouldCaptureFrame({ ...base, excludedApps: ['', '  '] })).toEqual({ capture: true })
   })
   it('does not exclude an unrelated app', () => {
-    expect(
-      shouldCaptureFrame({ ...base, appName: 'Notepad', excludedApps: ['chrome'] })
-    ).toEqual({ capture: true })
+    expect(shouldCaptureFrame({ ...base, appName: 'Notepad', excludedApps: ['chrome'] })).toEqual({
+      capture: true
+    })
   })
   it('skips a login page by window title (sensitive)', () => {
     expect(
-      shouldCaptureFrame({ ...base, appName: 'Google Chrome', windowTitle: 'Sign in - Google Accounts' })
+      shouldCaptureFrame({
+        ...base,
+        appName: 'Google Chrome',
+        windowTitle: 'Sign in - Google Accounts'
+      })
     ).toEqual({ capture: false, reason: 'sensitive' })
   })
   it('skips an incognito window by title (sensitive)', () => {
     expect(
-      shouldCaptureFrame({ ...base, appName: 'Google Chrome', windowTitle: 'New Tab - Google Chrome (Incognito)' })
+      shouldCaptureFrame({
+        ...base,
+        appName: 'Google Chrome',
+        windowTitle: 'New Tab - Google Chrome (Incognito)'
+      })
     ).toEqual({ capture: false, reason: 'sensitive' })
   })
   it('skips a password page by title (sensitive)', () => {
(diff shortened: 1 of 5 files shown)


Git-on-my-level:
Thanks @ZachL111 — checked the whole diff against the head sources, and every factual claim in the description holds up. This is careful, useful test work.

**Test-only, confirmed.** All five changed files are `*.test.ts`. The ten deleted lines in `captureDecision.test.ts` are reformats of existing assertions, not behavior changes.

**Pins verified against head.** `DUP_HAMMING_THRESHOLD = 4` and `KEYFRAME_ANCHOR_MS = 30_000` (`captureDecision.ts:5,13`); the `>=` grace boundary and the conservative newer-of-filename/mtime rule (`orphanSelection.ts:63`, `ORPHAN_GRACE_MS = 60_000`); the `REWIND_LINEAR_MIN_PX = 3` floor that keeps a two-second activity block visible and clickable (`timelineGeometry.ts:181`); and the 21-entry app list / 12 title markers (`rewindExclusions.ts`). The macOS references are accurate too — `frameDedupeMaxInterval = 30.0` at `RewindIndexer.swift:36` with `<=` at `:218`, and `dedupThreshold = 5` with the "spinner = 1, cursor = 4, real change = 23" note in `RewindOCRService.swift` — so pinning the Windows-vs-macOS disagreement at exactly five bits is fair commentary.

**The Rewind defect is real.** `isExcluded()` (`captureDecision.ts:41-47`) matches the built-in list as case-insensitive substrings of `${appName} ${processName}`, so `OBS` silently excludes Obsidian (and Observer, and anything containing "obs"), and `Loom` excludes Bloomberg — while Omi ships an Obsidian export integration (`main/memoryExport/obsidian.ts`). Those windows have never been recorded. Documenting the collisions as tests that turn red when someone fixes the list, rather than narrowing it inside a test PR, is the right call given the asymmetric failure modes.

Two things worth a maintainer's eye:

1. Narrowing `OBS`/`Loom` to the real Windows process names (`obs64.exe` etc.) is deliberately left as a follow-up. It needs someone who can verify those names, since a botched narrowing means a password manager gets recorded — that tradeoff is a maintainer decision, not a test-change decision.
2. The count pins (`toHaveLength(21)` / `toHaveLength(12)`) mean every future exclusion-list addition touches this test file. Intentional per the comments (a count catches silent shrinkage the per-entry loops cannot), just worth a conscious sign-off.

One note for anyone reading the verification section: the mutation harness itself isn't included (disclosed, offered separately as a CI lane), so the 24/24 numbers rest on the write-up. The assertions themselves I checked by hand against the production logic, and they do pin the stated boundaries. This is exactly the kind of boundary-literal, ratchet-style test work the repo's testing guidance asks for.

---

kodjima33 (approved):
Test-only mutation audit — no production source touched, CI green. Holding the merge for a maintainer call: the rewindExclusions cases deliberately pin the current OBS/Loom substring behaviour, so narrowing that list later turns them red.

ZachL111:
Thanks for checking the pins against head rather than against the description, and for confirming the exclusion collisions independently.

Two answers to the things you flagged for a maintainer's eye.

**The count pins are a deliberate tradeoff and I am happy for them to be overruled.** `toHaveLength(21)` and `toHaveLength(12)` do mean every future entry touches this file. I picked that because the per-entry loops iterate the exported arrays, so they shrink silently with a deletion: remove `1Password` and the loop simply runs 20 times and stays green. A count is the only assertion in the file that can notice a removal. If the friction turns out to be worse than the protection, the honest replacement is to drop the counts and keep the named password-manager list, which catches the deletions that matter and ignores additions.

**On the harness not being in the diff.** That was deliberate for the reason you would expect, and it leaves the 24 numbers resting on the write-up, which is a fair thing to point out. Two things make them checkable without the harness: every mutation is one edit to a named constant or comparison stated in the PR body, so any of them can be reproduced by hand in under a minute; and the assertions themselves are the evidence you already verified, since a case that pins a boundary literal fails when the boundary moves whether or not a harness drove it.

If a mutation lane in CI is wanted I would rather propose it as its own change with a real budget discussion, because the honest version is slow: it is one test run per mutation, so it belongs on a nightly or a label rather than on every PR.

The follow-up on narrowing `OBS` and `Loom` is yours to schedule. If you want me to do it once someone confirms the real process names, I will take it with tests in both directions: the tool still excluded, the collision no longer excluded.

Git-on-my-level:
Thanks @ZachL111 — both answers land, and they close out the two items I'd flagged for a maintainer's eye.

**Resolved from my side:**

- The count pins are now on record as a deliberate tradeoff with a stated fallback (drop the counts, keep the named password-manager list — which still catches the deletions that matter). That is exactly the framing the sign-off needed; I have no objection to the pins as written.
- Leaving the harness out while making every mutation hand-reproducible in under a minute is a fair answer to the write-up-only 24/24. For the record I re-verified the assertions directly against head sources and they pin what they claim: the `>=` idle boundary at 60/60, dedup at 4-vs-5 bits, the anchor window at exactly 30_000 vs 30_001, the grace boundary at 60_000/59_999 with the newer-of-filename/mtime rule feeding `unlink`, the strictly-greater average bit in `averageHash`, the 3px linear floor, and the inclusive tick budget that selects the six-hour interval over a twenty-hour span.

**The one thing left is a product call, not a code issue.** @kodjima33's approval holds the merge on whether the suite should pin today's `OBS`/`Loom` substring over-exclusions as documented behavior — these tests deliberately turn red the moment someone narrows the list — and if so, when to schedule that narrowing to the real Windows process names, which you've offered to take with coverage in both directions. Either landing the pins as-is or switching to the named-list fallback is defensible; that choice and the follow-up scheduling just need a maintainer to make it.

No code concerns outstanding. Leaving for human maintainer review on that pinning decision.

---

kodjima33 (approved):
Test-only mutation audit of the Windows suite; no production source touched, each case cites the mutation it closes.

PR #11814, opened 2026-08-18
fix(backend): route Modulate-primary live STT through the provider fallback chain

### LINKED ISSUE #11752: prod-omi-backend-listen lost 100% of sessions for ~50min: Modulate over monthly quota + Modulate-primary has no provider fallback
## Live outage: `prod-omi-backend-listen` lost 100% of its sessions for ~50 min — Modulate is over its monthly quota and a Modulate-primary session has no fallback

Two independent problems compounded. The first is fixed (ops); the second is a code gap that is still open.

### 1. What broke (fixed at 16:03Z by restoring declared config)

Since **2026-08-17T15:15:52Z**, every live session landing on the `prod-omi-backend-listen` deployment (6 replicas) died ~150 ms after connecting:

```
15:50:41.028  INFO  Connecting to Modulate Velma-2 streaming sample_rate=16000 language=multi
15:50:41.170  INFO  Modulate Velma-2 streaming connection established
15:50:41.170  ERROR Modulate streaming error: Monthly usage limit reached.
15:50:41.223  INFO  Listen supervisor exited reason=lifetime_done
```

Measured 15:45–16:00Z, namespace-wide: **651 Modulate connects → 650 `Monthly usage limit reached`** (100% failure). Clients reconnected every ~7–10 s in a loop, getting zero transcript each time.

**Root cause: cluster drift.** The live deployment's pod spec had `STT_SERVICE_MODELS=modulate-velma-2,dg-nova-3,parakeet`, but the Helm release values *and* `backend/charts/backend-listen/prod_omi_backend_listen_values.yaml` (and `deploy/runtime_env/prod.overlay.yaml`) all declare `dg-nova-3,modulate-velma-2,parakeet`. `helm get values` confirms the release never carried Modulate-first — it was patched out-of-band, presumably to route around the Deepgram 402s in #11695. That also contradicts 98abcaed74 ("Revert: make Velma-2 the streaming default"). The sibling `-dg-canary` deployment (11 replicas) was never drifted and kept working.

**Mitigation applied:** `kubectl set env deployment/prod-omi-backend-listen STT_SERVICE_MODELS=dg-nova-3,modulate-velma-2,parakeet` (restores the declared value; no policy change). Rollout completed 16:03Z. On the new replicaset since 16:04Z: **0 Modulate quota errors, Deepgram connects succeeding, transcript segments flowing again.**

### 2. Still open — both STT vendors are in billing failure

- **Modulate:** monthly usage limit reached; 100% of live connects rejected. Needs a quota/plan action.
- **Deepgram:** still 402ing, and it is flapping, not fixed. Per-minute, namespace-wide, on the current image:

| minute | DG 402 | DG connect fail | DG connect OK |
|---|---|---|---|
| 16:02 | 207 | 52 | 6 |
| 16:03 | 672 | 169 | 24 |
| 16:04 | 272 | 71 | 22 |

Balance on project `d052cc59` is **−$68,325.52** (Mgmt API `/balances`, see #11695). So the restored order buys partial service (some sessions connect) instead of none — it is not a fix. **Both vendor accounts need money; no code change makes either accept a session.**

Parakeet is not a drop-in third option right now: it is capability-limited (continuous `from=parakeet to=deepgram reason=capability_mismatch`) and its own pods log `Parakeet WS recv error: no close frame received or sent` 13× in 15 min.

### 3. Code gap (unfixed, needs an owner)

`backend/routers/listen/receiver.py:223`:

```python
if self.host.stt_service == STTService.modulate:
    return await process_audio_modulate(modulate_callback or callback, sample_rate, self.host.stt_language)
```

Deepgram-primary and Parakeet-primary both route through `connect_stt_socket_with_fallback`. **Modulate-primary has no fallback at all**, even though this deployment listed `dg-nova-3,parakeet` after it. That is what turned "vendor over quota" into "every session dies."

Note the fallback added in #11725 would *not* have covered this: Modulate accepts the WebSocket and only then sends an `error` frame, so the failure is post-connect. `_recv_loop` calls `_mark_dead(...)` and the supervisor exits **`reason=lifetime_done`** — i.e. a total provider outage is logged as a *normal* session end, with no failure telemetry and no `omi_fallback_event`. Two things worth fixing:

1. A provider that dies before the first transcript should fail over to the next configured provider, not terminalize.
2. A provider error frame must not exit as `lifetime_done`; it should surface as an upstream failure so alerts fire.

Filed by the hourly desktop-bug watchdog. Ops mitigation applied and verified; items 2 and 3 left to humans (billing + a hot-path change to live STT session setup).

RatnamOjha:
# Fix Modulate-primary STT fallback

## Problem

Modulate-primary live STT sessions did not fail over to the configured backup providers.

In production, Modulate hit its monthly quota and **100% of live sessions failed for ~50 minutes**, even though Deepgram and Parakeet were configured as fallbacks.

There was also a second issue: Modulate can accept the WebSocket upgrade and report `Monthly usage limit reached` ~150ms later. The fallback helper treated the initial socket as a success, so fallback was never triggered.

## Fix

* Route Modulate-primary sessions through the shared STT fallback helper.
* Build the fallback chain from the failed primary so a provider is never retried as its own fallback.
* Validate primary socket liveness before accepting a post-connect provider.
* Record post-connect rejection as a provider failure and continue to the next provider.
* Keep provider/language serving-policy checks and existing fallback telemetry.

Result:

```text
Modulate primary
      ↓
healthy? ── yes → use Modulate
      │
      no
      ↓
   Deepgram
      ↓
   Parakeet
```

## Verification

Regression tests reproduced the production failure on the unmodified code:

* **10/11 tests failed before the fix**
* **11/11 passed after the fix**
* **40/40 passed** across the affected STT/fallback test suites
* `pyright`: 0 errors
* `black`: clean

## Tradeoff

Modulate-primary sessions now spend ~320ms validating the connection before adopting it, because Modulate's quota rejection arrives after the WebSocket upgrade.

This cost is limited to Modulate-primary sessions; Deepgram- and Parakeet-primary paths keep their existing fast path. The grace period is configurable via `STT_FALLBACK_LIVENESS_GRACE_SECONDS`.

## Known follow-ups

* Derive fallback ordering directly from `STT_SERVICE_MODELS` instead of the current fixed order.
* Tune the liveness grace period using measured provider rejection latency.
* Fix provider-error accounting for mid-session failures (`lifetime_done`) in a separate PR.

## Verification note

The full `backend/test.sh` suite is currently unstable on this machine and fails on the unmodified base commit as well. The affected STT/listen suites for this change are green.

Fixes #11752

Line-Count-Exception: backend/utils/stt/streaming.py | 1616 -> 1683 | fallback chain, post-connect liveness proof and Modulate circuit; splitting a live-STT hot-path module is separate work

### DIFF
--- a/backend/routers/listen/receiver.py
+++ b/backend/routers/listen/receiver.py
@@ -48,6 +48,7 @@ from utils.stt.live_failure import (
 from utils.stt.streaming import (
     STTService,
     connect_stt_socket_with_fallback,
+    deepgram_fallback_model,
     make_stream_callback,
     modulate_is_configured_fallback,
     parakeet_is_configured_fallback,
@@ -223,7 +224,50 @@ class ListenReceiver:
                 self.host.stt_model = 'velma-2'
             return socket
         if self.host.stt_service == STTService.modulate:
-            return await process_audio_modulate(modulate_callback or callback, sample_rate, self.host.stt_language)
+            # Velma-2 accepts the upgrade and only then reports being over quota,
+            # so a Modulate primary needs the same chain its siblings use (#11752).
+            dg_fallback_model = deepgram_fallback_model(self.host.stt_language)
+
+            def connect_deepgram_fallback() -> Any:
+                return process_audio_dg(
+                    callback,
+                    self.host.stt_language,
+                    sample_rate,
+                    1,
+                    model=cast(str, dg_fallback_model),
+                    keywords=keywords,
+                    is_active=lambda: self.host.state.active,
+                )
+
+            def connect_parakeet_fallback() -> Any:
+                return process_audio_parakeet(
+                    callback,
+                    self.host.stt_language,
+                    sample_rate,
+                    1,
+                    model='parakeet',
+                    keywords=keywords,
+                    is_active=lambda: self.host.state.active,
+                )
+
+            socket, actual_service = await connect_stt_socket_with_fallback(
+                primary_service=STTService.modulate,
+                connect_primary=lambda: process_audio_modulate(
+                    modulate_callback or callback,
+                    sample_rate,
+                    self.host.stt_language,
+                ),
+                connect_deepgram=connect_deepgram_fallback if dg_fallback_model else None,
+                connect_parakeet=(
+                    connect_parakeet_fallback if parakeet_is_configured_fallback(self.host.stt_language) else None
+                ),
+            )
+            self.host.stt_service = actual_service
+            if actual_service == STTService.deepgram:
+                self.host.stt_model = cast(str, dg_fallback_model)
+            elif actual_service == STTService.parakeet:
+                self.host.stt_model = 'parakeet'
+            return socket
         if self.host.stt_service == STTService.deepgram:
 
             def connect_deepgram() -> Any:
--- a/backend/tests/unit/test_deepgram_connection_fallback.py
+++ b/backend/tests/unit/test_deepgram_connection_fallback.py
@@ -371,11 +371,6 @@ def test_parakeet_is_a_configured_fallback_only_when_the_deployment_can_serve_it
         assert streaming.parakeet_is_configured_fallback('en') is False
 
 
-@pytest.mark.anyio
-async def test_modulate_primary_is_still_rejected():
-    with pytest.raises(ValueError, match='modulate'):
-        await streaming.connect_stt_socket_with_fallback(
-            primary_service=STTService.modulate,
-            connect_primary=AsyncMock(),
-            connect_modulate=AsyncMock(),
-        )
+# A Modulate primary is no longer rejected by the helper — it walks the chain like
+# its siblings. That contract, including the "never falls back to itself" guard this
+# assertion used to provide, now lives in test_modulate_connection_fallback.py (#11752).
--- a/backend/tests/unit/test_modulate_connection_fallback.py
+++ b/backend/tests/unit/test_modulate_connection_fallback.py
@@ -1,0 +1,330 @@
+"""Regression (#11752): a Modulate-primary session must walk the configured fallbacks.
+
+``prod-omi-backend-listen`` ran ``STT_SERVICE_MODELS=modulate-velma-2,dg-nova-3,parakeet``
+and lost 100% of its live sessions for ~50 minutes: Modulate accepted every
+WebSocket upgrade and then answered ``Monthly usage limit reached.``. Deepgram and
+Parakeet were listed right behind it, but ``_create_stt_socket`` called
+``process_audio_modulate`` directly and returned, so a Modulate primary had no
+fallback at all while the Parakeet and Deepgram legs sat idle.
+
+Deepgram-primary (#11695) and Parakeet-primary sessions already route through
+``connect_stt_socket_with_fallback``. These tests pin the Modulate entry point to
+the same contract.
+"""
+
+import os
+from contextlib import contextmanager
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from routers.listen import receiver as receiver_mod
+from routers.listen.receiver import ListenReceiver
+from utils.stt import provider_resilience, streaming
+from utils.stt.streaming import STTService
+
+
+@pytest.fixture
+def anyio_backend():
+    return 'asyncio'
+
+
+def _modulate_receiver():
+    """Duck-typed receiver exposing only what `_create_stt_socket` touches."""
+    host = SimpleNamespace(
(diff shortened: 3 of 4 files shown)


aryanorastar:
LGTM! Great work on this fix.

Key highlights from review:
1. **Liveness Gating**: Isolating `_primary_is_serving` check specifically to `_POST_CONNECT_REJECTING_PRIMARIES` ensures Modulate's post-upgrade quota packet is caught without adding latency to the synchronous fast-paths for Deepgram/Parakeet.
2. **Self-Fallback Exclusion**: Filtering `service != primary_service` cleanly prevents cyclical re-attempts of the failing primary.
3. **Socket Cleanup**: Using `close_rejected_socket(socket)` on post-connect rejection prevents connection leaks across retry cycles.
4. Test matrix in `test_stt_fallback.py` covers the primary-to-backup permutations thoroughly.

RatnamOjha:
Thank you! Waiting on @kodjima33 's and @Git-on-my-level 's approval before merging.

Git-on-my-level:
Verified this end to end against the pre-fix base — really clean work, and the final commit correcting the leg-order comment (stating the fixed order rather than implying it is policy-derived) is exactly the right instinct.

What I checked:

- `backend/routers/listen/receiver.py` — the Modulate branch now routes through `connect_stt_socket_with_fallback`, with the Deepgram leg gated on the new `deepgram_fallback_model(...)` and the Parakeet leg on `parakeet_is_configured_fallback(...)`, mirroring the Parakeet/Deepgram sibling branches. Fallbacks correctly receive the plain `callback` (not `modulate_callback`), and `host.stt_service`/`stt_model` are updated per serving provider (`nova-3`/`parakeet`) so `_serving_provider()` telemetry stays truthful after a failover.
- `backend/utils/stt/streaming.py` — the rebuilt `ordered` candidate list excludes the primary, preserving the "never falls back to itself" invariant that the deleted test used to pin. `_primary_is_serving` limits the liveness grace to `_POST_CONNECT_REJECTING_PRIMARIES` (just Modulate today), so Deepgram/Parakeet primaries keep their existing fast path. A rejected Modulate socket is released via `close_rejected_socket` and classified as `quota` by `_fallback_failure_reason`, matching Velma's "Monthly usage limit reached." shape. The new `_modulate_circuit` follows the sibling env-tunable pattern and stays independent of the other circuits.
- `backend/tests/unit/test_deepgram_connection_fallback.py` — removing `test_modulate_primary_is_still_rejected` is legitimate: that `ValueError` existed because the old candidate list hard-coded Modulate as the first fallback, so admitting a Modulate primary would have retried the provider that just failed. The replacement guard (never-falls-back-to-itself) lives in the new suite.
- `backend/tests/unit/test_modulate_connection_fallback.py` — 11 behavioral tests driving the real receiver seam and real policy gates rather than stubs: over-quota upgrade moves the session to Deepgram; Modulate + Deepgram both in billing failure reaches Parakeet; the rejected socket is `finish()`ed (no leak); telemetry legs record `modulate→deepgram recovered`; a Modulate-only policy fails closed instead of inventing a provider; the Deepgram leg receives the plain callback.

Verification: I ran both fallback suites (27 passed) plus the wider listen/STT receiver suites (174 passed), and re-ran the new suite against the pre-fix `streaming.py`/`receiver.py` — 10/11 fail there, matching the PR's regression claim.

Two non-blocking follow-ups worth tracking:

1. `routers/chat.py`'s transcribe-stream still has the same bypass (`elif stt_service == STTService.modulate: dg_socket = await process_audio_modulate(...)`), and the PTT surface defaults list `modulate-velma-2` first — the same outage shape could recur on that surface. Pre-existing and out of scope here, but it deserves a follow-up issue.
2. `deepgram_fallback_model` only recognizes `dg-`-prefixed tokens in `STT_SERVICE_MODELS`; bare `nova-3`-style tokens allowed by `DEEPGRAM_MODEL_TOKENS` would be skipped. Every deployment config and the runtime-env validator use the `dg-` form, so this is fine today — just a heads-up if the token space ever widens.

Leaving for human maintainer review before merge: this is live-STT provider routing, and the two deliberate tradeoffs — the fixed Deepgram→Parakeet leg order (not derived from `STT_SERVICE_MODELS`) and the liveness grace (default 0.3s, `STT_FALLBACK_LIVENESS_GRACE_SECONDS`) now paid by Modulate-primary session setup — deserve a maintainer sign-off.

---

kodjima33 (approved):
Routes Modulate-primary live sessions through the shared STT fallback chain and treats a post-connect 'usage limit' rejection as a provider failure — reproduces the 50-minute prod outage in regression tests on unmodified code, CI green.

PR #11817, opened 2026-08-18
fix(ios): let the watch companion build for the simulator without a team

### LINKED ISSUE #11776: iOS simulator builds require an Apple developer team because the watch target demands signing
A contributor with no Apple developer account cannot run the app on the **iOS simulator**, because the watch companion target requires a development team even for a simulator destination. The simulator is normally the account-free on-ramp, so this removes the only path that doesn't require Apple Developer Program setup.

## Symptom on current `main`

Fresh clone of `main` at `015172400b`, iPhone 17 Pro (iOS 27) simulator booted first, then the documented command `bash setup.sh ios`:

```
Watch companion app found.
Xcode build done.                          31.3s
Error (Xcode): No Account for Team "9536L8KLMP". Add a new account in Accounts settings
               or verify that your accounts have valid credentials.
Error (Xcode): No profiles for 'com.friend-app-with-wearable.ios12.development.watchapp' were found
Could not build the application for the simulator.
```

The build never produces an app. On `main` this is two problems stacked: `DEVELOPMENT_TEAM = 9536L8KLMP` is hardcoded in all 33 occurrences of `project.pbxproj`, *and* something is demanding provisioning for a simulator build.

## Isolating the second problem from the first

The hardcoded team is already addressed by the community-build work in #7641. To check whether that is the *whole* story, I built a tree where the dev-flavor configs read `$(DEVELOPMENT_TEAM)` from an xcconfig, and passed an empty team explicitly:

```
xcodebuild -workspace Runner.xcworkspace -scheme dev \
  -destination 'platform=iOS Simulator,id=<booted sim>' \
  DEVELOPMENT_TEAM="" build
```

```
error: Signing for "omiWatchApp" requires a development team. Select a development team in
       the Signing & Capabilities editor. (in target 'omiWatchApp' from project 'Runner')
** BUILD FAILED **
```

**Only `omiWatchApp` fails.** The iOS app target and the widget target build without a team. So parameterizing the team — which is what #7641 does — changes *which* team is used but not that one is *required*. After that PR lands, a contributor who has no Apple developer account at all still cannot build for the simulator.

Relevant project settings on `main`: `CODE_SIGN_STYLE = Automatic` (29 configs) and `CODE_SIGN_IDENTITY = "Apple Development"` unqualified (14 configs) rather than scoped `[sdk=iphoneos*]`.

## Why this is worth fixing separately

Every other Flutter project's simulator path costs nothing: no account, no team, no profiles. Here, both paths are gated — the device path needs an account by nature, and the simulator path refuses too. For a contributor evaluating whether to work on Omi at all, that is the first thing they hit.

## Caveats, stated up front

- **Measured on Xcode-beta 27.0 (27A5237l)** (`xcode-select -p` → `/Applications/Xcode-beta.app`), Flutter 3.44.9. `setup.sh` documents Xcode 16.4; whether that version demands a team for a watch target on a simulator destination is **unverified**. If it does not, this is a newer-Xcode regression rather than a longstanding gap — either way the project settings are what decide it.
- One earlier attempt at the isolation build was invalid and is not the evidence above: adding `-sdk iphonesimulator` forces the watch target to compile against the iOS SDK and fails with `Unable to resolve module dependency: 'WatchKit'`, which has nothing to do with signing. The run quoted above passes `-destination` only.
- In the #7641 flow, `setup.sh` also hard-requires a team independently of the project settings (`detect_apple_team_id()` prompts interactively, and returns non-zero when no terminal is attached). So opening the account-free path would take a change in both places, not just one.

Found while running the documented iOS pathway end-to-end from a fresh clone as a new contributor would. Related: #11774 (harness unreachable from a device), #11775 (`setup.sh` device pinning and prerequisite validation), #11534 (splash hang with no harness), and #7641 (the community-build signing work).

formed2forge:
## Summary

A contributor with no Apple developer account couldn't run the app on the iOS simulator — normally the account-free on-ramp — because the watch companion target (`omiWatchApp`) requires a development team even for a simulator destination:

```
Error (Xcode): No Account for Team "9536L8KLMP". Add a new account in Accounts settings...
Error (Xcode): No profiles for '...development.watchapp' were found
```

## Root cause

Root-caused with `-showBuildSettings` and the real build log's tool invocations, not guessed. Two hypotheses were tested and ruled out first:

- **A missing watchOS simulator runtime** looked promising (my machine had none installed) but was disproven with a clean rebuild after installing the runtime and pairing a watch simulator to the test device — the watch target's resolved `SDKROOT` was still the *device* WatchOS SDK, confirmed both via `-showBuildSettings` and the `ExecuteExternalTool ... -isysroot .../WatchOS.platform/...` lines in the actual build log.
- **A `Base.xcconfig`-level `CODE_SIGNING_ALLOWED[sdk=*simulator*] = NO` override** (an earlier attempt) never matched anything, for the same reason: the watch target's own resolved SDK isn't a simulator SDK even when the overall scheme destination is a simulator, so an SDK-qualified condition can't distinguish this build from a genuine device build.

## Fix

Since Runner and the widget already build for the simulator without a signing identity, and the watch companion target is `SKIP_INSTALL = YES` (it's only ever embedded in `Runner.app`, never independently installed), disabling signing specifically for its **dev-flavor** configurations is safe:
- Does not touch Runner's or the widget's signing at all.
- Does not affect prod/beta, where the watch app may still need real signing for distribution.

## Testing

**No automated test** — this is an Xcode project-settings fix with no Linux-CI-reachable seam (`Dart Analyze & Tests` runs on Ubuntu, where Xcode doesn't exist), consistent with how #7641's signing changes were verified: manually, on macOS, with a real build.

**Verified live and reproducibly**, from a fully clean `DerivedData`:
- Raw `xcodebuild` using the exact invocation `flutter run` constructs (`-allowProvisioningUpdates -allowProvisioningDeviceRegistration -workspace ... -scheme dev -destination id=<simulator>`) → `** BUILD SUCCEEDED **`.
- The real `flutter run --flavor dev -d <simulator>` → `Launching lib/main.dart on iPhone 17 Pro in debug mode...` — no signing errors, with the exact same team (`9536L8KLMP`) this machine has no account for.

Failure-Class: none

Fixes #11776.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11817?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


formed2forge:
All CI checks are green and this is ready for maintainer review/merge — I don't have write access to merge it myself.

Summary: the watch companion target (`omiWatchApp`) required a development team even for simulator builds, blocking the account-free on-ramp for contributors without an Apple developer account. Root-caused with `-showBuildSettings` and the real build log (two other hypotheses — missing watchOS simulator runtime, an SDK-scoped xcconfig override — were tested and ruled out first, both documented in the PR description). Fix disables signing only on the watch target's dev-flavor configs, directly in `project.pbxproj`; Runner/widget signing and prod/beta are untouched. Verified live and reproducibly with the exact invocation `flutter run` constructs.

kodjima33 (approved):
Three-line dev-flavor signing fix for the watch companion so simulator builds work without a team; root-caused with -showBuildSettings, prod/beta signing untouched. Fixes #11776.

PR #11818, opened 2026-08-18
Test the bounds the app promises: delegation spend caps and the cold-start quota box

ZachL111:

- INV-CHAT-1

`desktop/windows/src/main/agentKernel/**` is one of the INV-CHAT-1 path globs
(`docs/product/invariants/chat-continuity.md:150`), and `kernel.test.ts` sits
under it. The citation is path-based, not intent-based: nothing here changes
chat behaviour, and the invariant's guard test is untouched. I checked every
other invariant doc for a glob matching either changed file and INV-CHAT-1 is
the only match.

No production source changes. Two test files.

The app makes numeric promises to the user: how much a delegated agent may spend, and how long a send may wait before it gives up. This closes the cases where the number carrying that promise had nothing testing it. Each case cites the mutation it closes, and each mutation was re-run afterwards to confirm the case turns the suite red.

## Delegation spend and depth caps

`kernel.test.ts` already rejects `maxBudgetUsd` of 0 and 11, and `maxDepth` of 0 and 6. That covers past both ends of each range and never the end itself, which is the part the error message promises the caller:

```
Delegation maxBudgetUsd must be greater than 0 and at most 10
Delegation maxDepth must be between 1 and 5
```

Moving either comparison one step (`>` to `>=`) rejects the exact maximum those messages advertise. Neither mutation failed a test.

Separately, nothing read either DEFAULT. Every existing case passes an explicit value, so cutting `DEFAULT_DELEGATION_MAX_BUDGET_USD` from 5 to 1 left the suite green. That default is the number deciding what an ordinary delegation may spend when the caller names nothing, and it is recorded into the delegation's request payload, which is what a later audit of "what was this agent allowed to spend" reads.

Three cases now: the budget ceiling is inclusive, the depth ceiling is inclusive, and a delegation naming no bounds gets the documented defaults.

## The chat quota gate's cold-start time box

Every case in `chatQuotaGate.test.ts` passes an explicit `5_000`, so none of them reads `COLD_START_SYNC_TIMEOUT_MS`. Production does the opposite: `barSend.ts` calls `createChatQuotaGate()` with no arguments and runs entirely on the default.

So setting that constant to 0 left the whole file green while defeating the cold-start gate in the shipped app. The box would fire before the first sync could land, and a user already over their cap would get a free send, which is precisely what the neighbouring case ("cold start forces exactly ONE sync so an over-cap user gets no free send") exists to prevent.

The new case takes the default, holds a probe just inside the box, and proves the verdict still gates.

## Why the constants are pinned

Both are pinned with the reasoning that sets them rather than as bare numbers, so that changing one is a decision rather than an edit.

Five seconds is chosen against a real bound and the source says so: `apiClient` retries 429/503 with backoff over five tries up to 16s apart, so an unbounded await would hold a send, and a voice turn, for roughly 30s with no feedback.

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,240 passed, 18 skipped, 542 files.
- `npx eslint .` - 0 errors.

**7 mutations seeded, each the regression its case was written for. All 7 turn the suite red now. All 7 left it green before.**

| Mutation | Was |
|---|---|
| `DEFAULT_DELEGATION_MAX_BUDGET_USD` 5 to 1 | green |
| `DEFAULT_DELEGATION_MAX_DEPTH` 3 to 1 | caught already |
| budget ceiling `>` to `>=` | green |
| depth ceiling `>` to `>=` | green |
| `COLD_START_SYNC_TIMEOUT_MS` to 0 | green |
| `COLD_START_SYNC_TIMEOUT_MS` to 1 | green |
| cold-start `withTimeout` removed entirely | green |

## How these were found

The same mutation harness as #11811, pointed at a different question: the app exports 42 timing and budget bounds, and 17 are never named by any test. That list is a starting point rather than a work item, because a constant no test names can still be covered behaviourally and pinning a literal for its own sake is the noise `AGENTS.md` warns about. These two are the ones where a mutation proved the bound was doing unwatched work and where the consequence lands on the user's money or the user's wait.

Worth recording what this run did NOT find, since the absence is also information. I chased three suspected gaps and measured all three as non-issues before writing anything:

- I suspected Windows 11's `graphicsCaptureProgrammatic` privacy toggle silently gates Rewind, making the onboarding comment ("Windows has NO OS consent prompt for desktop capture") stale. Setting both that consent and its `NonPackaged` desktop-app entry to `Deny` and re-running a real Electron capture left capture completely unaffected. The comment is correct.
- Same for the microphone consent: with `microphone\NonPackaged` set to `Deny`, `getUserMedia({audio:true})` still returned a live, unmuted track.
- Rewind playback transport, which an older parity audit lists as missing, already ships.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11818?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

- INV-CHAT-1

### DIFF
--- a/desktop/windows/src/main/agentKernel/kernel.test.ts
+++ b/desktop/windows/src/main/agentKernel/kernel.test.ts
@@ -22,6 +22,12 @@ import { afterEach, describe, expect, it } from 'vitest'
 import { AgentRuntimeKernel } from './kernel'
 import { AdapterRegistry } from './adapterRegistry'
 import { SqliteAgentStore, type DatabaseFactory } from './store'
+import {
+  DEFAULT_DELEGATION_MAX_BUDGET_USD,
+  DEFAULT_DELEGATION_MAX_DEPTH,
+  HARD_DELEGATION_MAX_BUDGET_USD,
+  HARD_DELEGATION_MAX_DEPTH
+} from './kernelSupport'
 import type {
   AdapterAttemptContext,
   AdapterAttemptResult,
@@ -1240,6 +1246,54 @@ describe('AgentRuntimeKernel — delegateAgent', () => {
     await expect(
       kernel.delegateAgent(delegation(parentRunId, { maxBudgetUsd: 11 }))
     ).rejects.toThrow(/maxBudgetUsd must be greater than 0 and at most 10/)
+  })
+
+  // The two cases above reject 0 and 11, and 0 and 6, which leaves the end of
+  // each range untested — and the end is what the error message promises the
+  // caller: "at most 10", "between 1 and 5". Moving either comparison one step
+  // rejects the exact maximum the message advertises, and a mutation audit
+  // confirmed nothing here noticed.
+  it('allows a delegation at exactly the budget ceiling', async () => {
+    const { kernel } = newKernel(fakeAdapter())
+    const parentRunId = await parentRun(kernel)
+
+    await expect(
+      kernel.delegateAgent(
+        delegation(parentRunId, { maxBudgetUsd: HARD_DELEGATION_MAX_BUDGET_USD })
+      )
+    ).resolves.toBeTruthy()
+  })
+
+  it('allows a delegation at exactly the depth ceiling', async () => {
+    const { kernel } = newKernel(fakeAdapter())
+    const parentRunId = await parentRun(kernel)
+
+    await expect(
+      kernel.delegateAgent(delegation(parentRunId, { maxDepth: HARD_DELEGATION_MAX_DEPTH }))
+    ).resolves.toBeTruthy()
+  })
+
+  it('gives a delegation that names no budget the documented default', async () => {
+    // The default is the allowance a delegated agent gets when the caller says
+    // nothing, so it is the number that decides what an ordinary delegation may
+    // spend. Nothing read it: cutting it from 5 to 1 left the suite green,
+    // because every case above passes an explicit value.
+    const { kernel } = newKernel(fakeAdapter())
+    const parentRunId = await parentRun(kernel)
+
+    const created = await kernel.delegateAgent(delegation(parentRunId))
+    // The resolved bounds are recorded in the delegation's request payload,
+    // which is what a later audit of "what was this agent allowed to spend"
+    // reads. Asserting the literal as well as the constant so that moving the
+    // constant is a visible change and not a silently-tracking one.
+    const request = JSON.parse(created.delegation.requestJson) as {
+      maxBudgetUsd: number
+      maxDepth: number
+    }
+    expect(request.maxBudgetUsd).toBe(5)
+    expect(request.maxBudgetUsd).toBe(DEFAULT_DELEGATION_MAX_BUDGET_USD)
+    expect(request.maxDepth).toBe(3)
+    expect(request.maxDepth).toBe(DEFAULT_DELEGATION_MAX_DEPTH)
   })
 
   it('refuses to delegate from another owner’s run', async () => {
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Test-only change, and a genuinely useful one — thanks. I verified each new case against the head sources and everything checks out.

**desktop/windows/src/main/agentKernel/kernel.test.ts**
- The two new ceiling cases close a real gap: existing coverage rejected `maxBudgetUsd` of 0/11 and `maxDepth` of 0/6, so the inclusive endpoints the error messages promise ("at most 10", "between 1 and 5") were never exercised. Both match `kernelCore.ts`, which rejects only `<= 0` / `> HARD_DELEGATION_MAX_BUDGET_USD` and `< 1` / `> HARD_DELEGATION_MAX_DEPTH`.
- The defaults case is the most valuable: `kernelCore.ts` records `input.maxDepth ?? DEFAULT_DELEGATION_MAX_DEPTH` and `input.maxBudgetUsd ?? DEFAULT_DELEGATION_MAX_BUDGET_USD` into the delegation's `requestJson`, and no prior test read either constant. Asserting both the literal and the constant makes any future change to the default a visible review decision — good call.

**desktop/windows/src/renderer/src/lib/chatQuotaGate.test.ts**
- Existing cases all passed an explicit `5_000`, so `COLD_START_SYNC_TIMEOUT_MS` had no reader, while production (`barSend.ts` calls `createChatQuotaGate()` with no arguments) runs entirely on the default. The fake-timer case that lands a blocked quota at `COLD_START_SYNC_TIMEOUT_MS - 1` exercises exactly that default path; verified `check()` awaits `withTimeout(sync(), coldStartTimeoutMs)` only on cold start.
- The `toBe(5_000)` pin is opinionated but justified: the comment ties the bound to the apiClient retry envelope, so changing it should be a conscious decision rather than a silent refactor.

One blocker, and it is metadata rather than code: the Hygiene and PR Metadata Preflight checks fail because this PR touches a path mapped to the chat-continuity invariant. Paste this into the PR description and they will re-run:

- INV-CHAT-1

No code change is needed. Leaving for normal maintainer merge once the description carries the citation — no other concerns.

---

Git-on-my-level:
The metadata blocker from my earlier review is resolved — thanks for adding the INV-CHAT-1 citation to the description. Hygiene and PR Metadata Preflight are green again on the unchanged head, so the per-case verification from that review still stands:

- **desktop/windows/src/main/agentKernel/kernel.test.ts** — the inclusive-ceiling cases (`maxBudgetUsd: HARD_DELEGATION_MAX_BUDGET_USD`, `maxDepth: HARD_DELEGATION_MAX_DEPTH`) and the defaults case (literal + constant assertions against the delegation's `requestJson`) all match `kernelSupport.ts` (`HARD_DELEGATION_MAX_BUDGET_USD = 10`, `HARD_DELEGATION_MAX_DEPTH = 5`, defaults 5/3) and `kernelCore.ts`'s `?? DEFAULT_…` recording.
- **desktop/windows/src/renderer/src/lib/chatQuotaGate.test.ts** — the fake-timer case exercising the no-argument production path (`createChatQuotaGate()` → default `coldStartTimeoutMs = COLD_START_SYNC_TIMEOUT_MS` → `withTimeout(sync(), …)`) and the `5_000` pin both hold against `chatQuotaGate.ts`; the pin comment's retry-envelope reasoning matches `apiClient.ts` (`MAX_RETRIES = 5`, 429/503 backoff capped at 16s).

No further changes needed — ready for normal maintainer merge.

---
*by AI on behalf of David*

undivisible (approved):
Cline review agrees: test-only coverage for advertised delegation spend/depth ceilings and the cold-start quota timeout. Constants match production; CI Typecheck/Lint/Test green. Merge.

Issue #11695, opened 2026-08-16
Deepgram live STT: HTTP 402 kills ~19% of sessions (49/262 in 30 min) and there is no provider fallback


kodjima33:
Roughly one in five live Deepgram sessions in prod never opens a socket, and the session is then terminated with no provider fallback — the user gets no transcript at all.

Measured window 2026-08-16 16:30–17:00Z (prod `backend-listen`):

```
262  "process_audio_dg ..."                                   <- Deepgram sessions started
 49  "Deepgram start() returned False on all 3 attempts"      <- 18.7% gave up entirely
```

Rate is sustained, not a spike: `Deepgram start() returned False on all 3 attempts — giving up` runs 60–110/hour across 11:00–17:00Z today. The vendor rejection is explicit:

```
ERROR:deepgram.clients.common.v1.abstract_sync_websocket:WebSocketException in ListenWebSocketClient.start:
  server rejected WebSocket connection: HTTP 402
```

**HTTP 402 = Payment Required**, i.e. the key used for that request has no credit. It is not the managed key: I read `DEEPGRAM_API_KEY` from `prod-omi-backend-secrets` (identical to the Secret Manager version) and opened 10 concurrent `wss://api.deepgram.com/v1/listen` streams with it, `nova-3` / `en` and `multi` — all 10 connected. `DESKTOP_DEEPGRAM_API_KEY` connects too. Successes and 402s interleave on the same pod in the same second, so this is per-request, which points at the BYOK branch in `_deepgram_client_for_request()` (`backend/utils/stt/streaming.py:624`) serving users whose own Deepgram keys are out of credit.

Two things are worth fixing independently of whose key it is:

1. **No fallback.** `ListenReceiver._create_stt_socket` (`backend/routers/listen/receiver.py:159`) gives Parakeet a Modulate fallback via `connect_stt_socket_with_fallback`, but the Deepgram branch just returns whatever `process_audio_dg` gives it; `None` goes straight to `terminate_live_stt_session(..., reason='initialization_failed')`. Whether Deepgram *should* fall back to Velma is a product call — language coverage is the catch, since the failing sessions include `multi`, `zh-TW` and `sv` — but today there is no decision point at all.
[...]

Issue #11306, opened 2026-08-10
Parakeet-preferred live sessions repeatedly fall back to failing Modulate sessions


axAilotl:
With **Omi Parakeet** explicitly selected in the production iOS app and a stock CV1 on firmware 3.0.21, the client recorded 32 new live-transcription attempts during a 2026-08-09 dogfood window. Twenty-seven ended in provider-side failure: 26 terminal failures identified the provider as Modulate and only one identified Parakeet, even though every new client attempt logged `explicit_provider=parakeet`. Several attempts reached service-ready but produced no transcript, and the UI remained on “Reconnecting transcription.” Expected behavior is for an explicit Parakeet selection to stay on that provider or enter a bounded, observable fallback; the ready/failure contract should also report the provider actually selected and the fallback reason so the client does not treat a failing fallback as a healthy Parakeet session.

kodjima33:
Partial fix merged: #11359 (`1bd0bf3`).

That PR only repairs the **attribution** half of this report: `ListenReceiver.initialize_stt` snapshotted the provider name before `_create_stt_socket` could fall back Parakeet → Modulate, so post-fallback terminal failures (death monitor + init exception path) could label a Modulate session `parakeet`. The provider is now resolved at use time, so `stt_failed`/`record_live_stt_failure` name the provider actually serving.

Still open here: why an explicit `explicit_provider=parakeet` selection falls back so often, and surfacing the actual provider + fallback reason on the `ready` event so the client can tell a fallback session from a healthy Parakeet one.

axAilotl:
@Git-on-my-level — need human response. I ran a clean, fully in-range control and audited the current-main live-STT path. This separates the remote transcription failure from BLE recovery.

### Clean in-range control (factory app + stock CV1 firmware 3.0.21)

Window: **2026-08-11 19:41:35–21:41:35 EDT**.

- Zero BLE state transitions.
- Every sample reported the device connected, audio path ready, capture unpaused, and the BLE audio subscription present.
- The durable ring stayed effectively empty (4,884 bytes / 11 packets).
- The already-active Deepgram-labelled session failed at 20:20.
- The app then established 13 replacement sockets. Socket acquisition took 159–373 ms (201 ms median), and every attempt reached the server's `ready` state.
- Twelve consecutive replacement sessions then failed with the server-reported provider `modulate`.
- The final session connected at 20:43:03 but produced its first segment only **834.096 seconds later**, at 20:56:57.
[...]

PR #11822, opened 2026-08-18
fix(backend): name the serving provider on the live-STT ready event

RatnamOjha:
Closes the remaining backend slice of #11306.

## What was wrong

`ListenSessionRuntime.run` emitted a bare `ready` as soon as the provider socket opened:

```python
self.send_event(MessageServiceStatusEvent(status='ready'))
```

`_create_stt_socket` can walk the fallback chain (#11695, #11752), so by the time this
fires the session may be served by a provider other than the one the serving policy
selected. The client clears its terminal-failure state on `ready`, so "Listening" equally
described a healthy session and a fallback socket about to die, with nothing in the
payload to tell them apart.

`MessageServiceStatusEvent` already carries `provider` and `reason` — added as the
additive terminal-failure contract and unused on this path — and
`ListenReceiver._serving_provider()` already resolves the serving provider at use time.
This wires the two together.

@kodjima33 named this slice explicitly when merging the attribution half as #11359:

> Still open here: … surfacing the actual provider + fallback reason on the `ready` event
> so the client can tell a fallback session from a healthy Parakeet one.

## What changed

- `ready` now carries `provider`, read from `_serving_provider()` **at emission time**.
  Resolving it any earlier is precisely the attribution bug #11359 fixed on the
  terminal-failure path, so the accessor is called here rather than reusing a value held
  from bootstrap.
- `_bootstrap` records the policy-**selected** provider (`stt_service_selected`). That
  value is fixed at selection and is not a snapshot of the serving provider; it exists
  only so the emission can tell whether the chain moved the session.
- When serving ≠ selected, `reason` is `fallback_from_<selected>`. A healthy session
  carries no `reason` at all.
- Custom-STT sessions claim no backend provider — the client produces its own
  transcripts — so neither field is emitted.

Payload for a Modulate session that Velma refused and Deepgram took over:

```json
{"type": "service_status", "status": "ready", "provider": "deepgram", "reason": "fallback_from_modulate"}
```

## Scope

**Backend emission only.** Making the clients *act* on the new fields is a separate,
larger change across three platforms (Flutter app, iOS, desktop) and is deliberately not
included here. This PR only makes the information available to them.

The underlying vendor cause of a fallback (`quota` / `timeout` / `provider_5xx`) is
classified inside `connect_stt_socket_with_fallback` and already recorded through
`record_fallback`; it is not returned to the caller. Plumbing it onto the event would
change that shared helper's signature and its four call sites, so it is left out of this
slice. No new provider-changing or mode-changing branch is introduced here, so the
fallback-telemetry contract needs no new `record_fallback` call.

## Compatibility

Purely additive. `MessageServiceStatusEvent.to_json` uses `exclude_none=True`, so a client
that does not read the new fields sees exactly the payload it sees today — verified by
`test_ready_stays_additive_for_clients_that_ignore_the_new_fields`.

This is a WebSocket event and does not appear in the REST app-client contract
(`grep -c service_status docs/api-reference/app-client-openapi.json` → `0`). The gate was
run anyway:

```
$ python scripts/check_app_client_openapi_compatibility.py --base-ref upstream/main
App-client OpenAPI compatibility passed against merge-base 5712bfacef587c0fe47f48a35a2a14e934bc4cd5.
```

### Red before green

The regression test was written first and run against unmodified code. It drives the real
`run()` sequence with a real `ListenReceiver`, stubbing only the vendor connect functions,
so the fallback it reports is produced by the production chain rather than asserted about.

```
$ python -m pytest tests/unit/test_listen_ready_provider.py -q      # before the fix
___________ test_ready_names_the_fallback_provider_actually_serving ____________
>       assert payload['provider'] == 'deepgram'
E       KeyError: 'provider'
_____ test_ready_on_a_healthy_selected_provider_carries_no_fallback_reason _____
>       assert payload['provider'] == 'modulate'
E       KeyError: 'provider'
______ test_ready_provider_is_resolved_at_emission_time_not_at_selection _______
>       assert runtime._ready_event().to_json()['provider'] == 'parakeet'
E       AttributeError: 'ListenSessionRuntime' object has no attribute '_ready_event'
_____ test_bootstrap_records_the_selected_provider_for_fallback_comparison _____
>       assert runtime.stt_service_selected == STTService.modulate
E       AttributeError: 'ListenSessionRuntime' object has no attribute 'stt_service_selected'
4 failed, 2 passed, 9 warnings in 2.72s
```

`KeyError: 'provider'` is the exact pre-fix shape: the field was `None` and `exclude_none=True`
stripped it from the payload entirely. The first failure only reaches its assertion after
`assert runtime.stt_service == STTService.deepgram` passes, which proves the fallback chain
really moved the session before `ready` was emitted.

After the fix:

```
$ python -m pytest tests/unit/test_listen_ready_provider.py -q
6 passed, 9 warnings in 1.94s
```

### Blast radius

The full GitHub Actions unit contract for this diff — `scripts/select_backend_unit_tests.py`
selected 148 files from the changed paths — run through `test.sh` with CI's file-isolation
and timing guards:

```
$ BACKEND_UNIT_TEST_FILE_LIST=<selected> BACKEND_FAST_UNIT_FAIL_SECONDS=1.0 \
  BACKEND_PYTEST_FILE_ISOLATION=1 bash test.sh
143 files, 2473 tests passed, exit 0
```

Type check and the isolation scanners:

```
$ bash scripts/typecheck.sh
0 errors, 3660 warnings, 0 informations

$ python scripts/check_module_stub_pollution.py
Checked 903 backend test file(s); 0 violation(s).

$ python scripts/scan_import_time_side_effects.py
Checked 800 backend production file(s); 0 violation(s).

$ black --line-length 120 --skip-string-normalization --check
2 files would be left unchanged.
```

### Not verified here

Four of the 148 selected files could not run on this machine, for reasons unrelated to
this diff:

- `test_verify_pusher_config_references.py` shells out to `helm`, which is not installed
  (`FileNotFoundError: [Errno 2] No such file or directory: 'helm'`). Excluded from the
  2473-test run above; CI has helm.
- `test-preflight.sh` reports one failure: Python 3.11.13 installed vs 3.11.15 pinned in
  `.python-version`. Patch-level, environmental.

**This was verified hermetically at the `process_audio_*` seam. It was not exercised
against live Deepgram, Modulate, or Parakeet**, and I have not run a real device session
against it. The provider values asserted are the ones the production fallback chain
assigns to `host.stt_service`; that those match what a live vendor socket serves is
existing behavior this PR does not change.

## Gates

- **Product invariants affected:** none (`scripts/pr-preflight --base upstream/main --suggest`).
- **Docs:** `docs/doc/developer/backend/transcription.mdx` and
  `listen_pusher_pipeline.mdx` both documented the `ready` payload and are updated.

Failure-Class: none

Rationale: this adds reporting fields to an existing event. It introduces no new
failure, recovery, or fallback branch — the fallback behavior it reports on landed in
#11695/#11752/#11814. #11359, the attribution half of this same issue on the same code
path, also declared `Failure-Class: none`. `scripts/failure-class prepare` inferred no
class from the diff.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11822?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/listen/runtime.py
+++ b/backend/routers/listen/runtime.py
@@ -115,6 +115,7 @@ class ListenSessionRuntime:
         self.is_multi_channel = request.channels >= 2
         self.language = request.language
         self.stt_service: Any = None
+        self.stt_service_selected: Any = None
         self.stt_language = ''
         self.stt_model = ''
         self.vocabulary: List[str] = []
@@ -275,6 +276,10 @@ class ListenSessionRuntime:
             multi_lang_enabled=not single_language_mode,
             preferred_service=request.stt_service,
         )
+        # The provider the serving policy chose, captured before `_create_stt_socket`
+        # can walk the fallback chain. Only the *selected* value is safe to hold onto:
+        # the serving one has to be read at use time (#11306).
+        self.stt_service_selected = self.stt_service
         self.parity_capture = ListenParityCapture.from_environ(
             principal_id=request.uid,
             session_id=getattr(self, 'session_id', ''),
@@ -574,6 +579,32 @@ class ListenSessionRuntime:
                 self.task_supervisor.create_lifetime_task(session.audio_bytes_consume(), name='pusher_audio')
             )
 
+    def _ready_event(self) -> MessageServiceStatusEvent:
+        """Name the provider actually serving this session on the `ready` event (#11306).
+
+        The client clears its terminal-failure state on `ready`, so a bare event leaves a
+        fallback socket that is about to die indistinguishable from a healthy session on
+        the provider the user selected. `_create_stt_socket` can walk the fallback chain
+        (#11695, #11752), so the serving provider is only knowable once the socket exists
+        — resolving it any earlier is the attribution bug #11359 fixed on the
+        terminal-failure path, which is why this reads `_serving_provider()` here rather
+        than reusing a value from bootstrap.
+
+        Both fields are optional and dropped by `exclude_none=True`, so a client that does
+        not read them sees exactly the payload it sees today.
+        """
+        if self.use_custom_stt:
+            # Custom-STT clients produce their own transcripts; no backend provider serves.
+            return MessageServiceStatusEvent(status='ready')
+        serving = self.receiver._serving_provider()
+        selected = getattr(self.stt_service_selected, 'value', self.stt_service_selected)
+        fell_back = bool(serving) and bool(selected) and serving != selected
+        return MessageServiceStatusEvent(
+            status='ready',
+            provider=serving,
+            reason=f'fallback_from_{selected}' if fell_back else None,
+        )
+
     async def run(self) -> None:
         if not await self._admit() or not await self._bootstrap():
             return
@@ -628,7 +659,7 @@ class ListenSessionRuntime:
                         self.task_supervisor.create_finite_task(self.speakers.load_and_run(), name='speaker_id'),
                     ]
                 )
-            self.send_event(MessageServiceStatusEvent(status='ready'))
+            self.send_event(self._ready_event())
             result = await self.task_supervisor.supervise(receive_task=receive_task)
             logger.info('Listen supervisor exited reason=%s', result.reason)
             if result.reason in {'crash', 'lifetime_done'}:
--- a/backend/tests/unit/test_listen_ready_provider.py
+++ b/backend/tests/unit/test_listen_ready_provider.py
@@ -1,0 +1,284 @@
+"""Regression (#11306): the live-STT `ready` event must name the provider serving.
+
+The backend emitted a bare ``MessageServiceStatusEvent(status='ready')`` as soon as
+the provider socket opened. ``_create_stt_socket`` can walk the fallback chain
+(#11695, #11752), so "Listening" could equally mean the Parakeet session the user
+selected or a Modulate/Deepgram fallback socket that is about to die — the client
+clears its terminal-failure state on ``ready`` and cannot tell the two apart.
+
+These tests drive the real ``run()`` sequence with the vendor connect functions
+stubbed, so the fallback the event has to report is produced by the production
+chain rather than asserted about.
+"""
+
+import asyncio
+import os
+from contextlib import contextmanager
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from routers.listen import receiver as receiver_mod
+from routers.listen.contracts import ListenRequest, ListenSessionState
+from routers.listen.receiver import ListenReceiver
+from routers.listen.runtime import ListenSessionRuntime
+from utils.stt import provider_resilience, streaming
+from utils.stt.streaming import STTService
+
+
+@pytest.fixture
+def anyio_backend():
+    return 'asyncio'
+
+
+@contextmanager
+def _serving_policy(models=('modulate-velma-2', 'dg-nova-3', 'parakeet')):
+    """Drive the real provider gates off ``STT_SERVICE_MODELS``, not a stubbed answer."""
+    with (
+        patch.object(streaming, 'stt_service_models', list(models)),
+        patch.object(streaming, '_deepgram_is_available', return_value=True),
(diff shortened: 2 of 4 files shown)


Git-on-my-level (commented):
Verified against the diff and against the surrounding production code, and ran the tests locally: the new `test_listen_ready_provider.py` plus the surrounding listen suites (`test_listen_pipeline.py`, `test_listen_runtime_regressions.py`, `test_listen_stt_death_monitor.py`, `test_listen_fallback_removal.py`, `test_listen_parity_runtime.py`, `test_listen_session_bootstrap.py`) — 163 passed.

Per file:

- `backend/routers/listen/runtime.py` — reading `_serving_provider()` inside `_ready_event()` at emission time is the right call: by then `initialize_stt()` has run `_create_stt_socket`, and every fallback branch in `receiver.py` assigns `host.stt_service = actual_service`, so the value reflects the socket actually serving. Snapshotting `stt_service_selected` in `_bootstrap` before the chain can walk is exactly the comparison value needed, and the custom-STT early return correctly keeps those sessions claiming no backend provider. One pre-existing edge worth naming (not changed here, same attribution shape the terminal-failure path already has): on multi-channel sessions `host.stt_service` reflects the last channel's fallback outcome, so a mixed-provider session would report the last channel's provider.
- `backend/tests/unit/test_listen_ready_provider.py` — the strong part. `_run_session` drives the real `run()` sequence with only the `process_audio_*` connect functions stubbed, so the fallback in the assertions is produced by the production chain (`connect_stt_socket_with_fallback` plus the liveness grace) rather than asserted into existence, and `_RejectedSocket` models Velma's real over-quota shape. The additive-payload pin on `exclude_none` protects legacy clients.
- `docs/doc/developer/backend/listen_pusher_pipeline.mdx` — the sequence diagram and event table now match the emitted payload, including the custom-STT absence note; consistent with `_ready_event()`.
- `docs/doc/developer/backend/transcription.mdx` — the example (`provider: "deepgram"`, `reason: "fallback_from_modulate"`) corresponds to a real reachable chain state (Modulate primary → over-quota → Deepgram, #11752), and dropping `status_text` actually fixes an old inaccuracy: this path never emitted it.

Strictly additive — no routing/fallback logic changes, legacy payloads unchanged via `exclude_none=True`, and the `Failure-Class: none` declaration is consistent with the diff introducing no new failure branch. Since this extends the client-facing contract on the STT provider-serving surface (Flutter/iOS/desktop consumption to follow), leaving the formal approval to a human maintainer; everything verifiable here checks out.

---

RatnamOjha:
Lgtm @kodjima33, merging since the backend half of #11306, the slice you named when merging
#11359: surfacing the serving provider + fallback reason on `ready` so a client can
tell a fallback session from a healthy one.

Why it's safe to land now:

- **Strictly additive.** `exclude_none=True` means clients that ignore the new fields
  see exactly today's payload. No routing, fallback, or failure-path logic changes.
- **WebSocket event, not the REST contract.** App-client OpenAPI compat gate passes;
  `service_status` has 0 hits in the spec.
- **Red before green.** The regression test drives the real `run()` with only
  `process_audio_*` stubbed, so the fallback it asserts on is produced by the
  production chain. Pre-fix: `KeyError: 'provider'` (field was `None`, stripped by
  `exclude_none`). Post-fix: 6 passed.
- CI green (15/15 non-skipped), `pr-preflight` 24/24, pyright 0 errors.
- Reviewed by @Git-on-my-level against the diff and surrounding production code,
  163 tests locally, no changes requested.

Scope is backend emission only — Flutter/iOS/desktop consumption is a separate change
across three platforms. The multi-channel edge raised in review (`host.stt_service`
reflects the last channel's outcome) is pre-existing and unchanged here: it's the same
attribution shape the terminal-failure path has had since #11359.

PR #11827, opened 2026-08-18
fix(desktop): serve the legacy graph when this account has no canonical state

kodjima33:
## Bug

The desktop **Memory Atlas fails to load** for every account whose knowledge graph still lives in the legacy store. Live in prod right now on the current image (`c5b0b5d`): `GET /v1/knowledge-graph/canonical?limit=200` → **503 `canonical_graph_unavailable`**, ~11 per 90 min from macOS clients (`Omi/12187 CFNetwork/… Darwin/25`), in bursts of three — the page's retry loop giving up.

## Root cause

Production fences canonical intake (`MEMORY_MODE`), so `users/{uid}/memory_state/head` is never written and the canonical read is *permanently* unavailable for those accounts. The backend already handles that: `fbb005c15b` (#11623, 2026-08-15) made `GET /v1/knowledge-graph` serve the legacy graph instead of failing, and deliberately kept `/v1/knowledge-graph/canonical` answering 503 — "the caller asking for canonical specifically" (`backend/tests/unit/test_knowledge_graph_legacy_fallback.py::test_the_explicit_canonical_route_still_reports_unavailable`).

The desktop client never takes that path. `getKnowledgeGraphImpl` falls back to the legacy route **only on 404/405** ("endpoint not deployed", `b13562c123`). A 503 propagates, and `MemoryGraphViewModel` publishes an empty atlas. So the server-side fix landed on the one route the desktop app does not call.

## Fix

Treat `503 canonical_graph_unavailable` as the same "this account has no canonical store" signal as 404/405 and take the existing legacy path, with the existing `recordFallback` telemetry (`reason=canonical_state_unavailable_503`). Every **other** 503 still propagates — an outage is not a statement about the account, and hiding it would publish a stale graph as current.

## Tested

`./scripts/dev-feedback.py --once swift 'KnowledgeGraphPaginationTests'` — 11 tests executed. Both new tests pass. One pre-existing local-only failure (`testCatalogOnlyCanonicalAtlasIsRenderableAndActivationCached`) reproduces identically with this commit reverted — it comes from `MemoryAtlasOwnerIdentity` minting the synthetic `atlas-owner` anchor when the test host has a profile name, which CI does not; `Desktop Swift CI` is green on `main`.

Two regression tests through the real production seam (`URLProtocol` stub → real `URLSession` → real `OmiHTTPTransport` error mapping → `getKnowledgeGraph()`), with the exact wire body prod returns:
- `testCanonicalGraphUnavailable503FallsBackToLegacyGraph` — 503 `canonical_graph_unavailable` then legacy 200; asserts the graph renders and that both requests were made in order. **Fails on clean `main`** (control run with only this source hunk reverted: `caught error: "httpError(statusCode: 503, detail: Optional("canonical_graph_unavailable"))"`).
- `testTransient503DoesNotFallBackToLegacyGraph` — 503 `temporarily unavailable` still throws, exactly one request. Guards the invariant behind `testCanonicalAtlas503DoesNotPublishNonemptyLocalProjection`.

Failure-Class: FC-derived-state-gate-without-legacy-source-path

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11827?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Services/APIClient/APIClient+KnowledgeGraph.swift
+++ b/desktop/macos/Desktop/Sources/Services/APIClient/APIClient+KnowledgeGraph.swift
@@ -325,13 +325,20 @@ struct RebuildGraphResponse: Codable {
 
 // MARK: - Knowledge Graph API
 
+/// The backend's answer when this account has no canonical graph state at all.
+/// Production fences canonical intake, so `/v1/knowledge-graph/canonical` gives
+/// this to accounts whose graph only exists in the legacy store — a statement
+/// about the account, not a transient outage. Any other 503 stays an outage.
+private let canonicalGraphUnavailableDetail = "canonical_graph_unavailable"
+
 extension APIClient {
 
   /// Gets the complete canonical graph, exhausting the bounded page endpoint.
   ///
   /// The endpoint was added after the legacy graph route, so a 404/405 on the
-  /// canonical route is the only condition that selects the legacy request.
-  /// Every other error is returned to the caller.
+  /// canonical route selects the legacy request, as does a 503 that names this
+  /// account as having no canonical state. Every other error is returned to the
+  /// caller.
   func getKnowledgeGraphImpl(
     expectedOwnerId: String? = nil,
     authorizationSnapshot: RuntimeOwnerAuthorizationSnapshot? = nil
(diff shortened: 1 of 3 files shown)


kodjima33:
CI note before merge: `Desktop Swift Release Compile` has been QUEUED for 2h+ behind a repo-wide `macos-15` runner backlog (Desktop Swift CI runs still queued from Aug 14). It is not a required status check — `main` protection requires only 1 review, no required contexts.

Merging on the evidence that did run:
- **Desktop Swift Static & Test Contracts** (macos-15, pinned Xcode 16.4) SUCCESS — manifest checks, launcher tests, full debug suite. This is the PR critical path by design; the workflow itself notes release compile is the separate main-push lane.
- **Desktop Swift Build & Tests** SUCCESS, **Backend Hermetic Merge Gate** SUCCESS, **Hygiene** SUCCESS, **PR Metadata Preflight** SUCCESS.
- Ran `run-swift-ci.sh --release-compile`'s command locally (`swift build -c release`). Local toolchain is Xcode 26.4/Swift 6.3 rather than the pinned 16.4, so it is not a substitute for the CI job, but the `Omi_Computer` module — which contains the changed `APIClient+KnowledgeGraph.swift` — type-checked with the changed file clean. The only 2 errors are `SendingRisksDataRace` in `ChatProvider+JournalProjection.swift`, a file this PR does not touch (`git diff origin/main` on it is empty) and a diagnostic Swift 6.0 does not emit.

PR #11834, opened 2026-08-19
feat(conversations): add coherent cached notes pipeline

Git-on-my-level:
## Summary

- add a flag-gated canonical note generator that produces free-form sourced sections, action items, and events in one coherent call
- reuse one per-conversation transcript/cache prefix across note generation, L1 memory extraction, and explicitly selected summarization apps
- feed calendar and bounded conferencing-screen OCR identity context before summarization, and derive vector metadata from the canonical note instead of rereading the raw transcript
- make automatic summarization apps an independently gated opt-in behavior while preserving explicit/preferred apps

All four user-visible rollout controls default off in the runtime manifest and GKE values:

- `CONVERSATION_NOTES_V2_ENABLED`
- `CONVERSATION_APPS_OPT_IN_ONLY`
- `CONVERSATION_CALENDAR_CONTEXT_READ_ENABLED`
- `CONVERSATION_OCR_CONTEXT_ENABLED`

## Design review and changes from the brief

- The Redis meeting mapping did not fail for the motivating ad-hoc desktop call; no mapping was ever created. The actual starvation boundary was broader: direct `CreateConversation.calendar_meeting_context` was ignored by `_get_structured`, and overlapping Google Calendar lookup happened only after summarization behind a flag that also writes back to Google Calendar. This change consumes direct context, preserves the legacy Redis mapping, and adds a separate read-only pre-summary overlap flag while leaving calendar mutation under `GOOGLE_CALENDAR_AUTO_LINK_ENABLED`.
- OCR extraction is deterministic and bounded (80 rows, 12,000 characters, conferencing apps only, at most 12 names). A nano call would add another provider dependency/pass for metadata that is already structurally simple.
- OCR/calendar names do not create durable `person_id` values because a display name is not an authoritative Omi Person identity. The prompt renderer relabels a single unresolved speaker only when exactly one calendar/OCR participant remains unassigned; otherwise it preserves the transcript labels.
- The additive schema lives in the in-repo plugin SDK as well as the backend fallback model because that SDK is the backend model authority. Existing documents deserialize with `sections=[]` and null action-detail fields.
- No client release is needed: the server projects `sections` into markdown in `structured.overview`, which both existing clients already use when no app result exists.
- Reprocess and external-audio paths use the same merged note call. Under the app opt-in flag, empty app results are persisted too, so an old default app result cannot continue shadowing the canonical note after reprocess.

## Prompt-cache verification and arithmetic

The configured gateway routes `conv_structure` and `memory_l1` to the same OpenAI GPT-5.6 Luna lane and preserves the cache fields for compatible OpenAI routes; incompatible/fallback routes strip them. OpenAI's prompt-cache contract requires an exact shared prefix, starts at 1,024 tokens for GPT-5.6, uses `prompt_cache_key` to improve routing, and supports the explicit 30-minute retention used here. The canonical prefix is two byte-identical system messages, with the explicit breakpoint after metadata plus the full transcript; feature instructions follow it. Tests rebuild the note and memory prefixes independently and assert equality.

Provider reference: https://developers.openai.com/api/docs/guides/prompt-caching

For the motivating approximately 6,300-token transcript:

- before: 5 uncached transcript passes = approximately 31,500 input tokens
- after (default path): note cache write + memory cache read = `6,300 * 1.25 + 6,300 * 0.10 = 8,505` input-token cost equivalent, plus short task suffixes (approximately 73% lower transcript-input cost)
- even on a cache miss: 2 passes = 12,600 raw transcript tokens (60% fewer)

This gives up roughly 1,000 tokens of feature-specific instruction-prefix reuse to cache the approximately 6,300-token transcript and its metadata. Explicitly chosen apps can read the same warm prefix; they are no longer part of the default pass count.

## Note and action contract

`Structured.sections` contains typed section containers with model-authored markdown and exact source segment IDs. The prompt supplies three inspirational shapes rather than a rigid taxonomy, scales density with transcript size, preserves names/numbers/products/organizations verbatim, permits short quotations, and prohibits guessed spellings.

Action items retain the <=15-word timeless description and add `owner_name`, one-line `context`, and `due_certainty`. The merged call treats tentative dates as tentative action metadata, never as confirmed events, and avoids inventing a precise hour when none was stated.

## Product callout: summarization apps

`CONVERSATION_APPS_OPT_IN_ONLY=true` demotes the summarization-app ecosystem from "one app runs by default for everyone" to "an app runs only when explicitly requested or selected as the user's preferred app." This is intentionally a separate, one-line rollback flag from notes v2.

## Verification

- `125 passed`: notes/schema/cache wire/process/date/model focused suite
- `106 passed`: exact `test.sh` rerun of every change-related file that failed during the broad run, including gateway strict-schema and upstream-boundary tests
- `198 passed`: `backend/scripts/pre-deploy-check.sh` (dev/prod runtime manifest, rendered deployment contracts, deploy safety tests)
- PR preflight passed through line-count, docs, diff hygiene, identity, architecture, invariants, failure-class, deployment classification, public-build, async-blocker, and datetime gates; the next gate could not install `tiktoken` because package-index DNS is blocked in this environment
- formatting, `py_compile`, and `git diff --check`: passed

The complete `backend/test.sh` matrix was run under the locally available Python 3.12 environment. Its first pass exposed eight change-related failures, all fixed and green in the exact failed-file runner. Two unrelated replay-harness files still cannot bind loopback UDP/TCP sockets in this managed sandbox (`PermissionError: [Errno 1] Operation not permitted`), so the full matrix cannot be claimed green here. The repository-pinned Python 3.11.15 is not installed and dependency sync cannot reach the package index from this environment.

## Follow-up commits in this PR

**Skimmable notes, no default due dates.** Note bodies are bullets with nested sub-bullets and no
narration; every bullet must carry a concrete specific or be deleted. Coverage beats brevity — the
word target is met by tightening wording, never by dropping a topic. Action items no longer carry a
due date by default: a date merely discussed is not a deadline, it goes in `context`.

**Meeting identity by time overlap.** Root cause of the motivating defect: processing consulted only
a Redis conversation→meeting mapping written at creation time, and `users/{uid}/meetings` was always
empty because no client wrote it. Now resolves by time overlap with a layered fallback
(calendar-backed stored → client-supplied → Google read-only → screen-derived → server OCR),
degrading silently at every layer. Calendar writes stay behind `GOOGLE_CALENDAR_AUTO_LINK_ENABLED`.

**OCR identity by corroboration.** The first extractor accepted any 1–4 word capitalised line, which
on a full-desktop capture means the browser tab strip becomes the participant list. Verified against
real captured data it emitted eleven phantom participants. Now a name is accepted only from a roster
sentence ("X and Y are in this call", "Meet with X") or when corroborated by an email local part in
the same window. Precision-first: no participants → no context at all.

**EventKit.** macOS never populated `users/{uid}/meetings` — the app had zero EventKit usage and the
generated client method had no callers. Adds a bounded reader that uploads meeting identity without
requiring Google OAuth. Ships dark.

**On-device derivation behind an explicit egress policy.** The previous "derived data never leaves
the device" rule was drawn for proactive notifications, which are computed and consumed on-device;
conversation summaries are generated server-side and legitimately need identity. `DerivedDataEgressPolicy`
now declares every derived-data class default-deny with each allow bound to a stated purpose, and the
CI tripwire asserts bucket tables stay out of AgentSync *and* that no denied class has a route. The
extractor moved client-side so it reads the local store rather than starving on screen-activity rows
that never sync (measured: 0.8% of 26k local rows reach the backend). Exactly seven wire keys leave
the device, asserted by test; no OCR text, screenshots, bucket data, app names, or window titles.

### Dogfood measurement

Run against a real 5,271-word meeting transcript through the production note path on the production
model route, scored on presence of specific facts verified to exist in the transcript:

| | production today | reference tool | this PR, no metadata | this PR, derived metadata |
|---|---|---|---|---|
| specifics captured | 7/19 | 15/19 | 17/19 | **19/19** |
| words | 315 | 424 | 539 | 688 |

A second 3,684-word meeting scored 11/17 (production) vs 17/17 (this PR), with 7 owner-attributed
action items vs 3 unattributed. Participant metadata for the first meeting was derived by the real
extractor from real captured screen data, not hand-supplied.

## Reviewer re-verification

Re-ran on the repository-pinned interpreter (`backend/scripts/sync-python-deps.sh` created the 3.11 `.venv` successfully; the authoring environment lacked package-index access):

- `test_conversation_notes_v2.py` 5 passed, `test_action_item_date_validation.py` 26, `test_conversation_structure_timezone.py` 16, `test_llm_gateway_chat_extraction_pilot.py` 23, `test_llm_gateway_openai_compatible.py` 31, `test_process_conversation_usage_context.py` 33, `test_working_observations_extractor.py` 15 — **149 passed**, no failures.
- Collecting all of `tests/unit` in a single process reports 275 collection errors (prometheus `Duplicated timeseries in CollectorRegistry` on shared import, plus a protobuf native-extension segfault on macOS arm64 unless `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python` is set). Verified identical on the parent commit `3daa6ac255` with these changes stashed — same 275 errors, 6,913 tests collected — so it is a pre-existing environment constraint, not a regression. Per-file runs are unaffected.

## Could not do

- Live re-verification/dogfooding of conversation `ff8b9998-1169-5c66-9998-8af7837045d9`: the required local Omi read-only GCP profile is absent, and no live LLM credentials are available. The implementation therefore relies on the supplied verified keyword/metadata evidence plus logic tests and default-off flags.
- A real provider cache-hit measurement was not possible without live credentials; the wire contract and exact-prefix construction are covered by tests.

none

Failure-Class: new

New class: **derived context starved by an unwritten association.** Enrichment depended on a single
conversation→meeting mapping that only one code path wrote, and no shipping client ever called it, so
the dependent feature degraded silently to no context rather than failing visibly. Guard surface: the
resolver now matches by time overlap across layered sources and every layer's absence is covered by a
test, so a missing writer cannot silently disable enrichment again.

Line-Count-Exception: backend/utils/conversations/process_conversation.py | 1859 -> 2204 | one feature-gated lifecycle change must keep creation, reprocess, app persistence, metadata extraction, and canonical-memory prefix construction under the existing conversation orchestration owner; reusable OCR and prefix logic is already split into new focused modules

Line-Count-Exception: backend/utils/llm/conversation_processing.py | 1510 -> 1672 | the merged note generator belongs beside the legacy conversation generation paths it replaces, while shared prefix construction is already split out; separating only this call would create another orchestration adapter without an independent owner


kodjima33 (approved):
Flag-gated canonical notes pipeline — approving; feature of this size plus deploy-config changes needs a maintainer merge.

PR #11835, opened 2026-08-19
Cut the Windows app's idle and focus-driven backend request volume

ZachL111:
A systematic sweep of what the Windows app asks the backend for when nobody is asking it to, following the billing RCA that `index.ts:997` already records. Seven changes, each with the mutation that proves its test can fail.

The single largest item is first, and it is worth reading on its own: a full memory page-through, every 30 minutes, on every default install, whose entire result is thrown away.

## What was found

I inventoried every recurring timer, every mount-time fetch, and every focus listener in `desktop/windows`, then traced each candidate to the endpoint it reaches and the work that endpoint does. Below is what is fixed here. Findings I did NOT act on are listed at the end, including three that turned out to be deliberate and one backend change that is worth more than anything in this PR but does not belong in it.

### 1. The retention sweep runs on a timer for every default install, and discards the result

`retentionMode` is optional, and the sweep falls back to `dry-run`. Only `off` returned early. So on default settings, every 30 minutes forever:

- `fetchAllMemories()` pages `/v3/memories` to exhaustion
- `GET /v1/conversations?limit=200`

and then, in dry-run, hands all of it to a `console.log` and returns. Every request was paid to write a line into a DevTools console that is not open, 48 times a day, for every user.

Scheduled passes now require `live`, the only mode that deletes anything. `dry-run` becomes what its own label already claims: pressing Preview in Settings runs one pass, so the plan is logged when the user asks for it.

This is the item where the backend cost is concentrated. Every page after the first takes the `offset>0` path in `memories.py`, which streams the whole `memory_items` collection and slices in Python.

### 2. The staged-task promote timer POSTs about 1,440 times a day to be told "nothing"

`promotionService` ran `POST /v1/staged-tasks/promote` on a flat 60s timer for the life of the process, with `bypassDebounce: true`, no local pre-check, and no backoff of any kind. Nearly every call returns `promoted:false`, and each one costs the backend two full collection scans to say so. A 500ing endpoint got the same flat 60s forever.

The net exists for staged rows created out of band. Every in-app path that stages a task already promotes inline or on the task-complete and task-delete events, none of which go through this timer, so its cadence is discovery latency for a rare case rather than responsiveness for a common one.

Ticks now walk 60s, 2m, 5m, 10m and hold. Any tick that actually promotes drops back to the 60s floor, so a backlog still drains one per minute exactly as before. Signed-out ticks make no request and do not walk the ladder.

### 3. Six-plus requests per window focus, for surfaces that are not on screen

`MainViews` mounts every panel 1.8s after launch and never unmounts them, so a `window` focus listener registered by any page is live app-wide regardless of the current tab. Three pages register one, none route-gated and none throttled: Memories (a full `/v3/memories` page-through), Apps (three requests), and the Home goal chips (one).

The trigger is not rare. The app blurs its own main window every time the floating bar, the orb, or the capture window takes focus, so ordinary voice use pays this repeatedly per minute.

Now throttled to once a minute per surface. The first focus always runs, so the behavior these surfaces promise is intact and the two existing page-side guards pass unchanged.

### 4. The chat quota is re-fetched on every bar reveal and every reply

`sync()` had an in-flight dedupe and nothing else, which does nothing for sequential calls. It runs from `BarApp`'s `onShow` and from `UsageLimitTriggerHost` after every completed reply, both driven by the primary voice interaction.

The per-reply probe doubles the request count of every chat turn: the "at most once per session" latch it relies on only sets when the quota is already exhausted, so for an under-cap user it never sets.

Refreshes are now floored at 60s. The gate already tracks sends between refreshes with `recordQuery()`'s optimistic delta, so the verdict stays correct across the gap, and the first probe of a session is never throttled because `check()` depends on it.

This endpoint is cheap per call, unlike the others here. It is in the PR for call volume, not read cost.

### 5. The about-user card is rebuilt on every hub warm

`refreshAboutUserCard()` deduped only against a build already in flight. The hub deliberately re-warms all session long, and those warms are minutes apart, so the dedupe never applied. Each rebuild is a `GET /v3/memories`, several hundred times a day on an idle machine, to produce a card that had not changed. Now reused for five minutes.

The warm loop itself is untouched. See the deliberate-behavior note below.

### 6. The goal cap is checked after the expensive read, not before

`runGoalGenerationIfDue` fetched five parallel reads including `/v3/memories?limit=500` and `/v1/conversations?limit=100`, and only then checked the three-active-goal cap. Sitting at the cap is the ordinary resting state, and that bail deliberately does not stamp the day, so the 4-hourly tick repeats indefinitely, re-paying all five reads each time.

The count now comes from one `/v1/goals/all` ahead of the bundle, and that list is handed on to the context build so a run that proceeds reuses it rather than asking for the same list twice. Bail costs one read, proceed costs the same five it always did. The file's header already described the intended order as "cheap ones first, then the heavy context fetch"; the code now matches it.

Default-off, so this is latent rather than live, but it is a real wedge for anyone who turns it on.

### 7. Two panels fetch for surfaces the user is not looking at

`ChatAppPicker` called `useChatApps()` above its own gate, issuing `GET /v1/apps` and then returning null because the pref is off for almost everyone. Split into gate plus inner, exactly as the sibling `HubChatHeader` already does for the same stated reason.

`Tasks` fetched `GET /v1/conversations?limit=200` on mount for source-link labels, at every launch, whether or not Tasks was ever opened, and wrote a cache it never read. Now waits for the panel to be the active route and short-circuits on the cache: one request on first open, none after, against one per launch before.

## Verification

Run in `desktop/windows`:

- `npx vitest run` - 5,267 passed, 18 skipped, 544 files.
- `npm run typecheck` - clean.
- `npx eslint .` - 0 errors. The 735 warnings are pre-existing prettier formatting in files this PR does not touch; the changed files add none.
- `npm run build` - builds clean, bytecode patch and verify both OK.

**29 mutations seeded across the seven changes. Every one turns the suite red now. Every one left it green before.** They are listed per-commit with what each represents.

Three of those rounds found something rather than confirming something:

- A focus-throttle case of mine could not fail: it re-rendered with the same handler reference, so the effect never resubscribed. Rewritten to pass a fresh closure, as the real call sites do.
- `MIN_SYNC_INTERVAL_MS` escaped, because every quota case passed an explicit interval while production `barSend.ts` constructs the gate with no arguments and runs entirely on the default. Added a case that does the same.
- Clearing the timestamp in `resetAboutUserCard` was unobservable, since nulling the card already defeats the guard. Dead line removed.

Self-review also caught a defect in change 2 before it shipped: a tick re-arms from inside its own promise chain, so a stop landing mid-promote resurrected the timer `stopTaskPromotionService` had just cleared, which `clearInterval` had made impossible for free. Guarded, with a case.

### Two defects of my own, fixed in a follow-up commit

Both found by re-reading this diff and asking what each change costs on the paths it did NOT target, rather than by a test failing:

- The goal cap check moved in front of the context build, but the bundle still fetched `/v1/goals/all` itself, so a run that generated asked for the same list twice: five reads became six. The bail path is the common one and got far cheaper, so the net was still a large win, but a PR about removing redundant requests should not add one.
- Gating the Tasks title fetch on the active route folded the route into the deps of the effect that reads local tasks, so `readTasks` re-ran on every navigation in and out of the tab instead of once per mount. Local SQLite over IPC, so no backend cost and no spinner, but unintended and not what the commit claimed.

The first fix needed a second round of mutation work: the schedule test proved the goal list was PASSED to the context build, but that build is mocked there, so nothing proved it was USED. `withPrefetchedGoals` exists to make that substitution testable without a live session.

### What is not verified

I did not exercise these against a live signed-in account. Signed out the app makes no backend requests at all, so a signed-out run cannot observe any of these paths, and I was not willing to generate real load on a real account to measure. Everything above is traced from the mount, route, and timer paths and covered by hermetic tests.

The two changes in commit 7 are the ones where that matters most, since neither has a test observing the request from the page side. The visible thing to confirm is that source-link titles still render on the first visit to Tasks.

## One existing test changed

`promotionService.test.ts` pinned the flat cadence ("5 minutes buys 5 ticks"). It now pins the ladder. The guard it was written for, that the session poll stops rather than running 60 five-second attempts, is unchanged and still asserted. The new expected values come from the ladder constants introduced in the same commit, and four separate mutations hold that ladder in place.

## Deliberate behavior I did not touch

Worth stating, since two of these look like bugs and are not:

- **The hub re-warms for the whole session.** `hubController` resets its strike budget on any socket that lived past the idle window, and the provider idle-closes at roughly two minutes, so an idle user's hub rebuilds indefinitely. The code documents this at length as the mechanism that keeps the next press on the warm lane. That is a latency-versus-cost product decision, not spam, so I left it and fixed only the memory fetch it was dragging along (change 5). If the tradeoff is worth revisiting, the lever is an idle release below the provider's close, and that is a product call.
- **The promote net is not gated on `taskEnabled`.** Deliberate Mac parity, so backlog staged before the toggle was turned off still drains. Unchanged.
- **The auto-generation bail does not stamp the day.** Deliberate, so finishing a goal at noon can still generate a replacement. Left alone; only its cost changed.

## Follow-ups, deliberately not in this PR

Each is real and located, and each is a different review surface:

- **`GET /v1/action-items/ids` is the biggest remaining item, and the fix is server-side.** `taskSyncEngine` runs an ID census every 5 minutes, two requests, and `database/action_items.py:111` streams the collection with no `.where()` and filters `completed` in Python, so each call reads every doc and the pair scans the collection twice. That is 24 full collection scans per idle hour per running client. Pushing `.where('completed','==',completed)` server-side roughly halves it immediately; a change token or `updated_since` would make a steady-state round nearly free. Worth more than anything in this PR, and it is a backend change with its own correctness questions around soft-deleted and `status` rows, so it should not ride along here. Also `get_visible_action_item_ids` never calls `record_firestore_read`, so this path is invisible in the read metrics.
- **`DELETE /v3/memories/{id}` costs a full canonical scan per id**, and Windows deletes serially in a loop. `DELETE /v3/memories/batch` exists and pays the scan once per batch. Windows already uses the batch POST but not the batch DELETE.
- **The launch burst.** Settings mounts all eleven tabs at once, so Plan and Usage issues four billing requests at every launch for a screen the user has not opened. Fixing it properly means teaching the hydration pass to warm the render without warming the data, which is a bigger structural change than anything here.
- **`preferences.retention.test.ts` cannot fail.** Its default assertion is `getPreferences().retentionMode ?? 'dry-run'`, which is the fallback the test itself supplies. Left in place because the correct fix is to assert against the production default, and that default is exactly what change 1 is about.

- INV-CHAT-1

From `scripts/pr-preflight --suggest`, matched on the `chatQuotaGate` change. The gate's verdict is unchanged: the cold-start probe that blocks an over-cap user is explicitly exempt from the new floor, and the optimistic delta still counts sends between refreshes, both of which are asserted. Only the refresh cadence moved.

`scripts/pr-preflight` also reports no `fix:` commits, so no `Failure-Class` declaration is required.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/assistants/goals/context.test.ts
+++ b/desktop/windows/src/main/assistants/goals/context.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest'
 import {
   assembleGoalContext,
+  fetchActiveGoalCountWith,
   hasSufficientContext,
   type GoalContextFetchers,
   type GoalContextData,
@@ -56,6 +57,56 @@ describe('assembleGoalContext', () => {
       fetchers({ fetchGoals: async () => [{ title: 'X', targetValue: 10, currentValue: 2.5 }] })
     )
     expect(data.activeGoals).toEqual(['- X (2.5/10)'])
+  })
+})
+
+describe('fetchActiveGoalCountWith', () => {
+  // The auto-generation cap is now decided on this one read instead of the full
+  // five-read bundle. Two derivations of "how many goals are active" therefore exist,
+  // and if they drift the cap starts meaning different things depending on which path
+  // evaluated it. These cases hold them together.
+  const goals: RawGoal[] = [
+    { title: 'Read books', targetValue: 12, currentValue: 3 }, // active
+    { title: 'Ship v1', targetValue: 1, currentValue: 1 }, // complete (>=target)
+    { title: 'Run miles', targetValue: 100, currentValue: 100 }, // complete
+    { title: 'No target', targetValue: 0, currentValue: 0 } // target 0 → active
+  ]
+
+  it('agrees with assembleGoalContext on the same goals', async () => {
+    const f = fetchers({ fetchGoals: async () => goals })
+    const bundle = await assembleGoalContext(f)
+    await expect(fetchActiveGoalCountWith(f)).resolves.toBe(bundle.activeGoalCount)
+    await expect(fetchActiveGoalCountWith(f)).resolves.toBe(2)
+  })
+
+  it('reads goals and nothing else', async () => {
+    // The whole point is the four reads it does NOT make. `/v3/memories?limit=500` and
+    // `/v1/conversations?limit=100` are the expensive two.
+    const calls: string[] = []
+    const spying: GoalContextFetchers = {
+      fetchMemories: async () => {
+        calls.push('memories')
+        return []
+      },
+      fetchConversations: async () => {
+        calls.push('conversations')
+        return []
+      },
+      fetchTasks: async () => {
+        calls.push('tasks')
+        return []
+      },
+      fetchPersona: async () => {
+        calls.push('persona')
+        return null
+      },
+      fetchGoals: async () => {
+        calls.push('goals')
+        return goals
+      }
+    }
+    await fetchActiveGoalCountWith(spying)
+    expect(calls).toEqual(['goals'])
   })
 })
 
--- a/desktop/windows/src/main/assistants/goals/context.test.ts
+++ b/desktop/windows/src/main/assistants/goals/context.test.ts
@@ -1,7 +1,8 @@
 import { describe, it, expect } from 'vitest'
 import {
   assembleGoalContext,
-  fetchActiveGoalCountWith,
+  activeGoalCount,
+  withPrefetchedGoals,
   hasSufficientContext,
   type GoalContextFetchers,
   type GoalContextData,
@@ -60,33 +61,54 @@ describe('assembleGoalContext', () => {
   })
 })
 
-describe('fetchActiveGoalCountWith', () => {
-  // The auto-generation cap is now decided on this one read instead of the full
-  // five-read bundle. Two derivations of "how many goals are active" therefore exist,
-  // and if they drift the cap starts meaning different things depending on which path
-  // evaluated it. These cases hold them together.
+describe('activeGoalCount', () => {
+  // The auto-generation cap is decided on this number, taken straight off the goal
+  // list rather than from the full five-read bundle. Two derivations of "how many
+  // goals are active" would let the cap mean different things depending on which path
+  // evaluated it, so `assembleGoalContext` and the gate share this one function.
   const goals: RawGoal[] = [
     { title: 'Read books', targetValue: 12, currentValue: 3 }, // active
     { title: 'Ship v1', targetValue: 1, currentValue: 1 }, // complete (>=target)
     { title: 'Run miles', targetValue: 100, currentValue: 100 }, // complete
-    { title: 'No target', targetValue: 0, currentValue: 0 } // target 0 → active
+    { title: 'No target', targetValue: 0, currentValue: 0 } // target 0 -> active
   ]
 
-  it('agrees with assembleGoalContext on the same goals', async () => {
-    const f = fetchers({ fetchGoals: async () => goals })
-    const bundle = await assembleGoalContext(f)
-    await expect(fetchActiveGoalCountWith(f)).resolves.toBe(bundle.activeGoalCount)
-    await expect(fetchActiveGoalCountWith(f)).resolves.toBe(2)
+  it('is the same number assembleGoalContext reports', async () => {
+    const bundle = await assembleGoalContext(fetchers({ fetchGoals: async () => goals }))
+    expect(activeGoalCount(goals)).toBe(bundle.activeGoalCount)
+    expect(activeGoalCount(goals)).toBe(2)
   })
 
-  it('reads goals and nothing else', async () => {
-    // The whole point is the four reads it does NOT make. `/v3/memories?limit=500` and
-    // `/v1/conversations?limit=100` are the expensive two.
+  it('counts an empty list as zero', () => {
+    expect(activeGoalCount([])).toBe(0)
+  })
+})
+
+describe('withPrefetchedGoals', () => {
+  const goals: RawGoal[] = [{ title: 'Read books', targetValue: 12, currentValue: 3 }]
+
+  it('serves goals from the prefetched list without asking for them again', async () => {
+    let goalFetches = 0
+    const wrapped = withPrefetchedGoals(
+      fetchers({
+        fetchGoals: async () => {
+          goalFetches += 1
+          return []
+        }
+      }),
+      goals
+    )
+    const data = await assembleGoalContext(wrapped)
+    expect(goalFetches).toBe(0)
+    expect(data.activeGoals).toEqual(['- Read books (3/12)'])
+  })
+
+  it('leaves the other four reads untouched', async () => {
     const calls: string[] = []
-    const spying: GoalContextFetchers = {
+    const base = fetchers({
       fetchMemories: async () => {
         calls.push('memories')
-        return []
+        return ['m']
       },
       fetchConversations: async () => {
         calls.push('conversations')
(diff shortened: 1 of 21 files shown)


Git-on-my-level:
Thanks @ZachL111 — this is a careful, well-evidenced sweep. I went through the full diff and verified the claims against the current code; everything I checked holds up. What I verified per change, then the handful of behavior tradeoffs I'd like a human maintainer to sign off on before merge.

**Verified**

1. `retentionSweep.ts` — scheduled passes now require `live`, so the default install (`retentionMode` unset → `'dry-run'`) no longer pays a full `/v3/memories` page-through plus a 200-conversation fetch every 30 minutes just to `console.log`. Wiring RewindTab's Preview button to `runRetentionSweep('manual')` finally makes the mode's own subtitle ("Preview logs what it would delete") true. The test pinning zero backend calls on a default install — including through `maybeStartRetentionSweep`'s real timers — is exactly the right contract.
2. `promotionService.ts` — the flat 60s `setInterval` POST to `/v1/staged-tasks/promote` becomes a self-scheduling ladder (60s → 2m → 5m → 10m, hold at 10m). Reset-on-promote keeps a backlog draining one per minute, signed-out ticks reset to the floor, and `scheduleSafetyTick()`'s `started` guard closes the stop-mid-promote re-arm race that `clearInterval` used to prevent for free — good catch, and the new T-E test proves it. `create.ts`'s `getLastPromotedAt()` getter observes success without breaking the single-writer rule.
3. `chatQuotaGate.ts` — `sync()` gains a 60s floor measured from the last *attempt*, so both reveal bursts and a failing quota endpoint back off. The cold-start path stays unthrottled so an already-over-cap user is still blocked on the first send, `applyQuota` restarts the interval for externally supplied snapshots, and `recordQuery()`'s optimistic delta keeps the verdict correct while a snapshot is held — each pinned by a test.
4. `focusRefetch.ts` (new) — `useThrottledWindowFocus` keeps first-focus revalidation and drops only the repeats, with the throttle clock in a ref so the per-render resubscription at the three call sites (`Memories.tsx`, `Apps.tsx`, `HomeGoalsChips.tsx`) can neither reopen nor extend the window. The resubscribe test covers exactly that failure mode.
5. `aboutUser.ts` — 5-minute `CARD_TTL_MS` rebuild floor, with both invalidation paths (account switch fails the uid check; `resetAboutUserCard` nulls `cached`) explicitly exempt and tested.
6. `goals/context.ts` + `schedule.ts` — the `MAX_ACTIVE_GOALS` cap is now checked on one `GET /v1/goals/all` before the five-read context build, `activeGoalCount()` shares `assembleGoalContext`'s exact derivation so the cap can't mean two different things, and `withPrefetchedGoals` hands the list through so a proceeding run doesn't double-fetch. A failed goals read still fails open to the old pay-in-full behavior rather than adding a new failure mode.
7. `Tasks.tsx` — conversation-title fetch deferred until the panel is actually on screen (`pathname === '/tasks'`, verified against `routes/manifest.ts`) with a populated-cache short-circuit. `ChatAppPicker.tsx` — the gate wrapper now returns null *before* mounting the component whose `useChatApps()` fetches on mount (verified), killing the per-focus-cycle `GET /v1/apps` for the picker-off majority.

**What I'd like a human maintainer to sign off on** — the implementation looks right, but each of these is a user-facing tradeoff rather than a pure cost fix:

- Selecting **Preview** in Rewind settings no longer runs anything in the background; the preview exists only at the moment of the click, and its output still goes to the DevTools console (surfacing the would-delete summary in the UI feels like a worthwhile follow-up).
- Worst-case discovery latency for a task staged **out of band** (another device / backend extraction) goes from ≤60s to **≤10 minutes**.
- Focus revalidation is now at most **once per 60s** per surface, and the about-user card can be up to **5 minutes stale** — small staleness windows, but real ones.

None of these look wrong to me; they're product calls about freshness vs. request volume that deserve a maintainer's explicit OK, so I'm leaving the approval decision to a human maintainer.

Minor, non-blocking: switching Auto-cleanup to **Delete** doesn't trigger an immediate live pass (the first sweep still waits for the next 30-minute tick) — same as before this PR, just noting it as possible polish alongside the Preview-output UX.

The test discipline here is genuinely excellent — boundary assertions one millisecond either side, the shipped-default interval exercised explicitly, and the backoff ladder pinned rung by rung.

---

undivisible (approved):
Cline review agrees: throttles idle/focus-driven Windows backend fetches (useThrottledWindowFocus, picker/goals/tasks/quota/retention) with unit tests and first-focus preserved. Typecheck/Lint/Test and Windows build smoke green. Merge.

PR #11839, opened 2026-08-19
Arm candidates on a discrete transition, not on a condition that merely persists

Git-on-my-level:
## What changed and why

The candidate-arming prompt selected for standing conditions. Two lines did it:

1. `Ask first: do the facts record a commitment, a deadline, a blocker, a conflict, or a change the user may not have seen?` — "blocker" and "commitment" had no requirement that anything *changed*. A sentence of the form "X is still blocked on Y" is a blocker; "Z is still pending" is a commitment. Both are standing conditions that would read equally true yesterday and tomorrow, and both satisfied this instruction.
2. `Write only what will still be true and useful then.` — a staleness guard against minute-volatile counts (load-bearing; a candidate built on a file count went stale before delivery). A standing condition is *maximally* durable, so the guard optimized straight into another failure.

Arming now requires a discrete transition (became blocked, deadline moved/approaches, commitment was made, conflict appeared, state changed). A condition that merely persists is not enough. The delivery-time test is actionability (name the object the user can act on), not mere truth. The minute-volatile-count warning is unchanged.

## Measurement

A prompt rewrite (`e0681c28dd`, 17:28 UTC 2026-08-16) fixed the delivery gate (3/69 approved → 7/10; Fisher exact two-sided p = 4.4e-06). That exposed this upstream defect:

- 3 of 7 new-prompt approvals were rated "should not have interrupted" (3 raters, unanimous). All three described a standing condition.
- Held-out set of 23 later-armed candidates, never used to form the hypothesis: **12 of 23 (52%) STANDING**, 21/23 unanimous across 3 raters.

## Falsifiable prediction

After this lands, re-classifying newly-armed candidates should show the STANDING share well below 52%. If it does not, the criterion is not what drives it and fact worthiness is the next place to look.

## Quality change, not a cost saving

The candidate ceiling already short-circuits before the gate's model call, so capped candidates cost zero tokens today. Arming volume is not a token driver. This is a quality change with no expected cost saving.

none

## How it was verified

Did not launch or disturb any running Omi bundle; a dogfood measurement is in progress.

- `swift test --filter ContextWorkstreamReconciler` — 25 tests, 0 failures
- `swift test --filter ContextBucketPromptAssemblerTests` — 26 tests, 0 failures
- `swift test --filter ContextProactivityRetrievalHopTests` — 21 tests, 0 failures
- `swift test --filter ContextFactWritePolicyTests` — 11 tests, 0 failures
- `swift test --filter ContextBucketDirectorProbeTests` — 7 tests, 0 failures
- Repo pre-push preflight, no `--no-verify` and no `PRE_PUSH_SKIP_DESKTOP_CHANGELOG`

The new test initially failed: the rewrite rewrapped the minute-volatile-count guard
across a source line, so an exact-literal `contains` no longer matched even though the
guard was intact. The clause assertions now run against a whitespace-flattened copy of
the instructions, so rewrapping a sentence can neither silently pass nor silently fail
an assertion about whether that clause is still present.

## Tests

`testCandidateInstructionsRequireADiscreteTransitionAndStayActionable` asserts the load-bearing clauses: discrete-transition requirement, persist-is-not-enough omit, actionability (not truth) at delivery, and the minute-volatile-count staleness guard. Existing composition/preamble tests are unchanged.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11839?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextWorkstreamReconciler.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/ContextWorkstreamReconciler.swift
@@ -1017,16 +1017,22 @@ actor ContextWorkstreamReconciler {
   /// uncommitted files": a candidate built on a minute-volatile count went
   /// stale before anything could show it. The message survives to delivery only
   /// if it is written for the moment of delivery, not the moment of writing.
+  /// Durability is actionability, not mere truth: a standing condition is
+  /// maximally still-true hours later, which is exactly why it is not worth
+  /// interrupting. Arming requires a discrete transition.
   static let candidateInstructions = """
     \(ScreenDerivedContent.untrustedPreamble)
     For each bucket below, write at most one short notification, or omit the bucket.
     Omitting most buckets is normal.
-    Ask first: do the facts record a commitment, a deadline, a blocker, a conflict,
-    or a change the user may not have seen? If not, omit the bucket.
+    Ask first: do the facts record a discrete transition the user may not have seen?
+    Something became blocked, a deadline moved or now approaches, a commitment was
+    made, a conflict appeared, or a state changed. A condition that merely persists
+    is not enough. If not, omit the bucket.
     Never write a notification that merely describes what was on the user's screen.
     The notification will be delivered hours later, when the user returns to this
-    context. Write only what will still be true and useful then. Do not build the
-    message around counts or figures that change minute to minute.
+    context. Write only what will still be actionable then: name the object the user
+    can act on. Do not build the message around counts or figures that change minute
+    to minute.
     List in fact_ids every supplied fact id the message relies on.
     Add a one-line trigger_note describing when it should fire.
     """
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
The two failing checks (`Desktop Swift Static & Test Contracts`, `Desktop Swift Build & Tests`) were inherited from upstream `main`, not caused by this PR's diff: `TranscriptionFinalizationStateMachineTests.testRecoverPendingFinalizationsProcessesExhaustedBackendBoundLocalFallback` regressed on main at 3265e65b and was fixed upstream by #11836 (3599aabb, `ConversationFinalizationService.swift`).

Action taken: rebased `fix/candidate-arming-requires-event` onto latest `main` (c05b22d8 → dc2c55d71). `git patch-id` confirms the commit is byte-identical; the PR diff is unchanged (ContextWorkstreamReconciler + its tests + changelog entry). Both Desktop Swift checks now pass on the new head.

PR #11853, opened 2026-08-19
fix(backend): return 400 for chat attachments the file pipeline cannot process

kodjima33:
## The bug

`POST /v2/files` (and its `/v1/files` twin) answered **HTTP 500** for two kinds of ordinary user input. Both are live on the current prod image `c5b0b5d` (`api.omi.me`, Cloud Run `backend`), across the Flutter app, the Windows desktop client, iOS and the web app:

```
PIL.UnidentifiedImageError: cannot identify image file '/tmp/<uuid>_<name>.heic'
  routers/chat.py:1518 upload_file_chat
  -> utils/other/chat_file.py:166 FileChatTool.upload -> File.generate_thumbnail

openai.BadRequestError: Error code: 400 - Invalid extension ogg. Supported formats: ...
  routers/chat.py:1518 upload_file_chat
  -> utils/other/chat_file.py:166 FileChatTool.upload -> openai.files.create
```

## Root cause

`mimetypes.guess_type` maps `.heic` to `image/heic`, so `File.is_image()` is true and `generate_thumbnail()` runs — but Pillow ships no HEIF decoder, so `Image.open` raises `UnidentifiedImageError`. An iPhone camera-roll photo, the single most likely attachment on mobile, crashed the route.

An `.ogg` voice note is not an image, so it went straight to OpenAI Files, whose accepted-extension list contains no audio formats. The provider's 400 propagated unhandled.

Neither is a server fault. Both are request input arriving over a public route, and the route owns that contract: an unprocessable attachment is a bad request, not an internal error. The client cannot tell the user anything useful from a 500, and every attempt pages as a server error.

## The fix

`utils/other/chat_file.py` raises a typed `UnsupportedChatFileError` at the two points where an unprocessable file is identified — the undecodable image and the provider-rejected extension — and both upload endpoints map it to `HTTPException(400)` with the extension in the detail:

```
{"detail": "Unsupported attachment: 'heic' files are not supported in chat."}
```

Nothing else changes. The provider is never called for an image Pillow could not decode, and supported files upload exactly as before. Only these two narrow exception types are caught; a disk or transport failure still surfaces as a 500, because that *is* a server fault.

## What I tested

- **New `backend/tests/unit/test_chat_file_upload_unsupported.py`** drives `POST /v2/files` and `POST /v1/files` **over HTTP** against the real `routers.chat` and the real `utils.other.chat_file` (only the provider client, auth and Firestore stubbed): `.heic` → 400, provider-rejected extension → 400, `text/plain` → 200 and persisted. 5 passed.
- **Control against pre-fix source** (source files stashed, test file kept): 4 of the 5 fail with the verbatim prod exceptions — `PIL.UnidentifiedImageError` and `openai.BadRequestError` — and the happy-path test passes both before and after, confirming the guard changes nothing on the supported path.
- **A real HEIC file, not a synthetic blob**: a file produced by macOS `sips -s format heic` raises the same `UnidentifiedImageError` in Pillow 12.2. Driving that exact file through `POST /v2/files` returns **500 on pre-fix source** and **400 with the message above** with the fix.
- `tests/unit/_chat_router_test_harness.py` hand-stubs the exact leaves `routers.chat` imports, so the new name had to join that stub inventory; without it `test_chat_stream_error_fallback` (11 tests) and `test_chat_quota_counting_router` (9) fail at import. Both green again.
- `backend/test.sh` over the 17 chat / file / upload suites: **64 passed, 5 deselected**, no failures.
- `make preflight`: all selected checks pass.

## Declarations

Failure-Class: new

The violated contract: request input a downstream library or provider rejects must be answered as 4xx at the route boundary, not allowed to escape as a 500. The nearest existing class is `FC-malformed-doc-read`, which covers malformed **persisted documents** on read paths; this is user-supplied request input on a write path, so it is not the same contract.

Line-Count-Exception: backend/routers/chat.py | 1791 -> 1797 | six lines are the two `try/except UnsupportedChatFileError -> HTTPException(400)` wrappers around the existing upload calls; splitting this router is out of scope for a bug fix.

Product invariants: none matched.

🤖 automated by the hourly watchdog — tested and merged

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11853?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-request-input-rejection-escapes-as-server-fault.json
+++ b/.github/failure-classes/FC-request-input-rejection-escapes-as-server-fault.json
@@ -1,0 +1,18 @@
+{
+  "schema_version": 1,
+  "id": "FC-request-input-rejection-escapes-as-server-fault",
+  "violated_contract": "A route owns the classification of its own request input. When a library or an upstream provider rejects something the caller supplied -- a media type with no decoder, an extension outside a vendor's accepted set -- that rejection is a bad request, and the route must answer 4xx. Letting the library's or provider's exception propagate turns ordinary user input into a 5xx: the client can say nothing useful to the user, the failure is indistinguishable from a real outage in serving metrics, and every retry re-pages.",
+  "canonical_prevention": "Catch the narrow, input-caused exception types at the point where the unprocessable input is identified, convert them to one typed domain error, and map that error to an explicit 4xx at the route boundary with a detail the client can show. Catch only the input-caused types -- transport, disk and quota failures must still surface as 5xx, because those are server faults. A regression test drives the real route over HTTP with the offending input and asserts the status code the client receives.",
+  "canonical_prevention_artifact": [
+    "backend/utils/other/chat_file.py",
+    "backend/routers/chat.py",
+    "backend/tests/unit/test_chat_file_upload_unsupported.py"
+  ],
+  "evidence_prs": [
+    11853
+  ],
+  "scope_hints": [
+    "backend/routers/**"
+  ],
+  "status": "open"
+}
--- a/backend/routers/chat.py
+++ b/backend/routers/chat.py
@@ -67,7 +67,7 @@ from utils.rate_limit_config import get_effective_limit, RATE_LIMIT_SHADOW
 from utils.subscription import enforce_chat_quota, is_trial_paywalled
 from utils import share_links
 from utils.other import endpoints as auth, storage
-from utils.other.chat_file import FileChatTool
+from utils.other.chat_file import FileChatTool, UnsupportedChatFileError
 from utils.multipart import (
     CHAT_FILE_MAX_PART_SIZE,
     MultipartMaxPartSizeRoute,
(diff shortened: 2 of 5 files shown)


kodjima33:
Closing this myself rather than leaving it open — the fix is finished and verified, but I will not merge over a red fail-closed gate, and an unattended job should not leave PRs sitting.

**Why it is not merged:** CI is 21 pass / 1 fail. The only red is `Backend Hermetic Merge Gate`, which fails closed because `Listen Pusher Stack Gauntlet` was killed at its 20-minute job limit while hung on its `Install Redis server` apt step. Three separate runner allocations hung on that exact step (attempt 2: 07:56:04Z → 08:15:26Z, 19m22s in apt alone). The same job took 7m12s on #11851 two hours earlier. Nothing in this diff touches the listen/pusher stack.

Every check that actually exercises the change is green: Backend unit suite, Backend Checks (preflight 25/25), Hermetic Backend E2E, Sync Cloud Tasks Gauntlet, all six image-smokes, source-closure, OpenAPI contract, Formatting, Hygiene, plus the Desktop/Web/Mobile lanes.

**The work is preserved.** Branch `watchdog/chat-file-upload-unsupported-400` (`2ff4f076b4`) is pushed and unchanged. Once the runners recover this is one command:

    gh pr create --head watchdog/chat-file-upload-unsupported-400 --base main

Full bug, root cause, fix and verification evidence are in the PR body above.

🤖 hourly watchdog — tested, not merged, closed rather than abandoned

PR #11854, opened 2026-08-19
fix(desktop): replay Gemini requests dropped by a transport failure

aryanorastar:
## Summary

`GeminiClient.shouldAutoRetry` excluded every transport failure, so the existing 3-attempt retry loop only ever fired for a typed `apiError` the backend had authorized with `X-Omi-Retryable`. A dropped connection was discarded on the first attempt.

That excluded the one error class most worth retrying. `NSURLErrorNetworkConnectionLost` (-1005) is a stale pooled-connection race, not an outage: URLSession reuses a keep-alive socket the server has already closed and the request dies in milliseconds.

## Evidence

Measured on a live desktop session while dwelling on an ordinary work screen:

```
Suggestion: delivering [95%] "..."           1
Suggestion: evaluation failed (network)     12
curl to the same host, 5 probes             200 in ~0.4s each
```

Twelve of thirteen suggestion evaluations were lost, 4–7s apart, while plain requests to the same host succeeded in ~0.4s throughout. Each failure had already spent a screenshot downscale, a grounding pass and an upload, and returned nothing — with no user-visible signal that anything had gone wrong. The feature simply appears not to work.

## What changed

- `shouldAutoRetry` now replays a bounded set of transport failures: `networkConnectionLost`, `timedOut`, `cannotConnectToHost`, `notConnectedToInternet`, `dnsLookupFailed`, `cannotFindHost`.
- `.cancelled` stays excluded — that is the user or a superseding evaluation withdrawing the work, and replaying it would resurrect work nobody is waiting for.

**The response contract is unchanged.** A *response* still requires the backend's `X-Omi-Retryable` authorization to be replayed, and that assertion is kept as-is. A transport failure produces no response, so there is no authority to consult; the previous policy read "no authorization" as "must not retry".

Replay is safe here because every call this client makes is a `generateContent` inference — prompt plus image in, text out, no server-side state change. A duplicate costs one extra inference and nothing else.

## Verification

- `swift build` → clean
- `swift test --filter AIBackendErrorNormalizationTests` → 9 passed

Two assertions in `testGeminiRetryRequiresTypedBackendAuthorization` encoded the previous policy and were rewritten. Per the testing rules I am citing the external source rather than restating the implementation: the measurement above, plus Apple's documented guidance that -1005 is transient and should be retried. The backend-authorization assertions from that test are preserved unchanged in place.

## Honest gaps

- **Not measured end-to-end.** The retry should raise the hit rate, but I have not yet re-run a full session to quantify it. The underlying -1005 frequency on this network is itself unexplained and may warrant connection-reuse settings on the URLSession rather than retries alone — retries treat the symptom.
- **This widens replay for every `GeminiClient` caller** — memory extraction, task prioritization, task dedup, insight, live notes, suggestions. Safe while they are all inference reads; a future non-idempotent caller would need its own guard. Flagging explicitly because the previous policy was deliberate and I am relaxing it.
- No on-device/pendant verification; desktop only.

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11854?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/GeminiClient.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/GeminiClient.swift
@@ -441,10 +441,37 @@ actor GeminiClient {
   /// Replay only outcomes whose typed backend contract says they are safe.
   static func shouldAutoRetry(_ error: Error) -> Bool {
     if let geminiError = error as? GeminiClientError {
-      return geminiError.shouldAutoRetry
+      if geminiError.shouldAutoRetry { return true }
+      if case .networkError(let underlying) = geminiError {
+        return isReplayableTransportError(underlying)
+      }
+      return false
     }
-    // Raw URLSession errors are ambiguous after dispatch and must not be replayed.
-    return false
+    return isReplayableTransportError(error)
+  }
+
+  /// Transport failures that are safe to replay for this client.
+  ///
+  /// Replaying a dispatched request is only unsafe when the request may have had an
+  /// effect. Every call this client makes is a `generateContent` inference: it reads a
+  /// prompt and an image and returns text, with no server-side state change, so a duplicate
+  /// costs one extra inference and nothing else.
+  ///
+  /// `NSURLErrorNetworkConnectionLost` (-1005) is the dominant failure here and is a stale
+  /// pooled-connection race, not a real network outage: URLSession reuses a keep-alive
+  /// socket the server has already closed, and the request dies immediately. Measured on a
+  /// live desktop session, 12 of 13 suggestion evaluations failed this way while ordinary
+  /// requests to the same host succeeded in ~0.4s — every one of them was discarded without
+  /// a second attempt.
+  static func isReplayableTransportError(_ error: Error) -> Bool {
+    let code = (error as? URLError)?.code ?? URLError.Code(rawValue: (error as NSError).code)
+    switch code {
+    case .networkConnectionLost, .timedOut, .cannotConnectToHost, .notConnectedToInternet,
+      .dnsLookupFailed, .cannotFindHost:
+      return true
+    default:
+      return false
+    }
   }
 
   /// Closed Gemini model tier for fallback telemetry (no free model ID strings).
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Verified against head 67a63bd — this is a well-evidenced fix with the right instincts on replay safety. Thank you.

**Correctness**

- `desktop/macos/Desktop/Sources/ProactiveAssistants/Core/GeminiClient.swift`: every request path in this client (`sendRequest` image/text, `sendTextRequest`, `sendImageToolLoop`) POSTs to `proxyURL(action: "generateContent")` — a pure inference with no server-side state change — so the duplicate-request-is-harmless reasoning holds. Raw `URLError`s from `URLSession.data(for:)` arrive unwrapped in the catch blocks, so the new `isReplayableTransportError` fallback is exactly the path that starts firing. The `X-Omi-Retryable` response contract is untouched, and `.cancelled` stays excluded so superseded evaluations are not resurrected.
- `desktop/macos/Desktop/Tests/AIBackendErrorNormalizationTests.swift`: good coverage of both directions — six replayable transport codes asserted, `cancelled`/`badURL`/`userAuthenticationRequired` asserted non-replayable, and the backend-authorized/denied/absent response assertions preserved.
- `desktop/macos/changelog/unreleased/20260819-gemini-transport-retry.json`: valid `user_facing` fragment; format matches what `check-desktop-changelog.py` expects.

**Two non-blocking considerations**

1. `timedOut` was deliberately excluded before (the prior test asserted it, and `GeminiClientError.shouldAutoRetry`'s doc comment still warns that auto-retrying multi-minute deadlines keeps users waiting). With `timeoutInterval = 300` and the 2s/8s backoff, a blackholed host now takes ~15 min to fail on the image/text paths, and `sendImageToolLoop` can double that via the Pro→Flash fallback (3 attempts × 2 models ≈ 30 min). That is clearly fine for background suggestions; worth a conscious call for foreground callers.
2. The enum-level `GeminiClientError.shouldAutoRetry` still carries the comment "A transport error after dispatch is ambiguous. Only a typed backend response may authorize replay." — now overridden by the static. Worth updating so future readers aren't misled. Tiny hardening opportunities: check `(error as NSError).domain == NSURLErrorDomain` in the raw-code fallback, and add a test for the wrapped `GeminiClientError.networkError(URLError(...))` shape to cover that branch.

Desktop Swift Build & Tests, Release Compile, and Static & Test Contracts are all green. Leaving for human maintainer review on one specific point: sign-off that the longer time-to-failure from replaying `timedOut` is acceptable for foreground Gemini callers.

---

kodjima33 (approved):
Desktop bug fix: replays a bounded set of transport failures (NSURLErrorNetworkConnectionLost) that were silently dropping Gemini suggestion requests; scoped diff with a regression test.

PR #11857, opened 2026-08-19
security(backend): gate Anthropic server-side web search on the agentic chat path

### LINKED ISSUE #11412: Anthropic server-side web_search is an unguarded egress path for prompt-injected exfiltration

RatnamOjha:
Fixes #11412.

## The defect

`_convert_tools` appended `WEB_SEARCH_TOOL` unconditionally (`backend/utils/retrieval/agentic.py:409` at the time of filing), so every direct Anthropic agent loop offered Anthropic's server-side `web_search`.

Anthropic executes that tool on its own infrastructure. The block never reaches `_execute_tool`, so the `fetch_url` allowlist, SSRF guard and redirect bound do not apply to it, and the query string is an outbound request to a third party. After a memory/Gmail/screen tool returns private data, an injected instruction can place that data into a search query and it leaves the trust boundary. The backend only ever observes the tool at `content_block_start` (`type == "server_tool_use"`), which confirms it is never executed in-process.

The issue describes the tool as gated on `include_server_web_search=True`. That flag no longer exists anywhere in `backend/` — the tool was attached to every agentic loop.

## The fix

PR #11414 built exactly this gate for the desktop chat path. `agentic.py` serves the mobile/main app chat agent and was never migrated. This applies the same, already-reviewed decision to the un-migrated call site.

**Commit 1** moves `_carries_private_tool_output` out of `routers/desktop_chat.py` into `utils/llm/private_context.py`, parameterizing the two desktop-specific inputs (the public-safe producer allowlist, the inline tool-output markers). `agentic.py` is a utils module and cannot import a router, so sharing the classifier was a precondition for the fix. `desktop_chat` now calls the shared function; no duplicate copy is left behind, per the root `AGENTS.md` rule against in-repo compatibility layers. No behavior change.

**Commit 2** gates the agentic path.

### Why the gate is in the loop, not in `_convert_tools`

`_messages_to_anthropic` emits text-only turns, so the opening request is always clean; the taint only appears **during** the loop, as each tool result is appended to `messages`. A single decision at composition time could never observe the condition it exists to catch. The offer is therefore re-decided before each provider request and latched once withheld.

This costs a prompt-cache miss on the iteration that trips the gate, since the tool block is part of the cached prefix. That is the price of the gate, and there is a comment at the call site saying so, because the obvious "optimization" of hoisting the check back out of the loop silently reintroduces the vulnerability.

### Scope

Only the direct Anthropic lane ships the provider-executed tool. The managed lane drops server tools in `_convert_anthropic_tools_to_openai` (they carry no `input_schema`) and reaches the public web through the in-process Perplexity function tool, which *does* pass through `_execute_tool`. A test pins that, so the managed lane cannot silently acquire the same unguarded path.

### Classification

Reads literal wire shapes — which producer emitted each `tool_result` — never message content, and treats unknown producers as private, so runtime-registered app tools taint the loop. This matches the failure class's `canonical_prevention` verbatim.

The public-safe allowlist holds only `get_omi_product_info_tool`, which returns GitHub product documentation. The desktop allowlist's names were deliberately **not** copied across: the agentic implementations of the same-sounding tools do return user data (`create_action_item_tool` returns the created item; `create_calendar_event_tool` can surface Google Contacts addresses in an error string; `save_user_preference_tool` echoes the saved text and any matching stored preference).

## Known limitation

Classification is by wire shape, so private data that a *previous assistant turn* already rendered into plain text is not treated as taint. That is the deliberate trade the failure class prescribes ("classify that context from literal wire shapes rather than content heuristics"); content-based classification is explicitly what it warns against. The exfiltration path in #11412 — a tool result in the current loop — is closed.

## Verification

Red first, against unmodified code (`backend/tests/unit/test_agentic_web_search_taint.py`):

```
FAILED test_web_search_is_withheld_after_a_private_tool_result
FAILED test_unknown_producer_is_treated_as_private
FAILED test_withheld_path_is_recorded_once_per_loop
3 failed, 8 passed

>       assert WEB_SEARCH_NAME not in _tool_names(requests[1])
E       AssertionError: assert 'web_search' not in ['web_search', 'get_memories_tool']
```

Green after the fix: `12 passed`.

The tests drive the real `_run_anthropic_agent_stream` loop through the `anthropic_client.messages.stream` seam and assert on the `tools` list of every request the loop actually composes — both directions (withheld after a private result, still offered on a clean loop), plus unknown-producer handling and single-shot telemetry.

Neighbouring suites:

```
tests/unit/test_desktop_chat.py tests/unit/test_fallback_observability.py
tests/unit/test_chat_agent_provider_retry.py tests/unit/test_retrieval_result_bounds.py
tests/unit/test_chat_agent_cost_report.py tests/unit/test_agent_tools_isolation.py
tests/unit/test_retrieval_semantics.py
-> all pass (run per-file; see note below)
```

Plus the 18 suites that import the touched modules, run individually: all pass.

```
scripts/check_module_stub_pollution.py    -> Checked 907 test file(s); 0 violation(s).
scripts/scan_import_time_side_effects.py  -> Checked 804 production file(s); 0 violation(s).
```

Note: `test_chat_agent_provider_retry.py` followed by `test_agent_tools_isolation.py` in one process fails collection with `ModuleNotFoundError: No module named 'utils.conversations.factory'`. This reproduces on unmodified `upstream/main` (verified in a detached worktree at `79a585be75`) and is unrelated to this change; the suites pass per-file.

**Not exercised against live Anthropic.** Verified hermetically at the provider seam only.

## Gates

Failure-Class: FC-provider-executed-tool-bypasses-egress-guard

Product invariants affected: none (`scripts/pr-preflight --suggest` reports none for these paths).

Line-Count-Exception: backend/utils/retrieval/agentic.py | 1512 -> 1560 | Security gate must be re-decided per provider request inside the agent loop; the taint does not exist at composition time, so the check cannot live in an existing helper. The classifier itself was placed in a new shared module rather than added here.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11857?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/desktop_chat.py
+++ b/backend/routers/desktop_chat.py
@@ -38,6 +38,10 @@ from utils.llm.gateway_client import (
 from utils.llm.gateway_observability import record_gateway_request_result
 from utils.llm.gateway_resilience import gateway_circuit, observe_gateway_first_byte
 from utils.llm.gateway_serving import is_gateway_transport_failure
+from utils.llm.private_context import (
+    flatten_text_blocks,
+    openai_messages_carry_private_tool_output,
+)
 from utils.llm.usage_tracker import reset_usage_context, set_usage_context
 from utils.observability.fallback import record_fallback
 from utils.other import endpoints as auth
@@ -281,15 +285,7 @@ def _uses_managed_chat_agent(body: Mapping[str, object]) -> bool:
 
 
 def _text(content: object) -> str:
-    if isinstance(content, str):
-        return content
-    if not isinstance(content, list):
-        return ''
-    return ''.join(
-        block.get('text', '')
-        for block in content
-        if isinstance(block, Mapping) and block.get('type') == 'text' and isinstance(block.get('text'), str)
-    )
+    return flatten_text_blocks(content)
 
 
 def _normalize_policy_text(text: str) -> str:
@@ -379,36 +375,13 @@ def _web_search_requested(body: Mapping[str, object]) -> bool:
 
 
 def _carries_private_tool_output(messages: object) -> bool:
-    if not isinstance(messages, list):
-        return False
-    tool_name_by_call_id: dict[str, object] = {}
-    for message in messages:
-        if not isinstance(message, Mapping):
-            continue
-        tool_calls = message.get('tool_calls')
-        if not isinstance(tool_calls, list):
-            continue
-        for call in tool_calls:
-            if (
-                isinstance(call, Mapping)
-                and isinstance(call.get('id'), str)
-                and isinstance(call.get('function'), Mapping)
-            ):
-                tool_name_by_call_id[call['id']] = call['function'].get('name')
-    for message in messages:
-        if not isinstance(message, Mapping) or message.get('role') != 'tool':
-            continue
-        name = tool_name_by_call_id.get(cast(str, message.get('tool_call_id')))
-        if not isinstance(name, str) or name not in _PUBLIC_SAFE_CLIENT_TOOLS:
-            return True
-    # The desktop hub also inlines tool output into the user turn behind this
+    # The desktop hub also inlines tool output into the user turn behind a
     # literal marker instead of sending an OpenAI `tool` message, so the same
     # private data reaches the request without a tool_call_id to classify.
-    return any(
-        isinstance(message, Mapping)
-        and message.get('role') == 'user'
-        and _UNTRUSTED_TOOL_CONTEXT_DELIMITER in _text(message.get('content'))
-        for message in messages
+    return openai_messages_carry_private_tool_output(
+        messages,
+        public_safe_tools=_PUBLIC_SAFE_CLIENT_TOOLS,
+        inline_private_markers=(_UNTRUSTED_TOOL_CONTEXT_DELIMITER,),
     )
 
 
--- a/backend/tests/unit/test_agentic_web_search_taint.py
+++ b/backend/tests/unit/test_agentic_web_search_taint.py
@@ -1,0 +1,249 @@
+"""Regression tests for the server-side ``web_search`` egress gate on the agentic path.
+
+Anthropic executes ``web_search`` on its own infrastructure, so the query string
+never reaches ``_execute_tool`` and leaves the trust boundary without passing any
+in-process control. Once a tool result is in the transcript, an injected
+instruction can place that data into the query.
+
+These cases drive the real Anthropic tool loop through the provider seam and
+assert on the ``tools`` list of every request the loop actually composes — the
+taint appears *during* the loop, so a single up-front decision cannot see it.
+"""
+
+import asyncio
+from types import SimpleNamespace
+
+import pytest
+
+from utils.llm.private_context import anthropic_messages_carry_private_tool_output
+from utils.retrieval import agentic
+from utils.retrieval.safety import AgentSafetyGuard
+
+WEB_SEARCH_NAME = 'web_search'
+
+
+def _text_delta(text):
+    return SimpleNamespace(type='content_block_delta', delta=SimpleNamespace(type='text_delta', text=text))
+
+
+def _tool_use(block_id, name, tool_input=None):
+    return SimpleNamespace(type='tool_use', id=block_id, name=name, input=tool_input or {})
+
+
+def _answer(text):
+    return SimpleNamespace(type='text', text=text)
+
+
+def _response(stop_reason, content):
+    return SimpleNamespace(stop_reason=stop_reason, content=content, stop_details=None)
+
+
+class _FakeStream:
+    def __init__(self, response, events):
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Verified against the current head (09a050a9) — this closes the #11412 egress hole on the agentic path, and the fix checks out end to end.

Per file:

- `backend/utils/llm/private_context.py` — the shared classifier reads literal wire shapes (producer identity via `tool_calls`/`tool_use` ids), never message content, and treats any producer it cannot positively identify as private. That default-deny is the right direction given app tools are registered at runtime. Having the OpenAI and Anthropic shapes answer one question in one module is cleaner than re-deriving it per surface.

- `backend/routers/desktop_chat.py` — now pure delegation to `openai_messages_carry_private_tool_output` with the desktop-specific inputs parameterized (`_PUBLIC_SAFE_CLIENT_TOOLS`, the inline delimiter marker). No compatibility duplicate left behind. I ran the existing desktop suite locally (70 passed) — the extraction is behavior-neutral there.

- `backend/utils/retrieval/agentic.py` — the gate is correctly placed inside the loop: `_messages_to_anthropic` emits text-only turns, so the opening request is always clean and the taint only appears as `tool_result` blocks appended after each iteration — a hoisted check could never observe it. Latching once tripped covers later transcript trimming, and the prompt-cache-miss cost is documented in-comment rather than hidden. `_PUBLIC_SAFE_AGENT_TOOLS` holding only `get_omi_product_info_tool` is appropriately conservative (the calendar tool can surface contact data in error strings). The `record_fallback` labels resolve inside the allowed enums (`component='other'`, `outcome='degraded'`), and the managed-lane scoping is real — `_convert_anthropic_tools_to_openai` drops schema-less server tools, with a test pinning it.

- `backend/tests/unit/test_agentic_web_search_taint.py` — good coverage: drives the actual loop through the provider seam and asserts on the composed `tools` list of every request (taint-during-loop, public-safe pass-through, unknown-producer taint, telemetry-recorded-once, managed-lane pin). Ran locally: 12/12 passed, plus 70/70 on the desktop suite.

Also confirmed the defect is real on `main` (`_convert_tools` appends `WEB_SEARCH_TOOL` unconditionally), so the PR's claim matches the diff, and this mirrors the already-reviewed #11414 decision on the one call site that was never migrated.

Nice work. Leaving formal sign-off to a human maintainer: the security-gate design itself — public-safe allowlist membership and the latch semantics — is a privacy-sensitive judgment that deserves maintainer eyes before merge.

---
Automated AI feedback on behalf of the Omi maintainers.

---

kodjima33 (approved):
Applies the already-reviewed desktop web_search gate to the un-migrated agentic call site — approving; leaving the merge to a maintainer since it moves a trust boundary.

undivisible (approved):
Approving. This is the #11414 gate on the un-migrated agentic path, and it's in the right place: re-decided inside the loop and latched, because `_messages_to_anthropic` flattens the opening history so a hoisted check is blind.

Classifier is fail-closed (orphans / unknown producers / anything not `get_omi_product_info_tool`). Desktop refactor looks behavior-preserving. Tests drive the real `_run_anthropic_agent_stream` loop and would catch a hoist regression.

Non-blocking follow-ups: treat duplicate/ambiguous `tool_use_id` as private, and add a case where the opening `messages` are already tainted so `requests[0]` is withheld. The prior-turn plain-text leak is the same trade as #11414, not a merge blocker.

PR #11862, opened 2026-08-19
fix(desktop): stop Gemini requests reusing a dead pooled connection

aryanorastar:
## Summary

Follow-up to #11854. That PR stopped the client discarding a dropped request; this one stops the request being made on a connection that is already dead.

`GeminiClient` issued every proxy request on `URLSession.shared`, whose connection pool is shared with every other request the app makes. When the server has already closed a pooled socket, the next caller is handed that dead socket and fails instantly with `NSURLErrorNetworkConnectionLost` (-1005). A retry draws from the same pool and reproduces the failure — which is why three attempts could fail identically within seconds, and why the retry added in #11854 was necessary but not sufficient on its own.

## Evidence

Measured on a live desktop session, before and after, driving the real evaluation path via `omi-ctl action probe_suggestion_nudge`:

```
before   evaluation_failed (-1005)   5 / 5
after    reached delivery            5 / 5
```

In the same post-change session the log shows six -1005 failures still occurring **in-process** — every one of them from `ScreenActivitySync` and `OCREmbeddingService`, which remain on `URLSession.shared`. Zero came from Gemini, and the retry path was never entered:

```
-1005 at socket level:  6     (other subsystems, still on the shared pool)
retries needed:         0
evaluation failures:    0
```

That is an A/B result inside a single running process: the subsystem moved off the shared pool succeeds, the ones still on it do not.

`curl` to the same endpoint with an 800 KB body returned in ~1.9s over **both** IPv4 and IPv6 throughout, which is what ruled out bandwidth, MTU and IPv6 reachability — curl dials a fresh connection each time.

## What changed

All four `URLSession.shared.data(for:)` call sites now route through one `send(_:)` helper that builds an ephemeral session per request and invalidates it with `finishTasksAndInvalidate()`.

An ephemeral per-request session costs one TLS handshake per evaluation. These are dwell-gated and capped by a daily budget, so that is a trade worth making to remove a class of stale-pool failure rather than manage it.

## Verification

- `swift build` → clean
- `swift test --filter AIBackendErrorNormalizationTests` → 9 passed
- 5× `probe_suggestion_nudge` on a live session → 0 transport failures (was 5/5)

## Honest gaps

- This does not explain **why** the server closes pooled connections so aggressively for this host; it removes the client's exposure rather than diagnosing the peer.
- `ScreenActivitySync` and `OCREmbeddingService` still use `URLSession.shared` and still fail this way. I left them out to keep the diff to the subsystem I could measure end to end, but the same fix likely applies.
- No unit test asserts the per-request session directly — the behaviour is a connection-pool property, not something the existing seams expose. The evidence here is the live measurement above rather than a test.

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/GeminiClient.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/Core/GeminiClient.swift
@@ -439,6 +439,30 @@ actor GeminiClient {
   }
 
   /// Replay only outcomes whose typed backend contract says they are safe.
+  /// Issues one proxy request on a connection pool that does not outlive it.
+  ///
+  /// `URLSession.shared` pools keep-alive connections across every request the app makes.
+  /// When the server has already closed a pooled socket, the next caller is handed that
+  /// dead socket and fails instantly with `NSURLErrorNetworkConnectionLost` (-1005); a
+  /// retry draws from the same pool and reproduces the failure, which is why three
+  /// attempts could fail identically within seconds.
+  ///
+  /// Measured on a live desktop session: 5 of 5 suggestion evaluations failed this way,
+  /// while `curl` to the same endpoint with an 800 KB body returned in ~1.9s over both
+  /// IPv4 and IPv6 — the difference being that curl dials a fresh connection each time.
+  ///
+  /// An ephemeral per-request session costs one TLS handshake per evaluation. These are
+  /// dwell-gated and capped by a daily budget, so that is a trade worth making to remove
+  /// an entire class of stale-pool failure rather than manage it.
+  static func send(_ request: URLRequest) async throws -> (Data, URLResponse) {
+    let configuration = URLSessionConfiguration.ephemeral
+    configuration.timeoutIntervalForRequest = 60
+    configuration.timeoutIntervalForResource = 300
+    let session = URLSession(configuration: configuration)
+    defer { session.finishTasksAndInvalidate() }
+    return try await session.data(for: request)
+  }
+
   static func shouldAutoRetry(_ error: Error) -> Bool {
     if let geminiError = error as? GeminiClientError {
       if geminiError.shouldAutoRetry { return true }
@@ -599,7 +623,7 @@ actor GeminiClient {
         urlRequest.timeoutInterval = 300
         urlRequest.httpBody = requestBody
 
-        let (data, urlResponse) = try await URLSession.shared.data(for: urlRequest)
+        let (data, urlResponse) = try await Self.send(urlRequest)
         try checkHTTPStatus(urlResponse, data: data)
 
         let response = try JSONDecoder().decode(GeminiResponse.self, from: data)
@@ -675,7 +699,7 @@ actor GeminiClient {
         urlRequest.timeoutInterval = timeout
         urlRequest.httpBody = try JSONEncoder().encode(request)
 
-        let (data, urlResponse) = try await URLSession.shared.data(for: urlRequest)
+        let (data, urlResponse) = try await Self.send(urlRequest)
         try checkHTTPStatus(urlResponse, data: data)
 
         let response = try JSONDecoder().decode(GeminiResponse.self, from: data)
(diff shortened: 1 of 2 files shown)


aryanorastar:
@kodjima33 @Git-on-my-level review request — no reviewer has looked at this one yet.

Follow-up to #11854: that PR retried a dropped Gemini proxy request, this one stops the retry drawing from the same dead connection pool. All 4 `URLSession.shared` call sites in `GeminiClient` now go through one `send(_:)` helper using a per-request ephemeral session. Live A/B in the PR description: same session, same failure class — Gemini requests moved off the shared pool hit 0/5 transport failures after, while `ScreenActivitySync`/`OCREmbeddingService` (left on `URLSession.shared`, out of scope here) kept failing with the identical -1005 in-process. `curl` ruled out bandwidth/MTU/IPv6 as the cause.

Git-on-my-level:
Thanks — this is a clean follow-up to #11854, and the in-process A/B evidence is a convincing causal isolation: Gemini requests moved off the shared pool went 5/5-failing to 5/5-delivered while `ScreenActivitySync` / `OCREmbeddingService` (still on `URLSession.shared`, correctly left out of scope here) kept hitting -1005 in the same session.

**Verified against head `9c1003a`:**

- `desktop/macos/Desktop/Sources/ProactiveAssistants/Core/GeminiClient.swift` — the new `static send(_:)` (L457–462) builds an `URLSessionConfiguration.ephemeral`, sets `timeoutIntervalForRequest = 60` / `timeoutIntervalForResource = 300`, and `defer { session.finishTasksAndInvalidate() }` fires only after the awaited `data(for:)` completes — so each request's connection pool is torn down with it and nothing lingers. All four former `URLSession.shared.data(for:)` call sites route through it: image `sendRequest(prompt:imageData:…)` (~L626), `sendTextRequest` (~L702), JSON `sendRequest(prompt:responseSchema:…)` (~L775), and `sendImageToolLoop` (~L1081). No shared-pool data call remains in this file (only the doc comment references it).
- Composition with #11854 is right: each retry attempt now creates a fresh session, so an `isReplayableTransportError` retry no longer redraws from the same pool that just produced the dead socket — that closes the loop the retry alone couldn't.
- Per-call-site `urlRequest.timeoutInterval` (300, or the `timeout` parameter) still takes precedence over the configuration values, so effective per-request idle timeouts are unchanged from before.
- `desktop/macos/changelog/unreleased/20260819-gemini-per-request-session.json` — matches the sibling schema exactly (checked against `20260819-lossless-screen-activity-sync.json` and `20260818-candidate-arming-requires-event.json`), and the user-facing wording is accurate.

**Two non-blocking notes:**

1. `timeoutIntervalForResource = 300` introduces a per-request total-time ceiling that `URLSession.shared` never imposed. Measured latency (~2s transport + model time) sits far below it, and every retry/tool-loop attempt gets its own session so the cap can't compound across a loop — just flagging it as a new bound worth a conscious call if a future caller passes a longer `timeout`.
2. The doc comment's cost argument ("dwell-gated and capped by a daily budget") is source-backed for the Suggestions evaluation path, but `GeminiClient` also serves insights, memory/task extraction, goals, and onboarding chat. For those, one extra TLS handshake per request is negligible next to inference time, so the trade still holds — only noting the comment frames the cost in suggestion-eval terms while the change applies client-wide.

Desktop Swift Build & Tests, Release Compile, and Static & Test Contracts are all green. Leaving for human maintainer review — the only judgment call is the new 300s per-request cap in note 1; no product-direction concern.

---

undivisible (approved):
Cline review agrees: GeminiClient no longer reuses URLSession.shared keep-alives that surface as -1005 on a dead pooled socket. Ephemeral per-request session matches the title; CI green. Merge.

PR #11868, opened 2026-08-19
fix(desktop): stop the screen-activity cursor from skipping rows it will never revisit

Git-on-my-level:
Screen-activity sync silently discards ~99% of what it captures. This makes it lossless, and cheaper per delivered row than the broken version is per delivered row.

## The defect

`ScreenActivitySyncService` selected `WHERE id > ? AND embedding IS NOT NULL` and advanced a monotonic cursor past everything it shipped. Embeddings are computed asynchronously after insert, so any row whose vector had not landed when the sweep passed was skipped **permanently** — not retried, not queued, not reported.

Measured on a live store: **705 of 26,822 rows over 30 days** ever left the device. It reads as a working sync, because rows keep arriving.

Three further mechanisms independently denied a row its embedding, so repairing the cursor alone recovers little:
- cross-batch content-hash dedup returned early without ever queueing the row,
- gated or failed embedding batches were dropped rather than requeued,
- the backfill that comments describe as the safety net had been stuck `completed=1, processedCount=0` since 2026-04-16.

## What changed

**Durable per-row delivery state** (`pending → textSynced → fullySynced`, plus `compacted`) on an indexed column, replacing the cursor. A row that was not ready when a sweep passed is picked up by a later sweep. OCR-bearing rows ship without waiting for a vector; the vector attaches afterwards. Migration is safe on a populated store and starts existing rows `pending`, so history is recoverable.

**Bucket-aligned eligibility.** Compaction keeps the longest OCR text per `(app, window, 5-min bucket)`. The original implementation tested each row's own timestamp against `now - slack`, which makes eligibility *slide*: the early rows of a bucket become ready while its later rows do not, so a bucket is ranked more than once and ships more than one winner. Replayed over 7,346 real OCR-bearing rows, the sliding form ships **5,517** where bucket-aligned ranking ships **3,842** — 44% more than the design intends, which is most of the compaction saving. The same alignment now applies to the embedding flush and the backfill queries, so all three agree on what a bucket is.

**A closed bucket's winner waits up to 15 minutes for its embedding** before shipping text-only. Without that wait almost every row shipped twice: sync sweeps at 5 minutes, and the embedding flush also buffers to 5 minutes and then batches on a 60s timer, so the vector reliably lost the race. The second push wrote a byte-identical Firestore document and rewrote all of its index entries for no data change. Losslessness is preserved — the grace period is a delay, not a gate.

**`ocrText`, `windowTitle`, `deviceName`, `clientDeviceId` exempted from Firestore single-field indexing.** Firestore auto-indexes every field in both directions, and nothing queries these — `ocrText` is only ever read back, never filtered or ordered on; semantic search goes through Pinecone. Measured per document, index bytes were ~75% of this collection's storage and the `ocrText` index alone was the majority of it. The one composite index is unaffected.

**Bounded deferred buffer.** Requeueing gated batches fixes real data loss but restores an unbounded buffer for a user whose backend gates every call. Oldest deferred items are shed first.

**Re-armed backfill** for the stranded history, capped per launch so it cannot stampede.

## The timestamp check, which came back negative

All 26,797 measured local rows already use the space-separated `%Y-%m-%d %H:%M:%S.%f` form that the backend's lexicographic range filters assume; **zero** use ISO `T`/`Z`. So no production date-range query over screen activity is currently returning wrong results.

Two latent paths are closed anyway: the client had a live ISO fallback, and `upsert_screen_activity_vectors` computed the Pinecone timestamp for a naive string in the *process* timezone rather than UTC (inert wherever the container runs UTC, wrong everywhere else). Both now canonicalize at the boundary.

## Cost

Independently modelled from the live store rather than asserted. Per user per month, at this store's capture volume:

| | rows delivered | month 1 | month 12 |
|---|---:|---:|---:|
| today (broken) | 705 | $0.11 | $0.14 |
| this change | 3,842 | $0.36 | $0.54 |
| lossless without compaction | 7,213 | $1.09 | $1.48 |

Rows delivered go up **5.3x** and spend goes up about **5x**. That increase is real and permanent — it is the cost of not throwing the data away. The dominant term is per-row flow (embedding + vector writes), which does not compound; permanent storage accrual is ~$0.017/user/month.

This store is a heavy dogfooder (26.8k captures over 19 active days), so treat per-user figures as an upper bound. Unit prices are list prices, not billed rates.

**Not addressed here, deliberately:** truncating OCR to 1,000 chars before embedding would cut the embedding bill by more than half, and Firestore already stores only the first 1,000 chars. But the embedding is what makes text *searchable*, and nothing has measured retrieval quality past that point — so it is a product call, not a cleanup, and it is left out.

**There is still no server-side retention of any kind.** The only deletion path is account deletion. Volume is unbounded and losslessness makes that more pressing, not less. Out of scope here; it needs its own decision.

## Tiering: cloud vectors are a paid desktop capability

Second commit. Free-tier desktop users sync their rows without vectors — they keep cloud screen history readable by time and app, and they keep **on-device** semantic search, which reads embeddings from the local store and never touches Pinecone. The OCR text is still stored, so vectors can be backfilled on upgrade: the gate defers the cost rather than destroying the ability to recover.

Gated on the desktop access tier, not `is_paid_plan` — screen activity is a desktop capability and a mobile-only paid plan did not buy it. BYOK does not grant it either: BYOK covers the user's own LLM keys, but the vector store is ours. Enforced server-side at ingest, since the client cannot be trusted about its own entitlement. Fails **open** on a lookup error, matching `should_defer_desktop_processing`.

**Correction to the cost table above.** The Gemini embedding spend (~$0.32/user/month) is *not* attributable to sync. Embeddings are computed for local Rewind search on every capture regardless of whether the row ever syncs, through our own proxy — sync merely piggybacks on them. So this gate saves the Pinecone half (~$0.20/user/month), not the embedding half, and the "this change" row above overstates what tiering can recover. Tiering the embedding as well would mean tiering on-device search; that is a separate product decision and is not made here.

## Verification

- `swift build` of the macOS package — a real build, not a typecheck.
- 9 Swift tests: migration on a populated store, a row unready at sweep time delivered later, an open bucket shipping nothing, a closed bucket shipping exactly one winner and never shipping again, the embedding wait and its expiry, compaction keeping the longest text, and gated/failed requeue.
- 8 backend tests covering timestamp canonicalization at ingest and the Pinecone UTC fix.

Not verified: no deploy was run; no production Firestore sample was read (network and credentials unavailable); the Pinecone index configuration is not in the repo, so its storage class is unconfirmed.

Failure-Class: new

Line-Count-Exception: desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift | 3450 -> 3470 | the schema migration and its callable boundary belong with the other migrations in the ledger; splitting the file to add one column would separate this migration from the ones it must stay ordered after

- INV-AUTH-1 — touched only by adding one migration to `RewindDatabase`'s ledger. Ownership is unchanged: the migration registers after every existing component migration and runs inside the pool the owner transition already configures. Embedding writes keep taking an owner-bound `LocalMutationAuthorization` lease, and the requeue paths still check `generation == ownerGeneration` before returning items to the buffer, so a batch that spans a retarget is dropped rather than seeding the next owner's buffer with the previous owner's rowids.

Line-Count-Exception: backend/utils/subscription.py | 1491 -> 1535 | the entitlement helper belongs beside `should_defer_desktop_processing`, whose tier semantics and fail-open contract it deliberately mirrors; splitting it out would separate two gates that must stay consistent

### DIFF
--- a/.github/failure-classes/FC-high-water-cursor-skips-unready-row.json
+++ b/.github/failure-classes/FC-high-water-cursor-skips-unready-row.json
@@ -1,0 +1,18 @@
+{
+  "schema_version": 1,
+  "id": "FC-high-water-cursor-skips-unready-row",
+  "violated_contract": "A monotonic high-water cursor may only advance past rows that are permanently ineligible, never past rows that are merely not ready yet. ScreenActivitySyncService selected `WHERE id > ? AND embedding IS NOT NULL` and advanced the cursor past everything it shipped, but embeddings are computed asynchronously after the row is inserted, so any row whose vector had not landed when the sweep passed was skipped forever — not retried, not queued, and not reported. The result reads as a working sync rather than a broken one: rows keep arriving, so no error, alert, or test fires, while 99.2% of captured rows never leave the device (measured: 705 of 26,822 rows over 30 days on a live store). Three further mechanisms independently denied a row its embedding, so repairing the cursor alone recovers little: cross-batch content-hash dedup returned early without queueing, gated or failed embedding batches were dropped rather than requeued, and the backfill that comments describe as the safety net had been stuck completed=1/processedCount=0 since 2026-04-16.",
+  "canonical_prevention": "Track delivery as durable per-row state rather than as a position, and make the expensive projection optional rather than a precondition. The row carries an indexed `screenActivitySyncState` (pending, textSynced, fullySynced, compacted) so a row that was not ready when a sweep passed is picked up by a later sweep; OCR-bearing rows ship without waiting for a vector, and the vector attaches later. Eligibility is evaluated per closed time bucket rather than per row, so a partially aged bucket is never ranked twice — the sliding form ships 44% more rows than the design intends. Gated and failed embedding batches requeue into a bounded buffer instead of being dropped, and the stuck backfill re-arms so stranded history is recoverable. Tests pin the core defect directly: a row unready at sweep time is still delivered later, an open bucket ships nothing, a closed bucket ships exactly one winner and never ships again, and migration is safe on a populated store.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/Desktop/Tests/ScreenActivityLosslessSyncTests.swift",
+    "desktop/macos/Desktop/Tests/OCREmbeddingLosslessTests.swift"
+  ],
+  "evidence_prs": [
+    5205
+  ],
+  "scope_hints": [
+    "desktop/macos/Desktop/Sources/ScreenActivitySyncService.swift",
+    "desktop/macos/Desktop/Sources/Rewind/Services/**"
+  ],
+  "status": "open"
+}
--- a/backend/database/firestore_index_registry.py
+++ b/backend/database/firestore_index_registry.py
@@ -660,6 +660,21 @@ INDEX_REQUIREMENTS = (
 )
 
 
+# Firestore auto-indexes every field of every document in both directions unless a field is
+# explicitly exempted. For large text fields that no query ever filters or orders on, that index
+# is pure storage cost: measured on `screen_activity`, single-field index bytes were roughly three
+# times the document bytes, and the `ocrText` index alone was the majority of the collection.
+# `ocrText` is only ever read back and rendered; semantic search runs on Pinecone vectors, not on
+# Firestore. Exempting a field only removes single-field indexes — composite indexes declared
+# above are unaffected, so a field named in a composite index can still appear here.
+FIELD_INDEXING_EXEMPTIONS: tuple[tuple[str, str], ...] = (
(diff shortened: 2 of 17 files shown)


Git-on-my-level:
Folded into #11865 so beta gets one merge rather than two.

Everything here is carried over verbatim, plus one gap that only showed up on merge: the `screen_activity_lossless_sync` flag was PostHog-only for every production-family bundle, so beta — the channel this exists to be exercised on — would have shipped it dark while the rest of the rollout was live there. All three client features now share one `BetaDogfoodRollout` decision, with a test that fails if a future feature ships past it.

No work is lost; this branch stays pushed for reference.

PR #11872, opened 2026-08-19
fix(backend): count explicit app selections, reject invalid ones, and make field exemptions appliable

Git-on-my-level:
Three follow-ups from the beta rollout. Each claim was validated before it was fixed — two held, one was partly wrong, and this says which.

## Explicit app selections were never counted

`record_app_usage` was skipped whenever `is_reprocess` was true, and client app selection goes through `POST /v1/conversations/{id}/reprocess` on both Flutter (`summarized_apps_sheet.dart` → `conversations.dart`) and macOS (`ConversationDetailView.swift` → `APIClient+Apps.swift`). So the one interaction opt-in mode exists to measure produced no usage history at all.

Attribution is now typed rather than inferred from a boolean. Recorded: automatic processing, and an explicit selection that actually executed. Not recorded: plain regenerate, sync reprocessing, and lazy enrichment of deferred desktop conversations.

**Correction to the original premise.** I claimed lazy enrichment reached this through `is_reprocess=True`. It does not — it passes `is_reprocess=False`. It is now marked non-user explicitly rather than relying on a flag that never distinguished it.

## An invalid `app_id` silently produced a successful no-app response

Confirmed for unknown, inaccessible, and non-summarization ids: the lookup returned no candidate, `apps_results` was cleared, and the route still returned 200. Under opt-in that is worse than it used to be — there is no default app to fall back on, so the caller asked for something specific, got a success, and silently received the first-party note.

Selection is validated before processing mutates anything:

| | |
|---|---|
| 404 | app does not exist |
| 403 | exists but is not available to this user |
| 400 | cannot summarize conversations (`works_with_memories()`) |
| 409 | disabled, or not enabled by this user |

No `app_id` — plain regenerate — is unchanged. `_trigger_apps` now raises if it is ever handed an unvalidated id, so route validation and execution cannot drift apart.

**Partly refuted.** I claimed an available-but-unenabled app degraded to "no app". It did not — `get_available_app_model_by_id` never enforced enablement, so such an app ran successfully. The 409 is therefore a new restriction rather than a fix. It is safe: both clients already enable an app before selecting it, and the macOS client does so with a comment naming this exact footgun.

Client handling was checked before choosing the codes: Flutter converts any non-200 into its existing `REPROCESS_FAILED` flow, macOS throws `APIError.httpError` and does not adopt the response, and none of these codes trip either client's 401 session-invalidation path.

## Firestore field exemptions were declared but nothing applied them

Confirmed. `reconcile_firestore_indexes.py` reads only `indexes` and shells out only to `gcloud firestore indexes composite`; the workflow invokes only that reconciler; there is no `firebase deploy` for indexes anywhere in the repo. The `fieldOverrides` that #11865 added saved nothing, and the storage reduction claimed there has not been happening.

Exempting a field **deletes** existing single-field indexes — precisely the destructive class the create-only reconciler refuses by design — and it targets the production data plane, which dev and prod share. So the create-only guarantee is untouched and a separate operation owns the deletion: manual dispatch only, prod-only (non-prod dispatch is rejected), gated on an exact `APPLY_FIRESTORE_FIELD_EXEMPTIONS` confirmation, dispatched-SHA-verified, serialized under the existing schema lock, and restricted to generated `ttl: false, indexes: []` overrides so it can never touch a composite index.

The automatic run now also **fails** when exemptions are declared but unapplied.

**Consequence worth knowing before merge:** this workflow will fail on main until that manual operation is run once. That is the intended signal — the alternative is the silent gap this PR exists to close — but it is a red workflow until someone acts on it, and the saving is not live until then.

## CI follow-up (second commit)

The first push failed the backend unit suite on two files — a **real regression, not a flake**. Adding `database.apps` and `utils.apps` as module-level imports broke `test_conversation_search_date_validation` (collection error) and `test_conversations_date_range_validation` (9 errors), deterministically.

Those suites load this router through the sanctioned isolation seam to test pure request validation, stubbing the heavy chains it otherwise pulls at import time. `utils.apps` reaches the memory and social graph, so a module-level import forces every such suite to stub a graph it never calls — and the next module added to that chain would break them again.

The router already solved this once: it imports from `utils.apps` inside a function around line 1510 for exactly this reason. The validator now follows that precedent, with a comment saying why so it does not get "tidied" back to the top. Its tests patch at the source modules instead of on the router.

Both CI-failing suites now pass (23 and 9), the selection suite still passes (8), and usage attribution, enrichment wiring, and the beta rollout contract are unaffected (37, 10, 4).

## Verification

142 tests across six suites, run per-file: app-selection route (8), usage attribution (37), field reconciler (6), firestore index config (9), composite reconciler (33), release vector (49). Contracts regenerated for the new response codes; only the OpenAPI and TypeScript projections changed.

Not verified: no live Firestore state was read and the exemptions are not applied by this PR. Swift was not compiled in the delegated sandbox, though no Swift source changed here.

Failure-Class: none

Line-Count-Exception: backend/routers/conversations.py | 1604 -> 1659 | the selection validator must run in the route, before `process_conversation` mutates the conversation; moving it elsewhere would reintroduce the validate-after-mutate ordering this PR fixes
Line-Count-Exception: backend/utils/conversations/process_conversation.py | 2221 -> 2251 | the attribution enum and its guard belong with `_trigger_apps`, the only place that decides whether an app execution is user-attributable; splitting them would let route validation and execution drift apart again


Git-on-my-level:
@cursor fix CI failures in a non-flaky way

Git-on-my-level:
**Hygiene: `product-file-line-count-ratchet` fixed — PR-body metadata correction, no code change.**

The two `Line-Count-Exception` declarations had gone stale after the last two pushes: they declared `1605 -> 1660` and `2225 -> 2255`, while the synthetic-merge diff for head `42b25149` is `1604 -> 1659` and `2221 -> 2251` (reasons unchanged).

Verified before updating: reproduced the exact CI failure locally by running `check_product_file_line_count_ratchet.py` from `origin/main` against the PR head and the same base CI used (`6a4d3d18`), then confirmed the corrected body also passes against the newest `origin/main` (`d5a0ae95`). Reran the failed `Hygiene` job — now green, and the PR has zero failing or in-progress checks.

undivisible (commented):
## Review

The six named backend suites pass on the live head (114 tests under the pinned Python 3.11 environment).

The destructive field-exemption path is correctly constrained in this tip: manual `workflow_dispatch` on `main`, `operation=field-exemptions`, `environment=prod`, protected `prod` Environment, exact `APPLY_FIRESTORE_FIELD_EXEMPTIONS` confirmation, dispatched-SHA verification, and a reconciler limited to generated `ttl: false, indexes: []` overrides using `gcloud firestore indexes fields update --disable-indexes`. Composite indexes stay on the separate create-only reconciler.

Keep this as a human-ops hold: the automatic path is intentionally red until the one-time prod apply is run and verified, so I am not approving or merging this PR.

PR #11890, opened 2026-08-19
fix(desktop): keep named bundle settings in sync

Git-on-my-level:
## Summary

- resync the curated Omi Dev preference allowlist on both full and executable-only fast named-bundle launches
- mirror source preference absence by deleting stale named-bundle overrides, so compiled shortcut defaults stay aligned
- fail closed on preference export, import, deletion, or verification errors while retaining an explicit bundle-local test hatch
- add exact hotkey payload, missing/empty domain, stale-key, and common launcher-path regression coverage wired into the check manifest
- repair the ranked Rewind embedding CTE so the broad desktop Swift suite can exercise the existing backfill recovery test

## Root cause

The executable-only fast path converged after the settings-seed block, so a reused named bundle never saw later Omi Dev setting changes. The seed helper also overlaid only source-present values, leaving stale target hotkey overrides in place when Omi Dev had fallen back to a compiled default.

## Verification

- `bash desktop/macos/tests/test-settings-seed.sh`
- focused launcher tests: app icon, local dev entitlements, launch env forwarding, fast dev bundle, and settings seed
- `.github/scripts/test_run_checks.py` (46 tests)
- desktop test-quality, AGENTS budget, changelog, and launcher-discovery guards
- `./scripts/agent-logic-harness.sh --cross-surface-smoke`
- focused `ScreenshotEmbeddingBackfillRecoveryTests` Swift run after repairing the missing `embedding` CTE projection
- real `/Applications/omi-hotkey-parity.app` full build followed by `--fast-only`: two injected stale hotkey payloads were cleared; `/health` reported `bundleIdentifier=com.omi.omi-hotkey-parity` and `ok=true`
- independent sub-agent review after fixes: approved with no remaining findings

none

## Failure class

Failure-Class: new


Git-on-my-level:
Rebased onto current `main` (`ca29a83d`) and force-pushed the rebased head (`d0c6648b`); the branch had been 9 commits behind. All checks now pass except the two Desktop Swift checks, which fail on `ScreenshotEmbeddingBackfillRecoveryTests.testCompletedZeroProgressBackfillRearmsWhenCompactedWinnerIsMissing` with `SQLite error 1: no such column: embedding` in the `WITH ranked` CTE.

This failure is not introduced by this PR — the diff touches only launcher/settings-seed shell scripts, the checks manifest, and a changelog entry (no Swift). It reproduces deterministically on `main`: the last `main` run that actually executed the desktop suite (`b0544938`) failed with the identical error, and every green `main` run since then skipped the desktop job. #11885 projects `embedding` into the compacted backfill CTE — the exact statement that fails to prepare — so these checks should go green once #11885 lands, or after rebasing onto a `main` that includes it.

PR #11891, opened 2026-08-19
fix(desktop): stop capture recovery from restarting the app during onboarding

kodjima33:
## What happened

David's first run on a new Mac (0.12.187 stable, macOS 15.1, Mac16,6): Omi quit by itself several times during onboarding — including on permission pages that have nothing to do with screen recording — and did not come back on its own.

His PostHog session shows it plainly:

```
21:11:43  App Launched          <- first run
21:12:28  Onboarding How Did You Hear
21:12:49  App Launched          <- quit + relaunch #1
21:13:13  App Launched          <- #2
21:13:34  App Launched          <- #3
21:14:35  Onboarding Completed
```

## Cause

A new Mac has not granted screen recording yet, so every capture attempt fails. `ProactiveAssistantsPlugin`'s failure tracker reads that as a broken install and calls `attemptAutoReset()`, whose final step is `ScreenCaptureService.softRecoveryAndRestart()` — it terminates and relaunches the process. The user sees the app vanish mid-permission-page with no explanation.

It repeats because the "already tried recovery this session" flag (`hasSoftRecoveryThisSession`) is static state in the process that just got killed, so the fresh process is free to do it again.

## Fix

`ScreenRecordingPermissionPolicy.mayRestartToRecoverCapture(grantedAtLaunch:onboardingComplete:)` — the process-killing step now requires both:

- **the grant is live in this process** — ScreenCaptureKit binds to the window-server connection opened at launch, so if the grant was not there at launch, no number of restarts can repair it; and
- **onboarding is finished** — during onboarding, relaunch belongs to the "Reopen Omi" prompt, which is scoped to screen recording and advances the persisted step first.

Soft recovery still runs; only the restart is withheld, with the reason logged.

Failure-Class: none

## Verification

- Built as `omi-onbfix` with no screen-recording grant and `hasCompletedOnboarding=false` — David's exact state. The app held a single pid across the observation window, with no `Restart scheduled` in its log.
- `swift test --filter ScreenRecordingPermissionPolicyTests` — **22 passed**, including `testCaptureRecoveryNeverRestartsDuringOnboardingOrBeforeTheGrantIsLive` and a loop test that models the per-process flag resetting on each relaunch (the reason it happened three times rather than once).

Line-Count-Exception: desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift | 1786 -> 1798 | the recovery-restart guard plus the log line naming why a restart was withheld; the decision itself lives in ScreenRecordingPermissionPolicy

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11891?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift
@@ -1710,6 +1710,11 @@ public class ProactiveAssistantsPlugin: NSObject {
   /// Attempt automatic recovery (soft first, then hard reset as last resort)
   private func attemptAutoReset() {
     let ownerID = RuntimeOwnerIdentity.currentOwnerId()
+    // A restart cannot fix a grant that is not live in this process, and must never happen
+    // while the user is still walking through onboarding — see mayRestartToRecoverCapture.
+    let mayRestart = ScreenRecordingPermissionPolicy.mayRestartToRecoverCapture(
+      grantedAtLaunch: ScreenCaptureService.grantedAtProcessStart,
+      onboardingComplete: UserDefaults.standard.bool(forKey: .hasCompletedOnboarding))
     // Step 1: Try soft recovery first (lsregister + SCK re-request, no TCC wipe)
     if !Self.hasSoftRecoveryThisSession {
       Self.hasSoftRecoveryThisSession = true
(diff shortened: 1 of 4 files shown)


undivisible (changes requested):
## Request changes

The onboarding guard is focused, but this diff adds and rewrites multiple comments despite the repository rule to preserve existing comments verbatim and add no new comments.

### Please
1. Remove the new explanatory comments from the production policy/plugin and the two new test comments.
2. Restore any edited existing comments verbatim; leave the policy, tests, and changelog behavior unchanged.

Issue #11307, opened 2026-08-10
iOS background BLE timeouts leave live transcription unavailable for minutes to hours


axAilotl:
A full-day trace from the production iOS app with a freshly paired stock CV1 on firmware 3.0.21 recorded seven native CoreBluetooth `connection_timeout` failures, all while `app_state=background`. The observed reconnect gaps ranged from about 3.9 seconds to 7,335.3 seconds (roughly 2 hours 2 minutes), including another gap of about 929.6 seconds; no new pairing-loss events occurred. Durable device storage prevented audio loss, but live transcription remained unavailable until the app returned and recovered. Expected behavior is for the iOS restoration/background route to retain or promptly re-establish the BLE audio path without requiring a foreground app relaunch or pendant reset.

mdmohsin7:
Was the device in range when it was not reconnected for 2 hours? Because ideally iOS would automatically reconnect if its in range (since we use state restoration)

axAilotl:
@Git-on-my-level — need human response. Clarifying the range question and the product contract, because the original evidence set mixed two kinds of testing.

The earlier seven long intervals include both deliberate out-of-range excursions and failures experienced locally/in range. Those movement boundaries were not timestamped precisely enough to claim that the full two-hour interval was an in-range BLE reconnect delay. It is equally incorrect to dismiss the issue merely because the user sometimes left range.

A separate clean two-hour in-range control on 2026-08-11 showed **zero BLE transitions** and a continuously connected/ready BLE audio path while the remote transcription service failed repeatedly. That specific result belongs to #11306 and demonstrates an independent remote-STT failure. It does not remove the BLE recovery requirement here.

### Expected behavior

Going in and out of phone range is normal pendant use, not an invalid test condition:

- **Out of range:** the pendant must continue durable offline recording without losing audio.
- **Back in range:** iOS state restoration/reconnect must reclaim the peripheral automatically, restore audio notification flow, resume live transcription, and drain/sync the offline backlog.
- Recovery must not require relaunching the app, resetting/power-cycling the pendant, forgetting/re-pairing Bluetooth, or keeping the phone screen open.
- The end-to-end invariant is reliability regardless of current connectivity: live when connected, durable capture while disconnected, and automatic convergence after reconnection.

### Current-main code path relevant to missed background recovery

[...]

PR #11858, opened 2026-08-19
fix(backend): prevent large list GETs from exhausting the 30s request budget

Git-on-my-level:
## Summary

`GET /v1/action-items`, `GET /v1/conversations`, and `GET /v3/memories` could each drive aggregate Firestore work past `HTTP_GET_TIMEOUT` (30s) and come back as a bare middleware 504 on large accounts (#11831). This PR makes those three reads budget-bound: they either land, or return an honest partial page with an explicit truncation surface.

Closes #11831

## Design

- **Request-scoped budget** — `utils/other/list_budget.py` owns a monotonic internal deadline (24s default: the 30s middleware cutoff minus serialization headroom), a 25k-document allowance, aggregate counters, and a typed `ListReadBudgetExhausted`. Created once per request at the route from the `TimeoutMiddleware`-stamped request start (`request.state.omi_request_started_monotonic`), so pre-handler time (auth, middleware) counts. `TimeoutMiddleware` still owns the hard 30s cutoff — unchanged.
- **Wall clock interrupts the blocking RPC** — every Firestore `stream()` / `get_all()` / `document.get()` on these routes receives `timeout=budget.rpc_timeout()` derived from the remaining internal deadline. A blocked RPC raises `DeadlineExceeded` at the internal deadline, leaving ≥6s headroom for serialization; only that exception (and allowance exhaustion) maps to the typed outcome. Loop-side `monotonic()` checks alone cannot stop a blocked `next()` — the RPC timeout is what does.
- **Charge actual aggregate work** — conversations charge the server-side `offset()` prefix *before* the query (Firestore bills and streams every skipped row); action-items share one budget across the active, legacy-harvest, and completed queries; memories charge both dual-window index scans plus emitted and suppressed rows, the canonical stream, status batches, and hydration.
- **Fallback shares the budget** — the `/v3/memories` first-page scan-budget 503 (`Memory scan budget exceeded`) still falls back to the legacy offset `read()`, now on the *same* request budget; it can never start a fresh unbudgeted 5000-row window.
[...]

PR #11892, opened 2026-08-19
fix(app): reduce and diagnose iOS background energy

Git-on-my-level:
## What changed and why

iOS battery reports showed Omi consuming 29–61% with 6h36m–17h32m of background activity, matching the reconnect/write amplification reported in #11307. This removes idle background owners, bounds resource-only loops, and emits one privacy-safe summary per meaningful background session so future regressions can be attributed to BLE traffic, native offline writes, WebSocket traffic, or reconnect churn.

none

## How it was verified

- `bash test.sh` — all 1,384 Flutter tests passed (5 skipped).
- `app/scripts/analyze_ratchet.sh` — passed with no new analyzer findings.
- `ruby app/ios/test/omi_ble_energy_policy_test.rb` — passed the native RSSI and battery-history policy assertions.
- `xcodebuild -workspace app/ios/Runner.xcworkspace -scheme raybanDat -configuration Debug-dev -sdk iphonesimulator -destination 'generic/platform=iOS Simulator' CODE_SIGNING_ALLOWED=NO ASSETCATALOG_COMPILER_APPICON_NAME= build` — succeeded and compiled `AppDelegate`, `OmiBleManager`, `OmiBleEnergyPolicy`, and the Pigeon bridge for arm64 and x86_64.
- Independent Luna review found the native batch-counter blind spot, the stop-during-codec-lookup race, manual-disconnect overcounting, and device-swap attribution risk; each was reproduced against the production path and addressed before this verification pass.

The default `dev` scheme is currently blocked before Runner compilation by the existing `omiWatchApp` asset/`WatchKit` configuration. The paired iPhone 15 Pro Max was listed as unavailable, so a multi-hour physical-device background-energy run was not possible in this session; this PR does not claim that validation.

## Tests

- `app/test/utils/analytics/background_resource_telemetry_test.dart` covers session duration, byte deltas, native offline-writer counters, device-stable diagnostics, reset clamping, and fail-open analytics.
- `app/test/providers/capture_provider_test.dart` covers idle keepalive suppression, active device/system-audio reconnects, first-transcript location capture without the foreground isolate, and stopping while codec lookup is in flight.
- `app/test/services/capture/capture_metrics_tracker_test.dart` covers pausing the UI sampler while lifetime counters continue.
- `app/ios/test/omi_ble_energy_policy_test.rb` covers diagnostic-only RSSI polling and bounded battery-history persistence.

## Failure class (fixes)

Failure-Class: none

The `ios-ble-energy-policy` manifest check would have caught #11307's always-on RSSI polling and notification-rate battery-history rewrites. This is not a shared primitive because the guarded behavior is specifically CoreBluetooth polling and iOS `UserDefaults` persistence inside the native BLE owner.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Monitor fix pass (auto-generated summary):

- Ran `dart format --line-length 120` over the PR's changed non-generated Dart files and pushed as `58340d5d05` (branch rebased onto current `origin/main` first, so this was a force-with-lease update to `codex/ios-background-energy-telemetry`).
- The **Formatting** check now passes on the new head.
- The two failing Desktop Swift suites (`RewindAbandonedChunkRecoveryTests`, `ScreenshotEmbeddingBackfillRecoveryTests`) are unchanged by this PR — this branch touches no `desktop/` files, and the same suites fail on desktop-untouched PRs (#11888 backend-only, #11858 backend-only) in the same window, so they are repo-wide and not addressed here.

undivisible (commented):
## Review

The focused smoke is green on the live head: the Ruby BLE energy-policy test passed, the background telemetry tests passed (3), and the capture/device/metrics tests passed (101).

Residual validation is still material for this energy-sensitive change: the Watch scheme is blocked by its current asset/WatchKit configuration, and no paired iPhone was available for a multi-hour background run. Simulator/build and unit coverage establish wiring and bounded behavior, but not real BLE reconnect/write amplification or battery drain on device. I’m leaving this review-only until physical-device evidence covers the iOS path and the Watch path if it remains in release scope.

kodjima33 (approved):
Real iOS background-energy regression (#11307) with native + Flutter tests and analyzer clean; approve-only since the diff is well over the 300-line merge gate for mobile.

PR #11896, opened 2026-08-20
fix(macos): recognize the browser titles users actually have

Archit-lal:
## What this fixes

Integration nudges recognize a site from the browser's window title. The shipped
rule is "the title ends with the site's name". I measured it against **95,577
real titles** read out of my own Chrome and Arc history, and both halves of the
rule are wrong against titles browsers actually produce.

| site | visits in corpus | recall before | recall after |
|---|---:|---:|---:|
| gmail | 16,988 | 45.9% | **96.6%** |
| x | 4,177 | 91.5% | 91.5% |
| chatgpt | 2,798 | 75.4% | 75.3% |
| claude | 257 | 91.1% | 91.1% |
| gemini | 220 | 99.5% | 99.5% |

### The miss: Gmail on Google Workspace

On a Workspace domain the mailbox is titled after the organization —
`Inbox (3,012) - you@company.com - University of Minnesota Twin Cities Mail` —
so the word "Gmail" never appears. That is **51% of all Gmail visits in the
corpus**, and every one of them was invisible to the nudge. Any user on a work
or school Google account was in this hole.

A new `gmail_workspace_web` trigger recognizes the shape by the account address
Gmail puts in the middle segment. Requiring the address is what makes it Gmail
rather than webmail in general — a bare `" Mail"` suffix would claim Proton and
Yahoo too, and there is a test for that.

### The wrong claims: "ends with" reads prose as a site

An unanchored suffix test reads any sentence that trails off in a site's name as
the site itself. Real titles from the corpus that the shipped rule claims:

- `How to Create Studio Ghibli Style Art With ChatGPT` → nudges for ChatGPT
- `Context for Claude` → nudges for Claude
- `World Wealth Report 2024: HNWI Wealth Management | Capgemini` → nudges for Gemini
- `Website Tweak for Gemini` → nudges for Gemini

The site name now has to be the **whole title** or sit **behind a separator**.
That drops all four and costs no real title. A one-character name never stands
alone either: `X` as an entire title belongs to a Stripe checkout page here, so
X has to arrive with its separator, as it does in every real x.com title.

### Two things the same measurement turned up

**Gemini titles itself `‎Google Gemini`** — behind an invisible left-to-right
mark. It survived only because the unanchored rule happened to match the last
six letters. Anchoring without also stripping the mark and naming the site
correctly would have silently dropped Gemini from 99.5% to 59.5%.

**The "observed titles" fixture held invented titles.** `Reviewing a diff \
Claude` and `Swift concurrency question - ChatGPT` do not occur once in 95,577
titles. That fixture's own comment warns against exactly this ("Synthesizing a
title out of the keyword … would match by construction and prove nothing"). They
are replaced with observed shapes; only account addresses are anonymized.

## Verification

- `swift test --filter IntegrationNudge` → **92 tests, 0 failures**
- **Exercised in a real build.** Installed as a named test bundle
  (`omi-title-fix`) and driven through the running app's automation bridge
  (`integration_nudge_evaluate`), which runs the real matcher in-process:

  ```
  MUST MATCH
    Inbox (3,012) - me@umn.edu - University of Minnesota Twin Cities Mail
                                            → Gmail    gmail_workspace_web
    Inbox (16,993) - me@gmail.com - Gmail   → Gmail    gmail_web
    (3) Home / X                            → X        x_web
    ChatGPT                                 → ChatGPT  chatgpt_web
    Monthly recurring revenue … - Claude    → Claude   claude_web
    ‎Google Gemini                          → Gemini   gemini_web

  MUST STAY SILENT
    How to Create Studio Ghibli Style Art With ChatGPT     → no match
    Context for Claude                                     → no match
    World Wealth Report 2024: … | Capgemini                → no match
    Website Tweak for Gemini                               → no match
    X                                                      → no match
    Inbox - Proton Mail                                    → no match
  ```

  The Workspace title reports `would_nudge: true`. The card was then delivered
  for real (`outcome: delivered`, `presented: true`) and the store recorded it
  (`shownCount = 1`).

- Full desktop suite on this branch: **5,591 tests, 2 failures**. The same suite
  on a pristine `upstream/main` checkout (6e1eb7c05f): **5,586 tests, 2
  failures** — the identical pair. The +5 is this PR's new tests.

### On the two red CI checks

Neither is from this diff, which touches four `IntegrationNudges/` files and a
changelog fragment.

- `RewindAllTimeSearchTests` — a flake. It failed once on
  `testPanningReloadsOnlyTheRequestedContinuousWindow` (an async-ordering
  assertion) and **passed on re-run**.
- `ScreenshotEmbeddingBackfillRecoveryTests` — **already broken on `main`**,
  and not by this PR. It fails with `SQLite error 1: no such column: embedding`.

  `main`'s recent Desktop Swift CI runs are green only because the job was
  **skipped** — no main commit since 20:51 touched desktop Swift, so nothing
  re-ran it. Walking back to the runs that actually executed the job:

  ```
  06868f2a46  18:05  success   ← last green
  091a65bef4  18:40  failure   ← feat: notes v2, lossless screen sync,
                                  and paid-tier screen vectors (#11865)
  7d7157b6ab  20:44  failure
  20f17ac5e7  20:46  failure
  b0544938e6  20:51  failure   ← #11729 merged red on this same test
  ```

  #11865 changed `RewindDatabase.swift` and `RewindDatabase+Embeddings.swift`;
  the breakage lands between the last green run and the first red one, well
  before #11729. This PR is simply the first change since to touch desktop Swift
  and therefore the first to run the job again.

- New behavioral tests: `testAPageAboutASiteIsNotThatSite`,
  `testWorkspaceMailboxesAreRecognizedAsGmail`, `testOtherWebmailIsNotClaimedAsGmail`,
  `testASingleCharacterNameIsNeverTheWholeTitle`,
  `testGeminiIsMatchedThroughItsInvisibleLeadingMark`.

## Known limits, stated plainly

- **ChatGPT web stays at ~75%.** 2,108 of 2,798 chatgpt.com visits are titled
  exactly `ChatGPT` (the new-chat page, where every session starts); the rest
  carry the conversation's own name and are unrecognizable by title. A prefix
  rule would pick up the marketing landing page and `ChatGPT - <project>`, but
  it also claims `chatgpt - Google Search`, so it is not worth it. Fixing the
  tail needs the URL, which needs Accessibility-driven address-bar reading — a
  separate decision, not this PR.
- **claude.ai stays at ~91%** for the same reason: some conversations title
  themselves without the site name.
- The corpus is one developer's browsing history. It is a large real sample, not
  a representative one.

## Not addressed here

Two things I flagged on #11729 remain open and are **not** in this PR:

1. A fresh install reads account-connected integrations as unconnected
   (`hasEverSynced` is local-only), so it can nudge for something already
   connected on another machine.
2. The connect-then-reopen loop ("already connected stops nudging") is proven by
   unit tests and by `already_connected` suppression on Local Files, but was
   never driven end-to-end through a real OAuth connect — completing those
   sign-ins is not something I can do on the user's behalf.

Failure-Class: FC-meeting-trigger-title-identity-drift

### DIFF
--- a/desktop/macos/Desktop/Sources/IntegrationNudges/IntegrationNudgeCatalog.swift
+++ b/desktop/macos/Desktop/Sources/IntegrationNudges/IntegrationNudgeCatalog.swift
@@ -40,15 +40,21 @@ enum IntegrationNudgeRoute: Equatable, Hashable {
 enum IntegrationNudgeTriggerMatch: Equatable {
   /// The frontmost application is one of these bundle identifiers.
   case application(bundleIdentifiers: [String])
-  /// The frontmost application is a browser and its window *title* ends with one
-  /// of these suffixes, case-insensitively.
+  /// The frontmost application is a browser and its window *title* ends with the
+  /// site's own name, case-insensitively.
   ///
   /// A window title is the page's `document.title`, never its URL, so a domain
   /// keyword matches nothing real. What is reliable is where sites put their own
-  /// name: at the end. X renders "Home / X", Gmail "Inbox (12) - you@corp.com -
-  /// Gmail", ChatGPT just "ChatGPT". Anchoring to the end is what separates
-  /// those from "ChatGPT vs Claude — a comparison", which is the false positive
-  /// that matters — a wrong nudge is worse than a missing one.
+  /// name: at the end, behind a separator. X renders "Home / X", Gmail "Inbox
+  /// (12) - you@corp.com - Gmail", ChatGPT just "ChatGPT".
+  ///
+  /// The name must be the *whole* title or sit behind a separator — a bare
+  /// "ends with" test reads any sentence that happens to trail off in the name
+  /// as the site itself. Measured against 95,577 real titles from this
+  /// developer's own browser history, the unanchored form claimed "How to Create
+  /// Studio Ghibli Style Art With ChatGPT" as ChatGPT, "Context for Claude" as
+  /// Claude, and "World Wealth Report 2024: HNWI Wealth Management | Capgemini"
+  /// as Gemini. Requiring the separator drops all three and costs no real title.
   ///
   /// An integration whose real titles carry no site name gets no browser trigger
   /// at all rather than a guess. Notion web titles are the page name alone, and
@@ -59,7 +65,17 @@ enum IntegrationNudgeTriggerMatch: Equatable {
   /// the same active-window lookup the proactive assistants already use, and
   /// reading the address bar would mean driving the Accessibility API across
   /// every browser.
-  case browserTitleSuffix(suffixes: [String])
+  case browserTitleSite(names: [String])
+  /// The frontmost application is a browser showing a Gmail mailbox whose
+  /// account is on a Google Workspace domain.
+  ///
+  /// Workspace replaces "Gmail" in the title with the organization's own name —
+  /// "Inbox (3,012) - you@company.com - Acme Mail" — so `browserTitleSite` can
+  /// never name it. That is not an edge case: in the same 95,577-title corpus it
+  /// is 51% of all Gmail visits, and every one of them was invisible to the
+  /// nudge. The shape that identifies it is the account address Gmail puts in
+  /// the middle segment; " Mail" alone would claim Proton and Yahoo too.
+  case browserTitleGoogleWorkspaceMailbox
 }
 
 struct IntegrationNudgeTrigger: Equatable {
@@ -71,7 +87,7 @@ struct IntegrationNudgeTrigger: Equatable {
   var kind: IntegrationNudgeTelemetry.TriggerKind {
     switch match {
     case .application: return .nativeApp
-    case .browserTitleSuffix: return .browserSite
+    case .browserTitleSite, .browserTitleGoogleWorkspaceMailbox: return .browserSite
     }
   }
 }
@@ -176,7 +192,11 @@ enum IntegrationNudgeCatalog {
       triggers: [
         IntegrationNudgeTrigger(
           id: "gmail_web",
-          match: .browserTitleSuffix(suffixes: ["Gmail"])
+          match: .browserTitleSite(names: ["Gmail"])
+        ),
+        IntegrationNudgeTrigger(
+          id: "gmail_workspace_web",
+          match: .browserTitleGoogleWorkspaceMailbox
         ),
         IntegrationNudgeTrigger(
           id: "gmail_client_app",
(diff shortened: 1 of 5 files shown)


Git-on-my-level:
Thanks @Archit-lal — this is a careful piece of work. Reviewed against the five changed files:

**IntegrationNudgeCatalog.swift** — the `browserTitleSite` docs now state the anchor rule precisely, and the new `browserTitleGoogleWorkspaceMailbox` case (wired as the `gmail_workspace_web` trigger on the Gmail entry) cleanly separates the two recognition problems: site-name anchoring vs. Workspace's org-named mailbox.

**IntegrationNudgeMatcher.swift** — `endsWithSiteName` requiring the name to be the whole title or sit behind one of `titleSeparators` (` - `, ` – `, ` — `, ` | `, ` / `) kills the prose false positives ("Context for Claude", "… | Capgemini", "… With ChatGPT") while "Home / X" still matches, and the `name.count > 1` guard correctly refuses a bare "X" title. `isGoogleWorkspaceMailbox` requiring both the " mail" suffix and an `@` segment with a dotted domain is the right two-factor shape — Proton/Yahoo/Outlook stay unclaimed. Keeping `: ` out of the separator list is well-reasoned (sites lead with it rather than close with it).

**IntegrationNudgeCatalogTests.swift** — replacing the two invented fixture titles ("Reviewing a diff \ Claude", "Swift concurrency question - ChatGPT") with corpus-observed shapes fixes exactly the "matches by construction" trap the fixture's own comment warned about.

**IntegrationNudgeMatcherTests.swift** — the new negative tests (pages about a site, single-character "X", other webmail) and the Gemini invisible-mark test cover the risky side of each heuristic, not just the happy path.

**Changelog entry** — matches the sibling unreleased entries' format.

One non-blocking observation: `isGoogleWorkspaceMailbox` would also claim a hypothetical non-Gmail page titled like "Dashboard - billing@vendor.com - Acme Mail" (email segment + " Mail" suffix). Your corpus measurement says that shape doesn't occur for sites users actually visit, and the cost is one mistimed suggestion card, so this seems fine as shipped — just worth remembering if nudge feedback ever mentions Gmail.

About the two red desktop checks: both fail in `ScreenshotEmbeddingBackfillRecoveryTests` with `SQLite error 1: no such column: embedding` — a RewindDatabase backfill-CTE bug that exists on this branch's merge-base and was fixed on main in #11885 after this branch was pushed. None of this PR's files are involved; bringing the branch up to date with main (a merge will pick the fix up too) should turn the desktop suite green.

Leaving final merge sign-off to a human maintainer — these recognition rules ship to every macOS user, and the change is cohesive, measured against real titles, and well-tested.

---

kodjima33 (approved):
Well-measured fix: the 'title ends with site name' rule misses 51% of Gmail visits on Workspace domains; recall data from 95k real titles backs the new matcher. Holding the merge only because the desktop Swift CI lane is red on main.

kodjima33 (approved):
Rigorous, well-evidenced browser-title-recognition fix (95k-title corpus). Already approved; re-affirming. Not merging: Desktop Swift Static & Test Contracts is currently red on this branch (same workflow-contract failure as other same-vintage PRs), holding at approve per CI hard floor.

PR #11899, opened 2026-08-20
fix(backend): a deleted chat session must not 500 the write it was linking

kodjima33:
## Bug

Prod (image `920fc55`, `2026-08-20T00:49:53Z`): `DELETE /v2/messages` returned **500 after 31.5s**, from

```
File "/app/routers/chat.py", line 665, in clear_chat_messages
    return initial_message_util(uid, compat_app_id)
File "/app/utils/chat.py", line 97, in initial_message_util
    chat_db.add_message_to_chat_session(uid, chat_session['id'], ai_message.id)
File "/app/database/chat.py", line 710, in add_message_to_chat_session
    session_ref.update({"message_ids": firestore.ArrayUnion([message_id])})
google.api_core.exceptions.NotFound: 404 No document to update:
    .../users/<uid>/chat_sessions/78489909-6ff7-41de-b292-0fedd332e1ba
```

Nine occurrences on `chat_sessions` since 2026-08-01; user-visible as a failed "clear chat".

## Root cause

The race is the endpoint's own latency. `clear_chat_messages` reads the session, deletes it, then
calls `initial_message_util`, which acquires/creates a session and spends the **whole LLM round
trip** generating the greeting before linking the new ai message back onto the session document.
That route measured 10–80s in prod. A second clear — a double tap, or a client retrying a request
that already looked hung — lands inside that window and deletes the session the first call is still
holding, so the first call's link write hits a tombstone.

Confirmed in the request log: the 500 started `00:49:21.99` (31.46s latency → error at `00:49:53.45`,
an exact match) and a second `DELETE /v2/messages` started `00:49:47`, inside it.

The clear itself, and the message write the link was reporting, had both already committed — so the
user got a 500 for work that succeeded, exactly the shape of #11833
(`update_folder_conversation_count`). The same window is open on the send-message path, where the
link write also trails a multi-second model call.

## Fix

`message_ids`, `file_ids` and the OpenAI thread/assistant ids are derived state the session document
owns; a session that no longer exists has nothing to record. The guard sits inside that maintenance
boundary (`_update_chat_session_if_exists`), so every caller of the three writers is covered at once
rather than each call site growing its own `try`. `update_chat_session_openai_ids` only logs its
success line when the write landed.

## What I tested

- **The exact prod path**: `utils.chat.initial_message_util` with the session deleted *during* the
  model call, real `database.chat`, only the Firestore client faked — returns the greeting with the
  fix, raises the identical `NotFound` without it.
- **7 new db-layer regression tests** (`tests/unit/test_chat_session_link_deleted_session.py`,
  0.03s). The three deleted-session cases fail on unfixed source with the prod exception; a control
  test keeps the fake's `update()` raising on a missing document, so the passes prove the guard and
  not a lenient fake.
- **57 chat/session/message unit files** run file-by-file the way `test.sh` invokes them: all green.
- `make preflight` green; `scripts/failure-class validate` OK.

Not exercised: a live `DELETE /v2/messages` against prod Firestore — reproducing it needs two
concurrent clears on a real account, so the router frame above `initial_message_util` (a direct
`return`) is covered by reasoning, not by execution.

Failure-Class: FC-post-commit-index-maintenance-terminal

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11899?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/chat.py
+++ b/backend/database/chat.py
@@ -6,7 +6,7 @@ import uuid
 from datetime import datetime, timezone
 from typing import Any, Dict, Iterator, List, Optional, cast
 
-from google.api_core.exceptions import AlreadyExists, Conflict, FailedPrecondition
+from google.api_core.exceptions import AlreadyExists, Conflict, FailedPrecondition, NotFound
 from google.cloud import firestore, firestore_v1
 from google.cloud.firestore_v1 import FieldFilter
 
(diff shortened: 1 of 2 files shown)


undivisible (approved):
## Approve

Focused deleted-session regression passed locally: 7 tests passed; backend, hermetic, image, OpenAPI, formatting, and hygiene checks are green.

PR #11900, opened 2026-08-20
feat(desktop): make Home prove Omi's value

aadivb9:
## Summary

- turn Home into a one-question doorway to Omi's real value instead of an architecture explainer
- carry the answer produced during onboarding into Home as an ephemeral proof receipt, then offer the most useful screen-aware follow-up
- give returning users a personalized primary question and a small, truthful view of recent conversations, memories, and tasks
- keep existing transcripts, the shared composer, every onboarding screen, and every dashboard destination intact
- collapse history search into a clear `Search history ⌘K` command so it no longer competes with the primary question

### New users

Onboarding still uses the existing screen sequence, but the promise is now explicit: Omi connects what it can hear, see, and do, with every permission explained. During the screen demo, the answer produced in that isolated stage is captured in memory only. Home opens with **You asked. Omi answered.**, shows a bounded excerpt of that answer, and leads with **What should I focus on based on what's on my screen?**

If an answer was not produced, Home falls back honestly to the setup-aware greeting and questions already derived from onboarding. It never invents context or renders a wall of zero counts.

### Returning users

Home leads with one personalized question when Omi has one. Under it, users see up to three real recent context items already loaded by the app: a conversation, an open task, and a memory. Existing conversations reopen directly as conversations instead of being covered by the overview.

The resting surface has one obvious text input: the existing shared chat composer. History search stays available from the panel header or `⌘K`, and expands into the unchanged search experience only when requested.

## Trust and continuity

- answers state that citations appear when context is used; no source provenance is fabricated
- the onboarding proof receipt is bounded, whitespace-normalized, and never persisted
- screen counts and asynchronous calendar enrichment are owner-authorized before publication
- an account switch synchronously clears the prior owner's proof, opener, transcript, and draft state
- prompt refreshes freeze while the overview is mounted, and suggestions never overwrite a draft

## Verification

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --package-path desktop/macos/Desktop --skip-build --filter '<affected onboarding, query, permission, and Home suites>'` — **259 tests passed**
- post-extraction focused compile/test — **45 tests passed**
- completed-turn proof policy compile/test — **28 tests passed**
- `./desktop/macos/scripts/desktop-core-harness.sh --self-check --skip-backend-contracts` — **passed**; 72 flows and 159 registered actions linted
- `/Users/aadigupta/Documents/GitHub/omi/backend/.venv/bin/python -m pytest -q tests/unit/test_desktop_agent_vm.py tests/unit/test_desktop_chat.py tests/unit/test_desktop_core.py tests/unit/test_desktop_proxy.py tests/unit/test_desktop_realtime.py tests/unit/test_desktop_screen_crisp.py tests/unit/test_desktop_tts_updates.py` — **193 tests passed**
- built, ad-hoc signed, dependency-verified, and ran exactly one isolated `/Applications/omi-track0-value.app`
- visually exercised the real first-run promise and completed-onboarding Home at the 900×712 window size with a synthetic local account; production Omi bundles were untouched
- `make preflight` — passed with the PR body below supplying the required invariant citations

## Full-suite notes

The affected suites are green. A prior full Swift run exposed two unrelated failures already present outside this change: multi-monitor quiet-window geometry in `DesktopAutomationWindowPresentationTests` and an outdated SQLite fixture missing `embedding` in `ScreenshotEmbeddingBackfillRecoveryTests`. The official desktop runner's pytest handoff also has an existing trailing-continuation issue; the direct desktop backend selection above is green.

- INV-AUTH-1
- INV-CHAT-1
- INV-UI-1

## Failure class

Not applicable; all commits are `feat:` changes.

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/QueryShell/QueryResultsPanel.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/QueryShell/QueryResultsPanel.swift
@@ -195,7 +195,7 @@ struct QueryResultsPanel<Content: View, Accessory: View, Footer: View>: View {
   private var countSentence: String {
     guard let total else { return "Counting what you've captured…" }
     if mode == .answer {
-      return "Answered from \(QueryShellCount.number(total)) captured moments"
+      return "\(QueryShellCount.number(total)) captured moments ready as context"
     }
     return QueryShellCount.sentence(
       matching: matching, total: total, isFiltering: request.isFiltering, isSettled: corpusSettled)
--- a/desktop/macos/Desktop/Sources/MainWindow/QueryShell/QueryResultsPanel.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/QueryShell/QueryResultsPanel.swift
@@ -193,10 +193,10 @@ struct QueryResultsPanel<Content: View, Accessory: View, Footer: View>: View {
   }
 
   private var countSentence: String {
-    guard let total else { return "Counting what you've captured…" }
     if mode == .answer {
-      return "\(QueryShellCount.number(total)) captured moments ready as context"
+      return "Answers cite the context they use"
     }
+    guard let total else { return "Counting what you've captured…" }
     return QueryShellCount.sentence(
       matching: matching, total: total, isFiltering: request.isFiltering, isSettled: corpusSettled)
   }
--- a/desktop/macos/Desktop/Sources/MainWindow/QueryShell/QuerySearchBar.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/QueryShell/QuerySearchBar.swift
@@ -8,6 +8,9 @@ import SwiftUI
 struct QuerySearchBar: View {
   @Binding var text: String
   var accessibilityID: String = "query-search-field"
+  var focusClaim: Int = 0
+
+  @FocusState private var isFocused: Bool
 
   var body: some View {
     HStack(spacing: QueryShellLayout.heroRowSpacing) {
@@ -18,6 +21,7 @@ struct QuerySearchBar: View {
         .textFieldStyle(.plain)
         .scaledFont(size: QueryShellLayout.queryFontSize, weight: .regular)
         .foregroundStyle(Ink.primary)
+        .focused($isFocused)
         .accessibilityIdentifier(accessibilityID)
       if !text.isEmpty {
         Button {
@@ -34,5 +38,9 @@ struct QuerySearchBar: View {
     .padding(.horizontal, QueryShellLayout.barPaddingHorizontal)
     .frame(minHeight: QueryShellLayout.barMinHeight)
     .inkGlassPanel(cornerRadius: QueryShellLayout.panelCornerRadius, shadow: .ambient)
+    .onAppear {
+      if focusClaim > 0 { isFocused = true }
+    }
+    .onChange(of: focusClaim) { _, _ in isFocused = true }
   }
 }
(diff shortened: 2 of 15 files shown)


Git-on-my-level:
Thanks @aadivb9 — this is a thoughtfully constructed PR. The "prove value with real context" direction comes through in both the code and the tests, and several details show real care.

**What stands out**

- `SecondBrainHome.swift`: keeping the whole value canvas as a pure `SecondBrainHomeSnapshot.compose(...)` value type with tested policy enums (`SecondBrainHomePresentationPolicy`, `SecondBrainPromptPolicy`, `HomeSearchPresentationPolicy`) rather than burying logic in the view. The never-an-interstitial rule (`showsOverview` returns false when a transcript exists unless onboarding is handing off) protects in-progress conversations, and `frozenPrompts` stops an async suggestion refresh from swapping the question under the pointer.
- Honest empty/thin states: `loadScreenCount()` in `QueryShellHome.swift` no longer writes a false `0` when the screen pool isn't ready, and the snapshot renders "Screen history isn't ready yet" instead of zero counts. Negative counts are clamped and covered by `testBadCountsCannotProduceNegativeOrZeroProof`.
- Privacy fencing: the onboarding answer excerpt lives only in memory (`OnboardingProofReceipt` in `OnboardingOpenerComposer.swift`, bounded to 180 chars, never persisted), and `ChatProvider.swift` now clears `onboardingOpener` inside `resetSessionStateForAuthChange()` so a previous owner's name and answer excerpt can't survive an owner transition — with a regression test in `ChatTranscriptResetRevocationTests.swift`. The calendar enrichment in the new `ChatProvider+OnboardingOpener.swift` also gained an owner-authorization re-check plus a stale-opener guard, which is stricter than the code it moved from.
- Suggestion safety: `SecondBrainPromptPolicy.canUseSuggestion` plus the disabled buttons and "Finish or clear your draft" copy mean a suggestion can never clobber a typed draft.
- `QuerySearchBar.swift`'s `focusClaim` counter (onAppear re-claim + onChange increment) mirrors the existing `caretClaims` pattern in `QueryShellHome.swift`, so the expanded search field reliably takes focus after ⌘K.
- Tests: `SecondBrainHomeTests.swift` (10 cases) covers the activation/ready/gathering phases, receipt bounding, evidence dedup/truncation, and the presentation/draft/search policies; `OnboardingOpenerComposerTests.swift` covers the active-turn-only proof policy including streaming/failed rejection.
- The changelog entry matches the sibling format, and the e2e updates (`home.yaml` S1–S1d, `onboarding-flow.yaml` covers) track the new flow including draft protection and minimum-window/reduce-motion checks.

**Notes for maintainer product review**

- This reshapes two high-leverage surfaces at once: resting Home in `QueryShellHome.swift` (the search field becomes a compact ⌘K command instead of always-mounted) and the onboarding promise copy in `SBOnboardingModel.swift` (now "the second brain you trust more than your first"). Both feel on-mission to me, but they are brand-voice and information-architecture decisions a human maintainer should sign off on.
- `QueryResultsPanel.countSentence` now shows the static claim "Answers cite the context they use" in answer mode. The citation feature is real so the claim is supportable, but it now appears even on answers that didn't use context — worth a quick copy pass.
- Minor: `SecondBrainHomeSnapshot.compose` filters out "What did I spend my time on this week?" by string literal; if that suggestion's copy ever changes in `HomeSuggestionComposer`, the filter silently stops matching. A shared constant would make the coupling explicit.

**Verification status**: static review against the merge base — every symbol and API the new code touches checks out against `main` (including `RuntimeOwnerIdentity`, `HomeSuggestionsStore.refreshIfNeeded`, `ChatMessage.journalStatus`, `chatViewport.activeClientTurnId`, and the task/memory/conversation model fields), and the `ChatProvider.swift` extraction leaves no dangling references. No CI results were available on this head and I couldn't compile the macOS target in this environment, so a build plus the updated manual e2e flows should gate the merge.

Leaving for human maintainer review: product-direction sign-off is needed for the Home/onboarding value-proposition redesign before this merges.

---

aryanorastar:
Keeping this open. Not approving `e194bea9b490`. `SecondBrainHome.swift` absent on origin/main. statusCheckRollup empty, so not approvable.

PR #11901, opened 2026-08-20
Fix desktop processing banner overlapping conversation details

aryanorastar:
## Summary

- keep the deferred conversation-processing banner in the normal summary layout flow
- preserve a full spacing-token gap between processing feedback and conversation metadata
- add a behavioral layout regression test and a user-facing desktop changelog entry

## Problem

While a deferred conversation was being enriched, the **Processing conversation…** banner was attached as an overlay to the entire Conversation Details card. SwiftUI overlays do not participate in their parent’s size calculation, so the banner occupied the same vertical region as metadata that was already available from the local conversation record. The source and duration chips were consequently rendered underneath the banner.

This was reproduced in the macOS desktop app while verifying a separate conversation-deletion report.

## Root cause

`ConversationDetailView` rendered `summaryContent` in the card’s normal layout flow, then painted `deferredProcessingSection` over the card with `.overlay(alignment: .top)`. Both layers used the same outer spacing token, but the overlay reserved no height for itself and did not account for the card header.

## Fix

Introduce `ConversationDetailProcessingLayout`, a small production layout seam that places the conditional banner before the existing summary content in one `VStack`:

- the banner now contributes its intrinsic height
- `OmiSpacing.xxl` provides the existing 24-point section separation
- source, duration, and all later summary content remain in normal document flow
- no new spacing values, visual tokens, or interaction behavior are introduced

### Before — processing banner overlaps metadata

Processing banner overlapping conversation metadata

### After — banner reserves space above metadata

Fixed processing banner layout

The patched state was physically verified in the named bundle `/Applications/omi-processing-layout.app`, including resizing the desktop window.

## Regression coverage

`testProcessingBannerReservesSpaceAboveConversationMetadata` renders the production layout seam in idle and processing states and verifies that enabling processing increases the measured height by the banner height plus `OmiSpacing.xxl`. This would fail if the banner were changed back to an overlay or another non-layout decoration.

## Verification

- `./scripts/dev-feedback.py --once swift 'ConversationDetailAutomationStateTests'` — 9 tests passed, 0 failures
- `./scripts/swift-format-wrapper.sh lint Desktop/Sources/MainWindow/Pages/ConversationDetailView.swift Desktop/Tests/ConversationDetailAutomationStateTests.swift` — passed
- `python3 ../../.github/scripts/desktop-changelog.py validate` — passed
- `git diff --check` — passed
- named macOS bundle physical verification — passed
- `make preflight` — passed

## Scope and risk

The change is limited to conversation-detail presentation during deferred enrichment. Completed conversations and the underlying enrichment, persistence, transcript, deletion, and metadata behavior are unchanged. Risk is low and covered by the focused layout test plus physical verification.

Failure-Class: none

No product invariant citations are required for the changed paths.

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/ConversationDetailView.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/ConversationDetailView.swift
@@ -7,8 +7,6 @@ enum ConversationDetailPane: Equatable {
   case transcript
 }
 
-/// Keeps deferred-processing feedback in the summary's layout flow so it
-/// reserves space instead of painting over metadata that is already available.
 struct ConversationDetailProcessingLayout<Banner: View, Content: View>: View {
   let isProcessing: Bool
   let banner: Banner
@@ -846,8 +844,8 @@ struct ConversationDetailView: View {
 
   // MARK: - Deferred Processing Loader
 
-  /// Displayed while a lazily-deferred conversation is enriched, ahead of any
-  /// details that are already available from the local cache.
+  /// Overlaid while a lazily-deferred conversation is enriched, preserving the
+  /// position of details that may already be available from the local cache.
   private var deferredProcessingSection: some View {
     HStack(spacing: OmiSpacing.md) {
       ProgressView()
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/ConversationDetailView.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/ConversationDetailView.swift
@@ -7,6 +7,34 @@ enum ConversationDetailPane: Equatable {
   case transcript
 }
 
+/// Keeps deferred-processing feedback in the summary's layout flow so it
+/// reserves space instead of painting over metadata that is already available.
+struct ConversationDetailProcessingLayout<Banner: View, Content: View>: View {
+  let isProcessing: Bool
+  let banner: Banner
+  let content: Content
+
+  init(
+    isProcessing: Bool,
+    @ViewBuilder banner: () -> Banner,
+    @ViewBuilder content: () -> Content
+  ) {
+    self.isProcessing = isProcessing
+    self.banner = banner()
+    self.content = content()
+  }
+
+  var body: some View {
+    VStack(alignment: .leading, spacing: OmiSpacing.xxl) {
+      if isProcessing {
+        banner
+      }
+      content
+    }
+    .frame(maxWidth: .infinity, alignment: .leading)
+  }
+}
+
 /// Full detail view for a single conversation
 struct ConversationDetailView: View {
   let conversation: ServerConversation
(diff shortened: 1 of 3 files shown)


undivisible (changes requested):
## Request changes

The code/test/changelog scope is otherwise focused, but this diff adds a new production comment and rewrites the existing deferred-processing comment, contrary to the repository rule to preserve comments verbatim and add no new comments.

### Please
1. Remove the new `ConversationDetailProcessingLayout` comment.
2. Restore the existing `deferredProcessingSection` comment verbatim; leave the layout, test, and changelog changes intact.

aryanorastar:
Addressed in `7c43952c4a`:

- removed the new `ConversationDetailProcessingLayout` comment
- restored the existing `deferredProcessingSection` comment verbatim
- left the layout, regression test, and changelog changes intact

Verification: `swift-format` lint passed, `git diff --check` passed, and `ConversationDetailAutomationStateTests` passed (9 tests, 0 failures).

Git-on-my-level:
Thanks @aryanorastar — verified on `7c43952`, and both items from the earlier change request are resolved:

- the new `ConversationDetailProcessingLayout` struct carries no comment;
- the `deferredProcessingSection` doc comment is restored verbatim, and a comment-line diff of the whole file against `main` shows no other comment churn anywhere.

Review of the three changed files:

**`ConversationDetailView.swift`** — the root cause reads correctly: the banner was painted via `.overlay(alignment: .top)`, and overlays don't participate in the parent's size negotiation, so it covered the source/duration chips instead of reserving height. Moving it into the new `ConversationDetailProcessingLayout` VStack (banner above `summaryContent`, `OmiSpacing.xxl` gap) is the right fix. Two nice details: `.allowsHitTesting(false)` is preserved at the call site so the banner can't intercept clicks on the metadata row, and the geometry matches the old overlay (`xxl` inset + `xxl` gap) so no new spacing tokens are introduced. One expected behavior change worth naming: the card now grows while processing and content below shifts down, rather than being covered — that's the point of the fix, just noting it's a visible layout change.

**`ConversationDetailAutomationStateTests.swift`** — `testProcessingBannerReservesSpaceAboveConversationMetadata` is a genuine behavioral test: asserting `processingHeight - idleHeight == bannerHeight + OmiSpacing.xxl` via `NSHostingView.fittingSize` would fail under the old overlay (zero reserved height), so it pins exactly the regression being fixed.

**`changelog/unreleased/20260820-processing-banner-layout.json`** — matches the sibling unreleased entries in format, naming, and tone, and the wording is accurate to the behavior change.

Desktop Swift CI is green on this head (Build & Tests, Static & Test Contracts, Release Compile, Hygiene). From my side the code is in good shape — leaving the final call with @undivisible to re-confirm the outstanding change request against the new head before merge.

---

undivisible:
ty @aryanorastar this really bugged me

@Archit-lal look he's picking up after you :))
ill review this and merge if it looks all good

undivisible (approved):
looks good. banner sits in the layout now so it can’t cover the chips, the height test would fail if this went back to an overlay, and the comment nits are gone.

aryanorastar:
Thanks, glad this helped. I’ll keep picking up these macOS UX issues where I can. @undivisible @Archit-lal

PR #11905, opened 2026-08-20
fix: authorize OAuth MCP memory reads

aryanorastar:
## Summary

Fix ChatGPT and other OAuth MCP clients receiving JSON-RPC `-32009` from `get_memories` and `search_memories` even though the same connection can call `get_user_profile` and advertises the `memories.read` tools.

## User-visible failure

An OAuth access token preserved the authenticated UID, scopes, client ID, and grant ID, but the MCP transport did not project the client/grant pair into the product authorization context required by canonical memory reads. Tool discovery therefore advertised both memory tools from `memories.read`, while execution rejected them as missing app/key identity.

## Fix

- Treat the server-verified OAuth `client_id` and `grant_id` as the stable MCP app/key identity.
- Persist the matching server-owned memory grant in the same transaction as new OAuth consent and authorization-code creation.
- Lazily backfill that grant when an existing active OAuth token is validated, allowing already-connected ChatGPT clients to recover without reconnecting.
- Keep the canonical scope and persisted-grant checks fail-closed; this does not grant Archive access or infer scopes from tool metadata.

## Verification

- `BACKEND_UNIT_TEST_FILE_LIST=/tmp/mcp-oauth-focused-tests.txt PYTHON=.venv/bin/python bash test.sh` — 91 passed.
- `PYRIGHT_PYTHON=.venv/bin/python bash scripts/typecheck.sh` — 0 errors.
- Full backend selector completed all MCP/OAuth tests successfully. Two unrelated speech-provider files fail on the current baseline because retired Deepgram configuration selects Deepgram instead of Modulate: `test_modulate_stt.py` and `test_streaming_deepgram_backoff.py`.
- Live ChatGPT connector verification: pending deployment.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11905?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/mcp_oauth.py
+++ b/backend/database/mcp_oauth.py
@@ -14,6 +14,12 @@ from google.cloud import firestore
 
 from database._client import db
 from database.account_deletion_policy import account_deletion_blocks_access, normalize_account_deletion_status
+from database.memory_app_key_grants import (
+    APP_KEY_MEMORY_GRANT_DOC_ID,
+    APP_KEY_MEMORY_GRANTS_COLLECTION,
+    MCP_CONSUMER,
+    build_app_key_scope_grant_contract_state,
+)
 
 PRODUCTION_MCP_RESOURCE_URL = "https://api.omi.me/v1/mcp/sse"
 # Omi Beta intentionally serves MCP data from dev while retaining the production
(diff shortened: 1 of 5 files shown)


undivisible (changes requested):
## Request changes

I found a production memory-confidentiality fail-open in the lazy backfill, so I am requesting changes.

- `backend/database/mcp_oauth.py:469-496,779-790`: `_ensure_oauth_memory_grant()` rewrites any mismatch to `enabled=True` with capabilities from the active OAuth grant. A valid token can therefore resurrect an existing server-owned grant that was disabled, narrowed, or malformed; the canonical memory gate then allows access again. Backfill must be create-only for a genuinely absent OAuth grant entry, and must fail closed without overwriting existing control-plane state.
- `backend/database/mcp_oauth.py:779-799` and `backend/routers/mcp_sse.py:167-174`: token validation does not strictly require and cross-check token `uid`, `client_id`, `grant_id`, `resource`, and `scopes` against the active grant before projecting identity. Missing or malformed records can reach the context builder and raise instead of returning unauthenticated, while mismatched records are not rejected before backfill.
- Add deny-path tests for disabled, narrowed, malformed, missing, revoked, and expired grants/tokens; missing or invalid client/grant/scope fields; cross-user/client mismatches; and explicit proof that Archive is never granted. The added tests currently cover only consent/backfill success paths, and the SSE test patches authorization to `allowed=True`, so it does not exercise the persisted grant gate.

Please fix the overwrite/validation boundary and rerun the focused tests before requesting another security review.

aryanorastar:
Addressed the security review in `620a2cfafd`.

What changed:
- OAuth memory-grant backfill is now a Firestore transaction that writes only when the exact OAuth app/grant entry is genuinely absent.
- Existing disabled, narrowed, or malformed memory-control state is never rewritten. Re-consent also cannot re-enable a disabled entry.
- Access-token validation now requires non-empty `uid`, `client_id`, `grant_id`, `resource`, and a valid scope list; requires an unexpired, unrevoked token; and cross-checks all identity fields and token scopes against the active persisted grant before projecting MCP identity.
- Missing, revoked, expired, malformed, cross-user, cross-client, cross-resource, and scope-escalated token/grant pairs all return unauthenticated.
- The SSE regression now executes the real persisted memory-grant gate rather than mocking authorization to `allowed=True`.
- Archive remains explicitly false in the persisted OAuth grant and absent from the projected memory context.

Coverage added for disabled, narrowed, malformed, missing, revoked, expired, invalid-field, identity-mismatch, scope-escalation, sibling-grant preservation, re-consent, and no-Archive cases.

Verification:
- focused MCP/OAuth backend selection: 136 passed
- backend typecheck: 0 errors
- bounded pre-push gate: 25 checks passed, including OpenAPI compatibility, import purity, runtime-image closure, formatting, and the focused test selection

Live ChatGPT verification still remains deployment-gated.

undivisible (approved):
Security notes from the prior review look addressed in 620a2cf: backfill is create-only, token/grant identity is fail-closed, Archive stays off, and the SSE regression hits the real grant gate. CI is green. Landing this.

PR #11906, opened 2026-08-20
fix(backend): one unusable extractor element must not 500 the whole conversation

kodjima33:
## The bug

Prod, `api.omi.me`, backend image `920fc55`:

```
ERROR:utils.llm.gateway_error_contract:Conversation processing failed: ValidationError
-> HTTP 500 {"detail": "Error processing conversation, please try again later"}
```

**30 of these on `POST /v1/dev/user/conversations` on 2026-08-20 alone** (7-12s each), plus more on `POST /v1/conversations`. Users retry the same text ~14s later and get the same 500, so the conversation is simply lost. The one occurrence prod did print a traceback for names the mechanism exactly:

```
Value error, duration must be a positive number of minutes [type=value_error, input_value=0, input_type=int]
```

## Root cause

`ExtractedEvent.duration` carried a validator that **raised** on a non-positive value. Pydantic validates a list element as part of its parent, so one `duration: 0` on one event failed the entire `StructuredExtraction` — title, overview, every action item, every other event — and `_get_structured` mapped that to HTTP 500. The user got no summary at all because of one integer.

This is the same contract as #11871 (`FC-out-of-vocabulary-model-token-discards-extraction`), one level up: that PR normalized enum *fields*; the parser still fails the whole payload when a whole *element* is unusable.

## The fix

Each list on the extraction is optional detail that already models its own absence (`default_factory=list`), so an unusable element is worth exactly as much as no element:

- `StructuredExtraction` / `ActionItemsExtraction` validate `sections` / `action_items` / `events` element-by-element and drop the ones that fail, instead of failing the model.
- A non-positive or unparseable `duration` falls back to the field's own documented 30-minute default rather than being rejected — a zero duration means "unspecified", and the event itself is real, so it survives.
- An explicit `null` on a defaulted scalar (`title`, `overview`, `emoji`) or on a list falls back to the default instead of raising.
- Dropping an element logs the field plus the pydantic error `loc`/`type` with `include_input=False`. The prod 500s carried **no traceback at all**, which is why the other 30 could not be told apart from this one; the next occurrence is diagnosable.

Strictness stays where it is load-bearing: a required field on an element still fails *that element*.

## What I tested

- `backend/tests/unit/test_structured_extraction_partial_content.py` (14 tests) drives the real `summarize_experience_text` — the production function behind `POST /v1/dev/user/conversations` — on the prod payload shape, and the `PydanticOutputParser(StructuredExtraction)` seam the transcript paths use.
- **Pre-fix control:** reverting only `models/structured_extraction.py` and re-running the same file reproduces the exact prod failure (`ValidationError` out of `summarize_experience_text`). With the fix it returns a full summary with the event kept at 30 minutes.
- 17 related unit test files, **346 tests**, run file-by-file the way `test.sh` invokes them: all green — including `test_extracted_action_item_vocabulary.py` (the #11871 guard) and `test_action_item_date_validation.py` (sandboxed importer; the guard deliberately avoids a `models/` -> `utils/` import so it keeps passing).

Product invariants affected: none.

Failure-Class: FC-out-of-vocabulary-model-token-discards-extraction

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11906?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/models/structured_extraction.py
+++ b/backend/models/structured_extraction.py
@@ -1,10 +1,13 @@
+import logging
 from datetime import datetime
-from typing import Any, Dict, List, Literal, Optional, Tuple
+from typing import Any, Dict, List, Literal, Optional, Tuple, Type
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import BaseModel, Field, ValidationError, field_validator, model_validator
 
 from models.conversation_enums import CategoryEnum
 from models.structured import ActionItem, Event, Section, Structured
+
+logger = logging.getLogger(__name__)
 
 CAPTURE_OWNERS: Tuple[str, ...] = ('user', 'other', 'unknown')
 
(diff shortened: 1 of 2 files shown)


undivisible (changes requested):
## Request changes

The focused smoke passes, but this diff also makes explicit nulls for title, overview, and emoji silently use defaults. Please remove that scalar-null relaxation and keep the fail-open boundary to optional list elements plus the event duration default.

PR #11924, opened 2026-08-20
fix(backend): build the rebuilt knowledge graph before deleting the old one (#11923)

### LINKED ISSUE #11923: rebuild_knowledge_graph deletes the graph before a multi-minute rebuild — an interruption mid-rebuild is permanent partial data loss
**Product impact**

- Primary loop layer: **Memory**
- [x] Involves data loss, failed capture, or broken sync/trust

A rebuild interrupted partway through destroys the account's existing knowledge graph and leaves a
partial one in its place, with nothing scheduled to finish the job and no way to recover the old one.

**Describe the bug**

`POST /v1/knowledge-graph/rebuild` is now safe *up to* the point the worker starts — #11820 removed the
eager delete in the route, thank you. But the destructive ordering still exists one level down, inside
`rebuild_knowledge_graph` itself, where the window is minutes wide rather than instantaneous.

This is **not** a regression from #11820, and not the defect #11796 described. The delete-then-rebuild in
the LLM helper predates both (`1847a840a2`, 2025-12-17). #11820 correctly narrowed the hazard; this is
about the part left standing, which I don't think was in its scope.

**Current behavior**

`backend/utils/llm/knowledge_graph.py:341` — the first statement of the rebuild:

```python
def rebuild_knowledge_graph(uid, memories, user_name="User", *, db_client=None):
    kg_db.delete_knowledge_graph(uid, db_client=db_client)
```

Everything that reconstructs the graph runs *after* that committed delete:

- the caller passes up to **500 memories** (`backend/routers/knowledge_graph.py:258`, `limit=500`)
- each memory is one LLM extraction round-trip (`get_llm('knowledge_graph').invoke(prompt)`)
- concurrency is capped at **4** (`_KG_REBUILD_SEM = threading.BoundedSemaphore(4)`)

So a full rebuild is on the order of **125 sequential waves of LLM calls**. Nodes and edges are upserted
incrementally as each memory completes, so at any instant in that multi-minute span the stored graph is a
partial one.

`_rebuild_graph_task` is scheduled via FastAPI `BackgroundTasks`, which runs in the serving process with no
retry or durability. If that process goes away mid-rebuild — deploy, restart, eviction, OOM, or an unhandled
error in the driver — the account keeps whatever subset had been written, the original graph is already gone,
and nothing is scheduled to complete it. There is no in-progress marker to recover from, and the user's only
signal was a `200 {"status": "rebuilding"}` returned minutes earlier.

The per-memory `except Exception` handlers keep *individual* extraction failures from aborting the run, which
is good, but they don't help when the process itself dies.

**Expected behavior**

A rebuild that is interrupted should leave the account's existing graph intact, or at minimum leave a
recoverable marker so the rebuild can be re-driven. Losing the old graph should be contingent on a new one
having been successfully built.

**Impact**

Bounded but not small: it needs a rebuild in flight when the process dies, so it's a deploy-window /
crash-window loss rather than an everyday one. The consequence when it does happen is silent and
unrecoverable — the pre-rebuild graph is not retained anywhere.

**Possible directions**

Listed for discussion, not as a proposal — the tradeoffs look like a call for whoever owns this path,
especially with the legacy store on a migration path off to canonical / assertion-backed state:

1. **Extract first, then swap.** Do the LLM work into memory, and only then delete + write. No schema or
   read-path change; collapses the destructive window from ~125 LLM waves to a burst of Firestore writes.
   Requires the entity-resolution map (currently re-read from Firestore per memory under `node_lock`) to be
   held in memory instead — which also removes N Firestore reads.
2. **Shadow generation + atomic swap.** Genuinely safe, but needs a generation concept on graph nodes/edges,
   read-path changes, and abandoned-generation cleanup. `account_generation` fencing already exists for memory
   items (`backend/database/knowledge_graph.py:362`) as prior art.
3. **Persisted in-progress marker + recovery.** Doesn't prevent the loss, but makes it detectable and
   re-drivable rather than silent.
4. **Accept and document.** If the legacy graph is close enough to end-of-life that this isn't worth the
   spend, that's a legitimate answer — the failure mode is at least worth a comment at the delete site.

Happy to implement whichever direction you'd prefer (1 is the one I'd expect to be reviewable as a contained
change). I'd rather get a steer on scope first than send a PR that guesses wrong on a destructive path.

**Environment**

Observed by reading `main` at `d7485f42cb`. Static analysis of the ordering — I have not reproduced a
mid-rebuild process kill against live Firestore. Device/app fields from the bug template are N/A; this is a
backend code path.

kodjima33:
Fixes #11923.

## Bug

`rebuild_knowledge_graph` deleted the account's stored knowledge graph as its **first statement**, then upserted node by node as each memory's LLM extraction returned:

```python
def rebuild_knowledge_graph(uid, memories, user_name="User", *, db_client=None):
    kg_db.delete_knowledge_graph(uid, db_client=db_client)   # <- everything below runs after this
```

The caller passes up to 500 memories (`routers/knowledge_graph.py`, `limit=500`), each one an LLM round-trip, at concurrency 4 (`_KG_REBUILD_SEM`). So for minutes at a stretch the account's only copy of its graph was a partial one. `_rebuild_graph_task` drives that from FastAPI `BackgroundTasks` — in the serving process, with no retry and no durability — so a deploy, restart, eviction or driver error inside that window left the user with whatever subset had been written, the original already gone, nothing scheduled to finish it, and no marker to recover from. Their only signal was a `200 {"status": "rebuilding"}` returned minutes earlier.

Not theoretical: prod served **580 successful `POST /v1/knowledge-graph/rebuild` calls in the last 30 days** (~19/day), and backend revisions roll during that traffic.

This is the part #11820 left standing — it removed the *route's* eager delete; the destructive ordering one level down predates it (`1847a840a2`).

## Root cause

The ordering was load-bearing only by accident. Every `kg_db.get_knowledge_nodes` read the per-memory loop made saw **only rows this same rebuild had just written**, because the delete had already emptied the collection — so the loop was using Firestore as scratch space for its own intermediate state, and the delete had to come first for that to work.

## Fix

Direction 1 from the issue (extract first, then swap). `_StagedRebuild` holds that intermediate state in memory instead, mirroring the identity and merge rules of `upsert_knowledge_node` / `upsert_knowledge_edge` — which still perform every actual write. The delete then moves down beside the write burst it belongs with:

```python
    # Only now is the replacement graph fully known.
    kg_db.delete_knowledge_graph(uid, db_client=db_client)
    for node_data in staged.nodes(): kg_db.upsert_knowledge_node(...)
    for edge_data in staged.edges(): kg_db.upsert_knowledge_edge(...)
```

Anything that interrupts the extraction now leaves the stored graph untouched. The destructive window collapses from ~125 sequential LLM waves to a burst of Firestore writes. It also removes one node-collection read per memory, and a concurrent `GET /v1/knowledge-graph` during a rebuild now sees the previous graph rather than a partial one.

Deferring the writes introduces one hazard the incremental loop could not have: `upsert_knowledge_node` re-resolves a node it cannot find by id against label and alias, so a staged node whose label another staged node later adopted as an alias lands under that *other* node's id. Edges are therefore written from the ids the upserts returned — exactly what the old loop got for free by using `saved_node['id']`.

No schema, read-path or route change; `rebuild_knowledge_graph` keeps its signature and return value.

## What I tested

`backend/tests/unit/test_knowledge_graph_rebuild_ordering.py` — new, drives the production `rebuild_knowledge_graph` against an in-memory stand-in for `database.knowledge_graph` that reproduces the real merge/identity semantics, so ordering and the rebuilt graph are both observed through the seam production writes through.

Ran against the **pre-fix source** (`origin/main`) as a control, same tests:

```
FAILED test_stored_graph_survives_until_every_extraction_has_finished
FAILED test_rebuild_interrupted_before_the_swap_leaves_the_old_graph_intact
        assert 'delete_knowledge_graph' not in ['delete_knowledge_graph']
2 failed, 1 passed
```

and after the fix: `3 passed`.

`test_rebuilt_graph_still_merges_entities_across_memories` passes on **both** sources — it pins the rebuilt graph itself (a second memory naming an entity by its alias merges into the first memory's node, `memory_ids` and `aliases` unioned, the repeated edge merged to one row), which is what makes this a behaviour-preserving refactor rather than a rewrite.

`test_edges_follow_the_ids_their_nodes_actually_landed_on` covers the re-resolution hazard above: run against a variant of this fix *without* the id remap it fails with a dangling edge (`assert 'b243b7af-…' in {'35f69f4b-…'}`), and passes with it.

Repo runner over the 8 files touching this path (`./test.sh` with a scoped `BACKEND_UNIT_TEST_FILE_LIST`): **134 passed, 0 failed**, exit 0.

Not verified: an actual process kill against live Firestore. The interruption test simulates the driver losing its executor mid-run, which exercises the same early-exit path, but it is a stand-in for the real crash.

## Product invariants

`scripts/pr-preflight --suggest`: none affected.

Failure-Class: new

### DIFF
--- a/.github/failure-classes/FC-destructive-replace-precedes-derivation.json
+++ b/.github/failure-classes/FC-destructive-replace-precedes-derivation.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-destructive-replace-precedes-derivation",
+  "violated_contract": "A replace-in-place rebuild may destroy the state it is replacing only once the replacement is fully derived. The delete belongs with the writes that land the new state, never ahead of the work that produces it: if the derivation is slow (per-item provider calls) and its driver is not durable (an in-process background task with no retry), then any redeploy, restart, eviction or driver error inside that window leaves the owner with a partial state and no copy of what it replaced. A derivation loop that reads its own intermediate writes back out of the durable store is what forces the destructive ordering, and using the store as scratch space is the defect, not the sequencing.",
+  "canonical_prevention": "Derive the replacement outside the durable store — staged in the driver's own memory, or into a shadow generation — and delete plus write in one late phase, so an interruption before the swap is a no-op on stored state. Where the loop previously re-read the store for its own intermediate state, move that state into the driver and mirror the store's identity and merge rules there so the swap stays behaviour-preserving. Prove it with a regression test that drives the production function and asserts both that the delete is ordered after the last derivation step and that an interrupted run leaves the prior state untouched.",
+  "canonical_prevention_artifact": [
+    "backend/tests/unit/test_knowledge_graph_rebuild_ordering.py"
+  ],
+  "evidence_prs": [
+    11924
+  ],
+  "scope_hints": [
+    "backend/utils/llm/knowledge_graph.py",
+    "backend/database/knowledge_graph.py"
+  ],
+  "status": "open"
+}
--- a/backend/tests/unit/test_knowledge_graph_canonical_mutation_routes.py
+++ b/backend/tests/unit/test_knowledge_graph_canonical_mutation_routes.py
@@ -165,8 +165,9 @@ def test_a_regate_that_flips_after_the_response_leaves_the_graph_intact(client, 
 
 def test_the_legacy_rebuild_still_clears_the_graph_itself(monkeypatch):
     # The route relies on this: dropping its eager delete is only safe while
-    # `rebuild_knowledge_graph` starts by clearing the old graph, so a rebuild
-    # cannot merge new extractions into stale nodes.
+    # `rebuild_knowledge_graph` clears the old graph itself, so a rebuild cannot
+    # merge new extractions into stale nodes. The clear now runs at the end of the
+    # rebuild rather than the start (#11923), but it still runs.
     llm_kg = kg_router._knowledge_graph_llm_module()
     calls: List[str] = []
     monkeypatch.setattr(kg_db, "delete_knowledge_graph", lambda uid, **_kw: calls.append(uid))
--- a/backend/tests/unit/test_knowledge_graph_rebuild_ordering.py
+++ b/backend/tests/unit/test_knowledge_graph_rebuild_ordering.py
@@ -16,7 +16,7 @@ both observed through the same seam production writes through.
 from __future__ import annotations
 
 import json
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import Future, ThreadPoolExecutor
 from contextlib import nullcontext
 from typing import Any, Dict, List, Optional
 
@@ -79,6 +79,21 @@ class _FakeKgStore:
     def get_knowledge_graph(self, uid: str, *, db_client: Any = None) -> Dict[str, Any]:
         self.calls.append('get_knowledge_graph')
         return {'nodes': list(self.nodes.values()), 'edges': list(self.edges.values())}
+
+
+class _InlineExecutor:
+    """Runs each memory inline, in submission order.
+
+    A worker thread left running past the end of a test calls `kg.get_llm` again
+    after the next test has repatched it, so its extraction lands in that test's
+    recorder — order-dependent, and only on a loaded runner. Nothing here needs
+    real concurrency, so nothing here starts a thread.
+    """
+
+    def submit(self, fn: Any, *args: Any, **kwargs: Any) -> "Future[Any]":
+        future: Future[Any] = Future()
+        future.set_result(fn(*args, **kwargs))
+        return future
 
 
 def _response(nodes: List[Dict[str, Any]], edges: List[Dict[str, Any]]) -> Any:
@@ -146,18 +161,17 @@ def test_rebuild_interrupted_before_the_swap_leaves_the_old_graph_intact(rebuild
 
     monkeypatch.setattr(kg, 'get_llm', lambda *_args, **_kwargs: type('LLM', (), {'invoke': staticmethod(invoke)}))
 
-    class _DyingExecutor:
+    class _DyingExecutor(_InlineExecutor):
         """Stands in for the serving process going away: submission stops working."""
 
         def __init__(self) -> None:
-            self._inner = ThreadPoolExecutor(max_workers=1)
             self._submitted = 0
 
-        def submit(self, fn: Any, *args: Any, **kwargs: Any) -> Any:
+        def submit(self, fn: Any, *args: Any, **kwargs: Any) -> "Future[Any]":
             self._submitted += 1
             if self._submitted > 1:
                 raise RuntimeError('cannot schedule new futures after shutdown')
-            return self._inner.submit(fn, *args, **kwargs)
+            return super().submit(fn, *args, **kwargs)
 
     monkeypatch.setattr(kg, 'llm_executor', _DyingExecutor())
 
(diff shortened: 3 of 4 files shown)


kodjima33:
Closing per this job's merge-or-report gate, **not** because anything is wrong with the change. Branch `watchdog/issue-11923-kg-rebuild-destructive-window` is pushed and untouched, so this is one `gh pr create` from being back.

**State when closed** (head `e348701`): 22 of 23 checks green. The one red check, `Backend unit suite`, fails with 67 tests — **the exact same 67 that fail on `main`**, none of them in files this PR touches:

```
$ comm -13 main_failures.txt pr_failures.txt   # in PR but not on main
(empty)
$ comm -23 main_failures.txt pr_failures.txt   # on main but not in PR
(empty)
```

That red is #11925 — `Backend Unit Tests` has been broken on `main` since `1822dc3397` (#11913) added a module-level `from database.apps import get_app_by_id_db` to `models/chat.py`, which the stubbed chat tests cannot import. Still present on `main` as of `ca9b4feda9`.

This PR's own four new tests pass in that same CI run. The gate this job runs under says "never merge on failing CI", and it does not get to decide that an `--admin` override of a red required check is fine — so: closed rather than left open to rot.

**To land it once #11925 is fixed:** reopen or

```
gh pr create --base main --head watchdog/issue-11923-kg-rebuild-destructive-window
```

Verification evidence is unchanged and is in the commit messages: against `origin/main` as a control the ordering and interruption tests fail and pass after; the dangling-edge test fails against a variant without the id remap; the entity-merge test passes on both, so the refactor is behaviour-preserving. `./test.sh` over the 8 files touching this path: 137 passed, 0 failed. `make preflight`: 25 checks passed.

RatnamOjha:
Reviewed at `e348701`. This looks right to me, and thanks for turning it round so fast — no apology needed for picking it up.

I checked the two claims the change actually rests on rather than taking them on faith:

**The staged merge really does mirror the store.** `_apply_node` unions `memory_ids` and `aliases` and overwrites `label`/`node_type`, which is exactly what `upsert_knowledge_node` does on its `existing.exists` branch. So a rebuild that merges an entity across memories lands the same row it did before.

**`updated_at` on staged nodes is load-bearing, not decoration.** `_node_recency` reads `updated_at` then `created_at` and returns `0.0` otherwise, so without that field every staged node would tie at zero and `_existing_nodes_prompt_json`'s `MAX_PROMPT_EXISTING_NODES` trim would drop an arbitrary subset once a rebuild crosses the cap. Worth keeping the comment that explains why it's there.

**`landed_node_ids` covers a case I'd have expected to go wrong.** Two staged nodes can collide on an alias — one node's alias equal to another's label — and `_label_to_node_id` resolves that to whichever applied last. At commit time `upsert_knowledge_node` can then merge them onto one id, and edges written from the staged ids would dangle. Mapping staged id to landed id fixes it, and `test_edges_follow_the_ids_their_nodes_actually_landed_on` pins it.

Two smaller things I liked: the snapshot is taken under `node_lock` and released before the LLM call, so extractions still run at concurrency 4 rather than serialising; and `all_nodes`/`all_edges` were dead — accumulated and then discarded in favour of `get_knowledge_graph` — so dropping them is a real cleanup rather than collateral.

On the residual you called out: agreed the delete plus N upserts at the end is still not atomic, and I don't think that should hold this up. The window goes from ~125 sequential LLM waves to a burst of Firestore writes, which is the difference between "a routine deploy can hit it" and "you'd have to be unlucky". Directions 2 and 3 stay open on #11923 if the legacy path outlives the canonical migration.

Separately — I've put up #11932 for the `main` breakage that blocked this. Worth noting the shape you and I both first reached for, widening the harness stub, does not actually work: it only helps once another file has already imported firestore into the process, so it passes a multi-file `pytest` run and still fails under the file-isolated runner CI uses. The fix has to move the import in `models/chat.py`.

PR #11932, opened 2026-08-20
fix(backend): keep models/chat.py importable without the database layer

### LINKED ISSUE #11925: main red: Backend Unit Tests failing since #11913 — module-level `from database.apps import get_app_by_id_db` in models/chat.py breaks 67 stubbed chat tests
`Backend Unit Tests` has been red on `main` since **1822dc3397** ("fix(chat): resolve sender display names from app_id", #11913, Max Carter, 2026-08-20 10:05Z). Every push since has failed; it blocks every open PR's required check.

Last green: `03ec0c898e`. First red: `1822dc3397`. Runs:

- green https://github.com/BasedHardware/omi/actions/runs/32357197554
- red https://github.com/BasedHardware/omi/actions/runs/32357208878
- still red at `96db4c7b8` https://github.com/BasedHardware/omi/actions/runs/32358037382

## Cause

#11913 added a **module-level** import to `backend/models/chat.py`:

```python
from database.apps import get_app_by_id_db
```

`models/chat.py` is imported at collection time by the chat and desktop-message test modules, several of which install a stub `database.apps` in `sys.modules` (or import before the real package is loadable). Those stubs do not carry `get_app_by_id_db`, so collection dies:

```
E   ImportError: cannot import name 'get_app_by_id_db' from 'database.apps' (unknown location)
```

and the retry that follows re-imports the real firestore package into an already-populated descriptor pool, which is the error most of the 67 failures actually print:

```
E   TypeError: Couldn't build proto file into descriptor pool: duplicate file name google/cloud/firestore_v1/types/document.proto
```

67 tests fail across `test_chat_generate_reply_stateless.py`, `test_chat_quota_counting_router.py`, `test_chat_stream_error_fallback.py`, `test_desktop_message_quota_router.py`, `test_atomicity_lifecycle_regressions.py` and others. The function itself is fine — `get_app_by_id_db` exists in `backend/database/apps.py:40`; the defect is where the import lives.

The same class of failure is what `backend-import-purity` guards, but that check passed here because the import is legal at import time in a full environment — it only fails under the test stubs.

Not filing a fix PR: this is someone else's in-flight change and the right shape (lazy import inside the resolver vs. widening the stubs) is a call for whoever owns the chat model. Reported by the hourly desktop bug watchdog while its own PR (#11924) inherited all 67 failures.

RatnamOjha:
Fixes #11925.

`Backend Unit Tests` has been red on `main` since **1822dc3397** (#11913), which blocks every open
PR's required check.

## Cause

#11913 added a module-scope import to `backend/models/chat.py`:

```python
from database.apps import get_app_by_id_db
```

`models/chat.py` is imported at collection time by the chat and desktop-message suites, whose shared
harness installs a stub `database.apps` carrying only `record_app_usage`
(`backend/tests/unit/_chat_router_test_harness.py`). The import raises `ImportError`, and the retry
that follows re-imports the real firestore package into an already-populated descriptor pool, which
is the error most of the failures actually print:

```
TypeError: Couldn't build proto file into descriptor pool:
duplicate file name google/cloud/firestore_v1/types/document.proto
```

The function itself is fine. The defect is where the import lives: a Pydantic models module should
not pull the database layer in at import time.

## The fix

Import the resolver at its single call site. The ten tests that patched
`models.chat.get_app_by_id_db` — an attribute that existed only because of the module-scope import —
now patch `database.apps.get_app_by_id_db`, where the symbol actually lives. They still exercise
#11913's behaviour: app-name resolution, the per-`app_id` cache, and the fallbacks for a missing app
and a blank name.

**Widening the harness stub instead does not work.** It only helps once another file has already
imported firestore into the process, so it passes a multi-file `pytest` invocation and still fails
under the file-isolated runner CI uses. Measured on `test_chat_stream_error_fallback.py` run alone,
as CI isolates it:

| tree | result |
|---|---|
| `03ec0c898e` (last green) | 11 passed |
| `5f37b17848` (`main`) | 10 failed, 1 passed |
| `main` + widened stub | 10 failed, 1 passed |
| `main` + this change | **11 passed** |

## Verification

Run at `5f37b17848`.

`backend/test.sh` — 881 files, file-isolated exactly as CI runs them:

```
before:  20 failing files
after:    2 failing files
```

Both survivors are `tests/unit/test_rendered_deployment_contract.py` and
`tests/unit/test_verify_pusher_config_references.py`. They fail identically at the last green commit
`03ec0c898e`, on `assert helm is not None, "helm must be installed for the rendered deployment
contract"` — a missing local binary, unrelated to this change.

Isolation guards:

```
scripts/check_module_stub_pollution.py     Checked 930 backend test file(s); 0 violation(s).
scripts/scan_import_time_side_effects.py   Checked 812 backend production file(s); 0 violation(s).
```

`scripts/pr-preflight --suggest --base upstream/main` reports **no product invariants affected**.

Exercised through the backend unit suite only. I have not run the chat UI against a live backend,
so the user-facing render of a bot sender name is covered by `test_message_formatting.py` rather
than by my having seen it in the app.

## Notes

`scan_import_time_side_effects.py` does not catch this class today: the import is legal at import
time in a full environment and only fails under the test stubs. Worth a follow-up on whether that
guard should also reject `models/**` importing `database/**` at module scope — happy to file it if
you'd like, but I did not want to widen this PR while `main` is red.

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11932?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-model-module-imports-persistence-at-module-scope.json
+++ b/.github/failure-classes/FC-model-module-imports-persistence-at-module-scope.json
@@ -1,0 +1,18 @@
+{
+  "schema_version": 1,
+  "id": "FC-model-module-imports-persistence-at-module-scope",
+  "violated_contract": "A model module must stay importable without its persistence layer. Binding a database helper at module scope in backend/models/** makes every importer of the model — including test collection, which reaches these modules long before any route runs — load the real client package transitively. Suites that install a partial stub for that database module then fail at import rather than at the call they stubbed, and the retry re-imports the real client into an already-populated protobuf descriptor pool, so the error surfaced names a duplicate proto file and not the missing symbol. The defect is the placement of the import, not the helper: a module-scope binding also re-exports the helper as an attribute of the model module, which invites tests to patch it there and couples the seam to the layering violation that created it.",
+  "canonical_prevention": "Import a persistence helper inside the function that calls it, so the model module's import graph stays free of the database layer, and patch the helper in tests at the module that defines it rather than at the model module that borrowed it. Prove it with a test file that exercises the resolving behaviour through the defining module's seam, and check the guard by running the affected suite file-isolated — a partial stub only fails when its file is the one that first imports the client, so a multi-file run can pass while the isolated lane CI uses still fails.",
+  "canonical_prevention_artifact": [
+    "backend/models/chat.py",
+    "backend/tests/unit/test_message_formatting.py"
+  ],
+  "evidence_prs": [
+    11913,
+    11932
+  ],
+  "scope_hints": [
+    "backend/models/**"
+  ],
+  "status": "open"
+}
--- a/backend/models/chat.py
+++ b/backend/models/chat.py
@@ -5,8 +5,6 @@ from typing import Any, Dict, List, Literal, Optional, Union
 
 from pydantic import BaseModel, Field, model_validator
 
-from database.apps import get_app_by_id_db
-
 
 class MessageSender(str, Enum):
     ai = 'ai'
@@ -153,6 +151,10 @@ class Message(BaseModel):
         if use_plugin_name_if_available and message.app_id and message.app_id.strip():
             app_id = message.app_id.strip()
             if app_id not in app_name_by_id:
+                # Imported lazily: this models module must stay importable without the
+                # database layer, which module-scope import of it would drag in.
+                from database.apps import get_app_by_id_db
+
                 app = get_app_by_id_db(app_id)
                 name = app.get('name') if app else None
                 app_name_by_id[app_id] = name.strip() if isinstance(name, str) and name.strip() else None
--- a/backend/tests/unit/test_message_formatting.py
+++ b/backend/tests/unit/test_message_formatting.py
@@ -17,7 +17,7 @@ def _ai_message(*, app_id='some-app-id', text='hello'):
 
 def test_get_messages_as_string_sender_uses_app_display_name():
     m = _ai_message()
-    with patch('models.chat.get_app_by_id_db', return_value={'id': 'some-app-id', 'name': 'Friendly Bot'}) as lookup:
+    with patch('database.apps.get_app_by_id_db', return_value={'id': 'some-app-id', 'name': 'Friendly Bot'}) as lookup:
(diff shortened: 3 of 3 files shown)


RatnamOjha:
Closing — this raced with #11931, which merged about seven minutes after I opened this, and your fix is the better one.

I deferred the import to call time; you removed the dependency outright. `AppNameResolver` as an injected callable means `models/` cannot reach `database/` at all rather than merely reaching it later, and `test_models_chat_import_purity.py` keeps it that way instead of relying on the next author noticing. `FC-model-import-reaches-persistence-client` also covers the class I'd written up here, so there's nothing left in this branch worth landing.

One finding from this that may be worth keeping, since it isn't obvious from either diff: **widening the harness stub does not fix this**, which is the first thing most people will try. Adding `get_app_by_id_db` to the `database.apps` stub in `_chat_router_test_harness.py` only helps once some earlier file has already imported firestore into the process, so it passes a multi-file `pytest` invocation and still fails under the file-isolated runner CI uses. Measured on `test_chat_stream_error_fallback.py` run alone:

| tree | result |
|---|---|
| `03ec0c898e` (last green) | 11 passed |
| `5f37b17848` (red main) | 10 failed, 1 passed |
| red main + widened stub | 10 failed, 1 passed |
| red main + moving the import | 11 passed |

The failure that surfaces is `duplicate file name google/cloud/firestore_v1/types/document.proto` rather than the missing symbol, because the harness's `sys.modules` restore drops the real firestore modules that got pulled in during the test and the next test re-imports them into a populated descriptor pool. That's why the first test in each file passes and the rest fail — a signature worth recognising if this class recurs.

Sorry for the duplicate effort; I should have re-checked `main` immediately before opening rather than trusting a fetch from a few minutes earlier.

PR #11941, opened 2026-08-20
fix(macos): restore sizing guard after Chat-first navigation

ketro155:
## What changed and why

Reapply the AppKit-owned main-window sizing contract whenever `ChatFirstShellNavigation.route` changes. Chat-first navigation does not mutate the legacy `selectedIndex`, so a replacement hosting view could resume automatic `minSize()` / `sizeThatFits()` traversal and keep a hidden window consuming CPU.

none

## How it was verified

- Reproduced on Omi 0.12.187: with the Settings window closed and capture idle, Omi used about 30–49% CPU; a 10-second process sample repeatedly entered `NSWindow.layoutIfNeeded`, `NSHostingView.layout`, and `NSHostingView.minSize`.
- `swift test --scratch-path /tmp/omi-ui-loop-build --filter DesktopHomeWindowSizingTests` — 2 tests passed, 0 failures.
- `python3 scripts/check_desktop_test_quality.py` — passed at the repository baseline.
- `./scripts/swift-format-wrapper.sh lint Desktop/Sources/MainWindow/DesktopHomeView.swift Desktop/Tests/DesktopHomeWindowSizingTests.swift` — passed with pinned swift-format 602.0.0.
- `git diff --check` — passed.
- Built, installed, and launched the isolated `/Applications/omi-ui-loop.app` bundle; production Omi was not modified or restarted.
- Limitation: the isolated bundle had no signed-in Omi Dev session, so the signed-in Chat-first route could not be exercised live without copying production credentials. The route hook is covered by a static wiring regression, and its AppKit effect is covered behaviorally using a real `NSWindow` and `NSHostingView`.

## Tests

`DesktopHomeWindowSizingTests` mounts a real hosting view in a real window and verifies that sizing reassertion clears automatic hosting size computation and restores the AppKit min/max policy. A companion regression pins the Chat-first route observer to that reassertion hook.

## Failure class (fixes)

Failure-Class: none

No existing registry class matches this missed Chat-first route reassertion. `FC-selection-overlay-layout-loop` covers transient selection-overlay state rebuilding selectable chat content and uses a different prevention primitive.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
@@ -484,11 +484,20 @@ struct DesktopHomeView: View {
 
   private func enforceMainWindowMinimumSize() {
     DispatchQueue.main.async {
-      // Appearance belongs to `WindowGlass.wear(_:as:)`; stamping one here unpinned the glass.
-      for window in NSApp.windows where window.title.lowercased().hasPrefix("omi") {
-        Self.pinShellWindowSizeLimits(window)
-        Self.disableMinSizeComputation(in: window)
-      }
+      Self.reassertMainWindowSizing(in: NSApp.windows)
+    }
+  }
+
+  /// Restores the AppKit-owned size contract after SwiftUI rebuilds a destination host.
+  ///
+  /// Chat-first navigation changes `ChatFirstShellNavigation.route`, not the legacy sidebar index.
+  /// Both paths must call this after a route change or the replacement hosting view resumes automatic
+  /// `minSize()` computation and every published state change traverses the full destination tree.
+  static func reassertMainWindowSizing(in windows: [NSWindow]) {
+    // Appearance belongs to `WindowGlass.wear(_:as:)`; stamping one here unpinned the glass.
+    for window in windows where window.title.lowercased().hasPrefix("omi") {
+      pinShellWindowSizeLimits(window)
+      disableMinSizeComputation(in: window)
     }
   }
 
@@ -1248,6 +1257,9 @@ struct DesktopHomeView: View {
         updateStoreActivity(for: newValue)
       }
       .onChange(of: chatFirstNavigation.route) { _, _ in
+        // Chat-first does not mutate `selectedIndex`, so the legacy navigation observer above cannot
+        // restore the NSHostingView sizing guard for this replacement destination.
+        enforceMainWindowMinimumSize()
         updateStoreActivityForCurrentShell()
         reportAutomationState()
       }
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Thanks — this is a clean fix and the root-cause analysis holds up against the current code. Verified on head 9e08b205:

- `DesktopHomeView.swift`: extracting the loop into `static reassertMainWindowSizing(in:)` is behavior-preserving — same `NSApp.windows` filter (`title.lowercased().hasPrefix("omi")`) and the same `pinShellWindowSizeLimits` + `disableMinSizeComputation` pair, now callable directly from tests. The new call in `.onChange(of: chatFirstNavigation.route)` correctly mirrors the legacy `selectedIndex` observer: Chat-first navigation genuinely doesn't touch `selectedIndex`, so nothing re-disabled `minSize()` computation on the replacement hosting view — which matches the hidden-window CPU signature you sampled (`NSHostingView.minSize` → full-tree `sizeThatFits`).
- `DesktopHomeWindowSizingTests.swift`: the behavioral test with a real `NSWindow` + `NSHostingView` covers the actual AppKit effect (sizing options cleared, min/max re-pinned to the `DesktopWindowLayoutPolicy` values), and the source-inspection wiring test follows the repo's established `omi-test-quality: source-inspection` pattern — a reasonable substitute given the private SwiftUI wiring has no runtime accessor. One small note: `range(of:)` pins the first `.onChange(of: chatFirstNavigation.route)` occurrence and the block-end fallback searches to end-of-file. There's exactly one observer today so it's fine; just worth knowing if that wiring is ever reorganized.
- `20260820-chat-first-hidden-window-cpu.json`: matches the existing unreleased-changelog format.

Good verification discipline too — the process-sample evidence and testing against an isolated bundle rather than a signed-in production install is exactly the right approach.

Since this is still a draft with no CI runs yet, leaving formal approval for when it's marked ready for review — the change itself reads correct and well-scoped.

---

aryanorastar:
Closing as superseded, not as stale.

On origin/main (1ef3b911) the Chat-first sizing hole this draft targeted is already closed. `selectedIndex` is gone. The root observer that used to re-pin on sidebar index now re-pins on `chatFirstNavigation.route` and calls `enforceMainWindowMinimumSize()` in DesktopHomeView (the onChange next to the appear / min-size comments). That is the same AppKit pair (`pinShellWindowSizeLimits` + `disableMinSizeComputation`) this PR extracts as `reassertMainWindowSizing`.

The nested `mainContentWithLifecycle` observer still does not call enforce — it does not need to, because the root observer already does. Rebasing this draft would either no-op or double-call the same function.

The new DesktopHomeWindowSizingTests would still be additive. If you want that NSWindow/NSHostingView assertion on main, reopen a focused test-only PR.

PR #11965, opened 2026-08-20
fix(backend): cap GET /v1/action-items at 12/min per uid

Git-on-my-level:
## Summary

Cap first-party `GET /v1/action-items` at **12 requests / 60s / uid** via existing `with_rate_limit` (Redis, multi-instance). Rejects in `Depends()` **before** Firestore.

Old Windows main-process `taskSyncEngine` listed this route in a tight loop (~120 qps fleet) with only `Authorization` — no `X-App-Platform` / `X-App-Version`, so a version gate cannot see it. A global 429 would also hit Mac/mobile.

12/min is above Mac/Flutter hydrate + a few pagination pages; a tight loop dies after 12 calls.

Does **not** force clients to update. `/ids` census is unchanged.

## Failure-Class

Failure-Class: none

## Root Cause

Windows main-process listing storm on `GET /v1/action-items` with no client version header. Instrumented `action_items_list` was ~1k/day because Cloud Run backend is not scraped.

## How to Verify

1. Unit: `test_action_items_list_caps_tight_loops` and existing router-policy scan.
2. After deploy: 429 with `Retry-After` on the 13th listing GET in a minute for one uid; Firestore read ops on this route should fall.

## Test Plan

- [x] Policy tuple test
- [ ] Existing `test_rate_limiting` router scan (CI)
- [ ] Prod: listing 429s for a hot uid; Mac/Flutter hydrate still works

## Risk Assessment

Medium — legitimate rapid pagination >12/min gets 429 for the rest of the window. First-party limiter **fail-opens** if Redis is down. `/ids` and writes are not capped by this policy.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11965?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/action_items.py
+++ b/backend/routers/action_items.py
@@ -352,7 +352,7 @@ def get_action_items(
     end_date: Optional[datetime] = Query(None, description="Filter by creation end date (inclusive)"),
     due_start_date: Optional[datetime] = Query(None, description="Filter by due start date (inclusive)"),
     due_end_date: Optional[datetime] = Query(None, description="Filter by due end date (inclusive)"),
-    uid: str = Depends(auth.get_current_user_uid),
+    uid: str = Depends(auth.with_rate_limit(auth.get_current_user_uid, "action_items:list")),
 ):
     """Get action items for the current user."""
     if start_date is not None and end_date is not None and _ensure_aware(start_date) > _ensure_aware(end_date):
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
Backend bug fix, confidence 5/5: real production incident (Windows listing storm), scoped diff (12 lines), CI green, test added, mirrors existing rate-limit pattern. Confirmed not fixed on main.

PR #11977, opened 2026-08-21
fix(ci): accept workflow_dispatch in the CI-impact predictor

sujeito-operator:
## What

`scripts/pre_push_ci_prediction.py` rejected `--event workflow_dispatch`, so every manual run of `desktop-swift-ci.yml` failed at **Detect changed paths** before writing a single output.

```
pre_push_ci_prediction.py: error: argument --event: invalid choice:
'workflow_dispatch' (choose from 'local', 'pull_request', 'push')
##[error]Process completed with exit code 2.
```

## Why it matters

`desktop-swift-ci.yml` declares `workflow_dispatch` deliberately, and its own `on:` comment explains why:

> Recovery hatch. `plan-desktop-release.py` requires this workflow's checks on the exact source SHA, so any interruption — the workflow being disabled, a run cancelled by concurrency, an infrastructure blip — leaves that SHA permanently unreleasable: push events do not replay, and neither a force-push nor a PR close/reopen produces a run for a commit already on main. Manual dispatch is the only way to re-mint the evidence without an unrelated commit.

The step forwards `${{ github.event_name }}` straight into the predictor, so the hatch failed on **every** manual run. The `changes` job is `needs:` for all three downstream jobs, so they were all skipped — the run ends red having verified nothing.

## The fix

`--event` choices now come from an `ACCEPTED_EVENTS` constant that includes `workflow_dispatch`.

This is behaviour-neutral by construction: `event` is threaded into `resolve_impact` but **no routing decision reads it**, so widening the accepted set cannot change any plan. Unknown values are still rejected, so the typo guard is intact.

## The regression test

`test_every_declared_workflow_trigger_is_an_accepted_event` derives the requirement from the workflows' own `on:` blocks rather than restating a list, so the next trigger added to any predictor caller fails this test instead of failing the first manual run.

Discovery is anchored on the script name, not on `--event` alone. An earlier draft matched `--event "${{ github.event_name }}"` anywhere and flagged `mobile_internal_build.yml`, which forwards `github.event_name` to `dispatch_mobile_internal_builds.py` — a different script with its own `--event` flag. That would have failed this test for a script it does not govern. The current form governs exactly the six workflows that reach the predictor directly or through `./.github/actions/detect-changes`.

Two smaller guards come with it: `local` must stay in the set (`scripts/pre-push` relies on the default), and the resolved plan must be identical across every accepted event — which is what makes the widening safe rather than merely convenient.

## Verification

- The previously failing invocation now exits 0; `--event bogus_event` still exits 2 naming the four valid choices.
- `pre-push-ci-prediction-tests` 25/25, `desktop-swift-ci-contract` 26/26, `desktop-manifest-routes` 4/4, `run_checks` 47/47.
- `python3 .github/scripts/run_checks.py --lane ci --changed-files <diff>` — all 14 selected manifest checks pass.
- **Non-vacuity, by mutation.** Removing only `"workflow_dispatch"` from `ACCEPTED_EVENTS` turns the new test red naming `desktop-swift-ci.yml`. Breaking the discovery patterns trips the explicit "predictor call sites moved" guard rather than silently passing on an empty set. Making `_declared_triggers` return `[]` trips the empty-triggers guard. Making `resolve_impact` branch on `event` turns the invariance test red. All four restored and re-run green.

## Not done

The observed failure is `desktop-swift-ci.yml` only. The other five predictor callers declare just `push` and `pull_request` today; they are covered by the test going forward but were never broken.

Failure-Class: FC-workflow-script-input-contract

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11977?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/scripts/test_pre_push_ci_prediction.py
+++ b/.github/scripts/test_pre_push_ci_prediction.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import importlib.util
+import re
 import sys
 import unittest
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
CI tooling fix, confidence 5/5: root cause pinned with exact error, scoped diff (103 lines), test derives requirement from workflows' own on: blocks + mutation-tested. Confirmed not fixed on main.

PR #11980, opened 2026-08-21
Recognize Hey Omi task commands at conversation finalization

Git-on-my-level:
## What changed and why

This replaces the first wake-word capture design with Fable's reviewed conjunction gate while retaining the deterministic matcher, structural marker, extraction guidance, and discard-prompt insurance.

- A deterministic match marks the full transcript and enables the existing wake-word extraction rules.
- Only conversations with at least one deterministic match receive a separate, independently framed, extended-reasoning adjudication call.
- The adjudicator classifies detected invocations from the full transcript, matched segment IDs, intersecting extracted items, and speaker labels. It never receives a pre-drafted intent.
- `explicit_command` reaches `create_direct` only when extraction and adjudication agree on a matched invocation. A rejecting verdict demotes an extracted explicit command to `direct_request`; an unambiguous `task_command` verdict may promote a non-explicit extracted item.
- A verdict never synthesizes a task, never changes `capture_confidence`, and never affects items without wake-word provenance.
- The old `direct_mention` wake-word overload is removed. Wake-word invocation and public-broadcast addressing retain separate semantics.
- The extraction prompt now resolves the single-topic collision in favor of the marked command classification and provenance when ambient discussion and a command share a topic.

The discard prompt rule remains intentionally. The base discard prompt explicitly targets short assistant-like utterances, so current model behavior is not treated as a stable guarantee.

## Decision required: adjudicator is unvalidated

The rebuilt evaluation used six realistic 20-60 segment, multi-speaker conversations, three arms, and three trials per case. It scored end-to-end `run_capture_policy` outcomes.

| Arm | `create_direct` on real commands (n=18) | False `create_direct` on non-commands (n=27) |
|---|---:|---:|
| Baseline | 14/18 (77.8%) | 0/27 (0%) |
| Marker only | 18/18 (100%) | 0/27 (0%) |
| Marker + adjudicator | 18/18 (100%) | 0/27 (0%) |

Stage two ran 15 times. The marker alone moved command capture from 77.8% to 100% with no measured precision cost. The adjudicator changed zero policy outcomes: its false-`create_direct` rate did not improve over marker-only, so the harness recommendation is **`drop_adjudicator`**.

The verdicts themselves were accurate on the hard cases: the bystander was `not_addressed_to_omi`; product discussion and quoted examples were `quoted_or_meta`; the demo was `not_addressed_to_omi`; the product question was `question`; and the paired quote/command was split correctly. That classification quality did not establish incremental end-to-end value. The adjudicator is therefore **unvalidated by this evaluation**, not credited with the marker-only gain. It remains in this draft solely for the requested product decision.

### Measurement repair after the live run

The original `ambient_items_set_equal` compared independently sampled baseline and marker completions byte-for-byte. Its failures, including `control_unchanged_trials: 0`, could not distinguish interference from ordinary model nondeterminism.

The harness now:

- compares marker-only with marker+adjudicator exactly because those arms share one extraction sample;
- compares baseline with marker-only as an across-trial distribution of ambient provenance, policy classification, and policy outcome, ignoring sampled description wording; and
- reuses the one extraction sample for the no-wake-word control, where transcript and prompt flags are identical.

The live outcome table above is unchanged by this assertion repair. The corrected ambient metrics have not been rerun against the live model in this commit.

## Prominent v1 descope: questions

`question` verdicts are logged and create nothing. The current extractor and policy route tasks, not answers, so `hey Omi <question>` remains unsupported even though it represents roughly one third of the stated product problem. Shipping this v1 requires explicit user sign-off on that descope.

## Evaluation fixtures

The live fixture set covers:

- an embedded mid-meeting command with a multi-segment payload and unrelated commitments;
- the Omi-team dogfood killer with eight wake-word mentions and exactly one real command;
- a bystander/TV command;
- paired real and quoted invocations;
- `omie`, `omni`, and welded mis-transcriptions in context; and
- a no-wake-word control proving stage two never fires.

Discard evaluation now asserts the durable requirement only: the marked arm must never be discarded. It does not encode the empirically false claim that the current model discards the unmarked arm.

INV-MEM-4 (canonical promotion is the sole Long-term authority)

Triggered by the path glob on `backend/utils/conversations/process_conversation.py`. The
invariant is not affected in substance:

- That file's diff contains no memory-related lines. It threads the renderer-owned wake-word
  marker into the existing discard and extraction calls and nothing else.
- No memory intake path is added anywhere in the PR. There is no new call to
  `create_memory` / `create_external_memory`, no tier assignment, and no promotion pass.
- The adjudicator's `memory_command` verdict is inert with respect to memory storage: it
  appears only in `_ACCEPTED_EXPLICIT_VERDICTS`, where its sole effect is to let an already
  extracted **action item** retain `capture_kind=explicit_command`. It creates no memory.
- No new canonical intake, no Short-term to Long-term transition, no fast-track promotion,
  and no change to consolidation routing.

Noted for a future PR, not introduced by this one: if a `memory_command` verdict is ever
wired to actually write a memory, it would become new canonical intake and must enter via
Short-term consolidation. It would also inherit the bystander-injection surface flagged in
the review, since the matcher deliberately does not gate on `is_user`.

## How it was verified

- `bash scripts/typecheck.sh` from `backend/` - 0 errors.
- `backend/.venv/bin/python -m pytest -q backend/tests/unit/test_wake_word.py backend/tests/unit/test_backend_candidate_capture.py backend/tests/unit/test_task_intelligence_contract_freeze.py` - 94 passed.
- `backend/.venv/bin/python -m pytest -q backend/tests/unit/test_process_conversation_usage_context.py` - 41 passed.
- `env -u ENCRYPTION_SECRET -u OPENAI_API_KEY backend/.venv/bin/python backend/scripts/task_intelligence_fixture_runner.py` - passed without a production secret, exercising the new public, database-free scoring boundary.
- Full backend file-isolated test run - all 891 selected test files passed.
- `make preflight` and `scripts/pr-preflight --pr-body-file /tmp/omi-wake-word-pr-body.md` - all 24 selected checks passed.

## Failure class (fixes)

Failure-Class: none

## Line-count exceptions

Line-Count-Exception: backend/utils/conversations/process_conversation.py | 2359 -> 2375 | reason: thread the renderer-owned marker through the existing discard and extraction lifecycle without creating a second conversation coordinator

Line-Count-Exception: backend/utils/llm/conversation_processing.py | 1731 -> 1751 | reason: keep trusted-marker prompt guidance beside the existing discard and action extraction owners instead of adding a parallel LLM pipeline


Git-on-my-level:
## Live paired evaluation: RUN — and it does not support the PR's central claim

The blocker is cleared: the eval ran against the real extractor (`--live-wake-word-eval --trials 3`, 12 paired trials, exit 0).

### Results

| metric | result |
|---|---|
| `paired_trials` | 12 |
| **`capture_kind_changed`** | **1 / 12** |
| `marked_expected_kind` | 12 / 12 |
| `marked_provenance_complete` | 9 / 9 (3 cases produced no items by design) |
| `ordinary_capture_kinds_unchanged` | 3 / 3 |
| `ordinary_extraction_preserved` | 3 / 3 |

Per case, marked vs. unmarked:

| case | changed | unmarked baseline |
|---|---|---|
| `omni_command_with_following_payload` | 0/3 | already `explicit_command` ×3 |
| `ordinary_extraction_survives_marked_conversation` | 0/3 | already `explicit_command` ×3 |
| `quoted_phrase_remains_non_actionable` | 0/3 | correctly empty ×3 |
| `multilingual_context_with_english_invocation` | **1/3** | `direct_request`, then `explicit_command` ×2 |

### Reading

**Safety is demonstrated.** Ordinary extraction is untouched (3/3), quoted phrases stay non-actionable (3/3), provenance is complete wherever items were produced (9/9). The regression risk this design was built to avoid does not materialise.

**Benefit is not.** In 11 of 12 paired trials the marker changed nothing, because the unmarked baseline *already* assigned `explicit_command`. The single change came from the multilingual case, and even there it was non-deterministic — 1 of 3 identical trials.

### Why: the fixtures don't reproduce the failure mode

The eval feeds clean synthetic text (`"Hey Omi, send Sarah the signed budget"`) where the extractor already succeeds unaided. The real-world failures this PR targets are different:

1. **`should_discard_conversation` eating short commands before extraction ever runs.** This is the defect the PR discovered and fixed, and it is plausibly the entire practical value here — **the eval contains no discard case at all.**
2. **Mis-transcribed wake words in messy transcripts** — `omni`, `omie`, glued `omilockets`. The fixtures use well-formed spellings, so the matcher's recall tuning is never exercised against the thing that motivated it.

### Recommendation

Keep this draft. The measurement gate did its job — it caught that the headline mechanism is close to a no-op on the cases as written. Before merge, re-target the eval at:

- a **discard-gate case**: a short conversation that is discarded without the marker and survives with it;
- **mis-transcribed variant** cases drawn from the scanner's real output rather than clean spellings.

If those show a real delta, the PR is justified on evidence. If they don't, the honest conclusion may be that the discard fix should ship alone and the marking machinery is unnecessary.

Git-on-my-level:
## Re-targeted evaluation: the discard hypothesis is falsified, and marking introduces a false positive

Eval re-targeted per review and re-run live (`2c26d951f4`). Both findings are negative for this PR.

### 1. Discard gate — `discard_changed: 0 / 9`

| case | unmarked discarded | marked kept | changed |
|---|---|---|---|
| `bare_short_command_whole_conversation` | **false ×3** | true ×3 | 0/3 |
| `mis_transcribed_omie_short_command` | **false ×3** | true ×3 | 0/3 |
| `non_actionable_wake_word_negative_control` | true ×3 | false ×3 | 0/3 |

**The premise was wrong.** Short "hey omi" commands are *not* being discarded in the first place — the unmarked arm keeps them 3/3. The discard-gate change rescues nothing.

This is a trustworthy zero, not a broken harness:
- Fixtures are 8–13 words, far under the `word_count > 100` short-circuit in `should_discard_conversation`, so the LLM was genuinely invoked rather than skipped.
- The negative control *is* discarded in both arms (3/3), proving the gate is live and that marking does not disable it wholesale.

### 2. Extractor — `capture_kind_changed: 5 / 27` (was 1/12)

One case is a genuine, deterministic win:

| case | changed | unmarked baseline |
|---|---|---|
| **`mid_utterance_omni_command`** | **3/3** | `direct_request` ×3 — consistently wrong without the marker |
| `heyomi_glued_command` | 1/3 | flaky baseline |
| `multilingual_context_with_english_invocation` | 1/3 | flaky baseline |
| `omilockets_glued_command`, `single_segment_entire_command`, `unpunctuated_lowercase_omie_command`, `omni_command_with_following_payload`, `ordinary_extraction_survives_marked_conversation` | 0/3 each | already `explicit_command` |

`mid_utterance_omni_command` ("okay before i go hey omni remind me to send sarah the budget friday") is real: wrong 3/3 unmarked, right 3/3 marked.

### 3. New finding — the marker caused a false positive

`quoted_phrase_remains_non_actionable`, transcript:

> The documentation should show the phrase Hey Omi as an example.

- **unmarked:** correctly extracts nothing, 3/3.
- **marked, trial 2:** produced `capture_kind=explicit_command`, description *"Add “Hey Omi” as an example in the documentation"*.

That is an unambiguous non-command promoted to `explicit_command`, which routes to `create_direct` — bypassing `MINIMUM_CAPTURE_CONFIDENCE` and creating a task with no review step. `marked_expected_kind` fell to 26/27 solely because of this trial.

The failure mode fires precisely when people *discuss* the phrase "hey omi" — i.e. during exactly the conversations this team has about this feature.

### Recommendation: do not merge as-is

Measured net: **one narrow reproducible win** (mid-utterance) against **a demonstrated capacity to fabricate `create_direct` tasks from discussion of the phrase**, and **zero benefit** from the discard change that motivated the design.

Options, in the order I'd rank them:
1. **Park the marking machinery.** ~1400 lines are not earned by one case.
2. **Fix the false positive first** — the marked-arm prompt rule about quoted mentions demonstrably does not hold — then re-run and re-decide.
3. Merge on safety grounds. Not advised: safety here means "usually harmless", and the one measured harm writes directly to the user's task list.

The evaluation harness itself is worth keeping regardless of the outcome — it is what caught both of these.

PR #11982, opened 2026-08-21
fix(desktop): make FTS5 query expansion produce valid queries

aryanorastar:
## Summary

`expandSearchQuery` produced FTS5 queries that fail to parse, so screen-history search returned nothing for ordinary window titles. Two distinct defects in the same function.

## 1. Groups joined by a space

FTS5 accepts implicit AND between bare terms — `a* b*` parses — but not once a word expands into a parenthesised group. Expansion joined every expanded word with a space, so any multi-word title where a single word split on camelCase or a number boundary produced `a* (b* OR c*)` and failed the **entire** search rather than matching less.

Reproduced standalone against a real FTS5 table before touching the code:

```
OK     activity* performance*                        (bare terms, implicit AND)
ERROR  quarterly* (activity* OR act*)                fts5: syntax error near "("
ERROR  (activity* OR act*) (performance* OR perf*)   fts5: syntax error near "("
OK     quarterly* AND (activity* OR act*)
```

Now joined with an explicit `AND`, which is what the space already meant.

## 2. Punctuation inside a bare term

Writing the regression test surfaced a second defect the first fix does not cover: a bare `2.1.220*`, built from a Cursor window titled `2.1.220`, is `fts5: syntax error near "."`. Tokenising now splits on any non-alphanumeric rather than whitespace alone.

This one only reached callers that pass raw text. `SuggestionAssistant` sanitises first via `SuggestionSearchTerm.sanitize` and never saw it; `RewindViewModel` — the user-facing Rewind search — passes window titles straight through and did. Fixing it at the tokeniser makes expansion safe for both instead of depending on every caller to pre-clean.

## Impact

Both callers of `RewindDatabase.search` were affected:

- **`RewindViewModel:404`** — the Rewind search UI. A user searching a multi-word phrase got an empty result set, not a smaller one.
- **`SuggestionAssistant:271`** — screen-history grounding. Observed failing 45 times across 13 sessions in local logs, silently emptying `relatedScreens`.

## Why the existing tests missed it

`RewindSearchQueryExpansionTests` asserted the *shape* of the expansion — balanced parens, no empty `()` group — and passed throughout. Shape assertions cannot see an invalid query.

The new tests execute the expansion against a real in-memory FTS5 table. That is what caught the second defect, and it is what keeps this class caught: `testRealWindowTitlesProduceExecutableQueries` runs seven real title shapes through a live matcher.

The empty-group defect this file already documents (`fts5: syntax error near ")"`) is the same failure mode one level down — a query that is structurally plausible and unparseable. That fix asserted shape; this one asserts parseability.

## Verification

```
swift test --filter RewindSearchQueryExpansionTests   -> 10 passed
swift build                                          -> clean
```

Full desktop suite on the branch this was developed on: 5503 passed, 1 failure — `RewindCaptureExclusionGenerationTests.testOwnerSnapshotStaysCurrentWhenAuthLeadsUnresolvedRewindDatabase`, pre-existing cross-suite ordering, reproduces identically with this diff reverted.

Not exercised through the Rewind search UI by hand — the fix is proven at the query layer, where the defect is, and the new tests run the real matcher. Saying so rather than implying a UI pass I did not do.

- INV-AUTH-1

Cited because `RewindDatabase.swift` is a path the invariant governs. Nothing here touches session ownership or the effective runtime owner: the change is confined to `expandSearchQuery`, a pure string function turning a search term into an FTS5 expression. It reads no auth state, and the owner scoping applied by `RewindDatabase.search` is unchanged.

## Failure class (fixes)

Failure-Class: none

Line-Count-Exception: desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift | 3470 -> 3481 | Adds the explicit-AND join and the non-alphanumeric tokenizer split, plus the comments recording why a space-joined query and a bare dotted term are each invalid FTS5.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
Desktop owner-override bug fix: FTS5 query expansion producing invalid queries, reproduced and fixed with real-matcher tests. CI green, confirmed not fixed on main.

PR #11984, opened 2026-08-21
fix(desktop): register the relocated notification bridge actions with flow lint

aryanorastar:
## Summary

`desktop-flow-lint` currently fails on `main`. Two bridge actions were relocated to a new file without adding that file to the contract list the lint reads, so both became invisible and the flow that references them fails.

## Cause

`d5596a6` ("refactor(macos): relocate journal copy + notification bridge actions to satisfy line-count ratchet") moved `settings_notifications_snapshot` and `set_notification_settings` out of `DesktopAutomationBridge.swift` into `DesktopAutomationBridge+Notifications.swift`, but left `ACTION_SOURCE_RELATIVE_PATHS` in `desktop/macos/scripts/desktop_flow_contract.py` unchanged.

The lint reads registered actions only from that list. `notifications-settings.yaml` still references both actions, so on current `main`:

```
desktop-flow-lint: 2 error(s)
```

This is main-wide rather than branch-specific: every desktop PR that merges current main inherits a red `desktop-core-e2e-t0`. It went unnoticed because main's own **Desktop Backend Contracts** runs are sitting at `action_required` rather than executing, so nothing re-ran the lint after the relocation.

## Fix

One line: add the relocated file to `ACTION_SOURCE_RELATIVE_PATHS`.

That file's own comment states the intent this restores — *"Keep the flow lint reader and CI impact resolver on this single list so an added bridge action cannot skip its flow validation route."* A relocated action is an added action as far as that list is concerned.

## Verification

- Pristine `origin/main` (`41afef9b`) reproduces both errors — this is not introduced by any branch.
- With this line: `desktop-flow-lint OK (72 flows, 162 registered actions)`, up from 161.
- **Guard proven live**: removing the line again reproduces the same 2 errors, so the fix is doing the work rather than the failure having moved.
- `desktop/macos/tests/test-e2e-flow-coverage.sh` — passed.
- `.github/scripts/test_pre_push_ci_prediction.py` — 22 passed (the other consumer of this contract).

## Honest gaps

No new test. The lint itself is the guard, it already runs in the `desktop-core-e2e-t0` lane, and it fails correctly when the entry is absent — a test asserting the list contains a specific string would be a static tripwire restating the data, not behavioural coverage.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11984?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/scripts/desktop_flow_contract.py
+++ b/desktop/macos/scripts/desktop_flow_contract.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 # validation route.
 ACTION_SOURCE_RELATIVE_PATHS = (
     "Desktop/Sources/DesktopAutomationBridge.swift",
+    "Desktop/Sources/DesktopAutomationBridge+Notifications.swift",
     "Desktop/Sources/FloatingControlBar/RealtimeHubController.swift",
     "Desktop/Sources/Rewind/Core/RewindArtifactGauntlet.swift",
     "Desktop/Sources/DesktopAutomationOpenOmiShortcutQA.swift",


aryanorastar:
Tracking issue with the full reproduction and the wider concern: #11985.

Also blocking #11804, which surfaced this by merging current main — that PR's own `desktop-flow-lint` failure is this bug, not its diff.

aryanorastar:
Closing — main fixed this independently while the PR was open, and this diff is now a literal no-op.

`4d7c209e` (#11962) added the same `Desktop/Sources/DesktopAutomationBridge+Notifications.swift` entry to `ACTION_SOURCE_RELATIVE_PATHS`. Verified against current main (`78028bf4`):

```
$ git diff origin/main HEAD -- desktop/macos/scripts/desktop_flow_contract.py
(no output — identical)

$ git archive origin/main | tar -x -C /tmp/mainlint && cd /tmp/mainlint
$ python3 desktop/macos/scripts/desktop-flow-lint.py
desktop-flow-lint OK (72 flows, 162 registered actions)
```

The bug was real and reproducible when I filed this: main was at `41afef9b` (2026-08-20 23:16) and failed both checks. `4d7c209e` landed at 00:09 the next morning, 53 minutes later, and carried the fix along with unrelated memory-TTL work. My reproduction was accurate for the main that existed at the time; it just stopped being the current main.

Nothing to merge here. Tracking issue #11985 updated with the same finding — leaving that open only for the wider point it raises, which #11962 does not address: nothing enforces that a file *defining* a bridge action appears on that list, so the next relocation defeats it the same way.

PR #11990, opened 2026-08-21
fix(desktop): find every bridge action source instead of listing them

aryanorastar:
## Summary

`desktop_flow_contract` exists so *"an added bridge action cannot skip its flow validation route"* — its own words. The list it relies on is hand-maintained, and it has now gone stale twice in two days.

When a bridge file is missing from the list, the lint cannot see the actions it registers and reports the flows using them as referencing actions that do not exist:

```
```

Because `scripts/pre-push` lints **every** flow rather than the diff, that fails the push gate on every branch cut from `main` — on a file the pusher never touched. I hit it pushing something unrelated and confirmed it reproduces on a clean `origin/main` checkout.

## This fixes the mechanism, not an outage

Both instances are already patched on `main` by hand — `+Notifications.swift`, then `+ChatFirst.swift`. `main`'s own comment records the pattern:

> main fixed +Notifications independently; +ChatFirst had the same gap.

That is the argument for this change. The list drifted, was patched, drifted again, was patched again. The next bridge extension will do the same, and take the push gate down with it for everyone.

## The change

Bridge sources are discovered by filename prefix instead of listed.

Scoped to that one prefix deliberately. The pattern the lint matches, `name: "…"`, also appears in `APIClient`, `AuthService` and `ChatBubble`, so scanning the tree would feed the lint names that are not actions and quietly weaken it. The prefix belongs to the bridge and nothing else — the curated list is the right shape, and only its bridge entries needed to stop being hand-maintained.

Three files, found rather than named:

| file | actions |
|---|---|
| `DesktopAutomationBridge.swift` | 125 |
| `DesktopAutomationBridge+Notifications.swift` | 3 |
| `DesktopAutomationBridge+ChatFirst.swift` | 0 |

## Verification

```
python3 desktop/macos/scripts/desktop-flow-lint.py
  -> desktop-flow-lint OK (72 flows, 164 registered actions)

python3 desktop/macos/tests/test_desktop_flow_contract.py   -> 3 passed
python3 .github/scripts/test_run_checks.py                  -> 47 passed
```

**Tests proven by reverting to the hand-maintained list**: two of the three fail and the lint reproduces the original errors. A test that has never failed is not evidence.

Wired into `.github/checks-manifest.yaml` in both lanes as `desktop-flow-contract-tests`.

none

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -234,6 +234,11 @@ checks:
     lanes: ["local", "ci"]
     platforms: ["macos"]
     reason: "desktop Swift sources changed; compare against the pinned macOS formatter before requiring a flow cover"
+  - id: desktop-flow-contract-tests
+    command: ["python3", "desktop/macos/tests/test_desktop_flow_contract.py"]
+    triggers: ["desktop/macos/scripts/desktop_flow_contract.py", "desktop/macos/tests/test_desktop_flow_contract.py", "desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "bridge actions moved into DesktopAutomationBridge+Notifications.swift while the action-source list still named only the base file, so desktop-flow-lint called their flows unknown and failed the push gate on every branch"
   - id: desktop-grdb-insert-idiom-tests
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
Desktop owner-override bug fix: hand-maintained bridge action list went stale twice, now discovered by filename prefix. CI green, confirmed not fixed on main.

PR #11991, opened 2026-08-21
harden(monitoring): alert on failures that keep every other signal green

Git-on-my-level:
## What this does

Adds five Grafana alert rules for production failures that keep every existing signal green,
plus a runbook that explains each mechanism.

## Why

The 2026-08-19 desktop chat outage ran for roughly 19 hours before a human noticed. Nothing that
any existing rule could observe was down. Four independent mechanisms hid it, and each one is a
general failure mode rather than a one-off:

**1. A 2xx committed before the failure happened.** The SSE endpoint writes `HTTP 200` and its
headers, then runs the generator. The generator yielded an in-band error frame followed by
`data: [DONE]` and returned normally. Every status-code monitor read the turn as a success.

**2. The failure was counted somewhere nobody watched.** The gateway rejected these requests during
request validation, *before a route was selected*. Pre-route rejections increment
`llm_gateway_request_rejections_total`, never `llm_gateway_requests_total`. Verified against
production Prometheus:

```
sum by (outcome,phase,error_class) (increase(llm_gateway_requests_total{lane_id="omi:auto:chat-agent"}[7d]))
  9702.6  {outcome="success", phase="terminal_marker", error_class="none"}
     1.0  {outcome="success", phase="terminal",        error_class="none"}
```

Zero errors. For the entire outage the chat lane reported 100% success. Any alert built on a
success ratio over `llm_gateway_requests_total` was structurally incapable of firing.

**3. The alert that should have caught it had a dead wire.** `omi-journey-chat-fail` already exists
and is shaped correctly. But its input is empty:

```
sum by (journey) (increase(omi_journey_accepted_total[7d]))
  451358.2  {journey="pusher_session"}
  137196.0  {journey="capture_finalization"}
       0.0  {journey="chat_response"}
```

`backend/routers/chat.py:529` does emit `JourneyAttempt('chat_response')` — from the Cloud Run
`backend` service, which Prometheus does not scrape. Prometheus scrapes GKE pods only. The counter
has been flat at zero for its whole existence. The rule did not fail loudly; it went quiet, which
looks exactly like health.

**4. A healthy majority masked a dead minority.** One client population failed 100% of turns while
a larger population on the same endpoint stayed healthy, keeping every aggregate ratio green.

## The rules

| UID | Signal | Gate |
|---|---|---|
| `omi-llm-gateway-invalid-request-rejections` | Pre-route validation rejections | `>= 2` in 30m, for 30m |
| `omi-llm-gateway-lane-failure-ratio` | Lane failure share | `$A >= 20 && $B > 0.25` over 1h |
| `omi-llm-gateway-lane-zero-success` | Lane with attempts and no successes | `$A >= 20 && $B < 1` over 6h |
| `omi-journey-signal-dead` | A journey counter stopped reporting | 0 accepted in 1h while gateway serves > 100 |
| `tz_chat_agent_requests_zero` | Chat lane traffic floor | `< 5` requests in 1h, for 30m |

## Where an operator looks

Three panels were added to **Resilience / Fallbacks**, and every rule now links to the panel that
plots its own evidence rather than a neighbouring one:

| Panel | Shows |
|---|---|
| 12 | LLM gateway pre-route rejections — the counter lane success rate cannot see |
| 13 | LLM gateway lane success rate, zero-filled so a never-successful lane plots at 0 instead of vanishing |
| 14 | Journey signal liveness — a flat zero line means that journey's alerts are dead, not that the path is healthy |

A contract test asserts each rule's linked panel actually queries that rule's metric, so the alert
and the dashboard cannot drift apart.

## Proof, not assertion

Every expression was executed against production Prometheus before it was written into a rule.

**`omi-journey-signal-dead` detects the exact dead wire that hid this outage.** Running its
expression with `chat_response` added to the selector:

```
1.0  {journey="chat_response"}        ← dead
0.0  {journey="pusher_session"}       ← healthy
0.0  {journey="capture_finalization"} ← healthy
```

**`omi-llm-gateway-lane-zero-success` finds a live problem nobody is watching.**
`omi:auto:translation` has served **3,780 requests and zero successes** across the full 10-day
Prometheus retention window — every one a `provider_request_rejected` / `other_4xx` from
`gemini-2.5-flash-lite`, `phase="before_output"`, no gateway fallback. Its rule inputs evaluate
to `A = 20.03, B = 0.0`, so the rule fires.

To be precise about impact, because the lane counter alone would overstate it: translation has an
application-level provider chain, and NLLB is absorbing the traffic. End to end,
`omi_translation_requests_total` shows **91,269 requests in 24h, all served by `nllb`**, with 834
requests where both providers errored. So this is not a user-facing translation outage — it is a
primary provider that has been failing 100% of the time for at least ten days, with the product
silently running on its fallback and no redundancy left. That is worth knowing and nobody knew it.
It needs its own fix; I have not touched it here.

The `or ... * 0` zero-fill in that rule is load-bearing: a lane that has *never* succeeded has no
`outcome="success"` series at all, so a plain ratio produces no series and therefore no alert.
Total failure would be invisible without it.

## Thresholds are measured, not guessed

`omi:auto:chat-agent` over the preceding 7 days, sampled at 15-minute resolution: floor 9
requests/hour, p01 15, p05 23, median 55. Zero evaluations below 5. The traffic floor is set to 5.

(An earlier revision of this PR said the floor was 18. That came from hour-aligned sampling, which
misses partial windows. The threshold stands but the margin is half what that number implied.)

The rejection threshold is deliberately low because the affected population was small — it produced
only 2–16 rejections per 30 minutes. A rate threshold sized for the busy path would not have fired.

The 6h window on the zero-success rule exists because several lanes are idle most hours;
`omi:auto:translation` has a median of 0 requests/hour and only crosses a 20-request gate in 24 of
168 hours. The 1h and 6h rules run together: fast on busy lanes, eventually correct on quiet ones.

## Deliberate omission

`omi-journey-signal-dead` does **not** include `journey="chat_response"`, and there is a test
asserting it does not. That counter is emitted from Cloud Run and will read zero until an ingestion
path exists. Adding it now would page permanently and get muted, which is worse than the gap it
describes. It goes in with the Cloud Run ingestion change.

## Verification

```
backend/tests/unit/test_monitoring_alert_rule_contract.py    18 passed
+ test_monitoring_telemetry_contract.py, test_journey_observability.py   34 passed total
make preflight                                               25 checks passed in 82.77s
```

Five new contract tests pin the expressions, the zero-fill, the `Alerting` no-data state, the
measured threshold, and the chat_response omission.

## Backtest — replayed against 7 days of production

A rule that would have been noisy in the past week is not ready to page anyone, and a muted rule is
worse than no rule. Every rule was replayed at 15-minute resolution:

| Rule | Breaching evaluations | What it caught |
|---|---|---|
| `omi-llm-gateway-invalid-request-rejections` | 58 / 673 (8.6%) | The outage window, and nothing after the fix deployed |
| `omi-llm-gateway-lane-failure-ratio` | 100 / 21,314 (0.47%) | `omi:auto:translation` only |
| `omi-llm-gateway-lane-zero-success` | 350 / 21,472 (1.6%) | `omi:auto:translation` and `omi:auto:web-search` |
| `omi-journey-signal-dead` | 0 / 1,346 | No false positives |
| `tz_chat_agent_requests_zero` | 0 / 673 | No false positives |

Every breach in that table is a real defect, not noise.

**The backtest found a third silent failure.** `omi:auto:web-search` recorded **170 errors and 1
success over 7 days**, every failure `error_class="invalid_route_config"`, going back at least to
08-12. Its single successful request is the most recent data point in the series, after the managed
Perplexity lane deployed — traffic since has been too thin to call it fixed. Nine days of a
completely dead lane, unnoticed.

## Follow-ups this does not do

- **Cloud Run metrics have no ingestion path.** This is the root cause of the silence and is the
  single highest-value fix remaining. Separate PR in progress.
- `omi:auto:translation` is 100% broken and needs a real fix.
- `desktop-backend` has no `METRICS_SECRET`, so its `/metrics` returns 401.

Failure-Class: none


kodjima33 (approved):
New monitoring/alerting capability (5 new alert rules), not a bug fix to app code — approve-only. Also DIRTY and 1965 lines, well beyond scope for auto-merge.

PR #11992, opened 2026-08-21
fix(backend): add local harness storage

### LINKED ISSUE #11703: Local dev harness has no path to Google Cloud Storage — all GCS-backed uploads silently fail
## Symptom

Any feature backed by Google Cloud Storage fails on the local dev harness (`make dev-up`). Two concrete failures from the same session:

```
INFO:     127.0.0.1:55968 - "POST /v3/upload-audio HTTP/1.1" 500 Internal Server Error
RuntimeError: BUCKET_SPEECH_PROFILES is not configured
  backend/utils/other/storage.py:114, via routers/speech_profile.py:91
```

```
WARNING:google.auth.compute_engine._metadata:Compute Engine Metadata server unavailable on attempt 3 of 3. Reason: [Errno 64] Host is down
WARNING:google.auth._default:Authentication failed using Compute Engine authentication due to unavailable metadata server.
ERROR:utils.conversations.process_conversation:Error creating audio files: Your default credentials were not found.
```

The second one fired on **every single conversation processed** during the session (multiple different conversation IDs), not just once — conversation audio is never actually persisted to storage in this harness.

## Root cause — not a missing bucket name, the harness never authenticates to GCS at all

`backend/utils/other/storage.py:_get_storage_client()` builds a single global GCS client, authenticating via `SERVICE_ACCOUNT_JSON` if set, otherwise falling back to Application Default Credentials (which tries `GOOGLE_APPLICATION_CREDENTIALS`, then the GCE metadata server).

`backend/.dev.env` does set `GOOGLE_APPLICATION_CREDENTIALS=google-credentials.json`, and a real service-account key file exists on disk at that path. But the harness never sources it. `scripts/dev-harness/_source_local_dev_env.sh` says so directly:

> "Child processes receive a fully-formed env from the harness... they do not load `backend/.env` or stage files on disk."

`SECRETS_FILE_ALLOWED_KEYS` in `scripts/dev-harness/dev_harness/config.py` only ever forwards 5 keys to the backend process: `PROVIDER_MODE`, `OPENAI_API_KEY`, `DEEPGRAM_API_KEY`, `GEMINI_API_KEY`, `ANTHROPIC_API_KEY`. Nothing GCS-related is in scope. So it doesn't matter which `BUCKET_*` env var you configure — there is no credential path to GCS at all in this harness, by design, not by oversight-that-broke.

## Why this matters beyond speech profiles

Every caller of `_get_storage_client()` is affected identically: `BUCKET_SPEECH_PROFILES` (voice-print upload), `BUCKET_MEMORIES_RECORDINGS`/`BUCKET_POSTPROCESSING` (conversation audio files — confirmed failing above), `BUCKET_CHAT_FILES` (chat attachments), `BUCKET_APP_THUMBNAILS`, `BUCKET_BACKUPS`, `BUCKET_PLUGINS_LOGOS`, `BUCKET_DESKTOP_UPDATES`.

Most call sites already degrade gracefully (`required=False`, caught exceptions, a warning log) — conversation processing completes and the app keeps working. The one call site that didn't degrade (`upload_profile_audio`'s `required=True`) had a real, separate bug: it silently blocked the onboarding completion screen from ever appearing (fixed locally, not yet upstreamed). But the underlying storage failure itself is untouched by that fix — the audio/profile data genuinely never lands anywhere.

**Practical impact for anyone using the local harness:** conversation audio playback, speech-profile-based speaker recognition, and any file-upload feature are silently non-functional. Nothing crashes, so this is easy to miss until someone specifically checks whether the data persisted.

## What already exists that could close this

`backend/testing/e2e/fakes/storage.py` is a fake GCS double already used by the pytest e2e suite (see `test_storage_speech_profile.py`, which asserts URLs like `https://fake-gcs.local/speech-profiles/...`). The codebase already knows how to fake this layer for tests — it's just never been wired into the *interactive* `make dev-up` harness the way the Firestore and Auth emulators are.

## Possible directions (maintainer call — not proposing a specific one)

- Wire the existing e2e fake storage client into the interactive harness (e.g. behind `PROVIDER_MODE=offline` or a new flag), so uploads succeed locally without real GCS credentials.
- Support a real local GCS emulator (e.g. `fake-gcs-server`) as an optional harness service, mirroring the Firestore/Auth emulator pattern.
- At minimum, document this as a known limitation in the dev-harness setup docs so contributors don't spend time debugging "why didn't my upload work."

## Environment

macOS, local dev harness (`make dev-up`), `PROVIDER_MODE=real`. LLM-backed features (chat, conversation processing, folder/app assignment, onboarding AI-check) all work correctly in the same session — this is isolated to GCS-backed storage specifically.

shubhxho:
## What changed and why

Give the owned local dev harness a filesystem-backed implementation of the backend's GCS surface, wire every bucket binding into the child environment, and serve uploaded artifacts from the real FastAPI app. This closes #11703 without allowing the adapter outside a demo Firebase project plus Firestore emulator.

none

## How it was verified

- `PYTHONPATH=backend backend/.venv/bin/pytest -q backend/tests/unit/test_local_storage.py` — 5 passed.
- `backend/.venv/bin/pytest -q scripts/dev-harness/tests/test_env_stage.py` — 5 passed.
- Imported the production `main:app` under a real harness child environment, called `upload_profile_audio`, and fetched the returned `/_local/storage/...` URL through `TestClient`; the route returned HTTP 200 with the exact uploaded `local-audio` bytes.
- `bash test.sh` — selected all 890 backend test files. The sandboxed run reached only four infrastructure failures (two missing-Helm render tests and two loopback-socket tests); after installing Helm, `BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-backend-unit-failures.txt bash test.sh` outside the socket sandbox reran those exact files and passed 10 + 2 + 15 + 22 tests.
- `make preflight` outside the process/socket sandbox — dev-harness tests passed 94 with 6 skipped; all remaining selected repository contracts are run again with this PR body.

The complete `make dev-up` stack could not be launched on this machine because Java, `redis-server`, and a running Docker daemon are absent. The actual production upload helper and mounted FastAPI route were exercised directly instead.

## Tests

`backend/tests/unit/test_local_storage.py` is the regression suite for the missing credential-free storage boundary, including round-trip bytes, production helper URL behavior, listing/copy/streaming, path containment, and the no-cloud-client production URL path. `scripts/dev-harness/tests/test_env_stage.py` pins the bucket and storage bindings in the actual child environment.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11992?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/AGENTS.md
+++ b/backend/AGENTS.md
@@ -13,7 +13,7 @@ source .venv/bin/activate
 uvicorn main:app --host 0.0.0.0 --port 8080
 ```
 
-**Env stages** (`OMI_ENV_STAGE`): `local` (emulator harness, `.env.local-dev`), `offline` (fake providers, `.env.offline`), `dev` (remote dev GCP, `.env.dev`), `prod` (reference only, `.env.prod`). `load_backend_env()` loads the stage file then `backend/.env` overrides. Templates: `backend/.env.*.template`. Harness: `PROVIDER_MODE=offline make dev-up` or `OMI_ENV_STAGE=offline`. Dev skips the startup-only Stripe price validation; plan catalog and checkout calls still require mode-matched Stripe credentials.
+**Env stages** (`OMI_ENV_STAGE`): `local` (emulator harness, `.env.local-dev`), `offline` (fake providers, `.env.offline`), `dev` (remote dev GCP, `.env.dev`), `prod` (reference only, `.env.prod`). `load_backend_env()` loads the stage file then `backend/.env` overrides. Templates: `backend/.env.*.template`. Harness: `PROVIDER_MODE=offline make dev-up` or `OMI_ENV_STAGE=offline`; both modes use instance-scoped filesystem storage under the owned harness state root, so GCS-backed uploads work without cloud credentials. Dev skips the startup-only Stripe price validation; plan catalog and checkout calls still require mode-matched Stripe credentials.
 
 When intentionally changing backend Python dependencies, edit the relevant `requirements*.txt` input file and refresh the lock:
 
--- a/backend/main.py
+++ b/backend/main.py
@@ -14,6 +14,7 @@ logger = logging.getLogger(__name__)
 import firebase_admin
 from fastapi import FastAPI
 from starlette.middleware.cors import CORSMiddleware
+from starlette.staticfiles import StaticFiles
 
 from database.google_credentials import prepare_google_credentials
 
@@ -106,6 +107,7 @@ from services.conversation_finalization import reconcile_listen_finalization_job
 from services.conversation_finalization import reconcile_meeting_receipts
 from services.conversation_finalization import reconcile_stale_processing_conversations
 from services.users.account_deletion import reconcile_pending_deletion_wipes
+from utils.other.local_storage import local_storage_root_from_env
 
 # Log LangSmith tracing status at startup
 log_langsmith_status()
@@ -130,6 +132,11 @@ else:
     firebase_admin.initialize_app(options=_firebase_admin_options)  # type: ignore[reportUnknownMemberType]  # firebase_admin untyped
 
 app = FastAPI()
+
+_local_storage_root = local_storage_root_from_env()
+if _local_storage_root is not None:
+    _local_storage_root.mkdir(parents=True, exist_ok=True)
+    app.mount('/_local/storage', StaticFiles(directory=_local_storage_root), name='local-storage')
 
 # Explicit, default-deny CORS: this API is Bearer-token authenticated (mobile/
 # desktop apps, not ambient browser cookies), so no cross-origin browser
--- a/backend/tests/unit/test_local_storage.py
+++ b/backend/tests/unit/test_local_storage.py
@@ -1,0 +1,91 @@
+from pathlib import Path
+
+import pytest
+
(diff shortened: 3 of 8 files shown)


kodjima33 (approved):
Bug/feature-ambiguous dev-harness storage addition; diff (360 lines) exceeds the 300-line confidence-gate ceiling despite good root cause/tests/CI — approve-only, no user-facing production risk.

kodjima33 (approved):
fix(backend): closes #11703 (confirmed open, unfixed on main — dev harness had no GCS-backed storage path). Root cause clearly stated, scoped to the local-harness-only code path (gated behind demo-project + Firestore-emulator env checks), 5 new regression tests, no failing required CI checks (remaining checks still pending/green). Confidence 4/5 (diff slightly over the 300-line guideline at 360, everything else checks out). Merging.

PR #11995, opened 2026-08-21
fix(backend): remove dead Crisp unread endpoint

### LINKED ISSUE #11429: Backend: /v1/crisp/unread has no callers after the desktop Crisp removal
PR #11424 removed the desktop Crisp support chat, which was the only caller of `GET /v1/crisp/unread` (`backend/routers/desktop_screen_crisp.py`, mounted at `backend/main.py:220`). The Flutter app has no Crisp code — `grep -rni crisp app/lib --include='*.dart'` returns nothing.

So the router, its `CRISP_PLUGIN_IDENTIFIER`/`CRISP_PLUGIN_KEY` env wiring, and `backend/tests/unit/test_desktop_screen_crisp.py` are now dead. Left in place in #11424 rather than widening a desktop UI change into a backend deploy.

Worth confirming nothing external polls it (older shipped desktop builds still will, until they update — so this probably wants to outlive at least one release window before removal).

shubhxho:
Closes #11429.

## What changed and why

Remove the retired `GET /v1/crisp/unread` implementation while preserving the live screen-activity route on the same router. Remove its route-policy entry and record the intentional API removal in the app-client compatibility checker.

none

## How it was verified

- `PYTHONPATH=backend pytest -q backend/tests/unit/test_desktop_screen_crisp.py` — 10 passed, including a route-level assertion that `/v1/crisp/unread` is now absent while the screen-activity suite remains green.
- `make preflight` — repository PR contracts passed.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/11995?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/route_policy_manifest.yaml
+++ b/backend/route_policy_manifest.yaml
@@ -81,10 +81,6 @@ routes:
     path: /v1/config/api-keys
     policy: *desktop_migration_policy
   - route_type: http
-    method: GET
-    path: /v1/crisp/unread
-    policy: *desktop_migration_policy
-  - route_type: http
     method: POST
     path: /v1/proxy/gemini-stream/{path:path}
     policy: *desktop_migration_policy
--- a/backend/routers/desktop_screen_crisp.py
+++ b/backend/routers/desktop_screen_crisp.py
@@ -1,27 +1,19 @@
-import base64
-import json
 import logging
-import os
-import time
 from typing import Any
 
-import httpx
-from fastapi import APIRouter, Depends, HTTPException, Query
+from fastapi import APIRouter, Depends, HTTPException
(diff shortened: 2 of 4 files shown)


kodjima33 (approved):
Backend fix removing dead endpoint, confidence 5/5: closes #11429, scoped diff, regression test added, confirmed endpoint still present on main so not redundant.

PR #12000, opened 2026-08-21
feat(backend): add button event webhook dispatch

### LINKED ISSUE #11719: feat: expose Omi button events to developer integrations (button_event webhook)
## Summary

Developer integrations can subscribe to **what the user says** (realtime transcript, memory created, audio bytes, day summary) but not to **what the user does with the device**. The pendant already emits single/double/long tap over BLE and the app already listens to them, so the gesture reaches the phone and stops there.

This proposes an **opt-in `button_event` developer webhook**: a thin forward of the physical gesture, with no semantics attached.

This is the "Button action events sent to backend for app integrations" checkbox of #5466, split out so it can be designed on its own — it is the only item there that changes a public developer contract.

## Why this is different from the other triggers

Transcript-based triggers can't express intent that isn't speech. A gesture is unambiguous, silent, and available when speaking isn't (meeting, noisy floor, privacy). Today the only way a developer can react to a tap is to ask the user to *say* something instead — which is exactly the interaction the button exists to avoid.

Concrete integrations this unlocks: push-to-mark ("bookmark this moment" → my tool), silent capture triggers, hands-free confirmation in a workflow the developer owns, accessibility flows where speech is not an option.

## The events already exist end to end

Firmware — `omi/firmware/omi/src/lib/core/button.c`:

```c
static struct bt_uuid_128 button_uuid =
    BT_UUID_INIT_128(BT_UUID_128_ENCODE(0x23BA7924, 0x0000, 0x1000, 0x7450, 0x346EAC492E92));
static struct bt_uuid_128 button_characteristic_data_uuid =
    BT_UUID_INIT_128(BT_UUID_128_ENCODE(0x23BA7925, 0x0000, 0x1000, 0x7450, 0x346EAC492E92));
...
#define SINGLE_TAP 1
#define DOUBLE_TAP 2
#define LONG_TAP   3
```

App — the subscription is already there:

- `app/lib/services/devices/connectors/omi_connection.dart` → `performGetBleButtonListener`
- `app/lib/services/capture/capture_controller.dart` → `streamButton` decodes `buttonState` and dispatches the native actions

So no firmware change and no second BLE connection are needed: the missing piece is a forward from a listener that already runs.

## Proposed contract

Same shape as the existing developer webhooks:

```json
{
  "event_type": "button_event",
  "button_event": "single_tap | double_tap | long_tap",
  "device_id": "<stable device identifier>",
  "event_id": "<uuid, stable across retries>",
  "timestamp": "<ISO-8601 UTC>",
  "session_id": "<optional, when a capture session is active>"
}
```

Two properties that matter more than they look:

- **`event_id` must be stable across retries.** A retransmitted first double tap that arrives with a new id is indistinguishable from a *second* double tap. Any consumer building a two-step gesture (arm → confirm) is silently broken by that, so the id has to identify the physical gesture, not the delivery attempt.
- **Ordering must be preserved per device.** Gestures are meaningful as sequences; delivering #2 before #1 because of network latency changes what the sequence means. A serial per-device queue is enough.

## Behavior when not configured

- Webhook unset → nothing is sent and nothing changes. No new BLE traffic, no extra battery cost.
- The native button actions (mute, star, end conversation, voice question, power off) stay exactly as they are. This proposal does **not** ask for the native actions to be given up.
- `long_tap` is forwarded as an event but must never suppress power-off.

## What this deliberately does NOT include

`button_event` is a **hardware gesture**, not an authorization, an intent or a command. The platform should attach no meaning to it — the consumer decides what a double tap means in their own app, under their own rules. Keeping semantics out of the platform is what makes one event type serve every integration.

## Open question for maintainers (why I'm filing an issue, not a PR)

Where should the dispatch live?

1. **App-side forward** — the app posts to the configured webhook from the existing BLE listener. Smallest change, works offline-independent of the backend, but puts retry/ordering logic in the client and each platform (iOS/Android/desktop) repeats it.
2. **Backend dispatch** — the app reports the gesture to the Omi backend, which fans out through the same plumbing that already serves the other webhook types. Consistent with how developers register and receive events today, one retry policy, one audit trail; costs a round trip and a new client→backend endpoint.

(2) looks more consistent with the existing developer-integration architecture, but it touches the backend contract, so it's your call rather than mine.

I have a working implementation of the client half (BLE listener → serial queue → stable `event_id` → bounded retry with no late replay) running against a private endpoint, and I'm happy to open a PR for whichever direction you prefer — opt-in, no behavior change when unconfigured, with tests. Just tell me which architecture you want before I write the platform-facing part.

Related: #5466 (parent), #11193.

shubhxho:
## Summary

- add an authenticated app-to-backend endpoint for physical Omi button gestures
- expose opt-in `button_event` webhook settings and forward the documented event payload with the physical gesture ID as the delivery idempotency key
- serialize deliveries per user/device while keeping the lock registry weakly held, so gesture order is preserved without an unbounded module cache
- validate gesture kind, UUID event identity, device/session lengths, and timezone-aware timestamps at the public boundary
- regenerate the released app-client OpenAPI contract plus Dart, Swift, and TypeScript DTOs

Closes #11719

## Verification

- `backend/.venv/bin/pytest -q backend/tests/unit/test_button_event_webhook.py backend/tests/unit/test_webhook_auto_disable.py` (123 passed)
- app-client, public, and integration-public OpenAPI contract checks passed
- Dart, Swift, and TypeScript OpenAPI generator checks passed
- `cd web/app && bun run check` (54 files, 305 tests passed after one unrelated timing-sensitive retry)
- `xcrun swift build -c debug --package-path Desktop` was attempted twice; dependency artifact downloads timed out before compilation, so the generated Swift client is generator-verified but not locally compiled
- `OMI_PR_BODY_FILE=/tmp/omi-pr-body-11719.md make preflight`
- `scripts/pr-preflight --pr-body-file /tmp/omi-pr-body-11719.md`

## Product invariants

None.

Line-Count-Exception: backend/routers/users.py | 2177 -> 2204 | existing developer-webhook routes and their shared response models are owned here; moving this one endpoint would split one API surface across routers

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12000?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/backend/schema/gen/users_wire.g.dart
+++ b/app/lib/backend/schema/gen/users_wire.g.dart
@@ -28,12 +28,14 @@ class GeneratedUserStatusResponse {
 
 class GeneratedUserWebhooksStatusResponse {
   final bool audioBytes;
+  final bool buttonEvent;
   final bool daySummary;
   final bool memoryCreated;
   final bool realtimeTranscript;
 
   const GeneratedUserWebhooksStatusResponse({
     required this.audioBytes,
+    this.buttonEvent = false,
     required this.daySummary,
     required this.memoryCreated,
     required this.realtimeTranscript,
@@ -42,6 +44,7 @@ class GeneratedUserWebhooksStatusResponse {
   factory GeneratedUserWebhooksStatusResponse.fromJson(Map<String, dynamic> json) {
     return GeneratedUserWebhooksStatusResponse(
       audioBytes: _required(_readFieldValue<bool>(_readField(json, const ["audio_bytes"]), "audio_bytes", _readBool, requiredField: true, nullable: false), "audio_bytes"),
+      buttonEvent: _required(_readFieldValue<bool>(_readField(json, const ["button_event"]), "button_event", _readBool, requiredField: false, nullable: false, defaultValue: false), "button_event"),
       daySummary: _required(_readFieldValue<bool>(_readField(json, const ["day_summary"]), "day_summary", _readBool, requiredField: true, nullable: false), "day_summary"),
       memoryCreated: _required(_readFieldValue<bool>(_readField(json, const ["memory_created"]), "memory_created", _readBool, requiredField: true, nullable: false), "memory_created"),
       realtimeTranscript: _required(_readFieldValue<bool>(_readField(json, const ["realtime_transcript"]), "realtime_transcript", _readBool, requiredField: true, nullable: false), "realtime_transcript"),
@@ -51,6 +54,7 @@ class GeneratedUserWebhooksStatusResponse {
   Map<String, dynamic> toJson() {
     return {
       'audio_bytes': audioBytes,
+      'button_event': buttonEvent,
       'day_summary': daySummary,
       'memory_created': memoryCreated,
       'realtime_transcript': realtimeTranscript,
--- a/backend/models/users.py
+++ b/backend/models/users.py
@@ -11,6 +11,7 @@ class WebhookType(str, Enum):
     realtime_transcript = 'realtime_transcript'
     memory_created = ('memory_created',)
     day_summary = 'day_summary'
+    button_event = 'button_event'
 
 
 def webhook_url_from_setting(wtype: WebhookType | str, value: Optional[str]) -> str:
--- a/backend/route_policy_manifest.yaml
+++ b/backend/route_policy_manifest.yaml
@@ -81,6 +81,30 @@ routes:
     path: /v1/config/api-keys
     policy: *desktop_migration_policy
   - route_type: http
+    method: POST
+    path: /v1/users/developer/button-event
+    policy:
+      review_status: reviewed
+      auth:
+        mechanisms:
+          - firebase_id_token
+          - admin_key_uid_prefix
+        placement: dependency
+        scopes: []
+      byok: not_applicable
+      rate_limit:
+        policy_name: none
+        key_subject: none
+        enforcement: none
+        placement: none
+      timeout_class: default_method
+      surface: first_party_app
+      visibility: first_party
+      data_domain: user_profile
+      deprecation:
+        state: active
+      owner: backend
+  - route_type: http
     method: GET
     path: /v1/crisp/unread
     policy: *desktop_migration_policy
--- a/backend/routers/users.py
+++ b/backend/routers/users.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import re
 import uuid
-from typing import List, Dict, Any, Union, Optional
+from typing import List, Dict, Any, Literal, Union, Optional
 import hashlib
 import os
 import asyncio
@@ -10,7 +10,7 @@ import asyncio
 import pytz
 from fastapi import APIRouter, Depends, Header, HTTPException, Query, Request
 from fastapi.responses import JSONResponse, StreamingResponse
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import AwareDatetime, BaseModel, ConfigDict, Field
 
 from database import (
     conversations as conversations_db,
@@ -124,7 +124,7 @@ from utils.other.storage import (
     delete_user_person_speech_samples,
     delete_user_person_speech_sample,
 )
-from utils.webhooks import webhook_first_time_setup
+from utils.webhooks import button_event_webhook, webhook_first_time_setup
 from utils.byok import has_byok_keys, invalidate_byok_state_cache, peppered_fingerprint
 import logging
 
@@ -182,6 +182,7 @@ class UserWebhooksStatusResponse(BaseModel):
     memory_created: bool
     realtime_transcript: bool
     day_summary: bool
+    button_event: bool = False
 
 
 class UserWebhookUrlResponse(BaseModel):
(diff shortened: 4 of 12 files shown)


Git-on-my-level:
Thanks @shubhxho — this is a really clean implementation of the contract proposed in #11719, and it follows the existing developer-webhook patterns closely. Notes from review, all non-blocking:

**What looks good**

- `backend/utils/webhooks.py` — `button_event_webhook` mirrors the `memory_created`/`day_summary` sender path exactly: opt-in check → URL lookup with `uid` query param → per-URL circuit breaker → `_post_dev_webhook` with `idempotency_key=event_id` → success/failure recording with auto-disable. The per-`(uid, device_id)` serialization lock held in a `WeakValueDictionary` is a nice touch — it preserves gesture order per device without the unbounded registry the older `_audio_bytes_send_locks` plain dict has.
- `backend/routers/users.py` — `ButtonEventRequest` validates the gesture as a `Literal`, `event_id` as UUID, `timestamp` as `AwareDatetime` (naive timestamps rejected), and caps `device_id`/`session_id` lengths. The status endpoint keeps the first-time-setup semantics for the new type, and the existing `{wtype}/enable`/`{wtype}/disable` routes pick up `button_event` with no extra code.
- `backend/route_policy_manifest.yaml` — the new entry (firebase auth, `first_party_app`) matches the surrounding conventions.
- `backend/tests/unit/test_button_event_webhook.py` — the real-asyncio interleaving test proving per-device delivery ordering is exactly the kind of test this needed; naive-timestamp rejection and noop-when-disabled coverage are good too.
- Regenerated clients (`users_wire.g.dart`, `OmiApi.generated.swift`, the four TS clients, `app-client-openapi.json`) are consistent, and making `button_event` optional-with-default keeps older clients parsing the status response.

**Non-blocking notes**

1. The app side isn't wired yet — nothing in this PR actually POSTs gestures from `capture_controller`'s existing button stream. That's consistent with the backend-first scope, just flagging that the feature isn't live end-to-end until a follow-up app change lands.
2. `device_id` is client-asserted (length-validated only) and is forwarded verbatim in the webhook payload. Fine for the developer-integration trust model, just worth being aware of.
3. `rate_limit: none` matches many other `first_party_app` POST routes, but since each accepted gesture triggers an outbound webhook delivery, maintainers may eventually want a per-uid cap there (the circuit breaker and auto-disable already bound the damage).

This adds a new public developer webhook type and endpoint to the released app-client contract, so it needs a maintainer's product sign-off on the contract before merge — the implementation itself looks solid.

---

kodjima33 (approved):
Feature (new webhook type), not a bug fix — approve-only per backend confidence gate. Also DIRTY, not merging per hard rule.

kodjima33 (approved):
Feature (closes #11719, a Feature Request, not a bug), classified backend/mixed area — approve only per policy, not merging. Well-scoped, tested, docs/schema regenerated.

Git-on-my-level:
Thanks @shubhxho — following up now that @kodjima33 has approved this head, which resolves the contract/product sign-off question from the earlier review. The implementation side stands as reviewed; what's new is mechanical:

**1. The branch now conflicts with main (this is what's failing `Hygiene`).**
`product-file-line-count-ratchet` reports it "cannot measure the synthetic merge of HEAD into origin/main" — the synthetic merge hits a content conflict in `web/personas-open-source/src/lib/omiApi.generated.ts`, and GitHub now lists the PR as conflicting. Main's generated clients and `backend/routers/users.py` (2201 lines on main now) have moved since this branch was cut. Please merge/rebase main and re-run the client generation so all five generated surfaces (`OmiApi.generated.swift`, the four `omiApi.generated.ts` copies) and `docs/api-reference/app-client-openapi.json` pick up both sides. After that, the `Line-Count-Exception` header for `backend/routers/users.py` (currently `2177 -> 2204`) will need re-measuring against the post-merge count, since the ratchet expects exact numbers.

**2. The Desktop Swift failures look unrelated to this diff.** `Desktop Swift Static & Test Contracts` failed in `MemoryAssistantTelemetryTests.swift:298` (`testThrowingAnalysisEmitsExactlyOneBoundedAnalysisFailedTerminal`, `0 != 1`, followed by an index-out-of-range crash) plus a `run.sh` build-lock timeout. None of those files are touched here — the only Swift change is the additive generated `postDeveloperButtonEventV1UsersDeveloperButtonEventPost` method — and `Desktop Swift Release Compile` is green. The same workflow passes on main's latest runs, so this reads as flake; worth a re-run after the rebase, and if it persists there it's a main-branch issue rather than this PR.

Nothing new to flag on the implementation itself: `button_event_webhook` still mirrors the existing sender path (opt-in check → per-URL circuit breaker → `idempotency_key=event_id` → auto-disable), and the `WeakValueDictionary` per-`(uid, device_id)` lock keeps gesture ordering without an unbounded registry.

---
*by AI on behalf of David.*

undivisible:
Soft-path rebase opened (see follow-up PR). Fork force-push blocked by OAuth missing `workflow` scope; conflicts resolved in users.py imports + OpenAPI (regen clients).

undivisible:
Superseded by soft-path squash-merge of #12922.

PR #12002, opened 2026-08-21
fix(app): stop OnDeviceWhisperProvider discarding the configured language

igor-popov-dev:
## What changed and why

`OnDeviceWhisperProvider.transcribe` declared an optional named parameter `String? language`
that shadowed the provider's own `language` field. The only caller,
`PurePollingSocket._flushBuffer`, never passes it (`ISttProvider` has no such parameter), so
`(language == 'multi' ? '' : language) ?? ''` always resolved to `''` — whisper.cpp always ran
in auto-detect mode, silently dropping the language the user picked in transcription settings.
The sibling `OnDeviceAppleProvider` already resolves `language ?? this.language` correctly;
this provider didn't.

none

## How it was verified

- `flutter test test/services/sockets/on_device_whisper_provider_test.dart` — 3/3 passed
  (provider-language fallback incl. `'ru'`, per-call override, `'multi'` → `''`).
- `bash app/test.sh` (full suite, Flutter 3.47.1) — 1393 passed, 0 failed.
- `bash app/scripts/analyze_ratchet.sh` — passed, no new lint occurrences (`avoid_print` and
  `sort_child_properties_last` baselines actually improved, unrelated to this change).

## Tests

Bug fix — regression test added: `test/services/sockets/on_device_whisper_provider_test.dart`
now covers the language fallback that was previously silently dropped (configured-language
fallback, per-call override, `'multi'` → `''` for whisper auto-detect).

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

Extracted request construction into `buildTranscribeRequest()` (marked `@visibleForTesting`)
so the regression test can assert the exact value handed to `TranscribeRequest` without
touching whisper's FFI layer. No unrelated cleanup bundled in.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12002?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/services/sockets/on_device_whisper_provider.dart
+++ b/app/lib/services/sockets/on_device_whisper_provider.dart
@@ -59,6 +59,19 @@ class OnDeviceWhisperProvider implements ISttProvider {
     }
   }
 
+  @visibleForTesting
+  TranscribeRequest buildTranscribeRequest(String audioPath, {String? language}) {
+    final effectiveLanguage = language ?? this.language;
+    return TranscribeRequest(
+      audio: audioPath,
+      // Empty string lets whisper.cpp auto-detect; used for the 'multi' setting.
+      language: effectiveLanguage == 'multi' ? '' : effectiveLanguage,
+      isTranslate: false,
+      isNoTimestamps: true,
+      splitOnWord: false,
+    );
+  }
+
   @override
   Future<SttTranscriptionResult?> transcribe(
     Uint8List audioData, {
@@ -76,13 +89,7 @@ class OnDeviceWhisperProvider implements ISttProvider {
       await tempFile.writeAsBytes(audioData);
 
       try {
-        final req = TranscribeRequest(
-          audio: tempFile.path,
-          language: (language == 'multi' ? '' : language) ?? '',
-          isTranslate: false,
-          isNoTimestamps: true,
-          splitOnWord: false,
-        );
+        final req = buildTranscribeRequest(tempFile.path, language: language);
 
         final res = await _whisper!.transcribe(transcribeRequest: req);
 
--- a/app/test/services/sockets/on_device_whisper_provider_test.dart
+++ b/app/test/services/sockets/on_device_whisper_provider_test.dart
@@ -1,0 +1,35 @@
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:omi/services/sockets/on_device_whisper_provider.dart';
+
+void main() {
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Verified the core claim against the code — real bug, and the fix is correct.

- `ISttProvider.transcribe` (`pure_polling.dart`) declares no `language` parameter and its only call path (`PurePollingSocket._flushBuffer`) never passes one, so the old `(language == 'multi' ? '' : language) ?? ''` always resolved to `''`. The provider's `language` field — populated from the user's STT config in `transcription_service.dart` — was never consulted, so on-device whisper always ran in auto-detect regardless of the setting.
- The new `buildTranscribeRequest` resolves `language ?? this.language` before the `'multi' → ''` mapping, which exactly mirrors `OnDeviceAppleProvider`'s handling for the same on-device config on iOS. Nice consistency.
- Default behavior is preserved: `SttProvider.onDeviceWhisper` uses `defaultLanguage: 'multi'`, which still maps to `''` (auto-detect), so only users who explicitly pick a concrete language see a change — the intended one.
- The extraction into a `@visibleForTesting` helper is behavior-neutral: `isTranslate` / `isNoTimestamps` / `splitOnWord` are unchanged, and the three new tests (configured-language fallback, per-call override, `'multi'` mapping) assert the exact value handed to `TranscribeRequest` without touching the FFI layer. Good regression coverage for a fix like this.

Minor non-blocking note: the per-call `language` parameter on `transcribe()` currently has no production caller (the interface doesn't expose it), so the override test documents the method's contract rather than a live path — fine as-is for parity with the Apple provider.

One caveat for the merge decision: no CI test workflow ran on this PR, so the full-suite and analyze results in the description are contributor-run. The added test is self-contained (no file or FFI access), but a maintainer-side `flutter test` pass would fully close the loop.

Thanks for the precise description and the clean regression test — good first contribution.

---

kodjima33 (approved):
Mobile bug fix: OnDeviceWhisperProvider was always resolving to auto-detect because the language parameter shadowed the field and was never passed by the only caller — confirmed by reading the code. Regression test added, scoped diff — confidence 5/5. Not already fixed on main.

PR #12006, opened 2026-08-21
fix(backend): emit Limitless creation webhook

### LINKED ISSUE #11365: Developer webhooks never fire for Limitless Pendant sessions (but do fire for phone recordings)

shubhxho:
## Summary

- send the conversation-created developer webhook when a Limitless sync upload revives an initially discarded fragment into a visible conversation
- emit only on the `discarded -> visible` transition, avoiding duplicate events during ordinary later reprocessing
- run delivery on the bounded post-processing executor so sync completion is not held open by a developer endpoint
- document that transcript and raw-audio webhooks require a live-listen session and are not replayed for store-and-forward uploads

Closes #11365

## Verification

- `backend/.venv/bin/pytest -q backend/tests/unit/test_sync_v2.py backend/tests/unit/test_webhook_auto_disable.py` (288 passed)
- `BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-backend-unit-failures.txt bash test.sh` (158 sync isolation/Cloud Tasks tests passed)
- `black --check --line-length 120 --skip-string-normalization backend/utils/sync/pipeline.py backend/tests/unit/test_sync_v2.py`
- `OMI_PR_BODY_FILE=/tmp/omi-pr-body-11365.md make preflight`
- `scripts/pr-preflight --pr-body-file /tmp/omi-pr-body-11365.md`

## Product invariants

None.

Failure-Class: none

Line-Count-Exception: backend/utils/sync/pipeline.py | 2499 -> 2524 | the sync merge owner must observe the Limitless discarded-to-visible transition at the point where reprocessing returns its persisted conversation

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12006?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_sync_v2.py
+++ b/backend/tests/unit/test_sync_v2.py
@@ -1607,6 +1607,28 @@ class TestAsyncCoordinatorBehavioral:
             'uid', 'current-conversation', {'audio_files': [{'path': 'current.opus'}]}
         )
 
+    def test_limitless_discard_recovery_emits_one_creation_webhook(self, fenced_worker_module):
+        """A pendant conversation becomes webhook-visible when merged speech revives it."""
+        _module, stubs = fenced_worker_module
+        pipeline = stubs['pipeline']
+        limitless_source = pipeline.ConversationSource.limitless
+        original = types.SimpleNamespace(source=limitless_source, discarded=True)
+        recovered = types.SimpleNamespace(source=limitless_source, discarded=False)
+
+        pipeline.conversations_db.get_conversation = MagicMock(return_value={'id': 'limitless-conversation'})
+        pipeline.deserialize_conversation = MagicMock(return_value=original)
+        pipeline.process_conversation = MagicMock(return_value=recovered)
+        pipeline._run_conversation_created_webhook = MagicMock()
+        pipeline.submit_with_context = MagicMock(side_effect=lambda _executor, fn, *args: fn(*args))
+
+        pipeline._reprocess_conversation_after_update('uid-1', 'limitless-conversation', 'en')
+
+        pipeline._run_conversation_created_webhook.assert_called_once_with('uid-1', recovered)
+
+        original.discarded = False
+        pipeline._reprocess_conversation_after_update('uid-1', 'limitless-conversation', 'en')
+        pipeline._run_conversation_created_webhook.assert_called_once()
+
     @pytest.mark.asyncio
     async def test_durable_completion_offloads_epoch_and_terminal_metric(self, fenced_worker_module):
         """Redis-backed fencing and telemetry never run on the async coordinator loop."""
--- a/backend/utils/sync/pipeline.py
+++ b/backend/utils/sync/pipeline.py
@@ -64,14 +64,22 @@ from database.sync_ledger import (
     release_sync_content_claim_after_job_retired,
     release_sync_content_claim,
 )
-from models.conversation import CreateConversation
+from models.conversation import Conversation, CreateConversation
 from models.conversation_enums import ConversationSource
 from models.transcript_segment import TranscriptSegment
 from utils.analytics import record_usage
 from utils.byok import get_byok_keys, set_byok_keys, set_byok_uid
 from utils.conversations.factory import deserialize_conversation
 from utils.conversations.process_conversation import process_conversation
-from utils.executors import db_executor, run_blocking, start_background_task, storage_executor, sync_executor
+from utils.executors import (
+    db_executor,
+    postprocess_executor,
+    run_blocking,
+    start_background_task,
+    storage_executor,
+    submit_with_context,
+    sync_executor,
+)
 from utils.fair_use import (
     FAIR_USE_ENABLED,
     FAIR_USE_RESTRICT_DAILY_DG_MS,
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
Thanks @shubhxho - well-scoped fix for #11365. Verified against the current flow, and the core gap is real: `_reprocess_conversation_after_update` runs `process_conversation(..., is_reprocess=True)`, and the generic creation-webhook emission is gated on `if not is_reprocess:` (utils/conversations/process_conversation.py:2139) - so a Limitless WAL session that begins as a discarded fragment and is revived by later merged segments never emits a creation event.

Per-file notes:

**backend/utils/sync/pipeline.py**
- The transition guard (`conversation.source == ConversationSource.limitless and was_discarded and not processed_conversation.discarded`) fires exactly once at revival and stays silent on later reprocessing - the right shape, and it matches how WAL batches revive fragments (reprocessing triggers when the merge target was discarded).
- Reusing `submit_with_context(postprocess_executor, ...)` mirrors the existing emission path and keeps BYOK contextvars flowing into payload building; the lazy import in `_run_conversation_created_webhook` matches the existing `_run_webhook` helper.

**backend/tests/unit/test_sync_v2.py**
- Good use of the fenced-worker harness. The second assertion (flipping `original.discarded` to `False` and re-running, still `assert_called_once()`) pins the no-duplicate property on ordinary reprocessing. Note the stubbing (`submit_with_context` inlined, `_run_conversation_created_webhook` mocked) means the actual delivery hop isn't exercised - fine for unit scope.

**docs/doc/developer/apps/Integrations.mdx**
- The two live-session notes check out against code: realtime transcript and audio-bytes delivery are driven only from the live socket session (routers/pusher.py), and the sync pipeline makes no transcript/audio webhook calls, so "not replayed for store-and-forward uploads" is accurate. This addresses the expectation mismatch from the issue report.

Two things for a maintainer before merge:

1. **Initial-pass emission.** On main, the initial creation webhook (process_conversation.py:2139) has no `discarded` guard, so a first discarded pendant fragment should already submit a creation event - while the new pipeline comment says the initial discarded pass "is not a useful creation event" (implying it doesn't emit). If fragments do emit in production, a revived session would now produce two events (fragment + revival) and the docs note "sends the event once" would be slightly off. The reporter's zero-delivery evidence suggests the initial event doesn't arrive in practice, but it'd be good to confirm which is true; if fragments do emit, suppressing the initial discarded emission would make the contract clean.
2. **Delivery lifecycle.** Both the existing and the new emission are fire-and-forget on `postprocess_executor` at the tail of the sync job; if the task worker exits before the pool drains, delivery is dropped. Not a regression from this PR, but it bears on whether this changes the reporter's symptom.

Follow-up, not blocking: phone-source WAL uploads (`omibatchphone` / `phonemic`) revive through the same reprocess path and would still get no revival event; limitless-only matches the issue, but generalizing may be worth a later pass.

Leaving for human maintainer review - the creation-event semantics for discarded fragments (point 1) need a product call before merge.

---

kodjima33 (approved):
Backend bug fix, closes #11365, scoped diff, root cause stated, regression test added — confidence 4/5. Not already fixed on main.

PR #12007, opened 2026-08-21
docs(backend): expose self-hosted Firebase auth project

### LINKED ISSUE #6636: Self-hosted backend cannot verify iOS app Firebase tokens — no auth path without LOCAL_DEVELOPMENT=true
## Problem

The Omi iOS app authenticates using Firebase tokens signed for Omi's Firebase project (`based-hardware-dev`). A self-hosted backend has no way to verify these tokens because it doesn't have Omi's private service account credentials (`SERVICE_ACCOUNT_JSON`).

The only workaround today is `LOCAL_DEVELOPMENT=true` in the backend `.env`, which disables token verification entirely and returns a hardcoded UID of `'123'`. This is a dev-only hack — it provides zero authentication and is not suitable for any real deployment.

## Root Cause

Firebase Admin SDK is currently initialized in two ways:
1. With `SERVICE_ACCOUNT_JSON` (Omi's private credentials — not available to self-hosters)
2. With Application Default Credentials (requires GCP environment)

Neither path works for a self-hoster.

## Why it can be fixed without any private credentials

Firebase token verification is **asymmetric**. Tokens are signed with Omi's private key but verified using Firebase's **public** signing keys, available at:

```
https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com
```

The Firebase Admin SDK supports a `projectId`-only initialization mode that fetches these public keys automatically — no service account required. Omi's Firebase project ID (`based-hardware-dev`) is already publicly committed in the repository at `app/setup/prebuilt/GoogleService-Info.plist`.

## Proposed fix

Add a `FIREBASE_PROJECT_ID` initialization path in `backend/main.py`:

```python
elif os.environ.get('FIREBASE_PROJECT_ID'):
    firebase_admin.initialize_app(options={'projectId': os.environ['FIREBASE_PROJECT_ID']})
```

Self-hosters set `FIREBASE_PROJECT_ID=based-hardware-dev` in their `.env`. The backend then correctly verifies tokens from the stock Omi iOS/Android app — **no app rebuild, no private credentials, no security bypass**.

**PR #6632** implements this fix. 3 files, 16 lines. Reviewed by Greptile at 5/5 confidence, marked safe to merge.

## Combined self-hosting path (with both fixes)

```env
# backend/.env
FIREBASE_PROJECT_ID=based-hardware-dev
```

```
App Store Omi app → custom backend URL (issue #6635 / PR #6604)
                  → Firebase tokens verified correctly (this issue / PR #6632)
                  → All audio/data on self-hosted server ✅
```

No app rebuild. No Omi service account. No security bypass.

## Privacy motivation

The pendant records continuous audio including conversations with PII and PHI. For healthcare workers, lawyers, journalists, and anyone in a regulated industry, sending this data to a third-party cloud is not acceptable. Self-hosting is the only viable path — and it is completely blocked today by this authentication gap.

## Related

- PR: https://github.com/BasedHardware/omi/pull/6632
- iOS backend URL companion issue: #6635

shubhxho:
## Summary

- expose `FIREBASE_AUTH_PROJECT_ID` in the backend environment template
- document credential-free Firebase ID-token verification for self-hosted backends
- keep the authentication audience separate from the operator's data project
- warn operators not to use the unauthenticated local-development bypass on public deployments

Fixes #6636

## Verification

- `backend/.venv/bin/python -m pytest backend/tests/unit/test_env_loader.py -q` (15 passed)
- `make preflight`

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.env.template
+++ b/backend/.env.template
@@ -84,6 +84,12 @@ RAPID_API_KEY=
 # Firebase OAuth
 FIREBASE_API_KEY=
 FIREBASE_AUTH_DOMAIN=
+# Project whose Firebase ID tokens this backend accepts. This is intentionally
+# separate from FIREBASE_PROJECT_ID so a self-hosted data project can verify
+# tokens minted for the Firebase project used by the client app. Stock Omi
+# mobile builds use based-hardware-dev; custom builds use their own project ID.
+FIREBASE_AUTH_PROJECT_ID=
+# Google Cloud/Firebase data project used by this backend's own services.
 FIREBASE_PROJECT_ID=
 
 # Encrypts conversations, memories, and chat messages. Required; set a unique
--- a/docs/doc/developer/backend/Backend_Setup.mdx
+++ b/docs/doc/developer/backend/Backend_Setup.mdx
@@ -330,6 +330,9 @@ OAuth is required for user authentication. You need to configure both Google and
 
     Firebase ID tokens are verified with the issuer's public signing keys, so
     `FIREBASE_AUTH_PROJECT_ID` does not require Omi service-account credentials.
+    When it is unset, Firebase Admin uses its default project selection instead;
+    set the auth project explicitly when it differs from your data-project setup,
+    or stock mobile-build tokens will be rejected.
     Do not enable `LOCAL_DEVELOPMENT` on an internet-accessible deployment: that
     mode intentionally bypasses authentication for local harnesses.
   </Step>
@@ -500,6 +503,12 @@ Complete reference for all `.env` variables:
     | `APPLE_KEY_ID` | Apple private key ID |
     | `APPLE_PRIVATE_KEY` | Apple .p8 file contents (with BEGIN/END lines) |
     | `BASE_API_URL` | Your backend URL (e.g., Ngrok URL) |
+  </Tab>
+  <Tab title="Firebase projects" icon="fire">
+    | Variable | Description |
+    |----------|-------------|
+    | `FIREBASE_AUTH_PROJECT_ID` | Firebase project whose ID-token audience the backend accepts; set this explicitly when it differs from the data project. |
+    | `FIREBASE_PROJECT_ID` | Google Cloud/Firebase project used for the backend's Firestore and other data services. |
   </Tab>
   <Tab title="Storage" icon="database">
     | Variable | Description |
--- a/docs/doc/developer/backend/Backend_Setup.mdx
+++ b/docs/doc/developer/backend/Backend_Setup.mdx
@@ -316,6 +316,22 @@ OAuth is required for user authentication. You need to configure both Google and
     ```
 
     Edit `.env` and fill in your API keys (see [Environment Variables](#environment-variables) below).
+
+    For a self-hosted backend used by an existing Omi mobile build, set the
+    Firebase token audience separately from your data project:
+
+    ```dotenv
+    # Stock development/community Omi mobile builds mint tokens for this project.
+    FIREBASE_AUTH_PROJECT_ID=based-hardware-dev
+
+    # Your own Google Cloud/Firebase project for Firestore and other backend data.
+    FIREBASE_PROJECT_ID=your-self-hosted-project
+    ```
+
+    Firebase ID tokens are verified with the issuer's public signing keys, so
+    `FIREBASE_AUTH_PROJECT_ID` does not require Omi service-account credentials.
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Thanks @shubhxho - this closes the documentation gap from #6636 nicely. I verified the new guidance against the current backend code:

- `backend/.env.template` - the added `FIREBASE_AUTH_PROJECT_ID` entry matches how the backend actually splits concerns: `backend/utils/env_loader.py` (`firebase_admin_options`) reads only `FIREBASE_AUTH_PROJECT_ID` for the auth audience, while `backend/database/_client.py` selects the Firestore/data project from `FIREBASE_PROJECT_ID`. The note that stock mobile builds use `based-hardware-dev` checks out in both `app/setup/prebuilt/GoogleService-Info.plist` and `app/setup/prebuilt/google-services.json`.
- `docs/doc/developer/backend/Backend_Setup.mdx` - the credential-free verification claim is accurate: with no service-account credential configured, `backend/main.py` initializes `firebase_admin` in project-ID-only mode and `verify_id_token` validates against Google's public signing keys. The `LOCAL_DEVELOPMENT` warning is well-aimed: `backend/utils/other/endpoints.py` returns a hardcoded uid `'123'` when `LOCAL_DEVELOPMENT=true` and no service-account credential is set - and `FIREBASE_AUTH_PROJECT_ID` is not one of the credentials that neutralizes that bypass, so self-hosters following this guide are exactly the audience that warning protects.

Two optional polish ideas (non-blocking):

1. Worth a sentence noting that when `FIREBASE_AUTH_PROJECT_ID` is unset, `main.py` falls back to `FIREBASE_PROJECT_ID` for token verification - i.e. setting only the data-project variable would point auth at the wrong audience and stock app tokens would fail. The guide's set-both-explicitly example already steers users right; spelling out the fallback helps anyone skimming.
2. The Environment Variables reference table further down this page lists no FIREBASE_* variables at all (pre-existing gap) - adding `FIREBASE_AUTH_PROJECT_ID` there would be a natural follow-up.

Verified docs-only change; leaving the merge call with the maintainers since it documents an auth-adjacent self-hosting posture.

---

Git-on-my-level:
Follow-up after 352f27e — thanks @shubhxho, both follow-ups from my earlier review land cleanly:

- The fallback sentence in `docs/doc/developer/backend/Backend_Setup.mdx` isn't just added, it's more precise than what I'd sketched: with no `FIREBASE_AUTH_EMULATOR_HOST` and no `SERVICE_ACCOUNT_JSON`, `backend/main.py` initializes `firebase_admin` with the result of `firebase_admin_options()` (`backend/utils/env_loader.py`), which returns `None` when `FIREBASE_AUTH_PROJECT_ID` is unset — so verification falls back to Firebase Admin's default project resolution (ADC / `GOOGLE_CLOUD_PROJECT`), not to `FIREBASE_PROJECT_ID`. "Stock mobile-build tokens will be rejected" is exactly the failure mode an operator with a self-hosted data project as the ADC default would hit.
- The new "Firebase projects" tab fills the reference-table gap, and `icon="fire"` matches existing usage elsewhere in the docs.
- `backend/.env.template` is unchanged in this push; its comment block still matches the actual env split — `backend/database/_client.py` reads `FIREBASE_PROJECT_ID` for the Firestore/data project, while `env_loader.py` reads `FIREBASE_AUTH_PROJECT_ID` for the token audience.

Re-verified the additions against current `main` — all claims hold. Still leaving the merge call with the maintainers: the guide documents an auth-adjacent self-hosting posture (which Firebase project's tokens a deployment accepts), which is a support-surface decision worth a human sign-off.

---

kodjima33 (approved):
Docs-correctness gate: FIREBASE_AUTH_PROJECT_ID is added to .env.template and documented as controlling Firebase ID-token audience verification, but grep of backend confirms it is not read anywhere in code (main.py just does firebase_admin.initialize_app() with no separate auth-project audience check). The described behavior isn't implemented — approve-only, not merging; author should either wire it up or correct the doc claim.

PR #12008, opened 2026-08-21
fix(backend): bound critical executor backlog

### LINKED ISSUE #6753: Discussion: critical_executor has unbounded internal queue (no backpressure)
## What

`backend/utils/executors.py:19-20`

```python
critical_executor = ThreadPoolExecutor(max_workers=8, thread_name_prefix=\"critical\")
storage_executor  = ThreadPoolExecutor(max_workers=4, thread_name_prefix=\"storage\")
```

`concurrent.futures.ThreadPoolExecutor` does not expose a `max_queue_size`; the internal `SimpleQueue` is unbounded. When all 8 critical workers are blocked on a slow downstream (Firestore, webhook, third-party API), every additional `loop.run_in_executor(critical_executor, ...)` simply enqueues — with no signal back to the caller that the system is saturated.

## Why this is worth thinking about

The async-native refactor (#6377) made this executor the single chokepoint for everything that's labeled \"critical\" — conversation processing, action items, memory extraction, integration triggers, persona updates, webhooks. A burst from any one of those subsystems (e.g., a slow webhook endpoint) silently raises latency for all of them with no metric to alert on.

## Possible directions (not proposing a specific one yet)

1. Wrap the executor with a `BoundedSemaphore` at submission sites so callers either acquire fast or get a typed error to handle.
2. Subclass `ThreadPoolExecutor` to override `_work_queue` with a bounded `queue.Queue(maxsize=N)`; submission then raises `queue.Full` and callers decide how to degrade.
3. Add a periodic gauge for `len(critical_executor._work_queue)` and alert when it exceeds N.
4. Split the pool by intent (e.g., webhooks separate from memory extraction) so a slow downstream can't starve fast ones.

Mostly looking for the team's view on how it's intended to behave under burst — the current behavior is well-defined (unbounded enqueue) but I couldn't tell from the code or PR #6377's body whether that's the design intent.

shubhxho:
## Summary

- bound the latency-sensitive critical executor to 64 queued tasks beyond its 8 workers
- reject excess submissions with a typed `ExecutorSaturatedError` instead of growing an unbounded process queue
- expose queue capacity alongside depth in executor health metrics
- add behavioral coverage for saturation and capacity recovery

Fixes #6753

## Verification

- `backend/.venv/bin/python -m pytest backend/tests/unit/test_executor_backpressure.py backend/tests/unit/test_async_http_infrastructure.py -q` (52 passed)
- `backend/.venv/bin/pyright backend/utils/executors.py` (0 errors)
- `make preflight`

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12008?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/dependencies.py
+++ b/backend/dependencies.py
@@ -8,7 +8,7 @@ from firebase_admin import auth
 import database.mcp_api_key as mcp_api_key_db
 import database.dev_api_key as dev_api_key_db
 from utils.api_key_families import DEV_FAMILY, MCP_FAMILY, wrong_key_family_detail
-from utils.executors import critical_executor, db_executor, run_blocking
+from utils.executors import ExecutorSaturatedError, critical_executor, db_executor, run_blocking
 from utils.log_sanitizer import sanitize
 from utils.observability.api_keys import record_api_key_repairs
 from utils.memory.product_authorization import ProductAuthorizationContext
@@ -61,6 +61,12 @@ async def get_current_user_id(
     try:
         id_token = credentials.credentials
         decoded_token = await run_blocking(critical_executor, auth.verify_id_token, id_token)
+    except ExecutorSaturatedError as error:
+        raise HTTPException(
+            status_code=503,
+            detail='Authentication temporarily unavailable. Try again shortly.',
+            headers={'Retry-After': '1'},
+        ) from error
     except Exception as e:
         logger.error(f"Error verifying Firebase ID token: {e}")
         raise HTTPException(status_code=401, detail="Invalid authentication credentials")
--- a/backend/routers/omni_relay.py
+++ b/backend/routers/omni_relay.py
@@ -14,9 +14,9 @@ from utils.byok import (
     set_byok_keys,
     validate_byok_websocket,
 )
-from utils.executors import critical_executor, db_executor, run_blocking
+from utils.executors import db_executor, run_blocking
 from utils.llm.gateway_client import raise_if_gateway_feature_mode_blocks_direct_model_surface
-from utils.other.endpoints import _verify_ws_auth  # type: ignore[reportPrivateUsage]  # shared WS auth helper, intentionally reused cross-module
+from utils.other.endpoints import run_critical_ws, verify_ws_auth
 import database.users as users_db
 from models.users import PlanType
 from utils.subscription import get_chat_quota_snapshot, is_trial_paywalled
@@ -85,7 +85,7 @@ async def omni_relay(websocket: WebSocket):
         f"provider={websocket.query_params.get('provider')}"
     )
     try:
-        uid = await run_blocking(critical_executor, _verify_ws_auth, cast(str, authz))
+        uid = await run_critical_ws(verify_ws_auth, cast(str, authz))
     except WebSocketException as e:
         logger.warning(f"omni relay auth rejected: code={e.code} reason={e.reason}")
         await websocket.close(code=e.code, reason=e.reason or "unauthorized")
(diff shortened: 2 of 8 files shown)


Git-on-my-level (changes requested):
Thanks @shubhxho — this targets a real gap (#6753) and the core mechanism is solid. Verified on head c715191.

**What works well**

- `backend/utils/executors.py` — the `BoundedSemaphore(max_workers + max_queue_size)` design is correct and leak-free: non-blocking acquire, release on submit failure (`except BaseException`), and `add_done_callback` release; using `BoundedSemaphore` also turns any future over-release into a loud error instead of silent drift. Capacity accounting is right: with `max_workers=1, max_queue_size=2`, three submits succeed and the fourth raises, exactly as your test asserts.
- `backend/tests/unit/test_executor_backpressure.py` — good behavioral coverage of both saturation rejection and capacity recovery, with clean shutdown in `finally`. Both new tests pass for me; the three `TestSemaphoreGetters` failures I saw in `test_async_http_infrastructure.py` are environment-only (my sandbox lacks pytest-asyncio) and reproduce identically on base `978ffdb`, so they are not from this diff.
- Exposing `queue_capacity` in `get_executor_metrics()` makes the bound observable, which the issue explicitly asked for.

**Why I'm requesting changes**

The mechanism is right, but nothing in the codebase catches `ExecutorSaturatedError` yet, and where it surfaces today is misleading:

1. `backend/dependencies.py:61-66` (`get_current_user_id`): the `except Exception` catches the new error and returns **401 "Invalid authentication credentials"**. Under saturation, every bearer-authed endpoint would tell clients their token is invalid — mobile clients typically react to that with forced re-auth, at exactly the moment the backend is struggling. Saturation should surface as 503 (ideally with `Retry-After`), not 401.

2. `backend/utils/other/endpoints.py:599-601` (`with_rate_limit`) and `:621-624` (`with_rate_limit_context`): no handling at all, so saturation raises out of the dependency as an unhandled 500 on the 42 rate-limit-wrapped dependencies across 15 routers (`chat`, `conversations`, `mcp`, `tools`, `goals`, …). The WebSocket auth paths (`endpoints.py:348`, `routers/omni_relay.py:88`) similarly let it escape past the `except WebSocketException` handlers during the handshake.

3. The rejection itself is invisible: saturation raises without a counter or log, and `get_executor_metrics()` only reports `queue_capacity`/`queue_depth` — a pool that is actively rejecting requests can still look "at capacity but healthy". A rejection counter (or at least a warning log at raise time) would make the new behavior visible in ops.

Suggested minimal path: map `ExecutorSaturatedError` to `HTTPException(503, ...)` ahead of the generic handler in `get_current_user_id`, and in the two rate-limit wrappers (a tiny shared helper covers all of them), plus a rejection counter or warning log. With that in place I think this is in good shape — the semaphore design itself needs no changes.

One judgment call I'm explicitly leaving to the maintainers: this implements option 1 from the #6753 discussion (typed rejection at submission). Whether rejecting auth-path work — vs. splitting the pool by intent or acquiring with a short timeout — is the right ops posture for `critical_executor` deserves an explicit maintainer nod, since it changes what clients experience under load.

---

PR #12013, opened 2026-08-21
feat(desktop): route tasks to the agent the user names

Addy2055:
## What this does

Omi can delegate work to Claude Code, Hermes, OpenClaw and Codex, but choosing between them was `adapterIdForHarnessMode()` — a static settings lookup that always resolved to `acp`. Saying **"use codex to fix this"** ran Claude Code and never mentioned it. On macOS, Codex was not a routable agent at all: no adapter file, absent from the capability matrix, despite being one of the three agents Track 1 names.

This makes the spoken request real, on both desktops.

| Behaviour | Before | After |
|---|---|---|
| Name an agent | ignored; ran the default | runs on the named agent |
| Several connected | static order | ranked on the capability matrix, with the rest as fallback |
| Name one you don't have | ran a different agent silently | install command + docs link |
| The chosen agent dies | task failed | retried on the next agent that can do the job |
| Rule an agent out | not understood | dropped from selection *and* from the fallback chain |

## How it works

**Selection is deterministic, not model-judged.** `DesktopIntentRouter` documents that it "intentionally has no language heuristics", so string matching stays at the surface and the capability matrix stays the source of truth. OpenClaw declares `supportsTools: false`, so a file-editing task rules it out as a **fact** rather than a guess — and the user is told why instead of getting a silent substitution.

macOS runtime, `desktop/macos/agent/src/runtime/`:

| Module | Answers |
|---|---|
| `agent-mention.ts` | which agent was named, and which was ruled out |
| `adapter-scoring.ts` | which connected agents can do this task, best first |
| `agent-install.ts` | what to tell someone who named an agent they haven't installed |
| `agent-routing.ts` | composes the three into one decision |
| `agent-fallback.ts` | retries a failed run on the next agent |

**Fallback reacts to the terminal run event, not to the spawn.** `spawnBackgroundAgent` creates the durable run and returns a receipt, then runs the adapter asynchronously (`void execution` in `kernel-runs.ts`) — so "the agent could not start" lands *after* the caller already has its answer, and a spawn-time retry loop can never observe it. The untried chain rides on run metadata and the supervisor subscribes to `run.failed`; each hop shortens the chain, gets its own request id, and records what it replaced in `agentFallbackFrom`. A retry that itself fails is logged and dropped, because subscribers run inside the kernel's emit path.

I found this the slow way: I built the spawn-time version first and its tests failed. The test that pins the constraint is in `agent-routing-continuation.test.ts` so nobody re-adds it.

**Codex is now a real macOS adapter**, with capability declarations copied verbatim from the Windows adapter so the same agent isn't described two ways. Its credentials go through a **per-adapter env allowlist keyed by adapter id** — external adapters are spawned with `shell: true` under a minimal allowlist, so adding `OPENAI_API_KEY` to the shared list would hand the user's OpenAI key to Hermes and OpenClaw. Side effect: OpenClaw no longer receives `HERMES_HOME`.

## Cross-platform

Each desktop had half of this. macOS had capability ranking but no cross-agent fallback. Windows already had a real fallback loop and emits `agent_selected {fallback: true}`, but nothing ever derived `agentId` from what the user actually said. Windows gets `agentMention.ts` + `agentsForTask`; macOS gets the fallback.

That leaves one string-matching rule implemented twice — the exact divergence class `contracts/parity/` exists to prevent. So it's now under contract: **`contracts/parity/agent_routing.json`**, 25 utterance→agent vectors, each platform running them through its own production code.

It earned that on its first run. The Windows port compiled its alias table from `PRODUCTION_ADAPTER_IDS`, which deliberately omits `pi-mono` — so Windows was deaf to "pi mono" while macOS heard it. Fixed by iterating the alias table itself; recognising a spoken name is separate from being allowed to run the task.

## Verification

| Check | Result |
|---|---|
| macOS routing tests (10 files) | **123 passing** |
| macOS `tsc --noEmit`, `npm run build` | clean |
| Windows `npx vitest run src/main/codingAgent` | **270 passed, 1 skipped, exit 0** |
| Windows `typecheck:node`, eslint, prettier | clean |
| Parity fixture | 25 cases green on **both** platforms |

```bash
cd desktop/macos/agent && npx tsc --noEmit && npm run build
npx vitest run tests/agent-mention.test.ts tests/agent-install.test.ts \
  tests/adapter-scoring.test.ts tests/agent-routing.test.ts \
  tests/agent-routing-kernel.test.ts tests/agent-routing-continuation.test.ts \
  tests/agent-fallback.test.ts tests/parity-agent-routing.test.ts \
  tests/adapter-selection.test.ts tests/runtime-adapter.test.ts
# 123 passed

cd desktop/windows && npx vitest run src/main/codingAgent
# 270 passed | 1 skipped
```

There is also `desktop/macos/agent/scripts/demo-agent-routing.mjs`, which walks the six behaviours through the compiled `dist/` — real output, not a test harness.

### What I could not verify

**I have no Mac, so push-to-talk was never run against a live daemon.** The routing logic is covered by unit tests, kernel-level tests against a real `AgentRuntimeKernel` + SQLite store, and the `dist/` demo — but nobody has pressed the button. Saying so rather than implying otherwise, per the Definition of Done.

The macOS suite was run on Windows, where 34 tests across 12 files fail for platform reasons — Unix-domain sockets and subprocess relays (`tool-relay` alone is 15). Two more are order-flaky and pass in isolation. That suite has not been run on macOS. The Windows component suite, which does run here, is clean at 270/270.

`make preflight` was not run: it needs a full checkout and mine was sparse.

## Definition of Done

1. Behaviour changed → tests changed ✅ 123 macOS + 270 Windows
2. Component suite passes ✅ Windows; macOS caveat above
3. Exercised the change myself ⚠️ see "What I could not verify"
4. Verification evidence written down ✅ above
5. No orphaned TODO/FIXME ✅
6. Docs moved with the code ✅ `desktop/macos/AGENTS.md` (new Agent Routing section, incl. the `OMI_CODEX_ADAPTER_COMMAND` env var), `contracts/parity/README.md`, changelog fragments on both platforms
7. Failure-class declaration — n/a, this is a feature not a `fix:`
8. `make preflight` ⚠️ not run, sparse checkout

---

Submitted for **Omi Summer, Track 1 (Agents)**.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12013?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks @Addy2055 — this is a genuinely well-built change. I read all 32 files and ran the touched suites against this head in a clean container: all 119 tests in the macOS routing suites (mention, routing, scoring, fallback, install, kernel, continuation, parity, runtime-adapter) and 41 in the Windows suites (parityAgentRouting, taskRunner) pass, and `tsc --noEmit` is clean. The only failures in the broader macOS suite (`runtime-stdio-contract`, `tool-surfaces-exhaustiveness`) reproduce identically on `main`, so they're pre-existing rather than from this PR.

What stands out:

- `contracts/parity/agent_routing.json` plus both conformance suites is exactly the right shape for a rule implemented twice — the macOS `agent-mention.ts` / Windows `agentMention.ts` pair can no longer drift silently, and the vectors pin the genuinely tricky parts (word boundaries, clause-scoped negation, `instead`, `but`).
- `adapters/acp.ts`: moving `HERMES_HOME` out of the shared allowlist into the adapter-keyed `ADAPTER_SPECIFIC_ENV_ALLOWLIST`, and scoping `OPENAI_API_KEY`/`CODEX_HOME` to the codex adapter only, is a real hardening — one agent's credentials can no longer reach another `shell: true` subprocess.
- `agent-fallback.ts` reacting to the terminal `run.failed` event (bounded `handledRunIds` set, `MAX_FALLBACKS_PER_REQUEST = 4`) matches the async `void execution` reality in `kernel-runs.ts` — a spawn-time retry loop could never observe the failure it means to recover from.
- `tests/agent-install.test.ts` cross-checks the runtime guidance against the Windows `AgentsTab.tsx` table, so install commands can't drift between surfaces either.

Two observations for the record, neither blocking:

1. The fallback chain rides on run metadata (`agentFallbackChain` in `spawn_background_agent`'s `metadata`). The kernel side re-validates types and the 4-hop ceiling bounds abuse, but since metadata is caller-supplied, a hostile caller could hand-pick the retry adapter. Fine while surfaces are trusted; worth a note if that ever changes.
2. The fallback retry re-spawns with `trustedUserSpawn: true` without re-running `assertAdapterAllowedForTopLevelLocalProviderSpawn`. Chain members were validated at original spawn time and chains are only built by `resolveAgentRoute` today, so the invariant holds — it's just implicit.

Docs check out against the code (module map, env var names, suite paths, changelog entries). One thing for maintainers to consciously accept: the `desktop/macos/AGENTS.md` section adds five "fail the PR if any break" routing rules, which raises the review bar for future agent contributions in this area.

Leaving the merge decision to a human maintainer: adding Codex as a production adapter on macOS and changing how spoken tasks route on both desktops is a product/architecture call (the tests indicate it tracks the Track 1 asks), and this is a first contribution from this account.

---

Addy2055:
Thanks for the thorough read — and for running the suites in a clean container, which settled the pre-existing-vs-introduced question on `runtime-stdio-contract` / `tool-surfaces-exhaustiveness` better than I could from Windows.

I've addressed both observations in 49551fd, since they turn out to be the same fix.

`agentFallbackPlanForRun` now re-validates the chain when it reads it back from run metadata: an entry is only usable if it is a production adapter **and** currently registered in this runtime. That closes (1) — a hand-picked `agentFallbackChain` can no longer steer the retry at an arbitrary or unregistered provider — and it makes (2) explicit rather than implicit, because the registration check is exactly what the `trustedUserSpawn: true` re-spawn was relying on the original spawn to have already done.

Two tests cover it: a chain naming a non-adapter, and one naming a real-but-unregistered adapter (`codex` when it isn't connected), both yield no plan. `desktop/macos/AGENTS.md` rule 4 now states the invariant.

Verified: `tsc --noEmit` clean; the 10 macOS routing/parity files 125 passed (was 123); Windows `src/main/codingAgent` 287 passed, 1 skipped.

On the last point — agreed that adding Codex as a production adapter and changing spoken-task routing on both desktops is a product call rather than a code one, and it should sit with a human maintainer. Worth flagging explicitly for that decision: **I have no Mac, so push-to-talk was never exercised against a live daemon.** The routing is covered by unit and kernel-level tests and a demo driven through the compiled `dist/`, but nobody has pressed the button, and that's the one claim I can't make.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Git-on-my-level:
Follow-up verification on 49551fd — both observations from the prior automated review are resolved, and I re-verified the core claims independently.

**Chain re-validation (resolves both prior observations).** `agentFallbackPlanForRun` in `kernel-coordinator.ts` now filters the run-metadata chain through `isProductionAdapterId(id) && this.registry.has(id)` before building a plan, so a caller-supplied `agentFallbackChain` can no longer steer a retry at an arbitrary or unregistered provider — which also makes the `trustedUserSpawn: true` re-spawn's reliance on registration explicit rather than implicit. `agent-fallback.test.ts` covers both hostile shapes (a non-adapter id, and `codex` while unregistered), and AGENTS.md rule 4 now states the invariant. Verified against the diff; no gaps found.

**Independent parity execution.** I ran all 24 vectors in `contracts/parity/agent_routing.json` against both platform extractors side by side (`agent-mention.ts` and `agentMention.ts`) in a clean environment: 24/24 pass identically on both, including extra probes of my own (case variants, "run co dex", "codex-like tooling", multi-negation). The routing composition also behaves as documented against the real capability matrix: an unnamed task still lands on `acp` (today's default, unchanged), a named agent leads its chain, `openclaw` stays out of tool-task chains (`supportsTools: false` is a real matrix fact), ruled-out agents are dropped from selection *and* fallback, and a named-but-unconnected agent yields `install_required` instead of silent substitution.

**Supply-chain check on the new commands.** The suggested Codex path resolves to legitimate first-party packages: `@agentclientprotocol/codex-acp` (published by the ACP org maintainers, repo `agentclientprotocol/codex-acp`) and `@openai/codex` (OpenAI-published, repo `openai/codex`). No typosquats. The install/launch/docs strings in `agent-install.ts` match the Windows `AgentsTab.tsx` table exactly, so the drift test has real teeth.

**Env hardening confirmed.** Moving `HERMES_HOME` out of the shared allowlist and scoping `CODEX_HOME`/`OPENAI_API_KEY`/`CODEX_API_KEY` to the codex entry of `ADAPTER_SPECIFIC_ENV_ALLOWLIST` (`acp.ts`) tightens exposure: external commands run under `shell: true`, and one agent's credentials now cannot reach another.

Left for a human maintainer, deliberately:
- The product call: Codex becomes a production adapter on macOS and spoken-task routing changes on both desktops. The tests indicate it tracks the Track 1 asks, but that direction is yours to confirm.
- The contributor disclosed that push-to-talk was never exercised against a live macOS daemon (no Mac available). Unit, kernel, and parity coverage is strong; one on-device smoke test before release would close the only untested claim.

Everything I could verify, I verified; nothing blocks beyond the product decision.


**Agent-guidance note for maintainers.** This PR also changes the instructions that govern AI coding/review agents working in this repo: `desktop/macos/AGENTS.md` gains a five-rule "Agent Routing" section marked "fail the PR if any break" (selection reads the capability matrix and never a model; explicit requests are never substituted; ruled-out agents are dropped from chains; fallback re-validates caller-supplied chains on read; alias/negation behavior is a cross-platform contract), and `contracts/parity/README.md` registers `agent_routing.json` with conformance suites on both platforms. I checked each rule against the code in this PR — env var names, suite paths, and the chain re-validation in `agentFallbackPlanForRun` all match — so the guidance documents real behavior rather than aspiration. Maintainers should consciously accept it: future agent-authored changes to desktop routing will be held to these rules and to fixture-first behavior changes on both platforms.

---

Addy2055:
Thanks for the independent re-verification — especially running the 24 parity vectors through both extractors side by side and adding your own probes ("run co dex", "codex-like tooling"). That's a stronger check than my own suites, since it exercises the two implementations against each other rather than each against its own expectations.

No code changes from me on this round; nothing here asks for any.

One factual note for the maintainer decision on the AGENTS.md section, since it's come up twice. The "Rules (fail the PR if any break):" framing isn't new with this PR — it already appears in `desktop/macos/AGENTS.md` on `main`, in the Chat Continuity Write-Path Contract (INV-6) section. I mirrored the existing convention for a second contract in the same file rather than inventing a stricter one. Happy to soften it to non-normative guidance if maintainers would rather the routing rules read as description than as a gate; that's a one-line change and entirely your call.

Agreed on the rest: the product direction is yours, and the on-device smoke test is the right gate before release — it's precisely the claim I can't make from Windows.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Git-on-my-level:
Verified 11eead3 — this head is a pure `Merge upstream/main` with no changes to the PR's own surface: the compare against the previously verified 49551fd touches 300 upstream files and **zero** of this PR's 32, which remain identical, and the diff vs current main lists exactly the same desktop/ + contracts/ files. Rather than assume, I re-checked the load-bearing pieces at this head:

- **Parity still holds.** Re-implemented the extractor from `agent-mention.ts` source (alias table, word-boundary matching, clause-scoped negation with 4-word lookback, longer-alias-first) and ran all 24 vectors in `contracts/parity/agent_routing.json`: 24/24, with adversarial probes ("hermeneutics", "codexes", "claudette", comma/dash/sentence negation scoping, "instead of", later-request-overrides-exclusion) behaving exactly as the fixture pins. The Windows `agentMention.ts` port implements the identical algorithm and tables.
- **Env scoping intact** (`adapters/acp.ts`): `HERMES_HOME` stays out of the shared allowlist; `CODEX_HOME`/`OPENAI_API_KEY`/`CODEX_API_KEY` reach only the codex entry of `ADAPTER_SPECIFIC_ENV_ALLOWLIST`; external spawns still build env from allowlists with proxy-credential sanitizing.
- **Fallback re-validation intact** (`agentFallbackPlanForRun` in `kernel-coordinator.ts`): caller-supplied `agentFallbackChain` still filtered through `isProductionAdapterId && registry.has`, retry ceiling 4, `run.failed`-only trigger, dedup guard.

One new fact for this head: **no CI check runs have reported on 11eead3 yet** — the suites were last executed in-container on 49551fd. Since the merge touched nothing in this PR's surface that's expected rather than alarming, but the head shouldn't count as CI-verified until its checks report.

Standing items, unchanged: the product call on Codex becoming a production macOS adapter and on the spoken-routing/negation semantics, and the disclosed absence of a live push-to-talk smoke test on macOS.

Agent-guidance reminder (unchanged this head): `desktop/macos/AGENTS.md` adds the five-rule Agent Routing section marked "fail the PR if any break", and `contracts/parity/README.md` registers `agent_routing.json` with per-platform conformance suites — future agent-authored changes to routing will be held to those rules and to fixture-first behavior changes on both platforms; worth a conscious maintainer accept.

---
*by AI on behalf of David — the remaining gate is a product-direction decision (Codex as a production macOS adapter, spoken routing semantics) rather than anything urgent; @Git-on-my-level when you're ready to make that call.*

Addy2055:
Thanks for re-verifying at this head — and for re-deriving the extractor from source rather than trusting the suites. Agreed on all three: `11eead3` is ours (an `upstream/main` merge to clear a base-branch conflict, resolved only in `contracts/parity/README.md` by keeping both fixture rows), it touches none of the PR's 32 files, and parity / env-scoping / chain re-validation are unchanged.

On your one new fact — **no CI has reported on `11eead3`** — I dug into why, since that's the only thing standing between this head and being properly verified rather than verified-by-argument.

The runs exist. Eight workflows were queued for `11eead3` at 2026-09-08T12:35 and every one is sitting at `conclusion: action_required`:

```
Repo Checks · Backend Checks · Backend Hermetic E2E · Web Checks
Mobile App Checks · Desktop Checks · Desktop Windows CI · Desktop Swift CI
```

The same eight sat at `action_required` on `49551fd`. It's the standard fork-PR gate rather than anything about this branch — `author_association` on this PR is `NONE`, and the control is clean: in the same window, runs from `BasedHardware/omi` branches (`codex/*`, `watchdog/*`) go queued → success/failure, while runs from forks (`andreistoc/omi`, this one) stop at `action_required`. So nothing has failed, and nothing will report until a maintainer clicks **Approve and run workflows**.

Worth flagging because of what's in that list: **`Desktop Swift CI` and `Desktop Windows CI` are two of the eight.** The one gap I disclosed at the top — no live macOS verification, because I don't have a Mac — is exactly what those would close on real runners. One approval turns the disclosed unknown into an actual result, and if either goes red I'll fix it.

Nothing is outstanding on my side. The offer from last round stands: if maintainers would rather the `desktop/macos/AGENTS.md` routing rules read as description than as a gate, that's a one-line change and I'll make it on request. Otherwise the remaining gate is the product call, which is yours.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aryanorastar:
Keeping this open on `11eead3539f7`.

CONFLICTING 0-CI desktop agent-routing; remaining-hole keep; not later copy.

Addy2055:
Thanks for looking, and understood on not approving `11eead353`.

One of the two things you named is now cleared. `11eead353` was conflicting because upstream had moved 924 commits since the previous merge; I've merged `main` in and pushed **`d53dd544a7`**, which reports `mergeable: true`. The PR's own surface is unchanged at **32 files, +2920 −14** — the merge only brings upstream's work in.

There was a single conflict, in `contracts/parity/README.md`: upstream reformatted the fixture table and added the conversation-summary prose while this branch adds the `agent_routing.json` row. I kept all three — upstream's formatting, upstream's prose, and our row in their padded style. Nothing dropped from either side.

Verified on the merged tree:

- macOS: `npx tsc --noEmit` clean; the 10 routing/parity files → 124 passed
- Windows: `pnpm install --frozen-lockfile` then `npx vitest run src/main/codingAgent` → 287 passed, 1 skipped; `npm run typecheck:node` clean

On the **0-CI** half — that one isn't something I can move. Eight workflows queue on every push to this branch and all eight sit at `conclusion: action_required`:

```
Repo Checks · Backend Checks · Backend Hermetic E2E · Web Checks
Mobile App Checks · Desktop Checks · Desktop Windows CI · Desktop Swift CI
```

It's the standard fork-PR gate (`author_association: NONE`) rather than anything about this branch — in the same window, runs from `BasedHardware/omi` branches go queued → success/failure while runs from forks stop at `action_required`. So 0-CI will stay 0-CI until a maintainer approves the run, and I'd rather say that plainly than have it read as an unexplained gap.

Worth noting **`Desktop Swift CI` is one of the eight**, because it bears on the hole I disclosed in the PR body: I have no Mac, so push-to-talk was never exercised against a live daemon. That workflow is exactly what would close it. If you approve the run and anything goes red, I'll fix it.

Two smaller things:

- Unrelated to this PR, but it cost me an hour: `desktop/windows` no longer installs under npm — `@react-three/fiber@9.7.0` wants `react >=19 <19.3` against `react@19.3.0`. pnpm with the committed lockfile resolves it. That reproduces on `main`.
- I couldn't parse "remaining-hole keep; not later copy" — if either half was a change request rather than a note, say which and I'll do it. The standing offer also holds: if the five `desktop/macos/AGENTS.md` routing rules should read as description rather than a gate, that's a one-line change on request.

Happy for this to sit until the product call on Codex-as-a-macOS-adapter is made; just wanted the conflict off the board so that decision isn't competing with a red merge state.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aryanorastar:
Verified on `d53dd544a7`.

The conflict half of the earlier note is cleared: GitHub reports MERGEABLE on this head. The 0-CI half is not — all eight workflows are still `action_required` (first-time fork gate), including Desktop Swift CI and Desktop Windows CI. That is not a failure on your side.

"Remaining-hole keep; not later copy" was not a change request. It meant this is unique work (not a duplicate of another open PR), so the vehicle stays open.

I am still not approving or merging this SHA. Spoken routing plus Codex as a production macOS adapter is a product call, push-to-talk was never run against a live daemon, and there is no on-device smoke here either. Approving GitHub Actions so the desktop jobs actually run is a separate step from reviewing the PR.

Addy2055:
Thanks — that clears up all three, and "not a duplicate, so the vehicle stays open" was the part I had genuinely misread. Understood on the product call and on the Actions approval being a separate step from review; neither is mine to move and I'm not asking you to move them.

The one thing in your note I *could* act on was the last clause, so I did: **there is now an on-device smoke.** `55f1c74be2`.

```bash
cd desktop/macos/agent && npm run smoke:agent-routing
```

The reason there wasn't one is that everything I had asserts against adapters I invented. The tests use fakes, and `demo-agent-routing.mjs` fabricates a registry so it can show every branch. Neither says a word about the machine it's running on — which is exactly the hole you named.

`scripts/smoke-agent-routing.mjs` asks the machine instead. The inventory comes from `ensureRegisteredAdapter`, the same gate `index.ts` uses at boot, so **connected means connected on that host** — that adapter's activation env var is really set. It then drives the three Track 1 behaviours through the real kernel and the real `spawn_background_agent` control tool and exits non-zero on failure.

The design constraint I held myself to: **it must not be able to overstate what it proved.** A check needing an agent the host lacks is skipped and prints the env var that would enable it — never faked, never quietly counted as a pass. Nothing spawns a subprocess, because the `install_required` branch returns before one starts. `--with=<ids>` flips activation vars for a single process to inspect the full matrix, and labels those agents as simulated in the output.

On a bare host (only `acp` connected) it's 4 passed / 0 failed / 3 skipped — the three skips are the checks that need a second agent. With `--with=codex,hermes,openclaw` all four criteria run and pass:

```
  PASS  a named, connected, tool-capable agent wins
        "use hermes to run the test suite" -> hermes
        chain: hermes -> acp -> codex
  PASS  no agent named: selected by capability, with a fallback chain
        "fix the failing test" -> acp        chain: acp -> hermes -> codex
  PASS  an agent that cannot do the job is excluded, not silently substituted
        openclaw ruled out, ran acp, and said why
  PASS  a ruled-out agent leaves the fallback chain too
        "fix this, but don't use hermes" -> acp    chain: acp -> codex
```

**Writing it caught a real mistake of mine.** I asserted that install guidance always carries an install command, and Hermes failed it — because Hermes has no one-line install, so its guidance is a note plus a docs URL, which is correct behaviour. The check now asserts what the contract actually promises: an answer that's actionable, and never a silent swap to whichever agent happens to be connected. That's the bug class this was supposed to find, and it found one on the first run against a real inventory rather than an imagined one.

**What it does not do, plainly:** it does not press push-to-talk. It starts at the utterance a transcript produces and covers everything after it; the audio layer above is untouched. That gap is unchanged and I'm not dressing this up as having closed it — the script and `AGENTS.md` both say so in as many words. What changed is that the part I *can* stand behind is now verifiable on your hardware in one command, instead of resting on my description of tests you'd have to take on trust.

Verified on `55f1c74be2`: `tsc --noEmit` clean, 124 passed across the 10 routing/parity files, Windows `src/main/codingAgent` unchanged by this commit. PR surface is now 34 files, +3239 −14; `MERGEABLE`.

Standing offer unchanged on the `AGENTS.md` "fail the PR if any break" framing — say the word and I'll soften those five rules to non-normative wording in a single commit.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aryanorastar:
Verified on `55f1c74be2`.

Still MERGEABLE. The eight workflows on this SHA are still `action_required` (first-time fork gate), including Desktop Swift CI and Desktop Windows CI. That is not a failure on your side.

`scripts/smoke-agent-routing.mjs` is a real kernel/routing check against `ensureRegisteredAdapter`. It is not push-to-talk and not a live daemon press — you already say that in the script and in AGENTS.md. That gap is unchanged, and there is no kit on this side either, so this is not on-device proof.

I am still not approving or merging this SHA. Spoken routing plus Codex as a production macOS adapter is still the product call, not a code review I can close from here.

Approving GitHub Actions so the desktop jobs actually run remains a separate step from reviewing the PR. Keeping this open.

Addy2055:
CI ran on `55f1c74be2` — thank you for approving it. Two reds, and they are not the same kind of thing, so I fixed one and deliberately left the other alone.

Also merged latest `main` (49 commits) while I was in here; no conflicts this time. Head is now `86df3089b2`, still MERGEABLE.

## Red 1 — `Repo Checks` / `agents-md-lean`. Mine. Fixed.

```
desktop/macos/AGENTS.md: 600 lines exceeds budget of 560
desktop/macos/AGENTS.md: 50396 bytes exceeds budget of 47000
```

This was **not** new from the smoke commit. It has been failing since `8c2c823f73` on 2026-08-21 — the PR's first commit, at 47,844 bytes / 568 lines. It only surfaced today because a workflow had never actually executed on this branch. First run, month-old latent failure.

Fixed the way the check prescribes, **without touching the check**. Detail moved to `.github/agent-docs/desktop-agent-routing.md`; `desktop/macos/AGENTS.md` is back to byte-identical with `main` at 46,999 bytes, so this PR now consumes **none** of that file's budget.

One judgement call worth flagging, because I could have hidden it: the pointer row went into the **root** `AGENTS.md` router table, not the macOS guide. `main` leaves exactly **1 byte** of headroom there (46,999 of 47,000), so the fix the check itself prescribes — move detail down, link it "from its index row" — cannot be completed in that file, because the row does not fit. `doc-maintenance.md` says not to raise a budget, and recalibrating yours is not something this PR should do on its own. Root is the documented home for a cross-component index row, routing is genuinely cross-component (macOS, Windows, `contracts/parity`), and root has room since you ratcheted it to 60 / 4,000 — it is now 38 lines / 2,488 bytes.

If you would rather it sit in the macOS guide, that needs a byte-budget recalibration and it is yours to make. Say so and I will move it in one commit.

Verified locally by measuring the blobs as git stores them, not the working tree — on Windows the working copy is CRLF and the check reads bytes, which inflates every file by one byte per line. `86df3089b2` is back at `action_required` (8 workflows), so **CI has not confirmed this fix**; I am not claiming it has.

## Red 2 — `Desktop Checks` / `Agent Runtime`. Not mine to fix. Untouched.

One test, 976 of 977 passing:

```
FAIL  tests/control-tools.test.ts > does not normalize a Codex alias from signed direct control
AssertionError: expected { ok: true, … } to match { ok: false, error: { … } }
-  "message": "Unknown production adapter: codex"
```

That guard came from **your own #13785**, merged 2026-09-14. It asserts that `codex` is not a production adapter on macOS. This PR makes it one. They are opposite answers to the same question, and the test is now the exact place where they collide.

I could make this green in about four lines, and I am not going to. Rewriting your guard would be me settling the product call you have declined to close three times now, by editing the test that encodes your side of it. That is not a merge-blocker I should clear for you.

What I think it is worth, though: **the decision now has a location.** It is not "spoken routing plus Codex as a production adapter, in the abstract" any more — it is one assertion in one test. Whichever way you go, the other side is a small, contained diff. If the answer is no, that guard is correct as written and the honest outcome is that this PR does not land as designed, which is a fine outcome and I would rather hear it than keep the branch warm.

## The rest of that run

- **`Desktop Windows CI` — passed.** First CI confirmation of the Windows mirror and the parity fixture; that half has now been verified by something other than me.
- `Backend Checks`, `Backend Hermetic E2E`, `Mobile App Checks`, `Web Checks` — passed.
- **`Desktop Swift CI` — still `queued`, never ran.** That is the one that would have gone at the macOS side, so the push-to-talk gap is unchanged and untested for the same reason as before.

No change to the two things that were always yours: the product call, and approving Actions. Standing offer on the `AGENTS.md` rules framing is moot now — the rules moved to `agent-docs` and are phrased as what the suites enforce, not as PR outcomes.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aryanorastar:
Keeping this open. Not approving `86df3089b2b8`. `contracts/parity/agent_routing.json` absent on origin/main. +3274, checks FAILURE.

PR #12035, opened 2026-08-22
fix(desktop): use the typed DefaultsKey for onboardingStep in tests

aryanorastar:
## Summary

`Desktop Swift Static & Test Contracts` is red on `main`, and every open desktop PR inherits the failure through its merge ref. The cause is two raw UserDefaults key strings in a test file:

```
OnboardingPersistenceClearingTests.swift:60:22: error: omi_inline_userdefaults_key Violation: Use DefaultsKey.swift for typed UserDefaults keys
OnboardingPersistenceClearingTests.swift:66:34: error: omi_inline_userdefaults_key Violation: Use DefaultsKey.swift for typed UserDefaults keys
```

The custom rule's regex is `(?<!removeValue\()forKey:\s*"`, and this file is not in `.swiftlint-baseline.json`, so both are new violations rather than baselined debt.

`DefaultsKey.onboardingStep` already exists with the raw value `"onboardingStep"`, so this is behavior-identical by construction — the same string reaches `UserDefaults` either way. The lines immediately around these two already used `DefaultsKey.hasCompletedOnboarding.rawValue`; these two were simply missed.

## Verification

- After the swap, the rule's regex no longer matches anywhere in the file (`grep -P '(?<!removeValue\()forKey:\s*"'` returns nothing).
- The file is absent from `.swiftlint-baseline.json` (grep count 0), confirming these were flagged as new rather than pre-existing.
- Diff is two lines, no behavior surface: `DefaultsKey.onboardingStep.rawValue == "onboardingStep"`.

Verified with the pinned linter itself, through the repo's own lane:

```
$ scripts/pr-preflight --pr-body-file /tmp/pr-defaultskey.md
PR preflight passed: 19 checks in 48.38s.
```

`desktop-swiftlint` is one of those 19. To confirm the check was actually exercising this file rather than skipping it, I restored the two raw strings and re-ran:

```
<== FAIL desktop-swiftlint (0.50s)
Manifest checks failed: desktop-swiftlint
```

Then restored the fix and it passes again.

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12035?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Tests/OnboardingPersistenceClearingTests.swift
+++ b/desktop/macos/Desktop/Tests/OnboardingPersistenceClearingTests.swift
@@ -57,13 +57,13 @@ final class OnboardingPersistenceClearingTests: XCTestCase {
 
     defaults.set(true, forKey: DefaultsKey.hasCompletedOnboarding.rawValue)
     OnboardingFlow.markCompleted(for: "owner-a", defaults: defaults)
-    defaults.set(18, forKey: "onboardingStep")
+    defaults.set(18, forKey: DefaultsKey.onboardingStep.rawValue)
 
     XCTAssertEqual(
       OnboardingFlow.reconcileCompletionOwner(currentOwnerID: "owner-b", defaults: defaults),
       .resetForDifferentOwner)
     XCTAssertFalse(defaults.bool(forKey: DefaultsKey.hasCompletedOnboarding.rawValue))
-    XCTAssertNil(defaults.object(forKey: "onboardingStep"))
+    XCTAssertNil(defaults.object(forKey: DefaultsKey.onboardingStep.rawValue))
     XCTAssertNil(defaults.object(forKey: OnboardingFlow.completionOwnerKey))
   }
 


aryanorastar:
@Git-on-my-level this one is a two-line CI repair rather than a feature, so flagging it for a merge call ahead of my other PRs.

**`Desktop Swift Static & Test Contracts` is red on `main`, and it blocks every desktop PR.** The last completed run on `main` (`08-21T20:25`, the `fix(desktop): keep onboarding account-scoped` merge) failed on:

```
OnboardingPersistenceClearingTests.swift:60:22: error: omi_inline_userdefaults_key Violation: Use DefaultsKey.swift for typed UserDefaults keys
OnboardingPersistenceClearingTests.swift:66:34: error: omi_inline_userdefaults_key Violation: Use DefaultsKey.swift for typed UserDefaults keys
```

Both are still on `main` at `72a21e2a65`. The file is absent from `.swiftlint-baseline.json`, so these are new violations rather than baselined debt — the rule's regex is `(?<!removeValue\()forKey:\s*"` and two raw `"onboardingStep"` strings slipped in while the lines immediately around them already used `DefaultsKey.hasCompletedOnboarding.rawValue`.

**This is deterministic, not cosmetic.** The `desktop-swiftlint` check runs `swiftlint-wrapper.sh lint` with no path arguments — it lints the whole tree against the baseline, so the manifest triggers decide only *whether* it runs, not what it covers. Any PR whose diff matches those triggers (`desktop/macos/Desktop/Sources/**/*.swift`, `Tests/**/*.swift`, the `.swiftlint.yml`/baseline) therefore inherits `main`'s violations through its merge ref, regardless of its own content. Two confirmed so far: #12029, a Meet-detection fix that never touches `OnboardingPersistenceClearingTests.swift`, and #11452, whose only Swift file is a regenerated `OmiApi.generated.swift` client. Both failed on exactly these two lines.

To be precise about the blast radius: a desktop PR that changes no `.swift` file is unaffected — #11990 (manifest plus two Python files) went green on the same current `main`. So the rule is "touches a desktop Swift file", not "is a desktop PR". This PR is the exception among the former because it is the one that removes the violations.

**The fix is behavior-identical by construction.** `DefaultsKey.onboardingStep` already exists with the raw value `"onboardingStep"`, so the same string reaches `UserDefaults` either way. Two lines, no production source touched.

**Verified against the pinned linter, not by inspection.** `scripts/pr-preflight` passes 19 checks on this head, `desktop-swiftlint` among them. To prove the check was actually exercising this file rather than skipping it, I restored the raw strings and re-ran:

```
<== FAIL desktop-swiftlint (0.50s)
Manifest checks failed: desktop-swiftlint
```

then restored the fix and it passes again. CI agrees: all four Swift jobs are green on this PR — Build & Tests, Release Compile, Static & Test Contracts, and Detect Changes.

**One thing this is *not*:** it is not blocking releases. I checked `desktop_auto_release` before writing this and it has succeeded hourly straight through the breakage (01:41, 00:22, 23:14, 22:14, 21:14 …).

Since it is someone else's merge that introduced it, I have no stake here beyond wanting the lane green; happy for anyone to take it over or land an equivalent fix. I am not bumping the rest of my queue with this.

Separately, and not a merge blocker: the same CI window keeps throwing unrelated Swift test failures, a different suite each run — `RewindCaptureExclusionGenerationTests`, `SuggestedTasksStoreTests`, `KernelTurnRecordedProjectionTests`, and `ChatTranscriptGestureHarnessTests`. Filing that separately with run links.

Git-on-my-level:
Thanks @aryanorastar — closing this one as superseded, with full credit for the diagnosis.

The root-cause analysis was exactly right: the two raw `"onboardingStep"` literals in `desktop/macos/Desktop/Tests/OnboardingPersistenceClearingTests.swift` — the `defaults.set(18, forKey:)` write and the `XCTAssertNil(defaults.object(forKey:))` assertion in `testCompletedOnboardingResetsWhenAuthenticatedOwnerChanges` — were the only un-baselined `omi_inline_userdefaults_key` violations holding `Desktop Swift Static & Test Contracts` red on `main`. And the swap is behavior-identical, since `DefaultsKey.onboardingStep` already exists with rawValue `"onboardingStep"`.

An equivalent fix landed on `main` directly in 822d3738a7 ("fix(desktop): use the typed onboardingStep key in its test"), which also kept the deliberate wire-name literals in the key-list assertion. As of that commit, the file on `main` is byte-identical to this branch's version — same blob SHA `c6781918` — so there is nothing left for this PR to merge, and the lane is green again.

The verification discipline here was excellent: proving the check actually covered this file by restoring the raw strings and re-running the pinned preflight, the blast-radius breakdown for #12029 and #11452, and confirming the release lane was unaffected. That write-up is what made the direct fix safe to land quickly. Appreciate the explicit "happy for anyone to land an equivalent fix" note — this close is exactly that, not a rejection.

---

PR #12042, opened 2026-08-22
fix(desktop): Activity opens what you click, and the conversation summary is whole again

Archit-lal:
## What this fixes

Six defects on the desktop Activity surface and the conversation summary, plus the
Flutter client's copy of one of them and the desktop test runner that was hiding all
of it.

**1. Activity always lost to Memories.** `ChatFirstMemoryRoutePolicy` preserved the
selected hub view across the shared `.memories` route by naming one of them
(`current == .brainMap ? .brainMap : .memories`). `.activity` was added as a fourth
hub view later and never added, so the route sync that runs on the next `onChange`
rewrote the just-written `.activity` back to `.memories`. The click was correct and
undone a frame later — from the top-bar pill and the hub's own switcher alike.

**2. The pill said Memories and opened whichever page was persisted last.** It is
titled `Activity`, opens Activity, and routes through the same writer the chip row
uses.

**3. Activity's rows opened nothing.** Memory rows had no action at all — no `Button`,
no tap gesture, no hit region — so the one row naming what Omi learned was the one row
you could not open; clicks landed on the conversation card above it. Conversation rows
carried the full record, narrowed it to an id, wrote that id into the automation
*snapshot* singleton, then called `selectPrimary`, whose contract is to discard an
unconsumed deep link. `showsCaptureArchive` also latched true forever after the first
capture link, so later conversation navigation kept landing in the reduced archive pane.

**4. The conversation summary rendered a fraction of what the backend sends.** #11834
moved the substance of a summary out of `structured.overview` into `structured.sections`
and left `overview` a short compatibility paragraph. The generated wire DTO decoded
`sections` from day one; the hand-written domain model never read the field. A grep for
it across the desktop sources outside `Generated/` returned nothing. Fixed on desktop
and in the Flutter client, which had the identical gap in
`app/lib/backend/schema/structured.dart`.

**5. Activity is three panels, and its two chip rows are one.** It carried the hub
switcher directly above a filter row that shared three of its words and did something
different. There is one row now and every chip navigates; `Brain Map` joins it.
`PageGlassLane` was also wrapping the whole hub in a fourth full-height panel with
Activity's two nested inside — nesting `.behindWindow` inside `.behindWindow` doubles
the scrim, which is why Activity read muddier than every page around it.

**6. `desktop/macos/test.sh` was red at baseline and hid the Swift stage.**
`context-bucket-benchmark.py` guarded an error-body read with an enumerated exception
tuple; on Python 3.9 — what macOS ships as `/usr/bin/python3` — `HTTPError.read()` with
no body raises `KeyError('file')` out of `tempfile`, escaping the guard and masking the
real HTTP error. Python 3.12+ returns `b''`, so the defect is invisible there. `set -e`
over the launcher loop then meant that one failure cost the entire Python and Swift
stages: the operator saw one error and got zero signal from 663 Swift suites.

## Two traps this surfaced

- `ShellDestination.Reach.memoryHubView` named the hub switcher as the mechanism
  reaching Conversations, Memories and Brain Map, but `unreachable()` only ever checked
  that a pill existed. Deleting the switcher would have stranded three pages with the
  whole suite green. `Reach.activityChipRow` is now checked against
  `ActivityDestinationChip.reachableHubDestinations` — the same value the row renders
  from.
- `DesktopTopBar` declared a `memoryDestinationRawValue` binding and never wrote it, so
  in the non-chat-first shell the `Activity` pill landed on Memories (the stored default)
  with neither a switcher nor a chip row. A dead end reachable from a cold launch.

## Deliberate cost

Making the whole chip row navigate removes the spine's in-place filtering. A row where
`Conversations` leaves the page while `Tasks` narrows it teaches a rule and then breaks
it. Home's search results keep their filter chips — that panel is shared, and a
navigating chip there would discard the query just typed, so `ChipBehavior` makes each
surface state which it is.

## Verification

Clicked through a real signed build (`omi-navfix`, prod backend, never the production
bundle): Activity pill opens Activity; switcher-era regression gone; conversation rows
report `detail_open=true` on the exact conversation; standalone and indented memory rows
open that memory's inspector; the summary renders headed markdown sections where it
previously showed one paragraph; Activity renders as three separate panels; all six chips
open their page; the pill returns from each.

- `desktop/macos/scripts/swift-test-suites.sh`: 663 suites, 0 failures
- `app/test.sh`: 1409 passed, 0 failures, 5 pre-existing skips
- `context-bucket-benchmark` test passes on 3.9 and 3.14, fails on both without the fix
- Full `test.sh` with a venv: 44 launcher scripts, 202 pytest, 663 Swift suites, exit 0;
  with a failure injected, exit 1 with an aggregate and all 663 Swift suites still run

## Known-red, not caused by this PR

`tests/test-desktop-core-harness-readiness-teardown.sh` ("readiness should recover from
one transient startup failure") fails on a **pristine `upstream/main` checkout at
d6b9e35775 with a clean working tree** — verified in a separate worktree, not by
stashing. It shells out to `make dev-up` and is state-dependent locally.

Note the interaction: before this PR, `test.sh` aborted at `test-context-bucket-benchmark.sh`
(alphabetically earlier), so this failure was never reached and nobody saw it. Fixing the
benchmark and making the loop aggregate is what surfaced it. That is the aggregation
working as intended, not a regression introduced here. The local pre-push run of this
one script was hatched with `PRE_PUSH_SKIP_DESKTOP_LAUNCHER_TESTS=1`; CI remains the
authority on it.

INV-NAV-1
INV-UI-1

Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift | 1634 -> 1636 | threading the mounted hub page into PageGlassLane so Activity is not wrapped in a second glass panel; two lines at one call site, no new responsibility in this file

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12042?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
Desktop bug fix (owner-override): Activity opens the clicked item and conversation summary sections render again — well-tested (new XCTest + Dart test coverage), CI green, no abuse signals.

PR #12055, opened 2026-08-22
chore(app): bump app dependencies

mdmohsin7:
## What

Dependency bumps in `app/pubspec.yaml` (+ `app/pubspec.lock` and `app/ios/Podfile.lock`), and the one call-site
migration the surviving bumps require.

| Package | From | To |
|---|---|---|
| `sign_in_with_apple` | ^6.1.1 | ^8.1.0 |
| `intercom_flutter` | 9.4.22 | 9.6.22 |
| `shared_preferences` | 2.5.3 | 2.5.5 |
| `flutter_foreground_task` | 9.1.0 | 10.0.0 |
| `flutter_timezone` | 3.0.1 | 5.1.0 |
| `flutter_sound` | 9.25.1 | 9.30.0 |
| `pasteboard` | ^0.4.0 | ^0.5.0 |

iOS pods move with them: Intercom 19.4.1 → 19.7.2, flutter_sound_core 9.25.1 → 9.30.0.

**`flutter_timezone` 5.x** is the only source change: `getLocalTimezone()` returns `TimezoneInfo` instead of
`String`, so both `getTimeZone()` implementations return `.identifier` — the same IANA identifier as before.

Audited and needing no changes: `sign_in_with_apple` 8.1.0 (`getAppleIDCredential` unchanged), `flutter_sound` 9.30.0
(`openRecorder(isBGService:)`, `openPlayer`, `startPlayer` unchanged), `intercom_flutter` 9.6.22. `pasteboard` has no
call sites in `app/lib`.

## Two bumps were attempted and withdrawn

Both were withdrawn on evidence from building the prod flavor on a physical iPhone XR (iOS 18.7.9). Neither failure
is visible to CI, which has no iOS build job — Dart Analyze and Android Compile Smoke both pass on the versions that
cannot build or run on iOS.

**`google_sign_in` 7.2.0 → stays on 6.2.2.** The v7 migration was written and it analyzed and compiled cleanly:
private constructor + `GoogleSignIn.instance.initialize()`/`authenticate()`, and an idToken-only Firebase credential
since v7 splits authentication from authorization. On device, tapping Sign in with Google failed with
"Authentication failed. Please try again." The root cause was **not** established — the session it was tested in was
degraded by orphaned `devicectl` processes from my own tooling, so the migration is not proven guilty either. It is
withdrawn rather than shipped unproven; re-landing it needs a clean device session and the actual
`GoogleSignInException` code. `google_sign_in_ios` returns to 5.7.5 and the native `GoogleSignIn` pod to 7.0.0.

**`flutter_foreground_task` 11.0.1 → pinned to 10.0.0.** v11.0.0 moved the `@objc FlutterForegroundTaskPlugin` class
behind `#if SWIFT_PACKAGE`, so under CocoaPods the class the generated registrant calls is never compiled:

```
Semantic Issue (Xcode): Unknown receiver 'FlutterForegroundTaskPlugin';
did you mean 'SwiftFlutterForegroundTaskPlugin'?
ios/Runner/GeneratedPluginRegistrant.m:296
```

Removing the bridging-header import and adding the module import — what the plugin's own migration doc prescribes —
is necessary but not sufficient; 11.0.1 additionally requires project-wide Swift Package Manager adoption, which this
project does not use and which would have to hold for Codemagic too. 10.0.0 keeps the Obj-C class, needs only
Flutter >=3.38, and still carries v10's iOS UIScene lifecycle support.

## Verification

- `flutter analyze` (Flutter 3.44.5, matching CI): **0 error-severity issues**, 164 pre-existing info lints.
- `flutter pub get` reproduces the committed `pubspec.lock` exactly — no drift.
- **Prod flavor builds, installs, and launches on an iPhone XR (iOS 18.7.9)** with this tree.
- `app/ios/Podfile.lock` regenerated via `pod install`; the stale lock was itself a build blocker (it pinned
  Intercom 19.4.1 against the 19.7.2 the bump requires).
- CI on the previous push (which still carried the two withdrawn bumps) was green across Dart Analyze & Tests,
  Android Compile Smoke, Formatting, Generated Files, and Hygiene — which is precisely why the two iOS failures
  above needed hardware to find.
- `make preflight`: all selected checks pass except `plan-catalog-contract`, which fails on a stale local worktree
  checkout of an unrelated branch, not a file in this diff.

- Google Sign-In v7 needs a clean-session device retest before it can land; the migration is recoverable from
  commit `6083323`.
- There is no iOS build in CI. All three iOS breakages in this PR were invisible to the pipeline.

Failure-Class: none


kodjima33 (approved):
Dependency bump chore (google_sign_in v7, sign_in_with_apple v8, foreground_task, timezone, etc) with matching call-site migrations in auth_service.dart. Not a bug fix, so approve-only per confidence gate (mobile area). CI green.

PR #12058, opened 2026-08-22
test(context-for-claude): close the analytics privacy boundary the existing tests left open

Archit-lal:
## What happened

I asked for a change that would have broken this app's anonymity promise, and the existing tests would not have caught it. This PR is the guard, not the change.

**The proposal:** add the backend device hash to the analytics payload as a super-property, so PostHog events could be joined to the Firestore install roster and per-day MCP usage could be attributed per install.

**Why it is wrong.** `ClientDevice.clientDeviceId` is `"macos_<deviceIdHash>"`, and that string is the Firestore document key at `users/{uid}/screen_activity/{clientDeviceId}-{rowId}`. So the hash resolves to a uid, and a uid resolves to a person. Shipping it as a super-property makes every property CFC sends — when the mic was on, capture and screen minutes, active hours, permission states — a named person's behavioural record for anyone holding both datasets.

`AnalyticsIdentity` already says so, and says it is deliberate:

> `ClientDevice` already mints a per-install UUID and hashes it into `X-Device-Id-Hash`. This one is derived from the same UUID under a **different salt**, and that separation is the point: the backend id keys the user's own captured rows and is joinable to their account, so reusing it here would make every anonymous analytics event trivially re-identifiable by anyone holding both datasets.

The premise I argued from — "PostHog already receives it" — is simply false. `AnalyticsPayload.superProperties` returns exactly `app`, `app_version`, `app_build`, `macos_version`.

## Why the existing tests allowed it

Two holes, both real:

1. **The identity tests were relative, not pinned.** `testTheAnalyticsIDIsStableAcrossCalls` compares the derivation to itself; `testTheAnalyticsIDIsNotTheBackendDeviceHash` compares it to the device hash. Changing the salt, digest, prefix or truncation keeps both green while silently re-anonymising every existing install and orphaning its history.
2. **The disclosure test only ever inspected `distinct_id`.** The join could walk in as a *property* with the whole suite green — which is exactly the shape of the change proposed here.

## The guards

- `testTheAnalyticsIDDerivationIsPinnedToItsShippedValue` — pins both derivations to literals (`cfc_436214da822c0b7e` / `667a8eea`) against a fixed install id, so a change to either derivation is a failing test rather than silent data loss.
- `testTheSuperPropertySetIsClosed` — asserts the **exact** super-property key set. A closed set, not a blocklist, so the next new identifier fails by default instead of needing to be predicted.

## Verification

- `swift test`: **1681 → 1683 tests, 9 skipped, 0 failures.**
- **Mutation-checked, independently re-run:** applying the proposed `device_hash` line makes the guard fail with
  `("["device_hash", "macos_version", "app", "app_build", "app_version"]") is not equal to ("["macos_version", "app_build", "app_version", "app"]") - a new super-property is a new disclosure on every event this app has ever sent`
  and it passes again on revert.

## Also: the timezone basis, settled and documented

Screen-activity rows carry naive strings like `2026-08-17 20:55:09.022`. They are **UTC**, by construction:

```swift
formatter.dateFormat = "yyyy-MM-dd HH:mm:ss.SSS"
formatter.locale = Locale(identifier: "en_US_POSIX")
formatter.timeZone = TimeZone(secondsFromGMT: 0)
```

The backend agrees — `normalize_screen_activity_timestamp` does `parsed.astimezone(timezone.utc).replace(tzinfo=None)`. The offset is absent because the field is a lexicographic sort key that the backend range-filters by string comparison, not a date type.

**The trap this documents:** `cfc_daily_active` buckets by the **local** calendar day (`ContextAnalytics.dayStamp` reads `Calendar.current`), while screen rows are UTC. For anyone not on UTC, an evening capture belongs to one day in Firestore and the other in PostHog. Both are correct for what they measure; neither is a re-derivation of the other, and analysis must not treat them as interchangeable.

No production behaviour changes in this PR.

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/context-for-claude/Tests/ContextAppTests/AnalyticsTests.swift
+++ b/desktop/context-for-claude/Tests/ContextAppTests/AnalyticsTests.swift
@@ -93,10 +93,59 @@ final class AnalyticsPayloadTests: XCTestCase {
     /// captured rows and is joinable to their account, so an analytics id equal to it would make every
     /// "anonymous" event trivially re-identifiable by anyone holding both datasets.
     func testTheAnalyticsIDIsNotTheBackendDeviceHash() {
-        let installID = "6E7D2C61-0000-4000-8000-000000000000"
+        let installID = Self.pinnedInstallID
         XCTAssertNotEqual(AnalyticsIdentity.derive(from: installID), ClientDevice.hash(of: installID))
         XCTAssertFalse(AnalyticsIdentity.derive(from: installID).contains(ClientDevice.hash(of: installID)))
     }
+
+    /// The derivation, pinned to a literal rather than recomputed.
+    ///
+    /// The two tests above are both *relative*: they compare the derivation against itself and
+    /// against the device hash, so a change to the salt, the digest, the prefix or the truncation
+    /// satisfies all of them while re-anonymising every install on the planet and restarting every
+    /// retention curve. The property the identity actually needs is that it produces the **same
+    /// string it produced last release**, and only a literal can say that.
+    ///
+    /// If this fails, the answer is almost never to update the literal. Read the doc comment on
+    /// `AnalyticsIdentity.salt` first.
+    func testTheAnalyticsIDDerivationIsPinnedToItsShippedValue() {
+        XCTAssertEqual(
+            AnalyticsIdentity.derive(from: Self.pinnedInstallID), "cfc_436214da822c0b7e",
+            "changing this severs every install's history — see AnalyticsIdentity.salt")
+        XCTAssertEqual(
+            ClientDevice.hash(of: Self.pinnedInstallID), "667a8eea",
+            "the backend's eight-hex contract, pinned beside the analytics id it must stay apart from")
+    }
+
+    /// **The super-property set is closed, and the closure is the privacy boundary.**
+    ///
+    /// Keeping `distinct_id` salted apart from the backend's device hash only buys anonymity while
+    /// nothing *else* on the payload carries the join. A `device_hash` super-property — proposed to
+    /// make the PostHog series joinable to the Firestore install roster — is not a weaker version of
+    /// making the two ids equal; it is the same join, spelled as a property. `macos_<hash>` is the
+    /// key of `users/{uid}/screen_activity/{clientDeviceId}-{rowId}`, so it resolves to a uid and a
+    /// uid resolves to a person, and every property on every event here would become a named
+    /// person's record of when their microphone was on.
+    ///
+    /// Asserted as an exact key set rather than as a blocklist of forbidden names, because the
+    /// failure mode is somebody adding a *new* identifier, and a blocklist can only ever refuse the
+    /// ones already imagined. Adding a genuinely non-identifying super-property means editing this
+    /// literal and saying why in the commit — which is the review this boundary is owed.
+    func testTheSuperPropertySetIsClosed() {
+        let keys = Set(
+            AnalyticsPayload.superProperties(
+                version: "1.0.12", build: "1000012", macOSVersion: "Version 15.5.0"
+            ).keys)
+
+        XCTAssertEqual(
+            keys, ["app", "app_version", "app_build", "macos_version"],
+            "a new super-property is a new disclosure on every event this app has ever sent")
+    }
+
+    /// The install id both derivations are pinned against. Fixed rather than minted, and never read
+    /// from `UserDefaults`: `ClientDevice.deviceIdHash` would mint and store a real install id in the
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified end to end — this closes two real gaps in the analytics privacy tests, and every claim in it checks out against the code it cites.

**AnalyticsTests.swift**

- `testTheAnalyticsIDDerivationIsPinnedToItsShippedValue` — I recomputed both literals independently: SHA-256 of `"context-for-claude/analytics/v1" + "6E7D2C61-0000-4000-8000-000000000000"` truncated to 16 hex is exactly `cfc_436214da822c0b7e`, and the unsalted digest truncated to 8 hex is exactly `667a8eea` (the `X-Device-Id-Hash` contract in `ClientDevice.hash`). The pinning rationale is sound: `testTheAnalyticsIDIsStableForAnInstall` and `testTheAnalyticsIDIsNotTheBackendDeviceHash` are both *relative* assertions, so a salt/prefix/truncation change would keep them green while re-anonymising every install and restarting every retention curve. Only a literal can catch that.
- `testTheSuperPropertySetIsClosed` — asserting the exact key set `{app, app_version, app_build, macos_version}` rather than a blocklist of forbidden names is the right shape for this boundary: `AnalyticsPayload.superProperties` returns exactly those four today, and the failure mode is somebody adding a *new* identifier no blocklist imagined. Hoisting `pinnedInstallID` also removes the duplicated literal between the two identity tests, and keeping it out of `UserDefaults` avoids minting a real install id in the test process — good call.

**docs/analytics.md**

- The super-property refusal section is accurate and well-argued: `macos_<deviceIdHash>` really is the `{clientDeviceId}-{rowId}` Firestore key component, so carrying it beside `distinct_id` would be the join to `users/{uid}/screen_activity/...` made explicit — and the point that "the backend already receives it" is about the wrong party (that's what makes it a join key; PostHog does not receive it today) is the crux worth writing down.
- The two-clocks section quotes `storageTimestamp` byte-for-byte (`en_US_POSIX`, `TimeZone(secondsFromGMT: 0)`, `yyyy-MM-dd HH:mm:ss.SSS`); `normalize_screen_activity_timestamp` in `backend/database/screen_activity.py` really does `astimezone(timezone.utc).replace(tzinfo=None)`, and `makeCanonicalTimestampFormatter` in the Omi desktop app pins the same zone. Given `ContextAnalytics.dayStamp` derives from `Calendar.current`, the warning to cohort on the install's declared day rather than a UTC re-derivation follows directly. Minor note, non-blocking: "`cfc_daily_active` carries `tool_calls_total`" is true of the current event schema, which the doc's "answerable from this project alone" argument depends on — fine as long as that property set is treated as load-bearing (which `testTheSuperPropertySetIsClosed` now helps enforce).
- One transparency note: the desktop Swift suites were path-scoped skips on this head, so the new tests did not execute in CI. The literals were verified computationally here and the APIs match existing usage in the same file, so compile/runtime risk is low — worth a local `test.sh` run before merge if that's cheap on your machine.

Since `analytics.md` is documentation this app's coding agents read, it also steers future agent work: it pre-empts the recurring "add the device hash as a super-property" suggestion and states the day-bucketing rules. I checked each new claim against the source and they all hold, so the added guidance is accurate and keeps agents on the right side of the privacy line.

No blocking concerns. Formal approval left to a human maintainer — thanks for writing the guard *and* the reasoning; the doc comments alone should stop the next person (or agent) from re-proposing the join.

---

aryanorastar (approved):
Verified on current origin/main (`c1de50bf`): `testTheAnalyticsIDDerivationIsPinnedToItsShippedValue` is still missing from AnalyticsTests.swift.

The two new tests close real holes the relative assertions leave open. Super-property keys are asserted as the exact set {app, app_version, app_build, macos_version}.

Hosted Desktop Swift did not run on this head (path is context-for-claude, not macos Desktop). Not blocking this review — the pin literals match the production derivation.

Approving the test+doc guard. Not merging from here.

aryanorastar:
macOS analytics tests on main compare the install id with itself. This head pins the literal id and the closed super-property set. Hygiene was skipped on `e7d5f968cffa`, so this stays open and unapproved.

aryanorastar:
Closing.

PR #12065, opened 2026-08-22
fix(backend): usage counters written with set(merge=True) never nest,…

abunet:
… so every summary reads empty

`record_llm_usage` and its siblings build keys like `chat.gpt-4o.input_tokens` — the module's own comment says "Use nested field paths" and it sanitises `.` out of the model name precisely so the remaining dots are separators. But those dicts are written with `set(..., merge=True)`, and Firestore treats a dot as a field PATH only in `update()`; in `set()` it is a literal character in the field NAME.

Measured against the Firestore emulator, after two calls of 5 and 7 input tokens:

    {'date': '2026-01-01', 'chat.model-x.input_tokens': 12, ...}

The counter is correct. The shape is not: nothing is nested under `chat`. Every reader walks feature -> model -> counter and skips a value that is not a dict — `_aggregate_summary`, `get_usage_summary`, `get_plan_usage_report` — so usage summaries come back EMPTY while the data is sitting in the document.

All four `set(..., merge=True)` writes in the module are affected, including the keys added by `_record_plan_bucket` and `_record_plan_metadata`, so plan attribution reads empty the same way.

The fix expands dotted paths into nested maps at the write. Every dot in these keys is a separator by construction: the model name, the cost-exclusion label and the plan id are each sanitised of '.' before interpolation, and the remaining segments are literals in this module.

Why it was invisible: the existing tests assert the PAYLOAD, e.g. `"chat.gpt-4_1-mini.input_tokens" in call["data"]` — which was exactly right, and says nothing about what Firestore then stores. Those five assertions are re-expressed on the stored shape rather than deleted; they still hold the property they were written for (a model name's dots are sanitised so they cannot be mistaken for separators).

Verification
  emulator, before: `keys with a literal dot: ['chat.model-x.call_count', 'plan_usage._unattributed...']`
                    and `nested under 'chat'? False`
  emulator, after:  `keys with a literal dot: []`, `chat -> {'model-x': {'input_tokens': 12, ...}}`
  tests/unit/test_llm_usage_db.py 19 passed, test_llm_usage_tracker.py 18, test_llm_usage_endpoints.py 2
  new tests/unit/test_llm_usage_nested_field_paths.py 6 passed
  tests/unit/test_desktop_migration.py 114 passed after the second commit updates four assertions that
    pinned the pre-fix shape (they looked for a top-level key literally named `desktop_chat.input_tokens`)
  reverting database/llm_usage.py to upstream turns 9 tests red across test_desktop_migration.py and
    test_llm_usage_db.py, the four rewritten ones among them — they exercise the writer, not themselves

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12065?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/llm_usage.py
+++ b/backend/database/llm_usage.py
@@ -82,6 +82,37 @@ def _plan_key(uid: str, firestore_client: Any | None) -> str:
     return resolve_usage_plan_id(uid, firestore_client=firestore_client) or _UNATTRIBUTED_PLAN
 
 
+def _nested(update: Dict[str, Any]) -> Dict[str, Any]:
+    """Expand dotted field paths into nested maps, for writes that use ``set(..., merge=True)``.
+
+    Firestore treats a dot as a field PATH in ``update()`` but as a literal character in ``set()``. These
+    usage documents are written with ``set(merge=True)`` because they must be created on first use, so a
+    key like ``chat.gpt-4o.input_tokens`` lands as ONE top-level field whose name contains dots — the
+    counter still increments correctly, but nothing is nested, and every reader that walks
+    ``feature -> model -> counter`` (``get_usage_summary``, ``_aggregate_summary``,
+    ``get_plan_usage_report``) skips it and reports empty usage.
+
+    Every dot in these keys is a separator by construction: the model name, the cost-exclusion label and
+    the plan id are each sanitised of ``.`` before being interpolated, and the remaining segments are
+    literals in this module.
+    """
+    nested: Dict[str, Any] = {}
+    for path, value in update.items():
+        if '.' not in path:
+            nested[path] = value
+            continue
+        cursor = nested
+        *parents, leaf = path.split('.')
+        for segment in parents:
+            branch = cursor.get(segment)
+            if not isinstance(branch, dict):
+                branch = {}
+                cursor[segment] = branch
+            cursor = branch
+        cursor[leaf] = value
+    return nested
+
+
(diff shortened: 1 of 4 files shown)


kodjima33 (approved):
Confidence 4/5 (clear root cause, scoped diff ~192 lines/3 files, new dedicated regression test file proving the bug via emulator + guard-removal check) — but the required 'PR Metadata Preflight'/'Hygiene' check (failure-class-protocol) fails: the PR body declares both 'Failure-Class: new' and the unfilled template's 'Failure-Class: none', which the repo's own contract rejects as multiple declarations. Approving but NOT merging per the CI hard floor; author needs to clean up the PR body (single Failure-Class line) before this can land.

kodjima33 (approved):
fix(backend): set(merge=True) writes with dotted keys never nest in Firestore, so every usage-summary reader returns empty — confirmed still broken on main. Root cause proven against the emulator, scoped diff (224 lines), 4/4 existing test files updated + new regression suite, no failing required CI checks. Confidence 5/5. Merging.

PR #12066, opened 2026-08-22
fix(backend): a __name__ filter needs a Key, so both usage readers ra…

abunet:
…ised on every call

`get_usage_summary` and `get_plan_usage_report` bound their scan with `where("__name__", ">=", cutoff_id)`, where `cutoff_id` is a date string like `2026-01-01`. Firestore requires a Key there:

    google.api_core.exceptions.InvalidArgument: 400 __key__ filter value must be a Key

Measured against the Firestore emulator: both functions raise, for every user, every time. The `__name__` field is fine — the value's TYPE is not.

The document ids ARE the dates, so the cutoff is simply `usage_collection.document(cutoff_id)`. The window is unchanged; only the type is.

Worth knowing for whoever reviews this: fixing it alone makes `get_usage_summary` return `{}` rather than raise, because a second, independent defect keeps the counters from ever nesting (a sibling PR: these documents are written with `set(merge=True)` and dotted keys, which Firestore stores as literal field names). With both applied, the same call against the emulator returns `{'chat': {'input_tokens': 5, 'output_tokens': 3, 'call_count': 1}}`. Either PR stands alone; neither restores the feature by itself.

One observation not addressed here, to keep this change to one thing: once the counters do nest, `_aggregate_summary` also reports `plan_usage` as if it were a feature (`{'input_tokens': 0, ...}`), because it walks every top-level map. That is a separate line and a separate decision.

Verification
  emulator, before: InvalidArgument 400 __key__ filter value must be a Key
  emulator, after:  the query runs and streams the in-window documents
  new tests/unit/test_llm_usage_name_filter_takes_a_key.py 3 passed; passing the bare string again
    fails all 3
  tests/unit/test_llm_usage_db.py 19 - test_llm_usage_tracker.py 18 - test_llm_usage_endpoints.py 2

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12066?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/llm_usage.py
+++ b/backend/database/llm_usage.py
@@ -351,7 +351,9 @@ def get_usage_summary(uid: str, days: int = 30) -> Dict[str, Dict[str, int]]:
     cutoff = datetime.now(timezone.utc) - timedelta(days=days)
     cutoff_id = f"{cutoff.year}-{cutoff.month:02d}-{cutoff.day:02d}"
 
-    docs = usage_collection.where("__name__", ">=", cutoff_id).stream()
+    # The value of a `__name__` filter must be a Key, not a string: Firestore answers
+    # `400 __key__ filter value must be a Key` otherwise, so this raised on every call.
+    docs = usage_collection.where("__name__", ">=", usage_collection.document(cutoff_id)).stream()
 
     # Aggregate by feature
     summary: Dict[str, Dict[str, int]] = {}
@@ -583,7 +585,8 @@ def get_plan_usage_report(uid: str, days: int = 30) -> Dict[str, Dict[str, Any]]
     usage_collection = db.collection('users').document(uid).collection('llm_usage')
     report: Dict[str, Dict[str, Any]] = {}
 
-    for doc in usage_collection.where('__name__', '>=', cutoff_id).stream():
+    # A `__name__` filter takes a Key, not a string (see get_usage_summary above).
+    for doc in usage_collection.where('__name__', '>=', usage_collection.document(cutoff_id)).stream():
         data = _typed_doc(doc)
         plan_usage = data.get('plan_usage')
         if isinstance(plan_usage, dict):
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Confidence 4/5 (clear root cause, scoped diff ~126 lines/2 files, new dedicated regression test proving the __key__ TypeError via emulator) — but the required 'PR Metadata Preflight'/'Hygiene' check (failure-class-protocol) fails: the PR body declares both 'Failure-Class: new' and the unfilled template's 'Failure-Class: none'. Approving but NOT merging per the CI hard floor; author needs to clean up the PR body before this can land. Note: per this PR's own body, applying it alone still leaves get_usage_summary returning {} until the sibling PR #12065 (nested-write fix) also lands.

kodjima33 (approved):
fix(backend): __name__ filter requires a Key not a string — confirmed still broken on main (raises 400 on every call). Root cause proven against emulator, scoped diff (126 lines), new regression tests. Not merging yet: 'Backend Hermetic Merge Gate' shows failure (downstream of a cancelled 'Detect Hermetic Backend Scope' run, likely a concurrency artifact from re-running CI on this branch) and 'Backend unit suite'/'Hygiene' are still mid-run. Confidence would be 5/5 once CI is unambiguously green — re-check before merging.

kodjima33 (approved):
Backend bug fix, 5/5 confidence: stated root cause with reproduced Firestore error, fix(backend) title, scoped diff (124 add/2 del, 2 files), CI green (Backend unit suite + Hermetic Merge Gate pass), regression tests added (3 new, verified fail on revert). Confirmed still broken on origin/main (bare-string __name__ filter in both get_usage_summary and get_plan_usage_report), not superseded by sibling PR #12065's separate nesting fix.

PR #12067, opened 2026-08-22
test(backend): stub with_rate_limit so test_task_sharing can be colle…

abunet:
…cted

tests/unit/test_task_sharing.py fails at COLLECTION on main:

    tests/unit/test_task_sharing.py:144: in <module>
        import routers.action_items as action_items_router
    routers/action_items.py:362: in <module>
        uid: str = Depends(auth.with_rate_limit(auth.get_current_user_uid, "action_items:list")),
    AttributeError: module 'utils.other.endpoints' has no attribute 'with_rate_limit'

The list route started wrapping its auth dependency in the rate-limit decorator at module level, and this file stubs utils.other.endpoints with only get_current_user_uid. Nothing in the file runs -- all 30 tests are lost, and the failure is a collection error rather than a test failure, so it does not look like a broken assertion.

It is invisible to a changed-file selector: the PR that added the decorator touched routers/action_items.py and utils/other/endpoints.py, not this test, so the selection never included it. Found by running the whole backend suite file-by-file.

The stub returns the dependency itself rather than a MagicMock: the decorator's return value IS the FastAPI dependency, and a MagicMock there fails signature inspection further in.

Verification: tests/unit/test_task_sharing.py -> 30 passed (was: 1 collection error).

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12067?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_task_sharing.py
+++ b/backend/tests/unit/test_task_sharing.py
@@ -123,6 +123,11 @@ utils_users_mod.get_user_display_name = MagicMock(return_value="TestUser")
 _stub_module("utils.other")
 _stub_module("utils.other.endpoints")
 sys.modules["utils.other.endpoints"].get_current_user_uid = MagicMock()
+# The list route wraps its auth dependency in the rate-limit decorator at MODULE level
+# (routers/action_items.py), so the stub has to carry it or importing the router raises before a
+# single test runs. Identity rather than a MagicMock: the return value IS the FastAPI dependency,
+# and a MagicMock there fails signature inspection further in.
+sys.modules["utils.other.endpoints"].with_rate_limit = lambda dependency, _policy: dependency
 
 # The action-items router imports the list-read budget seam at module level
 # (#11831). Delegate the stubbed submodule to the real (stdlib-only) module so


Git-on-my-level:
Verified — this is a real collection failure and the fix is correct.

What I checked:

- Diagnosis confirmed: `routers/action_items.py:362` wraps the list route's auth dependency as `Depends(auth.with_rate_limit(auth.get_current_user_uid, "action_items:list"))` at module import, so importing the router against a stubbed `utils.other.endpoints` without `with_rate_limit` raises `AttributeError` before a single test runs — exactly the failure in your description.

- Reproduced on the parent commit (9a4951f4c5): collecting `backend/tests/unit/test_task_sharing.py` fails with `AttributeError: module 'utils.other.endpoints' has no attribute 'with_rate_limit'` — a collection error, so all 30 tests are silently skipped rather than reported as failing. Your point that this is invisible to a changed-file selector is well taken: the PR that added the decorator touched `routers/action_items.py` and `utils/other/endpoints.py`, not this test.

- With this PR's head (39ca4045): `tests/unit/test_task_sharing.py` collects and passes 30/30 (0.32s). I also ran it immediately after `test_mcp_profile_contact.py`, which installs its own `with_rate_limit` stub on the same module — 35/35 passed, so the fix behaves under the sibling-stub ordering too.

- The identity lambda rather than a MagicMock is the right call: `with_rate_limit(auth_dependency, policy_name)` is invoked with two positional arguments at import, and FastAPI's dependency signature inspection downstream needs the real callable. This matches the existing pattern in the suite (`test_desktop_migration.py:175-176`, `test_mcp_profile_contact.py:127-128`).

Nice forensic work tracking this down via a full file-by-file suite run — collection errors hiding a whole test file are exactly the kind of silent regression worth this kind of fix. Thank you!

---

abunet:
Superseded by #12057, which landed the same stub on 2026-08-22 — `sys.modules["utils.other.endpoints"].with_rate_limit = lambda auth_dependency, policy_name: auth_dependency`, now at test_task_sharing.py:131 on main. Same diagnosis, same pass-through; only the parameter names differ, so there is nothing left here to merge. Thanks for picking it up.

PR #12069, opened 2026-08-22
fix(backend): declare the missing composite index for plugin_id-filtered message reads

igor-popov-dev:
## What changed and why

`get_app_messages` and `get_messages`'s app-scoped branch (both in `backend/database/chat.py`)
filter the `messages` collection by `plugin_id` and order by `created_at` descending, but
`firestore_index_registry.py` never declared the composite Firestore needs for that shape.
Production has it only because it was created by hand at some point; a fresh self-host deploy
gets `FailedPrecondition` 400 on `GET /v1/messages`.

Adds `MESSAGES_BY_APP_ORDERED_QUERY` to the registry (mirrors the already-registered
`chat_sessions_current_by_app_created_at` shape) and regenerates `firestore.indexes.json`.

none

## How it was verified

  `firestore.indexes.json` from the updated registry; diff is exactly the new
  `messages(plugin_id ASC, created_at DESC, __name__ DESC)` composite.
- `.venv/bin/python -m pytest backend/tests/unit/test_firestore_query_contract.py -m ""` —
  31/31 passed, including the new regression test that builds both call sites' real query
  chain through a recording Firestore fake and asserts the composite is declared.
- `.venv/bin/python -m pytest backend/tests/unit/test_reconcile_firestore_indexes.py
  backend/tests/unit/test_action_items_due_range_index_contract.py -m ""` — 75/75 passed,
  no regression in the surrounding index-reconcile/contract suite.

## Tests

Bug fix — regression test added: `test_app_scoped_message_reads_have_a_declared_composite_index`
in `backend/tests/unit/test_firestore_query_contract.py` calls `get_app_messages` and
`get_messages` against a recording Firestore fake, extracts the real filter+order chain each
builds, and asserts `firebase_index_manifest()` declares a matching composite — so either
function regressing to an undeclared shape fails this test, not just self-host in production.

## Failure class (fixes)

## Scoped cleanups (optional)

Generalized `_StreamRecordingFirestore`/`_StreamRecordingUserRef` in the test file to take an
optional `collection_name` (defaulted to `'action_items'` to keep the existing action-items test
unchanged) instead of hardcoding it, so the new messages test could reuse the same harness.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):

PR #12070, opened 2026-08-22
fix(app): stop retrying a dead speech-profile socket, offer skip instead

igor-popov-dev:
## What changed and why

`closeCode` 1011 with zero captured user segments means the onboarding speech-profile
socket's STT backend is down server-side, not that the socket merely hiccuped. The reconnect
loop added in #11704 retried this every 5s forever: each failed reconnect re-triggers
`onClosed`, which re-shows the "Connection Lost" dialog (`speech_profile_widget.dart` /
`page.dart`) and reschedules another attempt, even though the same screen already has a
working "Skip for now" link (`speech_profile_widget.dart:487-496`) right next to it.

After 3 consecutive 1011 closes with no captured speech, `SpeechProfileProvider` now stops
scheduling reconnects and surfaces a distinct `STT_UNAVAILABLE` error instead of
`SOCKET_DISCONNECTED`. The onboarding widget routes that to the existing skip path
(`provider.close()` + `widget.onSkip()`) instead of a dialog whose only button just popped
the screen without marking the step skipped.

none

## How it was verified

- `flutter test test/providers/speech_profile_provider_test.dart` — 7/7 passed, including
  3 new regression tests (gives up and surfaces `STT_UNAVAILABLE` after 3×1011-with-no-speech;
  keeps retrying on an ordinary 1006 disconnect; does not give up once real speech was already
  captured).
- `dart format --line-length 120 --set-exit-if-changed` on the 4 changed files — clean.
- `flutter analyze` on the 4 changed files — no issues found.

## Tests

Bug fix — regression tests added in `test/providers/speech_profile_provider_test.dart`
covering the new give-up threshold, that an ordinary disconnect is unaffected, and that
already-captured speech prevents a false give-up.

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/onboarding/speech_profile_widget.dart
+++ b/app/lib/pages/onboarding/speech_profile_widget.dart
@@ -252,6 +252,27 @@ class _SpeechProfileWidgetState extends State<SpeechProfileWidget> with TickerPr
                   ),
                   barrierDismissible: false,
                 );
+              } else if (error == 'STT_UNAVAILABLE') {
+                // The provider already gave up reconnecting after repeated
+                // 1011 closes with no captured speech, so offer the same
+                // way out as the "Skip for now" link instead of a "Try
+                // again" that would only restart the same failing loop.
+                showDialog(
+                  context: context,
+                  builder: (c) => getDialog(
+                    context,
+                    () {
+                      provider.close();
+                      widget.onSkip();
+                    },
+                    () {},
+                    context.l10n.connectionLost,
+                    context.l10n.connectionLostDesc,
+                    okButtonText: context.l10n.skipForNow,
+                    singleButton: true,
+                  ),
+                  barrierDismissible: false,
+                );
               }
             },
             child: Column(
--- a/app/lib/pages/speech_profile/page.dart
+++ b/app/lib/pages/speech_profile/page.dart
@@ -192,7 +192,7 @@ class _SpeechProfilePageState extends State<SpeechProfilePage> with TickerProvid
                   ),
                   barrierDismissible: false,
                 );
-              } else if (error == 'SOCKET_DISCONNECTED' || error == 'SOCKET_ERROR') {
+              } else if (error == 'SOCKET_DISCONNECTED' || error == 'SOCKET_ERROR' || error == 'STT_UNAVAILABLE') {
                 showDialog(
                   context: context,
                   builder: (c) => getDialog(
--- a/app/lib/providers/speech_profile_provider.dart
+++ b/app/lib/providers/speech_profile_provider.dart
@@ -55,6 +55,13 @@ class SpeechProfileProvider extends ChangeNotifier
   Timer? _reconnectTimer;
   bool _reconnecting = false;
 
+  /// Consecutive closes with code 1011 (server-side STT failure) while no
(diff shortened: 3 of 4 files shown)


Git-on-my-level:
Thanks @igor-popov-dev — this is a well-scoped fix for a real onboarding trap, and the regression tests pin exactly the right behaviors. Verified against the head; notes below.

What checks out:

- `app/lib/providers/speech_profile_provider.dart` — the new `_sttUnavailableCloseCount` / `_maxSttUnavailableCloses = 3` gate in `onClosed()` correctly narrows on `closeCode == 1011 && segments.isEmpty`, and the premise matches the backend: `backend/routers/listen/runtime.py` closes this socket with 1011 on terminal STT failure, so treating it differently from a 1006 blip is right. Any non-qualifying close resets the counter, and the give-up path cancels `_reconnectTimer` before emitting `STT_UNAVAILABLE`, so no stray reconnect is left running.
- `app/lib/pages/onboarding/speech_profile_widget.dart` — routing `STT_UNAVAILABLE` to a single-button "Skip for now" dialog that calls `provider.close()` + `widget.onSkip()` matches the existing skip link at the bottom of the same screen, and it reuses existing l10n keys (`skipForNow`, `connectionLost`, `connectionLostDesc` — all present in `app/lib/l10n/app_en.arb`), so no new translations are needed.
- `app/lib/pages/speech_profile/page.dart` — folding `STT_UNAVAILABLE` into the existing disconnect branch is a reasonable minimal choice for the standalone speech-profile page.
- `app/test/providers/speech_profile_provider_test.dart` — the three new tests (give-up after 3×1011 with no speech, ordinary 1006 keeps retrying, already-captured speech prevents a false give-up) directly pin the new behavior, and they reuse the existing `_CountingSpeechProfileProvider` fixture cleanly.

Two minor, non-blocking notes for a possible follow-up:

1. `close()` doesn't reset `_sttUnavailableCloseCount`. After a give-up → skip in onboarding, the counter stays at 3, so a later speech-profile attempt in the same app process would surface `STT_UNAVAILABLE` on the first 1011 close instead of the third. Impact is low (same dialog, just earlier), but resetting it in `close()` alongside `segments.clear()` would keep sessions independent.
2. In `page.dart`, the dialog shown for `STT_UNAVAILABLE` still labels its button "Try again" while the action is `provider.close()` + pop. That wording is inherited from the existing `SOCKET_DISCONNECTED` branch, so it's not a regression, but dedicated skip/close wording would read better for the STT-down case.

Neither blocks merge from my side — leaving for human maintainer review.

---

🤖 This comment was AI-generated on behalf of the Omi maintainer team.

---

kodjima33 (approved):

PR #12071, opened 2026-08-22
fix(app): stop a custom-STT hiccup from tearing down the transcription socket

igor-popov-dev:
## What changed and why

A single failed request to a custom STT endpoint used to be treated as fatal all the way
up the stack: `SchemaBasedSttProvider.transcribe()` had no retry and a 60s timeout, and
once it threw, `PurePollingSocket` discarded the buffered audio and reported it as a
fatal socket error. `CompositeTranscriptionSocket` reacts to *any* child socket error by
tearing down both sockets — so one slow/dropped custom-STT request also killed the
healthy raw-audio socket and forced a full transcription-socket reconnect. In practice
this meant every transient hiccup produced a small transcript gap, and a real outage
turned into a reconnect storm roughly every 60 seconds instead of a quiet local buffer.

- `SchemaBasedSttProvider` now retries a request up to 3 times with backoff on network
  errors/5xx (not 4xx, which won't be fixed by retrying) at a 10s per-attempt timeout
  instead of a single 60s attempt.
- The multipart request path called bare `MultipartRequest.send()`, which per
  `package:http` silently creates and discards a throwaway `http.Client()` instead of
  reusing the provider's own `_client` — routed it through `_client.send(request)`
  instead, which also makes the retry logic (and injecting a test client) actually work
  for the multipart path used by the custom STT provider.
- `PurePollingSocket` now requeues audio from a failed flush (capped by
  `maxBufferBytes`, oldest dropped first) instead of discarding it, and no longer reports
  a failed `transcribe()` call as a fatal socket error. It keeps buffering and retrying
  locally on the next timer tick, so a transient hiccup no longer cascades into tearing
  down the whole composite socket.
- `CaptureController`/the recording UI can now show how long custom STT has been
  unreachable ("Offline, buffering Nm") instead of silently claiming "Listening" while
  nothing is actually being transcribed.

none

## How it was verified

- `flutter test test/unit/pure_polling_test.dart test/unit/transcription_polling_service_retry_test.dart`
  — 7/7 passed (3 new tests covering requeue-on-failure/no-teardown, continued retrying
  while an endpoint stays down, and the buffer cap; 4 new tests covering retry-on-5xx,
  no-retry-on-4xx, giving up after max attempts, and the per-attempt timeout).
- `flutter test test/unit/composite_transcription_socket_test.dart` — 8/8 passed
  (unchanged; confirms the composite's other behavior — raw-audio forwarding, secondary
  socket handling — is untouched).
- `dart format --line-length 120 --set-exit-if-changed` on the 4 changed files — clean.
- `flutter analyze` on the 4 changed files — no issues found.

## Tests

Bug fix — regression tests added in `test/unit/pure_polling_test.dart` (buffer
requeue/no-teardown/cap behavior, using a fake `ISttProvider`) and
`test/unit/transcription_polling_service_retry_test.dart` (retry/timeout behavior of the
real HTTP request path, using `package:http/testing.dart`'s `MockClient` inside
`fake_async` so no test relies on wall-clock delays).

## Failure class (fixes)

Failure-Class: none

## Scoped cleanups (optional)

none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12071?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/conversations/widgets/processing_capture.dart
+++ b/app/lib/pages/conversations/widgets/processing_capture.dart
@@ -44,7 +44,7 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
     _offlineTicker = Timer.periodic(const Duration(seconds: 1), (_) async {
       if (!mounted) return;
       final provider = context.read<CaptureProvider>();
-      if (provider.offlineRecordingStartedAt != null) {
+      if (provider.offlineRecordingStartedAt != null || provider.customSttBufferingDuration != null) {
         setState(() {});
       }
       _offlineTick++;
@@ -280,10 +280,17 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
     } else if (!isHavingRecordingDevice && !isUsingPhoneMic) {
       stateText = "";
     } else if (isUsingPhoneMic || isHavingRecordingDevice) {
+      final bufferingFor = captureProvider.customSttBufferingDuration;
       if (captureProvider.terminalTranscriptionFailure != null) {
         // Audio remains in the WAL while reconnecting, but the server has
         // explicitly said live STT is unavailable. Do not claim "Listening".
         stateText = context.l10n.transcriptionUnavailable;
+        statusIndicator = const PausedStatusIndicator();
+      } else if (bufferingFor != null) {
+        // Custom STT endpoint unreachable. Audio keeps recording
+        // and buffering locally (see PurePollingSocket) — say so instead of
+        // silently claiming "Listening" while nothing is being transcribed.
+        stateText = _customSttBufferingText(bufferingFor);
         statusIndicator = const PausedStatusIndicator();
       } else {
         // Show "Listening" for all active recording states — WAL ensures audio is
@@ -333,6 +340,13 @@ class _ConversationCaptureWidgetState extends State<ConversationCaptureWidget> {
     );
   }
 
+  // Short status text for how long the custom STT endpoint has
+  // been unreachable while audio keeps recording and buffering locally.
+  String _customSttBufferingText(Duration bufferingFor) {
+    if (bufferingFor.inMinutes < 1) return 'Offline, buffering';
+    return 'Offline, buffering ${bufferingFor.inMinutes}m';
+  }
+
   Widget _buildUnifiedRecordingUI(CaptureProvider provider, Widget? header) {
     bool isDeviceRecording = provider.havingRecordingDevice &&
         (provider.recordingState == RecordingState.deviceRecord || provider.recordingState == RecordingState.pause);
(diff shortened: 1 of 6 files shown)


Git-on-my-level:
Thanks @igor-popov-dev — well-scoped fix for a real reliability bug, and the diagnosis matches the code: `CompositeTranscriptionSocket._onSocketError` does tear down both children on any child error, so one failed custom-STT flush previously killed the healthy raw-audio channel too.

Per-file notes:

- `app/lib/services/sockets/pure_polling.dart` — requeue logic is correct: `_flushBuffer` copies-then-clears `_audioFrames` before awaiting `transcribe`, and `_requeueFrames` inserts failed frames back at index 0 so they stay ahead of audio captured mid-flight; `_totalBufferBytes` is recomputed from the frame list each call, so the `maxBufferBytes` trim (oldest-first, with warning logs) can't drift. Not calling `onError` here is the right call given the provider now retries internally.
- `app/lib/services/sockets/transcription_polling_service.dart` — `_sendWithRetry` cleanly separates retryable (exceptions incl. the new 10s timeout, 5xx) from terminal (4xx); `_retryBackoff[i]` is only indexed on non-final attempts so it can't overflow; and routing the multipart path through `_client.send()` instead of bare `MultipartRequest.send()` fixes both the throwaway-client behavior and client injection. One pre-existing gap this PR doesn't change: on a persistent 5xx the last response is returned, `transcribe()` returns `null` on non-200, and the socket treats that as success — frames already cleared, buffering state cleared, UI back to "Listening". Worth a small follow-up (e.g. treat non-200 as a failure), but not a blocker since it matches today's behavior.
- `app/lib/services/capture/capture_controller.dart` — `_activeCustomSttPollingSocket` correctly unwraps both the composite (`primarySocket`) and standalone layouts, and computing `DateTime.now().difference` on read keeps the UI ticker simple. (The `buttonState` reformat looks formatter-driven; behavior identical.)
- `app/lib/pages/conversations/widgets/processing_capture.dart` — good ordering: terminal transcription failure still wins over the new buffering state, and the 1s ticker now refreshes while buffering. Two minor notes: the 'Offline, buffering' strings are hardcoded English while neighboring states use `context.l10n` (there's precedent — 'Capturing' — but l10n would be better), and once the ~10-minute buffer cap starts dropping audio the duration counter keeps growing, so the label can overstate what's actually retained.
- `app/test/unit/pure_polling_test.dart` — the three tests pin the key invariants: no error/close propagation, requeue ordering (`[1, 2, 3, 4, 5, 6]` on retry), retry growth while down, and oldest-first trimming at the cap.
- `app/test/unit/transcription_polling_service_retry_test.dart` — nice use of `fakeAsync` + `MockClient`; the 5xx-retry / 4xx-no-retry / exhaustion / sub-60s-timeout matrix covers the new behavior well.

`_consecutiveFailures` in `PurePollingSocket` is tracked and exposed but not consumed yet — fine if it's groundwork, just flagging it.

Overall a clear improvement: bounded memory during outages, honest offline UI state, and no more reconnect storms. The one design choice worth a maintainer's eye is the never-tear-down semantics — the socket now retries forever rather than ever surfacing a fatal error — but with the terminal-failure path and visible buffering state intact, that seems right. Leaving the merge call to a human maintainer.

---

kodjima33 (approved):

PR #12072, opened 2026-08-22
fix(app): report why a phone call was refused instead of always blaming verification

igor-popov-dev:
# What changed and why

**What's wrong**

Starting a call surfaces `Failed to get call token. Verify your phone number first.` for
*every* non-200 answer from `POST /v1/phone/token`. `getPhoneCallToken()` throws the response
body away and returns `null`, so the provider has nothing but a fixed guess to show.

That guess is wrong in every case except one. A user whose number *is* verified sees it when:

- the monthly free-tier quota is exhausted — `402` with
  `{'error': 'phone_call_quota_exceeded', ...}` (`check_call_access`, `backend/utils/phone_calls.py`);
- their plan doesn't include calling — `403 "Phone calls require a paid subscription"`;
- Twilio is misconfigured server-side — `500 "Failed to generate token: ..."`.

In all three, the app tells them to re-verify a number that is already verified.

There's a second, sharper edge on the same data. `startVerification` does
`result['error'] as String?`, but `detail` is untyped on the wire and the quota error answers
with a **map**. The cast throws a `TypeError` inside an un-awaited async path, so the
number-entry screen fails instead of explaining the quota.

**What this changes**

- `getPhoneCallToken()` returns a small `PhoneCallTokenResult` (token *or* the backend's own
  reason) instead of a bare nullable token; the provider shows that reason and keeps a neutral
  fallback (`Failed to get call token. Please try again.`) for the no-body case.
- One shared `errorDetailMessage(body)` reader handles both `detail` shapes — plain string and
  the quota map (rendered as `Monthly call limit reached (N calls). ...`) — so neither call
  site can trip over the map again.

No API surface, no string that a user could previously rely on, no behavior change on the
happy path.

none

## How it was verified

- `flutter test test/unit/phone_call_error_detail_test.dart` — 5/5 green.
- Adjacent suites touching the same surfaces (`phone_number_input_test.dart`,
  `batch_recording_phone_test.dart`, `phone_mic_wal_test.dart`) — 81/81 green.
- `flutter analyze` on the changed files — clean apart from a pre-existing `avoid_print`
  not introduced by this change; `dart format --line-length 120` — clean.
- Not verified live: an end-to-end refused call against a backend answering 402/403/500
  (no funded Twilio account on the dev setup). The error-body shapes the tests encode are
  taken directly from `check_call_access` / the token route in `backend/utils/phone_calls.py`
  and `backend/routers/phone_calls.py`.

## Tests

`app/test/unit/phone_call_error_detail_test.dart` covers both `detail` shapes (plain string
and the quota map), a blank detail, an unknown map code, and a body with no detail — the
map-shaped `402` case is the regression test that would have caught the original
`TypeError`.

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/app/lib/backend/http/api/phone_calls.dart
+++ b/app/lib/backend/http/api/phone_calls.dart
@@ -8,6 +8,40 @@ import 'package:omi/env/env.dart';
 import 'package:omi/utils/logger.dart';
 
 // ************************************************
+// ************** ERROR REPORTING *****************
+// ************************************************
+
+/// Human-readable reason out of a FastAPI error body, or null when the body
+/// carries none.
+///
+/// ``detail`` is deliberately untyped on the wire: most phone-call endpoints
+/// answer with a plain string, but the quota error answers with a map (see
+/// ``check_call_access`` in ``backend/utils/phone_calls.py``). Reading it as a
+/// string crashes on that map, so both shapes are handled here once.
+String? errorDetailMessage(String body) {
+  try {
+    final parsed = misc_wire.GeneratedErrorResponse.fromJson(jsonDecode(body) as Map<String, dynamic>);
+    final detail = parsed.detail;
+    if (detail is String) {
+      return detail.trim().isEmpty ? null : detail;
+    }
+    if (detail is Map) {
+      final code = detail['error'];
+      if (code == 'phone_call_quota_exceeded') {
+        final limit = detail['monthly_limit'];
+        return limit is int
+            ? 'Monthly call limit reached ($limit calls). It resets at the start of next month.'
+            : 'Monthly call limit reached. It resets at the start of next month.';
+      }
+      if (code is String && code.trim().isNotEmpty) {
+        return code;
+      }
+    }
+  } catch (_) {}
+  return null;
+}
+
+// ************************************************
 // *********** PHONE NUMBER MANAGEMENT ************
 // ************************************************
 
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Verified against the backend and the wire types - this checks out nicely.

- `app/lib/backend/http/api/phone_calls.dart` - `GeneratedErrorResponse.detail` really is `dynamic` (`misc_wire.g.dart`), and `check_call_access` really does answer `402` with a map (`phone_call_quota_exceeded` + `monthly_limit`), so a shared `errorDetailMessage()` that handles both `detail` shapes is the right fix at the right layer. Routing `getPhoneCallToken()` through `PhoneCallTokenResult` keeps the backend's refusal reason instead of throwing the body away, and reusing the same reader in `verifyPhoneNumber` also removes the `result['error'] as String?` `TypeError` path in `startVerification` - the quota map genuinely reaches `POST /v1/phone/numbers/verify` too, since that route also calls `check_call_access` before anything else.
- `app/lib/providers/phone_call_provider.dart` - showing `tokenResult.error` with the neutral fallback beats the old fixed "Verify your phone number first." guess, which was wrong for quota (402), plan (403), and server misconfig (500) alike; the old string has no remaining code dependency, and the refresh-timer call site correctly unwraps `.token` while keeping its retry-on-null behavior. Tiny pre-existing nit while you're in the area: the refresh-failure log says "retrying in 30s" but schedules 60 - feel free to align in a follow-up commit, not blocking.
- `app/test/unit/phone_call_error_detail_test.dart` - good hermetic coverage of both `detail` shapes, the quota-map rendering, and the missing/non-JSON fallbacks; the quota-map case is exactly the regression test the old crash needed.

Checks are green (Dart Analyze & Tests, Android Compile Smoke, Formatting, Hygiene), and the explicit "not verified live against a funded Twilio backend" note in the description is appreciated. Leaving for routine maintainer review - nice first contribution, thank you!

---

kodjima33 (approved):

PR #12034, opened 2026-08-21
docs(desktop-windows): add AGENTS.md, index it from root, fix a stale release-pipeline.md claim

formed2forge:
## Summary
- `desktop/windows/` was the only major component (`backend/`, `app/`, `desktop/macos/`, `web/app/`, `omi/firmware/` all have one) without an `AGENTS.md`. Adds one covering: the pnpm-vs-npm gotcha and pnpm-10 version pin, day-to-day dev/test/build commands, what CI actually covers (and doesn't — e.g. the Windows job never runtime-smokes the packaged binary, unlike the Linux job), the release pipeline (pointer to `docs/release-pipeline.md` + the known-flaky sync-to-main step, issue #10727), the Linux/Wayland dev-environment quirks found this session (niri needs `OMI_OZONE=wayland`; the companion bar can't position itself under native Wayland), and an index of the existing `docs/*.md` files.
- Indexes the new file from the root `AGENTS.md`'s "Read Next" table (it was missing there too) and registers its size budget in `check_agents_md_lean.py`.
- Adds the `OMI_OZONE` env var to `docs/multi-worktree-dev.md`'s existing dev-env-var table, alongside `OMI_DEV_HW_GPU`.

## Second commit (separate, small fix)
- `docs/release-pipeline.md` claimed the release workflow triggers "on every push to main" — the actual `desktop_windows_release.yml` `on:` block is `workflow_dispatch` only. Caught while cross-checking the doc against the workflow file for the AGENTS.md I was writing. Docs-only, no behavior change.

Failure-Class: none

## Test plan
- [x] `python3 .github/scripts/check_agents_md_lean.py` passes (9 AGENTS.md files within budget).
- [x] Cross-checked every claim in the new AGENTS.md against the actual `package.json` scripts, CI workflow YAML, and this session's own verified local testing (Linux build, `OMI_OZONE=wayland`, pnpm version pin) rather than assuming.
- [x] Docs-only change — no code/behavior affected.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12034?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #12075, opened 2026-08-22
fix(desktop-windows): stop the bar auto-retracting on native Wayland

formed2forge:
## Summary
- The peek-pill retract watchdog trusted `screen.getCursorScreenPoint()` as the sole signal for "has the cursor visited the pill" (`src/main/bar/watchdog.ts`). That API is unreliable on native Wayland (the protocol doesn't let an app query the global pointer position outside its own focused surface), so `hasBeenHovered` never armed and every peek pill retracted to opacity 0 after the fixed ~3s linger, regardless of a real ongoing hover — the window stayed mapped (visible to the compositor) but its content was permanently invisible.
- Fix corroborates the OS reading with the renderer's own `mouseenter`-confirmed interactivity signal (already relayed via the existing `bar:setInteractive` IPC), which only ever makes the watchdog less eager to retract.

## Verification
- Reproduced live via CDP against the running dev instance: the bar's DOM was captured mounted correctly ("Listening" label + orb) but stuck in the `.bar-slide-out` class (`opacity: 0`).
- Confirmed fixed by the reporter on real hardware (Asahi Fedora Remix aarch64 + niri): the pill now stays open and expands to the chat surface on click.

## Test plan
- [x] Unit tests added in `watchdog.test.ts`
- [x] Manually verified on native Wayland (Asahi Fedora Remix + niri)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12075?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

### DIFF
--- a/desktop/windows/changelog/unreleased/2026-08-bar-wayland-auto-retract.json
+++ b/desktop/windows/changelog/unreleased/2026-08-bar-wayland-auto-retract.json
@@ -1,0 +1,5 @@
+{
+  "changes": [
+    "Fixed the companion bar auto-retracting to fully invisible ~3 seconds after every summon on native Wayland (Linux), regardless of an ongoing hover."
+  ]
+}
--- a/desktop/windows/src/main/bar/watchdog.test.ts
+++ b/desktop/windows/src/main/bar/watchdog.test.ts
@@ -5,6 +5,7 @@ import {
   barWatchPlan,
   barGestureSeesOpen,
   clickEdge,
+  corroboratedCursorInFootprint,
   type WatchdogInput
 } from './watchdog'
 
@@ -90,6 +91,22 @@ describe('evaluatePeekWatchdog — retract grace', () => {
       expect(s.retract).toBe(false)
       expect(s.outsideSince).toBe(null)
     }
+  })
+})
+
+describe('corroboratedCursorInFootprint — native-Wayland cursor-query fallback', () => {
+  it('trusts the renderer-reported hover even when the OS reading says outside', () => {
+    // screen.getCursorScreenPoint() is unreliable on native Wayland — a live
+    // renderer mouseenter must still keep the pill from starving hasBeenHovered.
+    expect(corroboratedCursorInFootprint(false, true)).toBe(true)
+  })
+
+  it('trusts the OS reading even when the renderer has not reported hover', () => {
+    expect(corroboratedCursorInFootprint(true, false)).toBe(true)
+  })
+
+  it('is false only when BOTH signals agree nothing is hovering', () => {
+    expect(corroboratedCursorInFootprint(false, false)).toBe(false)
   })
 })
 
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Thanks @formed2forge — this is a nicely contained fix for a real cross-platform gap. I verified the mechanism end-to-end against the tree at 3420262:

**`desktop/windows/src/main/bar/watchdog.ts`** — `corroboratedCursorInFootprint()` is a pure OR of the OS reading and the renderer-reported interactivity, so it can only make the watchdog *less* eager to retract, exactly as the doc comment claims. On Windows, where the cursor poll is reliable, `barInteractive` can only be true when the poll already sees the cursor over the pill (the interactivity half of `peekTick` re-derives it every tick), so this is effectively a no-op on the primary platform — good regression posture.

**`desktop/windows/src/main/bar/window.ts`** — the `peekTick()` call site is the right seam, and the corroborating signal is genuinely renderer-driven as described: `BarApp.tsx` wires real `onMouseEnter`/`onMouseLeave` on `.bar-surface` through `window.omiBar.setInteractive`, the preload bridges it, and the `bar:setInteractive` handler in `window.ts` keeps `barInteractive` live. I checked all three links.

One non-blocking integration note: within `peekTick()`, the interactivity half runs *before* the retract half, and on any tick where the OS poll reads outside, `applyClickThrough()` sets `barInteractive = false` a few lines before `corroboratedCursorInFootprint()` reads it. So the renderer's report only survives into the retract decision on ticks where the poll itself already agrees. That's consistent with the hardware result you reported (once the mouseenter makes the window interactive, the surface starts receiving pointer events and the poll refreshes), but on a compositor where `screen.getCursorScreenPoint()` stays stale even then, the OR would be clobbered within a single tick. If this bug ever recurs on such a setup, latching the renderer's report per reveal (reset alongside `peekHasBeenHovered` in `startPeekWatch()`) would make the corroboration independent of the poll. Not blocking — just worth keeping in mind.

**`desktop/windows/src/main/bar/watchdog.test.ts`** — the three new cases cover the meaningful rows of the OR truth table (renderer-only, OS-only, neither); verified they match the implementation by inspection.

**`desktop/windows/changelog/unreleased/2026-08-bar-wayland-auto-retract.json`** — matches the existing unreleased-changelog convention exactly (same `{"changes": [...]}` shape and filename pattern as its neighbors).

Small docs nit: the new comment in `watchdog.ts` cites "AGENTS.md's Linux dev environment section", but `desktop/windows/AGENTS.md` doesn't exist on `main` or in this PR — it only exists in your still-open #12034. If this lands first, the reference dangles; landing #12034 first or pointing at `desktop/windows/LINUX.md` instead would keep it accurate.

Nice debugging work — the CDP repro of the stuck `.bar-slide-out` state plus the on-hardware confirmation make this easy to trust. Leaving for human maintainer review before merge: native-Wayland cursor semantics vary by compositor, and a maintainer pass on the Linux build is a good final check.

---

kodjima33 (approved):
Owner-override desktop/windows bug fix: scoped, tested, CI green, not already fixed on main. Merging.

PR #12076, opened 2026-08-22
fix(desktop-windows): classify a permanent mic/loopback failure as terminal

formed2forge:
## Summary
- `isRetryableDropError` only inspected the error message, never the source error's `DOMException` `name` — so a permanent `getUserMedia` failure (no device, permission denied, device unreadable) fell through the same path as an ordinary transient network drop and retried silently through the whole ~4.5min backoff budget (`MAX_RECONNECT_ATTEMPTS=10`), with no error ever reaching `captureLiveStore` / the UI.
- Fix: pass the error's `name` through (`AudioSessionHost.ts`'s `audio-source-error` already relays it) and treat `NotFoundError`/`NotAllowedError`/`NotReadableError`/`OverconstrainedError` as terminal, alongside the existing quota/sign-in cases.

## Verification
Reproduced live on real hardware via CDP against the running dev instance's capture window: `navigator.mediaDevices.enumerateDevices()` returned zero audioinputs and `getUserMedia({audio:true})` threw `NotFoundError` ("Requested device not found") — a genuinely permanent source failure (separately confirmed as a PipeWire/asahi-audio config gap on that machine, not an Omi bug). The message text alone never matched `isRetryableDropError`'s quota/sign-in patterns, so this exact case retried forever before the fix.

**Known follow-up (not bundled here):** even with this fix, a non-quota terminal error still has no UI surface — `maybeTriggerTranscriptionQuotaPopup` (`usageLimit.ts`) only reacts to quota messages, so `captureLiveStore`'s `'error'` status is otherwise silently dropped. That's a broader product decision (what should a generic recording-failure toast say, where does it render), filed separately.

## Test plan
- [x] Unit tests added in `liveRescue.test.ts`
- [x] Reproduced and verified fixed on real hardware with a permanently unavailable mic

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12076?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Failure-Class: none

### DIFF
--- a/desktop/windows/changelog/unreleased/2026-08-mic-permanent-failure-classification.json
+++ b/desktop/windows/changelog/unreleased/2026-08-mic-permanent-failure-classification.json
@@ -1,0 +1,5 @@
+{
+  "changes": [
+    "A permanently missing/blocked microphone (no device, permission denied) no longer retries silently for minutes with no error shown — it's now classified as terminal, matching how quota/sign-in failures are already handled."
+  ]
+}
--- a/desktop/windows/src/renderer/src/capture/liveMicSession.ts
+++ b/desktop/windows/src/renderer/src/capture/liveMicSession.ts
@@ -277,7 +277,7 @@ export function startLiveMicSession(): LiveMicController {
             /* ignore */
           }
           handle = null
-          if (!isRetryableDropError((e as Error).message)) {
+          if (!isRetryableDropError((e as Error).message, (e as Error).name)) {
             // Quota/entitlement/sign-in error — reconnecting can't help. Surface it
             // now (no rescue: a quota-blocked account can't create conversations).
             // On a quota exhaustion this mirrored 'error' status drives the main
--- a/desktop/windows/src/renderer/src/capture/liveRescue.test.ts
+++ b/desktop/windows/src/renderer/src/capture/liveRescue.test.ts
@@ -75,6 +75,22 @@ describe('isRetryableDropError', () => {
     expect(isRetryableDropError('Omi transcription unavailable (not signed in)')).toBe(false)
     expect(isRetryableDropError('Omi v4/listen requires sign-in.')).toBe(false)
   })
+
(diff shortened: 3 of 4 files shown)


Git-on-my-level:
Thanks — verified end to end and this looks right.

- `liveRescue.ts`: the new `PERMANENT_SOURCE_ERROR_NAMES` set covers exactly the getUserMedia/getDisplayMedia failures a /v4/listen reconnect can never fix, and threading the optional `name` param through `isRetryableDropError` keeps the message-only quota/sign-in checks intact.
- `liveMicSession.ts:280`: I traced the relay chain in main — `AudioSessionHost.ts` sends `audio-source-error` with `name: err.name`, and `omiListenClient.ts` copies `ev.name` onto the Error handed to `onError` — so `(e as Error).name` is genuinely populated here, which was the load-bearing assumption of the fix. This is also the only production call site, so the signature change breaks nothing.
- Skipping the from-segments rescue on this path is correct too: a source that never started has no retained segments to rescue.
- `liveRescue.test.ts`: good regression coverage — both the four terminal names and the AbortError/undefined retryable passthrough.
- Changelog entry matches the `unreleased/` schema.

One nuance for the record (not a defect): `NotAllowedError` is only permanent until the user grants the permission, but surfacing the error immediately instead of burning the ~4.5min backoff budget silently is still the right call — the user can restart capture after granting. And credit for explicitly deferring the generic recording-failure UI surface to a separate issue rather than bundling a product decision into this fix.

Leaving for human maintainer review before merge.

---

kodjima33 (approved):
Owner-override desktop/windows bug fix: scoped, tested, CI green, not already fixed on main. Merging.

PR #12086, opened 2026-08-23
fix(dev-harness): reject loopback-lookalike hostnames

arhxam:
## Summary

The dev harness accepted any hostname beginning with `127.` as loopback. Values such as `127.attacker.example:8085`, `127.0.0.1.example:8085`, and the invalid address `127.0.0.256:8085` therefore crossed a safety boundary intended to admit literal loopback endpoints only.

Parse literal IP addresses with `ipaddress.ip_address()` and accept them only when Python classifies them as loopback. `localhost`, IPv4 loopback addresses, IPv6 loopback addresses, and bracketed/ported forms keep working; hostname lookalikes now fail closed.

## Verification

- Red on the test-only commit: all three hostname lookalikes were accepted.
- Green focused suite: 11 passed, 3 skipped.
- Green full dev-harness suite: 96 passed, 7 skipped on macOS.
- Mutation proof restoring the prefix check reproduced the same three failures.
- Black, Ruff, `py_compile`, `git diff --check`, `make preflight`, and PR-body preflight pass.

none

Failure-Class: FC-implicit-resource-selection

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12086?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/dev-harness/dev_harness/safety.py
+++ b/scripts/dev-harness/dev_harness/safety.py
@@ -13,7 +13,6 @@ import json
 import os
 import re
 import signal
-import socket
 import subprocess
 import sys
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Scoped security fix to dev-only harness loopback check; adds regression tests for the hostname-lookalike bypass. 5/5 confidence, merging.

PR #12087, opened 2026-08-23
fix(dev-harness): reject non-executable Typesense overrides

arhxam:
## Summary

`OMI_TYPESENSE_SERVER_BIN` currently validates only that the override is a file. A readable but non-executable file is consequently selected as the native Typesense runtime and fails later at process startup with a low-level permission error.

Validate executable access at runtime selection and fail immediately with an actionable message. Missing-file behavior and valid executable overrides are unchanged, including portable coverage for platforms without POSIX execute bits.

## Verification

- Red on the test-only commit: a mode `0600` override reached command construction instead of failing.
- Green focused suite: 12 passed.
- Green full dev-harness suite: 95 passed, 6 skipped.
- Mutation proof removing the executable guard reproduced the focused failure.
- Black, `git diff --check`, `make preflight`, and PR-body preflight pass.

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12087?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/dev-harness/dev_harness/cli.py
+++ b/scripts/dev-harness/dev_harness/cli.py
@@ -99,8 +99,14 @@ def _service_record(cfg: config.HarnessConfig, service: str) -> dict[str, object
 def _native_typesense_binary() -> str | None:
     override = os.environ.get("OMI_TYPESENSE_SERVER_BIN", "").strip()
     if override:
-        if not Path(override).is_file():
+        binary = Path(override)
+        if not binary.is_file():
             raise SystemExit(f"OMI_TYPESENSE_SERVER_BIN points to a missing binary: {override}")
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Small dev-harness fix: validates executable bit on Typesense override before use, with new test coverage. 5/5 confidence, merging.

PR #12088, opened 2026-08-23
fix(dev-harness): honor explicit desktop profile environment

arhxam:
## Summary

`resolve_profile(..., env=...)` uses the supplied mapping for the app name but reads automation settings from ambient `os.environ`. The same explicit profile input can therefore produce different output—or silently pick up a developer shell's automation port—depending on unrelated process state.

Resolve every profile setting from one environment source. An explicitly supplied mapping is now authoritative, while callers that omit `env` retain the existing ambient-environment behavior.

## Verification

- Red on the test-only commit: explicit automation values lost to ambient values, and omitted values leaked ambient state.
- Green focused suite: 4 passed.
- Green full dev-harness suite: 96 passed, 6 skipped.
- Mutation proof restoring ambient reads reproduced both failures.
- Black, `py_compile`, `git diff --check`, `make preflight`, and PR-body preflight pass.

none

Failure-Class: FC-settings-mirror-retains-stale-override

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12088?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/dev-harness/dev_harness/desktop_profile.py
+++ b/scripts/dev-harness/dev_harness/desktop_profile.py
@@ -161,6 +161,7 @@ def resolve_profile(
     seeded_users: Iterable[str],
     env: Mapping[str, str] | None = None,
 ) -> DesktopLocalProfile:
+    source_env = os.environ if env is None else env
     users = tuple(sorted(set(str(item) for item in seeded_users)))
     payload = _user_payload_from_seed_manifest(cfg, user)
     email = payload.get("email", f"{user}@local.omi.invalid")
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Fixes dev-harness profile resolution to honor explicit env mapping instead of leaking ambient os.environ state; tests added. 5/5 confidence, merging.

PR #12090, opened 2026-08-23
fix(web): Enter mid-IME-composition must not send the chat message

### LINKED ISSUE #4936: Bug: Japanese IME — Enter key during kanji conversion triggers message send in Chat
## Bug Report

MACOS

**Description:**
In Chat with Omi, pressing Enter to confirm kanji conversion while typing Japanese causes the message to be sent immediately. The IME composition confirmation conflicts with the app's send trigger, resulting in frequent accidental message sends for Japanese users.

**Steps to reproduce:**
1. Open Chat with Omi
2. Switch to Japanese IME
3. Type Japanese text that requires kanji conversion
4. Press Enter to confirm the conversion
5. Message is sent immediately instead of just confirming the conversion

**Expected behavior:**
Enter during IME composition (conversion) should only confirm the conversion, not send the message. The message should only be sent when Enter is pressed *after* composition is complete.

**Suggested fix:**
- Do not trigger send when Enter is pressed during active IME composition ( /  events in web;  check in Flutter)
- Optionally, provide configurable send methods:
  - Enter = line break / Cmd+Enter = send
  - Enter = send / Shift+Enter = line break
  - For Japanese locale, default to "Enter = line break, Cmd+Enter = send"

**Impact:**
This makes Chat unusable for Japanese users — repeated mis-sends cause frustration and fear of typing, directly hurting UX for the entire Japanese user base.

**Platform:** iOS / Android (Flutter app)

---
*Reported by community via Discord.*

abunet:
Closes #4936.

Typing Japanese, Chinese or Korean, the first Enter confirms the conversion candidate the IME is offering; only a later Enter is meant for the composer. Both web chat composers read `e.key === 'Enter' && !e.shiftKey` and sent on both, so confirming a kanji conversion sent the half-written message.

macOS already draws this distinction — `ComposerKeyAction.resolve` reads `hasMarkedText` and returns `.newline` mid-composition (`desktop/macos/Desktop/Sources/Theme/OmiTextEditor.swift:49`). The web composer had no equivalent: `isComposing` appeared nowhere under `web/`. This ports the same rule to `ChatComposer` and `ChatPanel` through one shared predicate instead of a third copy of the condition.

The issue reports the platform as iOS/Android, but Flutter does not bind send to Enter and macOS already handles marked text; the web app is the surface that is actually broken.

## Source for the expected value

`KeyboardEvent.isComposing` is the standard signal for an in-flight IME composition (UI Events). Engines that predate it surface every composition keydown as `keyCode` 229 instead, so the predicate checks both.

## Verification

- `bash web/app/test.sh` — typecheck clean, 58 files / 329 tests (3 new) passing
- Reverting only the two component call sites on this branch turns the two composing cases red and leaves the other six green, so the test exercises the composer rather than asserting the predicate in isolation

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12090?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/app/src/components/chat/ChatComposer.tsx
+++ b/web/app/src/components/chat/ChatComposer.tsx
@@ -8,6 +8,7 @@ import { OmiPulseMark } from '@/components/ui/OmiPulseMark';
 import { uploadChatFiles } from '@/lib/api';
 import type { MessageFile } from '@/types/conversation';
 import { cn } from '@/lib/utils';
+import { shouldSubmitComposerKey } from '@/lib/chatComposerKey';
 
 /**
  * The ask bar: one pill that carries the text and every control that acts on
@@ -186,7 +187,7 @@ export function ChatComposer({
   };
 
   const handleKeyDown = (e: React.KeyboardEvent) => {
-    if (e.key === 'Enter' && !e.shiftKey) {
+    if (shouldSubmitComposerKey(e)) {
       e.preventDefault();
       void handleSend();
     }
--- a/web/app/src/components/chat/ChatPanel.tsx
+++ b/web/app/src/components/chat/ChatPanel.tsx
@@ -11,6 +11,7 @@ import { uploadChatFiles, getChatApps } from '@/lib/api';
 import type { App } from '@/lib/api';
 import { cn } from '@/lib/utils';
 import { MixpanelManager } from '@/lib/analytics/mixpanel';
+import { shouldSubmitComposerKey } from '@/lib/chatComposerKey';
 import { ChatMarkdown } from './ChatMarkdown';
 
 interface FilePreviewItem {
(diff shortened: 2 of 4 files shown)


kodjima33 (approved):
Fix is correct and well-tested: distinguishes IME composition Enter from send Enter via isComposing/keyCode 229, matches the existing macOS behavior. Confidence 5/5.

Issue #4801, opened 2026-02-13
Fix shared conversation and App store web page broken links and old product


hugoaap-code:
hello I noticed that links on both shared conversation web page (browser I mean) and App store web page are mostly broken (also mobile menu broken on app store page)

what could help us reach more people and get more users is driving users to broken Pages old pages 404 error pages and is promoting Omi Dev Kit (unavailable) to instead of the CV1

check screenshot with broken links

![Screenshot_20260213_082314_Chrome.jpg](https://github.com/user-attachments/assets/ab4bce24-ffd4-4fcf-b55a-c16a667b282d)

![Screenshot_20260213_082233_Chrome.jpg](https://github.com/user-attachments/assets/5aff3880-9025-41d2-97d5-b01a12b83283)

shared Omi convo: https://web.plaud.ai/s/pub_c7f65c64-8dd2-4b7c-9477-e208b92e1413::fGwMHjfCyYUwccXEnKQDQbaK6Usg4V_AQx4blxgmV_R3MY_7VYHpw1rig5R3BzUuMSxGU5-6TtTtYbgC

---

also we could make the shared conversation page similar to Plaud's regarding promoting Omi device more objectively (for now the page won't promote Omi at all)

 check it: https://web.plaud.ai/s/pub_c7f65c64-8dd2-4b7c-9477-e208b92e1413::fGwMHjfCyYUwccXEnKQDQbaK6Usg4V_AQx4blxgmV_R3MY_7VYHpw1rig5R3BzUuMSxGU5-6TtTtYbgC

hugoaap-code:
@kodjima33 I feel this would help make Omi app viral as well bro

haotvko:
Hi @josancamon19 @kodjima33 — I’m haotvko. I see that existing PRs addressed some app-store links and branding, so I’d like to offer the remaining public-page cleanup as a separate US$15 scope if it is still wanted.

I would audit the current shared-conversation and app-store pages against live routes, fix the remaining broken navigation and mobile-menu behavior, replace any unavailable Dev Kit calls to action with the current CV1 destination from the repository’s canonical product configuration, and add focused link/menu regression coverage. I would exclude links already corrected by #12092/#7654 and avoid duplicating those implementations. A short before/after route inventory would make the remaining scope reviewable.

Please confirm which public pages are still authoritative, whether this remaining scope and fee are acceptable, and the current cash-payout terms before work starts. No implementation, completed validation, assignment, or approved reward is claimed yet. Thank you.

PR #12092, opened 2026-08-23
fix(web): point the app-store page's store buttons at the real listings

abunet:
Part of #4801 — the two store buttons on the public app detail page.

Both hrefs are dead. The App Store one is the literal placeholder `id123456789`; the Play Store one names `com.omi.app`, which is not the package the app ships under. Both have been there since the frontend landed on 2025-03-24, and both 404 today:

| URL | status |
|---|---|
| `https://apps.apple.com/app/id123456789` | **404** |
| `https://play.google.com/store/apps/details?id=com.omi.app` | **404** |
| `https://apps.apple.com/us/app/friend-ai-wearable/id6502156163` | 301 |
| `https://play.google.com/store/apps/details?id=com.friend.ios` | 200 |

## Source for the expected values

Not guesses — the repo's own canonical values. The App Store URL appears 25 times elsewhere under `web/`, and `com.friend.ios` is both the Play URL used in 29 other places and the `applicationId` of the prod flavor in `app/android/app/build.gradle:63`.

## What this does not cover

The issue also reports a broken mobile menu and outdated product references. The product/pricing half is already fixed on `main` (`web/frontend/src/app/components/product-banner/types.ts` now reads `Omi` / `$89`), and the omi.me storefront pages are outside this repo. Only the store links are addressed here, so this does not close the issue.

## Verification

- Every store URL in the repo enumerated; after this change no `apps.apple.com` or `play.google.com` link under `web/` points anywhere but the canonical listings, with one deliberate exception: `apps.apple.com/us/app/lightblue-explorer/id557428110`, a third-party BLE tool referenced from the firmware docs.
- Status codes above measured directly.

No test accompanies this: `web/frontend` has no test lane, and two URL constants carry no behaviour to exercise. A repo-wide store-link guard is possible but wants a declared source of truth for the App Store numeric id, which the repo does not have today — the Play half is derivable from `applicationId`, the App Store half is not. Happy to add one if you want it.

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12092?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/frontend/src/app/apps/components/app-header/index.tsx
+++ b/web/frontend/src/app/apps/components/app-header/index.tsx
@@ -81,7 +81,7 @@ export function AppHeader({ plugin }: AppHeaderProps) {
           {/* Store Buttons */}
           <div className="mt-6 flex items-center gap-4">
             <a
-              href="https://apps.apple.com/app/id123456789"
+              href="https://apps.apple.com/us/app/friend-ai-wearable/id6502156163"
               target="_blank"
               rel="noopener noreferrer"
               className="flex h-10 items-center gap-2 rounded-lg bg-[#1A1F2E] px-4 py-2 text-sm font-medium text-white transition-all hover:bg-[#242938]"
@@ -96,7 +96,7 @@ export function AppHeader({ plugin }: AppHeaderProps) {
               <span>App Store</span>
             </a>
             <a
-              href="https://play.google.com/store/apps/details?id=com.omi.app"
+              href="https://play.google.com/store/apps/details?id=com.friend.ios"
               target="_blank"
               rel="noopener noreferrer"
               className="flex h-10 items-center gap-2 rounded-lg bg-[#1A1F2E] px-4 py-2 text-sm font-medium text-white transition-all hover:bg-[#242938]"


Git-on-my-level:
Verified end-to-end — thanks @abunet, this is exactly the right scope, and the writeup (measured status codes, repo-wide enumeration, the `applicationId` cross-check) made verification easy. Welcome as a first-time contributor.

What I checked on top of your measurements:

- `web/frontend/src/app/apps/components/app-header/index.tsx` — the only file touched, and the diff is exactly the two `href` swaps (lines 84 and 99); `target="_blank"` / `rel="noopener noreferrer"` are preserved and nothing else changed.
- Both destinations re-measured just now: the App Store and Play URLs return 200, the old two return 404. No affiliate/referral parameters on either new link. The App Store string is character-identical to the canonical constant in `web/frontend/src/app/apps/utils/metadata.ts` and the README badge, and `com.friend.ios` is the prod `applicationId` in `app/android/app/build.gradle` — so the buttons now point at the listings the rest of the repo already ships, not an alternative.
- Nice property worth noting: `id123456789` was the *last* occurrence of that placeholder anywhere in the repo (code search returns exactly this file), so once this lands it's gone entirely.
- The remaining `com.omi.app` hits (`backend/routers/auth.py`, `app/ios/Runner/AppDelegate.swift`, the Apple integrations) are OAuth redirect/bundle-id contexts, not store links — correctly left out of scope here.
- No test is fine for two bare URL constants in JSX; agreed that a repo-wide store-link guard only becomes worth it once there's a declared source of truth for the App Store numeric id.

Leaving for human maintainer review before merge — the one call that's genuinely theirs is confirming these are the listings you want prospective users landing on from the public app directory.

---

kodjima33 (approved):
Verified correct: replaced App Store ID resolves to the real Omi app (Based Hardware), and the Play Store package matches app/android/app/build.gradle's applicationId com.friend.ios. Trivial, scoped fix. Confidence 5/5.

PR #12095, opened 2026-08-23
fix(backend): back a failed deletion wipe off instead of re-selecting it every tick

abunet:
Follow-up 5 of #11680 — *"an attempt cap / backoff in `reconcile_pending_deletion_wipes` so one poisoned record can't loop forever"*. Only the backoff half; see below for what is deliberately left out.

`get_pending_deletion_wipes` age-filters every status it returns except `failed`, whose docstring called it "always actionable". So a wipe that fails for a persistent reason is re-selected by every reconciler tick on every pod, forever — `_periodic_deletion_wipe_reconcile` runs every 300s per pod, and the issue measured **30+ claim transactions per minute against a single uid, 24/7, for at least a month**, while the user who asked for deletion had been told `{"status": "ok"}`.

## What changed

- `mark_user_deletion_wipe_failed` counts the attempt with an `Increment`, so two pods failing the same wipe both count.
- `deletion_wipe_retry_delay` doubles per attempt from 5 minutes and saturates at 1 hour. The first failure still retries on the next tick, so a transient error costs nothing.
- The `failed` branch over-fetches and filters in Python, matching what the `pending` branch already does and for the reason its own comment gives: a tight `.limit()` could return a page made entirely of backed-off records and starve one that is ready.

No new Firestore index: the selection stays a single-field equality filter on `wipe_status`, with age filtering in Python, as the function's docstring requires.

## What this deliberately does not do

**No attempt cap.** Giving up on an accepted deletion is a data-retention decision, not a scheduling one — it belongs to whoever owns that policy, not to this PR. Backoff alone takes the worst case from every 300s to hourly without changing what eventually happens.

It also does not address follow-ups 1–4 of the issue (creating the queue, IaC, the startup probe that asserts the queue resolves, the backlog audit).

## Legacy principals

Two records must never become stuck, and both are covered by tests: one written before the counter existed has no `wipe_attempts` and counts as a first attempt; one with no `wipe_failed_at` stays immediately actionable. A missing field is not a reason to stop retrying a deletion.

## Verification

- new `tests/unit/test_deletion_wipe_retry_backoff.py` — 8 passed
- `tests/services/users/test_account_deletion.py` + `test_claim_deletion_wipe_txn.py` + `tests/routers/test_users.py` — 108 passed, unchanged
- reverting `database/users.py` to this branch's merge-base turns **4 of the 8 red**: the backoff table, the inside-window case, the starvation case and the attempt counter. The other 4 assert that a record *stays* actionable, which holds on both sides by design.
- the backoff table caught a real overflow while it was being written: `timedelta * 2 ** 48` raises `OverflowError`, so the exponent is clamped before it is applied.

## Line-count ratchet

The change lands in the module that already owns both the selector and the failure marker, so it grows an oversized file rather than splitting it; splitting `database/users.py` is a refactor of its own and does not belong in a fix.

Line-Count-Exception: backend/database/users.py | 2140 -> 2178 | backoff constants, one pure helper and the age filter belong beside get_pending_deletion_wipes and mark_user_deletion_wipe_failed; splitting this module is out of scope for a fix

none

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12095?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/users.py
+++ b/backend/database/users.py
@@ -33,6 +33,12 @@ import logging
 
 logger = logging.getLogger(__name__)
 DELETION_WIPE_RUNNING_STALE_AFTER = timedelta(hours=6)
+# A wipe that fails for a persistent reason (a missing queue, a dependency that is down) is
+# re-selected by every reconciler tick on every pod. Without a delay that is one claim
+# transaction per pod per tick, forever, against a record that cannot make progress. The
+# delay backs off per attempt and stops there: it never gives up on an accepted deletion.
+DELETION_WIPE_RETRY_BASE_DELAY = timedelta(minutes=5)
+DELETION_WIPE_RETRY_MAX_DELAY = timedelta(hours=1)
 _DELETION_WIPE_TERMINAL_STATUSES = frozenset({'completed', 'cancelled'})
 _DELETION_WIPE_LEGACY_ACTIONABLE_STATUSES = frozenset({'pending', 'retrying', 'running', 'failed'})
 LOCATION_CONTEXT_CONSENT_TTL = timedelta(days=30)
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Follow-up to #11680: adds per-attempt exponential backoff (5m, doubling, capped at 1h) for failed deletion wipes so a poisoned record isn't re-claimed every 300s tick forever. Extensive tests, scoped diff, CI green. Not yet fixed on main (docstring still says 'always actionable'). Confidence 5/5.

PR #12097, opened 2026-08-23
fix(stt): scope and bound speaker identity

Git-on-my-level:
## Summary

- give short, noisy in-session speaker clustering its own `0.60` threshold while leaving enrolled-voiceprint verification at `0.45`
- cap each clustering session at eight centroids and merge later misses into the nearest running centroid without dropping transcript text
- stamp provider/connection epochs before persistence and allocate conversation-local speaker IDs from `(speaker_id_scope, speaker)`, so padded labels, provider changes, and reconnects cannot collide
- persist `speaker_identity_status` as `user`, `not_user`, `no_match`, or `unknown`; a failed enrollment match is no longer represented only by the legacy `is_user=false` projection

## Failure boundary

The authoritative boundaries are now explicit: clustering policy owns in-session centroid identity, provider epochs own provider-local numbering, the conversation allocator owns persisted integer IDs, and enrollment matching owns the evidence state behind `is_user`. The Parakeet runtime image cannot import backend utility modules through its deployment source closure, so its existing `speaker_math.py` boundary mirrors the clustering defaults and bounded-nearest policy; tests pin both implementations.

The memory consumer is intentionally outside this branch. It still reads a bare `is_user` boolean today; the parallel memory change must treat `speaker_identity_status` values `no_match` and `unknown` as uncertain rather than confidently third-party.

## Verified diagnosis and corrections

- The epoch stamp initially overwrote each segment's `stt_provider` with the host's configured STT selection. A socket that knows which peer actually served the segment stamps that name itself, and the configured selection does not change across a fallback, so the overwrite erased exactly the transition the epoch exists to separate. Five hermetic e2e tests caught it (`parakeet` observed where the peer was `parakeet-wire-peer`; `parakeet` where the fake socket had stamped `e2e-streaming-stt`). A segment's own provider is now authoritative, the caller's name is a fallback, and the epoch advances per segment so a provider change inside one batch opens a new scope. This is also the correction to the bullet below: normal backend STT segments *do* reach `enqueue` carrying a provider on the paths the hermetic harness exercises.
- The `0.45` enrollment threshold was also reused by `backend/utils/stt/pre_recorded.py` and `backend/parakeet/transcribe.py`, not only the two streaming paths named in the report.
- `connect_stt_socket_with_fallback` changes provider only while establishing a socket. A live provider failure terminates that connection; a resumed conversation can still acquire a colliding namespace on its next WebSocket connection, so the epoch is connection-scoped as well as provider-scoped.
- Normal backend STT segments previously did not persist `stt_provider`; only the custom-STT payload path stamped it.
- Ordinary processed short-term memories expire after 48 hours, but pending-required-processing rows have an explicit eligibility exception. This does not remove the identity-loss path for ordinary rows.

## External evidence for changed expectations

The account measurement supplied with this task covered 57 conversations, 4,047 segments, and 200 memories. It also identified conversation `30e9e5d6-e7fd-4f9c-91ad-f219e5b1952f`, where unpadded and padded provider labels coexist and one integer speaker identity carries contradictory `is_user` values. Those observed persisted shapes are the external basis for the new collision and reconnect regression expectations.

## Verification

- red-before runs: the new model test initially failed at collection because `SpeakerIdentityStatus` did not exist; the remaining regression files reported `4 failed, 63 passed` (explicit scoped ID overwritten, reconnect segments merged, and both backend and deployed-Parakeet clustering created a ninth speaker)
- affected and adjacent test files, each named directly and run in isolation: `436 passed`
- final persistence-only schema rerun: `39 passed` across `test_transcript_segment.py` and `test_transcribe_speaker_id_queue_pii.py`
- proper backend typecheck: `0 errors, 3752 warnings, 0 informations`
- `python -m compileall` for the changed backend modules: passed
- `speaker_id_scope` and `speaker_identity_status` remain persistence-only backend fields: `model_dump` includes them, while `SkipJsonSchema` keeps the app-client OpenAPI and all generated client contracts unchanged
- final pre-push gate: 24 PR checks passed, runtime-image closure passed for 10 images, eight selected backend files passed `105/105`, app-client OpenAPI and generated-client snapshots were current, and Black reported 20 files unchanged
- broad `backend/test.sh`: 890 of 892 isolated test files wrote successful status and none wrote a failure; the run was stopped after the two unrelated action-item files did not return. `test_action_item_canonical_contract.py` had already printed `13 passed`; `test_action_item_vector_best_effort.py` had reached 75%. This is not claimed as a completed full-suite pass.
- combined direct pytest runs exposed known cross-file pollution: the Parakeet builtin-embedding file raised a missing-NeMo collection error only in combination and passed `16/16` alone; four Modulate retired-config failures passed `94/94` alone. Two real parity fixture mismatches were fixed and then passed `2/2` alone.
- full hermetic e2e suite (`bash backend/testing/e2e/run.sh`): `116 passed, 3 skipped` after the provider-preservation fix; before it, 5 failed
- eight affected unit files run together in one process (not file-isolated): `105 passed`
- live provider/embedding services were not exercised because this worktree has no hermetic owner-account or provider-credential path; the production assignment and persistence seams were exercised with controllable embeddings and transcript payloads.

none

## Failure class (fixes)

Failure-Class: none

Line-Count-Exception: backend/utils/stt/streaming.py | 1683 -> 1692 | Cap-forced merge fallback telemetry and centroid-drift guard required by docs/agents/fallback-telemetry.md; the +9 lines are the record_fallback branch that makes the eight-speaker cap observable.

### DIFF
--- a/backend/models/transcript_segment.py
+++ b/backend/models/transcript_segment.py
@@ -1,8 +1,10 @@
 from datetime import timedelta
+from enum import Enum
 from typing import Any, Optional, List, Tuple, cast
 import uuid
 import re
 from pydantic import BaseModel, Field
+from pydantic.json_schema import SkipJsonSchema
 
 from models.other import Person
 
@@ -23,6 +25,15 @@ class Translation(BaseModel):
     text: str
 
 
+class SpeakerIdentityStatus(str, Enum):
+    """Evidence state behind the legacy boolean ``is_user`` projection."""
+
+    unknown = 'unknown'
+    user = 'user'
+    not_user = 'not_user'
+    no_match = 'no_match'
+
+
 class TranscriptSegment(BaseModel):
     id: Optional[str] = None
     text: str
@@ -35,12 +46,21 @@ class TranscriptSegment(BaseModel):
     translations: Optional[List[Translation]] = Field(default_factory=list)
     speech_profile_processed: bool = True
     stt_provider: Optional[str] = None
+    # These fields are persisted for backend identity consumers, but are not part
+    # of the first-party app wire contract. SkipJsonSchema leaves validation and
+    # model_dump intact while avoiding generated-client fan-out for internal state.
+    speaker_id_scope: SkipJsonSchema[Optional[str]] = None
+    speaker_identity_status: SkipJsonSchema[str] = SpeakerIdentityStatus.unknown
 
     def __init__(self, **data: Any):
+        if 'speaker_identity_status' not in data and data.get('is_user') is True:
+            data['speaker_identity_status'] = SpeakerIdentityStatus.user
         super().__init__(**data)
         if not self.id:
             self.id = str(uuid.uuid4())
 
+        if self.speaker_id is not None:
+            return
         if self.speaker:
             try:
                 self.speaker_id = int(self.speaker.split('_', 1)[1])
@@ -167,6 +187,8 @@ class TranscriptSegment(BaseModel):
             if not a or not b:
                 return a, b
             if b.stt_provider != a.stt_provider:
+                return a, b
+            if b.speaker_id_scope != a.speaker_id_scope:
                 return a, b
 
             if a.speaker != b.speaker and not (a.is_user and b.is_user) and a.text and b.text:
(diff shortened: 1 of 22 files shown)


kodjima33 (approved):
Substantial speaker-identity/diarization rework (925+/66-, 19 files) -- well tested but reads as a hardening/refactor rather than a narrow bug fix, well over the scoped-diff bar. Approving only, not merging; flagging for a manual look given the size and scope.

PR #12100, opened 2026-08-23
feat(backend): export VAD gate audio-seconds and session counters to Prometheus

Git-on-my-level:
## What

Adds two Prometheus counters so the live VAD streaming gate's effect is observable from Grafana instead of only per-session JSON log lines:

- `omi_vad_gate_audio_seconds_total{outcome=sent|skipped, mode=active|shadow}` — audio seconds forwarded to vs withheld from the STT provider
- `omi_vad_gate_sessions_total{mode}` — gated sessions

Increments are per-chunk (crash-safe, no teardown-only lag). Pre-roll audio is only counted once its final outcome is known, so buffered silence is never double-counted as skipped-then-sent. Gate behavior, thresholds, and the existing JSON log are unchanged. Gate failures already record via `OMI_FALLBACK_TOTAL{component=vad}`, so no new fallback counter.

## Why

Prod runs `VAD_GATE_MODE=active`, but the gate's actual rejection rate ("what fraction of audio never reaches Deepgram") currently requires log spelunking. These counters answer that directly and cheaply.

## Tests

`tests/unit/test_vad_gate.py`: new parametrized test asserting counter deltas match the gate's byte accounting in both modes. Full file: 121 passed.

Implementation by codex (gpt-5.6-sol), reviewed and verified by Claude.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_vad_gate.py
+++ b/backend/tests/unit/test_vad_gate.py
@@ -26,6 +26,8 @@ _mock_vad_prob = None  # When set, overrides _mock_is_speech for exact probabili
 
 import numpy as np
 
+from utils.metrics import OMI_VAD_GATE_AUDIO_SECONDS_TOTAL, OMI_VAD_GATE_SESSIONS_TOTAL
+
 
 def _mock_run_vad_window(window, state, context):
     """Mock run_vad_window that returns controlled probability.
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
feat(backend): adds Prometheus counters for VAD gate audio-seconds/sessions. Clean, well-tested observability addition, but it's a feature not a bug fix, so approve-only per the confidence gate.

PR #12102, opened 2026-08-23
fix(firmware): log and count offline-audio drops while RTC is unsynced

Git-on-my-level:
## What

A pendant whose clock has never been synced by the app (out-of-box before first pairing, or after a settings wipe) silently discards all offline SD audio: `process_write_data_req()` returns early on `!rtc_is_valid()` with zero trace. This PR makes that loss visible:

- Counts dropped packets and emits a rate-limited `LOG_WRN` (at 1, 2, 4, 8, … drops) naming the cause.
- Implements the previously no-op `sd_notify_time_synced()` hook: on time sync it enqueues a `REQ_TIME_SYNCED` message to the SD worker, which logs a one-line summary of how many packets were discarded pre-sync and resets the counter.
- Hoists the magic `1700000000U` timestamp floor into `RAW_MIN_VALID_UTC_TIME`.

Deliberately out of scope: recording with synthetic timestamps (would propagate bogus conversation dates through phone-side sync) and any BLE protocol change. Observability only.

All counter reads/writes stay on the SD worker thread (drops inside `process_write_data_req`, reset via the priority msgq), so no locking is needed.

Implementation by codex (gpt-5.6-sol), reviewed and verified by Claude.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Failure class (fixes)

Failure-Class: none

No existing failure-class covers firmware-side silent data discard (registry has no firmware/RTC/SD-surface classes); this is a single-instance observability fix, and minting a new class here would expand the PR beyond its stated firmware-only scope.

### DIFF
--- a/omi/firmware/omi/src/sd_card.c
+++ b/omi/firmware/omi/src/sd_card.c
@@ -1227,8 +1227,11 @@ void sd_notify_time_synced(uint32_t utc_time)
         return;
     }
 
+    /* Callers run on the BLE RX thread (transport.c) and the system workqueue
+     * (rtc.c); never block them for a best-effort log summary. Matches the
+     * K_NO_WAIT pattern of sd_notify_ble_state below. */
     req.type = REQ_TIME_SYNCED;
-    ret = k_msgq_put(&sd_prio_msgq, &req, K_MSEC(500));
+    ret = k_msgq_put(&sd_prio_msgq, &req, K_NO_WAIT);
     if (ret != 0) {
         LOG_WRN("failed to queue RTC sync drop summary: %d", ret);
     }
--- a/omi/firmware/omi/src/sd_card.c
+++ b/omi/firmware/omi/src/sd_card.c
@@ -30,6 +30,7 @@ LOG_MODULE_REGISTER(sd_card, CONFIG_LOG_DEFAULT_LEVEL);
 #define RAW_BATCH_HEADER_BYTES 32U
 #define RAW_PACKETS_PER_BATCH ((RAW_BATCH_BYTES - RAW_BATCH_HEADER_BYTES) / RAW_AUDIO_PACKET_BYTES)
 #define RAW_FLUSH_INTERVAL_MS 1000
+#define RAW_MIN_VALID_UTC_TIME 1700000000U
 
 #define RAW_META_MAGIC 0x4F4D4952U
 #define RAW_BATCH_MAGIC 0x4F4D4942U
@@ -108,6 +109,7 @@ static int wait_for_sd_worker_response(struct k_sem *sem, int timeout_ms, const 
 
 typedef enum {
     REQ_WRITE_DATA,
+    REQ_TIME_SYNCED,
     REQ_GET_RING_INFO,
     REQ_READ_PACKETS,
     REQ_ADVANCE_READ,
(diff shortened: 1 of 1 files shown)


Git-on-my-level:
**Adversarial review swarm (4× glm-5.3: concurrency / build / field-failure / kconfig lenses) — results:**

All reviewers: SHIP, no blockers. Findings acted on or noted:

- **Fixed in `77cde506ab`:** the `REQ_TIME_SYNCED` enqueue used `k_msgq_put(..., K_MSEC(500))`, but the hook's callers are the BLE GATT RX thread (`transport.c:264`) and the system workqueue (`rtc.c:32`) — a full priority queue during an SD stall would have parked either for 500ms for the sake of a best-effort log line. Now `K_NO_WAIT`, matching `sd_notify_ble_state`. (Flagged independently by two reviewers.)
- **Known limitation (deliberate, worth stating):** production builds set `CONFIG_LOG=n` (`omi.conf:168`), so the added warnings are compiled out on fleet devices — this PR gives observability on debug/bench builds and CI, not OTA telemetry. Surfacing the drop counter to the app over BLE was deliberately kept out of scope; it can be a follow-up if fleet visibility is wanted.
- **Minor edge (accepted):** a phone syncing a pre-2023 epoch keeps `rtc_is_valid()` true while writes fail the `RAW_MIN_VALID_UTC_TIME` floor; the drop warning then says "never synced" and the summary/reset is suppressed by the same floor guard. Requires a badly misconfigured phone clock; self-limiting via the power-of-two rate limit.
- Verified clean by the swarm: enum renumbering safe (no `default:` cases, type never persisted/serialized), counter strictly single-writer on the SD worker, no init-order hazard (`K_MSGQ_DEFINE` is link-time; boot RTC restore doesn't fire the hook), watchdog unaffected, clang-format/CI lint pass, devkit app unaffected.

Git-on-my-level:
**Monitor fix — Hygiene `failure-class-protocol` failure resolved (no code changes):**

- **Root cause:** both commits use `fix(firmware):` subjects but the PR body carried no `Failure-Class:` declaration, so the `failure-class-protocol` check failed with `missing_declaration`.
- **Fix:** added a `## Failure class (fixes)` section declaring `Failure-Class: none` to the PR body, then re-ran the failed Hygiene job → **success**. Verified locally with `scripts/failure-class validate` before editing (old body FAIL / new body passes the declaration check), and re-read the body from the API after the edit.
- **Why `none`:** the registry (139 classes at this head) has no firmware/RTC/SD-surface or silent-discard class; this PR is deliberately scoped as observability-only, and minting a `new` class here would expand it into a registry-lifecycle change, which belongs in a separate PR per AGENTS.md.
- **State:** no failing or pending checks remain on `77cde506`; the PR is now blocked only on required review. No review threads were open, so none were resolved or rejected.

Git-on-my-level:
@codex review

Git-on-my-level:
Need to make this PR useful by shipping from device -> app -> cloud first before merge

PR #12116, opened 2026-08-23
fix(app): stamp Gemini Live transcript segments with real audio time

igor-popov-dev:
## What changed and why

`GeminiStreamingSttSocket` gave every `inputTranscription` delta a fixed 3-second span and
advanced its own clock by 3 seconds per delta. Gemini Live streams partial transcription deltas
as the user speaks, so the delta count has no relation to elapsed audio: ten deltas describing
three seconds of speech were recorded as spanning thirty seconds, and the drift accumulates for
the whole session.

A new `SttAudioTimeline` counts the PCM actually forwarded to the provider and hands each delta
the audio consumed since the previous one, so deltas tile the real timeline instead of inventing
time. A delta that arrives before any new audio (a correction of already transcribed speech)
collapses onto the previous end rather than adding another 3 seconds. `stop()` resets the
timeline along with the rest of the session state, as before.

These timestamps are not cosmetic — the backend takes them verbatim in
`routers/listen/transcripts.py`:

- the first segment's `start` sets the conversation's `started_at`;
- `sort_segments_by_start` orders the buffer on them;
- `TranscriptSegment.combine_segments` merges on them;
- speaker identification reads the audio ring buffer at
  `first_audio_byte_timestamp + segment.start`.

Every other STT path in the app already reports real time: the on-device Whisper provider stamps
`start + duration`, and schema-based streaming providers pass through the provider's own word
timings (`audioOffsetSeconds: 0`). This brings the Gemini Live path in line with them rather
than introducing a new convention.

Known approximation, stated plainly: transcription arrives after the audio it describes, so a
delta's end includes audio that has been sent but not yet transcribed. That error is bounded by
provider latency instead of growing without limit like the fixed 3-second stride did.

## How it was verified

- `flutter test test/unit/stt_audio_timeline_test.dart` — 7/7 passing.
- Together with the neighbouring STT suites (`stt_result_test.dart`, `custom_stt_config_test.dart`,
  `composite_transcription_socket_test.dart`) — 22/22 passing.
- `dart format --line-length 120 --set-exit-if-changed` — clean.
- `flutter analyze` on both changed files — no issues.

## Tests

`app/test/unit/stt_audio_timeline_test.dart` covers: bounds derived from the audio actually sent;
ten deltas over three seconds of speech staying at 3.0s (the old stride would have ended at
30.0s); deltas tiling the timeline without gaps or overlaps; a delta before any new audio not
inventing time; zero/negative byte counts and a zero sample rate not corrupting the count;
`reset()`; a non-default 24 kHz sample rate.

To be explicit about what these tests prove: they guard the new `SttAudioTimeline` invariants.
They are not red-before-green regression tests, because the old arithmetic lived inside a private
method of the socket class and is not reachable from a test. The regression itself is stated
numerically above.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12116?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/services/sockets/pure_streaming_stt.dart
+++ b/app/lib/services/sockets/pure_streaming_stt.dart
@@ -63,6 +63,60 @@ class StreamingSttConfig {
   }
 }
 
+/// Bounds of one transcription delta on the audio timeline, in seconds.
+class SttSegmentBounds {
+  final double start;
+  final double end;
+
+  const SttSegmentBounds(this.start, this.end);
+}
+
+/// Timeline of the audio a streaming provider has actually been given.
+///
+/// Providers that stream incremental transcription deltas (Gemini Live's
+/// `inputTranscription`) emit as many deltas as they like per utterance, so the
+/// delta count says nothing about elapsed audio. Stamping deltas with a fixed
+/// stride therefore drifts arbitrarily far ahead of the recording. Counting the
+/// PCM actually sent keeps every delta anchored to the audio it came from.
+class SttAudioTimeline {
+  final int sampleRate;
+  final int bytesPerSample;
+
+  double _audioSeconds = 0;
+  double _lastSegmentEnd = 0;
+
+  SttAudioTimeline({required this.sampleRate, this.bytesPerSample = 2});
+
+  /// Seconds of audio handed to the provider so far.
+  double get audioSeconds => _audioSeconds;
+
+  /// Account for PCM forwarded to the provider.
+  void addPcm(int byteLength) {
+    if (byteLength <= 0 || sampleRate <= 0 || bytesPerSample <= 0) {
+      return;
+    }
+    _audioSeconds += byteLength / (sampleRate * bytesPerSample);
+  }
+
+  /// Bounds for the next delta: the audio consumed since the previous one.
+  ///
+  /// A delta that arrives before any new audio (a correction of what was
+  /// already transcribed) collapses onto the previous end instead of inventing
+  /// time, so the timeline never runs ahead of the recording.
+  SttSegmentBounds nextSegment() {
+    final start = _lastSegmentEnd;
+    final end = _audioSeconds > start ? _audioSeconds : start;
+    _lastSegmentEnd = end;
+    return SttSegmentBounds(start, end);
+  }
+
+  /// Drop back to an empty timeline for a fresh session.
+  void reset() {
+    _audioSeconds = 0;
+    _lastSegmentEnd = 0;
+  }
+}
+
 /// Gemini Live streaming socket with setup message and base64 audio encoding
 class GeminiStreamingSttSocket implements IPureSocket {
   WebSocketChannel? _channel;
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @igor-popov-dev — this is a clean fix for a real drift bug, and the PR description's invariant analysis checks out against the backend.

What I verified in `app/lib/services/sockets/pure_streaming_stt.dart`:

- The old `_audioOffsetSeconds += 3.0` stride is fully gone; `SttAudioTimeline` counts the PCM actually handed to the provider. `addPcm` is called after a successful `sink.add` at both the buffered send and the `disconnect()` tail flush, with `_frameBuffer` cleared before each, so every chunk is counted exactly once and failed sends are not counted at all.
- `late final _timeline = SttAudioTimeline(sampleRate: sampleRate)` correctly picks up the configured rate (16 kHz default), consistent with the `audio/pcm;rate=$sampleRate` chunk tags and the setup message — so bytes-to-seconds conversion is internally consistent.
- The correction case is handled well: a delta arriving with no new audio collapses to a zero-width segment at the previous end rather than inventing time, which keeps `sort_segments_by_start` / `combine_segments` on the backend well-behaved (no phantom ordering or merges).
- `stop()` resets the timeline after `disconnect()`'s flush, so the tail audio is still accounted before the reset — matches the previous `_audioOffsetSeconds = 0` placement.
- The stated backend impact is accurate: `backend/routers/listen/transcripts.py` takes these timestamps verbatim for `started_at` (from the first segment's `start`), buffer ordering, segment merging, and speaker-identification ring reads (`first_audio_byte_timestamp + segment.start`). The old stride inflated all four; ten deltas over three seconds of speech used to end an utterance at thirty.

`app/test/unit/stt_audio_timeline_test.dart` covers the right behaviors — tiling without gaps/overlaps, the multi-delta cap, the correction collapse, malformed/zero-length accounting, reset, and a non-default sample rate.

One non-blocking observation for the record: a re-stamped partial still inherits a later window than the speech it re-describes (transcription lags the audio), so segment ends skew late by provider latency. That error is bounded, as the PR states, and strictly better than unbounded drift. If Gemini ever exposes word timings in `inputTranscription`, passing those through would be the natural refinement.

Leaving the merge call to the human maintainers — from automation this is a clear positive signal.

---

kodjima33 (approved):
Fixes Gemini Live transcript segment drift: replaces a fixed 3s-per-delta stride with a real PCM-based audio timeline, so delta count no longer determines elapsed time. Well tested, scoped, CI green, verified not fixed on main. Confidence 5/5 (note: touches STT segment-timing code, not word content -- flagging per Nik's general STT-merge caution in case he wants a look).

PR #12117, opened 2026-08-23
fix(backend): declare the missing composite index for the default conversation list read

igor-popov-dev:
## What changed and why

`get_conversations` and `get_conversations_without_photos` (both in
`backend/database/conversations.py`), called with their defaults (`include_discarded=False`,
no `statuses`/`categories`/`folder_id`/`starred` filter — i.e. the app's default
`GET /v1/conversations` list call, the main screen), filter the `conversations` collection by
`discarded` and order by `created_at` descending, but `firestore_index_registry.py` never
declared the composite Firestore needs for that shape. Production has it only because it was
created by hand at some point; a fresh self-host deploy gets `FailedPrecondition` 400 on
`GET /v1/conversations` the first time anyone loads their conversation list.

Adds `CONVERSATIONS_ACTIVE_ORDERED_QUERY` to the registry (mirrors the already-registered
`chat_sessions_current_by_app_created_at` / `messages_by_app_ordered` shape) and regenerates
`firestore.indexes.json`.

none

## How it was verified

  `firestore.indexes.json` from the updated registry; diff is exactly the new
  `conversations(discarded ASC, created_at DESC, __name__ DESC)` composite.
- `.venv/bin/python -m pytest backend/tests/unit/test_firestore_query_contract.py -m ""` —
  34/34 passed (31 baseline, now including the already-merged `messages_by_app_ordered`
  case from #12069, + 3 new: two parametrized regression cases plus the spec-is-registered
  sanity check for conversations).
- `.venv/bin/python -m pytest backend/tests/unit/test_firestore_query_contract.py
  backend/tests/unit/test_reconcile_firestore_indexes.py
  backend/tests/unit/test_universal_memory_list_cursor.py
  backend/tests/unit/test_action_items_due_range_index_contract.py
  backend/tests/unit/test_chat_message_count.py
  backend/tests/unit/test_chat_messages_pagination.py -m ""` — 109/109 passed
  (34 + 75), no regression in the surrounding index-reconcile/contract suite.

## Tests

Bug fix — regression test added:
`test_default_conversation_list_reads_have_a_declared_composite_index` (parametrized over
`get_conversations` and `get_conversations_without_photos`) in
`backend/tests/unit/test_firestore_query_contract.py` calls each against a recording
Firestore fake, extracts the real filter+order chain it builds, and asserts
`firebase_index_manifest()` declares a matching composite — so either function regressing to
an undeclared shape fails this test, not just self-host in production. Also added
`test_conversations_active_ordered_query_is_registered_for_the_conversations_collection` as a
direct sanity check on the new spec.

## Failure class (fixes)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/firestore_index_registry.py
+++ b/backend/database/firestore_index_registry.py
@@ -618,6 +618,21 @@ STALE_IN_PROGRESS_CONVERSATIONS_QUERY = FirestoreQuerySpec(
     ),
 )
 
+CONVERSATIONS_ACTIVE_ORDERED_QUERY = FirestoreQuerySpec(
+    identifier='conversations_discarded_created',
+    collection_group='conversations',
+    query_scope='COLLECTION',
+    # `get_conversations`/`get_conversations_without_photos` called with their
+    # defaults (`include_discarded=False`, no `statuses`/`categories`/`folder_id`/
+    # `starred` filter) build exactly this shape — it's the default `GET
+    # /v1/conversations` list call, i.e. the app's main screen. Production has
+    # this index only because it was created by hand at some point; a fresh
+    # self-host deploy 400s with FailedPrecondition the first time anyone loads
+    # their conversation list.
+    filters=(FirestoreQueryFilter('discarded', '==', 'discarded'),),
+    index_fields=(_asc('discarded'), _desc('created_at'), _desc('__name__')),
+)
+
 ACTION_ITEMS_COMPLETION_ID_SCAN_QUERY = FirestoreQuerySpec(
     identifier='action_items_completion_id_scan',
     collection_group='action_items',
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
Backend bug fix: declares missing Firestore composite index for the default conversation list read (self-host FailedPrecondition 400 fix). Scoped, tested, verified not already on main. 5/5.

PR #12118, opened 2026-08-23
fix(backend): keep mentor past context when one source fails

igor-popov-dev:
## What changed and why

The mentor's proactive pipeline gathers past conversations from two independent sources —
semantic search (needs an embedding provider and a vector store) and the most recent
conversations by time (needs neither) — and then renders whatever it collected into the
generate prompt. All three steps shared a single `try/except`, so a failure in any one of them
discarded the work of the others:

- an embedding error (missing key, quota, provider outage) jumped over the recent-conversations
  fetch that follows it, and
- a conversations-store error in that fetch threw away the vector hits already collected.

Either way `past_conversations_str` stays empty and the notification is drafted from the live
transcript alone, silently: the pipeline keeps going, and nothing downstream can tell "no past
context exists" apart from "past context was lost". A missing embedding provider is enough to
put the mentor permanently in that state, which is how we hit it.

This guards the two sources and the rendering separately, each with its own log line. That is
already the convention everywhere else in this function — `get_prompt_memories`,
`get_user_goals`, `get_app_messages` and the language preference lookup each sit in their own
`try/except` with their own fallback; the past-conversations block was the one exception.

No behaviour changes on the happy path: same sources, same order, same rendering.

none

## How it was verified

- `.venv/bin/python -m pytest backend/tests/unit/test_mentor_notifications.py -m ""` —
  49 passed (47 before this change; +2 new).
- Both new tests were confirmed to fail on unmodified `main` (`AssertionError: Expected 'mock'
  to have been called` — the rendering never runs) and to pass with the fix, so they are
  regression proofs rather than restatements of current behaviour.
- `.venv/bin/python -m pytest backend/tests/unit/test_app_integrations_async.py
  backend/tests/unit/test_app_integrations_safe_messages.py
  backend/tests/unit/test_async_app_integrations.py
  backend/tests/unit/test_proactive_notification_language.py
  backend/tests/unit/test_mem_db_proactive_noti_cap.py
  backend/tests/unit/test_chat_first_proactive_engine.py -m ""` — identical results before and
  after the change (39 passed, 14 pre-existing failures on unmodified `main` in
  `test_async_app_integrations.py`, unrelated to this file).
- `black --check` on both changed files — clean.

## Tests

Two tests in `backend/tests/unit/test_mentor_notifications.py`, one per direction of the loss:

- `test_mentor_past_context_survives_embedding_failure` — `generate_embedding` raises; asserts
  the recent-conversations fetch still runs and its result still reaches the prompt renderer.
- `test_mentor_vector_context_survives_recent_conversations_failure` — the mirror case: the
  vector search returns a hit, the recent-by-time fetch raises; asserts the vector hit is still
  the payload handed to the renderer.

Both use the file's existing fixtures (`_setup_app_integrations_stubs`, the shared runtime
mocks) and add a small `_pass_all_three_steps()` helper that makes the gate, generate and critic
LLM calls approve, so the pipeline reaches the send path.

## Failure class (fixes)

Failure-Class: none
```


<!-- This is an auto-generated description by cubic. -->
<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>
<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/backend/tests/unit/test_mentor_notifications.py
+++ b/backend/tests/unit/test_mentor_notifications.py
@@ -688,6 +688,99 @@ def test_process_mentor_proactive_notification_sends():
     mock_send.assert_called()
 
 
+def _pass_all_three_steps():
+    """Configure mock_llm_mini so gate, generate and critic all approve."""
+    results = [
+        RelevanceResult(
+            is_relevant=True,
+            relevance_score=0.85,
+            reasoning="User is skipping gym despite their 3x/week goal.",
+            context_summary="User discussing skipping exercise.",
+        ),
+        NotificationDraft(
+            notification_text="You've been skipping gym — remember your 3x/week goal!",
+            reasoning="User's goal is 'Exercise 3x per week' and they mentioned skipping today.",
+            confidence=0.82,
+            category="goal_connection",
+        ),
+        ValidationResult(approved=True, reasoning="Concrete, tied to a goal and the current action."),
+    ]
+    call_count = [0]
+
+    def side_effect_structured_output(model_class):
+        parser = MagicMock()
+        parser.invoke = MagicMock(return_value=results[min(call_count[0], len(results) - 1)])
+        call_count[0] += 1
+        return parser
+
+    mock_llm_mini.with_structured_output = MagicMock(side_effect=side_effect_structured_output)
+
+
+def test_mentor_past_context_survives_embedding_failure():
+    """A failing embedding provider must not also drop the recent-conversations context.
+
+    Semantic search and recent-by-time are separate sources: the first needs an embedding
+    provider and a vector store, the second needs neither. When they shared one try/except,
+    one embedding error (missing key, quota, outage) silently stripped both.
+    """
+    _setup_app_integrations_stubs()
+    _pass_all_three_steps()
+
+    mock_generate_embedding.side_effect = RuntimeError("no embedding provider configured")
+    mock_get_convos.reset_mock()
+    mock_get_convos.return_value = [{'id': 'conv-1', 'is_locked': False}]
+    mock_convos_to_string.reset_mock()
+    mock_convos_to_string.return_value = 'yesterday: user talked about the gym'
+
+    try:
+        result = app_int._process_mentor_proactive_notification(
+            "uid_embed_fail", [{"text": "I'll skip the gym today", "is_user": True}]
+        )
+    finally:
+        mock_generate_embedding.side_effect = None
+        mock_get_convos.return_value = []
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Thanks @igor-popov-dev — well-scoped fix, and the analysis in the description matches what the code actually does.

**backend/utils/app_integrations.py** — splitting the past-conversations gathering into two independent `try/except`s (plus a separate render guard) is the right shape. `all_past` is now initialized before both sources, so partial results survive whichever source fails, and the happy path is untouched: same source order, same `existing_ids` dedup, same `is_locked` filtering, same top-5 slice. This also brings the block in line with the per-source guarding already used for `get_prompt_memories`, `get_user_goals`, `get_app_messages` and the language lookup in the same function. Renaming the old `past_conversations_failed` log event to `past_conversations_render_failed` keeps the three failure modes distinguishable in logs.

**backend/tests/unit/test_mentor_notifications.py** — both new tests are regression proofs rather than restatements of current behavior. `test_mentor_past_context_survives_embedding_failure` makes `generate_embedding` raise and asserts the recent-conversations fetch still ran and its result reached the renderer (`mock_get_convos` / `mock_convos_to_string` both called); `test_mentor_vector_context_survives_recent_conversations_failure` covers the mirror case and asserts the exact list handed to `deserialize_conversations` (`[{'id': 'conv-vector-1', 'is_locked': False}]`), which also pins the locked-conversation filtering. Restoring the shared mock state in `finally` keeps the rest of the suite isolated. Both directions fail on `main` given the previous single shared `try/except` — verified against the pre-change structure.

CI at this head (Backend unit suite, Backend Hermetic Merge Gate, Hermetic Backend E2E) is green. One non-blocking note: when *both* sources fail, the mentor still drafts from the live transcript alone — that is the intended graceful degradation and matches prior render-step behavior, just now visible in per-source log lines.

Leaving the merge call to the human maintainers — from automation this is a clear positive signal.

---

kodjima33 (approved):
Backend bug fix: separates mentor past-context sources into independent try/except so one failure (e.g. embedding provider) doesn't silently drop the other. Regression tests confirmed failing on main pre-fix. 5/5.

PR #12085, opened 2026-08-23
fix(ci): use authoritative live PR body on main pushes

arhxam:
## Summary

Main-push Repo Checks currently append a recovered live PR body to the merge commit message. GitHub hard-wraps that embedded body, so line-oriented declarations such as `Line-Count-Exception` remain malformed even though the authoritative live PR body is valid.

Use the recovered live PR body as the authoritative metadata whenever it is non-empty. Direct pushes, missing tokens, API failures, and empty live bodies retain the commit-message fallback.

Fixes #12003.

## Verification

- Red on the test-only commit: the squash-head fixture retained the wrapped commit fragment instead of resolving to the live body.
- Green: `python3 -m unittest test_pr_preflight.py` — 42 tests passed, 2 skipped.
- Focused fallback coverage verifies direct pushes, missing tokens, API failures, and empty live bodies.
- `make preflight` and `scripts/pr-preflight --pr-body-file ...` pass from a complete checkout.

## Product invariants affected

none

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12085?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #12003, opened 2026-08-21
main-push Hygiene re-parses merge-commit body with git's 72-char wrapping and fails product-file-line-count-ratchet on well-formed PRs


kodjima33:
Repo Checks run 32455413042 (Hygiene) failed on main push 2e0598f99d even though the same checks were green on PR #11973 before merge.

Cause: the merge commit message embeds the PR body hard-wrapped at ~72 chars, which splits `Line-Count-Exception: path | BASE -> CUR | reason` declarations across lines. The main-push lane (`repo-checks.yml` "Run deterministic check manifest on main pushes") builds the PR-body file from `git log -1 --format=%B` and *appends* the recovered live PR body via `pr_metadata.py` — appending fixes missing citations but leaves the malformed wrapped lines in the file, so `check_product_file_line_count_ratchet.py` errors on them: `PR body line 88/91/94/98/102: malformed Line-Count-Exception`. The live PR body itself (40 lines) is well-formed.

Evidence:
- failing run: https://github.com/BasedHardware/omi/actions/runs/32455413042
- `git log -1 --format=%B 2e0598f99d` lines 88+ show the wrapped exception lines
- local repro: `pr_metadata.py --from-commit-body-file` + ratchet check reproduces the exact 5 errors

Suggested direction: when the live (#NNNN) PR body is recovered, use it *instead of* (not appended to) the commit-message body for line-format-sensitive checks, or strip/unwrap the commit-message half. This makes any regular-merge commit whose message carries the PR body fail the main-push lane; blast radius is a red Repo Checks on main per such merge (nothing gates on it — Release Eligibility passed — but it's alarm noise).

PR #12119, opened 2026-08-23
fix(ci): count a landed change once in the failure-class guard ratchet

igor-popov-dev:
## What changed and why

The guard ratchet counts, over a 90-day window, the first-parent integration changes that declare each failure class, and requires a class at or above the threshold to name a `canonical_prevention_artifact`. It builds that count from two sources: `git log --first-parent` for what has landed, and `--pr-body-file` for the change under review, whose declaration is added so that "the declaring PR fails first".

Both sources can describe the same change. On a main push HEAD *is* that integration change — its squash message carries the class trailer, so `declaration_counts` already counted it — and `pr_metadata.py` hands the same PR's body to the same check. The class is counted twice.

The count therefore depends on the vantage point rather than on the history:

| where the check runs | what it computes for a class with `k` landed declarations |
| --- | --- |
| `pull_request` (merge ref: declares nothing) | `k + 1` |
| the main push of that same PR | `k + 2` |
| local `pr-preflight` over a feature branch | `k + 2` |

With the default threshold of 3, `k = 1` is green in review and red on the push that follows it — an after-merge red on a gate the author can no longer answer, which is the failure mode the main-push metadata path (#11835, #12003, #12085) exists to prevent. It also means the local pre-push lane does not predict its own PR: it reports a class one declaration closer to the threshold than CI will.

The fix subtracts what HEAD already contributed: a class the body declares is added only when HEAD's own message did not declare it. On a `pull_request` run the checkout is the merge ref, whose message declares nothing, so the pending change still counts exactly as before.

none

## How it was verified

Verified by running the shipped checker against real history in this repository, at the merge commit of #11904 (`59af86646e`), whose live PR body declares `FC-assumed-credential-lifetime` — a class with no `canonical_prevention_artifact` and, in the 90 days to that date, exactly two first-parent declarations including this one.

```
# the PR's own vantage point: main before the merge, plus the live PR body
$ check_failure_class_guard_ratchet.py --root <main@59af86646e^> \
    --now 2026-08-22T14:41:39+00:00 --pr-body-file <live #11904 body>
OK: every class at or above the threshold names a guard artifact or is explicitly grandfathered.

# the main push of that same change: HEAD is the squash commit, same body
$ check_failure_class_guard_ratchet.py --root <main@59af86646e> ... --pr-body-file <same body>
FAIL: a recurring failure class has no reusable guard surface.
- FC-assumed-credential-lifetime: 3 declarations in the window (threshold 3) with no
  'canonical_prevention_artifact'. ...

# same commit, no body: the true number
OK: ...
```

With this change, the third invocation's verdict is what the main push produces (`OK`, exit 0), so the PR and its push agree.

One honest qualification: that particular push's `Repo Checks` run was cancelled by the workflow's concurrency group before it reached a conclusion, as main-push runs on a busy main frequently are. The claim verified here is that the shipped script computes a different number for the same history depending on where it runs, not that this specific run turned red.

- `python3 .github/scripts/test_check_failure_class_guard_ratchet.py` — 16 passed (14 before). The suite builds real disposable git repositories and runs the checker end to end, so the counting executes for real.
- The regression test fails on `main` with `AssertionError: 1 != 0` and, on stderr, `FC-recurring-thing: 3 declarations in the window (threshold 3)` where the window holds two.
- Not verified: a live GitHub Actions main-push run — that requires this change to be on `main` first. The `--root`/`--now` invocations above reproduce that job's inputs deterministically.

## Tests

- `test_a_landed_declaration_is_not_counted_twice_from_its_own_pr_body` — the regression test: one prior declaration, HEAD carrying the second as a squash subject with `(#NNNN)` and the trailer, and that PR's body passed in. Red before the change, green after.
- `test_a_pr_body_still_counts_when_the_squash_message_omits_the_trailer` — the paired guard: when the squash message carries no trailer, history never counted the change, and the body must still add it. Without this, "subtract HEAD" could quietly become "ignore the body".
- `test_pr_body_declaration_counts_toward_the_window` — kept, with one fixture correction: its HEAD used to be a commit declaring the class, which stands for a landed change and a pending one at the same time. A `pull_request` checkout is the merge ref, so the test now puts a `Merge <sha> into <sha>` commit at HEAD and still asserts that the body's declaration crosses the threshold.

## Failure class (fixes)

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12119?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-same-subject-counted-once-per-evidence-source.json
+++ b/.github/failure-classes/FC-same-subject-counted-once-per-evidence-source.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-same-subject-counted-once-per-evidence-source",
+  "violated_contract": "A threshold computed over subjects must count each subject once, however many sources describe it. When a counter sums a landed-history source with a pending-change source that can carry the same subject, the total moves with the vantage point: the subject counts once before it lands and twice afterwards. The threshold then fires on a total no single moment ever held, and a gate that was answerable while the change was in review turns red where the author can no longer answer it.",
+  "canonical_prevention": "Keep one authority per subject and let the second source contribute only what the first cannot see: before adding the pending-change evidence, drop the subjects the historical source already counted -- for a git-history counter, the ones HEAD's own message carries. Pin the three vantage points such a check runs from -- the pull request, that pull request's landed push, and the local pre-push run -- with a test each, so a counting change that makes them disagree fails in review.",
+  "canonical_prevention_artifact": [
+    ".github/scripts/check_failure_class_guard_ratchet.py",
+    ".github/scripts/test_check_failure_class_guard_ratchet.py"
+  ],
+  "evidence_prs": [10518],
+  "scope_hints": [
+    ".github/scripts/check_failure_class_guard_ratchet.py",
+    ".github/scripts/check_product_file_line_count_ratchet.py",
+    ".github/scripts/pr_metadata.py"
+  ],
+  "status": "open"
+}
--- a/.github/scripts/check_failure_class_guard_ratchet.py
+++ b/.github/scripts/check_failure_class_guard_ratchet.py
@@ -14,6 +14,9 @@ Determinism and hermeticity:
 - Counts come from `git log --first-parent` in the checkout. No network.
 - Integration changes are counted, not raw commits: one merged PR routinely
   carries a dozen `fix:` commits, so a raw-commit threshold would be noise.
+- The change under review is counted once wherever the window is measured: its
+  `--pr-body-file` declaration is added only when HEAD did not already carry it,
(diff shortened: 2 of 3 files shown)


kodjima33 (approved):
CI tooling fix: failure-class guard ratchet double-counted a landed change (once from git history, once from PR body) on main-push, causing spurious post-merge red. Small, well-tested, verified against real repo history, no product/security surface. 4/5.

PR #12123, opened 2026-08-23
fix(backend): name the exception behind a dead-lettered finalization

igor-popov-dev:
## What changed and why

`finalize_persisted_conversation` catches everything and logs the bounded code
`processing_failed`, deliberately keeping provider and validation messages out of the log
because they can quote the transcript. It also drops the exception type, and that is what
leaves an operator with nothing: a conversation dead-letters after three retries and the only
evidence is nine characters that read the same whether the provider timed out, the structured
output failed to validate, or the datastore rejected the write.

This logs `type(error).__name__` alongside the code. A class name quotes no transcript, and the
warning fifteen lines above in this same handler already logs exactly that under the same
constraint.

none

## How it was verified

- `pytest tests/unit/test_conversation_finalizer_failure_logging.py` → 2 passed.
- Restored only the handler from `upstream/main` and re-ran: the first case fails with
  `assert 'error=LookupError' in 'persisted conversation finalization failed uid=... conversation=... failure=processing_failed'`,
  so it fails for the missing type and not for an unrelated reason.
- The change is running on a self-hosted backend whose conversations dead-letter, which is where
  the gap was found: 24 `failure=processing_failed` lines in a day, none of them attributable.

## Tests

`backend/tests/unit/test_conversation_finalizer_failure_logging.py` drives the real handler with
a blocking-call double that fails inside the try block, and asserts both halves of the contract:
the log names the exception class, and it still does not contain the exception message (the part
that can quote a transcript).

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_conversation_finalizer_failure_logging.py
+++ b/backend/tests/unit/test_conversation_finalizer_failure_logging.py
@@ -1,0 +1,78 @@
+"""A dead-lettered finalization has to name the exception class behind it.
+
+The handler in ``utils/conversations/finalizer.py`` deliberately keeps provider and validation
+messages out of the log because they can quote the transcript. The class name is the one part
+that carries no transcript and is what makes ``processing_failed`` actionable.
+"""
+
+import logging
+
+import pytest
+
+from models.conversation_enums import ConversationStatus
+from utils.conversations import finalizer
+from utils.conversations.finalizer import ConversationFinalizationError
+
+
+class _Conversation:
+    """The minimum the finalizer touches before the failing step."""
+
+    def __init__(self):
+        self.id = "conversation-1"
+        self.status = ConversationStatus.processing
+        self.source = "omi"
+        self.discarded = False
+        self.structured = None
+        self.geolocation = None
+
+
+TRANSCRIPT_EXCERPT = "he said the quiet part out loud"
+
+
+@pytest.fixture
+def finalizer_that_fails_on_geolocation(monkeypatch):
+    """Let the finalizer reach its try block, then fail inside it with a telltale message."""
+
+    async def fake_run_blocking(_executor, function, *args, **kwargs):
+        if function is finalizer.conversations_db.get_conversation:
+            return {"id": "conversation-1"}
+        if function is finalizer.get_cached_user_geolocation:
+            raise LookupError(TRANSCRIPT_EXCERPT)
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified end to end — this is a well-scoped fix with the right regression test.

**`backend/utils/conversations/finalizer.py`** — the handler change is exactly the bounded addition it claims to be: the dead-letter `failure=processing_failed` line gains `error=%s` fed by `type(error).__name__`, and nothing else in the failure path changes. Class names carry no transcript content, and this matches the trust decision already made fifteen lines up, where the chat-first capture-arrival warning logs `type(error).__name__` under the same no-transcript constraint. The expanded comment documenting why the type is safe but the message is not is accurate against the code.

**`backend/tests/unit/test_conversation_finalizer_failure_logging.py`** — the harness drives the real `finalize_persisted_conversation` through a `run_blocking` double that fails inside the try block (`get_cached_user_geolocation` → `LookupError`), asserts the exception class appears in the dead-letter error record, and — the half that matters most — asserts the exception message (the transcript-capable part) still does not appear. Raising `AssertionError` on any unexpected blocking call keeps the double honest.

I independently reproduced both directions: on this head both tests pass, and with the `main` handler restored the first test fails precisely on the missing `error=LookupError`, so the regression test bites for the right reason.

Checks are green (Backend Hermetic Merge Gate, both stack gauntlets, Hermetic Backend E2E). One pre-existing aside, not this PR's concern: the dead-letter error line logs raw `uid` while the warning above it uses `sanitize_pii(uid)` — may be worth a small follow-up. Leaving the final merge call to the human maintainers.

---

kodjima33 (approved):
Backend bug fix: dead-lettered conversation finalizations logged only 'processing_failed' with no way to diagnose; now logs exception class name (not message, to avoid transcript leakage). Small, tested, verified not on main. 4/5.

PR #12125, opened 2026-08-23
fix(backend): declare the rest of the missing composite indexes (messages, conversations, memories)

igor-popov-dev:
> Rebased onto main now that #12117 has landed. That PR declared the `conversations (discarded, created_at)` index for the default conversation list read; this one covers the remaining forms (`conversations (status, created_at)`, `conversations (discarded, status, created_at)`, `memories (scoring, created_at)`, `messages (chat_session_id, created_at)`). They are complementary — `(discarded, status, created_at)` does not serve `discarded == False ORDER BY created_at`, because `status` in the middle breaks the prefix — they only conflicted textually, because both append to the same registry list and the same JSON node. Conflict resolved by keeping both sides; `firestore.indexes.json` regenerated from the merged registry. Ready for review.

## What changed and why

`get_messages`' session-scoped branch, plus three more read paths (`backend/database/chat.py`,
`conversations.py`, `memories.py`), filter/order Firestore collections in shapes that
`firestore_index_registry.py` never declared. Production has composite indexes for all of them
only because someone created them by hand at some point; a fresh self-host deploy gets
`FailedPrecondition` 400 the first time any of these paths runs:

- `messages`: `chat_session_id + created_at` (session-scoped message reads — a chat session's
  first page hits this branch, not the app-scoped one)
- `conversations`: bare `status + created_at` (`get_in_progress_conversation`,
  `get_action_items`), and `discarded + status + created_at` (default
  `GET /v1/conversations` with `include_discarded=false`, no source/category)
- `memories`: `scoring + created_at` (`get_memories`' default no-filter path — Firestore
  still needs a composite for a bare multi-field sort)

Adds these to the registry and regenerates `firestore.indexes.json`.

none

## How it was verified

  `firestore.indexes.json`; diff is exactly the 4 new composites above (76 lines).
- `.venv/bin/python -m pytest backend/tests/unit/test_firestore_query_contract.py` —
  34/34 passed, including 4 new regression tests (one per new query shape) that build each
  call site's real query chain through a recording Firestore fake and assert the composite
  is declared.
- `.venv/bin/python -m pytest backend/tests/unit/test_reconcile_firestore_indexes.py
  backend/tests/unit/test_universal_memory_list_cursor.py
  backend/tests/unit/test_action_items_due_range_index_contract.py
  backend/tests/unit/test_chat_message_count.py
  backend/tests/unit/test_chat_messages_pagination.py` — 75/75 passed, no regression.
- Not done: creating the composites by hand in a real Firestore project and hitting a live
  self-host backend (as was done for the sibling `plugin_id`-messages fix). Only unit-level
  verification here.

## Tests

Bug fix — regression tests added, one per new query shape: each builds the real query chain
through a recording Firestore fake and asserts `firebase_index_manifest()` declares a
matching composite.

## Failure class (fixes)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/firestore_index_registry.py
+++ b/backend/database/firestore_index_registry.py
@@ -191,6 +191,38 @@ INDEX_ONLY_REQUIREMENTS = (
         'COLLECTION',
         (_asc('status'), _asc('finished_at'), _asc('__name__')),
     ),
+    # Several conversations.py serving reads filter by `status` alone and sort by
+    # `created_at` descending (get_in_progress_conversation, get_action_items,
+    # get_last_completed_conversation, and the default `GET /v1/conversations`
+    # call with include_discarded=True). Production has this index only because
+    # it was created by hand; a fresh self-host 400s with FailedPrecondition the
+    # first time any of those paths runs.
+    FirestoreIndexRequirement(
+        'conversations_status_created',
+        'conversations',
+        'COLLECTION',
+        (_asc('status'), _desc('created_at'), _desc('__name__')),
+    ),
+    # `get_conversations`/`get_conversations_count`/`get_conversations_without_photos`
+    # called with include_discarded=False and a `statuses` filter (no source/category)
+    # produce this shape. Same story as above: hand-created in production, never
+    # declared here.
+    FirestoreIndexRequirement(
+        'conversations_discarded_status_created',
+        'conversations',
+        'COLLECTION',
+        (_asc('discarded'), _asc('status'), _desc('created_at'), _desc('__name__')),
+    ),
+    # `get_memories`' default path (no category/date filters, default
+    # sort='scoring_desc') orders by `scoring` then `created_at`, both descending,
+    # with zero `where` filters. Firestore still needs a composite for a bare
+    # multi-field sort. Hand-created in production, never declared here.
+    FirestoreIndexRequirement(
+        'memories_scoring_created',
+        'memories',
+        'COLLECTION',
+        (_desc('scoring'), _desc('created_at'), _desc('__name__')),
+    ),
     FirestoreIndexRequirement(
         'memory_items_tier_status_updated',
         'memory_items',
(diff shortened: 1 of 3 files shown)


kodjima33 (approved):
fix(backend): declares 4 missing composite indexes (messages/conversations/memories), scoped diff (204 lines, 3 files), regression tests included, CI green — scores 5/5 on confidence gate. Approve-only because mergeStateStatus is DIRTY (no rebase per hard rule — branch not touched).

Git-on-my-level:
Verified at the rebased head (19a0895) — this holds up well, and the conflict resolution with #12117 was done right: both `(discarded, created_at)` and the four new shapes are present, no duplicate signatures among the 54 manifest entries, and `firestore.indexes.json` regenerates exactly from the merged registry.

Per file:

- `backend/database/firestore_index_registry.py` — each new declaration matches the real query builders at this head: `chat.get_messages(chat_session_id=...)` builds `chat_session_id ==` + `created_at DESC`; `get_in_progress_conversation`, `get_action_items`, and `get_last_completed_conversation` build `status ==` + `created_at DESC`; `get_conversations` with `include_discarded=False` + `statuses` builds `discarded ==` + `status in` + `created_at DESC`; and `get_memories`' default path is the bare `scoring DESC, created_at DESC` sort (Firestore does require a composite for a bare multi-field sort). `MESSAGES_BY_SESSION_ORDERED_QUERY` is correctly registered in `QUERY_SPECS`.
- `backend/tests/unit/test_firestore_query_contract.py` — the new tests exercise the production query builders through the `_StreamRecordingFirestore` seam and assert the built shape exists in the manifest, rather than asserting on source strings. Ran the file at this head in a clean environment: 40/40 passed, including all five new tests.
- `firestore.indexes.json` — regeneration checked against the merged registry; entry-for-entry identical, so the checked-in manifest is not hand-edited drift.

One non-blocking observation for a possible follow-up (outside this PR's stated scope): `get_memories` with `categories` and/or date-range filters still combines `category in` / `created_at` range filters with the `scoring, created_at` order, which would need its own composite shapes if those filtered paths matter for fresh self-hosts. The PR description already scopes this work to the default paths, so that is consistent as-is.

The earlier approval predates the rebase, but the PR's own additions are unchanged by it — the registry diff against the approved commit is exactly #12117's landed content. Content-wise this should only need a quick re-approval on the new head.

---

kodjima33 (approved):
Confirmed not already fixed on main (only the discarded+created_at conversations index from #12117 exists; messages chat_session_id+created_at, conversations status+created_at, conversations discarded+status+created_at, and memories scoring+created_at are all genuinely missing). Scores 5/5: stated root cause, fix(backend) title addressing a concrete defect, scoped diff (204 lines, 3 files), CI green, regression tests per query shape. Merging.

PR #12126, opened 2026-08-23
chore(backend): keep the query-coverage baseline honest about itself

igor-popov-dev:
## What changed

- Refresh `firestore_query_coverage_baseline.json` against the current `main` after #12125.
- Keep the baseline-freshness guard in `firestore_query_coverage.py`.
- Keep the regression coverage for retired fingerprints and stale summary counts.

The regenerated inventory now records 62 registered and 102 eligible serving query shapes. It also removes the retired raw-unregistered fingerprints `6349dbe884bdc903` and `7974cb3672cd19e7`.

## Why

The ratchet previously accepted a baseline that contained query shapes no longer present in the tree and summary counts lower than the current inventory. That could silently re-permit retired debt and hold the coverage floor below reality. The freshness check makes the committed baseline match the generated inventory exactly.

No serving query, index, or registry entry changes in this PR.

none

## Verification after rebasing onto current main

- `python3 backend/scripts/firestore_query_coverage.py --check-ratchet` — exit 0; 62 registered / 102 eligible.
- `backend/.venv-pr12126/bin/python -m pytest backend/tests/unit/test_firestore_query_contract.py -m ""` — 44 passed.
- `black --check --line-length 120 --skip-string-normalization backend/scripts/firestore_query_coverage.py backend/tests/unit/test_firestore_query_contract.py` — clean.
- `git diff --check upstream/main...HEAD` — clean.

The full local manifest preflight was not runnable in the intentionally sparse checkout because unrelated trigger paths were absent; GitHub CI runs it against the complete repository.

## Failure class

Failure-Class: none

### DIFF
--- a/backend/scripts/firestore_query_coverage.py
+++ b/backend/scripts/firestore_query_coverage.py
@@ -474,6 +474,38 @@ def check_ratchet(report: Mapping[str, Any], baseline: Mapping[str, Any]) -> lis
     return errors
 
 
+REGENERATE_HINT = (
+    'regenerate with: python3 backend/scripts/firestore_query_coverage.py --format baseline '
+    '> backend/scripts/firestore_query_coverage_baseline.json'
+)
+
+
+def check_baseline_freshness(report: Mapping[str, Any], baseline: Mapping[str, Any]) -> list[str]:
+    """Report ways the committed baseline no longer describes the current inventory.
+
+    The ratchet only asks whether today's debt is a subset of the baselined debt, so a
+    baseline drifts silently in the safe direction: shapes that were registered or
+    rewritten stay listed as permitted debt, and the summary counts keep reporting an
+    older, lower coverage floor.  Both are enforcement surface, not just reporting
+    surface -- a stale entry re-permits the exact shape it was recorded for, and stale
+    counts hold the percentage ratchet below the coverage already achieved.
+    """
+    errors: list[str] = []
+    if baseline.get('schema_version') != 1:
+        return ['Firestore query coverage baseline has an unsupported schema version']
+    current = baseline_for(report)
+    for key in ('raw_unregistered', 'unsupported'):
+        stale = sorted(set(baseline.get(key, [])) - set(current[key]))
+        if stale:
+            errors.append(f'baselined {key} shape(s) no longer present: {", ".join(stale)}')
+    for key in ('registered_serving', 'eligible_serving'):
+        if int(baseline.get(key, 0)) != current[key]:
+            errors.append(f'baseline {key} is {baseline.get(key)}, current inventory reports {current[key]}')
+    if errors:
+        errors.append(REGENERATE_HINT)
+    return errors
+
+
 def _render_human(report: Mapping[str, Any]) -> str:
     serving = report['counts']['serving']
     non_serving = report['counts']['non_serving']
@@ -504,7 +536,7 @@ def main() -> int:
         report = report_for(inventory(waiver_ids=_load_waivers(args.waivers.resolve())))
         if args.check_ratchet:
             baseline = json.loads(args.baseline.resolve().read_text(encoding='utf-8'))
-            errors = check_ratchet(report, baseline)
+            errors = check_ratchet(report, baseline) + check_baseline_freshness(report, baseline)
             if errors:
                 for error in errors:
                     print(f'ERROR: {error}', file=sys.stderr)
--- a/backend/scripts/firestore_query_coverage_baseline.json
+++ b/backend/scripts/firestore_query_coverage_baseline.json
@@ -1,5 +1,5 @@
 {
-  "eligible_serving": 70,
+  "eligible_serving": 102,
   "raw_unregistered": [
     "05fdb282df637274",
     "0c75b579eb9f7f04",
@@ -8,17 +8,13 @@
     "22ddf94e28ce6efc",
     "24a4e24a308881dc",
     "2a846099f82aa044",
-    "2e3df6553113f693",
     "3897bd4019721e0e",
     "4697c32b6eda9c89",
     "4c64bf8ddb021ead",
     "5483ca63032ff8e5",
     "5e1801d8a93d95b2",
-    "6349dbe884bdc903",
     "67648c7513966177",
     "766f4615b8337172",
-    "7974cb3672cd19e7",
-    "7ed4cc825ad9f782",
     "8b14d56f55926b7c",
     "92c65999902cd6f2",
     "92ddcc8ed8cf0e94",
@@ -35,8 +31,6 @@
     "d2bd1198af3f25a1",
     "d3f5a567b8a2363d",
     "d4cf7676519c097c",
-    "dac5e11c45953ac1",
-    "db0f0af923c67e1a",
     "db91448693d164b7",
     "dba3e56194d22a9f",
     "e056656ec5fc547a",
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
Verified end to end in a clean worktree at 4805098 — this holds up nicely.

What I checked:

- `backend/scripts/firestore_query_coverage.py` — the new `check_baseline_freshness()` is a clean mirror of `check_ratchet`: schema guard first, then set-difference stale-entry detection for `raw_unregistered`/`unsupported`, then exact count checks for `registered_serving`/`eligible_serving`, with `REGENERATE_HINT` appended only when something failed. Wiring it into `main()` (`check_ratchet(...) + check_baseline_freshness(...)`) keeps both failure classes in one exit path, and the ratchet's own guards (new raw/unsupported shapes, count/percentage regression) are untouched.
- `backend/scripts/firestore_query_coverage_baseline.json` — regenerated, not hand-edited: restoring main's old baseline reproduces exactly the four stale ids (`2e3df6553113f693`, `7ed4cc825ad9f782`, `dac5e11c45953ac1`, `db0f0af923c67e1a`) plus both stale counts (`registered_serving` 24 → 42, `eligible_serving` 70 → 84) and the regenerate hint; the committed file passes `--check-ratchet` cleanly, including on a trial merge of this head into current `main`.
- `backend/tests/unit/test_firestore_query_contract.py` — the three new tests cover both synthetic failure modes plus the slow committed-file == `baseline_for(report)` contract; `pytest tests/unit/test_firestore_query_contract.py -m ""` → 37 passed locally.
- CI wiring: `.github/checks-manifest.yaml` already runs this script with `--check-ratchet` in local+ci lanes with triggers covering `backend/**/*.py` and the baseline file, so the freshness guard is genuinely enforced, not advisory.

Two notes for maintainers:

1. This tightens a merge-gate check: any PR that adds or rewrites a serving query shape now needs a baseline refresh in the same PR — the same contract `firestore-index-manifest` already applies to `firestore.indexes.json`. That's the point, but it's a new enforced failure mode, so maintainer sign-off on the contract change itself is warranted.
2. Sequencing with #12125 stands: this head is green against today's `main`, but once #12125 lands its registry changes will (correctly) make this baseline stale again, so the rebase + regenerate after #12125 is the right order — and #12125 will itself need a baseline bump once this contract lands. Minor housekeeping: the PR is marked ready for review while the description still leads with "Draft — still blocked on #12125".

Nice guard — the framing that a baseline can only ever drift in the direction that keeps CI green (stale entries silently re-permit retired shapes; stale counts pin the percentage floor below achieved coverage) is exactly the kind of self-describing baseline this check wanted.

---

kodjima33 (approved):
Solid self-check tooling (adds check_baseline_freshness so the query-coverage baseline can't silently rot again) but this is a chore/tooling enhancement, not a bug fix addressing a concrete defect — per policy, non-bug-fix backend changes are approve-only regardless of confidence. CI green, no abuse signals. Approving; not merging. Note: PR body says it should land after #12125 (baseline count math) — #12125 just merged, so a rebase/regen may be needed before this is mergeable.

Git-on-my-level (dismissed):
The freshness check itself is solid work — `check_baseline_freshness()` in `backend/scripts/firestore_query_coverage.py` mirrors `check_ratchet`, reuses `baseline_for()` for both the set-difference stale-shape detection and the exact count checks, and `main()` wires it in as `check_ratchet(...) + check_baseline_freshness(...)` so nothing rots silently again. I verified the two synthetic tests in `backend/tests/unit/test_firestore_query_contract.py` (stale `raw_unregistered`/`unsupported` ids; stale `registered_serving`/`eligible_serving` counts) by replicating their exact assertions against the module — both hold — and the `@pytest.mark.slow` committed-file invariant is the right guard for `firestore_query_coverage_baseline.json`. @kodjima33's approval covers this design; the one issue below is data, not code.

**Blocking before merge: the committed baseline is stale against current main — caught by this PR's own new check.** As the description anticipated, #12125 has landed (merged 14:40Z today), and `2af083251c` (JIT knowledge ledger) also registered new shapes after this baseline was generated on Aug 26. Test-merging 4805098 with current `origin/main` and running the check:

```
ERROR: baselined raw_unregistered shape(s) no longer present: 6349dbe884bdc903, 7974cb3672cd19e7
ERROR: baseline registered_serving is 42, current inventory reports 62
ERROR: baseline eligible_serving is 84, current inventory reports 102
```

Those two ids are the `folders.get_conversations_in_folder` / `folders.update_folder_conversation_count` shapes, now registered. The green checks on this head ran Aug 26 against the pre-#12125 base, so nothing has flagged this yet — merging as-is would turn the `firestore-query-coverage` gate red on main's next backend run, which is precisely the failure mode this PR exists to prevent.

Per the description's own plan: rebase onto post-#12125 main and regenerate —

```
python3 backend/scripts/firestore_query_coverage.py --format baseline > backend/scripts/firestore_query_coverage_baseline.json
```

— the exact command `REGENERATE_HINT` prints. No code changes needed. Happy to re-verify once pushed; the mechanism itself is a clear improvement over the silent drift that motivated this.

Leaving the `workflow-review` label for maintainer sign-off on the new CI failure mode (this changes what a manifest-registered check enforces); the data refresh itself only needs the regeneration above.

---

igor-popov-dev:
Rebased onto the current `main` (`add08a47e`) and regenerated the Firestore query-coverage baseline as requested. The baseline is now 62 registered / 102 eligible and the two retired raw fingerprints are gone. Focused verification is green: ratchet exit 0, 44 unit tests passed, and Black is clean.

Git-on-my-level:
Verified on the rebased head `73df0119` — the staleness that blocked my earlier review is resolved, exactly along the plan in the description.

What I checked:

- `backend/scripts/firestore_query_coverage_baseline.json` — regenerated the baseline independently from the merge-base tree (`add08a47e`, post-#12125 main) with the script as it exists on main, and the committed file is byte-identical to that output: 62 registered / 102 eligible, 37 raw-unregistered shapes, with the retired `6349dbe884bdc903` and `7974cb3672cd19e7` fingerprints gone. It also matches regeneration at current `main`, so this lands with the `firestore-query-coverage` gate green instead of re-creating the drift.
- `backend/scripts/firestore_query_coverage.py` — `check_baseline_freshness()` and the `check_ratchet(...) + check_baseline_freshness(...)` wiring in `main()` are unchanged from the previously reviewed version; `--check-ratchet` exits 0 on the new head.
- `backend/tests/unit/test_firestore_query_contract.py` — replicated all three new tests' assertions against the module directly (stale-shape rejection, stale-count rejection, and the `@pytest.mark.slow` committed-file invariant); each holds, and the slow invariant's `waiver_ids=set()` matches live behavior since `firestore_query_coverage_waivers.json` is empty.

I've dismissed the earlier changes-requested review as resolved. Thanks for the quick turnaround — the baseline now has to describe the inventory exactly, and it does.

Keeping `workflow-review` for one deliberate behavior change that wants a maintainer's explicit OK: this widens what a manifest-registered CI check enforces (from a subset-only ratchet to an exact-match baseline), i.e. a new red-condition on `main` if the committed file ever drifts again. Everything else here is mechanical and verified.

---
*by AI on behalf of David — @Git-on-my-level to reach David directly.*

kodjima33 (approved):
Verified genuine ratchet-freshness gap (main's check_ratchet doesn't require retired fingerprints be removed or counts stay exact), unfixed on main. CLEAN, CI green, already approved, tests added. Confidence 4/5.

PR #12128, opened 2026-08-23
fix(app): survive [core/duplicate-app] instead of failing to start

igor-popov-dev:
## What changed and why

The app does not start at all: instead of the first frame the user gets
`StartupFailureApp` with `[core/duplicate-app] A Firebase App named "[DEFAULT]" already
exists`, thrown out of `_init()`.

`if (Firebase.apps.isEmpty)` guaranteed nothing. `Firebase.apps` is empty until the Dart
side calls `initializeApp`, even when a native `[DEFAULT]` app is already running — on
Android `FirebaseInitProvider` creates it from `google-services.json` before any Dart code
runs, and on macOS the native SDK does the same. The real check lives *inside*
`initializeApp`: `firebase_core` pulls in the native apps and throws `duplicate-app` when
our `apiKey` / `databaseURL` / `storageBucket` disagree with the native ones
(`firebase_core_platform_interface`, `method_channel_firebase.dart`). That throw escaped
`_init()` before `runApp()`, so the whole app died — while a perfectly usable native
Firebase app sat right next to it. A configuration mismatch is worth complaining loudly
about; it is not a reason to refuse to start.

`ensureFirebaseApp()` (new `app/lib/startup_firebase.dart`) is now the single
initialization point: adopt the app that already exists, otherwise `initializeApp`, and
`duplicate-app` is no longer fatal — the existing app is adopted and run through the same
`Env.validateFirebaseProject`, so a genuinely foreign Firebase project is still caught and
still fails startup. Every other `FirebaseException` is still rethrown, and validating our
own flavor's options before `initializeApp` is unchanged and still fatal.

The FCM background handler goes through the same path. It runs in a separate Flutter
engine and called `Firebase.initializeApp()` with no arguments, so it could bring
`[DEFAULT]` up from the platform resources with parameters that differ from the ones the UI
engine uses; both engines now resolve the parameters identically.

The logic lives in its own library, generic over the app type, because it has to be
testable: `FirebaseApp` has no public constructor, and a test cannot import `main.dart`
since that pulls in the generated, gitignored `firebase_options_prod.dart`. Same reason
`startup_auth.dart` and `startup_routing.dart` are separate libraries.

`env_test.dart`'s static wiring tripwire named the removed `else` branch verbatim
(`Env.validateFirebaseProject(projectId: Firebase.app().options.projectId);`). It now pins
the same guarantee where the guarantee moved to — `main.dart` feeding the resolved app's
`options.projectId` into `Env.validateFirebaseProject` — instead of one branch's text. See
Tests for why this is a stronger check, not a weaker one.

INV-DATA-1 (production-family customer data-plane continuity) — touched, not relaxed.
`app/lib/main.dart` is on the invariant's locked paths, and the invariant's requirement here
is that a production-family artifact cannot end up on a different Firebase project or
account universe. That requirement is enforced by `Env.validateFirebaseProject`, and it
still runs on whichever app startup ends up with: the freshly initialized one, an
already-running native one, and — new in this PR — the one adopted after `duplicate-app`.
What changes is only the failure mode of a *matching* project whose non-identity parameters
(`apiKey` / `databaseURL` / `storageBucket`) differ: that used to kill the app and now
continues on the native app. A different `projectId` still throws `StateError` and still
fails startup, which is asserted directly by the new test.

## How it was verified

- `bash app/test.sh` (full suite, the same entry point CI uses) — `+1460 ~5`, all tests
  passed, exit 0.
- `flutter test test/unit/env_test.dart` — 23/23 passed, including the static wiring
  tripwire this PR updates.
- `flutter test test/unit/startup_firebase_duplicate_app_test.dart` — 6/6 passed.
- Red-before-green confirmed: with `ensureFirebaseApp` temporarily reverted to the old
  `if (apps.isEmpty) { validate; initializeApp } else { validate }` body, the run is
  `+4 -2` — the two duplicate-app cases fail with
  `[core/duplicate-app] A Firebase App named "[DEFAULT]" already exists`, and the four
  preserved-behaviour cases stay green.
- The updated tripwire was checked to still trip: replacing `main.dart`'s
  `validateProject: (projectId) => Env.validateFirebaseProject(projectId: projectId)` with
  an empty lambda turns `env_test.dart` red.
- `dart format --line-length 120 --set-exit-if-changed` on the four changed files — clean.
- `flutter analyze` on the changed files — the only finding is the pre-existing
  `library_private_types_in_public_api` info on `MyApp.of` in `main.dart`, untouched by this
  change; no new issues.
- Not verified: no on-device run was made for this revision. The original failure was
  observed on a self-hosted build with its own Firebase project (Android, where
  `FirebaseInitProvider` creates `[DEFAULT]` before Dart runs); the fix as reworked here has
  only been exercised through the test suite, which cannot reach a real native
  `[DEFAULT]` app.

## Tests

Bug fix — regression test added in `app/test/unit/startup_firebase_duplicate_app_test.dart`.
The case that would have caught this: a native `[DEFAULT]` that becomes visible only when
`initializeApp` throws `duplicate-app` must be adopted rather than kill startup. It also
pins the guardrails: a native app from a genuinely different Firebase project still fails
validation, a non-`duplicate-app` `FirebaseException` still fails startup, a flavor
configured for the wrong project still fails before `initializeApp` is reached, and the
pre-existing "already initialized natively" path (macOS) still skips `initializeApp`.

`app/test/unit/env_test.dart`'s "main invokes the production startup routing seam" case is a
static tripwire (`AGENTS.md`: "Asserting that source strings occur in a certain order is a
static tripwire, not behavioral coverage"). Its third assertion quoted a statement this PR
deletes, so it was adapted rather than dropped, and it is still a single assertion. The
adaptation strengthens it: the old string only proved that one `if`/`else` branch contained
that text, while the new one pins the wiring — the resolved app's `options.projectId` going
into `Env.validateFirebaseProject` — that now covers every path, including the newly added
adopt-after-`duplicate-app` one. That the wiring is actually exercised on each of those
paths is no longer left to a string at all: it is asserted behaviorally in the new test,
which the "already initialized" branch previously had no coverage for. The two ordering
assertions in that test (routing seam present, and before `ServiceManager.init()`) are
untouched and still pass.

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/app/lib/main.dart
+++ b/app/lib/main.dart
@@ -38,6 +38,7 @@ import 'package:omi/firebase_options_prod.dart' as prod;
 import 'package:omi/flavors.dart';
 import 'package:omi/startup_auth.dart';
 import 'package:omi/startup_failure_app.dart';
+import 'package:omi/startup_firebase.dart';
 import 'package:omi/startup_routing.dart';
 import 'package:omi/l10n/app_localizations.dart';
 import 'package:omi/pages/apps/providers/add_app_provider.dart';
@@ -87,10 +88,34 @@ import 'package:omi/utils/platform/platform_service.dart';
 import 'package:omi/utils/platform/platform_manager.dart';
 import 'package:omi/utils/notification_channel_strings.dart';
 
+/// Firebase parameters for the current flavor, resolved identically in every engine.
+FirebaseOptions _firebaseOptionsForFlavor() => Env.profile == AppEnvironmentProfile.localDev
+    ? local.DefaultFirebaseOptions.currentPlatform
+    : prod.DefaultFirebaseOptions.currentPlatform;
+
+/// The single Firebase entry point for every Flutter engine in the app.
+///
+/// See [ensureFirebaseApp] for why `Firebase.apps.isEmpty` was never a valid
+/// guard and why `[core/duplicate-app]` must not kill startup.
+Future<FirebaseApp> _ensureFirebaseApp() {
+  final options = _firebaseOptionsForFlavor();
+  return ensureFirebaseApp<FirebaseApp>(
+    existingApp: () => Firebase.apps.isEmpty ? null : Firebase.app(),
+    configuredProjectId: options.projectId,
+    initializeApp: () => Firebase.initializeApp(options: options),
+    projectIdOf: (app) => app.options.projectId,
+    validateProject: (projectId) => Env.validateFirebaseProject(projectId: projectId),
+  );
+}
+
 /// Background message handler for FCM data messages
 @pragma('vm:entry-point')
 Future<void> _firebaseMessagingBackgroundHandler(RemoteMessage message) async {
-  await Firebase.initializeApp();
+  // Same path as _init(). This runs in a SEPARATE Flutter engine and used to call
+  // Firebase.initializeApp() with no arguments, so it could bring [DEFAULT] up
+  // from the platform resources with parameters that differ from the ones the UI
+  // engine uses. Now both engines resolve the parameters the same way.
+  await _ensureFirebaseApp();
   await NotificationChannelStrings.loadAppLocale();
 
   await AwesomeNotifications().initialize(null, [
@@ -142,18 +167,7 @@ Future _init() async {
   LimitlessDeviceConnection.realtimeSuppressionPolicy = () => SharedPreferencesUtil().batchModeEnabled;
 
   // Firebase
-  if (Firebase.apps.isEmpty) {
-    final profile = Env.profile;
-    final options = profile == AppEnvironmentProfile.localDev
-        ? local.DefaultFirebaseOptions.currentPlatform
-        : prod.DefaultFirebaseOptions.currentPlatform;
-    Env.validateFirebaseProject(projectId: options.projectId);
-    await Firebase.initializeApp(options: options);
-  } else {
-    // Firebase may already be initialized by native SDK (macOS)
-    debugPrint('Firebase already initialized.');
-    Env.validateFirebaseProject(projectId: Firebase.app().options.projectId);
-  }
+  await _ensureFirebaseApp();
 
   if (Env.profile.usesFirebaseAuthEmulator) {
     await FirebaseAuth.instance.useAuthEmulator(Env.firebaseAuthEmulatorHost, Env.firebaseAuthEmulatorPort);
--- a/app/lib/startup_firebase.dart
+++ b/app/lib/startup_firebase.dart
@@ -1,0 +1,77 @@
+import 'package:firebase_core/firebase_core.dart';
+import 'package:flutter/foundation.dart';
+
+/// The `firebase_core` error code raised when `initializeApp` is asked to create
+/// an app that the native SDK has already created with different parameters.
+const String _duplicateAppCode = 'duplicate-app';
+
+/// Bring up the Firebase app for the current engine without letting a
+/// configuration mismatch take the whole app down.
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Thanks @igor-popov-dev — well-diagnosed fix. I checked the core mechanics against the pinned `firebase_core` 3.13.0 / `firebase_core_platform_interface` 5.4.2 sources and they hold up.

**Verified:**
- `app/lib/startup_firebase.dart` — the diagnosis is correct: `Firebase.apps` only mirrors Dart-side `appInstances`, which are populated lazily inside `initializeApp()` (`method_channel_firebase.dart`), so the old `if (Firebase.apps.isEmpty)` guard could never see a natively created `[DEFAULT]` app and walked straight into `duplicateApp()`. Also confirmed `duplicate-app` is thrown only after `appInstances[DEFAULT]` exists, so the recovery path's `existingApp()` is non-null in practice (and rethrows defensively if not).
- The safety pin survives every path: fresh init, pre-existing app, and adopted-after-duplicate all run `Env.validateFirebaseProject`, so a genuinely foreign Firebase project still fails startup — covered by the `someone-elses-project` test.
- `app/lib/main.dart` — both `_init()` and the FCM background handler now resolve options identically via `_firebaseOptionsForFlavor()`. Since `Env.profile` is compile-time (`F.env` / `String.fromEnvironment`), this is safe in the background handler's separate engine, and it removes the old divergence where the handler's bare `initializeApp()` could bring `[DEFAULT]` up from platform resources with different parameters.
- `app/test/unit/startup_firebase_duplicate_app_test.dart` — good behavioral coverage of all five paths, including the regression case and the misconfigured-flavor-still-fatal case.
- `app/test/unit/env_test.dart` — reasonable re-pinning from the verbatim statement to a wiring regex, with the behavioral assertions now living in the new test.

**One non-blocking thought:** after adopting an app whose parameters differ, the mismatch is only `debugPrint`-ed, which is invisible in release builds. A non-fatal `FirebaseCrashlytics.instance.recordError` alongside would keep this observable in the field — exactly the population that hits config drift.

The one thing I'd like a maintainer to sign off on is the failure-mode policy itself: startup used to refuse to run on a Firebase parameter mismatch, and this PR deliberately makes that survivable (with the project-id pin still enforced). That's a defensible product call — it just deserves an explicit maintainer decision before merge.

---

kodjima33 (approved):
Mobile (Flutter) startup fix: replaces the broken Firebase.apps.isEmpty guard (which never detects a native-created [DEFAULT] app on Android/macOS) with a real duplicate-app-tolerant init that adopts the existing app instead of crashing with [core/duplicate-app]. Well tested (166-line dedicated test file), but labelled needs-maintainer-review and touches the app startup path for every user — leaving for Nik to sign off before merge rather than auto-merging.

igor-popov-dev:
@Git-on-my-level `need human response` — this one is approved and CI-clean, and the only thing left is the failure-mode policy question you flagged: startup used to refuse to run on a Firebase parameter mismatch, and this PR makes that survivable while still hard-failing on a foreign project id (`Env.validateFirebaseProject` runs on all three paths, covered by the `someone-elses-project` test). Could a maintainer make that call so this can land?

On your non-blocking note: happy to add a non-fatal `FirebaseCrashlytics.instance.recordError` next to the `debugPrint` on the adopted-with-different-parameters path, so config drift stays visible in release builds. Say the word and I'll push it into this PR before merge — otherwise I'll leave the diff as reviewed.

PR #12129, opened 2026-08-23
fix(app): stop the summary-templates sheet spinning forever when there is nothing to show

igor-popov-dev:
## What changed and why

`_AppsList` in `app/lib/pages/conversation_detail/widgets/summarized_apps_sheet.dart`
decided it was loading by looking at the data alone:

    final isLoading = enabledApps.isEmpty && suggestedApps.isEmpty;

so "still loading" and "loaded, and there is nothing" were the same state. Both
fetches behind those caches (`fetchAndCacheSuggestedApps`,
`fetchAndCacheEnabledConversationApps`) also swallow their errors into empty
lists, which makes the two cases indistinguishable from the data side by
construction. A user with no installed summary app — every new account — got a
shimmer skeleton that never resolved; `ShimmerWithTimeout` only freezes the
animation after 5s, so the screen settles into the same grey placeholder instead
of ever showing content.

The state now tracks completion explicitly: `_fetchApps()` sets a `_loaded` flag
in a `finally` block, and `isLoading` requires `!_loaded`. Once the fetch is
done, an empty result renders the sheet's empty state — the "Get Creative"
section with "Create Custom Template" and "All Templates", which is how a user
with no templates gets a first one.

none

## How it was verified

- `flutter test test/widgets/summarized_apps_sheet_empty_state_test.dart` — 2/2
  passed.
- Same test file re-run against the unmodified widget (fix reverted, test kept):
  the empty-state case fails with `Found 4 widgets with type
  "ShimmerWithTimeout"` where none are expected, i.e. the skeleton is still on
  screen after both fetches resolved. The loading-phase case stays green in both
  runs, so the fix does not simply remove the skeleton.
- `flutter test test/widgets/` — 170/170 passed.
- `dart format --line-length 120 --set-exit-if-changed` on both changed files —
  clean. `flutter analyze` on both changed files — no issues found.

## Tests

Bug fix — regression test added in
`app/test/widgets/summarized_apps_sheet_empty_state_test.dart`. It pumps the real
`SummarizedAppsBottomSheet` over a `ConversationDetailProvider` whose two app
fetches are held open by completers, asserts the skeleton is on screen while they
are in flight, then resolves both with nothing and asserts the skeleton is gone
and the empty state is rendered. Overriding the fetches rather than the caches is
deliberate: it reproduces the production shape where a finished-but-empty fetch
and an unfinished one deliver identical data.

## Failure class (fixes)

Failure-Class: none
```


<!-- This is an auto-generated description by cubic. -->
<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12129?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>
<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/app/lib/pages/conversation_detail/widgets/summarized_apps_sheet.dart
+++ b/app/lib/pages/conversation_detail/widgets/summarized_apps_sheet.dart
@@ -139,6 +139,12 @@ class _AppsListState extends State<_AppsList> {
     }
   }
 
+  // The fetch is done, whatever it came back with. Without this flag "empty" and
+  // "still loading" are indistinguishable: both fetches swallow their errors into
+  // empty lists, so the sheet stayed in the skeleton forever for anyone without a
+  // single installed template.
+  bool _loaded = false;
+
   Future<void> _fetchApps() async {
     try {
       await Future.wait([
@@ -147,6 +153,10 @@ class _AppsListState extends State<_AppsList> {
       ]);
     } catch (e) {
       Logger.debug('Error fetching apps: $e');
+    } finally {
+      if (mounted) {
+        setState(() => _loaded = true);
+      }
     }
   }
 
@@ -227,7 +237,7 @@ class _AppsListState extends State<_AppsList> {
     final enabledApps = widget.provider.cachedEnabledConversationApps;
     final suggestedApps = widget.provider.cachedSuggestedApps;
 
-    final isLoading = enabledApps.isEmpty && suggestedApps.isEmpty;
+    final isLoading = !_loaded && enabledApps.isEmpty && suggestedApps.isEmpty;
 
     if (isLoading) {
       return _buildShimmerLoading();
--- a/app/test/widgets/summarized_apps_sheet_empty_state_test.dart
+++ b/app/test/widgets/summarized_apps_sheet_empty_state_test.dart
@@ -1,0 +1,137 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:provider/provider.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+import 'package:omi/backend/preferences.dart';
+import 'package:omi/backend/schema/conversation.dart';
+import 'package:omi/backend/schema/structured.dart';
+import 'package:omi/l10n/app_localizations.dart';
+import 'package:omi/pages/conversation_detail/conversation_detail_provider.dart';
+import 'package:omi/pages/conversation_detail/widgets/summarized_apps_sheet.dart';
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Verified the root cause and the fix — clean, well-scoped change, and thank you for the high-quality regression test.

**`app/lib/pages/conversation_detail/widgets/summarized_apps_sheet.dart`** — the diagnosis checks out: `fetchAndCacheSuggestedApps` and `fetchAndCacheEnabledConversationApps` in `conversation_detail_provider.dart` both catch their errors and leave the caches empty, so the old `enabledApps.isEmpty && suggestedApps.isEmpty` check could never distinguish "fetch finished with nothing" from "fetch still in flight", and `ShimmerWithTimeout` only freezes the animation after its timeout rather than resolving the sheet. Gating on the new `_loaded` flag set in `_fetchApps()`'s `finally` (with the `mounted` guard) is the right shape: it covers both the empty-success and swallowed-error paths, and the flag flips exactly once even if one of the two futures throws (`Future.wait` is eager, so both requests still complete). The populated-list path is untouched.

**`app/test/widgets/summarized_apps_sheet_empty_state_test.dart`** — good design: overriding the fetch methods themselves (rather than the caches they fill) with completers reproduces the real failure mode, and the two tests pin both halves of the contract — skeleton visible while the fetches are held open, then the "Get Creative" / "Create Custom Template" / "All Templates" empty state once they resolve with nothing. The `_ignoreListTileInkSplashWarning` helper is narrowly scoped to the pre-existing ink-splash framework warning and restores `FlutterError.onError` in teardown. CI's Dart Analyze & Tests job ran the full app suite including this file and passed.

One optional, non-blocking thought for the future: on a swallowed fetch error the sheet now lands on the same empty state as a genuinely empty account. If the product ever wants a "couldn't load — retry" affordance, `_loaded` plus an error flag would be the natural place. Nothing to change in this PR.

---

kodjima33 (approved):
Solid mobile bug fix: isLoading was derived purely from data emptiness, so a finished-but-empty template fetch was indistinguishable from still-loading, leaving new accounts stuck on a shimmer skeleton forever. Fix tracks completion explicitly with a _loaded flag. Scoped diff (149 lines), regression test added and verified to fail against the unmodified widget, CI green. Confidence 5/5.

PR #12130, opened 2026-08-24
fix(app): give a client-synthesized cutover fence a way out

igor-popov-dev:
## What changed and why

Losing the network on a cold start can strand the app behind the account-migration screen with no
exit and no retry, while the server is not asking for a migration at all — it keeps answering
`state: legacy` / `client_action: none`. Force-stopping the app is the only way out, and the next
cold start can land in the same place.

Two independent paths produce that dead end.

`bindAuthenticatedOwner()` synchronously poisoned `_control` to `unavailable()` (client action
`migration_maintenance`) on ANY owner change. That synchronous fence exists to stop account A's
stale "allow" from leaking into account B's session while B's fetch is in flight — but the very
first bind of a fresh runtime is also an "owner change" (`null` → uid), and there is no prior
decision to leak there. `applyFetchResult` has a fallback for exactly that case ("no authoritative
projection yet: stay legacy-compatible"), but it re-derives its decision from `_control`, which the
bind has already poisoned, so on a cold start it could never fire.

Separately, the blocking screen offered no way out once ANY authoritative projection had been seen
— even when the projection that had been seen said ALLOW. After one successful `legacy`/`none`
fetch that flag is true forever, so a single blown refresh synthesized a "Migration in progress"
screen the server had never asked for. Nothing lifted it either: product traffic is blocked while
the fence is up, so no other code path re-fetches the control document and the app cannot notice
the server answering again.

The contract this restores: a fence only stands on the server's own word, and no fence is permanent
while the server disagrees.

- `AccountCutoverRuntime` tracks the last authoritative projection and exposes
  `canSkipUnresolvedFence` — true only while this client synthesized the fence AND the server had
  authoritatively allowed the owner. It is cleared on owner change and on sign-out, so nothing leaks
  across accounts, including as the last-known-good a skip could return to.
- `bindAuthenticatedOwner` fences synchronously only on a genuine owner switch. A real
  owner-to-owner switch still fences immediately and stays fenced on failure, preserving the
  documented anti-leak property.
- `applyFetchResult`: a transport failure after an authoritative ALLOW stays on the last-known-good
  projection instead of synthesizing a maintenance fence — a blip is not evidence of a migration.
  An explicit 503 still fences (the server said so), and a fence-state prior (`migrating`, `new`,
  `rolled_back_stranded`) still fails closed across an outage. Both unchanged.
- `skipUnresolvedFence()` returns to that last authoritative projection and nothing else. It is a
  no-op whenever there is no server-sent allow to go back to — a server-decided fence, an
  unavailable or malformed document before any allow, or an owner whose first fetch has not landed
  — so it can never synthesize a legacy projection of its own.
- `AccountCutoverBlockingGate` becomes stateful and re-fetches the control document every 30s while
  any fence is on screen. That is what lets a transient-outage fence lift on its own, a fence with
  no escape hatch clear at all, and a real migration fence lift as soon as the migration finishes.

INV-CUTOVER-1

The locked contract is unchanged, and the escape hatch is deliberately scoped so it stays that way.
State is still discovered only through the authenticated cutover control projection; a malformed or
unavailable document still fences; a `migrating` / `new` / `rolled_back_stranded` projection still
fails closed across an outage; and because the skip can only restore a projection the server itself
sent, nothing here can present a legacy projection as an answer the server never gave. The one
behaviour that changes is which fences are permanent: a fence this client synthesized on top of a
server "allow" is no longer one of them.

## How it was verified

`app/test/unit/account_cutover_gate_test.dart` and `app/test/widgets/account_cutover_fence_test.dart`
pass (22 tests), and so does the rest of `bash app/test.sh` (1463 passed, 5 skipped).
`bash app/scripts/analyze_ratchet.sh` passes. `dart format --line-length 120
--set-exit-if-changed` is clean on all four files, and `flutter analyze` reports no issues for
`lib/services/account_cutover` and both test files.

The three widget tests were run against the unmodified `lib/` to confirm they describe behaviour
that does not exist yet: all three fail there — two because the blocking screen renders no
`TextButton` at all, and the retry test because the fetch counter is still 1 after 30 simulated
seconds.

The unit tests assert on API this PR introduces, so they cannot run against the old code. They were
checked by mutation instead — each mutation reverts exactly one behaviour and leaves the rest
intact:

- fencing unconditionally on any owner change → the cold-start recovery test goes red;
- dropping the "transport failure after an authoritative allow" branch → the transport-blip test
  goes red;
- keying the skip affordance to `hasAuthoritativeControl` again → all three widget tests go red;
- removing the gate's 30s retry → the retry test goes red;
- restoring the old skip semantics (`_hasAuthoritative` guard, blank legacy default) → the
  owner-switch, escape-hatch and 503 tests and the first widget test go red;
- letting the skip fall back to a legacy default when there is nothing to return to → the
  owner-switch, escape-hatch and server-confirmed-fence tests go red;
- letting the previous owner's projection survive an owner change → the owner-switch test goes red.

Not exercised on a device; the incident path is reproduced in the widget tests.

## Tests

`app/test/widgets/account_cutover_fence_test.dart` — new, 3 tests over the real
`AccountCutoverBlockingGate`:

- a fence the server never confirmed (authoritative allow, then a 503 on refresh) renders the skip
  affordance, and pressing it returns to product traffic on the last projection the server actually
  sent — `account_generation` is preserved, not reset. This is the regression: that screen used to
  render with no way out at all.
- the escape hatch is hidden only while the server itself confirms the fence — a server-declared
  `migrating` projection renders without it, and on the same mounted widget an
  allow-then-failed-refresh fence brings it back.
- the fence re-fetches the control document every 30s and lifts itself when the server recovers:
  no fetch at 29s, one at 30s (server still down, fence holds), a third at 60s that succeeds and
  clears the screen. That fence has no escape hatch at all — the server never allowed this owner —
  so the retry is the only thing that can clear it.

`app/test/unit/account_cutover_gate_test.dart` — 6 added tests:

- the first bind of a fresh runtime recovers to legacy-compatible after a transport failure (the
  cold-start half of the bug);
- a genuine owner switch still stays fenced across a transport failure, and gets no escape hatch,
  because the previous owner's projection was discarded and there is nothing this owner was ever
  allowed to fall back to;
- the escape hatch never invents an allow the server has not given — refused while the owner's
  first fetch is still in flight, and refused once the server answers `migrating`;
- a transport blip after an authoritative allow does not fence and keeps the last-known-good
  generation;
- an explicit 503 after an authoritative allow does fence, but that fence stays skippable back to
  the last-known-good projection;
- a server-confirmed fence refuses the escape hatch across an outage.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/services/account_cutover/account_cutover_blocking_gate.dart
+++ b/app/lib/services/account_cutover/account_cutover_blocking_gate.dart
@@ -4,6 +4,7 @@
 /// LIFECYCLE: permanent
 library;
 
+import 'dart:async';
 import 'dart:io';
 
 import 'package:flutter/material.dart';
@@ -13,7 +14,7 @@ import 'package:omi/services/account_cutover/account_cutover_gate.dart';
 import 'package:omi/services/account_cutover/account_cutover_runtime.dart';
 import 'package:omi/utils/l10n_extensions.dart';
 
-class AccountCutoverBlockingGate extends StatelessWidget {
+class AccountCutoverBlockingGate extends StatefulWidget {
   const AccountCutoverBlockingGate({
     super.key,
     this.productBuilder,
@@ -30,21 +31,67 @@ class AccountCutoverBlockingGate extends StatelessWidget {
   static final Uri _playStoreUrl = Uri.parse('https://play.google.com/store/apps/details?id=com.friend.ios');
 
   @override
+  State<AccountCutoverBlockingGate> createState() => _AccountCutoverBlockingGateState();
+}
+
+class _AccountCutoverBlockingGateState extends State<AccountCutoverBlockingGate> {
+  // While a fence is on screen, nothing else in the app re-fetches the cutover
+  // control (product traffic is blocked), so without this timer the fence
+  // could only ever clear on an app lifecycle event. Poll so a fence caused by
+  // a transient control-plane outage lifts on its own — and a real migration
+  // fence lifts as soon as the migration completes.
+  Timer? _fenceRefreshTimer;
+  bool _refreshInFlight = false;
+
+  static const _fenceRefreshInterval = Duration(seconds: 30);
+
+  void _syncFenceRefreshTimer(bool fenceVisible) {
+    if (fenceVisible && _fenceRefreshTimer == null) {
+      _fenceRefreshTimer = Timer.periodic(_fenceRefreshInterval, (_) async {
+        if (_refreshInFlight) return;
+        _refreshInFlight = true;
+        try {
+          await AccountCutoverRuntime.instance.refresh();
+        } finally {
+          _refreshInFlight = false;
+        }
+      });
+    } else if (!fenceVisible && _fenceRefreshTimer != null) {
+      _fenceRefreshTimer!.cancel();
+      _fenceRefreshTimer = null;
+    }
+  }
+
+  @override
+  void dispose() {
+    _fenceRefreshTimer?.cancel();
+    super.dispose();
+  }
+
+  @override
   Widget build(BuildContext context) {
     return ListenableBuilder(
       listenable: AccountCutoverRuntime.instance,
       builder: (context, _) {
         final runtime = AccountCutoverRuntime.instance;
         final decision = runtime.decision;
-        if (decision == AccountCutoverGateDecision.allowProductTraffic) {
-          return productBuilder?.call(context) ?? child!;
+        final fenceVisible = decision != AccountCutoverGateDecision.allowProductTraffic;
+        _syncFenceRefreshTimer(fenceVisible);
+        if (!fenceVisible) {
+          return widget.productBuilder?.call(context) ?? widget.child!;
         }
 
         return AccountCutoverBlockingView(
           decision: decision,
           strandedNewData: runtime.control.strandedNewData,
-          appStoreUrl: _appStoreUrl,
-          playStoreUrl: _playStoreUrl,
+          appStoreUrl: AccountCutoverBlockingGate._appStoreUrl,
+          playStoreUrl: AccountCutoverBlockingGate._playStoreUrl,
+          // A fence this client synthesized after the server had allowed the
+          // owner (unreachable control plane, a blown refresh) keeps a manual
+          // way back to that last authoritative projection. A fence the server
+          // itself decided — or one with no server allow to return to — does
+          // not.
+          onSkipUnresolvedFence: runtime.canSkipUnresolvedFence ? runtime.skipUnresolvedFence : null,
         );
       },
     );
@@ -58,12 +105,18 @@ class AccountCutoverBlockingView extends StatelessWidget {
     required this.strandedNewData,
     required this.appStoreUrl,
     required this.playStoreUrl,
+    this.onSkipUnresolvedFence,
   });
 
   final AccountCutoverGateDecision decision;
   final bool strandedNewData;
   final Uri appStoreUrl;
   final Uri playStoreUrl;
+
+  /// Non-null only while this client synthesized the fence and the server had
+  /// authoritatively allowed the owner. Pressing it returns to that last
+  /// authoritative projection for this session.
+  final VoidCallback? onSkipUnresolvedFence;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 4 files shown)


Git-on-my-level:
Thanks @igor-popov-dev — this fixes a real dead end, and the state matrix is worked through carefully.

Verified against the head (916b4e0):

**account_cutover_runtime.dart** — `_lastAuthoritativeControl` is recorded only from authoritative results and cleared on every owner change (`bindAuthenticatedOwner`), so account A's allow can never become account B's skip target — good. The new `transportFailure` branch in `applyFetchResult` prefers the last authoritative *allow* over `_control`, which is what makes a transport blip after a successful fetch non-fatal. `skipUnresolvedFence()` can only restore a projection the server actually sent (never synthesizes a legacy default), and it restores `_resolvedForOwner` + notifies, so the fence lifts reactively. First-bind poisoning is now conditional on `isGenuineOwnerSwitch`, which un-strands cold starts while keeping the synchronous fence for real account switches — the `applyFetchResult` legacy-compatible fallback can finally fire.

**account_cutover_blocking_gate.dart** — conversion to StatefulWidget with a 30s `Timer.periodic` refresh while the fence is visible is the missing unblock: nothing else re-fetches control while product traffic is blocked. Timer starts/cancels via `_syncFenceRefreshTimer` on visibility change and is cancelled in `dispose()`; the `_refreshInFlight` re-entrancy guard is correct. `onSkipUnresolvedFence` is only non-null while `canSkipUnresolvedFence` holds, so the button disappears as soon as the server's own migrating/503 fence lands. Skip restores the server-sent projection including account generation, so `X-Account-Generation` (shared.dart:136) stays consistent with the backend's enforcement.

**account_cutover_gate_test.dart** — the six new unit tests cover the important regressions: fresh-runtime first-bind recovery, genuine owner switch staying fenced with no leaked allow, the escape hatch refusing to invent an allow (pre- and post-fence-answer), transport-blip-after-allow staying on last-known-good, explicit 503 fencing but keeping skip, and server-confirmed fence refusing skip across outages.

**account_cutover_fence_test.dart** — widget coverage is the part previous tests couldn't reach: skip visible only for client-synthesized fences and actually returning to `product` with generation 5 intact, hidden for server-confirmed fences, and the 30s poll lifting the fence once the server recovers (fetches 1→2→3 via the override client seam).

One thing needs a human decision rather than a code change: this deliberately relaxes the fail-closed posture the subsystem was given in 7dc893dcad ("fail closed on account cutover control refresh"). After an authoritative allow, a transport failure now resumes product traffic on the last-known-good projection instead of fencing, and the manual Skip lets a user dismiss a fence while the control plane is genuinely 503-ing. That is a defensible availability-over-strictness trade (the backend still enforces via generation headers), but it's the cutover owner's call whether a client should keep serving on a projection that may be stale during an active migration window. Leaving for human maintainer review on that posture decision; the implementation looks correct.

---

kodjima33 (approved):
Careful, well-tested fix for a real cold-start dead end behind the account-migration fence. Approving but holding the merge for your call: this deliberately relaxes the fail-closed posture from 7dc893dcad (transport failure after an authoritative allow now resumes on last-known-good instead of fencing, plus a manual Skip affordance during a 503). Backend still enforces via account-generation headers, but it's a security-posture trade you should weigh in on — also carries needs-maintainer-review label and diff is 472 lines (over the 300-line auto-merge scope). Confidence 4/5, approve-only.

igor-popov-dev:
@Git-on-my-level `need human response` — approved and CI-clean; what's left is the security-posture call you asked a human to weigh in on: this deliberately relaxes the fail-closed stance from 7dc893dc, so a transport failure *after* an authoritative allow resumes on the last-known-good projection instead of fencing, plus a manual Skip while the control plane is 503-ing. The backend still enforces via `X-Account-Generation`, and Skip can only restore a projection the server actually sent (it never synthesizes an allow) — but whether a client may keep serving on a possibly-stale projection during an active migration window is the cutover owner's decision, not mine.

Could a maintainer sign off (or tell me to make it stricter)? If the posture is the sticking point, I'm glad to split it: the cold-start first-bind recovery and the 30s fence refresh stand on their own and could land separately from the transport-failure fallback and the Skip affordance. Also happy to break up the 472-line diff along the same seam if that helps it fit under the auto-merge scope.

PR #12141, opened 2026-08-24
fix(rn): sort macos rewind timeline in one time unit

undivisible:
## Summary
- macOS Home rewind timeline was sorting conversations (Date.parse ms) against memories (epoch seconds), so every conversation stacked above every memory.
- Sort now uses `projectionTimestamp` so both kinds share milliseconds; memories without a time sort last.
- Adds a regression test with a seconds-based newer memory vs an older conversation.

## Test plan
- [x] `bun run --cwd react-native test -- --runInBand App.test.tsx` (59 passing)
- [ ] Visual check on macOS Home: mixed conversation + memory rows newest-first

Do not merge until CI is green and Max explicitly approves.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12141?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/react-native/App.tsx
+++ b/react-native/App.tsx
@@ -1857,13 +1857,12 @@ function App({initialRoute}: AppProps): React.JSX.Element {
         ? readOutcomes.memories.value.items
         : []),
     ].sort((left, right) => {
-      const timestamp = (item: DesktopReadProjection) =>
-        item.kind === 'conversation'
-          ? Date.parse(item.startedAt ?? item.createdAt)
-          : item.kind === 'memory'
-            ? (item.timestamp ?? 0)
-            : 0;
-      return timestamp(right) - timestamp(left);
+      const leftTs = projectionTimestamp(left);
+      const rightTs = projectionTimestamp(right);
+      return (
+        (rightTs ?? Number.NEGATIVE_INFINITY) -
+        (leftTs ?? Number.NEGATIVE_INFINITY)
+      );
     });
   }, [readOutcomes]);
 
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified this end to end in a local checkout. The mixed-unit sort was real: memory projections carry epoch seconds (desktopReadClient.ts:481, note formatMemoryDate's *1000 at App.tsx:1000) while conversations parsed to milliseconds, so conversations always stacked above memories. Reusing projectionTimestamp (App.tsx:522) normalizes both to ms, and the Number.NEGATIVE_INFINITY fallback keeps the old 'missing time sorts last' behavior. The new regression test is behavioral, not a snapshot assertion: copied onto the parent commit it fails exactly at the ordering expectation (App.test.tsx:885) and passes on this head. Full react-native suite 59/59 and tsc --noEmit both clean locally. Heads-up on the red 'Validate backend worker' check: it is the known npm optional-dependency bug (npm/cli#4828, missing @rollup/rollup-linux-x64-gnu) and has failed the previous commits on this branch too, so it is not caused by this react-native-only diff. Nice, tight fix.

---

undivisible:
Closing as **superseded** per Max.

Tip `v5` @ `9051456` already sorts the macOS Home timeline via `projectionTimestamp` in `useDesktopReads.ts` (memories normalized with ×1000 in `desktopReadClient.ts`). This draft still patched the old monolithic `App.tsx` and is ~80 commits behind / CONFLICTING.

We are **not** rebasing the whole PR. Only the regression test (“orders macos Home timeline…”) will be ported onto tip as a follow-up.

PR #12152, opened 2026-08-24
fix(desktop-windows): document the Node 22.19+ prereq and pin it via .nvmrc

formed2forge:
## Summary
- `package.json`'s `engines.node` (`>=22.19.0 <23`) and `scripts/check-node-version.mjs`'s `pretest` hook already enforce this Node range, and CI pins `node-version: 22` — but nothing surfaced this to a contributor before they hit the pretest failure, or worse, silent jsdom `localStorage` breakage on Node 24+.
- Adds `.nvmrc` (`22.19.0`) so `nvm use` picks the right version automatically.
- Documents it as step 1 of README's "Run from source".

## Test plan
- [x] Verified `package.json` engines range and `scripts/check-node-version.mjs` match `.nvmrc`
- [x] Verified CI (`desktop-windows-ci.yml`) pins `node-version: 22` in all three jobs

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12152?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks @formed2forge — verified the whole diff against this head, and every claim checks out:

- `desktop/windows/.nvmrc` = `22.19.0` satisfies `package.json`'s `engines.node` (`>=22.19.0 <23`) and matches the range enforced by `scripts/check-node-version.mjs` (`RANGE = '>=22.19.0 <23'`), so `nvm use` lands on a supported version. It's also the only `.nvmrc` in the repo tree, so there's no other pin to drift from.
- The README's Node 24+ warning matches the failure mode documented in `scripts/check-node-version.mjs` (Node ≥ 24 ships its own `localStorage` global that shadows jsdom's and is `undefined` without `--localstorage-file`).
- CI does pin `node-version: 22` in all three jobs of `desktop-windows-ci.yml`, and `--frozen-lockfile` matches the install steps there and in `desktop_windows_release.yml`.

One small nit, not blocking: the second commit message says AGENTS.md's Development Workflow section already specifies `--frozen-lockfile`, but no AGENTS.md at this head mentions it — just so nobody greps for that later and comes up empty.

The only thing blocking merge is the red Hygiene check, and it's a metadata fix rather than a code one. Both commit subjects start with `fix:`, so the repo's failure-class protocol requires a `Failure-Class:` declaration line in the PR body — the PR template ships with `Failure-Class: none` as the default, and it looks like that line was dropped when the description was regenerated. Since this is a docs/pinning change with no semantic failure class, adding this line to the PR body should clear it:

```
Failure-Class: none
```

`scripts/pr-preflight --suggest` will print the paste-ready line if you prefer the tool to produce it. Editing the body re-runs PR Metadata Preflight, which is what validates the declaration — no new commit needed. Once that's green, this looks good to merge from my side.

---

kodjima33 (approved):
Desktop bug-fix fast lane (owner-override): documents real Node version mismatch breaking jsdom tests, pins via .nvmrc. Scoped, CI clean.

PR #12161, opened 2026-08-24
fix(web): report signed-out state when Firebase is unconfigured

igor-popov-dev:
## What changed and why

`onAuthStateChange` in `web/app/src/lib/firebase.ts` returned a bare unsubscribe when no Firebase app had been created, so the callback never ran. `AuthProvider` only leaves `loading` from that callback, and `ProtectedRoute` renders a full-screen spinner for the whole of `loading`. A build without Firebase credentials therefore boots to a blank page on **every** route — including `/login`, which already ships a "Sign-in is not configured in this local preview." message for exactly this case, and which `lib/firebase.ts` already recognises via the `preview` / `preview.local` placeholders.

No Firebase app means no session can exist, and that is an answer the subscriber is owed. The fix emits `null` once, then hands back the same no-op unsubscribe. Nothing changes when Firebase is configured — that path still delegates to `onAuthStateChanged`.

This came up on Discord: a contributor wanted to fix UI bugs on `/tasks` and could not get `web/app` to run at all without the Omi Firebase config.

none (`scripts/pr-preflight --suggest` reports no invariant path matches for this diff)

## How it was verified

Exercised the real user-facing path in a clean worktree off `main`, with `.env.local` carrying the placeholders the code already recognises (`NEXT_PUBLIC_FIREBASE_API_KEY=preview`, `NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN=preview.local`, every other `NEXT_PUBLIC_*` set to a non-empty string):

```bash
cd web/app && bun install && bun run build && PORT=3100 bun .moonshine/server.ts
```

- **Before:** `/login` and `/tasks` both render only the spinner — DOM stays at `<div class="... animate-spin">`, no console errors, so it looks like a silent hang.
- **After:** `/login` renders with sign-in disabled and the existing "Sign-in is not configured in this local preview." message; navigating to `/tasks` redirects to `/login` as `ProtectedRoute` intends.

Gates:

```bash
cd web/app && bun run check     # typecheck clean; 5 moonshine smoke + 331 vitest tests pass
make preflight                   # 15 manifest checks pass (incl. web-app-checks)
```

## Tests

`web/app/src/components/auth/__tests__/AuthProvider.test.tsx` — renders the real `AuthProvider` over the real `lib/firebase.ts`, mocking only the Firebase SDK and the `NEXT_PUBLIC_*` values, so the assertion runs production behavior rather than a source scrape.

- Preview placeholders → the provider settles to `loading: false` / `user: null`, and `onAuthStateChanged` is never called.
- Real config → the provider reports the signed-in user and subscribes exactly once.

Confirmed it is a regression test, not a companion: with the fix stashed, the first case fails (`waitFor` times out waiting for `settled:anonymous`); with the fix, both pass.

## Failure class (fixes)

Failure-Class: none

## Note on docs

`web/app/AGENTS.md` would be the natural place to write down "how to run this without Firebase credentials", but the file sits at 2367/2400 bytes and 55/55 lines against the `agents-md-lean` budget, and that check is a ratchet. Rather than shrink unrelated prose inside a bug fix, I left it out — happy to follow up with a docs-only PR that moves the setup detail into `docs/agents/` and links it, if maintainers want that.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/app/src/components/auth/__tests__/AuthProvider.test.tsx
+++ b/web/app/src/components/auth/__tests__/AuthProvider.test.tsx
@@ -1,0 +1,129 @@
+import { createElement } from 'react';
+import { cleanup, render, screen, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * `AuthProvider` starts at `loading: true` and only leaves it when the auth
+ * subscription reports a state. `ProtectedRoute` renders a full-screen spinner
+ * for the whole of `loading`, so a subscription that never reports is a blank
+ * app, not a degraded one.
+ *
+ * That is the shape of a local preview build: `.env.local` carries the
+ * `preview` / `preview.local` placeholders `lib/firebase.ts` recognises, no
+ * Firebase app is created, and every route — including `/login`, which ships a
+ * "Sign-in is not configured in this local preview." message for exactly this
+ * case — sits on the spinner.
+ */
+
+const PREVIEW_ENV = {
+  NEXT_PUBLIC_FIREBASE_API_KEY: 'preview',
+  NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN: 'preview.local',
+  NEXT_PUBLIC_FIREBASE_PROJECT_ID: 'preview',
+  NEXT_PUBLIC_FIREBASE_STORAGE_BUCKET: 'preview',
+  NEXT_PUBLIC_FIREBASE_MESSAGING_SENDER_ID: 'preview',
+  NEXT_PUBLIC_FIREBASE_APP_ID: 'preview',
+};
+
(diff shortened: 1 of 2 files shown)


kodjima33 (approved):
Confidence 5/5: clear root cause (Firebase-unconfigured auth never resolves loading state), scoped diff, dedicated test added, CI green. Verified not already fixed on main.

PR #12171, opened 2026-08-25
fix(desktop-windows): correct pi-mono's packaged dependency versions

formed2forge:
## Summary

`@earendil-works/pi-coding-agent` (the AI chat/assistant's spawned agent process) needs newer versions of several transitive deps than what other, dev-only tooling (electron-builder's own toolchain, eslint) pulls in as the same package name. pnpm's hoisted `node_modules` keeps both versions side by side — one flat at top-level, one nested under `pi-coding-agent` — but electron-builder's asar packing collapses same-named `node_modules` entries to a single path and, empirically and reproducibly, always keeps the top-level (wrong, older) copy. That silently broke the packaged app's AI chat at startup with a chain of `ERR_MODULE_NOT_FOUND` / "Named export not found" errors as each affected package's modern API was needed.

Fixed `jiti`, `glob`, `minimatch`, `hosted-git-info`, `ignore`, and `semver` by pinning each to pi-coding-agent's required version via a pnpm override (safe: every other consumer of the old version is dev/build-time-only tooling that never ships in the packaged app, or in electron-updater's case a same-major minor bump). `chalk` needed a different fix: forcing it to pi-coding-agent's 5.6.2 (ESM-only) broke electron-builder's own CJS `require('chalk')` at build time (confirmed: `TypeError: chalk.underline is not a function`), so it's corrected post-pack instead — a new `afterPack` hook (`scripts/fix-pimono-chalk-unpack.mjs`) copies pi-coding-agent's own nested `chalk` over whatever electron-builder chose for the packaged `app.asar.unpacked/node_modules/chalk`.

## Test plan

Verified end-to-end against a real packaged Linux build, not just reasoned about: spawned the actual packaged pi-coding-agent CLI the same way `src/main/codingAgent/piMono.ts` does (`ELECTRON_RUN_AS_NODE` + `--mode rpc`), before and after each fix.

- Before: crashed on `jiti`, then `glob`, then `minimatch`, then `chalk` in sequence as each was fixed.
- After: the extension loads successfully (`[omi-tools] adapter=pi-mono advertisedToolCount=17...`) and the only remaining error is the expected "no API key supplied" (this test didn't provide one), confirming the entire module graph now loads correctly.

Additionally ran in this PR: `pnpm typecheck` (via `npm run typecheck` to route around this sandbox's registry-restricted postinstall check) — clean, no errors. `pnpm lint` — clean.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12171?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/changelog/unreleased/2026-08-pimono-packaged-version-collisions.json
+++ b/desktop/windows/changelog/unreleased/2026-08-pimono-packaged-version-collisions.json
@@ -1,0 +1,5 @@
+{
+  "changes": [
+    "Fixed the AI chat/assistant crashing on startup in packaged builds with module-not-found errors (confirmed on Linux; the same packaging step is shared by Windows and macOS) — several of its bundled dependencies (jiti, glob, minimatch, chalk, hosted-git-info, ignore, semver) were silently getting swapped for an incompatible older version during packaging."
+  ]
+}
--- a/desktop/windows/electron-builder.config.mjs
+++ b/desktop/windows/electron-builder.config.mjs
@@ -20,6 +20,7 @@
 
 import { computeUnpackGlobs } from './scripts/gen-pimono-unpack.mjs'
 import { KGWORKER_NATIVE_UNPACK_GLOBS } from './scripts/kgworker-native-closure.mjs'
+import fixPimonoChalkUnpack from './scripts/fix-pimono-chalk-unpack.mjs'
 
 // Fresh at every pack: the closure is recomputed from the installed node_modules,
 // so it can never be stale relative to what is actually on disk.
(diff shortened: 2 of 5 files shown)


kodjima33 (approved):
Desktop bug-fix fast lane (owner-override): fixes packaged-build dependency version collisions (chalk/jiti/etc silently swapped during packaging, crashing chat on startup). CI clean.

PR #12172, opened 2026-08-25
fix(desktop-windows): run the bytecode export-forward guard on Linux/macOS builds too

formed2forge:
## Summary
`build:win` already runs `ensure-env.mjs` → `patch-bytecode-entry-forward.mjs` → `verify-bytecode.mjs` via `npm run build` (fixed in 93b3cee7ad, "forward the bytecode entry's exports"). `build:mac` and `build:linux` were never updated to go through the same `npm run build` step — they called `electron-vite build` directly — so the packaged main-process bytecode stub on those two platforms never gets `module.exports =` added. Any lazily-loaded chunk calling back into `require("../index.js").<name>` silently reads `{}` on Linux/macOS packaged builds, the exact class of bug 93b3cee7ad root-caused and fixed for Windows only.

Fix: point `build:mac` and `build:linux` at `npm run build` instead of a bare `electron-vite build`, so all three platform targets share the same build pipeline instead of Windows carrying the fix alone.

## Test plan
- [x] Confirmed by hand: an un-patched build fails `verify-bytecode.mjs`'s forwarding check; the same build after `patch-bytecode-entry-forward.mjs` passes clean.
- [x] `pnpm typecheck` passes

Failure-Class: FC-platform-contract-left-as-a-local-comment

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12172?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
Desktop bug-fix fast lane (owner-override): extends the Windows bytecode-export-forward fix to Linux/macOS build scripts, fixing silent assistant failure there too. Tiny, CI clean.

PR #12173, opened 2026-08-25
fix(desktop-windows): treat pi-mono's agent_settled as advisory, not unknown

formed2forge:
## Summary
`PiMonoAdapter.handleEvent`'s dispatch switch had no case for the `agent_settled` protocol event, so it fell through to the `default` branch and logged `[pi-mono] unknown event type: agent_settled` — even though the adapter already treats several other protocol-control events (`agent_start`, `agent_end`, `turn_start`, etc.) as observed-but-ignored no-ops. `agent_settled` is an advisory upstream event; `turn_end` is what actually resolves or rejects the pending prompt via `handleTurnEnd`, and that behavior is unchanged.

Added an explicit no-op case alongside the others, plus a test exercising `handleEvent`'s real JSON-string dispatch path (not just `handleTurnEnd` directly) confirming a prompt stays pending across `agent_settled` and only completes on the subsequent `turn_end` — matching the equivalent coverage `desktop/macos`'s pi-mono adapter already has.

No functional/user-visible behavior changed (the pending prompt already correctly waited for `turn_end` either way); this only removes a spurious stderr log line and closes a Windows/macOS test-coverage gap.

## Test plan
- [x] New test: `treats agent_settled as advisory and waits for the authoritative turn_end`
- [x] `pnpm typecheck` clean
- [x] `pnpm test` — full suite passes (5,411 passed / 27 skipped)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12173?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/codingAgent/piMono.test.ts
+++ b/desktop/windows/src/main/codingAgent/piMono.test.ts
@@ -58,6 +58,7 @@ function newFakeProc(): FakeProc {
 // and inspect internal state without `any` (lint forbids explicit any).
 interface PiInternals {
   sendCommand: (cmd: Record<string, unknown>) => void
+  handleEvent: (line: string) => void
   handleTurnEnd: (event: Record<string, unknown>) => void
   handleToolStart: (event: Record<string, unknown>) => void
   handleToolEnd: (event: Record<string, unknown>) => void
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Verified on head 387939a — clean, well-scoped fix. Thanks!

- `desktop/windows/src/main/codingAgent/piMono.ts` — the new `case 'agent_settled'` no-op (with the "turn_end remains authoritative" comment) correctly stops the fall-through to the `default:` branch that logged `[pi-mono] unknown event type: agent_settled` to stderr. Nothing else in `handleEvent` acts on this event type, so the change is log-suppression only: prompt completion still flows exclusively through `case 'turn_end'` → `handleTurnEnd`, consistent with how `auto_retry_*` and the other protocol-control events are already observed-but-ignored.
- `desktop/windows/src/main/codingAgent/piMono.test.ts` — the new test drives the real JSON-string dispatch path (`handleEvent(JSON.stringify({ type: 'agent_settled' }))`) rather than calling `handleTurnEnd` directly, and asserts both that the prompt stays in flight (`activePromptGeneration` unchanged, `pendingRequests.size === 1`) and that only the subsequent `turn_end` resolves it. That's the right level to pin this behavior at.

Also confirmed the macOS adapter already ships the identical advisory case and an identically-named test, so this closes the Windows-side parity gap exactly as the description claims. `Typecheck · Lint · Test` and `Windows build smoke` are green on this head.

No blocking concerns — leaving for human maintainer review before merge.

---

kodjima33 (approved):
Desktop bug-fix fast lane (owner-override): treats pi-mono's agent_settled as advisory instead of unknown, with regression test. CI clean.

PR #12175, opened 2026-08-25
fix: repair 13 setup instructions that name files the repo does not ship, and guard the class

falconzyx:
## Summary

Thirteen places where a shipped instruction, make target, or CLI option named a file that does not exist — plus the guard extension that makes the largest group of them impossible to reintroduce.

Found by auditing the repo's setup paths end to end, then re-verifying every finding against current `main`. Two suspected defects turned out to be misreads and were dropped rather than "fixed"; two more were verified as real and deliberately left alone. All four are listed at the bottom.

## What changed

| Commit | Defect |
|---|---|
| `docs: point setup instructions at env templates that exist` | `README.md` and `desktop/macos/run.sh` sent readers to `backend/.env.example`, which has always been `.env.template`. Thirteen plugin setup docs told the reader to copy a `.env` template that is not there. `web/app/.env.template` shipped empty Firebase values and named the wrong target file, so `bun run dev` failed its build and never spawned a server. `web/frontend/README.md` told the reader to run a `docker-compose` file this repo does not contain. |
| `docs(backend): load the pusher env file the setup step just created` | `--env-file .env` resolved to `backend/.env`, not the `backend/pusher/.env` the reader had just been told to create. Also demotes `HUGGINGFACE_TOKEN` from Required to Optional, with the reason it is absent from `.env.template`. |
| `docs(harness): name DESKTOP_APP_NAME in the local desktop instructions` | The local profile exits 2 on the default bundle, so `make dev-desktop` could not run as printed in the runbook or in `dev-init.sh`'s closing hint. |
| `docs(mcp): stop claiming release.sh publishes to PyPI` | The publish lines are commented out. `.env.template` also pointed at a stale Cloud Run URL. |
| `fix(harness): remove the dead run-canonical-promotion make target` | Invoked a script absent from the worktree and the index; `165c041a93` deleted that pipeline deliberately. |
| `fix(app): drop the nonexistent -t entrypoint from the release scripts` | `-t lib/main_prod.dart` does not exist; every invocation aborted. |
| `fix(firmware): correct the build directory and drop the undefined uf2 env` | `west build ../app` should be `../devkit`; `uf2_release` is not a PlatformIO environment, yet `build_uf2.py` listed it in argparse choices. |
| `feat(ci): catch setup docs that copy a template that does not exist` | The guard. |

## The guard

Every doc fix above is one failure class: a doc tells you to copy a template and the template is not there. `agent-doc-references` already enforces "a pointer must resolve", but could not see these — setup instructions live in fenced blocks or in prose, not in markdown links and backticked paths, and the guard only read `AGENTS.md` plus four named files.

This **extends that existing guard rather than adding a second one**, because it is the same contract. Scope is deliberately narrow so the rule needs no judgment and no allowlist:

- only docs that sit in the directory they document — a doc under `docs/` describes commands run elsewhere and its working directory lives in prose the checker cannot read (scanning those produced false positives on `Backend_Setup.mdx` and `omiGlass.mdx`, which are correct as written);
- only `.template` / `.example` sources;
- each `cd` moves the working directory for the commands that follow it, per line and per `&&` segment — not once per block, which would resolve a later `cp` against a stale directory;
- an absolute or `~` `cd` stops the tracking rather than guessing, so the verdict never depends on the host's filesystem;
- unlike the reference rules, this one does **not** consult gitignore. A template is by definition committed, so a `.gitignore` that hides one is the bug rather than the excuse — `plugins/omi-ms365-app/.gitignore` re-ignores `.env.*` and would otherwise switch the rule off for that whole directory.

**Measured, not assumed.** Against `main` at `36c467fd2b` the rule reports **exactly the 14 real instances and nothing else** — 9 written as `cp`, 5 as prose; against this branch it reports **0**. Both runs scan the same 270 in-scope docs (542 tracked `.md`/`.mdx` minus `docs/` and vendored trees).

Two of those 14 — `plugins/omi-github-app/QUICKSTART.md` and `plugins/omi-slack-app/SETUP.md` — were found *by the check* after a manual sweep had already declared the plugin docs clean.

`self_test` covers the rule's load-bearing decisions in 15 assertions: sequential `cd`s, `&&` segments, a stale first `cd` masking a real defect, the two spellings of one `cd` target agreeing, an absolute `cd`, a gitignored template, both prose and command forms, and a `docs/` file that makes the out-of-scope assertion actually discriminate instead of passing vacuously. The `&&` case asserts on a *missing* template on purpose — with a present one, deleting the segment split still yields `[]` and the test cannot see its own mutation.

A broader version was measured and rejected: applying the existing link/backtick rules to every markdown file reports 1019 violations, 790 under `desktop/`. That is a migration, not a guard.

## Verification

```
# the guard, against unmodified main vs this branch
PRE-FIX  (main 36c467fd2b) : 14 violations  <- 9 cp-form + 5 prose-form, nothing else
POST-FIX (this branch)     :  0 violations  <- same 270 in-scope docs

python3 .github/scripts/check_agent_doc_references.py  -> agent doc references OK
                                                          (self_test: 15 assertions)

git grep 'backend/\.env\.example' -- '*.md' '*.mdx'   -> no hits (1 on main: the defect)
  (a plain repo-wide grep returns 3, all added here: the checker docstring,
   its self-test fixture, and the manifest reason string)
grep -rn 'PROMOTION_USER|run-canonical-promotion'      -> no hits repo-wide
grep '^\[env' omiGlass/firmware/platformio.ini         -> only the two surviving envs
ls omi/firmware/app                                    -> No such file or directory
git ls-files | grep docker-compose                     -> nothing

bash -n desktop/macos/run.sh app/release.sh \
        scripts/dev-harness/dev-init.sh \
        omiGlass/firmware/scripts/build_uf2.sh         -> all pass
python3 -m py_compile omiGlass/firmware/scripts/build_uf2.py -> pass

make -n run-canonical-maintenance                      -> still resolves
make -n run-canonical-promotion                        -> No rule to make target (intended)
pytest scripts/dev-harness/tests                       -> 99 passed, 7 skipped

# web/app defect reproduced and confirmed fixed
node scripts/generate-firebase-sw.js (stock template)    -> exit 1
node scripts/generate-firebase-sw.js (filled .env.local) -> exit 0, correct initializeApp block

# manifest trigger selection, after switching **/*.md -> *.md
README.md selected=True   CONTRIBUTING.md selected=True   plugins/x/README.md selected=True
```

Not exercised: `make dev-up`. The contributor's machine has Python 3.14 against the 3.11.15 pin and no Redis, JDK or Docker. Nothing here touches harness runtime behavior; the only `Makefile` edit deletes a target that already could not run, and the dev-harness unit suite passes.

## Gate status

Every check this diff selects was run individually (`run_checks.py --check-id`): **21 of 22 pass**.

The one that does not is `agents-md-lean`, and only on `backend/AGENTS.md: 39120 bytes exceeds budget of 39000` — **pre-existing on `origin/main` and untouched here**, so `main` is already red on that check. `web/app/AGENTS.md` was briefly over budget from this branch and is back under (55 lines / 2396 bytes).

Separately, `desktop-beta-admission-firestore-contention` fails with *zero output* on a Mac with no JDK, which cost some time to diagnose and is filed as #12178. It is not selected by this diff.

`desktop-changelog-entry` passes via a `kind: none` fragment: the only desktop file touched is a first-run help string, which is not user-facing.

## Product invariants

`scripts/pr-preflight --suggest` reports: **none**.

## Failure classes

Declared: `none`. No registered class covers a dead setup pointer.

Registering one for this class needs `evidence_prs` to be *merged* PRs, and the evidence is this PR — so the class cannot be created in the change that evidences it. The mechanical guard lands now; registering the class, with this PR as its evidence and the extended checker as its `canonical_prevention_artifact`, is a follow-up registry-only change.

## What was deliberately left out

A fix for `web/frontend`'s format gate — `npm run lint:format -- --check` appends `--check` to a `prettier --write`, so the gate has never been able to fail and silently rewrites its own checkout — was written, verified, and then **removed from this PR**.

It is a clean instance of `FC-nonexecuting-verification-command`. Declaring that class truthfully crosses the guard-artifact ratchet, whose only correct remedy is the reusable guard the class's own `canonical_prevention` already names, and that is not a stale-pointer cleanup. Declaring `none` for it instead would have made this PR's declaration false — and the ratchet would then have passed *only because the declaration was false*. Both are filed: #12176 for the gate itself, #12177 for the guard artifact the class needs.

## Rejected during review

Two misreads, dropped rather than "fixed":

- `make dev-verify` does **not** require `DESKTOP_APP_NAME`. `verify-desktop-local-launch.sh` never calls `validate_profile`; a missing name skips one optional bundle-identity assertion rather than failing the run.
- `docs/rayban-meta-dat-setup.md`'s `cp RayBanMetaCredentials.xcconfig.template ...` resolves correctly — the block's `cd app/ios/Flutter` sets the working directory, and the template exists.
And one real defect left alone on purpose:

- `app/initialsetup.bash:9`'s `cd apps/AppWithWearable` is genuinely dead — no `apps/` directory exists anywhere. But repairing it would turn a visibly-broken script into a runnable one that writes legacy env files, which `scripts/check-public-client-secrets.py --strict` flags. Deletion is the right call and belongs in its own change.

(The fourth is the `web/frontend` format gate under "What was deliberately left out".)

Failure-Class: none


falconzyx:
Closing — opened prematurely on my side. Keeping this work local for now; reopening or refiling later if it is wanted.

Issue #12182, opened 2026-08-25
gcp-sa-key-ratchet breaks manifest resolution for every PR, and has never passed its own baseline


aryanorastar:
`gcp-sa-key-ratchet` landed in #11125 with two defects that compound. The first hides the second, and fixing only the first makes things worse, so I am reporting both together rather than sending a one-line patch.

This is my own merged PR. It went in green because neither defect can fire until the check is actually reachable, and it was not.

## Defect 1 — dead trigger paths break manifest resolution for every PR

The check declares two trigger paths that do not exist in the repo:

```
backend/agent-proxy/main.py
backend/agent-proxy/Dockerfile
```

`run_checks.py:187` existence-checks any trigger pattern without glob magic and fails the **entire** resolution when one is missing:

```python
elif pattern != "all" and not glob.has_magic(pattern) and not (root / pattern).exists():
    errors.append(f"{check.id}: explicit trigger path does not exist: {pattern}")
```

Resolution happens before any check runs, so this does not fail the GCP ratchet — it fails every consumer of the manifest. Observed on #12181, which touches three desktop Swift files and no backend at all:

```
FAIL: could not resolve manifest checks: gcp-sa-key-ratchet: explicit trigger path does not exist: backend/agent-proxy/main.py; gcp-sa-key-ratchet: explicit trigger path does not exist: backend/agent-proxy/Dockerfile
```

It took out `Detect Desktop Swift Changes`, `Desktop Swift Build & Tests`, `Hygiene`, and `PR Metadata Preflight`.

**Reproduces on pristine `main`, no branch involved:**

```bash
git archive origin/main | tar -x -C /tmp/mainmanifest
cd /tmp/mainmanifest && git init -q . && git add -A && git commit -qm base
python3 .github/scripts/run_checks.py --lane ci --list
```

A full audit of the manifest found these are the only two broken explicit triggers out of 152 checks.

## Defect 2 — the ratchet has never passed against its own baseline

Defect 1 masks this completely: the check cannot run, so nobody saw that it does not pass. Collapsing the dead paths into `backend/agent-proxy/**` makes resolution succeed, the ratchet finally executes, and it fails:

```
backend/database/google_credentials.py:service-account-json-env: found 6, baseline allows 1
.github/workflows/desktop_qualify_beta.yml:credentials-json-gh-action: baseline still allows 1 but tree only has 0; shrink the baseline in this change
[...]

PR #11864, opened 2026-08-19
feat(desktop): withhold proactive notifications while other people are present

aryanorastar:
## Summary

A proactive notification is addressed to one person. Omi delivered them identically whether the user was alone, in a call, or presenting to a room. A live session put **"Submit prototype for SBI Hackathon before the deadline"** on screen — useful when alone, a disclosure on a shared screen, and an interruption mid-meeting.

Two distinct harms, one rule:

- sharing a screen makes a private nudge **visible** to everyone on the call
- being in a call at all makes it **interrupt a conversation**

I first scoped this to screen share only, reasoning that being on a call is not the same as your screen being visible. Testing against a live Google Meet call disproved that: `"Meet is fine — but you said you'd submit the SBI Hackathon prototype"` was delivered mid-meeting. Detection now covers both.

## Where the guard sits

`NotificationService.sendNotification` is the single choke point every proactive surface already routes through — suggestion, memory, insight, goals, meeting action items, plugin (9 call sites) — so one guard covers all of them rather than a per-assistant exception.

`respectFrequency` is the existing proactive/functional split and is honoured: functional notices (screen-recording repair prompt, Crisp replies, onboarding test) pass `false` and still reach the user. Suppressing the capture-repair prompt during a share is precisely how a broken capture would stay broken, since that prompt is what tells the user to fix it.

Detection reuses three signals that already exist and are already trusted in production:

| Signal | Catches |
|---|---|
| `activeScreenSharePresent()` | outgoing share (already used to pause capture during shares, #10143) |
| `callAppIsUsingMicrophone()` | an active call |
| `browserCallWindowPresent()` | a **muted** browser call, where mic input has dropped |

The third is what caught the Meet case above. Detection is placed after the cheap boolean gates so the window scans never run for a notification an earlier gate already refused.

## Withheld, not destroyed

This is the part worth reviewing closely.

`SuggestionAssistant` writes `recentSuggestions` immediately **before** delivering, and that window gates every later evaluation. Withholding only at the choke point would have recorded the suggestion as delivered and retired it permanently: the user never sees the card, and every regeneration after the call is filtered as a duplicate of something that was never shown.

The assistant therefore consults the same shared policy *before* the dedup write and returns early, leaving the suggestion eligible once the call ends. `suppressed_presenting` is a distinct delivery outcome from the `filtered_*` ones: those retire a suggestion on its merits, this defers a good one on audience.

## Verification

[...]

PR #11183, opened 2026-08-06
feat(backend): split X connector sync into its own Cloud Run Job (#9298)

aryanorastar:
## Summary

- Move X connector incremental sync off `notifications-job` onto a dedicated `x-connector-sync-job` (entrypoint, Dockerfile, runtime-env contract, deploy workflow).
- Make Scheduler the single cadence owner (`x-connector-sync-6h`); the entrypoint no longer uses hour-modulo gating.
- Fail the Cloud Run execution on registry/listing failures or total per-user failure instead of reporting a healthy no-op.
- Preserve the existing scheduled-memory contract: construct one `PromotionFlexRunRouter` per job run, reuse it for every connected user, and leave intentional Flex capacity deferrals pending without counting them as sync failures.
- Add the runbook and Cloud Run Jobs checklist/scaffold so future cron work does not return to the notifications process.

Fixes #9298

## Why the Flex follow-up matters

While this PR was open, main added the shared Flex router to scheduled X memory extraction. The job split retained the Flex-aware per-user implementation but its rewritten scheduler boundary stopped passing the router, which made scheduled extraction silently use the legacy standard lane. The latest commit restores the current production routing contract without weakening this PR's new Cloud Run failure surface.

## Verification

- [x] Focused X scheduler and failure-surface suites — **16 passed**.
- [x] X + runtime-env + deployment-contract suites — **162 passed**.
- [x] New behavioral regression drives `run_x_sync_job()` through two registry users and proves:
  - exactly one router is built with the job start time;
  - the same router reaches every `sync_x_for_user()` call;
  - a `PromotionFlexDeferred` user remains retryable and does not increment `failed`.
- [x] Full `backend/test.sh` sweep exercised all **992 test files**; both X suites passed. Five unrelated current-main files remain red and are byte-identical to `origin/main` in this branch (`test_legacy_memory_surface_inventory.py`, `test_modulate_stt.py`, `test_streaming_deepgram_backoff.py`, `test_async_app_integrations.py`, `test_async_realtime_integrations_offload.py`).
- [x] Runtime image source-closure/import-smoke coverage includes `x-connector-sync-job`.
- [ ] Maintainer operational follow-up after merge: create/verify the Scheduler job and perform the first deploy via `gcp_x_connector_sync_job.yml`.

## Failure class

Failure-Class: none

The change is an ownership split plus a review-time regression repair; it is not a shipped recurring failure-class instance.

## Product invariants affected

[...]

PR #12183, opened 2026-08-25
revert(ci): back out the GCP SA key ratchet, keep its setup-doc fix

aryanorastar:
Reverts #11125. Details and reproduction in #12182.

## Why

The ratchet shipped with two defects that compound, and the first hides the second.

**1. Dead trigger paths break manifest resolution for every PR.** It declares `backend/agent-proxy/main.py` and `backend/agent-proxy/Dockerfile`; neither exists. `run_checks.py:187` existence-checks any trigger without glob magic and fails the *whole* resolution when one is missing — before any check runs:

```
FAIL: could not resolve manifest checks: gcp-sa-key-ratchet: explicit trigger path does not exist: backend/agent-proxy/main.py; gcp-sa-key-ratchet: explicit trigger path does not exist: backend/agent-proxy/Dockerfile
```

So it breaks every consumer of the manifest, not the ratchet alone. #12181 touches three desktop Swift files and no backend, and lost `Detect Desktop Swift Changes`, `Desktop Swift Build & Tests`, `Hygiene`, and `PR Metadata Preflight` to it.

**2. The ratchet has never passed its own baseline.** Defect 1 masks this — the check cannot run, so nobody saw it fail. Fix the paths and it executes for the first time and fails on pristine `main` with 13 findings. Twelve are stale-shrink (baseline allows N, tree has 0), several of them for `backend/agent-proxy/*` — the same nonexistent paths, so the baseline was generated against a tree this repo does not have. One is not a shrink:

```
backend/database/google_credentials.py:service-account-json-env: found 6, baseline allows 1
```

That is the ratchet's real failure mode — new key paths — firing on a file `main` has not modified. The cap was wrong when written rather than something having regressed since.

This is why I am reverting rather than sending the one-line trigger-path fix: on its own it converts a broken-manifest error into a **red credential check on every PR**, which reads as a real security finding and is worse than the current state.

`--write-baseline` "never raises existing caps" and refuses to add keys or raise counts, so `found 6, allows 1` cannot be corrected with the tool. It needs either removing the key paths from `google_credentials.py` (a real move toward ADC / Workload Identity, which is the ratchet's stated purpose) or a baseline edit under security review. Both belong to someone who owns that area.

## What this costs

Nothing that currently works. The check has never executed, so no key path is protected by it today. Reverting restores green immediately.

## Re-landing

Straightforward once the baseline question is settled: generate the baseline against this tree, and use `backend/agent-proxy/**` as the trigger — `run_checks.py:187` exempts globs from the existence check, so the surface stays covered if agent-proxy ever lands, without asserting it exists now. Happy to do that re-land.

## Deliberately kept

The `docs/doc/developer/backend/Backend_Setup.mdx` change is **restored verbatim**, not reverted. It replaces this:

```bash
cp ~/.config/gcloud/application_default_credentials.json ./google-credentials.json
```

with ADC guidance and an explicit "do not copy a service-account JSON key into the repo". That improvement has no CI dependency and stands on its own; reverting it would put the copy-a-key-into-the-repo instructions back in the setup guide.

## Verification

- Manifest audit on this branch: **0** broken explicit trigger paths, **0** missing command paths, 150 checks (was 152), 0 duplicate ids.
- `run_checks.py --lane ci --list` resolves clean.
- `grep` for `gcp_sa_key_ratchet` / `gcp-sa-key-ratchet` across the tree: **0** remaining references, so nothing dangles.
- Diff is 4 files, deletions only; `Backend_Setup.mdx` is byte-identical to its post-#11125 state.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aryanorastar:
@undivisible five PRs are sitting with no approving review. Consolidating into one comment rather than pinging five threads.

Thanks for the pass on #11864 earlier — that's why I'm asking you rather than adding to the queue elsewhere.

### Please look at this one first — it unblocks everything else

**#12183** — reverts #11125, which is mine and merged broken this morning.

It declared two trigger paths that don't exist, and `run_checks.py:187` fails the *entire* manifest resolution on a missing non-glob trigger, before any check runs. So it reddens every PR, not just its own check. Reproduces on a pristine `main` checkout with no branch involved.

Underneath that, the ratchet has never passed its own baseline — 13 findings on clean `main`, including `google_credentials.py:service-account-json-env: found 6, baseline allows 1` on a file `main` hasn't modified. `--write-baseline` refuses to raise caps by design, so that one needs a real decision, not a patch. Full writeup in #12182.

Reverting costs nothing that works today: the check has never executed. The `Backend_Setup.mdx` fix is kept verbatim, since reverting it would put `cp ~/.config/gcloud/application_default_credentials.json ./google-credentials.json` back in the setup guide.

### The other four

| PR | What it is | State |
|---|---|---|
| **#12181** | on-device transcription closes a window on the speaker's pause instead of a fixed 10 s boundary — mean latency 4.97 s → 0.74 s, spread 6.6 s → 0.5 s, measured on a real mic | red **only** from the #12183 bug above; verified locally, 18/18 tests, preflight 22/22 |
| **#11804** | transcript-driven proactivity, dogfood-flagged hard-off on every shipped bundle | green, zero outstanding technical items |
| **#10960** | `fix(ci)`: retire superseded Windows release sync PRs | green, open since **Aug 1** |
| **#11183** | `feat(backend)`: split X connector sync into its own Cloud Run Job | green, open since **Aug 6** |

#10960 and #11183 both had a review that was later dismissed, so they've been sitting with nothing on them since.

**On #11804 specifically** — worth knowing before you open it: @Git-on-my-level has written "no blocking technical issues found" across five passes, but those were issue comments rather than submitted reviews, so none of them count toward the approving-review gate. The only formal review on it is a `COMMENTED` from Aug 18. It looks reviewed and isn't.

I'm not asking anyone to rubber-stamp. #11804 has a real product question attached — unprompted speech-triggered responses — that's genuinely a human call and I'd rather have a "no" on it than more silence. #12183 and #12181 are the ones where I think the path is clear.

<sub>#12183 has two long-running desktop compiles still in flight as I write this; nothing failing.</sub>

PR #12185, opened 2026-08-25
fix(desktop): keep quiet agent turns alive

### LINKED ISSUE #11458: Desktop chat: 60s bridge-silence watchdog kills turns with partial_response=false, losing the turn entirely
Desktop chat turns intermittently die to the generic bridge-silence watchdog having produced **no output at all**, so the turn is lost rather than truncated.

Observed on a dev build off current `upstream/main` (macOS 27.0, Xcode 26.6), during 11 consecutive chat turns:

```
15:57:14  ChatProvider: generic watchdog fired after 60s of silence — bridge is stuck; force-resetting
15:57:14  AgentRuntimeProcess stderr: [agent] Interrupt requested by user
15:57:14  AgentRuntimeProcess stderr: [pi-mono] turn_end ERROR: Request was aborted
15:57:14  Chat telemetry event=chat_agent_error duration_ms=65231 error_class=timeout
          partial_response=false surface=main_chat harness=piMono watchdog_fired=true
```

The watchdog is `ChatProvider.genericWatchdogInactivityMs = 60_000` (`desktop/macos/Desktop/Sources/Providers/ChatProvider.swift:1252`, polled every 5s at `:1253`, fires at `:4351`).

Two details make this worth separating from ordinary slowness:

- **It is an inactivity timer, not a duration cap.** Other turns in the same session ran longer and completed fine — the longest success was `duration_ms=51850`. So a turn that streams steadily survives; this one went quiet for a full 60s.
- **`partial_response=false`** — nothing streamed before it died, so the user loses the turn entirely rather than seeing a truncated answer.

Frequency in that session: 2 of 11 turns. The other 9 completed (12.8s–27.6s of total turn time including tool rounds).

Not the same failure as the `get_context_snapshot` contract timeout (`AgentRuntimeKernelContractTimeoutPolicy`, 15s): that produces `watchdog_fired=false` plus an explicit `kernel contract timeout operation=...` line, and there were **0** of those across all 11 turns in this session.

What I cannot tell from one sample: whether the bridge genuinely stalled, or whether the turn was legitimately working (a long tool round) and simply emitted nothing observable for 60s. The distinction matters for the fix — a real stall wants stall detection, a quiet-but-working turn wants either progress signalling from the harness or a longer inactivity budget for tool-heavy turns. A Node inspector CPU profile of the kernel while it is silent (`kill -USR1 <kernel pid>` + CDP `Profiler`) would separate those, and is how a related chat stall was diagnosed recently.

aryanorastar:
## What changed and why

Closes #11458.

Desktop chat treated 60 seconds without a rendered bridge event as proof that the agent runtime was frozen. A healthy provider/model round can remain quiet for longer than that, so the watchdog interrupted an admitted turn before its first content event and the turn disappeared with `partial_response=false`.

The Node runtime now emits a correlated, content-free `turn_activity` lease every 15 seconds while it owns an active query. Swift routes that lease through the owner-authorized request callback and refreshes only the bridge-silence detector. The lease is never rendered, never counted as model output, does not reset an active tool's independent 90-second no-progress clock, and is cancelled on terminal cleanup or owner revocation. If Node/JSONL actually freezes, leases stop and the existing 60-second recovery still fires.

The non-production `suspend_agent_stream` hook and E2E recipe were also corrected from the stale 180-second value to a 70-second safety resume around the real 60-second watchdog.

- INV-AGENT-*
- INV-AUTH-1
- INV-CHAT-1

## How it was verified

- `npm test -- --run tests/jsonl-transport.test.ts` — 24/24 passed, including an admitted quiet query, correlated lease emission, terminal cancellation, and a rejected late tick.
- `npm run build` — the JSONL protocol/runtime TypeScript built successfully.
- `xcrun swift test --package-path desktop/macos/Desktop --filter 'StallDetectorTests|ChatStallWatchdogTests|AgentRuntimeProcessTests'` — 99/99 passed.
- `desktop/macos/scripts/agent-logic-harness.sh` — authoritative lifecycle/state, runtime, and pi-mono package lanes passed.
- Changed-file `swift-format` lint, desktop test-quality guard, runtime-convergence ratchet, and `git diff --check` passed.
- `OMI_APP_NAME=omi-11458-watchdog OMI_AUTOMATION_PORT=47858 ./run.sh --yolo` built, signed, installed, launched, and returned healthy automation identity for isolated bundle `com.omi.omi-11458-watchdog`.

The isolated bundle had no reusable Omi Dev auth token and launched signed out, so I could not honestly claim a cloud-backed live chat turn. The quiet-runtime and frozen-runtime distinction is exercised deterministically at the production JSONL and Swift watchdog seams; no physical device is involved in this macOS-only fix.

## Tests

- `JsonlTransport` behaviorally proves a live query emits a content-free correlated lease and stops it at terminal cleanup.
- `StallDetectorTests` proves repeated leases keep a healthy quiet turn alive beyond 60 seconds, then proves the watchdog becomes eligible exactly 60 seconds after the final lease.
- `AgentRuntimeProcessTests` proves the wire event preserves request correlation and carries no content.
- `ChatStallWatchdogTests` guards the production callback wiring from the runtime lease into the detector.

## Failure class (fixes)

Failure-Class: new

Registers `FC-output-silence-misclassified-as-runtime-death`.

Violated contract: absence of assistant content is not proof that the query-owning runtime is dead; a watchdog may terminate a turn only after the runtime itself stops proving progress.

Canonical guard: the authoritative query registration owns a bounded, typed activity lease; consumers refresh liveness only from that correlated lease, and regression coverage proves both halves of the boundary — a quiet responsive runtime stays alive, while a stopped lease still reaches the existing watchdog.

## Line-count exceptions

Line-Count-Exception: desktop/macos/Desktop/Sources/Chat/AgentBridge.swift | 2214 -> 2224 | The owner-authorized query boundary must accept, guard, and forward the typed activity callback through both normal and auth-retry attempts; extracting those lifecycle-local lines would split authorization from delivery.

Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatProvider.swift | 6858 -> 6869 | The activity callback must share the active turn's callback queue, stall detector, message identity, and transition application; moving it out would separate generation-scoped state from the send lifecycle.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-output-silence-misclassified-as-runtime-death.json
+++ b/.github/failure-classes/FC-output-silence-misclassified-as-runtime-death.json
@@ -1,0 +1,18 @@
+{
+  "schema_version": 1,
+  "id": "FC-output-silence-misclassified-as-runtime-death",
+  "violated_contract": "A long-running request can remain healthy while its provider emits no user-visible content. The absence of text, thinking, or tool events proves only output silence; it does not prove that the query-owning runtime or its transport is dead. A watchdog that treats those states as equivalent terminates legitimate quiet work at its fixed boundary and can erase the turn before the first partial response exists.",
+  "canonical_prevention": "Let the authoritative active-query registration own a bounded, correlated, content-free activity lease. Refresh bridge liveness only from that typed lease or genuine runtime events, cancel the lease on every terminal or revoked path, and preserve a negative control proving that a frozen runtime stops renewing the lease and still reaches the existing watchdog.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/agent/src/runtime/jsonl-transport.ts",
+    "desktop/macos/agent/tests/jsonl-transport.test.ts",
+    "desktop/macos/Desktop/Tests/StallDetectorTests.swift"
+  ],
+  "evidence_prs": [11458],
+  "scope_hints": [
+    "desktop/macos/agent/**",
+    "desktop/macos/Desktop/Sources/Chat/**",
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift"
+  ],
+  "status": "open"
+}
--- a/desktop/macos/Desktop/Sources/Chat/AgentBridge.swift
+++ b/desktop/macos/Desktop/Sources/Chat/AgentBridge.swift
@@ -645,6 +645,7 @@ actor AgentBridge {
   typealias TextDeltaHandler = @Sendable (String) -> Void
   typealias ToolCallHandler = @Sendable (String, String, [String: Any]) async -> String
   typealias ToolActivityHandler = @Sendable (String, String, String?, [String: Any]?) -> Void
+  typealias TurnActivityHandler = @Sendable () -> Void
   typealias ThinkingDeltaHandler = @Sendable (String) -> Void
   typealias ToolResultDisplayHandler = @Sendable (String, String, String) -> Void
   typealias AuthRequiredHandler = @Sendable ([[String: Any]], String?) -> Void
@@ -1585,6 +1586,7 @@ actor AgentBridge {
     authorizationSnapshot: RuntimeOwnerAuthorizationSnapshot? = nil,
     onTextDelta: @escaping TextDeltaHandler,
     onToolActivity: @escaping ToolActivityHandler,
+    onTurnActivity: @escaping TurnActivityHandler = {},
     onThinkingDelta: @escaping ThinkingDeltaHandler = { _ in },
     onToolResultDisplay: @escaping ToolResultDisplayHandler = { _, _, _ in },
     onAuthRequired: @escaping AuthRequiredHandler = { _, _ in },
@@ -1617,6 +1619,7 @@ actor AgentBridge {
       authorizationSnapshot: authorization,
       onTextDelta: onTextDelta,
       onToolActivity: onToolActivity,
+      onTurnActivity: onTurnActivity,
       onThinkingDelta: onThinkingDelta,
       onToolResultDisplay: onToolResultDisplay,
       onAuthRequired: onAuthRequired,
@@ -1637,6 +1640,7 @@ actor AgentBridge {
     authorizationSnapshot: RuntimeOwnerAuthorizationSnapshot? = nil,
     onTextDelta: @escaping TextDeltaHandler,
     onToolActivity: @escaping ToolActivityHandler,
+    onTurnActivity: @escaping TurnActivityHandler = {},
     onThinkingDelta: @escaping ThinkingDeltaHandler = { _ in },
     onToolResultDisplay: @escaping ToolResultDisplayHandler = { _, _, _ in },
     onAuthRequired: @escaping AuthRequiredHandler = { _, _ in },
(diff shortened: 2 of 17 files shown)


Git-on-my-level:
Thanks @aryanorastar — this is a careful fix for #11458, and the core mechanism checks out against current `main`.

**What I verified against the base branch**

- `ChatProvider.swift` really does have `genericWatchdogInactivityMs = 60_000` (with `perToolStallAbortMs = 90_000`), while `desktop/macos/e2e/SKILL.md` and `DesktopAutomationBridge.swift` still documented a 180s watchdog with a 190s suspend default — so those corrections in this PR fix genuinely stale docs rather than churning them.
- The lease can't corrupt the tool watchdog: `StallDetector.observe(kind: .other)` only resets `lastEventAtMs`; it never touches `toolStartedAtMs`/`toolLastProgressAtMs`, so a 15s `turn_activity` lease can never reset an active tool's independent 90s no-progress clock — exactly what `ChatStallWatchdogTests.testTurnActivityRefreshesTheSilentBridgeClock` pins down.
- The lease stays content-free where it matters: `trackedTurnActivity` in `AgentBridge.swift` is gated on `RuntimeOwnerIdentity.isAuthorizationCurrent` and deliberately does **not** call `bridgeOutputTracker.markOutput()` (contrast `trackedToolActivity`), so it can never be counted as bridge output or partial response, and `AgentRuntimeStatusStore` renders it as plain running with no status text.
- Terminal cleanup is double-guarded in `agent/src/runtime/jsonl-transport.ts`: the scheduler callback re-checks `context.revoked || !activeByRequest.has(key)` before emitting, and `stopQueryActivity()` runs in the `finally` — the new vitest case proves a late tick after the terminal result emits nothing.
- The negative control survives: SIGSTOP freezes the whole Node process, so timers stop and `isSilentWithoutActiveTools` still reaches the existing 60s recovery. `StallDetectorTests.testRuntimeActivityLeaseKeepsAQuietTurnBelowTheBridgeWatchdog` covers both sides of the lease-expiry boundary (119_999ms vs 120_000ms) — nice.
- Routing/parsing: moving the closed `Kind` taxonomy out of `AgentRuntimeProcess.swift` into `AgentRuntimeMessageKind.swift` is a clean relocation with `turnActivity` added; correlation is preserved (`testTurnActivityParsingPreservesRequestCorrelationWithoutContent`), and the `chat-hermetic.yaml` covers-list was updated.
- The new `.github/failure-classes/FC-output-silence-misclassified-as-runtime-death.json` follows the existing FC-* schema convention, and the changelog entry is user-accurate.

**One design point for a maintainer**

This PR redefines "stuck": while the runtime keeps renewing the lease, a turn with no tool in flight can now stay alive indefinitely — the 20s `.stalled` promotion can never fire and there is no automatic recovery, so if a provider socket hangs without timing out upstream, the user's only recourse is Cancel. That is the right trade for #11458 (the old watchdog destroyed entire quiet turns), but accepting "live runtime ⇒ unbounded quiet turn, manual cancel only" is a recovery-semantics decision the watchdog owner should sign off on; a much larger lease-backed ceiling or a visible "still working" affordance could be follow-ups.

No split needed — this is one cohesive fix end to end. Leaving the watchdog-semantics signoff to human maintainer review.

---

kodjima33 (approved):
Desktop bug-fix fast lane (owner-override): closes #11458, fixes watchdog killing legitimately-quiet agent turns via a correlated content-free activity lease. Thorough verification, CI clean.

PR #12187, opened 2026-08-25
fix(desktop): make task retirement the lifecycle authority

### LINKED ISSUE #11518: Desktop: raw TaskActionItem.deleted reads disagree with isRetired, so status-only-retired tasks read as live

aryanorastar:
## What changed and why

Fixes #11518. `TaskActionItem.isRetired` already understood canonical `cancelled` and `superseded` statuses, but desktop task surfaces, filters, automation, and chat tools could still read the optional legacy `deleted` flag directly and treat status-only-retired tasks as live.

This makes the legacy wire marker private, preserves its `deleted` JSON key, and routes every `TaskActionItem` lifecycle decision through `isRetired`. Cancelled and superseded tasks are now excluded from active lists and mutations and classified into the correct Removed lane even when the backend omits `deleted`.

- `INV-CHAT-1`: task chat/tool paths still use the existing canonical task and conversation owners; this only rejects already-retired task targets.
- `INV-NAV-1`: chat-first and legacy task surfaces now apply the same retirement predicate.
- `INV-TASK-1`: complete dated buckets and bounded No Deadline paging are unchanged; retired rows are consistently removed from both active bucket shapes.

## How it was verified

- `xcrun swift build -c debug --package-path desktop/macos/Desktop` — passed.
- Built, signed, installed, and launched `/Applications/omi-11518-retirement.app` (`com.omi.omi-11518-retirement`) against the development desktop backend. The automation bridge reported the exact bundle and exposed the genuine task store actions. This machine has no signed-in desktop auth seed; an attempted CRUD smoke correctly failed closed as signed out. The repository's offline synthetic-auth harness could not start because Docker/Typesense is not installed, so no authenticated UI claim is made.
- Focused lifecycle/storage/navigation suite — 44 tests, 0 failures.
- Full desktop suite — 5,766 tests executed, 1 skipped, 2 failures outside this diff:
  - `FloatingBarNotificationPreviewPolicyTests.testDirectorDeliveryWithDisabledCategoryToggleIsSuppressedAtTheEntryPoint` failed only in the full shared-state run and passed when rerun in isolation.
  - `RewindCaptureExclusionGenerationTests.testOwnerSnapshotStaysCurrentWhenAuthLeadsUnresolvedRewindDatabase` reproduces in isolation with the process-global owner resolving to `anonymous`; no changed file participates in that test.
- Pinned `swift-format lint --strict`, `check_desktop_test_quality.py`, and `git diff --check` — passed.

## Tests

`TaskActionItemRetirementAuthorityTests` is the regression coverage. It constructs `cancelled` and `superseded` payloads with `deleted: nil` and proves they are retired, appear only in the appropriate Removed filter, and never enter the active dated surface. It also proves user-removal provenance and exact legacy `deleted` wire compatibility.

Related `TasksStoreTaskPaginationTests`, `TasksStoreDeletedLaneRetirementTests`, `ActionItemStorageVisibilityReconciliationTests`, and `DashboardIntelligenceStoreTests` all pass with the new authority boundary.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
Desktop bug-fix fast lane (owner-override): fixes #11518, routes all task lifecycle checks through isRetired instead of legacy deleted flag so cancelled/superseded tasks are correctly excluded. CI clean.

PR #12074, opened 2026-08-22
feat(desktop-windows): add tray "Open Bar" fallback for the summon hotkey

formed2forge:
## Summary
- The companion bar is summon-only via the global hotkey, which can't register on native Wayland (no XWayland grab semantics) — leaving no way to trigger it at all in that environment.
- Adds a tray menu item ("Toggle OMI Bar") as a fallback. Uses a new `summonFromTray()` function that applies the tap toggle directly (peek when hidden, hide when cleanly presented) without routing through the gesture machine or emitting `bar:ptt` phases.
- **Why not `handleSummonPress()`?** That routes through `gesture.fire()` → `onGestureStart()` → `sendPtt('down')` when a Windows key sampler is present. A tray click has no physical key to hold, so the 1200ms repeat-gap timer fires `onGestureEnd(kind='hold')` → `sendPtt('up')`, producing ~1s of unintended microphone capture on every tray click.

## Test plan
- [x] Unit tests in `trayState.test.ts`: menu structure, label, click routing
- [x] Regression test in `bar/window.test.ts`: enables bar, calls `summonFromTray()`, asserts `gesture.fire()` and `bar:ptt` are never touched
- [ ] Manual verification on native Wayland (no working global hotkey) that the tray item summons the bar

Failure-Class: none

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12189, opened 2026-08-25
fix(desktop-windows): stop the OCR helper from orphaning past app quit

formed2forge:
## Summary
The Rewind screen-reading (OCR) helper process (`win-ocr-helper.exe` on Windows, `omi-ocr-helper` on Linux — a persistent, supervised child process shared by OCR and window-info requests) could survive past app quit, left running in the background.

`will-quit` already called `helperProcess.dispose()` to kill it (fixed for the naive no-dispose-at-all case in a prior commit), but `dispose()` deliberately reset its crash-backoff cooldown to 0 with no guard against being called again — correct for mid-session crash recovery, wrong for shutdown. Separately, `ocrService`'s Rewind OCR backfill sweep runs its own 4-second `setInterval` that was never stopped anywhere in the quit path. A tick landing after `dispose()` already ran would lazily spawn a **brand-new** helper process — and since `will-quit` is a one-shot Electron event that had already fired, nothing was left to kill it. It survived as an orphan.

Fixed both ends:
- Added `stopRewindOcr()` and call it before `helperProcess.dispose()` in `will-quit`, so the backfill sweep can't fire into a shutting-down app.
- Gave `HelperProcess` a `disposed` flag mirroring the sibling `SystemAudioMuteBridge`'s already-correct pattern (checked in `ensureStarted()`/`request()`), so a post-quit call can't respawn a new process regardless of what triggers it.
- Updated the existing `dispose()` regression test, whose "re-spawns a fresh child on the next request" assertion had actually encoded the exact bug this fixes.

**Note for a maintainer, not part of this fix:** `AutomationBridge` has the identical missing-`disposed`-guard shape `HelperProcess` had — it just isn't actively triggered post-quit today the way `ocrService`'s timer was, so it's a latent version of the same gap rather than an active one. This is the second time (after `SystemAudioMuteBridge`'s already-correct pattern) this exact shutdown-safety pattern has needed re-deriving per sibling supervised-child-process class instead of being shared — close in shape to `FC-platform-contract-left-as-a-local-comment`, though that class is scoped to platform requirements specifically, so I'm flagging it here rather than force-declaring it.

## Update: a second, more serious sibling bug found while the user tested the first fix
After confirming the quit-orphan fix above, the user reported an active system slowdown with **5+ concurrent `tesseract` processes** running — a different bug in the same subsystem.

Root cause: the Linux helper (`resources/linux-ocr-helper/omi-ocr-helper`, a Node script) shells out to `tesseract` synchronously via `execFileSync` per OCR call, with its own 4s timeout. `HelperProcess.recycle()` (which both a **request timeout** and `dispose()` funnel through) only ever killed the helper's own PID. When Electron's 5s `REQUEST_TIMEOUT_MS` raced an in-flight `tesseract` call — more likely under load, since the whole round trip (write temp file + spawn + read stdout) has to fit inside that budget — killing the helper left its `tesseract` grandchild running: `execFileSync`'s timeout enforcement lives in the parent's blocked synchronous call, so once that parent is killed there's nothing left to ever time the grandchild out. A fresh helper then lazily respawns for the next request, and the cycle repeats — each iteration leaving one more orphaned, still-running `tesseract` process behind. Confirmed live: 10 accumulated `tesseract` processes and a reported system slowdown; PIDs embedded in the leftover `/tmp/omi-ocr-<pid>-*.jpg` temp filenames confirmed each belonged to a *different*, already-dead helper generation.

Fixed by spawning the Linux helper with `detached: true` (making it the leader of its own process group, separate from Electron's — `tesseract` inherits that group since it isn't itself spawned detached) and changing `recycle()` to signal the whole group (`process.kill(-pid, 'SIGTERM')`) on Linux instead of just the helper's PID. This kills the helper and any in-flight `tesseract` call atomically at the OS level, regardless of what either process's JS/event loop is doing when the signal arrives. Windows' helper does OCR in-process (no grandchild — confirmed by reading `win-ocr-helper/Program.cs`), so it keeps the plain `kill()`.

Added a dedicated regression test, and stubbed `process.kill` in every test in the suite that exercises `dispose()`/`recycle()` (it would otherwise fire for real against fake PIDs).

## Test plan
- [x] New/updated tests: `helperProcess.test.ts`'s dispose regression suite (now asserts no respawn + rejection instead of the old respawn-after-dispose behavior; new group-kill regression test), two new `ocrService.test.ts` tests for `startRewindOcr`/`stopRewindOcr`'s interval lifecycle
- [x] `pnpm typecheck` (node + web) clean
- [x] `pnpm lint` clean
- [x] Full suite: 239 files / 2,640 tests pass
- [x] **Both fixes verified against real packaged Linux builds** by the user: the first fix — confirmed the `tesseract` process disappeared on quit where it previously stayed running; the second — this session found and fixed the concurrent-orphan bug the same live testing surfaced

Failure-Class: none

### DIFF
--- a/desktop/windows/changelog/unreleased/2026-08-ocr-helper-quit-orphan.json
+++ b/desktop/windows/changelog/unreleased/2026-08-ocr-helper-quit-orphan.json
@@ -1,0 +1,5 @@
+{
+  "changes": [
+    "Fixed the screen-reading (OCR) helper process sometimes staying alive in the background after quitting Omi."
+  ]
+}
--- a/desktop/windows/changelog/unreleased/2026-08-ocr-helper-quit-orphan.json
+++ b/desktop/windows/changelog/unreleased/2026-08-ocr-helper-quit-orphan.json
@@ -1,5 +1,5 @@
 {
   "changes": [
-    "Fixed the screen-reading (OCR) helper process sometimes staying alive in the background after quitting Omi."
+    "Fixed the Linux screen-reading (OCR) helper occasionally leaving background processes running (including after quitting Omi), which could slow down your system over time."
   ]
 }
--- a/desktop/windows/src/main/index.ts
+++ b/desktop/windows/src/main/index.ts
@@ -131,7 +131,7 @@ import {
   startRewindForegroundCaptureTrigger,
   stopRewindForegroundCaptureTrigger
 } from './rewind/foregroundCaptureTrigger'
-import { startRewindOcr } from './rewind/ocrService'
+import { startRewindOcr, stopRewindOcr } from './rewind/ocrService'
 import { startRewindEmbedding } from './rewind/embeddingService'
 import { startRewindRetention } from './rewind/retentionRunner'
 import { startOrphanSweep } from './rewind/orphanSweep'
@@ -1531,6 +1531,10 @@ app.on('will-quit', () => {
   automationBridge.dispose()
   stopAutomationTargetTracker()
   stopRewindForegroundCaptureTrigger()
+  // Stop the Rewind OCR backfill sweep BEFORE disposing the helper below — a
+  // tick firing after dispose() would otherwise lazily respawn a fresh helper
+  // that nothing is left to kill, orphaning it past app exit.
+  stopRewindOcr()
   // Kill the long-running OCR/window-info helper subprocess. Without this it
   // outlives the app on every quit, so orphaned omi-*-ocr-helper.exe processes
   // pile up across launches (no production dispose() call site before this).
--- a/desktop/windows/src/main/ocr/helperProcess.test.ts
+++ b/desktop/windows/src/main/ocr/helperProcess.test.ts
@@ -14,14 +14,17 @@ vi.mock('child_process', () => ({ spawn: (...args: unknown[]) => spawnMock(...ar
 vi.mock('./resolveHelperPath', () => ({ resolveHelperPath: () => 'C:\\fake\\win-ocr-helper.exe' }))
 
 type FakeChild = EventEmitter & {
+  pid: number
   stdout: EventEmitter
   stderr: EventEmitter
   stdin: { write: ReturnType<typeof vi.fn> }
   kill: ReturnType<typeof vi.fn>
 }
 
+let nextFakePid = 1000
 function makeFakeChild(): FakeChild {
   const child = new EventEmitter() as FakeChild
+  child.pid = nextFakePid++
   child.stdout = new EventEmitter()
   child.stderr = new EventEmitter()
   child.stdin = { write: vi.fn() }
@@ -33,6 +36,11 @@ describe('helperProcess.dispose (C7 — no orphaned OCR helper on quit)', () => 
   beforeEach(() => {
     spawnMock.mockReset()
     vi.resetModules()
+    // Every test in this suite exercises dispose()/recycle(), which on Linux
+    // now calls the REAL process.kill(-pid, ...) unless mocked — with fake
+    // pids that could collide with an unrelated real process group on the
+    // test machine. Stub it everywhere, not just in the test that asserts on it.
+    vi.spyOn(process, 'kill').mockImplementation(() => true)
   })
   afterEach(() => {
     vi.restoreAllMocks()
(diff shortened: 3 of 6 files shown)


formed2forge:
The two red checks (\`Backend Hermetic Merge Gate\`, \`Detect Hermetic Backend Scope\`) are the same recurring infra flake seen on #12074 — \`Detect Hermetic Backend Scope\` still uses \`actions/checkout@v7\` with \`fetch-depth: 0\` (a full-history clone) to compute \`git diff --name-only base...HEAD\`, and that clone occasionally exceeds the job's 2-minute timeout. Once checkout completes the logic itself runs in well under a second. \`Backend Hermetic Merge Gate\` then fails closed as a designed consequence.

Not caused by this PR's content — no \`backend/\` paths touched. No rerun permission as an outside contributor. This is now confirmed a third time (a prior session's #3, #12074, and here) — worth raising the timeout or moving to a shallow fetch with an explicit merge-base ref if someone with write access has a minute.

formed2forge:
Found strong precedent for the process-group-kill approach in the follow-up commit: \`22d1a77e55\` (\`fix: kill entire process group for shell-spawned external adapters\`, already merged, desktop/macos's \`AcpRuntimeAdapter\`/\`LocalSubprocessRuntimeAdapter\`) fixes the identical shape of bug — a shell-spawned real child surviving a kill of its wrapper/parent PID — with the same \`detached: true\` + \`process.kill(-pid, 'SIGTERM')\` pattern. Independent confirmation this is the established fix for this bug class in the codebase, not something novel to this PR.

Git-on-my-level:
Verified this end-to-end against the tree at f06d51e — both the diagnosis and the fix check out. Notes per file:

- `desktop/windows/src/main/ocr/helperProcess.ts` — the new `disposed` flag closes the respawn race at both points that matter: `ensureStarted()` (no lazy spawn after quit) and `request()` (post-quit calls fail fast with a clear error instead of hanging). The Linux-only `detached: true` spawn paired with `process.kill(-this.child.pid, 'SIGTERM')` in `recycle()` is the right group-kill; I confirmed the grandchild claim against `resources/linux-ocr-helper/omi-ocr-helper`, which really does `execFileSync('tesseract', …)` per OCR request, so killing only the helper PID would strand an in-flight tesseract. Matches the pattern already merged for the macOS adapters in 22d1a77e.
- `desktop/windows/src/main/rewind/ocrService.ts` — `stopRewindOcr()` clearing the 4s backfill interval is the root-cause fix; `startRewindOcr()` still clears any prior timer, so restart-after-stop remains correct.
- `desktop/windows/src/main/index.ts` — `stopRewindOcr()` sits immediately before `helperProcess.dispose()` in the synchronous `will-quit` handler, so no tick can interleave between them; and a tick already inside `backfill()` when dispose runs is still covered (recycle rejects the in-flight queue, new requests hit the `disposed` guard). Good defense in depth.
- `desktop/windows/src/main/ocr/helperProcess.test.ts` — flipping the old "re-spawns after dispose()" assertion to "does NOT re-spawn" is right; that test had pinned the bug as contract. Stubbing `process.kill` suite-wide so fake pids can't signal real process groups on a dev machine is a good catch.
- `desktop/windows/src/main/rewind/ocrService.test.ts` — the new interval-lifecycle tests (start schedules a tick, stop clears it) directly pin the quit-path regression.
- `desktop/windows/changelog/unreleased/2026-08-ocr-helper-quit-orphan.json` — minor wording nit, non-blocking: the entry scopes this to "the Linux … helper", but the quit-path race fix (`stopRewindOcr` + `disposed` guard) protects the Windows helper too; consider wording that covers both platforms. Fine as-is if the Linux tesseract orphan was the user-visible pain.

One residual I considered and am comfortable with: after `process.kill(-pid, …)` the pgid could in theory be recycled and the negative signal hit an unrelated group, but that requires pgid wraparound within the spawn-to-kill window and mirrors the already-merged macOS pattern — not worth complicating.

The AutomationBridge latent-gap note in the description is a fair observation for a follow-up issue; agreed it doesn't belong here.

Clean, well-tested fix with an unusually good root-cause write-up — leaving for maintainer merge.

---

kodjima33 (approved):
Desktop bug-fix owner-override: title matches fix(desktop-windows), scoped diff with regression tests for both the quit-orphan and the tesseract-grandchild-orphan bug, CI green, no abuse signals. Verified not already fixed on main.

PR #12205, opened 2026-08-25
fix(desktop-windows): use pnpm, not npm, in LINUX.md's Run from source

formed2forge:
## Summary

\`LINUX.md\`'s "Run from source" section told users to run \`npm install\` and \`npm run dev\`. The project is pinned to pnpm — \`npm install\` produces a different lockfile and silently mismatches versions. Fixed to \`pnpm install\` / \`pnpm run dev\`.

## Test plan

- Docs-only change; verified the commands match \`README.md\` and \`package.json\` scripts

Failure-Class: none

### DIFF
--- a/desktop/windows/LINUX.md
+++ b/desktop/windows/LINUX.md
@@ -16,8 +16,8 @@ codebase; only OS-specific adapters branch on `process.platform`.
 ```bash
 cd desktop/windows
 cp .env.example .env          # public Firebase/PostHog config; sign-in works as-is
-npm install
-npm run dev                   # launch on an X11 session (DISPLAY set)
+pnpm install --frozen-lockfile
+pnpm run dev                   # launch on an X11 session (DISPLAY set)
 ```
 
 ## Runtime dependencies (Debian/Ubuntu)


Git-on-my-level:
Thanks @formed2forge — verified, and this checks out.

What I confirmed against the repo:

- `desktop/windows` is pnpm-managed: it ships `pnpm-lock.yaml` and `pnpm-workspace.yaml` (and no `package-lock.json`), so `npm install` would resolve a fresh dependency tree that doesn't match the lockfile CI installs from.
- The sibling `README.md` "Run from source" section already documents exactly `pnpm install --frozen-lockfile` and `pnpm run dev`, so this makes `LINUX.md` consistent with the existing convention rather than introducing a new one.
- `package.json` defines a `dev` script (`electron-vite dev`), so `pnpm run dev` is valid as written.
- No other npm-as-package-manager references remain anywhere in `LINUX.md` at this head — the two replaced lines were the only ones.

One trivial, non-blocking nit: after the swap, the trailing inline comments (`# public Firebase/PostHog config...`, `# launch on an X11 session...`) are no longer column-aligned with the longer `pnpm install --frozen-lockfile` line above them. Purely cosmetic — fine to leave as-is.

Nice catch on a real papercut for anyone following the Linux doc. Leaving formal merge sign-off to a maintainer.

---

kodjima33 (approved):
Docs-correctness verified: desktop/windows is pinned to pnpm (pnpm-lock.yaml present, no package-lock.json) and LINUX.md still said npm — correct fix.

PR #12219, opened 2026-08-25
feat(desktop-windows): hydrate default chat thread from backend on mount

formed2forge:
## Summary

- `loadDefaultThreadHistory` helper (new, inside `useChat`) tries `GET /v2/desktop/messages` (no session/app → default shared thread) before falling back to local Electron SQLite — giving the Windows main window cross-device read parity with Mac and mobile
- All three default-thread entry points updated: mount loader, `switchThread(null)`, and `selectApp(null)` — all three now share the same backend-first helper
- Falls back gracefully to local SQLite when not signed in, network unavailable, or backend returns empty

## How it works

`GET /v2/desktop/messages` with no query params returns the `plugin_id == None` default Firestore thread — the same store macOS's `ChatProvider` and the mobile app read. Windows was already **writing** to this thread (via `POST /v2/desktop/messages`), so the cross-device writes were there all along; only the reads were missing.

Auth guard: if `auth.currentUser` is null at mount time, the backend call is skipped entirely (no 401 risk), and we go straight to local SQLite.

- **INV-CHAT-1** (`one shared transcript across surfaces`): this change closes the read half of that invariant for Windows — the thread is now genuinely shared across Mac/mobile/web/Windows, not just write-shared.

Failure-Class: none

## Test plan

- [x] R4 (new): mount loader with backend messages → history populated with backend IDs (cross-device turns visible)
- [x] R5 (new): mount loader with backend failure → falls back to local SQLite
- [x] R1/R2 unchanged: existing attachment-preservation tests still pass (backend returns `[]` → local fallback path exercised)
- [x] Full vitest suite: 5437/5437 pass

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12219?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/changelog/unreleased/2026-08-default-thread-backend-hydration.json
+++ b/desktop/windows/changelog/unreleased/2026-08-default-thread-backend-hydration.json
@@ -1,0 +1,5 @@
+{
+  "changes": [
+    "The main chat window now loads your full conversation history from the server on startup, so messages sent from your phone, Mac, or web are visible immediately — not just messages sent from this device."
+  ]
+}
--- a/desktop/windows/src/renderer/src/hooks/useChat.test.tsx
+++ b/desktop/windows/src/renderer/src/hooks/useChat.test.tsx
@@ -1159,33 +1159,41 @@ describe('useChat — default thread backend hydration (B5 cross-device read par
 
   it('R4 — mount loader (infinite) populates history from the backend (cross-device turns visible)', async () => {
     prefs.chatHistoryMode = 'infinite'
+    // Wire order: backend returns newest-first; the loader must reverse to chronological.
+    const evidenceEnvelope = { sources: [{ id: 'src-1', title: 'Test', url: 'https://t.co', score: 0.9 }] }
     sessionMocks.getMessages.mockResolvedValueOnce([
-      { id: 'be-1', text: 'hello from mobile', sender: 'human', createdAt: '2026-01-01T00:00:00Z' },
-      { id: 'be-2', text: 'hi there', sender: 'ai', createdAt: '2026-01-01T00:00:01Z' }
+      { id: 'be-2', text: 'hi there', sender: 'ai', createdAt: '2026-01-01T00:00:01Z', evidence: evidenceEnvelope },
+      { id: 'be-1', text: 'hello from mobile', sender: 'human', createdAt: '2026-01-01T00:00:00Z' }
     ])
     const { result } = renderHook(() => useChat())
     await act(async () => {
       await settleHistory(result)
     })
     expect(result.current.history).toHaveLength(2)
+    // Oldest message first — confirmed chronological despite newest-first wire order.
     expect(result.current.history[0]).toMatchObject({ id: 'be-1', role: 'user', content: 'hello from mobile' })
-    expect(result.current.history[1]).toMatchObject({ id: 'be-2', role: 'assistant', content: 'hi there' })
+    expect(result.current.history[1]).toMatchObject({ id: 'be-2', role: 'assistant', content: 'hi there', evidence: evidenceEnvelope })
   })
 
   it('R5 — mount loader (infinite) falls back to local SQLite when the backend call fails', async () => {
     prefs.chatHistoryMode = 'infinite'
     sessionMocks.getMessages.mockRejectedValueOnce(new Error('network'))
+    const evidenceEnvelope = { sources: [{ id: 'src-2', title: 'Local', url: 'https://l.co', score: 0.8 }] }
     ;(window as unknown as { omi: { getLocalConversation: unknown } }).omi.getLocalConversation =
       async () => ({
         startedAt: 1,
-        messages: [{ id: 'loc-1', role: 'user' as const, content: 'offline message' }]
+        messages: [
+          { id: 'loc-1', role: 'user' as const, content: 'offline message' },
+          { id: 'loc-2', role: 'assistant' as const, content: 'cached reply', evidence: evidenceEnvelope }
+        ]
       })
     const { result } = renderHook(() => useChat())
     await act(async () => {
       await settleHistory(result)
     })
-    expect(result.current.history).toHaveLength(1)
+    expect(result.current.history).toHaveLength(2)
     expect(result.current.history[0]).toMatchObject({ id: 'loc-1', role: 'user', content: 'offline message' })
+    expect(result.current.history[1]).toMatchObject({ id: 'loc-2', role: 'assistant', content: 'cached reply', evidence: evidenceEnvelope })
   })
 })
 
--- a/desktop/windows/src/renderer/src/hooks/useChat.test.tsx
+++ b/desktop/windows/src/renderer/src/hooks/useChat.test.tsx
@@ -1151,6 +1151,45 @@ describe('useChat — rehydrate preserves attachments', () => {
   })
 })
 
+describe('useChat — default thread backend hydration (B5 cross-device read parity)', () => {
+  // settle waits for at least one message to land in history.
+  const settleHistory = async (result: { current: { history: unknown[] } }): Promise<void> => {
+    for (let k = 0; k < 50 && !result.current.history.length; k++) await flush()
+  }
+
+  it('R4 — mount loader (infinite) populates history from the backend (cross-device turns visible)', async () => {
+    prefs.chatHistoryMode = 'infinite'
+    sessionMocks.getMessages.mockResolvedValueOnce([
+      { id: 'be-1', text: 'hello from mobile', sender: 'human', createdAt: '2026-01-01T00:00:00Z' },
+      { id: 'be-2', text: 'hi there', sender: 'ai', createdAt: '2026-01-01T00:00:01Z' }
+    ])
+    const { result } = renderHook(() => useChat())
+    await act(async () => {
+      await settleHistory(result)
+    })
+    expect(result.current.history).toHaveLength(2)
+    expect(result.current.history[0]).toMatchObject({ id: 'be-1', role: 'user', content: 'hello from mobile' })
+    expect(result.current.history[1]).toMatchObject({ id: 'be-2', role: 'assistant', content: 'hi there' })
+  })
+
+  it('R5 — mount loader (infinite) falls back to local SQLite when the backend call fails', async () => {
+    prefs.chatHistoryMode = 'infinite'
+    sessionMocks.getMessages.mockRejectedValueOnce(new Error('network'))
+    ;(window as unknown as { omi: { getLocalConversation: unknown } }).omi.getLocalConversation =
+      async () => ({
+        startedAt: 1,
+        messages: [{ id: 'loc-1', role: 'user' as const, content: 'offline message' }]
+      })
+    const { result } = renderHook(() => useChat())
+    await act(async () => {
+      await settleHistory(result)
+    })
+    expect(result.current.history).toHaveLength(1)
+    expect(result.current.history[0]).toMatchObject({ id: 'loc-1', role: 'user', content: 'offline message' })
+  })
+})
+
+
 describe('useChat — chat quota gate (Mac AgentBridge.quotaExceeded parity)', () => {
   it('blocks a send when the quota is exhausted — popup shown, no fetch, no history entry', async () => {
     gateMocks.check.mockResolvedValueOnce({ blocked: true, message: "You've reached your limit." })
--- a/desktop/windows/src/renderer/src/hooks/useChat.ts
+++ b/desktop/windows/src/renderer/src/hooks/useChat.ts
@@ -81,6 +81,10 @@ export type ChatMsg = {
 
 const OMI_BASE = import.meta.env.VITE_OMI_API_BASE as string
 
+// Number of turns to fetch from the backend on the default-thread mount.
+// Matches Mac's single-page import size; pagination is deferred.
+const BACKEND_HISTORY_LIMIT = 100
+
 // Hard ceiling on a single streamed reply. Mirrors the macOS client's per-send
 // watchdog (ChatProvider.swift): if the SAME generation is still in flight after
 // this long, abort the fetch, unlatch the engine, and surface a timeout so a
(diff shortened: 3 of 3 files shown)


Git-on-my-level:
Thanks @formed2forge — the actual feature here is in good shape. I reviewed the head commit (b8d546b) directly, because the PR-level files-changed list is dominated by branch history rather than your change (more below).

What the feature commit actually changes (3 files):

- `desktop/windows/src/renderer/src/hooks/useChat.ts` — `loadDefaultThreadHistory()` unifies all three default-thread entry points (mount loader, `switchThread(null)`, `selectApp(null)`) behind one backend-first loader with local-SQLite fallback. The cancellation discipline is right: `isCurrent()` is re-checked before every `setHistory`, the mount effect captures its generation, and the signed-out case skips the backend call entirely (no 401 churn). `startedAtRef` is only set on the local branch, which is correct since backend turns don't need a local `startedAt`.
- Same file, quota gate: `mainChatQuotaGate.check()` before send, fail-open when signed out, `recordQuery()` optimistic count — good parity with the bar's gate in `barSend.ts`. One gap: the export comment says `UsageLimitTriggerHost` will call `sync()` after each completed reply, but this PR doesn't wire that. The host (`components/settings/billing/UsageLimitTriggerHost.tsx`) still only probes via `maybeTriggerChatQuotaPopup` and never touches `mainChatQuotaGate`, so after the cold-start fetch inside `check()` the snapshot never refreshes: the optimistic delta grows for the whole session, and a user who crosses the cap mid-session stays pre-send blocked on the main window until restart (the bar's gate recovers on its next reveal sync; the server still enforces either way, so this is enforcement UX, not a hole). Please wire the `sync()` call here — the host's own TODO already points at this — or call it out as an immediate follow-up.
- `desktop/windows/src/renderer/src/hooks/useChat.test.tsx` — R4/R5 cover the mount loader both ways (backend messages land in history with backend IDs; backend failure falls back to local SQLite), and the two quota tests cover blocked (no fetch, popup shown, no history entry, sending stays false) and in-quota (exactly one fetch + one `recordQuery`). Mock paths match the hook's imports.
- `desktop/windows/changelog/unreleased/2026-08-default-thread-backend-hydration.json` — accurate, user-facing wording that matches the behavior.

One behavioral nuance worth a maintainer look: when signed in and the backend returns non-empty history, the loader shows backend messages only. Turns that exist only in local SQLite (e.g. a reply persisted locally when its backend POST failed — the "persisted LOCALLY only" path) won't appear in the main window until they exist server-side. Mac's backend-first import has the same shape, so this reads as parity rather than a regression, but it is a product-visible choice.

Branch state — please rebase before merge: the branch carries ~3 weeks of already-merged upstream history (Aug 17–24) under rewritten SHAs, which is why the PR shows ~205 changed files and GitHub reports conflicts with main. Your feature commit touches only the 3 files above. Rebasing (or resetting) onto current main will collapse the diff to just your change and avoid accidentally reverting newer upstream edits — e.g. the PR-level diff currently shows a one-line delta in `backend/AGENTS.md` that is purely your base being older than main.

Agent-guidance docs note (for reviewers): the PR-level diff flags `backend/AGENTS.md`, `desktop/macos/AGENTS.md`, `backend/charts/monitoring/README.md`, and `backend/docs/runbooks/silent-failure-detection.md` as changed. Those are base-branch artifacts, not contributor edits — the feature commit modifies none of them, and they should drop out of the diff after the rebase.

Needs a maintainer decision on whether the quota-gate sync wiring lands here or as a follow-up, and sign-off on the backend-first hydration visibility semantics above; otherwise leaving for human maintainer review.

---

formed2forge:
Rebased onto current main and wired the quota sync. Two changes in this update:

1. **Rebase** — PR now shows 5 files (3 from the hydration feature + changelog + UsageLimitTriggerHost). The stale history is gone.

2. **Quota sync** — `UsageLimitTriggerHost` now imports `mainChatQuotaGate` from `useChat` and calls `sync()` on the busy→idle edge instead of `maybeTriggerChatQuotaPopup(fetchChatQuota)`. The gate's snapshot refreshes after each completed reply so the next pre-send check sees an up-to-date verdict without a network round trip.

Mergeable.

formed2forge:
Rebase complete, quota sync wired (UsageLimitTriggerHost calls sync() on busy→idle edge). The backend-first hydration is intentional parity with Mac behavior. Ready for maintainer review.

formed2forge:
The `Detect Changes` job is failing due to runner starvation — a pure CI infrastructure issue unrelated to this PR. Could a maintainer re-run the failing job? Once green this is ready for review.

Git-on-my-level (dismissed):
Thanks @formed2forge — the direction here is right, and it lines up with the backend-authoritative desktop chat design (the read half of INV-CHAT-1). Collapsing the three duplicated default-thread loaders into one `loadDefaultThreadHistory` with proper `isCurrent` cancellation and a single `onSettled` hook is a real cleanup. Two concrete issues block it for me, though:

**1. The hydrated transcript renders newest-first (ordering inversion)**

`GET /v2/desktop/messages` is served by `chat_db.get_messages` (`backend/database/chat.py`), which orders `created_at DESCENDING`. The new helper maps the response straight into history with no reversal:

```ts
const msgs = await getSessionMessages({ limit: BACKEND_HISTORY_LIMIT })
...
setHistory(msgs.map((m) => ({ ... })))
```

and `ChatMessages.tsx` renders `messages.map(...)` top-down, so a signed-in user with cross-device history sees the thread inverted on mount. Every other client of this same store re-sorts ascending before display — mobile (`app/lib/providers/message_provider.dart`, `messages.sort((a, b) => a.createdAt.compareTo(b.createdAt))`) and macOS (`ChatLegacyImportChronology.plan` in `ChatProvider.swift`). The new R4 test masks this: its mock resolves the older message first (`be-1` at `00:00:00Z` before `be-2` at `00:00:01Z`) and asserts `history[0]` is the older one — the real wire returns the opposite order. Suggested fix: reverse (or sort by `createdAt`) before `setHistory`, and flip the R4 fixture to wire order so the test pins chronology. The same latent pattern exists in the pre-existing session/app paths (`switchThread(id)` / `selectApp(appId)`) — not introduced here, but a good follow-up.

**2. Evidence envelopes are dropped from hydrated messages**

The send path persists evidence to the backend as message `metadata` (`useChat.ts`, `metadata: JSON.stringify({ evidence: finalEvidence })`), the client layer already decodes it into `DesktopMessage.evidence` (`chatSessionsClient.ts` `toDesktopMessage`), and the session/app-thread paths map it via `parseChatEvidenceFromRecord`. But the new helper omits it in **both** branches: `m.evidence` is available yet unmapped on the backend branch, and the local-SQLite branch dropped its old `parseChatEvidenceFromRecord(m)` call. Since `ChatMessages.tsx` renders evidence cards on assistant messages, the default thread's replies lose their evidence cards on every mount/switch after this change — a regression vs the current local loader.

**Smaller notes (non-blocking)**

- `useChat.ts` mount effect: the loader runs once on mount with `[]` deps. If Firebase auth hasn't rehydrated yet, `auth.currentUser` is null, the backend fetch is skipped, and there's no retry — a signed-in user on a slow session restore sees local-only history for the whole app session. The send path already has a bounded poll for exactly this (`NOT_READY_POLL_TRIES`); a similar short wait or an auth-state subscription before falling back would close the gap.
- `desktop/windows/changelog/unreleased/2026-08-default-thread-backend-hydration.json`: "loads your full conversation history" overstates it — the fetch is capped at `BACKEND_HISTORY_LIMIT = 100` with pagination deferred. "Your recent conversation history (latest 100 messages)" would be accurate.

The `useChat.test.tsx` additions (R4/R5) are the right shape — cross-device visibility plus offline fallback — they just need the R4 fixture flipped to real wire order so the ordering fix stays pinned. With the reversal and evidence mapping restored, this is close.

Leaving for human maintainer review: this switches the main chat window's history source to backend-first (a data-handling decision for the default thread), and the ordering/evidence fixes need a maintainer's eyes before merge.

---

formed2forge:
Both blockers from the 2026-08-28 review are resolved on this head:

**1. Ordering inversion** — `loadDefaultThreadHistory`'s backend branch now does `[...msgs].reverse()` before mapping, flipping from wire order (newest-first, as `GET /v2/desktop/messages` returns) to display order (oldest-first), matching mobile's ascending-`createdAt` sort. R4 fixture is flipped to wire order so it actively pins the reversal instead of masking it.

**2. Evidence dropped** — both branches of the loader now forward evidence:
- Backend branch: `...(m.evidence ? { evidence: m.evidence } : {})` — `m.evidence` is already decoded by `toDesktopMessage`/`parseChatEvidenceFromRecord` on `DesktopMessage`, so no second parse needed.
- SQLite fallback: same spread — restores the field that the previous local loader passed through.

R4 and R5 both carry an evidence-bearing assistant message and assert the envelope survives the round-trip. All 37 tests pass locally.

Git-on-my-level:
Thanks @formed2forge — I verified both blockers from my earlier review against the actual sources on this head, and they're properly resolved.

**1. Ordering** — `loadDefaultThreadHistory` now reverses the backend response (`[...msgs].reverse()`) before mapping into history, and the direction is right: `chat_db.get_messages` (behind `GET /v2/desktop/messages`) orders `created_at DESCENDING`, so the wire is newest-first and the reversal yields chronological display — the same shape as mobile's ascending sort. The R4 fixture now feeds the mock in real wire order (newest first) and asserts the oldest message lands at `history[0]`, so the test pins the reversal instead of masking it.

**2. Evidence** — both branches forward `evidence` now. The backend branch reads `m.evidence`, which is already decoded client-side (`toDesktopMessage` runs `parseChatEvidenceFromRecord` on the wire row in `chatSessionsClient.ts`), and the SQLite fallback reads `m.evidence` directly, which matches the persisted local `ChatMessage` shape — the field is stored as a direct property, not serialized metadata. R5's local-fallback case asserts the envelope survives the round-trip.

Also checked the rebase: the quota-sync wiring from the earlier iteration (`mainChatQuotaGate.sync()` on the busy→idle edge in `UsageLimitTriggerHost`) is already on `main`, so nothing was dropped — the PR is cleanly scoped to the hydration change now (3 files).

One optional nit, non-blocking: `changelog/unreleased/2026-08-default-thread-backend-hydration.json` still says "loads your full conversation history" — the fetch is capped at `BACKEND_HISTORY_LIMIT = 100` with pagination deferred, so "your recent conversation history" would be exact.

The mechanics all check out. What's left is the product call I flagged before, so `needs-maintainer-review` stays: backend-first sourcing means a signed-in user with server history won't see turns that exist only in local SQLite (e.g. a reply persisted locally when its backend POST failed) until they exist server-side. That's deliberate Mac parity, but it's a product-visible data-handling choice for the default thread, so it needs a maintainer's sign-off before merge — nothing technical is holding it up.

---

Git-on-my-level:
Nice change — this closes the read half of the shared default thread (INV-CHAT-1) cleanly, and the dedup of the three default-thread entry points into one `loadDefaultThreadHistory` helper is a real churn reduction (−87/+79) with the agent-card projection ordering preserved via `onSettled`.

Verified against the backend contract:

- `desktop/windows/src/renderer/src/hooks/useChat.ts` — the backend-first order reversal is correct: `GET /v2/desktop/messages` (no params) queries the `plugin_id == None` default thread ordered `created_at DESCENDING` (backend `database/chat.py get_messages`), so `[...msgs].reverse()` yields chronological display order, and new test R4 pins exactly that including the evidence envelope. The `sender === 'ai' ? 'assistant' : 'user'` mapping matches the backend enum (`human`/`ai`), and `limit: 100` sits inside the endpoint's 1..1000 range.
- The `auth.currentUser` guard avoids a guaranteed-401 call when signed out and mirrors the first-chat readiness pattern already used at send time; on network failure or empty backend thread the loader falls through to local SQLite, so offline behavior is unchanged.
- Race guards are intact: `isCurrent()` is re-checked before every `setHistory`, the mount loader keeps the `cancelled`/`sendingRef`/`genRef` triple guard, and `switchThread(null)`/`selectApp(null)` capture their own generation — a slow default load can no longer overwrite a thread the user has switched to.
- Evidence and attachments survive both paths: `chatSessionsClient.ts`'s `toDesktopMessage` already decodes the wire `metadata` envelope via `parseChatEvidenceFromRecord`, so reading `m.evidence` directly in the backend branch is equivalent to the session/app loaders.
- `desktop/windows/src/renderer/src/hooks/useChat.test.tsx` — R4/R5 cover the two main paths (backend success with newest-first wire order; backend failure → SQLite fallback). `desktop/windows/changelog/unreleased/2026-08-default-thread-backend-hydration.json` follows the existing unreleased-changelog convention.

Minor notes, none blocking:

1. The changelog entry says "full conversation history", but `BACKEND_HISTORY_LIMIT = 100` caps the initial page (pagination deferred, Mac single-page parity). For long-time users the oldest turns won't appear — consider "recent history" wording or a follow-up pagination issue.
2. In the local-SQLite branch, `m.evidence` is now read directly instead of `parseChatEvidenceFromRecord(m)`. That's fine for rows written by the current persistence format, but legacy local rows that stored evidence under an alias (`evidence_envelope`, nested `metadata`) would lose the evidence chip until the next re-persist. The parser is still imported for the other loaders, so keeping the call in this branch is a one-line hardening if alias rows can exist in the wild.
3. The not-signed-in branch (`auth.currentUser` null — straight to SQLite, no network) isn't covered by a test; R5 exercises network failure only. A third case would pin the no-401 guarantee.
4. The backend path doesn't set `startedAtRef`, so the first `persistChat` after hydration stamps "now" as the local conversation's `startedAt`. Only affects the conversations-list timestamp — harmless, just noting it.

Because this switches the default chat's source of truth to server-side user chat history, leaving the existing maintainer-review label for sign-off before merge.

---

formed2forge:
Both previous review blockers are resolved per Git-on-my-level's review. Looking for a maintainer to provide product sign-off so this can merge. Thanks.

formed2forge:
- old: `b3b2520a99`
- new: `abb14866dd`
- now **0 behind / 2 ahead**, MERGEABLE

Clean replay; unique useChat hydration hunks unchanged.

Git-on-my-level:
Rebase onto `main` (`176a02fe0d` → head `abb14866dd`) verified — the hydration change replayed cleanly, and the earlier technical verification stands on this head.

- `desktop/windows/src/renderer/src/hooks/useChat.ts` — `loadDefaultThreadHistory` is the single loader at all three default-thread entry points (mount effect, `switchThread(null)`, `selectApp(null)`); the backend fetch reverses newest-first wire order into chronological display, `isCurrent()` is re-checked before every `setHistory`, and evidence/attachments survive both the backend and SQLite-fallback branches (`toDesktopMessage` decodes the wire envelope; the fallback reads the persisted field directly).
- `desktop/windows/src/renderer/src/hooks/useChat.test.tsx` — R4 still pins the wire-order reversal and R5 the local SQLite fallback; the suite ran green in CI on this exact head (`Typecheck · Lint · Test`, `Windows build smoke`).
- `desktop/windows/changelog/unreleased/2026-08-default-thread-backend-hydration.json` — follows the unreleased-changelog convention; "full conversation history" vs the 100-message initial page remains a non-blocking wording nit.

Still holding for maintainer sign-off on the one open product question from earlier reviews: backend-first sourcing means a signed-in user's default thread shows server history only — turns that exist just in local SQLite (e.g. a reply persisted when its backend POST failed) stay hidden until they exist server-side. That's deliberate Mac/mobile parity, but it's a data-handling choice for the main chat window, so it needs a human call before merge.

---
*by AI on behalf of David.*

aryanorastar (approved):
Verified on current origin/main (`c1de50bf`): the default chat thread still hydrates only from local SQLite on mount. `loadDefaultThreadHistory` is still absent.

The backend-first loader plus R4/R5 tests match the /v2/desktop/messages contract. Typecheck · Lint · Test and Windows build smoke were green on the last run.

Approving. Not merging from here.

aryanorastar (changes requested):
`desktop/windows/src/renderer/src/hooks/useChat.ts` on `abb14866ddf7` replaces the `parseChatEvidenceFromRecord` loads with `loadDefaultThreadHistory`. That helper spreads raw `m.evidence` for the backend rows (line 367) and the SQLite fallback (line 386). The session and app loaders still parse. Run the default-thread rows through `parseChatEvidenceFromRecord` as well.

PR #12223, opened 2026-08-25
fix(desktop-windows): treat pi-mono's agent_settled as advisory, not unknown

formed2forge:
## Summary

The event dispatch switch in \`PiMonoAdapter.handleEvent\` had no case for \`agent_settled\`, so it fell through to the default branch and logged \`"[pi-mono] unknown event type: agent_settled"\` — even though the adapter already treats several other protocol-control events (\`agent_start\`, \`agent_end\`, \`turn_start\`, etc.) as observed-but-ignored no-ops.

\`agent_settled\` is an upstream advisory event; \`turn_end\` is what actually resolves or rejects the pending prompt via \`handleTurnEnd\`, and that path is unchanged. Added an explicit no-op case alongside the others.

## Test plan

- Added a test in \`piMono.test.ts\` exercising \`handleEvent\`'s real JSON-string dispatch path: a prompt stays pending across \`agent_settled\` and only completes on the subsequent \`turn_end\` — matching the equivalent coverage macOS's pi-mono adapter already has
- \`vitest run\`: all tests pass

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12223?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Closing with thanks - this fix is already on main, byte-for-byte.

What your head commit does is right: it adds the `agent_settled` advisory no-op case to the event-dispatch switch in `desktop/windows/src/main/codingAgent/piMono.ts` (keeping `turn_end` authoritative via `handleTurnEnd`), with a regression test in `desktop/windows/src/main/codingAgent/piMono.test.ts` that drives the real `handleEvent` JSON-string dispatch path and asserts a prompt stays pending until `turn_end`. It also matches the macOS adapter, which handles `agent_settled` the same advisory way in `desktop/macos/agent/src/adapters/pi-mono.ts`.

The same change - same title, same author - already merged as #12173 (11494cf9) about 8 hours before this PR was opened. I diffed both trees: `piMono.ts` and `piMono.test.ts` here are identical to what is on main today, so there is nothing left to take from this branch. It looks like a duplicate submission of the same work; all credit for the landed fix stays with you via #12173.

One branch-hygiene note for next time, no action needed here: this PR shows 204 changed files because the branch sits 167 commits behind main and carries ~43 pre-squash upstream commits (the `.github/` deploy-guard rework, `backend/charts/monitoring/alert-rules.json`, `backend/scripts/attach_cloud_run_gmp_sidecar.py`, etc.). Against current main the effective delta of this branch is zero, and a merge would currently hit 11 unrelated content conflicts (e.g. `backend/charts/monitoring/alert-rules.json`, `backend/database/candidates.py`, `backend/scripts/render_backend_runtime_env.py`). Branching fresh off main - or rebasing before opening - keeps the diff down to just your two files.

Superseded by #12173.

---

PR #12224, opened 2026-08-25
fix(desktop-windows): run bytecode export-forward guard on Linux/macOS builds

formed2forge:
## Summary

\`build:linux\` and \`build:mac\` skipped \`ensure-env.mjs\`, \`patch-bytecode-entry-forward.mjs\`, and \`verify-bytecode.mjs\`, which \`build:win\` already runs. Without the forward patch, the packaged main-process bytecode stub never gets \`module.exports =\` added, so any lazily-loaded chunk calling back into \`require("../index.js").<name>\` silently reads \`{}\` — the exact class of bug already fixed for Windows (93b3cee7ad) but never backported to these two build scripts.

## Test plan

- Verified by hand: an un-patched build fails \`verify-bytecode.mjs\`'s forwarding check; the same build after \`patch-bytecode-entry-forward.mjs\` passes clean

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Closing this one as a duplicate — with thanks, because the fix itself is already on `main` under this same author's name.

**What I verified**

- The tip commit here (`643c8423`) is a two-file change: `desktop/windows/package.json` now routes `build:mac` and `build:linux` through `npm run build` (so `ensure-env.mjs`, `patch-bytecode-entry-forward.mjs`, and `verify-bytecode.mjs` run, matching `build:win`), plus the changelog entry `desktop/windows/changelog/unreleased/2026-08-linux-mac-bytecode-export-forward.json`.
- That identical change was already merged as `9d1ee50775` via #12172 — same author — merged 2026-08-25 14:43 UTC, about eight hours before this PR was opened. I compared both patches file-by-file and they are identical.
- This branch's merge base with `main` dates to Aug 22, so the 204-file diff shown above is almost entirely upstream work that has already landed on `main`. The branch now conflicts with `main` in 14 files (`firestore.indexes.json`, `backend/scripts/scan_async_blockers.py`, `backend/scripts/render_backend_runtime_env.py`, and others) purely because `main` carries newer versions of those files, so merging would bring nothing new and the conflict resolutions would risk reverting newer `main` content.

No action needed from you — full credit for the bytecode export-forward fix on Linux/macOS builds is yours via #12172, and it is already protecting those build paths. If you intended this PR to carry something additional, please open a fresh branch off the latest `main` and it will get a proper review.

---

PR #12230, opened 2026-08-25
fix(desktop-windows): reset onboarding when a different account signs in

formed2forge:
## Summary

- `reconcileAccountForSignIn` detected a UID change and wiped all user data, but `onboardingCompletedAt` survived because `teardownUserData` preserves onboarding state (device-scoped, intentional for same-user re-auth). Result: a second user signing in on the same machine skipped the wizard and landed directly in the home screen.
- Fix: call `resetOnboarding()` in the `stored !== uid` branch of `reconcileAccountForSignIn`, just before `teardownUserData()`. Same-user re-auth and first-ever sign-in are unaffected.

## Verification

```
node_modules/.bin/vitest --run src/renderer/src/lib/authTeardown.test.ts
✓ 7 tests pass (3 assertions updated to pin resetOnboarding call-count per branch)
```

Cannot exercise the live multi-account sign-in path (no app build in this environment); tests cover the four guard branches via `localStorage` + IPC mocks.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12230?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level (changes requested):
Thanks @formed2forge - the actual change here is a good, well-scoped fix worth landing. One mechanical problem first: the branch needs a rebase onto current main.

The fix itself, verified on head d4a939a0:

- `desktop/windows/src/renderer/src/lib/authTeardown.ts` (+5/-2): `reconcileAccountForSignIn` now calls `resetOnboarding()` in the `stored !== uid` branch, before `teardownUserData()`. This is correct and ordering-safe: `resetOnboarding` (already on main in `preferences.ts`) deletes `onboardingCompletedAt`/`onboardingStep` and persists the preferences blob, and teardown's final `clearUserScopedPreferences()` re-reads that persisted blob and only strips `displayName`/`goal`, so the reset survives the wipe. Same-user re-auth and first-ever sign-in still skip the wipe, matching the file's documented account-switch contract.
- `desktop/windows/src/renderer/src/lib/authTeardown.test.ts` (+11/-2): `resetOnboarding` is added to the `./preferences` mock, and the assertions pin the call count per branch (once on account switch, never on same-uid re-auth or no-stored-uid) - exactly the regressions this fix could reintroduce.
- `desktop/windows/changelog/unreleased/2026-08-onboarding-per-account-reset.json` (+5): changelog entry present, matching the repo's per-fix convention.

Why a rebase is needed: the branch contains a merge of roughly three weeks of upstream history (Aug 17-24) under rewritten SHAs, so the PR displays 205 changed files that are not your edit, and GitHub now reports it as conflicting - the fork base is 167 commits behind main. Merging as-is is not possible, and hand-resolving those conflicts from the stale side risks reverting newer upstream versions of files like `.github/scripts/verify_auto_backend_release_admission.py` and `backend/scripts/render_backend_runtime_env.py`. I cherry-picked your single commit onto current main and it applies cleanly with no conflicts, so your fix itself does not conflict at all - rebasing (or resetting the branch onto main and re-committing just this change) will collapse the PR to exactly the three files above.

For reviewers: the PR-level diff flags `backend/AGENTS.md`, `desktop/macos/AGENTS.md`, `backend/charts/monitoring/README.md`, and `backend/docs/runbooks/silent-failure-detection.md` as agent-guidance/documentation changes. These are base-branch replay artifacts, not contributor edits - the head commit touches none of them, and versus current main the net change to all four is zero (two are byte-identical; two differ only because main carries newer versions). This PR introduces no new agent guidance.

Requesting changes only for the rebase - the underlying fix is good as-is, and it should land quickly once the branch is current.

---

formed2forge:
Rebased onto current main — PR now shows 4 files (the 3 fix files + changelog). Mergeable.

kodjima33 (approved):
Legit, narrowly-scoped fix: teardownUserData intentionally preserves onboarding state for same-user re-auth, but that also incorrectly applied to a genuinely different account signing in, letting them skip the wizard. resetOnboarding() is now called only on the stored!==uid branch. Verified not already fixed on main. Desktop bug-fix fast lane, owner-override merge.

PR #12248, opened 2026-08-26
Fix/web app tasks UI

Yny4ii:
## What changed and why

Three Tasks UI bugs on `web/app` `/tasks`: the due-date popover was clipped by `overflow-hidden`, a double-click on `+1d` entered row selection because `click` `stopPropagation` does not stop `dblclick`, and deleting a task unmounted the whole list because cache-invalidation refetch flipped `loading` to true.

none

## How it was verified

- `cd web/app && bun run typecheck` — pass
- `bunx vitest run --config vitest.config.mts src/components/tasks/__tests__/TaskRow.test.tsx src/hooks/__tests__/useActionItems.test.tsx` — 2 files, pass
- `NODE_OPTIONS=--no-experimental-webstorage bun run check` — typecheck + 328 tests pass (local Node 25 needs that flag; CI Bun 1.3.14 is the suite authority)
- `bun run build` — pass (firebase SW generated, 26 routes)
- Pre-push `scripts/pre-push` — pass; pushed `fix/web-app-tasks-ui`

Manual: exercised the list-view date picker (popover no longer clipped), double-click on `+1d` (snooze without selection mode), and delete (remaining rows stay on screen). Could not sign in against production Firebase from a contributor env; those checks were on the local Tasks UI with loaded tasks.

## Tests

- Date picker clip: no test — CSS `overflow-hidden` on the pending-list wrapper; clipping only shows in layout. Asserting a class string would be a static tripwire.
- Double-click `+1d`: `web/app/src/components/tasks/__tests__/TaskRow.test.tsx` — dblclick snooze does not call `onEnterSelectionMode`.
- Delete unmount: `web/app/src/hooks/__tests__/useActionItems.test.tsx` — cache-invalidation refetch keeps `loading === false` and existing items mounted.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/app/src/components/tasks/TaskListView.tsx
+++ b/web/app/src/components/tasks/TaskListView.tsx
@@ -84,7 +84,7 @@ export function TaskListView({
   return (
     <div className="flex flex-col">
       {/* Pending tasks - flat list */}
-      <div className="border border-bg-tertiary rounded-lg overflow-hidden">
+      <div className="border border-bg-tertiary rounded-lg">
         <AnimatePresence mode="popLayout">
           {filteredPending.map((task, index) => (
             <TaskRow
--- a/web/app/src/components/tasks/TaskRow.tsx
+++ b/web/app/src/components/tasks/TaskRow.tsx
@@ -355,6 +355,7 @@ export function TaskRow({
             animate={{ opacity: 1 }}
             exit={{ opacity: 0 }}
             transition={{ duration: 0.1 }}
+            onDoubleClick={(e) => e.stopPropagation()}
             className="flex items-center gap-0.5 flex-shrink-0"
           >
             <button
--- a/web/app/src/components/tasks/__tests__/TaskRow.test.tsx
+++ b/web/app/src/components/tasks/__tests__/TaskRow.test.tsx
@@ -1,0 +1,39 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { describe, expect, it, vi } from 'vitest';
+import { TaskRow } from '@/components/tasks/TaskRow';
+import type { ActionItem } from '@/types/conversation';
+
+function task(overrides: Partial<ActionItem> = {}): ActionItem {
+  return {
+    id: 'task-1',
+    description: 'Preview task',
+    completed: false,
+    due_at: new Date().toISOString(),
+    ...overrides,
+  };
+}
+
+describe('TaskRow', () => {
+  it('does not enter selection mode when double-clicking +1d', async () => {
+    const onSnooze = vi.fn();
+    const onEnterSelectionMode = vi.fn();
+    const user = userEvent.setup();
+
+    render(
+      <TaskRow
+        task={task()}
+        onToggleComplete={vi.fn()}
+        onSnooze={onSnooze}
+        onDelete={vi.fn()}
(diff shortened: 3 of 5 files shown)


Git-on-my-level:
Thanks @Yny4ii — nicely scoped set of fixes, and the verification section of the description is exactly what makes this fast to review.

I traced all three fixes through the code and ran the full web/app vitest suite against this head (60 files / 328 tests, all green, including the two new test files — useful since CI here only ran the AI reviewer).

Per-file notes:

- `web/app/src/hooks/useActionItems.ts` — the `hasLoadedRef` guard is the right fix for the delete-unmounts-list bug. The chain checks out: `deleteActionItem` invalidates the action-items cache (`lib/api.ts`) → the `onCacheInvalidation` subscription refetches → the old `setLoading(true)` unmounted the populated list in TaskHub, whose render gate is `!loading && items.length > 0`. Keeping `loading` as first-paint-only is correct stale-while-revalidate behavior, and the error path now preserving stale items instead of `setItems([])` is a real improvement. I compared the rest of the file (snooze / setDueDate / updateDescription / removeItem / bulk*) line-by-line — it's Prettier reformatting only, no behavior change. `useActionItems` has a single consumer (TaskHub), so the loading-semantics change is fully contained.
- `web/app/src/components/tasks/TaskRow.tsx` — `onDoubleClick={(e) => e.stopPropagation()}` on the hover-actions wrapper is the correct layer for this: `click`-level `stopPropagation` doesn't stop the separate `dblclick` event from reaching the row's `handleRowDoubleClick`, and putting it on the wrapper covers every hover button, not just `+1d`.
- `web/app/src/components/tasks/TaskListView.tsx` — dropping `overflow-hidden` un-clips the absolutely-positioned date popover (`top-full right-0 z-50`). One small thing to keep an eye on: the wrapper still has `rounded-lg` and rows animate `height: 0` on exit — if exiting-row content is ever visible past the rounded corners during the animation, a clip on the row itself would restore it without re-clipping the popover. Cosmetic at most; not a blocker.
- `web/app/src/components/tasks/__tests__/TaskRow.test.tsx` and `web/app/src/hooks/__tests__/useActionItems.test.tsx` — both assert the user-visible regression (dblclick snooze doesn't enter selection mode; invalidation refetch keeps `loading === false` with items mounted) rather than implementation details, which is exactly what these tests should do. Skipping a layout test for the CSS clip, with the rationale in the description, is a reasonable call.

No blockers from my side — leaving the merge sign-off to a maintainer.

---

kodjima33 (approved):
Three real, well-scoped UI bug fixes (date-popover clip, dblclick selection, delete-unmount) with tests for 2 of 3. Confidence gate score ~2/5 (no linked issue, diff over 300-line threshold) — approving only per policy, not merging.

kodjima33 (approved):
Confidence 4/5: stated root cause for 3 concrete UI bugs, fix-shaped title, CI green, tests added for the double-click and unmount regressions. Merging.

PR #12250, opened 2026-08-26
fix(web-app): stop Home from snapping chat history back down

Yny4ii:
## What changed and why

Home shares one scroller for history, Currents, and the live exchange. After history loaded it called `scrollIntoView` on Currents, and live-edge follow used `scrollIntoView` on the same ancestor, so scrolling history upward yanked the viewport back down. Launch placement now runs once per session and yields as soon as the reader moves; live follow only runs while pinned to the bottom (or on a newly opened exchange).

none

## How it was verified

- Scrolled chat history up on `/home` (worktree at http://localhost:3000): the viewport stayed on the history instead of snapping back to Currents.
- `cd web/app && NODE_OPTIONS=--no-experimental-webstorage bun run check` — typecheck clean, 334 tests passed.
- Pre-push: 15 manifest checks passed, including `web-app-checks` and `failure-class-protocol`.

## Tests

- `web/app/src/components/home/__tests__/HomePage.test.tsx` — after a wheel on history, delayed `loadHistory` must not `scrollIntoView` Currents.
- `web/app/src/components/chat/__tests__/ChatTranscript.test.tsx` — autoScroll does not follow a stream once the scroller has left the live edge; it still places a new exchange and follows while pinned.
- `web/app/src/lib/__tests__/scrollEdges.test.ts` — `shouldFollowLiveEdge` pins the follow/force cases without a DOM.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/app/src/components/chat/ChatTranscript.tsx
+++ b/web/app/src/components/chat/ChatTranscript.tsx
@@ -6,6 +6,11 @@ import Image from '@tschk/moonshine-next/image';
 import { Brain } from 'lucide-react';
 import type { ClientMessage } from '@/types/conversation';
 import { cn } from '@/lib/utils';
+import {
+  nearestVerticalScroller,
+  scrollEdgesOf,
+  shouldFollowLiveEdge,
+} from '@/lib/scrollEdges';
 import { ChatMarkdown } from './ChatMarkdown';
 
 /**
@@ -65,11 +70,31 @@ export function ChatTranscript({
   autoScroll = true,
 }: ChatTranscriptProps) {
   const messagesEndRef = useRef<HTMLDivElement>(null);
+  const placedExchangeRef = useRef(false);
 
-  // Scroll to bottom when messages change or streaming text updates
+  // Follow the live edge only while the reader is already there. Home's
+  // history sits in the same overflow ancestor, so a blanket scrollIntoView
+  // here is what yanks an upward history gesture back down.
   useEffect(() => {
-    if (autoScroll) messagesEndRef.current?.scrollIntoView({ behavior: 'smooth' });
-  }, [autoScroll, messages, streamingText]);
+    if (!autoScroll) {
+      placedExchangeRef.current = false;
+      return;
+    }
+    const end = messagesEndRef.current;
+    if (!end) return;
+    const scroller = nearestVerticalScroller(end.parentElement);
+    const pinnedToBottom = scroller
+      ? scrollEdgesOf({
+          scrollTop: scroller.scrollTop,
+          scrollHeight: scroller.scrollHeight,
+          clientHeight: scroller.clientHeight,
+        }).atBottom
+      : true;
+    const force = !placedExchangeRef.current;
+    if (!shouldFollowLiveEdge({ pinnedToBottom, force })) return;
+    placedExchangeRef.current = true;
+    end.scrollIntoView({ behavior: force ? 'auto' : 'smooth' });
+  }, [autoScroll, messages, streamingText, isStreaming, currentThinking]);
 
   if (isLoading && messages.length === 0) {
     return (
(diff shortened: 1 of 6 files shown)


Git-on-my-level:
Thanks @Yny4ii — clean diagnosis, well-scoped fix, and the tests actually reproduce the bug. Reviewed all six changed files:

**Root cause check:** the old effect in `ChatTranscript.tsx` called `messagesEndRef.scrollIntoView()` on every `messages`/`streamingText` update, and `scrollIntoView` walks every overflow ancestor — on Home that ancestor is the page scroller that also holds chat history, so any stream update while reading history yanked the viewport back down. Confirmed the new gating addresses exactly that.

- `web/app/src/lib/scrollEdges.ts` — `shouldFollowLiveEdge` (`force || pinnedToBottom`) and `nearestVerticalScroller` (closest ancestor with computed `overflow-y: auto|scroll|overlay`) are small, DOM-free testable, and reuse the existing `SCROLL_EDGE_EPSILON` convention. Good home for the rule.
- `web/app/src/components/chat/ChatTranscript.tsx` — live-edge follow now reads the shared scroller's `atBottom` via `scrollEdgesOf` and only scrolls when pinned or on first placement (`placedExchangeRef`). `pinnedToBottom` defaults to `true` when no scroller is found, preserving prior behavior on non-scrolling hosts, and the `autoScroll=false` early-return defensively resets placement. Adding `isStreaming`/`currentThinking` to the deps also means the thinking indicator now keeps a pinned reader pinned, which is a small improvement.
- `web/app/src/components/home/HomePage.tsx` — `readerOwnsViewportRef` + `onWheel`/`onTouchMove`/`onPointerDown` suppress the one-time post-`loadHistory` `currentsRef.scrollIntoView` after any reader interaction, and the ref resets per app/session change so launch placement still works when switching threads. Keying the effect to selection via `loadHistoryRef` is fine (`loadHistory` is already identity-stable while the selection holds). One nuance worth knowing, not blocking: a pointer-down anywhere in the scroller — e.g. tapping a quick prompt before history resolves — also counts as reader-owned and skips the Currents placement.
- `web/app/src/components/chat/__tests__/ChatTranscript.test.tsx` — pins all three behaviors with a fake scroller: force-place at `scrollTop: 100`, no follow mid-stream away from the edge, follow while pinned at 1200 (= 2000 − 800). Solid.
- `web/app/src/components/home/__tests__/HomePage.test.tsx` — the delayed-`loadHistory` + wheel test reproduces the reported bug precisely; the added `loadHistory.mockReset()`/`mockResolvedValue(undefined)` in `beforeEach` also prevents leakage into the existing Currents-ordering tests.
- `web/app/src/lib/__tests__/scrollEdges.test.ts` — DOM-free pinning of the follow/force truth table.

I also checked consumers: `ChatTranscript` has exactly one mount point in the web app (Home, with the history transcript on `autoScroll={false}`), so the behavior change is contained to the current-exchange section.

One ask before merge: GitHub only shows the cubic review check on this PR — the manifest/test runs cited in the description were pre-push, so please make sure the web-app checks run (or are re-run) on this head so the suite is visible to maintainers.

Leaving final sign-off to a human maintainer; no urgency — this is solid work and a nice first contribution.

---

kodjima33 (approved):
Confirmed bug fix (Home snapping chat history back to Currents), well-scoped diff with root cause explained and 3 new/updated test files, no CI failures, not already fixed on main. Merging.

PR #12253, opened 2026-08-26
fix(desktop-windows): treat pi-mono agent_settled as advisory

formed2forge:
## Summary

The pi-mono library emits an `agent_settled` event to signal that the
transcription model has stabilised. The desktop app was treating this
as a hard unknown state, blocking the UI unnecessarily.

**Change:** downgrade `agent_settled` from an unknown/blocking state to
an advisory signal so the UI continues rendering while settlement occurs.

## Files changed

- `desktop/windows/src/` — classify `agent_settled` as advisory, not unknown

## Test plan

- [ ] `pnpm test` passes
- [ ] Live transcription session with a pi-mono device shows no UI freeze
  during the settlement window

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks for this — and good news: the fix is already in `main`. :)

**What I checked**

- The one novel commit at the tip of this branch (`9690d7f`) adds a `case 'agent_settled':` advisory no-op to the RPC event switch in `desktop/windows/src/main/codingAgent/piMono.ts` (+4, with a comment that `turn_end` remains authoritative), plus a correlation test in `piMono.test.ts` (+24) asserting that the active prompt generation and pending request survive `agent_settled` and only resolve when the authoritative `turn_end` arrives.
- `main` already contains exactly that change: it landed via #12173 — authored by you — merged Aug 25, the day *before* this PR was opened. The resulting file blobs are byte-identical on both sides, so full credit for the fix is already on record. :tada:

**Why this PR shows ~200 changed files**

Your branch's merge-base with `main` predates the v0.12.208 releases and it carries 43 upstream commits that were squash-merged separately, so GitHub's three-dot diff counts them all. Merging would conflict and effectively *revert* newer work — GitHub reports the branch as conflicting with `main`, and for example `backend/AGENTS.md` and `backend/charts/monitoring/README.md` on this branch are strictly older than the current copies on `main`.

Since the substance is already merged, I'm closing this as superseded by #12173 — nothing further is needed from you. For your next PR, syncing your fork first (`git fetch upstream && git rebase upstream/main`) will make the diff show only your own changes. Thanks again — the advisory-vs-authoritative distinction in the event switch is exactly right, and pinning `turn_end` as the sole completion authority in the test is a nice touch!

---

PR #12192, opened 2026-08-25
test(desktop): isolate Rewind owner snapshot authority

aryanorastar:
## What changed and why

Addresses the Rewind/auth occurrence reported in #12039. The regression test changed `auth_userId` directly even though runtime owner authorization is process-wide and deliberately revokes out-of-band owner changes. That made the result depend on which owner-bound suite ran first and produced the observed `anonymous` versus `auth-leading-…` failure.

The test now recreates that revoked-authority state deliberately, then establishes and restores its owner through `RuntimeOwnerAuthorityTestFixture`. This follows the same serialized transition boundary as production and automatically places the suite in the runner's derived owner-isolation cluster.

## Product invariants affected

none

## How it was verified

- `xcrun swift test --package-path desktop/macos/Desktop --filter 'RewindCaptureExclusionGenerationTests'` — 9/9 passed.
- Repeated the exact contamination-recovery regression 50 times with `--skip-build` — 50/50 passed.
- Ran the full 5,769-test macOS Swift process three times. The repaired test passed under real process-order contamination in every run; each full run retained one unrelated baseline failure elsewhere, consistent with #12039's broader multi-suite report.

This is test-harness isolation only; there is no user-facing path or physical-device behavior to exercise.

## Tests

`RewindCaptureExclusionGenerationTests.testOwnerSnapshotStaysCurrentWhenAuthLeadsUnresolvedRewindDatabase` now starts from the exact revoked authority that previously arrived nondeterministically, then proves the auth-leading owner remains current while `RewindDatabase.currentUserId` is unresolved.

## Failure class (fixes)

Failure-Class: FC-hand-listed-test-isolation-membership

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #12039, opened 2026-08-22
Desktop Swift Static & Test Contracts: Kernel projection flake remains


aryanorastar:
## Current scope — updated 2026-09-10

Three legs from the original report are now addressed:

- Rewind owner isolation — fixed in #12192
- Chat transcript gesture ownership — fixed in #12406
- Suggested Tasks feedback-owner isolation — fixed in #12650

The remaining unresolved leg is:

- `KernelTurnRecordedProjectionTests`

PR #12664 added a leak-proof diagnostic seam and ruled out two hypotheses, but explicitly did not fix this Kernel projection flake. Keep this issue open for that investigation.

Next step: reproduce the Kernel failure on current `main` both in isolation and in the mixed suite, use the #12664 diagnostic seam to capture owner/revocation state around the reset, identify the ordering/shared-state/async cause, then add a deterministic regression test and link the fixing PR here.

---

`Desktop Swift Static & Test Contracts` has failed on four different PRs in one afternoon, each time on a **different** test suite, none of them in a subsystem the PR under test touched. Every one passed on a later run or is unrelated to the diff, so this reads as flakiness in the desktop Swift test lane rather than four separate regressions.

## Occurrences (2026-08-21 → 08-22, all UTC)

| PR | Run | Failing test |
|---|---|---|
| #12029 (Meet title detection, 3 files, no Rewind/auth code) | [32517586976](https://github.com/BasedHardware/omi/actions/runs/32517586976) | `RewindCaptureExclusionGenerationTests.testOwnerSnapshotStaysCurrentWhenAuthLeadsUnresolvedRewindDatabase` — `("anonymous")` is not equal to `("auth-leading-E9B86F1F-…")` |
| #11452 (dev-API `ask` endpoint + CLI, backend/Python) | [32520765434](https://github.com/BasedHardware/omi/actions/runs/32520765434) | `SuggestedTasksStoreTests.testNotMineAndAlreadyHandledPersistReasonAndResolveCandidate` — `("nil")` is not equal to `("Optional(…TaskIntelligenceFeedbackReason.not_mine)")` |
[...]

PR #12259, opened 2026-08-26
test(desktop): isolate FloatingBar notification owner authority

aryanorastar:
## What changed and why

The same defect #12192 fixes in `RewindCaptureExclusionGenerationTests`, in a suite #12039 does not name. I found it while checking whether the other suites in that issue shared a root cause — three of them don't, this one does.

Both owner-seeding tests wrote `auth_userId` directly:

```swift
defaults.set(owner, forKey: DefaultsKey.authUserId.rawValue)
```

Runtime owner authorization is process-wide and deliberately fails closed on exactly that write — `RuntimeOwnerIdentity.swift:51`:

```swift
} else if self.ownerID != normalized {
  revokeUnexpectedOwnerMismatch()
  return nil
}
```

Once any earlier owner-bound suite has bootstrapped a different owner, the raw write revokes the authority and it stays revoked without a real `beginTransition`/`endTransition` pair. So the outcome depended on which suite ran first.

Both tests now establish their owner through `RuntimeOwnerAuthorityTestFixture`, crossing the same serialized transition boundary production uses. Restore happens in `tearDown` rather than a `defer` — a `defer` can't `await`, and more importantly a failed assertion must not leave the process-wide authority revoked for whatever runs next, which is the contamination this is fixing. Adopting the fixture also enrolls the suite in the runner's derived owner-isolation cluster, since `swift-test-suites.sh` greps for that symbol.

## How it was verified

I checked the guard actually does the work rather than assuming it. Identical command, same three owner-bound suites, only the fix differing:

```
before:  Executed 28 tests, with 1 failure
         FloatingBarNotificationPreviewPolicyTests.swift:218: error:
         XCTAssertEqual failed: ("rejectedOwnerChange") is not equal to ("suppressed")

after:   Executed 28 tests, with 0 failures
```

```bash
xcrun swift test --package-path Desktop \
  --filter 'CaptureScreenToolTests|FloatingBarNotificationPreviewPolicyTests|ChatToolExecutorPolicyTests'
```

In isolation the suite passes both before and after (19/19) — which is exactly why the flake read as nondeterministic and why running it alone never reproduced it.

`xcrun swift build -c debug --package-path Desktop` clean.

Worth noting the symptom is not where I first expected it. I predicted `XCTUnwrap(RuntimeOwnerIdentity.currentOwnerId())` would throw on a nil owner; it doesn't — `currentOwnerId()` returns fine and the revocation surfaces later at the delivery gate as `rejectedOwnerChange`. Same root cause, different surfacing point, and the `rejectedOwnerChange` payload is precisely the "production error payload that looks like a real defect" the failure class describes.

This is test-harness isolation only; no production code is touched and there is no user-facing path to exercise.

none

## Relationship to the other #12039 suites

`ChatTranscriptGestureHarnessTests`, `KernelTurnRecordedProjectionTests`, and `SuggestedTasksStoreTests` have no `authUserId` writes and don't use the fixture, so they flake for some other reason. This PR does not address them and I have not diagnosed them.

Failure-Class: FC-hand-listed-test-isolation-membership

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Tests/FloatingBarNotificationPreviewPolicyTests.swift
+++ b/desktop/macos/Desktop/Tests/FloatingBarNotificationPreviewPolicyTests.swift
@@ -12,6 +12,23 @@ import XCTest
 /// enabled is the one case that falls back to a native system banner so the
 /// notification is never fully silenced.
 final class FloatingBarNotificationPreviewPolicyTests: XCTestCase {
+  /// Runtime owner authorization is process-wide and fails closed on an
+  /// out-of-band `authUserId` write, staying revoked for every later suite in
+  /// the xctest process. The two owner-seeding tests below therefore establish
+  /// their owner through the production transition boundary, and restore runs
+  /// in `tearDown` rather than a `defer` so a failed assertion cannot leave the
+  /// authority revoked for whatever runs next.
+  private var ownerFixture: RuntimeOwnerAuthorityTestFixture?
+
+  override func setUp() async throws {
+    ownerFixture = await RuntimeOwnerAuthorityTestFixture()
+  }
+
+  override func tearDown() async throws {
+    await ownerFixture?.restore()
+    ownerFixture = nil
+  }
+
   func testPreviewsAndBarEnabledShowsPreviewWithNoForcedBanner() {
     XCTAssertTrue(
       FloatingBarNotificationPreviewPolicy.shouldShowInBarPreview(
@@ -166,11 +183,9 @@ final class FloatingBarNotificationPreviewPolicyTests: XCTestCase {
   /// `presentContextDirectorNotification` makes this call return `.queued` from the
   /// banner path instead of `.suppressed`, failing the test.
   @MainActor
-  func testDirectorDeliveryWithDisabledCategoryToggleIsSuppressedAtTheEntryPoint() throws {
+  func testDirectorDeliveryWithDisabledCategoryToggleIsSuppressedAtTheEntryPoint() async throws {
     let defaults = UserDefaults.standard
     let pinnedKeys = [
-      DefaultsKey.authUserId.rawValue,
-      DefaultsKey.automationOwnerOverride.rawValue,
       NotificationService.masterEnabledDefaultsKey,
       NotificationService.frequencyDefaultsKey,
       DefaultsKey.desktopIsPaywalled.rawValue,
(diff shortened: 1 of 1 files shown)


Git-on-my-level:
Verified against the current head — this checks out cleanly, and the root-cause diagnosis is accurate.

- `FloatingBarNotificationPreviewPolicyTests.swift` is the only changed file, and the fix is exactly what the flake class in #12039 calls for: both owner-seeding tests previously wrote `auth_userId` directly into `UserDefaults`, which trips the fail-closed owner-mismatch guard in `RuntimeOwnerIdentity.capture(...)` (`RuntimeOwnerIdentity.swift:51-53`) once an earlier owner-bound suite has bootstrapped a different owner — so the outcome depended on suite order.
- The new `setUp`/`tearDown` pair adopting `RuntimeOwnerAuthorityTestFixture` matches the fixture's contract in `RuntimeOwnerAuthorityTestSupport.swift`: `establish(authOwnerID:)` crosses the same `performEffectiveOwnerTransition` boundary production uses, and `restore()` re-crosses it instead of raw-writing defaults. Moving owner restoration out of `defer` into `tearDown` is the right call — `defer` can't `await`, and more importantly a failed assertion must not leave the process-wide authority revoked for whatever runs next.
- Confirmed side benefit: `swift-test-suites.sh` derives its owner-isolation cluster by grepping for `RuntimeOwnerAuthorityTestFixture`, so this suite is now serialized with the other owner-bound suites automatically, matching the canonical prevention recorded in the hand-listed-test-isolation-membership failure class.
- Dropping `authUserId`/`automationOwnerOverride` from `pinnedKeys` is consistent — the fixture now owns those keys' lifecycle.

The before/after run transcript in the description (1 failure → 0 failures across the three owner-bound suites) is a credible verification of the guard doing the work. Thanks for tracing this through the guard mechanism rather than just pinning more keys — it pairs nicely with #12192 for the Rewind suite. Leaving for human maintainer review before merge.

---

PR #12261, opened 2026-08-26
fix(macos): restore shell clicks after reopening

### LINKED ISSUE #12249: macOS: transparent main shell intermittently ignores clicks over visible controls
**Product impact**

- Primary loop layer: UX-Polish

**Describe the bug**

On macOS, the signed-in transparent main shell can render normally while visible controls stop accepting clicks. Navigation buttons do not select, and search or composer fields do not reliably receive focus, making the shell feel like it is not a real application window.

The frameless appearance itself is understood to be intentional. The defect is that visible content sometimes behaves like the transparent dead area and passes mouse events through instead of remaining interactive.

**To Reproduce**

The failure is intermittent and currently appears to depend on shell runtime state:

1. Launch Omi Desktop while signed in.
2. Open the main shell from the floating/notch control.
3. Navigate among Chat, Brain, Tasks, Rewind, or Apps and try to focus a search or composer field.
4. In the affected state, the shell remains visible but clicks over visible controls do not select or focus them.

**Current behavior**

Visible controls can become non-interactive while the transparent main shell remains on screen.

This is not a general input failure on the Mac. On the same machine, a side-by-side named comparison bundle opened the same signed-in transparent shell, navigated from Chat to Brain, accepted text in Brain search, and remained at low CPU. That points to a build, route, or window-state boundary rather than a machine-wide macOS failure.

**Expected behavior**

Only transparent dead margins should pass clicks through. Every visible panel, navigation control, button, search field, and composer should keep the main window interactive.

**Screenshots**

The affected and working states can look identical, so a static screenshot does not show the mouse-event failure. A focused diagnostic capture can be added once the state is reproduced deterministically.

**user ID (can we access the user info to validate the bug?):**

Not included in a public issue. It can be provided privately if needed.

**Smartphone + device (please complete the following information):**

- Device: MacBook Pro (Mac16,5), Apple M4 Max, 36 GB
- OS: macOS 27.0, build 26A5421a
- Browser: N/A, desktop app
- App Version: Omi Desktop 0.12.213, build 12213

**Additional context**

Current upstream `ShellMouseInterceptionSync` updates the whole window's `ignoresMouseEvents` state at 30 Hz from `InkGlassHitRegions`. A missing or stale visible-surface registration during a shell rebuild could therefore classify visible content as dead air. This is a suspected boundary, not yet a confirmed root cause.

A local diagnostic build using conventional AppKit window chrome restored stable interaction, but that changes Omi's documented transparent-shell product design and is not proposed as the upstream fix. The upstream fix should preserve click-through for dead margins while behaviorally proving that visible controls remain interactive through route and hosting-view changes.

Related work:

- #7516 / #7593 fixed a separate transparent Omi window that intercepted clicks after a menu closed.
- #11941 is a separate draft fix for a hidden-window sizing/layout CPU loop and should remain independently scoped.

aryanorastar:
## What changed and why

Fixes #12249.

The macOS shell reuses the same `NSWindow` when it is dismissed and summoned again. While the window was hidden, mouse-interception reconciliation correctly switched `ignoresMouseEvents` on, but the synchronizer then stopped polling because the window was no longer visible. Ordering that same window back on screen did not trigger another reconciliation, so the visible shell could remain click-through until an unrelated pointer event repaired it. That made visible controls intermittently ignore the first click—or every click if no local event could reach the window.

This change observes AppKit's authoritative `NSWindow.isVisible` transition and immediately reruns the existing interception policy whenever visibility changes. It also releases that observation with the synchronizer's other lifecycle resources.

The regression test mounts a real AppKit window and exercises the production sequence directly: visible → ordered out → reconciliation enables pass-through → ordered front. It proves the shell restores mouse interception before the first user click, without depending on pointer movement or test-only state.

none

## How it was verified

- Reproduced test-first: the new AppKit lifecycle test failed against the previous implementation because the reopened window still ignored mouse events.
- Passed the exact regression after the fix.
- Passed 52 focused shell/window tests across `ShellClickThroughPolicyTests`, `GlassPanelHitRegionTests`, `ShellSummonTests`, and `ShellWindowChromeTests`.
- Repeated the visibility-transition regression 30 times: 30/30 passed.
- Passed the cross-surface agent contract smoke harness.
- Built and launched an isolated signed-in bundle (`com.omi.omi-12249-clicks`), dismissed and summoned the production shell, then used a real native pointer click on **Brain**; the selected route changed from Chat to Brain on the first click.
- Ran the complete macOS component suite: all launcher checks, 214 backend tests, and all shell suites passed. Three unrelated Swift cases failed or timed out under concurrent machine load; each exact case then passed serially (`ChatTranscriptGestureHarnessTests`, `MemoryAtlasPerformanceHarnessTests`, and `RewindArtifactGauntletTests`).

## Tests

```text
xcrun swift test --jobs 2 --package-path Desktop --filter ShellClickThroughPolicyTests/testOrderingShellBackOnScreenRestoresMouseInterception
xcrun swift test --jobs 2 --package-path Desktop --filter 'ShellClickThroughPolicyTests|GlassPanelHitRegionTests|ShellSummonTests|ShellWindowChromeTests'
./scripts/agent-logic-harness.sh --cross-surface-smoke
./test.sh
```

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12261?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/ShellClickThrough.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/ShellClickThrough.swift
@@ -49,6 +49,7 @@ enum ShellClickThroughPolicy {
 @MainActor
 final class ShellMouseInterceptionSync {
   private nonisolated(unsafe) var monitors: [Any] = []
+  private var visibilityObservation: NSKeyValueObservation?
   private var pollingCancellable: AnyCancellable?
   private weak var window: NSWindow?
 
@@ -72,6 +73,9 @@ final class ShellMouseInterceptionSync {
     {
       monitors.append(local)
     }
+    visibilityObservation = window.observe(\.isVisible, options: [.new]) { [weak self] _, _ in
+      MainActor.assumeIsolated { self?.sync() }
+    }
     sync()
   }
 
@@ -86,6 +90,7 @@ final class ShellMouseInterceptionSync {
     window?.ignoresMouseEvents = false
     window = nil
     pollingCancellable = nil
+    visibilityObservation = nil
   }
 
   func sync() {
--- a/desktop/macos/Desktop/Tests/ShellClickThroughPolicyTests.swift
+++ b/desktop/macos/Desktop/Tests/ShellClickThroughPolicyTests.swift
@@ -7,6 +7,7 @@ import XCTest
 /// frame — the reserved title-bar band, lane margins, and gaps between panels swallowed clicks
 /// aimed at other apps, which never activated (the shell-window dead zone). The policy passes the
 /// pointer through everywhere except visible content, the modal barrier's host, and the resize rim.
+@MainActor
 final class ShellClickThroughPolicyTests: XCTestCase {
   private let windowSize = NSSize(width: 960, height: 712)
 
@@ -59,5 +60,34 @@ final class ShellClickThroughPolicyTests: XCTestCase {
         isResizable: false,
         contentContains: { _ in false }),
       "a fixed-size window has no resize affordance to preserve")
+  }
+
+  /// The shell reuses one window across dismiss/summon. Reconciliation while that window is ordered
+  /// out must not leave it ignoring mouse events when AppKit puts it back on screen: while ignored,
+  /// neither its visible controls nor its local mouse monitor can recover the window.
+  func testOrderingShellBackOnScreenRestoresMouseInterception() {
+    let mouse = NSEvent.mouseLocation
+    let window = NSWindow(
+      contentRect: NSRect(x: mouse.x + 5_000, y: mouse.y + 5_000, width: 320, height: 240),
+      styleMask: [.borderless],
+      backing: .buffered,
+      defer: false)
+    window.orderFront(nil)
+    let sync = ShellMouseInterceptionSync(window: window)
+    defer {
+      sync.detach()
+      window.orderOut(nil)
+    }
+
+    XCTAssertFalse(window.ignoresMouseEvents)
+    window.orderOut(nil)
+    sync.sync()
+    XCTAssertTrue(window.ignoresMouseEvents, "the hidden shell previously entered pass-through mode")
+
+    window.orderFront(nil)
+
+    XCTAssertFalse(
+      window.ignoresMouseEvents,
+      "a visible shell must recover before the first click, without waiting for pointer movement")
   }
 }
--- a/desktop/macos/changelog/unreleased/20260826-shell-click-recovery.json
+++ b/desktop/macos/changelog/unreleased/20260826-shell-click-recovery.json
@@ -1,0 +1,3 @@
+{
+  "change": "Fixed the main window occasionally ignoring clicks after reopening"
+}


Git-on-my-level:
Thanks @aryanorastar — this is a textbook small fix: correct root-cause diagnosis, a five-line production change, and a regression test that exercises the real AppKit lifecycle instead of a mock.

Traced through all three files at head:

- `desktop/macos/Desktop/Sources/MainWindow/ShellClickThrough.swift` — the diagnosis checks out against the pre-existing code: `sync()` calls `setPollingActive(window.isVisible)`, and `FloatingBarMouseInterceptionPolicy.shouldIgnoreMouseEvents` returns `true` for a hidden window, so a dismissed shell ends up with `ignoresMouseEvents == true` and no polling timer. The only remaining reconciliation triggers were the mouse-movement monitors, which need pointer motion — so a re-summoned shell could sit there click-through until the mouse moved. Observing `\.isVisible` closes exactly that gap, and since the handler just calls the existing idempotent `sync()` (guarded `ignoresMouseEvents` assignment), both directions of the transition are handled with no new state. `[weak self]` avoids a retain cycle on the observation token, `MainActor.assumeIsolated` is the right tool for `NSWindow` KVO (always delivered on the main thread), and releasing `visibilityObservation` in `detach()` alongside `pollingCancellable` keeps teardown symmetric, with deinit auto-invalidation as the backstop.
- `desktop/macos/Desktop/Tests/ShellClickThroughPolicyTests.swift` — the new test pins the actual failure sequence: visible → `orderOut` + `sync()` (asserting the hidden shell really is in pass-through mode, which validates the test's precondition rather than assuming it) → `orderFront` → asserts recovery without any pointer movement. Placing the window at `mouse.x + 5_000` keeps the pointer outside the frame so the `ignoresMouseEvents` assertions are deterministic, and `defer { sync.detach(); window.orderOut(nil) }` tears down the 30 Hz polling timer the KVO path restarts. Adding `@MainActor` to the class is required and correct given the sync is main-actor-isolated.
- `desktop/macos/changelog/unreleased/20260826-shell-click-recovery.json` — matches the existing unreleased-entry convention (dated slug, single user-facing `change` line) and the wording tracks the fixed behavior. One honest-scoping note in your favor: the issue says "intermittently", and this mechanism (dismiss → summon with no pointer movement) explains that intermittency; if a repro path ever shows up that doesn't involve a dismiss/summon cycle, that would be a separate root cause — nothing here overclaims coverage.

The desktop CI lanes are green on this head, including the Swift build & tests with the new lifecycle test. No blockers from my side — leaving the formal approval and merge sign-off to a maintainer.

---

kodjima33 (approved):
Small, well-scoped desktop bug fix (fixes #12249) with regression test; holding at approve-only because Desktop Swift Static & Test Contracts is currently red (RuntimeOwnerIdentityTests failure, unrelated to this diff's ShellClickThrough.swift change) — hard floor blocks merge on a failing required check.

aryanorastar:
No @kodjima33 @Git-on-my-level @undivisible — the implementation is complete and this PR is ready from the author side.

The current red state is an unrelated CI flake in `RuntimeOwnerIdentityTests.testReplacementOwnerWaitsForNonCooperativePhysicalEffectToDrain`; this PR does not modify that test or its production path. Evidence:

- The same code at `c6048400bc` already passed the complete Desktop Swift CI run: https://github.com/BasedHardware/omi/actions/runs/32991584530
- The shell-click regression itself remains green.
- I ran the exact failing RuntimeOwner test 30 times on this PR base and another 30 times after refreshing against current `main`: **60/60 passed**.
- The downstream `Desktop Swift Build & Tests` check only failed because it mirrors the failed static/test-contract result; release compile passed.
- The PR is conflict-free and has maintainer approval.

I attempted to rerun the failed GitHub job directly, but GitHub restricts reruns to repository admins. A maintainer rerun of https://github.com/BasedHardware/omi/actions/runs/33064363595 is the remaining action; no product-code change is warranted for this failure.

Git-on-my-level:
Follow-up on the current red CI, since the flake analysis above asked for a maintainer-side check — I verified it independently against the Actions history:

- This head (`18e8b7f`) is code-identical to `c604840`: the only commit between them is the empty `chore(ci): retrigger` commit, and the full desktop CI suite passed on that exact tree in run [32991584530](https://github.com/BasedHardware/omi/actions/runs/32991584530).
- The failing suite, `RuntimeOwnerIdentityTests.testReplacementOwnerWaitsForNonCooperativePhysicalEffectToDrain`, asserts an async drain ordering (`effect_cancelled_and_finished` vs `effect_finished`) in the runtime-owner handoff path — nothing this diff touches. The change observes `NSWindow.isVisible` and re-runs the idempotent `sync()` in `ShellClickThrough.swift`; it has no interaction with owner transitions, and `ShellClickThroughPolicyTests` (including the new `testOrderingShellBackOnScreenRestoresMouseInterception`) ran green inside the same red run.
- The desktop test lane is flaking repo-wide today: several `main` commits (e.g. [c48562c](https://github.com/BasedHardware/omi/commit/c48562c7b8), [a62b15e](https://github.com/BasedHardware/omi/commit/a62b15e143)) failed the same two jobs within the same hours — one on an entirely different suite (`AgentPillLifecycleTests`) — while other `main` commits went green.
- `Desktop Swift Build & Tests` is the aggregation gate mirroring the static/test-contract result (`VERIFY_RESULT: failure`); the release-compile lane passed.

So no PR-side change is warranted for the red state. The one remaining step is operational: an admin rerun of [run 33064363595](https://github.com/BasedHardware/omi/actions/runs/33064363595) so the required checks can re-evaluate this unchanged tree. With the approval already on this head and the diff itself previously reviewed clean, nothing else is pending from my side.

---

PR #7093, opened 2026-04-29
fix(action_items): content-hash idempotency on POST /v1/action-items

builtbycnob:
## Summary

`database.action_items.create_action_item` allocated a fresh Firestore document id on every call (`action_items_ref.add(...)`), so a flaky-network retry from any HTTP client produced a duplicate row. This PR adds opt-in idempotency:

- **db layer** — `create_action_item(uid, data, idempotency_key=...)` looks up an existing document with that key before writing; returns the existing id on hit, persists the key on miss.
- **router layer** — `POST /v1/action-items` now computes `sha256(f"{uid}:{normalized_description}")` and passes it through. A retried POST of the same task collapses to the original document.

## Why

Production observation on an affected account: 8 distinct task descriptions × 5–6 backend duplicates each (~50 phantom rows). The dominant duplicating path is the staged-task promotion gap (#7092), but the desktop client also has a `retryUnsyncedItems` loop that re-pushes manual tasks whose initial sync didn't complete. Without a server-side idempotency key, every retry produced a fresh Firestore document — and once the row had a `backendId`, the desktop's `syncTaskActionItems` couldn't dedup it any longer.

This PR closes the retry-amplification gap so even an aggressive client cannot mint duplicates by hammering the endpoint.

## Diff

`database/action_items.py`:

```python
def create_action_item(uid: str, action_item_data: dict,
                        idempotency_key: Optional[str] = None) -> str:
    ...
    if idempotency_key:
        existing_query = action_items_ref.where(
            filter=FieldFilter('idempotency_key', '==', idempotency_key)
        ).limit(1)
        for doc in existing_query.stream():
            return doc.id

    ...
    if idempotency_key:
        action_item_data['idempotency_key'] = idempotency_key

    doc_ref = action_items_ref.add(action_item_data)[1]
    return doc_ref.id
```

`routers/action_items.py`:

```python
def _content_idempotency_key(uid: str, description: str) -> str:
    normalized = (description or '').strip().lower()
    return hashlib.sha256(f"{uid}:{normalized}".encode('utf-8')).hexdigest()


@router.post("/v1/action-items", ...)
def create_action_item(request: ..., uid: str = ...):
    ...
    idempotency_key = _content_idempotency_key(uid, request.description)
    action_item_id = action_items_db.create_action_item(
        uid, action_item_data, idempotency_key=idempotency_key
    )
```

## Backwards compatibility

[...]

PR #12266, opened 2026-08-26
Fix/web app duplicate task title

Yny4ii:
## What changed and why

`POST /v1/action-items` hashed the task title as an idempotency key, so creating a second task named like an existing one (`123`) returned the original document (same due date, gone after reload). Create is now idempotent only when the client sends `Idempotency-Key`; the web app sends a per-submit UUID, and the proxy forwards it.

none

## How it was verified

- `BACKEND_UNIT_TEST_FILE_LIST=tests/unit/test_action_item_idempotency.py BACKEND_PYTEST_XDIST=0 ./test.sh` → 11 passed
- `NODE_OPTIONS=--no-experimental-webstorage bunx vitest run --config vitest.config.mts src/lib/__tests__/actionItemList.test.ts` → 2 passed
- Did **not** exercise live `/tasks` against this backend: production still hashes titles until this change is deployed. Could not confirm two same-title tasks persist after reload on a real account.

## Tests

Regression: `backend/tests/unit/test_action_item_idempotency.py::test_router_does_not_hash_description_as_idempotency_key` — two POSTs with the same description and no header must both insert.

Web defensive list merge: `web/app/src/lib/__tests__/actionItemList.test.ts` — if the API still returns an existing id, the list replaces that row instead of prepending a ghost duplicate.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/database/action_items.py
+++ b/backend/database/action_items.py
@@ -287,9 +287,9 @@ def create_action_item(
             retry on flaky networks or duplicate event delivery — the previous
             behaviour silently allocated a fresh Firestore id on every call,
             producing user-visible duplicates. The key is stored on the
-            document so future calls can find it. Callers that want
-            content-based idempotency typically pass
-            ``hashlib.sha256(f"{uid}:{normalized_description}".encode()).hexdigest()``.
+            document so future calls can find it. Pass a per-attempt client
+            key (for example an ``Idempotency-Key`` header), not a hash of
+            the description: task titles are not unique.
         document_id: Optional caller-reserved Firestore document id. Reusing
             the id returns the existing document without rewriting it, making
             a crash-retried create deterministic.
--- a/backend/routers/action_items.py
+++ b/backend/routers/action_items.py
@@ -1,12 +1,11 @@
 import asyncio
-import hashlib
 import logging
 import uuid
 
 from utils.executors import postprocess_executor, submit_with_context
 
-from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response
-from typing import Optional, List
+from fastapi import APIRouter, Depends, Header, HTTPException, Query, Request, Response
+from typing import Annotated, Optional, List
 from datetime import datetime, timezone
 
 import database.action_items as action_items_db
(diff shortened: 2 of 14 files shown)


Git-on-my-level:
Thanks @Yny4ii — this is a clean fix for a real correctness bug, and the verification section is exactly how I wish every PR reported testing.

**What I verified**

- `backend/routers/action_items.py` — `_content_idempotency_key()` (the sha256 over uid + normalized title from #7093) is replaced by `_client_idempotency_key()`, which only strips/blank-checks a caller-supplied `Idempotency-Key` header (`Annotated[Optional[str], Header(alias='Idempotency-Key', max_length=256)]`; blank → None → always insert). The lookup stays scoped to `users/{uid}/action-items` with equality on the stored `idempotency_key` field, so there is no cross-user surface.
- `backend/database/action_items.py` — docstring-only change; the db-layer opaque-key support already existed.
- `backend/tests/unit/test_action_item_idempotency.py` — `test_router_does_not_hash_description_as_idempotency_key` (two POSTs, same description, no header → both insert) pins exactly the bug being fixed.
- `web/app/src/lib/api.ts` — `createActionItem` sends `Idempotency-Key: crypto.randomUUID()` per submit, and `fetchWithAuth` merges caller headers after the defaults, so the key survives.
- `web/app/src/app/api/proxy/[...path]/route.ts` — forwards `Idempotency-Key` the same way as the existing FCM headers.
- `web/app/src/lib/actionItemList.ts` + `__tests__/actionItemList.test.ts` — `prependOrReplaceById` is a good defensive touch: while production still runs the old hash behavior, a replayed id replaces the row instead of prepending a ghost duplicate that vanishes on reload.
- `web/app/src/hooks/useActionItems.ts` — `addItem` now routes through that helper. Note the +226/−194 here is almost all arrow-paren reformatting; the functional change is one line.
- `docs/api-reference/app-client-openapi.json` and the regenerated clients (macOS Swift, Windows TS, web app, admin, personas) expose the optional header consistently.

**One decision for maintainers**

Removing the server-side title-hash fallback changes the contract for every client, and only the web app sends the key today:

- macOS: `APIClient+TaskCatalog.createActionItem` posts to `v1/action-items` with no `Idempotency-Key` (the `taskIntelligenceMutation` helper used for work-intents/checkpoints does send one, but plain task creates don't).
- Flutter: `app/lib/backend/http/api/action_items.dart` `createActionItem` sends no key.

The hash was originally added (#7093) to stop flaky-network retry duplicates from the desktop client, so until desktop/Flutter adopt the header, a network-level retry on those clients can again produce a duplicate. On net this is still the right trade — current behavior silently returns the wrong task on same-titled creates — but whether to merge ahead of client adoption or together with it is a maintainer sequencing call. The regenerated clients already support the header, so a follow-up on the desktop side is straightforward.

Minor, non-blocking: the db idempotency lookup filters `completed == False` (pre-existing, unchanged here), so a retry that lands after the original task was completed will insert a new one. Worth knowing; not this PR's bug.

**CI note**

The failing checks are not from this branch: `web-app-checks` fails on a frozen `bun.lock` drift introduced on main (prettier was added to `web/app/package.json` without a lockfile update — other PRs fail identically), the Desktop Swift jobs fail with the same format-drift/sentinel signature seen on maintainer PRs today, and the backend `Build` failure is a pre-existing pyright unused-import in `backend/utils/subscription.py` that also fails on main.

---

kodjima33 (approved):
Real bug (title-hash idempotency collision) and clear fix, but diff exceeds the confidence-gate size ceiling and CI has multiple failing required checks, so approve-only per policy.

Git-on-my-level:
Follow-up on my earlier review — the head is unchanged (`c8c3231`), and @kodjima33's approval settles the one open sequencing call (merging ahead of desktop/Flutter adopting the `Idempotency-Key` header — the regenerated clients already support it, so follow-ups there are easy).

State check since that review:

- The core change still checks out: `backend/routers/action_items.py` `_client_idempotency_key()` (optional header, `max_length=256`, blank → always-insert) replacing the title-hash from #7093, with the web wiring intact — per-submit `crypto.randomUUID()` in `web/app/src/lib/api.ts`, forwarding in `web/app/src/app/api/proxy/[...path]/route.ts`, and the `prependOrReplaceById` merge from `web/app/src/lib/actionItemList.ts` wired into `addItem`.
- The five red checks all predate the current base, not this diff: PR Metadata Preflight / Build fail on the frozen `web/app` `bun.lock` drift, which main fixed in #12288 *after* these checks ran; the Desktop Swift jobs fail on format-lint/sentinel errors and an `AppState+Transcription.swift` compile issue — a file this PR never touches — and main's Desktop Checks are green as of today. A rebase onto current main should clear all five; worth doing before merge so this doesn't land carrying red CI.

Label updates: removed `needs-maintainer-review` (human approval landed on this head) and added `positive-signal` + `ci-failing-pre-existing`. Thanks again @Yny4ii — the reproduction and regression tests made this an easy approve.

---

undivisible:
Soft-path rebase onto current `main` opened as a follow-up PR (cherry-pick of the 2 commits; fork push blocked by OAuth missing `workflow` scope). Conflicts: router imports (JSONResponse+Header/Annotated) and `prependOrReplaceById` on create.

undivisible:
Superseded by soft-path squash-merge of #12921 (cherry-picked onto main; fork push blocked by OAuth missing workflow scope).

Issue #12275, opened 2026-08-27
main's Desktop Swift CI has read green since 08-26 18:24Z while the alertPresenter SILGen crash from d49f978512 sits unfixed


Git-on-my-level:
# `main`'s Desktop Swift CI has read green since 08-26 18:24Z while a real compiler crash sits unfixed underneath

## TL;DR

Two stacked defects, found while rebasing three unrelated telemetry PRs (#12267, #12269, #12271) onto current `main`:

1. **A live compiler crash.** `d49f978512` ("fix(desktop): present microphone alerts without blocking") landed a stored-property default initializer in `AppState.swift` that segfaults `swift-frontend` (signal 11) on the CI-pinned Xcode 16.4. It is still on `main` right now.
2. **A required-check gap that hides it.** `desktop-swift-ci.yml`'s path filter treats "no desktop files touched by the last commit's diff against a prior baseline" as license to skip `Desktop Swift Static & Test Contracts` / `Release Compile` / `Build & Tests` — and the aggregate check accepts `skipped` as equivalent to `success`. Four consecutive `main` pushes since the crash landed happened not to touch `desktop/**`, so the required checks read green through all four, even though the crash was never fixed.

Any PR that touches `desktop/**` and rebases onto current `main` inherits the crash immediately — which is how we found this.

## Part 1: the crash is real and reproduced three times on `main` itself

`AppState.swift` currently reads (line 322, unchanged since `d49f978512`):

```swift
var alertPresenter: any DesktopAlertPresenting = AppKitSheetAlertPresenter()
```

Emitting the `any DesktopAlertPresenting` existential erasure through the stored-property default-value generator crashes `swift-frontend` SILGen on Xcode 16.4 / Swift 6.1.2. Confirmed on the actual CI runner log (job [98274550806](https://github.com/BasedHardware/omi/actions/runs/32998616632/job/98274550806)):

```
error: emit-module command failed due to signal 11 (use -v to see invocation)
Please submit a bug report (https://swift.org/contributing/#reporting-bugs) and include the crash backtrace.
Stack dump:
```

Three consecutive `main` pushes that actually touched `desktop/**` all ran the real job and all failed on this:

| commit | touches `desktop/**` | `Desktop Swift Static & Test Contracts` | run |
| --- | --- | --- | --- |
| `d49f978512` (introduces the crash) | yes | **failure** | [32998616632](https://github.com/BasedHardware/omi/actions/runs/32998616632) |
[...]

PR #12279, opened 2026-08-27
fix(desktop): dodge the Xcode 16.4 SILGen segfault on alertPresenter (#12275)

kodjima33:
## Bug

`main`'s Desktop Swift CI has been failing on every commit that touches `desktop/**` since `d49f978512` landed. `AppState.alertPresenter`'s `any DesktopAlertPresenting` stored-property default-value initializer crashes `swift-frontend` SILGen (signal 11) on the CI-pinned Xcode 16.4 / Swift 6.1.2 toolchain. The crash is masked on `main`'s status badge only because the required-check gate treats a path-filtered `skipped` job as satisfying the check — see #12275 for the full trace across four `main` pushes.

## Root cause

```swift
var alertPresenter: any DesktopAlertPresenting = AppKitSheetAlertPresenter()
```

Emitting the `any DesktopAlertPresenting` existential erasure through the stored-property default-value generator crashes SILGen on this toolchain (confirmed via the CI job log linked in #12275, and reproduced locally: `xcrun swift build -c debug --package-path Desktop` on `main` HEAD before this fix fails the same way).

## Fix

Move the same default value to the same lifecycle point — `AppState.init()` — instead of the stored-property default. No behavior change; `AppState` has exactly one initializer, and `alertPresenter` is assigned before anything else runs in it. This is the same workaround #12275 says is already carried on three unrelated branches.

Fixing that unmasked a second, pre-existing bug in `AppState+Transcription.swift:848`: `alertPresenter.pauseQueueUntilAppActive()` referenced the property without explicit `self` inside an escaping closure, which is invalid under this package's `-strict-concurrency=complete -warnings-as-errors` settings. It was never reached by the compiler because the SILGen crash aborted first. Added the required `self.`.

## What I tested

- `xcrun swift build -c debug --package-path Desktop` — full app target, exit 0 (previously crashed with signal 11 on `main` HEAD).
- `xcrun swift test -c debug --package-path Desktop --filter AppStateAlertPresentationTests` — 11/11 tests pass, 0 failures (previously could not even compile the test target).

Running the pinned `swift-format` lint scope after both fixes above surfaced one more piece of the same masked breakage: `AppState+Permissions.swift:106` had formatting drift introduced by the same `d49f978512` commit — CI's static-lint step never ran against it because the SILGen crash on the same commit aborted the job before reaching that step. Applied the pinned formatter (mechanical output only, no behavior change) so `desktop-swift-format-lint` passes clean.

## Scope

Part 2 of #12275 (the CI gate that let `skipped` satisfy a required check) is out of scope here — it lives in `.github/workflows/`, which is out of bounds for this fix.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12279?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33:
Closing — superseded. Both fixes this PR proposed are already on `main`:

- The `alertPresenter` SILGen-crash workaround (moving the default init into `AppState.init()`) landed via a different PR in the meantime — `main` HEAD already has it (same fix, different wording).
- The pre-existing `AgentRuntimeProcessTests`/`BYOKPaywallTests` failures this PR's CI run hit were fixed by #12277 (merged 2026-08-27T03:49:10Z), which this branch predates.

Diffing this branch against current `main` now shows only a cosmetic comment-style difference plus a stale `desktop/macos/changelog/releases/0.12.223.json` that this branch would incorrectly delete (main has since shipped that release). Nothing left to land.

🤖 automated by hourly watchdog — closed as superseded, not merged (main already carries the equivalent fix)

PR #12293, opened 2026-08-27
feat(stt): route explicit-English sessions to Velma-2 English Fast v2

mdmohsin7:
## Why

The multilingual Velma-2 streaming endpoint is failing for our account. A direct
WebSocket probe with our production key, outside our infrastructure, reproduces it:

```
wss://.../api/velma-2-stt-streaming            → ConnectionClosedError 1011 (internal error)
wss://.../api/velma-2-stt-streaming-english-v2 → connects, returns partial_utterance
```

That matches production, where Modulate has been rejecting ~90% of streams for days
while Deepgram carries the load on rate-capped accounts.

The English-only model answers normally with the same credentials, so sessions that
declare English can be served today instead of waiting for the multilingual endpoint.

## What changed

`process_audio_modulate` picks the endpoint from the session language:

- `en` (incl. `en-US`, `en_GB`) → `velma-2-stt-streaming-english-v2`
- everything else, including `multi` → `velma-2-stt-streaming`, unchanged

`multi` stays on the multilingual model. Auto-detect resolves to any language and this
model is English-only, so routing auto-detect there would produce wrong transcripts
rather than an error.

## Known tradeoff: diarization

The documented parameter table for `velma-2-stt-streaming-english-v2` is only
`api_key`, `audio_format`, `sample_rate`, `num_channels`, `endpointing`, and its
documented utterance payload has no `speaker` field. `_handle_utterance` maps a
missing speaker to `SPEAKER_00`, so **English sessions may lose speaker separation**
and return single-speaker transcripts.

Probing showed the endpoint ignores undocumented parameters rather than rejecting
them, so `speaker_diarization=true` and `partial_results=true` are sent anyway in case
they are honoured. Whether diarization actually works there is unverified — it needs a
real two-speaker sample, which I have not run.

If diarization is lost, that is a quality regression for English users traded against
a currently ~90% failure rate. `MODULATE_ENGLISH_FAST_ENABLED=false` reverts routing
without a deploy.

## Expected impact

Measured over 20 minutes of production traffic:

```
multi   1,784   82%   unchanged
en        240   11%   → English Fast v2
other     151    7%   unchanged
```

Roughly a ninth of Modulate traffic. Useful while the multilingual endpoint is broken;
not a fix for the outage.

## Testing

- `test_modulate_english_fast_routing.py`: 10 cases covering English variants, the
  `multi`/non-English guard, parameters sent per endpoint, and the disable flag.
- Existing Modulate and STT policy suites pass (41 tests).
- Both endpoints probed live against the production key, including parameter
  acceptance and first-text-frame config.

### DIFF
--- a/backend/tests/unit/test_modulate_english_fast_routing.py
+++ b/backend/tests/unit/test_modulate_english_fast_routing.py
@@ -38,9 +38,13 @@ async def test_english_sessions_use_english_fast_model(language):
     uri = await _connect(language)
     assert streaming.MODULATE_ENGLISH_FAST_PATH in uri
     params = _params(uri)
+    assert params['endpointing'] == ['true']
+    # The model rejects a language parameter; it is English-only by construction.
     assert 'language' not in params
-    assert 'speaker_diarization' not in params
-    assert 'partial_results' not in params
+    # Undocumented for this model but ignored rather than rejected, so we ask for
+    # diarization in case it is honoured.
+    assert params['speaker_diarization'] == ['true']
+    assert params['partial_results'] == ['true']
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


mdmohsin7:
Closing: the English Fast v2 endpoint's documented utterance payload has no speaker field, so English sessions would lose diarization. Not worth trading speaker separation for coverage of ~11% of traffic.

PR #12294, opened 2026-08-27
fix(desktop): ground legacy sidebars and keep them clickable

igor-popov-dev:
## What changed and why

Give the old Home sidebar slot one real `InkGlass` surface. The primary navigation and the Settings menu now share that surface, so the area has both a visible frosted ground and the same registered mouse-hit ownership.

Previously the legacy slot sat directly on the transparent top-level window. Its menu rows added only a faint tint, while the shell knew only about the adjacent page glass; the sidebar therefore looked almost transparent and clicks could be passed through to Finder or another app behind Omi.

Modern Settings remains inside its existing `PageGlassLane`, so it does not gain a second material or a rectangular hit region over the panel's rounded corner cut-outs.

INV-NAV-1

## How it was verified

- Mounted the real `SidebarView` and `SettingsSidebar` in a transparent `NSWindow`. The regression test verifies that both legacy menus have an actual `NSVisualEffectView` using behind-window glass, a registered surface, and an accepted shell hit; modern Settings inherits its page panel and adds neither.
- Built and launched the separately identified `/Applications/omi-legacy-sidebar-clicks.app` against the isolated local Omi harness with the old Home design enabled. The sidebar rendered as its own frosted lane.
- Exercised the live UI: Settings `About` → `General` → `Account & Plan`, Back to the primary sidebar, then `Memories`. Every selection changed the visible destination while Omi remained active.
- Probed screen points inside both live legacy menus through `debug_hit_probe`; each reported `shellAccepts=true` with `glassSurfaces=2`.
- `GlassPanelHitRegionTests`: 4 passed.
- `ShellClickThroughPolicyTests`, `PageGlassLaneTests`, `SettingsGlassChromeTests`, and `GlassSurfaceReviewRegressionTests`: 21 passed.
- Pinned Swift formatting lint and `scripts/check_desktop_test_quality.py` passed.
- Full `desktop/macos/test.sh` from the first commit: launcher/contract checks passed and backend tests passed (214/214). The overall local run remained non-zero on four unrelated existing suites: three assert English-formatted byte/count strings while this Mac's `Locale.current` is `ru_KZ`, and `ProactiveListenEventTests` crashes inside `UNUserNotificationCenter` when XCTest is launched from Xcode's command-line bundle. The impacted hit-region and shell-policy suites passed both inside that run and in focused runs.

## Tests

`GlassPanelHitRegionTests.testLegacySidebarSlotOwnsGlassAndHitsForBothMenus` is the behavioral regression test. It fails if a legacy menu loses either its actual behind-window glass or the registered surface used by shell click-through policy, and it protects modern Settings from nested material.

## Failure class (fixes)

Failure-Class: none

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift
@@ -1324,26 +1324,24 @@ struct DesktopHomeView: View {
   @ViewBuilder
   private var sidebarSlot: some View {
     if showsPrimarySidebar {
-      ZStack {
-        SidebarView(
-          selectedIndex: $selectedIndex,
-          isCollapsed: $isSidebarCollapsed,
-          memoryDestinationRawValue: $memoryDestinationRawValue,
-          appState: appState,
-          ownsShellHitRegion: !isInSettings
-        )
-        .opacity(isInSettings ? 0 : 1)
-        .allowsHitTesting(!isInSettings)
-        if isInSettings { settingsSidebar }
+      LegacySidebarSurface {
+        ZStack {
+          SidebarView(
+            selectedIndex: $selectedIndex,
+            isCollapsed: $isSidebarCollapsed,
+            memoryDestinationRawValue: $memoryDestinationRawValue,
+            appState: appState
+          )
+          .opacity(isInSettings ? 0 : 1)
+          .allowsHitTesting(!isInSettings)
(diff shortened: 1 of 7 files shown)


kodjima33 (approved):
Desktop bug fix (owner-override): grounds the legacy sidebar with a real glass surface and fixes click-through to apps behind Omi. Regression tests included, CI not red. Merging.

PR #12304, opened 2026-08-27
fix(app): align Firebase plugins for SwiftPM

### LINKED ISSUE #11676: Align Firebase plugin versions so Swift Package Manager can be re-enabled
Flutter 3.44 turned Swift Package Manager on by default for iOS/macOS (`swiftPackageManager` gained `enabledByDefault: true` on stable in `flutter_tools/lib/src/features.dart`). When Codemagic moved to 3.44.5, every iOS archive started failing at SPM resolution:

```
xcodebuild: error: Could not resolve package dependencies:
  Failed to resolve dependencies Dependencies could not be resolved because
  'firebase_crashlytics-4.3.2' depends on 'flutterfire' 3.11.0-firebase-core-swift and
  'firebase_messaging-15.2.5' depends on 'flutterfire' 3.13.0-firebase-core-swift.
```

The conflict predates the Flutter bump — CocoaPods tolerated the skew, SPM cannot, because it will not hold two versions of the same package.

| Pinned in `app/pubspec.yaml` | Declares |
|---|---|
| `firebase_core: 3.13.0` | — |
| `firebase_messaging: 15.2.5` | `firebase_core: ^3.13.0` |
| `firebase_crashlytics: 4.3.2` | `firebase_core: ^3.11.0` |

`firebase_crashlytics` is the outlier, two minor versions behind the pinned core.

**Stopgap already landed:** SPM is disabled at the project level in `app/pubspec.yaml` (`flutter: config: enable-swift-package-manager: false`), restoring the CocoaPods behaviour that has been shipping. Flutter warns that disabling "will not be allowed in a future version", so this is borrowed time.

**What this issue tracks:** bump `firebase_crashlytics` to the release built against `firebase_core 3.13.0` so both plugins want the same `flutterfire` package, then remove the flag. Per pub.dev, `firebase_crashlytics 4.3.5` declares `firebase_core: ^3.13.0`; 4.3.6+ move to 3.13.1 and beyond, so the whole Firebase set may be worth moving together rather than nudging one pin.

Acceptance is an actual iOS archive with the flag removed — a green `pub get` proves nothing here, since the failure is in Xcode's SPM resolution.

arhxam:
## What changed and why

Align `firebase_crashlytics` with the repository's `firebase_core` 3.13 family and remove the temporary Flutter Swift Package Manager opt-out. A contract test locks the compatible plugin set so the skew cannot silently return.

Closes #11676

none

## How it was verified

- `flutter pub get --enforce-lockfile` — resolved the checked-in lockfile without changes.
- `flutter test test/unit/flutterfire_spm_contract_test.dart` — the aligned Firebase versions and enabled SPM contract passed.
- Inspected the resolved `firebase_auth`, `firebase_messaging`, and `firebase_crashlytics` plugin manifests — each now requires the `firebase_core` 3.13 family.
- `pod install`, then `flutter build ipa --release --flavor dev --no-codesign` — Flutter added SwiftPM integration and resolved the native graph successfully, including Firebase iOS SDK 11.15.0 plus aligned `firebase_core` 3.13.0, `firebase_auth` 5.5.3, `firebase_crashlytics` 4.3.5, and `firebase_messaging` 15.2.5. No Firebase/SwiftPM version conflict occurred.
- Xcode reached the archive action but this machine lacks the watchOS 26.2 platform required by the scheme's embedded Watch app. The remaining failure is `watchOS 26.2 must be installed in order to archive the scheme`, not package resolution or compilation skew. A signed maintainer/Codemagic archive is still the final acceptance gate.

## Tests

- [x] Added a regression contract for the Firebase version family and SPM opt-out removal.
- [x] Resolved dependencies from the committed lockfile and checked the native plugin compatibility family.
- [x] Generated the CocoaPods workspace and resolved Flutter's hybrid SwiftPM/CocoaPods dependency graph.
- [ ] Completed signed iOS + Watch archive (blocked locally by missing watchOS 26.2 platform and signing credentials).

Failure-Class: none

### DIFF
--- a/app/pubspec.lock
+++ b/app/pubspec.lock
@@ -541,10 +541,10 @@ packages:
     dependency: "direct main"
     description:
       name: firebase_crashlytics
-      sha256: "6273ed71bcd8a6fb4d0ca13d3abddbb3301796807efaad8782b5f90156f26f03"
+      sha256: f3fa4a17c2f061b16b2e3ac7aaed889ae954b8952d0fd3e0009a9870cde7bbd2
       url: "https://pub.dev"
     source: hosted
-    version: "4.3.2"
+    version: "4.3.5"
   firebase_crashlytics_platform_interface:
     dependency: transitive
     description:
--- a/app/pubspec.yaml
+++ b/app/pubspec.yaml
@@ -42,7 +42,7 @@ dependencies:
   firebase_core: 3.13.0
   firebase_auth: 5.5.3
   firebase_messaging: 15.2.5
-  firebase_crashlytics: 4.3.2
+  firebase_crashlytics: 4.3.5
 
   # Auth
   google_sign_in: 6.2.2
@@ -182,14 +182,6 @@ flutter_launcher_icons:
   image_path: "assets/images/app_launcher_icon.png"
 
 flutter:
-  # Flutter 3.44 turned Swift Package Manager on by default for iOS/macOS.
-  # firebase_crashlytics 4.3.2 is built against firebase_core 3.11.0 while
-  # firebase_messaging 15.2.5 is built against 3.13.0, so SPM cannot resolve a
-  # single `flutterfire` package version and every iOS archive fails. CocoaPods
-  # tolerates the skew. Re-enable once the Firebase plugin versions are aligned;
-  # Flutter will eventually stop honouring this. https://github.com/BasedHardware/omi/issues/11676
-  config:
-    enable-swift-package-manager: false
   generate: true
   uses-material-design: true
   assets:
--- a/app/test/unit/flutterfire_spm_contract_test.dart
+++ b/app/test/unit/flutterfire_spm_contract_test.dart
@@ -1,0 +1,18 @@
+import 'dart:io';
+
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  test('FlutterFire pins share firebase_core 3.13 and SPM is enabled', () {
+    final source = File('pubspec.yaml').readAsStringSync();
+
+    expect(source, contains('firebase_core: 3.13.0'));
+    expect(source, contains('firebase_messaging: 15.2.5'));
+    expect(source, contains('firebase_crashlytics: 4.3.5'));
+    expect(
+      source,
+      isNot(contains('enable-swift-package-manager')),
+      reason: 'the temporary SPM opt-out must be removed',
+    );
+  });
+}


Git-on-my-level:
Verified the version algebra behind this fix, and it checks out:

- `firebase_crashlytics` 4.3.2 -> 4.3.5 in `app/pubspec.yaml` is the right pin: on pub.dev, 4.3.5 declares `firebase_core: ^3.13.0`, which now matches `firebase_auth` 5.5.3 and `firebase_messaging` 15.2.5 (both `^3.13.0`) against the pinned `firebase_core` 3.13.0. That removes the `flutterfire` 3.11.0-vs-3.13.0 SPM conflict from #11676 while leaving the rest of the Firebase set untouched (4.3.5 specifically, since 4.3.6+ already requires core 3.13.1+).
- `app/pubspec.lock` changes only the crashlytics entry (version + sha256), and the sha256 matches the published pub.dev archive hash for 4.3.5 — no other lockfile drift.
- The `enable-swift-package-manager: false` block removed from `app/pubspec.yaml` was the only reference to the opt-out in the repo, so nothing is left stale. The new contract test in `app/test/unit/flutterfire_spm_contract_test.dart` is a nice touch — it locks the aligned pin set and the opt-out removal so the skew can't silently return. (Minor note: it reads `pubspec.yaml` via a cwd-relative `File(...)`, which is fine under `flutter test` from `app/`, just cwd-dependent.)

One merge gate before this lands: #11676 defines acceptance as an actual iOS archive with the flag removed, and neither this PR (local disk-full, honestly disclosed — thanks) nor CI (the Build/iOS jobs are skipped on this PR) has produced one. SPM resolution happens in Xcode, so a green `pub get` and Dart CI don't exercise it — and re-enabling SPM moves the whole iOS/macOS native resolution path, not just Firebase. A maintainer-run archive (e.g. the Codemagic dev flavor) should gate the merge; since the description says "Closes #11676", if that archive surfaces any other plugin skew, please reopen the issue rather than letting it close with the remainder untracked.

Tight, well-scoped fix — thanks, and credit for adding the regression contract.

---

kodjima33 (approved):
Verified genuine Firebase/SwiftPM version-skew bug (unfixed on main, SPM opt-out still present), scoped diff, contract test added, CI green. Confidence 5/5.

mdmohsin7:
This broke iOS builds. `flutter build ipa` on macOS now fails at link:

```
Error (Xcode): 40 duplicate symbols
Error (Xcode): Linker command failed with exit code 1 (use -v to see invocation)

Your project uses both CocoaPods and Swift Package Manager, which can cause the above
error. It may be caused by there being both a CocoaPod and Swift Package Manager
dependency for the following module(s): GTMSessionFetcherCore.
```

The crashlytics 4.3.2 → 4.3.5 alignment is correct. The problem is dropping
`enable-swift-package-manager: false` alongside it — that assumed the version skew was the
only reason the flag existed. It wasn't. The flag was also masking a second SPM problem
that 06e59e7921 never named, because resolution failed before anything reached the linker.

**Why it duplicates.** With SPM on, plugin resolution splits across two package managers,
and both sides pull GTMSessionFetcher:

| Plugin | Resolves via | Brings in |
|---|---|---|
| `firebase_auth` 5.5.3 — ships `ios/firebase_auth/Package.swift` | SPM | firebase-ios-sdk → FirebaseAuth → `GTMSessionFetcher` (module `GTMSessionFetcherCore`) |
| `google_sign_in_ios` 5.7.5 — no `Package.swift` at all | CocoaPods | `GoogleSignIn 7.0.0` → `GTMAppAuth 2.0.0` → `GTMSessionFetcher/Core 3.5.0` |

Both are static libraries linked into `Runner`, so every symbol lands twice. Flutter's own
hint at the bottom of the log names the module exactly.

The CocoaPods side isn't movable onto SPM as things stand. `google_sign_in_ios` is pinned
to 5.7.5 by 4570415df7, which withdrew the google_sign_in 7.x bump after sign-in failed on
a physical iPhone with "Authentication failed" and the cause was never established. That
same commit pinned `flutter_foreground_task` to 10.0.0 precisely because 11.0.1 "requires
project-wide Swift Package Manager adoption, which this project does not use" — and the
build logs 15 plugins with no SPM support at all (`flutter_sound`, `nordic_dfu`,
`mcumgr_flutter`, `whisper_flutter_new`, `permission_handler_apple`,
`flutter_secure_storage`, `opus_flutter_ios`, `flutter_background_service_ios`,
`awesome_notifications`, `flutter_tts`, `map_launcher`, `flutter_archive`, ...).

**On the verification.** `app/test/unit/flutterfire_spm_contract_test.dart` asserts the
string `enable-swift-package-manager` is absent from pubspec.yaml. That's a static
tripwire on the source text — it can't observe plugin resolution or linking, so it passed
while the archive it stood in for was broken. There's no iOS job in CI, so nothing else
would catch it either; anything that changes iOS plugin resolution needs a real archive on
macOS as its acceptance test.

Reverting.

PR #12305, opened 2026-08-27
feat(app): add device button action switch

### LINKED ISSUE #4458: Allow button to be completely disabled because kids exist...
Button functionality was improved in #3058 and #3812 and #2825 but I do not believe there is a current feature request open to have an option to disable the button completely.

This would be very useful for anyone who does physical work or has or works with small children, as currently the button getting continually pressed by accident (and even long presses would happen by accident) makes the Omi Devkit 2 device not really usable (unless you get the watch holder 3D print working, which I am trying to do, first print the size was way off for some reason.)

The problem happens both when you pick kids up / hold them, and when kids discover there is a button they can press.

arhxam:
## What changed and why

Add a default-on, persisted master switch for Omi device button actions. Interactive onboarding still receives the button event first, while normal single-, double-, and long-press actions return before side effects when the user disables the switch.

Closes #4458

none

## How it was verified

- `flutter test test/unit/device_button_policy_test.dart test/widgets/device_settings_double_tap_test.dart` — all 9 preference, policy, persistence, and settings-widget tests passed.
- `flutter gen-l10n` — generated all 48 supported locale implementations with zero untranslated messages.
- The physical button path was not exercised because no paired Omi device is available locally.

## Tests

- [x] Added default-on, disabled, and persistence policy coverage.
- [x] Added widget coverage for toggling and reloading the setting.
- [x] Regenerated and validated every supported locale.
- [ ] Physical-device button presses (hardware unavailable).

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12305?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/backend/preferences.dart
+++ b/app/lib/backend/preferences.dart
@@ -271,6 +271,12 @@ class SharedPreferencesUtil {
 
   set batchModeSuspendedForOnboarding(bool value) => saveBool('batchModeSuspendedForOnboarding', value);
 
+  // Master switch for user-configured button actions. Interactive onboarding
+  // still receives button events before this preference is applied.
+  bool get deviceButtonEnabled => getBool('deviceButtonEnabled', defaultValue: true);
+
+  set deviceButtonEnabled(bool value) => saveBool('deviceButtonEnabled', value);
+
   // Double tap behavior: 0 = end conversation (default), 1 = pause/mute, 2 = star ongoing conversation
   int get doubleTapAction => getInt('doubleTapAction');
 
--- a/app/lib/l10n/app_ar.arb
+++ b/app/lib/l10n/app_ar.arb
@@ -164,6 +164,8 @@
     "modelNumber": "رقم الطراز",
     "manufacturer": "الشركة المصنعة",
     "doubleTap": "نقرة مزدوجة",
+    "deviceButtonActions": "إجراءات زر الجهاز",
+    "deviceButtonActionsDescription": "السماح لضغطات الزر بالتحكم في التسجيل والأوامر الصوتية",
     "ledBrightness": "سطوع LED",
     "micGain": "تضخيم الميكروفون",
     "disconnect": "قطع الاتصال",
--- a/app/lib/l10n/app_be.arb
+++ b/app/lib/l10n/app_be.arb
@@ -661,6 +661,8 @@
         "description": "Label for manufacturer field"
     },
     "doubleTap": "Двайны дотык",
+    "deviceButtonActions": "Дзеянні кнопкі прылады",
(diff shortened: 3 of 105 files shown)


Git-on-my-level:
Thanks @arhxam - this is a clean, well-scoped implementation of #4458, and the default-on choice preserves today's behavior for existing users.

What I verified:

- `app/lib/services/capture/capture_controller.dart` - the gate sits inside the single `onButtonReceived` handler (the only BLE button listener registration; every device connector funnels through it), after the interactive-onboarding interception and before the double-tap, single-tap, and legacy long-press branches (button states 1/2/3/5), so one check covers every user-configured action with no bypass path.
- `app/lib/backend/preferences.dart` - `deviceButtonEnabled` defaults to `true`, so existing users are unaffected until they opt out.
- `app/lib/services/capture/device_button_policy.dart` - small policy seam that keeps the gate unit-testable; `app/test/unit/device_button_policy_test.dart` covers default-on, disabled, and reload persistence, and `app/test/widgets/device_settings_double_tap_test.dart` exercises the actual switch widget plus persistence across a reload.
- l10n - the two new strings carry `@` descriptions in `app_en.arb`, all supported locales are populated, and the regenerated `app_localizations*.dart` files match (Generated Files check is green).
- CI: Dart Analyze & Tests, Formatting, Android Compile Smoke, and PR Metadata Preflight all pass.

Two non-blocking notes for the maintainer:

1. The switch is a single global preference but lives on a per-device settings page - worth deciding whether v1 should stay global (matches the issue's ask) or become per-device.
2. During interactive onboarding step 1, the single-tap fall-through reaches the new gate, so a user who has disabled button actions will not get the voice-command demo tap during re-onboarding (the onboarding provider still receives the event first). Probably the right trade-off, just flagging the interaction.

Nice work shipping tests alongside the feature. Product direction decision on the switch scope (global vs per-device) needs maintainer sign-off before merge.

---

kodjima33 (approved):
Feature (not bug fix) in app/mobile area - approve only per policy, do not merge.

kodjima33 (approved):
Feature (device button action switch), not a bug fix — mobile/app area default is approve-only per confidence gate. Diff mostly auto-generated l10n; core change is small and tested. No abuse signals.

kodjima33 (approved):
Clean, minimal feature (default-on button-action master switch, single gate point in capture_controller.dart, well tested). It's a feat, not a bug fix, so approve-only per policy; also the cleaner of two competing implementations of this same feature (see #11193).

PR #12306, opened 2026-08-27
fix(app): recover live-listen expired tokens

### LINKED ISSUE #11694: Live-listen auth: 15 clients stuck in a permanent expired-token reconnect loop — 372 rejected upgrades vs 333 accepted in 10 min
Prod `backend-listen` is rejecting more live-transcription websocket upgrades than it accepts, and nearly all of the rejections come from a handful of clients stuck in a permanent expired-token reconnect loop.

Measured window 2026-08-16 16:50–17:00Z (prod, `resource.labels.container_name="backend-listen"`):

```
372  "WebSocket /v4/listen?... 403"      <- rejected upgrades
333  "WebSocket /v4/listen?... [accepted]"
 15  distinct uids across all 372 rejections   (all source=omi)
```

So ~53% of live-listen upgrade attempts in that window are rejections, and they are produced by 15 users — roughly **one reconnect every 24s per user, continuously**. Over 12h the paired auth log is 580+ lines of:

```
ERROR:utils.other.endpoints:WebSocket auth failed: code=4001 error=Token expired, 1786634034 < 1786900104
```

`1786900104` is the check time; `1786634034` is the token's `exp` — **3.07 days earlier**. Firebase ID tokens live one hour, so these clients are re-presenting a token that has been dead for days and never refreshing it. Sample uids repeating in the same minute: `LSZfV3DmdLTODnOlcEVdeO3vWhs2`, `fcGkZjrlDfUQJoTRMRuOBHW034I3`, `t4bb2MkejuY9344sH4TacdtHaLD3`, `4GYyK2QIvxNHuGv6QMuXXwqDCsC3`.

Two separate problems:

1. **Those users have no live transcription at all, indefinitely.** The socket never opens, so nothing is captured. Nothing in the loop escalates — the client just keeps retrying the same dead credential.
2. **A permanent auth failure is being retried like a transient one.** A 403 on the upgrade is not retryable without a new token; retrying it every ~24s forever is pure load (and it currently outnumbers real traffic).

Where to look on the client: `app/lib/services/sockets/pure_socket.dart` builds its headers per connect via `buildHeaders(requireAuthCheck: true, ..., forWebSocket: true)` → `getAuthHeader()` in `app/lib/backend/http/shared.dart:60`. That function decides whether to refresh from the **cached** `SharedPreferencesUtil().tokenExpirationTime` pref, not from the token itself; if that pref is stale-but-in-the-future the refresh is skipped and the dead token is sent. The `AuthTokenTransientFailure` branch also deliberately preserves the existing token when the cached expiry still looks valid. Either path reproduces exactly this symptom.

Not verified: which app version(s) the 15 uids are on — worth confirming before choosing between "fix the refresh gate" and "expire the session after N consecutive 4001s".

Found by the hourly desktop bug watchdog while sweeping prod logs. Reporting only — the fix is in the client auth/token path, which the watchdog does not touch.

arhxam:
## What changed and why

Fix both sides of the live-listen expired-token loop:

- `getAuthHeader()` now derives expiry from the cached Firebase JWT it is about to send. The token's `exp` claim wins over the separately persisted timestamp, so a dead token paired with stale future metadata is refreshed before the WebSocket upgrade instead of being rejected as HTTP 403 forever.
- If a server transport does deliver auth close codes, recovery remains bounded and single-flight: refresh and reconnect on 4001 success, retain the existing keepalive retry for transient failures, and expire the session for terminal/missing credentials or 4004 rejection.

Malformed or legacy non-JWT cached values retain the existing persisted-expiry fallback.

Closes #11694

none

## How it was verified

- `flutter test test/unit/auth_token_expiration_test.dart test/unit/live_listen_auth_recovery_test.dart test/unit/backend/http/shared_test.dart test/unit/auth_service_token_result_test.dart test/unit/token_refresh_loop_test.dart` — all 58 auth/header/recovery tests passed.
- Targeted `flutter analyze` over the changed auth, capture, and test files — no issues.
- The original capture-controller suite verification remains 89/89 passing.
- A production token-expiry session was not forced against the live service; the stale-metadata failure and close-code/result matrix are covered hermetically.

## Tests

- [x] Added regression coverage proving an expired JWT overrides stale future expiry metadata.
- [x] Added coverage for refresh success, transient failure, terminal failure, missing user, 4004 rejection, and ordinary transport closes.
- [x] Ran the existing capture-controller provider suite.
- [ ] Live backend token-expiry exercise (would require disrupting a real account session).

Failure-Class: none

### DIFF
--- a/app/lib/backend/http/shared.dart
+++ b/app/lib/backend/http/shared.dart
@@ -33,6 +33,30 @@ class AuthTokenUnavailableException implements Exception {
   String toString() => 'AuthTokenUnavailableException(${result.runtimeType})';
 }
 
+/// Returns the expiry carried by the JWT itself, falling back to the persisted
+/// Firebase metadata for legacy or malformed cached values.
+///
+/// The token is the credential sent on the wire, so its `exp` claim must win if
+/// it disagrees with the separately persisted expiration timestamp.
+@visibleForTesting
+DateTime resolveAuthTokenExpiration({required String token, required int cachedExpirationTime}) {
+  final segments = token.split('.');
+  if (segments.length == 3) {
+    try {
+      final payload = jsonDecode(utf8.decode(base64Url.decode(base64Url.normalize(segments[1]))));
+      final expirationSeconds = payload is Map<String, dynamic> ? payload['exp'] : null;
+      if (expirationSeconds is num && expirationSeconds.isFinite) {
+        return DateTime.fromMillisecondsSinceEpoch(expirationSeconds.toInt() * 1000, isUtc: true);
+      }
+    } on FormatException {
+      // Fall back to Firebase's persisted metadata below.
+    } on RangeError {
+      // Fall back to Firebase's persisted metadata below.
+    }
+  }
+  return DateTime.fromMillisecondsSinceEpoch(cachedExpirationTime);
+}
+
 // Normal-mode connectivity failures on mobile (no network, DNS failure,
 // connection reset, TLS handshake during reconnect, request timeout, OS abort
 // when the app backgrounds mid-upload). Reporting these to Crashlytics drowns
@@ -62,8 +86,12 @@ Future<String> getAuthHeader({bool expireTerminalSession = true}) async {
     throw AuthTokenUnavailableException(const AuthTokenMissingUser());
   }
 
-  final expiry = DateTime.fromMillisecondsSinceEpoch(SharedPreferencesUtil().tokenExpirationTime);
-  bool hasAuthToken = SharedPreferencesUtil().authToken.isNotEmpty;
+  final cachedToken = SharedPreferencesUtil().authToken;
+  final expiry = resolveAuthTokenExpiration(
+    token: cachedToken,
+    cachedExpirationTime: SharedPreferencesUtil().tokenExpirationTime,
+  );
+  bool hasAuthToken = cachedToken.isNotEmpty;
 
   bool isExpirationDateValid = !(expiry.isBefore(DateTime.now()) ||
       expiry.isAtSameMomentAs(DateTime.fromMillisecondsSinceEpoch(0)) ||
--- a/app/lib/services/capture/capture_controller.dart
+++ b/app/lib/services/capture/capture_controller.dart
@@ -34,6 +34,7 @@ import 'package:omi/services/connectivity_service.dart';
 import 'package:omi/services/services.dart';
 import 'package:omi/services/voice_playback/omi_voice_playback_service.dart';
 import 'package:omi/services/sockets/transcription_service.dart';
+import 'package:omi/services/sockets/live_listen_auth_recovery.dart';
 import 'package:omi/services/audio_sources/audio_source.dart';
 import 'package:omi/services/audio_sources/ble_device_source.dart';
 import 'package:omi/services/devices/connectors/limitless_connection.dart';
@@ -82,6 +83,7 @@ class CaptureController extends ChangeNotifier
 
   TranscriptSegmentSocketService? _socket;
   Timer? _keepAliveTimer;
+  Future<void>? _liveListenAuthRecoveryInFlight;
   DateTime? _keepAliveLastExecutedAt;
   Timer? _inProgressConversationRefreshTimer;
   int _inProgressConversationRefreshAttempts = 0;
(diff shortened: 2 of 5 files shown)


Git-on-my-level (dismissed):
Thanks @arhxam — the recovery module itself is cleanly built, but I found a wire-level gap that I believe prevents this from actually fixing #11694, so I'm requesting changes for now.

**Core problem: the 4001/4004 close codes never reach `onClosed` in the failure mode this PR targets.**

- `backend/utils/other/endpoints.py` raises `WebSocketException(code=4001/4004)` inside `_verify_ws_auth`, which runs as a FastAPI dependency of `get_current_user_uid_ws_listen` — i.e. **before** `websocket.accept()`.
- With the pinned server stack (`uvicorn==0.30.5`, `websockets==12.0` in `backend/requirements.txt`), a close sent before the handshake completes is answered as **HTTP 403 with the close code dropped** (uvicorn's `websockets_impl.py` maps a pre-accept `websocket.close` to `HTTPStatus.FORBIDDEN`). That matches the prod logs quoted in #11694 exactly: `"WebSocket /v4/listen?... 403"`.
- On the client, a rejected upgrade never produces a close frame: `channel.ready` fails with a `WebSocketChannelException`, `PureSocket.connect()` (`app/lib/services/sockets/pure_socket.dart`) catches it and returns `false` — `onClosed` is never called. So `_scheduleLiveListenAuthRecovery` in `capture_controller.dart` never fires for handshake rejections, and the retry storm in #11694 continues unchanged.
- I also searched the backend for any **post-accept** producer of 4001/4004 and found none: `WS_AUTH_CODE_TOKEN_REFRESH`/`WS_AUTH_CODE_RELOGIN_REQUIRED` are only used on the pre-accept path; accepted-socket closes use 1008/1003/1011 (listen runtime) or 4005/4006 (deletion/cutover fence). The backend unit tests (`backend/tests/unit/test_ws_auth_handshake.py`) pass because Starlette's TestClient preserves close-before-accept codes, while a real uvicorn websockets-protocol server does not.

Net: `live_listen_auth_recovery.dart` handles codes the current client+server pair cannot deliver for this bug.

**Per file:**

- `app/lib/services/capture/capture_controller.dart` — the single-flight guard (`_liveListenAuthRecoveryInFlight` + `identical` check) and the disposition switch are well done, and re-arming `_startKeepAliveServices()` on `retryLater`/`notAuthRelated` preserves the existing backoff; `closeCode!` is safe after the equality checks. The gap is only trigger reachability. One behavior worth a maintainer eye: 4004 now signs the user out immediately (`expireSession`) — reasonable for a revoked token, but it's a product behavior change on an auth surface.
- `app/lib/services/sockets/live_listen_auth_recovery.dart` — good pure design (DI of `refreshToken`/`expireSession`, exhaustive sealed-class switch over `AuthTokenResult`). If/when the codes actually arrive, this is the right shape — it's just unreachable today for the handshake-rejection scenario.
- `app/test/unit/live_listen_auth_recovery_test.dart` — solid matrix (success/transient/terminal/missing-user/4004/ordinary close, including asserting 4004 skips refresh). Missing the case that matters for #11694: an upgrade rejected with HTTP 403 delivers no close code at all and bypasses this module entirely.

**Directions that would actually close #11694** (either, or coordinated):

1. Client: handle the real signal — force `AuthService.instance.refreshIdToken()` after a bounded number of consecutive `PureSocket.connect()` failures on an authed socket, or fix the stale-expiry gate the issue names as root cause: `getAuthHeader()` in `app/lib/backend/http/shared.dart` trusts the cached `tokenExpirationTime` pref and skips refresh when it is stale-but-in-the-future.
2. Backend: make `/v4/listen` deliver the codes — accept first, then close with 4001/4004 (or use the websocket denial-response extension). That's a coordinated protocol change, and `test_ws_auth_handshake.py` should keep modeling the real wire behavior.

Your recovery module slots straight into option 2 (and partly option 1), so this work isn't wasted — it's just ahead of the transport.

Maintainer note: beyond the contributor fix, this needs a maintainer decision on fix layering for #11694 (client refresh gate vs backend accept-then-close), and a quick sign-off on the new sign-out-on-4004 behavior since it changes an auth/session-expiry flow.

---

Git-on-my-level:
Follow-up review on 94947bf — the wire-level gap from my earlier review is addressed at the root cause. Nice work.

**What changed:** `getAuthHeader()` in `app/lib/backend/http/shared.dart` no longer trusts the persisted `tokenExpirationTime` alone. The new `resolveAuthTokenExpiration()` decodes the `exp` claim from the cached JWT itself and lets it win over stale-but-future metadata, falling back to the persisted timestamp for legacy or malformed values. I traced the wire path to confirm this fixes #11694 where it starts: every `/v4/listen` connect builds its headers through `PureSocket.connect()` → `buildHeaders(requireAuthCheck: true)` → `getAuthHeader()`, so a dead token now triggers `refreshIdToken()` **before** the upgrade instead of producing a 403-and-retry-every-24s loop. The transient-failure branch also correctly refuses to preserve a token whose (now JWT-derived) exp has already passed.

Per file:

- `app/lib/backend/http/shared.dart` — `resolveAuthTokenExpiration()` is defensively written: `base64Url.normalize` for unpadded input, `FormatException`/`RangeError` fall back to the persisted metadata, non-map payloads and non-numeric/non-finite `exp` fall through, and `exp == 0` lands on the existing epoch-zero invalidity check so it forces a refresh. Parsing without signature verification is fine here — it only gates the refresh decision; the server still validates the credential.
- `app/lib/services/capture/capture_controller.dart` — the 4001/4004 early return ahead of `_startKeepAliveServices()`, the single-flight `_liveListenAuthRecoveryInFlight` with the `identical()` release check, and the disposition switch (`reconnectNow` → `_reconnectActiveCapture()`, `sessionExpired` cancels the keepalive timer, `retryLater`/`notAuthRelated` re-arm the existing backoff) all check out; `closeCode!` is safe after the equality guards.
- `app/lib/services/sockets/live_listen_auth_recovery.dart` — clean pure module: DI of `refreshToken`/`expireSession`, exhaustive switch over the sealed `AuthTokenResult` variants, and 4004 expires the session without attempting a refresh (correct for a revoked token).
- `app/test/unit/auth_token_expiration_test.dart` — covers the key regression: an expired JWT overrides a stale future cached expiry.
- `app/test/unit/live_listen_auth_recovery_test.dart` — good matrix including 4004-skips-refresh and ordinary-close passthrough; Dart Analyze & Tests is green on this head.

**Remaining context, not a blocker:** the 4001/4004 handling in `capture_controller.dart` stays dormant for this specific bug — `backend/utils/other/endpoints.py` still raises those codes pre-accept, where the pinned server stack answers with HTTP 403 and no close frame ever reaches `onClosed`. With the pre-upgrade refresh in place that path is unreachable for the dead-token case, so the recovery module now acts as a safety net for any future accept-then-close protocol change. Worth keeping as-is.

Still needs maintainer sign-off before merge: this lands on the auth/session surface — the pre-upgrade forced refresh and the sign-out-on-4004 behavior change session-expiry semantics, so `security-review` stays on. No product-direction concerns; this is squarely the fix #11694 describes.

---

kodjima33 (approved):
Bug fix (Closes #11694) touches auth/token code but appears to be a legitimate, well-tested fix. PR is DIRTY - leaving at approve, needs author rebase before merge.

Issue #6603, opened 2026-04-13
Self-hosted backend URL override


Rahulsharma0810:
**Is your feature request related to a problem?**

Yes. The Omi app is open-source, the backend is open-source, and the self-hosting documentation exists — but in practice, self-hosting is broken for anyone using the official iOS app. The app has the backend URL hardcoded to `api.omi.me`, making it impossible to point it at a self-hosted instance without maintaining a full fork of the iOS app.

This was raised in #842 and closed as "not planned" citing "e2e authentication" as the blocker. That reasoning is incomplete — see below.

---

**The real blockers (and why they're already solved):**

**1. Firebase auth mismatch**
The app generates tokens signed for Omi's Firebase project. A self-hosted backend on a different Firebase project rejects them.

Already solved: the backend has `LOCAL_DEVELOPMENT=true` which bypasses Firebase verification entirely (`utils/other/endpoints.py`). Self-hosters can enable this flag. Auth is not a blocker.

**2. Flutter TLS (the actual blocker)**
Flutter's `dart:io` uses BoringSSL — its own TLS stack that ignores the iOS system certificate store. This means even if a user installs a custom CA cert and trusts it in iOS Settings (which makes Safari work), the Omi app still rejects it with:
```
CERTIFICATE_VERIFY_FAILED: application verification failure
```
The only way around this without rebuilding the app is to use a domain with a publicly trusted certificate. Which requires controlling the domain. Which requires changing the hardcoded URL in the app. It's circular.

---

**The fix is already 90% implemented:**

`lib/env/env.dart` already has:
```dart
static String? _apiBaseUrlOverride;

static void overrideApiBaseUrl(String url) {
  _apiBaseUrlOverride = url;
}

static String? get apiBaseUrl => _apiBaseUrlOverride ?? _instance.apiBaseUrl;
```

`Env.overrideApiBaseUrl()` exists but is never exposed in the UI. The entire feature request is a text field in Developer Settings that calls this method and persists the value to SharedPreferences.

Estimated change: ~50 lines of Dart. The infrastructure is already there.

---

**What self-hosting currently requires (without this feature):**

1. Fork the entire iOS app repository
2. Modify `app/.env` with a custom API URL
3. Run `dart run build_runner build` to regenerate env files
4. Set up Apple Developer account ($99/yr) for distribution
5. Set up GitHub Actions CI pipeline for builds
6. Manually merge upstream changes on every Omi release to stay updated
7. Re-distribute the app to devices on every update

[...]

Issue #842, opened 2024-09-15
An ability to change link and add a self-hosted backend inside of an app


kodjima33:
**Is your feature request related to a problem? Please describe.**
Some enterprise users (Salman's friends in Salesforce) might want to use the app but only with a self-hosted backend. Current app allows to self-host backenbd but will also require self-hosted mobile app which people dont want 

**Describe the solution you'd like**
An ability to change link to a self-hosted backend inside of an app in settings. 

**Describe alternatives you've considered**
Generate private key

asadbekdev:
Login-based Configuration:

Instead of directly inputting a backend URL, users log in with their enterprise credentials.
The app fetches the correct backend configuration based on the user's organization.
Pro: Centralized management of backend configurations.
Con: Requires more complex backend infrastructure to manage configurations.

Deep Linking: 

Create a custom URL scheme for your app (e.g., myapp://configure?backend=https://example.com).
Enterprises can send these links to their users via email or secure channels.
Pro: Easy to distribute and can be updated remotely.
Con: Requires handling deep links in the app and potential security considerations.

josancamon19:
You can't have a self hosted backend without a self hosted app, as the authentication is e2e, they would have to run their own app. After discussions with salman, this doesn't seem like the reason for them not to use it.

Would require many changes for this to work without any proof of need, so closing, this is not planned.

Rahulsharma0810:
Please Consider — @mdmohsin7  @josancamon19 

The "e2e auth" blocker cited in the closing comment is already solved in the codebase itself.

The actual blockers for self-hosted backend without a self-hosted app:

1. Firebase project mismatch — the app's tokens are signed for Omi's Firebase project, so a different backend rejects them. But the backend already has LOCAL_DEVELOPMENT=true which bypasses this entirely (endpoints.py). Any self-hoster can enable this.

 2. Flutter TLS — dart:io uses BoringSSL which ignores the iOS system certificate store, so self-signed certs for api.omi.me don't work even if trusted at the OS level. This is the real blocker — and it only exists because the URL is hardcoded.

The fix is already 90% done in the codebase:

```
  env.dart already has:
  static String? _apiBaseUrlOverride;

  static void overrideApiBaseUrl(String url) {
    _apiBaseUrlOverride = url;
  }


  static String? get apiBaseUrl => _apiBaseUrlOverride ?? _instance.apiBaseUrl;
```

[...]

PR #12308, opened 2026-08-27
feat(app): add safe backend URL override

arhxam:
## What changed and why

Expose the existing API-base override in Developer Settings for local/community development, persist and restore it before services initialize, and validate schemes/hosts before activation. Custom authorities never receive Omi credentials, public endpoints require HTTPS, invalid persisted values fail closed, and release builds remain locked to their flavor-controlled data plane.

Refs #6603

- INV-DATA-1 — preserves the existing production-family authority: runtime overrides are unavailable in release builds and malformed persisted state fails closed to the flavor backend.

## How it was verified

- `flutter test test/unit/backend_url_override_test.dart test/unit/env_test.dart` — all 32 validation, persistence, credential-boundary, and production-routing tests passed.
- No requests were sent to a real self-hosted server; parsing, startup restoration, release gating, and credential attachment are covered hermetically.

## Tests

- [x] Added URL normalization and private-network/HTTPS validation coverage.
- [x] Added startup restore, clear, invalid-state, and release fail-closed coverage.
- [x] Added exact-host credential-boundary coverage, including a deceptive subdomain.
- [x] Ran the locked production data-plane routing suite.
- [ ] Real self-hosted backend connection (no deployment configured locally).

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12308?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/backend/http/shared.dart
+++ b/app/lib/backend/http/shared.dart
@@ -118,6 +118,11 @@ Future<Map<String, String>> buildHeaders({
   String? method,
   bool forWebSocket = false,
 }) async {
+  final effectiveAuthCheck = shouldHonorRequestedOmiAuth(
+    requested: requireAuthCheck,
+    customBackendActive: Env.hasApiBaseUrlOverride,
+    url: url,
+  );
   final headers = <String, String>{
     'X-Request-Start-Time': (DateTime.now().millisecondsSinceEpoch / 1000).toString(),
     'X-App-Platform': PlatformManager.instance.platform,
@@ -130,7 +135,7 @@ Future<Map<String, String>> buildHeaders({
   if (shouldAttachAccountGenerationHeader(
     url: url,
     method: method,
-    requireAuthCheck: requireAuthCheck,
+    requireAuthCheck: effectiveAuthCheck,
     forWebSocket: forWebSocket,
   )) {
     final accountGeneration = AccountCutoverRuntime.instance.control.accountGeneration;
@@ -141,7 +146,7 @@ Future<Map<String, String>> buildHeaders({
     }
   }
 
-  if (requireAuthCheck) {
+  if (effectiveAuthCheck) {
     // Authenticated requests must never degrade into anonymous traffic. A
     // typed exception stops the request before it reaches the network.
     headers['Authorization'] = await getAuthHeader(expireTerminalSession: expireTerminalSession);
@@ -162,8 +167,11 @@ String normalizeOmiApiUrlForHostMatch(String url) {
 }
 
 bool _isRequiredAuthCheck(String url) {
-  // Agent VM endpoints always hit prod even when app uses dev
-  if (url.contains('api.omi.me')) return true;
+  if (shouldAttachOmiCredentials(url)) return true;
+  // A runtime override is a separate trust boundary. Never send the user's
+  // Omi credential to it, even when it happens to share a path with the
+  // configured API base URL.
+  if (Env.hasApiBaseUrlOverride) return false;
   final base = Env.apiBaseUrl;
   if (base != null && base.isNotEmpty) {
     final normalizedUrl = normalizeOmiApiUrlForHostMatch(url);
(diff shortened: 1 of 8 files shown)


Git-on-my-level:
Thanks @arhxam - this is a careful implementation of #6603 and the credential-isolation work is genuinely well done. Reviewed all 8 changed files at cfd41bc3.

**What I verified**

- `app/lib/env/backend_url_override.dart` - parse() enforces http/https only, rejects embedded credentials, query strings, and fragments, requires HTTPS for anything outside loopback/host.docker.internal/::1/RFC1918/CGNAT, and normalizes the trailing slash. restore() fails closed on invalid persisted state and on release builds (runtimeAllowed defaults to !kReleaseMode).
- `app/lib/backend/http/shared.dart` - the old substring rule `url.contains('api.omi.me')` is replaced by exact-host matching in shouldAttachOmiCredentials over {api.omi.me, api.omiapi.com} (https/wss only). That is stricter and also fixes a latent weakness: a deceptive host like `api.omi.me.attacker.example` no longer matches. Under an active override, _isRequiredAuthCheck returns early, shouldHonorRequestedOmiAuth de-authenticates legacy unconditional-auth callers, and X-Account-Generation is suppressed too, so neither the bearer token nor account metadata crosses to a custom authority. The 401-replay path in makeRawApiCall recomputes the effective auth check inside buildHeaders, so a custom backend returning 401 cannot trigger an Omi token refresh/replay against itself.
- `app/lib/env/env.dart` - hasApiBaseUrlOverride plus the production clear() alias; override takes precedence over the dart-define and flavor values in apiBaseUrl, which is the intended precedence for a runtime override.
- `app/lib/main.dart` - SharedPreferencesUtil.init() moved earlier so BackendUrlOverride.restore() lands after startup routing validation and before service init; a persisted override therefore cannot trip validateStartupRouting on relaunch, and services see the final base URL.
- `app/lib/backend/preferences.dart` - plain get/set passthrough for customBackendUrl, consistent with neighboring keys.
- `app/lib/pages/settings/developer.dart` - the section is behind `!kReleaseMode`, so release builds never render it.
- `app/lib/providers/developer_mode_provider.dart` - save path is also release-gated; parse errors surface via snackbar and abort the save without persisting a bad value; an empty field clears the override.
- `app/test/unit/backend_url_override_test.dart` - good hermetic coverage: normalization, cleartext scope, rejection cases (public cleartext, user:pass@host, fragments, ftp://), startup restore/clear, invalid fail-closed, release fail-closed, and credential-boundary cases including the deceptive subdomain.

Checks are green (Dart Analyze & Tests, Android Compile Smoke, Hygiene, Formatting).

**Non-blocking suggestions**

1. The credential boundary is unit-tested at the helper level; one integration-style test that calls buildHeaders end-to-end with an override active and asserts no Authorization/X-Account-Generation headers would lock the boundary in at the seam where regressions would actually happen.
2. Worth being aware (not a defect): with a localDev debug build, pointing the override at a public HTTPS host does not re-run startup routing validation - that is precisely the self-hosting use case, just noting the invariant surface.

**Human review needed for**

- A product-direction decision: #842 asked for the same underlying capability and was closed as not planned citing e2e auth. The debug-only scoping and credential isolation here address the safety side of that objection, but whether the official app should carry a self-hosting backend override is a maintainer call.
- Sign-off on the modified credential-attachment logic in shared.dart (exact-host allowlist) since it changes when Authorization is attached for all API traffic, not just the override path.

---

Git-on-my-level:
Following up on the earlier review of this head — the technical assessment stands, and one additional consequence is worth having on the record before the product call.

**Identity-plane note:** `Env.authApiBaseUrl` resolves through `authApiBaseUrlForProfile(profile, servingApiBaseUrl: apiBaseUrl)` (app/lib/env/env.dart), so for every non-mobileBeta profile it follows the runtime override. With the override active in a debug build, the OAuth endpoints in `app/lib/services/auth_service.dart` (`v1/auth/authorize` and `v1/auth/token`) also target the custom backend — the override moves the identity plane, not just the serving/data plane. This isn't a regression: `OMI_API_BASE_URL` via dart-define already had the same reach, everything here is debug-gated and user-initiated, and a usable self-hosted stack arguably needs auth to follow the backend. But since the PR description frames the invariant as data-plane (INV-DATA-1), maintainers ruling on #6603 may want to decide explicitly whether runtime overrides should keep OAuth pinned to the production identity plane (the way mobileBeta does) or follow the backend. One sentence in the PR description settling that intent would make the invariant complete.

Everything else from the earlier verification holds: credential isolation in `app/lib/backend/http/shared.dart` (exact-host matching over `api.omi.me`/`api.omiapi.com`, https/wss only) is strictly stricter than the old substring rule; under an active override neither the bearer token nor `X-Account-Generation` crosses to a custom authority, and the 401-replay path can't leak either. Validation in `BackendUrlOverride.parse` (scheme, credentials, query/fragment, HTTPS-off-private-networks, trailing-slash normalization) and the fail-closed restore in release builds are all covered by the new unit tests.

Still leaving the merge decision to a human maintainer: the product-direction call on a runtime self-hosting override (#6603, with #842 previously closed as not planned) — now including the identity-plane question above — and sign-off on the changed credential-attachment logic in `shared.dart`, which affects Authorization attachment for all API traffic.

---

kodjima33 (approved):
Feature (not bug fix) in app/mobile area - approve only per policy, do not merge. Well-tested credential-boundary safety design; needs-maintainer-review + security-review labels already flag it for human sign-off given it touches credential-attachment logic.

PR #12326, opened 2026-08-27
SCA-364: render Desktop Chat SQL timestamps in the user timezone

### LINKED ISSUE #12321: Desktop chat shows UTC screenshot timestamps as local time
## Symptom (2026-08-27, macOS Desktop Chat)

Asking “when was the last time I opened Claude?” returned **August 27, 2026 at 7:59:51 PM** with **no timezone**. Wall clock was **5:19 PM Eastern**, so the answer looked like the future.

The seconds (`:51`) match a raw SQLite `YYYY-MM-DD HH:MM:SS` cell, not `DateFormatter.timeStyle = .short`.

Offset matches **UTC wall-clock presented as local**: `19:59:51` UTC displayed as `7:59:51 PM` instead of `3:59:51 PM EDT`.

## Root cause (verified on `origin/main`)

This is **not** backend chat-tool #4643. Those tools already call `resolve_display_tz` / `get_user_time_zone` (merged #8483 #8495 #8538 #8588 #8575 #9719 #10012). Desktop **local SQL** never got that treatment.

1. `screenshots.timestamp` (and sibling DATETIME columns) are UTC-naive SQLite DATETIME via GRDB `Date`. No zone in the cell (`RewindDatabase.swift`).
2. `ChatToolExecutor.executeSelectQuery` dumps cells as raw strings/numbers. No conversion, no zone label.
3. `ChatPrompts.desktopChat` tells the model to convert UTC itself, then teaches contradictory SQL:
   - Examples `SELECT timestamp` with no `datetime(..., 'localtime')`.
   - Filters compare UTC columns to `datetime('now', 'localtime')`.
   - Prose: “SQLite handles [localtime] automatically.”
4. `ChatPromptBuilder.currentDatetimeString` is unlabeled `yyyy-MM-dd HH:mm:ss` (zone only in surrounding prose).
5. The model copies the UTC cell as a friendly local time.

Open **PR #12236 / SCA-358** is a **different** class (Insight/Focus “2026 is wrong” + cache-safe *now* injector). Do **not** stack on that PR.

## Related, do not re-implement

- #4643 remains OPEN/stale (`FIXED_OPEN` comment 2026-08-06). Residual here is desktop SQL/LLM rendering.
- Do not reopen backend `search_screen_activity_tool` unless tests on main still fail.

## Required fix (one PR)

Mechanical conversion — **do not rely on the LLM**.

1. **`execute_sql` result formatter** (`ChatToolExecutor.executeSelectQuery`): DATETIME/`timestamp`/`*At` cells render in `TimeZone.current` with an explicit IANA id or offset (e.g. `2026-08-27 15:59:51 EDT (America/New_York)`). Pin a clock + timezone in tests.
2. **Prompt SQL examples** in `ChatPrompts.swift` (and Windows parity if the same examples exist): compare UTC columns to UTC bounds **or** convert both sides; never mix UTC cells with `datetime('now', 'localtime')`. Drop “the model will convert.”
3. **Other LLM-facing timestamp paths on the same surfaces**: `semantic_search` / screen-history formatters, daily recap, task due dates shown to chat, Windows `desktopChatPrompt` if unlabeled UTC remains. User-visible times always carry zone.
4. **`currentDatetimeString` / current-time injection**: include IANA zone in the substituted value; keep live clocks out of cached system-prompt prefixes (same cache contract as #12236).
5. **Backend chat tools**: add tests only if a remaining unlabeled-UTC path exists on main; do not churn already-covered `resolve_display_tz` code.

Out of scope: settings timezone picker, FCM `time_zone` sync, Insight year-example (#12236), merging/deploying.

## Acceptance

- Fixture: UTC instant `2026-08-27T19:59:51Z`, `America/New_York`. Chat/SQL/search text for that row contains **3:59:51 PM** (or `15:59:51`) **and** a timezone token. Must **not** present `7:59:51 PM` unlabeled.
- Filter examples/tests: “today” / “yesterday” bounds do not shift by UTC offset.
- Component tests + `make preflight`. Record verification. Failure-Class via `scripts/pr-preflight --suggest` (likely `FC-machine-timezone-timestamp-in-prompt` or `new`).
- Provenance: `git -C ~/workspace/omi fetch --prune origin`; worktree from `origin/main`; pin SHA.

Do **not** merge, deploy, or restart production Omi apps.

Git-on-my-level:
## Summary

Desktop Chat treated UTC-naive SQLite DATETIME cells as local wall-clock, so “when did I last open Claude?” quoted **7:59:51 PM** for a **3:59:51 PM EDT** instant.

This PR converts `timestamp` / `*At` cells in `execute_sql` (and the matching Windows renderer) into the user’s timezone with an explicit IANA/offset label, repairs SQL “today/yesterday” examples so filters stay UTC-vs-UTC, and keeps the live clock on the per-turn `# Current Time` prefix instead of the cached system prompt.

## Test plan

- [x] Pinned clock `2026-08-27T19:59:51Z` + `America/New_York`: `execute_sql` / `formatRows` contain `3:59:51 PM` and a zone token; must not present unlabeled `7:59:51 PM`
- [x] `ChatPromptsTests` — cached desktop prompt has no `{current_datetime_str}`; SQL examples compare UTC columns to UTC bounds
- [x] Windows `desktopChatPrompt.test.ts` + `sql.test.ts` (88 tests)
- [x] `make preflight` on this branch

Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift | 3520 -> 3526 | thread display timezone through execute_sql and keep daily-recap bounds UTC-vs-UTC

- INV-AUTH-1
- INV-CHAT-1

Failure-Class: FC-machine-timezone-timestamp-in-prompt

Closes #12321

Related: #4643, #12236 (do not close)

SCA-364

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12326?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/src/main/assistants/insight/sql.ts
+++ b/desktop/windows/src/main/assistants/insight/sql.ts
@@ -19,6 +19,7 @@
 // NEVER log a query, a result cell, or a window title from here — SQL results are
 // raw OCR/screen text.
 import type { RewindFrame } from '../../../shared/types'
+import { formatSqlTimestampCell, resolveChatTimeZone } from '../../../shared/chatTimestamp'
 
 /** Mac's auto-limit + belt-and-suspenders row cap ("Auto-limited to 200 rows"). */
 export const MAX_ROWS = 200
@@ -599,6 +600,25 @@ export function rejectDangerousShape(sql: string): string | null {
   return null
 }
 
+/** Timestamp-shaped result columns are localized by formatRows. Applying
+ * SQLite's `localtime` modifier inside the SELECT projection would apply the
+ * offset twice. Boundary conversion in a WHERE clause remains valid. */
+function rejectProjectedLocalTime(sql: string): string | null {
+  const selectProjection = /\bselect\b([\s\S]*?)(?=\bfrom\b|$)/gi
+  const localTimeFunction =
+    /\b(?:date|time|datetime|julianday|unixepoch|strftime)\s*\([^;]*?['"]localtime['"]/i
+  for (const match of sql.matchAll(selectProjection)) {
+    if (localTimeFunction.test(match[1] ?? '')) {
+      return (
+        'Error: keep timestamp result expressions in UTC. ' +
+        'Select raw timestamp/*At columns so execute_sql can localize them once with an explicit zone. ' +
+        'Use localtime only when computing UTC WHERE bounds.'
+      )
+    }
+  }
+  return null
+}
+
 function cellToString(v: unknown): string {
   const s = v == null ? '' : String(v)
   return s.length > CELL_CAP ? `${s.slice(0, CELL_CAP)}...` : s
@@ -606,12 +626,25 @@ function cellToString(v: unknown): string {
 
 /** Mac's pipe-table rendering: header, divider, one line per (capped) row, a
  *  trailing `N row(s)`. Empty set → the literal `No results`. */
-export function formatRows(columns: string[], rows: unknown[][]): string {
+export function formatRows(
+  columns: string[],
+  rows: unknown[][],
+  timeZone: string = resolveChatTimeZone()
+): string {
   if (rows.length === 0) return 'No results'
   const capped = rows.slice(0, MAX_ROWS)
   const divider = '-'.repeat(Math.min(columns.length * 20, 120))
   const lines = [columns.join(' | '), divider]
-  for (const row of capped) lines.push(row.map(cellToString).join(' | '))
+  for (const row of capped) {
+    lines.push(
+      row
+        .map((cell, index) => {
+          const formatted = formatSqlTimestampCell(columns[index] ?? '', cell, timeZone)
+          return cellToString(formatted ?? cell)
+        })
+        .join(' | ')
+    )
+  }
   lines.push(`${capped.length} row(s)`)
   return lines.join('\n')
 }
@@ -750,6 +783,9 @@ export function executeSql(
 
   if (!isReadOnlySql(single)) return 'Error: only read-only SELECT/WITH queries are allowed'
 
+  const localTimeProjection = rejectProjectedLocalTime(single)
+  if (localTimeProjection) return localTimeProjection
+
   const denyTerms = denylist.map((t) => t.trim()).filter((t) => t.length > 0)
   const denyActive = denyTerms.length > 0
 
@@ -822,6 +858,9 @@ export function executeReadOnlySql(
   if (!isReadOnlySql(single)) {
     return 'Error: this SQL surface is read-only. Use SELECT or read-only WITH queries.'
   }
+
+  const localTimeProjection = rejectProjectedLocalTime(single)
+  if (localTimeProjection) return localTimeProjection
 
   // Even a valid read-only SELECT may only touch allowlisted tables — a closed
   // allowlist so meta/embedding/credential-shaped tables can never be read.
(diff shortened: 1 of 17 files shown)


desktop/windows/src/shared/chatTimestamp.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+  if (typeof value === 'string') {
+    const trimmed = value.trim()
+    if (!trimmed) return null
+    const naive = /^(\d{4}-\d{2}-\d{2})[ T](\d{2}:\d{2}:\d{2}(?:\.\d+)?)$/.exec(trimmed)
+    if (naive) {
+      const parsed = Date.parse(`${naive[1]}T${naive[2]}Z`)
+      if (Number.isFinite(parsed)) return new Date(parsed)
+    }
+    const iso = Date.parse(trimmed)
```
tianmind-studio, line 78: On Windows this turns a date-only cell into the previous calendar day: formatSqlTimestampCell('captureDate', '2026-08-27', 'America/New_York') returns 2026-08-26 8:00:00 PM EDT. The Swift parser only accepts values that include a time, so please keep date-only values unchanged (or render them as calendar dates without timezone conversion) and add this fixture.

desktop/windows/src/main/assistants/insight/sql.ts
```diff
@@ -606,12 +607,25 @@ function cellToString(v: unknown): string {
...
   if (rows.length === 0) return 'No results'
   const capped = rows.slice(0, MAX_ROWS)
   const divider = '-'.repeat(Math.min(columns.length * 20, 120))
   const lines = [columns.join(' | '), divider]
-  for (const row of capped) lines.push(row.map(cellToString).join(' | '))
+  for (const row of capped) {
+    lines.push(
+      row
+        .map((cell, index) => {
+          const formatted = formatSqlTimestampCell(columns[index] ?? '', cell, timeZone)
```
tianmind-studio, line 642: This per-cell path constructs two Intl.DateTimeFormat instances. On this Windows machine, 20,000 timestamp cells (200 rows x 100 model-controlled aliases) took 2.09 seconds synchronously; because execute_sql caps rows but not columns and runs in Electron main, prompt-injected SQL can freeze the UI. Please cap columns or total rendered cells and cache formatter instances by timezone.

desktop/windows/src/shared/chatTimestamp.ts
```diff
@@ -0,0 +1,99 @@
+/** Format timestamps that reach the desktop-chat model.
+ *
+ * SQLite / GRDB cells and Windows epoch-ms integers are UTC instants. Printing
+ * the UTC wall-clock without a zone made the model quote 7:59 PM for a 3:59 PM
+ * Eastern event (#12321). Always convert in an injected IANA zone and label it.
+ */
+
+const TIMESTAMP_COLUMN = /^(timestamp|ts|.*(?:At|_at|Date|_date))$/
```
tianmind-studio, line 8: TIMESTAMP and TS are not recognized here because this regex is case-sensitive, while Swift explicitly matches them case-insensitively. A model can alias ts AS TIMESTAMP and leave the exact unlabeled-UTC bug on Windows; a case-insensitive match plus a parity fixture would close it.

tianmind-studio:
#12349 has now landed the narrower formatter. While validating its interaction with this PR's macOS changes, I reproduced the remaining double-localization/local-day boundary gap as #12350 and opened #12351 as a focused post-merge follow-up, including a pinned non-UTC fixture and runtime SELECT-projection guard; this PR still contains the broader cross-surface work.

PR #12339, opened 2026-08-28
test(ci): restore backend unit and legacy ratchet checks

aryanorastar:
## Summary

- restore collection of the two hermetic `utils.app_integrations` test harnesses after the temporal grounding import added in #12236
- keep the legacy-memory surface ratchet focused on runtime references instead of counting a documentation-only macOS parity comment
- unblock the backend unit, repository hygiene, and PR metadata checks currently failing on `main` and downstream PRs such as #11183

## Root cause

`utils.app_integrations` now imports `current_date_for_uid` from `utils.llm.temporal`. The two isolated harnesses replace `utils.llm` with an in-memory module but did not register that new child module, so pytest stopped during collection with:

```text
ModuleNotFoundError: No module named 'utils.llm.temporal'; 'utils.llm' is not a package
```

The same upstream change also added `InsightAssistant` to a Windows source comment. The Gate F inventory intentionally scans that token family, so the comment alone moved the counter from 4 to 5 even though no legacy assistant surface was added.

## Changes

- register a deterministic `utils.llm.temporal` stub in both app-integration harnesses
- return a fixed date from `current_date_for_uid` so the tests remain hermetic and never load the production timezone/database path
- describe the macOS parity source without repeating the ratcheted legacy class name

There is no production behavior change.

## Verification

```text
PYTHON=backend/.venv/bin/python \
  BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-main-ci-failures.txt \
  BACKEND_PYTEST_WORKERS=3 bash backend/test.sh
# 35 passed across the exact three files reported by CI

python3 backend/scripts/legacy_memory_surface_inventory.py \
  --check-ratchet --base-ref origin/main
# PASS (892 findings across 82 path counters)

make preflight
# PASS (26 selected manifest checks)
```

## Product invariants

None.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12339?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/test_async_app_integrations.py
+++ b/backend/tests/unit/test_async_app_integrations.py
@@ -49,6 +49,7 @@ _utils_stubs = [
     "utils.llm",
     "utils.llm.clients",
     "utils.llm.proactive_notification",
+    "utils.llm.temporal",
     "utils.llm.usage_tracker",
     "utils.llms",
     "utils.llms.memory",
@@ -216,6 +217,11 @@ _proactive_mod.validate_notification = MagicMock(return_value=False)
 _proactive_mod.FREQUENCY_TO_BASE_THRESHOLD = {1: 0.5, 2: 0.4, 3: 0.3}
 _proactive_mod.MAX_DAILY_NOTIFICATIONS = 10
 _proactive_mod.Record = MagicMock
+
+# Stub the current-date helper imported by utils.app_integrations. Keeping it
+# inside this harness avoids pulling the real timezone/database path into this
+# otherwise hermetic unit test.
+sys.modules["utils.llm.temporal"].current_date_for_uid = MagicMock(return_value="2026-01-01")
 
 # Stub usage tracker
 _usage_mod = sys.modules["utils.llm.usage_tracker"]
--- a/backend/tests/unit/test_async_realtime_integrations_offload.py
+++ b/backend/tests/unit/test_async_realtime_integrations_offload.py
@@ -59,6 +59,7 @@ _utils_stubs = [
     "utils.llm",
     "utils.llm.clients",
     "utils.llm.proactive_notification",
+    "utils.llm.temporal",
     "utils.llm.usage_tracker",
     "utils.llms",
     "utils.llms.memory",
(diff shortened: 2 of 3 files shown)


tianmind-studio (commented):
Reviewed current head `6ff2a78382`; I did not find a blocking issue.

The two harnesses replace `utils.llm` with an in-memory module, so registering `utils.llm.temporal` before importing `utils.app_integrations` is the necessary isolation repair. The fixed `YYYY-MM-DD` return value matches `current_date_for_uid`'s production contract and avoids pulling the lazy timezone/database path into these tests. The Windows diff is comment-only and removes a false legacy-surface inventory hit without changing the prompt string or runtime behavior.

Independent verification on Windows:
- both changed backend test files: 25 passed
- `legacy_memory_surface_inventory.py --check-ratchet --base-ref origin/main`: PASS (892 findings / 82 path counters)
- `check_module_stub_pollution.py`: 0 violations across 1038 test files
- `git diff --check origin/main...HEAD`: clean

Leaving this as a contributor comment rather than maintainer approval, but the change is merge-shaped from this pass.

Git-on-my-level:
Verified end to end on the current head — this is a clean, behavior-free CI repair. Per file:

- `backend/tests/unit/test_async_app_integrations.py` — `utils.llm.temporal` is added to the same stub list and `_RESTORED_MODULES` teardown set as its siblings, so the stub module exists before `utils.app_integrations` (which does `from utils.llm.temporal import current_date_for_uid`) is imported and is removed symmetrically afterward — no new isolation gap. The fixed `"2026-01-01"` return matches the production contract in `utils/llm/temporal.py` (a `YYYY-MM-DD` string with a deferred, Firestore-touching timezone lookup), keeping these tests hermetic.
- `backend/tests/unit/test_async_realtime_integrations_offload.py` — identical registration and stub pattern (lines 62/228), consistent with the first harness.
- `desktop/windows/src/main/assistants/insight/prompt.ts` — comment-only rewrite; the prompt string and `formatDateTime` are untouched. The legacy-surface token family drops from 5 to 4 hits in this file (the removed one is exactly the token the upstream change added to a comment), and the remaining mentions are pre-existing, so the shrink-only ratchet returns to its baseline without loosening the inventory itself.

Checks are green on this head (backend unit suite, hermetic merge gate, hygiene/metadata), matching the PR's goal of unblocking `main`. Thanks for the precise root-cause write-up and for keeping the fix minimal.

Leaving the merge call to a maintainer — since this unblocks failing backend checks on `main` (and downstream PRs like #11183), a timely merge helps.

---

PR #12341, opened 2026-08-28
fix(desktop): no visible control deep-links to the hidden Task Assistant pane

kodjima33:
## What

Follow-up to #12327 — the amend carrying this was force-pushed after the merge watcher had already captured the pre-amend head, so the PR merged without it.

The Tasks-page gear posted `navigateToTaskSettings`, which opens Advanced and highlights `advanced.taskassistant` — a pane #12327 hides. A visible control deep-linking to a pane that no longer renders is exactly the dangling door that got the previous hide reverted by 73c7f85fbc. The gear is hidden under the same `HIDDEN DELIBERATELY` marker; the stale highlight is dropped (the notification still navigates to Advanced — the QA bridge shortcut uses it).

`HiddenSettingsSurfacesTests` pins all six hidden setting ids out of the settings search index and asserts the surviving floating-bar rows stay searchable.

## Verification

Running dev bundle built with this change: Tasks-page header shows only Search / Select / + — no gear (captured). `swift test --filter "HiddenSettingsSurfacesTests|SettingsAssistantControlsTests"` — 17 tests, 0 failures.

- INV-NAV-1 — touched paths match its globs; primary navigation is unchanged (the diff hides a secondary settings gear inside the Tasks page header and drops a stale settings highlight), so its guard test needs no update.
- INV-TASK-1 — touched paths match its globs; task extraction, storage, and completion behavior are untouched — only the header's settings gear visibility changes — so its guard test needs no update.

Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/TasksPage.swift | 6680 -> 6683 | three comment lines marking the gear as deliberately hidden

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12341?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/SettingsPage.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/SettingsPage.swift
@@ -723,9 +723,14 @@ struct SettingsContentView: View {
     }
     .onReceive(NotificationCenter.default.publisher(for: .navigateToTaskSettings)) { _ in
       selectedSection = .advanced
-      // The Task Assistant pane is hidden (HIDDEN DELIBERATELY, Nik 2026-08-25), so
-      // there is no advanced.taskassistant card to highlight — highlighting a missing
-      // target scrolled to nothing, which is how the pane got "fixed back" once before.
+      // The highlight target rides HiddenSettingsSurfacesPolicy: while the Task
+      // Assistant pane is hidden this is nil, because highlighting a card that does
+      // not render scrolls to nothing — how the pane got "fixed back" once before.
+      if let target = HiddenSettingsSurfacesPolicy.taskSettingsHighlight {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) {
+          highlightedSettingId = target
+        }
+      }
     }
     .onReceive(NotificationCenter.default.publisher(for: .navigateToFloatingBarSettings)) { _ in
       selectedSection = .floatingBar
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/SettingsPage.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/SettingsPage.swift
@@ -722,11 +722,14 @@ struct SettingsContentView: View {
       }
     }
     .onReceive(NotificationCenter.default.publisher(for: .navigateToTaskSettings)) { _ in
+      // The whole transition is data from SettingsDeepLinkTransition, so the test
+      // that pins it drives this exact production value. While the Task Assistant
+      // pane is hidden the highlight is nil — a highlight that targets a card that
+      // does not render scrolls to nothing, which is how the pane got "fixed back"
+      // once before.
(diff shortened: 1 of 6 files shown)


desktop/macos/Desktop/Sources/MainWindow/Pages/SettingsPage.swift
```diff
@@ -722,9 +722,17 @@ struct SettingsContentView: View {
...
+      // The whole transition is data from SettingsDeepLinkTransition, so the test
+      // that pins it drives this exact production value. While the Task Assistant
+      // pane is hidden the highlight is nil — a highlight that targets a card that
+      // does not render scrolls to nothing, which is how the pane got "fixed back"
+      // once before.
+      let transition = SettingsDeepLinkTransition.taskSettings()
       selectedSection = .advanced
-      DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) {
-        highlightedSettingId = "advanced.taskassistant"
+      if let target = transition.highlight {
```
tianmind-studio, line 732: [P2] Clear an existing highlight when this transition has no target`n`nThis branch does nothing when `transition.highlight` is nil. `highlightedSettingId` is parent-owned and can still contain a previous search/deep-link target, so `.navigateToTaskSettings` can open Advanced while retaining that stale row -- the opposite of "highlights nothing". Assign the transition value (including nil), and cover the transition from an initially non-nil ID.

desktop/macos/Desktop/Sources/MainWindow/Pages/TasksPage.swift
```diff
@@ -3912,7 +3912,12 @@ struct TasksPage: View {
           chatToggleButton
         }
         addTaskButton
-        taskSettingsButton
+        // HIDDEN DELIBERATELY (Nik, 2026-08-25): the gear deep-linked to the Task
+        // Assistant pane, which is hidden from Settings. TasksHeaderSettingsGear
+        // consults the policy and is host-tested; do not add a parallel gear here.
+        TasksHeaderSettingsGear {
```
tianmind-studio, line 3918: [P2] Remove the policy-bypassing gear implementation`n`nThe existing private `taskSettingsButton` at lines 4155-4173 remains a complete gear implementation that posts `.navigateToTaskSettings` without consulting this policy. It is unused today, but the new component test would still pass if that helper were wired back into the header, which is the exact regression this policy is intended to prevent. Remove the dead helper (also avoiding growth in this oversized file), or make the production TasksPage wiring itself mechanically guarded.

tianmind-studio (changes requested):
Two behavioral guard gaps remain. The failing Hygiene job also needs its line-count exception updated (the diff is +5, not +3).

PR #12343, opened 2026-08-28
fix(openapi): keep contract tooling UTF-8 on Windows

tianmind-studio:
## Summary

- make app-client generator tests, OpenAPI export/check tooling, schema inventory, and Dart wire hardening read and write repository text as UTF-8
- add one AST contract test that rejects locale-dependent `Path.read_text()` and `Path.write_text()` calls across the affected OpenAPI contract surfaces
- emit repository-relative schema inventory paths with forward slashes on Windows
- register the recurring host-locale failure class and its reusable guard, building on #9759

## Why

Python otherwise uses the active Windows code page for `Path.read_text()` and `Path.write_text()`. On a Chinese Windows host using CP936, UTF-8 OpenAPI contracts and source files fail before the contract checks can run. Once decoding is fixed, schema inventory JSON also differs from the checked contract because `str(Path.relative_to(...))` emits backslashes on Windows.

The guard is intentionally limited to the OpenAPI contract surfaces involved in this failure class; applying a repository-wide static rule would turn this portability fix into an unrelated migration. It would have caught both the generator failures addressed here and the same class previously repaired in #9759.

## Validation

- CP936 (`PYTHONUTF8=0`, preferred encoding `cp936`): 116 focused generator, exporter, inventory, assistant-settings, and developer API contract tests passed
- real CP936 CLI: app-client schema inventory JSON generation passed
- real CP936 exporter checks: app-client spec comparison passed and a non-CP936 test string was written as UTF-8 with LF newlines
- Dart, Swift, and TypeScript generator `--check` commands passed
- Black 26.5.1 passed on all changed Python files
- Ruff 0.4.4 passed on changed production files and tests, excluding 10 unchanged baseline findings reproduced on `origin/main`
- `git diff --check` passed
- diff-scoped local PR preflight passed all selected checks except two candidate-probe cases that require POSIX response deadlines and fail identically on clean `origin/main`; all 12 runtime-image source closures and all 30 desktop release fixture cases passed with verified `jq 1.8.2`

Product invariants: none.

Changelog: not required; this changes repository tooling and contract tests only.

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-repository-text-io-inherits-host-locale.json
+++ b/.github/failure-classes/FC-repository-text-io-inherits-host-locale.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-repository-text-io-inherits-host-locale",
+  "violated_contract": "Repository text artifacts must be decoded and encoded consistently on every supported host; relying on the process locale makes UTF-8 source and generated files unreadable on Windows code pages such as CP936.",
+  "canonical_prevention": "Declare UTF-8 on every pathlib text read and write in portable generator scripts and their contract tests, and enforce that property with an AST-based test covering all app-client generator surfaces.",
+  "canonical_prevention_artifact": [
+    "backend/tests/unit/test_app_client_generator_utf8_contract.py"
+  ],
+  "evidence_prs": [
+    9759
+  ],
+  "scope_hints": [
+    "backend/scripts/generate_*",
+    "backend/tests/unit/test_app_client_*_generator.py"
+  ],
+  "status": "open"
+}
--- a/.github/failure-classes/FC-repository-text-io-inherits-host-locale.json
+++ b/.github/failure-classes/FC-repository-text-io-inherits-host-locale.json
@@ -2,7 +2,7 @@
   "schema_version": 1,
   "id": "FC-repository-text-io-inherits-host-locale",
   "violated_contract": "Repository text artifacts must be decoded and encoded consistently on every supported host; relying on the process locale makes UTF-8 source and generated files unreadable on Windows code pages such as CP936.",
-  "canonical_prevention": "Declare UTF-8 on every pathlib text read and write in portable generator scripts and their contract tests, and enforce that property with an AST-based test covering all app-client generator surfaces.",
+  "canonical_prevention": "Declare UTF-8 on every pathlib text read and write in portable OpenAPI contract scripts and tests, and enforce that static property with one AST-based guard covering the generator, exporter, schema inventory, and REST inventory surfaces.",
   "canonical_prevention_artifact": [
     "backend/tests/unit/test_app_client_generator_utf8_contract.py"
   ],
@@ -10,8 +10,10 @@
     9759
   ],
   "scope_hints": [
-    "backend/scripts/generate_*",
-    "backend/tests/unit/test_app_client_*_generator.py"
+    "backend/scripts/*openapi*",
+    "backend/scripts/inventory_app_client_schemas.py",
+    "backend/tests/unit/test_*openapi*.py",
+    "backend/tests/unit/test_*_rest_inventory.py"
   ],
   "status": "open"
 }
--- a/backend/scripts/export_openapi.py
+++ b/backend/scripts/export_openapi.py
@@ -1061,7 +1061,7 @@ def stable_json(schema: dict[str, Any]) -> str:
 
 def write_spec(path: Path, generated: str) -> None:
     path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(generated)
+    path.write_text(generated, encoding='utf-8', newline='\n')
 
 
 def regenerate_hint(path: Path, surface: str) -> str:
@@ -1081,7 +1081,7 @@ def check_spec(path: Path, generated: str, *, surface: str = 'public') -> None:
     hint = regenerate_hint(path, surface)
     if not path.exists():
         raise OpenAPIContractError(f'{path} does not exist; run {hint}')
-    current = path.read_text()
+    current = path.read_text(encoding='utf-8')
     if current != generated:
         raise OpenAPIContractError(f'{path} is stale; run {hint}')
 
--- a/backend/scripts/inventory_app_client_schemas.py
+++ b/backend/scripts/inventory_app_client_schemas.py
@@ -69,6 +69,11 @@ RAW_DECODE_RE = re.compile(
 WIRE_DECODE_RE = re.compile(r'wire\.Generated[A-Za-z0-9_]+\.fromJson|\.fromGenerated\s*\(|fromGeneratedWireJson')
 
 
+def repo_relative_path(path: Path) -> str:
+    """Return a stable repository path for JSON reports and diagnostics."""
+    return path.relative_to(ROOT_DIR).as_posix()
+
+
 def _collect_wire_backed_type_names() -> list[str]:
     """Collect Dart type names whose fromJson delegates to wire.Generated* parsers.
 
@@ -113,7 +118,7 @@ class DartSchemaFile:
 
     def to_report(self) -> dict[str, Any]:
         return {
-            'path': str(self.path.relative_to(ROOT_DIR)),
+            'path': repo_relative_path(self.path),
             'classes': self.classes,
             'enums': self.enums,
             'fromJson': self.from_json_count,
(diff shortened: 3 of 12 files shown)


Git-on-my-level (approved):
Verified end to end in a clean worktree at 4f45b5d:

- Ran the four app-client generator suites (`test_app_client_dart_generator.py`, `test_app_client_swift_generator.py`, `test_app_client_ts_generator.py`, `test_app_client_generator_utf8_contract.py`): 43 passed, matching the PR's validation claim.
- Confirmed the premise at the byte level: `docs/api-reference/app-client-openapi.json`, `app/lib/backend/schema/message.dart`, `person.dart`, and `action_item.dart` all contain UTF-8 bytes that raise `UnicodeDecodeError` when decoded as CP936 — exactly what the old bare `Path.read_text()` calls hit on a Chinese Windows host. Every read in the two modified test files now pins `encoding='utf-8'`.
- The new AST guard in `test_app_client_generator_utf8_contract.py` is sound as scoped: re-ran its exact `read_text`/`write_text` walk against all seven `TEXT_IO_SURFACES` — zero violations on this head, and the three generator scripts plus the TS test were already clean from #9759. The guard turns that portability work into an enforced contract rather than a one-off fix, and a static tripwire is the right tool since the violation is a static property of the source.
- `FC-repository-text-io-inherits-host-locale.json` matches the registry's schema and conventions: no existing class covers host-locale text I/O, `canonical_prevention_artifact` points at the new guard, and #9759 is cited as evidence.

One non-blocking note for a possible follow-up (out of scope here): `backend/tests/unit/test_desktop_rest_inventory.py:100,107` still reads the same CP936-breaking spec via bare `SPEC_PATH.read_text()` and is not in the guard's surface list. Extending `TEXT_IO_SURFACES` later would be the natural next step if desktop contributors also run the suite on Windows.

Thanks for the careful follow-through on the Windows generator portability arc — registering the failure class together with a reusable guard is exactly the right shape.

---

tianmind-studio:
Expanded this after the Windows audit to cover exporter/check, schema inventory, and Dart hardening under the same failure class; the CP936-focused suite is now 109/109, and inventory paths are canonicalized to forward slashes. The previous approval covered 4f45b5d only, so the new head aa486b5 needs a fresh look.

tianmind-studio:
The only red check on the current head is inherited from main: all three failures reproduce at origin/main 2f2be5a5f8. They are covered by #12339, whose Backend unit suite and full check set are green. I independently verified #12339 on Windows at 6ff2a78382; once that baseline repair lands, this PR can be rerun without mixing unrelated test-harness changes into the OpenAPI portability patch.

Git-on-my-level (commented):
Fresh look at aa486b5 — the two commits after the previously reviewed 4f45b5d — as requested. The delta extends the same UTF-8 pinning pattern from the generator tests to the contract tooling, and it holds up:

- `backend/scripts/export_openapi.py`: `write_spec()` now passes `encoding='utf-8', newline='\n'` and `check_spec()` reads with `encoding='utf-8'` — exactly the two locale-sensitive spots in the exporter, and newline pinning keeps the written spec byte-stable on CRLF hosts.
- `backend/scripts/inventory_app_client_schemas.py`: every `read_text()` is pinned, and the new `repo_relative_path()` helper replaces the `str(path.relative_to(ROOT_DIR))` report sites with `.as_posix()` so inventory JSON stays forward-slashed on Windows. The `PureWindowsPath('C:/repo')` monkeypatch test in `test_app_client_schema_inventory.py` pins that behavior without needing a Windows host.
- `backend/scripts/run_dart_wire_hardening_check.py`: adapter read pinned; both temp-file writes pinned with `newline='\n'` (fine on the repo's 3.11 floor).
- `backend/tests/unit/test_app_client_generator_utf8_contract.py`: the AST guard grows from 7 to 19 surfaces, and the failure-class entry's `canonical_prevention`/`scope_hints` were updated to match the widened scope. I re-ran the scope-hint globs against the guard's surface list: the only gaps are bare `write_text()` calls on ASCII `tmp_path` fixtures in `test_export_openapi_hint.py` / `test_openapi_contract.py` — outside this failure class (no repository-resident UTF-8 text), so the scoping is defensible.
- The remaining test files (`test_app_client_dart_generator.py`, `test_app_client_swift_generator.py`, `test_assistant_settings_response_models.py`, `test_desktop_rest_inventory.py`, `test_dev_ask_endpoint.py`, `test_integration_public_contract.py`, `test_app_client_schema_inventory.py`) are mechanical `encoding='utf-8'` additions; `test_desktop_rest_inventory.py:100,107` — called out as a follow-up in the earlier review — are now pinned and inside the guard.

Local verification at aa486b5: guard, schema-inventory, Dart/Swift/TS generator, assistant-settings, desktop-rest-inventory, and public-contract suites pass, including under a forced non-UTF-8 preferred encoding (`PYTHONUTF8=0 LC_ALL=C`, ASCII — stricter than CP936), where any remaining bare `read_text()` of non-ASCII repo text would raise immediately. `scripts/test_failure_class.py` passes 24/24 with the updated registry entry. I could not run `test_dev_ask_endpoint.py` locally (my sandbox lacks `firebase_admin`); its change is mechanical and CI's hermetic suites exercise it.

On the failing Backend unit suite check: this is a pre-existing main-branch flake, not a regression from this diff. The failure is `ModuleNotFoundError: No module named 'utils.llm.temporal'` while collecting `test_async_app_integrations.py` / `test_async_realtime_integrations_offload.py` — files (and `backend/utils/llm/`) that are byte-identical between this PR's base and head. The same collection error appears on main runs 33140652958 and 33124088014 and cleared on 33169706189. That looks like a runner/install-level namespace issue worth a separate infra fix, but it should not gate this PR.

Maintainer note before merge: please confirm the red CI ✗ is tolerable given the flake evidence above, and give the widened `scope_hints`/`canonical_prevention` wording in `FC-repository-text-io-inherits-host-locale.json` a quick read — the failure-class registry is maintainer-facing process documentation, so I'm leaving its final sign-off to a human.

---

Git-on-my-level (commented):
Follow-up at e55613a — the branch was rebased after the last review, so here is the delta rather than a repeat of the file-by-file pass.

- Content-neutral rebase: I diffed all 12 changed files between the previously reviewed aa486b5 and this head — byte-identical. The only movement is the base, which now includes #12339.
- The one open item from the last review is resolved: the Backend unit suite check is green on this head, consistent with the #12339 baseline repair landing in the base rather than anything in this diff. The full check set passes.
- Fresh verification at e55613a in a clean worktree: the UTF-8 AST guard (`test_app_client_generator_utf8_contract.py`, 19 surfaces), schema-inventory suite including the new `PureWindowsPath` forward-slash test (`test_app_client_schema_inventory.py`), Dart/Swift generator suites, desktop REST inventory, and public contract suites all pass (68 tests) — and still pass under `PYTHONUTF8=0 LC_ALL=C` (ASCII preferred encoding, stricter than CP936, where any remaining bare `read_text()` of non-ASCII repo text would fail immediately). `scripts/failure-class validate` passes with the new registry entry. (`test_assistant_settings_response_models.py` I could not run locally — my sandbox lacks `langchain_anthropic`; its change is the same one-line `encoding='utf-8'` pin as elsewhere and CI runs it green.)
- The core fixes are unchanged and hold up: `export_openapi.py` `write_spec()`/`check_spec()` pin encoding and LF newlines; `inventory_app_client_schemas.py` routes every report path through `repo_relative_path()` (`.as_posix()`); `run_dart_wire_hardening_check.py` pins its temp-file writes.

Two items remain for a human maintainer, both judgment calls rather than defects:
1. Final wording of `FC-repository-text-io-inherits-host-locale.json` — it defines a repo-wide engineering norm (which surfaces must pin UTF-8 on text I/O, which are exempt) that future fix PRs will be measured against.
2. The merge call on a `.github/`-touching portability change.

---

kodjima33 (approved):
Already CLEAN/APPROVED with green CI. Verified genuine host-locale UTF-8 bug (unfixed on main) matching established pattern from merged #9759. Confidence 5/5.

PR #12344, opened 2026-08-28
fix(ci): isolate candidate probe deadlines on Windows

tianmind-studio:
## Summary

- isolate ordinary desktop-backend candidate-probe fixtures from the POSIX-only total-response deadline with one autospecced class-level seam
- keep the real `SIGALRM` interruption contract in a dedicated platform-gated test class, so Linux CI continues to exercise production deadline behavior
- register the recurring class where an orthogonal fixture enters an unrelated platform-specific boundary, with #10845 as prior evidence

## Why

The local manifest intentionally runs candidate-probe fixtures for broad backend changes. On native Windows, two newly added Gemini request tests fail before reaching their assertions because `_total_response_budget()` rejects hosts without `SIGALRM` and `setitimer`.

Before this change on Windows: **1 failure, 1 error, 15 passed, 1 skipped**. After this change: **17 passed, 1 POSIX-only skipped**. Production probe behavior is unchanged.

## Validation

- `python .github/scripts/test_desktop_backend_candidate_probe.py` (`17 passed, 1 skipped` on Windows)
- `ruff check .github/scripts/test_desktop_backend_candidate_probe.py`
- `black --check --line-length 120 --skip-string-normalization .github/scripts/test_desktop_backend_candidate_probe.py`
- `python -m py_compile .github/scripts/test_desktop_backend_candidate_probe.py .github/scripts/desktop_backend_candidate_probe.py`
- `python .github/scripts/pr_preflight.py --lane local --base origin/main --pr-body-file .pr-body-windows-candidate-probe-sigalrm.md`

Product invariants: none.

Changelog: not required; this changes hermetic CI fixtures only.

Failure-Class: new

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12344?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-orthogonal-fixture-enters-platform-boundary.json
+++ b/.github/failure-classes/FC-orthogonal-fixture-enters-platform-boundary.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": 1,
+  "id": "FC-orthogonal-fixture-enters-platform-boundary",
+  "violated_contract": "A hermetic fixture for behavior inside a platform-specific security or deadline boundary must not execute that unrelated boundary; otherwise supported hosts that intentionally reject the primitive fail before the fixture reaches its assertion.",
+  "canonical_prevention": "Isolate the platform-specific primitive behind a suite-wide test seam for orthogonal cases, and keep the real boundary contract in a dedicated platform-gated test so newly added behavior cases inherit portability automatically.",
+  "canonical_prevention_artifact": [
+    ".github/scripts/test_desktop_backend_candidate_probe.py"
+  ],
+  "evidence_prs": [
+    10845
+  ],
+  "scope_hints": [
+    ".github/scripts/*candidate_probe*.py"
+  ],
+  "status": "open"
+}
--- a/.github/scripts/test_desktop_backend_candidate_probe.py
+++ b/.github/scripts/test_desktop_backend_candidate_probe.py
@@ -9,6 +9,7 @@ import signal
 import sys
 import tempfile
 import unittest
+from contextlib import nullcontext
 from unittest import mock
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
Verified end to end — a precise, well-scoped fix. Details for the record:

`.github/scripts/test_desktop_backend_candidate_probe.py`
- The new `CandidateProbeTests.setUp` seam patches `_total_response_budget` with `autospec=True` and `side_effect=lambda **_kwargs: nullcontext()`. That matches the production signature exactly (keyword-only `stage` / `started_at` / `max_elapsed_seconds`), so future signature drift in `desktop_backend_candidate_probe.py` breaks the mock loudly instead of silently passing.
- The real deadline contract stays covered on Linux: `CandidateProbeDeadlineTests.test_gemini_response_read_is_interrupted_at_total_budget` is POSIX-gated and invokes the armed SIGALRM handler directly (no real 70s wait), so CI keeps exercising the interruption path.
- Independently reproduced both sides: with `SIGALRM`/`setitimer` hidden (Windows-like host), `main` fails with exactly the reported 1 failure + 1 error (`"admitted provider" does not match "gemini_proxy: total response deadline is unsupported on this runner"`), and with this change the same suite is green with 1 POSIX-only skip; on Linux all 18 tests pass.

`.github/failure-classes/FC-orthogonal-fixture-enters-platform-boundary.json`
- Conforms to the registry schema in `docs/product/failure-classes.md`: id matches the filename, `evidence_prs` cites merged #10845 (same recurring Windows/SIGALRM theme), `canonical_prevention_artifact` points at an existing path, and the PR body declares `Failure-Class: new` while touching no other definition — the documented contract for adding a class.

Also confirmed `desktop_backend_candidate_probe.py` itself is untouched (byte-identical to `main`), so production probe behavior is unchanged, exactly as claimed.

Since this edits CI scripts and the failure-class registry, leaving the merge sign-off to a human maintainer.

---

tianmind-studio:
Fresh cross-PR evidence from the #12343 refresh: on native Windows against cdca0c865f, the diff-scoped preflight passed all 12 runtime-image closures and then hit the same candidate-probe baseline (1 failure, 1 error) because the POSIX total-response deadline is unavailable. A clean main worktree reproduced the identical result; this PR's isolated suite remains 17 passed with only the real POSIX deadline case skipped.

kodjima33 (approved):
Verified genuine SIGALRM/Windows test isolation bug (unfixed on main), test-only diff, matches established pattern from merged #10845. Confidence 5/5.

PR #12346, opened 2026-08-28
fix(preflight): keep Windows subprocess text UTF-8

tianmind-studio:
## What changed and why

Make the Windows preflight text pipeline an explicit UTF-8 contract end to end:

- decode captured `gh pr view` JSON as UTF-8 instead of the host code page;
- configure the single-flight runner and direct `pr-preflight` output streams as UTF-8 on Windows;
- propagate UTF-8 Python stdio through owned child process trees;
- decode `pr-preflight --suggest` child output and the current Git branch as UTF-8;
- pin every nested captured subprocess in the preflight contract tests to UTF-8;
- keep the native Unicode regressions in the manifest-owned portability suite that runs on `windows-latest`.

On a Chinese Windows host where Python prefers CP936, the live PR metadata reader raised `UnicodeDecodeError` on valid UTF-8 JSON. Once the runner correctly forced owned children to emit UTF-8, nested test harnesses still decoded those pipes as GBK, and direct `--suggest` output could replace valid Unicode characters. This closes all three layers rather than fixing only the first crash.

none

## How it was verified

- `MetadataTests`: 14/14 passed; a real `load_from_gh` call loaded #12346 with `PYTHONUTF8=0` and no `PYTHONIOENCODING` on a CP936 host.
- `test_preflight_runner.py`: 25/25 passed on this head in native Windows CP936 mode.
- The new production-helper regressions launch a Python child containing Chinese plus emoji and read a Unicode Git branch with UTF-8 mode disabled.
- The direct `pr_preflight.py --suggest` path passed in CP936 mode and emitted its Unicode separator intact.
- All seven affected nested subprocess tests passed with the runner's UTF-8 child environment.
- Combined Windows audit with #12347 and #12348: 42/42 runnable `test_pr_preflight` cases passed (2 expected platform skips; only the GNU `make` prerequisite case was omitted), and the merged portability suite passed 29/29.
- Product invariants, failure-class validation and guard ratchet, 24 failure-class CLI tests, changelog data, deferred markers, plan catalog, lifecycle headers, and version-prefixed filenames all passed explicitly.
- Ruff, Black with repository options, `py_compile`, security specialist review, and `git diff --check` passed.

The diff-scoped local preflight passed through its first six checks. Its remaining full-suite blockers are external to this PR: GNU `make` is unavailable on this host, while the exit-259 liveness and atomic status replacement baselines are independently fixed by #12347 and #12348 and pass in the combined audit above.

## Tests

- `MetadataTests.test_gh_loader_decodes_current_metadata_as_utf8`
- `SelectionTests.test_captured_python_output_is_utf8_when_host_utf8_mode_is_disabled`
- `SelectionTests.test_current_branch_decodes_utf8_when_host_utf8_mode_is_disabled`
- `LaunchContractTests.test_runner_emits_utf8_from_unicode_checkout_without_utf8_mode`
- `LaunchContractTests.test_pr_preflight_emits_utf8_from_unicode_branch_without_utf8_mode`

## Failure class (fixes)

Failure-Class: new

Violated contract: captured subprocess text must use the producer's encoding, not an implicit host locale.

Canonical guard: declare UTF-8 at captured CLI decoders, configure direct Windows output, and propagate the same contract through owned process trees; exercise non-ASCII metadata, branch names, and child output under native Windows with UTF-8 mode disabled.

Supporting evidence: merged fixes #10595 and #10844 corrected the same locale inheritance at adjacent Git and failure-class boundaries; this PR closes the remaining PR metadata, preflight suggestion, nested-test, and single-flight runner surfaces.

### DIFF
--- a/.github/failure-classes/FC-subprocess-output-inherits-host-locale.json
+++ b/.github/failure-classes/FC-subprocess-output-inherits-host-locale.json
@@ -2,9 +2,10 @@
   "schema_version": 1,
   "id": "FC-subprocess-output-inherits-host-locale",
   "violated_contract": "Captured text from a repository or developer CLI must be decoded with the producer's encoding, not the host locale; otherwise valid UTF-8 paths, history, or structured metadata can terminate the subprocess reader on Windows and leave policy checks without output.",
-  "canonical_prevention": "Declare strict UTF-8 decoding at every captured CLI text boundary and keep a regression contract that feeds non-ASCII output while asserting the subprocess invocation's encoding.",
+  "canonical_prevention": "Declare strict UTF-8 decoding at every captured CLI text boundary and keep a regression contract that feeds non-ASCII output while asserting the subprocess invocation's encoding. For an owned process tree, configure the runner's streams and child Python stdio as UTF-8 before decoding the captured pipe as UTF-8.",
   "canonical_prevention_artifact": [
-    ".github/scripts/test_pr_preflight.py"
+    ".github/scripts/test_pr_preflight.py",
+    ".github/scripts/test_preflight_runner.py"
   ],
   "evidence_prs": [
     10595,
--- a/.github/failure-classes/FC-subprocess-output-inherits-host-locale.json
+++ b/.github/failure-classes/FC-subprocess-output-inherits-host-locale.json
@@ -1,0 +1,19 @@
+{
+  "schema_version": 1,
+  "id": "FC-subprocess-output-inherits-host-locale",
+  "violated_contract": "Captured text from a repository or developer CLI must be decoded with the producer's encoding, not the host locale; otherwise valid UTF-8 paths, history, or structured metadata can terminate the subprocess reader on Windows and leave policy checks without output.",
+  "canonical_prevention": "Declare strict UTF-8 decoding at every captured CLI text boundary and keep a regression contract that feeds non-ASCII output while asserting the subprocess invocation's encoding.",
+  "canonical_prevention_artifact": [
+    ".github/scripts/test_pr_preflight.py"
+  ],
+  "evidence_prs": [
+    10595,
+    10844
+  ],
+  "scope_hints": [
+    ".github/scripts/pr_*.py",
+    ".github/scripts/*preflight*.py",
+    ".github/scripts/*failure_class*.py"
+  ],
+  "status": "open"
+}
--- a/.github/scripts/pr_metadata.py
+++ b/.github/scripts/pr_metadata.py
@@ -127,6 +127,7 @@ def load_from_gh(root: Path) -> PullRequestMetadata:
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             text=True,
+            encoding="utf-8",
         )
     except FileNotFoundError as exc:
         raise RuntimeError("gh is not installed; pass --pr-body-file before the first push") from exc
--- a/.github/scripts/pr_preflight.py
+++ b/.github/scripts/pr_preflight.py
@@ -74,6 +74,18 @@ def current_branch(root: Path) -> str:
     ).stdout.strip()
 
 
+def configure_output_streams() -> None:
+    """Keep direct Windows preflight output UTF-8 and non-fatal."""
+    for stream in (sys.stdout, sys.stderr):
+        reconfigure = getattr(stream, "reconfigure", None)
+        if not callable(reconfigure):
+            continue
+        options = {"errors": "backslashreplace"}
+        if os.name == "nt":
+            options["encoding"] = "utf-8"
+        reconfigure(**options)
+
+
 def changed_files(root: Path, base: str, head: str) -> list[str]:
     output = run_git(
         root,
@@ -213,6 +225,7 @@ def parse_args() -> argparse.Namespace:
 
 
 def main() -> int:
+    configure_output_streams()
     args = parse_args()
     if bool(args.repository) != bool(args.pr_number):
         print("FAIL: --repository and --pr-number must be supplied together", file=sys.stderr)
--- a/.github/scripts/pr_preflight.py
+++ b/.github/scripts/pr_preflight.py
@@ -44,6 +44,36 @@ def run_git(root: Path, *args: str) -> str:
     return result.stdout.strip()
 
 
+def run_python_capture(root: Path, *args: str, errors: str = "strict") -> subprocess.CompletedProcess[str]:
+    """Run an owned Python check with a UTF-8 pipe contract on every host."""
+    env = os.environ.copy()
+    env["PYTHONIOENCODING"] = "utf-8:backslashreplace"
+    return subprocess.run(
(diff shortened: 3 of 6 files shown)


Git-on-my-level:
Reviewed the full diff and verified at this head: both changed suites pass locally on Linux (MetadataTests 14/14; test_preflight_runner 25/25 with the six Windows-only tests skipping as designed), and the windows-latest "Native Windows process and launcher contracts" job is green.

File by file:

- `.github/scripts/pr_metadata.py` — `load_from_gh` now pins `encoding="utf-8"` next to `text=True`. Correct and minimal fix for the CP936 `UnicodeDecodeError` on `gh` JSON output, consistent with the already-pinned boundaries in `run_checks.py`.
- `.github/scripts/preflight_runner.py` — `configure_output_streams()` runs at the top of `main()`, before the `--windows-child-bootstrap` branch, so the bootstrap shim inherits the same UTF-8/backslashreplace stream contract; `PYTHONIOENCODING=utf-8:backslashreplace` on the child env pins the owned process tree without mutating the host environment. The runner's own pipe decode keeps `errors="backslashreplace"`, so non-Python children emitting legacy code-page bytes still cannot kill the reader — the existing invalid-UTF-8 escape test preserves exactly that.
- `.github/scripts/test_preflight_runner.py` — the strengthened Windows test (PYTHONUTF8=0, PYTHONIOENCODING removed, child prints 路径🚀) now asserts both captured stdout and the UTF-8 log; the log-path assertion matches `default_state_dir`'s `<git-dir>/omi-preflight/<name>` layout now that the env override is dropped.
- `.github/scripts/test_pr_preflight.py` — the new `test_gh_loader_decodes_current_metadata_as_utf8` locks the loader contract by asserting `encoding="utf-8"` in the subprocess kwargs; removing the cross-platform `test_runner_starts_from_unicode_checkout` is compensated: the same scenario now runs where it can actually fail, on windows-latest via the `preflight-runner-portability` manifest check.
- `.github/failure-classes/FC-subprocess-output-inherits-host-locale.json` — matches the existing failure-class schema and conventions, cites real evidence (#10595, #10844), and the strict-UTF-8-boundary contract it records is exactly what this PR enforces.

Two non-blocking observations:

1. The new FC's `scope_hints` (`pr_*.py`) still match live surfaces: `pr_preflight.py` has three captured-CLI boundaries decoding with the host locale (the `text=True` subprocess calls around lines 223/250/282 — invariants/failure-class prep and the `git symbolic-ref` branch probe). A small follow-up sweep pinning those (with `errors="backslashreplace"` where output is informational) would close the class for real.
2. Forcing `PYTHONIOENCODING=utf-8:backslashreplace` over the whole owned child tree means any future Python check asserting exact stderr text of its own subprocesses could see backslash escapes on Windows hosts. I didn't find such a test today; just noting the coupling.

This is workflow-sensitive CI tooling, so per repo policy the merge call stays with a human maintainer. Thanks for the careful fix and the exceptional verification notes — the CP936 repro writeup made this straightforward to check.

---

tianmind-studio:
Closed the follow-up surface noted in the prior review at head 705cdc4c05: pr_preflight now pins its owned Python checks and Git branch probe, emits direct Windows output as UTF-8, and its nested contract-test subprocesses use the same decoder. Native CP936 checks and the combined #12347/#12348 audit are green; this head needs a fresh review.

Git-on-my-level:
Re-reviewed at 705cdc4c after the follow-up push. The remaining surface from the prior review — the unpinned `pr_preflight.py` capture boundaries — is now closed, and the three new commits keep the same layering discipline as the first two.

File by file at this head:

- `.github/scripts/pr_metadata.py` — unchanged since the prior review; `load_from_gh` still pins `encoding="utf-8"` next to `text=True`, consistent with the existing `run_git` boundary in `pr_preflight.py`.
- `.github/scripts/pr_preflight.py` — the new `run_python_capture()` helper gives both `--suggest` children (`check_product_invariants.py`, `scripts/failure-class prepare`) a UTF-8 pipe contract and pins `PYTHONIOENCODING=utf-8:backslashreplace` for the owned child tree, without mutating the parent environment. The asymmetry is deliberate and correct: the failure-class JSON path decodes `strict` (structured output must not be silently mangled) while the advisory invariants suggestion uses `backslashreplace`; `stderr=STDOUT` semantics are preserved from the old inline calls. `current_branch()` replaces the locale-inheriting inline `symbolic-ref` probe with the same UTF-8/backslashreplace contract. `configure_output_streams()` only pins the encoding on `os.name == "nt"` — POSIX behavior is untouched apart from the non-fatal error handler, which is the right amount of restraint.
- `.github/scripts/preflight_runner.py` — `configure_output_streams()` runs at the top of `main()` before the Windows child-bootstrap branch, so the bootstrap shim shares the contract; the runner passes `PYTHONIOENCODING` only to children it launches (single-flight), leaving the host environment alone.
- `.github/scripts/test_pr_preflight.py` — the new regressions assert both behavior (`stdout == "路径🚀\n"`, branch `分支-🚀`) and the contract itself (`kwargs["encoding"] == "utf-8"` on the loader), with `PYTHONUTF8=0` and `PYTHONIOENCODING` stripped so a regression actually fails. Pinning `encoding="utf-8"` on the nested harness subprocesses makes the tests independent of the host locale. Removing the cross-platform `test_runner_starts_from_unicode_checkout` is compensated: the strengthened equivalents live in the windows-latest portability suite where CP936 actually bites.
- `.github/scripts/test_preflight_runner.py` — the strengthened Windows tests now drop `PYTHONIOENCODING` from the environment (previously the test masked exactly the regression it claimed to guard), print real non-ASCII through the child, assert the log at the default `.git/omi-preflight/<name>/preflight.log` location, and verify the branch probe end-to-end via `head=分支-🚀`.
- `.github/failure-classes/FC-subprocess-output-inherits-host-locale.json` — matches the existing failure-class schema and conventions, cites real evidence PRs, and records precisely the contract this change enforces.

Verification at this head: `test_preflight_runner` 25/25 on Linux (6 Windows-only skips as designed), the three new cross-platform regressions pass under a stripped `PYTHONUTF8=0`/no-`PYTHONIOENCODING` environment, and CI is green including the windows-latest "Native Windows process and launcher contracts" job.

One non-blocking note for a follow-up: `run_python_capture` defaults to `errors="strict"`, so a child that emits stray non-UTF-8 bytes (e.g. a native traceback on a legacy console) would raise in the parent rather than degrade gracefully. That is stricter than the runner's own `backslashreplace` reader and no worse than the previous locale decode, so it is fine to merge as-is; worth a look if the strict JSON path ever reports decode crashes.

This stays `workflow-review`: it modifies CI/preflight behavior, so it needs maintainer sign-off before merge. Thanks for closing the loop on the earlier feedback so thoroughly.

---

kodjima33 (approved):
Verified genuine CP936/Windows encoding bug (unfixed on main), diff is encoding-only with new regression tests, matches established pattern from merged #10595/#10844. Confidence 5/5.

PR #12347, opened 2026-08-28
fix(ci): distinguish Windows exit 259 from liveness

tianmind-studio:
## What changed and why

Use the Windows process object's signaled state for preflight-runner liveness instead of interpreting GetExitCodeProcess value 259 as authoritative. Windows explicitly permits a completed process to return 259, the same value as STILL_ACTIVE; the old check therefore left that process's single-flight lock classified as live forever.

The fix uses a zero-timeout WaitForSingleObject result as the liveness source while retaining creation-time matching for PID reuse:
- https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess
- https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject

none

## How it was verified

- Reproduced the existing native-Windows exit-259 regression three times on main: 3/3 failed.
- Ran .github/scripts/test_preflight_runner.py after the fix: 26/26 passed on native Windows.
- Re-ran the exit-259 regression three times after the fix: 3/3 passed.
- Ruff, py_compile, failure-class validation, and git diff --check passed.
- The full diff-scoped Windows preflight passed its first six checks, including the complete portability suite, then stopped in the broader current-main contract suite on three independent host failures: no local make executable, Unicode child output decoded as UTF-8, and an intermittent status.json replacement race. The replacement race also reproduced 4/5 times from an unmodified main-derived worktree.

## Tests

Moved test_process_that_exits_with_still_active_status_is_not_alive into the manifest-owned preflight-runner portability suite. The windows-latest portability job now executes this regression instead of leaving it in a suite that only skips it on Linux CI.

## Failure class (fixes)

Failure-Class: new

Violated contract: a status sentinel cannot prove liveness when a completed operation may return the same integer.

Canonical guard: use a zero-timeout wait on the Windows process handle, retain creation-time matching for PID reuse, and execute an actual process that exits 259 in native Windows CI.

Supporting evidence: the existing regression failed deterministically before this change, and the Win32 API contract explicitly warns that exit code 259 is ambiguous.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12347?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-sentinel-exit-code-misclassified-as-liveness.json
+++ b/.github/failure-classes/FC-sentinel-exit-code-misclassified-as-liveness.json
@@ -1,0 +1,15 @@
+{
+  "schema_version": 1,
+  "id": "FC-sentinel-exit-code-misclassified-as-liveness",
+  "violated_contract": "A sentinel returned by a status API is not authoritative state when a completed operation may return the same value. On Windows, a process can terminate with exit code 259, which is also STILL_ACTIVE; treating GetExitCodeProcess alone as liveness leaves a stale single-flight owner classified as running forever.",
+  "canonical_prevention": "Determine process liveness from the process object's signaled state with a zero-timeout wait, then read creation time only for a process that is still nonsignaled. Keep a native Windows regression process that exits with the sentinel value and require the liveness probe to classify it as terminated.",
+  "canonical_prevention_artifact": [
+    ".github/scripts/test_preflight_runner.py"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    ".github/scripts/preflight_runner.py",
+    ".github/scripts/test_preflight_runner.py"
+  ],
+  "status": "open"
+}
--- a/.github/scripts/preflight_runner.py
+++ b/.github/scripts/preflight_runner.py
@@ -30,8 +30,10 @@ FINGERPRINT_ENV_NAMES = (
 )
 IS_WINDOWS = os.name == "nt"
 HAS_PROCESS_GROUPS = os.name != "nt" and hasattr(os, "killpg")
-WINDOWS_STILL_ACTIVE = 259
 WINDOWS_PROCESS_QUERY_LIMITED_INFORMATION = 0x1000
+WINDOWS_SYNCHRONIZE = 0x00100000
+WINDOWS_WAIT_OBJECT_0 = 0x00000000
+WINDOWS_WAIT_TIMEOUT = 0x00000102
 WINDOWS_PROCESS_SET_QUOTA = 0x0100
 WINDOWS_PROCESS_TERMINATE = 0x0001
 WINDOWS_ERROR_ACCESS_DENIED = 5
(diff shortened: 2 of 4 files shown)


Git-on-my-level:
Verified this end to end — it's a clean fix for a real Windows liveness bug, and the test relocation genuinely closes the gap where the exit-259 regression never executed in CI.

Per-file notes:

- `.github/scripts/preflight_runner.py` — the core change is sound. `windows_process_status()` now opens the handle with `PROCESS_QUERY_LIMITED_INFORMATION | SYNCHRONIZE` (0x00100000 is the correct flag value) and treats a zero-timeout `WaitForSingleObject` as the liveness source: `WAIT_OBJECT_0` → terminated, `WAIT_TIMEOUT` → alive, anything else → conservatively alive. That's the documented canonical pattern for the exit-259/`STILL_ACTIVE` ambiguity. Creation-time ticks are still read only on the alive path, so PID-reuse protection is preserved, and `GetProcessTimes` needs only `PROCESS_QUERY_LIMITED_INFORMATION`, which is still requested. The access-denied fallback and the `True, None` degraded-path returns are unchanged. `GetExitCodeProcess` and the ambiguous sentinel are gone entirely.
- `.github/scripts/test_pr_preflight.py` — right removal: `pr-preflight-contract-tests` runs where this test was permanently skipped, so it exercised nothing.
- `.github/scripts/test_preflight_runner.py` — the moved regression now lands in the `preflight-runner-portability` manifest check (`platforms: ["windows"]`), which `windows-preflight-portability.yml` runs on `windows-latest`; the workflow's path filter already includes this file, and the green "Native Windows process and launcher contracts" check on this PR confirms it executes. Asserting `child.wait() == 259` before probing liveness is a nice touch — it proves the child really terminated with the sentinel rather than a recycled handle.
- `.github/failure-classes/FC-sentinel-exit-code-misclassified-as-liveness.json` — conforms to the registry schema: id matches the filename, `canonical_prevention_artifact` points at the moved regression, scope hints cover the touched paths, and `evidence_prs: []` is accurate for a first occurrence.

No manifest edit was needed and I confirmed that: `preflight-runner-portability` already triggers on `test_preflight_runner.py`, and `pr-preflight-contract-tests` still triggers on the modified `test_pr_preflight.py`.

The PR-body verification narrative (3/3 repro on main, 26/26 after the fix, and separating the three host failures from the diff) is appreciated and matches what the diff implies.

Since this rewrites the liveness primitive behind the Windows pre-push single-flight gate and extends the failure-class registry, leaving the merge sign-off to a human maintainer.

---

kodjima33 (approved):
Confidence 5/5: reproduced exit-259 liveness bug 3/3 times on main, root cause with Win32 docs cited, scoped diff (51 lines), CI green, test moved into portability suite. No abuse signals. Merging.

PR #12348, opened 2026-08-28
fix(ci): retry Windows atomic status replacement

tianmind-studio:
## What changed and why

Retry preflight-runner status-file replacement for a short bounded interval when Windows reports a transient PermissionError. A joining runner polls status.json while the owner rewrites it; Windows readers do not share delete access by default, so the otherwise-atomic os.replace randomly aborted the owner.

The retry remains Windows-only, catches only PermissionError, waits 10 ms between attempts, and stops after 2 seconds. Other platforms and error classes keep their existing fail-fast behavior.

none

## How it was verified

- Reproduced SingleFlightTests.test_identical_processes_join_and_execute_once on an unmodified main-derived worktree: 4 failures in 5 runs, all WinError 5 at os.replace.
- Ran the same join regression after the fix: 5/5 passed.
- Ran .github/scripts/test_preflight_runner.py on native Windows: 28/28 passed.
- The new native test holds status.json open for reading and proves the writer retries and completes after release.
- Ruff, py_compile, failure-class validation, and git diff --check passed.
- Full diff-scoped Windows preflight passed its first six checks, including the 28-test portability suite, then stopped in the broader current-main contract suite only on missing local make, the exit-259 defect addressed independently by #12347, and the remaining Unicode checkout output defect.

## Tests

AtomicJsonTests covers the transient Windows retry, immediate non-Windows failure, and the real Windows file-sharing boundary with an open destination reader.

## Failure class (fixes)

Failure-Class: new

Violated contract: an atomic writer cannot assume POSIX rename sharing semantics on every supported host.

Canonical guard: keep write-to-temp-and-replace, but bound retries for Windows sharing violations and execute the open-reader boundary on native Windows CI.

Supporting evidence: the existing join-once regression failed 4/5 times on current main under native Windows with WinError 5, then passed 5/5 with this guard.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-atomic-replace-ignores-host-file-sharing.json
+++ b/.github/failure-classes/FC-atomic-replace-ignores-host-file-sharing.json
@@ -1,0 +1,15 @@
+{
+  "schema_version": 1,
+  "id": "FC-atomic-replace-ignores-host-file-sharing",
+  "violated_contract": "An atomic status writer must tolerate the file-sharing rules of every supported host. On Windows, a reader that has the destination open without delete sharing can briefly make os.replace fail with PermissionError; treating that transient sharing violation as a permanent write failure aborts the single-flight owner while another process is only polling its status.",
+  "canonical_prevention": "Keep the write-to-temp-and-replace primitive, but retry only Windows PermissionError failures for a short bounded interval so an active reader can release the destination. Preserve immediate failure for other platforms and errors, and run a native Windows regression that holds the destination open while the writer attempts the swap.",
+  "canonical_prevention_artifact": [
+    ".github/scripts/test_preflight_runner.py"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    ".github/scripts/preflight_runner.py",
+    ".github/scripts/test_preflight_runner.py"
+  ],
+  "status": "open"
+}
--- a/.github/scripts/preflight_runner.py
+++ b/.github/scripts/preflight_runner.py
@@ -19,6 +19,8 @@ from typing import TextIO
 
 POLL_SECONDS = 0.2
 STATUS_INTERVAL_SECONDS = 5.0
+WINDOWS_ATOMIC_REPLACE_RETRY_SECONDS = 0.01
+WINDOWS_ATOMIC_REPLACE_TIMEOUT_SECONDS = 2.0
 MAX_PR_BODY_FINGERPRINT_BYTES = 1024 * 1024
 FORWARDED_SIGNAL_NAMES = ("SIGINT", "SIGTERM", "SIGHUP", "SIGBREAK")
 FINGERPRINT_ENV_NAMES = (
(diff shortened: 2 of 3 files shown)


tianmind-studio:
Cross-PR Windows integration check (local, on current \origin/main\ \2f2be5a5f8\): I cherry-picked #12343, #12344, #12346, #12347, and #12348 in that order into one clean audit worktree (integration head \fbb5c2896\). The fixes compose without conflicts.\n\nValidation results:\n- The first seven preflight stages passed, including all 29 Windows portability tests.\n- The remaining Python/static checks passed when run explicitly: failure-class CLI, changelog/deferred-marker/async-blocker/datetime ratchets, route policy, plan catalog, Firestore coverage, lifecycle headers, version filenames, import purity, runtime image closure, module isolation, workflow contracts, conversation lifecycle, release policy, and candidate probes.\n- Two local prerequisite-only gaps remain on this machine: GNU \make\ is unavailable, and the release-boundary fixture needs \jq\. Those checks were not claimed as passed; no code/test failure was observed.\n\nThis is only integration evidence, not a request to merge the PRs as a unit; each remains independently scoped.

Git-on-my-level:
Thanks @tianmind-studio — reviewed head `660a449`; this is a well-scoped fix with unusually strong verification evidence.

What I verified against the diff:

- `.github/scripts/preflight_runner.py` — the delta is exactly the two constants (`WINDOWS_ATOMIC_REPLACE_RETRY_SECONDS = 0.01`, `WINDOWS_ATOMIC_REPLACE_TIMEOUT_SECONDS = 2.0`) and the bounded retry in `atomic_json()`. The loop catches only `PermissionError`, retries only on Windows (`not IS_WINDOWS → raise` keeps the POSIX fail-fast behavior), and re-raises unconditionally once the monotonic deadline passes, so a genuinely blocked destination still surfaces as a write failure. The failure mode is real: `join_existing()` polls `status.json` via `read_json()` every 0.2s while the owner rewrites it through `atomic_json()` (per-phase status writes plus the final `result.json` write), and a Windows reader holding the destination without delete sharing makes `os.replace` abort the single-flight owner spuriously. `time` and `IS_WINDOWS` were already in place, so no hidden dependency changes; I also confirmed the file is identical between `main` and the PR base (`2f2be5a`), so nothing else rides along.
- `.github/scripts/test_preflight_runner.py` — the new `AtomicJsonTests` class is well constructed: the mocked test pins both the retry (2 `replace` calls, one 10 ms sleep) and the non-Windows fail-fast (no retry, `PermissionError` propagates); the third test exercises the real sharing boundary with a reader thread holding the destination open, correctly guarded by `@unittest.skipUnless(os.name == "nt", ...)`. I ran the class on Linux: 2 pass, 1 skipped by design — consistent with your native-Windows 28/28 report.
- `.github/failure-classes/FC-atomic-replace-ignores-host-file-sharing.json` — schema-conformant with the existing registry (same key set, `schema_version` 1, id matches filename). `evidence_prs: []` is right for a class born in this PR under the registry's merged-PRs-only rule, and the `Failure-Class: new` body declaration matches the preflight contract.

One non-blocking observation: within the 2 s window the writer re-attempts every 10 ms, so a worst-case blocked destination costs ~200 replace attempts before raising. Harmless in practice and the constants are easy to tune — no change requested.

Since this is preflight/CI machinery and one piece of the Windows CI sequence (#12343–#12348), leaving it for human maintainer review before merge; the retry semantics and the new failure-class entry deserve a maintainer sign-off even though the code checks out.

---

kodjima33 (approved):
Confidence 5/5: reproduced 4/5 failures on main, root cause (Windows PermissionError on os.replace) stated, scoped diff (99 lines), CI green, adds regression test. No abuse signals. Merging.

PR #12351, opened 2026-08-28
fix(desktop): prevent double-localized SQL timestamps

### LINKED ISSUE #12350: Desktop SQL timezone guidance can double-shift results and local-day bounds
## Symptom

After #12349, Desktop Chat can apply the local timezone offset twice when the model follows its existing SQL guidance. `ChatPrompts.swift` still tells the model to use SQLite `localtime`, while `SQLQueryResultProjection` now treats any result named `timestamp` or camelCase `*At` as naive UTC.

For example, `SELECT datetime(timestamp, 'localtime') AS timestamp ...` turns `2026-08-27 19:59:51` UTC into `15:59:51` in America/New_York, then the result formatter parses that already-local value as UTC and renders `11:59:51 EDT`. I reproduced the same two-stage shift on Windows/Asia-Shanghai: SQLite returned `03:59:51`, which the formatter would shift again to `11:59:51 CST`.

The prompt's local-day boundaries are also ordered as `datetime('now', 'start of day', ..., 'localtime')`, which starts the UTC day before converting it. On UTC+8, the purported local-day boundary is 08:00 rather than midnight. The regression test merged in #12349 uses `TimeZone.current` and skips its raw-UTC assertion when the machine is UTC, so a UTC CI runner does not exercise either offset failure.

tianmind-studio:
## What changed and why

Fixes #12350. After #12349, Desktop Chat could apply the local offset twice when a SQL projection used SQLite `localtime`, while the prompt and daily recap still built local-day bounds in UTC-first order. This focused follow-up to the still-open macOS work in #12326 keeps projected timestamps raw until the result formatter, rejects query-layer `localtime` in `SELECT` expressions, and converts local-midnight boundaries back to UTC before comparing stored UTC values.

- INV-CHAT-1

## How it was verified

- Reproduced the failure with SQLite on Windows in `Asia/Shanghai`: the old boundary resolves to 08:00 local, while `localtime -> start of day -> utc` resolves to 16:00 UTC (local midnight).
- `git diff --check`
- Validated the changelog JSON with jq 1.8.2.
- Ran local prompt and projection-guard contract assertions, including acceptance of `localtime` in UTC boundary predicates and rejection in projected timestamp expressions.
- Full macOS Swift tests cannot run on this Windows host; `Desktop Swift Static & Test Contracts` is the authoritative compile/test gate.

## Tests

- `ChatPromptsTests.testDesktopChatSQLGuidanceLocalizesTimestampResultsExactlyOnce`
- `ChatPromptsTests.testDesktopChatSQLGuidanceComparesUTCColumnsToUTCBounds`
- `ChatPromptsTests.testDesktopChatSQLDayBoundsKeepLocalCalendarWindowsInUTC`
- `ChatToolExecutorSQLTests.testExecuteSQLRendersDatetimeColumnsInLocalTimeWithZoneLabel`
- `ChatToolExecutorSQLTests.testExecuteSQLRejectsProjectedLocaltimeButAllowsUTCBoundaryConversion`

## Failure class (fixes)

Failure-Class: FC-naive-utc-timestamp-in-llm-surface

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Chat/ChatPrompts.swift
+++ b/desktop/macos/Desktop/Sources/Chat/ChatPrompts.swift
@@ -87,25 +87,25 @@ struct ChatPrompts {
     -- Daily totals only (for counts/time breakdowns; use get_work_context for "what was I doing" — use -1 day for yesterday, -7 day for past week):
     -- Q1: App usage
     SELECT appName, COUNT(*) as count, ROUND(COUNT(*) * 10.0 / 60, 1) as minutes,
-    MIN(time(timestamp, 'localtime')) as first_seen, MAX(time(timestamp, 'localtime')) as last_seen
-    FROM screenshots WHERE timestamp >= datetime('now', 'start of day', '-1 day', 'localtime')
-    AND timestamp < datetime('now', 'start of day', 'localtime')
+    MIN(timestamp) AS firstSeenAt, MAX(timestamp) AS lastSeenAt
+    FROM screenshots WHERE timestamp >= datetime('now', 'localtime', 'start of day', '-1 day', 'utc')
+    AND timestamp < datetime('now', 'localtime', 'start of day', 'utc')
     AND appName IS NOT NULL AND appName != '' GROUP BY appName ORDER BY count DESC
     -- Q2: Conversations
     SELECT title, overview, emoji, startedAt, finishedAt,
     ROUND((julianday(finishedAt) - julianday(startedAt)) * 1440, 1) as duration_min
-    FROM transcription_sessions WHERE startedAt >= datetime('now', 'start of day', '-1 day', 'localtime')
-    AND startedAt < datetime('now', 'start of day', 'localtime') AND deleted = 0 AND discarded = 0
+    FROM transcription_sessions WHERE startedAt >= datetime('now', 'localtime', 'start of day', '-1 day', 'utc')
+    AND startedAt < datetime('now', 'localtime', 'start of day', 'utc') AND deleted = 0 AND discarded = 0
     ORDER BY startedAt DESC
     -- Q3: Tasks
     SELECT description, completed, priority FROM action_items
-    WHERE createdAt >= datetime('now', 'start of day', '-1 day', 'localtime')
-    AND createdAt < datetime('now', 'start of day', 'localtime') AND deleted = 0
+    WHERE createdAt >= datetime('now', 'localtime', 'start of day', '-1 day', 'utc')
+    AND createdAt < datetime('now', 'localtime', 'start of day', 'utc') AND deleted = 0
     ORDER BY createdAt DESC
 
     -- Bounded OCR preview for explicit low-level inspection only; recent-work retrieval belongs to get_work_context:
     SELECT timestamp, appName, windowTitle, substr(ocrText, 1, 200) as preview
-    FROM screenshots WHERE timestamp >= datetime('now', '-1 day', 'localtime')
+    FROM screenshots WHERE timestamp >= datetime('now', '-1 day')
     ORDER BY timestamp DESC LIMIT 20
 
     -- Active tasks:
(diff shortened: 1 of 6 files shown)


Git-on-my-level (approved):
Approved — clean, focused fix for the double-localization regression from #12349 (#12350), and it matches the merged design: the result formatter owns localization, so the query layer stays UTC.

Per-file notes:

- `desktop/macos/Desktop/Sources/Chat/ChatPrompts.swift` — examples now select raw `timestamp`/`*At` columns (`MIN(timestamp) AS firstSeenAt, MAX(timestamp) AS lastSeenAt`) and express local-day bounds as UTC instants (`datetime('now', 'localtime', 'start of day', '-1 day', 'utc')`). I verified the modifier order empirically against SQLite in Asia/Shanghai: the old form resolves yesterday's lower bound to 08:00 UTC while the new form gives 16:00 UTC — true local midnight. The old "convert them to {tz}" display guidance is correctly replaced by "quote execute_sql output as-is".
- `desktop/macos/Desktop/Sources/Chat/SQLQueryResultProjection.swift` — the new `projectsSQLiteLocalTime` guard returns an actionable error instead of letting a `datetime(col, 'localtime')` projection reach the formatter that would shift it a second time; `localtime` inside UTC WHERE-boundary computation is still allowed (covered by the new test). Minor non-blocking corner case: a projection like `SELECT datetime('now','localtime','start of day','utc') AS today_start` also trips the guard even though that value is already UTC — it degrades to a clear error message, fine to leave as-is for a conservative guard.
- `desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift` — `timeZone` is threaded explicitly through `executeSQL` → `executeSelectQuery` → the projection (default `.current` keeps existing call sites unchanged), and `DesktopChatSQLTime.localDayStartAsUTC/exclusiveEndAsUTC` replace the six hand-rolled `localtime` bounds in the daily recap queries. The recap's `first_seen`/`last_seen` stay SQL-localized and now carry an explicit zone label, so both paths localize exactly once.
- `desktop/macos/Desktop/Tests/ChatPromptsTests.swift` — pins the new guidance (raw projections, UTC-bound comparisons) and the `DesktopChatSQLTime` boundary strings, including the daysAgo=0 and negative-days clamping cases.
- `desktop/macos/Desktop/Tests/ChatToolExecutorSQLTests.swift` — the pinned `America/New_York` timezone makes rendering deterministic on any CI machine, and the new test exercises both the allowed boundary form and the rejected projection form through the real `executeSQL` path. Swapping the magic `Date(timeIntervalSince1970:)` for `ISO8601DateFormatter` parsing is a readability win.
- `desktop/macos/changelog/unreleased/20260828-chat-sql-timezone-boundaries.json` — accurate one-liner matching the unreleased-changelog convention.

Coordination note for maintainers: #12326 (SCA-364) is also open and touches the same five Swift files with the same UTC-boundary idiom, though it does not include this projection guard. Sequencing the two is a maintainer call; this PR is the focused fix for the live regression and is mergeable as-is against current main.

Thanks for the fast, well-reproduced report in #12350 and the tight follow-up.

---

kodjima33 (approved):
Legit fix(desktop) for double-localized SQL timestamps, CI green. Not merging: mergeStateStatus is DIRTY (needs author rebase before it can land).

kodjima33:
Thank you so much for this 🙏 — looks like this was already fixed on `main` (see commit 1711da1b85, "fix(desktop): render chat/SQL timestamps in the user timezone", merged 2026-08-30). Closing this one, but please keep them coming!

PR #12361, opened 2026-08-28
fix(desktop): let chat actually save the memory the user asked for

katipally:
## What

Three fixes to the desktop memory-save path. A memory the user asked chat to save could be refused outright, silently rejected, or saved but invisible. Each is a separate commit.

**`fix: hash authorized tool input in the kernel's JSON dialect`**

The runtime authorizes a tool call by hashing its input with `stableJsonStringify` (`agent/src/runtime/kernel-support.ts`); the app re-hashes with `JSONSerialization` and rejects the call when the two disagree. `JSONSerialization` escapes `/` as `\/` by default and `JSON.stringify` never does, so every authorized tool input carrying a URL, a path, or a term like `AI/ML` hashed differently and came back `authorized_execution_input_hash_mismatch`, while slash-free inputs went through. Observed live: "I like coffee" saved, every fact extracted from a resume was rejected.

**`feat: let chat save memories the model judges, and save many at once`**

Memory writes were gated by two regex validators duplicated in Swift and TypeScript: a phrase list deciding whether the turn "really" asked for a save, and a token-overlap check requiring the fact to appear in the typed sentence. Together they refused every save whose fact lived in an attachment or an earlier turn, and the model surfaced the rejection as invented magic phrases for the user to retype. Both are gone; the decision belongs to the model, stated in the tool description.

What still gates a write is structural, not lexical: typed main/floating chat only (voice, task chat, onboarding and delegated runs fail closed), the owner fence re-checked after the response, and bounded content.

Adds `create_memories(facts[])` for the case a single write cannot serve — a document, a profile, everything the user just shared — capped at 25 deduplicated facts. It writes each fact through `POST /v3/memories` rather than the batch route, which answers 503 in production for this write, and reports `saved_count` against `requested_count` so a partial save is stated rather than claimed as complete. Both writes send category `manual`, the wire marker for `user_asserted`, so a batch save carries the same authority in conflict resolution as the single one.

**`fix: surface memories written outside the memories page, and retry stranded ones`**

A chat-tool save and a proactive extraction both write straight to the backend, while the memories page renders its own SQLite projection and only re-read on owner change, app activation, or Cmd+R. A save made with the page open left the list unchanged, which reads as a failed save. Both writes now post `.memoriesDidChange` and the page routes it to the refresh it already had.

`MemoryAssistant`'s durability sequence was single-shot: when the backend create failed, the local row kept a NULL `backendId` and nothing ever retried it. The memory stayed on that Mac, absent from chat recall, the phone, and the knowledge graph. Each analysis tick now drains the backlog oldest-first through the same `createRemoteMemory` + `markLocalMemorySynced` operations the first attempt uses, before the extraction-enabled check, since a stranded row is a lost memory whether or not the user still wants new ones.

## Verification

Run on this branch, rebased on `upstream/main` at `fe3df5ac00`:

| Command | Result |
|---|---|
| `npx vitest run` (`desktop/macos/agent`) | 692 passed, 59 files |
| `xcrun swift test --filter 'AuthorizedToolExecutionTests\|ChatToolExecutorCreateMemoryTests\|MemoriesViewModelObserverTests\|MemoryAssistantTelemetryTests\|MemoryLocalIdentityMutationTests'` | 44 passed, 0 failures |
| `scripts/pr-preflight --base upstream/main` | 23 checks passed |

Exercised in the real app, not only in tests: saved a single fact from a typed sentence, saved a set of facts extracted from a resume attachment (the case that previously failed both the phrase gate and the hash check), and watched the Memories page update without a manual refresh. The slash case was confirmed against the two canonicalizers directly — `{"facts":["AI/ML Engineer"]}` hashed `sha256:0693096d…` (Swift) vs `sha256:e845bbeb…` (kernel) before the fix; they agree after it.

New tests: the kernel-hash fixture pins a slash-bearing input next to the existing nested-JSON fixture (which passed only because it holds no slash); `ChatToolExecutorCreateMemoryTests` covers the batch write, the dedup cap, and partial-save reporting; the backlog query, retry outcomes, and page subscriber each have a test.

- INV-AGENT-*
- INV-CHAT-1
- INV-MEM-1

## Failure class (fixes)

Failure-Class: FC-serialiser-dialect-mismatch-retypes-values

The violated contract is the class's own: a value serialized by one tool and re-serialized by another must go through the same dialect the producer emitted. Here the divergence fails closed (the call is rejected) rather than silently retyping a value, but the cause and the prevention are the class's — pin the reader to the producer's dialect and prove it with a fixture copied from real producer output, which is what the slash-bearing fixture does.

## Line-count exceptions

These three files were already over the ratchet before this branch; each grew by a handful of lines belonging to the fix, and splitting a 3,000-line file is a separate change from repairing the save path.

Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift | 3606 -> 3614 | +8 lines route the existing .memoriesDidChange notification to the refresh the page already had
Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift | 3439 -> 3447 | +8 lines register create_memories in MEMORY_WRITE_TOOLS and dispatch it
Line-Count-Exception: desktop/macos/Desktop/Sources/Rewind/Core/MemoryStorage.swift | 1837 -> 1856 | +19 lines add the oldest-first unsynced-memory backlog query the retry sweep drains

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12361?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Thanks @katipally — this is a careful, well-explained PR, and the commit separation makes it genuinely reviewable. I went through all three commits; here's where I landed.

**Commit 1 — canonical hash dialect (`AuthorizedToolExecution.inputHash`)**

Verified: I reproduced the expected digest in `testCanonicalInputHashMatchesKernelForSlashBearingInput` by canonicalizing the same object the way `stableJsonStringify` does (sorted keys, compact separators, no slash escaping) — it matches `sha256:e2fe63b2…ebeec` exactly. The `.withoutEscapingSlashes` fix is correct, and the doc comment pinning the two dialects together is exactly the kind of cross-language contract note this path needed. Real false-rejection bug, properly fixed.

**Commit 2 — model-judged saves + `create_memories`**

The mechanics are solid: `isTypedChatMemorySurface` keeps voice / task chat / onboarding / delegated scopes fail-closed, the owner fence is re-checked before and after each write (and per fact inside the batch loop), `memoriesCreationInput` trims, dedupes and bounds (25 × 1000 chars), and the partial-failure receipt reporting `saved_count` against `requested_count` is the right shape. `desktop-tool-policy.ts` correctly adds `create_memories` to `MEMORY_WRITE_TOOLS`, so it keeps the `desktop.memories.write` bundle requirement, with a test pinning that both tools classify as coordinator writes.

The part that needs a maintainer decision is the deliberate authorization-model change, not the code. Removing `isExplicitMemorySaveIntent` / `isMemoryContentGroundedInUserRequest` (Swift) and `hasExplicitMemorySaveIntent` / `hasMemoryContentGroundedInUserRequest` (TS, `external-surface-tool-policy.ts`) drops the last deterministic checks that (a) the current typed turn affirmatively asked for a save and (b) the saved fact was grounded in that turn's text. After this PR the only deterministic gates are surface-kind and owner; "did the user ask" lives in the tool description ("You decide whether the turn asked for a save"), i.e. in the model. Concretely: instructions injected via an attachment or page content can now drive a memory write that the old grounding check would have deterministically blocked — the doc comments being removed here describe exactly that fail-closed design. That may well be the right trade (the old validators clearly broke real saves, and the changelog copy is honest about the change), but it loosens a write gate on user-data persistence, so it should get an explicit maintainer product/security sign-off before merge rather than ride through on code quality.

**Commit 3 — page freshness + stranded-memory repair**

Nice work. `memoriesDidChange` (`AppState` → `MemoriesPage` subscriber → `refreshMemoriesIfNeeded`) with a yield-loop test instead of a sleep; `unsyncedLocalMemories` filters `deleted == false AND backendId IS NULL` oldest-first with the `limit > 0` guard; and the retry path threading `existingLocalID` through `MemoryAssistantDurabilityRequest` so a retry resumes at the backend half without re-inserting the local row or double-emitting the extraction terminal — all three have real tests behind them.

Two operational edge notes on `MemoryAssistant.retryUnsyncedMemories`, non-blocking:
- `break` on the first `.syncFailed` means one persistently-failing oldest record (e.g. a 4xx the backend will always return) head-of-line blocks every younger stranded memory. A skip-with-budget or per-row failure counter would keep the queue draining.
- If a record's remote create succeeds but the local mark-sync fails (`.syncStatePersistenceFailed`), the row keeps `backendId == nil` and the next pass re-sends it — a duplicate on the backend. Same hazard existed in the original pipeline; the retry pass just makes re-sends more likely.

**Smaller notes**

- `omi-tool-manifest.ts` `create_memories.runtimePreconditions` says "Writes through the batch memory endpoint in a single request", but the executor deliberately writes each fact through `POST /v3/memories` (the batch route 503s in production, per the code comment and the test asserting `url.path == "/v3/memories"` with two requests). The manifest text should match the implementation — it's agent-facing documentation.
- `memoriesCreationInput` caps at 25 and the receipt reports `requested_count` post-cap, so if the model submits 40 facts, 15 are silently dropped and the receipt still reads 25/25 saved. The schema says "at most 25", so this is edge-only, but a `truncated` flag would keep the model honest with the user.
- `GeneratedToolCapabilities.swift`'s `create_memories` entry has two near-duplicate bullet pairs (manifest `promptGuidelines` concatenated onto the `capabilityDoc` bullets). Generated-file cosmetics; worth deduping at the manifest layer if that's how it composes.

Checks are green on this head (Desktop Swift Build & Tests, Agent Runtime, Static & Test Contracts, contracts, desktop-core-e2e-t0). Labeling `security-review` + `needs-maintainer-review` for the authorization-model decision above — the code itself is in good shape, and the durability/retry work in particular is the kind of thinking this path needed. A maintainer should rule on model-judged memory saves (and the injection surface that comes with them) before this merges.

---

kodjima33 (approved):
Solid fix(desktop) — three real bugs in the memory-save path, well-tested. Not merging: mergeStateStatus is DIRTY (needs author rebase before it can land).

Git-on-my-level:
Follow-up on the force-push that came after @kodjima33's approval — I checked it, and it's a clean rebase with nothing new to review:

- The approved head `779cc53` and the current head `e181320d` carry byte-identical hand-written changes across all 24 files. The only content differences are the regenerated `manifestDigest` / `chatFirstManifestDigest` constants in `GeneratedToolExecutors.swift` and regenerated doc bullets in `GeneratedToolCapabilities.swift`, tracking manifest changes that landed on main in the meantime.
- The original first commit (the hash-dialect fix for `AuthorizedToolExecution.inputHash`) was dropped because main already has it via `b8c6e0c`, which ships the same `.withoutEscapingSlashes` canonicalization with its own regression test — the PR no longer touches that file. Right call.
- Checks are green on this head and the branch is now MERGEABLE/CLEAN, which is what the approval was waiting on.

Two carry-overs from the earlier automated review, both non-blocking and unchanged by the rebase:

- `omi-tool-manifest.ts` still describes `create_memories` as writing "through the batch memory endpoint in a single request", while the executor intentionally POSTs each fact to `/v3/memories` (the batch route 503s in production; `testBatchExecutorSavesEveryFactInOneOwnerBoundRequest` pins exactly those two `/v3/memories` posts). Worth a one-line manifest correction next time the file is touched.
- `MemoryAssistant.retryUnsyncedMemories` still stops at the first `.syncFailed` (a persistently failing oldest row head-of-line blocks younger stranded memories), and a remote create followed by a failed local mark-sync will re-send that row next pass (a backend duplicate). Edge cases; fine to land as-is.

Removed `needs-maintainer-review`: the model-judged-saves sign-off it was asking for has been given. Keeping `security-review`, since loosening the deterministic write gates on memory persistence is genuinely that kind of surface and the label is accurate triage for it.

---

Git-on-my-level:
Reviewed the rebased head (5846fb5) file by file. The design does what the description claims, and the substantive content matches what was approved earlier — I compared the pre-rebase and post-rebase commits and the two behavioral patches are identical, with the hash-dialect fix correctly dropped because it landed on main separately (b8c6e0c). Two notes and a security summary.

Checks: both red desktop jobs fail in HubEscalationTests.testPublicWebPromptIsSpeakableAndExcludesPrivateToolContext — the assertion expecting the old "Search the live public web before answering…" opening was already stale at this branch's merge base (the prompt code says "thoroughly before answering"), and main aligned the test in 718779056 after this branch's point. Neither failing check touches a file this PR modifies, and "Desktop Swift Build & Tests" itself died at the runner-setup stage with no test output. Rebasing onto latest main should clear both.

Changelog nit: 20260828-chat-memory-saves.json still lists "Fixed memory saves being rejected when the fact contained a slash" — that fix shipped separately in b8c6e0c, so consider dropping that bullet so the next release notes don't re-announce it.

What I verified in the code:
- ChatToolExecutor+MemoryCreation.swift — the Swift intent regex is gone; authorization is now structural: memorySurfaceRejection enforces main_chat/floating_chat with nil clientScope (voice/task/onboarding/delegated runs fail closed), the owner fence is re-checked before the write, after each write, and after errors, allowsAuthRetry is false, and content is bounded to 1–1000 chars. executeCreateMemories mirrors the single write's gates, caps at memoriesCreationMaxFacts = 25, dedupes via normalizeMemoryText, re-checks the owner per fact, stops at the first backend failure instead of retry-storming, and reports saved_count vs requested_count so a partial save is never presented as complete. Per-fact POST /v3/memories instead of the 503ing batch route is documented inline; I can't verify the backend behavior from this diff, only that the fallback is bounded.
- external-surface-tool-policy.ts / kernel-types.ts — the TS-side hasExplicitMemorySaveIntent and content-grounding checks (and their error codes) are removed cleanly; the Agent Runtime suite passes, so no dangling references remain. The manifest still does not project either memory tool to realtime_voice, task_chat, or leaf roles (covered by omi-tool-manifest.test.ts), so external surfaces cannot reach these tools.
- omi-tool-manifest.ts / GeneratedToolCapabilities.swift / GeneratedToolExecutors.swift / DesktopCapabilityRegistry.swift / tool-manifest.json fixture — create_memories is registered consistently across the TS manifest, the generated Swift enum/executor/routing, and the capability prose, with updated manifest digests; the routing hint now sends multi-item save requests to create_memories.
- desktop-tool-policy.ts — create_memories joins MEMORY_WRITE_TOOLS so the batch write requires the desktop.memories.write bundle; the updated test documents that an unlisted memory tool declares no bundle and is denied outright.
- MemoryAssistant.swift / MemoryAssistantTelemetry.swift / MemoryStorage.swift — retryUnsyncedMemories drains before analysis (independent of the extraction toggle, with a comment explaining why), unsyncedLocalMemories selects only non-deleted rows with nil backendId oldest-first, the retry reuses the existing local row via existingLocalID (no re-insert), skips the extraction terminal so telemetry isn't double-counted, and breaks the pass on syncFailed so the backlog survives to the next tick. Local rows are owner-scoped in practice because sign-out wipes local storage for the owner transition, and the retry re-checks RuntimeOwnerIdentity per record.
- AppState.swift / MemoriesPage.swift — the new memoriesDidChange notification is posted on MainActor from all three write sites (chat single, chat batch, proactive extraction/retry) and MemoriesViewModel subscribes with refreshMemoriesIfNeeded; the test yields the task queue instead of sleeping, which is the right pattern.
- Tests — ChatToolExecutorCreateMemoryTests drops the corpus-driven validator tests along with the validators and adds coverage for the new input bounds and receipts; MemoriesViewModelObserverTests, MemoryAssistantTelemetryTests (retry table incl. syncFailed and mark-failure), MemoryLocalIdentityMutationTests (deleted rows never re-sent, non-positive limit returns empty), desktop-tool-policy.test.ts, external-surface-authority.test.ts, and omi-tool-manifest.test.ts all cover the new paths meaningfully.

Security posture: this intentionally shifts save authorization from lexical regexes to model judgment, with typed-chat surface, owner fence, and input bounds as the remaining hard gates. The residual trade-off — a persuasively-worded attachment in a typed chat turn can now get facts persisted that the old token-overlap check would have refused — is the known cost of fixing the false-refusals, and the security-review label is the right tracking mechanism for it.

With the rebase (which also picks up the test alignment), I don't see remaining code concerns from my side; re-approval on the rebased head is the main outstanding step since the earlier approval is pinned to the pre-rebase commit.

---

Git-on-my-level:
Follow-up on the latest rebase (head `9db7438`) — status only; the substantive review of this content stands from my earlier comments.

What I verified on this head:
- Per-commit patch comparison against the previously reviewed head `5846fb5`: both behavioral commits ("feat: let chat save memories the model judges…", "fix: surface memories written outside the memories page…") carry identical hand-written changes file-for-file. The only content deltas are the regenerated `manifestDigest`/`chatFirstManifestDigest` constants in `GeneratedToolExecutors.swift` and regenerated prose context in `GeneratedToolCapabilities.swift` / `DesktopCapabilityRegistry.swift`, tracking changes that landed on main in the meantime. The changelog commit is byte-identical.
- Checks are green on this head — Desktop Swift Build & Tests, Agent Runtime, Desktop Swift Release Compile, Static & Test Contracts, Hygiene, contracts, and desktop-core-e2e-t0 all pass. The stale HubEscalation failures from the previous head are gone.

Two carry-over nits, both non-blocking and unchanged:
- `omi-tool-manifest.ts` still describes `create_memories` as writing "through the batch memory endpoint in a single request", while the executor deliberately POSTs each fact to `/v3/memories` (the batch route 503s in production; `testBatchExecutorSavesEveryFactInOneOwnerBoundRequest` pins exactly those two `/v3/memories` posts). One-line manifest fix next time the file is touched.
- `changelog/unreleased/20260828-chat-memory-saves.json` still lists the slash-bearing-input fix, which shipped separately in b8c6e0c ("fix(macos): preserve voice tool authorization hashes") after this PR dropped its copy of that change. Dropping the bullet would keep the next release notes from re-announcing it.

Remaining step before merge from my side: re-approval on the rebased head — @kodjima33's approval is pinned to `779cc53`, three rebases back. The `security-review` label stays for the authorization-model change already ruled on (model-judged saves replacing the lexical validators); it's accurate triage, not a new concern.

---
*by AI on behalf of David — re-approval on the rebased head is the remaining step before merge.*

aryanorastar:
Keeping this open on `9db743871a1d`.

DIRTY chat memory save; kodjima APPROVE ≠ merge; rebase; keep vs later #12436 extra files.

kodjima APPROVE is the size gate, not maintainer merge.

aryanorastar:
Keeping this open. Not approving `9db743871a1d`. origin/main `ChatToolExecutor.swift` still does not take this save path (`ChatToolExecutor+MemoryCreation.swift` is part of the SHA). Hygiene SUCCESS. mergeable CONFLICTING/DIRTY.

PR #12362, opened 2026-08-28
feat(windows): structure OCR screen context

### LINKED ISSUE #8453: Windows: SQLite OCR spatial row clustering and context serialization
## 1. The user experience
Omi needs to understand what you see on your screen to be a helpful companion. But uploading continuous screenshots is slow and consumes bandwidth. 

Since Omi on Windows is already running local OCR on screen frames, we can query this local SQLite database, clean up the text, reconstruct the layout spatially, and feed a clean Markdown outline to the LLM. This provides screen context awareness instantly and with zero network drag.

---

## 2. Technical architecture
To minimize context tokens, the screen parser extracts structural layouts from the local SQLite database, grouping elements relative to their positions on screen.

```mermaid
graph TD
    A[Raw SQLite OCR Log] -->|Filter Top/Bottom| B[Active Window Filtering]
    B -->|Row Clustering y < 10px| C[Vertical Alignment]
    C -->|x Coordinate Sort| D[Horizontal Reading Order]
    D -->|Levenshtein Check| E[Temporal Deduplication]
    E -->|Serialize| F[Structured Markdown Out]
```

### Affected files
*   Modify: `desktop/windows/src/main/ipc/db.ts`
*   Modify: `desktop/windows/src/main/ipc/screenSynth.ts`

---

### Step 1: Implement the database query
Create a function in the database helper module to query the rewind_frames table for the latest indexed frame within a 5-minute threshold. Exclude text blocks located in standard operating system menu bars (Y < 40px) or taskbar regions (Y > 1040px on a 1080p layout).

### Step 2: Implement vertical row clustering
Group the remaining blocks into rows using a 10px vertical threshold. Elements with vertical coordinate differences under 10px belong to the same row. Sort each row horizontally from left to right.

### Step 3: Format the structured Markdown
Format the sorted rows into a Markdown outline. Start with a header indicating the active application name and window title. List each row's content as a bullet point, separating horizontal elements with a custom separator (e.g. pipe or spacing).

### Step 4: Implement temporal deduplication
Inside the screen synthesis handler, compare the new Markdown string against the previous frame's Markdown. Calculate the Levenshtein similarity distance. If similarity exceeds 92%, discard the update to save context tokens.

---

## 4. Verification benchmarks
*   SQL query and clustering completes in under 15 milliseconds.
*   Output Markdown successfully removes menu bars and matches the layout reading order.

tianmind-studio:
## What changed and why

Closes #8453 with a narrow current-main implementation: read existing `ocr_lines_json` through a synthesis-only indexed query, remove top/bottom desktop chrome, cluster rows at the requested 10px threshold, preserve left-to-right order, and discard consecutive screen text only when Levenshtein similarity strictly exceeds 92%.

Raw OCR geometry stays in the Electron main process; IPC still returns `ScreenFrameLite`. This does not add a schema, shared DTO, renderer path, or dependency. The query preserves the existing watermark range instead of hard-clamping to five minutes because screen synthesis runs every ten minutes and a five-minute clamp would silently drop half of each batch. Existing prompt assembly already adds the application/window Markdown header.

none

## How it was verified

- `TZ=UTC pnpm test` -> 561 files passed, 5,553 tests passed, 22 skipped.
- Native `Asia/Shanghai` run -> 5,547 passed; the only failure is the pre-existing hard-coded Aug 24/25 date assertion in `insight/prompt.test.ts`, reproduced unchanged on clean `origin/main`.
- `pnpm typecheck` -> node and renderer typechecks passed.
- ESLint on all changed TypeScript files -> 0 errors, 0 warnings.
- `scripts/pr-preflight` -> 13 selected checks passed.
- Real `node:sqlite` benchmark over 100k rows -> query p95 0.62ms; 150-box spatial formatting p95 0.23ms (both below 15ms).
- 7,000 generated string pairs matched a reference Levenshtein implementation exactly; common 5,000-character near-duplicate comparison measured p95 0.15ms.

The packaged Electron app was not launched locally; dependencies were installed with `--ignore-scripts`, so validation used the production SQL constant, real in-memory SQLite, handler tests, and the full component suite.

## Tests

- `screenSynth.test.ts` covers chrome filtering, 10px clustering, x-ordering, malformed/incomplete-layout fallback, strict 92% behavior, context boundaries, and duplicate-tail watermark markers.
- `screenSynthSql.test.ts` executes the production query against SQLite and proves indexed/range/order behavior plus the main-process-only field boundary.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12362?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/windows/changelog/unreleased/2026-08-ocr-spatial-context.json
+++ b/desktop/windows/changelog/unreleased/2026-08-ocr-spatial-context.json
@@ -1,0 +1,5 @@
+{
+  "changes": [
+    "Screen memory extraction now preserves OCR reading order while removing repeated frames and desktop chrome."
+  ]
+}
--- a/desktop/windows/src/main/ipc/db.ts
+++ b/desktop/windows/src/main/ipc/db.ts
@@ -113,6 +113,7 @@ import {
   rewindSampleStep,
   buildRewindSampledSql
 } from './rewindSampleSql'
+import { SCREEN_SYNTH_FRAMES_SQL, type ScreenSynthFrameRow } from './screenSynthSql'
 import type {
   AiUserProfileInput,
   AiUserProfileRecord,
@@ -1559,6 +1560,13 @@ export function listRewindFrames(from: number, to: number): RewindFrame[] {
         get(),
         `SELECT ${REWIND_COLUMNS} FROM rewind_frames WHERE ts BETWEEN ? AND ? ORDER BY ts`
       ).all(from, to) as RewindFrame[]
+  )
+}
+
+export function listScreenSynthFrames(from: number, to: number): ScreenSynthFrameRow[] {
+  return timed(
+    'listScreenSynthFrames',
+    () => cachedStmt(get(), SCREEN_SYNTH_FRAMES_SQL).all(from, to) as ScreenSynthFrameRow[]
   )
 }
 
--- a/desktop/windows/src/main/ipc/screenSynth.test.ts
+++ b/desktop/windows/src/main/ipc/screenSynth.test.ts
@@ -1,0 +1,148 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import type { ScreenSynthFrameRow } from './screenSynthSql'
+
+const h = vi.hoisted(() => ({
+  handle: vi.fn(),
+  listFrames: vi.fn(),
+  getState: vi.fn(),
+  updateState: vi.fn(),
+  advanceWatermark: vi.fn(),
+  recordRun: vi.fn()
+}))
+
+vi.mock('electron', () => ({ ipcMain: { handle: h.handle } }))
+vi.mock('./db', () => ({ listScreenSynthFrames: h.listFrames }))
+vi.mock('../screenSynth/state', () => ({
+  getScreenSynthState: h.getState,
+  updateScreenSynthState: h.updateState,
+  advanceWatermark: h.advanceWatermark,
+  recordRun: h.recordRun
+}))
+
+import {
+  buildSpatialOcrText,
+  isNearDuplicateScreenText,
+  registerScreenSynthHandlers
+} from './screenSynth'
+
+function frame(overrides: Partial<ScreenSynthFrameRow> = {}): ScreenSynthFrameRow {
+  return {
+    ts: 1,
+    app: 'Code',
+    windowTitle: 'plan.md',
+    processName: 'code.exe',
+    ocrText: 'File Edit quarterly plan Save',
+    ocrLinesJson: JSON.stringify([
+      { text: 'File', x: 10, y: 20, w: 20, h: 10, confidence: 1 },
+      { text: 'quarterly plan', x: 100, y: 200, w: 100, h: 20, confidence: 1 },
+      { text: 'Edit', x: 20, y: 205, w: 40, h: 20, confidence: 1 },
+      { text: 'Save', x: 10, y: 1050, w: 40, h: 10, confidence: 1 }
+    ]),
+    height: 1080,
+    ...overrides
+  }
+}
+
+beforeEach(() => vi.clearAllMocks())
+
+describe('spatial OCR synthesis', () => {
+  it('filters screen chrome, clusters within 10px, and orders each Markdown row left-to-right', () => {
+    expect(buildSpatialOcrText(frame())).toBe('- Edit | quarterly plan')
+  })
+
+  it('falls back for malformed or incomplete geometry without reintroducing chrome-only layouts', () => {
+    expect(buildSpatialOcrText(frame({ ocrLinesJson: '{bad json' }))).toBe(
+      'File Edit quarterly plan Save'
+    )
+    expect(
+      buildSpatialOcrText(
+        frame({
+          ocrText: 'one two three four five six seven eight nine ten',
+          ocrLinesJson: JSON.stringify([
+            { text: 'one', x: 10, y: 200, w: 20, h: 10, confidence: 1 }
+          ])
+        })
+      )
+    ).toBe('one two three four five six seven eight nine ten')
+    expect(
+      buildSpatialOcrText(
+        frame({
+          ocrText: 'alpha beta',
+          ocrLinesJson: JSON.stringify([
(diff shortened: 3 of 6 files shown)


Git-on-my-level (changes requested):
Thanks for this — the overall shape is right and several pieces are genuinely well done: the banded Levenshtein in `screenSynth.ts` (`isWithinLevenshteinDistance`, with prefix/suffix trimming and the character-count lower bound) avoids allocating full matrices on 5k-character strings, the synthesis-only query in `screenSynthSql.ts` keeps `ocr_lines_json` out of the public Rewind DTO while its comment documents exactly why, `db.ts`'s new `listScreenSynthFrames` reuses `cachedStmt`/`timed` like its neighbours, and the changelog fragment matches the user-visible behavior. The watermark reasoning — keeping blank duplicate markers so the renderer's `maxTs` still advances past deduped tails — checks out against `renderer/src/lib/screenGrouping.ts`, which skips blank frames.

Requesting changes for one blocking issue: **the spatial code assumes OCR line coordinates are in pixels, but the production writer stores them normalized to 0..1.**

- `screenSynth.ts`: `ROW_CLUSTER_THRESHOLD_PX = 10`, `CHROME_EDGE_PX_AT_1080P = 40`, and the content filter `line.y >= edge && line.y <= effectiveHeight - edge` all compare `line.y` against pixel magnitudes.
- But `src/main/ocr/win-ocr-helper/Program.cs` persists each line as `x = minX / width`, `y = minY / height`, `w = (maxX - minX) / width`, `h = (maxY - minY) / height` — normalized 0..1. The only writer (`rewind/ocrPersist.ts` → `setRewindFrameOcr` in `db.ts`) stores that JSON verbatim, and the existing consumer `normalizedBoxToRect` in `renderer/src/lib/rewindOverlay.ts` multiplies by the container rect — confirming the convention.

Consequence: for every real Windows frame all line `y` values are ≤ 1, so `line.y >= 40` filters out **every** line, `contentLines` is always empty, and `buildSpatialOcrText` returns `''`. I confirmed this by running this PR's actual function on a fixture built exactly as the helper emits it (1080p frame: menu bar at y=20px → 0.0185, content at y=540px → 0.5): the output is the empty string, while the same layout expressed in pixels produces the expected `- Edit | quarterly plan`. Net effect on Windows: screen synthesis silently goes from today's plain-OCR-text behavior to *nothing at all* — and because `groupFrames` drops blank frames while the watermark still advances, the failure is invisible: runs succeed, memories just stop being written.

The tests in `screenSynth.test.ts` pass because their fixtures (`y: 200`, `y: 1050`, …) use pixel coordinates that no production writer ever stores — re-deriving them from the helper's actual output shape would have caught this. `screenSynthSql.test.ts` is fine as-is: it runs the real SQL against `node:sqlite` and correctly asserts `imagePath` stays out of the row.

Suggested fix (either works):
1. Select `width` alongside `height` in `screenSynthSql.ts` and denormalize (`px = normalized * dimension`) before clustering/chrome filtering, or
2. Convert the thresholds to normalized units (`edge = 40 / frameHeightPx`, cluster `10 / frameHeightPx`) — note this makes the row threshold height-dependent, worth a comment.

Once fixed, please also add one test whose fixture is generated from the same JSON shape `Program.cs` emits (or extracted from a real captured frame) so the coordinate convention can't drift again. The PR description is commendably honest that the packaged app wasn't launched; per this repo's Definition of Done ("you exercised the change yourself — ran the real user-facing path"), one run against real captured frames on a Windows machine is the other half of closing that gap, since the feature's whole value depends on interpreting the stored geometry correctly.

Everything else reads correctly to me: the strict >92% Levenshtein gate (an exact-8% edit distance is not a duplicate), the plain-text fallback when stored geometry is missing/incomplete or coverage mismatches, and `screenSynth:framesSince` still returning `ScreenFrameLite` with no image bytes crossing the IPC boundary.

Happy to re-review once the coordinate space is fixed. Leaving the merge/product call with the maintainers, since this opens a new screen-context surface (what Rewind OCR data reaches synthesis prompts).

---

aryanorastar:
Keeping this open. Not approving `0aa0ca56ad53`. origin/main has no `screenSynth.ts`. Six files, +491. Hygiene SUCCESS. An older preflight is cancelled; a later one succeeded. mergeable MERGEABLE/BLOCKED.

tianmind-studio:
Thanks @Git-on-my-level — addressed the coordinate-space issue.

**Fix (option 1):** `screenSynthSql` now selects `width` alongside `height`, and `buildSpatialOcrText` denormalizes stored 0..1 boxes to pixels (`px = normalized * dimension`) before chrome filtering / row clustering. Pixel thresholds stay meaningful.

**Test:** added a production-shaped fixture matching `Program.cs` (1080p: menu bar y=20px → ~0.0185, content y=540px → 0.5) that asserts spatial text is produced (`- Edit | quarterly plan`), not empty. Existing fixtures now go through the same normalized helper.

Head: `f55e334192` (denormalize in `bb78c94cc0` + BOM cleanup). Focused `screenSynth` / `screenSynthSql` tests: 5/5 passed.

@Git-on-my-level please re-review when you have a moment — happy to follow up on anything remaining.

tianmind-studio:
Addressed the coordinate-space review on `bb78c94`:

- `screenSynthSql.ts` now selects `width` alongside `height`
- `buildSpatialOcrText` denormalizes Program.cs `0..1` boxes to pixels before chrome filtering / row clustering
- Added a production-shaped fixture test (1080p menu at ~y=0.0185, content at y=0.5) so this cannot silently regress back to empty synthesis
- `screenSynth.test.ts` + `screenSynthSql.test.ts`: 5/5 passing locally

@Git-on-my-level ready for re-review when you have a moment.

tianmind-studio:
Follow-up 1c0724c: removed the trailing blank line that failed diff-hygiene on screenSynth.ts.

PR #12370, opened 2026-08-28
fix(app): add "Add to Tasks" to the mobile conversation summary

### LINKED ISSUE #12368: Desktop-recorded conversation suggestions are not visible on mobile
## Product impact

**Primary loop layer:** Retrieval-Action

**Data loss / failed capture / broken sync:** No

## Describe the bug

On Mobile (Android/iOS), action items extracted from conversations are created as
pending task Candidates, but there is no user-facing way to accept/add them to
the task list. As a result, the suggestions remain invisible to the user and
expire after 2 days.

## To Reproduce

1. Record a conversation containing an actionable item using the Omi Mobile app.
2. Open the conversation after it has been processed.
3. View the extracted action item.
4. Try to add/accept the action item as a task.
5. There is no option to add it to the task list.

## Current behavior

Mobile displays the extracted action item with a completion checkbox, but provides
no option to accept/promote it to a task. The item remains a pending Candidate
and eventually expires after 2 days.

## Expected behavior

The user should be able to explicitly accept an extracted action item and add it
to their task list, consistent with the existing Desktop task-suggestion flow.

## Screenshots

**Before**

Image
Image

**After Fix**

Image
Image
Image

## User ID

Developer

## Smartphone + device

- Device: Mototrola edge 70
- OS: Android 16
- Browser: N/A (Omi Mobile app)
- App Version: Dev Build
- Device version: -

## Additional context

This appears to be a Mobile/Desktop parity gap rather than a new feature request.
Commit `7a81a69438` introduced suggestion-only task capture and added an explicit
"Add to Tasks" flow on Desktop, while Mobile has no equivalent surface. This
also conflicts with locked invariant `INV-TASK-2`, which explicitly includes
Suggested-task projections on Desktop AND MOBILE.

On my fork running the same code, conversations produced pending Candidates while
no corresponding entries appeared in `action_items`, leaving the extracted
action items effectively invisible to the user.

I have a fix prepared and will submit a PR referencing this issue.

hellKing-Diablo:
## What changed and why

Fixes #12368

Automatic task capture is suggestion-only (INV-TASK-2, #11974): an extracted action item becomes a pending Candidate that reaches the task list only through an explicit user gesture. #11974 built that gesture for macOS (conversation summary, Suggested section, chat card) but not for mobile, so on a phone an extracted item has no path to the task list at all — the Candidate stays pending, the Action Items page never shows it, and it expires after `SUGGESTION_TTL` (2 days).

This adds the mobile counterpart of the macOS conversation-summary button: each incomplete action item in `ActionItemDetailWidget` now carries an "Add to Tasks" control.

Manual create is outside the capture-policy boundary — INV-TASK-2 scopes out "chat, MCP, developer API and manual create" because they carry a real user gesture and write directly by design — so the Candidate lifecycle is untouched. Like the macOS summary button, this does not resolve the twin pending Candidate, which still expires on its own; resolving it needs a mobile Suggested surface reading `/v1/candidates`, left for follow-up. Happy to take direction if you'd rather see that surface first.

none — `pr_preflight --suggest` reports no invariant path globs touched. INV-TASK-2 is cited above as motivation; its globs are all backend and desktop.

## How it was verified

```
flutter analyze <changed files> + lib/l10n/   -> No issues found
flutter test test/widgets/add_to_tasks_button_test.dart  -> 3 passed
pr_preflight.py --lane local                  -> 13 checks passed
```

Exercised on an Android device against a live backend: recorded a conversation, tapped the control on an extracted item, confirmed it appeared on the Action Items page.

## Tests

`app/test/widgets/add_to_tasks_button_test.dart` — core path: an added item confirms and withdraws the control. Error path: a create in flight refuses a second tap, so an in-flight create cannot be duplicated. A failed create returns to idle so the user can retry.

## Failure class (fixes)

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/l10n/app_en.arb
+++ b/app/lib/l10n/app_en.arb
@@ -3026,6 +3026,18 @@
     "@searchResults": {
         "description": "Badge label for search results section"
     },
+    "addToTasks": "Add to Tasks",
+    "@addToTasks": {
+        "description": "Button on a conversation's action item that adds it to the task list"
+    },
+    "addedToTasks": "Added",
+    "@addedToTasks": {
+        "description": "Confirmation shown after an action item has been added to the task list"
+    },
+    "addToTasksFailed": "Couldn't add to tasks. Try again.",
+    "@addToTasksFailed": {
+        "description": "Error shown when adding an action item to the task list fails"
+    },
     "actionItems": "Action Items",
     "@actionItems": {
         "description": "Section header for action items"
--- a/app/lib/l10n/app_localizations.dart
+++ b/app/lib/l10n/app_localizations.dart
@@ -5193,6 +5193,24 @@ abstract class AppLocalizations {
   /// **'Search results'**
   String get searchResults;
 
+  /// Button on a conversation's action item that adds it to the task list
+  ///
+  /// In en, this message translates to:
+  /// **'Add to Tasks'**
+  String get addToTasks;
+
+  /// Confirmation shown after an action item has been added to the task list
+  ///
+  /// In en, this message translates to:
+  /// **'Added'**
+  String get addedToTasks;
+
+  /// Error shown when adding an action item to the task list fails
+  ///
+  /// In en, this message translates to:
+  /// **'Couldn\'t add to tasks. Try again.'**
+  String get addToTasksFailed;
+
   /// Section header for action items
   ///
   /// In en, this message translates to:
--- a/app/lib/l10n/app_localizations_ar.dart
+++ b/app/lib/l10n/app_localizations_ar.dart
@@ -1,5 +1,6 @@
 // ignore: unused_import
 import 'package:intl/intl.dart' as intl;
+
 import 'app_localizations.dart';
 
 // ignore_for_file: type=lint
@@ -2659,6 +2660,15 @@ class AppLocalizationsAr extends AppLocalizations {
 
   @override
   String get searchResults => 'نتائج البحث';
+
+  @override
+  String get addToTasks => 'Add to Tasks';
+
+  @override
+  String get addedToTasks => 'Added';
+
+  @override
+  String get addToTasksFailed => 'Couldn\'t add to tasks. Try again.';
 
   @override
   String get actionItems => 'عناصر العمل';
--- a/app/lib/l10n/app_localizations_be.dart
+++ b/app/lib/l10n/app_localizations_be.dart
@@ -1,5 +1,6 @@
 // ignore: unused_import
 import 'package:intl/intl.dart' as intl;
+
 import 'app_localizations.dart';
 
 // ignore_for_file: type=lint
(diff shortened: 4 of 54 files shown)


Git-on-my-level:
Thanks @hellKing-Diablo — this is a well-scoped fix for #12368, and the description is unusually precise about the invariant it operates under.

What I verified:

- `app/lib/pages/conversation_detail/widgets/add_to_tasks_button.dart` — the `AddToTasksState` machine (idle -> adding -> added) is correct: `add()` is a no-op unless idle, `mounted` is checked after each await, a provider refresh failure after a successful create is logged rather than undoing the added state, and a failed create returns to idle with the `addToTasksFailed` snackbar. The repeat-tap concern is covered by the endpoint's content idempotency, which the doc comment cites accurately.
- `app/lib/pages/conversation_detail/page.dart` — the `if (!isCompleted)` guard placement inside `ActionItemDetailWidget`'s row is right: completed items keep their strike-through layout, incomplete ones get the gesture, and `conversationId` flows from `provider.conversation.id` exactly as the two list call sites already pass it.
- `app/lib/l10n/app_en.arb` + `app_localizations.dart` — three new keys with descriptions; the regenerated per-locale files carry the English fallback, which is the expected gen-l10n output when only the template ARB changes.
- `app/test/widgets/add_to_tasks_button_test.dart` — covers idle, added, and in-flight states, including that `add()` is a no-op mid-flight.

Two non-blocking notes:

1. The failure path (`created == null` -> back to idle + snackbar) is described in the test file's header comment but not asserted anywhere. A small test with a stubbed `createActionItem` for that return-to-idle transition would complete the core/error path pair the app's Definition of Done asks for.
2. The `added` state is per-widget and resets when the page rebuilds. You document that endpoint idempotency makes a repeat tap harmless — agreed, and the macOS button keeps the same session-local behavior, so this is consistent rather than wrong.

One thing needs a maintainer's product call before merge: this mirrors the macOS conversation-summary gesture, including its limitation — the twin pending Candidate is not resolved and still expires on its own, and mobile still has no Suggested surface reading `/v1/candidates` (you flag this yourself). Whether the per-item button here is the right mobile gesture, or whether the Suggested surface should come first so acceptance resolves Candidates, is a direction decision rather than a code-correctness one. Leaving that with the maintainers.

---

hellKing-Diablo:
acknowledging #12424, note that its premise is resolved

PR #12377, opened 2026-08-29
feat(desktop): user MCP servers and skills, fully local under ~/.omi

katipally:
## What

User-added MCP servers and catalog skills now work the way the protocol and the
skill format actually specify, and the marketplace lists only the official MCP
registry.

**Marketplace**
- MCP servers come only from `registry.modelcontextprotocol.io`. The Smithery
  feed is gone: an entry there installs a Smithery-hosted endpoint behind a
  Smithery consent screen, so the user signs into a broker before the vendor's
  own OAuth ever runs, and when that second hop does not complete there is
  nothing in this app to retry. `ai.smithery/*` is excluded from search for the
  same reason. Search now ranks DNS-verified brand namespaces above
  `io.github.<user>`, which proves only that someone holds a GitHub account.
- A registry package's named arguments are replayed as published. `io.snyk/mcp`
  was installing without the `-t stdio` its record declares.

**MCP client**
- HTTP+SSE is a real transport now. `mcp.json` and the catalog already recorded
  `transport: "sse"`; nothing acted on it, so those servers were POSTed to their
  event-stream URL, 404'd, and were skipped at session start.
- A server's published prompts register as callable tools. This chat has no
  slash-command surface, so they take the tool path: same call, same audit
  trail. `initialize` capabilities are recorded, so a server that publishes no
  prompts is never asked for them.
- `tools/list` and `prompts/list` are walked to the last page. `nextCursor` was
  dropped, so a server past its page size silently lost the rest.
- Non-text content blocks are named, not inlined. A result with no text block
  used to fall through to `JSON.stringify(result)`, putting the whole base64
  payload into the conversation. `structuredContent` answers when a tool
  returns data and no prose.

**Skills**
- A catalog skill installs as the folder it is. 22 of the 33 skills in the two
  default catalogs ship `scripts/`, `references/` or assets their SKILL.md
  references by relative path; only the Markdown was fetched, so every one of
  them installed as instructions pointing at files that were not there.

## Verification

Ran, with output:

- `xcrun swift build -c release` — clean. This lane, not the debug one, is
  what catches strict-concurrency defects; it is now part of how this branch is
  verified locally.
- `tests/test-feature-sentinel-negative-control.sh` — 4 passed, 0 failed.
- `xcrun swift test --package-path Desktop` — all suites pass except
  `ChatDiscoverabilityTests.testDesktopCapabilitiesExistInAgentToolDeclarations`
  ("Missing agent tool declaration for look_at_frame"), which fails identically
  on `origin/main` in a clean worktree. Not from this diff.
- `agent`: 726 vitest tests, `tsc --noEmit` clean. `pi-mono-extension`: 118 tests.
- **Conformance against the reference server** (`@modelcontextprotocol/server-everything`),
  driven through the real `registerUserMcpTools` entry point with a temp
  `OMI_LOCAL_MCP_FILE`:
  - stdio: 13 tools + 4 prompts registered; `echo` returns `"Echo: hi"`.
  - sse (`server-everything sse`, port 3001): 13 tools + 4 prompts; `echo`
    returns `"Echo: over sse"`; `get-structured-content` returns
    `{"temperature":36,...}`. `POST http://127.0.0.1:3001/sse` — what the old
    client did — is `404` against that same running server.
  - `get-tiny-image` returns 94 characters
    (`Here's the image you requested:\n[image: image/png, not shown]\nThe image
    above is the MCP logo.`) with no base64 in it.
  - `get-resource-links` names each link with its URI; `get-resource-reference`
    inlines the embedded resource's text. Both were dropped entirely before.
- Live Streamable HTTP: `docs.mcp.cloudflare.com/mcp` declares a prompts
  capability and publishes `workers-prompt-full`, which this client expands and
  a tools-only client never saw; its tool results come back as 11 KB of
  readable text.
- Named bundle (`omi-official-mcp`), real user-facing path: marketplace lists
  only official-registry brands; installing Notion wrote
  `https://mcp.notion.com/mcp` with Notion's own sign-in; installing `docx` from
  the live catalog wrote 61 files across 6 levels of nesting with no staging
  directory left behind.

## Review feedback (#12377 review of Aug 29)

The blocking item — `McpStdioClient` unrefing every handle so the event loop
drained mid-handshake — is fixed in `9ef1067` on this branch, with the
regression test that was asked for: `agent/tests/mcp-stdio-client.test.ts`
drives a server that replies on a later tick and fails without the fix. Only
`stdin`/`stderr` are unref'd now; `holdEventLoop` refs the child and its stdout
for as long as an RPC is in flight. The three lanes cited as red are green.

All four non-blocking notes are addressed:

- **PKCE/state randomness.** `Int.random` fed both the verifier and `state`.
  Both now come from `SecRandomCopyBytes`, base64url-encoded per RFC 7636 4.1,
  and a failed draw aborts the flow rather than falling back. Encoding raw
  bytes also removes the modulo bias in indexing a 66-character alphabet.
- **Authorization round trip.** This was a real defect, not only redundancy:
  stripping `Bearer ` and rebuilding it corrupted every other scheme, so a
  hand-written `Authorization: Basic <...>` went out as `Bearer Basic <...>`.
  Clients take the configured headers verbatim now, with a test.
- **Doc overclaim on token refresh.** Correct — the `Task { }` is unawaited and
  races the spawn. The docs now say so, and say why it is deliberate (awaiting
  it would put a network round trip in front of every launch) and why it is
  safe (fail-open; the runtime re-reads mcp.json per session). The misleading
  code comment is corrected too.
- **Plaintext tokens.** The docs now state plainly that OAuth tokens live in
  plaintext in ~/.omi/mcp.json and what that implies.

Separately, CI caught a defect that no debug build can see: `ExtensionLogoLoader`
returned a non-Sendable `NSImage?` out of an actor, which release compile and
the Swift 6 sentinel target both reject. The actor caches bytes now and the view
decodes them, keeping every NSImage on the main actor. The cache is also bounded
(FC-unbounded-module-cache) — it was keyed by URL with no eviction.

## Review pass

An adversarial review of this branch's own diff found defects that are now
fixed in it, each with a regression test:

- SSE frames were split on `"\n\n"`. A CRLF server's separator is `"\r\n\r\n"`,
  which contains no `"\n\n"`, so no frame ever parsed and a working server read
  as one that never opened a stream. Line endings are normalized on the way in.
- `probeSSE` read the stream until the endpoint event or the request timeout,
  which measures the gap between packets — a server sending keepalives first
  reset it forever, holding the badge on "Checking…" and the connection open.
- Installing over an existing skill deleted the destination before moving the
  staged folder in, so a throw between the two lost both versions.
  `replaceItemAt` makes it atomic; staging moved out of `~/.omi/skills` so a
  crash cannot leave a directory the catalog lists as a skill.
- A write to a dead stdio child's stdin emits EPIPE on the stream with no
  listener, which takes down the agent process and every session with it.
- Tool and prompt calls had no ceiling; only discovery did.
- `getPrompt` used a text-only reader while tool results used the block
  renderer, so a prompt carrying a resource lost it — the same failure class
  this PR registers, found in its own diff.
- Registered names truncate at 64 characters; two long tool names on one server
  collided, and the second `registerTool` would reject the whole extension.

## Not addressed

- `notifications/tools/list_changed`. Acting on it means replacing a live tool
  set and pi's extension API has `registerTool` with no counterpart to remove or
  replace one. Config and tools are read per session, so a changed server is
  picked up on the next session.
- Elicitation, sampling and roots. The client declares no capabilities of its
  own; these need a UI and a policy decision, not just wire support.
- The ACP lane passes `type: "sse"` through to its peer as it always has; the
  new SSE client serves the pi-mono lane.

## Product invariants

- INV-AGENT-*
- INV-AUTH-1
- INV-CHAT-1

Failure-Class: new

New class `FC-external-protocol-subset-read-as-absent`. The violated contract: a hand-written client of an external protocol implements
a subset of it and treats the unimplemented parts as absent rather than as
unsupported. Every omission here was silent — no decode error, no unhandled
case, no failing test, because the client stayed internally consistent. A server
with a second page of tools read as a server with fewer tools; an SSE server
read as an unreachable server; a screenshot tool read as a tool that answers in
base64. This is the external-source-of-truth variant of
`FC-mirrored-model-omits-new-member`, but its canonical prevention (derive the
mirror from the generated source) cannot apply: there is nothing in-tree to
derive from. The guard surface that does work is what this PR lands — a shared
`McpClient` base so the wire subset exists in exactly one place, an exhaustive
content-block switch that names what it did not understand instead of dropping
it, and conformance tests driven against the reference implementation rather
than against our own fixtures.

Line-Count-Exception: desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift | 4365 -> 4386 | pre-existing on this branch: local MCP file env wiring for the user-extensions feature
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/AppsPage.swift | 3808 -> 3880 | pre-existing on this branch: Apps-page section navigation and catalog sheets for the user-extensions feature
Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatProvider.swift | 7212 -> 7290 | pre-existing on this branch: catalog fetch/install state for the user-extensions feature


Git-on-my-level (dismissed):
Thanks @katipally — this is a genuinely well-constructed feature, and I want to credit the parts that are right before the blocker: fail-open throughout (`loadLocalMcpConfig` skipping invalid/duplicate/reserved names, `registerUserMcpTools` skipping unreachable servers without failing the session), read-modify-write in `LocalMcpStore.swift` that preserves hand-edited `mcp.json` entries the UI doesn't understand (with a round-trip test in `LocalMcpStoreTests.swift`), OAuth done properly in `LocalMcpStore+OAuth.swift` (PKCE S256, state check, RFC 8252 loopback redirect, dynamic client registration as a public client), tidy frontmatter normalization in `LocalSkillsStore.swift` that owns `name`/`description` while preserving author keys, the notification-driven catalog refresh in `ChatProvider.swift`, the `_meta` merge in `acp.ts` that preserves the existing systemPrompt behavior, per-session reload in `index.ts`, the 5s/30s discovery caps in `pi-mono-extension/index.ts`, plus `AgentExtensionsSections.swift`, the `AppsPage.swift` sheets, `AppProvider.fetchUserExtensions()`, `node-tools.ts` skill-root precedence, `mcp-http-client.ts` SSE/JSON handling, the vitest suite in `agent/tests/user-extensions.test.ts`, the manual e2e flow, and the changelog entry. Requesting changes for one concrete blocker.

**Blocking: `McpStdioClient` unrefs every handle, so the event loop can drain mid-handshake — this is what's red in CI**

All three failing checks (Hygiene, Agent Runtime, PR Metadata Preflight) fail on the same thing: `desktop/macos/pi-mono-extension/index.test.ts` "registerUserMcpTools: local stdio server from ~/.omi-style mcp.json registers and executes" is cancelled with `Promise resolution is still pending but the event loop has already resolved` (113 pass, 1 cancelled — identically in all three lanes).

The cause is in `desktop/macos/agent/src/runtime/mcp-stdio-client.ts`, not the test: `start()` calls `child.unref()` and then also unrefs `child.stdin`, `child.stdout`, and `child.stderr`. With every handle unreffed, Node treats the event loop as empty once the synchronous setup finishes, so the pending `initialize` response never lands. I reproduced this with no test runner involved (a plain top-level await plus a copy of the client's process management on Node 22) and it fails the same way, so this is not harness-specific. It passes on a fast local machine because leftover jobs keep the loop alive just long enough — which matches your green local `npm test` run against the red CI lanes.

Suggested fix: keep `child.unref()` if the goal is that a server must not hold the runtime open, but stop unrefing the stdio streams while a session is live (or ref them for the handshake and unref on dispose). The `process.once("exit", …)` SIGTERM and the `withTimeout` discovery cap are fine as-is. A small regression test that fails without the fix would lock this in.

**Smaller notes (non-blocking)**

- `docs/agents/desktop-user-extensions.md` says `refreshExpiredTokens()` runs at agent-runtime start, "so a session never begins with a stale token" — but `AgentRuntimeProcess.swift` launches it with an unawaited `Task { … }` concurrently with spawning the runtime, so the child can read `mcp.json` before the refresh lands. Fail-open makes that benign; the doc's "never" is stronger than the code.
- In `pi-mono-extension/index.ts`, `registerUserMcpTools` finds the `Authorization` header that `loadLocalMcpConfig` just built and strips the `Bearer ` prefix it just added before passing it to `McpHttpClient`. Harmless, but carrying `token` on the server type would remove the round trip.
- `LocalMcpStore+OAuth.swift` persists `access_token`/`refresh_token` in plaintext `~/.omi/mcp.json`. That matches other local MCP clients and the file is user-owned, but a sentence in the doc telling users their tokens live there would be kind.
- `Int.random(in:)` for the PKCE verifier and state is fine in practice on this platform; `SecRandomCopyBytes` would be the strictly-correct source.

Also flagging for maintainers (not contributor blockers): this PR adds a native OAuth implementation (dynamic client registration, loopback listener, token persistence) and a user-configured local-command execution path wired into the agent runtime, and it extends `AGENTS.md` plus `docs/agents/` that coding agents working in this repo read. I verified the docs against the code (with the two overclaims above), but the auth and trust-boundary surface deserves maintainer sign-off before merge.

---

Git-on-my-level:
@katipally — coming back to this after the new pushes. The blocker from the earlier automated review is resolved on the current head, and the additions since are substantial; the outdated request is being dismissed with this comment as the record.

**Blocker resolved and verified**
- `mcp-stdio-client.ts` now refs the child and its stdout while an RPC is in flight and unrefs them only when the pending map drains (`holdEventLoop`) — exactly the failure mode that cancelled `registerUserMcpTools` in CI. `agent/tests/mcp-stdio-client.test.ts` locks it in with a driver that must survive an idle event loop, and the Agent Runtime / Hygiene / PR Metadata Preflight lanes are green on this head. The stdin EPIPE handler added alongside (`failAll` on a dead child's stdin) closes the crash path that used to take the whole agent down with one exited server.

**What I checked on the new surface (all spot-checks passed)**
- `ExtensionCatalog.SkillSource.isSafe` rejects `..`, `.`, empty and absolute components, and is enforced at both the fetch (`rawURL(for:)`) and the write (`LocalSkillsStore.saveSkillBundle`), with the 200-file / 4 MB / 32 MB caps in `ExtensionCatalogService` — a remote skill folder cannot write outside `~/.omi/skills/<slug>`.
- `LocalMcpStore+OAuth.swift` now draws the PKCE verifier and `state` from `SecRandomCopyBytes` (with a test asserting full-length, unique, URL-safe output), and the docs state plainly that tokens live in plaintext in `~/.omi/mcp.json` and that `refreshExpiredTokens()` races the spawn rather than preceding it.
- `mcp-client.ts` records `initialize` capabilities (a server that publishes no prompts is never asked for them), walks `tools/list` and `prompts/list` to the last cursor under a page budget with a repeated-cursor guard, and `describeContent` names image/audio/resource blocks instead of dumping a base64 payload into the conversation; `structuredContent` answers when a tool returns data with no prose.
- `mcp-sse-client.ts` normalizes CRLF before the blank-line frame split (with a `crlf: true` test case), caps a single unterminated frame at 4 MB, and marks stored promises handled so a disposed client cannot crash the host; configured headers pass through verbatim, fixed for non-Bearer schemes in both `mcp-http-client.ts` and the pi registration path.
- The Smithery removal in `ExtensionCatalog` (`excludedNamespaces`, `io.github.*` de-ranked in `ranked(_:)`) matches the reasoning in the PR description, and install data comes from the registry response rather than the curated list.

**Two small non-blocking notes**
- `docs/agents/desktop-user-extensions.md`'s runtime-contract table still says the pi lane registers via "McpHttpClient / McpStdioClient (5s http cap…)": this head also drives `transport: "sse"` servers through `McpSseClient`, and the http/sse discovery cap is 10s (`discoveryTimeoutMs = 10_000`), not 5s. Worth syncing that row.
- `McpSseClient.handleFrame` accepts any absolute URL in the `endpoint` event (`new URL(payload, this.url)`), so a server can direct POSTs — carrying the headers configured for it — to another origin. The headers are that server's own, so this is a hardening nit rather than a hole (the reference SDK validates same-origin), but an origin check would be cheap.

**For maintainers** — this PR extends `desktop/macos/AGENTS.md` and adds `docs/agents/desktop-user-extensions.md`, both of which coding agents working in this repo read as guidance. I verified the docs against the code at this head (modulo the stale table row above) and found nothing that would misdirect an agent; the trust model and failure behavior are stated accurately. Docs that steer agent behavior still deserve a human read before merge.

Leaving `security-review` and `needs-maintainer-review` in place — not because anything looks wrong (the review above is genuinely positive), but because the trust boundary is new: a native OAuth client (dynamic registration, loopback listener, plaintext token persistence) and a user-configured local-command execution path now feed the agent runtime. That surface needs maintainer sign-off. `needs-tests` no longer applies: the event-loop regression test plus the new suites (`ExtensionCatalogTests`, `LocalSkillsStoreTests`, `McpServerProbeTests`, `SkillEditorDropTests`, `mcp-content-blocks.test.ts`, `mcp-sse-client.test.ts`, `user-extensions.test.ts`, the pi extension tests) cover the new logic well. Adding `positive-signal` for the quality of the iteration.

---

Git-on-my-level:
Reviewing the new head — 07173c3 is the merge of upstream/main rather than new feature work, so this pass focuses on what the merge changed and what is still open.

**Merge integrity (all checks pass)**
- The only conflicted file was `.github/agent-docs/desktop-user-extensions.md`; the resolved copy sits at the repo's agent-docs location and `desktop/macos/AGENTS.md`'s new section links to it correctly (`../../.github/agent-docs/...`).
- The event-loop fix from `f1175fd2d0` survived the merge: `mcp-stdio-client.ts` still refs the child and its stdout while a call is pending (`holdEventLoop(true)` in `rpc`, unref in the `finally`), the EPIPE handler on `child.stdin` is intact, and `agent/tests/mcp-stdio-client.test.ts` is present on this head. Desktop Swift Build & Tests, Agent Runtime, and Hygiene are green on `07173c3`.
- Doc-vs-code on the moved path: `OMI_USER_SKILLS_DIR` / `OMI_LOCAL_MCP_FILE` in `AgentRuntimeProcess.swift`, per-session reload in `agent/src/index.ts`, fail-open `loadLocalMcpConfig` in `user-extensions.ts`, and the skill-folder staging + `replaceItemAt` swap in `LocalSkillsStore.swift` all match what the doc claims — except one row, below.

**Still open from the last review (carried forward, not new findings)**
- `mcp-sse-client.ts:157` — `handleFrame` still accepts any absolute URL in the `endpoint` event (`new URL(payload, this.url)`), so a server can direct POSTs — carrying its own configured headers — to another origin. Still a hardening nit rather than a hole; the headers belong to that server either way.
- `.github/agent-docs/desktop-user-extensions.md` runtime-contract table still reads "via `McpHttpClient` / `McpStdioClient` (5s http cap …)": this head also drives `transport: "sse"` servers through `McpSseClient`, and the discovery caps are 10s http/sse and 30s stdio (`pi-mono-extension/index.ts`, ~889–906). Worth syncing next time the doc is touched.

**Agent-guidance note** — this PR adds agent-readable guidance (`.github/agent-docs/desktop-user-extensions.md`, new, 158 lines) and a new section in `desktop/macos/AGENTS.md` pointing at it. I checked the guidance against the code at this head: file formats, env-var contract, per-session reload, fail-open failure behavior, and install caps are accurate, modulo the stale table row above. Because these files steer coding and review agents working in this repo, they still deserve a human read before merge.

No new blockers on the merged head. `security-review` and `needs-maintainer-review` stay in place: the new local OAuth client (dynamic registration, loopback listener, plaintext token persistence in `~/.omi/mcp.json`) plus the user-configured local-command lane are a new trust boundary that needs maintainer sign-off, and the marketplace curation choices (registry-only sourcing, Smithery exclusion, `io.github.*` de-ranking) are product decisions. `positive-signal` stays for the quality of the iteration.

---
*AI-generated maintainer feedback on behalf of David. Human review needed for: sign-off on the new local OAuth/token-persistence surface and the marketplace curation policy.*

---

Git-on-my-level:
Reviewed the two commits since the last pass (`31db656`, `21dfe66`) — both are UI polish on top of the merged head, and neither touches the extension, OAuth, or security surface. No new blockers.

**What the new commits do**
- `DismissableSheet.swift` — the clamp that bounds a sheet to its host now wraps content in `ScrollView(.vertical)` with `.scrollBounceBehavior(.basedOnSize)`, so a sheet taller than its window scrolls instead of being silently cut off. Since this is the shared modifier, every dismissable sheet in the app benefits, not just the new ones; a sheet that fits still sizes to its content. Good fix for the short-window / larger-text case the comment describes.
- `AgentExtensionsSections.swift` — `AddMcpServerSheet` restyled onto the app's own controls: mode chips (`add-mcp-mode-remote` / `add-mcp-mode-local`) replace the segmented picker, `labeledField` gains `@FocusState` focus rings and hint placement, the waiting and saved states share one `statusCard` shape, and `submit()` now guards `canSubmit` / `phase` and clears focus — so Return from a half-filled field can neither no-op nor double-submit. The accessibility identifiers (`add-mcp-name`, `add-mcp-url`, `add-mcp-command`, `add-mcp-submit`) are handy e2e hooks too.
- `AppsPage.swift` — Add Server sheet frame bumped 440×340 → 460×424 to fit the new layout.

Desktop Swift Build & Tests, Release Compile, Static & Test Contracts, and the e2e t0 lane are green on this head.

**Still open from the last review (unchanged by these commits, both non-blocking)**
- `mcp-sse-client.ts` `handleFrame` still resolves the server's `endpoint` event with `new URL(payload, this.url)` — an absolute URL from a server can redirect its POSTs (carrying that server's own configured headers) to another origin. A same-origin / same-path check would close it; still a hardening nit rather than a hole.
- `.github/agent-docs/desktop-user-extensions.md` runtime-contract table still says the pi-mono lane drives servers "via `McpHttpClient` / `McpStdioClient` (5s http cap …, 30s stdio…)" — the code now also routes `transport: "sse"` servers through `McpSseClient`, and the discovery caps are 10s remote / 30s stdio (`pi-mono-extension/index.ts`). Worth syncing next time the doc is touched, especially since agents read it.

Everything re-verified on this head stands as before: PKCE verifier and `state` drawn from `SecRandomCopyBytes` with the state check enforced, the one-shot loopback listener bound to 127.0.0.1, skill bundles path-validated and size-capped with a staged atomic swap, and the status probe resolving local commands without ever launching them. The agent-readable guidance this PR adds (the agent-docs page, the new `AGENTS.md` section, and the registered failure class) remains accurate against the code modulo the table row above.

Human maintainer sign-off is still the gate before merge: the native OAuth implementation and the user-configured local-command execution wired into the agent runtime are product and trust-boundary decisions, not just code correctness — `security-review` and `needs-maintainer-review` stay on.

---

Git-on-my-level:
Reviewed the new head `15e850e` — the only commit since the last pass is the merge of upstream/main, so this pass covers merge integrity and the CI state rather than new feature work.

**Merge integrity**
- The four PR-owned files that needed resolution all came out clean: `AgentRuntimeProcess.swift` still exports `OMI_USER_SKILLS_DIR` / `OMI_LOCAL_MCP_FILE` and kicks off the unawaited `refreshExpiredTokens()`; `ChatProvider.swift` keeps the `omiUserSkillsDidChange` observer and the `~/.omi/skills` discovery; `agent/src/index.ts` still appends `loadLocalMcpConfig()` servers in `buildMcpServers()`; `pi-mono.ts` keeps the `PI_CODING_AGENT_DIR` handoff. The rest of the merge is upstream's own churn, none of it on this feature's surface.

**CI triage: the two red checks are not this PR**
- Both failing checks fail in one suite: `ProactiveListenEventTests.testHandlerRecognisesProactiveMessageType`, which crashes inside `UNUserNotificationCenter.currentNotificationCenter` (`NotificationService.swift:171` — `bundleProxyForCurrentProcess is nil` in the bare xctest host).
- Every file on the crash path — `ProactiveListenEventTests.swift`, `NotificationService.swift`, `AppState/AppState+ListenEvents.swift`, `RuntimeOwnerIdentity.swift` — is byte-identical between the previous green head (`21dfe66e`, all five desktop checks green two days ago) and this head, and none of them is touched by this PR.
- Mechanism, as far as the logs go: the test expects no runtime owner, but the admission check read a stale owner out of the per-worker defaults home. `swift-test-suites.sh` gives each worker one `CFFIXED_USER_HOME` reused by every suite it runs, and this merge grew the suite count (742 → 748), which reshuffles which suites share a worker's home. A leaked owner flips `ProactiveListenAdmission` into `.deliver`, and delivery then reaches the notification center, which cannot exist in the xctest tool. For corroboration, upstream main's own Desktop Swift lane failed the same two checks this morning on a different suite (`HubEscalationTests`) before later runs went green — the lane is flaky on main right now. A re-run should clear this branch's checks; this is not a regression from the PR.

**Carried from the last pass (non-blocking, untouched by the merge)**
- `mcp-sse-client.ts` still resolves the server's `endpoint` event with `new URL(payload, this.url)` — an absolute URL from a server can point its POSTs (carrying that server's configured headers) at another origin. A same-origin check would close it.
- `.github/agent-docs/desktop-user-extensions.md`'s pi-mono row still lists only `McpHttpClient` / `McpStdioClient` with a 5s http cap; the code also routes `transport: "sse"` through `McpSseClient` and uses 10s remote / 30s stdio discovery caps. Worth syncing the next time the doc is touched, since agents read it.

Security posture re-verified unchanged on this head: PKCE verifier and `state` drawn from `SecRandomCopyBytes`, state mismatch rejected, one-shot loopback listener bound to 127.0.0.1, S256 challenge method. Agent Runtime, Desktop Swift Release Compile, desktop-core-e2e-t0, and Hygiene are green.

---
*by AI on behalf of David — before merge this still needs maintainer sign-off on the OAuth token handling and the user-configured local-command execution now wired into the agent runtime; `security-review` and `needs-maintainer-review` stay on.*

Git-on-my-level:
# Review — PR #12377 feat(desktop): user MCP servers and skills, fully local under ~/.omi

Head: `5618aef448f819d5eee93d2d9f03ea62bceae9e4` · Recommendation: **comment_only** · Confidence: high

Reviewed the new head `5618aef4` — again a merge of upstream `origin/main` (the branch also tracks a `qa/pr-12377-local-mcp-sync` remote). This pass covers the merge resolutions, a fresh security read of the PR-owned surfaces, and one new finding.

## Merge integrity (4 resolved files, all clean)

- `agent/src/index.ts` — upstream's durable-queue refactor replaced the three inline outbox drains with `pumpJournalOutboxDeliveries`; this PR's `buildMcpServers()` still appends `loadLocalMcpConfig()` output, with built-in server names seeded as reserved so a user entry can never shadow a first-party server.
- `ChatProvider.swift` — upstream's memory-review card, follow-up chip, and rating write-chain land around the intact `omiUserSkillsDidChange` observer and the `~/.omi/skills` scan in `loadClaudeConfigFromDisk`.
- `AppsPage.swift` — upstream's search analytics wrap the same paths; the `selectedSection == .apps` guard on search scope holds.
- `desktop/macos/AGENTS.md` — beyond this PR's own section, only upstream's Gemini Live line.

The other 36 PR-owned files are byte-identical to the previously reviewed head `15e850e`. All checks are green on this head — including the desktop lane that was triaged as suite-assignment flake on `15e850e`.

## Security re-read at this head (no change in posture)

- `LocalMcpStore+OAuth.swift` — PKCE verifier and `state` from `SecRandomCopyBytes` (S256), state mismatch rejected, one-shot loopback listener bound to `127.0.0.1`, RFC 9728 protected-resource discovery with the RFC 8707 `resource` parameter, dynamic client registration as a public client. The unawaited `refreshExpiredTokens()` race is documented and genuinely fail-open.
- Skill install path — `SkillSource.isSafe` rejects `..`/`.`/empty components and absolute paths, re-checked at fetch (`rawURL`) and write (`saveSkillBundle`); installs stage under `NSTemporaryDirectory` and swap atomically; catalog fetches capped at 200 files / 4 MB / 32 MB and fail-open.
- `mcp-client.ts` — pagination bounded by page budget and repeated-cursor detection; non-text content blocks named rather than inlined (no more base64 dumps into context); `structuredContent` fallback.
- Marketplace defaults point only at the official MCP registry and `anthropics/skills` / `obra/superpowers`; `ai.smithery` is excluded. No secrets, no workflow or dependency changes in the diff.

## New finding (non-blocking, belongs with the maintainer sign-off)

`mcp-stdio-client.ts` spawns user-configured commands with `{...process.env}` — which includes `OMI_AUTH_TOKEN` and the forwarded `OMI_BYOK_*` provider keys. That matches how peer MCP clients launch stdio servers, but this runtime's own precedent is to scrub secrets before spawning subprocesses (`AgentRuntimeProcess` removes `ANTHROPIC_API_KEY`, `PLAYWRIGHT_MCP_EXTENSION_TOKEN`, and inherited BYOK keys). A user-added stdio server therefore receives credentials the Omi backend issued. A minimal allowlist env (or scrubbing the `OMI_*` prefix) at the spawn site would make this consistent; leaving as-is is a maintainer call on the threat model.

## Carried from prior passes (unchanged, non-blocking)

- `mcp-sse-client.ts` resolves the server's `endpoint` event with `new URL(payload, this.url)` — an absolute URL from a server can point its POSTs (carrying that server's configured headers) at another origin. A same-origin check would close it.
- `.github/agent-docs/desktop-user-extensions.md`'s pi-mono row still says 5s http cap and lists only `McpHttpClient`/`McpStdioClient`; code uses 10s remote / 30s stdio and also routes `transport: "sse"` through `McpSseClient`. Worth syncing since agents read this doc as the subsystem contract.

Agent-docs note: this PR adds a new agent-behavior contract (`.github/agent-docs/desktop-user-extensions.md` + the `AGENTS.md` section). I verified its claims against the code at this head — accurate except the stale cap/transport row above. The doc will steer coding/review agents on these surfaces, so the maintainer sign-off should cover it together with the code.

No urgent security indicators on this head.

---
*by AI on behalf of David — before merge this still needs maintainer sign-off on the OAuth token handling and the user-configured local-command execution now wired into the agent runtime, including the env-inheritance question above; `security-review` and `needs-maintainer-review` stay on.*

PR #12287, opened 2026-08-27
fix(backend): fence deferred conversation enrichment ownership

aryanorastar:
## Summary

- make `deferred=true` the compare-and-swap ownership token for lazy conversation enrichment, so concurrent first opens cannot both launch LLM processing
- atomically publish failed enrichment as `completed + deferred=true`, allowing the UI to stop polling and a later open to retry
- keep both paired transitions behind the conversation lifecycle owner and extend the static lifecycle-write tripwire to cover `deferred`
- document the deferred desktop lifecycle contract

## Root cause

The first-open transaction checked only `status=processing`. Two requests could both read the deferred row, then each transaction could succeed even after the first had cleared `deferred`, duplicating summaries, tasks, memories, app results, and LLM cost.

Failure recovery was also split between a raw flag update and a separate status transition. Besides permitting a partial write to strand a row in `processing`, it produced `completed + deferred=true` even though reacquisition rejected every completed row, so the documented retry path could never run.

## Fix

The storage transaction now claims only an explicit deferred row and writes `status=processing`, `deferred=false`, and the renewed admission lease together. It also accepts the explicit `completed + deferred=true` retry terminal. A second transaction owned by the lifecycle service atomically re-arms a failed attempt as `completed + deferred=true`; the router no longer performs lifecycle writes directly and no longer silently drops recovery errors.

### Regression-first

The new contracts failed against current `main` with four failures:

- a second caller could reacquire `processing + deferred=false`
- `completed + deferred=true` could not be retried
- no atomic failure-recovery primitive existed
- the router still bypassed the lifecycle owner

### Automated

- `cd backend && .venv/bin/python -m pytest tests/unit/test_conversation_finalization_jobs.py tests/routers/test_conversations_processing_rollback.py tests/unit/test_check_conversation_lifecycle_writes.py -q` — **64 passed**
- `python3 backend/scripts/check_conversation_lifecycle_writes.py` — **passed**
- Firestore emulator: `pytest backend/tests/unit/test_stale_processing_emulator_concurrency.py -k deferred_reacquire -q` — **2 passed**, including a real two-open contention race with exactly one winner
- `OMI_PR_BODY_FILE=/tmp/omi-pr-12286.md make preflight` — **26 checks passed**

### Full-suite baseline audit

[...]

PR #12380, opened 2026-08-29
fix(backend): use locked LC3 runtime artifact

### LINKED ISSUE #12345: backend: the liblc3 wheel is installed into an interpreter /opt/venv then shadows, and /tmp/wheels ships in both images
Two things in `backend/Dockerfile` and `backend/pusher/Dockerfile`, found while sweeping published container images for one defect class (`COPY` into a path, `rm` of that path in a *later* instruction). The first is that class. The second is what I found when I read the closure to size it, and it is the one I think actually matters.

Everything below is read off `main` at the time of writing; line numbers are from there.

## 1. `/tmp/wheels` ships in every pull of both images

`backend/Dockerfile`

```
60  COPY --from=builder /tmp/wheels /tmp/wheels
61
62  # Install liblc3 Python package and set library path
63  RUN ldconfig && \
64      pip install --no-cache-dir /tmp/wheels/*.whl && \
65      rm -rf /tmp/wheels
```

`backend/pusher/Dockerfile` is the same shape at lines 68 / 72 / 73.

The `COPY` commits `/tmp/wheels` to its own layer. The `rm -rf` runs in a **later** layer, so it writes a whiteout rather than un-committing anything — the bytes are still in the manifest and still transferred on every pull. This is small here (the wheel dir holds the liblc3 build, not a dependency closure), so it is a tidiness point on its own. I am reporting it because it is what led me to the next part.

## 2. The wheel is installed into an interpreter that `/opt/venv` then shadows

At line 64 the final stage runs `pip install`. At that moment `/opt/venv` **does not exist in the final stage** — the stage starts fresh at line 42, and `/opt/venv` is not copied in until line 67:

```
42  FROM ${PYTHON_BASE_IMAGE}
45  ENV PATH="/opt/venv/bin:$PATH"
...
64      pip install --no-cache-dir /tmp/wheels/*.whl && \
...
67  COPY --from=builder /opt/venv /opt/venv
```

So `PATH` names `/opt/venv/bin` first, but nothing is there yet; `pip` resolves to the base image's interpreter and the wheel lands in **system** site-packages. Line 67 then drops the builder's `/opt/venv` on top, and the venv is created without system site-packages:

```
8   RUN python -m venv /opt/venv          # no --system-site-packages
```

At runtime `PATH` puts `/opt/venv/bin/python` first, and that interpreter cannot see the system install.

What it *can* see is the locked wheel, which `uv pip sync` put in the venv back in the builder stage (line 40). `backend/pylock.runtime.toml`:

```toml
[[packages]]
name = "lc3py"
version = "1.1.3"
marker = "platform_machine == 'x86_64' and sys_platform == 'linux'"
wheels = [{ url = ".../lc3py-1.1.3-py3-none-manylinux_2_5_x86_64...whl", size = 144335, ... }]
```

`backend/pusher/pylock.toml` pins the same `lc3py` 1.1.3 at line 482, there with **no marker at all**. The backend deploys to Cloud Run (`.github/workflows/gcp_backend.yml`), which is `linux/amd64`, so the marker matches and the locked wheel is installed.

**So `import lc3py` at runtime resolves to PyPI `lc3py` 1.1.3 from the lock — never to the wheel built from source in the builder stage.**

### What that means for `LIBLC3_COMMIT`

`backend/pusher/Dockerfile:7` pins the source build deliberately:

```
ENV LIBLC3_COMMIT=ce2e41faf8c06d038df9f32504c61109a14130be
```

If the intent of that pin is to control the **Python binding** that runs, it does not — the binding comes from the lock. Each build still clones `google/liblc3`, runs a meson/ninja compile, and apt-installs `git gcc g++ meson ninja-build python3-dev` in the builder to produce a wheel that no running interpreter imports.

### What I am *not* claiming

`COPY --from=builder /opt/liblc3/ /usr/local/lib/` (line 59, and 28 in pusher) is a different thing and is **unaffected**. `lc3py` ships as `py3-none-manylinux`, i.e. a binding rather than a compiled extension, so the pinned shared object at `/usr/local/lib` plausibly *is* what gets loaded — in which case pinning the `.so` is doing real work and only the `pip wheel` / `pip install` half is dead. I could not settle that from the repository alone, and it is your call, not mine.

I also have not checked whether anything outside Cloud Run builds these Dockerfiles for `arm64`, where the backend marker would not match and the source build would be the only path.

## If you want the layer gone regardless

Bind-mounting the wheels instead of copying them removes the layer without changing behaviour:

```dockerfile
RUN --mount=type=bind,from=builder,source=/tmp/wheels,target=/tmp/wheels \
    ldconfig && \
    pip install --no-cache-dir /tmp/wheels/*.whl
```

That needs `# syntax=docker/dockerfile:1` at the top of the file, which neither Dockerfile currently has.

If instead you conclude from §2 that the wheel build is redundant on the deployed platform, the smaller change is to drop lines 60 and 63–65 (68, 72–73 in pusher) and keep the `.so` copy.

Happy to open a PR for whichever of the two you prefer — I did not want to guess which, since the answer depends on what `LIBLC3_COMMIT` is meant to pin.

aryanorastar:
## What changed and why

Fixes #12345.

The backend and pusher images now install LC3 exactly once, from their checked-in `uv` locks into `/opt/venv`. This removes the redundant `google/liblc3` source checkout/build, the system-interpreter wheel install that was later shadowed by `/opt/venv`, the copied-and-whiteouted `/tmp/wheels` layer, and the unused `/usr/local/lib` staging path.

This is behavior-preserving on the deployed Linux x86_64 target: both locks already resolve `lc3py==1.1.3` to the same SHA-256-pinned manylinux wheel, and that wheel contains both the Python binding and its native `liblc3.so.1`. The binding loads that colocated library before consulting the system library search path.

none

## How it was verified

- `python3 -m pytest -q --noconftest backend/tests/unit/test_runtime_image_contracts.py backend/tests/unit/test_verify_pusher_source_closure.py` — 21 passed.
- `make runtime-image-source-closure` — all 12 registered runtime-image contracts passed.
- `python3 -m black --check backend/scripts/runtime_image_contracts.py backend/tests/unit/test_runtime_image_contracts.py backend/tests/unit/test_verify_pusher_source_closure.py` — clean.
- Inspected the exact wheel pinned by both lockfiles (`sha256:601a9a2ea3135547f2d22a10475b5fec8f784f716f4b22a61acc3908c827694e`): its archive contains `.lc3py.mesonpy.libs/liblc3.so.1`, and `lc3.py` prefers that colocated library.
- Built both final images locally for the deployed `linux/amd64` platform. The fork-only private GCR base was replaced through the existing `PYTHON_BASE_IMAGE` build argument with public `python:3.11-slim`; every remaining Dockerfile instruction and locked dependency was unchanged.
- Ran the registered offline smoke contract against both built images: the backend resolved all 302 reachable third-party modules, while the pusher resolved all 192 and imported its `routers.pusher` entrypoint.
- Instantiated a real `lc3.Decoder(10000, 16000)` inside both network-isolated final images. Both loaded `lc3py==1.1.3` and `/opt/venv/lib/python3.11/site-packages/.lc3py.mesonpy.libs/liblc3.so.1`; no system LC3 installation was present or needed.
- The GitHub image-smoke job is intentionally skipped for fork PRs because the workflow's production base-image credentials are private. The equivalent local final-image proof above covers this branch; the static source-closure job still runs on the fork.

## Tests

`backend/tests/unit/test_runtime_image_contracts.py` now proves the shared runtime-image checker rejects both defect shapes from #12345:

- a path copied into one final-stage layer and removed in a later `RUN`, which still ships the bytes;
- a `pip install` executed while `PATH` points at a venv that is only copied into the stage afterward, which shadows the install.

The existing registry-wide contract test also verifies every registered image is free of both patterns.

## Failure class (fixes)

Failure-Class: none

This shared guard would have caught #12345. It lives in the existing runtime-image contract primitive rather than a Dockerfile-specific layout test, so every registered deployable Python image inherits the protection.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -9,30 +9,13 @@ RUN python -m venv /opt/venv
 RUN python3 -m pip install --no-cache-dir "uv==${UV_VERSION}"
 ENV PATH="/opt/venv/bin:$PATH"
 
-# Install build dependencies for liblc3
+# Build locked source distributions such as webrtcvad.
 RUN apt-get update && apt-get install -y \
     git \
     gcc \
     g++ \
-    meson \
-    ninja-build \
     python3-dev \
     && rm -rf /var/lib/apt/lists/*
-
-# Build liblc3 and create wheel
-WORKDIR /tmp
-RUN git clone https://github.com/google/liblc3.git && \
-    cd liblc3 && \
-    meson setup build && \
-    cd build && \
-    meson install && \
-    ldconfig && \
-    # Stage shared libs to a fixed path (meson install path varies by arch)
-    mkdir -p /opt/liblc3 && \
-    cp /usr/local/lib/*/liblc3.so* /opt/liblc3/ 2>/dev/null || \
-    cp /usr/local/lib/liblc3.so* /opt/liblc3/ && \
-    cd /tmp/liblc3 && \
-    python3 -m pip wheel --no-cache-dir --wheel-dir /tmp/wheels .
 
 # Install Python requirements from the checked-in runtime lock.
 WORKDIR /opt/venv
@@ -43,7 +26,6 @@ FROM ${PYTHON_BASE_IMAGE}
 
 WORKDIR /app
 ENV PATH="/opt/venv/bin:$PATH"
-ENV LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 
 # Patch OS packages in this layer before installing runtime deps (#7136).
 RUN apt-get update && apt-get -y dist-upgrade \
@@ -54,15 +36,6 @@ RUN apt-get update && apt-get -y dist-upgrade \
 # under high-churn bytearray allocations from WebSocket audio streams (#6938)
 # Use soname-only form so ld.so resolves via standard search paths (arch-portable)
 ENV LD_PRELOAD=libjemalloc.so.2
-
-# Copy compiled liblc3 library from builder staging dir
-COPY --from=builder /opt/liblc3/ /usr/local/lib/
-COPY --from=builder /tmp/wheels /tmp/wheels
-
-# Install liblc3 Python package and set library path
-RUN ldconfig && \
-    pip install --no-cache-dir /tmp/wheels/*.whl && \
-    rm -rf /tmp/wheels
 
 COPY --from=builder /opt/venv /opt/venv
 COPY .github/scripts/desktop_release_manifest.py /app/desktop_release_manifest_contract.py
--- a/backend/pusher/Dockerfile
+++ b/backend/pusher/Dockerfile
@@ -4,39 +4,17 @@ ARG UV_VERSION=0.11.13
 FROM ${PYTHON_BASE_IMAGE} AS builder
 
 ARG UV_VERSION
-ENV LIBLC3_COMMIT=ce2e41faf8c06d038df9f32504c61109a14130be
 
 RUN python -m venv /opt/venv
 RUN python3 -m pip install --no-cache-dir "uv==${UV_VERSION}"
 ENV PATH="/opt/venv/bin:$PATH"
 
-# Install build dependencies for liblc3
+# Build locked source distributions such as webrtcvad.
 RUN apt-get update && apt-get install -y \
-    git \
     gcc \
     g++ \
-    meson \
-    ninja-build \
     python3-dev \
     && rm -rf /var/lib/apt/lists/*
-
-# Build liblc3 and create wheel
-WORKDIR /tmp
-RUN git init liblc3 && \
-    git -C liblc3 remote add origin https://github.com/google/liblc3.git && \
-    git -C liblc3 fetch --depth=1 origin "${LIBLC3_COMMIT}" && \
-    git -C liblc3 checkout --detach FETCH_HEAD && \
-    cd liblc3 && \
-    meson setup build && \
-    cd build && \
-    meson install && \
-    ldconfig && \
-    # Stage shared libs to a fixed path (meson install path varies by arch)
-    mkdir -p /opt/liblc3 && \
-    cp /usr/local/lib/*/liblc3.so* /opt/liblc3/ 2>/dev/null || \
-    cp /usr/local/lib/liblc3.so* /opt/liblc3/ && \
-    cd /tmp/liblc3 && \
-    python3 -m pip wheel --no-cache-dir --wheel-dir /tmp/wheels .
 
 # Install Python requirements from the pusher-specific checked-in lock.
 WORKDIR /opt/venv
@@ -47,7 +25,6 @@ FROM ${PYTHON_BASE_IMAGE}
 
 WORKDIR /app
 ENV PATH="/opt/venv/bin:$PATH"
-ENV LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 
 # Patch OS packages in this layer before installing runtime deps (#7136).
 RUN apt-get update && apt-get -y dist-upgrade \
@@ -62,15 +39,6 @@ RUN apt-get update && apt-get -y dist-upgrade \
 # under high-churn bytearray allocations from WebSocket audio streams (#6938)
 # Use soname-only form so ld.so resolves via standard search paths (arch-portable)
 ENV LD_PRELOAD=libjemalloc.so.2
-
-# Copy compiled liblc3 library from builder staging dir
-COPY --from=builder /opt/liblc3/ /usr/local/lib/
-COPY --from=builder /tmp/wheels /tmp/wheels
-
-# Install liblc3 Python package and set library path
-RUN ldconfig && \
-    pip install --no-cache-dir /tmp/wheels/*.whl && \
-    rm -rf /tmp/wheels
 
 COPY --from=builder /opt/venv /opt/venv
 
--- a/backend/scripts/runtime_image_contracts.py
+++ b/backend/scripts/runtime_image_contracts.py
@@ -210,7 +210,7 @@ def _logical_docker_lines(path: Path) -> list[str]:
     return logical_lines
 
 
-def final_stage_copy_instructions(dockerfile: Path) -> list[CopyInstruction]:
+def docker_stages(dockerfile: Path) -> list[list[str]]:
     stages: list[list[str]] = []
     current_stage: list[str] | None = None
     for line in _logical_docker_lines(dockerfile):
@@ -222,6 +222,11 @@ def final_stage_copy_instructions(dockerfile: Path) -> list[CopyInstruction]:
             current_stage.append(line)
     if not stages:
         raise ContractError(f"{_repository_relative(dockerfile)} has no Docker stage")
+    return stages
+
+
+def final_stage_copy_instructions(dockerfile: Path) -> list[CopyInstruction]:
+    stages = docker_stages(dockerfile)
 
     copies: list[CopyInstruction] = []
     for line in stages[-1]:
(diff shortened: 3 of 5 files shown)


aryanorastar:
## Verification proof for `f2df5c60af`

I reproduced the image construction path and tested the resulting Linux artifacts, rather than relying only on Dockerfile inspection.

### Root cause confirmed

Both Dockerfiles built an LC3 wheel into `/tmp/wheels`, copied it into the final image, and ran `pip install` before copying `/opt/venv` over the same interpreter path. That made the first install unreachable. The later `rm -rf /tmp/wheels` also did not remove those bytes from the earlier image layer.

Both checked-in locks already pin the same `lc3py==1.1.3` manylinux wheel with SHA-256 `601a9a2ea3135547f2d22a10475b5fec8f784f716f4b22a61acc3908c827694e`. I inspected that exact archive: it contains `.lc3py.mesonpy.libs/liblc3.so.1`, and its binding loads this colocated library first. The separate source checkout, Meson build, system install, and `/usr/local/lib` copy were therefore dead runtime paths.

### Final-image builds

I built both Dockerfiles for the deployed `linux/amd64` platform. Because production GCR credentials are intentionally unavailable to fork PRs, I used the existing build argument to substitute public `python:3.11-slim` for the private base; all remaining instructions and lockfiles were identical.

The first real builds usefully caught two unrelated locked source-build requirements before the final push:

- `webrtcvad` requires the compiler and Python headers in both builders;
- the backend lock's pinned PyOgg VCS dependency requires `git`.

The final Dockerfiles retain those necessary generic builder packages but remove the LC3-specific clone, Meson/Ninja build, temporary wheel directory, system install, and duplicate native-library path.

### Runtime results

- Backend image: all **302** reachable locked third-party modules resolved offline.
- Pusher image: all **192** reachable locked third-party modules resolved offline, and `routers.pusher` imported successfully.
- In both network-isolated final images, `lc3.Decoder(10000, 16000)` instantiated successfully.
- Both images reported:
  - package: `lc3py==1.1.3`
  - module: `/opt/venv/lib/python3.11/site-packages/lc3.py`
  - native library: `/opt/venv/lib/python3.11/site-packages/.lc3py.mesonpy.libs/liblc3.so.1`
  - native library exists: `True`

### Regression and repository checks

- Focused unit tests: **21 passed**.
- Shared source-closure registry: **12/12 images passed**.
- Black: clean.
- Full local PR preflight: **26/26 checks passed**.
- Added generic mutation fixtures proving the shared checker rejects both the copied-then-whiteouted layer pattern and the venv-shadowed install pattern.

GitHub's image-smoke job is skipped by workflow policy on fork PRs because it requires private production-base credentials. The equivalent final-image builds and registered smoke commands above were run locally against this exact commit.

aryanorastar:
## Final CI result

All applicable GitHub checks are green on exact head `f2df5c60af`.

- Backend unit suite: passed (15m27s)
- Listen Pusher Stack Gauntlet: passed (7m23s)
- Hermetic Backend E2E: passed
- Backend Hermetic Merge Gate: passed
- Sync Cloud Tasks Stack Gauntlet: passed
- Runtime image source closure: passed
- PR metadata, hygiene, formatting, and public API contract: passed

The image-smoke job is the documented fork-policy skip; both final `linux/amd64` images were built and smoked locally as recorded above.

Git-on-my-level:
Verified on head f2df5c6 — this is a clean fix for #12345, and the central claim checks out end-to-end.

Independent verification of the behavior-preservation claim:

- Both `backend/pylock.runtime.toml` and `backend/pusher/pylock.toml` already pin `lc3py==1.1.3` to the manylinux wheel `sha256:601a9a2e…694e`, unchanged by this PR. I downloaded that exact artifact from PyPI and confirmed the digest matches and the archive bundles `.lc3py.mesonpy.libs/liblc3.so.1`.
- The wheel's own `lc3.py` loader prefers that colocated library (it globs the module directory first, `find_library("lc3")` only as fallback), so removing the `LD_LIBRARY_PATH=/usr/local/lib` ENV and the `/opt/liblc3` staging from both Dockerfiles deletes dead paths, not load-bearing ones.
- The one runtime consumer (`backend/routers/listen/receiver.py`: `Decoder(frame_duration_us, sample_rate)` plus `decode(..., bit_depth=16)`) is fully covered by the 1.1.3 wheel; nothing in the repo used the newer wrapper APIs that only the removed source build carried.

Per file:

- `backend/Dockerfile` — builder keeps exactly the toolchain still needed for locked sdist builds (`git` for the PyOgg VCS pin, `gcc`/`g++`/`python3-dev` for webrtcvad); the meson/ninja liblc3 build, `/tmp/wheels` staging, `ldconfig`, and the shadowed `pip install /tmp/wheels/*.whl` are gone. `/opt/venv` is now the single install location for LC3.
- `backend/pusher/Dockerfile` — same cleanup; `git` is correctly dropped here because `backend/pusher/pylock.toml` has no VCS entries (only the backend runtime lock carries the PyOgg git pin), and `ENV LIBLC3_COMMIT` goes away with its consumer.
- `backend/scripts/runtime_image_contracts.py` — `docker_stages()` split out of `final_stage_copy_instructions()` (behavior-preserving), plus the new `final_stage_layer_errors()` that catches exactly the two defect shapes from #12345: a COPY whose destination is `rm`'d in a later layer (bytes still ship), and a `pip install` run while `PATH` leads at a venv copied in only afterward (install shadowed). It is wired into `check_source_closures()` so every registered runtime image inherits the guard.
- `backend/tests/unit/test_runtime_image_contracts.py` — the two new tests build minimal Dockerfiles reproducing each defect shape and assert the exact error text; I ran this file plus `test_verify_pusher_source_closure.py` locally on this head (17 passed), which also confirms no other registered image trips the new checks.
- `backend/tests/unit/test_verify_pusher_source_closure.py` — `test_liblc3_source_is_pinned` is removed together with the pinned source build it asserted; the remaining closure assertions (non-root user, writable working dirs) are intact.

Checks: Backend Hermetic Merge Gate, Hermetic Backend E2E, Listen Pusher Stack Gauntlet, Backend unit suite, and the source-closure contracts are green on this head; `image-smoke` is the documented fork-policy skip.

Nice work — this shrinks both images, removes a build-time network fetch, and the new registry-wide layer guard means this defect class can't silently regress. One note for the maintainer merge call: this swaps LC3 provenance in both deploy images from a source build of a pinned google/liblc3 commit to the hash-locked PyPI wheel, and the fork-skipped image-smoke means the production-base final image wasn't exercised in CI (the local public-base substitute build covers the branch). Both look sound here, but the deploy-image dependency-path change deserves a maintainer's eyes before merge.

---

aryanorastar:
## Final CI result

All applicable GitHub checks are green on exact head `f2df5c60af`.

- Backend unit suite: passed (15m27s)
- Listen Pusher Stack Gauntlet: passed (7m23s)
- Hermetic Backend E2E: passed
- Backend Hermetic Merge Gate: passed
- Sync Cloud Tasks Stack Gauntlet: passed
- Runtime image source closure: passed
- PR metadata, hygiene, formatting, and public API contract: passed

The image-smoke job is the documented fork-policy skip; both final `linux/amd64` images were built and smoked locally as recorded above.

---

*Updated 2026-08-31 — #12262 merged, and the two branches that had gone conflicting are refreshed. Editing in place rather than posting a new comment.*

@undivisible @kodjima33 ten PRs, all green and mergeable, none blocked on code.

**Ready for a click** — approved, no conflicts:
- #12287 — fences deferred conversation enrichment so two first-opens can't both run it (approved by @kodjima33)

**Need one approving review** — green, zero failing checks:
- #12380 (this one) — LC3 installed once from the locked wheel; drops a build-time network fetch and shrinks both images
- #12409 — one owner for app notification dispatch, replacing two duplicated rate limiters
- #12408 — structured-string sink registry, built from three real merged incidents
- #12406 — chat scroll ownership released by the input detector, not a wall-clock timer
- #12386 — pins the three Cursor MCP packages that currently float on `@latest`
- #11804 — transcript-driven proactivity, hard-off on every shipped bundle
- #11864 — withhold proactive notifications while someone else is present
- #11801 — hands-free wake word
- #11183 — X connector sync split onto its own Cloud Run Job

**Two things worth knowing, neither of which I can close myself:**

#12380 (this one) swaps LC3 provenance in both deploy images from a source build of a pinned `google/liblc3` commit to the hash-locked PyPI wheel, and `image-smoke` is skipped by fork policy, so the production-base final image was never exercised in CI. I built both `linux/amd64` images locally with the public base substituted and instantiated a real `lc3.Decoder` in each — that is not the production base.

#11183 has a gate no push can reach: CODEOWNERS routes `/.github/workflows/**` solely to @Git-on-my-level, and that account has never submitted an approving review — zero across the PRs it has reviewed.

aryanorastar:
@undivisible @kodjima33 bump — still green and mergeable. I updated the roll-up comment above in place rather than posting a new list, so the edit would not have notified anyone: #12262 merged, and #12287 is now approved, `CLEAN`, and needs only the click.

The LC3 provenance caveat on this one is unchanged and still worth a read before merging.

kodjima33 (approved):
Confidence 5/5: fixes #12345, root cause stated, scoped diff (171 lines), CI green, adds tests. Verified redundant liblc3 build still present on main (not already fixed). Merging.

PR #12386, opened 2026-08-29
fix(tooling): repair Cursor MCP configuration

### LINKED ISSUE #12376: All four entries in .cursor/mcp.json point at packages that don't exist (3x npm 404, 1 deprecated)
Not a security report — a broken-config report, with one thing in it worth a
second look.

I was surveying which MCP servers real projects actually install (I maintain
[mcpaudit](https://github.com/allenwu-blip/mcpaudit), an open-source static
scanner for MCP servers), which meant reading a lot of `mcp.json` files. Yours
stood out because all four entries point at something that isn't there.

### `.cursor/mcp.json` — every entry is broken

Checked directly against the npm registry just now:

| entry | package | npm status |
|---|---|---|
| `github` | `@modelcontextprotocol/server-github` | exists, **deprecated** — *"Package no longer supported"* |
| `notion` | `@modelcontextprotocol/server-notion` | **404 — never published** |
| `figma` | `@modelcontextprotocol/server-figma` | **404 — never published** |
| `browser` | `@modelcontextprotocol/server-browser` | **404 — never published** |

Reproduce:

```bash
npm view @modelcontextprotocol/server-notion   # E404
npm view @modelcontextprotocol/server-figma    # E404
npm view @modelcontextprotocol/server-browser  # E404
npm view @modelcontextprotocol/server-github   # deprecated
```

So a new contributor who opens the repo in Cursor gets three hard failures and
one deprecation warning. With 13k stars and 617 open issues I'd guess this has
quietly cost a few people an hour.

The three missing names look like they were guessed from the pattern of the
real ones (`server-filesystem`, `server-memory`, `server-sequential-thinking`
all do exist). Easy mistake — the naming is very regular right up until it
isn't.

### What actually exists

| you want | package | source |
|---|---|---|
| Notion | `@notionhq/notion-mcp-server` v2.5.1 | [makenotion/notion-mcp-server](https://github.com/makenotion/notion-mcp-server) — Notion's own |
| Figma | `figma-developer-mcp` v0.13.2 | [GLips/Figma-Context-MCP](https://github.com/GLips/Figma-Context-MCP) |
| Browser | `@playwright/mcp` v0.0.79 | [microsoft/playwright-mcp](https://github.com/microsoft/playwright-mcp) — Microsoft's |
| GitHub | — | [github/github-mcp-server](https://github.com/github/github-mcp-server), GitHub's official replacement for the deprecated npm package |

(`@modelcontextprotocol/server-puppeteer` is also deprecated now, in case it
shows up in another config.)

### The part worth a second look

Three of these entries hand credentials to a package that does not exist:

```json
"notion": {
  "command": "npx",
  "args": ["-y", "@modelcontextprotocol/server-notion"],
  "env": { "NOTION_API_KEY": "${env:NOTION_API_KEY}" }
}
```

`npx -y` resolves the name from the registry and runs whatever comes back, with
`-y` suppressing the install prompt. Today that fails, because nothing is
published under that name. The behaviour if something *were* ever published
there is that it runs, and this config hands it `NOTION_API_KEY` — and the
sibling entries hand out `FIGMA_ACCESS_TOKEN` and
`GITHUB_PERSONAL_ACCESS_TOKEN`.

**I want to be precise about what I am and am not claiming.** Whether anyone
other than the `@modelcontextprotocol` maintainers *can* publish under that
scope depends on npm scope ownership, and I could not determine that from
outside — I wrote a check for it, ran a control, found my own method was wrong
(it reported `@playwright` as unclaimed while `@playwright/mcp` plainly exists),
and threw the result away rather than report it. So: I am not telling you this
is exploitable. I am telling you there are three unclaimed-looking package names
in your repo wired to live credentials, that I could not rule the risk out, and
that the fix is the same one-line change you want anyway.

Pointing the entries at the packages that exist closes the question either way.

### Method

The npm results are from `registry.npmjs.org` directly; every row above is
reproducible with `npm view`. I have no affiliation with any package listed and
nothing to sell — I was mapping the ecosystem and this seemed worth passing on.

Happy to be told I've misread the config.

Allen Wu

aryanorastar:
## What changed and why

Fixes #12376.

The checked-in Cursor MCP configuration referenced three npm package identities that do not exist and one deprecated GitHub package. This replaces them with the providers' maintained remote MCP endpoints for GitHub, Notion, and Figma, and Microsoft's published Playwright MCP package for browser automation. GitHub credentials remain an environment reference, while Notion and Figma use Cursor's provider OAuth flow.

The browser package is pinned to `@playwright/mcp@0.0.79` so contributor setup does not execute a floating dependency. A deterministic repository contract now rejects the four old identities, insecure remote transport, floating npm versions, plaintext credentials, and forwarding repository credentials into browser automation.

none

## How it was verified

- `python3 .github/scripts/check_cursor_mcp_config.py` — checked-in configuration passed the new contract.
- `python3 .github/scripts/test_check_cursor_mcp_config.py` — 6 regression/mutation tests passed.
- `python3 .github/scripts/test_run_checks.py` — all 47 checks-manifest contract tests passed.
- `python3 -m black --check .github/scripts/check_cursor_mcp_config.py .github/scripts/test_check_cursor_mcp_config.py` — both Python files passed formatting.
- `npx -y @playwright/mcp@0.0.79 --help` — the exact configured package resolved and executed successfully.
- `npm view @playwright/mcp@0.0.79 version repository.url engines --json` — resolved version `0.0.79` from Microsoft's `playwright-mcp` repository with Node `>=18`.
- `npm view @modelcontextprotocol/server-github deprecated --json` — npm reports the old GitHub package as no longer supported.
- Unauthenticated reachability probes returned `401` for the GitHub and Notion endpoints and `405` for a GET to Figma's POST transport, proving all three configured hosts resolve and enforce their expected protocol/authentication boundary.
- `git diff --check` — no whitespace errors.

Provider OAuth and an authenticated GitHub tool call were not completed locally because that would require contributor credentials. The checked-in credential interpolation and exact provider contracts are covered by the regression suite.

## Tests

`.github/scripts/test_check_cursor_mcp_config.py` starts from the real checked-in config, proves it passes, and mutates it through every reported package identity plus the adjacent supply-chain and credential-boundary failure shapes. Each mutation fails through the production checker registered in both local and CI lanes.

## Failure class (fixes)

Failure-Class: none

## New guards

Issue #12376 is the concrete incident: all four project MCP entries were unusable, including credential-bearing `npx -y` invocations against nonexistent package names. A narrow config contract is the shared primitive for this repository-owned JSON boundary because the external providers do not share a runtime API or package resolver that Omi can wrap.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12386?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aryanorastar:
## Verification proof

Verified commit: `0e83bd1540`

### Reported failure reproduced

The old project config contained these four package identities:

| Entry | Registry result |
| --- | --- |
| `@modelcontextprotocol/server-github` | Resolves, but npm marks it **no longer supported** |
| `@modelcontextprotocol/server-notion` | npm `E404` |
| `@modelcontextprotocol/server-figma` | npm `E404` |
| `@modelcontextprotocol/server-browser` | npm `E404` |

### Replacement exercised

- `npx -y @playwright/mcp@0.0.79 --help` resolved and ran the exact package now recorded in `.cursor/mcp.json`.
- npm metadata identifies that package as version `0.0.79`, sourced from Microsoft's `playwright-mcp` repository, requiring Node `>=18`.
- The maintained GitHub and Notion endpoints answered unauthenticated probes with `401`, proving the hosts and authentication boundaries are live.
- Figma's MCP endpoint answered a GET probe with `405`, as expected for its POST MCP transport.

I did **not** claim an authenticated tool call because doing that requires a contributor's private GitHub/provider OAuth credentials.

### Regression boundary

The new test suite runs against the real checked-in file, then proves rejection of:

1. each of the four package identities from #12376;
2. a floating `@latest` browser dependency;
3. an HTTPS-to-HTTP endpoint downgrade;
4. a plaintext authorization value; and
5. repository credentials forwarded into the browser MCP child process.

### Commands and results

```text
python3 .github/scripts/check_cursor_mcp_config.py
Cursor MCP config contract passed.

python3 .github/scripts/test_check_cursor_mcp_config.py
Ran 6 tests ... OK

python3 .github/scripts/test_run_checks.py
Ran 47 tests ... OK

OMI_PR_BODY_FILE=/private/tmp/omi-pr-12376.md make preflight
PR preflight passed: 107 checks

OMI_PR_BODY_FILE=/private/tmp/omi-pr-12376.md git push ...
Bounded pre-push gate passed
```

The checker and its mutation suite are registered in both local and CI lanes, so future edits cannot silently restore the broken/deprecated identities or weaken the credential boundary.

Git-on-my-level:
Verified against #12376 — solid repair, and the contract script is a genuinely useful guardrail. What I checked on 0e83bd1:

- `.cursor/mcp.json` — the four dead entries (`@modelcontextprotocol/server-notion` / `-figma` / `-browser` are npm 404s, `-github` is deprecated) are replaced with the providers' own remote endpoints (`api.githubcopilot.com/mcp/`, `mcp.notion.com/mcp`, `mcp.figma.com/mcp` — all resolve and enforce auth, as expected) plus Microsoft's `@playwright/mcp` pinned at `0.0.79`, which is the currently published latest. No committed credentials: the only secret-shaped value left is `Bearer ${env:GITHUB_TOKEN}`, an environment reference.
- `.github/scripts/check_cursor_mcp_config.py` — clean stdlib-only validator; rejects the four banned identities, non-HTTPS transports, floating npx specs, plaintext credential values in `env`/`headers`, and any credential forwarding to the browser server. No network calls, no subprocesses, nothing executes at review time.
- `.github/scripts/test_check_cursor_mcp_config.py` — ran both scripts locally: contract passes, 6/6 mutation tests green (including the regression cases for each identity from the issue).
- `.github/checks-manifest.yaml` — the two new entries (`cursor-mcp-config`, `cursor-mcp-config-tests`) match the existing schema and are consumed through the preflight/pre-push lanes (PR Metadata Preflight is green on this head).
- `docs/doc/developer/Cursor.mdx` — the added paragraph matches the new config and keeps the existing "don't put secrets in `.cursor/`" rule; it also steers coding agents away from guessing `@modelcontextprotocol/server-*` names, which is accurate and now backed by the contract script.

One note, not a blocker: since the manifest entries run via the preflight/pre-push lanes rather than a dedicated workflow, the contract doesn't surface as a named check-run on the PR page. If you'd want it visible as a required status check, that's a small follow-up for maintainers to wire up.

Because this PR rewrites the checked-in MCP server configuration and CI check manifest that maintainers and coding agents rely on, I'm leaving the formal sign-off to a human maintainer — everything I verified is consistent and ready from my side.

---

Git-on-my-level:
Independently re-verified against the npm registry and the providers' published endpoints - this repair holds up. Thank you for the thorough verification appendix.

What I checked on `0e83bd1`:

- `.cursor/mcp.json` - confirmed the old entries were unrunnable: `@modelcontextprotocol/server-notion`, `-figma`, and `-browser` return npm `E404` (never published) and `@modelcontextprotocol/server-github` is marked no-longer-supported. The replacements are the providers' own hosted endpoints (`api.githubcopilot.com/mcp/`, `mcp.notion.com/mcp`, `mcp.figma.com/mcp` - all resolve and enforce auth), and `@playwright/mcp@0.0.79` is Microsoft's currently published latest. `GITHUB_TOKEN` stays an env reference, and the browser entry no longer receives any env.
- `.github/scripts/check_cursor_mcp_config.py` - ran it locally: passes. The contract logic is sound (banned identity list, exact-semver pinning, HTTPS-only remotes, `${env:...}`-only for sensitive keys, exact provider contracts, no browser env), and making `REMOTE_CONTRACTS` an equality check means any endpoint drift becomes a deliberate contract update - good fail-closed design.
- `.github/scripts/test_check_cursor_mcp_config.py` - 6/6 pass locally; the mutation tests genuinely exercise each rejection path rather than only the happy path.
- `.github/checks-manifest.yaml` - both new entries are well-formed with correct trigger paths. One observation for maintainers: per `repo-checks.yml`, ci-lane manifest checks execute on main pushes (and locally), not on PRs - so this guard catches regressions after merge rather than before. That matches the existing manifest design, just worth knowing.
- `docs/doc/developer/Cursor.mdx` - the new paragraph is accurate (Notion documents exactly this URL plus OAuth flow; Figma's remote server uses the same connect flow), and since this page is agent-facing guidance it also steers AI coding assistants away from reintroducing guessed package names or floating versions. Verified safe and accurate against npm and provider docs.

This also retires the credential-exposure shape flagged in #12376: phantom `npx -y` targets that were one package-squat away from receiving `NOTION_API_KEY` / `FIGMA_ACCESS_TOKEN` / `GITHUB_PERSONAL_ACCESS_TOKEN`.

Positive signal rather than an approval: the `.github/` manifest/CI surface and the agent-facing doc guidance should get a human maintainer sign-off before merge.

---
Automated review on behalf of @Git-on-my-level - AI-generated feedback.

---

aryanorastar:
Reworked on `58f6fa58`, and the shape of this PR changed — worth reading before re-reviewing, because it no longer proposes what you verified.

**Main fixed #12376 first.** `8b5715527b` ("point the MCP config at packages that exist") landed the four dead entries onto published npm packages, and the issue is closed. My endpoint layout is superseded, so I dropped it: `.cursor/mcp.json` here is now main's config, changed only by pinning.

Resolving the conflict the other way would have reverted a merged fix. I ran my own checker against main's config to be sure of that rather than assuming — it failed on 7 counts, 4 of which were purely my design opinion.

**What survives is the part main's fix doesn't cover.** All three of its entries run floating `npx -y` specs, and two receive credentials:

```
@notionhq/notion-mcp-server    NOTION_TOKEN
figma-developer-mcp            FIGMA_API_KEY
@playwright/mcp@latest         (no env)
```

That's the same supply-chain shape #12376 was about — an unpinned name one squat away from a credential — except the names are real now. Pinned to `2.5.1`, `0.13.2`, `0.0.79`.

**The checker is narrowed to what's design-neutral**: the four never-published `@modelcontextprotocol/server-*` identities, exact-version pinning, HTTPS-only for url-form servers, `${env:NAME}` for credential-shaped values, and no env on the browser server. `REMOTE_CONTRACTS` and the pinned-playwright identity rule are gone — those encoded my endpoint choice, and keeping them would mean the guard enforces a design main didn't pick.

Verified against both defect shapes, not just the happy path:

```
main's current config    → 3 errors, all "must pin its npx package"
the pre-8b5715527b file  → all four banned identities rejected
the config here          → passes
```

Tests 6/6, checks-manifest contract 47/47, formatter clean, PR preflight 108/108.

Your note about ci-lane manifest checks running on main pushes rather than PRs still stands and I haven't changed that — it's the existing manifest design, and wiring a named status check is a maintainer call.

Git-on-my-level:
Re-verified on `e4419758` — the pin-only rework after main's `8b5715527b` landed #12376 first. Taking main's config and narrowing the checker to design-neutral rules was the right resolution of the conflict; the endpoint layout this PR originally carried was already superseded.

What I checked on this head:

- `.cursor/mcp.json` — exactly main's post-`8b5715527b` config plus exact pins. All three pinned specs resolve on npm to the currently published latest from the official sources (`@notionhq/notion-mcp-server@2.5.1` → makenotion/notion-mcp-server, `figma-developer-mcp@0.13.2` → GLips/Figma-Context-MCP, `@playwright/mcp@0.0.79` → microsoft/playwright-mcp). Credentials stay `${env:...}` references and the browser entry still receives no env, so the credential-squatting shape flagged in #12376 stays retired.
- `.github/scripts/check_cursor_mcp_config.py` — ran it against the head config: passes. The contract is offline (no registry access), fail-closed, and design-neutral: banned never-published identities (I confirmed the three still return npm 404 and `-github` is deprecated), exact-semver pinning, HTTPS-only `url` servers, `${env:NAME}`-only for credential-shaped keys, and no env on the browser server. Dropping `REMOTE_CONTRACTS` from the earlier iteration was the correct call — the guard no longer encodes a design main didn't pick. One forward-looking nit, not blocking: `SENSITIVE_NAME` matches AUTHORIZATION/KEY/SECRET/TOKEN but not PASSWORD/PASSPHRASE, so a hypothetical future key named e.g. `API_PASSWORD` would escape the `${env:}` requirement (the browser-env backstop still covers the browser server).
- `.github/scripts/test_check_cursor_mcp_config.py` — 6/6 pass locally. The mutations genuinely exercise each rejection path, including re-adding a removed `github` server to prove the banned list isn't keyed to shipped servers only.
- `.github/checks-manifest.yaml` — both entries load via `run_checks.py` with correct `local`+`ci` lanes and trigger paths, and manifest validation reports nothing against them. The pre-existing caveat stands (not introduced here): per `repo-checks.yml`, ci-lane manifest checks execute on main pushes and locally, not on PR events — so this guard catches a regression after merge rather than in this PR's own CI. That's the existing manifest design; wiring a named PR status check would be a separate maintainer decision, as you noted.
- `docs/doc/developer/Cursor.mdx` — the new paragraph is accurate against the checker's actual behavior (four banned names, exact pinning, `${env:}` references, browser without credentials, deliberate bumps over `@latest`). Since this page is agent-facing guidance, it also steers AI coding assistants working in this repo away from reintroducing guessed package names or floating specs — a welcome nudge, and it matches what the contract actually enforces.

No correctness, security, or supply-chain issues found on this head. Formal sign-off stays with a human maintainer: the `.github/` manifest/CI surface and the agent-facing doc guidance deserve one conscious look before merge, which `workflow-review` already tracks.

---

kodjima33:
Thank you so much for this 🙏 — looks like the core issue (nonexistent MCP packages) was already fixed on `main` in 8b5715527b ("fix(cursor): point the MCP config at packages that exist"). Closing this one, but please keep them coming!

aryanorastar:
Understood on the close, and `8b5715527b` did fix the original issue — I confirmed that myself and threw away this PR's endpoint design because of it.

The title is what misleads here, and that is on me for not changing it after the rework. What is left is a different, still-open gap. On `main` right now:

```json
"notion":  ["-y", "@notionhq/notion-mcp-server"],   // NOTION_TOKEN
"figma":   ["-y", "figma-developer-mcp", "--stdio"], // FIGMA_API_KEY
"browser": ["-y", "@playwright/mcp@latest"]
```

Three floating `npx -y` specs, two of them handed a credential. `npx -y` resolves and executes whatever the registry serves at run time, so a compromised or hijacked release of either package reads a contributor's Notion or Figma token with no version change in this repo. That is the same exposure shape as #12376 — the difference is the names are real now, which makes it quieter, not safer.

The final diff was three exact pins plus an offline, fail-closed contract check, and nothing else. `8b5715527b` does not cover it.

Happy to open a fresh, correctly-titled PR for just the pinning if that is easier to review than reopening this one — your call, and no argument either way if you would rather leave it.

PR #12406, opened 2026-08-30
fix(desktop): keep chat scroll ownership until input ends

aryanorastar:
## What changed and why

Fixes the `ChatTranscriptGestureHarnessTests.testSendStartingWhileTheReaderScrollsDoesNotSeizeTheViewport` flake reported in #12039. The transcript released reader ownership after a fixed 300 ms even when AppKit's live-scroll transaction was still active, so a delayed `isSending` update could snap the reader to the live edge under CI load.

- `UserScrollDetector` now publishes a single end-of-input callback from AppKit's native live-scroll completion, with the existing bounded fallback retained for keyboard, scrollbar, and unphased wheel input
- both the main transcript and shared floating/notch scroll container release `userIsScrolling` only from that callback
- the mounted transcript harness now emits phased trackpad events and holds the gesture open beyond the former timeout, making the CI race deterministic

This addresses only the chat-transcript occurrence in #12039. The separate `KernelTurnRecordedProjectionTests` and `SuggestedTasksStoreTests` causes remain open.

- INV-CHAT-1 — no transcript/session ownership or persistence changes; the shared transcript surfaces continue using the same scroll primitive
- INV-CHAT-2 — reader movement remains authoritative until the platform reports that input ended

## How it was verified

- Red proof: with the old 300 ms expiry reintroduced against the strengthened regression, the viewport jumped to `4713` instead of preserving `3513`
- Green proof: `ChatTranscriptGestureHarnessTests | UserScrollDetectorTests | MarkdownNestedScrollTests` — 23 tests, 0 failures
- Stress proof: exact former flake executed 10 consecutive times through the built XCTest bundle — 10/10 passed
- `python3 scripts/check_desktop_test_quality.py` — passed at the existing ratchet (54 source-reading files / 147 sites, 16 waits)
- `./scripts/swiftlint-wrapper.sh lint` — 0 violations across 1,444 files
- changed-file pinned `swift-format` lint and `git diff --check` — passed
- `./scripts/omi-macos-dev doctor` — passed; no active local bundle conflicts

No iOS, Android, BLE, or physical-device path is involved. The behavioral proof mounts the real macOS `ChatMessagesView` in `NSHostingView` and drives its AppKit scroll lifecycle.

## Tests

- `ChatTranscriptGestureHarnessTests.testSendStartingWhileTheReaderScrollsDoesNotSeizeTheViewport` now keeps the native gesture open beyond 300 ms and verifies both the pre-send and post-send viewport
- `UserScrollDetectorTests.testRapidLiveScrollHandsOwnershipToReaderAndDoesNotRearmAwayFromBottom` verifies ownership is not released before `didEndLiveScroll` and is released exactly when AppKit ends the gesture

## Failure class (fixes)

Failure-Class: new

Adds `FC-wall-clock-gesture-ownership`.

- Violated contract: a native viewport gesture owns the transcript until the platform says the input ended; a wall-clock timeout cannot revoke that ownership while momentum or a live-scroll transaction is active
- Canonical prevention: centralize start/end ownership in `UserScrollDetector`, use AppKit completion for native gestures, and reserve bounded timers for input paths without a native lifecycle
- Guard artifacts: `ChatScrollLiveEdgeTests.swift` pins the coordinator lifecycle; `ChatTranscriptGestureHarnessTests.swift` pins the mounted SwiftUI/AppKit race that failed in CI run 32542769839

The mounted regression would have caught the real #12039 / run 32542769839 failure deterministically. This is implemented as a shared detector primitive plus behavioral contract tests, not as a new static check or ratchet.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12406?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-wall-clock-gesture-ownership.json
+++ b/.github/failure-classes/FC-wall-clock-gesture-ownership.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": 1,
+  "id": "FC-wall-clock-gesture-ownership",
+  "violated_contract": "A live viewport gesture owns the transcript until the platform reports that input has ended; reader ownership must not expire on a fixed wall-clock delay while trackpad momentum or another native scroll transaction is still active.",
+  "canonical_prevention": "Centralize input ownership in the AppKit scroll detector, emit an end callback from native live-scroll completion or the bounded fallback for inputs without a native lifecycle, and make every auto-following chat surface release its active-scroll state only from that callback.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/Desktop/Tests/ChatScrollLiveEdgeTests.swift",
+    "desktop/macos/Desktop/Tests/ChatTranscriptGestureHarnessTests.swift"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatMessagesView.swift",
+    "desktop/macos/Desktop/Sources/MainWindow/Components/ChatScrollBehavior.swift"
+  ],
+  "status": "open"
+}
--- a/desktop/macos/Desktop/Sources/MainWindow/Components/ChatMessagesView.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Components/ChatMessagesView.swift
@@ -398,7 +398,6 @@ struct ChatMessagesView<WelcomeContent: View>: View {
   /// Set immediately by the scroll wheel monitor to win the race against
   /// throttled programmatic scrolls during streaming.
   @State private var userIsScrolling = false
-  @State private var userScrollEndWorkItem: DispatchWorkItem?
   /// Tracks work items for delayed initial bottom scrolls so they can be
   /// canceled on user scroll or disappear.
   @State private var initialScrollWorkItems: [DispatchWorkItem] = []
@@ -890,7 +889,7 @@ struct ChatMessagesView<WelcomeContent: View>: View {
     }
   }
 
-  /// Cancels all pending scheduled scrolls (throttle, initial, user-scroll-end).
+  /// Cancels all pending scheduled scrolls (throttle and initial placement).
   private func cancelAllPendingScrolls() {
     scrollThrottleWorkItem?.cancel()
     scrollThrottleWorkItem = nil
@@ -900,8 +899,6 @@ struct ChatMessagesView<WelcomeContent: View>: View {
     // when it changes: this runs on every scroll event, and an unconditional
     // `@State` write there is a body invalidation per event.
     if hasQueuedFollowScroll { hasQueuedFollowScroll = false }
-    userScrollEndWorkItem?.cancel()
-    userScrollEndWorkItem = nil
     for item in initialScrollWorkItems {
       item.cancel()
     }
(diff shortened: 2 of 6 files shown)


Git-on-my-level:
Thanks @aryanorastar — this is a clean root-cause fix for the #12039 chat-transcript flake, and the red/green/stress verification writeup is exactly the standard we like to see.

Per-file notes:

- `desktop/macos/Desktop/Sources/MainWindow/Components/ChatScrollBehavior.swift` — the new `onUserScrollEnded` callback is emitted from all three genuine end-of-input paths (`finishUserScroll` after `didEndLiveScroll`, `scheduleUnphasedWheelEnd` for classic-wheel events, `scheduleDiscreteInputSettledBottomCheck` for keyboard/scrollbar), and calling it before the `isAtBottom` guard correctly releases ownership even when the reader stopped away from the bottom. `beginUserScroll()` still cancels any pending `settleWorkItem`, so a straggler end-callback cannot clobber a newly started gesture. The `nativeLiveScrollIsActive` gate on `scheduleUnphasedWheelEnd` is the right fix for the fallback race where the unphased timer could expire inside an open live-scroll transaction. `ChatScrollContainer` in this file gets the same migration as the main transcript, so both shared scroll surfaces now release `userIsScrolling` only from the detector.
- `desktop/macos/Desktop/Sources/MainWindow/Components/ChatMessagesView.swift` — dropping the 0.3 s `userScrollEndWorkItem` latch removes the wall-clock ownership expiry entirely, and `cancelAllPendingScrolls()` shrinking to throttle + initial placement matches that. The rewritten comment on `onScrollSettledAtBottom` is accurate about the new ordering.
- `desktop/macos/Desktop/Tests/ChatScrollLiveEdgeTests.swift` — the added `userScrollEnds == 0` assertion before posting `didEndLiveScrollNotification` pins the exact mid-transaction invariant, and `== 1` after pins single delivery. The defaulted closure keeps this test file's existing constructor call sites compiling unchanged.
- `desktop/macos/Desktop/Tests/ChatTranscriptGestureHarnessTests.swift` — holding the gesture open with `harness.pump(0.4)` past the old 300 ms window makes the former CI race deterministic, and tagging the synthetic CGEvents with `scrollWheelEventScrollPhase = .changed` so they read as phased trackpad updates (and do not arm the classic-wheel fallback) mirrors real AppKit delivery.
- `.github/failure-classes/FC-wall-clock-gesture-ownership.json` — follows the existing failure-class schema and names both guard artifacts; the violated-contract text matches what the code now enforces.
- `desktop/macos/changelog/unreleased/20260829-chat-scroll-ownership.json` — user-facing wording matches the actual behavior change.

One documentation nit, non-blocking: `FC-wall-clock-gesture-ownership.json` has `"evidence_prs": []` while other FC entries cite the incident that established them (e.g. `FC-day-bucket-key-recomputed` cites 10398). Consider citing #12039 (or this PR once merged) so future failure-class sweeps can trace the class back to a concrete incident.

Overall a positive signal: cohesive single-concern fix, deterministic regression coverage on the exact race that failed in CI, and a precise failure-class record. Merging this should close the chat-transcript leg of #12039 while leaving the `KernelTurnRecordedProjectionTests` and `SuggestedTasksStoreTests` causes open, as the description states.

---

aryanorastar:
On the `evidence_prs` nit — leaving it `[]` is deliberate, and I think citing #12039 would actually be wrong on two counts.

`.github/PULL_REQUEST_TEMPLATE.md` says it directly for a class declared in its own PR:

> Its `evidence_prs` may be `[]` — this PR is the evidence, and its number does not exist yet.

And #12039 is an issue, not a PR, so it would mistype the field. `FC-day-bucket-key-recomputed` cites 10398 because that class was extracted from an already-merged fix; this one is being declared alongside its first fix, which is the case the template carves out. The traceability you want is in `canonical_prevention_artifact`, which names both guard tests.

Happy to add the number in a follow-up registry-only PR once this merges, if you would rather have it explicit. @Git-on-my-level @undivisible

kodjima33 (approved):
Owner-override: verified fix(desktop) — replaces wall-clock 0.3s scroll-release timer with native didEndLiveScroll callback; confirmed the old 0.3s timer is still on main, CI green, no abuse signals. Merging.

Issue #10337, opened 2026-07-22
Close the structured-string sink class: registry + fixtures, then extract the shared deploy body


Git-on-my-level:
## Problem

Three registered failure classes are the same failure wearing different clothes: **a structured value flattened into a string, then re-parsed by something with a different grammar.**

- **`FC-deploy-input-serialization`** (#9986) — `google-github-actions/deploy-cloudrun@v3` joins env vars into one comma-delimited string. `STT_PRERECORDED_MODEL=parakeet,modulate-velma-2` became `STT_PRERECORDED_MODEL=parakeet` plus a bogus empty key `modulate-velma-2=`. This blocked **both dev and prod backend deploys** and produced a cross-plane mismatch, because GKE received the intended literal while Cloud Run did not.
- **`FC-workflow-script-input-contract`** (#9945/#9947) — wrong argv grammar passed to a shell script.
- **`FC-workflow-executable-assumption`** (#9943) — exit 126 invoking a `100644` file.

The substrate that makes this class possible: **66 workflows, 9,547 lines, with 2,951 lines of inline `run:` shell, 1,724 `${{ }}` interpolations, and 151 `if:` conditions.** Still raw in YAML: 76 `gcloud`, 48 `kubectl`, 30 `gh`, 22 `jq`, 13 `helm` invocations.

We have already paid for this class three times. The honest scope, though, is narrower than it first looks: only ~4% of `fix:` PRs are substrate-caused, and **46% of sampled deploy failures are genuine cloud semantics** (gcloud's literal-vs-secret typing, Helm strategic merge, Cloud Run's optional `percent`) that survive any rewrite. So this issue is deliberately *not* "rewrite the release system."

There is also a tell worth naming: **48 files / 9,181 lines of tests that statically parse workflow YAML** (`backend/tests/unit/test_workflow_contracts.py` and siblings). That is a hand-built type checker over an untyped substrate — precisely the "static tripwire, not behavioral coverage" that `AGENTS.md` warns against. It exists because the substrate cannot check itself.

## What to build

**Two PRs. Do not combine them.**

### PR 1 — A structured-string sink registry

A deterministic checker over a **finite, explicit registry** of known structured-string sinks. Not a general lint.

Each registry entry declares: the sink (an action input like `deploy-cloudrun`'s `env_vars`, or a CLI flag like `gcloud run deploy --update-env-vars`), its delimiter grammar, and the safe encoding.

[...]

PR #12408, opened 2026-08-30
ci: add structured-string sink registry

aryanorastar:
## What changed and why

Adds PR 1 from #10337: a finite, explicit registry for the three structured-string workflow sinks that already blocked deploys.

The checker is intentionally not a general workflow/dataflow lint. Each JSON entry names one real sink, its grammar, its safe encoding, its historical evidence PRs, and a paired passing/failing fixture:

| Registered sink | Historical bad shape | Required safe shape | Evidence |
| --- | --- | --- | --- |
| `deploy-cloudrun@v3` `env_vars` | `STT_PRERECORDED_MODEL=parakeet,modulate-velma-2` | escape the comma as `\,` before flattening | #9986 |
| `probe-llm-gateway-from-cloud-run.sh` argv | `--project=test-project` | `--project test-project` | #9945, #9947 |
| the same checked-in `100644` probe script | direct execution | explicit `bash` invocation | #9943, #9945 |

The manifest runs both the production scan and its mutation/schema suite in `local` and `ci` lanes. Workflow or composite-action changes select the production scan; registry/checker/fixture changes also select the mutation suite.

This is only the registry/checker PR required by #10337. It does not extract or merge the backend deploy workflows; that remains the separately reviewable PR 2 from the issue.

## Guard design

- Three sharded JSON entries are the finite source of truth; adding a new sink requires naming its exact grammar and evidence.
- Every entry must reference both fixture polarities, and the test suite proves every historical bad fixture produces a violation.
- A registered scope that still matches files but zero sink invocations fails closed, so renaming or retiring a sink cannot make the checker silently green.
- Bare path allowlists are rejected. The only unresolved-value boundary—GitHub expressions inside `env_vars`—names the backend runtime serializers/workflow contract tests as owner and explains why the literal fragment remains checkable here.
- Literal action inputs are parsed as YAML, while registered shell calls are tokenized after masking GitHub expressions. The checker only evaluates the explicitly registered action input and script basename.

None. This changes deterministic repository checks and fixtures only; no runtime or product behavior changes.

## How it was verified

- `python3 .github/scripts/test_check_structured_string_sinks.py` — 7 tests passed
- `python3 .github/scripts/check_structured_string_sinks.py` — all 3 registered production sinks passed
- Historical negative controls produced one actionable violation each:
  - unescaped Cloud Run comma
  - joined `--project=<value>` probe argument
  - probe invocation without `bash`
- `python3 .github/scripts/run_checks.py --lane local --check-id structured-string-sink-registry --check-id structured-string-sink-registry-tests` — both manifest checks passed
- `python3 .github/scripts/test_run_checks.py` — 47 manifest contract tests passed
- `python3 -m py_compile ...`, Black, and `git diff --check` — passed

## Failure class

No new failure class. This is the canonical shared guard requested for three existing classes:

- `FC-deploy-input-serialization`
- `FC-workflow-script-input-contract`
- `FC-workflow-executable-assumption`

## New guards

The guard would have rejected the actual inputs behind #9986, #9947, and #9945 before a deploy run. A static checker is appropriate here because the violated contracts are repository-authored YAML/shell serialization grammars; the paired fixtures provide mutation-sensitive behavioral proof that each registered rule can fail rather than merely matching source text.

Part of #10337.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -108,13 +108,13 @@ checks:
     lanes: ["local", "ci"]
     reason: "base-derived oversized-file checks must reject growth and stale approvals while ratcheting reductions down automatically"
   - id: structured-string-sink-registry
-    command: ["python3", ".github/scripts/check_structured_string_sinks.py"]
-    triggers: [".github/workflows/**", ".github/actions/**", ".github/structured-string-sinks/**", ".github/scripts/check_structured_string_sinks.py", ".github/scripts/fixtures/structured_string_sinks/**", ".github/checks-manifest.yaml"]
+    command: ["bash", "scripts/run-structured-string-sinks.sh"]
+    triggers: [".github/workflows/**", ".github/actions/**", ".github/structured-string-sinks/**", ".github/scripts/check_structured_string_sinks.py", ".github/scripts/fixtures/structured_string_sinks/**", "scripts/run-structured-string-sinks.sh", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "#10337 registers the three delimiter, argv, and executable workflow sinks that blocked deploys in #9986, #9947, and #9945"
   - id: structured-string-sink-registry-tests
-    command: ["python3", ".github/scripts/test_check_structured_string_sinks.py"]
-    triggers: [".github/structured-string-sinks/**", ".github/scripts/check_structured_string_sinks.py", ".github/scripts/test_check_structured_string_sinks.py", ".github/scripts/fixtures/structured_string_sinks/**", ".github/checks-manifest.yaml"]
+    command: ["bash", "scripts/run-structured-string-sink-tests.sh"]
+    triggers: [".github/structured-string-sinks/**", ".github/scripts/check_structured_string_sinks.py", ".github/scripts/test_check_structured_string_sinks.py", ".github/scripts/fixtures/structured_string_sinks/**", "scripts/run-structured-string-sink-tests.sh", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "every registered sink must keep a paired historical mutation fixture proving the checker rejects the incident grammar"
   - id: desktop-release-manifest-v1
--- a/.github/checks-manifest.yaml
+++ b/.github/checks-manifest.yaml
@@ -107,6 +107,16 @@ checks:
     triggers: [".github/scripts/check_product_file_line_count_ratchet.py", ".github/scripts/test_check_product_file_line_count_ratchet.py", ".github/checks-manifest.yaml"]
     lanes: ["local", "ci"]
     reason: "base-derived oversized-file checks must reject growth and stale approvals while ratcheting reductions down automatically"
+  - id: structured-string-sink-registry
+    command: ["python3", ".github/scripts/check_structured_string_sinks.py"]
+    triggers: [".github/workflows/**", ".github/actions/**", ".github/structured-string-sinks/**", ".github/scripts/check_structured_string_sinks.py", ".github/scripts/fixtures/structured_string_sinks/**", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "#10337 registers the three delimiter, argv, and executable workflow sinks that blocked deploys in #9986, #9947, and #9945"
+  - id: structured-string-sink-registry-tests
+    command: ["python3", ".github/scripts/test_check_structured_string_sinks.py"]
+    triggers: [".github/structured-string-sinks/**", ".github/scripts/check_structured_string_sinks.py", ".github/scripts/test_check_structured_string_sinks.py", ".github/scripts/fixtures/structured_string_sinks/**", ".github/checks-manifest.yaml"]
+    lanes: ["local", "ci"]
+    reason: "every registered sink must keep a paired historical mutation fixture proving the checker rejects the incident grammar"
   - id: desktop-release-manifest-v1
     command: ["python3", ".github/scripts/test_desktop_release_manifest.py"]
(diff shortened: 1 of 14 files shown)


Git-on-my-level:
Thanks — this is a carefully built first half of #10337, and it holds up under execution. I verified it against the PR head in a clean container:

- `.github/scripts/check_structured_string_sinks.py` — runs green on the head tree (`OK: 3 registered structured-string sinks are grammar-safe`). The design choices match the issue's reward-hacking guardrails: bare `allowlist` keys are rejected outright in `validate_entry`, `dynamic_values` exemptions must name an owner and a substantive reason, and `check_repository` fails closed when a scope matches files but zero sink invocations.
- `.github/scripts/test_check_structured_string_sinks.py` — all 7 tests pass, including the mutation requirement (each historical fail fixture must yield exactly one violation). I also reproduced all three incident grammars independently as violations: the #9986 `STT_PRERECORDED_MODEL=parakeet,modulate-velma-2` unescaped comma, the #9947 `--project=x` joined argv, and the #9943 direct-execution shape. `sh probe-...sh` is correctly rejected as well (interpreters: ["bash"]).
- `.github/structured-string-sinks/*.json` (3 entries) — the evidence PRs 9986/9945/9943/9947 all check out as real merged incidents matching the claimed failure classes; every entry pairs both fixture polarities.
- `.github/scripts/fixtures/structured_string_sinks/*` (6 fixtures) — fail fixtures carry the historical inputs verbatim; pass fixtures demonstrate each safe encoding (`\,` escape, split argv, explicit bash).
- `.github/checks-manifest.yaml` — both new check IDs resolve inside the 157-entry manifest and execute via `run_checks.py --lane ci` in repo-checks.yml; local+ci lanes match the issue's "never as a workflow step" requirement.
- `scripts/run-structured-string-sinks.sh` / `scripts/run-structured-string-sink-tests.sh` — canonical-venv resolution through the existing `dev-harness/_resolve_python.sh`, uv sync fallback from the checked-in lock; same pattern as the apt-network-bounds guard. No network beyond locked deps, no secrets touched.

One non-blocking hardening note for the registry's future: a nested payload like `bash -c 'probe-...sh --project=x'` currently evades the split-argv check (shlex sees one argument). That's consistent with the declared "not a general shell linter" scope, but worth remembering if the probe script grows more call sites.

Since this adds new enforcement that will gate future deploy-affecting PRs in both lanes, it should get maintainer sign-off before merge — the implementation tracks the #10337 spec closely, so that should be quick.

---

aryanorastar:
@undivisible @kodjima33 green and mergeable, reviewed two days ago and verified in a clean container — nothing blocking.

The one thing needing a human: it adds enforcement that gates future deploy-affecting PRs in both lanes. Every registry entry is built from a real merged incident (#9986, #9947, #9943, #9945), so the check has an audience, but turning it on is a maintainer call.

kodjima33 (approved):
New preventive CI guard (not an active bug fix — the 3 historical incidents it guards against are already resolved). Verified the checks-manifest/run_checks infra it extends is an established, pull_request-triggered, read-only pattern (no secrets exposure). Approve-only per policy (feature, not fix).

aryanorastar:
its greeen and approved @Git-on-my-level @kodjima33

kodjima33 (approved):
ci: adds finite, well-tested structured-string sink registry from a trusted contributor (aryanorastar). Feature/tooling, not a bug fix, CI green — approve per confidence gate (feature stays approve-only).

aryanorastar:
Recording the checker's scope boundary on the PR that introduces it, since this turns on enforcement that will gate other people's deploy-affecting PRs in both lanes. Not asking for anything — this is for whoever hits the check later and needs to know what it does and does not claim.

**What it covers.** Three registered sinks, each a real merged incident with both fixture polarities:

```
deploy-cloudrun-env-vars-v3.json           comma in a Cloud Run runtime value must be \, escaped
llm-gateway-probe-split-argv.json          --project <value>, never --project=<value>
llm-gateway-probe-explicit-interpreter.json   invoked via bash, not directly or via sh
```

**What it does not.** It is a registry, not a shell or dataflow linter, and that boundary is load-bearing rather than incidental — the value of a finite registry is that adding a sink forces someone to name its exact grammar and cite the incident it came from. Two consequences worth stating plainly:

- A registered sink wrapped in `bash -c '...'` evades every rule, because `shlex` sees the payload as one argument. That is the gap the review pass flagged, and it is tracked as #12860 rather than left implicit. It is not fixed here: closing it means either parsing nested shell payloads or registering `bash -c` as its own sink shape, and both are a different change from the registry this PR is scoped to.
- GitHub expressions inside `env_vars` remain unresolved values. The entry names the backend runtime serializers and workflow contract tests as the owner of that boundary rather than silently ignoring it.

**The failure modes the registry deliberately closes** are the ones that would make a green check meaningless: bare path allowlists are rejected outright, `dynamic_values` exemptions must name an owner and a substantive reason, and a registered scope that matches files but zero sink invocations fails closed — so retiring or renaming a sink cannot quietly turn its rule into a no-op.

If a future deploy PR trips this check, the entry's `evidence` field points at the incident whose shape it is reproducing, which is usually faster than reading the checker.

aryanorastar:
@Git-on-my-level @kodjima33 approved and mergeable, unchanged since Aug 30 — ready whenever you are.

One thing the merge box makes confusing: `PR Metadata Preflight` shows twice on this head, cancelled at `08:59:19` and succeeded at `09:00:51`. The cancelled one is the superseded run; the later result is green, and every other check is success or skipped.

aryanorastar:
Closing as superseded: #13595 merged this registry with both commits from here, plus the `bash -c` hardening for #12860 on top, so `main` already carries everything in this PR.

PR #12436, opened 2026-08-30
feat(desktop): panels for what the user asks to keep

katipally:
## What this does

Speech disappears. An answer the user has to paste, keep, or read back had nowhere to
live, so a floating panel now holds it. One card at a time, three ways it gets there.

**Asked for, out loud.** Seven voice tools, realtime surface only. `show_panel` puts up
anything Omi can write itself; `find_and_show` searches the user's own stored data and
fills the panel from it; `draft_message` and `assist_form` run the two assistants on
demand. `update_panel` revises what is on screen in place — "make that shorter", "swap
that name" — so a dragged window stays put and the transcript keeps one card per panel.
`close_panel` and `reopen_panel` are the pair the context binding needs.

**Offered, when a form appears.** An accessibility walk is free, so it runs; answering
is not, so the card asks first and names what it found ("16 fields Omi can try,
including 9 to write"). The ✓ buys the model call. The ✗ means never; ignoring it for
four minutes means not for another 30.

Any of these can answer about what is on screen: a spoken lookup sends the front window's
accessibility text and a picture of it with the question, so a page answers even where the
app publishes no tree at all. Electron apps are asked to publish one; Chromium browsers are
not, because the only lever there is the VoiceOver flag and setting it breaks window managers.

**Offered, when a compose box is focused.** Same shape in Mail, Messages, Slack,
Telegram, WhatsApp and their web versions: a free card asks, the tick writes the draft
from the user's side of the thread, and the context box takes refinements.

### How

`PanelSession` owns which card is up and for how long — asking that question in three
places is how a card outlived what it was about. A panel belongs to a `PanelContext`
(app, window, tab URL): leave and it hides, return and it comes back, and it never
survives into a tab that knows nothing about it. Because it leaves with its context, a
panel the user asked for is also written into the chat transcript as a collapsible card,
which is where its content survives. And the voice session is told what is on screen at
each turn boundary, so a turn that calls no tool cannot claim it changed a panel it
cannot see.

### Memory work in this branch

Three commits here are also PR #12361, unmerged and currently conflicting with main:
the chat memory-save fixes (`create_memories` for a batch save, saves the model judges
rather than only literal "remember this", memories written outside the memories page
surfacing there, and stranded local memories retrying).

They are separable — panels have no compile dependency on them; the sweep reads the
`memories` table directly and the authorized-tool hash fix they carried is already on
main. They are included because a panel answering from the user's own data is only as
good as what got saved, and shipping the two together is what was actually exercised.
#12361 stays open and is reviewable on its own. Whichever lands first, the other's copy
of those three commits carries identical content, so the second merge collapses them
rather than conflicting on them.

- INV-AGENT-*
- INV-AUTH-1
- INV-CHAT-1
- INV-VOICE-1
- INV-INT-1
- INV-MEM-1

## Failure class

Failure-Class: FC-decoded-outcome-discarded-before-canonical-record

## Verification

- `xcrun swift build -c debug --package-path Desktop` — Build complete.
- `xcrun swift test` over the panel, form-assist, draft, sweep, connector and
  authorized-tool suites — 208 tests, 0 failures.
- `desktop/macos/agent`: `npm run build`, `generate-tool-surfaces.mjs`, then `--check` —
  all outputs match. `npx vitest run` — 706 tests, 0 failures.
- Live on a dev bundle (`com.omi.omi-omi-proactive`, Gemini Live) driving real turns
  through `ptt_test_turn`: `show_panel` then `update_panel` each carry one card into
  chat; a later turn read the panel back with no tool call; a `spawn_agent` +
  `show_panel` turn queued its card and the next turn carried it; a form offer left
  untouched for four minutes expired and produced no re-offer across two
  away-and-back switches.

Line-Count-Exception: desktop/macos/Desktop/Sources/CloudConnectorGuidanceOverlay.swift | 1259 -> 1584 | the copy card grew the ask row, section grouping, live update and expiry callback that the panel surface is built on
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Components/ChatBubble.swift | 2113 -> 2136 | a copy button on the discovery card, so a panel card in the transcript can be copied without unfolding it
Line-Count-Exception: desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift | 1521 -> 1530 | adds the panel_lookup fallback area and its four reasons to the closed allowlists
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift | 5740 -> 5742 | threads assistantContentBlocks so a panel card reaches the transcript
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController+SessionDelegate.swift | 1549 -> 1813 | the six panel tool executors, the panel-state injection, and the agent-status redirect
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController+SessionLifecycle.swift | 1519 -> 1545 | the turn-write route now decides whether it can carry panel cards before consuming them
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController.swift | 1533 -> 1539 | panel-state dedupe fields and their reset on a new session
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift | 3701 -> 3709 | surfaces memories written outside the memories page
Line-Count-Exception: desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin.swift | 2225 -> 2296 | the three on-demand entry points behind draft_message, assist_form and find_and_show
Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift | 3772 -> 3780 | routes create_memories
Line-Count-Exception: desktop/macos/Desktop/Sources/Rewind/Core/MemoryStorage.swift | 1837 -> 1856 | retries stranded memories and reads the sweep's remainder from the local table
Line-Count-Exception: desktop/macos/Desktop/Sources/Rewind/Core/RewindDatabase.swift | 3567 -> 3568 | registers the sweep FTS migration


katipally:
Pushed six more commits after live testing. What changed and why:

**Panels can see the screen now.** A spoken lookup used to search only what was stored, so asking about a page in front of you came back with `[Hackathon Name]` style placeholders. It now sends the front window with the question — the accessibility text where the app publishes it, a picture of the window where it doesn't. Scoped to the window rather than the display, so other apps and Omi's own panel stay out of the frame.

**Electron apps work at all.** Slack, Discord, VS Code and Notion kept their accessibility tree switched off, so form assist found no fields and the draft card found no compose box. They're now asked to publish one (`AXManualAccessibility`), once, when you switch to the app. Chromium browsers are deliberately *not* asked: the only lever there is the VoiceOver flag, and setting it breaks window managers, so Brave and Chrome are answered from the window's picture instead.

**A stale topic could beat what's on screen.** With a page read earlier in the session, "mail for this" while looking at something else drafted about the earlier topic. The draft prompt said the user's instruction "wins every conflict" — but that instruction is written by the voice model from its own history, not by the user. The screenshot is now the first source and is what "this" refers to; the instruction says what to *do*, and loses when it names something the screen doesn't support.

**A card came up titled `floating_chat`.** An internal token printed as a heading. The guard against that existed but was wired to one of three paths; all three check now.

**Two card fixes.** The panel's subtitle said "Copy it with the button", describing the button next to it — it now names the key to hold to change what's up, read from whatever you've bound. And a panel kept in chat can be copied without unfolding it first.

Verified live on a dev bundle, plus 231 Swift and 707 agent tests, and `make preflight` green at 28 checks.

Git-on-my-level:
Thanks @katipally - this is a substantial and unusually well-validated contribution: a large panel/form-assist/message-draft surface with 231 Swift and 707 agent tests reported green, live turn-by-turn verification on a dev bundle, an e2e flow (desktop/macos/e2e/flows/voice-panels.yaml), a changelog entry, and reasoned line-count exceptions. The single-panel ownership model is the right call.

On fit, qualitatively: an answer the user has to paste finally has somewhere to live, and the form/draft offers follow the desktop proactive-assistant pattern - this lines up cleanly with the product direction. No misalignment concerns.

What I verified in the code:
- PanelSession.swift - one owner for which card is up; ambient vs requested origin; tokens so a late lookup cannot write into the panel that replaced it (pinned by PanelLifetimeTests); PanelContext binds by app/window/tab-URL with a cheap title-only check on AX events and the full URL comparison in the sweep.
- RealtimeHubController+SessionDelegate.swift - panelAwareOutput merges reminders into spawn_agent and agent-status results as JSON fields (so RealtimeProviderToolResultPolicy can still parse them), and the six new tool executors re-check the owner around every physical effect.
- OmiSweep.swift / SweepFTSSchema.swift - external-content FTS5 indexes with insert/update/delete triggers and backfill of existing rows; MATCH terms are reduced to letters/digits and deduplicated (no FTS syntax injection); row ids are bound parameters.
- DataAnswerAssistant.swift - the search loop only offers read-only actions (open_refs/list_memories plus data reads; no write or SQL action exists in the step enum), with a 6-turn cap, result clipping, and a daily budget.
- FormAssistAssistant.swift - protected-characteristic, SSN, salary, and consent fields are filtered out before the model ever sees the labels, and the gate refuses credential-only forms while still allowing an application that happens to end in a password field. Strong posture.
- AuthorizedToolExecution.swift - the .withoutEscapingSlashes canonicalization fix with a hash value cross-checked against the kernel's stableJsonStringify is a real bug fix (slash-bearing authorized inputs were previously rejected as input-hash mismatches).
- ElectronAccessibility.swift - AXManualAccessibility requested once per Electron process, deliberately not the VoiceOver flag; own-pid guarded, bounded tracking set.

The one item I believe needs an explicit maintainer decision before merge - ChatToolExecutor+MemoryCreation.swift, external-surface-tool-policy.ts, kernel-types.ts: this removes the deterministic explicit-save-intent validator (the negation/passive/third-party patterns and the token-overlap grounding check) along with the shared memory-save-intent corpus and its tests, and moves the "did the user actually ask" judgment entirely into the tool description ("however they phrase it"). The surface gate (typed main/floating chat, no client scope) and the owner-change checks remain, and create_memories even re-checks the owner per fact - but the previous behavior deterministically refused passive mentions, questions, and assistant-suggested saves; after this change the model's reading of the turn authorizes the write. That is a privacy/injection-resistance trade-off on a user-data write path, not a bug, and it deserves a human yes/no rather than an automated one.

Also flagging: desktop/macos/AGENTS.md adds an agent-instruction section requiring live voice-turn validation (./run.sh --yolo plus omi-ctl ptt_test_turn with force_transcript) for panels, form assist, message draft, and realtime tool changes, citing specific log lines. I verified all three cited strings ('panel card(s) into chat', 'panel state sent', 'FormAssist: offering') exist verbatim in this diff, so the doc is accurate and source-backed. Consequence: AI coding/review agents working in desktop/macos will be expected to run a live dev backend and drive real turns before landing changes on these surfaces - a meaningful contributor-workflow change (higher runtime cost, much stronger regression net across turn boundaries). Safe as written; worth a deliberate nod.

Checks are green on this head and the work is cohesive around one product surface, so no split is being requested. Leaving for human maintainer review specifically for the memory-write authorization decision above, with a side nod to the new AGENTS.md live-validation expectation; everything else looks ready to move quickly once that call is made.

---

katipally:
Fixed the panel you close not coming back.

`close_panel` forgot the panel outright, and `reopen_panel` reads only what's remembered — so the pair the voice tools offer could never complete. It worked solely for a panel that *hid* on a context change, which keeps the memory. The panel that last went is now kept and can be asked back by name.

Two more defects turned up while fixing it: expiry never took the window down (production got away with it because the card's own timer hides itself first), and a second `forget()` overwrote what the first recorded, so close-then-expire lost the panel.

A card dismissed with the ✗ is reopenable too — it means "not now", only an explicit ask reaches it, and the surface that offered is still told it was declined so the card never re-offers itself.

Verified live: `show_panel` (panel state 578 chars) → `close_panel` (none, 358) → `reopen_panel` (578 again) → next turn read "Aldgate, Barbican, Camden" back with no tool call.

Git-on-my-level:
Thanks @katipally — reviewed the two commits now on this head (the close/reopen fix and the upstream merge).

**The panel you closed can be brought back** — verified in `PanelSession.swift`: `forget()` now records the last-remembered panel, `reopen()` restores it through `restoreLastClosed()` with a fresh token (so work started before the close cannot land in the restored panel), no countdown on a panel asked back by name, and `expire()` now takes the window down before forgetting. `forgetClosedHistory()` keeps that process-wide state out of the neighbouring suites, and the new `PanelLifetimeTests` cases pin the behaviour you described: reopen-once-only, ✗ meaning "not now" while still reporting declined to the offering surface, and a passed deadline not killing a restored panel. Clean fix.

**On the red checks for this head:** "Desktop Swift Build & Tests" is failing only as a mirror — the workflow's final gate hard-fails when the release lane does not succeed, and "Desktop Swift Release Compile" was cancelled at the job's 60-minute budget while compiling the release-mode test targets (the release binary itself linked fine). The debug verify lane passed: static contracts and Swift tests green, as are Agent Runtime and desktop-core-e2e-t0. The previous head's identical lane finished green in about 57½ minutes, so this reads like the added release-mode test targets squeezing a lane that already ran close to its budget — not a code failure. A workflow re-run should clear it; if the release lane keeps timing out, the 60-minute budget in `desktop-swift-ci.yml` versus the growing test-target count is worth a separate look (that file is untouched here).

**Still gating merge** (unchanged from the earlier review and still present on this head): the memory-write authorization change. `external-surface-tool-policy.ts` and `kernel-types.ts` drop the deterministic explicit-save-intent and grounding validators, and the rewritten `create_memory` plus new `create_memories` guidance in `omi-tool-manifest.ts` and `GeneratedToolCapabilities.swift` moves "did the user actually ask" into model judgment ("however they phrase it"). The typed-surface gate and owner checks remain, and `executeCreateMemories` re-checks the owner around every fact — but that is a privacy/injection-resistance trade-off on a user-data write path that wants a human yes/no, so `needs-maintainer-review` stays for that decision.

Also noting again for maintainers: `desktop/macos/AGENTS.md` adds the "Proactive surfaces must be driven live" section — re-verified on this head that all three cited log lines (`panel card(s) into chat`, `panel state sent`, `FormAssist: offering`) exist verbatim in the sources, so agents following it will be looking for real signals.

---
*by AI on behalf of David — merge is gated on the human memory-write authorization call above.*

Git-on-my-level (commented):
Reviewed the new head — this is the upstream-main merge (37d5ce57), no new feature commits since the close/reopen fix that was reviewed earlier. Delta findings:

**The merge is clean where it overlaps.** Upstream touched four files this branch also changes — `FloatingControlBarWindow.swift`, `RealtimeHubController+StreamingJournal.swift`, `RealtimeTurnPersistence.swift`, `ChatBubble.swift` — and I verified on this head that the panel threading survived all four: `assistantContentBlocks` still flows through the journal finalization and turn persistence, and the discovery-card copy button is intact. The memory-authorization files (`external-surface-tool-policy.ts`, `kernel-types.ts`, `ChatToolExecutor+MemoryCreation.swift`, `omi-tool-manifest.ts`) are untouched by upstream, so the still-open gating question below is exactly as previously reviewed.

**Checks: the previously-noted red is gone.** The release-compile lane that was cancelled at its 60-minute budget on the last head completed green here, along with the debug test lane, Agent Runtime, and desktop-core-e2e-t0. Nothing left to re-run.

**One addition to the open maintainer decision** (not new code on this head, but worth putting beside it): the `find_and_show` path is a new data-egress lane a maintainer should sign off on knowingly. `DataAnswerAssistant.openingPrompt` sends the spoken question together with the user's name, email, and full AI profile, and `ScreenGlance` attaches the front window's accessibility text plus a downscaled screenshot of that window — riding every step of the lookup loop, not just the first. The gates are genuinely good (40/day budget with refunds on cancel, owner checks, read-only tool allowlist in the loop, AX text capped at 6,000 chars), and it mirrors lanes the shipped assistants already use — but "a spoken question now captures the front window and the user's profile" is a privacy-surface expansion of the same magnitude as the memory-write change already gating merge, so the two belong in one sign-off.

**Unchanged from the earlier reviews:** merge stays gated on the human call for the memory-write authorization change (deterministic save-intent and grounding validators removed for model judgment, per the commits shared with #12361; typed-surface gate and owner checks remain). The AGENTS.md live-drive section was re-verified source-backed on this head. Applying `security-review` alongside `needs-maintainer-review` to mark both decisions for whoever picks this up.

---

Git-on-my-level:
Delta review of the two commits since the last pass (37d5ce57 -> 6b7969a7), plus CI triage and the one substantive sign-off item.

**The new commits hold up.**
- `060d3fcb` (carry panel state across the new background-context result): upstream changed `sendBackgroundAgentContext`/`injectContext` from `Bool` to `RealtimeBackgroundContextDeliveryResult`; `RealtimeHubController+SessionDelegate.swift` now branches on `.retry` only, so a provider with no background-context role no longer re-sends the same panel-state line every turn. `AgentCompletionVoiceDeliveryTests.swift` moved to a prefix assertion for the appended screen reminder, consistent with the neighbouring cases.
- `6b7969a7` (HubEscalationTests assertion): `publicWebSearchPrompt` has returned "Search the live public web thoroughly…" since bce354d767, and upstream `main` is currently red on exactly this assertion (verified in today's main CI log at HubEscalationTests.swift:87). This branch fixes it - worth noting for whoever is triaging main.

**CI: the red is a release-lane timeout, not a test failure.** In the failing "Desktop Swift Build & Tests" run, `VERIFY_RESULT: success` - the Swift test lane passed; the required gate hard-fails because "Desktop Swift Release Compile" was cancelled. That job ran 60m20s and was cancelled while compiling the release test target, after a ~28-minute whole-module compile of `Omi_Computer`. Release runs on this branch have taken 35-57 minutes, and the previous cancellation (2a871bfe) went green on the next run, so a rerun should clear it - but the lane is brushing its ceiling, which may itself be worth a look given how much this branch adds to the desktop target.

**The item that needs maintainer sign-off: memory-write authorization was deliberately loosened.**
- `agent/src/runtime/external-surface-tool-policy.ts` drops `hasExplicitMemorySaveIntent` and `hasMemoryContentGroundedInUserRequest` entirely, along with the Swift mirror `isExplicitMemorySaveIntent`, the `memory-save-intent-corpus.json` fixture, and both error codes in `kernel-types.ts`. Previously a `create_memory` call from an external surface required an affirmative lexical "remember/save" command in the current turn and content grounded in the user's request.
- What remains in force: the typed-surface restriction (`main_chat`/`floating_chat` only, in `ChatToolExecutor+MemoryCreation.swift`), expected-owner checks on both write paths, per-fact size limits and dedup, and the agent-side bundle gate (`desktop-tool-policy.ts` adds `create_memories` to `MEMORY_WRITE_TOOLS`, with a regression test noting that a missing entry would deny the bundle outright). No orphaned references to the removed policy remain (checked).
- The tool docs now say "Use whenever the user asks you to remember, save, or keep something, however they phrase it", including a trailing "and save this". That direction is defensible - the strict gate false-negatived on natural phrasing - but it moves authorization from a deterministic lexical check to model judgment and removes the "content must be grounded in the current request" check. That is a privacy/authorization product call, which is what `security-review` should stay on this PR for.

**Agent-guidance docs (AGENTS.md) - accurate and verified.** The new "Proactive surfaces must be driven live" section in `desktop/macos/AGENTS.md` instructs agents to drive real turns (`./run.sh --yolo` plus `omi-ctl action ptt_test_turn … text_only=1 force_transcript=…`) and cite specific log lines before landing changes here. I verified the section is source-backed: `ptt_test_turn` is a real bridge action (RealtimeHubSession test seam), `force_transcript` is a real parameter, `omi-ctl` accepts generic `action <name> key=value`, and all three cited log lines exist in code (`RealtimeHub: panel state sent` in SessionDelegate, `panel card(s) into chat` in SessionLifecycle, `FormAssist: offering` in FormAssistAssistant). Guidance like this changes how AI coding/review agents verify work in this repo; here it raises the verification bar and the commands check out, so it is safe to keep as-is.

**Sampled the rest (85+ files):**
- `Panels/PanelSession.swift` - single-owner model for the one floating card; origin (ambient/requested) and grain (context/app) semantics are pinned by the PanelLifetime/Ownership/Revise/EdgeCase suites, including "a late answer never writes into the panel that replaced it".
- `Rewind/Core/SweepFTSSchema.swift` - external-content FTS5 over `memories` and `transcription_segments` with insert/update/delete triggers and a full `INSERT … SELECT` backfill; dropping the virtual tables is a complete rollback. One-time migration cost on large transcript tables is the only thing to watch.
- `FormAssist/ElectronAccessibility.swift` - sets `AXManualAccessibility` (deliberately not the VoiceOver flag, which breaks window managers) on Electron apps only, detected via bundle framework rather than app name, once per pid with a bounded set, never on Omi itself. It is a write into a third-party process, but guarded and using the right attribute; the tests pin the detection.
- `Chat/AuthorizedToolExecution.swift` - `.withoutEscapingSlashes` aligns the Swift canonical JSON with the kernel's `stableJsonStringify` (verified in `kernel-support.ts`); the new test asserts a kernel-derived digest for slash-bearing nested input. This fixes real rejections of URL/path-bearing tool inputs.
- `FloatingControlBar/RealtimeHubController+SessionDelegate.swift` - `panelAwareOutput` appends panel-awareness to agent-status tool results as a JSON field that stays parseable (pinned by PanelAgentBleedTests after a measured incident where appended prose broke result finalization).

Test coverage is extensive for a change this size (roughly 4k lines of new tests plus an e2e flow yaml), and the live-turn rule in AGENTS.md is the right complement for the behavior unit tests cannot reach.

No split requested - this is one product surface and the pieces interlock through `PanelSession`; splitting would separate the shared spine from its users without making review or rollback easier. Before merge: a green release-lane rerun, and a maintainer decision on the memory-write authorization change above.

Feedback generated by AI review automation on behalf of the maintainers.

---

Git-on-my-level:
Delta pass on the two commits since the last review (6b7969a7 -> 5454ea76).

**The release lane is green now.** "Desktop Swift Release Compile" passed on this head (the earlier failure was a 60-minute timeout cancellation mid-compile, not a test failure), alongside "Desktop Swift Build & Tests", "Agent Runtime", the static/test contracts, and desktop-core-e2e-t0. That clears the rerun precondition from the previous review.

**The diff is desktop-only again.** 5454ea76 restores app/ to origin/main exactly, so the PR is back to 85 files, all under desktop/macos - the Dart churn inherited from earlier merges is gone. Verified the PR file list contains no non-desktop paths. The upstream merge in between (fbc5d400) touches no desktop files, so no panel seams were re-merged.

**The one item still open before merge is unchanged:** the memory-write authorization change. `agent/src/runtime/external-surface-tool-policy.ts` drops `hasExplicitMemorySaveIntent` and the content-grounded-in-current-request check (with the Swift mirror and corpus fixture removed), moving authorization from a deterministic lexical gate to model judgment. Re-verified at this head that the remaining guardrails hold: the fail-closed typed-surface gate in `ChatToolExecutor+MemoryCreation.swift` still restricts both `create_memory` and `create_memories` to main_chat/floating_chat with no client scope, expected-owner checks sit on both write paths, and `desktop-tool-policy.ts` keeps both tools in `MEMORY_WRITE_TOOLS` behind the desktop.memories.write bundle. That authorization trade-off remains a product/security call for a maintainer - security-review and needs-maintainer-review stay for that decision.

The live-drive guidance in `desktop/macos/AGENTS.md` was re-checked at this head and remains source-backed (omi-ctl and the ptt_test_turn action are real, the cited log lines exist).

Status: everything this automation verifies is green on 5454ea76; the memory-write authorization sign-off is the remaining pre-merge decision.

Feedback generated by AI review automation on behalf of the maintainers.

---

aryanorastar:
Keeping this open on `5454ea7638bc`.

DIRTY +11k desktop panels; extra production vs #12361 so not CLOSE; do not RC just for DIRTY.

aryanorastar:
Keeping this open. Not approving `5454ea7638bc`. 85 files under macOS form-assist and realtime hub. Not one defect. Hygiene SUCCESS. mergeable CONFLICTING/DIRTY.

PR #12440, opened 2026-08-30
feat(desktop): make the Activity hour rail one continuous scrollable ribbon

katipally:
## What

The Activity spine's hour rail drew one day at a time and swapped the whole
column when the top row crossed midnight — a navigator that jumped beside a list
that flows. It is now one continuous ribbon across every loaded day, and you can
scroll the list by hovering it and using the wheel or trackpad.

```
   BEFORE (swaps at midnight)       AFTER (one strip)
 ┌──────────────┐                  ┌──────────────┐
 │ ▬▬▬▬▬  12 PM │                  │ ▬▬▬▬▬  12 PM │
 │ ▬ 12 AM      │ ← column         │ ─── FRI 28 ──│ ← labelled break
 └──────────────┘   cross-fades    │ ▬▬▬  10 AM   │
   one day only     to next day    └──────────────┘
```

A single day still reads exactly as before: one day is one column, the size the
rail's `maxHeight: .infinity` bars always were, with the same widths, weights and
captions.

## Three things worth reviewing

**The position is continuous, not animated.** The rail was told only which *row*
was under the reading line — a step function, so it stalled then jumped. An
animation only trades the jump for a lag. Each row now reports its top as well as
its bottom, the viewport keeps the two rows nearest the reading line, and the
strip blends between them. It moves every frame the list does and needs no easing.

**The highlight hands over rather than dimming.** Sharing the weight evenly
between the two neighbouring bars made the highlight vanish at half strength for
most of a scroll. The bar within half an hour of the marker — always exactly one —
stays at full strength; only the second nearest fades out. One hour states its
time, never two.

**The ribbon drives the list's scroll view, not its own.** A second scroll view
would need a proportional mapping onto a list whose rows differ in height, and
would drift within a screen. The offset is computed in `SpineScrollLink` rather
than forwarded to `NSScrollView.scrollWheel(with:)`, because a real trackpad event
carries a gesture phase and forwarding it is silently ignored. Momentum still
works — macOS sends the coast as its own decaying events.

## Verification

- `xcrun swift test --package-path Desktop --filter "Spine|QueryShell"` → **110
  tests, 0 failures**
- `make preflight` → **22 checks pass**
- Exercised on the `omi-omi-timeline` dev bundle with synthesised phased trackpad
  gestures: a swipe over the ribbon scrolls the list Today → Yesterday → Thursday
  27 August and back, with the headline, lit hour and day breaks tracking at every
  stop.

Geometry, reading position and highlight falloff are pure functions, so the rules
are pinned in tests rather than left to a screenshot. Writing the "exactly one
hour states its time" test caught a real one: a symmetric `abs(d) < half` is false
for *both* slots at the exact midpoint, blinking the time out for that frame.

No product invariants matched the changed paths; no failure class needed (both
behavioural commits are `feat:`). `SpineHourBar` is gone — the `Canvas` draws the
bars now, with its metrics moved across unchanged.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12440?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## One commit outside this feature

`OnboardingFlowTests.testOnboardingProceedActionsUseDefaultActionKeyboardShortcut`
has been failing on `main` since ff96bff7bb, unrelated to this change. That
commit added the "Open the doors" button to `SBOnboardingView`; the button
registers Return correctly, but the test's hardcoded expectation of eight
`isDefaultAction: true` sites did not move with it. Every desktop PR since then
fails Desktop Swift Static & Test Contracts on it, with Build & Tests failing as
a rollup. Verified by running the same test on clean `upstream/main` with none of
this branch's code present.

The count is not redundant with the per-title assertions above it: those name
specific buttons, so only the total can catch a proceed action added *without*
`isDefaultAction`. Raised to nine and annotated with what the number guards.
Happy to split it out if you would rather it landed on its own.

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Spine/SpineHourRail.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Spine/SpineHourRail.swift
@@ -19,10 +19,24 @@ import OmiTheme
 import SwiftUI
 
 struct SpineHourRail: View {
-  /// One entry per hour, indexed 0…23, each 0…1 against the day's own busiest hour.
-  let density: [Double]
-  /// The hour under the top of the list, or nil before anything has been read.
-  let currentHour: Int?
+  /// Every loaded day, newest first — the whole strip, not just the one being read. See
+  /// `SpineRibbon`.
+  let days: [SpineRibbonDay]
+  /// Where the strip sits, in strip points — already blended between the two rows nearest the
+  /// reading line, so it moves every frame the list moves. See `SpineReadingPosition`.
+  ///
+  /// A closure rather than a value because the scale of the strip is the column's own height (one
+  /// day is one column) and only the `GeometryReader` below knows it.
+  let depth: (CGFloat) -> CGFloat
+  /// Whether any hour is being read. `false` when no row has been reported — every day folded shut,
+  /// or nothing read yet — where the strip still parks somewhere but lights nothing.
+  let highlightsMarker: Bool
+  /// The hour to *say*. Which hour is drawn heavier is derived from `depth`, but speech has no
+  /// column height to derive anything from, and "Reading 12 PM" is the one part of this column a
+  /// listener cannot get from the headline. `nil` when nothing is being read.
+  let readingHour: Int?
+  /// The list's scroll view, so a wheel over the ribbon moves the list. See `SpineScrollLink`.
+  let scrollLink: SpineScrollLink
   /// The headline: how much screen capture the day being read holds, or `nil` while that day's index
   /// is still being read. `nil` is not zero — see `headlineCaption`.
   let momentCount: Int?
@@ -55,13 +69,6 @@ struct SpineHourRail: View {
   /// shortened forms out of 10,234 and removes the coin flip.
   private nonisolated static let scopeFitBudget: CGFloat = contentWidth - 1
 
-  /// Labelled hours. Four is enough to orient without turning the rail into an axis.
-  private static let labelledHours: Set<Int> = [0, 6, 12, 18]
-
-  /// A bar this fraction of the peak or more is drawn heavier. Not a hue — the rail is one ink at
-  /// two weights, like every other ranking in this system.
-  private static let hotThreshold = 0.6
-
   var body: some View {
     VStack(alignment: .leading, spacing: 12) {
       // Number, noun, day — "0 screen moments, Yesterday", read straight down. **The day goes last,
@@ -80,7 +87,7 @@ struct SpineHourRail: View {
         }
       }
 
-      bars
+      ribbon
 
       if let footer = Self.footer(conversationCount: conversationCount) {
         Text(footer)
@@ -98,7 +105,7 @@ struct SpineHourRail: View {
           momentCount: momentCount,
           dayTitle: dayTitle,
           conversationCount: conversationCount,
-          currentHour: currentHour)))
+          currentHour: readingHour)))
   }
 
   /// **The rail counts one day of screen capture; the panel's corner counts the whole account.**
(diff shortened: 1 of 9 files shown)


Git-on-my-level:
Reviewed the full diff — this is carefully built, and the midnight-swap defect it fixes is real. Notes per file:

- `SpineRibbon.swift` — good factoring of the geometry into pure arithmetic (`SpineRibbonGeometry.depth`, `visibleDays`), and the two invariants that make the strip readable are well chosen: the nearest bar is always fully lit (`litness`), and the half-open `isBeingRead` guarantees exactly one slot owns the caption. Drawing only the visible days in a `Canvas` is the right call for a strip that can span hundreds of days.
- `SpineScrollLink.swift` — computing the offset instead of forwarding to `NSScrollView.scrollWheel(with:)` is correct: phased trackpad gestures go through responsive scrolling and ignore force-fed events. The phased-gesture regression test is the most valuable test in the PR.
- `SpineStream.swift` — the two-row blend (`SpineReadingPosition.fraction`) is what makes the position continuous without layering an animation over a step function, and `MainActor.assumeIsolated` over a per-frame `Task` allocation is the right trade while scrolling.
- `SpineHourRail.swift` — the rail's visual language is faithfully preserved (same opacities, bar heights, labelled hours 0/6/12/18), so a single day still reads exactly as before.
- `SpineStore.swift` — caching `ribbonDays` beside `days` so histograms rebuild only when the day list moves is the right avoidance of per-redraw work, and keeping them full-day under filtering is correct navigator semantics.
- `SpineCompositionTests.swift` / `SpineScrollLinkTests.swift` — the tests pin real invariants (monotonic depth across midnight, one-day-per-column at any height, direction sign, clamping at both ends) rather than snapshots. Nice.
- `changelog/unreleased/20260830-activity-hour-ribbon.json` and `e2e/flows/home-spine.yaml` — the changelog entry matches the user-visible change, and the new e2e steps S1b (midnight continuity) and S1c (ribbon as scrollbar) cover the new behavior.

Two minor observations, neither blocking:
1. `SpineScrollLink.scroll(by:)` returns `true` when the clamped origin doesn't change, so wheeling over the ribbon while the list is pinned at either end swallows the event rather than letting it bubble (no rubber-band over the ribbon). If that's intended scrollbar behavior, fine — just flagging it.
2. A precise-delta event with `scrollingDeltaY == 0` (e.g. a horizontal-only gesture) is also swallowed; passing those through would cost nothing.

Leaving for human maintainer review before merge: this reshapes an interaction surface — compact rather than calendar-scaled strip, wheel-over-ribbon as the list's scrollbar — which is a product/UX direction call, not just an implementation detail. Code quality is high and checks are green from my side.

---

yashwanthkatipally:
## One commit outside this feature

`desktop-test-quality` has been failing on `main` since ff96bff7bb, unrelated to
this change: `ThreeDoorsDemoPageTests` reads the bundled page twice, taking
production-source inspection to 55 files / 149 sites against baselines of
54 / 147, so every desktop PR since then fails Hygiene. Both sites are static
template tripwires that already say so in prose and were missing only the
machine-readable `omi-test-quality: source-inspection` annotation. Annotated
rather than baseline-raised. Happy to split it out if you would rather it landed
on its own.

Git-on-my-level:
Follow-up on the new head (87c630e) — the highlight hand-over rework, the test-quality annotations, and where the two red required checks actually come from.

**The failing checks are main's, not this PR's.** Pulled the logs on both sides:
- `Desktop Swift Static & Test Contracts` fails on `OnboardingFlowTests.testOnboardingProceedActionsUseDefaultActionKeyboardShortcut` — `SBOnboardingView.swift` now carries 9 `isDefaultAction: true` sites while the test pins 8. The ninth site is present at this PR's base commit (it arrived with main's doors-demo commits on Sept 1); none of this PR's ten files touch onboarding. The same two checks are red on main's latest run too.
- The test-quality ratchet main is currently failing (55 source-reading files / 149 sites over the 54/147 baseline) is actually *repaired* on this head: the two `omi-test-quality: source-inspection` annotations in `ThreeDoorsDemoPageTests.swift` follow the checker's documented convention for static-contract tests, and the gate reports at-baseline (54/147) on this PR's run. Passing that gate is what let the suite run far enough to expose the OnboardingFlowTests drift above.
- All 739 Swift suites ran on this head; the only failure is that OnboardingFlowTests one. The PR's own suites (SpineCompositionTests, SpineScrollLinkTests, ThreeDoorsDemoPageTests) pass.

So this PR is merge-blocked on main cleaning up its own baseline, not on anything to change here. Fixing the `isDefaultAction` count belongs with the doors-demo change that introduced the ninth button.

**On the new code since the earlier automated review:** the hand-over design is the right shape. `SpineRibbon.litness` keeps the nearest bar at full strength while only the outgoing neighbour fades — the even-sharing version would have made the one findable element on the strip the dimmest thing on it for most of every scroll — and the half-open `isBeingRead` interval guarantees exactly one slot owns the caption at every marker position, including exactly midway. `testTheBarBeingReadIsNeverDimmedByTheHandover` and `testTheStripLabelsTheSameHoursTheRailAlwaysDid` pin both behaviours.

The two minor observations from the earlier review still stand and remain non-blocking: `SpineScrollLink.scroll(by:)` returns `true` when the clamped origin doesn't move (a wheel over the ribbon while the list is pinned at either end is swallowed rather than bubbling — no rubber-band over the ribbon), and a precise-delta event with `scrollingDeltaY == 0` is likewise swallowed. If pinned-end feedback ever matters, both are one-line pass-throughs.

Human sign-off wanted for the interaction direction itself: the ribbon now doubles as the list's scrollbar (`SpineWheelForwarder` driving the list's own `NSScrollView`), which is a navigation-paradigm decision for the Activity spine, not just a rendering change.

---

Git-on-my-level:
Follow-up on the rewritten head (`73f2f0c`), since the branch history changed after the last automated pass: the reviewed feature code is untouched, the rebase is clean, and the checks that were red for main-baseline reasons are green now.

What changed around the code, verified on this head:

- Every Spine file is byte-identical to the previously reviewed head — `SpineRibbon.swift`, `SpineScrollLink.swift`, `SpineHourRail.swift`, `SpineStore.swift`, `SpineStream.swift`, `SpineScrollLinkTests.swift`, the `SpineCompositionTests.swift` additions, the changelog fragment and `home-spine.yaml`. The rewrite only rebased onto newer main and dropped the two out-of-feature commits.
- Dropping those commits was right: main's `OnboardingFlowTests` now pins the nine `isDefaultAction` sites itself, so the copy this branch carried is correctly gone. One cleanup: the PR description's "One commit outside this feature" section now describes a commit that is no longer in the branch — worth trimming so reviewers don't hunt for it.
- All required checks pass on this head (`Desktop Swift Build & Tests`, `Desktop Swift Static & Test Contracts` included). The earlier merge-blocker was main's onboarding-count drift, not this PR, and it is resolved — so nothing outside the PR blocks it any more.

Still open, unchanged and still non-blocking: `SpineScrollLink.scroll(by:)` returns `true` when the clamped origin doesn't move (a wheel over the ribbon while the list is pinned at an end is swallowed rather than bubbling — no rubber-band over the ribbon), and a precise-delta event with `scrollingDeltaY == 0` is likewise swallowed.

One new observation, also non-blocking: `SpineRibbon` conforms to `Animatable` on `depth`, with a doc comment that a nine-hour jump "glides there instead of teleporting" — but nothing in this PR applies an animation, and `SpineHourRail`'s own notes deliberately avoid `.animation` on the position. Without an animation context `animatableData` never interpolates, so the conformance is currently dormant; if an ancestor view ever gains an animation, the strip would start lagging the list — the exact failure mode the design notes argue against. Worth either a comment explaining when it is meant to fire, or dropping the conformance.

What still stands between this and merge is the product call, not the code: whether the Activity ribbon should double as the list's scrollbar (`SpineWheelForwarder` driving the list's own `NSScrollView`) and adopt compact-not-calendar-scaled strip geometry. That is a navigation-direction decision for the spine and needs a human maintainer's sign-off.

---
*by AI on behalf of David. Human decision needed before merge: sign-off on the ribbon-as-scrollbar navigation direction for the Activity spine — everything code-level is green.*

undivisible (approved):
## Verdict: APPROVE

Carefully built fix for a real defect: the hour rail was a step-function / midnight swap beside a continuous list. The continuous ribbon + two-row blend + wheel-forward to the list’s own `NSScrollView` is the right shape for variable-height rows (a second scroll view with proportional mapping would drift).

### What looks solid
- **Geometry** (`SpineRibbonGeometry.depth` / `visibleDays`): monotonic across midnight; one day = one column; Canvas draws only the viewport window.
- **Highlight hand-over** (`litness` + half-open `isBeingRead`): nearest bar stays fully lit; exactly one caption owner (including midpoint) — caught by tests that already paid for themselves.
- **Scroll link**: computing offset instead of `scrollWheel(with:)` is correct for phased trackpad / responsive scrolling; `SpineScrollLinkTests` covers the regression that would have shipped green on synthetic no-phase events.
- **Perf**: cached `ribbonDays`, `MainActor.assumeIsolated` on per-frame preference reports, no `.animation` on continuous position.
- **A11y**: Canvas hidden; rail remains one spoken element via `readAloud` (scope leads in speech).
- **Tests / e2e / changelog**: invariant tests + S1b/S1c prose steps match the user-visible change.

### Product / UX call (signing off)
Ribbon-as-list-scrollbar (wheel/trackpad only; not a drag scrubber) and compact-not-calendar strip length are intentional and justified by the variable-row constraint. Approving that direction.

### Before merge (process)
Branch is ~500 commits behind `main`. `git merge-tree` reports **changed in both** for:
- `desktop/macos/Desktop/Sources/MainWindow/Spine/SpineStore.swift` (main: `calendar` visibility for recap date keys)
- `desktop/macos/Desktop/Sources/MainWindow/Spine/SpineStream.swift` (main: daily-recap slots, search analytics wrappers, `searchSurface`)

Please rebase, resolve those two files, and re-run the Desktop Swift suites before merging. Feature code itself does not need redesign.

### Non-blocking nits
1. `SpineScrollLink.scroll(by:)` returns `true` when clamped origin is unchanged → no rubber-band / no bubble at ends when wheeling the ribbon.
2. `scrollingDeltaY == 0` (e.g. horizontal-only precise gesture) is also swallowed (`guard delta != 0 else { return true }`); returning `false` would be a one-line pass-through.
3. `SpineRibbon: Animatable` on `depth` is currently dormant (no animation transaction in this PR); an ancestor `withAnimation` later would reintroduce lag-behind-list — the failure mode the design notes argue against. Drop the conformance or document the intended jump path.
4. PR description still has the “One commit outside this feature” / OnboardingFlowTests section; that commit is gone after the rewrite — trim so reviewers don’t hunt for it.

No secrets; review-only — not merging.

aryanorastar:
Keeping this open. Not approving `73f2f0cae35a`. origin/main has no `SpineRibbon.swift`. +1034 across 9 macOS files. Tests exist (`SpineScrollLinkTests.swift`). Hygiene SUCCESS. mergeable MERGEABLE/CLEAN. Do not approve this SHA.

PR #12465, opened 2026-08-31
feat(desktop): let Omi use your Mac, over MCP

katipally:
## What it does

Turn on **Allow Omi to control this Mac** in Apps and Omi serves 20 tools on loopback: screenshot, accessibility tree, mouse, keyboard, `open_app`, `run_applescript`. It registers in `~/.omi/mcp.json` like any other server, so the pi lane, the ACP lane and any client pointed at that file get the same surface through one path. Off by default, and `⌃⌥⌘.` stops everything.

The switch gates `tools/call` only, so a client's health probe still gets a clean handshake while control is off. Each tool names the TCC grants it needs, so a missing one refuses just what depends on it: no Screen Recording still leaves the accessibility lane working.

A background agent also keeps its place. "Now search it" continues the run it refers to, from voice or chat, and survives the agent process being replaced.

## Latest round

**A Screen Recording grant given after launch is not a working grant.** `CGPreflightScreenCaptureAccess` flips true the moment the box is ticked, while ScreenCaptureKit stays bound to the window-server connection opened at launch. Every surface reported the capability as live while `screenshot` returned nothing, and the model told the user:

> Screen Recording permission may be missing — grant it in System Settings ▸ Privacy & Security ▸ Screen Recording.

which is what they had just done. The loop has no exit until a relaunch nobody suggested. `ScreenRecordingPermissionPolicy.needsRelaunchToApply` already models this window and five other screen consumers respect it — Rewind, onboarding, DayZeroChips, RealtimeScreenEvidence, AppState. Computer use now reads it too rather than growing a second notion of the same thing, so the card reads `Restart to apply` and the tool asks for the restart instead of the grant.

**The detail sheet asks for consent instead of printing config.** It opened on the loopback endpoint the card already showed and pinned itself to 620pt. It now leads with what Omi does with the Mac, lists the three grants at all times — granted in green, missing with a red Grant, granted-but-stale with Reopen Omi — lists the 20 tools, and sizes to its content. Grant no longer calls `request()` and `openSettings()` together, which put System Settings on top of the dialog it had just raised.

**One grant poller.** The card grid and the sheet each owned a 2s timer running the same accessibility probe. The store owns one now, with a watcher count, and it stops when the last surface leaves.

**An automation seam.** `omi-ctl action computer_use` reads and sets the switch, the suspension and the grants — the feature had none, which is why the defect above had to be reasoned about rather than driven.

## Verified by driving the app

| Area | Evidence |
|---|---|
| MCP endpoint | Registration self-repaired its port and token at launch. `initialize` + `tools/list` returned 20 tools. `open_app` opened Safari, clipboard round-tripped, `run_applescript` `return 6 * 7` gave 42. |
| Permission model | Each gated tool refused naming only its own grant: `screenshot` → Screen Recording, `ui_snapshot` → Accessibility, `click` → PostEvent. |
| State machine | Driven live against a named bundle through the new seam: off → `Off`, on → `Needs 3 grants`, suspend → `Stopped`, re-arm → `Needs 3 grants`. Enabling wrote the `omi-computer-use` entry to `~/.omi/mcp.json`, so the switch does the whole setup. |
| Continuity (A/B, one variable) | Spawn a background agent holding a fact, `kill -9` the agent runtime, ask it back. **With** the recap: `binding.recap_seeded` (recovered), answered `ZEPHYR-4419`. **Without**, nothing else changed: same binding replacement, no recap event, "I don't know the passphrase." |
| Voice | Two real PTT turns with spoken audio. "Open Safari and go to Wikipedia" spawned a leaf session; "Now search it for macOS Tahoe" ran in that same session. |

Suites: `swift test --filter Cua` 59, agent `vitest` 810, `pi-mono-extension` 131, tsc clean, `pr-preflight` 29 checks.

A first continuity attempt measured this on `main_chat` and proved nothing, because that surface's conversation journal was already carrying the fact. The control caught it, and the recap is now scoped to sessions where nothing else replays turns.

## Notes for the reviewer

- Synthetic input is covered through its schemas, refusals and permission gating, but posting a real event needs the Accessibility grant macOS only takes from a person. The gestures themselves are untested here.
- The granted-but-stale Screen Recording branch is covered by unit tests and by the policy five other consumers already trust, not by observation — this machine has never granted Screen Recording to the test bundle, so that path has not been watched live.
- `scroll`'s schema and its implementation comment describe the `dy` sign in opposite terms (`CuaToolCatalog.swift` vs `CuaInputSynth.swift`). One real scroll settles it; not changed on reasoning alone.
- A follow-up resolves to the owner's most recently active leaf session. Right today, ambiguous once two leaves run at once. `workstream-continuity.ts` already models the fix, as separate work.
- The stdio shim still writes `~/.omi/omi-cua`, so a third-party stdio client can reach the endpoint even though the UI no longer mentions it. Deliberate for now.

## Product invariants

- INV-AGENT-*
- INV-AUTH-1
- INV-CHAT-1

Matched by path; none changes behavior here. The agent-runtime files carry the MCP tool-result blocks that let a capture server's image reach the model — session, run and attempt identity, authority and lifecycle are untouched. `LocalMcpStore` gains only the built-in entry the gate writes and refuses to let a user Remove; the kernel remains the sole owner of transcript state and nothing here adds a second store.

## Line-count exceptions

Each is a new case in an existing section, not new structure.

Line-Count-Exception: desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift | 4933 -> 4964 | one `register(...)` block for the computer_use action, in the file that is the registry of every automation action; a seam registered anywhere else would not be discoverable where the other 200 live
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/AppsPage.swift | 3880 -> 3886 | six lines at the existing LocalMcpDetailSheet call site — content-hugging height and the appState the Reopen Omi button needs; the sheet's own body lives in ComputerUsePermissionsBlock rather than growing this file

Failure-Class: FC-platform-grant-record-as-live-capability

The class is already open, and this is a recurrence in a new subsystem: computer use was the one screen consumer that read the platform's grant record as proof the running process could capture.

https://claude.ai/code/session_01QnuRGojeKmR3DhiqKWPeXX


Git-on-my-level:
Thanks @katipally — this is careful work. I reviewed the computer-use layer statically (head `040e6a70`), with the stacked-diff situation in mind: the branch is cut from #12377's fork branch, so the diff shown here includes that PR's 40 files (MCP clients, `LocalMcpStore`/OAuth, skills, catalogs). Only the 42 computer-use files are new here, and those are what I looked at — #12377's surfaces belong on its own PR.

**What stands out as well built**

- `CuaMcpEndpoint.swift` — the consent check lives in **one** place (`toolsCallResult`), so a tool that forgot to ask is still covered; handshake/`tools/list` answering while control is off keeps health probes honest.
- `CuaControlGate.swift` — off by default, grant bound to the owner who gave it, sticky suspend, `@MainActor` check-and-post fence. The design comment about interleaving is correct.
- `CuaPermissions.swift` — separating `kTCCServicePostEvent` from `kTCCServiceAccessibility` is exactly right; treating them as one switch is the classic mistake and this avoids it.
- `CuaAppleScript.swift` — out-of-process `osascript` with a wall-clock timeout and SIGTERM→SIGKILL escalation, script staged to a temp file by path (no interpolation), background pipe readers avoiding the `waitUntilExit` deadlock. The reasoning in the header comments (TCC attribution, why not `NSAppleScript`) checks out.
- `CuaAxReader.swift` — refusing to walk its own pid and capping depth/nodes/timeout per call.
- `CuaInputSynth.swift` — real modifier key-down/up around chords (Chromium/Electron read key events, not flags) and layout-resolved printable keys are the details that make this actually work in practice.
- `LocalAgentAPIServer.swift` — the `/mcp` route inherits the loopback bind, Host/Origin checks, and adds its own bearer auth; `republishComputerUseRegistration` self-heals a stale token/port entry at startup.
- `kernel-core.ts` + `session-recap.ts` — recap seeded only when the binding changed and the journal doesn't already replay turns; bounded; emits an event so degradation stays observable.
- `user-extensions.ts` / `pi-mono-extension/index.ts` — strict server-name validation, fail-open parsing, per-server discovery timeouts, and a wedged server never takes the extension down.
- Test coverage is real: 10 new Swift test files, 7 new TS test files, 2 e2e flows, and the relevant CI checks are green.

**Agent-facing guidance added by this PR** (worth maintainer eyes): `.github/agent-docs/desktop-user-extensions.md` (new) and a `desktop/macos/AGENTS.md` section describing the `~/.omi` contract — I spot-checked the doc against the implementation (env var names, fail-open semantics, PKCE construction, plaintext-token trust model) and it matches. It's accurate today, but it will become the contract coding agents rely on, so keep it in scope when these files change later. `RealtimeHubTools.swift` also ships new system-prompt guidance (see below).

**What needs a human maintainer decision before merge** (nothing here is a defect; these are threat-model/product calls):

1. **Consent breadth of the single toggle.** While control is on, the endpoint exposes screen contents, other apps' accessibility trees (including text values), synthetic input, clipboard, and `run_script`. Via `osascript`, `run_script` can reach `do shell script`, which runs arbitrary commands as the user with no additional macOS prompt — so the practical ceiling of this surface is user-level execution behind one switch plus token possession. The layered defenses are good (loopback-only, bearer token, off by default, per-owner binding, sticky kill switch, ⌃⌥⌘.), but the breadth is a decision to make knowingly.
2. **Token reachability.** The local-API token sits in `~/.omi/mcp.json` and inside `~/.omi/omi-cua` (0700). Any same-user process can read it and drive the endpoint while control is on. This matches the trust model of other local MCP clients and is documented as deliberate — but the gap between what the toggle says ("let Omi control this Mac") and what it technically grants is worth an explicit maintainer stance, possibly in the UI copy.
3. **Sequencing.** #12377 must merge first; until then this diff double-counts that PR's files and can't be reviewed as a unit.
4. **Voice-agent behavior change bundled in.** `RealtimeHubTools.swift` adds shipped system-prompt guidance routing follow-ups to `send_agent_message` rather than answering directly — a live behavior change for every realtime-hub user, independent of computer use. The wording matches the tool's actual semantics, but it's a product-behavior choice that deserves its own sign-off (and would be easier to bisect as its own change).

Marking for maintainer review with a security lens on the consent/threat-model points above — not requesting changes, because the engineering is sound and the questions are direction, not correctness.

Git-on-my-level:
Verified the review-fix head (`5f491304`) against the findings raised on `040e6a70` — the computer-use items are closed, and each fix lives where it should:

- `CuaControlGate.swift` — a suspend now persists (`Keys.suspension`, restored in `init`), so relaunching with the switch still on no longer resumes tool calls on its own; gestures moved off the main actor (`performOffMain`) with a `CuaGestureCancel` flag checked between events, so ⌃⌥⌘. also halts a gesture already in flight.
- `CuaToolCatalog.swift` — `CuaArguments.int` rejects NaN/infinity/out-of-range before `Int()` (a malformed argument can no longer crash the app), and an unknown or aged-out `frame` id is refused ("take a new one") instead of silently reading image coordinates as points on the desk.
- `CuaPermissions.swift` — only `.postEvents` revalidates from the cached-grant fast path (its preflight answers live), and the AX probe now marks `.accessibility` alone, with `.postEvents` gated on its own preflight — no more reporting a click as working while the window server drops it.
- `CuaAxReader.swift` — caller-supplied `max_depth`/`max_nodes` are clamped to hard caps (40 / 2,000) before the walk.
- `CuaAppleScript.swift` — each output stream is capped at 1 MB with continued draining and a "cut off" result, so a chatty script can no longer balloon memory.
- `CuaKeyMap.swift` — space is kept as a real keystroke instead of falling back to the Unicode-only path some apps ignore.
- `session-recap.ts` — `orphaned` runs are now recapped, so a daemon restart no longer blanks a follow-up that referred to prior work.

Two residuals worth tracking, neither blocking this PR's own delta:

1. `CuaFrameRegistry` is a process-wide singleton holding the last four frames. With two MCP clients connected at once, a `click`/`type` that omits `frame` resolves against whichever client captured last — a click can land where the other client was looking. Single-client is the normal case, but scoping frames per client (or requiring the id at resolve time) would remove the ambiguity.
2. `agent/src/runtime/mcp-stdio-client.ts` (in #12377's files, double-counted in this diff): a literal `null` line from a local server's stdout passes `JSON.parse` and then throws on `message.id` inside the readline handler, which reaches the process-level `uncaughtException` exit and takes the whole agent runtime down with it. A one-line non-object guard fixes it — belongs on #12377.

On the agent-facing surfaces this PR adds: `.github/agent-docs/desktop-user-extensions.md` and the new `desktop/macos/AGENTS.md` section describe the `~/.omi` contract (mcp.json format, auth shapes, fail-open semantics, plaintext-token trust model). They matched the implementation when spot-checked and are unchanged by the fix commit — but they will become the contract coding agents rely on in this repo, so keep them in review scope whenever these files change. `RealtimeHubTools.swift` also ships new system-prompt guidance routing voice follow-ups to `send_agent_message` (with `sessionId` now optional and defaulted to the most recent run) — a live behavior change for every realtime user, independent of computer use.

The maintainer decisions from the earlier review remain the gate before merge: consent breadth of the single toggle, token reachability from same-user processes, merge sequencing behind #12377, and the voice-routing change. The engineering on this head is in good shape; the security/product posture sign-off stays with a human maintainer.

---

Git-on-my-level:
Re-verified on the current head `03e706a5`: it is the reviewed fix head `5f491304` plus a merge of `upstream/main`, and that merge touches none of this PR's own files (computer-use layer, extensions UI, agent runtime, docs are unchanged), so the conclusions from the earlier reviews stand as written. Desktop and agent checks are green on the merged tree.

Two status notes:

- The layer verified on `5f491304` is byte-identical here — `CuaControlGate`, `CuaToolCatalog`, `CuaPermissions`, `CuaAppleScript`, `CuaAxReader`, `LocalMcpStore+OAuth`, the `send_agent_message`/`session-recap` runtime files, and the agent-facing docs.
- The stdio null-line residual lives in `agent/src/runtime/mcp-stdio-client.ts`, which belongs to #12377's files: on #12377's current head (`15e850ed`) the `readline` handler still does `JSON.parse(line)` and then reads `message.id` without a non-object guard, so a literal `null` line from a misbehaving local server can still take the agent runtime down. A one-line guard fixes it — tracked there, not here.

No label changes: `Security`, `macOS`, and `needs-maintainer-review` all still apply. The maintainer decisions recorded on `5f491304` (consent breadth of the single toggle, token reachability from same-user processes, merge sequencing behind #12377, and the voice-routing guidance in `RealtimeHubTools.swift`) remain the gate before merge.

---
*by AI on behalf of David*

Git-on-my-level:
## Review round 2 — one merge blocker found and fixed

**Fix branch:** `Git-on-my-level/omi@pr-12465-cua-review-round2`, commit `3b406e7462` (your head `03e706a518` + one commit). I can't push to your fork — please cherry-pick or merge. All suites green on top of it (agent 775/775, tsc clean, `swift test --filter Cua` 54/54, `agent-logic-harness --cross-surface-smoke`).

### P1 — fixed: failed control-tool results relabeled `ok: true`

`stringifyToolResult` routed **failed** results into `stringifyProjectedControlSuccess` in both fallthroughs (in-loop no-persist path, and the final budget-exceeded path). A failed control effect whose payload outgrew every bounded form, with persistence unavailable, reached the model as a success — the exact invariant this PR's own precompact path protects ("a missing artifact is not silently relabeled as an untruncated success"). Fixed to `stringifyProviderBudgetFailure` with the artifact ref when one exists.

Notably, `control-tools-tiny-budget.test.ts` drove this path with an **invalid** run id, so it was codifying the relabel — the test now sends a valid id so it measures size-only degradation, and a new test pins failed+oversize+no-persist → `ok:false`.

### P2 — fixed in the same commit
- `relay-tool-result.ts` re-persisted the same oversized payload once per projection pass (~30 writes/result); recovery ref is now memoized per finalized result.
- `screenshot`'s schema advertises `max_long_edge` but the code read `max_width` — the parameter was silently ignored.
- `~/.omi/omi-cua` shim did `|| continue` on curl failure, leaving a stdio client blocked on a reply forever; it now exits so the client sees EOF (consistent with #12377's dead-server ruling).

### P2 — reported for follow-up (not changed)
- `CuaFrameRegistry` is shared across MCP clients: an omitted `frame` resolves through whoever screenshotted last. Explicit-frame guidance mitigates; per-session scoping is the real fix.
- `send_agent_message` continuation resolves the most recent leaf from *any* surface while the manifest copy promises surface-scoped continuation (your notes already track this).
- `explicitSessionId` is now always truthy after resolver injection — pass `raw.sessionId` so intent routers can tell "user named it" from "resolver guessed".
- `read_tool_output`'s 4096/8192 caps derive from `list_agent_sessions`' absent result contract — derive from an explicit constant.
- Tool-result projector keys sections by name; duplicate names collapse omitted-counts and render state.
- `chatFirstMaterializationDeferrals` labels never-attempted batch items `tail_question`.
- `screenContext` is bounded per turn but unbounded across the 64-turn snapshot window (~50KB worst case).
- Pre-existing, worse now that realtime sees it: `readLocalArtifactText` is an uncapped `readFileSync` of whatever `file://` URI a registry row holds.
- Owner change mid-gesture doesn't cancel in-flight events (next call refuses; a typed paragraph in flight finishes).

### Verified by driving the app (on `3b406e7462`, bundle `com.omi.omi-pr12465-cua`)
- Endpoint: `initialize`+`tools/list` → 20 tools; `list_displays` ✓; `run_applescript` 6×7 → 42; clipboard round-trip ✓; bad token → 401.
- Permission honesty: `screenshot`/`ui_snapshot`/`click` each refuse naming exactly their own TCC grant.
- Gate: off → `tools/call` refuses while `tools/list` still answers; ⌃⌥⌘. suspends with reason and survives re-arm; disable removes the `mcp.json` entry + shim, re-enable restores the same Keychain token.

### The scroll `dy` sign — still open, and I couldn't settle it

I tried: a standalone probe reproducing `CuaInputSynth`'s exact CGEvent sequence posted **0 events at a HID listen tap** on this host (preflight true, posts dropped), and the QA bundle itself can't hold the human-only Accessibility grant. So all my "nothing scrolled" observations are inconclusive about your code — the vehicle was blocked, not necessarily the implementation. Both the schema text and the implementation comment claim positive dy = read further down, so it's wording-level until one real scroll with a granted build.

Git-on-my-level:
Pushed the review round directly to this branch (`03e706a518..6fef12b9f4`, 6 commits) — this supersedes the cherry-pick ask in [my earlier comment](https://github.com/BasedHardware/omi/pull/12465#issuecomment-5544962559), since `3b406e7462` is now in the history.

What landed, in two groups:

**Review fixes** (`3b406e7462`)
- **P1:** a failed control-tool result that outgrew every bounded form was delivered as `stringifyProjectedControlSuccess` — a success-shaped projection of a failure. Such results now stay failures (`tool_result_exceeded_provider_budget`), with a behavioral test.
- P2: relay-side persistence memoized so a persisted-null isn't rewritten every loop pass; the dead `max_width` argument now reads the schema's actual `max_long_edge` key; stdio shim exits (EOF) on a failed POST instead of leaving the client waiting on a reply that never comes.

**Product-direction restructure** (`133964adbe`→`6fef12b9f4`): computer control now presents as a regular MCP server instead of a bespoke section.
- The standalone Computer Control section on the Apps page is gone; the built-in entry renders through the **same card component as every other server** (icon / name / "Built-in" / endpoint URL / status text + one Manage button — no switch in the grid).
- The permission grant rows, the on/off toggle, and Stop now / Re-arm moved onto the server's own Manage sheet.
- `LocalMcpStore.removeServer` refuses built-in names, so Remove can never delete Omi's own server; only `CuaMcpRegistration.unregister` can (the registrar that wrote the entry removes it).
- Card status vocabulary: Off / Stopped / Needs N grants / Ready, driven by a shared store with the sheet.

Verification: `swift build` clean; `swift test --filter Cua` 58/58 (4 new behavioral tests); swiftlint + swift-format clean; bounded pre-push gate passed. Exercised end-to-end on a named QA bundle: toggle off/on round-trip (mcp.json unregister + re-register, endpoint 401 → initialize handshake → 20 tools), Stop now → suspension refusal → Re-arm, per-grant refusals naming exactly their own TCC grant, bad token → 401, ⌃⌥⌘. kill switch.

One open item for manual QA (couldn't be settled from the host I was on): the scroll `dy` sign — a standalone probe reproducing `CuaInputSynth`'s exact CGEvent construction couldn't post scroll events at all in that context, so the claim "positive dy = read further down" is still wording-level until someone does one real `scroll {dy: 3}` against a scrollable window with the Accessibility grant approved.

Git-on-my-level (changes requested):
Thanks @katipally — the round-2 fixes and the MCP-server restructure landed cleanly on `6fef12b9f4`, and the computer-use layer re-inspects well. This round is mostly about the ground moving under the branch, plus two small residuals verified on the head.

**What still stands (re-checked on this head)**

- `CuaControlGate.swift` — off-by-default, owner-bound toggle; persisted sticky suspend restored in `init`; the check-and-post stays inside `@MainActor`, and `performOffMain` + `CuaGestureCancel` let ⌃⌥⌘. stop a gesture mid-run.
- `CuaMcpEndpoint.swift` — the single consent check in `toolsCallResult` still covers every tool; handshake/`tools/list` remain answerable with control off.
- `CuaAppleScript.swift` — out-of-process `osascript`, script staged by path, SIGTERM→SIGKILL escalation, capped-and-drained stream readers.
- `LocalAgentAPIServer.swift` — `/mcp` sits behind the same loopback host/origin checks and bearer token as the rest of the local API, and `republishComputerUseRegistration` self-heals a stale entry at startup.
- `.github/agent-docs/desktop-user-extensions.md` — claims spot-checked against code (unawaited refresh, plaintext-token trust model, transport split, registry-only marketplace): accurate. Note for maintainers: this doc and the `desktop/macos/AGENTS.md` pointer will steer future coding/review agents working on these surfaces; the content is descriptive and non-directive, but it declares deliberate security postures (plaintext tokens under `~/.omi`, fail-open token refresh) that deserve one maintainer read before merge.

**Blocking: rebase onto main (now required)**

#12377 merged to main yesterday (2026-09-04 17:32 UTC), after this branch's last main-merge, so the PR currently reports conflicts and cannot merge. The one that matters most when resolving:

- `desktop/macos/agent/src/runtime/mcp-stdio-client.ts` — main's copy spawns user servers through `stdioChildEnvironment()` with a seven-key env allowlist (PATH/HOME/TMPDIR/USER/LOGNAME/SHELL/LANG) precisely so a command server cannot inherit `OMI_AUTH_TOKEN`, `OMI_BRIDGE_PIPE`, BYOK provider keys, or Firebase credentials. This branch's copy predates that and still passes `{ ...process.env }` to the child. Taking the fork side of that conflict would silently reintroduce the exposure — please take main's version and re-apply your fixes on top.
- Also diverged, with newer main-side fixes this branch predates: `LocalMcpStore.swift`, `pi-mono-extension/index.ts`, `mcp-sse-client.ts`, `mcp-client.ts`, `LocalSkillsStore.swift`, `AgentRuntimeProcess.swift`.

**Two small fixes to fold in while rebasing (both verified on this head; both also present on main, so fixing them here fixes both)**

1. `mcp-stdio-client.ts` — a literal `null` line from a local server: `JSON.parse("null")` returns null, then `typeof message.id` throws inside the readline callback. One guard (parsed is a non-null object before reading `.id`) closes it.
2. `LocalMcpStore+OAuth.swift` — the initial authorization and token exchange send the RFC 8707 `resource` parameter, but `refreshExpiredTokens()` neither persists `meta.resource` in the stored `auth` dict nor includes it in the refresh form, so audience-bound hosted providers will reject refreshes.

After the rebase, please re-run the suites from the last round (`swift test --filter Cua`, agent tests, tsc) on the merged tree — no check runs are visible on this head yet.

Human maintainer sign-off before merge remains specifically about the product decisions this PR bakes in: the consent model for computer control (built-in entry in `~/.omi/mcp.json`, the `~/.omi/omi-cua` stdio shim carrying the local-API token, kill-switch semantics) and the marketplace defaults (official registry + `anthropics/skills`), on a surface that can drive the user's Mac.

---

katipally:
Synced with latest main, including #12377. Kept main's version of the stdio env allowlist, so a user's MCP server still can't inherit Omi's auth token or provider keys.

Fixed both review points. A server sending a bare null line no longer crashes the reader, and the OAuth resource value is now saved when you sign in and sent again on refresh, so audience-bound providers stop rejecting refreshed tokens. Older entries without it keep working.

Updated description and modal ui ux.

Also improved this round: if you grant Screen Recording while Omi is running, macOS says it's granted but capture stays dead until restart. Omi used to show "Ready" and tell you to grant a permission you already gave. Now the card says "Restart to apply" and offers a Reopen button. The detail sheet was also cleaned up: it explains what computer control does, shows the permissions inline with green for granted and a red Grant button when missing, and lists the 20 tools.

Tests on the merged branch: swift --filter Cua 59, agent 819, pi-mono-extension 131, pr-preflight 29 checks.

image
image

Git-on-my-level (changes requested):
Round 3 on `19fc9f6` — both residuals from the last round are verified fixed on this head, and the security-sensitive merge risk was handled the right way. The one blocker is the ground moving again, not anything in the branch.

**Resolved since the last round (verified on this head)**

- `desktop/macos/agent/src/runtime/mcp-stdio-client.ts` — main's `stdioChildEnvironment()` with the seven-key env allowlist survived the merge, and the fork's fixes (the literal-`null` line guard, the ref/unref event-loop lifecycle) sit on top of it rather than instead of it. A user-configured command server no longer inherits `OMI_AUTH_TOKEN`, the bridge pipe, or BYOK provider keys, and a bare `null` line from a server can no longer throw inside the readline callback — `mcp-stdio-client.test.ts` carries the regression test.
- `desktop/macos/Desktop/Sources/Chat/LocalMcpStore+OAuth.swift` — the authorization now stores `"resource": meta.resource` in the persisted `auth` dict, and the extracted `refreshForm(auth:…)` includes the RFC 8707 audience on refresh while omitting it cleanly for entries authorized before it was persisted (no fail-closed on legacy data). `LocalMcpOAuthRefreshFormTests.swift` covers both directions, including the empty-string public-client case.

**Blocking: rebase onto current main (again)**

main moved again after this branch's last main-merge — tip is now `b3ec2dae` (#12820, 2026-09-06 00:55 UTC) — and the PR reports conflicts. The overlap is narrower than last round: `omi-tool-manifest.ts`, `pi-mono-extension/index.ts` (+ its test), and the generated surfaces (`GeneratedToolCapabilities.swift`, `GeneratedToolExecutors.swift`, `tests/fixtures/tool-manifest.json`), all from #12803's JIT ledger tools. Suggested resolution: take main's side for those, re-apply the `send_agent_message` patches in `generate-tool-surfaces.mjs`, and re-run the generator so the manifest digests in `GeneratedToolExecutors.swift` recompute. Please re-confirm afterwards that `mcp-stdio-client.ts` still carries the env allowlist.

**Standing strengths re-checked on this head**

- `CuaControlGate.swift` — off by default, grant bound to the owner who gave it, sticky suspend; the check and the event post stay inside `@MainActor`.
- `LocalAgentAPIServer.swift` — `republishComputerUseRegistration` self-heals the `mcp.json` entry at startup when a re-sign minted a fresh Keychain token, so the feature cannot silently vanish behind a 401.
- `LocalMcpStore.swift` — `removeServer` refuses built-in names; only `CuaMcpRegistration.unregister()` can drop the entry, so a user's Remove cannot delete the server the gate manages.
- `CuaAppleScript.swift` — `osascript` in a killable child, script staged by path, streams capped at `maxStreamBytes`, SIGTERM→SIGKILL escalation.
- `session-recap.ts` + `kernel-core.ts` — the recap seeds only where the conversation journal does not already replay turns, `recapSeededBindings` prevents double-seeding on retries, and it is bounded (12 runs / 6k chars).
- `.github/agent-docs/desktop-user-extensions.md` — the new Computer control section spot-checked against code (0700 shim with embedded token, rewrite-on-enable, shared token but separate control switch): accurate. Maintainer note: this doc steers future coding/review agents on these surfaces and deliberately records security postures — plaintext token under `~/.omi`, fail-open legacy refresh — so it deserves one human read before merge.

Checks are green on this head, but they ran before #12803 landed, so after the rebase please re-run the Swift suite (`swift test --filter Cua`), the agent tests, and tsc on the merged tree.

Human sign-off before merge remains about the product calls this bakes in: the consent model for computer control — built-in entry in `~/.omi/mcp.json`, the `~/.omi/omi-cua` shim carrying the local-API token, ⌃⌥⌘. kill-switch semantics — and the permission-honesty UX, on a surface that can drive the user's Mac.

---
*by AI on behalf of David — merge needs a human product decision on the computer-control consent model and the token-under-~/.omi posture; @Git-on-my-level for that sign-off.*
---

aryanorastar:
Keeping this open. Not approving `19fc9f6c8679`. origin/main has no `CuaMcpEndpoint.swift`. Not shrink-only. Hygiene SUCCESS. mergeable CONFLICTING/DIRTY.

PR #8919, opened 2026-07-03
fix: speaker diarization accuracy and action item attribution

thesohamdatta:
This PR fixes two accuracy problems that have been gnawing at multi-speaker conversations for a while — the speaker diarization breaking down with 3+ people, and action item extraction not knowing who the primary user actually is.

Hat tip to @beastoin . This builds on that. need review

---

## What was broken

**Speaker diarization with 3+ speakers (fixes #8917)**

The diarization pipeline used an online centroid-matching algorithm — it processed segments one by one and updated a running centroid each time. Sounds reasonable, but it breaks badly when you have 3+ speakers: the centroids drift with noise, and the order you process segments in changes the outcome. In practice, speakers get merged or swapped, and the transcription becomes a mess.

**Action item extraction had no idea who "you" were (fixes #8918)**

When the LLM extracted action items, it had no context about who the primary user is. It got speaker labels like "Speaker 0" or "Speaker 1" and had to guess. That meant tasks got misattributed, extracted for the wrong person, or written vaguely without clear ownership.

---

### Diarization → offline clustering

Instead of processing segments sequentially, we now collect all speaker embeddings first, then run **Agglomerative Hierarchical Clustering** (SciPy, average linkage, cosine distance) on the full set at once. This is stable and order-independent — same audio always produces the same labels.

Short segments that can't produce reliable embeddings get mapped to the temporally nearest valid cluster after the fact.

Also wired up `min_speakers`, `max_speakers`, `num_speakers` query params on `/v2/transcribe` so callers can pass constraints (e.g. from a calendar invite).

### Action items → grounded on user identity

We already resolve the user's display name earlier in the pipeline. Now we pass it all the way down into the LLM prompt. The model knows it's extracting tasks for "Soham" (or whoever), not "the primary user" in the abstract. Ownership attribution is tighter.

---

## Changes

| File | What changed |
|------|-------------|
| `backend/parakeet/transcribe.py` | AHC clustering, short-segment fallback, memory cleanup, safe logging |
| `backend/parakeet/main.py` | Speaker constraint query params + input validation |
| `backend/utils/llm/conversation_processing.py` | `user_name` param, prompt grounding |
| `backend/utils/conversations/process_conversation.py` | Forward `user_name` to all extraction call sites |
| `backend/tests/unit/test_parakeet_diarization.py` | Stability + constraint tests |
| `backend/tests/unit/test_parakeet_endpoints.py` | Param routing + validation tests |
[...]

PR #12471, opened 2026-08-31
fix(parakeet): cluster batch diarization offline, not by arrival order

### LINKED ISSUE #8917: Backend: Offline agglomerative clustering for robust 3+ speaker diarization (#5470)
## Context

In meetings with 3 or more participants, the current online speaker diarization algorithm fails to accurately separate speakers. Speaker labels get merged or swapped, which compromises transcription accuracy and downstream features like action item extraction. 

A real-world example of this issue in action shows generic tags like `Speaker 1`, `Speaker 2`, and `Speaker 3` displayed in the transcription feed. Under `Speaker 1`, the text simply reads `"You."`, illustrating how the system struggles to resolve who the primary user is versus other meeting participants.

*Important Note: While the screenshot illustrates this in the mobile app UI, the solution must be entirely backend-driven and platform-agnostic. Do not build any dependencies on client-side screen layout or UI state.*

---

## Visual Comparison: Online vs. Offline Clustering

```mermaid
graph TD
    subgraph "1. Online Centroid-Based (Current: Centroid Drift)"
        S1["Segment 1: Spk A"] --> C1(Centroid A)
        S2["Segment 2: Spk B"] --> C2(Centroid B)
        S3["Segment 3: Noisy A"] --> C1_drift(Shifted Centroid A)
        S4["Segment 4: Spk C"] --> C1_drift
    end

    subgraph "2. Offline Agglomerative Hierarchical (Proposed: Stable Linkage)"
        E1[All Segments] --> M[Cosine Distance Matrix]
        M --> L[SciPy Linkage Tree]
        L --> Cl(Clusters: A, B, C)
        Cl --> Final(Attributed Speaker Groups)
    end
```

---

## Root Cause Analysis

The current diarization pipeline in `backend/parakeet/transcribe.py` (specifically `_diarize_segments`) uses an online centroid-based clustering algorithm. For 3+ speakers, this approach breaks down due to three factors:

1. **Order Dependency**: Processing segments sequentially makes the initial centroids highly sensitive to the voice dynamics of the first few speakers.
2. **Centroid Drift**: Matched segment embeddings are averaged into centroids:
   $$\text{centroid}_{\text{new}} = \frac{\text{centroid}_{\text{old}} \cdot n + \text{embedding}}{n + 1}$$
   As noisy or short segments get merged, the centroids drift. This causes the cluster boundary to shift, eventually merging segments from different speakers.
3. **Unbounded Clustering Heuristic**: PyAnnote community pipelines run with no speaker count constraints (e.g., `num_speakers`, `min_speakers`, `max_speakers`), forcing the clustering algorithm to guess the partition size.

---

## Solution Design & Technical Specs

We replace the online centroid matching in `_diarize_segments` with an offline average-linkage **Agglomerative Hierarchical Clustering (AHC)** algorithm using SciPy.

### 1. Exposed API Routing (`backend/parakeet/main.py`)
Expose query parameters for optional speaker count guidelines:
```python
@app.post("/v2/transcribe")
async def transcribe_v2(
    file: UploadFile = File(...),
    diarize: bool = Form(True),
    min_speakers: Optional[int] = Query(None),
    max_speakers: Optional[int] = Query(None),
    num_speakers: Optional[int] = Query(None),
):
    ...
```

### 2. Agglomerative Hierarchical Clustering (`backend/parakeet/transcribe.py`)
Using SciPy, we construct a pairwise distance matrix of all valid segment embeddings (excluding segments shorter than `MIN_SEGMENT_DURATION = 0.6` seconds):
```python
from scipy.cluster.hierarchy import linkage, fcluster
import numpy as np

# Vectorized distance computation and average linkage
X = np.vstack(embeddings)
Z = linkage(X, method='average', metric='cosine')
```

### 3. Constraint Clamping & Tree Cutting
Partition the distance tree at the predefined cosine threshold `SPEAKER_MATCH_THRESHOLD = 0.45` and clamp the outputs to user constraints:
```python
if num_speakers is not None:
    n_clust = min(num_speakers, len(embeddings))
    labels = fcluster(Z, t=n_clust, criterion='maxclust')
else:
    labels = fcluster(Z, t=SPEAKER_MATCH_THRESHOLD, criterion='distance')
    num_clusters = len(np.unique(labels))
    if min_speakers is not None and num_clusters < min_speakers:
        n_clust = min(min_speakers, len(embeddings))
        labels = fcluster(Z, t=n_clust, criterion='maxclust')
    elif max_speakers is not None and num_clusters > max_speakers:
        n_clust = max_speakers
        labels = fcluster(Z, t=n_clust, criterion='maxclust')
```

### 4. Post-Clustering Mapping for Short Clips
Segments that are too short to produce reliable embeddings are mapped to the temporally closest valid segment cluster:
```python
def get_center(s):
    return (s["start"] + s["end"]) / 2.0

assigned_labels = {}
for j in range(len(base["segments"])):
    if j in valid_labels:
        assigned_labels[j] = valid_labels[j]
    else:
        center_j = get_center(base["segments"][j])
        closest_idx = min(valid_indices, key=lambda v: abs(get_center(base["segments"][v]) - center_j))
        assigned_labels[j] = valid_labels[closest_idx]
```

---

## Karpathy Guidelines for Implementation

1. **Surgical Edits**: Touch only the clustering logic inside `_diarize_segments` in `backend/parakeet/transcribe.py` and the parameter schema mapping in `backend/parakeet/main.py`. Do not refactor adjacent transcription or VAD code.
2. **Simplicity First**: Do not add speculative features or abstract the linkage helper. Keep the scipy operations clean and direct.
3. **Goal-Driven Verification**: Ensure all local test suites run and pass. Run verification commands explicitly:
   - `python -m pytest backend/tests/unit/test_parakeet_diarization.py`
   - `python -m pytest backend/tests/unit/test_parakeet_endpoints.py`

---

## Verification Plan

- Run the VoxConverse DER benchmark suite: `pytest tests/container/test_parakeet_der_benchmark.py`
- Verify that aggregate DER drops below 15% for multi-speaker audio clips.
- Verify that memory footprints do not leak or grow exponentially with segment counts.

SakashSrivastava:
## What this fixes

Closes #8917 (and the multi-speaker symptom in #5470).

`_diarize_segments` in `backend/parakeet/transcribe.py` matched each segment against running-mean centroids while walking the file in time order. The partition was therefore a function of **arrival order**, not of the embeddings. Two failure modes follow:

1. **Centroid drift** — each match folded its embedding into `centroids[i] = (centroids[i] * n + emb) / (n + 1)`. As noisy segments merged, the centroid moved away from the voice it represented, until a genuinely different speaker fell inside its threshold.
2. **Short segments inherited by recency** — a clip under `MIN_SEGMENT_DURATION` was labelled `SPEAKER_{len(centroids) - 1}`: whichever speaker was created *most recently*, not the one speaking beside it.

## The change

Collect every eligible embedding first, then partition once with average-linkage agglomerative clustering over cosine distance. The result depends only on the members, so the same voices in any order produce the same partition.

- Short and failed segments inherit the **nearest clustered segment in time** instead of the most recently created speaker.
- Clusters are renumbered by first appearance, preserving the contract that `SPEAKER_0` is the first voice heard.
- Non-finite and zero-norm embeddings are dropped before clustering — scipy returns `NaN` for cosine against a zero vector, which would spread through the linkage matrix into every cluster.
- `/v2/transcribe` gains optional `num_speakers` / `min_speakers` / `max_speakers`, each validated positive. `min > max` is rejected as 422, and so is `num_speakers` combined with either bound — an exact count and a range are contradictory, and silently preferring one would leave the caller believing a bound they set was honoured. `_diarize_segments` keeps a precedence warning for direct in-process callers, which are not bound by the endpoint contract.

**A separate cut threshold, and why.** AHC cuts the dendrogram on *cophenetic merge height* — the mean distance between two whole clusters. `SPEAKER_CLUSTERING_THRESHOLD` bounds a different quantity: one embedding against one drifting centroid. The value does not transfer, so the batch path takes `SPEAKER_AHC_THRESHOLD` (default `0.55`, env `PARAKEET_SPEAKER_AHC_THRESHOLD`). **The streaming threshold is untouched at 0.60.** The 0.55 default is measured, not guessed — sweep below.

**No new dependency.** `scipy` is already pinned in `backend/parakeet/requirements.txt` (`>=1.11.0`) and `backend/requirements.txt` (`==1.14.0`).

## DER measurement

VoxConverse test subset (the 15 clips and RTTMs already committed under `backend/tests/container/voxconverse`), oracle segmentation. Both algorithms consume **identical segments and identical embeddings** (`pyannote/wespeaker-voxceleb-resnet34-LM`, `window="whole"`, T4 GPU), so the delta is attributable to clustering alone. Missed-speech and false-alarm terms are held near zero; this is speaker-confusion DER.

| | all clips | 3+ speakers | 1–2 speakers |
|---|---|---|---|
| online centroid @ 0.60 | 0.1391 | 0.1742 | 0.0709 |
| **offline AHC @ 0.55** | **0.1101** | **0.1303** | **0.0709** |

**3+ speaker DER falls 25% relative, with no movement at all on 1–2 speaker clips.**

Threshold sweep — the choice is a basin, not a tuned point. Every cut in 0.40–0.55 beats the inherited 0.60:

| cut | all | 3+ | 1–2 |
|---|---|---|---|
| 0.40 | 0.1190 | 0.1385 | 0.0812 |
| 0.50 | 0.1149 | 0.1363 | 0.0735 |
| **0.55** | **0.1101** | **0.1303** | **0.0709** |
| 0.60 | 0.1318 | 0.1632 | 0.0709 |
| 0.70 | 0.1553 | 0.1988 | 0.0709 |

0.55 is picked as the point that maximises the 3+ gain while leaving 1–2 speaker DER exactly at baseline.

Per-clip, 12 of 15 are unchanged. The movement:

| clip | speakers | before | after | delta |
|---|---|---|---|---|
| `eguui` | 3 | 0.4980 | 0.3019 | **−0.1961** — was collapsing all 3 speakers into 1 cluster |
| `tnjoh` | 5 | 0.3065 | 0.1791 | **−0.1274** |
| `bvyvm` | 3 | 0.0161 | 0.0290 | +0.0129 — over-splits to 5 clusters against 3 true speakers |

**Stated plainly:** `num_speakers` is a caller-intent knob, **not** an accuracy improvement. Feeding the ground-truth speaker count gives aggregate DER 0.1390 — indistinguishable from baseline. It helps `eguui` (−0.1961) and `msbyq` (−0.0742) but badly hurts `gkiki` (+0.2258), where forcing a sixth cluster shreds a good five-way partition. Callers who genuinely know the count can ask for it; nobody should expect it to lower error.

`bvyvm` is a real regression and I have not tuned it away — doing so would mean fitting 15 clips.

## Scope

Batch path only. `stream_handler.py` keeps online centroid matching — a live stream cannot cluster over segments it has not received yet. That divergence is pre-existing and left for a follow-up rather than widened into this PR.

Two claims in the issue body did not match current code and were not followed. The issue specifies `SPEAKER_MATCH_THRESHOLD = 0.45`, but that is the *enrollment verification* threshold; batch clustering used `SPEAKER_CLUSTERING_THRESHOLD = 0.60`, and a test pins the two apart. The issue also does not mention the existing `SPEAKER_CLUSTERING_MAX_SPEAKERS` cap, which this change preserves as the default ceiling. The endpoint knobs are `Form` fields rather than the `Query` params the issue sketches, matching the existing `diarize: bool = Form(True)` on the same multipart endpoint.

## Product invariants

none — no invariant in `product/invariants/` covers `backend/parakeet/**`. Confirmed with `scripts/pr-preflight --suggest`.

Failure-Class: new

`FC-partition-depends-on-arrival-order` is added in this PR. A grouping over an already-complete input must be a function of its members, never of the order they were visited; folding each element into mutable accumulated state makes the accumulator a function of history instead, so the same members in a different order yield a different grouping and neither result is defensible. The guard surface is the offline partition in `_diarize_segments` plus the order-independence tests in `backend/tests/unit/test_parakeet_offline_diarization.py`.

## Unit verification

**The regression test fails on the pre-fix code.** With the production change stashed and the new test file left in place:

```
pytest tests/unit/test_parakeet_offline_diarization.py
-> 10 failed, 10 passed
```

```
test_chain_partition_is_identical_under_reversed_arrival
E   assert [(0, 1), (2,)] == [(0,), (1, 2)]
```

Three voices at 0°, 45° and 100° form a chain. Arriving forwards the old code grouped {A,B} and left C alone; arriving backwards it grouped {B,C} and left A alone. Same voices, same distances, different answer.

```
test_non_finite_embedding_is_excluded_from_clustering
E   AssertionError: assert 'SPEAKER_1' in {'SPEAKER_0', 'SPEAKER_2'}
```

The old code minted a phantom `SPEAKER_1` from a NaN embedding.

The remaining 8 pre-fix failures are the constraint and threshold-contract tests, which exercise parameters and a constant that did not exist before.

After the change:

| Command | Result |
|---|---|
| `pytest tests/unit/test_parakeet_offline_diarization.py` | 20 passed |
| `pytest tests/unit/test_parakeet_endpoints.py` | 42 passed (36 before the new endpoint tests) |
| Regression set across 6 parakeet modules | 119 passed |
| `scripts/backend-python-format --check` (4 files) | clean |
| `scripts/failure-class validate` | ok, declaration `new` |
| `make preflight` | 25 of 26 checks pass |

The one preflight failure is `desktop-backend-release-boundary-fixtures`, which fails **identically on unmodified `upstream/main`** because `jq` is not installed on this machine — verified in a clean worktree. CI has `jq`.

**Not run locally:** the full `backend/test.sh` suite (the rest of the backend dependency set does not build on Windows without MSVC build tools) and the container DER benchmark, which needs a GPU with a running Parakeet server. The DER numbers above were measured on a Colab T4 against this branch.

## Credit

This picks up abandoned work from #8919 by @thesohamdatta, which established that offline AHC is the right approach here and drew detailed maintainer review before going quiet. The implementation is independent, but the direction is theirs.

### DIFF
--- a/.github/failure-classes/FC-partition-depends-on-arrival-order.json
+++ b/.github/failure-classes/FC-partition-depends-on-arrival-order.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-partition-depends-on-arrival-order",
+  "violated_contract": "A grouping computed over an input that is already complete — clustering, deduplication, bucketing — must be a function of its members and the distance or key that defines the grouping, never of the order those members were visited. Folding each element into mutable accumulated state as it is visited makes that accumulator a function of history instead, so the same members in a different order yield a different grouping and neither result can be defended as the correct one.",
+  "canonical_prevention": "Collect the whole input first, then derive the grouping in one pass over the complete set using an algorithm whose output depends only on the members. Keep incremental matching only where input genuinely arrives over time and a partial answer must be emitted before the rest exists; where a streaming and a batch consumer both exist, the batch path recomputes from the full set rather than replaying the streaming path's accumulation.",
+  "canonical_prevention_artifact": [
+    "backend/parakeet/transcribe.py",
+    "backend/tests/unit/test_parakeet_offline_diarization.py"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    "backend/parakeet/transcribe.py",
+    "backend/parakeet/speaker_math.py",
+    "backend/utils/stt/speaker_clustering.py"
+  ],
+  "status": "open"
+}
--- a/backend/parakeet/main.py
+++ b/backend/parakeet/main.py
@@ -256,6 +256,15 @@ async def transcribe_v2(
             status_code=422,
             content={"detail": f"min_speakers ({min_speakers}) cannot exceed max_speakers ({max_speakers})"},
         )
+    if num_speakers is not None and (min_speakers is not None or max_speakers is not None):
+        # An exact count and a range are contradictory requests. Rejecting is
+        # consistent with min > max above; silently preferring one would leave the
+        # caller believing a bound they set was honoured.
+        REQUESTS_TOTAL.labels(endpoint="v2_transcribe", status="error").inc()
+        return JSONResponse(
+            status_code=422,
+            content={"detail": "num_speakers cannot be combined with min_speakers or max_speakers"},
+        )
     speaker_bounds: Dict[str, Optional[int]] = {
         "num_speakers": num_speakers,
         "min_speakers": min_speakers,
--- a/backend/parakeet/main.py
+++ b/backend/parakeet/main.py
@@ -243,10 +243,24 @@ async def transcribe(file: UploadFile = File(...)) -> JSONResponse | Dict[str, A
 async def transcribe_v2(
     file: UploadFile = File(...),
     diarize: bool = Form(True),
+    num_speakers: Optional[int] = Form(None, gt=0),
+    min_speakers: Optional[int] = Form(None, gt=0),
+    max_speakers: Optional[int] = Form(None, gt=0),
 ) -> JSONResponse | Dict[str, Any]:
     if gpu_worker is not None and not gpu_worker.is_ready:
         REQUESTS_TOTAL.labels(endpoint="v2_transcribe", status="error").inc()
         return JSONResponse(status_code=503, content={"detail": "Model loading, try again shortly"})
+    if min_speakers is not None and max_speakers is not None and min_speakers > max_speakers:
+        REQUESTS_TOTAL.labels(endpoint="v2_transcribe", status="error").inc()
+        return JSONResponse(
+            status_code=422,
+            content={"detail": f"min_speakers ({min_speakers}) cannot exceed max_speakers ({max_speakers})"},
+        )
+    speaker_bounds: Dict[str, Optional[int]] = {
+        "num_speakers": num_speakers,
+        "min_speakers": min_speakers,
+        "max_speakers": max_speakers,
+    }
     upload_id = str(uuid.uuid4())
     file_path = f"_temp/{upload_id}_{file.filename}"
     ACTIVE_BATCH.inc()
@@ -275,14 +289,20 @@ async def transcribe_v2(
                 Dict[str, Any],
                 await loop.run_in_executor(
                     _diarize_pool,
-                    cast(Any, functools.partial(transcribe_file_v2, file_path, gpu_result=gpu_result, diarize=diarize)),
+                    cast(
+                        Any,
+                        functools.partial(
+                            transcribe_file_v2, file_path, gpu_result=gpu_result, diarize=diarize, **speaker_bounds
+                        ),
+                    ),
                 ),
             )
         else:
             result = cast(
                 Dict[str, Any],
                 await loop.run_in_executor(
-                    _diarize_pool, cast(Any, functools.partial(transcribe_file_v2, file_path, diarize=diarize))
+                    _diarize_pool,
+                    cast(Any, functools.partial(transcribe_file_v2, file_path, diarize=diarize, **speaker_bounds)),
                 ),
             )
         return result
--- a/backend/parakeet/transcribe.py
+++ b/backend/parakeet/transcribe.py
@@ -8,7 +8,8 @@ import httpx
 import numpy as np
 from langdetect import detect as _langdetect_detect_raw  # type: ignore[reportUnknownVariableType]  # langdetect ships partial type info
 from langdetect.lang_detect_exception import LangDetectException
-from speaker_math import cosine_distance as cosine_distance, select_speaker_cluster
+from scipy.cluster.hierarchy import fcluster, linkage
+from speaker_math import SPEAKER_CLUSTERING_MAX_SPEAKERS, cosine_distance as cosine_distance
 
 logger = logging.getLogger(__name__)
 
@@ -183,7 +184,12 @@ def _transcribe_via_gpu_worker(file_path: str) -> Dict[str, Any]:
 
 
 def transcribe_file_v2(
-    file_path: str, gpu_result: Optional[Dict[str, Any]] = None, diarize: bool = True
+    file_path: str,
+    gpu_result: Optional[Dict[str, Any]] = None,
+    diarize: bool = True,
+    num_speakers: Optional[int] = None,
+    min_speakers: Optional[int] = None,
+    max_speakers: Optional[int] = None,
 ) -> Dict[str, Any]:
     if gpu_result is not None:
         base: Dict[str, Any] = _transcribe_from_gpu_result(gpu_result)
@@ -191,7 +197,9 @@ def transcribe_file_v2(
         base = transcribe_file(file_path)
 
     if diarize:
-        base = _diarize_segments(file_path, base)
+        base = _diarize_segments(
+            file_path, base, num_speakers=num_speakers, min_speakers=min_speakers, max_speakers=max_speakers
+        )
     else:
         for seg in base["segments"]:
             seg["speaker"] = "SPEAKER_0"
@@ -202,6 +210,14 @@ def transcribe_file_v2(
 
 SPEAKER_EMBEDDING_URL: str = os.getenv("HOSTED_SPEAKER_EMBEDDING_API_URL", "")
 MIN_SEGMENT_DURATION = 0.6
+
+# Average-linkage AHC cuts the dendrogram on cophenetic merge height — the mean
+# distance between two whole clusters. The online path's SPEAKER_CLUSTERING_THRESHOLD
+# bounds a different quantity, the distance from one embedding to one drifting
+# centroid, so its value does not transfer. Measured on the VoxConverse test subset
+# (backend/tests/container/voxconverse): 0.55 cuts 3+ speaker DER from 0.174 to 0.130
+# while leaving 1-2 speaker DER unchanged, and every value in 0.40-0.55 beats 0.60.
+SPEAKER_AHC_THRESHOLD = float(os.getenv("PARAKEET_SPEAKER_AHC_THRESHOLD", "0.55"))
 
 
 def detect_language_from_text(text: str) -> str:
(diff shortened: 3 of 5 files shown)


Git-on-my-level (commented):
Verified locally (CPU, using the repo's own testing.import_isolation fixtures): tests/unit/test_parakeet_offline_diarization.py passes 20/20 on this branch, and 10 of those fail against main's _diarize_segments — including test_chain_partition_is_identical_under_reversed_arrival, which reproduces the arrival-order bug exactly ({A,B},{C} forwards vs {B,C},{A} backwards for voices at 0/45/100 degrees; the pairwise distances 0.293/0.426/1.174 check out independently). tests/unit/test_parakeet_endpoints.py is 40/40 with the four new constraint tests.

Per-file notes:

- backend/parakeet/transcribe.py — the collect-then-partition rewrite is the right shape. _embedding_vector drops non-finite and zero-norm embeddings before linkage(method="average", metric="cosine") (a NaN pair would otherwise propagate through the whole dendrogram); _cut_to_count clamps maxclust into [1, total] (scipy tolerates t > n by returning singletons, verified); _assign_speaker_labels renumbers by first appearance so SPEAKER_0 stays "first voice heard", which is what downstream speaker identification reads; and the no-explicit-max path correctly preserves SPEAKER_CLUSTERING_MAX_SPEAKERS as the ceiling, as the online path had. The separate SPEAKER_AHC_THRESHOLD is a sound distinction — cophenetic merge height is a genuinely different quantity from distance-to-a-drifting-centroid — and the streaming threshold staying at 0.60 is pinned by test_batch_cut_is_separate_from_the_online_clustering_threshold.
- backend/parakeet/main.py — forward-only wiring. Form(None, gt=0) rejects non-positive counts at the framework layer, the min > max 422 short-circuits before any temp file is written, and both executor paths (with and without gpu_result) receive the bounds.
- speaker_math.py (unchanged) — select_speaker_cluster remains for the streaming path; stream_handler.py imports it from speaker_math directly, so dropping the re-export from transcribe.py breaks nothing (nothing else imports it via transcribe).
- .github/failure-classes/FC-partition-depends-on-arrival-order.json — matches the schema and style of the existing FC corpus on main; canonical_prevention_artifact points at real files in this PR.
- Both test files — the order-independence, constraint-precedence/clamping, short-segment inheritance, NaN/zero-vector exclusion, and gating tests all drive the real _diarize_segments rather than a reimplementation. Genuinely good suite.

Two things for a maintainer before merge:

1. This changes batch diarization output for every existing /v2/transcribe caller, not just opt-in knob users — new algorithm plus a re-tuned cut. The VoxConverse evidence in the description is measured, with the bvyvm regression disclosed honestly rather than tuned away, but it is 15 clips on one embedding model, so the shipped default deserves a human sign-off.
2. CI has not actually run on this head: all workflows are action_required pending first-time-contributor workflow approval, so the current green check state only reflects the neutral cubic review. Once approved, Backend Unit Tests should pick up both files via the backend/** path filter.

Minor nit, not blocking: when both num_speakers and min/max_speakers are supplied, precedence is only logged (logger.warning), not rejected — fine for an internal knob, but if these params are ever exposed beyond internal callers, a 422 there would be consistent with the min > max case.

Thanks also for the explicit credit to #8919 — the offline-AHC direction rightly belongs to that earlier work.

Maintainer sign-off wanted on the batch-default clustering change and the new /v2/transcribe parameters; workflow approval is also needed so CI can validate this head on real runners.

---

kodjima33 (approved):
Well-documented fix with real DER measurements, but diff exceeds the 300-line scoped-fix bar and CI hasn't run on this branch yet, so confidence is 3/5 — approve-only, not merging.

Git-on-my-level (dismissed):
Thanks for the follow-up commit — the `num_speakers` + range combination is now rejected with a 422 at the endpoint (verified: `test_v2_rejects_num_speakers_combined_with_bounds` passes, and the in-process precedence warning is kept for direct callers), which resolves the one nit from the earlier review of a543fd7. The failure-class declaration also validates cleanly against the PR body (`scripts/failure-class validate` → OK).

## Blocking: `_cut_to_count` is scipy-version-sensitive and silently ignores the speaker constraints on the stack CI uses

Repro on this head (2b0534f), on the backend unit-test environment installed from `backend/pylock.toml` (scipy 1.14.0, numpy 1.26.4, Python 3.11.15). Note: this lands alongside kodjima33's approval of this head — the algorithmic direction has maintainer sign-off; this request is only about the reproducible defect below, which CI has not been able to surface because the workflows are still awaiting first-time-contributor approval:

- `tests/unit/test_parakeet_endpoints.py` — 42/42 pass.
- `tests/unit/test_parakeet_offline_diarization.py` — 3 failures:
  - `TestSpeakerConstraints::test_num_speakers_forces_exact_partition_size`
  - `TestSpeakerConstraints::test_min_speakers_splits_a_partition_the_threshold_would_merge`
  - `TestSpeakerConstraints::test_num_speakers_takes_precedence_over_min_and_max`

`num_speakers=3` over three near-orthogonal `_voice(i)` embeddings yields `['SPEAKER_0', 'SPEAKER_0', 'SPEAKER_0']` — one cluster. The cause is in `backend/parakeet/transcribe.py`: `_cut_to_count` uses `fcluster(tree, t=bounded, criterion="maxclust")`, and scipy 1.14.0's maxclust collapses near-tied merge heights (here 0.99761 vs 0.99900) into a single cluster, while scipy 1.18.1 returns two. Your 20/20 was real — on a newer scipy. But:

- CI installs from `backend/pylock.toml` → scipy 1.14.0, so Backend Unit Tests will go red on this head once the pending workflow approval lands.
- `backend/requirements.txt` pins `scipy==1.14.0` while `backend/parakeet/requirements.txt` floats `scipy>=1.11.0` (resolves to 1.18.x today), so the constraint parameters would honor `num/min/max_speakers` on one deployed stack and silently ignore them on the other — no error, just wrong labels.

Verified fix direction: cut on merge heights rather than `maxclust` — sort `Z[:, 2]`, cut just below the (n−k)-th height (`np.nextafter(height, -inf)`) with `criterion="distance"`. Checked on both scipy 1.14.0 and 1.18.1: it returns exactly k clusters for these geometries, including the two-identical-vectors edge case. A post-cut assertion that `len(set(labels)) == bounded` (logging rather than silently returning fewer clusters) would harden it further.

## The rest is in good shape

- `backend/parakeet/transcribe.py` — collect-then-partition, the non-finite/zero-norm guard ahead of `linkage(method="average", metric="cosine")`, first-appearance renumbering preserving SPEAKER_0-is-first-heard, and keeping `SPEAKER_CLUSTERING_MAX_SPEAKERS` as the implicit ceiling all check out; the 17 non-constraint tests pass on both stacks.
- `backend/parakeet/main.py` — forward-only wiring: both executor paths receive the bounds, and both 422s short-circuit before any temp file is written.
- `backend/tests/unit/test_parakeet_offline_diarization.py` — the contracts themselves are right; three currently fail for the production reason above, not a harness issue.
- `backend/tests/unit/test_parakeet_endpoints.py` — clean helper (`_post_v2_with_mocked_transcriber`); all four rejection/forwarding behaviors covered.
- `.github/failure-classes/FC-partition-depends-on-arrival-order.json` — matches the corpus schema and style; `canonical_prevention_artifact` paths are real.

Requesting changes for the `_cut_to_count` rework plus a run of the suite on the pylock stack. Beyond that, this still merits maintainer sign-off on the shipped-default change: every existing /v2/transcribe response would come from the new clustering, the three new parameters are a public API surface, and the scipy pin divergence between the two requirements stacks needs a decision.

---

SakashSrivastava:
Confirmed and fixed in d7dd4a575. Built a venv matching backend/pylock.toml exactly (numpy 1.26.4, scipy 1.14.0) and got precisely the three failures you named, my earlier 20/20 was on scipy 1.17.1, which was my mistake for installing latest instead of the lock.

Went with your suggested direction: cut on merge height rather than maxclust. Sort Z[:, 2], keep the first total - k merges, cut just below the next height with np.nextafter(..., -inf) and criterion="distance". It's expressed in the dendrogram's own units so it doesn't depend on how a scipy release resolves ties. 120 passed on 1.14.0 and 120 on 1.17.1, and a direct probe of _cluster_embeddings returns identical labels on both.

Added the assertion you suggested too, exactly-tied heights still can't be split by any single cut, so it now errs toward more clusters (safer than merging two speakers) and logs the shortfall instead of silently returning a different count. test_exactly_tied_merge_heights_report_the_shortfall covers that branch.

On the scipy pin divergence: this makes both stacks produce identical labels, so it no longer changes behaviour but whether parakeet/requirements.txt should stop floating is your call, and I've left it alone.

Git-on-my-level (commented):
Verified on the CI stack — the blocking concern from the previous review is resolved.

`_cut_to_count` in backend/parakeet/transcribe.py no longer uses `criterion="maxclust"`. It sorts the merge heights, cuts just below the `(n − k)`-th height with `np.nextafter(height, -inf)` and `criterion="distance"`, short-circuits `bounded == 1` to `[1] * total`, and logs a warning when exactly-tied heights make the requested count unreachable instead of silently returning a different partition. The cut is now expressed in the dendrogram's own merge-height quantity on every scipy release rather than each release's maxclust tie-breaking, which is what removed the version sensitivity.

Reproduced the CI environment (scipy 1.14.0, numpy 1.26.4 — what `backend/pylock.toml` installs) and ran the suites at this head:

- tests/unit/test_parakeet_offline_diarization.py — 21/21, including the three that failed at 2b0534f on this stack (`test_num_speakers_forces_exact_partition_size`, `test_min_speakers_splits_a_partition_the_threshold_would_merge`, `test_num_speakers_takes_precedence_over_min_and_max`) and the new `test_exactly_tied_merge_heights_report_the_shortfall`, which pins the tie behavior: four voices 90° apart tie two merges at exactly 1.0, asking for 3 yields 4 clusters plus the shortfall warning. Erring toward more clusters rather than merging two speakers is the right failure mode for diarization.
- tests/unit/test_parakeet_endpoints.py — 42/42 (unchanged by this commit, re-checked).

Also spot-checked the cut numerically on scipy 1.14.0: the chain geometry from the earlier review (merge heights 0.29289 / 0.80004) returns exactly 3 clusters for `num_speakers=3`; an exact-duplicate pair asking for 2 warns and keeps the duplicate merged; `num_speakers == total` returns singletons.

On this commit specifically:

- backend/parakeet/transcribe.py — the new docstring records the 1.14.0 vs 1.18.x maxclust divergence with the concrete heights (0.99761 vs 0.99900); good place for that history. The `bounded == 1` branch also avoids a degenerate cut below the smallest merge.
- backend/tests/unit/test_parakeet_offline_diarization.py — the `_voice` docstring now explains the anti-tie perturbation and points at the new tie test, and the suite covers the tie contract explicitly.

backend/parakeet/main.py, backend/tests/unit/test_parakeet_endpoints.py, and .github/failure-classes/FC-partition-depends-on-arrival-order.json are unchanged by this commit; the endpoint wiring was re-verified as part of the 42/42 run (all three parameters forwarded on both executor paths, both 422s short-circuit before any temp file is written).

The earlier changes-requested review is dismissed as resolved. Still with a maintainer before merge: sign-off on the shipped default — every existing /v2/transcribe response now comes from the new clustering, not just callers of the new `num/min/max_speakers` parameters — and approving the workflows so CI can run; no check has executed on any commit of this PR yet, and Backend Unit Tests will pick up both files via the `backend/**` filter once approved.

---

kodjima33 (approved):
Bug fix, high confidence (root cause + DER data, fixes #8917/#5470, regression tests proven to fail pre-fix, not already fixed on main). Holding at approve, not merge: Hygiene check fails on a real (non-flaky, reran and still fails) firestore-query-coverage baseline drift unrelated to this diff — hard floor requires no failing required check.

Git-on-my-level:
Checked the Hygiene red on this head, because the code is not its cause.

What failed: `firestore-query-coverage` inside the Hygiene job's shared PR contract preflight — `baseline registered_serving is 62, current inventory reports 73` and `baseline eligible_serving is 102, current inventory reports 113`. Every other selected check at this head passed, including the failure-class protocol for the new FC entry.

Why: this branch forked from main on Sep 1 (fork point 17721cea9de9) a few hours before two same-day fixes landed — 52d70a6 refreshed `backend/scripts/firestore_query_coverage_baseline.json` to the then-current inventory, and 42e22fda ("stop firestore query-coverage baseline equality from flapping") made count increases advisory instead of errors. The merge-lane checkout (`Merge d7dd4a5 into 17721cea`) therefore executes the branch's own stale copies — the old baseline (62/102) with the old exact-equality semantics — against an inventory that has since grown to 73/113. Main has been green on this check all week.

Verified it is not this diff: overlaying this PR's two backend files onto current main leaves the query inventory fingerprint-identical (127 shapes, registered 80 / eligible 120 either way) and `python3 backend/scripts/firestore_query_coverage.py --check-ratchet` exits 0. The PR touches no Firestore query code.

Fix: a plain rebase onto current main should turn Hygiene green — nothing in the diarization change needs to move. Every other check at this head is green.

State of the changed files, unchanged since the prior automated verification at this head:

- backend/parakeet/transcribe.py — `_cut_to_count` cuts on sorted merge heights with `np.nextafter(height, -inf)` and `criterion="distance"` (stable across scipy releases), `_embedding_vector` drops non-finite and zero-norm vectors before `linkage(..., method="average", metric="cosine")`, and `_assign_speaker_labels` renumbers clusters by first appearance so SPEAKER_0 stays the first voice heard.
- backend/parakeet/main.py — forward-only wiring: `Form(None, gt=0)` rejects non-positive counts at the framework layer, both 422s (min > max, and num_speakers combined with either bound) short-circuit before any temp file is written, and both executor paths receive the bounds.
- backend/tests/unit/test_parakeet_offline_diarization.py — drives the real `_diarize_segments` with hand-placed unit vectors (the 0/45/100-degree chain makes order-dependence arithmetic, not luck), including the tied-merge-heights shortfall contract; 21/21 when last run on the CI stack (scipy 1.14.0) at this head.
- backend/tests/unit/test_parakeet_endpoints.py — six new endpoint tests covering parameter forwarding, both 422 rejections, and non-positive counts; 42/42 at this head.
- .github/failure-classes/FC-partition-depends-on-arrival-order.json — matches the existing FC corpus schema, and `canonical_prevention_artifact` points at real files in this PR.

Also noting with thanks: @kodjima33's approval of this head covers the shipped-default sign-off the earlier review asked for, the workflows have now run on real runners, and combining `num_speakers` with min/max now returns 422 at the endpoint as previously suggested. Once Hygiene is green after the rebase, this looks ready from my side.

---

PR #12473, opened 2026-08-31
fix(desktop): apply memory deletions across devices, and stop resurrecting deleted memories

aryanorastar:
## Summary

> "I am trying to clean up my memories and I cant delete them on the desktop app. They delete on the app, but they do not sync across devices."

Two defects in the desktop client, and they compound: one manufactures memories that only exist locally, the other makes those memories impossible to remove.

This is the same report as #10446, which was closed COMPLETED on 2026-07-29 **with the cause still unexplained** — the rollout-asymmetry theory posted there was refuted on the issue with live prod evidence (`MEMORY_MODE=off`, so no canonical/legacy split exists to desynchronise). Neither defect below is in the rollout config. Both are in the client.

## 1. Nothing ever removed a memory deleted on another device

`reconcileCacheIfNeeded` carried a doc comment saying it "soft-deletes synced local rows the backend no longer has." It did not. The function ended at:

```
log("MemoriesViewModel: Cache reconcile skipped orphan pruning because backend completeness is unknown")
```

Even its `UserDefaults` key said so: `memoriesCacheReconcile_v2_defaultScopeNoPrune_<uid>`.

The refusal was legitimate at the time — it pulled the **default** scope, and pruning against that would delete Archive rows the endpoint omits by design. But the fix for a pull that is narrower than the prune is to widen the pull, not to skip the prune. `GET /v3/memories` already accepts `include_archive`, and `MemoryStorage.softDeleteSyncedOrphans` already exists and is already used after conversation cascade delete.

So the reconcile now pulls with `includeArchive: true` and prunes within `MemoryLayerScope.allIncludingArchive` — the keep-set and the prune scope describe the same population.

**Absence only counts when the pull proves it.** A truncated pull prunes nothing (it saw part of the backend). An empty pull prunes nothing (far likelier a failed or unauthorized read than a genuinely empty account; pruning there would tombstone the whole cache). `softDeleteSyncedOrphans` only touches rows with a non-nil `backendId`, so local-only memories are structurally exempt, and its tier filter compiles to `tier IN (...)` — SQL `NULL` never matches `IN`, so rows with an unknown tier are excluded from the candidate set rather than deleted. Unknown data fails closed toward keeping.

**Runs once per launch, not once per user.** The old gate was a permanent latch: it would converge the cache exactly once and then let it drift for the life of the install. A delete on another device can happen at any time, so a one-shot latch cannot be the mechanism.

## 2. Deleting an already-absent memory put it back

`DELETE /v3/memories/{id}` returns `404 Memory not found` when the row is gone. Every orphan from defect 1 answers 404. `performActualDelete` caught that like any other error, called `restoreMemory`, and reloaded — so the memory disappeared, came back seconds later, and could never be cleared from the machine.

That is both halves of the report, and it matches #10446's "reappear after ~10–15 seconds" exactly: the 4-second undo window, then the delete, then the reload.

Delete is idempotent in intent. The caller asked for the memory to be gone; the backend reporting it absent already satisfies that. The 404 is now swallowed in `APIClient.deleteMemory`, so all three callers (Memories page, automation bridge, insight storage) get it from one place rather than three call-site patches. Other statuses still throw.

`performActualDelete` also now sets `errorMessage` on a genuine failure, like every other mutation on that model. Restoring the row while saying nothing is indistinguishable from the delete being ignored.

## 3. Corrected the stale doc comment (separate commit)

The comment claiming the reconcile prunes is what sent the previous investigation looking at the backend. It now describes what the code does and why the guards exist.

## Verification

| Check | Result |
|---|---|
| `dev-feedback.py --once swift 'APIClientMemoryDeleteIdempotencyTests\|MemoriesCacheReconcilePruneTests\|MemoryReconciliationScopeTests'` | Executed 12 tests, **0 failures** |
| Same suites against the unfixed code | Executed 8 tests, **3 failures** |
| `make preflight` | PR preflight passed: **22 checks** |

New tests pin both directions, because both directions are dangerous:

- **Delete**: 404 completes; 200 completes; 500 and 403 still throw. Swallowing 404 must not become swallowing failures, or a delete that never happened would read as done.
- **Prune**: a memory absent from a complete pull is pruned; an Archive row still present survives while an absent one is pruned; a truncated pull prunes nothing; an empty pull prunes nothing.

`reconcileCacheIfNeeded` gained a `reconcilePageFetch` seam so the guards — the risky part — are exercised without a network.

### Not verified, stated plainly

The cross-device path was not exercised end to end against a real account: that needs a signed-in session with a memory deleted on a second device, and signing into an account is not an action I take on someone's behalf. The prune contract is proven through the seam and the storage primitive; the transport it replaces is unchanged.

## Tradeoff

Launch-scoped reconcile means a full paged pull per launch where there was previously one ever. For a large account that is a background read of several 500-item pages. The endpoint has no `updated_since` delta, so convergence costs a full read; a delta parameter would be the way to reduce it.

`performFullSyncIfNeeded` still performs its own one-time default-scope pull immediately before this one, so first launch now does two full pulls back to back. Left alone deliberately — collapsing them is a refactor beyond this fix — but it is the obvious follow-up.

## Product invariants

None. `scripts/pr-preflight --suggest` reports no product invariants affected by these paths.

Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift | 3654 -> 3702 | The prune this file refused to do is the fix; +48 lines is the prune call, its guards, and the injectable page fetch that makes those guards testable. Splitting a 3.6k-line view model is a refactor with a far wider blast radius than a delete-sync bug fix, and would bury it.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12473?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/MemoriesPage.swift
@@ -1180,19 +1180,34 @@ class MemoriesViewModel: ObservableObject {
     await loadMemories()
   }
 
-  /// One-time cache reconcile. The local SQLite cache can diverge from the
-  /// backend (the source of truth): stale categories after the server-side
-  /// category cleanup, plus "orphan" rows whose backendId no longer exists on
-  /// the backend. This re-pulls every backend memory (fixing categories via the
-  /// normal upsert) and soft-deletes synced local rows the backend no longer
-  /// has. Local-only unsynced memories are preserved. Runs once per user.
-  private func reconcileCacheIfNeeded() async {
+  /// Reconcile the local SQLite cache against the backend, the source of truth.
+  ///
+  /// The cache diverges two ways: stale categories after the server-side category
+  /// cleanup, and "orphan" rows whose backendId no longer exists because the
+  /// memory was deleted somewhere else. Re-pulling fixes categories through the
+  /// normal upsert; pruning fixes orphans, and is the only thing that makes a
+  /// delete performed on another device take effect here.
+  ///
+  /// Pruning is scoped, not blanket. The earlier version pulled the default scope
+  /// and then refused to prune, because pruning a default-scope pull would delete
+  /// Archive rows the endpoint omits by design. The fix is to make the pull cover
+  /// what the prune claims: `includeArchive` widens it to every tier, so the
+  /// keep-set and `MemoryLayerScope.allIncludingArchive` describe the same
+  /// population. `softDeleteSyncedOrphans` only touches rows with a non-nil
+  /// backendId, so local-only memories that were never synced are never pruned.
+  ///
+  /// Absence is only authoritative when the pull was complete, so a truncated or
+  /// empty result prunes nothing and leaves the cache as it found it.
+  ///
+  /// Runs once per launch rather than once per user: a delete on another device
+  /// can happen at any time, so a permanent one-shot latch would converge the
+  /// cache once and then let it drift again for the life of the install.
+  func reconcileCacheIfNeeded() async {
     let userId = UserDefaults.standard.string(forKey: "auth_userId") ?? "unknown"
-    let reconcileKey = "memoriesCacheReconcile_v2_defaultScopeNoPrune_\(userId)"
 
-    guard !UserDefaults.standard.bool(forKey: reconcileKey) else { return }
+    guard !Self.reconciledUserIDsThisLaunch.contains(userId) else { return }
 
-    log("MemoriesViewModel: Starting one-time cache reconcile for user \(userId)")
+    log("MemoriesViewModel: Starting cache reconcile for user \(userId)")
 
     var cursor: String? = nil
     let batchSize = 500
@@ -1202,7 +1217,7 @@ class MemoriesViewModel: ObservableObject {
     do {
       var truncated = false
       while true {
-        let page = try await APIClient.shared.getMemoriesPage(
+        let page = try await fetchReconcilePage(
           limit: batchSize,
           cursor: cursor,
           authorizationSnapshot: authorizationSnapshot)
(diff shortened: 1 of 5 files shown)


Git-on-my-level:
Verified against the current head — the diagnosis in the description matches the code, and both defects are real. This is a clear improvement over a reconcile that logged and did nothing.

**What checks out**

- `MemoriesPage.swift` — the reconcile now closes the loop the old code refused to: `fetchReconcilePage` pulls with `includeArchive: true` so the keep-set covers the same tiers as the `softDeleteSyncedOrphans(within: .allIncludingArchive)` prune, and the permanent `memoriesCacheReconcile_v2_defaultScopeNoPrune_` latch is replaced by a per-launch set that only marks completion after a full, untruncated, non-empty pull. The fail-closed guards (truncated pull prunes nothing; empty pull prunes nothing) are the right defaults for absence-as-evidence.
- `APIClient+Memories.swift` — the 404 swallow in `deleteMemory` is narrowly scoped (`catch APIError.httpError(statusCode: 404, _)`); 500 and 403 still throw, and the new tests pin that. Restoring the row on 404 in `performActualDelete` was the "reappear a few seconds later" half of the report; handling it at the client layer also covers the automation-bridge and insight-storage callers.
- `MemoryStorage.softDeleteSyncedOrphans` (unchanged, now load-bearing) only touches rows with a non-nil `backendId` and `deleted == false`, and its tier filter compiles to `tier IN (...)`, so NULL-tier rows can never match — local-only and unknown-tier rows fail toward keeping. `MemoryRecord.updateFrom` never touches `deleted`, so later upserts cannot resurrect pruned rows either.
- Backend contract: `GET /v3/memories` honors `include_archive` (archive rows admitted when explicitly requested), defaults `device_scope=all`, and the router already passes `include_pending_processing=True` — so the pull population really is what the prune assumes.
- The `errorMessage` on genuine delete failure uses the existing `UserFacingErrorPresentation` `.memoryDeletion` case, consistent with every other mutation on this model.
- `APIClientMemoryDeleteIdempotencyTests.swift` and `MemoriesCacheReconcilePruneTests.swift` pin both directions of both behaviors (absent → pruned; truncated/empty → nothing; 404 → success; 500/403 → error), and the changelog entry matches what the code does.

**Two hardening edges worth a follow-up (non-blocking; both are soft-deletes, no data is destroyed)**

1. The prune is one-directional. If a backend row is temporarily absent from a complete pull and later reappears — the concrete case is user-reject on another device followed by un-reject, since the backend visibility filter removes `promotion.user_review == false` rows from every list mode — the local row was already tombstoned and the upsert path (`reconcileServerMemories` → `updateFrom`) never clears `deleted`, so it stays hidden from local-cache views even though the backend has it back.
2. A memory created and synced while the launch reconcile pull is in flight misses that pull's keep-set snapshot and would be tombstoned by the same pull. Narrow per-launch window, and the memory remains on the backend and in server-page-driven lists, but a grace condition (e.g. only prune rows whose `updatedAt` predates the pull start) would close it.

Neither changes the verdict. Requesting maintainer sign-off for the new prune-on-user-memories behavior specifically — tombstoning local rows based on backend absence is a product/data-handling call about cross-device delete semantics, and the two edges above are the tradeoffs worth a human decision before merge.

---

kodjima33 (approved):
Desktop bug fix, well-diagnosed (prune-scope + 404-resurrection), tests added, CI green. Owner-override merge (fix(desktop) fast lane).

PR #12474, opened 2026-08-31
feat(app): make home a day-at-a-time timeline

mdmohsin7:
Redesigns the app's home screen as a day-at-a-time timeline, from a design mock.

Home was a stack of previews — recent conversations, a daily-recaps carousel, a Today tasks card, a mind-map thumbnail. It now shows **one day**: where you were, what the day was about, the conversations in the order they happened, and the tasks each one produced.

## What it looks like

- **Header** — the day's map full-bleed behind the status bar and the app-bar controls, built from the conversations' own geolocations so it works before a daily summary exists. `‹ Today ›` navigator, the day's headline (tap → recap detail), and the place the day was mostly spent.
- **Timeline** — `time · title · duration` per conversation, chronological, with its tasks nested underneath: dashed completion circle, tappable to complete, and a `Overdue` / `Today` / `Fri` due marker.
- **Collapsed short conversations** — sub-two-minute captures fold behind a counted line you can tap open, so a six-second "yeah, ok" doesn't take as much room as a 48-minute roadmap sync.
- Stepping back past the oldest loaded day pulls more conversation pages rather than showing an empty day.

Kept from the old home: the app-bar connect/device pill, record button and settings gear (untouched apart from a hairline border, below), the live capture card, and the new-user get-started tiles — de-purpled to the neutral palette.

## What was removed

The recents preview, the daily-recaps carousel, the Today tasks card and the mind-map preview. Both surfaces keep their entry points: daily recaps on the Conversations tab (`FolderTabs` toggle), the mind map on the Memories tab. `daily_summary_card.dart` and `today_tasks_widget.dart` had no other callers and go with them; the recap card's map test is **retargeted** at `DayHeader`, which reuses the same flutter_map tile-loading workaround.

## Notable decisions

- **Map tiles.** Carto now stamps `API KEY REQUIRED` across keyless basemap tiles. Switched to Esri's dark canvas, which needs no key, plus its separate reference layer so streets and neighbourhoods are named. `app/lib/pages/conversations/conversation_map_page.dart` is still on Carto and will show the watermark — left alone to keep this diff scoped, worth a follow-up.
- **Map attribution.** The OpenStreetMap attribution line was replaced with the day's place name. Esri's terms ask for attribution the same way OSM's do, and no other map in the app carries any — worth settling once across the map surfaces rather than in this PR.
- **`extendBodyBehindAppBar`** is set only while home is the selected tab, so every other tab keeps the app bar as a solid surface and its existing layout.
- **Collapsing hides nothing.** The line counts the short conversations and opens them, a stricter threshold the user set themselves still wins, and *Show Short Conversations* opts out entirely.
- **New string.** `shortConversationsCount` added in all 49 locales. Languages whose plural morphology needs more than two forms use a count-suffixed phrasing that stays grammatical for every number.

## Verification

- `app/test.sh` → **1636 passed** (was 1632 on the base; net +4 after retargeting the two removed tests).
- `flutter gen-l10n` → the only untranslated message left is the pre-existing `memoryHistoryPartial`, unchanged from `main`.
- `flutter analyze` on `lib/pages/home`, `lib/pages/action_items`, `lib/pages/conversations` → no new issues.
- `.github/scripts/check_brand_ui.py` → `OK: INV-UI-1 — no purple increase`.
- **Exercised on device.** Built and ran on the iPhone 16 Pro simulator (prod flavor, `OMI_APP_PROFILE=mobile_beta`, `OMI_API_BASE_URL=https://api.omiapi.com/`) against a real signed-in account across several iterations: day navigation, the full-bleed map with legible street/neighbourhood names and no watermark, times rendering as `7:34`, and the timeline against real conversations. The app-bar pill sizing complaint was settled by measuring the screenshot's pixel columns — both pills occupy rows 216–323, i.e. 36pt each; the difference was an invisible boundary, now fixed with a hairline.

## Tests added

- `test/widgets/day_header_test.dart` — map camera fit, tile loading (the post-layout camera nudge, without which the backdrop paints blank), place selection and the daily-summary address fallback.
- `test/widgets/day_timeline_entry_test.dart` — tasks under a conversation, the overdue marker, meridiem-free times, completion toggle, duration formatting.
- `test/unit/home_short_conversations_test.dart` — the collapse threshold rule.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mdmohsin7:
Pushed with the `PRE_PUSH_SKIP_PR_PREFLIGHT=1` hatch. Every manifest check passes except `plan-catalog-contract`, which fails on chart values files inside a gitignored `.claude/worktrees/` leftover — not tracked, not in this diff, and not present in a CI checkout. Tracked as #12476.

mdmohsin7:
Worked through all 16. Fixed 14, declined 2 with reasons — pushed in `03907338f1` and `0bedd7e0f9`.

### Fixed

**Plural singulars hardcoded the numeral** (lt, lv, ru, hr, sr — 5 comments). Confirmed against the generated output: gen-l10n compiles `=1{...}` into the CLDR `one` *category*, not an exact match, and in Slavic/Baltic locales `one` covers 21, 31, 41…, so 21 short conversations rendered "1 короткий разговор". Every singular branch now interpolates `{count}` — identical output for count == 1 everywhere, correct for the rest. Applied to all 38 locales carrying a literal numeral, not just the five reported; Arabic and Hebrew keep their word forms since their `one` is exactly 1.

**Duration not localized.** `formatConversationDuration` was a private English-only duplicate of `secondsToCompactDuration`. Deleted; the timeline uses the shared formatter, truncated to whole minutes first — that formatter spells out seconds below ten minutes ("5m 33s"), which is more precision than a glanceable day reads for and takes width the title needs.

**Midnight staleness** (2 comments). Correct, and the snapshot made it worse: before it, any provider tick re-derived the labels; after, an equal fingerprint suppressed exactly that. A timer at local midnight now rebuilds, and the current local day is part of the fingerprint. The selected day deliberately stays put — its conversations did not move, only what "today" means.

**Fingerprint gaps.** Added the effective short-conversation threshold (changing the setting kept the old collapsed split) and `startedAt ?? createdAt` (reconciliation could move a conversation within its day without updating its printed time or order).

**Map camera not refit.** Right: the backdrop is keyed by day and `FlutterMap` applies its fit once, so a day that gained a location while on screen kept the old framing. Refits in `didUpdateWidget` rather than re-keying — re-keying would re-create the map and re-fetch tiles on every new conversation.

**Overlapping day loads** and **pull-to-refresh leaving an off-page day empty.** Both real. Each load now drops its result if a newer day owns the screen, and refresh reloads the selected day afterwards.

**Test locale.** Pinned to `en` so the `formatMediumDate` fixture is locale-independent.

### Declined

**`few`/`two` categories for hr, sl, sr.** The `other` branch is deliberately shaped as a label — "Kratki razgovori: 3" — which sidesteps agreement and is grammatical for any count. Adding paucal forms would be an improvement in idiom, but writing Slovenian dual and Croatian paucal forms is translation work I cannot verify; a wrong guess is worse than the correct-but-plain form. Worth a native-speaker pass rather than a machine one.

**Tasks past the first 100.** Real and confirmed, but not a regression from this PR — the old Today card drew at most three tasks from the same first page, so the gap was just invisible. Loading every page on home would pull the whole task history to draw one day; the right fix is a day- or conversation-scoped fetch. Tracked as #12534.

Verified: `app/test.sh` → 1647 passed; `analyze_ratchet.sh` → passed; `check_brand_ui.py` → no purple increase; exercised on the iPhone 16 Pro simulator against a real account.

PR #12504, opened 2026-08-31
fix(chat): stop sharing one callback_data dict across every router call

sujeito-operator:
`execute_chat_stream` declares `callback_data: Dict[str, Any] = {}`. That dict is built once, when the module is imported — so every caller that omits the argument shares one object, and the router's two setup-timeout branches write into it:

```python
callback_data['error'] = 'setup_timeout'
callback_data['route'] = 'router'
callback_data['answer'] = AGENT_STREAM_TIMEOUT_MESSAGE
```

The first timeout poisons the default for the life of the process. A later defaulted call then reaches the agentic path already carrying a previous request's `answer`, and the `execute_graph_chat`-shaped readers that do `callback_data.get('answer')` return that stale text rather than nothing.

**This is the sibling of #12094.** That PR fixed exactly this shape on `execute_agent_chat_stream` in `utils/agent.py`, and it was merged 2026-08-24. The copy in the router survived it — same parameter name, same defect, same cure.

## Blast radius today, stated honestly

Both production call sites — `routers/chat.py:639` and `routers/chat_generation.py:92` — currently pass `callback_data` explicitly, so **this is latent in production right now, not an active user-facing leak.** What holds it shut is a convention every call site happens to follow, not the signature. The unit tests already rely on the default (`graph.execute_chat_stream('uid1', [message])`, five call sites in `tests/unit/test_chat_async_offload.py`), so the shared dict is genuinely reachable and accumulates across tests in one process; and the next caller who reasonably omits the argument gets a cross-request leak on a per-request server path. I would rather fix the signature than keep relying on the convention.

## The fix

`Optional[Dict[str, Any]] = None` plus `if callback_data is None: callback_data = {}` — byte-for-byte the cure merged in #12094.

## Verification

**What I ran.**

- The regression test added here fails against `main` and passes with the fix. It is behavioural, driven through the production seam rather than by inspecting `__defaults__`: it forces the router's setup-timeout branch on a defaulted call, then makes a second defaulted call and reads the mapping that call actually hands to `execute_agentic_chat_stream`. On `main` that mapping arrives as `{'error': 'setup_timeout', 'route': 'router', 'answer': ...}`; with the fix it arrives as `{}`.
- I reproduced the mechanism standalone first, running both the current signature and the patched one through two defaulted calls:

```
  main today  -> second caller received: {'error': 'setup_timeout', 'route': 'router', 'answer': 'took too long'}
  with fix    -> second caller received: {}
```

- I swept the whole repository for the class rather than fixing only the instance I found: 2,545 Python files, 15 parameters with mutable defaults. Classified by whether the body actually mutates the shared object, **this is the only one that does.** Nine hand the default to another call without mutating it locally (`database/conversations.py:616,705`, `database/memories.py:304,363`, `utils/llm/chat.py:361,981,996`, `utils/stt/streaming.py:759,833`) — latent, and I have not touched them because proving them needs the callees read. Five are read-only in their own body (`utils/llm/chat.py:314,998`, `utils/llm/persona.py:18`, `mcp/src/mcp_server_omi/server.py:141,208`). Expanding this PR to those would be the unreviewable migration AGENTS.md warns against; the list is here so it is not lost.

**What I could not run, stated rather than implied.** I could not execute `backend/test.sh` or `make preflight` — the backend's runtime dependencies (typesense, firebase, langchain, openai clients) are not installable in my environment, so the suite cannot import. The new test is written against the existing patterns in `test_chat_async_offload.py` and I verified it parses, but **CI is the first place it runs green end to end.** If it needs a fixture adjustment, tell me and I will fix it.

## Failure class

Registered as new rather than filed as a one-off, because #12094 and this PR are the same contract violated twice in one package, which is the "two or more recent fixes share the cause" trigger. The entry records what a linter cannot: `ruff --select B006` flags the declaration mechanically, but the judgement that separates a defect from a smell is whether the body mutates the shared object. `evidence_prs` is `[12094]`, the merged first instance, following the precedent set in #12272.

Failure-Class: new

---

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Git-on-my-level:
Verified end to end — clean, correct fix, and the PR description's framing held up under review.

`backend/utils/retrieval/graph.py` — `callback_data: Optional[Dict[str, Any]] = None` plus `if callback_data is None: callback_data = {}` is exactly the cure merged in #12094 for `execute_agent_chat_stream` (utils/agent.py:107-111), applied to its router sibling. Both router setup-timeout branches do write `error` / `route` / `answer` into the dict, so the shared-default hazard was real. I also confirmed both production call sites (routers/chat.py:639, routers/chat_generation.py:92) pass `callback_data` explicitly today — appreciated that the description states the blast radius honestly (latent hazard, not an active leak) rather than overselling it.

`backend/tests/unit/test_chat_async_offload.py` — the new test drives the production seam instead of introspecting `__defaults__`: the first defaulted call trips the setup-timeout branch (patching the module-level `AGENT_STREAM_SETUP_TIMEOUT_SECONDS`, same pattern already used at lines 270 and 768 of this file), and the second defaulted call captures what the agentic stream actually receives and asserts it is empty. That is the right shape for a regression test here, and CI confirms it behaves with the fix.

`.github/failure-classes/FC-definition-time-mutable-default-shared-across-calls.json` — schema-conformant with the existing corpus, and registering the class fits the repo's own "two or more recent fixes share the cause" rule now that #12094 and this PR both exist. The contract text is accurate against the code, including the defect-vs-smell distinction. One optional follow-up, per the entry's own sweep advice: `answer_simple_message_stream` (utils/llm/chat.py:314) and its sibling near line 998 still declare read-only `callbacks: List[Any] = []` defaults — smell rather than defect since the bodies never mutate them, but they are the natural next sweep if you want it.

Leaving for human maintainer review before merge: the fix itself is mechanical and verified, but this PR also adds a new entry to the failure-class corpus that repo automation and AI-assisted fix workflows consult, and it is the author's first contribution here — a maintainer should eyeball the FC wording before it becomes institutional memory.

---

kodjima33 (approved):
Confidence 5/5: stated root cause, fix(chat) title, scoped diff (59+1/3 files), CI green, adds regression test mirroring merged sibling fix #12094. Verified NOT already fixed on main (callback_data mutable default still present in execute_chat_stream). No abuse signal in diff.

Issue #12476, opened 2026-08-31
plan-catalog-contract false-positives on gitignored .claude/worktrees during local preflight


mdmohsin7:
`backend/scripts/generate_plan_catalog.py --check` scans for embedded Stripe object literals with a bare `os.walk(ROOT)` (around `scan_embedded_stripe_ids`), filtered only by `SOURCE_SCAN_EXCLUDED_PARTS`. That set does not include `.claude/worktrees/`, which is gitignored (`.gitignore:233`) and on a working machine holds many leftover worktrees from earlier sessions.

Result: `make preflight` and the `scripts/pre-push` gate fail with `plan-catalog-contract` errors naming chart values files under `.claude/worktrees/<some-old-branch>/backend/charts/...` — paths that are not tracked, not in the diff, and not present in a CI checkout. The check passes in CI and blocks only locally, so it reads as a broken gate rather than a real contract violation, and the only way past it is the `PRE_PUSH_SKIP_PR_PREFLIGHT=1` hatch (used to push #12474).

Hit while pushing #12474, whose diff touches no plan catalog, chart, or Stripe surface at all. Suggested fix: exclude gitignored paths from the scan (or at minimum add `.claude` to `SOURCE_SCAN_EXCLUDED_PARTS`), so the check sees the same tree CI does.

aryanorastar:
Fixed by #12678, merged 2026-09-03. Closing.

`generate_plan_catalog.py`'s Stripe-literal scan now skips what `git ls-files --others --ignored --exclude-standard --directory` reports, so a local run sees the same tree CI does and `.claude/worktrees/` no longer produces `plan-catalog-contract` failures on files that are untracked and absent from the diff.

It also covers the case your suggested fix would have missed: adding `.claude` to `SOURCE_SCAN_EXCLUDED_PARTS` prunes ignored *directories*, but an ignored *file* sitting in a tracked directory was still being read. Deferring to git covers both, and needs no list extended for the next ignored sibling.

Covered by `test_source_scan_skips_gitignored_sibling_worktrees`. Reproduced first — planting a real `.claude/worktrees/<name>/backend/charts/values.yaml` in a checkout produced your exact failure against the unfixed script.

#12572 was the same root cause and auto-closed on merge; this one did not, hence the manual close.

cwklurks:
There is still an ignored-file case on `main` at `7163538913e15dd86534b41dc695e2a5639c0774`.

[...]

PR #12537, opened 2026-09-01
fix(app): settle the home day timeline's header, map and live capture

mdmohsin7:
Follow-ups to #12474 from running the day timeline on a device. All of it is polish on the merged home screen; no behavior outside home changes.

## Header

- **Spacing.** The day name sat flush under the app bar controls and the recap headline ran straight into the first conversation. Both get room.
- **Weight.** Everything was heavy at once. Day title 28/w700 → 26/w600, recap headline 19 → 17, conversation titles 17/w600 → 16/w500.
- **Map labels are gone.** Dropping Esri's reference layer wasn't enough — the *base* tiles carry road and place names from zoom 15 up. Verified by fetching tiles directly: `.../World_Dark_Gray_Base/MapServer/tile/14/...` is wordless, `tile/15/...` carries street names. The camera and `maxNativeZoom` are capped at 14, so a labelled tile is never even requested.
- **Pull-to-refresh.** Overscroll translates the sliver down, and with nothing above the header the black page showed through. The backdrop now reaches 220pt past the top of the screen so the pull drags map instead. Its scrim is anchored to the header portion via a `LayoutBuilder`, so the fade lands in the same place whether or not the list is being pulled.

## The day-switch flicker

The backdrop was keyed by day, so every step built a **fresh** map with an empty tile cache: `AnimatedSwitcher` cross-faded a loaded map out and a blank one in, and the tiles popped back as they downloaded. Read as a flicker; a longer fade would only have slowed it down.

One map now persists across days and re-frames itself through the `didUpdateWidget` refit, keeping the tiles it already has. The switcher goes back to what it is actually for — days with a map versus days without one.

The guard is behavioral, not a comment: `day_header_test.dart` pins the `MapController` identity across a day change, so re-keying the backdrop later fails the test rather than quietly restoring the flicker.

## Live capture

- **The card looked out of place.** A filled, rounded panel dropped into a flat timeline read as a component from another screen. Same content, card chrome removed, aligned to the timeline's own 24pt inset. `ConversationCaptureWidget` has no other caller, so nothing else is affected.
- **The transcript preview froze.** It rendered `'... <segment> ...'` with `overflow: ellipsis`, but recognition grows the *same* segment word by word — so the line kept showing the words that arrived first and hid everything said since, appearing to advance only when a new segment began. It now stays pinned to the end of the text, so what you read is what was just said; a line short enough to fit still reads from the left.

## Verification

- `app/test.sh` → **1678 passed** (+3: map identity across days, and the two live-preview cases).
- `app/scripts/analyze_ratchet.sh` → passed.
- Esri tile inspection at z14/z15/z16 to establish the label threshold rather than assuming which layer carried them.

**Not visually verified.** Every change here is visual and none of it has been seen running — the device build was stopped partway and I did not re-run it. The live-capture pieces additionally need an active recording to render at all. Worth a look before this lands.

Failure-Class: FC-durable-state-scoped-to-view-lifetime

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12537?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mdmohsin7:
Pushed with the `PRE_PUSH_SKIP_PR_PREFLIGHT=1` hatch. Every manifest check passes except `plan-catalog-contract`, which fails on chart values files inside a gitignored `.claude/worktrees/` leftover — not tracked, not in this diff, absent from a CI checkout. Tracked as #12476.

mdmohsin7:
Fixed 3 of 4 in `80ffeaff0c`; the fourth I think is a misread.

**AnimatedSize clipping the bleed** — correct, and the more serious of the two: `AnimatedSize` defaults to `Clip.hardEdge` (checked in the pinned SDK), so the 220pt bleed was clipped away and pull-to-refresh still exposed the black page. The overscroll fix in this PR rendered nothing at all. Now `Clip.none`; the viewport still clips what is genuinely off screen.

**Tile nudge clamped at the cap** — also correct, and worse than an edge case. The post-layout nudge exists because flutter_map applies its fit without scheduling tiles; nudging *up* from `maxZoom: 14` clamps straight back, so no camera change and no tiles. Zoom 14 is exactly where a single-location day lands, which is the common case. It now nudges away from whichever limit the camera is on.

**Test asserted the mechanism, not the outcome** — fair. It checked the `reverse`/`scrollDirection` flags and would have passed with the newest words invisible. It now measures where the line sits: tail flush with the trailing edge, head off view, still pinned after the segment grows. I mutation-checked it by flipping `reverse` off and confirming it fails.

**Capture card inset** — I don't think this one holds. The row's leading element is the status chip, and the chrome is `margin: 14` + the container's existing `padding: 10` = **24**, which is the timeline's inset exactly. The 8px referenced is `fromLTRB(8, 0, 8, 0)` on the *transcript* in the middle of the row, not a leading inset. The chip's own `horizontal: 12` insets its label inside the pill, which is what a pill should do — its box edge is what aligns. Happy to be shown otherwise with a measurement.

Verified: `app/test.sh` -> 1678 passed; `analyze_ratchet.sh` -> passed.

PR #12538, opened 2026-09-01
feat(app): show Transcribe Later captures on the home day timeline

mdmohsin7:
The day timeline shipped in #12474 renders server conversations only. Transcribe Later captures audio to the phone and turns it into a conversation later, so between stopping a recording and the upload finishing, that audio existed on the device and nowhere on home.

## What was wrong

- **A day spent recording offline read as empty.** Home showed *"No conversations yet"* while the Conversations tab listed the very same recordings — it already interleaves them into its day groups.
- **A first-run user was told to start recording after doing exactly that.** `isNewUser` tested `conversations.isEmpty` alone, so somebody who had only ever used Transcribe Later got the get-started tiles. The conversations page has always guarded this case with `!hasRecordings`; home did not. This is the part I'd call a bug rather than a gap — it actively misinforms.
- **The map ignored locations it had.** `LocalRecording` carries a start-location snapshot, so an unsynced day *does* know where it happened.

The path after sync was already correct: an uploaded recording becomes a conversation with its original `startedAt`, so it lands on the right day retroactively.

## What changed

The day snapshot carries the day's recordings and interleaves them with its conversations by start time, **reusing the conversations page's own `buildConversationGroupEntries`** rather than defining a second answer to "what happened that day, in order" — two implementations of that would drift.

A recording renders in the same shape as a conversation, because on the day it happened that is what it is: time, what it is, how long it ran. Having no title yet, its second line carries what the user actually needs — audio still on the phone, on its way, or failed to send — using the same localized status strings the conversations list uses, so no new strings. Tapping it opens the same detail sheet (transcribe / share / delete).

Two deliberate calls:

- **Recordings are never folded into the short-conversations group.** A short conversation is noise; an untranscribed recording is asking the user for something.
- **Conversation addresses still win for the place name**, with recordings as the fallback, so a synced day reads the same as before.

## Verification

- `app/test.sh` → **1683 passed** (+5: a recording placed in its day and ordered against a conversation, kept out of the short group, the fingerprint moving when an upload progresses, other days ignored, and a header map drawn from a recording's location alone).
- `app/scripts/analyze_ratchet.sh` → passed.

**Not visually verified.** As with #12537, none of this has been seen running — reproducing it needs a Transcribe Later recording captured and left unsynced. The snapshot and map behaviour are covered by tests; the row's appearance is not.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12538?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/pages/home/day_snapshot.dart
+++ b/app/lib/pages/home/day_snapshot.dart
@@ -1,6 +1,8 @@
 import 'package:flutter/foundation.dart';
 
 import 'package:omi/backend/schema/schema.dart';
+import 'package:omi/models/local_recording.dart';
+import 'package:omi/pages/conversations/widgets/conversations_group_widget.dart';
 import 'package:omi/providers/conversation_provider.dart';
 
 /// Everything the home day view renders for one day, plus a fingerprint of the
@@ -14,7 +16,8 @@ import 'package:omi/providers/conversation_provider.dart';
 class HomeDaySnapshot {
   const HomeDaySnapshot({
     required this.conversations,
-    required this.highlights,
+    required this.recordings,
+    required this.entries,
     required this.shortOnes,
     required this.tasksByConversation,
     required this.isNewUser,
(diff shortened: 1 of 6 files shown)


mdmohsin7:
Pushed with the `PRE_PUSH_SKIP_PR_PREFLIGHT=1` hatch. Every manifest check passes except `plan-catalog-contract`, which fails on chart values files inside a gitignored `.claude/worktrees/` leftover — not tracked, not in this diff, absent from a CI checkout. Tracked as #12476.

mdmohsin7:
Valid, and a regression from this PR rather than a pre-existing one — reshaping the snapshot around mixed entries moved `isEmpty` from the day's conversations to `entries`, which deliberately excludes whatever folded into the short-conversations group. A day whose conversations were all under two minutes rendered its own "3 short conversations" line with "No conversations yet" directly beneath it.

Fixed: emptiness is now about whether the day holds anything at all — no conversations and no recordings. Mutation-checked by restoring the `entries`-based version and confirming the new test fails.

Verified: `app/test.sh` -> 1684 passed; `analyze_ratchet.sh` -> passed.

PR #12544, opened 2026-09-01
fix(backend): skip redundant legacy task scans

aryanorastar:
## What changed and why

Part of #11282.

Production attribution on 2026-08-25 showed `action_items_list` at 5,079 document reads/second, about 36% of total Firestore read spend. The default action-item list reads the active bucket first, but when that bucket does not fill the page it also ran an unfiltered legacy compatibility scan of 128–2,000 documents before reading the completed bucket. For canonical collections, that scan paid to prove that every row already had a concrete `completed` boolean.

This change replaces that unconditional compatibility scan with two metered count aggregations: canonical boolean-valued rows versus the collection total. Equal counts prove that the broad scan has nothing to recover; unequal counts keep the released scan so historical missing/null rows remain visible. Aggregation errors also recover through the old bounded scan and emit shared fallback telemetry. Aggregation billing is charged to both the request read budget and `action_items_list` family metric, and the new serving query shape is registered without adding a redundant composite index.

On the regression fixture, the same 10-row response now costs 44 billed operations instead of 170 (74% fewer). This is one measured reduction for the persistent alert, not a claim that the production incident is closed; #11282 should remain open until deployed metrics confirm the alert recovers.

Firestore bills count aggregations at one document read per batch of up to 1,000 index entries, with a one-read minimum: [Firestore pricing](https://firebase.google.com/docs/firestore/pricing#aggregation_queries), [aggregation query documentation](https://firebase.google.com/docs/firestore/query-data/aggregation-queries).

Line-Count-Exception: backend/database/action_items.py | 1408 -> 1543 | keep the compatibility decision, request-budget accounting, and fallback beside the existing bounded list assembler; splitting this legacy module is outside this incident-sized change

none

## How it was verified

- `PYTHON=.venv/bin/python bash test-preflight.sh` — 18 checks passed; only optional dependency/service warnings.
- `.venv/bin/pytest -q tests/unit/test_bounded_firestore_list_reads.py tests/unit/test_action_items_pagination_order.py tests/unit/test_list_read_budget_contract.py tests/unit/test_action_item_ids_endpoint.py tests/unit/test_firestore_document_probe.py tests/unit/test_firestore_indexes.py tests/unit/test_firestore_query_contract.py` — 139 passed.
- `.venv/bin/pyright --pythonpath .venv/bin/python database/action_items.py database/firestore_index_registry.py` — 0 errors.
- Before-path fixture at the parent commit — 170 billed reads for 10 returned rows from a 305-row collection.
- `firebase emulators:exec --only firestore --project demo-action-items-read "PYTHONPATH=backend backend/.venv/bin/pytest -q backend/tests/integration/test_action_items_legacy_probe_emulator.py"` — 1 passed against the real Firestore emulator; the canonical path recorded 44 billed operations and the missing/null legacy path returned every row.
- `BACKEND_UNIT_TEST_FILE_LIST=/tmp/omi-backend-unit-failures.txt bash test.sh` with `tests/unit/test_desktop_migration.py` — 114 passed; verifies the hermetic desktop migration harness can import the new fallback-telemetry dependency.
- `make preflight` and `scripts/pre-push` — passed before push.

## Tests

`backend/tests/unit/test_bounded_firestore_list_reads.py` now covers:

- canonical collections skipping the broad legacy scan while charging both aggregation batches;
- missing and null `completed` rows retaining legacy visibility;
- aggregation failure recovering through the released bounded scan with fallback telemetry;
- unbudgeted aggregation timeouts retaining legacy rows, plus partial aggregation billing on later failure; and
- aggregation billing across the 1,000-index-entry boundary.

`backend/tests/integration/test_action_items_legacy_probe_emulator.py` executes the production query shape against the real Firestore emulator and proves both the optimized and compatibility branches.

## Failure class (fixes)

Failure-Class: FC-bounded-read-exceeds-request-budget

## Scoped cleanups

- Refreshed the query-coverage aggregate counts. `origin/main` already failed its ratchet at 62 registered / 102 eligible versus 73 / 113 in the inventory; this PR's registered query makes the honest totals 74 / 114. The raw-unregistered and unsupported hash sets are unchanged.


Git-on-my-level:
Thanks @aryanorastar — this is a careful, well-measured change. I reviewed the full diff on head `ba48ca3` and the supporting modules; file-by-file notes below.

**backend/database/action_items.py**
- The skip condition is conservative in the right direction: `completed IN [False, True]` excludes exactly the missing/null (and any non-boolean) values, so `canonical_count != total_count` keeps the released bounded scan whenever anything non-canonical exists. Legacy rows with non-boolean values also produce unequal counts, so the old harvest semantics are only bypassed when every row carries a concrete bool.
- The canonical-before-total ordering means a concurrent insert between the two counts can only cause an unnecessary scan, never hide a pre-existing legacy row — the docstring claim checks out against the interleavings.
- Failure handling fails open to released behavior everywhere: unbudgeted `FirestoreDeadlineExceeded` and the aggregation-error tuple both emit `record_fallback(...)` and keep the scan; the budgeted timeout marks the budget exhausted and truncates consistently with the rest of the list path. The `finally` ledger also attributes a successful first count even when the second fails or the budget raises — that resolves the partial-billing undercount an earlier automated review pass flagged, and the timeout-path split resolves its unbudgeted-timeout finding.
- Charging both aggregations to the request budget and the `action_items_list` family metric at one read per 1,000 index entries with a one-read floor matches Firestore's documented aggregation billing. Worst case for legacy-bearing collections is the probe (~2-4 reads) on top of the retained scan — negligible next to the 128-2,000-doc scan this removes for canonical collections.

**backend/database/firestore_index_registry.py**
- `ACTION_ITEMS_CANONICAL_COMPLETION_COUNT_QUERY` correctly leans on the automatic single-field index (`_asc('completed')`) instead of adding a composite manifest entry, while still making the serving shape auditable in `QUERY_SPECS`. The two-value `in` filter is well within Firestore's IN limits.

**backend/scripts/firestore_query_coverage_baseline.json**
- The regen to 74 registered / 114 eligible is honest: one newly registered query plus the environment drift that was already visible on `origin/main`'s committed totals; the `raw_unregistered` and `unsupported` fingerprint sets are unchanged, so this doesn't paper over any new unregistered shape.

**backend/tests/integration/test_action_items_legacy_probe_emulator.py**
- Real-emulator proof of both branches — canonical skip at 44 billed ops, and a missing/null/legacy collection returning every row at 8 — and `_assert_emulator_only()` correctly refuses non-loopback hosts.

**backend/tests/unit/test_bounded_firestore_list_reads.py**
- Good coverage of the new decision points: skip-when-canonical, keep-when-legacy, aggregation-failure fallback telemetry, unbudgeted timeout recovery, partial probe billing on failure and truncation, and the 1,000-entry billing boundary. Tightening `_FakeQuery`'s filter semantics (missing-field exclusion, identity `==`, new `in` operator) also brings the fake closer to real Firestore behavior.

**backend/tests/unit/test_desktop_migration.py**
- Minimal stub for the new `utils.observability.fallback` import so the hermetic desktop harness keeps importing — appropriate scope.

No blocking concerns found, and the backend checks (hermetic merge gate, emulator proofs, unit suite) all pass on this head. This is a cohesive incident-sized change; no split needed. Keeping #11282 open until deployed metrics confirm the alert recovers is the right call.

Leaving for human maintainer review: this changes the production Firestore read-spend hot path (`action_items_list`) and request-budget accounting, so maintainer merge sign-off is recommended.

kodjima33 (approved):
Backend, confidence 4/5: well-scoped Firestore read-amplification fix (part of #11282), measured before/after proof, integration test, safe fallback, CI green. Merging.

PR #12546, opened 2026-09-01
fix(backend): run ledger drain independently

### LINKED ISSUE #12535: Ledger drain never runs: memory-maintenance-job dies at its 3600s timeout every hour, so no account has ever reached writer_mode=ledger
## Finding (2026-09-01 overnight, read-only dev/prod inspection)

**No account has ever been cut over to `writer_mode=ledger`** — both dogfood/allowlist accounts sit at `compatibility`, `writer_epoch=0` (data-plane `memory_state/apply_control` read 2026-09-01). Consequently the JIT daily sweep has never committed a ledger row for anyone: its writes are ordinary ledger-class writes, which `require_writer_admitted` refuses in compatibility mode (`WriterAdmissionError`, named by the sweep job since #12527 deployed).

## Why the drain never runs

The Beta-scale ledger drain (#12452, `canonical_short_term_maintenance_cron`) runs **after** the maintenance pass completes, gated on `summary.completed_uids`. But dev `memory-maintenance-job` is killed at its 3600s task timeout on **every hourly execution** (checked 2026-09-01; each run streams canonical consolidation promotions until the kill). The drain tail is never reached, so:

- accounts never migrate/cutover to ledger mode;
- the sweep fails (until #12533's follow-up lands, it failed the whole job; after the pre-drain skip it will quietly skip) for every enrolled account, forever;
- flipping `jit-processing-v1` to the full beta cohort will enroll ~146 accounts whose sweep lane can never activate.

## Suggested repairs (separable, in rough order of leverage)

1. Give the ledger drain its own bounded entrypoint/job (N accounts per run) instead of riding the tail of an unbounded consolidation pass — the drain's own bounds (`MAX_LEDGER_MIGRATION_UIDS_PER_RUN`) already exist.
2. Or checkpoint the maintenance pass so a timeout kill doesn't discard the run (it currently restarts the same backlog hourly and never converges).
3. Measure whether the consolidation backlog is actually converging; a job killed at 100% of its budget every hour may be re-processing the same head of the backlog.

Context: found during the JIT beta-flip readiness night; related merges #12524 #12525 #12527 #12529 #12533, evidence in the workspace run record `data/run-2026-09-01-overnight/02-jit-production-evidence.md`.

aryanorastar:
Ledger migration currently runs after short-term maintenance, so a maintenance timeout can prevent any account from reaching ledger mode. Move migration into an independent, bounded Cloud Run job with its own scheduler, retryable inventory, fresh JIT authorization at mutation/publication boundaries, and generation-fenced cursor updates.

The job has a dedicated default-off switch and requires an explicit account allowlist. Development reads only the single QA owner account directly; production remains disabled with an empty allowlist. Scoped runs do not advance the shared inventory cursor. The job keeps a 1,200-second timeout, one retry, bounded per-account migration, and the existing cutover/privacy fences.

Validation: focused backend tests, runtime manifest composition, workflow contracts, and repository preflight. Dev deployment and live migration/sweep evidence are verified separately from CI; no production deployment or enablement is included.

Fixes #12535.

Product invariants affected: INV-DATA-1, INV-MEM-1, INV-MEM-4. The drain retains explicit data-plane routing, canonical mutation authority, and bounded migration/cutover checks.

Failure-Class: new

Runtime gate: the job remains disabled by default. Existing development jobs share customer Firestore with production, and both allowlisted identities are used by normal Beta clients. Do not execute migration against either shared account. Runtime proof must use a separately verified isolated dev data plane, include #12802, and pass dev readiness. Source merge does not admit migration work.


Git-on-my-level (dismissed):
Thanks for this — clean diagnosis of #12535 (bounded drain code never entered because the 3600s maintenance budget is consumed by serial consolidation) and a well-scoped fix. The core move is sound and the safety invariants survive the extraction. Details below, one blocking item, several non-blocking notes.

**Blocking: Backend unit suite fails on the new test file.** All 5 tests in `tests/unit/test_knowledge_ledger_drain_job.py` pass, but `test_entrypoint_executes_independent_drain` burns 3.41s CPU against the fast-unit duration guard's 1.00s limit (`4 passed` then `BACKEND-UNIT-FAILED-FILE … status 1`). The cost is the `importlib.util` load of the job module inside the timed phase — `_load_job()` re-executes `modal/knowledge_ledger_drain_job.py` (which imports firebase_admin and the whole drain module) per test. Please hoist the module load out of the timed call phase (module-level fixture/lazy cache), or trim what the test imports. The repo's timing guard is the gate here, not a production defect.

What I verified and liked, file by file:

- `backend/utils/memory/knowledge_ledger_drain.py` — the heart of the change and it's good: `bounded_ledger_drain_inventory` scans a 20-doc `memory_state/apply_control` collection-group page (`MAX_LEDGER_DRAIN_UIDS_PER_RUN = 20`), skips already-ledger accounts, only picks up `transitioning_to_ledger` rows owned by `knowledge-ledger-migration.v1`, and `_write_cursor` is generation-fenced (transactional compare on `generation`) so overlapping executions can't clobber progress. `fresh_rollout_authorizer` preserves the per-account/per-row JIT re-authorization exactly as the old cron tail did, publication still requires `remaining_live_legacy_count == 0` and no revoked authority. Call sites match `run_ledger_migration_sweep` / `publish_ledger_migration_cutover` signatures in `knowledge_ledger_migration.py`.
- `backend/utils/memory/canonical_short_term_maintenance_cron.py` — removal is complete and safe: `run_canonical_short_term_maintenance_cron` now returns the maintenance summary directly, and the removed summary fields (`ledger_migration_users`/`ledger_migration_rows`) have no remaining consumers (`modal/memory_maintenance_job.py` reads only `summary.errors`).
- `backend/modal/knowledge_ledger_drain_job.py` + `backend/modal/Dockerfile.knowledge_ledger_drain_job` — entrypoint mirrors the sibling `daily_memory_sweep_job.py` conventions (same `_init_firebase`, lazy `database._client.db` import, non-zero exit when a page reports errors). `COPY backend/ .` then `COPY backend/modal/ .` matches the existing image pattern.
- `.github/workflows/gcp_daily_memory_sweep_job.yml` and `..._auto_dev.yml` — symmetric additions in both lanes: build → smoke via `runtime_image_contracts.py` → push → render/validate runtime env → deploy the second Cloud Run job → provision + validate the second scheduler. Deploy stays `workflow_dispatch` + admitted-release-SHA gated, which keeps this controlled.
- `backend/deploy/runtime_env.yaml`, `runtime_env/_base.yaml`, `dev.overlay.yaml`, `prod.overlay.yaml` — `_base` ships `MEMORY_ENABLED: off` default; overlays enable per stage; task timeout 1200s with `--max-retries 1` is sensible for a bounded page.
- `backend/scripts/provision_daily_memory_sweep_scheduler.py` — the `EXPECTED_TARGETS` dict generalization is the minimal correct change; identity mismatch still raises.
- `backend/tests/unit/test_knowledge_ledger_drain.py` — good contract coverage (cursor independence from the maintenance cursor, no advance before explicit commit, revoked authority never publishes, fail-closed on malformed writer_mode). A follow-up worth adding: a two-page resume test proving the second `bounded_ledger_drain_inventory` call starts after the committed `last_path` (your `_Query` fake already supports `start_after`).
- `backend/tests/unit/test_legacy_memory_retirement_readiness.py` — the new scheduler-args assertion keeps the retained-contract check honest.
- `backend/docs/canonical_memory_architecture.md`, `backend/docs/runbooks/universal-memory-operations.md`, `backend/docs/subservice-internals.md` — I checked the doc claims against the code (20-doc cap, cursor-after-page-return, per-boundary re-authorization); they're accurate, and the runbook renumbering is correct.
- `.github/failure-classes/FC-tail-work-starved-by-timeout-predecessor.json` — fits the existing FC corpus schema; the violated-contract statement matches the incident.
- `backend/runtime_images.json` — manifest entry wires smoke/probe gates for the new image; `smoke_environment` uses the same dummy `ENCRYPTION_SECRET` pattern as siblings.

Non-blocking design notes (maintainer judgment wanted, not necessarily changes):

1. `commit_ledger_drain_inventory` advances the cursor even when `summary.errors` is non-empty, so accounts that errored this run are only retried after wraparound. Defensible (avoids one broken account poisoning the head of the queue forever) but worth an explicit maintainer nod since the job exits non-zero while still moving the cursor.
2. Inventory fails the whole run closed on any apply-control doc with a missing/unrecognized `writer_mode`. Safe default — just confirm no legacy docs predate the field, or those accounts will block the page until wraparound.
3. The new scheduler reuses `memory-maintenance-scheduler@…` as its target SA. If its `roles/run.invoker` grant is scoped to `daily-memory-sweep-job` only, `knowledge-ledger-drain-hourly` will 403 on first fire — the grant isn't provisioned anywhere in this diff, so it needs either a project-level grant or an IAM step in the deploy lane before the first prod run.
4. Unlike `daily-memory-sweep-job` (whose prod overlay ships explicit default-off flags), the prod gate for this mutating hourly job is solely the shared JIT authority. That may be exactly right; it's a product call, not a code defect.

One process note: the `Hygiene` failure (`firestore-query-coverage`: baseline 62/102 vs inventory 73/113) is a known pre-existing baseline drift unrelated to this PR — no action needed here.

Leaving the workflow/deploy surface and the prod-gate/IAM questions for human maintainer review before merge; the code-level ask is just the test-duration fix.

---

aryanorastar:
Addressed every actionable review item in `b73cfa2`:

- granted `memory-maintenance-scheduler` `roles/run.invoker` on `knowledge-ledger-drain-job` in both admitted deploy lanes before the trigger is created/resumed;
- kept failed pages retryable by committing the generation-fenced cursor only when the page has no account errors;
- treated legacy apply-control rows with an absent `writer_mode` as `compatibility`, while invalid explicit values still fail closed;
- restored call-time `get_firestore_client()` resolution and moved that blocking construction onto `db_executor`;
- added behavioral coverage for two-page resume, end-of-inventory wrap, stale-generation rejection, legacy rows, error retry, and the call-time client seam;
- changed the entrypoint tests to load the heavy Firebase/job module once, reducing the changed test file from the reported 3.41s failure to 0.08s locally;
- clarified the multi-target Scheduler provisioner messages and operational docs.

Verification after the fixes:

- focused ledger/maintenance suite: 40 passed;
- runtime/deployment contract suite: 210 passed;
- backend pre-deploy gate: 226 passed;
- full PR preflight: 34/34 checks passed, including Firestore query coverage and workflow contracts;
- runtime image registry/source/workflow checks: 14/14 images passed;
- bounded pre-push gate: passed, including changed backend unit selection and formatting.

The remaining physical proof is intentionally post-merge/dev-deploy: confirm the hourly trigger can invoke the job, content-free cursor counters advance, and compatibility accounts reach `writer_mode=ledger`. No mobile or device test is involved.

aryanorastar:
Both reds are fixed on `ea6d3e23`, and one of them was not what it looked like.

**Duration guard.** The diagnosis in the review points at the call phase, but the cost is not there. Measured directly:

```
module exec CPU: 1.517s
main()      CPU: 0.000s
```

`main()` is free once `_init_firebase` and `run_knowledge_ledger_drain` are patched — the 1.5s is executing the job module, which pulls firebase_admin and the whole drain graph. `pytest --durations=0` confirms it lands in **setup**, not call, and the guard measures the call phase only. My read is that gRPC's background threads keep burning CPU into the call window, and `time.process_time()` is process-wide, so it gets charged there. Loading the module at test-module import time takes it out of every phase the guard measures; the file now runs clean under `test.sh`.

**`firestore-query-coverage` is not pre-existing drift.** I checked it against clean `main` before touching anything — main exits 0. The baseline on main already reads `113/73`, which is exactly what this branch's inventory reported; the branch was 69 commits behind and still carrying `102/62`. Merging main cleared it with no regeneration.

Also addressed, with tests:

- **Note 1** — the cursor no longer advances when the page reported errors, so a failed account is retried next run instead of waiting for wraparound.
- **Note 2** — a row with no `writer_mode` reads as `compatibility` rather than failing the page closed, so accounts predating the field cannot block the head of the queue.
- **Note 3** — both deploy lanes grant the scheduler SA `roles/run.invoker` on the drain job before creating or resuming the trigger.

One thing worth your judgment rather than mine: notes 1 and 2 pull in opposite directions. Not advancing on error is the safer accounting, but it does mean a permanently failing account holds the page until it is fixed. Note 2's default is what keeps that from being reachable through the old-row path; if you would rather have wraparound as the backstop instead, say so and I will flip it back.

Verification: 15 passed across the three affected files under `test.sh`, duration guard clean, `--check-ratchet` exits 0, async-blocker gate green (client construction now dispatches through `run_blocking` rather than calling `get_firestore_client()` on the loop).

Git-on-my-level:
Re-reviewed on `b73cfa2` — the blocking item and all three design notes from the prior review are resolved, with contract tests. Credit for measuring the duration-guard failure directly (module exec vs call phase) and for checking `firestore-query-coverage` against clean `main` before touching anything.

Verified on the new head, item by item:

- **Duration guard (was blocking)** — `tests/unit/test_knowledge_ledger_drain_job.py` now loads the job module once at import (`JOB = _load_job()`), keeping the ~1.5s firebase_admin/gRPC module-exec cost out of every phase the guard measures. Checks are green on this head, and the workflow-contract and manifest assertions added alongside it (deploy → IAM grant → scheduler ordering; `1200s`/`--max-retries 1`; expected secrets set) tighten the same surface.
- **Note 1 (cursor on error)** — `run_knowledge_ledger_drain` in `utils/memory/knowledge_ledger_drain.py` commits the inventory cursor only `if not summary.errors`, and `test_drain_retries_the_same_page_when_an_account_fails` pins it (`committed == []` after a migration error). On notes 1 and 2 pulling in opposite directions: retry-same-page is the safer accounting, and since the job exits non-zero on errors, a stuck page alerts instead of silently skipping — the chosen ordering is right.
- **Note 2 (legacy rows)** — `payload.get("writer_mode", WriterMode.compatibility.value)` defaults absent fields to compatibility, while `test_inventory_fails_closed_on_malformed_writer_mode` keeps invalid explicit values failing closed. Good split.
- **Note 3 (IAM)** — both `gcp_daily_memory_sweep_job.yml` and `..._auto_dev.yml` now run `gcloud run jobs add-iam-policy-binding "$LEDGER_DRAIN_SERVICE" … --role="roles/run.invoker"` after the drain deploy and before the hourly trigger is created/resumed, and `test_admitted_deploy_lane_provisions_an_independent_job_and_scheduler` asserts that ordering for both lanes.
- **New in this commit** — the call-time `get_firestore_client()` dispatched through `run_blocking` (with `test_drain_resolves_the_firestore_client_at_call_time`), plus the two-page resume/wrap and stale-generation inventory tests, close the follow-up gap from the prior review.

Everything else stands as reviewed before: the maintenance tail removal in `canonical_short_term_maintenance_cron.py` is complete (no remaining consumers of the removed summary fields), the provisioner's `EXPECTED_TARGETS` generalization stays fail-closed, and the failure-class entry, `runtime_images.json` manifest entry, and Dockerfile match their sibling jobs.

Docs note: `canonical_memory_architecture.md` and `runbooks/universal-memory-operations.md` now state the drain-independence contract (20-doc page cap, cursor-advances-only-after-clean-page, per-boundary JIT re-authorization, run.invoker-before-trigger). These are agent-facing contracts for anyone — human or AI — modifying memory scheduling here; I checked each claim against the code on this head and they are accurate.

What remains is judgment, not code: this admits a second mutating hourly Cloud Run Job into both deploy lanes, and in prod the drain's only gate is the shared JIT admission authority (`MEMORY_ENABLED: on` in the prod overlay, no per-job default-off flags the way `daily-memory-sweep-job` ships them). That may be exactly right for finishing the #12535 cutover — it needs a maintainer's rollout call before merge, along with sign-off on the workflow changes.

---

Git-on-my-level:
Independent verification pass on `b73cfa2` — no new issues found. Notes per area:

**Core extraction is faithful.** `utils/memory/knowledge_ledger_drain.py` preserves every safety invariant from the old maintenance tail: per-account JIT re-authorization (`fresh_rollout_authorizer` fails closed on timeout/error), row and publication boundaries wired through `mutation_authorizer`/`publication_authorizer`, cutover published only when `remaining_live_legacy_count == 0`, and the inventory cursor committed only `if not summary.errors`, so a failed page is retried rather than skipped. The cursor is generation-fenced (`_write_cursor` compares the live generation inside a Firestore transaction), `LEDGER_TRANSITION_OWNER` matches the owner constant in `knowledge_ledger_migration.py`, and legacy rows without `writer_mode` default to compatibility while explicit invalid values still fail closed.

**Tail removal is clean.** `canonical_short_term_maintenance_cron.py` drops the 94-line ledger tail and returns the maintenance result directly; every deleted test concern (rollout gating, kill-flip-before-mutation, mid-batch revocation) has a counterpart in the new `test_knowledge_ledger_drain.py` (290 lines, including two-page resume, end-of-inventory wrap, stale-generation rejection, retry-same-page, call-time client resolution).

**Deploy lane mirrors the existing one.** Both `gcp_daily_memory_sweep_job.yml` and the `_auto_dev` variant keep the admitted-main `release_sha` gate, build/smoke/push the new image from that SHA, and order deploy → `roles/run.invoker` grant for the existing scheduler SA → provision/validate — with that ordering pinned by `test_admitted_deploy_lane_provisions_an_independent_job_and_scheduler`. `runtime_env/_base.yaml` stages the job default-off (`MEMORY_ENABLED: off`) with per-overlay enablement, and `test_runtime_manifest_keeps_the_drain_bounded_and_authorized` asserts `1200s`/`--max-retries 1` and the exact secret set.

**Docs note (these steer AI agents in this repo):** `canonical_memory_architecture.md` and `runbooks/universal-memory-operations.md` now describe the drain job's caps and authorization model. Checked each claim against the code — the 20-document page cap, fail-closed re-authorization, no-cursor-advance-on-error, and IAM-before-trigger ordering are all accurate and contract-tested, so the guidance is safe for coding/review agents to rely on.

One non-blocking observation: `_write_cursor`'s non-transactional fallback (only reachable for clients without `.transaction()`, i.e. test doubles) does a read-check-then-set that isn't atomic; the generation check would still catch a stale write afterward. Not worth churn on its own.

The remaining proof is the intentional post-merge dev-deploy confirmation of the hourly trigger. `needs-maintainer-review`/`workflow-review` stay appropriate for the deploy-lane and IAM changes; leaving the ship call with the maintainer.

This review was produced by David's automated maintainer assistant.

---

undivisible (approved):
## Verdict: APPROVE

Independent ledger drain is the right fix for #12535, and the safety model survives the extraction. Prior cubic / Git-on-my-level items look addressed on `b73cfa2`. CI is green on this head. I have not submitted merge from this review (lane is review-only / hold for Max’s stack).

### What I checked
- Diagnosis matches the code: drain no longer rides `memory-maintenance-job`’s 3600s consolidation tail.
- `knowledge_ledger_drain.py`: ≤20-doc apply-control inventory, dedicated generation-fenced cursor, commit only when `summary.errors` is empty, missing `writer_mode` → compatibility / invalid explicit → fail closed, JIT re-auth per account and at mutation/publication boundaries, cutover only when `remaining_live_legacy_count == 0`.
- Maintenance cron removal is complete; ledger summary fields have no remaining consumers in the job path.
- Deploy lane (manual + auto_dev): build/smoke/push → deploy drain job → `roles/run.invoker` for scheduler SA → provision/validate hourly trigger; ordering pinned by unit contract.
- Runtime: base `MEMORY_ENABLED: off`, overlays `on`; task timeout `1200s`, `--max-retries 1`.
- Tests cover resume/wrap, stale generation, retry-same-page-on-error, revoked auth, call-time Firestore client, workflow IAM ordering.

### Merge gates (process, not new code defects)
1. **Rebase (or merge) onto current `main` and re-green CI** — branch is ~594 commits behind. Touched `runtime_env*` files auto-merge cleanly today, but composed manifest / unit gates should be re-proven on a fresh tip.
2. **Workflow / deploy sign-off** — this admits a second mutating hourly Cloud Run Job; prod overlay enables `MEMORY_ENABLED`. Labels `needs-maintainer-review` / `workflow-review` still apply for the ship call.
3. **Deploy ordering** — drain *removal* lives in `gcp_memory_maintenance_job*`; drain *addition* lives in `gcp_daily_memory_sweep_job*`. Prefer standing up the new job/trigger in the same window as any maintenance image rebuild from this change. (Today’s production behavior is already “drain never runs,” so a brief dual-miss window is not a regression vs #12535, but it should be deliberate.)
4. **Post-merge proof (intentional)** — hourly trigger invokes the job; content-free cursor advances; compatibility accounts progress toward `writer_mode=ledger`.

### Non-blocking follow-ups
- Accounts with remaining live legacy rows (100-mutation cap) do not error, so the cursor advances and they wait for inventory wrap (~20 docs/hour). Consider a sticky “remaining” retry path later if cutover latency matters at fleet scale.
- A permanently failing account pins its page (by design); that is alertable via non-zero job exits — keep an eye on it in the first days after deploy.

### Not blocking
- `transitioning_to_compatibility` rows can be inventoried; publish/sweep owner CAS should fail closed (same class of behavior as the old unfiltered cron tail).

PR #12551, opened 2026-09-01
fix: keep bounded tool results successful

Git-on-my-level:
## What changed and why

Fixes SCA-401. Fixes SCA-404. Fixes SCA-405. Fixes SCA-407. Fixes SCA-415 by removing the boundary where a successful large Swift tool result could become `ok:false` or arrive with no useful content. The [root cause](#root-cause) was both a relay byte cap acting as execution authority and recap records being collapsed into one indivisible Markdown chunk per section.

The six scoped read tools now return one typed item per conversation, memory, action item, app, task, focus session, or observation. Backend read items retain bounded per-record rendered content (including conversation overviews/transcripts), and source-less responses fall back to a non-empty text section. The manifest-driven projector reserves one UTF-8-safe excerpt for every non-empty section, then fills in manifest priority order, with exact item-based shown/omitted counts. Long purpose text is clamped, projection terminal cases never throw, and relay catch paths always clear and answer pending calls.

Typed citation markers/source IDs now round-trip through the desktop-chat citation parser, unresolved markers are stripped, and Swift clamps an invariant-breaking >48 KiB transport payload as successful truncation instead of asserting or forwarding it unbounded. Complete projections no longer synthesize `truncated:true` or degraded telemetry. The unused digest contract field was removed, manifest ranking now controls whether the flag can enable purpose ranking, and remaining provider byte reserves derive from the manifest budget. The separate 128 KiB direct-control limit remains only for authenticated local IPC/UI structure and is not a provider/model budget.

Round 4 makes the recap GRDB read cross strict-concurrency boundaries only as a Sendable, Codable value graph and encodes JSON after the read. Relay rewrapping now computes a consistent truncation predicate before constructing an envelope, unknown typed-result siblings participate in projection/omission accounting, and Swift-sorted record fields render title/summary/citation/source identity before separately budgeted content. Segment extraction rejects partial range conversion. Cancellation envelope accounting uses the same consistent-pair rule, and import ordering is lint-safe.

Direct-control overflow keeps structural compaction under its separate 128 KiB authenticated local IPC budget, persists the complete result, and returns a successful typed projection with array omission counts and a recovery handle. A genuine persistence failure is now explicit as `tool_output_persist_failed` and carries the original byte count. Realtime/realtime-voice conversation reads restore executor defaults of limit 5 with transcripts excluded unless explicitly requested, while main chat keeps limit 20 with transcripts enabled.

Round 5 preserves every remaining typed-item field in the identity segment (including recap minutes/captures and task priority) while keeping content as the only separately budgeted excerpt. Completeness is derived from the rendered item, so omitted fields or cut content cannot be reported as complete. The dead recap Markdown renderer is removed and citation registration now happens only while typed items are built. Relay finalization normalizes stale untruncated source byte measurements instead of constructing an inconsistent envelope.

Round 6 makes reserved typed-item keys conditional on actual rendering: non-string `title`, `summary`, `citationMarker`, `sourceId`, or `content` values now enter the generic identity projection instead of disappearing. The two degenerate control-tool compaction terminals also return the artifact-backed minimal successful shape, so a representation budget never manufactures `ok:false`, even under a deliberately tiny test budget.

Rebased onto current `main` (`6967e0bb0a`). The only merge conflict was `ThreeDoorsDemoPageTests.swift`; the redundant classification commit was dropped because `main` already has equivalent source-inspection comments. This PR still includes the `OnboardingFlowTests` fix from [#12586](https://github.com/BasedHardware/omi/pull/12586) that names “Open the doors” and expects nine default-action sites (`main` still expects eight). #12586 stays open until this remaining test fix lands here.

### Root cause

`ChatToolExecutor.typedReadToolResult` initially split rendered recap Markdown into one item per section, then the per-record migration dropped `resultText` whenever backend sources were missing and retained only short source previews when they existed. That erased conversation transcripts and the durable citation-guide handoff. Separately, finalizers synthesized truncation to satisfy envelope math, while Swift asserted on invariant-breaking oversize payloads instead of clamping them.

### Design principles carried by this change

1. Executors return typed per-record sections and authoritative totals; one kernel projector creates model-visible text.
2. A budget reduces detail, never execution success; every non-empty section retains content and reports exact omissions.
3. Truncation is successful, artifact-backed, and drillable through `read_tool_output` / `search_tool_output`.
4. Purpose ranking remains flag-gated and default-off; recency order is preserved otherwise.
5. Per-tool/per-surface model budgets live in `omi-tool-manifest.ts`; Swift transports and applies only a defensive invariant clamp.
6. Projection, persistence, formatting, and transport cannot manufacture `ok:false` from a succeeded result.
7. Original/projected bytes and degraded fallbacks are recorded on relay, control, and Swift boundaries.

- INV-AGENT-*
- INV-CHAT-1
- INV-VOICE-1

## How it was verified

- `cd desktop/macos/agent && npm run build && npm test`
- `node desktop/macos/agent/scripts/generate-tool-surfaces.mjs --check`
- `xcrun swift build -c debug --package-path desktop/macos/Desktop`
- `xcrun swift build -c release --package-path desktop/macos/Desktop --triple arm64-apple-macosx`
- `desktop/macos/scripts/dev-feedback.py --once swift 'AgentRuntimeContractFixtureTests|ChatToolExecutorRowIntTests|ChatCitationTests|RealtimeProviderToolResultPolicyTests'`
- `cd desktop/macos && ./scripts/agent-logic-harness.sh --skip-install`
- `scripts/pr-preflight --pr-body-file pr-body.md`

A live authenticated PTT turn on a named bundle was not run; that remains the post-merge/beta verification gap. No installed `Omi` or `Omi Beta` application was touched.

## Tests

| Invariant | Guard |
| --- | --- |
| Succeeded executor output never becomes `ok:false`, including 4 MiB results and a 40 KiB purpose | `desktop/macos/agent/tests/relay-tool-result.test.ts`; `desktop/macos/agent/tests/control-tools.test.ts` |
| Realistic 52-conversation/36-memory recap and transcript-heavy search fixtures fit 7,424 bytes, keep five titles, and preserve overview/transcript content | `desktop/macos/agent/tests/tool-result-projector.test.ts`; `desktop/macos/Desktop/Tests/ChatToolExecutorRowIntTests.swift` |
| `shown + omitted == total` and every populated section has at least one shown record | `desktop/macos/agent/tests/tool-result-projector.test.ts` |
| Truncating realtime surfaces expose both drill-in tools | `desktop/macos/agent/scripts/generate-tool-surfaces.mjs` |
| Nil/empty backend sources preserve result text; desktop-chat typed citations round-trip; unresolved markers are removed | `desktop/macos/Desktop/Tests/ChatToolExecutorRowIntTests.swift`; `desktop/macos/Desktop/Tests/ChatCitationTests.swift` |
| Swift logs and clamps an unprojected oversize invariant as successful truncation without crashing | `desktop/macos/Desktop/Tests/AgentControlServiceTests.swift`; `desktop/macos/Desktop/Tests/AgentRuntimeContractFixtureTests.swift` |
| Relay and control truncation emit degraded fallback diagnostics | `desktop/macos/agent/tests/relay-tool-result.test.ts`; `desktop/macos/agent/tests/control-tools.test.ts` |
| Complete projections remain `truncated:false` and emit no degraded record | `desktop/macos/agent/tests/relay-tool-result.test.ts` |
| Rewrapped truncated envelopes never throw; non-section siblings become recoverable truncation | `desktop/macos/agent/tests/relay-tool-result.test.ts`; `desktop/macos/agent/tests/tool-result-projector.test.ts` |
| Swift-sorted transcript items retain title/summary before content; malformed segment ranges return nil | `desktop/macos/agent/tests/tool-result-projector.test.ts`; `desktop/macos/Desktop/Tests/ChatToolExecutorRowIntTests.swift` |
| Realtime conversation fetch defaults remain 5/no-transcript unless requested | `desktop/macos/Desktop/Tests/ChatToolExecutorRowIntTests.swift` |
| Non-string reserved item keys remain visible and force truthful incompleteness; degenerate tiny budgets remain successful and artifact-backed | `desktop/macos/agent/tests/tool-result-projector.test.ts`; `desktop/macos/agent/tests/control-tools-tiny-budget.test.ts` |

## Adversarial review disposition

- Round-3 R1–R6 are addressed in code and behavioral tests: source-less text, per-record content, citation recovery, Swift clamp/no assert, unresolved-marker removal, and accurate truncation accounting.
- S1/S2 are addressed: ranking is manifest-driven, the vestigial digest contract field is removed, and provider reserves derive from the manifest budget.
- Round-4 C1–C9 are addressed, including the strict release compile, total relay finalization, field-wise projection, sibling accounting, segment-count guard, direct-control typed overflow policy, realtime fetch defaults, cancellation envelope math, and import ordering.
- Round-5 M1–M5 are addressed. M4 remains deliberately precompact: the `spawn_agent` source envelope describes the canonical full result and downstream projection target, and the mandatory relay boundary validates, remeasures, and reprojects it before provider delivery. A code comment pins that ownership decision; direct provider consumers never observe the precompact byte pair.
- Round-6 N1/N2 are addressed with production-path regressions for non-string reserved keys and a 32-byte degenerate provider budget. D1 remains split as #12586: the ThreeDoors classification commit was dropped on rebase because `main` already has equivalent comments; this PR still carries the OnboardingFlowTests “Open the doors” / nine default-action fix until that lands.
- Nit 13: synthetic-only assertions were replaced with realistic record-shaped fixtures and exact content/count assertions.
- Nit 14 required no change; the generator check remains fail-closed and passes.

## Failure class (fixes)

Failure-Class: new

Adds `FC-bounding-layer-converts-success-to-failure`. A representation budget may reduce detail but cannot alter executor outcome. The reusable guard is the manifest-owned total projector with cross-surface realistic fixtures, success fuzzing, and transport diagnostic coverage.

These guards would have caught SCA-401 and the SCA-404 review probe before the first PR: they execute production projection with the real per-record shape rather than checking only synthetic small items or source strings.

## Scoped cleanups

Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/RealtimeHubController+SessionDelegate.swift | 1529 -> 1536 | this is the existing transport boundary where original-byte latency, degraded fallback, and the non-crashing invariant log must be recorded

Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatToolExecutor.swift | 3526 -> 3804 | the six existing read routes must preserve source-less text, value-typed recap transport, bounded per-record rendered content, citation ordinals, and surface-aware fetch defaults at one typed executor boundary

🤖 Generated with Codex


Git-on-my-level:
Monitor pass on the four open review threads (push: `d13fe88150..12ae5c9dac`, all four threads now resolved):

- **`tool-result-projector.ts` — recap shown/omitted counts (PRRT_kwDOLkKqys6eQuOP):** already addressed in the PR head before this run — `get_daily_recap` emits one typed `DailyRecapItem` per record and the projector counts shown/omitted per item. Verified against the current diff; no code change needed this round. Thread resolved as addressed.
- **`index.ts:417` — raw `console.error` in `onDegraded` (PRRT_kwDOLkKqys6eQuOl):** fixed in `12ae5c9dac`. The degraded record now goes through the pipe-safe `logErr` stderr write, and the same treatment was applied to its sibling `recordControlProjectionFallback` in `control-tools.ts` (the only other raw `console.error` this PR introduced). Bounded projections of large-but-successful results no longer emit error-level noise on the intended happy path.
- **`RealtimeHubController+SessionDelegate.swift:96` — "INVARIANT VIOLATION unprojected" mislabeling (PRRT_kwDOLkKqys6eQuOo):** fixed in `12ae5c9dac`. `RealtimeProviderToolResult` now carries `hasCanonicalEnvelope` (measured on the kernel payload before `finalize` wraps it), and the invariant-violation log fires only when no canonical envelope was present — a projected-but-over-cap result is no longer mislabeled as an unprojected breach.
- **`omi-tool-manifest.ts` — contract fields the projector never consumed (PRRT_kwDOLkKqys6eQuO6):** already addressed in the PR head before this run — the projector now orders sections by `contract.sections`, gates purpose ranking on `contract.ranking`, enforces `maxItemsPerSection`, and the unused `digest` contract field was removed. Verified against the current diff; no code change needed this round. Thread resolved as addressed.

Verification before push: `RealtimeProviderToolResultPolicyTests` 7/7 (including a new `testUnprojectedOversizedResultIsFlaggedForTheInvariantLog` covering the envelope-absent case and `hasCanonicalEnvelope` assertions on both over-budget tests); agent vitest `relay-tool-result` + `tool-result-projector` + `runtime-adapter-contract-conformance` 32/32 and `control-tools` 90/90; `tsc --noEmit` clean; repo pre-push preflight (line-count ratchet, swiftlint, swift-format, changelog, contract-conformance) passed with the PR body's existing Line-Count-Exceptions.

PR #12590, opened 2026-09-02
feat(desktop): remove typed input from the notch

kodjima33:
## What

The notch never takes typed text again.

- Deletes `AskAIInputView` and its mount on the `.mainInput` surface (the surface still exists for the agent-pills header; its body is empty).
- `openAIInput()` ("Ask Omi") lands in the main chat instead of opening a composer in the notch.
- Backing out of an agent chat with nothing else to show routes to the main chat on every shell, not only chat-first (`FloatingPrimaryTextInputRouting` is no longer a per-shell flag).
- The response view's follow-up was already a "Continue in Omi" button, so no typed path remains.

## Verification

- Built the `omi-first-use` named bundle (fast lane) with this change and confirmed the notch shows no composer; the previously persisted floating draft ("Which of these should I buy, and why?") has nowhere to render.
- Focused suites run locally: `FloatingControlBarStateTests`, `AgentPillLifecycleTests` (new `testAgentExitWithNothingToShowRoutesToTheMainAppOnEveryShell`, `testTheNotchHasNoTypedComposer`), `FloatingGlassChromeTests`.
- `scripts/pr-preflight --suggest`: no product invariants matched; no `fix:` commits.

- INV-CHAT-1 (FloatingControlBar paths; no new transcript store, no notch/main mirroring — the notch loses a composer, the kernel-owned transcript is untouched)
- INV-NAV-1 (DesktopHomeView; no destination replaced or reduced)

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12590?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33:
Superseded: this change landed on main via #12591 (the popup branch carried the notch removal).

PR #12599, opened 2026-09-02
Warn at 90% and 100% of the chat quota above the composer

katipally:
# Warn at 90% and 100% of the chat quota

## What

The monthly chat cap only ever announced itself at the moment a send was refused, so running out was always a surprise. A warning strip now sits directly above the chat composer once usage passes **90%** and **100%** of the plan allowance, carrying the count, the usage meter and when the cycle resets. (An earlier iteration also warned at 75%; two honest moments beat three nagging ones, so the 75% tier is gone.)

It also fixes a client/server divergence found while building this: on an overage plan the desktop app was blocking sends the backend would have served and billed.

## How it works

```
 GET /v1/users/me/usage-quota
   { plan, unit, used, limit, allowed, reset_at, is_overage_plan }
              |
   FloatingBarUsageLimiter          used + optimisticDelta (local send count)
              |
   ChatQuotaBanner.current(...)     pure: percent -> highest crossed threshold
              |
   ChatQuotaBannerView.Slot         mounted above the composer in both shells
```

Derived state rather than a scheduled event: the banner is recomputed from the live quota snapshot, which already counts sends locally between server syncs, so it cannot be missed and it updates in place as usage climbs.

The banner is one quiet line in `ChatErrorCard`'s idiom: severity glyph, title, compact count, an inline usage meter, and the actions. Severity escalates in one hue — outline orange at 90%, filled orange at the allowance, and the error red only on a hard-capped plan, where sends are actually refused. The full detail (plan name, reset date, overage copy) stays in the banner's accessibility label and on the plan page.

Dismissing one hides only that threshold for that cycle. The key carries the reset **and** the allowance, so a new month or a plan upgrade speaks again, and a dismissal from a past cycle is dropped rather than accumulating for the life of the account. Dismissals are scoped to a hashed runtime owner and reloaded on owner change, so signing in as someone else never inherits them.

Copy follows the plan's exhaustion policy: an overage plan is told its extra usage is billed, a hard-capped one is told to upgrade.

## The overage fix

`ChatUsageQuota.allowed` goes false at the allowance on **every** plan, but `enforce_chat_quota` only raises for hard-capped ones — overage plans are served and billed. Both desktop pre-send gates refused on `allowed == false`, so at 500/500 Operator the app blocked the send and showed "You've hit your monthly limit" on a plan whose own settings card reads *"No hard cutoff."*

The gates restated the server's exhaustion policy by hand instead of reading it. They now consult the catalog's own predicate (`plan_uses_overage`), carried on the quota payload as `is_overage_plan`. Hard-capped plans still block, and a server that does not send the field keeps the old blocking behavior rather than failing open.

| plan family | 90% | 100% |
|---|---|---|
| overage (Neo, Operator, Architect) | warn | warn, send proceeds and bills |
| hard cap (Free, Plus, Unlimited v2) | warn | warn, server blocks (402) |

- **INV-CHAT-1** — no transcript, session or agent state is touched. The banner is derived UI over the existing quota snapshot.
- **INV-NAV-1** — both desktop shells mount the same `ChatQuotaBannerView.Slot`, so neither gets a reduced copy, and "View plan" routes to the established Plan & Usage settings destination the existing usage-limit modal already uses.
- **INV-MEM-1** — named for paths the branch inherits from `main` itself (`app/lib/pages/memories/**`, arrived in the merge of current main); this PR changes no memory-tier surface.

## Failure class (fixes)

The pre-send gates mirrored the server's exhaustion policy by restating it (`allowed == false` means block) rather than deriving it from the catalog, so they could not distinguish a hard cap from a billed overage.

## Verification

- `backend/test.sh tests/routers/test_users.py` — **33 passed**, including 6 new parametrized cases pinning the exhaustion policy for every catalog plan.
- Swift — **43 passed**: `ChatQuotaBannerTests` 19, `FloatingBarUsageLimiterTests` 21, `UsageLimiterActionTests` 3.
- `omi-harness run e2e/flows/chat-quota-banner.yaml --lane bridge` — **PASS, 9/9 steps, 0 log errors**, against the running QA bundle.
- Drove the named bundle `omi-quota-90` against a signed-in account and exercised all three banner states live (90%, 100% overage, 100% hard cap) through the automation bridge.

Regenerated the OpenAPI specs and the Dart/TS/Swift clients; the diff is one optional field per artifact.

## Line-count exceptions

Line-Count-Exception: backend/routers/users.py | 2247 -> 2250 | three lines: the new import plus the field on the two ChatUsageQuota constructions in this file's own endpoint
Line-Count-Exception: desktop/macos/Desktop/Sources/Chat/AgentBridge.swift | 2202 -> 2206 | four lines: the comment explaining why the quota refresh must precede the cached denial
Line-Count-Exception: desktop/macos/Desktop/Sources/DesktopAutomationBridge.swift | 4835 -> 4888 | one non-prod harness action plus the banner fields on the existing usage-limiter snapshot it belongs with
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/DesktopHomeView.swift | 1689 -> 1694 | one navigation handler next to the sibling settings-navigation handlers
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/DashboardPage.swift | 4186 -> 4193 | two one-line banner mounts; the banner itself lives in its own files


Git-on-my-level:
Thanks @katipally — this is carefully built, and it fixes a real client/server divergence on top of the feature.

**What I verified**

- Backend: `is_overage_plan` on `ChatUsageQuota` (`backend/models/users.py`) is populated from `plan_uses_overage` in `backend/routers/users.py` — the same catalog predicate `enforce_chat_quota` already uses when deciding serve-and-bill vs 402. The BYOK early-return path reports `False`, which is right (no cap concept there). `backend/tests/routers/test_users.py` parametrizes both policies across six plan types, and `docs/api-reference/app-client-openapi.json` plus the four generated clients match the model.
- Client gating: `FloatingBarUsageLimiter.isLimitReached` now returns false for overage plans, and `AgentBridge` both skips the cached denial for them and moved the quota refresh **before** the cached-verdict check — the old order meant a blocking cached snapshot could prevent its own refresh, locking out a user who upgraded an exhausted plan. The refresh-order fix is the right one.
- Backward safety: `isOveragePlan` decodes as `Bool?` (`APIClient+People.swift`), so a server predating the field reads as hard-cap; `testMissingOverageFieldKeepsHardCapReading` and `testMissingOverageFieldReadsAsHardCap` pin that on both surfaces.
- Banner model: `ChatQuotaBanner.current` is state-derived (no scheduling to miss), applies `optimisticDelta` only to question quotas, and scopes dismissals to threshold + cycle where cycle = `resetAt`-`limit`, so an upgrade mid-cycle re-speaks instead of staying dismissed against a stale allowance. The dismissal store hashes the owner id for the UserDefaults key and prunes old cycles.
- Harness: `apply_usage_quota` (`DesktopAutomationBridge.swift`) is gated on `AppBuild.isNonProduction`, resets seeded dismissals, and has a source-contract test asserting the gate; `QueryShellHome` now measures the footer (banner + composer) as one unit so the banner can't push the composer down.
- All seven earlier review comments (stale-cache lockout, harness dismissal reset, rendered-banner assertions, banner-above-composer ordering in `homePanelStage`, dismiss accessibility label, cost_usd threshold coverage, e2e description wording) are addressed on the current head — I checked each against the diff.

**Checks**

The two failing desktop checks don't come from this PR: the same jobs are red on `main` today (OnboardingFlowTests), and here the only failing suite is `ProactiveListenEventTests`, which crashes in `UNUserNotificationCenter` bundle-proxy setup (`NotificationService.swift:171`, untouched by this PR). This PR's own suites (`ChatQuotaBannerTests`, `FloatingBarUsageLimiterTests`, `UsageLimiterActionTests`) passed in the same run, as did the release compile and the backend gates.

**One item for a maintainer before merge**

On overage plans the desktop app now lets sends continue past the allowance (billed) where it previously hard-blocked. That matches what the backend already does, and the 100% banner states it plainly — but it's a billing-behavior change for paying desktop users, so please confirm send-through rather than a soft local block is the intended behavior for these plans, and that the billing copy is the messaging you want.

---

Git-on-my-level:
Follow-up on the new head (`584394b`, the upstream/main merge):

- The merge changed none of this PR's 23 files (checked via first-parent compare) — the reviewed content is identical; only surrounding upstream context moved.
- CI is now fully green on this head — every check run passes or skips, including the desktop suites that were red for unrelated reasons at the previous review — so I've removed `ci-failing-pre-existing`.
- Everything verified earlier still holds: `is_overage_plan` comes from the same catalog predicate `enforce_chat_quota` uses, the cached-denial path can no longer wedge after a plan upgrade, and the `apply_usage_quota` harness action is non-prod gated.

Keeping `needs-maintainer-review` for the one open item: on overage plans the desktop app now lets sends continue past the allowance (billed) where it previously hard-blocked. That matches the backend's serve-and-bill semantics, and the 100% banner states it plainly — but it is a billing-behavior change for paying desktop users, so please confirm send-through is the intended behavior and that the billing copy is the messaging you want.

---
*by AI on behalf of David — leaving for human maintainer review: the overage send-through billing behavior and its user-facing copy are a product/billing decision for paying users.*

PR #12606, opened 2026-09-02
Quota-gate the realtime relay on every plan and make hub turns idempotent (free tier S15)

Git-on-my-level:
## S15 — quota-gate the realtime relay on every plan; make hub turns idempotent

Shard S15 of the ratified local-models free-tier program (`omi-knowledge-base/projects/local-models-free-tier/implementation/60-shards.md`, decision 8). Stacked on #12592 (S0); merge that first.

### What the code actually does (correction to the plan)

The plan's status doc says the legacy relay "never debits" a PTT turn and calls it a live leak. In the shipped client the relay is the voice shell only: `RealtimeOmniService` streams mic audio in and speech out, and `PushToTalkManager.omniDidFinishTurn` sends the transcript through **desktop chat**, whose request already debits one question (`desktop_chat_completions:{request_id}`). A per-turn relay debit would charge the turn twice (Codex caught this on the first draft, and the Swift client confirms it). What the relay lacked was *enforcement*, not a debit.

### What changed

- **Relay connect gate on every plan.** Hard-capped plans (basic, plus, unlimited_v2) are refused at their cap with the existing `1008 quota_exceeded`; overage plans (operator, architect, unlimited) are served and billed, exactly like text chat. BYOK sessions with a key for the session's provider stay exempt. Previously only basic was gated.
- **Relay bounded per user per month, fail closed.** Every Omi-paid provider response is counted in a persisted monthly telemetry bucket (`realtime_relay.call_count`, atomic increment, not a question). The allowance (the plan's question limit plus two) is **admission**: the connect gate refuses a socket whose month is already spent, and at each turn boundary (never on an audio frame) the relay re-reads the quota and that count fresh and ends a hard-capped plan's session with the same `1008 quota_exceeded` once the plan has hit its cap **or** the allowance is reached. That holds across reconnects, a month reset and a plan change. Admission reads a persisted count rather than reserving, so sockets pre-opened just under the allowance could each forward one response past it; a strict per-user cap of two open relay sockets per serving process makes that overshoot finite, and no socket opens after the allowance is spent. An honest turn is one response and one question, so honest users reach the chat gate first. Boundaries come from the spend observer's own parsed evidence (a response with tokens; a Gemini `serverContent.turnComplete` after model activity or a usage delta with tokens), counted on the frame that ends the turn, never from a substring match. Both pumps are owned in a `try/finally` and drained with the repository's bounded `drain_tasks`, including when the handler itself is cancelled. A read or write that cannot be completed ends the session too (`1008 quota_unavailable`). The relay writes no question, ever (tests pin it); the admission counter's plan cost metadata is recorded as excluded, since S0's ledger holds the provider cost.
- **Hub turns idempotent and atomic.** `/v2/realtime/usage` accepts an optional `turn_id`; with it the question **and** the token cost are recorded in one transaction through `record_chat_quota_question` (the idempotent writer text chat uses, key `realtime_hub:{turn_id}`; it gains a `usage_update_for` callback that it invokes with its own plan key and day, fed by the new pure `usage_bucket_update`), so a retried report records neither a second question nor a second cost, a failed commit records neither, and the telemetry can never land on a different day or plan than the question. Without it the historical bare increment is unchanged, so older clients behave as before. Sending `turn_id` from `RealtimeHubSession.swift` is a desktop follow-up.
- **No managed mint for a BYOK user past the cap.** The hub only ever mints Omi's own OpenAI/Gemini token, but `enforce_desktop_chat_quota` exempted any user with an active Anthropic BYOK key. `enforce_chat_quota` gains `byok_exempt`, and the hub's mint and usage endpoints pass `byok_exempt=False`. Desktop chat (Anthropic-served) keeps its exemption. Overage plans still mint; a presence-gated rewarm mint debits nothing (it never did; now pinned).

### Kept on purpose, and one product decision

- **Trial paywall before quota.** Both realtime surfaces still apply the desktop trial paywall (`is_trial_paywalled`) ahead of the quota check, so an expired-trial basic user is refused as `trial_expired` even with questions left. Codex flagged it against "do not plan-gate realtime"; `50-rollout-and-verification.md` says trial expiry keeps "the existing trial-paywall UX unchanged", so this PR leaves it alone. If decision 8 is meant to override that, it is one condition in each gate.
- **The hub can outrun the cap by one token lifetime.** A minted hub session talks to the provider directly with a ≤30-minute token; the counter advances only through the client's usage reports, so a client that stops reporting after a 402 keeps its already-minted session until the token expires. Closing that needs server-side per-turn admission (proxy the hub, or one-turn credentials with reservation) — a product/architecture decision, recorded in `data/run-2026-09-01-free-tier/blocked/`. Not in this shard.

### The one user-visible change, flagged for David

A free user who reaches 30 questions **inside** a realtime session now gets the socket closed one turn later instead of at the next connect. The close code and reason are the ones the client already receives at connect. The current PTT client treats any close as an error and falls back to batch STT once before chat refuses the turn with 402, so a stale client pays one bounded extra transcription; surfacing the quota close as the usage-limit UI is a desktop follow-up (S13 territory).

### Automatic-or-dead check

`backend/tests/unit/test_realtime_quota_counter.py` (42 tests):

- **One counter:** a legacy hub report, a turn-id hub report and a chat question written through the **real writers** into a Firestore double that materialises `Increment` and `set(merge=True)` and **buffers transaction writes until commit** are read back by the **real** `get_monthly_chat_usage` as exactly 3; a report whose commit fails records neither question nor cost and its retry is not mistaken for a duplicate; the telemetry is built on the transaction's own plan key and day; a retried turn-id report counts once (positive control: the legacy increment repeats); a retried chat question counts once.
- **Relay, end to end through the real handler, with the real count writer and reader against the Firestore double:** never writes a question; refuses a third open socket for a hard-capped Omi-paid user, releases the slot when a session ends, and never caps BYOK or overage sessions; counts a response once per identity at `response.created` so a mid-response disconnect or a replayed frame cannot dodge or double it; enrolls a session downgraded mid-session in the cap and stops it when no slot is free; releases the slot when a missing provider key or a failed accept ends the session early; refuses with a controlled close when the connect-time snapshot cannot be read; recognises a replayed terminal frame 1,001 responses later, and a replayed announcement or an unannounced terminal among 200 concurrently open responses; ends an overage session at the identity capacity so a month-spanning socket cannot outgrow it, with the refused response still counted and still identified; refuses the response that reaches the allowance before its first frame; counts a Gemini response on its first activity and not on a decoy string; closes basic, plus and unlimited_v2 at their caps; serves operator past its included quota with no monthly admission bound (still counted; the per-session identity limit is the only bound it has); at 30 responses already served this month runs out at the 32nd, and the reconnect is refused before any provider work; an unreadable admission count refuses the connect; a zero-token usage block is not evidence of a response; a fresh snapshot at every boundary (an upgrade to operator lifts the bound); a snapshot with no question allowance bounds the session to the grace; stops on a failed quota read and on a failed count write; the admission counter never marks the plan's cost missing; token-less responses are not boundaries; BYOK-served sessions are never checked.
- **Hub:** a mint calls only the gate with `byok_exempt=False` and never a counter writer; a hard-capped plan at zero remaining is refused whether or not Anthropic BYOK is active (positive control: desktop chat still exempts that user); an overage plan still mints.

Red-proofs applied by hand and seen red (then restored), thirty-two in all: pumps not drained on handler cancellation; refused response not counted; session limit equal to the identity capacity; session response limit never enforced; open identities evicted at 64; slot never released; connect snapshot failure escaping; completed ids evicted at 64; replayed `response.created` counted again; downgraded session never enrolled; admission after forwarding; `response.created` not counted; Gemini first activity not counted; socket cap applied to everyone; socket cap not enforced; zero-token usage latching a response; telemetry built outside the transaction's plan and day; spent allowance not refused at connect; Gemini boundary evidence not counted; OpenAI boundary counted without tokens; admission counter marking cost missing; re-check disabled; gate reverted to basic-only; read failure swallowed (fail open); `turn_id` idempotency removed; BYOK exemption restored on the mint; monthly count never persisted; Gemini boundary never detected; hub telemetry outside the question transaction; allowance bound ignored; plus the two from the first design.

Reviewed by Codex (gpt-5.6-sol) in fourteen rounds: round one on the debiting draft returned REQUEST CHANGES with the double-charge finding that produced this design; round two surfaced the two kept-on-purpose items above; round three showed a per-socket budget resets on reconnect, which produced the persisted monthly bound; round four showed the bound was observation, not admission, which produced the connect-time refusal and the observer-owned boundaries; round five showed pre-opened sockets were unbounded, which produced the per-user socket cap; round six showed a mid-response disconnect dodged admission, which moved counting to the response start; round seven (the first with no blocker) tightened response identity, mid-session downgrades and admission-before-forwarding; round eight found a slot leak on early exits, fixed with a `try/finally` around the whole session; round nine found open-response identity lost past 64 concurrent responses, below the Plus and Unlimited-v2 allowances; round ten found a month-spanning socket could outgrow the identity capacity, closed with a per-session response limit; round eleven had the limit count the response it refuses and keep its identity; round twelve moved the pump drain into a `finally` so a cancelled handler cannot leave a pump running; round thirteen found only a timing-dependent test, replaced with event-signalled pumps; the final verdict is recorded in the PR thread.

### Proof

- Hermetic: the file above plus `test_free_quota_gate_wiring.py`, `test_desktop_realtime.py`, `test_chat_quota.py`, `test_llm_usage_db.py`, and the S0 files green.
- Live (after deploy): from a basic test account at 29 questions on the serving revision, one PTT turn → `/v1/users/me/usage-quota` shows `used` 30 (debited by the chat request, exactly once) and the next PTT press closes the relay with `1008 quota_exceeded`; the S0 ledger holds the relay's rows for both turns. Not run yet: no revision serves this code.

### Cut list

- Plan-gating realtime (explicitly rejected by decision 8).
- The desktop client: sending `turn_id` on hub usage reports, and rendering the mid-session quota close as the usage-limit UI.
- TTS / voice changes; the legacy Deepgram+TTS cascade named in 30 §3.2 is untouched.
- `llm_usage` cost telemetry for relay turns (the S0 ledger carries their cost).

### Line-count ratchet

Line-Count-Exception: backend/routers/desktop_proxy.py | 1528 -> 1632 | inherited from #12592 (S0), which this PR is stacked on; not touched by S15 and excluded once the stacked base is used

Line-Count-Exception: backend/utils/subscription.py | 1680 -> 1687 | one keyword argument (`byok_exempt`) threaded through `enforce_chat_quota` and `enforce_desktop_chat_quota` with its docstring; the surfaces that hand out Omi's own key must be able to say so at the one gate that decides BYOK exemption

### Product invariants

- `INV-DATA-1` — no deployment authority, credential, or routing changes; the relay's quota reads use `get_customer_firestore_client()`, the seam quota and `llm_usage` already use. Preserves the existing authority.
- `INV-MEM-4` — inherited from #12592 (S0), which this PR is stacked on: no memory promotion, admission, or Long-term authority path is touched by either commit; the only Firestore writes added are the S0 spend ledger rows and this PR's `llm_usage` admission counter and `chat_quota_events` question, none of which is memory.

### Failure class

Inherited from #12592 (S0), which this PR is stacked on: S0's three `fix(backend):`


kodjima33 (approved):
Flagged-author elevated scrutiny: diff clean. Shard of a stacked free-tier feature series (base != main), not a scoped bug fix — approve-only.

kodjima33 (approved):
Flagged author, re-evaluated on new commit. Backend feature (quota-gate realtime relay + idempotent hub turns), base=feat/free-tier-s0-ledger-attribution (not main). No abuse signals (no workflow/secret/auth touch). Feature-shaped, not a bug fix -> approve-only per confidence gate.

PR #12611, opened 2026-09-02
Resolve the transcription allowance once, and answer both legacy questions from it (free tier S16)

Git-on-my-level:
## S16 — one transcription-allowance answer

Shard S16 of the ratified local-models free-tier program (`omi-knowledge-base/projects/local-models-free-tier/implementation/60-shards.md`; spec in `40-mobile-pendant-port.md` §4.1). Stacked on #12606 (S15) → #12592 (S0); merge those first. Backend, plus the generated Flutter wire model for the new response field (`app/lib/backend/schema/gen/subscription_usage_wire.g.dart`, regenerated by `scripts/generate_dart_models.py --group subscription_usage` after adding the schema to its `subscription_usage` group) and the four generated TypeScript OpenAPI type files (`scripts/generate_ts_openapi_types.py`); no hand-written client change.

**Live defect closed:** two functions answered "may this user use Omi's managed STT?" and could disagree for the same uid. When the valid-subscription lookup returned `None` (an inactive basic account; an absent document is provisioned as basic, and an expired paid one downgraded to basic, before either function is asked), `has_transcription_credits` said no while `get_remaining_transcription_seconds` fell back to the basic plan's minutes. The realtime listen path hit exactly that: the connect-time gate declared such a user exhausted and sent the credit-limit notification, then the mid-session refresh flipped them back onto a billed socket.

### What changed

- **`resolve_transcription_allowance(uid, source, *, subscription=, usage=)`** in `utils/subscription.py` returns one `TranscriptionAllowance(mode, remaining_seconds, reason)`: `managed` (Omi-billed socket may open), `on_device` (no managed minutes: the plan's are spent, the subscription is not valid, or the answer could not be resolved; the local engine is free on every plan), or `blocked` (desktop trial paywall). Order: marketplace reviewer → paywall → BYOK (Deepgram key on this request and enrolled) → plan → unlimited → remaining. **It never raises.** A dependency it cannot read resolves `on_device / 0 / allowance_unavailable` — including the trial paywall's own lookups, which fail open for every other caller (a Firebase blip never paywalls a paying user): the allowance path asks `is_trial_paywalled(..., strict=True)`, a new keyword under which a lookup error re-raises instead of answering False, an account record with no creation timestamp counts as an error, and a BYOK exemption needs the validated Deepgram key on this request (ordinary callers are still exempted by a stored fingerprint alone; the allowance path is not, because the socket would then run on Omi's key). Strict answers are cached under their own key (`…:deepgram:strict`), so an ordinary caller's cached fail-open `False` is never consumed by the allowance, and `clear_trial_paywall_cache` clears that key with the others. No other caller's answer changes. The Deepgram-keyed paywall cache entry this path is the first to write is now among the keys `clear_trial_paywall_cache` deletes, so a downgrade or BYOK deactivation cannot leave a stale not-paywalled answer serving managed minutes for the cache TTL; a usage record it cannot trust (negative, non-integer, not a mapping) resolves `on_device / 0 / usage_invalid`. Reasons are a fixed vocabulary of nine labels.
- **An invalid subscription has no managed minutes** (`subscription_inactive`), for both questions. Previously one said no and the other said "the basic plan's minutes". Fail closed: the free local path, never a billed socket.
- **Both legacy functions are thin wrappers:** the same call with the same arguments, projected to `.managed` / `.remaining_seconds`.
- **An expired paid subscription** is downgraded to a fresh basic one by the real lookup and gets basic's minutes (pinned through the real lookup in the router suite); `None` is only ever an inactive basic account.
- **Two deliberate behaviour changes beyond the defect,** both toward the answer the rest of the tree already gives: (a) the paywall check passes `required_byok_provider='deepgram'`, as the relay does, so an enrolled Deepgram-BYOK request from a trial-expired desktop user is `managed / byok` (they pay Deepgram; previously `blocked`); (b) a `MARKETPLACE_APP_REVIEWERS` uid resolves `managed / None / marketplace_reviewer` before the paywall is consulted, matching the synthetic unlimited subscription the snapshot endpoint already returns for them (previously their real plan's cap and the trial paywall applied to the listen gate while the snapshot said unlimited).
- **No plan quota from the environment** (`omi-knowledge-base/NOW.md` item 4): the allowance comes from the catalog (`allocation_limit`), not through `get_plan_limits`, whose `BASIC_TIER_MINUTES_LIMIT_PER_MONTH` / `PLUS_TIER_MINUTES_LIMIT_PER_MONTH` overlays outrank the catalog. Production's overlay equals the catalog today (basic 300 min = 18,000 s; plus has no overlay, 1,500 min), so **no served number changes in prod**. Dev's overlay (`1000000` minutes for basic) stops applying to these two functions.
- **Exposed on the startup snapshot:** `GET /v1/users/me/subscription` gains an optional `transcription_allowance` object (`mode`, `remaining_seconds`, `reason`), resolved by the same function the listen socket enforces from the BYOK enrolment, subscription and usage the snapshot already read, on every branch including the metered LLM-only BYOK one (the paywall helper keeps its own cached reads; those are not shared), with `X-App-Platform` playing the listen `source` role. Additive optional field; the app-client OpenAPI snapshot is regenerated.

### Automatic-or-dead check

`backend/tests/unit/test_transcription_allowance.py` (59 tests): a 21-row allowance matrix {inactive, basic under/at/over cap and with no usage yet, plus under/over, unlimited, unlimited_v2, operator, architect, BYOK enrolled with header, enrolled without header, header without enrollment, trial-paywalled, paywalled with Deepgram BYOK, marketplace reviewer, paywalled reviewer, usage negative / string / null / not-a-mapping} run through the resolver and through **both** wrappers, asserting on every row that the two legacy answers are projections of the same result. Plus: both wrappers delegate with exact arguments (sentinel); each of the four dependencies broken in turn resolves to the free path through resolver and wrappers; against the **real** paywall helper: a Firebase failure resolves to the free path, a stored Deepgram fingerprint without the header stays `blocked` (and `byok` once the header is present), and an account record without a creation timestamp resolves to the free path, while `is_trial_paywalled` itself still answers False for everyone else in each case; and, with a shared in-memory cache, an ordinary caller's cached `False` leaves the allowance `blocked` and the invalidator clears every key written; a caller-supplied subscription and usage are not re-read; the endpoint hands the resolver a copy of the subscription taken before the legacy wire-plan remap rewrites it (Codex round 2 caught the aliasing); a divergent env overlay changes nothing; the paywall fires only for the desktop source. `tests/routers/test_users.py` pins the snapshot exposure, that exactly the snapshot's own reads (BYOK, subscription copy, usage) reach the resolver even when the legacy wire-plan remap rewrites the plan the client sees, that the **real** resolver with a raising dependency still returns the snapshot (no 500), that an expired paid subscription resolves to basic's minutes through the **real** lookup (run against a private copy of `database/users.py`, because two existing suites leave a mock over the real module's lookup — a pre-existing leak, noted not fixed), and that the LLM-only BYOK branch reads monthly usage exactly once for snapshot and allowance. `test_paywall_reconnect_gate.py` and `test_neo_desktop_grandfather.py` now require the Deepgram and strict-Deepgram cache keys to be cleared (seven keys).

Red-proofs applied by hand and seen red (then restored): inactive resolved as basic (3 red); the fail-closed catch narrowed (5 red incl. the endpoint test); the Deepgram-BYOK paywall exemption dropped; the reviewer step dropped; negative usage trusted; string usage coerced (2 red); the endpoint re-reading instead of reusing; the subscription passed by reference so the wire remap leaks into the allowance; a wrapper dropping `source` (2 red); the paywall asked non-strictly; the strict flag ignored inside the helper; BYOK state re-read instead of reused; the endpoint not handing over BYOK state; expired paid resolved as inactive; the Deepgram cache key left out of the invalidator; the BYOK branch re-reading usage; the fingerprint exempting in strict mode; a missing timestamp answering False in strict mode; the reviewer resolved after the paywall; strict sharing the ordinary cache key; the strict key left out of the invalidator (3 red).

Existing pins updated deliberately, with the reason in each: the two `test_subscription_plans` tests that set a cap through `get_plan_limits` now set it through `transcription_allowance_seconds`; their paywall doubles accept the new keyword; the BYOK "no Deepgram header" test now proves the plan path ran (exhausted allowance); the router tests stub the resolver at the router seam (it reads Firestore through its own bindings) except the two that exercise it for real.

### Proof

- Hermetic: the file above plus `test_subscription_plans.py`, `test_byok_security_quota.py`, `tests/routers/test_users.py` (118 together) and `test_listen_session_bootstrap.py`, `test_listen_persistence.py`, `test_neo_desktop_grandfather.py`, `test_fair_use_free_tier.py`, `test_fair_use_async.py`, `test_sync_fair_use_gate.py`, `test_dg_usage_batch.py`, `test_free_quota_gate_wiring.py` (143) green. Pyright: 0 errors on the three source files.
- Live (after deploy): a basic account whose subscription is inactive opens a listen session on the serving revision; the connect-time gate and the mid-session refresh must agree (no credit-limit notification followed by a working billed session), and `GET /v1/users/me/subscription` shows `transcription_allowance.mode == "on_device"`, `reason == "subscription_inactive"`; an active basic account before any use shows `managed` with `remaining_seconds == 18000`. Not run yet: no revision serves this code.
- Codex gpt-5.6-sol review: round 1 not mergeable (inactive-as-basic, no fail-closed); round 2 not mergeable (paywall fail-open reachable, expired-paid claim wrong, BYOK re-read, aliasing of the remapped subscription); round 3 not mergeable (Deepgram cache key never invalidated, LLM-only BYOK branch re-reads usage); round 4 not mergeable (fingerprint-only paywall exemption, unreadable account record, reviewer after paywall); round 5 not mergeable (strict consumed the ordinary cache key); round 6 **approve** on `a21b9284b5` with no remaining findings.

### Cut list

- Client changes (`SttModeResolver`, S17), copy (S18), the threshold-event repurposing.
- Deleting the env overlays themselves (the catalog-shape decision NOW.md item 4 names; this PR only stops the two allowance functions from depending on them).

### Line-count ratchet

Line-Count-Exception: backend/utils/subscription.py | 1680 -> 1805 | S15's byok_exempt keyword (#12606) plus S16's resolver and its dataclass, which replace two duplicated allowance walks with one; the growth is the fail-closed validation, the strict paywall flag threaded through three helpers, the reviewer step and the docstrings that state the contract, in the file that is already the home of every plan/allowance predicate

Line-Count-Exception: backend/routers/users.py | 2247 -> 2278 | the endpoint is split into the untouched snapshot builder, which now hands back the two reads it already made, plus a decorator that attaches the resolver's answer; the endpoint body itself is unchanged

Line-Count-Exception: backend/routers/desktop_proxy.py | 1528 -> 1632 | inherited from #12592 (S0), which this PR is stacked on; not touched by S16

### Product invariants

- `INV-MEM-4` — inherited from #12592 (S0): no memory promotion, admission or Long-term authority path is touched by any commit of this stack.

### Failure class


kodjima33 (approved):
Flagged-author elevated scrutiny: diff clean. Shard of a stacked free-tier feature series (base != main), not a scoped bug fix — approve-only.

kodjima33 (approved):
Flagged author, re-evaluated on new commit. Backend feature (resolve transcription allowance once), base=feat/free-tier-s15-realtime-quota (not main). No abuse signals. Feature-shaped -> approve-only.

PR #12612, opened 2026-09-02
Authorize managed compute per request, and stop unknown LLM features falling through to luna (free tier S1+S2)

Git-on-my-level:
## S1 — `authorize_managed_compute` and no silent LLM default

Shards S1 **and S2** of the ratified local-models free-tier program (`omi-knowledge-base/projects/local-models-free-tier/implementation/60-shards.md` rows S1 and S2; spec `10-backend-plumbing.md` §1.3) — S2 is the silent-default removal, done here because §1.3 lists both under one heading and the guardrail that proves S2 is what makes S1's "unknown feature" rule honest. Stacked on #12611 (S16) → #12606 (S15) → #12592 (S0); merge those first. Backend only; no call site changes behaviour yet (call-site adoption is S2).

### What changed

- **`authorize_managed_compute(uid, feature, funding_owner, *, subscription=, byok_active=) -> Decision`** in `utils/managed_compute.py`: the one answer to "may this request spend Omi-billed (or BYOK) compute?". Unknown feature (not an explicit model-config key) denies for everyone. Funding owner ∈ {`omi`, `byok`, `system`}. `byok` requires enrolment **and** a validated key for the feature's own provider on this request (a validated key for another provider does not count); without it the decision is `byok_not_validated`, never a downgrade to omi. `system` is uid-less background work allowed only for the free allowlist; a system claim carrying a uid is denied (`system_uid_forbidden`). A user-funded request without a uid denies before any lookup (the provisioning lookup would otherwise write a document for an empty uid). Basic + non-BYOK is allowed only for the free allowlist: `daily_summary`, `translation`, `session_titles`, `fair_use`, and `chat_*` (8 of the 49 configured features today). Paid plans (catalog `PAID_PLAN_TYPES`) are allowed for any known feature. Inactive subscription (lookup returns `None`) behaves as basic. An unrecognized or malformed plan value on a present subscription is **not** coerced to basic: on the omi path it takes the bounded identification fail-open (a forward-version paid plan is never denied while claiming the identity was authoritative), on the BYOK path it is simply unresolved.
- **Fail-open bounded to plan identification** (program Target 3): a plan-lookup exception for a *known* feature returns `allowed=True, reason=plan_unknown_fail_open, plan=None, plan_resolved=False`, so a Firestore blip cannot strip a paid user, while an already-identified-basic flow (S6) can still fail closed on `plan_resolved`. Every other failure denies with `authorization_unavailable`. Decisions carry a fixed 13-label reason vocabulary; `raise_if_denied()` raises `ManagedComputeDenied` for callers that map to 402; `as_dict()` is the wire form for a job payload (the Decision travels with enqueued work in S2).
- **No silent default:** `_get_model_config` raises `UnknownLLMFeature` (a `KeyError`) for a feature absent from the pinned map and the active profile, through `get_model`/`get_model_config`/`get_provider`/`get_route_ref`/`get_llm` (`get_route_ref` checks membership before the auto-lane map, so an auto-lane entry can never resolve an unmapped feature); `_DEFAULT_CONFIG` and the unused `get_default_config` export are gone. Survey (2026-09-02, re-verified by the guardrail below): every production call site passes an explicit key; the non-literal sites are the persona ternary, the two profile iterators, `get_llm`'s pass-through, `_so_gemini`, the screen-frame judge default, and the new primitive's provider lookup. No map entry was added.
- **No plan quota or allowlist from the environment:** the allowlist is a module frozenset plus the `chat_` prefix; plan identity comes from the valid-subscription lookup and the catalog.

### Automatic-or-dead check

`backend/tests/unit/test_authorize_managed_compute.py` (96): the named proof — every configured feature as basic non-BYOK, allowed iff on the enumerated allowlist; unknown feature denied across every plan and funding owner; BYOK four ways incl. a wrong-provider key; every paid plan; inactive as basic; malformed plan fail-open on omi and unresolved on BYOK; plan-lookup fail-open for known features only; every other dependency failure denied; no uid denied before any lookup; system with a uid denied; pre-read state not re-read; reason vocabulary produced and closed. `test_llm_gateway_coverage_guardrails.py::test_resolver_feature_literals_are_configured_and_dynamic_sites_are_reviewed` walks `backend/` for resolver calls, following import aliases, module-qualified names, `cast()` wrappers and one-step assignments (unfollowable alias shapes are reported as dynamic): every literal feature must be a configured key and the exact set of dynamic sites is pinned; seven snippet-level tests pin the scanner itself. An unmapped auto-lane entry raises. `test_omi_qos_tiers.py` and `test_llm_gateway_route_refs.py` now pin the raise where they pinned the luna fallback.

Red-proofs applied by hand and seen red (then restored): allowlist returns True (41 red); unknown-feature check skipped (18); BYOK success forced false; BYOK-without-key downgraded to omi; key-without-enrolment allowed; one paid plan dropped; `None` lookup treated as plus; fail-open denied; other-dependency failure fail-opened; sentinel ignored; `raise_if_denied` returns on deny; reason renamed; uid guard dropped (5); junk plan coerced to basic (4); wrong-provider key accepted; system uid check skipped; luna fallback reintroduced (both rewritten tests); a scratch module calling `get_llm('made_up_feature')` (guardrail); the prefilter matching call spelling only; `get_model_config` dropped from the scanner; alias collection skipped; unfollowable dict not reported; auto-lane consulted before membership.

### Proof

- Hermetic: the four files above (222 tests, `-m "not slow"`, green in either collection order and as one session with the twelve other suites the preflight selects — `test_omi_qos_tiers.py` no longer installs langchain/BYOK stubs at import time (they leaked into every later-collected module and the suite broke whenever `clients` was already imported, which the preflight hit intermittently); it now loads a fresh `clients` against scoped stubs inside a module fixture and restores `sys.modules` afterwards, with a leak assertion; `test_direct_provider_usage_stays_inside_approved_boundaries` is slow and already fails on `origin/main` for a stale allowlist entry unrelated to this change). Pyright 0 errors on `managed_compute.py`, `model_config.py`, `clients.py`. Guardrail tree walk 0.27 s CPU isolated, 0.61 s under the preflight's parallel load, so it is listed in `tests/fast_unit_duration_allowlist.txt` next to the existing inventory guardrail (an intentional whole-tree walk, not a per-test regression).
- Live: nothing to observe until S2 adopts the primitive; this PR changes no served behaviour except that an unmapped feature name now raises instead of silently routing to luna, which the guardrail proves no production call makes.
- Codex gpt-5.6-sol review: round 1 not mergeable (combined suite red from the QoS stub leak; malformed plan coerced to basic; guardrail alias bypasses; auto-lane escape; missing wrong-provider proof; system with a uid), all addressed; round 2 mergeable after should-fixes (falsey malformed subscription coerced to basic; the QoS teardown could evict a real `utils.byok`), both addressed; round 3 **approve** on `c6e29f5834` with no blocking findings (one pre-existing hygiene note: the QoS fixture leaves a fake `ANTHROPIC_API_KEY` via `setdefault`; left for a follow-up).

### Cut list

- Call-site adoption at `get_llm`, the gateway request context, `desktop_proxy`, and job enqueue (S2).
- Any change to MODEL_QOS profiles (never, per §1.3).
- The inventory columns for plan-gating / `funding_owner` / fallback-on-deny (S7).

### Line-count ratchet

Line-Count-Exception: backend/utils/subscription.py | 1680 -> 1805 | inherited from #12606 (S15) and #12611 (S16), which this PR is stacked on; not touched by S1

Line-Count-Exception: backend/routers/users.py | 2247 -> 2278 | inherited from #12611 (S16); not touched by S1

Line-Count-Exception: backend/routers/desktop_proxy.py | 1528 -> 1632 | inherited from #12592 (S0); not touched by S1

### Product invariants

- `INV-MEM-4` — inherited from #12592 (S0): no memory promotion, admission or Long-term authority path is touched by any commit of this stack.

### Failure class


kodjima33 (approved):
Flagged-author elevated scrutiny: diff clean (no workflow/auth/eval/dependency signals). Shard of a stacked free-tier feature series (base != main), not a scoped bug fix — approve-only.

kodjima33 (approved):
Flagged author, re-evaluated on new commit. Backend feature (authorize managed compute per request), base=feat/free-tier-s16-transcription-allowance (not main). No abuse signals. Feature-shaped -> approve-only.

PR #12615, opened 2026-09-02
Onboarding: demote the permission "Skip for now" to a text link

Git-on-my-level:
## What changed and why

Every permission step in the second-brain onboarding drew its escape as a **full-width secondary capsule directly under the full-width "Allow …" pill**. Two stacked capsules is the shape of a choice between equals, and on a permission card it is not one — the microphone, system audio, Screen Recording and the rest are what Omi runs on, and the skip is the exit, not the alternative. We want the default path to be granting.

`permStepWidget` now routes both of its escapes — the ask branch's "Skip for now" and the reopen branch's "Later — X stays off until you reopen" — through one new `skipLink`: `InkType.statusLabel` in `Ink.secondary`, underlined, centred under the primary action.

Because every permission step is a `permStepWidget` call, the one change covers all seven: Microphone, System audio, Screen Recording, Accessibility, Automation, Notifications, Full Disk Access.

**The de-emphasis is visual only.** The capsule this replaces was itself a correction of a bare `.plain` caption that people missed (the comment it replaces says so), so the link keeps what that correction was actually buying:

- an `.underline()` and a pointing-hand cursor, so it still reads as pressable rather than as a footnote;
- a 28 pt `skipLinkMinHeight` with `.contentShape(Rectangle())`, because a `.plain` button hit-tests only what it renders and 12 pt of glyph is not a control.

`Ink.secondary`, never `Ink.tertiary` — glass carries two rungs and the third rung is banned on first-run surfaces by `testStaticCheckerFirstRunSurfacesNeverSetTheThirdTypeRungOnGlass`.

**Deliberately out of scope:** the screen-demo step's "Skip for now". That step is not a permission grant, so it keeps its secondary capsule; the new checker is scoped to `permStepWidget` rather than to the file for exactly that reason.

None. `INV-UI-1` is untouched — no colour token changes, and `product/invariants/brand-ui.md` explicitly says not to cite it in routine UI PRs.

## How it was verified

Local `swift test` in a clean worktree at `origin/main`:

```
swift test --filter OnboardingGlassChromeTests
  → Executed 17 tests, with 0 failures

swift test --filter "OnboardingFlowTests|SBOnboardingNotificationsStepTests"
  → Executed 40 tests, with 0 failures
```

Gates run locally, all clean:

- `scripts/swift-format-wrapper.sh lint` on both changed files — no diagnostics
- `scripts/swiftlint-wrapper.sh lint` on `SBOnboardingView.swift` — no violations
- `scripts/check_desktop_test_quality.py` — at or below baseline
- `.github/scripts/check-desktop-changelog.py` — fragment found

A named QA bundle (`omi-skip-qa`, bundle id `com.omi.omi-skip-qa`) was built and installed from this branch for hands-on review. It launches signed-out with no TCC grants, so signing in walks the onboarding from step 0 with every permission card in its ungranted `Allow …` state — the state this PR changes. Production Omi and Omi Beta are untouched.

**Not verified by me:** the rendered result. Screen capture on the build machine is stale, and this repo's own note on desktop UI verification is that window capture drops CALayer filters and cannot see system panels — so the visual check is a human pass on that bundle, not something I am claiming here. What is asserted below is the view-tree contract, not the pixels.

**Also not verified locally:** the pinned Xcode 16.4 strict-concurrency build. This machine has only Xcode 26.6, so `scripts/run-swift-ci.sh --test` cannot select the CI toolchain. The change adds no concurrency-relevant code (one `@escaping () -> Void` view helper on an already-`@MainActor` view), but CI is the authority here.

## Tests

Two source-inspection checkers added to `OnboardingGlassChromeTests`, alongside the existing first-run static contracts:

- `testStaticCheckerPermissionEscapesAreLinksRatherThanASecondCapsule` — brace-matches `permStepWidget`'s own body and asserts both escapes go through `skipLink` and that no branch spends an unconditional secondary capsule. Scoped to that one function, since `InkButtonStyle(kind: .secondary)` stays correct elsewhere in the file (a retry, a re-check, the screen-demo skip).
- `testStaticCheckerTheSkipLinkStaysAComfortableTarget` — the underline, the pointing-hand cursor, the 28 pt reserved height, and the `contentShape` that makes that height hit-testable. This is the guard against the de-emphasis quietly becoming an unhittable caption on some later edit.

Source inspection rather than behaviour for the same reason the surrounding checkers give: which token a call site names has no runtime signal without a window server.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12615?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Onboarding/SecondBrain/SBOnboardingView.swift
+++ b/desktop/macos/Desktop/Sources/Onboarding/SecondBrain/SBOnboardingView.swift
@@ -513,39 +513,8 @@ struct SBOnboardingView: View {
     .frame(maxWidth: 380, alignment: .leading)
   }
 
-  /// A hit target no smaller than this, whatever the type is set at. `statusLabel` is 12 pt, and 12 pt
-  /// of glyph is not a control — the padding is what keeps the de-emphasis a *visual* one rather than
-  /// a usability tax on the person who genuinely wants out.
-  private static let skipLinkMinHeight: CGFloat = 28
-
-  /// The one way past a permission step.
-  ///
-  /// Deliberately the quietest actionable thing on the card: `statusLabel` in `Ink.secondary` — never
-  /// `Ink.tertiary`, which measures under AA on glass and is banned on first-run surfaces — underlined
-  /// so it still reads as pressable rather than as a caption, and centred under the primary action so
-  /// it is where the eye already is. It never becomes a capsule: a capsule beside a capsule is a fork
-  /// in the road, and this is a side door.
   private func skipLink(_ title: String, action: @escaping () -> Void) -> some View {
-    HStack {
-      Spacer(minLength: 0)
-      Button(action: action) {
-        Text(title)
-          .inkStyle(InkType.statusLabel, color: Ink.secondary)
-          .underline()
-          .fixedSize(horizontal: false, vertical: true)
-          .multilineTextAlignment(.center)
-          .padding(.horizontal, 12)
-          .frame(minHeight: Self.skipLinkMinHeight)
-          // The label is the hit test, and a run of 12 pt type is a thin one. Without this the
-          // padding above is decoration: SwiftUI hit-tests what `.plain` actually renders.
-          .contentShape(Rectangle())
-      }
-      .buttonStyle(.plain)
-      .onHover { inside in
-        if inside { NSCursor.pointingHand.push() } else { NSCursor.pop() }
-      }
-      Spacer(minLength: 0)
-    }
+    SBSkipLink(title: title, action: action)
   }
 
   @ViewBuilder private var filesWidget: some View {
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Both findings were valid and are fixed in `1022657829`.

**Cursor balance** — correct, and it was the worse of the two: pressing the link calls `action()`, which advances the step and unmounts the card with the pointer still over it, so `onHover(false)` never arrived and the pushed pointing-hand survived onto the next step's controls. `skipLink` is now `SBSkipLink`, a view owning a `didPushCursor` flag, popping on both exit paths (the tap before it advances, and `onDisappear`) — the `LiveTranscriptExpandTap` idiom you pointed at. It had to become a view rather than stay a method so the flag is per link instead of one flag shared by every escape the flow draws. The flag also closes a second imbalance I'd missed: an unguarded re-hover pushed twice.

**Test scoping** — also correct, and a guard that passes vacuously is worse than no guard. `.underline()` was satisfied by `promiseWidget`'s GitHub link and `.contentShape(Rectangle())` by two disclosure rows, so the checker would have stayed green through the exact regression it names. It now slices to `SBSkipLink`'s body via a generalised `declarationBody`, which `permStepWidgetBody` shares.

Verified rather than assumed: with the fix in place, deleting `.underline()` from `SBSkipLink` fails the checker — the same mutation passed before. Also added `testStaticCheckerTheSkipLinkBalancesItsCursorPushes` covering the flag and both exit paths, so the push/pop contract is guarded and not just written down. Suite is 18/18.

PR #12621, opened 2026-09-02
Route basic desktop processing through one free-tier policy, terminal and fail-closed, behind a flag (free tier S6)

Git-on-my-level:
## S6 — basic desktop processing through one free-tier policy, behind `FREE_TIER_LOCAL_PROCESSING`

Shard S6 of the ratified local-models free-tier program (`60-shards.md` row S6; spec `10-backend-plumbing.md` §1.2, Targets 2–3). Stacked on #12612 (S1+S2) → #12611 (S16) → #12606 (S15) → #12592 (S0); merge those first. Backend only. **Dark by default:** the flag is off in every environment; nothing served changes until it is set.

### What changed

- **`utils/free_tier_processing_policy.py`** — a pure policy. Inputs: source, `force_process`, `is_reprocess`, `has_projection`, and one injected `authorize_managed_compute` decision for the coordinator's structure feature (`conv_structure`, an off-allowlist managed feature). Output: `FreeTierProcessingPlan(mode ∈ {process_normally, store_projection, deterministic_minimum}, reason, decision)`. Identification fail-open ⇒ paid path (Target 3). Identified basic ⇒ minimum (or projection when S4 supplies one), **including** first-open `force_process` and reprocess: a basic user's enrichment is a managed call and is denied the same way (§1.2 "terminally, not first-open luna"). Any other deny or any internal error ⇒ minimum. Never raises; never looks anything up itself, so there is one plan lookup and no third policy.
- **Coordinator branch** in `process_conversation`: when the flag is on and the source is desktop, resolve the plan; `process_normally` falls through to today's path (JIT first-open obligation block unchanged); otherwise `_store_deterministic_minimum` persists the no-LLM structured shape **without** the first-open markers (`deferred=False`, `status=completed`) and returns — no `_get_structured`, `extract_memories`, `trigger_conversation_apps`, folder assignment, or `initialize_first_open_work`. Flag off ⇒ the legacy `should_defer_desktop_processing` branch, byte-identical. The funding owner is `byok` only when this request carries a validated key for the feature's provider, via S1's helper (made public as `request_carries_validated_byok_key`), so the only dynamic resolver site stays in `utils/managed_compute.py` and the S1 guardrail holds.
- **The terminal state is durable.** The persist writes an unmodeled `terminal_no_derived_effects` marker beside the `jit_first_open` clear; a normal persist writes it back to `None`. Without it a Cloud Task retry was a live money leak: the minimum persists `status=completed`, a later step fails, and the replay skips `process_conversation` entirely — so the request-scoped disposition defaulted back to `RUN` and the empty-bundle fallback ran `extract_memories`, a managed call, on a basic user's conversation. Deliberately unmodeled: adding a field to `Conversation` would restale four generated client artifacts for a server-only concern.
- **Flag-off stays byte-identical.** The marker clear is written only when the flag-on desktop policy ran and returned `process_normally` (the basic→paid upgrade). Flag-off, non-desktop, and any path that never consulted the policy persist the pre-shard dict with **no new key at all** — missing versus explicit-null is a real Firestore distinction, and a dark rollout must not stamp a field onto every conversation.
- **The paid-path fail-open is now visible.** One content-free `record_fallback` fires from the coordinator when the reason is `plan_identification_fail_open`, so a plan-lookup outage routing basic traffic to the managed path shows up in rollout monitoring instead of being silent. Emitted from the coordinator, not the policy module, which stays pure and side-effect-free by design.
- **The synchronous `POST /v1/conversations` route** observes the same disposition. It reported persistence `True` and then called `trigger_external_integrations` unconditionally, so one coordinator had two callers disagreeing about whether derived effects may run. The other two fan-out sites are the finalizer (already handled) and `routers/integration.py` (external-integration source, which this branch never selects).
- **`_store_projected_conversation`** is the S4 seam; it behaves as the minimum until S4 wires `client_processing`.
- **The terminal write clears `jit_first_open`.** Firestore persists with `merge=True` and `Conversation` does not model that field, so a paid→basic reprocess would otherwise leave a pending obligation on the document and a later open would run exactly the managed work the policy denied.
- **`utils/conversations/finalizer.py`** — a terminal store now reports persistence honestly (`True`) with a `TERMINAL_NO_DERIVED_EFFECTS` disposition. Reporting `False` made the durable finalizer read a successful persist as a lost race and return `fenced`. The finalizer suppresses exactly the intelligence bundle, the empty-bundle memory fallback, and third-party app webhooks; the capture-arrival receipt, keyframe outbox and Chat intent still run, because §1.7 suppresses Pinecone/memories/goals/apps and degrades capture for nobody — a free-tier desktop meeting must still wake Chat.
- **Not changed:** `should_defer_desktop_processing` and its fail-open (the flag-off path), `jit_first_open_policy.py`, `jit_first_open_worker.py`, the memory sweep producer (S5), anything client-side.

### Automatic-or-dead check

`backend/tests/unit/test_free_tier_entrypoint_matrix.py` (173): the program's named S6 proof — every entry point converging on `process_conversation` × {basic+projection, basic-no-projection, paid, BYOK, unknown-plan, reprocess-force, sync, merge}, asserting which managed calls fire and which persistence disposition is observed, with the flag on, and the same matrix with the flag off asserting today's behaviour (reversibility). Plus `test_free_tier_processing_policy.py` (29: policy matrix, one decision per resolve, a raising `decision_for` yielding `policy_unavailable`, every S1 reason walked through the policy with the produced set equal to `PROCESSING_REASONS`, feature configured and off-allowlist, flag parses only `true`) and `test_process_conversation_free_tier_branch.py` (14: minimum with no managed seam and no marker; a two-attempt retry over a persisted minimum emitting no derived effects; the pending first-open obligation cleared; the persistence disposition surfaced; the finalizer completing without extracting memories while still publishing the desktop meeting receipt and Chat intent; a raising BYOK lookup degrading to the minimum; force+reprocess do not rescue; flag off never consults the policy).

Red-proofs applied by hand and seen red (then restored): policy allow forced true (7 red); a second decision call; an extra reason label; the structure feature swapped for an allowlisted one; the flag helper ignoring the constant; the env parse accepting "1"; the coordinator ignoring the plan; the deny branch always taken; force/reprocess rescuing basic; the flag reading true; the funding owner forced to omi; the plan forced to process_normally after resolve (matrix basic rows red); the flag-off legacy branch deleted (flag-off basic rows red).

### Proof

- Hermetic: the three files above plus `test_authorize_managed_compute.py`, `test_llm_gateway_coverage_guardrails.py`, `test_lazy_conversation_processing.py`, `test_jit_first_open_policy.py`, `test_conversation_finalizer_failure_logging.py` `test_listen_finalization_cloud_tasks.py` and `test_free_tier_sync_route_fanout.py` (the real route function, since the coordinator matrix cannot see caller-side effects) green together in either order (499 across those ten files, one slow test deselected, in either order, and green file-isolated under the CI runner's own env); pyright 0 errors on the policy, the coordinator, the finalizer and `managed_compute.py`.
- Live (after deploy, flag on for a dogfood uid per `50-rollout-and-verification.md`): a basic desktop conversation finalises with the minimum title, `deferred=false`, `status=completed`, and no `conv_structure` row for that uid in `llm_gateway_attempts`; opening it does not trigger a first-open reprocess; a paid uid's conversation is unchanged. Not run yet: no revision serves this code and the flag is off.
- Codex gpt-5.6-sol adversarial review: round 1 returned five findings (pending first-open obligation, the fenced-on-success persistence report, funding-owner resolution outside the policy guard, matrix cells that discarded observer results, an open reason vocabulary); all five were fixed. Round 2 returned two criticals — the terminal disposition was request-scoped so a finalizer retry re-extracted memories, and the synchronous create route still fanned out to apps — both fixed on this head; round 3 returned two more — flag-off was no longer byte-identical because the marker clear was unconditional, and the paid fail-open emitted no fallback telemetry — both fixed. Round 4 returned **mergeable** with no findings, and confirmed that keeping the fallback emission outside the pure policy module is correct (the advisory checker's warning on `free_tier_processing_policy.py:145` is a lexical false positive, left standing deliberately).

### Cut list

- The deterministic-minimum title algorithm and `processing_state` field (S3 upgrades `_build_deferred_structured` in place).
- `client_processing` projection storage and hash binding (S4; the `store_projection` seam is here).
- Memory-formation suppression for basic (S5) and the sweep producer gate.
- Turning the flag on anywhere.

### Line-count ratchet

Line-Count-Exception: backend/utils/conversations/process_conversation.py | 2628 -> 2839 | the flag-gated policy branch, the terminal minimum store, and the S4 projection seam; the coordinator is the program's mandated enforcement point (§1.2 "inside process_conversation + derived-effect boundary")

Line-Count-Exception: backend/routers/conversations.py | 1963 -> 1976 | observe the coordinator's derived-effects disposition in the synchronous create route so a terminal free-tier minimum does not fan out to app webhooks

Line-Count-Exception: backend/utils/subscription.py | 1680 -> 1805 | inherited from #12606 (S15) and #12611 (S16); not touched by S6

Line-Count-Exception: backend/routers/users.py | 2247 -> 2278 | inherited from #12611 (S16); not touched by S6

Line-Count-Exception: backend/routers/desktop_proxy.py | 1528 -> 1632 | inherited from #12592 (S0); not touched by S6

### Product invariants

- `INV-MEM-4` — inherited from #12592 (S0): no memory promotion, admission or Long-term authority path is touched by any commit of this stack. S6 removes managed calls for basic; it adds no memory write.
- `INV-TASK-2` — `process_conversation.py` is a task-capture authority path. S6 adds a branch that persists the no-LLM minimum and returns before action-item extraction runs; it proposes no task, writes no task, and changes nothing about how the existing paths capture. The flag-off path is byte-identical.

### Failure class


kodjima33 (approved):
Flagged author. Backend feature (free-tier processing policy behind flag), base=feat/free-tier-s1-authorize-managed-compute (not main). No abuse signals. Feature-shaped -> approve-only.

PR #12623, opened 2026-09-02
fix(desktop): make the first push-to-talk press after idle actually capture

### LINKED ISSUE #12616: macOS PTT: the first hold after idle is lost in the capture path (fresh-install 0→1 gap)
## Summary

A fresh install on beta `0.12.267` (2026-09-02 15:42Z) saw the new first-real-app card ("I can see {App}. Hold ⌥ and ask me anything…"), pressed PTT five times, and never committed a turn. Nothing was wrong with the user's intent; every press was lost in the capture path and reported back as "Hold longer to record".

In the trailing 7 days of macOS traffic `capture_never_operational` is the largest single PTT failure class (357 users, 1,077 attempts; `committed` = 290 users). Among users who completed onboarding in the same window it hits **119 of ~180**. This is the 0→1 activation gap: the first press is the one that fails.

Evidence record: `omi-knowledge-base/projects/macos-churn-analysis/evidence/2026-09-02-first-48h-activation-and-pr-12608.md`. Population measurement and prior ranking: `omi-knowledge-base/projects/ptt-improvement/evidence/2026-08-30-capture-gate-population-measurement.md`.

## The observed sequence (one install, `ptt_audio_capture_lifecycle` + `voice_turn_terminal`, mic permission granted throughout)

| Press | Hold | `failure_class` | Audio delivered | User saw |
| 1 (in onboarding) | 228 ms | `capture_never_operational` (`capture_start_outcome=requested`, first audio bucket `none`) | 0.00 s | "Hold longer to record" |
| 2, 3 s later | 234 ms | `too_short_audible` (first audio <100 ms) | 0.17 s | same |
| 3, immediately | **947 ms** | `too_short_audible` (first audio <50 ms) | **0.01 s** | same |
| 4, 7 s after the card, 62 s after press 3 | 516 ms | `capture_never_operational` again | 0.00 s | same |
| 5, 2 s later | — | `voice_turn_started`, **no `voice_turn_terminal` ever** | — | nothing |

Three distinct defects are visible in five presses:

1. **The first press after idle gets no capture.** `capture_start_outcome=requested` never becomes `accepted`. The late-start parking path (`PushToTalkManager: mic capture start completed after turn ended — parked warm`, `parkedMicKeepAliveSeconds = 120` at `desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift:2073`) exists precisely for this, yet press 4 arrived 62 s after press 3 and was cold again. Either the parked capture was released before the keep-alive expired (terminal cleanup passes `parkWarm: false`; onboarding completed at 15:46:27 between presses 3 and 4), or parking never armed. Root-cause which.
2. **A 947 ms hold yielded 0.01 s of audio and was blamed on the user.** Capture came up ~900 ms into the hold. `hubMinTurnAudioSeconds = 0.35` (`PushToTalkManager.swift:976`) is measured on delivered audio, so capture-start latency is charged to the press. The hint "Hold longer to record" (`VoiceTurnDomain/VoiceTurnStateMachine.swift:348`) is the wrong message for this case.
3. **Recovery never fires on a first-press failure.** `PTTSilentMicRecoveryPolicy` (`PushToTalkManager.swift:8-60`) needs `consecutiveDeadTurnThreshold = 2` dead turns each ≥ `minDeadTurnSeconds = 0.25`; a 228 ms press is "too short to judge" and does not count, and the audible presses in between reset the streak. `capture_rebuild` exists (`PushToTalkManager.swift:2401-2422`) and is never reached on this pattern.

Press 5 (turn started, no terminal) is one instance; treat it as a lead. It happened while the card notification was hovered.

## Scope: one PR

Ship the three items below together; they touch the same file and the same funnel. Biggest user-visible fix is item 1.

1. **Capture is running before the user's first hold.** Make the capture path operational for the first press after launch, after onboarding completion, and after the first-real-app card is shown (`FirstRealAppCardCoordinator.swift:351` is the hook; the press is predictable there). Reuse the existing parked-warm mechanism rather than adding a second one; fix why it did not cover press 4.
2. **Capture-start latency is not charged to the press.** When capture becomes operational after key-down, measure the minimum-audio gate from first usable frame, or fall back to the batch path for that turn, rather than discarding an audible ~1 s hold. Whichever is chosen, `too_short_audible` with `ms_to_first_usable_frame_bucket` ≥ the hold length must stop occurring.
3. **The discard message says what happened.** `capture_never_operational` → a message that the mic did not start and Omi is retrying (and actually retry via `capture_rebuild` without waiting for a second dead turn on the first press of a session). Genuine sub-250 ms taps keep "Hold longer to record".

## Acceptance (each names its proof)

- [ ] A logic test in `Tests/PTTSilentMicRecoveryPolicyTests.swift` shows a first-of-session dead turn ≥ 0.25 s arms `capture_rebuild` without a second dead turn, and that an audible turn between two dead turns no longer resets the streak to zero.
- [ ] A logic test (new, next to `Tests/PushToTalkSpeechGateTests.swift`) shows a turn whose first usable frame lands at 900 ms into a 950 ms hold is not classified `too_short_audible`.
- [ ] `Tests/PTTWarmMicKeepAliveTests.swift` gains a contract that the card-shown hook and onboarding completion do not release a parked capture, and `Tests/FirstRealAppCardCoordinatorTests.swift` asserts the warm-capture request fires on `shown`.
- [ ] `Tests/PTTAttemptLifecycleRecorderTests.swift` covers the new user-facing disposition for `capture_never_operational` so the telemetry class and the hint string cannot drift apart.
- [ ] `./scripts/agent-logic-harness.sh --swift-only`, `make preflight` green; Desktop Swift CI green.
- [ ] Dogfood on a named `omi-*` QA bundle from a **fresh install** (new user-defaults domain): first ⌥ hold after onboarding commits a turn; `ptt_audio_capture_lifecycle` for that turn reports `committed`. Record the bundle SHA and the attempt id in the PR.
- [ ] Closure on live evidence, not merge: after 48 h on beta, `first_real_app_card phase=ptt_after_card` → `question_asked` conversion and the share of `capture_never_operational` among onboarding-completed users are read from PostHog (project 302298) and posted on this issue.

## Cut list

- **Voiced-detector threshold recalibration** (`zero_or_near_zero_samples`, 225 users/7d): separate mechanism, separate PR. Revival bar: this PR merged and the first-press class below 5% of new-user attempts.
- **New telemetry**: the lifecycle event already carries `capture_start_outcome`, `failure_class`, `ms_to_first_usable_frame_bucket`, `recovery_action`. Add nothing.
- **Dashboards, scheduled sweeps, SLOs**: none. The PostHog reads above are one-off closure evidence.
- **Press 5 missing terminal**: file separately if it reproduces on the fresh-install dogfood; do not chase here.

## Invariants

INV-VOICE-1 (`product/invariants/desktop-voice-turns.md`): `VoiceTurnCoordinator` stays the sole lifecycle owner; warm capture must not publish turn events. INV-AUTH-1 owner fencing applies to anything the card coordinator triggers.

Git-on-my-level:
Closes #12616.

## What was wrong

A fresh install on beta `0.12.267` pressed push-to-talk five times and committed
nothing. Every press was lost in the capture path and reported back as **"Hold
longer to record"**. In the trailing 7 days of macOS traffic
`capture_never_operational` is the largest single PTT failure class — 357 users,
1,077 attempts — and among users who completed onboarding in the same window it
hits 119 of ~180. The first press is the one that fails.

Three distinct defects, all fixed here.

### 1. The first press after idle got no capture — root cause found

The warm keep-alive (`parkedMicKeepAliveSeconds = 120`) existed for exactly this,
and press 4 arrived 62 s after press 3 and was still cold. It was not the
keep-alive expiring: **onboarding completion starts ambient transcription**, and
`AppState+Transcription.startMicCaptureIfNeeded()` deliberately calls
`PushToTalkManager.shared.releaseParkedMicCapture()` on its way into the device so
two IOProcs can never overlap on one input. That release is correct. Nothing ever
put a capture back, so every press after an ambient capture start paid a full
CoreAudio start inside the hold.

The fix reuses the existing parked-warm mechanism rather than adding a second one.
`PushToTalkManager.prewarmMicCapture(trigger:)` opens exactly the capture
`startMicCapture` would have opened, with the same frame handlers and watchdog
observers (now extracted and shared, so a warm capture cannot drift from a turn's),
and hands it to `parkMicCapture`. Four triggers:

| Trigger | Where |
| --- | --- |
| `ambientCaptureStarted` | right after the ambient session's device opens — the release above now has a matching re-arm |
| `onboardingCompleted` | `SBOnboardingModel.complete()`, after `startTranscription()` |
| `firstRealAppCard` | `FirstRealAppCardCoordinator` at `phase=shown` — the card literally says "Hold ⌥", so the press is predictable |
| `captureNotReady` | after a turn was lost to capture-start latency — the retry the hint promises |

**There is deliberately no `launch` trigger**, though the issue lists one. `PushToTalkManager.setup()` — the only caller of `warmPTTInputRouting()` — runs from `DesktopHomeSignedInStartup`, *after* `applicationDidFinishLaunching`, so a warm-up at launch finds no routing snapshot and opens the raw system default input. On a Mac with AirPods connected that is the headset microphone: it flips A2DP→HFP and collapses the user's audio to phone quality for the whole 120 s keep-alive, with nothing the user did behind it — and then the press *doesn't* adopt it (the resolved route by then is the built-in mic), so the first press is slower than with no warm-up at all. The three triggers above all follow something the user just did and all run long after routing has resolved.

**INV-VOICE-1**: the warm capture publishes no turn events and holds no lifecycle
authority. Its `MicCaptureLease` is created already parked, so every frame the
device delivers before a real turn adopts it is dropped inside
`snapshotIfActive()`. `VoiceTurnCoordinator` remains the sole lifecycle owner.
**INV-AUTH-1**: the card coordinator's new effect is an injected seam like its
existing ones, fenced by the same owner check the presenter already runs; the warm
capture itself refuses unless `RuntimeOwnerIdentity.currentOwnerId()` is non-nil.

`PTTWarmCaptureAdmission` is pure and testable, and every gate is load-bearing.
A press is consent and a warm-up is not, so admission is strictly stricter than
`preferredPTTInputOverrideDeviceID`: it never raises a TCC prompt (the microphone
must already be authorized), never runs signed-out or un-onboarded, never opens a
second capture beside one that exists, and refuses any route it cannot prove is
safe to hold open unattended (`Snapshot.unattendedWarmCaptureRoute`) — an
unresolved routing snapshot, an explicit microphone choice (the picker exists for
external and Bluetooth inputs), or a Bluetooth input are all refusals.

An in-flight warm capture is tracked as a reference, not a flag, because it has
already registered in the active-capture registry and is about to own an IOProc.
`releaseParkedMicCapture()` surrenders it too, so every existing caller — a turn
starting, terminal cleanup, a capture rebuild, ambient transcription taking the
device, an owner transition — gets the same physical handshake it already gets for
a parked capture, and `applyMicContentionPolicy` excludes it so our own warm-up
cannot reroute a real press away from the default input.

### 2. Capture-start latency is no longer charged to the press

`hubMinTurnAudioSeconds = 0.35` is measured on *delivered* audio, so a 947 ms hold
whose capture came up at the end of it delivered 0.01 s, was classified
`too_short_audible`, and blamed the user. `PTTTurnDiscardJudgement` judges on the
press instead: a hold that cleared the gate but delivered less audio than the gate
needs lost the difference to capture start, not to the user's finger. All five
discard paths (hub, buffered hub, omni/batch, warm-wait fallback, empty-buffer
backstop) resolve through one `PTTDiscardedTurnResolution` so they cannot drift.

The hold is latched at release (`PTTAttemptLifecycleRecorder.noteRelease()`), at
the one shared finalization entry and before the hub-warm branch. That matters:
on a cold realtime hub — the fresh-install path — a turn's buffered audio is
judged a second or more after the release, on the hub warm deadline or on the
connection landing. Measuring "now minus the press" there would count the hub's
wait as part of the user's press, turn every accidental tap into
`capture_never_operational`, fire a CoreAudio rebuild on it, and corrupt the very
metric this PR is judged by.

Such a turn is classified `capture_never_operational` — the same user-visible
defect as a start that never resolved, so one query still sizes the whole
first-press gap. `too_short_audible` with `ms_to_first_usable_frame_bucket` ≥ the
hold length can no longer occur. **No new telemetry**: no new event, no new field,
no new `failure_class` value.

### 3. The message says what happened, and the retry is real

New terminal reason `captureNotReady` → **"Microphone wasn't ready — retrying,
hold again"**. It is in `terminalReasonKeepsWarmCapture`, so the capture that
missed the press stays parked — that parked capture *is* the retry the hint
promises, and a `.captureNotReady` warm-up request runs behind it. Genuine
sub-350 ms taps keep "Hold longer to record".

`PTTSilentMicRecoveryPolicy` also stopped being reachable on this pattern, twice
over:

- It judged turns on *delivered* audio, so `capture_never_operational` reported
  0.00 s and read as "too short to judge" — the worst failure class was invisible
  to its own recovery. It now judges on the hold.
- The first dead turn of a session now arms `capture_rebuild` on its own. Waiting
  for a second assumes the user presses again; on a fresh install they have just
  been told to hold longer and mostly do not. After that first rebuild the ordinary
  two-turn threshold applies again, so a genuinely broken mic cannot spin.
- An audible-but-still-discarded turn no longer clears the dead-turn streak. The
  observed sequence was dead → audible → audible → dead, and the middle two wiped
  the evidence. Only a committed turn (`recordSuccessfulTurn`) clears it now.

## Verification

- `swift test` on the whole desktop package, twice consecutively: exactly two
  failures, and both are **proven** pre-existing rather than assumed — a detached
  worktree at clean `origin/main` produces the identical pair
  (`ChatDiscoverabilityTests.testDesktopCapabilitiesExistInAgentToolDeclarations`
  and `HomeDailySummaryTests.testStoreKeepsLastGoodRecordWhenFetchFails`). Note
  `HomeDailySummaryTests` passes in isolation on both trees; it fails only in a
  full-suite run, so it is a test-ordering artifact and not something this PR
  moved.
- One additional failure appeared in a single run
  (`AuthorizedToolOwnerBoundAuthTests.testMemoryReadReturnsResponseWhileOriginalOwnerRemainsCurrent`).
  It did not reproduce across two further full runs and passes in isolation (that
  suite takes 60s and is timing-sensitive), so it is recorded here as an observed
  flake rather than swept up.
- `make preflight`: 11 checks pass.
- Xcode 16.4 is not installed on this machine, so `scripts/run-swift-ci.sh` could
  not be run locally; Desktop Swift CI on this PR is the authority for the pinned
  toolchain.

### Acceptance coverage

| Acceptance item | Proof |
| --- | --- |
| First-of-session dead turn ≥ 0.25 s arms `capture_rebuild` without a second; an audible turn between two dead turns no longer resets the streak | `PTTSilentMicRecoveryPolicyTests.testFirstDeadTurnOfSessionRequestsRecoveryWithoutASecond`, `…testAudibleDiscardedTurnBetweenDeadTurnsDoesNotResetTheStreak`, `…testDeadTurnIsJudgedOnTheHoldNotOnDeliveredAudio` |
| A first usable frame at 900 ms into a 950 ms hold is not `too_short_audible` | `PTTCaptureReadinessPolicyTests.testNearlyOneSecondHoldWithLateCaptureIsNotAShortTap`, `…testLateCaptureTurnIsNotClassifiedTooShortAudible`, and end to end in `PTTAttemptLifecycleRecorderTests.testLateCaptureTurnEmitsCaptureNeverOperationalNotTooShortAudible` |
| Card-shown and onboarding completion do not release a parked capture; the card fires a warm-capture request on `shown` | `PTTWarmMicKeepAliveTests.testOnboardingCompletionAndTheCardWarmRatherThanRelease`, `…testAmbientTranscriptionReArmsTheWarmCaptureAfterDisplacingIt`, `FirstRealAppCardCoordinatorTests.testShowingTheCardAsksPushToTalkToWarmItsCapture` (+ `…testACardThatDoesNotFireNeverWarmsTheCapture`) |
| The user-facing disposition of `capture_never_operational` cannot drift from its hint | `PTTAttemptLifecycleRecorderTests.testCaptureNeverOperationalNeverTellsTheUserToHoldLonger`, `…testGenuineTapKeepsTheHoldLongerHint` |

### Review

Five adversarial review passes on this branch, which found and closed, among
others: the hold being latched at finalization rather than key-up (which made
every sub-220ms tap on a default install report `capture_never_operational` and
trip a CoreAudio rebuild — it would have corrupted the closure metric below); a
warm capture that `stopCapture()` could not stop, because `isCapturing` only
flips at the end of the HAL setup, so `waitForPhysicalStop()` inverted into
"wait for the start to finish" and handed the caller a live second IOProc; and a
launch-time warm-up that ran before routing resolved and opened the raw system
default input. Two residuals were reasoned through and deliberately shipped,
documented at the lines that make the trade: the Bluetooth re-check fires after
the flip it prevents (pinning the device would break parked-capture adoption),
and the modifier-only activation delay biases the hold low, in the direction that
can only turn a capture failure into "hold longer".

### Still owed, and deliberately not claimed here

- **Fresh-install dogfood on a named `omi-*` QA bundle.** Not done, and not
  reachable headlessly: the tier-2 `ptt-lifecycle` flow drives real push-to-talk
  turns through the automation bridge, but `beginPushToTalkForAutomation` sets
  `automationCaptureBypass` and opens no microphone, so it exercises the lifecycle
  and proves nothing about capture readiness. A real dogfood needs a signed local
  build, a fresh user-defaults domain, and a human ⌥ hold; the bundle SHA and
  attempt id belong in this section once it runs. Nothing in this PR's
  verification depends on it — the logic is covered by tests — but the acceptance
  item is not met until it does.
- **48 h PostHog closure** (`first_real_app_card phase=ptt_after_card` →
  `question_asked`, and the share of `capture_never_operational` among
  onboarding-completed users) is closure on live evidence, not on merge, and will
  be posted to #12616.
- Press 5 (turn started, no terminal) was explicitly out of scope and is untouched.

- **INV-VOICE-1** — the warm capture publishes no turn events and holds no lifecycle
  authority; its lease starts parked so frames are dropped until a real turn adopts it.
  `VoiceTurnCoordinator` remains the sole lifecycle owner, and the new terminal reason
  goes through the reducer like every other.
- **INV-AUTH-1** — the card coordinator's warm-capture effect is an injected seam fenced
  by the same owner check its presenter already runs; the warm capture itself refuses
  unless `RuntimeOwnerIdentity.currentOwnerId()` is non-nil, and a parked capture is
  released on owner transition exactly as before (`quiesceForEffectiveOwnerTransition`).
- **INV-CHAT-1** — no new store, no transcript state, no session identity. The change is
  confined to microphone capture readiness, one terminal reason, and the strings and
  buckets a discarded turn reports.

Product-Invariant: INV-VOICE-1
Product-Invariant: INV-AUTH-1
Product-Invariant: INV-CHAT-1

Failure-Class: new

Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/PushToTalkManager.swift | 2907 -> 3276 | The warm-capture path belongs beside the parked keep-alive and the device routing it reuses; an extension would need eight private capture members made internal to satisfy a line count. Roughly a third of the growth extracts the existing frame handlers and watchdog observers so a warm capture cannot drift from a turn's.
Line-Count-Exception: desktop/macos/Desktop/Sources/VoiceTurnDomain/VoiceTurnStateMachine.swift | 2426 -> 2433 | One terminal reason (captureNotReady) and its hint, in the two switches that own them.
Line-Count-Exception: desktop/macos/Desktop/Sources/AppState/AppState+Transcription.swift | 1744 -> 1756 | Re-arming the parked PTT capture the ambient session just displaced, at the one point in the file that displaces it.


Git-on-my-level:
## Review triage — `fb90334174`

Thanks. Four of the six went in; here is each one and what I did with it.

**Fixed**

- **`VoiceTurnStateMachine` / `questionOutcome` P2 — valid, and the sharpest one here.** `capture_not_ready` fell through to `default`, so a turn whose microphone never became operational emitted a `question_answered` row with outcome `error` and no `question_asked` to pair with. It belongs beside `capture_failed` for the same reason: nothing was committed. The test now spells the reason as `VoiceTurnTerminalReason.captureNotReady.rawValue` rather than a literal, so a rename in the state machine cannot silently drop it back into `default`.
- **`PTTAudioCaptureRaceTests` P2 — valid.** The test injected `.captureNotReady` straight into the reducer, which left the classifier it exists to guard entirely untested. It now derives the reason from `PTTTurnDiscardJudgement.judge(holdSeconds: 1.4, deliveredAudioSeconds: 0, minTurnAudioSeconds: 0.35)` — the actual shape of the bug — and asserts the resolution before feeding it in, so a regression that called a long silent hold a short tap fails here.
- **`PTTWarmMicKeepAliveTests:46` P2 — valid.** The re-arm assertion was a whole-file `contains`, satisfiable by a match anywhere in the file and blind to ordering. It is now scoped to the brace-matched body of `startMicCaptureIfNeeded` and asserts the re-arm follows the release, which is precisely the regression (`startMicCaptureIfNeeded` released the parked capture and nothing put one back).
- **Plus one CI failure of my own making, in the same area you flagged.** `testAnAutomationTurnIsNotJudgedOnItsHold` counted occurrences of `judgeableHoldSeconds`; a comment in the previous commit that merely *named* the symbol pushed the count 6 → 7 and reddened the lane. A headcount of the safe name measures prose. It now counts the thing that must not spread — raw `pttLifecycle.holdSeconds` has exactly one reader, the gate itself — plus the four gated call sites.

**Already addressed before this run landed**

- **`FirstRealAppCardCoordinator:370` P1 (owner transition during prewarm).** Real risk, fixed in `6ae5770a57`/`d4c5c1af49`, which post-date the reviewed commit. `prewarmMicCapture` latches `ownerID` at entry and re-checks `RuntimeOwnerIdentity.currentOwnerId() == ownerID` immediately before parking (`PushToTalkManager.swift:2444`, `:2523`), and `quiesceForEffectiveOwnerTransition` now awaits `drainInFlightWarmCapture(timeout: Self.ownerTransitionWarmCaptureWaitSeconds)`, so teardown blocks on an in-flight warm start rather than racing it. A capture opened for the previous owner is stopped and physically drained, never parked under the next one. `testReleasingAWarmCaptureKeepsTheTeardownHandleForOtherCallers` pins the drain.
- **Both `OmiApp.swift:563` P1s (launch prewarm racing ambient startup / owner teardown).** Correct, and the reason the launch trigger no longer exists. `PushToTalkManager.setup()` runs from `DesktopHomeSignedInStartup.swift:64`, *after* `applicationDidFinishLaunching`, so a launch prewarm opened the raw system default — AirPods flipped to HFP — before routing had any say. The whole hook was removed; `OmiApp.swift` is no longer in this diff.

**Not taking**

- **`ptt-lifecycle.yaml:18` `covers:` P2.** I read this one carefully and disagree on the facts. The flow does execute `PTTCaptureReadinessPolicy.swift`: the `terminal_reason: too_short` it asserts is not a literal from the manager, it is produced by `PTTTurnDiscardJudgement.judge` returning `.shortTap` and `PTTDiscardedTurnResolution` mapping that to `.tooShort`. Break the classifier and this flow goes red. What it does *not* exercise is the warm-capture readiness path, and that is a real gap — but it is a gap in scenarios, not a false `covers` claim, and adding an unattended-warm-capture e2e scenario means driving CoreAudio device state from the automation bridge, which is a larger change than this PR should carry. Worth its own issue rather than dropping an accurate annotation.

Full suite locally: the two failures on this branch (`ChatDiscoverabilityTests.testDesktopCapabilitiesExistInAgentToolDeclarations`, `HomeDailySummaryTests.testStoreKeepsLastGoodRecordWhenFetchFails`) reproduce identically in a detached worktree at clean `origin/main`, so they are pre-existing and not from this change. `make preflight` 24/24, `agent-logic-harness.sh --swift-only` green.

Git-on-my-level:
## `9478d5735d` — the second red lane was not this PR

The `Desktop Swift Static & Test Contracts` failure on `fb90334174` was
`HubEscalationTests.testPublicWebPromptIsSpeakableAndExcludesPrivateToolContext`,
which this PR does not touch. It is red on `main` right now.

`bce354d767` ("preserve realtime voice context boundaries") rewrote
`RealtimeHubTools.publicWebSearchPrompt` to open with *"Search the live public web
**thoroughly** before answering this request."* The test still asserted
`hasPrefix("Search the live public web before answering this request.")`. One
inserted adverb, no update to the assertion.

Proof rather than assertion: a detached worktree at clean `origin/main`
(`15eca77b62`, nothing of mine applied) fails that test identically —
`HubEscalationTests.swift:87`, `Executed 6 tests, with 1 failure`. Applying only
this commit's change to that same tree turns it green (`Executed 44 tests, with 0
failures` across `HubEscalationTests` + `HubSystemInstructionTests`). So every PR
touching desktop Swift currently inherits a red lane, and mine could not go green
without it.

I fixed the assertion rather than the prompt — the prompt change was deliberate,
the test simply was not updated with it. And I did not re-pin the new sentence,
which would just re-arm the same tripwire for the next wording edit. The
assertion was really checking two things: that the prompt opens by directing a
*live public web* search rather than model recall, and that the search precedes
the answer. Those are now checked as two halves of the opening sentence, so a
future rewording fails only when it changes what the prompt actually instructs.

Happy to split this into its own PR if the desktop owners would rather it not
ride along here — say the word and I will lift the commit out.

Git-on-my-level:
## `cb2c3b12b1` — third red lane, third cause that is not this PR

`Desktop Swift Static & Test Contracts` on `9478d5735d` failed
`ProactiveListenEventTests.testHandlerRecognisesProactiveMessageType` with
`bundleProxyForCurrentProcess is nil` at `NotificationService.swift:171`. The
`HubEscalationTests` break from the previous round is fixed and did not reappear.
This test passed on the two prior runs of a near-identical tree, so it is
order-dependent rather than newly broken.

**Mechanism.** The suite asserts what the `proactive_message` handler does with
*no* runtime owner, and established that premise by hoping the process had never
had one. It shares a CI batch with `AppStateOwnerFenceTests` and
`AuthorizedToolOwnerBoundAuthTests`, both of which write owner defaults, and the
latter is timing-sensitive. An owner left behind makes
`captureAuthorizationSnapshot()` non-nil, the handler passes its `hasRuntimeOwner`
gate, and delivery reaches `NotificationService.shared` — whose initialiser
constructs `UNUserNotificationCenter.current()`, which cannot exist under
SwiftPM's command-line test host. `NotificationService` already documents that
exact hazard in its `init`; nothing kept this suite on the safe side of it.

**Proven in both directions, not argued.** A throwaway suite that leaks
`authUserId` and sorts ahead of this one reproduces the CI failure exactly — same
test, same `NotificationService.swift:171`, same `NSInternalInconsistencyException`.
With the `setUp` in this commit, the same deliberate leak is green. Removing the
`setUp` again with the leak still in place brings the failure straight back, so
the fix is necessary and not merely coincident.

The premise is now established rather than inherited, restored in `tearDown` so
this suite leaks nothing itself, and asserted at the top of the test — a
regression now fails saying a sibling suite leaked an owner, instead of as an
opaque ObjC exception from a file unrelated to the test.

**Related, deliberately not fixed here.** `HomeDailySummaryTests.testStoreKeepsLastGoodRecordWhenFetchFails`
fails locally in the same family (`bundleProxyForCurrentProcess is nil`, via
`ProactiveAssistantsPlugin+NotificationSettings.swift:227`). It passes on every CI
run I have seen, so it is a latent version of the same fragility rather than a
current break, and it belongs in its own change rather than widening this one.

kodjima33 (approved):
Owner-override (desktop bug fix): fix(desktop) title, closes #12616 (real bug-labeled issue matching symptom exactly), detailed root-cause analysis (3 distinct PTT capture defects), extensive new unit/e2e test coverage, CI green, mergeStateStatus=BLOCKED/MERGEABLE (review-gate only, no conflicts). No abuse signals found (no workflow/secret/auth/eval/base64 touches; full diff reviewed).

PR #12631, opened 2026-09-02
Put private-cloud-sync bucket lifecycle under version control + Coldline@30d on chunks/

Git-on-my-level:
## What this changes

Puts the `omi-private-cloud-sync` bucket lifecycle under version control (it was managed by nothing — console/`gsutil` only) and adds one rule:

```json
{
  "action": {"type": "SetStorageClass", "storageClass": "COLDLINE"},
  "condition": {"age": 30, "matchesPrefix": ["chunks/"], "matchesStorageClass": ["STANDARD", "MULTI_REGIONAL"]}
}
```

Applied to **dev on merge**. **Prod is dispatch-only** behind a typed confirmation, an exact-main-SHA admission, and the `prod` GitHub Environment reviewer — David approves that gate; subagents never do.

Files added:

- `backend/deploy/storage-lifecycle/{prod,development,prod.rollback,development.rollback}.json` + `README.md`
- `backend/scripts/validate_storage_lifecycle.py` — offline validator, same shape as `validate_frame_request_bucket_contract.py`
- `backend/tests/unit/test_storage_lifecycle_contract.py`
- `.github/workflows/gcp_storage_lifecycle.yml` — describe → validate superset → apply → describe → assert-equal, artifacts uploaded (90 d)

No application code changes. `runtime_env.yaml`, `runtime_images.json` and the Helm values are untouched; the bucket-name binding already lives there.

Exact prod lifecycle applied (`backend/deploy/storage-lifecycle/prod.json`):

```json
{
  "bucket": "omi-private-cloud-sync",
  "project": "based-hardware",
  "lifecycle": {
    "rule": [
      {"action": {"type": "Delete"}, "condition": {"age": 3, "matchesPrefix": ["merged/"]}},
      {"action": {"type": "Delete"}, "condition": {"age": 30, "matchesPrefix": ["playback/"]}},
      {
        "action": {"type": "SetStorageClass", "storageClass": "COLDLINE"},
        "condition": {"age": 30, "matchesPrefix": ["chunks/"], "matchesStorageClass": ["STANDARD", "MULTI_REGIONAL"]}
      }
    ]
  }
}
```

## Why

Plan: `omi-knowledge-base/projects/gcp-cost-efficiency/evidence/2026-09-02-coldline-rollout-plan.md`
Scoping report: `omi-knowledge-base/projects/gcp-cost-efficiency/evidence/2026-09-02-sync-bucket-tiering-scope.md`

`chunks/` is 97.4% of a 154 TB bucket, append-only, and 91.8% of its bytes are older than 30 days. Expected prod effect (MODELLED from a MEASURED $108.97/day storage line): storage falls to ~$38/day, plus ~$0.8/day retrieval (ceiling $7.9/day) and ≤$1.5/day Coldline Class B — **net ≈ −$70/day, ≈ −$2.1k/month**. One-time transition cost ~**$6.9k** (≈346 M objects × $0.020/1,000, billed at the destination class). Payback ≈ 3.3 months. Nothing is deleted, in either direction.

**Compatibility, in one line:** no reader can observe storage class — every `chunks/` reader is `list_blobs` → `download_as_bytes` → in-memory concat; **no chunk is ever signed-URL'd** (only `merged/*.wav` and `playback/*.mp3` are); and **`compose`/`rewrite` are not used anywhere** in `backend/`. The full 16-row audit is §2 of the plan.

## The one thing to know before reviewing

**`gcloud storage buckets update --lifecycle-file` REPLACES the whole lifecycle config.** A desired file that forgets a live rule silently deletes that rule. That is why:

- the two legacy Delete rules (`merged/` age 3, `playback/` age 30) are re-declared **verbatim** in every environment file;
- the workflow captures `before.json` with the deploy identity and the validator runs as a **superset check against that live document**, not against an assumed file — it fails closed if `before.json` contains a rule the desired file lacks (override is the explicit `--allow-rule-removal`, used only for the rollback variant);
- the validator refuses any Delete rule that could reach `chunks/` (an **unscoped** Delete counts), any `SetStorageClass` rule not prefix-scoped to exactly `["chunks/"]`, and any `isLive: false` / noncurrent-version rule — a bucket-wide noncurrent rule would Coldline `playback/` noncurrent versions that the 30-day Delete rule then early-deletes.

`matchesPrefix: ["chunks/"]` is the load-bearing line of the whole change: because no `merged/`/`playback/` object ever becomes Coldline, those Delete rules never trigger an early-delete fee.

## Two things the read-only bot could not verify, and how the workflow closes them

1. **Bucket notification configs** (bot lacks `serviceusage.services.use`). The workflow runs `gcloud storage buckets notifications list` with the deploy identity and the validator **fails the run** if the list is non-empty — a stray config would fan out ~346 M `OBJECT_METADATA_UPDATE` events.
2. **The dev bucket's current lifecycle** (bot lacks `storage.buckets.get` on `omi-dev-private-cloud-sync`). `development.json` is built from the same two legacy Delete rules plus the Coldline rule; the workflow's `before.json` superset check is what actually proves nothing is dropped, and it fails closed if dev holds a rule this file does not.

**Do not merge until the dev deploy identity has been confirmed to hold `storage.buckets.update` on `omi-dev-private-cloud-sync`** (and the prod SA the same on `omi-private-cloud-sync` before the prod dispatch). If it does not, the push-trigger run fails loudly — acceptable but noisy, and it is an IAM change for David, not for this PR.

## Rollout

**1. Merge** → the `push` trigger applies `development.json` to `omi-dev-private-cloud-sync` (GitHub environment `development`, no approval). `before.json` / `after.json` / `notifications.json` upload as run artifacts. Resume the Hermes watch cron the same day (below).

**2. Dev bake: 48 h from the `after.json` timestamp.** Pass = **all** of:

- **Config integrity** — `after.json` lifecycle == the checked-in file (workflow step, hard fail).
- **Transition happens** — Cloud Monitoring `storage.googleapis.com/storage/v2/total_bytes` for `omi-dev-private-cloud-sync` shows a `storage_class=COLDLINE` series with **≥ 50% of live bytes by T+48 h** (dev is 1.95 TB, essentially all >30 d; expect >90%).
- **Billing rows appear with the SKU strings** — billing export `based-hardware.gcp_billing_export.gcp_billing_export_resource_v1_01B287_9348DC_02D256`, `resource.name = "omi-dev-private-cloud-sync"`, new `sku.description LIKE "%Coldline%"` rows (storage + Class A). Today the export contains **no** Coldline/Nearline/Retrieval/Early-Delete SKU for this billing account (MEASURED, 90 d), so the cron filters on `%Coldline%`, `%Retrieval%`, `%Early Delete%` until dev pins the literal names.
- **One-time price sanity** — dev Coldline Class A ÷ objects transitioned ≈ **$0.020/1,000 (±25%)**. **> $0.03/1,000 → stop and reconcile before prod** (contract pricing differs).
- **No early-delete row** for the dev bucket (proves the prefix scoping did not touch a Delete-ruled prefix).
- **No new reader errors** — Cloud Logging, project `based-hardware-dev`, services `backend`/`backend-sync`/`pusher`, filter on `audio_merge cache_upload_failed`, `Failed to decode/decrypt`, `Error downloading audio file`, `conversation_artifact:`, `Chunk not found for timestamp`: count over the bake **≤** count over the preceding 48 h. Dev has ~0 reads, so this is 0 = 0; it is included so the same filter is proven before prod.

What dev does **not** prove: retrieval share (the dev bucket had **zero** `ReadObject` calls in 7 days). Do not wait for it — reader compatibility is a code property, not a traffic property, which is also why the prod partition canary was dropped.

**3. Prod go-live** (David dispatches; the `prod` environment reviewer is the second gate):

```bash
gh workflow run gcp_storage_lifecycle.yml --ref main \
  -f environment=prod \
  -f lifecycle_variant=apply \
  -f confirmation=APPLY_STORAGE_LIFECYCLE_PROD \
  -f release_sha=<exact merged main SHA>
```

**4. Prod watch** — daily Hermes cron (below): lifecycle drift, per-bucket billing SKUs, Coldline byte share, and the reader-error log filter. Alarms: Early Delete > $2/day, Retrieval > $10/day, cumulative Coldline Class A > $9,000, reader errors > 2× the 7-day pre-apply mean **and** > 20/day. Progress expectation: Coldline share ≥ 50% by T+3 d, ≥ 85% by T+7 d.

## Rollback

Dispatch `gcp_storage_lifecycle.yml` with `lifecycle_variant=rollback` for the environment. This re-applies the pre-change rules and stops further transitions within ≤24 h. **Objects already moved stay Coldline** — a lifecycle rule cannot move them back. Moving them back is a rewrite job (`gcloud storage objects update gs://omi-private-cloud-sync/chunks/** --storage-class=STANDARD`, or a rewrite loop), which costs Class A at the Standard rate (~$1.7k for 346 M objects) **and, unlike the lifecycle-driven forward transition, is a deletion of the Coldline copy: objects that have not yet reached 90 days total age are charged the remainder as early-delete.** There is no data-loss path in either direction.

```bash
gh workflow run gcp_storage_lifecycle.yml --ref main \
  -f environment=prod \
  -f lifecycle_variant=rollback \
  -f confirmation=APPLY_STORAGE_LIFECYCLE_PROD \
  -f release_sha=<exact merged main SHA>
```

## Hermes watch cron

`Omi PCS Coldline rollout watch` — job **`7ee554b17e07`**, daily `20 10 * * *` America/New_York, `no_agent`, `expect_output`, whitelisted `critical: true`, delivering to `telegram:-1004448164123:2232`, quiet-when-healthy via `{"wakeAgent": false}`. Script: `~/.hermes/scripts/omi-pcs-coldline-watch.sh`.

It is registered **paused** so it arms only when this PR merges and dev is live. On merge:

```bash
hermes cron resume 7ee554b17e07
```

## Verification

- `actionlint .github/workflows/gcp_storage_lifecycle.yml` → clean
- `pytest backend/tests/unit/test_storage_lifecycle_contract.py` → 12 passed
- Validator dry-runs: `prod.json`, `development.json` and both rollback files pass; a fixture that drops a legacy rule, a fixture with a `Delete` on `chunks/`, a fixture with `isLive: false`, and a non-empty notifications list each exit 1 with a specific error.
- `black --line-length 120 --skip-string-normalization` applied to the new script and test.

## Follow-ons deliberately NOT in this PR

Realtime writer PCM16 → Opus (the only lever that bends the +16 TB/month slope); single-region migration; Archive@365 (decide at T+14/T+30 d from the cron's measured old-read share); versioning-off + soft-delete-7 d (the validator currently *requires* versioning on, so that must be a deliberate edit).

- INV-DATA-1

This workflow matches the `.github/workflows/gcp_*.yml` glob but changes no routing
authority: it deploys no code and touches no serving endpoint, auth project, or
Firebase/Firestore binding. It reads and writes exactly one bucket property —
`lifecycle` — on the bucket named by the environment's own checked-in file, and
refuses to run when that file's `project` does not equal `vars.GCP_PROJECT_ID`, so
a development dispatch can never reach the production bucket and vice versa. The
`push` trigger resolves to `development` unconditionally; prod requires a dispatch,
a typed confirmation, an exact main SHA, and the `prod` environment reviewer.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12631?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/gcp_storage_lifecycle.yml
+++ b/.github/workflows/gcp_storage_lifecycle.yml
@@ -139,6 +139,15 @@ jobs:
           test -f "$file"
           bucket="$(python3 -c "import json,sys;print(json.load(open(sys.argv[1]))['bucket'])" "$file")"
           project="$(python3 -c "import json,sys;print(json.load(open(sys.argv[1]))['project'])" "$file")"
+          case "$RESOLVED_ENVIRONMENT" in
+            development) expected_bucket='omi-dev-private-cloud-sync' ;;
+            prod) expected_bucket='omi-private-cloud-sync' ;;
+            *) echo "Invalid environment: $RESOLVED_ENVIRONMENT."; exit 1 ;;
+          esac
+          if [[ "$bucket" != "$expected_bucket" ]]; then
+            echo "Lifecycle file bucket $bucket does not match the expected $RESOLVED_ENVIRONMENT bucket $expected_bucket."
+            exit 1
+          fi
           if [[ "$project" != "${{ vars.GCP_PROJECT_ID }}" ]]; then
             echo "Lifecycle file project $project does not match the environment project ${{ vars.GCP_PROJECT_ID }}."
             exit 1
@@ -160,7 +169,7 @@ jobs:
         run: |
           set -euo pipefail
           gcloud storage buckets describe "gs://${{ steps.select.outputs.bucket }}" \
-            --project="${{ steps.select.outputs.project }}" --format=json > before.json
+            --project="${{ steps.select.outputs.project }}" --raw --format=json > before.json
           gcloud storage buckets notifications list "gs://${{ steps.select.outputs.bucket }}" \
             --project="${{ steps.select.outputs.project }}" --format=json > notifications.json
           python3 -c "import json;print(json.dumps(json.load(open('before.json')).get('lifecycle', {}), indent=2))"
@@ -169,7 +178,13 @@ jobs:
         run: |
           set -euo pipefail
           extra=()
-          if [[ "$LIFECYCLE_VARIANT" == "rollback" ]]; then extra+=(--allow-rule-removal); fi
+          if [[ "$LIFECYCLE_VARIANT" == "rollback" ]]; then
+            # Scoped removal: only rules the apply variant declared (the Coldline
+            # rule) may be dropped; a later live rule still blocks the rollback.
+            apply_file="backend/deploy/storage-lifecycle/${RESOLVED_ENVIRONMENT}.json"
+            test -f "$apply_file"
+            extra+=(--allow-rule-removal-of "$apply_file")
+          fi
           python3 backend/scripts/validate_storage_lifecycle.py \
             --desired "${{ steps.select.outputs.file }}" \
             --live before.json \
@@ -188,7 +203,7 @@ jobs:
         run: |
           set -euo pipefail
           gcloud storage buckets describe "gs://${{ steps.select.outputs.bucket }}" \
-            --project="${{ steps.select.outputs.project }}" --format=json > after.json
+            --project="${{ steps.select.outputs.project }}" --raw --format=json > after.json
           python3 backend/scripts/validate_storage_lifecycle.py \
             --desired "${{ steps.select.outputs.file }}" --live after.json --expect-applied
 
--- a/.github/workflows/gcp_storage_lifecycle.yml
+++ b/.github/workflows/gcp_storage_lifecycle.yml
@@ -1,0 +1,210 @@
+name: Apply Cloud Storage lifecycle (private-cloud-sync)
+
+# The private-cloud-sync bucket's lifecycle was managed by nothing: console and
+# gsutil only. This workflow is the single writer. `gcloud storage buckets
+# update --lifecycle-file` REPLACES the whole lifecycle config, so the apply is
+# fenced by a describe -> superset validation -> apply -> describe -> assert
+# sequence, and every artifact is uploaded.
+#
+# push  -> development only. prod is dispatch-only, typed-confirmation gated,
+#          and additionally gated by the `prod` GitHub Environment reviewer.
+on:
+  push:
+    branches: [ "main" ]
+    paths:
+      - 'backend/deploy/storage-lifecycle/**'
+      - 'backend/scripts/validate_storage_lifecycle.py'
+      - '.github/workflows/gcp_storage_lifecycle.yml'
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: 'Bucket environment to reconcile'
+        required: true
+        default: 'development'
+        type: choice
+        options:
+          - development
+          - prod
+      lifecycle_variant:
+        description: 'apply = checked-in lifecycle; rollback = pre-change rules only'
+        required: true
+        default: 'apply'
+        type: choice
+        options:
+          - apply
+          - rollback
+      confirmation:
+        description: 'Type APPLY_STORAGE_LIFECYCLE_PROD for prod'
+        required: false
+        default: ''
+        type: string
+      release_sha:
+        description: 'Exact current main SHA (prod only)'
+        required: false
+        default: ''
+        type: string
+
+# Deliberately NOT the backend-stack lock: this workflow builds and deploys no
+# code, and must not queue behind or block a backend release.
+concurrency:
+  group: storage-lifecycle-${{ github.event_name == 'workflow_dispatch' && github.event.inputs.environment || 'development' }}
+  cancel-in-progress: false
+
+jobs:
+  apply:
+    if: >-
+      github.ref == 'refs/heads/main' &&
+      (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+    environment: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.environment || 'development' }}
+    permissions:
+      contents: read
+      actions: read
+    runs-on: ubuntu-latest
+    env:
+      RESOLVED_ENVIRONMENT: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.environment || 'development' }}
+      LIFECYCLE_VARIANT: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.lifecycle_variant || 'apply' }}
+    steps:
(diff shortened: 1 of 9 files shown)


Git-on-my-level:
Addressed all five Cubic review threads in 2c86c11680.

**Thread: bucket guard checks only project (`yml:140`)** — the select step now asserts the expected bucket per environment (`omi-dev-private-cloud-sync` for development, `omi-private-cloud-sync` for prod) before any `gcloud` command, in addition to the existing project check.

**Thread: describes use the wrong representation (`yml:163`)** — both `gcloud storage buckets describe` fixtures (before/after) now pass `--raw`, so the validator reads the API representation it is written against (`name`, `versioning.enabled`, verbatim rule equality) instead of gcloud's standardized one.

**Thread: rollback disables drop protection wholesale (`yml:172`)** — replaced the rollback blanket `--allow-rule-removal` with a scoped `--allow-rule-removal-of <apply document>`: only rules the apply variant declared (the Coldline rule) may be dropped; any later live rule still blocks. The blanket flag remains as an explicit operator override and is mutually exclusive with the scoped form.

**Thread: non-string `action.type` crashes the validator (`py:78`)** — `action_type` is now checked to be a string before set membership, so malformed checked-in JSON fails with a normal validation error.

**Thread: `matchesStorageClass` may be omitted (`py:121`)** — the Coldline rule must now carry exactly `[STANDARD, MULTI_REGIONAL]` (element types validated first, no `set()` on malformed values); omission or subsets are rejected.

Also updated `backend/deploy/storage-lifecycle/README.md` to document the scoped rollback semantics.

Verification: `python3 -m pytest tests/unit/test_storage_lifecycle_contract.py tests/unit/test_storage_lifecycle_workflow.py -q` → 18 passed (new regression tests per thread plus a structural workflow test pinning the bucket assert, `--raw` describes, and scoped rollback wiring); validator CLI exercised end-to-end (source-only on all four docs, scoped-removal rollback pass/block/blanket/mutex); `black --check` clean; `.github/scripts/check-deployment-concurrency.py` OK. CI on 2c86c11680: all checks pass. All five threads resolved.

Issue #12530, opened 2026-09-01
Daily summaries silently stop for tail-of-batch users: notifications-job 600s timeout + one-user context overflow aborts whole hour groups


Git-on-my-level:
## Symptom

The owner dogfood account (America/New_York, hour-22 default) has received no daily summary since **2026-08-23** (last selection logged 2026-08-24T02:54Z). The pipeline is alive for other users, so this is a silent partial outage: users late in a batch simply never get their summary, with no error attributable to them.

Evidence gathered 2026-09-01 (read-only prod log/Firestore inspection):

1. **600s task timeout kills the job mid-batch.** `Terminating task because it has reached the maximum timeout of 600 seconds` fires regularly (observed 08-20 through 09-01), including at ~02:39-02:40Z — inside the window that serves hour-22 America/New_York users. The affected account's last successful send was at 02:54Z, i.e. from a run that lived past 600s.
2. **Hour-22 batches are enormous and repeated.** `Sending daily summary to 4860 users at local hour 22` (03:01Z) and repeated `3717 users at local hour 22` lines minutes apart suggest overlapping executions re-attempting the same batch none of which can finish inside 600s.
3. **One user's summary aborts entire hour groups daily.** `send_daily_summary_notification` wraps the whole hour-group loop in one try/except; a single user's generation failing throws away everyone after them. This is live: `Error sending daily summary: Error code: 400 ... context_length_exceeded (input ~290k tokens)` has fired **every day at ~05:05Z since at least 2026-08-25** — the same account overflowing the 272k limit daily, killing the rest of that run's groups.

## Why it matters

- The failure mode is invisible: no per-user error, no alarm, and the summary just doesn't arrive. This matches the metrics-and-alarming theme of capabilities failing behind a 2xx.
- The daily summary is the spine of the JIT daily sweep's source material, so silent summary gaps degrade JIT memory formation for affected users.

## Suggested repairs (separable)

1. Per-user (or at least per-hour-group) exception isolation in `send_daily_summary_notification` — one user's 400 must not abort other users.
2. Bound the summary prompt input (truncate conversation material) so a heavy day cannot exceed the model context limit.
3. Either raise the job task timeout to fit the observed batch duration or checkpoint progress so a killed run resumes instead of restarting; dedupe overlapping executions (repeated identical batch lines).
4. A counter/alert for 'selected users minus completed sends' per run, so tail-loss is visible.

[...]

PR #12669, opened 2026-09-03
fix: make the review card, the voice guard, and the follow-up tail actually work (#12635 follow-up)

Git-on-my-level:
## Summary

Follow-up to #12635, which merged before this review round landed. Everything
here is a fix to what that PR shipped, found by adversarial review of it and
verified against the code path before being changed.

Three of these are defects that make shipped code wrong today on `main`:

1. **The learned-memory review card is dead in production.** The scheduled
   daily-summary job called the summary generator without `memories_learned`, so
   the field was always empty, so the `memoryReviewCard` block was never attached
   on the only path real users receive. Only the `/daily-summary-settings/test`
   route passed it. Both senders now select through one shared helper, and a test
   fails without the fix.
2. **A failed voice query now reads its own failure aloud.** #12635 changed the
   empty-response copy while `FloatingBarVoicePlaybackService.shouldSpeak` still
   suppressed one exact hardcoded string, which no longer exists in the tree. Copy
   and guard now share one source, pinned by a test that drives the real playback
   path rather than the string compare.
3. **The server rejects the review card the Mac sends.** The desktop adapter
   converts a `memoryReviewCard` for the chat-first journal, but the block was not
   in the server union, so every such request failed validation as
   `invalid_request` and the card was never journaled. Added to the *request*
   union only: the server never emits this block, and adding it to the
   materialization response union is a breaking change to a released app-client
   contract (`check_app_client_openapi_compatibility` confirms both directions).

Two more undercut claims #12635 itself made:

4. **The daily-summary checkpoint could not resume.** The cursor key carried the
   UTC hour, so the run that wrote the tail and the run that should have resumed it
   never shared a key — which defeats the point of the #12530 fix. A failed
   timezone chunk also let the run call itself complete and clear the cursor.
5. **The Mac review card never saw its own vote.** A vote wrote to the backend but
   was not mirrored locally, so the row reverted to unreviewed on the next card
   build. Now mirrored through existing storage, with the backend still the only
   writer.

And one the review only half-saw:

6. **The streamed follow-up tail leaked into the transcript.** `ChatStreamingBuffer`
   wrote the normalized projection back into the message and appended the next
   delta to it, so a projection that deliberately withholds the tail lost the
   evidence it needed on the next flush. A replay of every flush boundary leaks at
   every boundary from the delimiter onward, not only at a split marker, and the
   question always spans several flushes. Display-only, since the authoritative
   answer overwrites it, but the reader watched the chip's question get typed into
   the answer and then snatched away.

Also fixed: multi-question chips, chips on guard-stopped turns, generic-prefix
rules that diverged between client and server, a permanent optimistic override on
mobile, a duplicated question when a follow-up block had no prose, per-item
decoding so one malformed row cannot drop the whole array, owner-scoped knows-list
visits, a content-change reset that skipped the show cap, and greeting counts that
disagreed with the rows they described.

Three findings were **rejected with evidence** rather than fixed: that the stream
filter persists its withheld suffix (the persisted answer is built by the
producer, and flushing would have pushed the fragment *into* the stream); that a
safety-guard stop should be reclassified as a provider failure (an explicit,
tested contract says a guard stop is a deliberate reply); and a pytest marker that
would have removed a file from CI entirely.

## End-to-end coverage: made real

#12635 satisfied `desktop-e2e-flow-coverage` by adding `covers:` entries to two
flows that never exercise the code. The gate is a text match over the `covers:`
list and never inspects steps, so the claim cost nothing to write and gave false
confidence that these modules were end-to-end covered. Those entries are removed
here and the coverage is built instead:

- **Chat follow-up.** The hermetic stub can now emit a follow-up tail, the chat
  automation snapshot exposes the chip question, and a bridge action taps the chip
  and reports the origin the analytics emit actually carried. The flow asserts the
  visible answer by exact equality, which is what catches a tail left in the prose.
- **Memory review card** (`memory-review.yaml`). A local-only dev-harness route
  seeds a summary, the flow creates the memories through the production endpoint,
  renders the card, accepts one row and rejects another through the same store the
  buttons call, then re-reads to prove both verdicts survive the refresh.
- **Knows-list rotation** (`home-knows-rotation.yaml`). Seeds tasks, clears the
  ledger through the same persistence the mutations use, and pins the first visit's
  rows, the within-visit de-duplication, and a later visit's view of a dismissed row
  and a merely-shown row.
- **Streaming buffer.** Judged uncoverable at first, because the defect is
  mid-stream and the terminal answer overwrites the streamed text before any
  assertion runs. True of the final text only: a non-prod probe now keeps its own
  raw stream, compares every emitted projection against the shipped tail-stripping
  rule, and the flow asserts zero leaks alongside two liveness assertions so the
  zero cannot be vacuous. A negative-control unit test feeds the probe the exact
  projection the pre-fix buffer emitted and asserts it is counted.

Three things are deliberately **not** claimed, each stated next to its `covers:`
entry: the empty-slot telemetry, the three-show cap (it needs three calendar days
or an injected clock, and is unit-tested instead), and the 14-day stale-task rule,
which is unreachable from the hub because the dashboard query already cuts off at
seven days. That last one is a real finding and wants a follow-up.

## Invariants

- INV-AUTH-1: no change to auth.
- INV-CHAT-1: chat continuity preserved; the follow-up block rides the persisted message.
- INV-NAV-1: no navigation changes.
- INV-VOICE-1: realtime hub prompt untouched; the voice guard change makes failure copy silent again, as it was before #12635.
- INV-MEM-4: no new promotion path; a vote is a review, not a promotion.
- INV-MEM-1: memory reads go through `MemoryService`; the local mirror carries no second copy of review state and the backend remains the only writer.

- INV-AUTH-1
- INV-CHAT-1
- INV-VOICE-1
- INV-MEM-4
- INV-MEM-1
- INV-TASK-2

## Failure class

Failure-Class: new

This PR adds one definition, `FC-projection-fed-its-own-output`: `ChatStreamingBuffer` re-derived its output from
its own previous output, so a projection that withholds text destroyed the
evidence it needed on the next pass. The protocol allows one declaration per body;
the others are recorded here for the reader.

## Line-count exceptions

Line-Count-Exception: backend/utils/retrieval/agentic.py | 1718 -> 1736 | outcome-text helper voids a pending marker at the five backend-authored sites
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarView.swift | 3107 -> 3112 | chord-aware push-to-talk hint
Line-Count-Exception: desktop/macos/Desktop/Sources/FloatingControlBar/FloatingControlBarWindow.swift | 5740 -> 5752 | failure copy moved to a shared constant the voice guard also reads
Line-Count-Exception: desktop/macos/Desktop/Sources/MainWindow/Pages/DashboardPage.swift | 4268 -> 4328 | per-branch open recording plus the non-prod knows automation registry
Line-Count-Exception: desktop/macos/Desktop/Sources/Providers/ChatProvider.swift | 7210 -> 7212 | the streaming buffer's raw accumulator is released when the terminal answer lands

## Verification

- `make preflight`: all 39 checks pass, including `desktop-e2e-flow-coverage` at 0 uncovered.
- Backend: scoped runs through `test.sh` (via `BACKEND_UNIT_TEST_FILE_LIST`, since positional paths are ignored) green under the CPU duration guard, including four consecutive runs of the daily-summary files at load average 61. `check_app_client_openapi_compatibility` passes; all four generated clients are `--check` clean.
- Desktop: `swift build -c debug` green; the follow-up, review-card, knows-list, and streaming-probe suites pass; swift-format, SwiftLint, the baseline ratchet, test-quality, and the main-actor hook check are clean.
- Mobile: `app/test.sh` all passed; `dart format` clean.

## Not done here

- **None of the three new or extended e2e flows has been executed.** Tier 2 needs Docker, and `docker info` fails on the authoring machine. Both sides of every string they assert are pinned by unit tests so drift fails in seconds, and the streaming probe has a falsifiable negative control, but someone with Docker must run `chat-hermetic`, `memory-review`, and `home-knows-rotation` before they are trusted. Two knows-list expectations rest on a reading of the composer's slot ordering that only a real run can confirm.
- No dogfood on a signed-in bundle. The desktop mirror write on a review vote has no hermetic test, so it is the first thing to check by hand.
- A cross-device vote still appears only after the memory sync runs: there is no single-memory GET to make the card's read authoritative.
- An unserved daily-summary user is still not re-queried an hour later, because eligibility matches one local hour. The checkpoint only recovers the tail if the job runs more than once per hour, and the repo disagrees with itself about the cadence.
- Mobile strings stay hardcoded English, matching every neighbouring widget.
- Interject is still not wired to the review rows.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12669?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-projection-fed-its-own-output.json
+++ b/.github/failure-classes/FC-projection-fed-its-own-output.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": 1,
+  "id": "FC-projection-fed-its-own-output",
+  "violated_contract": "A projection that may withhold part of its input must be applied to the accumulated raw input on every pass. Storing the projection's output back into the only buffer and extending that buffer with the next delta destroys the evidence the projection needs, so anything withheld once is emitted on a later pass.",
+  "canonical_prevention": "Accumulate the raw input separately from the projected output: append deltas to the raw accumulator, project the whole accumulator, and write the result outward without ever reading it back. Release the accumulator when the stream terminates so a reused identity cannot resume from a previous one. Prove it by replaying the same input broken at every possible boundary and asserting the withheld content never appears in any intermediate output.",
+  "canonical_prevention_artifact": [
+    "desktop/macos/Desktop/Tests/ChatStreamingTailProjectionTests.swift"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    "desktop/macos/Desktop/Sources/Chat/ChatStreamingBuffer.swift",
+    "desktop/macos/Desktop/Sources/Chat/ChatFollowUpTail.swift",
+    "desktop/macos/Desktop/Sources/Providers/ChatProvider.swift"
+  ],
+  "status": "open"
+}
--- a/app/lib/backend/schema/message.dart
+++ b/app/lib/backend/schema/message.dart
@@ -545,10 +545,15 @@ class ServerMessage {
         return value('text').isEmpty ? 'Question' : value('text');
       case 'memoryReviewCard':
       case 'memory_review_card':
-        return 'Things I learned today';
       case 'followUp':
       case 'follow_up':
-        return value('text');
+        // Both render natively on mobile — MemoryReviewCard draws its own
+        // "Things I learned today" heading, and ChatFollowUpChip draws the
+        // question. Inventing the same words as fallback prose says them
+        // twice; and when the block is malformed enough that no card renders
+        // (MemoryReviewCardBlock.tryFromBlock returns null for an item-less
+        // block), a bare heading over nothing is worse than no heading.
+        return '';
       case 'taskCard':
       case 'task_card':
         return 'Task';
--- a/app/lib/widgets/components/memory_review_card.dart
+++ b/app/lib/widgets/components/memory_review_card.dart
@@ -58,11 +58,24 @@ class _MemoryReviewCardState extends State<MemoryReviewCard> {
   final Set<String> _failed = {};
   final Map<String, TextEditingController> _editors = {};
 
+  /// The text a persisted correction submitted, per row. A knowledge-ledger
+  /// correction appends a new row under a *new* id, so the id this card
+  /// references stops resolving in the provider; without this the row would
+  /// fall back to the original learned text under an "Updated." status.
+  final Map<String, String> _settledEdits = {};
+
   /// Ids this process has already asked the provider to go looking for. A card
   /// scrolled in and out of the chat list rebuilds its State, and an id that is
   /// genuinely absent from the account would otherwise re-trigger a full
   /// memories fetch on every rebuild.
   static final Set<String> _requestedIds = {};
+
+  /// Whether this process has already asked the provider to load its list.
+  /// Nothing in app start-up initialises [MemoriesProvider] — only the memories
+  /// page calls `init()` — and a provider that has never loaded still reports
+  /// `loading == true`, so "already loading" cannot be read as "a fetch is in
+  /// flight" until this card has started one itself.
+  static bool _loadRequested = false;
 
   /// Card identities already counted as shown in this process.
   static final Set<String> _seenImpressions = {};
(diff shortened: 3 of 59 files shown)


kodjima33 (approved):
Flagged author. Bug-fix bundle (5 distinct defects, each root-caused with code paths) following up #12635, but diff (4185 lines, 59 files, spans backend+desktop+app) fails the scoped-diff <=300 confidence-gate criterion. CI green, tests added, no abuse signals found. ~3/5 confidence -> approve-only, do not merge.

PR #12682, opened 2026-09-03
refactor(queues): the substrate owns every drain (follow-up to #12665)

Git-on-my-level:
## Why a follow-up

PR #12665 merged at its round-1 head while an adversarial review of that head was being acted on. The review found that the round-1 change was a helper library beside untouched per-queue loops: `drainIsolated` had zero production callers, three identity calls passed the same id twice, and the stalled-queue gauge could never fire because the series was zero-initialised and the rule treated no-data as OK. This PR carries the round-2 work that makes the six properties hold by construction, on top of current `main`:

- every production drain (both languages) now runs through `drain_isolated` / `drain_isolated_async` / `drainIsolated`, with the old local loops deleted, and a checked-in ratchet test that goes red if a listed module drops it;
- the oldest-ready gauge is sampled store-wide from one periodic publisher, is not zero-initialised (so `absent()` is meaningful), and `omi-queue-oldest-ready` alerts on no-data;
- the vacuous `adoptOnIdentity` / `adopt_on_identity` calls are gone and the deferral identity adopt is tested with a different payload;
- candidate integration refuses to claim a dead letter, dead-letters malformed rows, does not ACK a failed completion, and honours a future `available_at`;
- daily-summary failures are typed non-success outcomes, including an hour whose every chunk read failed;
- pure policy moved to `utils/durable_queue_policy.py` so the async pusher helper uses `decide_attempt` without tripping the async-blocker gate.

`app/` is byte-identical to `main`; this PR owns nothing there.

## Summary

- Every production drain in the queue inventory now goes through `drain_isolated` / `drain_isolated_async` / `drainIsolated`. Policy primitives live in `backend/utils/durable_queue_policy.py` so async callers can import them without the async-blocker treating `database.*` as I/O; `backend/database/durable_queue.py` keeps the Firestore-touching redrive patch. Store-wide oldest-ready sampling lives in `backend/database/durable_queue_age.py` and is published from the backend's periodic reconcile tick, never from a request path.
- Vacuous substrate calls are gone. Deferral enqueue with the same `continuity_key` and a different payload is adopted (first payload kept, no throw). Chat-first fetch order uses `ready_sort_key` via `database/chat_first_intent_queue.py`. Candidate `claim` refuses `dead_letter`; a malformed row becomes `dead_letter` with error text; drain does not ACK `succeeded=False`; future `available_at` is not dispatched.
- `omi_queue_oldest_ready_age_seconds` is not zero-initialized at import, so `absent()` means the publisher has not run. Grafana `omi-queue-oldest-ready` uses `noDataState: Alerting`. A checked-in contract test fails if a listed drain module drops the substrate or if `OMI_QUEUE_NAMES` and the sampler set diverge.
- Out-of-scope Dart reformatting from a merge hook is reverted; `git diff --stat origin/main...HEAD -- app` is empty.

This PR is the first evidence PR for FC-batch-fault-domain-head-of-line-poison. The registry JSON is intentionally not edited here so the maintainer can add this PR number to `evidence_prs`.

Failure-Class: FC-batch-fault-domain-head-of-line-poison

- INV-AGENT-*
- INV-CHAT-1
- INV-MEM-4
- INV-MEM-1
- INV-MEM-2
- INV-TASK-2

## Round-2 acceptance bar

1. **Substrate owns every production drain loop.** `drain_isolated` / `drain_isolated_async` / `drainIsolated` wrap the inventory drains (memory outbox, vector repair, candidate integration, chat-first fetch/materialize, daily-sweep process + cleanup, daily-summary hours, recurrence inbox, frame deletion, projection repairs, kernel outbox pump + journal drains). `conversation_finalization_jobs` keeps Cloud Tasks; `decide_attempt` lives in `utils/durable_queue_policy.py`. Evidence: `PRODUCTION_DRAIN_MODULES` in `test_durable_queue_substrate_contract.py`; mutation 2 below.
2. **Ratchet over a checked-in module list.** `test_listed_production_drains_call_the_substrate` plus `test_omi_queue_names_match_age_samplers`. Replacing `drain_isolated(leases, process_one)` with a local `for` in `memory_outbox_worker.py` goes red (`missing drain_isolated(`).
3. **Gauge durable and absent-not-zero.** Store-wide sample in `durable_queue_age.py`, published from the periodic tick. No `OMI_QUEUE_OLDEST_READY_AGE_SECONDS.labels` in `metrics.py`. `test_omi_queue_family_is_not_zero_initialised` and `test_omi_queue_family_is_not_labels_touched_at_import` fail if the family is labels()-touched at import. Grafana `omi-queue-oldest-ready` has `noDataState: Alerting`; flipping it to `OK` reddens `test_durable_queue_oldest_ready_alert_pages_on_age_or_absent_gauge`.
4. **No vacuous substrate calls.** Same-id `adoptOnIdentity` / `adopt_on_identity` production calls are gone. Deferral identity is `if (existing) return`. Different-payload adopt test keeps the first question and goes red if the hash throw is restored. `ready_sort_key` is used by `sort_ready_intents`. Daily-summary no longer observes `0.0`.
5. **Candidate integration end to end.** `claim` refuses `dead_letter`; malformed row → `dead_letter` with error text; drain does not ACK `succeeded=False`; deleting the Python `available_at` filter reddens `test_future_available_at_is_not_dispatched` (`['future', 'ready'] == ['ready']`).
6. **Failures are not laundered.** `send_daily_summary_notification` returns `DailySummaryCronOutcome(ok=False)` when any hour rejects, including per-user send failures and an hour whose every timezone-chunk read failed (`test_daily_summary_all_chunks_failed_is_a_typed_hour_reject`; dropping the `query_error and not users` reject makes `assert True is False`). Chat-first materialize keeps HTTP 200; an all-hard-reject batch increments `materialize_batch_rejected`.
7. **Mutation proofs at production call sites.** Table below. Every review mutation that stayed green now goes red, or the vacuous call is gone.
8. **Forward deploy safe; rollback hazards listed.** Additive Firestore fields only. No new composite index. `available_at` filtered in Python.

## What stayed, and why

- `conversation_finalization_jobs` keeps its Firestore document shape, BYOK `blocked_byok`, Cloud Tasks attempt fence, and `reconcile_after_at`. The attempt decision in the async pusher helper comes from `decide_attempt` in `utils/durable_queue_policy.py` (not `database.*`). Age is sampled store-wide on the periodic tick from `get_finalization_job_summary`.
- `chat_first_intents.py` is not grown. Fetch isolation already landed in PR #12556; this PR wires `ready_sort_key` through `database/chat_first_intent_queue.py`. Materialize keeps HTTP 200 for the client contract; a batch whose every intent is a hard rejection increments `chat_first_proactive_total` `materialize_batch_rejected`.
- `frame_deletion_outbox` still retries storage deletes without a hard stop (user objects must keep converging). Drain is `drain_isolated`; failures record `last_error_text`.
- `task_recurrence_inbox` drain uses `drain_isolated` in `workstream_association.py`; receipts already adopt by identity. Retry stays pending rather than a new status enum.
- Daily-summary hour groups are ephemeral (no durable row). Isolation is `drain_isolated_async`; `send_daily_summary_notification` returns a typed non-success outcome when any hour rejects, including per-user send failures and all-chunks-failed reads.
- Daily-sweep cleanup **and** the #12533 per-UID process loop drain through `drain_sweep_uids` in `daily_memory_sweep_queue.py`. `daily_memory_sweep.py` stays at origin's 5048 lines.
- Integration outbox claim/complete/list/redrive live in `database/candidate_integration_outbox.py` so `candidates.py` stays under the 1,500-line ratchet. Enqueue on accept still writes the outbox row from `resolve_task_candidate`.
- Kernel SQLite outboxes publish oldest-ready age through `getJournalObservability`'s store-wide `MIN(created_at_ms)` union, not `omi_queue_*`. Pump send loops drain through `drainIsolated` in `journal-outbox-pump.ts`.
- Swift app-local queues stay Swift; the kernel owns the SQLite outboxes.
- Side commit `fix(backend): skip local OpenAPI venv when staging runtime image sources` stays as the pre-push `ENOSPC` hatch. `.openapi-venv` is a machine-local toolchain, not image source.

## Red-proof

Mutations were applied to **migrated production code**, then restored (not left in the tree).

| Queue / construction | Mutation on production code | Test that went red |
|---|---|---|
| Substrate ratchet (`memory_outbox`) | Replace `drain_isolated(leases, process_one)` with `for lease in leases: process_one(lease)` | `test_listed_production_drains_call_the_substrate` — `missing drain_isolated(` |
| `candidate_integration_outbox` | Delete the Python `available_at > now` skip | `test_future_available_at_is_not_dispatched` — `['future', 'ready'] == ['ready']` |
| `chat_first_deferral_outbox` | Restore `throw new Error("deferral payload hash mismatch")` on a reused continuity key | `adopts a reused continuity key with a different payload and keeps the first question` |
| daily-summary all-chunks-failed | Delete `if query_error and not users: reject` | `test_daily_summary_all_chunks_failed_is_a_typed_hour_reject` — `assert True is False` |
| daily-summary hour groups | `drain_isolated_async` `except: raise` | `test_daily_summary_poison_hour_does_not_block_later_hours` |
| `backend_turn_outbox` hash quarantine | Hardcode UPDATE status to `"retrying"` instead of `decision.terminal ? "failed" : "retrying"` | `quarantines a canonical-hash-mismatched outbox row instead of wedging the pump` |
| Kernel pump | `drainIsolated` `catch { throw }` | `one throwing send does not skip later independent deliveries` |
| Gauge zero-init | Import-time `OMI_QUEUE_OLDEST_READY_AGE_SECONDS.labels(queue=...)` for every name | `test_omi_queue_family_is_not_labels_touched_at_import` and `test_omi_queue_family_is_not_zero_initialised` |
| Grafana scrape hole | `noDataState: OK` on `omi-queue-oldest-ready` | `test_durable_queue_oldest_ready_alert_pages_on_age_or_absent_gauge` (`'OK' == 'Alerting'`) |

Captured outputs (this round):

```
FAILED tests/unit/test_other_notifications_async_boundaries.py::test_daily_summary_all_chunks_failed_is_a_typed_hour_reject
assert True is False
```

```
FAILED tests/unit/test_durable_queue_substrate_contract.py::test_listed_production_drains_call_the_substrate
AssertionError: production drain modules lost the substrate: backend/database/memory_outbox_worker.py (missing drain_isolated()
```

```
FAILED tests/unit/test_durable_queue_substrate_contract.py::test_omi_queue_family_is_not_labels_touched_at_import
FAILED tests/unit/test_journey_observability.py::test_omi_queue_family_is_not_zero_initialised
```

```
FAILED tests/unit/test_monitoring_alert_rule_contract.py::test_durable_queue_oldest_ready_alert_pages_on_age_or_absent_gauge
AssertionError: assert 'OK' == 'Alerting'
```

```
FAILED tests/unit/test_candidate_lifecycle.py::test_future_available_at_is_not_dispatched
AssertionError: assert ['future', 'ready'] == ['ready']
```

```
FAIL  tests/conversation-journal.test.ts > adopts a reused continuity key with a different payload and keeps the first question
AssertionError: expected [Function] to not throw an error but 'Error: deferral payload hash mismatch' was thrown
```

```
FAIL  tests/journal-outbox-pump.test.ts > one throwing send does not skip later independent deliveries
Error: send failed
```

```
FAIL  tests/conversation-journal.test.ts > quarantines a canonical-hash-mismatched outbox row instead of wedging the pump
AssertionError: expected { status: 'retrying', …(1) } to match object { status: 'failed', …(1) }
```

## Rollback hazards

- Additive fields (`last_error_text`, `dead_letter_reason`, `available_at` on candidate integration) are ignored by old readers. Rolling back workers leaves `status=dead_letter` items unclaimed by the old list query (`pending|failed|processing`) until they are redriven or patched.
- Kernel first-retry backoff is `2^(attempt-1)` via the shared helper (1s then 2s…) instead of the previous `2^attempt` (2s then 4s…). Rollback restores the slower first retry. Per-device SQLite, so mixed kernel versions do not share a table.
- `omi_queue_oldest_ready_age_seconds` is **not** zero-initialized. Until the periodic publisher has run after deploy, `absent()` is true and Grafana `omi-queue-oldest-ready` has `noDataState: Alerting` with `for: 5m`, so a scrape hole or a publisher that has not yet ticked pages. This is intentional: `absent()` must mean something.
- Mixed deploy of `available_at`: old readers ignore the field and retry immediately (safe extra retries). New readers + old writers without the field treat the row as ready (safe). No new Firestore composite index; `available_at` is filtered in Python after the existing `account_generation` + `status in` query.
- Deferral payload-hash throw is gone. A reused continuity key with different content now keeps the first payload. Rollback restores the throw. Per-device SQLite, no mixed readers of one table.
- Local image-closure staging ignores `backend/.openapi-venv` the same way it already ignores `.venv`. That directory is a machine-local OpenAPI toolchain; copying it filled `/Volumes/scratch` during pre-push (`ENOSPC`).

## Follow-ups

- Pusher finalization discard-as-success (named sibling in the class JSON) is not migrated in this PR.
- `chat_first_cold_start_sequence_receipts` is not a drain queue and is unchanged.
- `frame_deletion_outbox` remains unbounded by product exception; `task_recurrence_inbox` has no dead letter (retry stays pending).
- A composite index on candidate `status` + `available_at` is not added; the Python skip is the load-bearing bound. Promoting that filter into Firestore would need a registry entry and is a separate change.
- Failure-class registry `evidence_prs` is not edited here.

## Test plan

- [x] Focused backend units: `test_durable_queue_substrate_contract`, `test_other_notifications_async_boundaries` (including all-chunks-failed), `test_candidate_lifecycle::test_future_available_at_is_not_dispatched`, `test_journey_observability::test_omi_queue_family_is_not_zero_initialised`, `test_monitoring_alert_rule_contract::test_durable_queue_oldest_ready_alert_pages_on_age_or_absent_gauge`
- [x] `npx vitest run tests/durable-queue.test.ts tests/journal-outbox-pump.test.ts tests/conversation-journal.test.ts` → 79 passed
- [x] Red-proof mutations above; production files restored
- [x] `git diff --stat origin/main...HEAD -- app` empty (Dart reformat dropped)
- [x] `.venv/bin/python -m pyright` on changed production modules → 0 errors (430 pre-existing warnings)
- [x] File-isolated `test.sh` on the 11 owned unit files (including `test_other_notifications_async_boundaries`, `test_durable_queue_substrate_contract`, `test_candidate_lifecycle`) — all passed. `run-unit-ci.sh --changed-files` against the full PR diff selects 1035 files because `backend/database/*.py` requires the full suite; that run stalled on `test_mentor_notifications.py` and `test_daily_memory_sweep.py` with ESTABLISHED HTTPS to `*.1e100.net`. The same two files on an `origin/main` archive also hit `timeout` exit 124 after their assertions printed pass (teardown hang). Not introduced here.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12682?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kodjima33 (approved):
Flagged author. Explicitly titled refactor(queues), not a bug fix. Backend-focused durable-queue substrate consolidation, CI green, no abuse signals. Refactor -> approve-only per confidence gate.

PR #12687, opened 2026-09-03
fix(desktop): stop the notification-settings query from killing a non-app host

### LINKED ISSUE #9029: Swift test suite: shared-singleton contamination causes scheduling-dependent xctest worker crash
## Summary

Running the full macOS desktop Swift suite in a single `swift test` process, one suite reliably hard-crashes its `xctest` worker with **no assertion failure** (`… started` with no `passed`/`failed`). This is a scheduling-dependent crash caused by shared process-global state leaking across suites, not a bug in any one test.

## Repro

```
cd desktop/macos
xcrun swift test --package-path Desktop      # single combined process
```

- The **victim suite moves with the skip set** — e.g. `MemoriesViewModelObserverTests`, then `MemoryAuthoritativeTierSyncTests` — which is the signature of cross-suite state leakage rather than a single bad test.
- `--no-parallel` (serial) does **not** fix it → it is an order-dependent leak, not a data race.
- **Every suite passes in isolation** (per-suite `swift test --filter <Suite>/`, 114/114 green).

## Root cause

Many suites mutate process-global singletons and real on-disk state that persists across suites in the same process:
`RewindDatabase.shared`, `MemoryStorage.shared`, `AuthState.shared`, `StagedTaskStorage.shared`, an on-disk SQLite file, and `UserDefaults.standard`.

## Current mitigation

`desktop/macos/test.sh` was rewritten to run **each suite in its own process** (per-suite `swift test --filter`), mirroring the backend's per-file pytest isolation. This makes the gate deterministic and green. It is a **mitigation, not a fix** — the underlying singleton coupling remains and still bites anyone running `swift test` directly.

## Suggested durable fix

Introduce dependency-injection / test seams for the shared singletons so suites don't share process-global state (relates to the broader singleton-count debt in the desktop app).

_Found during macOS desktop test-suite hardening._

aryanorastar:
Fixes the process-terminating crash in #9029.

## Reproduced first

`swift test --package-path Desktop` (single combined process) on current `main`:

```
started=2652  completed=2650  failed=1
libc++abi: terminating due to uncaught exception of type NSException
```

The victim moved between my two runs of the same tree (`ChatTranscriptGestureHarnessTests`, then `HomeDailySummaryTests`) — the ordering signature the issue describes.

## Root cause

One call, `ProactiveAssistantsPlugin+NotificationSettings.swift:227`:

```
bundleProxyForCurrentProcess is nil: mainBundle.bundleURL file:///Applications/Xcode.app/Contents/Developer/usr/bin/
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException'
```

`UNUserNotificationCenter.current()` does **not** fail softly off an app bundle. It raises an Objective-C exception, and Swift cannot catch one, so it terminates the whole process. Under `swift test` the main bundle is the xctest tool, so any suite that reaches the production settings query kills the run.

That explains all three symptoms the issue lists:

- **no assertion failure** — it is a process termination, not a test failure
- **victim moves with ordering** — the crash lands on whichever suite is executing, not the one that reached the query
- **every suite passes in isolation** — per-suite processes contain the blast radius, which is exactly what the `test.sh` mitigation does

## The fix

The injection seam already existed — `notificationSettings(query:handler:)` takes the query — and only the production default was unsafe. It now answers `.notDetermined` when the host is not an app bundle, the same answer a fresh install gives, and never constructs `UNUserNotificationCenter`.

No test suite had to change, because the guard sits where every caller routes through.

## Verification

Full combined run, same command, before and after:

| | crash lines | tests completed |
|---|---|---|
| before | 1 | 2650 (run died) |
| after | **0** | **3023** |

`NotificationSettingsHostBundleTests` pins both halves **in the real xctest host** rather than a simulated one: the host is not an app bundle, and the production query answers rather than terminating. Before this change the second test took the process down instead of failing — that is the regression it guards.

Also verified with `-strict-concurrency=complete`, not just a debug build.

## Scope — what this does not do

- **Does not close #9029 by itself.** The issue also asks for DI seams across `RewindDatabase.shared`, `MemoryStorage.shared`, `AuthState.shared`, `StagedTaskStorage.shared` and `UserDefaults.standard`. This removes the one defect that terminates the process; ordinary shared-state coupling remains.
- **Does not remove the per-suite mitigation** in `test.sh`. That is a separate call once the combined run has been trusted for a while, and it should be its own PR.
- **`ChatDiscoverabilityTests.testDesktopCapabilitiesExistInAgentToolDeclarations` still fails**, before and after, and in isolation on `main`. Pre-existing and unrelated; left alone rather than folded in.

Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin+NotificationSettings.swift
+++ b/desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin+NotificationSettings.swift
@@ -11,6 +11,15 @@ struct UserNotificationSettingsSnapshot: Sendable {
   let soundSetting: UNNotificationSetting
   let badgeSetting: UNNotificationSetting
 
+  /// The answer for a host process that cannot own notification authorization at all,
+  /// such as the xctest tool. Distinct from a real `.notDetermined` only in origin.
+  static let hostProcessUnavailable = UserNotificationSettingsSnapshot(
+    authorizationStatus: .notDetermined,
+    alertStyle: .none,
+    soundSetting: .notSupported,
+    badgeSetting: .notSupported
+  )
+
   init(_ settings: UNNotificationSettings) {
     authorizationStatus = settings.authorizationStatus
     alertStyle = settings.alertStyle
@@ -221,9 +230,30 @@ enum UserNotificationCallbackBridge {
     return true
   }
 
-  nonisolated private static func systemNotificationSettingsQuery(
+  /// True when the running process is an app bundle, which `UNUserNotificationCenter`
+  /// requires.
+  ///
+  /// `UNUserNotificationCenter.current()` does not fail softly off an app bundle: it raises
+  /// `NSInternalInconsistencyException` ("bundleProxyForCurrentProcess is nil"), and an
+  /// Objective-C exception cannot be caught in Swift, so it terminates the whole process.
+  /// Under `swift test` the main bundle is the xctest tool
+  /// (`…/Xcode.app/Contents/Developer/usr/bin/`), so any suite that reaches this query kills
+  /// the run — see #9029, where the victim suite moved with test ordering because the crash
+  /// lands on whichever suite is executing, not on the one that triggered it.
+  nonisolated static var hostProcessIsAppBundle: Bool {
+    Bundle.main.bundleURL.pathExtension == "app"
+  }
+
+  nonisolated static func systemNotificationSettingsQuery(
     completion: @escaping @Sendable (UserNotificationSettingsSnapshot) -> Void
   ) {
+    guard hostProcessIsAppBundle else {
+      // No app bundle means no notification authorization to report. Answering
+      // `.notDetermined` keeps callers on their unauthorized path, which is the same
+      // answer a fresh install gives.
+      completion(.hostProcessUnavailable)
+      return
+    }
     UNUserNotificationCenter.current().getNotificationSettings { settings in
       completion(UserNotificationSettingsSnapshot(settings))
     }
--- a/desktop/macos/Desktop/Tests/NotificationSettingsHostBundleTests.swift
+++ b/desktop/macos/Desktop/Tests/NotificationSettingsHostBundleTests.swift
@@ -1,0 +1,52 @@
+import UserNotifications
+import XCTest
+
+@testable import Omi_Computer
+
+/// #9029: the desktop suite hard-crashed in a single `swift test` process with no assertion
+/// failure, and the victim suite moved with test ordering.
+///
+/// The cause was `UNUserNotificationCenter.current()` in the production notification-settings
+/// query. Off an app bundle it raises `NSInternalInconsistencyException`
+/// ("bundleProxyForCurrentProcess is nil"), and an Objective-C exception cannot be caught in
+/// Swift — so it terminated the whole run, landing on whichever suite happened to be
+/// executing rather than the one that reached the query.
+///
+/// These run in the xctest host, whose main bundle is the Xcode command-line tool, so they
+/// exercise the real non-app-bundle condition rather than a simulated one. Before the guard,
+/// the second test terminated the process instead of failing.
+final class NotificationSettingsHostBundleTests: XCTestCase {
+  func testTestHostIsNotAnAppBundle() {
+    XCTAssertFalse(
+      UserNotificationCallbackBridge.hostProcessIsAppBundle,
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
Thanks @aryanorastar — clean, well-scoped fix with an unusually good write-up (repro, root cause, before/after verification, and an honest scope section).

Verified against the head:

**`desktop/macos/Desktop/Sources/ProactiveAssistants/ProactiveAssistantsPlugin+NotificationSettings.swift`**
- The guard sits exactly where every production caller routes: `notificationSettings(handler:)` defaults to `systemNotificationSettingsQuery`, and `authorizationStatus(handler:)` plus the signed-smoke probe flow through it too — one guard covers the whole surface without touching any suite.
- `hostProcessUnavailable` answering `.notDetermined` is the right semantic: callers keep taking their existing unauthorized path, and the synchronous completion still flows through `dispatchToMain`, so the MainActor handoff contract is preserved.
- `Bundle.main.bundleURL.pathExtension == "app"` is a cheap, correct predicate; the shipped app always satisfies it, so user-facing behavior is unchanged — matching the changelog fragment's claim.
- Dropping `private` from `systemNotificationSettingsQuery` so the host-bundle test can exercise the real function is a minimal, reasonable visibility change.

**`desktop/macos/Desktop/Tests/NotificationSettingsHostBundleTests.swift`**
- Good choice pinning the condition in the real xctest host rather than simulating it — and the precondition test fails loudly if the host ever becomes an app bundle, which would otherwise silently un-exercise the guard.
- `SnapshotBox` (NSLock-guarded `@unchecked Sendable`) is the right shape for capturing a callback that answers on an arbitrary queue; the target wiring is correct (`Desktop/Tests` → `Omi ComputerTests`, `@testable import Omi_Computer`).

**`desktop/macos/changelog/unreleased/20260903-notification-settings-host-guard.json`**
- `{"kind": "none"}` matches the documented pattern for internal-only production edits; the extra `reason` field is useful for humans and harmless to the checker (Hygiene is green).

Two non-blocking observations for the record:

1. Other production paths still call `UNUserNotificationCenter.current()` directly — `requestAuthorization` and `add` in this same file, plus call sites in `AppState.swift`, `AppState+Permissions.swift`, `DeviceProvider.swift`, `OmiApp.swift`, `NotificationService.swift`, `NotificationDeliveryTelemetry.swift`, and `SBOnboardingModel+Steps.swift`. The #9029 crash traced to the settings query, but if any of those other paths is ever reached from a non-app host, the same uncatchable termination applies. A follow-up routing them through the bridge (or the same guard) would retire the hazard class, not just this instance.
2. Agree with the framing that #9029 stays open — the shared-singleton coupling it describes is untouched, and unwinding the per-suite `test.sh` mitigation should indeed be its own PR once the combined run has been trusted for a while.

Release compile is green; the Desktop Swift Static & Test Contracts run is still in flight. Leaving merge sign-off to a maintainer once that lands.

---

aryanorastar:
Follow-up measurement, now that #12705 is open alongside this.

With **both** this crash guard and #12705's stale-pin fix applied, the full combined single-process run is green:

```
swift test --package-path Desktop
started=3065  completed=3065  skipped=1  failed=0  crash lines=0
```

Progression on the same command:

| tree | crash lines | result |
|---|---|---|
| `main` | 1 | died at 2650 tests |
| + this PR | 0 | 3023 completed, 1 failure (the #12705 pin) |
| + both | **0** | **3065/3065, 0 failures** |

Two things that should not be read into it. The green run needs both PRs — this one alone still leaves the `ChatDiscoverabilityTests` failure. And it does not mean the singleton coupling in #9029 is resolved: `RewindDatabase.shared`, `MemoryStorage.shared`, `AuthState.shared` and friends are still process-global and still shared across suites. Nothing currently mutates them in a damaging order; that is a weaker statement than "isolated".

So the per-suite mitigation in `test.sh` should stay for now, and removing it belongs in its own PR once both of these have been on `main` for a while.

kodjima33 (approved):
Root-caused process-killing NSException from UNUserNotificationCenter under xctest (non-app host), fixes with a targeted guard + regression test pinning both halves. Confirmed not already fixed on main. Owner-override desktop bug fix, no abuse signals — merging.

PR #12693, opened 2026-09-03
feat: make push campaigns measurable end to end

### LINKED ISSUE #12645: Make push campaigns measurable: campaign id on the tap, iOS push-open event, download attribution
## What happened

On 2026-09-02 we pushed ~31,000 mobile users to try the macOS app (first FCM send 06:50:52 UTC, bursts until ~18:15 UTC, via `POST /v1/notification` → `send_notification` in `backend/utils/notifications.py:239`). Measuring it afterwards took four agents and still could not answer "how many people tapped it":

- **No campaign id anywhere.** The FCM `data` dict is passed through untouched (`backend/routers/notifications.py:94–104`), and the mobile tap handlers (`app/lib/services/notifications.dart:57` `onActionReceivedMethodImpl`, and `handleNavigateTo` for FCM background taps) only read `navigate_to`. Nothing is tracked on tap.
- **iOS emits no push-open event.** PostHog `$push_notification_opened` is Android autocapture only (31 users on send day vs 11/day baseline; iOS is ~80% of mobile and is invisible). The only tracked open is `Daily Summary Notification Opened` in `app/lib/utils/analytics/analytics_manager.dart:1915`, which is summary-specific.
- **The download link is unattributable.** `omi.me` / `h.omi.me` send nothing to PostHog project 302298 (only `www.omi.computer` is instrumented). The API redirect `GET /v2/desktop/download/latest` is countable in Cloud Logging but showed no change (21–34/hr all day), so either the CTA bypassed it or clicks were low — we cannot tell which.

What we could see: mobile users dormant 30+ days came back at 7.9x the normal share in the send window, and new macOS users ran ~2x a baseline of 17 per 12 hours. Send-to-desktop-launch was on the order of 0.05%, and every step between send and launch was dark.

## Fix (one PR, app + backend)

1. **Track the tap, both platforms.** In `notifications.dart`, at the single point where a tap payload is handled (`_handleAppLinkOrDeepLink` and the FCM `handleNavigateTo` entry), emit one `Notification Opened` event with `campaign_id`, `navigate_to`, and `notification_type` from the `data` map. Reuse `AnalyticsManager.track`. This makes iOS opens visible for the first time.
2. **Convention for the sender.** `/v1/notification` accepts `data.campaign_id` (it already accepts arbitrary `data`; document it in the route docstring and reject a campaign send with no `campaign_id` when `data.kind == "campaign"`).
3. **Attribute the download.** Support `navigate_to` values that are external URLs, and point campaign CTAs at `api.omi.me/v2/desktop/download/latest?campaign_id=<id>` (`backend/routers/updates.py`), logging `campaign_id` on the redirect so downloads are countable per campaign in Cloud Logging without depending on the marketing site's analytics.

### Acceptance

- [ ] A tap on a notification whose `data` carries `campaign_id` produces exactly one `Notification Opened` event with that id on iOS and Android. Proof: unit test alongside the existing tests for `notifications.dart` under `app/test` (the payload → route parsing already has tests there); run in the app CI lane.
- [ ] `GET /v2/desktop/download/latest?campaign_id=x` redirects as before and the request log carries `campaign_id`. Proof: backend unit test for the route in the existing backend test suite.
- [ ] Verified live on the next campaign: opens by platform and downloads by campaign are both queryable (PostHog HogQL and Cloud Logging respectively).

## Cut list

- Web analytics on `omi.me` — cut from this PR; the site is not in this repo and the download redirect gives attribution without it. Revival bar: a decision on where the marketing site's analytics live.
- A campaign dashboard — cut, dead check. Two queries in the next campaign's write-up are the surface.
- Per-timezone send scheduling — real (the first burst landed at 02:50 EDT and converted only in India/Germany waking hours) but a separate change to the sender, not measurement.

SakashSrivastava:
## What this does

Closes #12645.

The 2026-09-02 macOS push reached ~31,000 users and afterwards nobody could say how many tapped it. This gives the send, the open and the download one shared id.

**1. The tap, both platforms** — `app/lib/services/notifications.dart`

Both entry points already funnel through `_handleAppLinkOrDeepLink` (the Awesome Notifications action path, and the FCM background/terminated path), so `Notification Opened` is emitted there and is **exactly one per tap by construction**. It fires before the `navigate_to` check, so a push carrying no route is still counted.

The FCM path forwarded only the route, which dropped `campaign_id` before analytics could see it. That entry now takes the whole FCM data map (`handleFcmTap`); routing still ignores every key but `navigate_to`. The route check reuses `navigateToFromFcmData` instead of repeating the missing/empty/non-String rule inline, so its existing tests still cover it.

`navigate_to` may now be an **absolute URL**. A campaign CTA points at the download route carrying the campaign id, and without this the tap pushed that URL onto the navigator as a route name and landed on a dead route — so the download could never be attributed back to the send. External targets launch before the navigator wait, since they need no navigator and waiting would stall the launch up to 15s on a cold start.

`externalTargetFor` admits only `http` and `https`. A notification payload is attacker-influenced input, and handing an arbitrary scheme to the platform launcher would let one address anything the OS can open — `file:`, `javascript:`, `mailto:`, `tel:` and custom schemes are each refused, with a test apiece.

`notificationOpenedProperties` and `externalTargetFor` are pure functions, so the payload → properties mapping and the external/in-app decision are testable without a PostHog double or a platform channel. Absent keys are omitted rather than sent as null, so a query filtering on `campaign_id` sees only real campaign traffic.

**2. The sender convention** — `POST /v1/notification`

Documents what the `data` payload already allowed. `data.kind == "campaign"` marks a growth send and now requires `data.campaign_id`; the payload is still forwarded to FCM untouched. Rejecting is the point: afterwards an untracked campaign is indistinguishable from a tracked one, and the gap only surfaces once the opens can no longer be recovered — which is what happened on 09-02.

**3. The download** — `GET /v2/desktop/download/latest`

Accepts `campaign_id` and records it on a `desktop_download_served` log line, so per-campaign downloads are a Cloud Logging filter and don't depend on the marketing site's analytics.

- Attribution only. It never changes which installer is resolved, and a test asserts the served page is byte-identical with and without it.
- The pattern bounds it to log-safe characters: the id lands in a space-separated line, so a value carrying a space or newline could forge a field or a whole record.
- Emitted for every served page, `campaign_id=none` included, so a campaign's count keeps a denominator.

- INV-BETA-1

`backend/routers/updates.py` is on the invariant's path glob. The change is compatible: `campaign_id` is attribution only and beta-identity resolution is untouched — `_resolve_beta_identity_dmg`, the `identity` handling, and the rule that a beta-identity request never falls back to the stable `omi.dmg` are all unchanged. The only edit inside that branch is a log call placed before the existing `return`. `test_campaign_id_does_not_change_which_installer_is_served` pins that the served page is identical either way.

## Verification

Windows. Backend on `backend/.venv-ci` matching `backend/pylock.toml`; app on **Flutter 3.44.5**, the version `mobile-app-checks.yml` pins.

**Backend (`20e9d7a`)**

| Command | Result |
|---|---|
| `pytest tests/unit/test_desktop_updates.py tests/unit/test_integration_notification_validation.py` | **145 passed** |
| `scripts/backend-python-format --check` on all 4 files | clean |

Reverting only the two router edits and keeping the tests: 3 of the 4 download tests fail (`campaign_id` absent from the log line; the log-separator id accepted instead of rejected), and `test_campaign_send_without_campaign_id_returns_400` fails. The two that still pass are invariance assertions by design — that `campaign_id` does not change the served installer, and that a non-campaign send is forwarded untouched.

**App (0fc1a73)**

| Command | Result |
|---|---|
| `flutter test test/unit/notification_fcm_navigate_to_test.dart` | **18 passed** — 11 new, plus the 4 existing `navigateToFromFcmData` and 3 `waitUntilNonNull` cases |
| `flutter test` (full suite) | **1668 passed**, 5 skipped, 3 failed |
| `git diff --exit-code -- lib` after codegen | passes — `build_runner` produced no content change |
| `dart analyze` | **0 ERROR-severity issues**; my three files contribute **0** new issues |

The 3 suite failures are `on_device_transcription_upgrade`, `plans_sheet_l10n` and `voice_recorder_mic_contention`. I checked them out at `upstream/main` for the two source files I changed and re-ran: identical `+14 -3`. Pre-existing and unrelated — the last is a Windows file-lock (`PathAccessException ... used by another process`).

`scripts/analyze_ratchet.sh` itself needs `jq`, which this machine lacks, so I reproduced its comparison directly: every lint is at or under `analysis_baseline.json` except `unused_element` (2 vs 0), and both instances are in `lib/utils/manifest/manifest.g.dart` — a generated file that is byte-identical to `upstream/main` and absent from this diff. The four analyzer notes inside my files (`avoid_print` ×3, one unnecessary `dart:ui` import) are all on lines that predate this change.

**Preflight:** `pr_preflight.py --base upstream/main` over all 7 files passes, including `product-invariants` with INV-BETA-1 named. Its one failure is `desktop-backend-release-boundary-fixtures`, which needs `jq` and fails identically on unmodified `upstream/main`.

## What closes the loop

Send carries `campaign_id`, the tap emits it, the CTA carries it into the download route. Then "opens by platform" and "downloads by campaign" are the two queries the issue asks for — PostHog HogQL and Cloud Logging respectively.
Failure-Class: none

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12693?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/lib/services/notifications.dart
+++ b/app/lib/services/notifications.dart
@@ -57,10 +57,10 @@ class NotificationUtil {
   }
 
   static Future<void> onActionReceivedMethodImpl(ReceivedAction receivedAction) async {
-    if (receivedAction.payload == null || receivedAction.payload!.isEmpty) {
-      return;
-    }
-    await _handleAppLinkOrDeepLink(receivedAction.payload!);
+    // A payloadless notification is still a tap. Returning early here would drop
+    // it from the open count entirely, so pass an empty payload through and let
+    // the handler emit the event and then no-op on routing (#12645).
+    await _handleAppLinkOrDeepLink(receivedAction.payload ?? const <String, dynamic>{});
   }
 
   /// Public entry for FCM background/terminated notification taps (#5126).
@@ -92,7 +92,11 @@ class NotificationUtil {
     put('campaign_id', payload['campaign_id']);
     put('navigate_to', payload['navigate_to']);
     // Senders set one or the other; `notification_type` is the newer spelling.
-    put('notification_type', payload['notification_type'] ?? payload['type']);
+    // `??` alone would keep an empty `notification_type` and lose a valid legacy
+    // `type`, since an empty string is not null but is dropped by `put`.
+    final notificationType = payload['notification_type'];
+    final hasNotificationType = notificationType is String && notificationType.isNotEmpty;
+    put('notification_type', hasNotificationType ? notificationType : payload['type']);
     return properties;
   }
 
@@ -168,7 +172,12 @@ class NotificationUtil {
     final externalTarget = externalTargetFor(navigateTo);
     if (externalTarget != null) {
       try {
-        await launchUrl(externalTarget, mode: LaunchMode.externalApplication);
+        // launchUrl reports "no handler" by returning false, not by throwing, so
+        // the result has to be checked or a CTA that never opened looks succeeded.
+        final launched = await launchUrl(externalTarget, mode: LaunchMode.externalApplication);
+        if (!launched) {
+          Logger.debug('No handler opened external navigate_to=$navigateTo');
+        }
       } catch (e) {
         Logger.debug('Failed to open external navigate_to=$navigateTo: $e');
       }
--- a/app/lib/services/notifications.dart
+++ b/app/lib/services/notifications.dart
@@ -8,8 +8,10 @@ import 'package:awesome_notifications/awesome_notifications.dart';
 
 import 'package:omi/app_globals.dart';
 import 'package:omi/pages/home/page.dart';
+import 'package:omi/utils/analytics/analytics_manager.dart';
 import 'package:omi/utils/l10n_extensions.dart';
 import 'package:omi/utils/logger.dart';
+import 'package:url_launcher/url_launcher.dart';
 
 // Re-export the main notification service for backward compatibility
 // All notification functionality is now handled by the platform-aware service
@@ -62,9 +64,53 @@ class NotificationUtil {
   }
 
   /// Public entry for FCM background/terminated notification taps (#5126).
-  static void handleNavigateTo(String route) {
+  ///
+  /// Takes the whole FCM `data` map rather than just the route: the tap event
+  /// reads keys beyond `navigate_to` — notably `campaign_id` — and forwarding
+  /// only the route dropped them before they reached analytics (#12645).
+  static void handleFcmTap(Map<String, dynamic> data) {
     // Fire-and-forget: waits for navigator readiness before pushing (terminated cold start).
-    unawaited(_handleAppLinkOrDeepLink({'navigate_to': route}));
+    unawaited(_handleAppLinkOrDeepLink(Map<String, dynamic>.from(data)));
+  }
+
+  /// Properties for the single `Notification Opened` event emitted per tap.
+  ///
+  /// Senders put `campaign_id` on the FCM `data` map, which reaches the client
+  /// untouched, so an open can be tied back to the send that caused it. Keys
+  /// absent from the payload are omitted rather than sent as null, so a query
+  /// filtering on `campaign_id` sees only real campaign traffic (#12645).
+  @visibleForTesting
+  static Map<String, dynamic> notificationOpenedProperties(Map<String, dynamic> payload) {
+    final properties = <String, dynamic>{};
+
+    void put(String key, dynamic value) {
+      if (value is String && value.isNotEmpty) {
+        properties[key] = value;
+      }
+    }
+
+    put('campaign_id', payload['campaign_id']);
+    put('navigate_to', payload['navigate_to']);
+    // Senders set one or the other; `notification_type` is the newer spelling.
+    put('notification_type', payload['notification_type'] ?? payload['type']);
+    return properties;
+  }
+
+  /// The external destination [navigateTo] addresses, or null for an in-app route.
+  ///
+  /// Campaign CTAs point at an absolute URL — the desktop download route carrying
+  /// the campaign id — so that tap has to leave the app rather than be pushed onto
+  /// the navigator as a route name, which lands on a dead route (#12645).
+  ///
+  /// Only http and https qualify. A notification payload is attacker-influenced
+  /// input, and handing an arbitrary scheme to the platform launcher would let one
+  /// address anything the OS knows how to open.
+  @visibleForTesting
+  static Uri? externalTargetFor(String navigateTo) {
+    final uri = Uri.tryParse(navigateTo);
+    if (uri == null || !uri.hasScheme || !uri.hasAuthority) return null;
+    if (uri.scheme != 'http' && uri.scheme != 'https') return null;
+    return uri;
   }
 
   /// Extract a chat/conversation deep-link from an FCM data map.
@@ -103,9 +149,29 @@ class NotificationUtil {
   static Future<void> _handleAppLinkOrDeepLink(Map<String, dynamic> payload) async {
     WidgetsFlutterBinding.ensureInitialized();
 
-    final navigateTo = payload['navigate_to'];
-    if (navigateTo is! String || navigateTo.isEmpty) {
+    // Exactly one event per tap: both entry points — Awesome Notifications
+    // actions and FCM background/terminated taps — funnel through here. Emitted
+    // before the route check, so a push whose payload carries no `navigate_to`
+    // is still counted; PostHog's `$push_notification_opened` is Android
+    // autocapture only, so until now iOS opens were invisible (#12645).
+    AnalyticsManager().track('Notification Opened', properties: notificationOpenedProperties(payload));
+
+    final navigateTo = navigateToFromFcmData(payload);
+    if (navigateTo == null) {
       Logger.debug('Navigate To is null');
+      return;
+    }
+
+    // A campaign CTA points outside the app — the download route carrying the
+    // campaign id. Handled before the navigator wait: an external target needs no
+    // navigator, and waiting would stall the launch for up to 15s on a cold start.
+    final externalTarget = externalTargetFor(navigateTo);
+    if (externalTarget != null) {
+      try {
+        await launchUrl(externalTarget, mode: LaunchMode.externalApplication);
+      } catch (e) {
+        Logger.debug('Failed to open external navigate_to=$navigateTo: $e');
+      }
       return;
     }
 
--- a/app/lib/services/notifications/notification_service_fcm.dart
+++ b/app/lib/services/notifications/notification_service_fcm.dart
@@ -265,10 +265,9 @@ class _FCMNotificationService implements NotificationInterface {
 
     void handleNotificationTap(RemoteMessage? message) {
       if (message == null) return;
-      final navigateTo = NotificationUtil.navigateToFromFcmData(message.data);
-      if (navigateTo != null) {
-        NotificationUtil.handleNavigateTo(navigateTo);
-      }
+      // The whole data map goes through, not just the route: the tap event reads
+      // `campaign_id` from it, and routing still ignores everything else (#12645).
+      NotificationUtil.handleFcmTap(message.data);
     }
 
     // Background: app is backgrounded and the user taps a push notification (#5126).
--- a/app/lib/services/notifications/notification_service_fcm.dart
+++ b/app/lib/services/notifications/notification_service_fcm.dart
@@ -210,6 +210,17 @@ class _FCMNotificationService implements NotificationInterface {
         if (navigateTo != null && navigateTo.toString().isNotEmpty) {
           payload['navigate_to'] = navigateTo.toString();
         }
+        // Attribution keys ride along with the foreground notification too. A tap
+        // on it re-enters through onActionReceivedMethodImpl rather than
+        // handleFcmTap, so without these the Android foreground case — the user
+        // already has the app open, which is when a campaign matters most — would
+        // emit a `Notification Opened` carrying no campaign_id (#12645).
+        for (final key in const ['campaign_id', 'notification_type', 'type']) {
+          final value = data[key];
+          if (value != null && value.toString().isNotEmpty) {
+            payload[key] = value.toString();
+          }
+        }
 
         // Handle action item data messages
         final messageType = data['type'];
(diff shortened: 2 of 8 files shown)


SakashSrivastava:
All seven addressed in 217c57d1 — every one was real, no false positives.

The Android foreground gap was the serious one. `handleFcmTap` covered the background and terminated paths, but a foreground push on Android is displayed through AwesomeNotifications with a payload built from `navigate_to` alone, so the tap re-entered via `onActionReceivedMethodImpl` and emitted an event with no campaign_id — undercounting opens exactly when the user already has the app open. The foreground payload now carries campaign_id/notification_type/type, so both tap
paths are symmetric.

On the shared validation: both ends now read `CAMPAIGN_ID_PATTERN` and `CAMPAIGN_ID_MAX_LENGTH` from `backend/campaign_attribution.py`. I put it at the backend root rather than under `utils/` because `tests/unit/test_integration_notification_validation.py` stubs the whole `utils` package, so the constants arrived as MagicMocks and the validation silently no-opped under test. Root-level modules have precedent there (`desktop_download_page`). There's a test asserting both routers reference the same constant objects, so the two ends cannot drift back apart.

The sentinel is now `(absent)` rather than `"none"` — parentheses sit outside the pattern, so no valid campaign id can spell it.

Backend 145 → 153 passed (new: the "none" collision, `data: null`, five rejected campaign-id shapes, and the shared-constant assertion). App 1669 passed, same three pre-existing failures. `dart analyze` 0 errors, no new issues in the changed files.

Git-on-my-level (dismissed):
Thanks @SakashSrivastava — this is careful work. The single-tap funnel through `_handleAppLinkOrDeepLink`, the `or {}` guard for `"data": null` (with a regression test), and the `(absent)` sentinel that no valid id can spell all show real attention to the failure modes in #12645. I ran the two touched backend test files locally: 153 passed, including all the new campaign cases.

**One blocking issue: the "one shared rule" invariant has a trailing-newline gap.**

`backend/campaign_attribution.py` documents that the send end must never accept an id the download CTA rejects, and both routers import the same `CAMPAIGN_ID_PATTERN = r'^[A-Za-z0-9._:-]+$'` — but the two ends don't apply it the same way:

- `backend/routers/updates.py` validates via `Query(..., pattern=CAMPAIGN_ID_PATTERN)`, and pydantic matches the **whole** string, so a trailing newline is rejected (422).
- `backend/routers/notifications.py` uses `re.match(CAMPAIGN_ID_PATTERN, campaign_id)`, and Python's `$` also matches just before a trailing newline — so `'macos-push\n'` **passes the send-side gate**.

Verified locally: `re.match(r'^[A-Za-z0-9._:-]+$', 'ab\n')` is truthy while the same string gets a 422 from the download route. That is exactly the "send looks fine, CTA fails, attribution silently lost" scenario the module docstring says the shared rule exists to prevent — and a newline in `campaign_id` would also break the space-separated `desktop_download_served` log line it feeds.

Fix is one character: `r'^[A-Za-z0-9._:-]+\Z'` (or `re.fullmatch`). The parametrized rejection tests in `backend/tests/unit/test_integration_notification_validation.py` cover `'has\nnewline'` but not a *trailing* newline — please add `'ab\n'` there, and consider asserting in `backend/tests/unit/test_desktop_updates.py` that no served log line ever contains a newline.

**Per-file notes** (non-blocking):

- `app/lib/services/notifications.dart` — emitting `Notification Opened` before the route check is the right call (payloadless taps finally count; iOS opens become visible), and `externalTargetFor` restricting the launcher to http/https with a test per hostile scheme (`file:`, `javascript:`, `mailto:`, `tel:`, custom) is the right treatment of an attacker-influenced payload. Handling external targets before the 15s navigator wait avoids a real cold-start stall.
- `app/lib/services/notifications/notification_service_fcm.dart` — forwarding the whole `message.data` map via `handleFcmTap` and mirroring `campaign_id`/`notification_type`/`type` into the foreground-notification payload closes the Android foreground gap symmetrically. I checked the head tree: `handleNavigateTo` has no remaining callers, so the rename is clean.
- `backend/routers/notifications.py` — the 400 gate only fires for `kind == "campaign"`, so existing sends are unaffected, and `data.get('data') or {}` keeps `"data": null` tolerated instead of turning it into a 500 (tested).
- `backend/routers/updates.py` — `campaign_id` is attribution-only (test asserts byte-identical responses with and without it), and `_log_download_served` fires on both the beta-identity and stable paths so counts keep a denominator.
- `backend/campaign_attribution.py` — dependency-free and root-level was the right call given the stubbed `utils` package in the notification tests; the identity assertion that both routers bind the same constant objects is a nice touch.
- `backend/tests/unit/test_desktop_updates.py` — the `'none'`-is-not-the-sentinel test and the separator-rejection test both encode real invariants.
- `app/test/unit/notification_fcm_navigate_to_test.dart` — pinning the empty-string `notification_type` falling back to legacy `type` is a genuinely subtle case, good to have locked.

Two nits, not blocking: `_log_download_served`'s docstring still says "campaign_id=none" while the sentinel is `(absent)`; and on Android a foreground FCM push shown via AwesomeNotifications should not re-enter `onMessageOpenedApp` (data messages don't), so the two tap paths look mutually exclusive — only worth a comment if you've observed a double count.

Requesting changes just for the one-character pattern fix plus the trailing-newline test. Beyond that, two things need a human maintainer before merge: a product call on whether the 400-on-missing-`campaign_id` gate should be launch-blocking for admin sends (it changes operator behavior the first time a campaign payload is malformed), and final sign-off on the app launching external browsers from push payloads — the http/https allowlist looks right, but it is a new trust boundary and deserves maintainer eyes.

---

SakashSrivastava:
Fixed in 778fa6fc5. You're right, and the two ends genuinely disagreed.

One thing on the suggested fix: `\Z` in the shared pattern doesn't work. pydantic validates with the Rust regex crate, which has no `\Z` escape, so putting it in CAMPAIGN_ID_PATTERN makes the download route's Query fail to build — a `Field(pattern=r'^[A-Za-z0-9._:-]+\Z')` raises SchemaError at import. So I went with `re.fullmatch` instead, which keeps one pattern string valid on both engines.

Added `'ab\n'` to the parametrised rejection cases. Reverting only fullmatch back to match fails exactly that one parametrisation and nothing else, so it pins the reported gap rather than passing incidentally.

Also added `test_the_served_line_never_contains_a_newline`, asserting both emitted lines are free of CR and LF, so the record stays one-per-download as you suggested.

Nit fixed: `_log_download_served`'s docstring still said "campaign_id=none" after the sentinel became `(absent)`.

On the Android double-count nit — I haven't observed one, and I can't reproduce it without a device, so I've left the code alone rather than add a comment asserting behaviour I haven't verified. Happy to dig in if you've seen it in practice.
Backend 153 → 155 passed. Format check clean.

Git-on-my-level:
Verified on 778fa6fc — the trailing-newline gap is closed, and cleanly.

The two ends of the shared rule now agree, and not by accident: `backend/routers/notifications.py` validates with `re.fullmatch(CAMPAIGN_ID_PATTERN, campaign_id)`. I re-confirmed the exact divergence locally: `re.match(r'^[A-Za-z0-9._:-]+$', 'ab\n')` is truthy, `re.fullmatch` on the same pattern rejects it, and pydantic's whole-string match on the download route rejects it too. Thanks also for the note on `\Z` — right call: keeping one pattern string both engines accept is the sturdier invariant, and `re.fullmatch` buys exactly that. And `'ab\n'` in the parametrised rejections pins the reported case rather than passing incidentally — reverting only `fullmatch` back to `match` fails exactly that one parametrisation.

I ran the two touched backend files on this head in a clean worktree: **155 passed** (`tests/unit/test_desktop_updates.py`, `tests/unit/test_integration_notification_validation.py`), including `test_the_served_line_never_contains_a_newline` (CR and LF asserted out of every served line) and `test_sender_and_download_route_share_one_campaign_id_rule` (both routers bind the same constant objects).

Per-file, current head:

- `backend/campaign_attribution.py` — unchanged this commit; still the single dependency-free definition, and `(absent)` remains unspellable by the pattern.
- `backend/routers/notifications.py` — the gate reads well: `isinstance` + length + `fullmatch`, firing only for `kind == "campaign"`, and the `or {}` keeps `"data": null` tolerated rather than turning into a 500 (regression-tested).
- `backend/routers/updates.py` — `campaign_id` stays attribution-only (byte-identical responses with and without it), `_log_download_served` fires on both the beta-identity and stable paths, and the docstring now matches the `(absent)` sentinel.
- `backend/tests/unit/test_integration_notification_validation.py` — the `'ab\n'` case, the null-data no-500 case, and the shared-constants identity test all ran green here.
- `backend/tests/unit/test_desktop_updates.py` — newline-free served lines, `'none'` ≠ sentinel, separator rejection: all green here.
- `app/lib/services/notifications.dart` — one event per tap before the route check, http/https-only `externalTargetFor`, external launch handled before the 15s navigator wait. Reviewed statically; I have no Flutter toolchain on this box, so the Dart suite was not run by me.
- `app/lib/services/notifications/notification_service_fcm.dart` — whole-map `handleFcmTap` plus the foreground payload mirror keep both tap paths symmetric; also statically reviewed only.
- `app/test/unit/notification_fcm_navigate_to_test.dart` — good hostile-scheme coverage (`file:`, `javascript:`, `mailto:`, `tel:`, custom) and the empty-string→legacy-`type` fallback pin; not run locally, same reason. The pure-function split that makes these testable without a PostHog double or platform channel is the right shape.

The changes I requested are resolved, so I'm clearing that review. Two items from my earlier pass still want a maintainer before merge, unchanged: (1) the product/ops call on whether a 400 for a campaign send missing `campaign_id` should be launch-blocking for admin sends — it changes operator behavior the first time a campaign payload is malformed; (2) final sign-off on the app launching external http/https URLs from push payloads — the allowlist looks right, but it is a new trust boundary. `security-review` and `needs-maintainer-review` stay for those two; adding `positive-signal` since the technical side is now verified clean on my end.

---
*by AI on behalf of David — needs a maintainer's call on the two items above: the campaign 400-gate's operator behavior, and sign-off on the new external-launch trust boundary.*

PR #12697, opened 2026-09-03
chore: dead-code cleanup (phases 0/1/R) + dead-code & duplicate-route ratchets — 261 files, −63k lines

Git-on-my-level:
# Dead-code cleanup (phases 0/1/R) + dead-code ratchet checks

One PR: six cleanup commits plus focused review-fix commits from the 2026-09-03
independent high-risk audit and review round 2. Executed from the
2026-09-03 monorepo-cleanup inventory (10 scoped reports, 118 findings, every item
here grep-verified before deletion and machine-verified after).

## What this does

1. **chore(app)** — removes 56 unreachable Flutter screens (~11k LOC; old auth/onboarding,
   goals, settings, PayPal, sdcard pages — proven unreachable by import-graph reachability
   from `main.dart`, which is authoritative because routing is widget-direct), the
   wifi-permission vertical dead on both Dart+iOS+Android ends, two single-ended bridge
   channels, an orphaned 567-line decompiled BLE tool, 9 unused pubspec deps, and stale
   e2e `covers:` metadata.
2. **fix(backend) → REVERTED** — the shadowed-route change was attempted
   (`461f1a75ed` removed the duplicate `DELETE /v1/conversations/{id}/action-items`
   registration) and then **reverted** (`9648f161b6`) per the independent high-risk
   audit: the live Flutter swipe-delete chain still sends the description-body form
   to the first-mounted conversations handler and expects 204, so the shadowing is
   load-bearing for shipped clients — un-shadowing it would turn a single-item
   swipe-delete into delete-all-items-for-conversation on already-shipped builds.
   The file is restored byte-identical to base; the OpenAPI contract + four
   generated TS clients were regenerated back to net-empty vs base. The durable
   outcome is the **duplicate-route guard** (`.github/scripts/check_duplicate_routes.py`
   + self-test + checks-manifest entries) with an explicit, evidenced
   `KNOWN_DUPLICATES` exemption for this load-bearing duplicate; migrating the
   Flutter callers to per-item `DELETE /v1/action-items/{id}` and deduping is
   tracked follow-up (audit FINDING B).

Failure-Class: none

The shadowed-route instance is audited and guarded by the duplicate-route ratchet kept in
this PR; its class definition follows in a separate registry-only PR per the protocol
(#12698). No fix commit stands (the semantic change was reverted), so the declaration is
`none`; the `fix(backend)` subject earlier in this branch's history is reverted by
`9648f161b6`.
3. **chore(backend)** — deletes dead-on-arrival `utils/agent.py` (zero importers in all of
   git history, yet maintained until August), the WS-G11 shim trio **and** the
   promotion_routes/proposals modules whose only consumers were those shims (the new
   ratchet surfaced this transitive death the moment the shims went), the orphaned
   postprocess pipeline, phantom summary model, never-wired daily_reconciliation, dead
   v1 vector-db family, plus numbering/moves/pin cleanups.
4. **chore(desktop)** — removes retired agent-cloud residue (24 files), the 1,067-line
   compiled acp-bridge leftover, superseded agent adapters (~2.7k LOC with tests), the
   dead MCP server that shipped in every signed bundle, 8+5 never-imported Windows
   modules (incl. benchmark finds: BrainMap, the self-declared-unused insight helpers),
   and the d3-force dep.
5. **chore(web,plugins)** — 24 unused shadcn wrappers + 21 deps in admin, dead
   deps/components elsewhere, README fixes, plugins husks. **Security: `iq_rating` no
   longer hardcodes OMI_APP_ID/OMI_APP_SECRET** (public ~9.5 months) — it now fails
   fast at app **startup** unless both come from env (per the audit, the check moved
   off import time in `b3abb9bf91`: the plugins image-import smoke imports the module
   without env or app lifecycle, and an import-time raise would red CI / crash-loop
   Cloud Run before the env is set; a real deployment still refuses to serve
   unconfigured).
6. **chore (root/CI/docs) + test(ci)** — root strays, broken Makefile target, disabled
   docs-deploy cutover residue, orphaned scripts, stale path citations across
   workflows/docs, **and the phase-R dead-code ratchet**: four area scanners
   (import-graph reachability for flutter/agent/windows; zero-importer buckets with
   demotion passes for backend) wired into checks-manifest (local+ci) with committed
   baselines seeded from this cleanup and reason-bearing allowlists. New-guard
   justification: this is a dedicated tree-wide reachability check, not a shared
   primitive — the incident it would have caught is the 57-file Flutter accumulation
   (a June 2026 security fix landed on unreachable code) and DMC-4's dead server
   shipping in signed bundles. Validated against the hand-verified inventory before
   landing: flutter 56/56 with zero false positives; trap canaries (string-spawned
   omi-tools-stdio.ts, dynamically-imported backendTools.ts) correctly stay alive.

- **INV-AUTH-1, INV-DATA-1, INV-BETA-1** (triggered by `codemagic.yaml`): the change is
  comment-only path-citation fixes (`omi/desktop/macos/` → `desktop/macos/`) plus the
  approved raw-digest fixture refresh required by the release-process guard. The
  semantic digest of the document is byte-identical (verified with the guard's own
  canonicalization before the fixture was updated); no workflow, credential, routing,
  or beta-identity semantics changed.
- **INV-MEM-1** (`app/lib/pages/memories/**`): two memory pages deleted as unreachable
  (zero importers from `main.dart`; verified by reachability scan and per-basename rg).
  No memory behavior changed — the screens were unreachable dead code.
- **INV-MEM-4** (`backend/utils/memory/**`): `daily_reconciliation.py` deleted — it was
  a planning seam with no producer ever wired (only its own test imported it);
  `v3/archive_visibility_readiness.py` moved (not deleted) next to its only caller in
  `scripts/`. Promotion/consolidation authority logic is untouched.
- **INV-TASK-2** (`backend/config/task_intelligence_sources_v1.json`): `240172c5` drops
  the source-manifest path of `app/lib/pages/action_items/widgets/action_item_tile_widget.dart`,
  a file this PR deleted (the swipe-to-delete widget whose caller chain the audit used to
  revert BRT-2). Manifest hygiene only — the capture-authority guard's source inventory
  must not list paths that no longer exist; no capture, Candidate, acceptance, or writer
  semantics change.

## Verification

- `flutter analyze` byte-identical before/after (0 new errors); analyze ratchet passes
  with 9 rules improved; `flutter pub get` lock refresh (removals only).
- 500+ targeted backend unit tests green (venv) incl. conversations/action-items,
  memory contracts, desktop surface parity, workflow contracts; import-time side-effect
  scan 0 violations; route-policy baseline current.
- Duplicate-route guard: self-suite 5/5 (incl. the new exempted-known-duplicate case);
  guard exits 0 on the post-revert tree, printing the one exempted load-bearing
  duplicate.
- iq_rating: import without env succeeds; its startup hook raises without env and
  passes with env (verified against the backend venv's FastAPI).
- Generated `FlutterGeneratedPluginSwiftPackage/Package.swift` restored byte-identical
  to base (`fc005f748f`); net diff vs base empty.
- Review round 2 folds: `dead-code-ratchet` triggers widened to every `backend/**/*.py`
  (R2-1 — the scanner reads all of them; manifest validator green, `test_run_checks`
  47/47); pusher `pylock.toml` regenerated via the repo's uv tooling without the removed
  typesense pin (R2-2 — delta is exactly the typesense block, TOML re-validated);
  login e2e flow expects updated to the live auth consent text (audit nit). Pre-existing
  stragglers left deliberately: `integrations.mdx` `getTypeOfBluetoothDevice` mention
  (line untouched by this PR) and one stale entry in
  `backend/tests/.import_time_side_effects_legacy` (shrink-only file).
- CI-fold from this PR's own Hygiene run: `scan_ts` built its bare-stem map over
  filesystem-dependent `rglob` order, so with three windows files sharing the stem
  `index` a sibling ref's resolution — and with it reachability verdicts — could flip
  between machines (observed as a CI-only `dead-code[windows]` failure on identical
  content). Stem resolution is now sorted/deterministic, and the shipped pi-mono
  payload entry (`jiti`-loaded via a constructed runtime path the scanner cannot see)
  plus its static import are allowlisted with those reasons (`b62b21fdfb`); windows
  floor 7, self-suite 10/10.
- AST-based dangling-import scan across all Dart/Python/TS source+tests: zero unresolved
  in-repo imports caused by this PR (one *pre-existing* finding flagged:
  `scripts/memory-continuity-gauntlet.py` imports a nonexistent test — left for its owner).
- 3,829 Windows/agent TS import specifiers resolve; edited package.json/lockfiles parse
  and are internally consistent.
- Dead-code ratchet `--check` exit 0 on the post-cleanup tree (floors: flutter 0,
  backend 5 pending manual review, agent 0, windows 0; all survivors allowlisted with
  reasons); its 10-test hermetic self-suite green; manifest validator: 160 checks,
  0 errors; `test_run_checks` 47/47.
- Full local pre-push preflight green after the fixture/digest and hygiene fixes above
  (release-guard digest fixture for the comment-only codemagic.yaml edit — semantic digest
  proven identical; changelog `{"kind": "none"}` fragment; lifecycle headers on three
  touched backend/scripts files; OpenAPI contract + four generated TS clients regenerated,
  then regenerated back to net-empty vs base after the revert; migration rename reverted
  rather than fight the shrink-only import-purity allowlist).
  **One hatch used, disclosed:** `PRE_PUSH_SKIP_WINDOWS_KGWORKER_NATIVE_CLOSURE` — the
  pi-mono unpack verifier fails identically on pristine `origin/main` in this environment
  (this machine's seeded node_modules corrupt pnpm's hoisted layout; reproduced on a clean
  main worktree with full `--force` install). CI's clean Linux install is the arbiter.

## Not in this PR (deliberately deferred — gated on external evidence or decisions)

chat v1 quartet + sync v1 + advice/focus routers + legacy appcast (need one Cloud Run
traffic-log review) · web/frontend page retirements + marketplace unification · plugins
monolith retirement (dispatch-history check) · account-cutover (kept by ruling; bind
when the rewrite serves accounts) · universal-memory legacy-row retirement milestone ·
macOS dual-logging consolidation (Phase 3) · hardware archives (~97 MB). Backend
baseline carries 5 zero-importer candidates (`utils/llm/trends.py`,
`utils/memory/projections.py`, `canonical_kg_promotion.py`, `legacy_backfill_inventory.py`,
`utils/other/task.py`) pending one manual review each.

## Coordination notes

- **iq_rating deploy**: set `OMI_APP_ID` + rotated `OMI_APP_SECRET` env on the Cloud Run
  `plugins` service before the next manual deploy (workflow is dispatch-only; nothing
  breaks until then). Note the fail-fast moved to app startup (`b3abb9bf91`): a deploy
  without the env now refuses to start instead of silently running with hardcoded
  defaults. The leaked secret must be rotated regardless of this PR.
- `pnpm-lock.yaml` (windows) and `bun.lock` (web/app) were hand-edited surgically and
  validated; CI's frozen-lockfile install is the final arbiter and regenerates the same
  delta if it disagrees.
- `app/ios/Flutter/ephemeral/.../Package.swift` is tracked but machine-generated — expect
  churn on local `pub get`; this PR only removed the `in_app_review` entries.
- Two dispatch-only workflows deleted (`entellegence_*`) and the disabled docs-deploy
  workflow removed with its dead npm package — flagged for maintainer attention per
  workspace rules on `.github` changes.


Git-on-my-level:
Monitor run — review round 3 (all 14 unresolved Cubic threads addressed, 12 with new commits, 2 verified as already-fixed at head and resolved as such).

**`fix(ci)`: duplicate-route ratchet — closed six route-discovery gaps** (commit `8da3c3d0`, review threads on `check_duplicate_routes.py`):
- Named-attribute mounts (`api_key_management.mcp_router` / `.developer_router`) are now resolved; route tables are keyed by `(module, router_var)` so two routers from one module mounted together are not a false duplicate, while cross-module shadowing via named routers is caught.
- Module-level constant paths (`jit_*` `_SNAPSHOT_PATH` pattern) and both legacy `add_api_route` forms (literal args; the `desktop_deprecated` `_ROUTES` dict + loop) are now parsed — verified live: the checker now sees **784** mounted `(app, method, path)` combinations vs 654 before, still green with the single evidenced `KNOWN_DUPLICATES` exemption.
- `api_route` now parses `methods=[...]` instead of conservatively expanding to all five verbs — no more false duplicates for undeclared methods.
- `{param}` placeholders normalize to `{}` so `/t/{a}` vs `/t/{b}` shadowing is detected (FastAPI matches them identically).
- Double-mounting one router is caught (`len(v) > 1`, not `len(set(v)) > 1`).
- Dead helpers/unused regexes removed so the checker adds no dead code of its own.
- Self-test extended 5 → 11 cases, each covering one fixed gap. All pass; real-repo run exits 0.

**`docs`: stale references corrected** (commit `62daa4a4`):
- `apps-marketplace.yaml` covers now lists `category_section.dart` — the flow's S1–S7 browsing renders `CategorySection`/`SectionAppItemCard`; `AppListItem` only renders in search-filter results the flow never enters.
- `goals-tracking.yaml` + `goals_widget.dart`: GoalsWidget renders after the processing-conversations widget in `conversations_page.dart` (TodayTasksWidget lives in `home_content.dart`); with zero goals the widget hides and the home-screen Add Goal entry point is `ActionItemsPage._buildGoalsRow`.
- `00-INDEX.md`: `focus/stats.ts` and `AutoCreatedTasksStep.tsx` entries updated to match the deletion note (marked resolved-by-deletion, no longer presented as actionable).
- `plugins/README.md`: monolith (`plugins/requirements.txt` → `./omi-plugin-sdk`) vs per-service (own `requirements.txt` → `../omi-plugin-sdk`) SDK install paths documented separately.

**Already fixed at previous head, verified and resolved:** pi-mono allowlist entries (`windows.allowlist.json`, commit `dcb9b967`) and pusher `pylock.toml` typesense regen (commit `ed57e17cb`) — both confirmed present locally and the `dead-code-ratchet` exits 0.

**Verification:** checker self-test 11/11 OK; `check_duplicate_routes.py` real-repo exit 0 (784 combos); `check_dead_code.py` exit 0 across all four areas; yaml parse OK; black-formatted. Pre-push gate was run manually against `origin/main` (6 PASS); `check-manifest-contract` only fails locally because the manifest hardcodes `python3` = 3.9 on this host (`str | None` needs 3.10+) — the equivalent check passes under python3.11, so CI is unaffected. Push used the documented narrow `--no-verify` bypass for that host limitation. CI on the new head: fast lanes green, long lanes in progress, 0 failures so far.

PR #12705, opened 2026-09-04
test(desktop): exempt the get_screenshot alias from the tool-declaration check

aryanorastar:
> **Scope changed while this was open.** `8234c835fe` (#12607) landed the three-name exemption list on `main` on 2026-09-05, so the red assertion this PR opened to fix is **already green there**. What remains is a different, smaller thing: the exempt set is now *derived* from the manifest fixture rather than restated as a literal, so it cannot go stale again. Judge it on that, not on the original framing.

## What this does now

`ChatDiscoverabilityTests.testDesktopCapabilitiesExistInAgentToolDeclarations` computes its Local-API-only exemptions from `agent/tests/fixtures/tool-manifest.json` — the same fixture the assertion already reads — instead of carrying a hand-maintained list:

```swift
var localApiOnlyTools = Set<String>()
for entry in fixture {
  guard let name = entry["name"] as? String,
    let executor = entry["executor"] as? [String: Any],
    executor["kind"] as? String == "localApiOnly"
  else { continue }
  localApiOnlyTools.insert(name)
  for alias in (entry["aliases"] as? [String]) ?? [] {
    localApiOnlyTools.insert(alias)
  }
}
```

A tool is exempt when the manifest marks it Local API only; an alias inherits its canonical entry's exemption.

## Why it is still worth landing

The hardcoded list is what broke the test. `look_at_frame` reached the desktop registry in #12084 on 2026-08-27 while the list had not changed since 2026-08-01, so a manifest addition became a red assertion — and stayed red until #12607 hand-edited the list again. Deriving it means the next Local-API-only tool or alias needs no edit here at all.

This was [cubic's P3](https://github.com/BasedHardware/omi/pull/12705) on the original head, and it was right. My reply at the time said deriving was not possible because `GeneratedToolCapabilities.swift` carries no alias data. **That was wrong** — this test does not read that file. It reads the fixture, which carries both halves:

```
get_screenshot:   executor.kind = localApiOnly, aliases = ["look_at_frame"]
get_local_status: executor.kind = localApiOnly
```

## Guard on the derivation itself

`XCTAssertFalse(localApiOnlyTools.isEmpty, ...)` fails loudly if the fixture ever stops marking executor kinds. That is the failure mode a derived set introduces and a literal does not — an empty derivation would silently exempt nothing and pass.

## Verification

Equivalence checked against `main`'s own fixture rather than assumed:

```
localApiOnly entries: {'get_local_status': [], 'get_screenshot': ['look_at_frame']}
→ derived set = {get_local_status, get_screenshot, look_at_frame}
```

Exactly `main`'s literal list, computed instead of maintained.

- `swift test --filter ChatDiscoverabilityTests` — **25 passed, 0 failures**
- Mutation-checked twice: dropping alias inheritance fails with `Missing agent tool declaration for look_at_frame`; dropping a canonical name fails with its own missing-declaration error. The assertion has teeth in both directions.
- Merged current `main` (0 behind). The conflict was this file; I kept the derivation and carried `main`'s comment rationale into it, since #12607 explains *why* the alias inherits better than the original text did.

## On the earlier history objection

Resolved: the reviewer withdrew it and confirmed the dates independently. Worth recording the cause, since it generalises — the GitHub commit API caps its file listing at 300 entries and #12084 has exactly 300, so its `desktop/macos` files were silently dropped from the listing. "No desktop files in that commit" was truncation, not fact. Content-at-ref (`git cat-file -e <ref>:<path>`) cannot be truncated that way; `log -S`, `--name-only` and the commit API each can.

Test-only. `{"kind": "none"}` changelog fragment unchanged.

Failure-Class: none


Git-on-my-level (dismissed):
Verified the core fix — this exemption is correct, and the mutation check is genuine.

- In `agent/tests/fixtures/tool-manifest.json`, `look_at_frame` appears only as an alias on `get_screenshot` (`"adapters": {"local-agent-api": {"advertised": true, "aliases": ["look_at_frame"]}}`). The test only collects names advertised via `pi-mono`/`omi-tools-stdio` or with `runtimeControl` executors, so `look_at_frame` can never land in `declaredTools` — while `GeneratedToolCapabilities.swift` does register it as a `.desktopChat` capability, so the assertion fails on `main` exactly as described.
- I re-derived the loop statically against the fixture and the generated registry: with the old two-name list, `look_at_frame` is the only missing name; with your three-name list, zero; dropping `get_local_status` instead reports `get_local_status`. The check keeps its teeth — this is an exemption, not a blanket pass.
- `LocalAgentAPIServer.swift:318` mapping `look_at_frame` → `get_screenshot` checks out, as does the `{"kind": "none"}` fragment per `check-desktop-changelog.py`'s accepted fragment kinds.

One correction needed before merge, though: the history recorded in the added comment (and the PR body) doesn't match what the repo shows.

- "It reached the registry in #12084 (2026-08-27)" — #12084 created the *backend* `look_at_frame` tool (`backend/utils/retrieval/tools/frame_request_tools.py`); its diff contains no `desktop/macos` files at all. On the desktop side, `git log -S 'look_at_frame' -- desktop/macos` returns a single commit: the #12469 merge (`e1eca38ff2`, 2026-08-31), which is where `GeneratedToolCapabilities.swift` picked up the entry.
- "this list had not changed since 2026-08-01 … the assertion below has failed ever since" — the test file itself (with the two-name list) landed in that same #12469 merge, i.e. the assertion has been red since birth (~4 days), not since 2026-08-27.

Since the comment's entire job is to record why the exemption exists, pointing at the wrong PR and dates will misdirect the next person who greps here. Could you update those lines to the #12469 / 2026-08-31 born-red story (and the PR body table to match)? The one-line exemption itself can land unchanged.

No maintainer escalation needed — once the comment matches the history this is a straightforward merge; the diagnosis of *why* the exemption is correct (alias of a Local-API-only tool) stands.

---

kodjima33 (approved):
Correct, narrow fix for a test assertion that's been red on main since the look_at_frame alias landed (#12469/#12084). Bug-fix fast-lane score is 2/5 (small diff + symptom described, but no fix/bug title vocabulary, no bug label, title doesn't start with fix(desktop)) so holding at approve rather than owner-override merge.

aryanorastar:
I checked the history against the repo before changing the comment, and it holds as written — so I'd rather not edit it to the #12469 story, because that would put an incorrect history into the file the comment exists to explain.

Three claims, each re-derived on current `main` (`176a02fe0d`):

**1. #12084 does touch `desktop/macos`.**

```
$ git show --stat --format="" 2af083251c | grep desktop/macos
 desktop/macos/Desktop/Sources/APIClient.swift      |    18 +
 desktop/macos/Desktop/Sources/OmiApp.swift         |     5 +
 desktop/macos/agent/src/adapters/interface.ts      |     2 +
 desktop/macos/agent/src/adapters/pi-mono.ts        |     4 +
 ...
```

`2af083251c` is `feat: add JIT knowledge ledger foundation and guarded adoption (#12084)`, dated 2026-08-27.

**2. The alias search returns #12084, not #12469.**

```
$ git log -S 'look_at_frame' --oneline -- desktop/macos/Desktop/Sources/Generated/GeneratedToolCapabilities.swift
2af083251c 2026-08-27 feat: add JIT knowledge ledger foundation and guarded adoption (#12084)
```

Same single commit for `ScreenContextTelemetry.swift`, and the same commit for `-- desktop/macos` as a whole.

I suspect the empty result came from the path: the file is under `Sources/**Generated**/GeneratedToolCapabilities.swift`, and a path spec without that directory segment matches nothing and returns silently. I hit exactly that a few minutes ago and briefly drew your conclusion from it.

For completeness, `e1eca38ff2` is `fix(listen): fail over at the audio send path, not just the death-monitor poll (#12469)` — an audio-failover change, and it does not touch `GeneratedToolCapabilities.swift` at all:

```
$ git show --stat --format="" e1eca38ff2 | grep -c GeneratedToolCapabilities
0
```

**3. The test file predates the alias by nearly four weeks, and was green for all of it.**

```
$ git log --diff-filter=A --format="%h %ad %s" --date=short -- desktop/macos/Desktop/Tests/ChatDiscoverabilityTests.swift
0d3dfd6ff1 2026-08-01 fix(app): label import-history rows from the viewer's local day (#10988)

$ git show 0d3dfd6ff1:desktop/macos/Desktop/Tests/ChatDiscoverabilityTests.swift | grep localApiOnlyTools
    let localApiOnlyTools: Set<String> = ["get_local_status", "get_screenshot"]
```

So the two-name list landed 2026-08-01 and was correct on the day it was written — `look_at_frame` did not exist in the registry yet. It became stale on 2026-08-27 when #12084 generated the alias into `GeneratedToolCapabilities.swift` without updating the exemption list. Red since 2026-08-27, not born red.

That is the story the comment already tells, so I've left it and the PR body unchanged. Happy to be shown otherwise if you're reading a different commit than `2af083251c` — but the born-red framing would misdirect the next person here, which is the outcome we both want to avoid.

On the separate cubic P3 (derive the exemptions instead of listing them): I agree it is the better design and said so in the PR body. The blocker is that the alias relationship exists only in the TypeScript manifest and is not carried into `GeneratedToolCapabilities.swift`, so there is nothing on the Swift side to derive from without changing the generator. That is worth doing as its own change; I did not want to put a generator change in the path of unbreaking a red lane.

Git-on-my-level:
My earlier history objection was wrong — @aryanorastar, your rebuttal checks out and the comment stands as written. Withdrawing that request; from automation's side this is merge-ready under the recorded approval.

I re-derived the history independently (content-at-ref, not path-filtered logs) before replying:

- `GeneratedToolCapabilities.swift` at `2af083251c^` (#12084's parent) contains no `look_at_frame`; at `2af083251c` (2026-08-27) it does. That commit is when the alias entered the desktop registry.
- `e1eca38ff2` (#12469, the audio-failover fix) touches 7 files, none under `desktop/` — it never touched `GeneratedToolCapabilities.swift`.
- The two-name `localApiOnlyTools` list exists at #12084's parent and is unchanged through it, so the assertion has genuinely been red since 2026-08-27 — not born-red in #12469.

For the record, where my earlier claim came from, since it's a failure mode worth remembering when auditing large squash merges: the commit API's file listing caps at 300 entries and #12084 has exactly 300, so its `desktop/macos` files were silently dropped from the listing I checked — "no desktop files" was truncation, not fact — and my `log -S` miss was the omitted `Sources/Generated/` path segment, the same trap you flagged.

Everything else from my earlier review stands unchanged: in `agent/tests/fixtures/tool-manifest.json`, `look_at_frame` exists only as a `local-agent-api` alias on `get_screenshot`, so it can never enter `declaredTools` (no `pi-mono`/`omi-tools-stdio` advertisement, no `runtimeControl` executor); `LocalAgentAPIServer.swift:318` maps it back to `get_screenshot`; the mutation check showed the assertion keeps its teeth; and the `{"kind": "none"}` changelog fragment is valid per `.github/scripts/check-desktop-changelog.py`. The three-name list is the correct narrow fix.

Thanks for pushing back with receipts rather than editing a correct comment to match a wrong review — that was exactly the right call, and it kept the repo's recorded history accurate.

---

aryanorastar:
Appreciated — and the root cause you found is the more useful half of this.

The GitHub commit API capping its file listing at 300 entries, with #12084 landing on exactly 300, is a genuinely nasty auditing trap: the response looks complete, there is no truncation marker in what you read, and "no `desktop/` files in this commit" comes back as a confident fact rather than a gap. My own first pass hit an adjacent version of it — I searched `Sources/GeneratedToolCapabilities.swift` and got an empty result, because the file is actually at `Sources/`**`Generated/`**`GeneratedToolCapabilities.swift`. A pathspec that matches nothing and a listing that silently truncates fail identically: both return "not there" and neither says why.

The lesson I am taking from it is the one you applied — content-at-ref beats path-filtered history. `git cat-file -e <ref>:<path>` and reading the blob at the commit and its parent cannot be truncated or defeated by a wrong pathspec, where `log -S`, `--name-only` and the commit API can each quietly return empty for reasons unrelated to the question.

On the merge state: the formal `CHANGES_REQUESTED` review is still recorded, so branch protection continues to block even though the objection is withdrawn. That one needs a maintainer to dismiss it or submit an approving review — nothing I can do from the author side.

aryanorastar:
Merged current `main` on `9f43ab1924`; the branch is 0 behind and GitHub now reports `MERGEABLE`.

**The scope of this PR changed while it was open, and the description now says so up front.** `8234c835fe` (#12607) landed the three-name exemption list on `main` on 2026-09-05, so the red `ChatDiscoverabilityTests` assertion this opened to fix is already green there. What remains is the derivation: the exempt set is computed from `agent/tests/fixtures/tool-manifest.json` rather than restated as a literal, so it cannot go stale again — which is what broke the test in the first place, and what [cubic's P3](https://github.com/BasedHardware/omi/pull/12705) asked for.

**The conflict was this file, and I kept both sides' intent.** `main` rewrote the exemption comment while this branch replaced the list with a derivation. I kept the derivation and carried `main`'s rationale into it — #12607 explains *why* the alias inherits better than the original text did (the local API is the only adapter either name is advertised to, so neither can appear in a chat adapter's declarations), and that reasoning holds regardless of how the set is built.

**Equivalence verified against `main`'s own fixture**, not assumed:

```
localApiOnly entries: {'get_local_status': [], 'get_screenshot': ['look_at_frame']}
→ derived set = {get_local_status, get_screenshot, look_at_frame}
```

Exactly `main`'s literal list, computed instead of maintained.

`swift test --filter ChatDiscoverabilityTests` — **25 passed, 0 failures** on the merged head.

**One correction carried into the description.** My earlier reply said deriving was not possible because `GeneratedToolCapabilities.swift` carries no alias data. That was wrong: this test never reads that file. It reads the fixture, which carries both the `localApiOnly` executor kind and the alias list. The derivation exists because that objection was right and my reason for dismissing it was not.

Only remaining gate is the recorded `CHANGES_REQUESTED`. That objection was withdrawn — the reviewer re-derived the history independently and confirmed it — but the formal review still stands and needs a maintainer to dismiss it. Nothing I can push moves that.

kodjima33 (approved):
Test-only hardening (derives Local-API-only exemption set from the tool-manifest fixture instead of a hand-maintained list) — bug-fix heuristic scores 2/5 (small surgical diff + clear before/after symptom description, but no fix/bug title vocabulary or bug label), below the 3/5 desktop fast-lane bar. Original red-assertion bug it opened to fix is already independently fixed on main (#12607); this PR's remaining value is preventing recurrence. Nik already personally approved the substantive commit; only a trivial merge-from-main landed since. Approve only, not merging.

Git-on-my-level:
Verified on the current head (`9f43ab1`): the fixture-derived exemption is equivalent to `main`'s literal list and keeps its teeth. With #12607 already green on `main`, this head is the recurrence-prevention half — it removes the hand-maintained list that let the assertion go stale (cubic's original P3), and the merge from `main` kept both sides' intent: the derivation plus #12607's rationale comment.

- **`desktop/macos/Desktop/Tests/ChatDiscoverabilityTests.swift`** — the derivation reads the same fixture the assertion loop already loads. At this head the manifest has exactly two `executor.kind == "localApiOnly"` entries: `get_local_status` (no aliases) and `get_screenshot` (top-level `aliases: ["look_at_frame"]`), so the derived set is `{get_local_status, get_screenshot, look_at_frame}` — identical to the literal #12607 landed on `main`. Equivalence, not a behavior change.
- **Teeth, re-checked by mutation** — deriving canonical names only (dropping the alias-inheritance loop) fails the assertion on `look_at_frame`; an empty derived set trips the new `XCTAssertFalse(localApiOnlyTools.isEmpty)` guard before failing on all three names. The exempt set is also disjoint from `declaredTools` (57 names via `pi-mono`/`omi-tools-stdio`/`runtimeControl`), so the derivation can't silently weaken the check.
- **`desktop/macos/changelog/unreleased/20260903-chat-discoverability-alias-exemption.json`** — `{"kind": "none"}` validates per `check-desktop-changelog.py`'s accepted fragment kinds, and the reason text is accurate. Minor note, not a change request: `Desktop/Tests/` already sits in that script's exempt prefixes, so the fragment is belt-and-braces rather than required — fine to keep.
- **CI at this head** — `Desktop Swift Build & Tests` (and the static/test-contract lane) green.

The recorded CHANGES_REQUESTED is dismissed below: its sole objection — the history attribution in the exemption comment — was withdrawn on 2026-09-05 after the history was re-derived from content-at-ref (the objection had come from a truncated 300-file commit listing). With that cleared and the approval already recorded on this head, nothing further is needed from review's side; the merge call stays with a maintainer.

*by AI on behalf of David — no escalation needed.*

PR #12666, opened 2026-09-03
fix(desktop): make tagged memory imports provider-independent

aryanorastar:
## What changed and why

Fixes the macOS ChatGPT/Claude memory import failure reported in #12442. Omi's export prompt already asks the provider for one durable, explicitly tagged memory per line, but the desktop app sent that final-form output through a second model extraction request; any extraction outage or rejection collapsed into the same generic error.

The importer now parses the documented `[date]`, `[recent]`, `[earlier]`, `[long-term]`, and `[unknown]` line format locally, then sends the resulting artifacts through the existing authenticated batch persistence path. Unstructured paste still uses the managed extractor. Extraction/save failures are classified into bounded categories and shown as actionable messages, and partial saves report the unsaved count.

Closes #12442

## Product invariants affected

- INV-INT-1 — connector imports continue to use the canonical authenticated evidence batch path; only the redundant extraction hop is skipped for already-structured exports.

## How it was verified

- Reproduced the reported input shape with the exact 10-line `[long-term]` sample from #12442; the regression test parses all 10 memories without calling a provider.
- Built and launched the named macOS bundle `com.omi.omi-memory-import` against the offline local auth/backend/Firestore/Typesense stack.
- Ran `connector-import.yaml` through the desktop automation bridge: PASS in 542.72 ms, zero log errors.
- The app made a real `POST /v3/memory-imports/batch`; an independent Firestore emulator query found exactly one active `chatgpt_memory_log` artifact containing `[long-term] [[MARKER:connector-import]] hermetic connector import probe` for the signed-in synthetic user.
- `desktop-core-harness.sh --self-check --skip-backend-contracts`: PASS (74 flows, 188 registered actions).
- `make preflight`: PASS with this PR body.

## Tests

- `./scripts/dev-feedback.py --once swift 'OnboardingMemoryLogImportServiceTests|ConnectorImportOperationsTests|DesktopAutomationSecondaryActionTests'`
  - 54 tests passed, 0 failures on the current `origin/main` head.
- `OnboardingMemoryLogImportServiceTests.testIssue12442TenLineChatGPTExportParsesWithoutProvider` is the issue-specific regression guard.
- Parser coverage includes export wrappers, bullets, numbered items, case normalization, unknown tags, deduplication, unstructured fallback, and bounded network/backend failure classification.
- Outcome mapping coverage includes full success, partial save reporting, empty extraction guidance, and actionable network failure copy.

## Failure class (fixes)

Failure-Class: none

[...]

Issue #12442, opened 2026-08-30
macOS: ChatGPT memory import still fails with “The import couldn't run. Try again.” even for minimal input


StormySailoryeg:
## Summary

The ChatGPT memory importer on macOS is still failing with the generic error:

> The import couldn't run. Try again.

This reproduces with both:

1. A full ChatGPT memory export generated using Omi's own prompt.
2. A deliberately tiny 10-line plain-text memory sample.

Because the minimal sample also fails, this does not appear to be caused by export size, formatting complexity, or the 40,000-character truncation in the current import path.

## Environment

- Platform: macOS desktop app
- Date observed: 2026-08-30
- Surface: ChatGPT memory import / connector import

## Steps to reproduce

1. Open Omi on macOS.
2. Go to the ChatGPT memory import flow.
3. Paste a normal ChatGPT memory export generated using the prompt Omi provides.
4. Run the import.
5. Observe: `The import couldn't run. Try again.`
6. Repeat with this minimal sample:

```text
[long-term] My name is Dennis Cox.
[long-term] Call me Captain Stormy or Dennis.
[long-term] I live in Edmonton, Alberta, Canada.
[long-term] I am a Canadian Navy veteran.
[long-term] I run Stormy Sailors Maker Lab.
[long-term] I use Linux, macOS, WordPress, servers, QR systems, and maker technology.
[long-term] I value direct communication and honest disagreement.
[long-term] I prefer concise answers for simple questions and detailed explanations for complex ones.
[long-term] My long-term goal is to live aboard a motor vessel on Canada's west coast.
[long-term] The Tactical Pirate is one of my major maker projects.
```

7. Observe the same failure.

## Expected behavior

The minimal sample should extract and save at least several durable memories, or the UI should surface a specific actionable failure reason.

## Actual behavior

The import fails with the generic `.failed` message:

`The import couldn't run. Try again.`

## Relevant current code

The macOS connector maps `OnboardingMemoryLogImportService.ImportOutcome.failed` to this exact message in:

- `desktop/macos/Desktop/Sources/MainWindow/Pages/ConnectorImportOperations.swift`

The import service currently:

- trims the pasted text,
- sends up to 40,000 characters to `APIClient.shared.extractMemoryLog(...)`,
- converts extracted memories into batch items,
- saves them through `OnboardingImportEvidenceService.save(...)`,
- returns `.failed` when extraction throws or when no memories are successfully saved.

Relevant file:

[...]

PR #12711, opened 2026-09-04
fix(desktop): preserve mixed-format memory imports

aryanorastar:
## Summary

Follow-up to #12666 that addresses every actionable finding from the post-merge Cubic review.

- fall back to the managed extractor when a paste mixes tagged and untagged memory lines, preventing silent data loss
- preserve a distinct `server` telemetry class for backend 5xx failures
- surface partial-save counts during onboarding as well as on the Apps connector surface
- clarify the user-facing changelog entry and align test names with what they prove

## Root cause

The local parser treated any paste containing at least one recognized `[tag]` line as fully structured. That bypassed the managed extractor and discarded every unmatched non-empty line. Separately, onboarding ignored the service's partial-save count, and server failures were collapsed into `invalid_response` telemetry.

## Verification

- exact #12442 ten-line ChatGPT export remains covered
- mixed tagged/untagged input now returns to the complete-text managed extraction path
- partial saves retain their unsaved count in onboarding messaging
- backend 5xx errors emit the bounded `server` telemetry class
- focused Swift suite: 75 tests passed, 0 failed for the identical review patch before rebasing onto current `main`
- `git diff --check`: clean

## Scope

macOS desktop only. No iOS, Android, Windows, or hardware behavior changes.

Product-Invariant: INV-INT-1

Failure-Class: none

Line-Count-Exception: desktop/macos/Desktop/Sources/Onboarding/OnboardingPagedIntroCoordinator.swift | 1933 -> 1940 | surface the existing bounded partial-save outcome during onboarding without creating a second result-mapping path

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12711?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/Integrations/IntegrationConnectTelemetry.swift
+++ b/desktop/macos/Desktop/Sources/Integrations/IntegrationConnectTelemetry.swift
@@ -79,6 +79,7 @@ enum IntegrationConnectTelemetry {
     case permission = "permission"
     case authentication = "authentication"
     case conflict = "conflict"
+    case server = "server"
     case invalidResponse = "invalid_response"
     case resourceExhausted = "resource_exhausted"
     case unknown = "unknown"
@@ -140,7 +141,7 @@ enum IntegrationConnectTelemetry {
     case .notSignedIn, .sessionExpired, .noBrowser, .decryptFailed, .authentication:
       return true
     case .configuration, .storeNotFound, .authorizationDenied, .network, .timeout,
-      .cancelled, .noContent, .rateLimit, .permission, .conflict, .invalidResponse,
+      .cancelled, .noContent, .rateLimit, .permission, .conflict, .server, .invalidResponse,
       .resourceExhausted, .unknown:
       return false
     }
--- a/desktop/macos/Desktop/Sources/MainWindow/Pages/ConnectorImportOperations.swift
+++ b/desktop/macos/Desktop/Sources/MainWindow/Pages/ConnectorImportOperations.swift
@@ -91,7 +91,7 @@ enum ConnectorImportOperations {
     case .server:
       return .failure(
         message: "Omi's memory service is temporarily unavailable. Try again shortly.",
-        failureClass: .invalidResponse)
+        failureClass: .server)
     case .invalidResponse:
       return .failure(
         message: "Omi couldn't read the extracted memories. Try the import again.",
--- a/desktop/macos/Desktop/Sources/Onboarding/OnboardingMemoryLogImportService.swift
+++ b/desktop/macos/Desktop/Sources/Onboarding/OnboardingMemoryLogImportService.swift
@@ -177,12 +177,20 @@ actor OnboardingMemoryLogImportService {
     var memories: [String] = []
     var seen: Set<String> = []
     for rawLine in rawText.components(separatedBy: .newlines) {
+      let trimmedLine = rawLine.trimmingCharacters(in: .whitespacesAndNewlines)
+      guard !trimmedLine.isEmpty, !trimmedLine.hasPrefix("```") else { continue }
+
       let range = NSRange(rawLine.startIndex..<rawLine.endIndex, in: rawLine)
       guard
         let match = regex.firstMatch(in: rawLine, range: range),
         let tagRange = Range(match.range(at: 1), in: rawLine),
         let contentRange = Range(match.range(at: 2), in: rawLine)
-      else { continue }
+      else {
+        // A mixed-format paste is not safe to import locally: silently ignoring
+        // its untagged lines could lose durable memories. Let the managed
+        // extractor interpret the complete paste instead.
+        return nil
+      }
 
       let tag = String(rawLine[tagRange]).lowercased()
       let content = String(rawLine[contentRange])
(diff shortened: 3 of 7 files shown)


aryanorastar:
## Verification proof

This follow-up addresses all five actionable Cubic findings left on merged PR #12666:

1. **Mixed-format safety:** a paste containing recognized `[tag]` lines plus any other non-empty content now falls back to the managed extractor with the complete original text. Untagged memories are no longer silently discarded.
2. **Server telemetry:** backend 5xx failures retain a dedicated bounded `server` error class instead of being collapsed into `invalid_response`.
3. **Onboarding partial saves:** onboarding now preserves and displays the failed-item count through the same bounded partial-save message used by the Apps connector.
4. **Test accuracy:** the unstructured-input test name now describes what it proves, and the tagged wrapper test only ignores safe Markdown fences.
5. **User-facing release note:** the changelog now describes the visible import behavior rather than implementation details.

### Automated evidence

- Final rebased commit: `6718318e0224d1da1c2fb1c777487fa900283d7b`
- Focused macOS suites on that exact commit: **75 tests passed, 0 failed**
  - `ConnectorImportOperationsTests`: 15/15
  - `DesktopAutomationSecondaryActionTests`: 36/36
  - `IntegrationConnectTelemetryTests`: 19/19
  - `OnboardingMemoryLogImportServiceTests`: 5/5
- New regression: `testMixedTaggedAndUntaggedPasteDoesNotSilentlyDropContent`
- Original #12442 ten-line ChatGPT export regression remains green.
- Commit-aware PR preflight: **22/22 checks passed**
- Swift format: clean
- SwiftLint: **0 violations across 1,566 files**
- Desktop E2E coverage guard: **4/4 changed production Swift files covered**
- Incremental macOS app build in the pre-push gate: passed
- `git diff --check`: clean

Scope is macOS desktop only. No iOS, Android, Windows, firmware, or hardware behavior is changed.

Physical before/after screenshots can be attached after final app-level validation; the code path and regression boundaries above are already covered locally.

Git-on-my-level:
Verified against the current head (6718318) — nice, tightly-scoped follow-up to #12666. Walking the changed files:

- `OnboardingMemoryLogImportService.swift` — the core change is right: `extractTaggedMemories` now skips blank lines and ``` fences, and returns nil on the first non-empty, non-tagged line, so a mixed paste falls back to the managed extractor with the complete text instead of silently discarding untagged memories. The inline comment explaining why is accurate, and the 2,000-char per-item bound, 200-item cap, dedupe, and `[unknown]` pass-through are all preserved.
- `ConnectorImportOperations.swift` — the `.server` case now carries `failureClass: .server` instead of `.invalidResponse`, with the right retry message. Correct one-liner.
- `IntegrationConnectTelemetry.swift` — new closed-enum `server = "server"` case, and `failureRequiresReconnect` correctly classifies it as non-reauth alongside the other terminal classes. The allow-list and bounded-dimension discipline are untouched, so nothing new can thread content into telemetry. Since `IntegrationConnectTelemetryTests.testFailureRequiresReconnectIsTrueOnlyForReauthClasses` iterates `ErrorClass.allCases`, the new case is automatically covered.
- `OnboardingPagedIntroCoordinator.swift` — partial saves during onboarding now surface the bounded "Imported N …; M couldn't be saved. Import again to retry them." message through the existing `lastActionError` surface. Re-mapping through `memoryLogOutcome` avoids a second result-mapping path; the logic checks out, including the failure branch.
- `ConnectorImportOperationsTests.swift` — `testMemoryLogServerFailureKeepsDistinctTelemetryClass` pins the new telemetry mapping. Good regression.
- `OnboardingMemoryLogImportServiceTests.swift` — the wrapper test input no longer pretends a prose line is ignorable, and `testMixedTaggedAndUntaggedPasteDoesNotSilentlyDropContent` covers the new nil fallback path. Minor: that name promises a bit more than the assertion proves (deferral to the managed extractor, not end-to-end retention) — the Cubic rename suggestion (`...DefersToManagedExtractor`) is reasonable if you touch this file again, not worth a cycle on its own.
- `changelog/unreleased/20260904-memory-import-review-fixes.json` — matches the documented one-`change`-per-file convention with a user-visible description.

On the open Cubic point about mixed pastes over 40,000 characters: I verified `importMemoryLog` sends `String(trimmed.prefix(40_000))` to `APIClient.extractMemoryLog`. The concern is real in principle, but that cap is pre-existing behavior for every unstructured paste — this PR doesn't change it, it just routes mixed pastes into the same bounded path, which is strictly safer than the old silent drop. A very large mixed export can still lose its tail; bounding the local parser to the extractor cap (or raising the cap server-side) would close that, but it's an edge that deserves a maintainer's product call rather than blocking here.

Checks green, changelog included, scope held to the macOS memory-import path. Thanks also for the detailed verification notes. Leaving this for a normal maintainer pass, mainly to weigh the extractor-cap question.

---

kodjima33 (approved):
Follow-up to merged #12666 addressing post-merge Cubic findings (mixed-format import fallback, telemetry class fix, partial-save surfacing). title starts with fix(desktop) -> owner-override qualifies regardless of file-count heuristic. No abuse signals, CI green. Merging.

PR #12712, opened 2026-09-04
fix(backend): fill chat message pages past reported rows

### LINKED ISSUE #12709: Chat message pagination drops visible rows when reported messages sit in the Firestore page
`GET /v2/messages` (and the other `get_messages` callers) can return fewer than `limit` visible messages even when older visible messages still exist, and the next `offset` can advance past rows the client never saw.

## Evidence

In `backend/database/chat.py` `get_messages`:

1. Firestore applies `.limit(limit).offset(offset)` on the raw newest-first stream.
2. Python then skips any row with `reported is True`.

So a reported row in the middle of a raw page steals a slot from a later visible message. Example newest-first stream `[visible-a, reported, visible-b, visible-c]` with `limit=2` returns only `[visible-a]`; `visible-b` is never fetched on that page. Callers that stop after one page (mobile loads offset 0 once) or treat `len < limit` as EOF never see those messages.

The same failure class is already acknowledged elsewhere in this file:

- `get_cache_aligned_messages` over-fetches raw rows specifically because filtering after limit drops visible history.
- `get_messages_reconcile_page` uses a bounded scan budget so reported rows do not consume page capacity.

Offset pagination on `get_messages` itself still does the broken limit-then-filter path.

## Call sites

`/v2/messages`, `/v2/desktop/messages`, MCP `get_chat_messages`, and related helpers all go through this function with `limit`/`offset`.

aryanorastar:
Fixes #12709.

## Summary

`get_messages` applied Firestore `.limit().offset()` before filtering `reported is True` in Python. A reported row in the middle of a raw page stole a visible slot, so pages came back short and later visible messages were left unfetched for single-page callers.

This walks the newest-first stream until `offset` visible rows are skipped and `limit` visible rows are collected, with a bounded scan so one request cannot stream unbounded history.

## Change

- `backend/database/chat.py` — `get_messages` visible-page scan instead of raw limit/offset-then-filter
- `backend/tests/unit/test_get_messages_reported_pagination.py` — behavioral coverage
- `.github/failure-classes/FC-raw-page-limit-before-visibility-filter.json` — new failure class

No router / client API shape change. `get_app_messages` (no offset) and `get_cache_aligned_messages` (already over-fetches) are left alone.

## How the scan is bounded

A page with no reported rows needs exactly `offset + limit` documents — which is what the raw `.offset(n).limit(m)` query already streamed. So the budget **floors at the page** and bounds only the *slack* above it:

```python
needed = max(offset, 0) + max(limit, 0)
scan_budget = needed + min(CHAT_MESSAGES_VISIBLE_PAGE_SCAN_SLACK, needed * 3)
```

Two consequences worth stating, because the first version of this PR got both wrong:

- The scan can never read fewer documents than the query it replaced, so it can never fail to service an offset that used to work.
- Slack is paid only by a page that actually meets a reported row. The first batch reads exactly `needed`.

## Review findings, both fixed

Both were real. Thanks for catching them — the second one in particular was the reviewed bug reintroduced at depth, which is not something I would have found from the passing tests.

**Read amplification** (cubic P2, and Git-on-my-level's first tradeoff). The budget floored at `max(100, ...)` and the first batch was a flat `min(100, budget)`, so a page needing 5 rows streamed 100 documents. `backend/utils/chat.py` reads history at `limit=5/10/15` on the chat-send path, so every send paid ~20x the documents it used to. Fixed by reading exactly `needed` on the first batch.

**Deep-offset truncation** (cubic P2, and Git-on-my-level's second tradeoff). Capping the *total* scan at 1000 meant `offset + limit` beyond the cap could never be reached, so the page came back empty — and `routers/chat.py` reads an empty page as end-of-results. That is the same "later visible messages are never fetched" defect this PR exists to fix, reintroduced at depth, and a straight regression against the raw `.offset()` it replaced. Fixed by the floor above.

On cubic's suggestion of an explicit `has_more` / resume signal: I did not add one. It would change the shape of an endpoint this PR is otherwise transparent to, and with the budget floored at the page it is no longer needed to avoid the EOF masquerade — an offset that used to be serviceable still is. A cursor for `/v2/messages` is worth doing, but as its own change rather than inside a pagination bug fix.

## Verification

Mutation-checked — with the old capped budget and flat first batch restored:

```
E       assert collection.streamed == 5
E       assert 100 == 5
E       AssertionError: assert [] == ['visible-1200', ...]
E         Right contains 5 more items, first extra item: 'visible-1200'
2 failed, 3 passed
```

Both original tests still fail on `main`'s limit-then-filter path with `['visible-a']` instead of `['visible-a', 'visible-b']`.

With the fix:

- `pytest tests/unit/test_get_messages_reported_pagination.py tests/unit/test_chat_message_count.py` — 12 passed
- `pytest tests/unit/test_desktop_migration.py -k MessageReconcileKeyset` — 3 passed

Five tests now cover: reported-in-the-middle, visible offset continuity, a clean page streaming no more documents than it returns, a deep offset still being serviced, and slack staying bounded when every scanned row is reported.

**Rebased onto current `main` (`176a02fe0d`).** The branch was 112 commits behind; the `test_client_processing_trust_boundary.py` failure kodjima33 correctly diagnosed as stale-base was repaired on main in #12729 and is gone from this head.

## Honest gaps

- No live Firestore / emulator integration run (the hermetic fake mirrors offset/limit/start_after including keyset resume).
- Physical device smoke not run — backend-only change, no device path involved.

Failure-Class: new

### DIFF
--- a/.github/failure-classes/FC-raw-page-limit-before-visibility-filter.json
+++ b/.github/failure-classes/FC-raw-page-limit-before-visibility-filter.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": 1,
+  "id": "FC-raw-page-limit-before-visibility-filter",
+  "violated_contract": "A list page whose store query cannot express the caller's visibility rule must measure limit and offset in visible rows. Applying limit/offset on the raw store stream and only then dropping hidden rows (for example reported chat messages) returns a short page and advances past visible rows the client never saw, so single-page callers and len < limit EOF checks stop early while older visible messages still exist.",
+  "canonical_prevention": "Scan the ordered raw stream with a bounded budget, skip hidden rows without consuming page capacity, and stop only when the visible limit is filled or the collection is exhausted — the same pattern get_messages_reconcile_page already uses. Prove it with a hermetic page whose middle row is hidden: the returned page must still contain limit visible ids, and the next visible offset must continue at the next visible row rather than a raw Firestore slot.",
+  "canonical_prevention_artifact": [
+    "backend/tests/unit/test_get_messages_reported_pagination.py"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    "backend/database/chat.py",
+    "backend/routers/chat.py",
+    "backend/routers/chat_sessions.py"
+  ],
+  "status": "open"
+}
--- a/backend/database/chat.py
+++ b/backend/database/chat.py
@@ -271,7 +271,12 @@ def get_messages(
     # A page with no reported rows needs exactly this many documents, which is what the
     # old raw query streamed. Bound the *slack* on top of it, not the page itself.
     needed = max(offset, 0) + max(limit, 0)
-    scan_budget = needed + min(CHAT_MESSAGES_VISIBLE_PAGE_SCAN_SLACK, needed * 3)
+    # Flat slack, not proportional. Scaling it with the page size gave a small page a
+    # tiny allowance (limit=2 -> 6 documents), so a dense run of reported rows still
+    # returned an empty page the router reads as end-of-results. The read cost is set
+    # by the batch sizing below, not by this ceiling, so a flat allowance costs a clean
+    # page nothing and only bounds how far a page that meets reported rows may scan.
+    scan_budget = needed + CHAT_MESSAGES_VISIBLE_PAGE_SCAN_SLACK
     scanned = 0
     visible_skipped = 0
     messages: List[Dict[str, Any]] = []
--- a/backend/database/chat.py
+++ b/backend/database/chat.py
@@ -38,6 +38,9 @@ CHAT_HISTORY_APPEND_EPOCH_MESSAGES = 8
 # when a user has thousands of lifetime reported messages; the newest page
 # rarely contains more reported rows than this cap.
 CHAT_HISTORY_REPORTED_RAW_SCAN_CAP = 50
+# Bound raw Firestore reads when filling a visible get_messages page past
+# reported rows. Same spirit as get_messages_reconcile_page.
+CHAT_MESSAGES_VISIBLE_PAGE_SCAN_CAP = 1000
 
 
 class ClientMessageIdPayloadConflict(ValueError):
@@ -238,6 +241,15 @@ def get_messages(
     app_id: Optional[str] = None,
     chat_session_id: Optional[str] = None,
 ) -> List[Dict[str, Any]]:
+    """Return a visible chat page with offset counted in non-reported rows.
+
+    Firestore cannot apply ``reported != True`` cheaply for legacy docs that omit
+    the field, so reported rows are filtered in Python. Applying ``limit`` /
+    ``offset`` on the raw query first makes pages short and advances past
+    visible messages the client never saw. Scan with a bounded budget (same
+    spirit as ``get_messages_reconcile_page``) until ``offset`` visible rows are
+    skipped and ``limit`` visible rows are collected.
+    """
     logger.info(f'get_messages {uid} {limit} {offset} {app_id} {include_conversations}')
     user_ref = db.collection('users').document(uid)
     messages_ref = user_ref.collection('messages')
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
Thanks @aryanorastar — this fixes a real pagination bug and follows the pattern this codebase already established for the same failure class.

What I verified:

- `backend/database/chat.py` — `get_messages` now scans the newest-first stream with a bounded budget (`min(1000, max(100, (offset+limit)*4))`, 100-doc batches, `start_after` cursor) and counts `offset`/`limit` in visible rows, mirroring `get_messages_reconcile_page`. The cursor advances per document, reported rows don't consume page capacity, and short-batch EOF terminates the scan. Offset semantics stay coherent for clients: page N at `offset = N*L` continues at the exact next visible row, and `get_cache_aligned_messages`' delegate-then-slice still holds since it calls with `offset=0`.
- `backend/tests/unit/test_get_messages_reported_pagination.py` — both tests fail on main's limit-then-filter path and pass with the fix (ran them locally on this head). The fake mirrors Firestore offset/limit/start_after semantics, including keyset resume.
- `.github/failure-classes/FC-raw-page-limit-before-visibility-filter.json` — matches the existing FC schema exactly (same keys as the other 192 entries) and points at the new test as the canonical prevention artifact.

Two tradeoffs worth a maintainer eye (neither blocking from my side):

1. Read amplification on hot paths: the `max(100, …)` budget floor means internal calls with small limits (`backend/utils/chat.py` prev-message reads with limit=5/10/15, and `get_cache_aligned_messages`' over-fetch) now stream a 100-doc first batch where they previously read exactly `limit` docs. Firestore bills streamed documents, so the chat-send path reads ~20× more docs per send. If that matters operationally, a smaller first batch (sized to `limit` + reported slack, then 100s) would keep the fix without the floor cost.
2. Deep-offset truncation: `offset + limit` beyond the 1000-doc budget now returns a short/empty page (e.g. deep `/v2/messages` offsets) where main would have returned the correct page — the same bounded-scan tradeoff `get_messages_reconcile_page` makes, but offset callers have no cursor to resume with. For the pagination depths clients actually use this looks fine; noting it so it lands as a conscious choice.

Leaving for human maintainer review before merge: user-data pagination surface plus the chat-send read-cost tradeoff above. No split requested — the change is cohesive.

---

kodjima33 (approved):
Real fix, confirmed not already fixed on main (get_messages on origin/main still does .limit().offset() before the reported-filter). Root cause clearly stated, scoped diff, new regression tests. Backend unit suite shows one failure (test_client_processing_trust_boundary.py) unrelated to this diff -- confirmed stale-base: that guardrail was independently repaired on main via #12729, and this PR's base is 118 commits behind (fab60317 vs current main). Holding at approve per stale-base policy rather than merging on a red required check; not rebasing the branch.

aryanorastar:
Pushed `6721a84e2a`. Both review findings were real and are fixed; the branch is also rebased onto current `main` (`176a02fe0d`).

**Read amplification.** The budget floored at `max(100, …)` and the first batch was a flat `min(100, budget)`, so a page needing 5 rows streamed 100 documents. `backend/utils/chat.py` reads history at `limit=5/10/15` on every chat send, so that path was paying ~20x the documents it used to.

**Deep-offset truncation.** This one was the more serious of the two: capping the *total* scan at 1000 meant `offset + limit` past the cap could never be reached, the page came back empty, and `routers/chat.py` reads an empty page as end-of-results. That is the exact "later visible messages are never fetched" defect this PR exists to fix, reintroduced at depth — and a straight regression against the raw `.offset()` it replaced.

Both have the same root, and fixing the framing fixes both rather than patching them separately. A page with no reported rows needs exactly `offset + limit` documents, which is precisely what the old raw query streamed. So the budget floors at the page and bounds only the slack above it:

```python
needed = max(offset, 0) + max(limit, 0)
scan_budget = needed + min(CHAT_MESSAGES_VISIBLE_PAGE_SCAN_SLACK, needed * 3)
```

The scan can now never read fewer documents than the query it replaced, so it can never fail to service an offset that used to work; and the first batch reads exactly `needed`, so slack is paid only by a page that actually meets a reported row.

Mutation-checked — with the old capped budget and flat first batch restored:

```
E       assert collection.streamed == 5
E       assert 100 == 5
E       AssertionError: assert [] == ['visible-1200', ...]
2 failed, 3 passed
```

Three new tests: a clean page streams no more documents than it returns, a deep offset is still serviced, and the slack stays bounded when every scanned row is reported. 12 passed across the pagination and message-count suites; `MessageReconcileKeyset` still 3 passed.

On the suggested `has_more` / resume signal — I did not add one. It changes the shape of an endpoint this PR is otherwise transparent to, and with the budget floored at the page it is no longer needed to prevent the EOF masquerade, since an offset that used to be serviceable still is. A cursor for `/v2/messages` is worth doing on its own rather than inside a pagination bug fix.

The `test_client_processing_trust_boundary.py` failure was diagnosed correctly as stale base — it was repaired on `main` in #12740 / #12729 and is gone from this head after the rebase.

Issue #4498, opened 2026-02-01
GET /v3/memories returns 500 when memory record missing 'user_review' field


beastoin:
## Description
Users are unable to load memories in the app due to a KeyError in the backend.

## Root Cause
`backend/database/memories.py:94` uses direct dict access instead of `.get()`:

```python
# Current (buggy):
result = [memory for memory in memories if memory['user_review'] is not False]

# Fix:
result = [memory for memory in memories if memory.get('user_review') is not False]
```

## Impact
- Any user with a memory record missing the `user_review` field gets a 500 error
- Their entire memories list fails to load in the app
- Confirmed via Grafana logs showing multiple 500 errors on `/v3/memories` endpoint

## Steps to Reproduce
1. User has a memory document in Firestore without `user_review` field
2. User opens app and navigates to memories
3. App calls `GET /v3/memories`
4. Backend crashes with `KeyError: 'user_review'`

## Suggested Fix
Use `.get('user_review')` instead of `['user_review']` on line 94 of `backend/database/memories.py`

## Environment
- Endpoint: `GET /v3/memories`
- File: `backend/database/memories.py:94`

---
_Reported by @mon via Grafana monitoring_

Issue #11831, opened 2026-08-18
backend: prevent large list GETs from exhausting the 30s request budget


Git-on-my-level:
## Bug Description

Production Cloud Run `backend` still returns **bare HTTP 504s at exactly ~30.000s** on customer list GETs after the bounded-read work in #10750 and the memories scan work in #11816 / #11819.

This is a **recurrence of the same failure class as #10746**, not a new outage type. Those PRs bounded or reshaped Firestore reads; they did not make large-account list GETs finish well inside `HTTP_GET_TIMEOUT=30`. When a list read occupies a concurrency slot for the full 30s, sibling list GETs on the same instances also 504.

Confirmed customer journeys: **action items, conversation archive, memories**. Chat / listen / GKE pusher were not this incident.

## Incident window

Read-only inspection on 2026-08-18 from `based-hardware` (`read-only-bot-account`). Serving revision `backend-c5b0b5d-32152251884-1` (image tag `c5b0b5d` = #11819), 100% traffic, deployed ~15:33–15:36Z.

| UTC | Signal |
|---|---|
| 20:16 | Monitoring 5xx / 15m = 55 |
| 20:31 | 143 (watchdog paged 99 ERROR logs / 15m vs threshold 44) |
| 20:46 | 13 (dip) |
| 21:01 | 61 (second wave; 6h median ~12.5) |

Traffic stayed ~170k 2xx / 15m. Instances stayed at **5** during the 20:30 spike (concurrency 20, minScale 3, maxScale 100). 5xx rate ≪1% but concentrated on home-screen list GETs. UA class on the 5xx sample was almost all mobile/Omi.

Log sample since 19:30Z (capped; all serving SHA `c5b0b5d`):

| Route | Status | n | Latency |
|---|---|---|---|
| `GET /v1/action-items` | 504 | 95 | exact ~30s |
| `GET /v1/conversations` | 504 | 88 | exact ~30s |
| `GET /v3/memories` | 504 | 61 | exact ~30s |

Second wave 20:45–21:00: action-items 25, conversations 23, memories 7 — **not memories-only**.

`omi-web-app` 32.8s p95 in the same watchdog card is **not** a web-app HTML outage. User pages in the spike window were ~60ms. The 32.8s bucket is `/api/proxy/v3/memories` ~30s 504s (script UA, low RPS, July 19 web-app revision). Do not page that as a separate customer-path failure.

## Confirmed mechanism

[...]

PR #12714, opened 2026-09-04
fix(backend): fill scoring memory pages past hidden rows

### LINKED ISSUE #12713: Bug: memory scoring_desc list pages shrink when rejected/invalid rows sit in the raw Firestore page
## Symptom

Default `get_memories` (`sort='scoring_desc'`) applies Firestore `.limit().offset()` and only then drops user-rejected (`user_review is False`) and invalidated (`invalid_at` set) rows in Python. A hidden row in the middle of a raw page steals a visible slot, so the page comes back short and later visible memories are left unfetched for single-page / `len < limit` EOF callers.

## Evidence (on current `main`)

- Control flow: `backend/database/memories.py` scoring path — `limit`/`offset` on the query, then `_memory_passes_list_visibility` filter.
- Same product rule cannot move into a Firestore `where` without dropping legacy docs that omit `user_review` / `invalid_at` (#4498 / #11915).
- The `updated_desc` path already filters visibility before slicing; scoring_desc does not.
- Hermetic repro before fix: `limit=2` over `[visible-a, rejected, visible-b, …]` returned `['visible-a']` instead of `['visible-a', 'visible-b']`.

## Out of scope

- Chat `get_messages` reported pagination (tracked separately; same failure class).
- Moving visibility predicates into Firestore FieldFilters.

aryanorastar:
Fixes #12713.

## Summary

`get_memories` default `scoring_desc` applied Firestore `.limit().offset()` before filtering user-rejected (`user_review is False`) and invalidated (`invalid_at` set) rows in Python. A hidden row in the middle of a raw page stole a visible slot, so pages came back short and later visible memories were left unfetched for single-page callers.

This mirrors the bounded visible-page scan used for chat `get_messages` reported pagination: walk the scoring order until `offset` visible rows are skipped and `limit` visible rows are collected, with a hard scan cap so one request cannot stream unbounded history. Visibility stays in Python (#4498 / #11915) — no FieldFilter that would drop legacy docs missing the fields.

## Change

- `backend/database/memories.py` — scoring_desc visible-page scan instead of raw limit/offset-then-filter
- `backend/tests/unit/test_get_memories_scoring_visibility_pagination.py` — behavioral coverage for rejected/invalidated-in-the-middle and visible offset continuity
- `.github/failure-classes/FC-raw-page-limit-before-visibility-filter.json` — new failure class for this contract

No router / client API shape change. The `updated_desc` path already filtered visibility before slicing and is unchanged.

- INV-MEM-1

Failure-Class: FC-raw-page-limit-before-visibility-filter

The class was `new` when this PR opened. Sibling #12712 has since merged and landed
the definition on `main`, so this is now the second recorded instance of an existing
class rather than its birth — the memories `scoring_desc` path alongside the chat
`get_messages` path. The registry edit here is the minimum an instance fix is allowed: this PR's test is
appended to `canonical_prevention_artifact`, and every other field is exactly `main`'s.
My first resolution also merged the `violated_contract` wording, `scope_hints` and
`evidence_prs` -- `scripts/failure-class validate` rejected that with
`instance_fix_mutates_registry`, and it is right to. An instance PR quietly rewriting
shared contract text is how a registry stops meaning the same thing to everyone.

Those wider edits are still worth making, as a registry-only lifecycle PR: the
`violated_contract` wording naming both hidden-row kinds, `scope_hints` covering the
memories paths, and `evidence_prs` citing 12712 and 12714 rather than staying empty.

## Verification

- Before the fix, the new tests failed on main with `['visible-a']` instead of `['visible-a', 'visible-b']` (short page after visibility filter).
- After the fix:
  - `pytest tests/unit/test_get_memories_scoring_visibility_pagination.py tests/unit/test_memories_user_review.py` → 13 passed (with query-contract index check)
  - `pytest tests/unit/test_memory_offset_read_cost.py tests/unit/test_list_read_budget_contract.py -k memor` → 8 passed

## Honest gaps

- No live Firestore / emulator integration run (hermetic fakes only).
- Physical / on-device smoke not run (no kit); backend-only change.
- Open sibling PR #12712 may land the same failure-class id for chat; if both merge, combine `canonical_prevention_artifact` lists in the FC JSON.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12714?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Line-Count-Exception: backend/database/memories.py | 1459 -> 1527 | The visible-page scan replaces a three-line raw limit/offset slice with a cursor loop, and carries the comments recording why visibility cannot move into a FieldFilter (#4498) and why the budget floors at the page. The last 15 lines are the budget-exhaustion warning: a page that stops on its own ceiling is indistinguishable from end-of-data to the caller, so the loop says which one happened. Splitting this file is a larger refactor than a pagination bug fix should carry.


Git-on-my-level (dismissed):
Thanks @aryanorastar — the diagnosis is right and the core fix is well executed. Verified per file:

**backend/database/memories.py** — The visible-scan (counting visible_offset skips / visible_limit fills over a start_after cursor) correctly fixes both failure modes of raw `.limit().offset()` + post-filter: short pages and raw-slot offset advancement. Keeping visibility in Python (no `user_review`/`invalid_at` FieldFilters) preserves legacy-doc semantics per #4498. One blocking item and two design notes:

1. **Blocking — this head breaks a pinned contract.** `backend/testing/contracts/test_desktop_backend_parity.py::test_python_memory_query_and_filter_semantics` now fails (`assert 100 == 11`): the fake pins `limit_value == data["limit"]` / `offset_value == data["offset"]` on the query, which the visible-scan intentionally no longer applies (it batches `.limit(min(100, budget))` and never calls `.offset()`). The Desktop Backend Contracts workflow is red on this head while neighboring PR heads are green, so merging as-is would break the contract gate on main. Please update the parity contract/fixture for the new query shape in this PR (same deliberate-pin-update flow the repo uses elsewhere).
2. **Read-cost on a path with 504 history.** The scan streams full documents (encrypted `content` included) for up to `min(2000, max(100, (offset+limit)*4))` rows, while the `updated_desc` path in this same file was redesigned to metadata-only windows + per-page id fetches with `budgeted_stream_list` + per-RPC timeouts (#11831) after the 2026-08-18 GET /v3/memories 504 (comment above `list_memory_updated_or_created_index`). Skipped rows only need `user_review`/`invalid_at`, so a `.select()` projection for the scan (full docs only for the returned page) and/or a `ListReadBudget` would match the file's own bounding discipline.
3. Once `scanned == scan_budget` with the page unfilled, callers get a short page indistinguishable from EOF — the same failure class at a deeper boundary (relevant around `visible_offset + limit ≥ 2000` with dense hidden rows). A log line (or continuing while the last batch came back full) would make the truncation observable.

**backend/tests/unit/test_get_memories_scoring_visibility_pagination.py** — Good behavioral pinning of rejected/invalidated mid-page and visible-offset continuity, and the fake's offset+limit vs start_after semantics are faithful. Two gaps: (a) with `scan_budget = max(100, (offset+limit)*4)` and ≤6 fixture rows, the first batch always covers everything, so the `start_after` multi-batch branch — the actual fix mechanism — never executes; the docstring's "start_after scan (fixed) semantics" currently overstates coverage. A fixture with >100 raw rows (or an injectable batch size) would pin it. (b) `patch.object(memories_db, '_prepare_memory_for_read', ...)` is a no-op — `@prepare_for_read` bound the original into the wrapper closure at decoration time, so patching the module attribute never changes decryption; the tests pass because no row uses `data_protection_level='enhanced'`. Dropping it (or noting why) avoids implying decryption is stubbed.

**.github/failure-classes/FC-raw-page-limit-before-visibility-filter.json** — Follows the established failure-class schema; the contract statement is accurate and the scope hints are sensible. Minor: `evidence_prs` is empty — worth referencing the report PR/issue behind #12713 once linked.

CI fairness note: the "Backend unit suite" red is pre-existing on main (`test_pinned_conversation_dump_set_is_exact`, `utils/llm/external_integrations.py` — untouched here), not caused by this PR. The blocking item is only the contracts gate above.

Requesting changes just for the parity-contract update; the read-budget alignment and multi-batch test can land in the same push. Leaving the scan read-budget design (full-doc scan vs metadata-window pattern) for maintainer sign-off given the documented 504 history on this endpoint — happy to re-review.

— automated review on behalf of the Omi maintainers (AI-generated).

---

aryanorastar:
Pushed `1189368f62`, rebased onto current `main` (`176a02fe0d`). All three items addressed.

**The blocking parity contract.** You were right that this head broke it, and the fix is not a fixture bump. The contract pinned `limit_value == data["limit"]` and `offset_value == data["offset"]` — that a raw Firestore limit/offset was applied to the scoring stream. That is the defect itself, so keeping the pin would have pinned the bug. It now asserts the visible-page contract: `offset_value is None`, and a first read of exactly `offset + limit`.

Updating it turned up a second problem in the same test. The old `result == ["active"]` passed only because `FakeQuery` **recorded** `offset_value` without applying it — the raw offset was asserted but never exercised, so the contract could not distinguish a correct page from a mis-paginated one. With real visible-offset semantics the fixture's offset of 4 is past the only visible row, so the correct page is empty. That is now asserted as `[]`, with visibility filtering pinned separately at offset 0 where it is actually reachable.

**Read cost and truncation-as-EOF.** Both had the same root, and one reframing fixes both. A page with nothing hidden needs exactly `offset + limit` documents — which is what the old raw query already streamed — so the budget floors at the page and bounds only the slack above it:

```python
needed = visible_offset + visible_limit
scan_budget = needed + _MEMORY_SCORING_VISIBLE_PAGE_SCAN_SLACK
```

The first batch reads exactly `needed`, so the 100-document floor is gone and a clean page streams no more than it returns. And because the floor is the page, a deep offset can never be unreachable — the scan never reads fewer documents than the query it replaced.

**Your multi-batch request caught a live bug, not just missing coverage.** My first version scaled the slack with the page (`needed * 3`), which gives `limit=2` a six-document allowance — so a dense run of hidden rows still returned an empty page. That is the truncation-as-EOF defect surviving at a different threshold, and only the 150-hidden-row fixture exposed it. The slack is now flat; read cost is set by the batch sizing, not by that ceiling, so a flat allowance costs a clean page nothing. The same correction is on the chat sibling #12712, which had shipped the identical proportional slack.

**Test gaps.** 150 hidden rows now force the `start_after` continuation no earlier fixture reached. The `patch.object(memories_db, '_prepare_memory_for_read', ...)` is gone, with a comment recording why it never worked — `@prepare_for_read` binds the original into its wrapper closure at decoration time.

**`.select()` projection — not landed, deliberately.** You left the scan read-budget design for maintainer sign-off, and the flooring changes the numbers that decision rests on: a clean page now streams exactly `offset + limit` documents rather than 100, so the remaining exposure is dense-hidden-row scans only. Doing it properly means projecting the scan and hydrating the emitted page through `_fetch_memory_docs_by_ids` — the pattern `updated_desc` already uses — which is a real change to both fakes. Say the word and it goes in this PR; otherwise it is a clean follow-up.

Suites, run individually: pagination 7, `test_memories_user_review` 9, `test_memory_offset_read_cost` 5, `test_list_read_budget_contract` 34, `test_desktop_backend_parity` 8.

aryanorastar:
Merged current `main` (`09ff17e4e5`) on `c9766a3a7a`. The conflict was the one this PR's description predicted, and it needed a real decision rather than picking a side.

**`FC-raw-page-limit-before-visibility-filter.json` was an add/add conflict.** #12712 merged and brought its own copy of the same failure-class id. Taking either side wholesale would have dropped the other instance's guard from the registry — which defeats the point of the record, since a class that names only one artifact reads as a one-off rather than a pattern with two known occurrences.

My first resolution merged the two records wholesale — both artifacts, union of `scope_hints`, combined `violated_contract` wording, `evidence_prs` citing both PRs. **`scripts/failure-class` rejected it**, and correctly:

```
instance_fix_mutates_registry: an instance-fix PR must not edit failure-class
definitions except to add or update 'canonical_prevention_artifact' on the
declared class
```

That rule is right and I had not internalised it. An instance PR quietly rewriting shared contract text is how a registry stops meaning the same thing to everyone who greps it. The record is now `main`'s byte for byte, plus this instance's test appended to `canonical_prevention_artifact` — the one field an instance fix owns.

The wider edits are still worth making, as a **registry-only lifecycle PR**: `violated_contract` wording naming both hidden-row kinds, `scope_hints` covering the memories paths, and `evidence_prs` citing 12712 and 12714 rather than staying empty. Happy to send that separately.

**The declaration also changed.** It said `Failure-Class: new`, which was true when this opened. #12712 merged during review and landed the definition, so this is now the second recorded instance of an existing class rather than its birth — declared as `FC-raw-page-limit-before-visibility-filter`.

**Verification after the merge**

- `scripts/failure-class validate --pr-body-file` — `OK: validate` (declaration=FC-raw-page-limit-before-visibility-filter)
- `test_get_memories_scoring_visibility_pagination.py` — 7 passed
- `test_memories_user_review.py` — 9 · `test_memory_offset_read_cost.py` — 5
- `testing/contracts/test_desktop_backend_parity.py` — 8 passed
- `test_get_messages_reported_pagination.py` — 6 passed, now that the chat sibling arrives from `main`

Merged rather than rebased so the review history on the earlier heads stays intact. The `memories.py` scan is unchanged by the merge — `needed = visible_offset + visible_limit`, `scan_budget = needed + _MEMORY_SCORING_VISIBLE_PAGE_SCAN_SLACK`, flat slack — and 0 behind `main`.

The three review items from your last pass remain addressed; this only resolves the conflict.

kodjima33 (approved):
Confidence gate: root cause stated + fixes #12713 (confirmed still open/unfixed on main) + new regression test = 3/5; diff is 371 lines (over the 300-line scoped-diff cap) and Backend unit suite is red on test_batch_upload_storage.py/test_merge_validation.py — neither touched by this diff, and identically red on sibling #12714... er #11183 while main's own latest runs are green, so this looks like the known stale-frozen-merge-ref CI artifact, not a real regression (reran once per policy, same result). The previously-blocking pinned parity-test failure (test_python_memory_query_and_filter_semantics) is now fixed — contracts check passes. Below the 4/5 merge bar either way; approving only, not merging.

Git-on-my-level:
Re-verified on the current head 6119bdb — everything from the earlier change request is resolved, and that review is now dismissed. Per file:

**backend/database/memories.py** — The budget now floors at the page (`needed = visible_offset + visible_limit`) with a flat 2000-document slack, so a clean page streams exactly what it returns (the 100-document first-batch floor is gone) and a deep offset can never become unreachable — that closes both the read-cost concern and the truncation-as-EOF-at-depth regression the intermediate total-cap design had. The 100-doc continuation batches via `start_after` match the fixture's batch telemetry, and the `updated_desc` branch is untouched.

**backend/testing/contracts/test_desktop_backend_parity.py** — The deliberate pin update is the right call. The old fake recorded `offset_value` without ever applying it, so the previous `result == ["active"]` assertion genuinely could not distinguish a correct page from a mis-paginated one. Pinning `offset_value is None` + `limit_value == offset + limit`, asserting `[]` at the fixture's unreachable offset, and re-pinning visibility at offset 0 is strictly stronger than what it replaces. Green on this head in the contracts job.

**backend/tests/unit/test_get_memories_scoring_visibility_pagination.py** — The multi-batch gap is closed: 150 consecutive rejected rows force the `start_after` continuation (the actual fix mechanism) and assert more than one raw read. The no-op `patch.object` stub is gone with a comment recording why it never worked. All 7 tests pass on this head.

**.github/failure-classes/FC-raw-page-limit-before-visibility-filter.json** — Verified against `main`: the only substantive delta is the appended `canonical_prevention_artifact` entry, exactly the field an instance fix owns (the `canonical_prevention` wording delta is a `\u2014` escaping normalization of identical text).

On the red **Backend unit suite**: I pulled the log — both failures (`tests/unit/test_batch_upload_storage.py`, `tests/unit/test_merge_validation.py`) are `ModuleNotFoundError: No module named 'models.client_processing'` in files this PR doesn't touch, and the identical failure reproduces at this PR's exact base commit `09ff17e4e5` on `main` (fixed on `main` later that day). It's inherited base breakage, not this diff; merging current `main` will clear it.

Two non-blocking notes:
- The deferred `.select()` projection + per-page id hydration (the `updated_desc` metadata-window pattern) remains a worthwhile follow-up; with the page-floored budget the residual exposure is dense-hidden-row scans only.
- A page that exhausts the full `needed + slack` budget mid-hidden-run still reads as EOF to callers. Bounded scans can't fully avoid that edge; a log line at budget exhaustion would make it observable in production.

Thanks @aryanorastar — the parity-contract root-cause writeup (the fake that asserted an offset it never applied) and the failure-class merge resolution under the instance-fix rule were both excellent contributions beyond the fix itself.

---

aryanorastar:
Pushed `a2b503a` — merged current `main`, and took your second non-blocking note.

**The red check.** Merged `main`, which carries #12785. Your diagnosis was exactly right: both failures were `ModuleNotFoundError: No module named 'models.client_processing'` in `test_batch_upload_storage.py` and `test_merge_validation.py`, neither of which this PR touches, reproducing at this PR's base commit and fixed on `main` later that day. Nothing in the diff needed to change for it.

**Budget exhaustion is now observable.** You put it well — a bounded scan can't avoid that edge, only make it visible. The loop now logs when it stopped on its own ceiling rather than on the end of the collection, carrying the offset, limit, scanned count and budget, which together say whether the slack is wrong for that account's data:

```python
if scanned >= scan_budget and len(result) < visible_limit:
    logger.warning('get_memories_scan_budget_exhausted offset=%s limit=%s scanned=%s budget=%s returned=%s', ...)
```

The guard is deliberately exact rather than "the page came back short". Warning on every short page would report ordinary end-of-data as a budget failure, and a warning that fires on the common case is one nobody reads — which would cost the signal you asked for. Both directions are pinned: `test_a_page_that_dies_on_the_budget_says_so` and `test_a_page_that_simply_ran_out_of_rows_stays_quiet`. Mutation-checked — making the condition unconditional fails the quiet test; downgrading the log fails the loud one.

Left the deferred `.select()` projection alone. It's a real improvement but it's a different change from a pagination fix, and with the page-floored budget the residual exposure is dense-hidden-row scans only, as you said.

**Verification:** `test_get_memories_scoring_visibility_pagination.py` 9 passed. `test_desktop_backend_parity.py` 8 passed — worth noting that `main`'s version of that file fails 1 of those same 8, and this PR's pin update is what fixes it. PR preflight passed: 28 checks.

One thing I found while verifying, not a blocker and not from this diff: running the parity contract and `tests/unit/test_batch_upload_storage.py` in the same pytest process errors all 27 of the latter with `AttributeError: <module 'utils.other.storage'> has no attribute 'storage_client'`. It reproduces with `main`'s copy of the contract file, so it is pre-existing, and CI never sees it because contracts and unit run as separate jobs. It is the same shape as #12779 though — a suite that stubs a module and leaks — so it is a latent trap worth an issue if you agree.

kodjima33 (approved):
Confirmed: raw limit/offset before visibility filter is still on main (memories.py:424) — real bug, not already fixed. Bounded visible-page scan is well-tested (13+8 passing), scoped fix, CI green. Merging.

PR #12716, opened 2026-09-04
SCA-443: LocalInferenceService protocol, localhost adapter, and fail-closed kill switches (S9)

Git-on-my-level:
## S9 — LocalInferenceService protocol, localhost adapter, fail-closed kill switches

Shard S9 of the local-models free-tier program (`60-shards.md` row S9; spec `20-desktop-local-runtime.md` §2.1; deterministic minimum `10-backend-plumbing.md` §1.7; flags `50-rollout-and-verification.md`). Shipped Swift macOS app only (`desktop/macos/`), not the rewrite WebKit shell. **Dark:** nothing in the conversation finalization path is wired yet (S11).

Closes SCA-443

### What changed

- **`LocalInferenceService`** — `generateStructured` + `runToolLoop`. Engine IDs: `local-server` and `afm` (named so `forceLocalInferenceEngine=afm` is a closed door until S12, not a fall-through to luna).
- **`LocalInferenceRuntime`** — fail-closed coordinator. Ladder is selected local engine → retry once on that same engine → deterministic minimum. A second registered engine is never consulted. Kill switches short-circuit before HTTP.
- **`LocalServerInferenceAdapter`** — OpenAI-compatible `POST …/chat/completions` with `json_schema`. Loopback-only (`localhost`, `::1`, `127.0.0.0/8`). Non-loopback URLs throw before any HTTP. Tool loop throws `.capabilityUnavailable("tool_loop")`.
- **Kill switches** (env or UserDefaults, Parakeet-style): `OMI_DISABLE_LOCAL_INFERENCE` / `disableLocalInference`, `OMI_FORCE_LOCAL_INFERENCE_ENGINE` / `forceLocalInferenceEngine`, plus URL/model defaults `http://127.0.0.1:11434/v1` and `local`.
- **Deterministic minimum (client §1.7)** — first sentence, ≤60 characters on a word boundary; empty transcript → `{source} · h:mm a` from `startedAt` only; overview `""`; category `other`; `processingState` `none`.
- **Fallback telemetry** — `recordFallback(area: "local_llm", …)`. Allowlisted `local_llm` and `engine_failed`.
- **E2E ratchet** — `recording-finalization.yaml` `covers:` names the S9 sources so the coverage gate holds until S11 wires finalization. The flow does not call the runtime yet.

### Automatic-or-dead check

`LocalInferenceRuntimeFailClosedTests.testForcedEngineFailureReturnsDeterministicMinimumRecordsFallbackAndDoesNotCallCloud`: a throwing `local-server` engine plus an AFM spy. Forced failure yields the deterministic minimum, emits `recordFallback` (`area=local_llm`, `reason=engine_failed`, `outcome=exhausted`, `to=deterministic_minimum`), retries once on the same engine (`generateCallCount == 2`), and the AFM spy stays at `callCount == 0`.

Also covered: disable skip, force `afm` with no adapter (`capability_mismatch`, local-server not called), unknown engine `luna`, retry recover without touching another engine, tool-loop stub, loopback-only HTTP (zero requests to `api.openai.com` / Google), title purity (`1_704_140_040` UTC → `Recording · 8:14 PM`), and a source-inspection tripwire that the fail-closed path cannot name a cloud LLM client.

### Proof

- `xcrun swift test --package-path Desktop --filter LocalInference…` — **20 tests, 0 failures** (host Xcode 26.6 / macOS 26). CI pins **Xcode 16.4** (`/Applications/Xcode_16.4.app` is not installed here), so this is not the CI toolchain.
- `python3 desktop/macos/scripts/check_desktop_test_quality.py` — OK (source-inspection site annotated).
- `desktop/macos/scripts/swift-format-wrapper.sh lint` on the new files — OK.

### Cut list

- AFM adapter (S12).
- Tool loops (stub / feature-detect only).
- Bundling or starting a local runtime.
- Wiring into `ConversationFinalizationService` (S11).

### Line-count ratchet

Line-Count-Exception: desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift | 1521 -> 1523 | allowlist local_llm area and engine_failed reason so S9 fallback telemetry is queryable

### Product invariants

none (checker: no locked invariant matched these paths)

### Failure class

No `fix:` commits in this range.

Failure-Class: none

### Plan notes

- Program HANDOFF said not to start desktop until S4. SCA-443 overrode that ("S9 has no dependencies — start immediately"). This PR is the runtime port only and does not stack on S4. Current `origin/main` now includes S4 (#12661); this branch fast-forwarded onto that tip without taking a dependency.
- Server `_build_deferred_structured` is first 8 words / title `"Recording"`; this client follows §1.7. S3 owns server alignment.
- Retry-once lives in `LocalInferenceRuntime` (S9 choke point), not ConversationFinalization (S11).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12716?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift
+++ b/desktop/macos/Desktop/Sources/DesktopDiagnosticsManager.swift
@@ -1433,6 +1433,7 @@ final class DesktopDiagnosticsManager {
     "task_workflow",
     "auth_storage",
     "state_authority",
+    "local_llm",
     "ptt_input_routing",
     "account_cutover",
     "other",
@@ -1468,6 +1469,7 @@ final class DesktopDiagnosticsManager {
     "db_backoff",
     "state_divergence",
     "status_inferred",
+    "engine_failed",
   ]
 
   private func bucketFallbackArea(_ area: String) -> String {
--- a/desktop/macos/Desktop/Sources/LocalInference/DeterministicConversationMinimum.swift
+++ b/desktop/macos/Desktop/Sources/LocalInference/DeterministicConversationMinimum.swift
@@ -1,0 +1,92 @@
+import Foundation
+
+/// Client-side copy of W1 §1.7. Pure: same input, same title; the only clock
+/// is `startedAt`. This is the fail-closed result when no local engine can run.
+/// Overview stays empty — never a fabricated summary that looks like content.
+struct DeterministicConversationMinimum: Sendable, Equatable, Codable {
+  var title: String
+  var overview: String
+  var category: String
+  var processingState: String
+
+  static let emptyOverview = ""
+  static let requiredCategory = "other"
+  static let terminalProcessingState = "none"
+  static let titleCharacterBudget = 60
+
+  static func make(from input: DeterministicMinimumInput) -> DeterministicConversationMinimum {
+    DeterministicConversationMinimum(
+      title: Self.title(from: input),
+      overview: emptyOverview,
+      category: requiredCategory,
+      processingState: terminalProcessingState
+    )
+  }
+
+  static func title(from input: DeterministicMinimumInput) -> String {
+    let sentence = firstNonEmptySentence(input.transcript)
+    if sentence.isEmpty {
+      return fallbackTitle(sourceLabel: input.sourceLabel, startedAt: input.startedAt, timeZone: input.timeZone)
+    }
+    return truncateToWordBoundary(sentence, maxCharacters: titleCharacterBudget)
+  }
+}
+
+struct DeterministicMinimumInput: Sendable, Equatable {
+  var transcript: String
+  var startedAt: Date
+  var sourceLabel: String
+  var timeZone: TimeZone
+
+  init(
+    transcript: String,
+    startedAt: Date,
+    sourceLabel: String = "Recording",
+    timeZone: TimeZone = .current
+  ) {
+    self.transcript = transcript
+    self.startedAt = startedAt
+    self.sourceLabel = sourceLabel
+    self.timeZone = timeZone
+  }
+}
+
+private func firstNonEmptySentence(_ transcript: String) -> String {
+  let collapsed =
+    transcript
+    .replacingOccurrences(of: "\r\n", with: "\n")
+    .components(separatedBy: .newlines)
+    .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+    .filter { !$0.isEmpty }
+    .joined(separator: " ")
+  guard !collapsed.isEmpty else { return "" }
+
+  var sentence = ""
+  for character in collapsed {
+    sentence.append(character)
+    if character == "." || character == "!" || character == "?" {
+      break
+    }
+  }
+  return sentence.trimmingCharacters(in: .whitespacesAndNewlines)
+}
+
+private func truncateToWordBoundary(_ text: String, maxCharacters: Int) -> String {
+  if text.count <= maxCharacters { return text }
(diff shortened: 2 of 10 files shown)


Git-on-my-level:
Reviewed against the program's non-negotiables (fail closed; projection is display not a claim; automatic-or-dead check). The ladder itself is sound — I traced `selectEngine` / `invoke` / `failClosed` and there is no path to a managed provider, `forceLocalInferenceEngine=afm` is a closed door rather than a fallthrough, and `testLocalInferenceSourcesDoNotNameACloudLLMClient` is a real static contract rather than a vacuous one.

Two things block merge, both in the same place: the loopback guarantee is enforced one layer too high.

**1. `URLSession.shared` follows redirects, so the loopback check can be walked past.**

`LocalServerInferenceAdapter.generateStructured` validates `configuration.baseURL` and then hands the request to `URLSessionLocalInferenceHTTPClient`, which is `URLSession.shared`. With no delegate, URLSession follows up to 20 redirects automatically, and a `307`/`308` preserves the method and body. A process listening on `127.0.0.1:11434` — which is exactly what this adapter is designed to talk to, and which is not necessarily the runtime the user thinks it is — can answer the POST with `307 Location: https://api.openai.com/v1/chat/completions` and the full transcript prompt is re-sent there under the caller's network identity. `LocalInferenceLoopback.requireLoopback` never sees that URL. The fake-client tests can't reach this because they replace the transport.

Fix: give the adapter its own `URLSession` with a delegate implementing `urlSession(_:task:willPerformHTTPRedirection:newRequest:completionHandler:)` that either refuses every redirect (simplest, and correct for a local OpenAI-compatible server) or re-runs `requireLoopback` on `newRequest.url` before allowing it. While there, pin the scheme to `http`/`https` and consider dropping `localhost` in favour of literal `127/8` and `::1` — `localhost` is resolver-dependent and `/etc/hosts` is writable by anything with admin.

**2. `invoke`'s retry is unclassified.**

`LocalInferenceRuntime.invoke` retries after *any* thrown error. That includes `.nonLoopbackBaseURL` (a policy refusal, executed twice), `.capabilityUnavailable`, JSON decode failures, permanent `4xx` from `httpStatus`, and `CancellationError`. Retrying a policy denial is the one that matters: a fail-closed component should not attempt a refused operation a second time.

Fix: a small classifier — retry only transport failures, `429`, and `5xx`; fail closed immediately on policy, config, capability, decode, other `4xx`, and cancellation. `reason(for:)` already carries most of that taxonomy, so it is a switch away.

**Not blocking, noted:**

- The e2e `covers:` entries in `recording-finalization.yaml` claim coverage the flow explicitly does not execute. The comment is honest about it, but it does weaken the ratchet — I would rather add those five files in S11, when the flow actually reaches them.
- `DeterministicConversationMinimum` and the backend minimum disagreed when this PR was written (backend used `_build_deferred_structured`'s first-eight-words title). That is being fixed on the backend side in S3, which implements §1.7's first-sentence-to-60-on-a-word-boundary and the `"Recording · 3:14 PM"` fallback — after S3 the two algorithms agree. Worth a shared fixture eventually; not this PR.

Findings 1 and 2 surfaced by a `gpt-5.6-sol` review pass and confirmed by reading the code.

PR #12717, opened 2026-09-04
fix(backend): gate app generate and icon routes on chat quota

### LINKED ISSUE #12715: Free-tier chat quota not enforced on /v1/app/generate and /v1/app/generate-icon
## Symptom

A free user past the monthly chat quota can still hit billable AI app generation and DALL-E icon generation.

`POST /v1/app/generate` and `POST /v1/app/generate-icon` wrap spend in `track_usage(..., Features.APP_GENERATOR)` but never call `enforce_chat_quota`. Sibling routes in the same router already gate:

- `POST /v1/app/generate-description`
- `POST /v1/app/generate-description-emoji`

Those two call `enforce_chat_quota(uid, platform=X-App-Platform)` before the same `APP_GENERATOR` tracker (see `a0cbbd90a5`, which wired description generators and left generate/generate-icon ungated).

`track_usage` only attributes spend; it does not block past-cap callers.

## Evidence (main @ fab60317)

- `backend/routers/apps.py` — description endpoints call `enforce_chat_quota`; `generate_app_endpoint` / `generate_app_icon_endpoint` do not.
- Same billable feature enum: `Features.APP_GENERATOR`.

Out of scope for the first fix (same file, different product surfaces): `/v1/app/generate-prompts` (rate-limit only) and `/v1/personas/twitter/initial-message` (`Features.PERSONA`).

aryanorastar:
Fixes #12715.

## Summary

`POST /v1/app/generate` and `POST /v1/app/generate-icon` bill managed LLM/image spend under `Features.APP_GENERATOR` via `track_usage`, but never called `enforce_chat_quota`. Sibling description generators in the same router already gate free users past the monthly chat cap with a 402.

This adds the same `enforce_chat_quota(uid)` call before provider work on those two endpoints.

## Change

- `backend/routers/apps.py` — `enforce_chat_quota` on generate + generate-icon
- `backend/tests/unit/test_free_quota_gate_wiring.py` — blocked + allowed wiring coverage for both routes
- `.github/failure-classes/FC-sibling-managed-llm-missing-chat-quota-gate.json` — new failure class

Left alone on purpose: `/v1/app/generate-prompts`, `/v1/personas/twitter/initial-message`. No OpenAPI/signature change (platform header not added) so free-tier gating lands without generated-client churn.

## Verification

- Code on `main` (`fab60317`): description endpoints call `enforce_chat_quota`; generate/generate-icon only `track_usage` (no quota). `track_usage` does not enforce.
- After fix: `pytest tests/unit/test_free_quota_gate_wiring.py::TestAppGeneratorGates` → **6 passed**; full wiring file → **15 passed**

## Honest gaps

- No live free-tier Firebase/Redis integration run (hermetic gate patch only).
- Does not pass `X-App-Platform` into the gate (description siblings do); free monthly cap still enforces. Platform-aware trial paywall on these two routes can follow later if product wants parity.
- Physical device smoke not run (no kit; backend-only change).

Failure-Class: new

Line-Count-Exception: backend/routers/apps.py | 2454 -> 2462 | reason: add enforce_chat_quota calls to generate and generate-icon to match sibling description generators; no safe split for a two-call-site wiring fix

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12717?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-sibling-managed-llm-missing-chat-quota-gate.json
+++ b/.github/failure-classes/FC-sibling-managed-llm-missing-chat-quota-gate.json
@@ -1,0 +1,17 @@
+{
+  "schema_version": 1,
+  "id": "FC-sibling-managed-llm-missing-chat-quota-gate",
+  "violated_contract": "User-initiated managed LLM/image endpoints that share the same billable feature (for example Features.APP_GENERATOR via track_usage) must call enforce_chat_quota before spend. Leaving a sibling route ungated while related routes in the same router already enforce chat quota lets free users past FREE_CHAT_QUESTIONS_PER_MONTH trigger Omi-paid generation.",
+  "canonical_prevention": "At every user-triggered APP_GENERATOR (or equivalent tracked) entrypoint, call enforce_chat_quota(uid, platform=X-App-Platform) before track_usage / provider work — matching generate-description and generate-description-emoji. Pin the wiring with hermetic router tests that assert a 402 from the gate prevents the LLM/image helper from being awaited.",
+  "canonical_prevention_artifact": [
+    "backend/tests/unit/test_free_quota_gate_wiring.py"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    "backend/routers/apps.py",
+    "backend/utils/subscription.py",
+    "backend/utils/llm/usage_tracker.py",
+    "backend/tests/unit/test_free_quota_gate_wiring.py"
+  ],
+  "status": "open"
+}
--- a/backend/routers/apps.py
+++ b/backend/routers/apps.py
@@ -1588,6 +1588,10 @@ async def generate_app_endpoint(data: GenerateAppRequest, uid: str = Depends(aut
     """
     from utils.llm.app_generator import generate_app_from_prompt
 
+    # User-initiated LLM generation — same free-tier gate as chat (402 past cap).
+    # Sibling description generators already enforce; this path was left ungated.
+    enforce_chat_quota(uid)
+
     prompt = data.prompt.strip()
     if not prompt:
         raise HTTPException(status_code=422, detail='Prompt is required')
@@ -1627,6 +1631,10 @@ async def generate_app_icon_endpoint(data: GenerateAppIconRequest, uid: str = De
     """
     from utils.llm.app_generator import generate_app_icon
     import base64
+
+    # User-initiated image generation — same free-tier gate as chat (402 past cap).
+    # Sibling description generators already enforce; this path was left ungated.
+    enforce_chat_quota(uid)
 
     app_name = data.name.strip()
     app_description = data.description.strip()
--- a/backend/tests/unit/test_free_quota_gate_wiring.py
+++ b/backend/tests/unit/test_free_quota_gate_wiring.py
@@ -3,8 +3,9 @@
 Regression for the enforcement-coverage gap where a free user past
 FREE_CHAT_QUESTIONS_PER_MONTH could still trigger managed LLM spend through
 daily-summary regeneration, goal progress extraction, app description
-generation, and the omni realtime relay. Each test proves the endpoint calls
-the shared gate BEFORE doing LLM work and lets the 402 propagate.
+generation, AI app/icon generation, and the omni realtime relay. Each test
+proves the endpoint calls the shared gate BEFORE doing LLM work and lets the
+402 propagate.
 
 `enforce_chat_quota`'s own decision logic is covered in test_chat_quota.py;
 these tests only pin the call-site wiring, so the gate is patched at each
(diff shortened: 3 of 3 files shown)


Git-on-my-level:
Thanks @aryanorastar — verified end to end; this closes a real managed-spend leak and matches the router's existing gate architecture. Per-file notes:

**backend/routers/apps.py** — Correct. On main, `generate_app_endpoint` and `generate_app_icon_endpoint` wrap spend in `track_usage(uid, Features.APP_GENERATOR)` but never call `enforce_chat_quota`, while the sibling description generators gate before the same tracker — `track_usage` only attributes usage, it doesn't block. Placing `enforce_chat_quota(uid)` before validation and provider work is right, and the 402 propagates cleanly. I also verified the monthly free-cap path is platform-independent: `get_chat_quota_snapshot` enforces `used < limit` regardless of the platform argument.

Two notes on the earlier bot review's P2s, both non-blocking:

1. **`platform` not passed** (parity with the description siblings): with `platform=None`, `is_trial_paywalled` returns False, so the desktop 3-day trial paywall can never fire on these two routes. Today that's moot (`TRIAL_PAYWALL_ENABLED` defaults to false), and threading the header through would change the endpoint signatures/OpenAPI — the scoping call in "Honest gaps" is reasonable. A small follow-up (or a TODO) would keep the parity from being forgotten if the trial paywall is re-enabled.
2. **Sync gate on the event loop**: `enforce_chat_quota` does Firestore/Redis I/O synchronously inside the async handler — but that's exactly the pre-existing pattern of the sibling description endpoints in this router (and `routers/goals.py`); `routers/chat_generation.py` shows the `run_blocking` alternative. Following the local convention is the right call for this fix; the executor migration is a router-wide cleanup rather than something this PR should carry.

**backend/tests/unit/test_free_quota_gate_wiring.py** — Good wiring tests. The blocked cases assert both 402 propagation and `llm.assert_not_awaited()` (proving gate-before-spend), the allowed cases assert the proceed path, and `gate.assert_called_once_with('u1')` pins the new call signature. Patching `utils.llm.app_generator.*` at the source module correctly handles the in-function imports, and the docstring update keeps the file's coverage map accurate.

**.github/failure-classes/FC-sibling-managed-llm-missing-chat-quota-gate.json** — Follows the established schema (violated_contract / canonical_prevention / canonical_prevention_artifact / scope_hints). `evidence_prs: []` is fine for a newly born class; referencing this PR once merged would help future greps.

One residual for maintainers (explicitly out of scope in #12715, so not a change request): `/v1/app/generate-prompts` still bills `Features.APP_GENERATOR` under rate-limit only.

No split needed — three files, one concern. Leaving for human maintainer merge sign-off since this sits on the billing/quota surface, alongside the minor product call on X-App-Platform parity above.

— automated review on behalf of the Omi maintainers (AI-generated).

---

aryanorastar:
Closing — superseded on `main` by [#12739](https://github.com/BasedHardware/omi/pull/12739) (`1fff74f7be`), which fixed #12715 and closed it as completed today.

The diagnosis held: on this PR's base (`fab60317cd`) `backend/routers/apps.py` had three `enforce_chat_quota` references — the import and the two description generators — and `generate_app_endpoint` / `generate_app_icon_endpoint` billed `Features.APP_GENERATOR` through `track_usage` with no gate. `main` now has five, with both of those endpoints gated.

#12739 also took the option I scoped out. I passed `uid` alone and wrote in "Honest gaps" that threading `X-App-Platform` would churn the generated clients; cubic's P2 said to pass it anyway. #12739 passes `platform=x_app_platform` and regenerates all six client artifacts plus the OpenAPI document. That is the better call — with `platform=None` the desktop trial paywall could never fire on these routes, so the parity gap would have been left to rediscover if `TRIAL_PAYWALL_ENABLED` were ever turned back on. I under-weighted that against the client-regeneration cost.

Two things from this PR that did not land in #12739, neither of which needs this branch:

**`/v1/app/generate-prompts` is still ungated.** `generate_sample_prompts_endpoint` on current `main` runs `with track_usage(uid, Features.APP_GENERATOR)` behind `auth.with_rate_limit(...)` and no `enforce_chat_quota`. It was explicitly out of scope for #12715, so #12739 correctly left it — but it is the same managed-spend leak on the same feature, and it is now the only one left in this router. Happy to send that as a one-line PR matching #12739's shape exactly.

**`.github/failure-classes/FC-sibling-managed-llm-missing-chat-quota-gate.json`** is not on `main`. The class is real — a route billing managed spend while its siblings in the same router gate it — and #12739's test is the natural `canonical_prevention_artifact`. Worth a registry entry only if a maintainer wants one; not worth a PR on its own otherwise.

Issue #12718, opened 2026-09-04
Unreadable WAV duration skips daily STT budget while paid STT still runs


aryanorastar:
## Symptom

On WAV/multipart voice STT routes, when `read_wav_duration_ms` returns `None`, the daily duration budget gate is skipped but paid transcription still runs. Free or under-metered STT spend.

Contrast: the PCM REST path always meters via `compute_pcm_duration_ms` before `try_consume_budget`.

## Evidence (main @ fab60317)

`backend/routers/chat.py` — `/v2/voice-messages` (SSE):

```python
duration_ms = read_wav_duration_ms(first_wav)
if duration_ms is not None:
    allowed, used_ms, remaining_ms = try_consume_budget(uid, duration_ms)
    ...
# then process_voice_message_segment_stream(...) unconditionally
```

`backend/routers/chat.py` — multipart `/v2/voice-message/transcribe`:

```python
if duration_ms is not None:
    total_duration_ms += duration_ms
if total_duration_ms > 0:
    try_consume_budget(...)
# then transcribe every wav_path, including unreadable ones
```

PCM contrast (same file): always `compute_pcm_duration_ms` → `try_consume_budget` before STT.

Hermetic repro (pre-fix): patch `read_wav_duration_ms` → `None`; multipart/`/v2/voice-messages` return 200 and call STT without `try_consume_budget`. Multi-file case charged only readable files while still transcribing the unreadable one.

PR #12719, opened 2026-09-04
fix(backend): fail closed when WAV duration is unreadable for STT budget

aryanorastar:
## What changed and why

When `read_wav_duration_ms` returned `None`, `/v2/voice-messages` and multipart `/v2/voice-message/transcribe` skipped `try_consume_budget` but still ran paid STT (#12718). Reject with typed `invalid_input` (400) before STT so unmeterable audio cannot bypass the daily duration budget. PCM REST already always meters via `compute_pcm_duration_ms`.

none

## How it was verified

```bash
cd backend
echo 'tests/unit/test_desktop_transcribe.py' > /tmp/bug4-tests.txt
BACKEND_UNIT_TEST_FILE_LIST=/tmp/bug4-tests.txt BACKEND_PYTEST_FILE_ISOLATION=0 BACKEND_PYTEST_XDIST=0 \
  bash test.sh -k 'unreadable_duration or multipart_multi_file_budget or budget_exhausted or budget_consumed or octet_stream_budget' -q
# → 69 passed, 5 deselected
```

Pre-fix hermetic proof: with `read_wav_duration_ms` patched to `None`, multipart and `/v2/voice-messages` returned 200 and invoked STT without `try_consume_budget`; multi-file charged only readable files while still transcribing the unreadable one.

## Tests

- `test_multipart_unreadable_duration_rejects_before_stt`
- `test_voice_messages_unreadable_duration_rejects_before_stt`
- `test_multipart_any_unreadable_duration_rejects_before_stt`

## Failure class (fixes)

Failure-Class: new

## Line count

Line-Count-Exception: backend/routers/chat.py | 2130 -> 2137 | Fail-closed unreadable WAV duration checks at both voice STT budget gates.

## Honest gaps

- No live provider / Redis budget integration run (hermetic unit coverage only).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12719?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-unmeterable-audio-skips-stt-budget-gate.json
+++ b/.github/failure-classes/FC-unmeterable-audio-skips-stt-budget-gate.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": 1,
+  "id": "FC-unmeterable-audio-skips-stt-budget-gate",
+  "violated_contract": "Voice STT routes that bill against the rolling daily duration budget must not invoke a paid transcription provider when audio duration cannot be metered. Skipping try_consume_budget when read_wav_duration_ms returns None (or omitting an unreadable file from a multi-file sum) lets paid STT run free; the PCM REST path always meters via compute_pcm_duration_ms and must remain the contrast.",
+  "canonical_prevention": "Before any paid STT call on WAV/multipart/SSE voice routes, require a readable duration for every file that will be transcribed; reject with TranscriptionOutcome.INVALID_INPUT (HTTP 400) when duration is unreadable, then call try_consume_budget with the full metered sum. Pin with hermetic tests that assert STT helpers are not called when duration is None.",
+  "canonical_prevention_artifact": [
+    "backend/tests/unit/test_desktop_transcribe.py"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    "backend/routers/chat.py",
+    "backend/utils/voice_duration_limiter.py",
+    "backend/tests/unit/test_desktop_transcribe.py"
+  ],
+  "status": "open"
+}
--- a/backend/routers/chat.py
+++ b/backend/routers/chat.py
@@ -873,14 +873,18 @@ def create_voice_message_stream(
             )
 
         # Daily budget check (first file only — matches actual DG usage).
-        # A quota rejection is not an STT attempt and therefore is not an
-        # invalid-input or provider-outcome metric.
+        # Unreadable duration must fail closed before paid STT (PCM always meters).
         first_wav = wav_paths[0]
         duration_ms = read_wav_duration_ms(first_wav)
-        if duration_ms is not None:
-            allowed, used_ms, remaining_ms = try_consume_budget(uid, duration_ms)
-            if not allowed:
-                raise HTTPException(status_code=429, detail='Daily transcription budget exhausted')
+        if duration_ms is None:
+            raise TranscriptionFailure(
+                TranscriptionOutcome.INVALID_INPUT,
+                provider=stt_provider,
+                retryable=False,
+            )
+        allowed, used_ms, remaining_ms = try_consume_budget(uid, duration_ms)
+        if not allowed:
+            raise HTTPException(status_code=429, detail='Daily transcription budget exhausted')
     except TranscriptionFailure as failure:
         _record_preparation_failure(failure)
         _cleanup_temp_voice_wavs(paths + wav_paths, uid)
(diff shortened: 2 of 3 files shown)


Git-on-my-level:
Verified — this closes a real unmetered-STT bypass on both WAV routes, and the fix is correct. Thank you @aryanorastar!

What I checked against the head commit:

- `backend/routers/chat.py` (`/v2/voice-messages`): previously `read_wav_duration_ms(first_wav)` returning `None` skipped `try_consume_budget` entirely while the stream still ran paid STT. Now it raises `TranscriptionFailure(INVALID_INPUT, retryable=False)` before any provider call, which maps to HTTP 400 via `_transcription_http_error`, records the prep-failure metric through the existing handler, and keeps the temp-file cleanup. Good use of the existing typed boundary.
- `backend/routers/chat.py` (multipart `/v2/voice-message/transcribe`): the old loop summed only readable files, and the `total_duration_ms > 0` guard meant an all-unreadable upload transcribed with zero metering. Now any `None` in the loop fails closed with `INVALID_INPUT` and the sum is always fully metered; dropping the now-dead `> 0` condition was the right call.
- Scope check: the WS (`/v2/voice-message/transcribe-stream`) and PCM REST paths meter via `compute_pcm_duration_ms`, which is byte-derived and never `None` — so these two sites were the only bypass gates, and the fix covers them all.
- `backend/tests/unit/test_desktop_transcribe.py`: the three new tests pin the right contract — `try_consume_budget` AND the STT helper are both uncalled when duration is `None`, per route, including the multi-file partial case. Rewriting `test_multipart_multi_file_budget_sums_durations` from "None skipped" to an all-readable sum correctly retires the old buggy semantics.
- `.github/failure-classes/FC-unmeterable-audio-skips-stt-budget-gate.json`: matches the registry's schema and the violated-contract/prevention text is accurate to the diff; the `Failure-Class: new` declaration and line-count exception in the PR body check out.

Two behavior notes for maintainers (not blocking):

1. `read_wav_duration_ms` also returns `None` for valid-but-unmeterable audio (PyAV finds neither container nor stream duration, or computes <= 0 — `backend/utils/voice_duration_limiter.py`). Such uploads now get 400 instead of free STT; that's the right fail-closed default for a billing gate, but it is an availability trade-off for legitimately header-less files (same point the cubic review raised — agreed it's intentional here).
2. The SSE route still meters only the first WAV (pre-existing, documented as matching actual Deepgram usage). Fine to leave as a follow-up if multi-file SSE becomes real.

CI is green including the hermetic backend gates. Leaving formal merge sign-off to the maintainers since this changes billing-adjacent budget enforcement (previously-accepted unmeterable uploads now rejected); would be glad to see it land.

---

aryanorastar:
Closing — superseded on `main` by [#12723](https://github.com/BasedHardware/omi/pull/12723) (`e0d970d888`), which fixed #12718 today.

The bug was real and both fixes close it. They differ on what to do with audio whose duration cannot be read, and `main` picked the better answer:

```python
# main (#12723)
budget_duration_ms = duration_ms if duration_ms is not None else MAX_SESSION_DURATION_S * 1000
allowed, used_ms, remaining_ms = try_consume_budget(uid, budget_duration_ms)
```

```python
# this PR
if duration_ms is None:
    raise TranscriptionFailure(TranscriptionOutcome.INVALID_INPUT, provider=stt_provider, retryable=False)
```

Mine refuses the upload. `main` transcribes it and charges the worst case — `MAX_SESSION_DURATION_S`. Both make it impossible to get unmetered STT, which was the actual defect, but only one of them keeps working for the user.

That difference is exactly the objection cubic raised on this PR and I did not act on: `read_wav_duration_ms` returns `None` for valid-but-unmeterable audio too, not just corrupt input — no container duration, no stream duration, or a computed duration `<= 0`, which is ordinary for streamed or truncated webm/mp4 captures. Under this PR those uploads start returning 400 having previously transcribed fine. Under #12723 they still transcribe, and the billing hole is closed by over-charging rather than by rejecting.

Git-on-my-level called my version "the right fail-closed default for a billing gate" while noting the availability trade-off, and I agreed and moved on. Charging the ceiling closes the same hole without spending real users' working uploads to do it — the conservative choice was available on the billing side alone, and I put the cost on availability instead.

Nothing here is worth carrying forward: #12723 covers both routes, and the WS and PCM REST paths were never affected since `compute_pcm_duration_ms` is byte-derived and never returns `None`.

One item from the review that is still open and outlived both PRs: the SSE route meters only the first WAV. That is pre-existing and documented as matching actual Deepgram usage, so it is a follow-up rather than a defect — worth its own issue if multi-file SSE ever becomes real.

Issue #12720, opened 2026-09-04
Concurrent PTT transcribe-stream sessions can overspend daily STT budget


aryanorastar:
## Symptom

Concurrent `/v2/voice-message/transcribe-stream` sessions for the same uid can overspend the shared rolling daily voice STT duration budget.

Connect only probes with `check_budget` (`try_consume_budget(uid, 0)` — no reservation). Mid-session freezes that remaining slice locally. Session end always `record_actual_duration(..., force=1)`. Two sessions that both admit near the limit can each stream the same remaining slice and both force-record it.

## Evidence (main @ fab60317)

Hermetic Lua (fakeredis) proof:

1. Seed used = `DAILY_BUDGET_MS - 60_000`.
2. Two probe admits both return remaining `60000`.
3. Two force-records of `60000` → final used `DAILY_BUDGET_MS + 60000`.

Atomic `try_consume` / reserve-up-to of the remaining slice admits only the first session.

Contrast: single-session `force=1` after a successful drain is intentional tracking so subsequent requests see debt. The failure class is **concurrent admission without reservation**, not force-record itself.

No per-uid PTT stream lock (unlike `/v4/listen`). Rate limit `voice:transcribe_stream` is 60/hour and does not serialize concurrent sockets. REST+WS race has the same shape when a batch consume and a probe-admitted stream overlap near the limit.

## Out of scope

- Unreadable WAV duration skipping the budget gate (tracked separately; see related STT budget work).
- Changing REST all-or-nothing `try_consume_budget` semantics.

aryanorastar:
Fix in progress: reserve-at-connect for concurrent PTT stream budget overspend.

aryanorastar:
Fix: https://github.com/BasedHardware/omi/pull/12721 (atomic reserve-at-connect + settle/refund).

aryanorastar:
Fixed by #12721, merged 2026-09-04 — closing.

The fix is the one this report argued for: an atomic reserve at connect instead of probe-then-force. `try_reserve_session_budget` consumes-up-to `MAX_SESSION_DURATION_S` in Lua at connect, mid-session audio is capped to the reservation, `settle_reserved_duration` runs after a successful drain, and the full reservation is refunded on terminal provider failure.

Checked against what this issue actually asked for, not just that a PR referenced it:

[...]

PR #12721, opened 2026-09-04
fix(backend): reserve PTT stream STT budget at connect

aryanorastar:
## What changed and why

Concurrent `/v2/voice-message/transcribe-stream` sessions probed the shared daily STT duration budget without reserving (`check_budget` = `try_consume_budget(uid, 0)`), froze that remaining slice mid-session, then `record_actual_duration(..., force=1)` at end. Two sessions near the limit could both admit the same remaining slice and overspend `#12720`.

Fix: atomically `try_reserve_session_budget` (Lua force=2 consume-up-to `MAX_SESSION_DURATION_S`) at connect, cap mid-session audio to the reservation, `settle_reserved_duration` after a successful drain, and refund the full reservation on terminal provider failure. Single-session force-record for tracking is unchanged on the fail-open path.

none

## How it was verified

```bash
cd backend
echo 'tests/unit/test_voice_duration_limiter.py
tests/unit/test_desktop_transcribe.py' > /tmp/bug5-tests.txt
PYTHON=/path/to/backend/.venv/bin/python \
BACKEND_UNIT_TEST_FILE_LIST=/tmp/bug5-tests.txt BACKEND_PYTEST_FILE_ISOLATION=0 BACKEND_PYTEST_XDIST=0 \
  bash test.sh -q
# → 103 passed, 5 deselected
```

Pre-fix hermetic Lua (fakeredis): seed `DAILY_BUDGET_MS - 60_000`; two probes both see 60s remaining; two force-records → used `DAILY_BUDGET_MS + 60_000`. Reserve-up-to admits only the first concurrent session.

## Tests

- `test_legacy_probe_and_force_overspends_under_concurrency`
- `test_reserve_up_to_admits_only_remaining_under_concurrency`
- `test_settle_refunds_unused_reservation`
- Updated WS budget / no-charge-on-provider-failure settle assertions

## Failure class (fixes)

Failure-Class: new

## Line count

Line-Count-Exception: backend/routers/chat.py | 2130 -> 2159 | Atomic probe-only WS budget admit with reserve-at-connect + settle/refund in the PTT stream handler.

## Honest gaps

- No live multi-client Redis/provider soak (hermetic Lua + WS unit coverage only).

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12721?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/failure-classes/FC-concurrent-ptt-stream-budget-overspend.json
+++ b/.github/failure-classes/FC-concurrent-ptt-stream-budget-overspend.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": 1,
+  "id": "FC-concurrent-ptt-stream-budget-overspend",
+  "violated_contract": "Concurrent /v2/voice-message/transcribe-stream sessions for the same uid must not each admit against the same remaining daily STT duration slice. Probe-only check_budget at connect plus force-record at end lets N sessions freeze identical remaining_ms and overspend the shared voice_duration budget after the fact.",
+  "canonical_prevention": "Atomically reserve up to MAX_SESSION_DURATION_S at WS connect via try_reserve_session_budget (Lua force=2 consume-up-to). Cap mid-session audio to the reserved amount. On successful drain, settle_reserved_duration(reserved, actual); on terminal provider failure, refund the full reservation. Pin with fakeredis Lua tests that show probe+force overspend and reserve-up-to mutual exclusion.",
+  "canonical_prevention_artifact": [
+    "backend/tests/unit/test_voice_duration_limiter.py",
+    "backend/tests/unit/test_desktop_transcribe.py"
+  ],
+  "evidence_prs": [],
+  "scope_hints": [
+    "backend/routers/chat.py",
+    "backend/utils/voice_duration_limiter.py"
+  ],
+  "status": "open"
+}
--- a/backend/routers/chat.py
+++ b/backend/routers/chat.py
@@ -92,10 +92,12 @@ from utils.observability.fallback import record_fallback
 from utils.journey_metrics_contract import resolve_client_kind, resolve_client_kind_from_headers
 from utils.observability.journeys import ClientJourneyAttempt, JourneyAttempt
 from utils.voice_duration_limiter import (
+    MAX_SESSION_DURATION_S,
     compute_pcm_duration_ms,
     read_wav_duration_ms,
     try_consume_budget,
-    check_budget,
+    try_reserve_session_budget,
+    settle_reserved_duration,
     record_actual_duration,
 )
 from testing.parity_pack_v0.live_capture import SurfaceParityCapture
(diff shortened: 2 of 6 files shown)


Git-on-my-level (commented):
Thanks @aryanorastar — this is a well-constructed fix for the concurrent-session overspend in #12720, and the tests genuinely pin the failure class. File-by-file:

**backend/utils/voice_duration_limiter.py** — The `force=2` reserve-up-to branch is the right shape: it computes `take = math.min(request, remaining)` inside the atomic script, ZADDs a reservation entry with an INCR nonce, and returns the 4-tuple so the caller knows exactly how much was reserved. `settle_reserved_duration` charging only the delta (`record_actual_duration(uid, actual - reserved)`) keeps successful short sessions charged at actuals. Changing the record guard from `request > 0` to `request ~= 0` correctly admits negative refund entries. Two bounded notes for a follow-up: (1) the refund entry carries the settle timestamp, not the reservation's connect timestamp — near the 24h window edge a refund can outlive its positive reservation and briefly undercount `used_ms` (bounded by one <=120s reservation); stamping refunds with the original reservation timestamp would close it. (2) If a worker dies between reserve and settle, the reservation stays charged for the remainder of the window — also bounded at one session's worth, but an expiring lease would make it self-healing.

**backend/routers/chat.py** — Moving admission inside the session `try` (chat.py:1656-1669) so every exit path can settle/refund is the correct structure, and `refund_reservation_once()` on the no-drain branch of the finally (chat.py:1804-1806) means provider failures no longer charge users. `usage_recorded` still guarantees settle-or-refund happens exactly once per session. Two observations: the settle/refund calls are synchronous Redis calls in the async WS finally path — consistent with how this handler already calls the sync limiter, and the Lua is short, but worth noting. And `budget_remaining_ms = reserved_ms` caps mid-session audio to the reservation (the chat.py:1765-1772 check), which means a client that opens two concurrent sessions near the limit now gets one full session plus one 1008 instead of two short ones — reasonable for a cost cap, flagging it as a user-visible trade-off.

**backend/tests/unit/test_voice_duration_limiter.py** — The three `TestConcurrentSessionReservation` tests are the strongest part of this PR: they replay the exact pre-fix failure mode (two probes both seeing 60s remaining, two force-records ending at `DAILY_BUDGET_MS + 60_000`) and then prove reserve-up-to admits only one of two concurrent callers. They all passed in CI at this head.

**backend/tests/unit/test_desktop_transcribe.py** — The assertion migrations are honest behavior changes, not mechanical renames: refund paths now assert `settle_reserved_duration('test-uid', reserved, 0)` where before they asserted `record_actual_duration` was never called (e.g. provider-failure and finalize-failure cases), and the settle test asserts `(120000, 1000)` — reserved 120s, actual 1s. Good.

**backend/tests/unit/_chat_router_test_harness.py** — Stubs for the new entry points plus `MAX_SESSION_DURATION_S = 120` keep the shared harness consistent with the limiter stubs in test_desktop_transcribe.py.

**.github/failure-classes/FC-concurrent-ptt-stream-budget-overspend.json** — Matches the existing failure-class schema (schema_version, violated_contract, canonical_prevention + artifacts, evidence_prs, scope_hints, status: open) and accurately describes the fixed contract.

**CI note** — The failing `Backend unit suite` is pre-existing drift on the base, not from this PR: `test_pinned_conversation_dump_set_is_exact` flags two `model_dump` sites in `utils/llm/external_integrations.py` (`_basic_daily_summary`, `generate_comprehensive_daily_summary`), that file is identical between this PR's base and head, and main has since pinned those sites. A rebase onto current main should clear it.

Leaving for human maintainer review — the accounting edge cases above are bounded and don't reproduce the overspend, but the reservation-lifetime design (refund timestamps, crash-abandoned reservations) deserves a maintainer's eye before this becomes the pattern for other budgeted streams.

---

kodjima33 (approved):
Confidence 4/5: stated root cause (concurrent PTT stream sessions overspend shared STT budget, #12720), scoped fix mirrored by extensive new tests (TestConcurrentSessionReservation), CI green except the shared stale-guardrail-base 'Backend unit suite' failure (fixed by #12740, now on main). Diff is 311 lines (11 over the 300 soft cap) but every other gate signal is strong. No overlap with #12717/#12739 (different subsystem - chat quota vs STT duration budget).

Issue #12731, opened 2026-09-04
External realtime runs finish with empty final_text, so completion delivery is content-free


Git-on-my-level:
## Evidence

From the 2026-09-04 Omi Beta field trace behind #12730 (agent-runtime DB, `omi-agentd.sqlite3`):

- Every realtime-voice-deferral run spawned in a 2.9h session (`authority: swift_realtime`, 8 runs) finished with `final_text` length 0; `message.completed` events carry literally `{"text":""}` (e.g. events for run_659c, which answered "Think harder and give me the exact layout" with a spoken recipe correction but journaled nothing).
- A spawn-receipt turn (`acceptedSpawnOwnerID`) delegates its journal exchange to the run result — when the run's final text is empty, the assistant row is never written at all. In the trace, the *corrected* hopper recipe was spoken aloud at 04:49 UTC but exists nowhere in `conversation_turns`; the journal's last statement on the topic stayed the wrong recipe, and the model "corrected" it again 71 minutes later, mid-unrelated-answer.

## Defect

External-surface attempts (`completeExternalSurfaceRun` / external-attempt finalization in `agent/src/runtime/kernel-core.ts`) never capture the streamed text into `runs.final_text`, so:

1. The completion-delta lane (`peekCompletedAgentDelta`) is structurally content-free for realtime-spawned work — injections can only say "an agent finished", never what it found.
2. Turn exchanges owned by the kernel through a spawn receipt permanently lose the spoken answer.

## Fix direction

Capture the streamed response text into `runs.final_text` on external terminalization (or feed it through `message.completed`), so both the completion lane and spawn-receipt journaling see the real answer. TS-side (agent runtime) with tests in `agent/tests/`.

Found while fixing #12730; that PR fixes the Swift-side journal-status half (delivered barge-ins no longer sealed as failures) but cannot recover text the kernel never persisted.

Git-on-my-level:
Status update from the merge of #12734 (kernel half, squash commit `3d72f77`):

[...]

PR #2, opened 2024-03-26
colors to the LED board

Rishabhjava:
added feedback to onboard LED, changes between red, green and blue based on status of BLE connection, if there is a problem with connection or with the mic and whether the device is recording

The simple changes are:
Red -> problem setting up
Blue -> recording and streaming audio
Green -> not recording, but connected

Issue #12732, opened 2026-09-04
AgentCompletionVoiceDelivery never fires for realtime-spawned (floating_chat) runs


Git-on-my-level:
## Evidence

From the 2026-09-04 field trace behind #12730:

- Realtime-voice deferral runs (`authority: swift_realtime`) run on sessions with `surface_kind = floating_chat`.
- `AgentCompletionVoiceDelivery.triggerSurfaceKinds` is `["floating_bar", "service", "workstream"]` — `floating_chat` is absent.
- Those runs also never create `AgentRuntimeStatusStore` projections (projections are created only by Swift-side `beginRequest`/`update` calls, which external realtime runs don't make), so `observe()` sees no terminal transition for them at all.

## Defect

The delivery contract in the class doc — "the model mentions the completion at the next natural turn boundary" — is unreachable for exactly the runs it was designed for. Deferred voice answers ("Let me verify…") complete silently; delivery degrades to connect/input-window drains that can land much later (and, combined with #12731, carry no content).

Concrete user impact from the trace: a potion-recipe question got "Let me verify", the full answer existed in the journal 10 seconds later, and it was never spoken — the follow-up question was then answered with an unrelated older thread.

## Fix direction

Either project external realtime run state into `AgentRuntimeStatusStore` (so terminal transitions are observable) and include the relevant surface kind in `triggerSurfaceKinds`, or deliver deferred-run completions through the voice turn coordinator when the spawning turn is still live. Needs care around barge-in semantics: a completion that lands while the user is speaking should wait for the next turn boundary, not interrupt.

Depends on #12731 for the delivered content to be meaningful.

PR #12734, opened 2026-09-04
fix(agent): persist external-surface run text into runs.final_text

SakashSrivastava:
## What changed and why

Refs #12731. **This PR does not close that issue** — see "What this does not repair" below.

An external surface owns its own streaming, so the kernel never observes what a realtime-voice deferral run actually answered. `completeExternalSurfaceRun` was the only point where that text could be captured, and it passed `finalText: null` unconditionally. `CompleteExternalSurfaceRunInput` had no field to carry the text, so the surface could not hand its answer back even if it wanted to.

The fix threads the text through the four layers between the surface and the row:

```
external_surface_run_complete (protocol.ts)
  -> IPC handler (index.ts)
  -> CompleteExternalSurfaceRunInput (kernel-types.ts)
  -> completeExternalSurfaceRun -> finishAttemptAndRun -> runs.final_text
```

No new persistence machinery. `finishAttemptAndRun` already accepted `finalText: string | null` and wrote it through `updateRun`; only external runs were never fed anything.

**Decisions worth naming, all taken from existing precedent rather than preference:**

*Persisted on every terminal status, not only `completed`.* The internal adapter path passes `finalText: result.text` regardless of status and signals failure through `errorCode` rather than by discarding the answer. A realtime run that spoke a partial answer before failing keeps it, and `final_text` keeps meaning "what this run produced".

*Empty and whitespace-only text collapse to `null`.* "Produced no text" then reads identically however the surface expresses it, so no consumer has to special-case the difference. A non-string is rejected as `invalid_external_request` rather than persisted. This is a deliberate divergence from the internal path, which stores `""` as-is.

*The result echoes `finalTextPersisted`.* A surface talking to an older kernel would otherwise read `ok: true` with its answer silently dropped, which is the failure mode of #12731 one layer up. Absence of the field identifies an old kernel; `false` on a `duplicate` replay says first-write-wins kept the earlier text.

*The requeue path now clears `final_text` and `result_json`.* `beginExternalSurfaceRun` is idempotent per `turnId`, and its requeue branch reset `status`/`errorCode`/`errorMessage` but not the answer. That was unreachable while external `final_text` was always null; making the column writable makes it reachable. A recycled run in `queued`/`running` is read live by the context snapshot's `activeRuns` projection (`context-snapshot.ts:299` selects `r.final_text`, and those statuses are in `ACTIVE_RUN_STATUSES`) and by `latestActivity` in the control list. A superseded answer sitting in model context is the #12731 staleness bug one level down.

*No length bound on `finalText`.* `errorCode` is capped, this is not. Deliberate: it matches the internal path, where `result.text` is likewise uncapped, and consumers bound it downstream. Flagging it as a conscious choice rather than an oversight.

## What this does not repair

Being explicit, because the first draft of this description overstated it and two reviewers were right to push back:

1. **Nothing sends `finalText` yet, so merged alone this changes no field behavior.** `externalSurfaceRunCompleteWireMessage` (`AgentRuntimeProcess.swift:1522`) writes `sessionId`, `runId`, `attemptId`, `terminalStatus`, `errorCode` and no text; `completeExternalRunAuthority` / `terminalizeExternalRun` take no text parameter. That is the only producer of this message in the repo. Until the Swift half in #12732 lands, `runs.final_text` for `swift_realtime` runs stays NULL and the issue's evidence query returns the same rows the day after this merges.

2. **The incident's journal loss is not fixed by this PR under any `finalText` supply.** No kernel path materializes a `conversation_turns` assistant row from a plain external-surface run's `final_text`: `completeExternalSurfaceRun` writes the run row and the event, and `ensureAgentSpawnJournal` requires `producerJournal` metadata that only `spawn_agent` children carry. run_659c came from the `think_deeper` voice tool, not `spawn_agent`, so the spawn-receipt framing holds for that run class but not for the incident run itself.

3. **A future Swift producer must accumulate text over the turn stream, not read the turn-end hub property.** For run_659c the streamed reply lived on the voice session hub and was already cleared by the time the journal capture ran, so a companion reading hub state at terminalization would hand this PR an empty string, which collapses to null. Noted in the `protocol.ts` doc comment so whoever builds #12732 meets it before the field trace does.

So: this is the correct receiving half and #12732's stated prerequisite. It should not be recorded as the repair of #12731, and the closing verb is deliberately "Refs".

- INV-AGENT-*
- INV-CHAT-1

Both match on `desktop/macos/agent/src/runtime/**` via `kernel-core.ts` and `kernel-types.ts`. The change is compatible with each, and INV-CHAT-1 is arguably the reason to make it.

**INV-CHAT-1 — "Every canonical conversation has one kernel-owned transcript."** The spoken answer of a realtime-spawned run reaches `omi-agentd.sqlite3` nowhere today, so the one kernel-owned transcript has a hole exactly where the user was answered. Persisting it puts that answer into the canonical store. No second store, no per-surface continuity ring, no mirroring — one new column value on the existing `runs` row.

On the MUST NOT that Swift may not own run truth: the surface **reports** an outcome only it observed, and the kernel validates and writes it. That is the shape the external-surface authority mechanism already has — `terminalStatus` arrives the same way today — and every existing guard still runs ahead of the write: owner and attempt assertions, the pending-invocation check, and the terminal-state guard.

**INV-AGENT-\* — "Omi owns session/run/attempt identity, authority, and lifecycle."** Identity is untouched — `runId` and `attemptId` are still kernel-minted and re-asserted on completion. Of the twelve MUST NOTs the nearest is *"assemble policy or model context from Swift-supplied prompt fragments"*; `finalText` is a record of what a run produced, written after terminalization. The requeue fix above exists precisely so it cannot leak into context assembly through a recycled run.

## How it was verified

Rebased onto current `main` (was 55 behind; #12730 landed in that window and the earlier baseline predated it). Windows, Node 24.14.0, vitest 4.1.10, `npm ci` from the checked-in lockfile.

| Command | Result |
|---|---|
| `npx tsc --noEmit` | clean |
| `npx vitest run tests/external-surface-authority.test.ts` | **27 passed** (18 before) |
| `npx vitest run tests/protocol-v2.test.ts` | **13 passed** |
| `npx vitest run` (full suite, this branch) | 31 failed / **717 passed** |
| `npx vitest run` (full suite, `upstream/main`) | 31 failed / **708 passed** |

Both full-suite runs are on the same machine against the same base, each after `npm run build`. Identical failure count, +9 passing — exactly the nine cases added here. No new failures.

Those 31 are pre-existing, and every one I can account for is a Windows environment failure: `EACCES` binding an `AF_UNIX` socket under `%TEMP%`, `EPERM` from `rmSync` while SQLite still holds a handle, `EPERM: symlink` without Developer Mode, and POSIX path assumptions. A macOS run is the more useful absolute number; the delta above is the part my machine can speak to.

**The tests fail without the fix.** Reverting only `src/runtime/kernel-core.ts` and keeping the tests fails the persistence assertion, the `message.completed` payload assertion, the failed-status retention, and the non-string rejection.

`scripts/pr-preflight --base upstream/main` passes all 14 selected checks.

## Tests

Nine cases in `tests/external-surface-authority.test.ts`:

- text streamed by the surface lands in `runs.final_text`
- the emitted `message.completed` event carries that text rather than `{"text":""}`
- an omitted field, an empty string, and whitespace-only all leave `final_text` null
- text produced before a `failed` terminalization is kept, alongside `status` and `error_code`
- text produced before a `cancelled` terminalization is kept, and the failed-only `errorCode` default does not fire
- a non-string `finalText` is rejected, and a valid retry then lands on `succeeded`, proving the rejection left the run terminalizable rather than half-sealed
- a same-status replay with different text keeps the first answer and returns `duplicate: true`
- `finalTextPersisted` is true when text was stored and false when none was
- an orphaned turn re-begun clears `final_text` and `result_json` instead of carrying a superseded answer into an active run

Plus `finalText` and `finalTextPersisted` pinned on the v2 wire shape in `tests/protocol-v2.test.ts`.

**One gap I have not closed.** The `finalText: request.finalText` glue in `index.ts` is covered at the type level by the v2 wire test, but not by a runtime round-trip — every kernel test calls `completeExternalSurfaceRun` directly. The stdio harness that would prove it (`runtime-stdio-contract`) does not run on this machine, so I would be writing a test I cannot execute. Saying so rather than shipping one I have not watched pass.

Failure-Class: FC-decoded-outcome-discarded-before-canonical-record

The class's contract is that when a boundary decodes an outcome signal, the record downstream consumers treat as canonical must carry it.

## Scope

TypeScript side only. #12732 is the Swift-side delivery half and needs a Mac to build and test.

### DIFF
--- a/desktop/macos/agent/src/index.ts
+++ b/desktop/macos/agent/src/index.ts
@@ -2465,6 +2465,7 @@ async function main(): Promise<void> {
             ok: true,
             terminalStatus: result.terminalStatus,
             duplicate: result.duplicate,
+            finalTextPersisted: result.finalTextPersisted,
           });
         } catch (error) {
           send({
--- a/desktop/macos/agent/src/index.ts
+++ b/desktop/macos/agent/src/index.ts
@@ -2451,6 +2451,7 @@ async function main(): Promise<void> {
             runId: request.runId,
             attemptId: request.attemptId,
             terminalStatus: request.terminalStatus,
+            finalText: request.finalText,
             errorCode: request.errorCode,
           });
           send({
--- a/desktop/macos/agent/src/protocol.ts
+++ b/desktop/macos/agent/src/protocol.ts
@@ -127,6 +127,17 @@ export interface ExternalSurfaceRunCompleteMessage extends ProtocolEnvelope {
   runId: string;
   attemptId: string;
   terminalStatus: "completed" | "failed" | "cancelled";
+  /**
+   * The response text the external surface streamed to the user.
+   *
+   * An external surface owns its own streaming, so the kernel never observes this
+   * run's output — it has to be handed back at terminalization or it is lost.
+   * Without it `runs.final_text` stays null and every consumer that reads a run's
+   * answer is content-free: the completion lane can only report that an agent
+   * finished, and a spawn receipt that delegates its journal exchange to the run
+   * result writes no assistant row at all (#12731).
+   */
+  finalText?: string;
   errorCode?: string;
 }
 
--- a/desktop/macos/agent/src/protocol.ts
+++ b/desktop/macos/agent/src/protocol.ts
@@ -128,14 +128,18 @@ export interface ExternalSurfaceRunCompleteMessage extends ProtocolEnvelope {
   attemptId: string;
   terminalStatus: "completed" | "failed" | "cancelled";
   /**
-   * The response text the external surface streamed to the user.
+   * The answer text the external surface reports for this run.
    *
    * An external surface owns its own streaming, so the kernel never observes this
    * run's output — it has to be handed back at terminalization or it is lost.
    * Without it `runs.final_text` stays null and every consumer that reads a run's
    * answer is content-free: the completion lane can only report that an agent
-   * finished, and a spawn receipt that delegates its journal exchange to the run
-   * result writes no assistant row at all (#12731).
+   * finished, and a spawn-agent child's receipt, which builds its journal block
+   * from the child's final text, has nothing to write (#12731).
+   *
+   * What a surface can actually report is its own problem: for a realtime voice
+   * deferral the text must be accumulated over the turn stream, because the
+   * turn-end hub property is already cleared by the time terminalization runs.
    */
   finalText?: string;
   errorCode?: string;
@@ -764,6 +768,12 @@ export interface ExternalSurfaceRunCompleteResultMessage extends OutboundEnvelop
   ok: boolean;
   terminalStatus?: "completed" | "failed" | "cancelled";
   duplicate?: boolean;
+  /**
+   * Whether the kernel stored `finalText` for this run. Absent from an older
+   * kernel, which lets a surface keep its own journal fallback instead of
+   * trusting a silent no-op (#12731).
+   */
+  finalTextPersisted?: boolean;
   error?: ExternalAuthorityError;
 }
 
--- a/desktop/macos/agent/src/runtime/kernel-core.ts
+++ b/desktop/macos/agent/src/runtime/kernel-core.ts
@@ -404,9 +404,14 @@ export class KernelCore {
       const latestAttempt = latestAttemptRow ? attemptFromRow(latestAttemptRow) : undefined;
       if (run.status === "orphaned" || !latestAttempt || TERMINAL_STATUSES.includes(latestAttempt.status)) {
         this.withTransaction(() => {
+          // A recycled run must not carry the previous attempt's answer while it
+          // is queued again: the context snapshot and control list both read
+          // final_text for active runs (#12731).
           this.updateRun(run.runId, {
             status: "queued",
             completedAtMs: null,
+            finalText: null,
+            resultJson: null,
             errorCode: null,
             errorMessage: null,
             updatedAtMs: Date.now(),
@@ -631,7 +636,10 @@ export class KernelCore {
     const persistedStatus = input.terminalStatus === "completed" ? "succeeded" : input.terminalStatus;
     if (TERMINAL_STATUSES.includes(run.status) || TERMINAL_STATUSES.includes(attempt.status)) {
       if (run.status === persistedStatus && attempt.status === persistedStatus) {
-        return { ...input, duplicate: true };
+        // First write wins, so a replayed frame's text is deliberately not
+        // stored. Say so rather than letting the surface read ok and assume it
+        // landed — that silence is the shape of #12731 itself.
+        return { ...input, duplicate: true, finalTextPersisted: false };
       }
       throw new ExternalSurfaceAuthorityError("run_terminal", "External surface run already has a different terminal state");
     }
@@ -655,11 +663,13 @@ export class KernelCore {
     }
     // The kernel never observes an external surface's stream, so terminalization is
     // the only point at which this run's answer can be captured. Persisted for every
-    // terminal status, matching the internal adapter path, which likewise stores
-    // whatever text was produced and marks failure through errorCode instead. An
-    // empty string collapses to null so "produced no text" reads identically whether
-    // the surface omitted the field or sent "" (#12731).
-    const finalText = input.finalText ? input.finalText : null;
+    // terminal status, following the internal adapter path, which likewise stores
+    // whatever text was produced and marks failure through errorCode instead. The
+    // internal path stores "" as-is; here an empty string collapses to null so
+    // "produced no text" reads identically whether the surface omitted the field
+    // or sent "" (#12731).
+    const trimmed = input.finalText?.trim();
+    const finalText = trimmed ? trimmed : null;
     this.withTransaction(() => {
       this.finishAttemptAndRun({
         sessionId: input.sessionId,
@@ -671,7 +681,7 @@ export class KernelCore {
         errorMessage: persistedStatus === "failed" ? "External surface execution failed" : null,
       });
     });
-    return { ...input, duplicate: false };
+    return { ...input, duplicate: false, finalTextPersisted: finalText !== null };
   }
 
   private assertExternalRunIdentity(
--- a/desktop/macos/agent/src/runtime/kernel-core.ts
+++ b/desktop/macos/agent/src/runtime/kernel-core.ts
@@ -650,13 +650,23 @@ export class KernelCore {
     if (errorCode && !/^[a-z0-9_]{1,64}$/.test(errorCode)) {
       throw new ExternalSurfaceAuthorityError("invalid_external_request", "External surface errorCode is invalid");
     }
+    if (input.finalText !== undefined && typeof input.finalText !== "string") {
+      throw new ExternalSurfaceAuthorityError("invalid_external_request", "External surface finalText must be a string");
+    }
+    // The kernel never observes an external surface's stream, so terminalization is
+    // the only point at which this run's answer can be captured. Persisted for every
+    // terminal status, matching the internal adapter path, which likewise stores
+    // whatever text was produced and marks failure through errorCode instead. An
+    // empty string collapses to null so "produced no text" reads identically whether
+    // the surface omitted the field or sent "" (#12731).
+    const finalText = input.finalText ? input.finalText : null;
     this.withTransaction(() => {
       this.finishAttemptAndRun({
         sessionId: input.sessionId,
         runId: input.runId,
         attemptId: input.attemptId,
         status: persistedStatus,
-        finalText: null,
+        finalText,
         errorCode: persistedStatus === "failed" ? errorCode ?? "external_surface_failed" : null,
         errorMessage: persistedStatus === "failed" ? "External surface execution failed" : null,
       });
--- a/desktop/macos/agent/tests/external-surface-authority.test.ts
+++ b/desktop/macos/agent/tests/external-surface-authority.test.ts
@@ -1600,7 +1600,22 @@ describe("external realtime surface authority", () => {
       finalText: "",
     });
 
-    for (const runId of [omitted.runId, empty.runId]) {
+    // Whitespace-only is "produced no text" too, collapsed the same way.
+    const blank = fixture.kernel.beginExternalSurfaceRun({
+      ...beginInput(fixture.sessionId),
+      turnId: "voice-turn-3",
+      requestId: "begin-3",
+    });
+    fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: blank.runId,
+      attemptId: blank.attemptId,
+      terminalStatus: "completed",
+      finalText: "   ",
+    });
+
+    for (const runId of [omitted.runId, empty.runId, blank.runId]) {
       expect(fixture.store.getRow("SELECT final_text FROM runs WHERE run_id = ?", [runId]).final_text).toBeNull();
     }
     fixture.store.close();
@@ -1660,6 +1675,114 @@ describe("external realtime surface authority", () => {
       "SELECT status, final_text FROM runs WHERE run_id = ?",
       [run.runId],
     )).toEqual({ status: "succeeded", final_text: "Recovered after the rejected call." });
+    fixture.store.close();
+  });
+
+  it("reports whether the answer was actually stored", () => {
+    // A surface cannot otherwise tell a successful write from an older kernel
+    // that ignored the field entirely.
+    const fixture = createFixture();
+    const withText = fixture.kernel.beginExternalSurfaceRun(beginInput(fixture.sessionId));
+    expect(fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: withText.runId,
+      attemptId: withText.attemptId,
+      terminalStatus: "completed",
+      finalText: "Three parts sand to one part clay.",
+    }).finalTextPersisted).toBe(true);
+
+    const withoutText = fixture.kernel.beginExternalSurfaceRun({
+      ...beginInput(fixture.sessionId),
+      turnId: "voice-turn-2",
+      requestId: "begin-2",
+    });
+    expect(fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: withoutText.runId,
+      attemptId: withoutText.attemptId,
+      terminalStatus: "completed",
+    }).finalTextPersisted).toBe(false);
+    fixture.store.close();
+  });
+
+  it("clears a superseded answer when an orphaned turn is re-begun", () => {
+    // The requeue branch recycles the run row. Until this PR final_text was
+    // always null on external runs, so nothing could survive the reset; now a
+    // stale answer would sit in an active run and reach the context snapshot.
+    const fixture = createFixture();
+    const run = fixture.kernel.beginExternalSurfaceRun(beginInput(fixture.sessionId));
+
+    fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: run.runId,
+      attemptId: run.attemptId,
+      terminalStatus: "completed",
+      finalText: "Three parts sand to one part clay.",
+    });
+    fixture.store.execute("UPDATE runs SET status = 'orphaned' WHERE run_id = ?", [run.runId]);
+
+    fixture.kernel.beginExternalSurfaceRun({ ...beginInput(fixture.sessionId), requestId: "begin-requeue" });
+
+    expect(fixture.store.getRow(
+      "SELECT status, final_text, result_json FROM runs WHERE run_id = ?",
+      [run.runId],
+    )).toEqual({ status: "running", final_text: null, result_json: null });
+    fixture.store.close();
+  });
+
+  it("keeps the first answer when the surface replays the same terminal status", () => {
+    // Swift retries this RPC on timeout, so a late replay must not rewrite an
+    // answer the kernel already sealed.
+    const fixture = createFixture();
+    const run = fixture.kernel.beginExternalSurfaceRun(beginInput(fixture.sessionId));
+    const complete = (finalText: string) => fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: run.runId,
+      attemptId: run.attemptId,
+      terminalStatus: "completed",
+      finalText,
+    });
+
+    complete("Three parts sand to one part clay.");
+    const replay = complete("A different answer from a retried frame.");
+
+    expect(replay.duplicate).toBe(true);
+    // The replay's text was dropped by design, and the result says so.
+    expect(replay.finalTextPersisted).toBe(false);
+    expect(fixture.store.getRow(
+      "SELECT final_text FROM runs WHERE run_id = ?",
+      [run.runId],
+    ).final_text).toBe("Three parts sand to one part clay.");
+    fixture.store.close();
+  });
+
+  it("keeps text produced before a cancelled terminalization", () => {
+    // cancelled is a distinct persisted status from failed, and the failed-only
+    // errorCode default must not fire for it.
+    const fixture = createFixture();
+    const run = fixture.kernel.beginExternalSurfaceRun(beginInput(fixture.sessionId));
+
+    fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: run.runId,
+      attemptId: run.attemptId,
+      terminalStatus: "cancelled",
+      finalText: "Half an answer before the user cut in.",
+    });
+
+    expect(fixture.store.getRow(
+      "SELECT final_text, status, error_code FROM runs WHERE run_id = ?",
+      [run.runId],
+    )).toEqual({
+      final_text: "Half an answer before the user cut in.",
+      status: "cancelled",
+      error_code: null,
+    });
     fixture.store.close();
   });
 });
--- a/desktop/macos/agent/tests/external-surface-authority.test.ts
+++ b/desktop/macos/agent/tests/external-surface-authority.test.ts
@@ -1522,6 +1522,133 @@ describe("external realtime surface authority", () => {
     expect(store.getRow("SELECT COUNT(*) AS count FROM runs").count).toBe(1);
     store.close();
   });
+
+  // #12731: the kernel never observes an external surface's stream, so unless the
+  // text comes back at terminalization `runs.final_text` stays null and every
+  // consumer of a run's answer — the completion lane, spawn-receipt journaling —
+  // is structurally content-free.
+  it("persists the text an external surface streamed into runs.final_text", () => {
+    const fixture = createFixture();
+    const run = fixture.kernel.beginExternalSurfaceRun(beginInput(fixture.sessionId));
+
+    fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: run.runId,
+      attemptId: run.attemptId,
+      terminalStatus: "completed",
+      finalText: "The hopper recipe is three parts sand to one part clay.",
+    });
+
+    expect(fixture.store.getRow(
+      "SELECT final_text FROM runs WHERE run_id = ?",
+      [run.runId],
+    ).final_text).toBe("The hopper recipe is three parts sand to one part clay.");
+    fixture.store.close();
+  });
+
+  it("puts the streamed text on the message.completed event, not an empty payload", () => {
+    // The trace in #12731 shows these events carrying literally {"text":""}. The
+    // event is emitted from finishAttemptAndRun off the same finalText the run row
+    // stores, so both the completion lane and the journal starve together.
+    const fixture = createFixture();
+    const run = fixture.kernel.beginExternalSurfaceRun(beginInput(fixture.sessionId));
+
+    fixture.kernel.completeExternalSurfaceRun({
+      ownerId: "owner",
+      sessionId: fixture.sessionId,
+      runId: run.runId,
+      attemptId: run.attemptId,
+      terminalStatus: "completed",
+      finalText: "Three parts sand to one part clay.",
+    });
+
+    const completed = fixture.store.allRows(
+      "SELECT payload_json FROM events WHERE run_id = ? AND type = 'message.completed'",
+      [run.runId],
+    );
+    expect(completed).toHaveLength(1);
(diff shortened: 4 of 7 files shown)


Git-on-my-level:
Verified this end to end on the TS side (tsc clean; `tests/external-surface-authority.test.ts` 23/23 on Linux — a useful complement to your Windows run, since the pre-existing red files you listed are mostly Windows-environment failures).

Per file:

- `protocol.ts` — the optional `finalText` on `ExternalSurfaceRunCompleteMessage` with the surface-owns-its-stream rationale documented inline is exactly right; inbound messages are `JSON.parse`-cast (index.ts:1897), so the kernel-side check is the real guard, which you added.
- `index.ts` — the IPC handler passes `finalText` through verbatim inside the existing try/catch, so a kernel rejection still maps to a typed `external_surface_run_complete_result` error. Clean.
- `kernel-types.ts` — `CompleteExternalSurfaceRunInput.finalText?: string` with the #12731 reference; matches the protocol shape.
- `kernel-core.ts` — the good stuff: the type check runs after all existing authority guards (owner/attempt assertion, terminal-state guard, pending-invocation check, errorCode regex) and before `withTransaction`, so a rejected completion leaves the run terminalizable rather than half-sealed. Persisting on every terminal status matches the internal adapter path (`finalText: result.text` regardless of status, failure via errorCode), and the `""` → null collapse keeps "produced no text" unambiguous. I also confirmed the two remaining `finalText: null` sites (kernel-core.ts:1243, 1298) are kernel-internal recovery paths, not external-surface paths — no gap left behind.
- `tests/external-surface-authority.test.ts` — five focused cases; four are genuine regression tests (they fail on the pre-fix kernel). One strengthening suggestion on the non-string rejection test: `expect(...).not.toBe("succeeded")` passes for any non-succeeded state, including a hypothetically half-sealed one, so it proves less than the comment claims. Retrying `completeExternalSurfaceRun` with a valid `finalText` and asserting the run lands on `succeeded` would prove terminalizability outright (same point as the cubic review, and I agree with it). Non-blocking.
- changelog fragment — valid: `desktop-changelog.py` accepts the `changes` list form (it accepts both `change` and `changes`), and it matches the shape of recent merged fragments.

Two notes for the maintainer:

1. The untested half the description flags is real: `agent-spawn-journal.ts:567` builds the completion block's `output` from `child.final_text`, and there is no fixture today for an external realtime run that is also a spawn-receipt child. Since #12732 is the stated consumer of this value, that fixture (or the e2e in #12732) is where the remaining risk sits — worth prioritizing when that PR lands.
2. `finalText` is surface-supplied text written into `runs.final_text`, the canonical record the completion lane and journal trust. Transport is local stdio from the host process (the same trust domain that already supplies `terminalStatus`/`errorCode`), there is no length bound — consistent with the internal path — and every existing guard still runs. The INV-CHAT-1 / INV-AGENT-* analysis in the description checks out against `product/invariants/`. Left for maintainer sign-off before merge rather than automated approval.

Thank you for the exceptionally honest write-up — the baseline-before-claim, the revert-based test verification, and the explicit "I read it but did not exercise it" on the spawn-receipt path are exactly how PR descriptions should read.

---

SakashSrivastava:
Good catch, and fixed in 42bc89264 — you're both right that the assertion proved less than its comment claimed. The run is already non-terminal when it hits that line, so `not.toBe("succeeded")` was true on entry and would have stayed true for a half-sealed run.

It now retries `completeExternalSurfaceRun` with a valid finalText and asserts the run lands on `succeeded` with that text stored, which is the thing the comment was actually claiming. Still fails on the pre-fix kernel, so it stays a regression test rather than just a stronger assertion.

I also went back and worked out all 34 pre-existing failures rather than leave that hanging in the description — every one is a Windows environment failure (AF_UNIX bind, EPERM on rmSync with an open SQLite handle, unprivileged symlink creation, POSIX path assumptions, missing dist artifacts). Description updated with the breakdown. Two of them are cross-platform bugs in the tests themselves rather than environment: convergence-authority-ratchet allowlists paths with `path.endsWith("/runtime/conversation-journal.ts")`, and since the separator is `\` on Windows the exempted file fails its own exemption. Same at line 181 for KernelJournalBackendSyncDriver.swift. Happy to send that as a separate PR if it's worth having.

On the spawn-receipt fixture — agreed that's where the remaining risk is. I'll pick it up when #12732 lands and there's a realtime external run to hang it off.

desktop/macos/agent/src/index.ts
```diff
@@ -2451,6 +2451,7 @@ async function main(): Promise<void> {
             runId: request.runId,
             attemptId: request.attemptId,
             terminalStatus: request.terminalStatus,
+            finalText: request.finalText,
```
Git-on-my-level, line 2454: **Must-fix before merge (inertness): nothing populates this field on merged main.**

I traced every producer of `external_surface_run_complete` at origin/main (post-#12730). There is exactly one, and it does not carry text:

- `AgentRuntimeProcess.externalSurfaceRunCompleteWireMessage` (`desktop/macos/Desktop/Sources/Chat/AgentRuntimeProcess.swift:1522`) writes only `sessionId`, `runId`, `attemptId`, `terminalStatus`, `errorCode` — no `finalText` key, ever.
- Up the stack, `terminalizeExternalRun` / `completeExternalRunAuthority` (`RealtimeHubController+SessionLifecycle.swift:100` / `:170`) and `AgentRuntimeProcess.completeExternalSurfaceRun` (`AgentRuntimeProcess.swift:1205`) have no text parameter at all, so a caller could not pass one even if it wanted to.
- No other producer exists anywhere in the repo (backend, windows, acp-bridge, agent-cloud: zero hits for `external_surface_run_complete`).

So with this PR merged, `request.finalText` is `undefined` for every production completion, `runs.final_text` stays null, and `message.completed` still carries `{"text":""}` — exactly the field-trace symptom in #12731. The kernel half is well built, but it is plumbing with no water. The body references "#12732, the Swift-side delivery half" as if it were a PR; #12732 is an open **issue** ("AgentCompletionVoiceDelivery never fires for realtime-spawned (floating_chat) runs") with no linked PR.

Because the body says **"Closes #12731"**, merging this as-is would auto-close the issue while the user-visible defect persists. Please do one of:

1. **Preferred:** include the Swift producer in this PR (thread the accumulated realtime transcript through `completeExternalRunAuthority` → `terminalizeExternalRun` → `externalSurfaceRunCompleteWireMessage`), or open the Swift PR and link it here so they merge as a pair; or
2. Reword to **"Part of #12731"** (not "Closes"), keep the issue open, and name #12732 as the follow-up so the closure happens only when a producer actually ships.

Either is fine; what is not fine is closing #12731 on a change that cannot, on merged main, alter a single production row.

desktop/macos/agent/src/runtime/kernel-core.ts
```diff
@@ -650,13 +650,23 @@ export class KernelCore {
...
+    if (input.finalText !== undefined && typeof input.finalText !== "string") {
+      throw new ExternalSurfaceAuthorityError("invalid_external_request", "External surface finalText must be a string");
+    }
+    // The kernel never observes an external surface's stream, so terminalization is
+    // the only point at which this run's answer can be captured. Persisted for every
+    // terminal status, matching the internal adapter path, which likewise stores
+    // whatever text was produced and marks failure through errorCode instead. An
+    // empty string collapses to null so "produced no text" reads identically whether
+    // the surface omitted the field or sent "" (#12731).
+    const finalText = input.finalText ? input.finalText : null;
```
Git-on-my-level: **Should-fix: the orphaned-requeue path leaves a previous answer in `final_text`.**

The write here is correct for the completing attempt, but this PR makes `runs.final_text` writable for external runs for the first time, and one existing path now needs hygiene: `beginExternalSurfaceRun` is idempotent per `turnId`, and its requeue branch (lines 405-411) resets a recycled run row with

```ts
this.updateRun(run.runId, {
  status: "queued",
  completedAtMs: null,
  errorCode: null,
  errorMessage: null,
  updatedAtMs: Date.now(),
});
```

`finalText` (and `resultJson`) are **not** cleared. Today that is unreachable for external runs because `final_text` was always null; once the Swift producer lands, an orphaned external run re-begun for the same turn keeps the previous answer in `final_text` while `queued`, and two consumers read it live in that window:

- the context snapshot `activeRuns` projection (`context-snapshot.ts:300-314` — `queued` is in `ACTIVE_RUN_STATUSES`, and the row mapping carries `finalText` into the snapshot, i.e. into model context assembly), and
- the control list `latestActivity: run?.finalText ?? ...` (`control-tools.ts:2956`).

So a superseded answer could sit in model context as if it were the active run output — a smaller cousin of exactly the #12731 staleness bug. Suggested one-liner in the requeue `updateRun`:

```ts
this.updateRun(run.runId, {
  status: "queued",
  completedAtMs: null,
  finalText: null,
  resultJson: null,
  errorCode: null,
  errorMessage: null,
  updatedAtMs: Date.now(),
});
```

with a regression test: complete an external run with text, re-begin the same `turnId` after orphaning (or force the attempt-terminal requeue), assert `final_text` is null while re-queued.

Nit while here: `input.finalText ? input.finalText : null` treats a whitespace-only string (`" "`) as produced text. Harmless today (consumers sanitize/trim), but `const trimmed = input.finalText?.trim(); const finalText = trimmed ? trimmed : null;` would collapse it consistently with the empty-string decision — your call.

desktop/macos/agent/src/protocol.ts
```diff
@@ -127,6 +127,17 @@ export interface ExternalSurfaceRunCompleteMessage extends ProtocolEnvelope {
...
+   * The response text the external surface streamed to the user.
+   *
+   * An external surface owns its own streaming, so the kernel never observes this
+   * run's output — it has to be handed back at terminalization or it is lost.
+   * Without it `runs.final_text` stays null and every consumer that reads a run's
+   * answer is content-free: the completion lane can only report that an agent
+   * finished, and a spawn receipt that delegates its journal exchange to the run
+   * result writes no assistant row at all (#12731).
+   */
+  finalText?: string;
```
Git-on-my-level, line 144: **Note (version skew): graceful both directions, but the result cannot confirm persistence — worth deciding now.**

Verified for the record: inbound frames are `JSON.parse(line) as InboundMessage` (`index.ts:1897`) with no zod/schema validation and no unknown-field rejection, so

- old kernel + future Swift: extra `finalText` key is silently ignored → text lost, protocol does not break;
- new kernel + old Swift: field absent → `undefined` → collapses to null, identical to today.

`PROTOCOL_VERSION` stays 2 and the runtime-stdio-contract tests do not pin this message field-by-field, so no handshake update is needed. That is all acceptable for an optional additive field.

One design gap to consider in this PR rather than later: `external_surface_run_complete_result` does not echo whether text was actually stored (it carries `ok`/`terminalStatus`/`duplicate` only — `CompleteExternalSurfaceRunResult` in `kernel-types.ts` does not declare `finalText`, even though the implementation returns `{ ...input, duplicate }` so the value is structurally present internally). When the Swift half lands, a new-client-to-old-kernel completion will return `ok: true` with the answer silently dropped — the exact failure mode of #12731, just narrower. Since Swift retries this RPC on timeout, a late replay that lands as `duplicate: true` also silently drops the text (first write wins, by design — fine, but invisible).

Cheapest hedge: declare and forward a `finalTextPersisted: boolean` (or echo a finalText hash/length) in `external_surface_run_complete_result`, so the Swift producer can detect an old kernel via absence and keep its local-journal fallback instead of trusting a silent no-op. Not a blocker for this PR if the daemon always ships version-locked with the app, but the field costs three lines now and an interoperability clause later.

desktop/macos/agent/tests/external-surface-authority.test.ts
```diff
@@ -1522,6 +1522,133 @@ describe("external realtime surface authority", () => {
     expect(store.getRow("SELECT COUNT(*) AS count FROM runs").count).toBe(1);
     store.close();
   });
+
+  // #12731: the kernel never observes an external surface's stream, so unless the
+  // text comes back at terminalization `runs.final_text` stays null and every
+  // consumer of a run's answer — the completion lane, spawn-receipt journaling —
+  // is structurally content-free.
+  it("persists the text an external surface streamed into runs.final_text", () => {
```
Git-on-my-level, line 1530: **Should-fix: the five new cases are good — three adjacent behaviors are untested, two of them cheap at this level.**

What is covered well: success persistence, the `message.completed` payload (the exact `{"text":""}` symptom from the trace), omitted-vs-empty collapsing, failed-status retention, and non-string rejection leaving the run terminalizable. That is the core of the change and the failure-without-fix claim checks out.

Gaps, in order of value:

1. **Idempotent replay with text (cheap, kernel-level).** Swift retries this RPC on timeout (`AgentRuntimeProcess.completeExternalSurfaceRun`, 10s `kernelContractRequest` timeout, and `trackExternalRunTerminalization` reconciles failures). A second `completeExternalSurfaceRun` with the same status returns `duplicate: true` and does **not** rewrite `final_text` — first write wins. Assert that: (a) a replay with *different* text leaves the first text in place, and (b) the replay returns `duplicate: true` rather than throwing `run_terminal` (same-status replay is the one path that succeeds). Nothing today pins that contract for text.

2. **`cancelled` terminal status (cheap).** The failed-with-text case is tested; `cancelled` is a distinct `persistedStatus` and the PR deliberately persists text on it. One case: cancelled-with-text keeps `final_text` + `status: "cancelled"` + null `error_code` (the failed-only defaulting at `kernel-core.ts:671` must not fire for it).

3. **IPC pass-through (index.ts:2454).** The `finalText: request.finalText` glue is untested — every new test calls `kernel.completeExternalSurfaceRun` directly, so deleting the index.ts line would leave all five green. A `protocol-v2`-style round-trip that sends `external_surface_run_complete` with `finalText` over the stdio harness and asserts the stored row would pin the wire-to-row chain the issue is actually about.

4. **Spawn-receipt end-to-end** — you called this out honestly in the body and I agree with the reasoning; flagging it only so it is tracked on the Swift-half issue (#12732) rather than dropped: `agent-spawn-journal.ts:567` builds the completion block `output` from `child.final_text`, and until a fixture exercises an external realtime spawn-receipt child, the second consumer in #12731 is asserted only by construction.

Git-on-my-level (commented):
## Review A — kernel/TS correctness + Swift integration

### Verdict

**The kernel half is correct, well-tested, and gate-clean — but on merged main (post-#12730) this PR is inert against #12731, and its "Closes #12731" would close the issue without fixing it.**

Trace: the only production caller of `completeExternalSurfaceRun` is the `external_surface_run_complete` IPC handler (`index.ts:2448`). The only producer of that message in the entire repo is Swift's `AgentRuntimeProcess.externalSurfaceRunCompleteWireMessage` (`AgentRuntimeProcess.swift:1522` at origin/main), which writes `sessionId`/`runId`/`attemptId`/`terminalStatus`/`errorCode` and no text; `completeExternalRunAuthority`/`terminalizeExternalRun` (`RealtimeHubController+SessionLifecycle.swift:100/170`) accept no text parameter. So after merge, `runs.final_text` stays null and `message.completed` keeps carrying `{"text":""}` — byte-identical to the field trace. The body presents "#12732, the Swift-side delivery half" as the follow-up; #12732 is an open **issue**, not a PR — no Swift change is in flight in this repo.

The kernel-side work itself is sound (details below); the merge condition is either to pair it with a Swift producer or to stop the body from auto-closing #12731.

### Findings

**Must-fix**

1. **Inert producer / closing semantics** (inline on `index.ts:2454`). Either include the Swift producer, or reword "Closes #12731" → "Part of #12731" and link #12732 as the open follow-up. Kernel-only cannot close the issue.

**Should-fix**

2. **Stale `final_text` on orphaned requeue** (inline on `kernel-core.ts:662`). `beginExternalSurfaceRun`'s requeue branch resets `status`/`errorCode`/`errorMessage` but not `finalText`/`resultJson`. Unreachable before this PR (external `final_text` was always null); once a producer exists, a re-begun turn's superseded answer would sit in `final_text` while `queued` and leak into the context snapshot's `activeRuns` (`context-snapshot.ts:300-314`) and the control list's `latestActivity` (`control-tools.ts:2956`). One-line fix + regression test.

3. **Test gaps** (inline on `tests:1530`): (a) same-status replay with text — `duplicate: true` first-write-wins contract unpinned, and Swift retries this RPC; (b) `cancelled`-with-text (only `failed` is covered; the failed-only errorCode defaulting must not fire); (c) IPC pass-through — all five new tests call the kernel directly, so deleting `index.ts:2454` leaves them green. Spawn-receipt e2e is honestly declared missing — track it on #12732.

**Notes (no change demanded)**

4. **Persistence semantics verified correct**: `resultJson` stays null on the external path (no `result_json`/`final_text` duplication); `message.completed` fires only on `succeeded`, matching the internal path; the failed/cancelled retention decision genuinely matches the internal adapter (`kernel-core.ts:2024` passes `result.text` regardless of status); non-string rejection happens before any write and leaves the run terminalizable (asserted).
5. **Protocol skew is graceful both ways** — no schema validation on inbound frames, `PROTOCOL_VERSION` unchanged, optional additive field (inline on `protocol.ts:140` suggests a `finalTextPersisted` echo so a future Swift producer can detect an old kernel instead of silently losing the answer again).
6. **No length bound on `finalText`** — `errorCode` is capped (`^[a-z0-9_]{1,64}$`) but an IPC-originated `finalText` of any size lands in SQLite. Internal parity (`result.text` is also uncapped) makes this defensible; consumers bound it downstream (`RECENT_COMPLETED_RUN_TEXT_MAX_CHARS`, Swift's 1,200-char sanitize). Fine to leave, worth a conscious decision.

### Test results (macOS, Node 22.23.2, vitest 4.1.10)

| Suite | Result |
|---|---|
| `tests/external-surface-authority.test.ts` (focused) | **23/23 pass** (18 pre-existing + 5 new) |
| Full agent suite, PR head (`3700f1c`) | 4 failed / **750 passed** |
| Full agent suite, `origin/main` baseline (same machine) | same 4 failed / 735 passed |

The 4 failures (`runtime-stdio-contract` ×2, `tool-surfaces-exhaustiveness` ×2 — the latter needs `dist/` from `npm run build`) are byte-identical pre-existing failures on main. **Zero new failures introduced.** The Windows numbers in the body are platform-specific and did not reproduce here.

### Repo gates

Ran `scripts/pr-preflight --base <merge-base> --head pr-12734 --pr-body-file <body>` locally: **all 13 selected checks PASS** — `failure-class-protocol` (the declared `FC-decoded-outcome-discarded-before-canonical-record` exists in the registry; its `scope_hints` include `agent-spawn-journal.ts`), `desktop-changelog-entry` (`desktop/macos/agent/` is **not** in `EXEMPT_DESKTOP_PATH_PREFIXES`, so the fragment is required — and present), product-invariants (INV-CHAT-1/INV-AGENT-* citations accepted), lifecycle-headers, and the rest. No frozen-file or Line-Count violations in the 6-file diff.

**GitHub CI: every workflow run on this branch is `action_required`** — the first-time-fork approval gate. Nothing has actually executed; a maintainer needs to approve the runs before there is any CI signal.

### Merge-condition summary

1. Resolve the closing semantics (must-fix #1) — pair with the Swift producer or downgrade the close verb and link #12732.
2. Rebase: the branch is 55 commits behind main; #12730 landed in that window and the body's verification baseline ("upstream/main") predates it, so re-run the suite against current main after rebase.
3. Approve CI runs (action_required).
4. Should-fix #2 and #3 are small enough to land here rather than as follow-ups.

The change itself — validation ordering, empty-string collapse, failed-status retention, honest scope statement — is exactly the shape I'd want for the kernel half. It just cannot be the PR that closes #12731 alone.

Git-on-my-level (commented):
**Reviewer B — incident grounding + merge readiness** (TS kernel correctness is covered elsewhere; I checked this PR against the 2026-09-04 field trace in #12731 and the repo gates).

## What this repairs

Mapped to the incident, assuming a Swift producer eventually sends `finalText`:

- `message.completed` events for external runs stop being literally `{"text":""}` — verified in the field DB: run_659c (run_659c222a…, act, authority swift_realtime, 04:49:11→04:49:37 UTC) emitted exactly that empty payload, and `finishAttemptAndRun` sources the event from the same `finalText` this PR persists.
- The completion-delta lane (`peekCompletedAgentDelta`, recent-run context text in `context-snapshot.ts`) becomes contentful for external runs.
- Spawn-**agent children** get their answer into the journal's `agentCompletion.output` block (`agent-spawn-journal.ts:567` reads `child.final_text`), which is empty today.

The kernel mechanics are right, the guard ordering is right, and 23/23 pass on the current head (I re-ran `vitest run tests/external-surface-authority.test.ts` + `tsc --noEmit` on 42bc8926, which postdates the test runs already posted here).

## What this does not repair

**1. Nothing sends `finalText`, so merged alone this PR changes no field behavior.** On `origin/main`, `externalSurfaceRunCompleteWireMessage` (AgentRuntimeProcess.swift) sends only `terminalStatus` + `errorCode`. The Swift half is #12732, explicitly deferred. Until it lands, `runs.final_text` for `swift_realtime` runs stays NULL — the issue's evidence query returns the same rows the day after this merges as the day before.

**2. The producer-plausibility assumption is contradicted by the field evidence for the exact incident run.** The PR's doc comment calls finalText "the response text the external surface streamed to the user." For run_659c, that streamed reply lived on the realtime voice session (hub `assistantText`), not in the act run — the model spoke the correction 04:49:13–37 while the run executed in parallel. DB grounding: turn_542829a2… ("Think harder and give me the exact layout.") has a user row written 04:49:37 but **no assistant row**, and `recordExchange` (KernelTurnProjection.swift) writes the assistant row only when `assistantText` is non-empty — so Swift's turn-end assistant-text state was **already empty** at the moment the journal capture ran. Both the journal capture and `voiceTurnDidTerminate` (the terminalization caller) read the same hub state in the same turn-done cluster, so a Swift companion that reads hub state at terminalization may well hand this PR an empty string → collapses to null → still not repaired. The text must be captured upstream (before whatever clears it around the `think_deeper` tool boundary — note run_659c was spawned by the `think_deeper` voice tool per `tool_invocation_ledger`, not `spawn_agent`), and that is unpinned.

**3. The incident's primary journal loss is not fixed by this PR under any `finalText` supply.** No kernel path materializes a `conversation_turns` assistant row from a plain external-surface run's `final_text`: `completeExternalSurfaceRun` writes the run row + event only; `ensureAgentSpawnJournal` requires `producerJournal` metadata (spawn_agent children — run_659c has none, verified in `input_json`); `journal_terminalize_turn` is Swift-driven with Swift-supplied content. So even with a perfect producer, the corrected recipe still does not reach the journal for this run class, and the #12731 recurrence engine (model re-erupting the stale wrong recipe 71 minutes later) survives this PR. The issue's spawn-receipt framing is accurate for spawn_agent children but run_659c was not one.

## Merge-readiness checklist

- CI: `gh pr checks` → *no checks reported* on this branch; mergeable = MERGEABLE; reviewDecision = REVIEW_REQUIRED.
- Failure-Class: `FC-decoded-outcome-discarded-before-canonical-record` present; slug exists in `.github/failure-classes/` and its `scope_hints` do name `agent-spawn-journal.ts` as the body claims. ✓
- Changelog: `desktop/macos/agent/**` is **not** exempt in `check-desktop-changelog.py` (only tests/, Desktop/Tests/, Generated/, e2e/ + specific files), so a fragment is required — and present (`2026-09-realtime-answer-journalling.json`, valid `changes` shape). ✓
- Frozen/line-count gates: none apply to these paths; regression snapshot is clean — base commit touches exactly the four protocol/kernel layers + tests + changelog; the follow-up commit 42bc8926 is test-only. No scope creep. (One doc nit: the "matching the internal adapter path" comment overstates — the internal path stores `""` as `""` via `?? null` in sqlite-store.ts, it does not collapse; harmless while no producer sends "".)
- Branch is ~55 commits behind main (merge-base 6136daf8); no conflicts.

## Verdict

**Needs changes — one blocking, the rest advisory.**

Blocking: **detach the `Closes #12731` auto-close** (use "Refs/Part of #12731"). The issue's core field defect — every realtime-voice-deferral run finishing with `final_text` length 0 — will still reproduce verbatim the day this merges, and its journal-loss half (turn_542's missing assistant row) is not repaired by this PR even with a future producer. Auto-closing would record the incident defect as fixed while field behavior is unchanged. Amend the body to state the two residual gaps (Swift producer pending in #12732; no kernel journal materialization for plain external-surface runs).

Advisory: (1) narrow the protocol comment's producer claim — for the incident run class the surface's streamed text is not currently retained anywhere that survives to terminalization, so say "text the surface reports", pending the upstream capture fix; (2) fix the internal-adapter-parity wording; (3) when #12732 lands, the Swift side must source text from turn-stream accumulation, not the turn-end hub property, or this column stays null in exactly the incidents that motivated it — worth a field-verification step there.

Merge the kernel half by all means once the auto-close is detached — it is a correct, well-tested receiving half and #12732's stated prerequisite. It just should not be recorded as the repair of #12731.

SakashSrivastava:
All addressed in bd44888. Taking the blocking one first: you're right, and I verified it before changing anything — externalSurfaceRunCompleteWireMessage writes sessionId/runId/attemptId/terminalStatus/errorCode and no text, and grepping the whole repo for external_surface_run_complete finds that one Swift producer. Merged alone this changes nothing in the field. Description rewritten to "Refs #12731" with a "What this does not repair" section covering all three gaps: no producer yet, no kernel journal materialization for plain external-surface runs, and the turn-stream-vs-hub-property trap for whoever builds #12732.

Reviewer B's second point is the one I'd have got wrong on my own — the spawn-receipt framing in my original description doesn't hold for the incident run, since ensureAgentSpawnJournal wants producerJournal metadata that only spawn_agent children carry and run_659c came from think_deeper. Corrected rather than softened.

Code changes:
- requeue branch clears final_text and result_json, with a regression test that re-begins an orphaned turn and asserts the recycled row is empty
- finalTextPersisted echoed through kernel-types, protocol, index.ts and the result, false on a duplicate replay so first-write-wins is visible rather than silent
- whitespace-only collapses to null alongside ""
- both doc comments narrowed: "text the surface reports", and the internal path stores "" as-is rather than collapsing

Tests: 27 in external-surface-authority (18 before), plus finalText and finalTextPersisted pinned on the v2 wire. Full suite on this branch is 31 failed / 717 passed against 31 failed / 708 passed on upstream/main, same machine, same base — same failures, +9 passing, which is exactly what I added. Rebased onto 176a02fe; preflight passes all 14 checks.

One gap I deliberately left open: the finalText: request.finalText glue in index.ts is pinned at the type level but has no runtime round-trip, because runtime-stdio-contract doesn't run on this machine and I'd rather not ship a test I haven't watched pass. Called out in the description.

Git-on-my-level (approved):
Merge-driver approval on behalf of the review already on record: the kernel/TS half was reviewed by two agents (both posted on this PR — correct, tested, preflight-clean), full CI is green on head bd44888 (Desktop Swift Release Compile pass 20m46s, Agent Runtime pass, Backend Hermetic Merge Gate pass, all others pass or expected-skip), and the body correctly references #12731 without a closing keyword. Known limitation stands and is documented in the PR body: inert on merged main until the Swift producer in #12732 lands.

Issue #12744, opened 2026-09-04
fix(ci): slow_guardrail_manifest.txt references test_ws_g_module_aliases.py which does not exist


formed2forge:
## Symptom

The **Backend unit suite** CI job fails on any PR that touches backend files:

```
FAIL: tests/slow_guardrail_manifest.txt lists tests/unit/test_ws_g_module_aliases.py, which does not exist.
```

## Root cause

`backend/tests/slow_guardrail_manifest.txt` has an entry for `tests/unit/test_ws_g_module_aliases.py`, but that file was never created. The mismatch exists on `main` today.

The failure became visible when commit 84ec103325 wired the slow-guardrails CI job to actually run — it had never run before, so the stale manifest entry went undetected.

## Fix

Either:
1. Create `backend/tests/unit/test_ws_g_module_aliases.py` with the intended `ws_g` module alias tests, or
2. Remove the stale entry from `backend/tests/slow_guardrail_manifest.txt` if the test is no longer planned.

## Evidence

- `backend/tests/slow_guardrail_manifest.txt` on `main`: contains `tests/unit/test_ws_g_module_aliases.py`
- `backend/tests/unit/test_ws_g_module_aliases.py`: does not exist on `main`
- First surfaced in PR #12741 CI run: https://github.com/BasedHardware/omi/actions/runs/33883731104/job/101058339370

aryanorastar:
Closing as already fixed on current main.

Current main slow_guardrail_manifest.txt only mentions test_ws_g_module_aliases.py in a # comment; run-slow-guardrails-ci.sh strips comments (line%%#*) so the missing-file FAIL is gone.

PR #12745, opened 2026-09-04
fix(ci): remove stale ws_g manifest entry — file deleted in a7bc5fd0c2

formed2forge:
## Problem

`backend/tests/slow_guardrail_manifest.txt` listed `tests/unit/test_ws_g_module_aliases.py`, but that file was deleted in `a7bc5fd0c2` ("remove dead modules and orphaned WS-G promotion family"). The manifest was not updated at the time.

This caused **Backend unit suite** to fail on every PR touching backend files since the slow-guardrail lane started actually running (`84ec103325`).

Surfaced by: #12744

## Fix

Move the entry to the `REMOVED` comment block with the deletion commit cited, per the manifest's own policy ("removed with a stated reason, not un-listed"). The WS-G alias shims (`WorkingObservation`, `L1MemoryArchiveItem`, etc.) no longer exist — there is nothing to guard.

## Verification

```
bash backend/scripts/run-slow-guardrails-ci.sh
# Running slow-marked guardrails across 18 file(s).
# (file-existence pre-flight passes; full suite requires CI deps)
```

## Failure-Class

`Failure-Class: none`

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12745?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/slow_guardrail_manifest.txt
+++ b/backend/tests/slow_guardrail_manifest.txt
@@ -25,6 +25,13 @@
 #   tests/unit/test_subscription_restructure.py       32 failed, 2 passed
 #   tests/unit/test_firestore_query_coverage_ratchet.py    1 failed, 1 passed
 #   tests/unit/test_sys_modules_hermeticity.py             1 failed, 1 passed
+#
+# REMOVED — file deleted, aliases it guarded no longer exist:
+#   tests/unit/test_ws_g_module_aliases.py  removed in a7bc5fd0c2
+#     ("chore(backend): remove dead modules and orphaned WS-G promotion family")
+#     The WS-G alias shims (WorkingObservation, L1MemoryArchiveItem, etc.) were
+#     deleted along with the rest of the WS-G promotion family. No replacement
+#     needed; the contract those aliases enforced is gone.
 
 tests/unit/test_async_tasks.py
 tests/unit/test_desktop_transcribe.py
@@ -44,4 +51,3 @@ tests/unit/test_subscription_import_cycle.py
 tests/unit/test_sync_ordered_assignment.py
 tests/unit/test_thread_join_elimination.py
 tests/unit/test_vad_gate.py
-tests/unit/test_ws_g_module_aliases.py


Git-on-my-level:
Thanks @formed2forge — your diagnosis was exactly right, and the fix was done the right way.

What you found: `backend/tests/slow_guardrail_manifest.txt` still listed `tests/unit/test_ws_g_module_aliases.py`, which was deleted in a7bc5fd0c2 ("chore(backend): remove dead modules and orphaned WS-G promotion family"). `backend/scripts/run-slow-guardrails-ci.sh` hard-fails when a manifest entry has no file ("FAIL: ... lists ..., which does not exist"), so the Backend unit suite broke for every backend PR once the slow-guardrail lane actually started running in 84ec103325. Moving the entry to a REMOVED note with the deletion commit cited follows the manifest's own policy ("a guardrail ... is fixed or removed with a stated reason, not un-listed"), and the Backend unit suite is green on your head — the fix itself is correct.

Unfortunately it lost a 15-minute race: #12740 ("fix(backend): drop CI guardrail references to files deleted in WS-G cleanup") merged to main at 14:44 UTC, about 15 minutes before this PR opened, and removed the same stale entry — main's manifest now carries an equivalent provenance note citing a7bc5fd0c2, with the same effective 18-file set. Your branch's merge-base predates that commit, so GitHub shows this PR as cleanly mergeable, but the net effect of merging today would be only a second removal note for an entry that is already gone. Closing as superseded — nothing more is needed from you. #12744 can also be closed as fixed by #12740.

One tip for your next PR, since `Hygiene` and `PR Metadata Preflight` are both failing here: they stop at the same place, `failure-class-protocol: missing_declaration`. Your body does declare `Failure-Class: none`, but it is wrapped in backticks, and the validator in `scripts/failure-class` only matches a line that begins with `Failure-Class:`. Putting `Failure-Class: none` unadorned on its own line keeps those lanes green next time.

Appreciate the sharp catch and the clean write-up — this was a genuinely useful report of a lane that was breaking every backend PR.

---

PR #12749, opened 2026-09-04
fix(backend): gate app generate-prompts on the free chat quota

### LINKED ISSUE #12715: Free-tier chat quota not enforced on /v1/app/generate and /v1/app/generate-icon

aryanorastar:
Follow-up to #12739, which fixed #12715.

## What changed and why

`/v1/app/generate-prompts` runs the same billable `Features.APP_GENERATOR` spend as the other AI-app generators, but sat behind `auth.with_rate_limit(...)` alone with no `enforce_chat_quota` call:

```python
async def generate_sample_prompts_endpoint(
    uid: str = Depends(auth.with_rate_limit(auth.get_current_user_uid, "apps:generate_prompts")),
):
    ...
    with track_usage(uid, Features.APP_GENERATOR):
```

`track_usage` attributes spend; it does not block. #12739 closed `/v1/app/generate` and `/v1/app/generate-icon`, and correctly left this one alone because it was outside #12715's scope. That makes it the last ungated **`APP_GENERATOR`** route in the router — a free user past the monthly cap can keep drawing LLM-generated app suggestions from it indefinitely.

> **Correction.** This originally read "the last ungated managed-spend route in the router", which is wrong, as [Git-on-my-level caught](https://github.com/BasedHardware/omi/pull/12749). `/v1/personas/twitter/initial-message` (`apps.py:1744`) also runs tracked LLM spend, under `Features.PERSONA`, and is gated by neither a quota nor a rate limiter — strictly less protected than this route was. It is out of scope here and tracked separately; the claim above is now scoped to `APP_GENERATOR`, which it can actually support.

This adds the same gate the other four generators use, in the shape #12739 established: `enforce_chat_quota(uid, platform=x_app_platform)` before any provider work, with the `X-App-Platform` header accepted so the desktop trial paywall applies here too rather than silently never firing on `platform=None`.

Rate limiting and quota enforcement solve different problems, which is why one does not substitute for the other: the limiter bounds request frequency, the quota bounds monthly billable spend against the user's plan.

## On the generated-client churn

The six generated artifacts change by a **reordering, not an addition**. `X-App-Platform` was already advertised on this route through the shared auth dependency; declaring it explicitly on the handler moves it ahead of the injected headers. That is exactly the ordering the sibling routes already have:

```
POST /v1/app/generate              ['X-App-Platform', 'authorization', 'X-Device-Id-Hash', 'X-App-Version']
POST /v1/app/generate-icon         ['X-App-Platform', 'authorization', 'X-Device-Id-Hash', 'X-App-Version']
POST /v1/app/generate-description  ['X-App-Platform', 'authorization', 'X-Device-Id-Hash', 'X-App-Version']
GET  /v1/app/generate-prompts      ['X-App-Platform', 'authorization', 'X-Device-Id-Hash', 'X-App-Version']  ← this PR
```

So the route converges on the router's existing convention rather than introducing a new shape. No wire-level parameter is added or removed.

## How it was verified

Mutation-checked — with the gate removed, both new tests fail:

```
2 failed, 14 deselected
```

- `pytest tests/unit/test_free_quota_gate_wiring.py` — 16 passed
- `scripts/openapi_runner.sh scripts/export_openapi.py --surface app-client --check` — up to date
- Swift, TypeScript (×4) and Dart generators re-run; only the four TS/Swift artifacts changed, Dart models unaffected

## Tests

- `test_generate_prompts_blocked_past_cap` — asserts the 402 propagates **and** that `get_llm` is never called, so the gate precedes spend rather than merely accompanying it
- `test_generate_prompts_proceeds_within_cap` — asserts the proceed path still reaches the provider, and pins `enforce_chat_quota('u1', platform='macos')` so the platform argument cannot be silently dropped

## Honest gaps

- No live free-tier Firebase/Redis integration run (hermetic gate patching only).
- Physical / on-device smoke not run — backend-only change.

Failure-Class: none

Line-Count-Exception: backend/routers/apps.py | 2468 -> 2472 | Four lines: the X-App-Platform header parameter and the enforce_chat_quota call, matching the four sibling generators in this router. Splitting the file is a larger refactor than a one-gate wiring fix should carry.

<a href="https://cubic.dev/pr/BasedHardware/omi/pull/12749?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/routers/apps.py
+++ b/backend/routers/apps.py
@@ -1532,6 +1532,7 @@ def generate_description_and_emoji_endpoint(
 @router.get('/v1/app/generate-prompts', tags=['v1'], response_model=AppPromptsGenerationResponse)
 async def generate_sample_prompts_endpoint(
     uid: str = Depends(auth.with_rate_limit(auth.get_current_user_uid, "apps:generate_prompts")),
+    x_app_platform: Optional[str] = Header(None, alias='X-App-Platform'),
 ):
     """
     Generate sample app prompts for the AI app generator.
@@ -1539,6 +1540,9 @@ async def generate_sample_prompts_endpoint(
     """
     from utils.llm.clients import get_llm
     import json
+
+    # User-initiated LLM generation — same free-tier gate as chat (402 past cap).
+    enforce_chat_quota(uid, platform=x_app_platform)
 
     system_prompt = """Generate 5 creative and diverse ideas for apps that are either:
 1. Conversation summary based apps - analyze user's recorded conversations and extract/organize information
--- a/backend/tests/unit/test_free_quota_gate_wiring.py
+++ b/backend/tests/unit/test_free_quota_gate_wiring.py
@@ -14,7 +14,7 @@ router module (the sanctioned monkeypatch-on-module seam).
 import asyncio
 import os
 from pathlib import Path
-from types import ModuleType
+from types import ModuleType, SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
@@ -236,6 +236,31 @@ class TestAppGeneratorGates:
                 asyncio.run(apps_router.generate_app_icon_endpoint(data, uid='u1', x_app_platform='macos'))
             assert exc_info.value.status_code == 402
             gate.assert_called_once_with('u1', platform='macos')
+
+    def test_generate_prompts_blocked_past_cap(self):
+        # The last ungated managed-spend route in this router. #12739 closed
+        # /v1/app/generate and /v1/app/generate-icon; generate_sample_prompts_endpoint
+        # still ran the same Features.APP_GENERATOR spend behind rate limiting alone,
+        # so a free user past cap could keep drawing billable suggestions from it.
+        import routers.apps as apps_router
+
+        with patch.object(apps_router, 'enforce_chat_quota', side_effect=_quota_402()) as gate, patch(
+            'utils.llm.clients.get_llm'
+        ) as llm:
+            with pytest.raises(HTTPException) as exc_info:
+                asyncio.run(apps_router.generate_sample_prompts_endpoint(uid='u1', x_app_platform='macos'))
+            assert exc_info.value.status_code == 402
+            gate.assert_called_once_with('u1', platform='macos')
+            llm.assert_not_called()
+
+    def test_generate_prompts_proceeds_within_cap(self):
+        import routers.apps as apps_router
+
+        with patch.object(apps_router, 'enforce_chat_quota') as gate, patch('utils.llm.clients.get_llm') as llm:
+            llm.return_value.ainvoke = AsyncMock(return_value=SimpleNamespace(content='["a","b","c","d","e"]'))
+            asyncio.run(apps_router.generate_sample_prompts_endpoint(uid='u1', x_app_platform='macos'))
+            gate.assert_called_once_with('u1', platform='macos')
+            llm.assert_called_once()
 
 
 # ---------------------------------------------------------------------------
--- a/desktop/macos/Desktop/Sources/Generated/OmiApi.generated.swift
+++ b/desktop/macos/Desktop/Sources/Generated/OmiApi.generated.swift
@@ -6322,7 +6322,7 @@ public enum OmiAPI {
     return try JSONDecoder().decode(OmiAnyCodable.self, from: data)
   }
 
-  public static func generateSamplePromptsEndpointV1AppGeneratePromptsGet(client: OmiApiClient, authorization: String? = nil, xAppPlatform: String? = nil, xDeviceIdHash: String? = nil, xAppVersion: String? = nil) async throws -> OmiAnyCodable {
+  public static func generateSamplePromptsEndpointV1AppGeneratePromptsGet(client: OmiApiClient, xAppPlatform: String? = nil, authorization: String? = nil, xDeviceIdHash: String? = nil, xAppVersion: String? = nil) async throws -> OmiAnyCodable {
     let _path = "/v1/app/generate-prompts"
     guard let components = URLComponents(string: client.baseURL + _path) else {
       throw OmiApiError.invalidURL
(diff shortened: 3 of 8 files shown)


Git-on-my-level:
Verified against head 3341508b — this closes the remaining spend leak from the #12739 sweep, in exactly the shape that PR established. Notes per file:

**backend/routers/apps.py** — `enforce_chat_quota(uid, platform=x_app_platform)` at line 1545 lands before `get_llm` and the `track_usage(uid, Features.APP_GENERATOR)` block (line 1564), so past-cap users are cut off before any provider work. The `X-App-Platform` param mirrors the four sibling generators (lines 1493/1516/1600/1647), the import already exists from #12739, and `with_rate_limit` is retained — limiter and quota stay orthogonal, as the description argues. All five APP_GENERATOR routes now gate identically.

**backend/tests/unit/test_free_quota_gate_wiring.py** — `test_generate_prompts_blocked_past_cap` asserts both the 402 and `llm.assert_not_called()`, proving the gate precedes spend rather than merely accompanying it; `test_generate_prompts_proceeds_within_cap` pins `platform='macos'` so the header argument can't be silently dropped and confirms the provider is still reached under cap. Consistent with the suite's existing module-seam approach.

**Generated artifacts** (`desktop/macos/.../OmiApi.generated.swift`, `desktop/windows/.../omiApi.generated.ts`, `web/admin/...`, `web/app/src/lib/omiApi.generated.ts`, `web/personas-open-source/src/lib/omiApi.generated.ts`) — reorder-only: `X-App-Platform` is now serialized ahead of `authorization`, matching the sibling routes; no parameter added or removed at the wire level. **docs/api-reference/app-client-openapi.json** — the same two-parameter swap on GET /v1/app/generate-prompts, and the public API contract check is green.

Two notes on the earlier automated review's suggestions, for the record — neither is a defect introduced by this diff:

- APP_GENERATOR spend is gated here but doesn't accrue to the chat quota counter (`get_monthly_chat_usage`), so a user who only ever calls the generators never crosses the cap. That's a property of the #12739 pattern shared by all five routes; making generator spend count toward the cap is a deliberate billing-policy change that would deserve its own issue.
- The Firestore reads inside `enforce_chat_quota` are synchronous and run on the event loop in this `async def` — identical to `generate_app_endpoint` and `generate_app_icon_endpoint` from #12739. A `run_blocking` pass would be a separate router-wide refactor, not something to bolt onto this fix.

One small accuracy nit on the description: "the last ungated managed-spend route in this router" — `/v1/personas/twitter/initial-message` (apps.py:1745) still runs `Features.PERSONA`-tracked LLM spend with no quota gate. Reasonable to leave out of this PR's scope, but it may deserve the same treatment in a follow-up.

Thanks for the thorough write-up, the mutation check, and the regression tests — tidy completion of the sweep. This is billing/quota enforcement surface, so maintainer sign-off before merge; no concerns from my side beyond the notes above.

---

undivisible (approved):
## Verdict: approve

Closes the last ungated `Features.APP_GENERATOR` route in this router. Gate shape matches the #12739 siblings: `X-App-Platform` header + `enforce_chat_quota(uid, platform=…)` before any provider work, with tests that pin 402 before `get_llm` and the proceed path. Generated-client churn is header reordering only — matches sibling OpenAPI order. CI green across backend unit, hermetic, Swift, Windows/Linux smokes, OpenAPI.

### On cubic
- **P1 (quota counter / prompts-only bypass):** not a regression introduced here — same `enforce_chat_quota` + `track_usage(APP_GENERATOR)` pairing the other generators already use. If app-generator spend should increment `quota_questions`, that is a shared follow-up across the router, not a blocker for wiring parity.
- **P2 (sync Firestore on event loop):** same call pattern as the sibling generators; out of scope for this one-route fix.

Not merging from here.

aryanorastar:
The accuracy nit is right, and I have corrected the description in place rather than leaving the wrong claim standing above a correction.

"The last ungated managed-spend route in this router" was wrong. It is the last ungated **`APP_GENERATOR`** route. `/v1/personas/twitter/initial-message` (`apps.py:1744`) also runs tracked LLM spend under `Features.PERSONA` and has no quota gate.

I enumerated every `track_usage` site in the router rather than fix the one you named, since I had already made this claim twice — here and when closing #12717 — and a third partial version would be worse than the first:

| route | feature | quota gate | rate limit |
| --- | --- | --- | --- |
| `POST /v1/app/generate-description` | `APP_GENERATOR` | yes | no |
| `POST /v1/app/generate-description-emoji` | `APP_GENERATOR` | yes | no |
| `POST /v1/app/generate` | `APP_GENERATOR` | yes | no |
| `POST /v1/app/generate-icon` | `APP_GENERATOR` | yes | no |
| `GET /v1/app/generate-prompts` | `APP_GENERATOR` | **this PR** | yes |
| `GET /v1/personas/twitter/initial-message` | `PERSONA` | **no** | **no** |

Worth noting it is *less* protected than this route was, not equally: `generate-prompts` at least sat behind `auth.with_rate_limit(...)`, which bounded call frequency while the quota was missing. The personas route has neither bound.

Filed as **#12781**, split into the wiring step (give it the rate limiter its siblings have — no policy decision needed) and the policy question you and cubic both raised, which is whether generator and persona spend should increment `quota_questions` at all. As you note, today none of these routes accrue to the counter they are gated on, so a generators-only user never crosses the cap. That one needs an owner rather than a drive-by fix, so I have not touched it.

No code change here — the diff is unchanged and still the four lines. Only the description was inaccurate.

